PR #58, opened 2015-08-06
Resolve #55: Retry nnet3/nnet-optimize-test on failure with optimizations turned off one by one.

### LINKED ISSUE #55: nnet3- small task
This task is for anyone who can do it and has free time.
If you check out the nnet3 branch of the code and build it, you'll notice that there is a test failure in nnet3-optimize-test:

LOG (UnitTestNnetOptimize():nnet-optimize-test.cc:111) Output sum (not optimized) is -33558.6
LOG (UnitTestNnetOptimize():nnet-optimize-test.cc:113) Output sum (optimized) is -34012.6
ERROR (UnitTestNnetOptimize():nnet-optimize-test.cc:115) Non-optimized and optimized versions of the computation give different outputs.
terminate called after throwing an instance of 'std::runtime_error'
  what():  ERROR (UnitTestNnetOptimize():nnet-optimize-test.cc:115) Non-optimized and optimized versions of the computation give different outputs.

[stack trace: ]
kaldi::KaldiGetStackTrace()
kaldi::KaldiErrorMessage::~KaldiErrorMessage()
kaldi::nnet3::UnitTestNnetOptimize()
./nnet-optimize-test(main+0xe3) [0x54903b]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xfd) [0x7f3de3f45ead]
./nnet-optimize-test() [0x547b69]

What needs to be done is a refactoring and change to  UnitTestNnetOptimize(), so that when it detects a mismatch like this, it will turn on the optimizations in the NnetOptimizeOptions config one by one and show the output-sum for each one separately, and hopefully, as a convenience, try to inform the user which of those optimizations seems to be the issue.
Vijay was going to do this but he is very busy right now and I thought it's the kind of thing we could find someone from the list to do.

Dan

kkm000:
Had to add a missing semicolon to `nnet-example-utils.cc`.

### DIFF
--- a/src/nnet3/nnet-example-utils.cc
+++ b/src/nnet3/nnet-example-utils.cc
@@ -70,7 +70,7 @@ static void GetIoSizes(const std::vector<NnetExample> &src,
                   << dims[i] << " vs. " << this_dim << " for '"
                   << io.name << "'.";
       }
-      KALDI_ASSERT(io.features.NumRows() == io.indexes.size());      
+      KALDI_ASSERT(io.features.NumRows() == io.indexes.size());
       int32 this_size = io.indexes.size();
       (*sizes)[i] += this_size;
     }
@@ -99,7 +99,7 @@ static void MergeIo(const std::vector<NnetExample> &src,
     io.name = names[f];
     io.indexes.resize(size);
   }
-  
+
   std::vector<std::string>::const_iterator names_begin = names.begin(),
                                              names_end = names.end();
   std::vector<NnetExample>::const_iterator iter = src.begin(), end = src.end();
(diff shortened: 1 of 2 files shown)


kkm000:
Wait, I'll rebase the change to fix the conflict.

danpovey:
Wow, that was fast!
Thanks!

kkm000:
Done. @danpovey, PTAL. Do you think it would be less error prone to wrap each test re-run in a preprocessor macro? I'd rather change to a macro since it would be hard to notice an error like

```C++
  optimize = optimize_all;
  optimize.initialize_undefined = false;
  bool succ_no_initialize_undefined = UnitTestNnetOptimizeWithOptions(optimize);

  optimize = optimize_all;
  optimize.initialize_undefined = false;
  bool succ_no_move_sizing_commands = UnitTestNnetOptimizeWithOptions(optimize);
```

danpovey:
I'd prefer it as it is... more readable without macro
dan

kkm000:
> Wow, that was fast!

Says the guy who wrote [838,648 lines of code](https://github.com/kaldi-asr/kaldi/graphs/contributors) in a couple years... :astonished:

PR #61, opened 2015-08-06
Unify .gitignore patterns

kkm000:
`.gitignore` turned into a green scaly monster!

Ignore extension-less files in `/src/*/` directories only (not recursive), `.depend.mk` and directories `/egs/*/s*/{mfcc,plp.data.exp}`. This covers a kerzillion of former individual entries.

Git will not ignore files matching `.gitignore` that are already tracked; the entries for README and Makefile are to ease up future additions, so that newly added files thus named are not forgotten unintentionally.


jtrmal:
Let's let Dan decide.  I like the fact the .gitignore is smaller.
I'd be a little bit worried that too "smart" position-dependent rules might
cause hard to debug problems in cases when users would be adding new files.
y.

On Thu, Aug 6, 2015 at 12:30 PM, Kirill Katsnelson <notifications@github.com

kkm000:
Usually you do not debug Git ignores, you just work around them with `git add -f` :)  But certainly `git check-ignores` exists to the rescue of the adventurous.

Currently, the rules would exclude one tracked file: the link `src/nnet2bin/raw-nnet-init -> nnet-init`. Methinks if you need an oddity like this, just `git add -f` it. Otherwise, the structure and contents of /src/*/ is very regular, so I would not expect any problems there.

jtrmal:
Very well -- I know about git add -f, didn't know about git check-ignores

danpovey:
OK with me- Kirill has more git experience so I trust that he knows the
best way to do things.
Dan

jtrmal:
Ok, merged

PR #65, opened 2015-08-07
Adding some unit test for CPU version of SparseMatrix

chenguoguo:
Plus various style fixes. I haven't applied "explicit" keyword to all the constructors with single argument, as that causes some problem with vector initialization, will do later.


danpovey:
Thanks.
You're right, the self-constructors generally have to have the "explicit" keyword left off otherwise it causes problems with stl vector.
Dan

PR #66, opened 2015-08-10
blstm remove bug

nichongjia:
remove a bug in blstm


vesis84:
Hi Ni, 
many thanks for the pull-request. You did the first one which is related to the 'nnet1' code! :)
I went through your code, and I have following suggestions:

1) It would be better **NOT TO** extend the 'Component' interface by 'Component::SetSeqLengths(...)', 
  * it is used only by '<BLstmProjectedStreams>', while does not make sense for components. This means it is not a good function for an interface. 
  * a more sensible solution is to add a method to 'Nnet::SetSeqLengths(...)', and look for data-type 'BLstmProjectedStreams', like is done for 'Nnet::ResetLstmStreams(...)', 
2) Int-type int32,
  * the default kaldi int-type is 'int32', normal 'int' can be both 32 and 64 bit, int32 is cross-platform,  
  * Nnet::SetSeqLengths(std::vector<int> &sequence_lengths) -> Nnet::SetSeqLengths(std::vector<int32> &sequence_lengths), 
  * std::vector<int> sequence_lengths_ -> std::vector<int32> sequence_lengths_;
2) Inconsistent name of training tool : nnet-train-blstm-parallel -> nnet-train-blstm-streams
  * make it consistent with LSTM tool        

Can you please create a new pull-request which incorporates the changes?
Please let me know if you need more info.
Thank you!
Karel

kkm000:
I think that GitHub does not notify the owners that a PR has been updated, only when it is created. @nichongjia, it can be a good idea to post a comment, simply "Done" if nothing else, to keep the review going.

nichongjia:
Hi Karel,
According to your suggestion, I have modified the code, and re-create a new pull-request.

dophist:
Hi Chongjia,

I just had a very quick review of your blstm code, got one question:

I notice you use zero-padded frames in blstm multi-stream training, please check if these zeros have undesired effect on the reverse-activation-pass.

Karel (@vesis84) used "0 frame-weight" to block the back-prop signal because derivative-pass is linear and a 0 in a multiplicative chain will set the result to 0.  However in blstm, reverse-activation-pass is not linear, zeroed frames will be reversely propagated through the whole utterance due to bias and sigmoid and tanh, hence affecting all network activation states.

just a quick review,  please correct me if I missed the details : )

best,
Jiayu (Jerry)

src/nnet/nnet-component.h
```diff
@@ -91,7 +91,7 @@ class Component {
   /// Convert component type to marker
   static const char* TypeToMarker(ComponentType t);
   /// Convert marker to component type (case insensitive)
-  static ComponentType MarkerToType(const std::string &s);
+  static ComponentType MarkerToType(const std::string &s);
```
kkm000: The only change to this file is the adding of EOL whitespace to this line.

src/nnet/nnet-blstm-projected-streams.h
```diff
@@ -19,8 +19,8 @@
 
 
 
-#ifndef KALDI_NNET_NNET_BLSTM_PROJECTED_STREAMS_H_
-#define KALDI_NNET_NNET_BLSTM_PROJECTED_STREAMS_H_
+#ifndef KALDI_NNET_BLSTM_PROJECTED_STREAMS_H_
```
kkm000: NNET_NNET is not a mistake: the file is referred to as "nnet/nnet-blstm-projected-streams.h". Please revert these guard names.

danpovey: BTW, please try to get the whitespace correct.  Run your code through
cpplint.py to check.
Here:
nrecur_(int32(output_dim/2)),
there are two issues: for casting you should static_cast<int32> instead of
a C-style cast.
And, like many other places, there is not the recommended whitespace around
operators; should be
output_dim / 2.
I'm not really checking the code for content; I assume Karel and others
will do that.
Dan


On Thu, Aug 13, 2015 at 6:54 PM, Kirill Katsnelson <notifications@github.com

vesis84:
Looks good! Merging-in...

vesis84:
Oooops, there was a compilation issue intruduced with this pull-request...
It's fixed now.
K.

(Next time I'll try the dry-run comilation with downloadable patch : https://github.com/kaldi-asr/kaldi/pull/66.patch)

kkm000:
@vesis84, GitHub has a machinery that makes applying the patch totally unnecessary. Pull requests are actually refs (you can think of them as special invisible branches) in the repository, and there are up to two per pull request. First is the pointer to the original branch that the requester wants to merge, and the second is _a branch already merged_. Yes, by the time GitHub tells you "This PR can be automatically merged". it _has been_ automatically merged! The post-merge reference is `pull\$ID\merge`, where `$ID` is the pull request number. You can see these special refs with the `git ls-remote` command, and [this article](https://help.github.com/articles/checking-out-pull-requests-locally/) has a bit of additional info.

So all you need to do is to fetch this special pre-merged reference, and check it out:
```
kkm@yupana:~/work/temp[master]$ git fetch origin pull/1/merge
remote: Counting objects: 1, done.
remote: Total 1 (delta 0), reused 0 (delta 0), pack-reused 0
Unpacking objects: 100% (1/1), done.
From git://github.com/kkm000/kaldi-old
 * branch            refs/pull/1/merge -> FETCH_HEAD
kkm@yupana:~/work/temp[master]$ git checkout FETCH_HEAD
Note: checking out 'FETCH_HEAD'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at f31ddcc... Merge a172e8794cf9747bdd95c6efd0c80661bd68dba4 into 20a88df464b2b8b5fb6038101487a5e520aadb4b
kkm@yupana:~/work/temp[(f31ddcc...)]$ make .... # Ready for verification
```

Your worktree now is in a so called _detached head_ state, which means there is no branch pointer for it. But this is exactly what you want: you do not intend to keep this merge in your local clone, only test if there are any problems with it, and then discard it. As soon as you go back to one of your own branches, the whole chain of commits you just pulled becomes liable to garbage collection. Indeed, if you want to attach the head to a branch pointer, `git checkout -b` suggested by git above is the way to do it.

You can also see the current state is indeed the proposed merge of the pull request:

```
kkm@yupana:~/work/temp[(f31ddcc...)]$ git log -5 --graph --oneline --decorate
*   f31ddcc (HEAD) Merge a172e8794cf9747bdd95c6efd0c80661bd68dba4 into 20a88df464b2b8b5fb6038101487a5e520aadb4b
|\
| * a172e87 (origin/winbuild) Default to KALDI_CYGWIN_COMPAT
| * e363c6c Fixes to the above 2 cherry-picks (manually applied to source as well)
| * f7a008c Changes from https://github.com/kaldi-asr/kaldi/commit/5d2cc3#commitcomment-11321531
| * eadd779 Changed per the discussion at https://github.com/kaldi-asr/kaldi/pull/3
```

HTH.

vdp:
It's interesting that GitHub's documentation only talks about pull/$ID/HEAD (which I assume is the (pre-merge) head of the pull requester's branch), but not pull/$ID/merge, although it's mentioned in some [third party docs](http://fiji.sc/GitHub_Tricks#Testing_Pull_Requests_before_merging_them). Perhaps it would be useful to add the above info to Kirill's tutorial for future reference...

danpovey:
This is maintainer-level documentation, not developer-level, so let's leave
it in this email thread for now.
search terms for later: "applying pull request", "kirill's instructions"
Dan


On Fri, Aug 14, 2015 at 10:33 PM, Vassil Panayotov <notifications@github.com

PR #76, opened 2015-08-17
Add wav-reverberate.cc and modify the scripts using it.

tomkocse:
Please comment.

### DIFF
--- a/egs/aspire/s5/local/multi_condition/reverberate_data_dir.sh
+++ b/egs/aspire/s5/local/multi_condition/reverberate_data_dir.sh
@@ -1,6 +1,7 @@
 #!/bin/bash 
 
 # Copyright 2014  Johns Hopkins University (Author: Vijayaditya Peddinti)
+#           2015  Tom Ko
 # Apache 2.0.
 # This script processes generates multi-condition training data from clean data dir
 # and directory with impulse responses and noises
@@ -9,11 +10,8 @@
 set -e
 
 random_seed=0
-num_files_per_job=100
 snrs="20:10:15:5:0"
 log_dir=exp/make_reverb
-max_jobs_run=50
-dest_wav_dir=
 
 . ./path.sh;
 . ./utils/parse_options.sh
@@ -29,13 +27,8 @@ src_dir=$1
 impnoise_dir=$2
 dest_dir=$3
 
-if [ -z $dest_wav_dir ]; then
-  dest_wav_dir=$dest_dir/wavs
-fi
-
 mkdir -p $dest_dir
 mkdir -p $log_dir
-mkdir -p $dest_wav_dir
 
 wav_prefix="rev${random_seed}_"
 utt_prefix="rev${random_seed}_"
@@ -48,17 +41,9 @@ cat $src_dir/utt2spk | awk -v p=$utt_prefix '{printf("%s%s %s\n", p, $1, $1);}' 
 # create the wav.scp files
 cat $src_dir/wav.scp | sed -e "s/^\s*//g" | \
   cut -d' ' -f1 | \
-  awk -v p1=$dest_wav_dir -v p2=$wav_prefix \
-  '{printf("%s%s%s.wav\n", p1, p2, $1);}'> $log_dir/corrupted_${random_seed}.list
-
-python -c "
-import re
-file_ids = map(lambda x: x.split()[0], open('$src_dir/wav.scp').readlines())
-dest_file_names = map(lambda x: x.split()[0], open('$log_dir/corrupted_${random_seed}.list'))
-for file_id, dest_file_name in zip(file_ids, dest_file_names):
-  print '$wav_prefix{0} cat {1} |'.format(file_id, dest_file_name)
-" > $dest_dir/wav.scp
-
+  awk -v p2=$wav_prefix \
+  '{printf("%s%s\n", p2, $1);}'> $log_dir/corrupted_${random_seed}.list
+  
 # modify segments file to point to the new wav files
 cat $dest_dir/segments | awk -v p=$wav_prefix \
   '{printf("%s %s%s %s %s\n", $1, p, $2, $3, $4);}' > $log_dir/segments_temp
@@ -71,13 +56,9 @@ for file in cmvn.scp feats.scp reco2file_and_channel; do
   rm -f $dest_dir/$file
 done
 
-python local/multi_condition/get_reverberate_parameter_lists.py \
-  --snrs $snrs --num-files-per-job $num_files_per_job --random-seed $random_seed \
+python local/multi_condition/reverberate_wavs.py \
+  --snrs $snrs --random-seed $random_seed \
 $src_dir/wav.scp $log_dir/corrupted_${random_seed}.list $impnoise_dir \
-$log_dir/corrupt_wavs.${random_seed}.list > $log_dir/num_corruption_jobs || exit 1;
-
-num_jobs=$(cat $log_dir/num_corruption_jobs)
-$decode_cmd -V --max-jobs-run $max_jobs_run JOB=1:$num_jobs $log_dir/corrupt_wavs.${random_seed}.JOB.log  \
- python local/multi_condition/corrupt.py --temp-file-name $log_dir/temp_JOB.wav $log_dir/corrupt_wavs.${random_seed}.JOB.list || exit 1;
+$dest_dir/wav.scp || exit 1;
 
 echo "Successfully generated corrupted data and stored it in $dest_dir." && exit 0;
--- a/src/feat/signal-test.cc
+++ b/src/feat/signal-test.cc
@@ -1,0 +1,63 @@
+// feat/signal-test.cc
+
+// Copyright 2015  Tom Ko
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
+void UnitTestBlockConvolution() {
+  for (int32 i = 0; i < 5; i++) {
+    int32 signal_length = 4000000 + Rand() % 400000;
+    int32 filter_length = 10000 + Rand() % 1000;
+    Vector<BaseFloat> signal(signal_length);
+    Vector<BaseFloat> filter(filter_length);
+    signal.SetRandn();
+    filter.SetRandn();
+    Vector<BaseFloat> signal_test(signal);
+    FFTbasedConvolveSignals(filter, &signal_test);
+    FFTbasedBlockConvolveSignals(filter, &signal);
+    AssertEqual(signal, signal_test, 0.000001 * signal.Dim());
+  }
+}
+
+void UnitTestConvolution() {
+  for (int32 i = 0; i < 5; i++) {
+    int32 signal_length = 40000 + Rand() % 4000;
+    int32 filter_length = 100 + Rand() % 100;
+    Vector<BaseFloat> signal(signal_length);
+    Vector<BaseFloat> filter(filter_length);
+    signal.SetRandn();
+    filter.SetRandn();
+    Vector<BaseFloat> signal_test(signal);
+    ConvolveSignals(filter, &signal_test);
+    FFTbasedBlockConvolveSignals(filter, &signal);
+    AssertEqual(signal, signal_test, 0.0001 * signal.Dim());
+  }
+}
+}
+
+int main() {
+  using namespace kaldi;
+  UnitTestBlockConvolution();
+  UnitTestConvolution();
+  KALDI_LOG << "Tests succeeded.";
+
+}
--- a/src/feat/signal.cc
+++ b/src/feat/signal.cc
@@ -1,0 +1,123 @@
+// feat/signal.cc
+
+// Copyright 2015  Tom Ko
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
+void ElementwiseProductOfFft(const Vector<BaseFloat> &a, Vector<BaseFloat> *b) {
+  int32 num_fft_bins = a.Dim() / 2;
+  for (int32 i = 0; i < num_fft_bins; i++) {
+    // do complex multiplication
+    ComplexMul(a(2*i), a(2*i + 1), &((*b)(2*i)), &((*b)(2*i + 1)));
+  }
+}
+
+void ConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal) {
+  int32 signal_length = signal->Dim();
+  int32 filter_length = filter.Dim();
+  Vector<float> signal_padded(signal_length + filter_length - 1);
+  signal_padded.SetZero();
+  for (int32 i = 0; i < signal_length; i++) {
+    for (int32 j = 0; j < filter_length; j++) {
+        signal_padded(i + j) += (*signal)(i) * filter(j);
+    }
+  }
+  signal->CopyFromVec(signal_padded.Range(0, signal_length));
+}
+
+
+void FFTbasedConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal) {
+  int32 signal_length = signal->Dim();
+  int32 filter_length = filter.Dim();
+
+  int32 fft_length = RoundUpToNearestPowerOfTwo(signal_length + filter_length - 1);
+  KALDI_VLOG(1) << "fft_length for full signal convolution is " << fft_length;
+
+  SplitRadixRealFft<BaseFloat> srfft(fft_length);
+
+  Vector<float> filter_padded(fft_length);
+  filter_padded.Range(0, filter_length).CopyFromVec(filter);
+  srfft.Compute(filter_padded.Data(), true);
+
+  Vector<float> signal_padded(fft_length);
+  signal_padded.Range(0, signal_length).CopyFromVec(*signal);
+  srfft.Compute(signal_padded.Data(), true);
+
+  ElementwiseProductOfFft(filter_padded, &signal_padded);
+
+  srfft.Compute(signal_padded.Data(), false);
+  signal_padded.Scale(1.0 / fft_length);
+
+  signal->CopyFromVec(signal_padded.Range(0, signal_length));
+}
+
+void FFTbasedBlockConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal) {
+  int32 signal_length = signal->Dim();
+  int32 filter_length = filter.Dim();
+
+  KALDI_VLOG(1) << "Length of the filter is " << filter_length;
+
+  int32 fft_length = RoundUpToNearestPowerOfTwo(4 * filter_length);
+  KALDI_VLOG(1) << "Best FFT length is " << fft_length;
+
+  int32 block_length = fft_length - filter_length + 1;
+  KALDI_VLOG(1) << "Block size is " << block_length;
+  SplitRadixRealFft<BaseFloat> srfft(fft_length);
+
+  Vector<float> filter_padded(fft_length);
+  filter_padded.Range(0, filter_length).CopyFromVec(filter);
+  srfft.Compute(filter_padded.Data(), true);
+
+  Vector<float> temp_pad(filter_length - 1);
+  temp_pad.SetZero();
+  Vector<float> signal_block_padded(fft_length);
+
+  for (int32 po = 0; po < signal_length; po += block_length) {
+    // get a block of the signal
+    int32 process_length = std::min(block_length, signal_length - po);
+    signal_block_padded.SetZero();
+    signal_block_padded.Range(0, process_length).CopyFromVec(signal->Range(po, process_length));
+
+    srfft.Compute(signal_block_padded.Data(), true);
+
+    ElementwiseProductOfFft(filter_padded, &signal_block_padded);
+
+    srfft.Compute(signal_block_padded.Data(), false);
+    signal_block_padded.Scale(1.0 / fft_length);
+
+    // combine the block
+    if (po + block_length < signal_length) {       // current block is not the last block
+      signal->Range(po, block_length).CopyFromVec(signal_block_padded.Range(0, block_length));
+      signal->Range(po, filter_length - 1).AddVec(1.0, temp_pad);
+      temp_pad.CopyFromVec(signal_block_padded.Range(block_length, filter_length - 1));
+    } else {
+      signal->Range(po, signal_length - po).CopyFromVec(
+                        signal_block_padded.Range(0, signal_length - po));
+      if (filter_length - 1 < signal_length - po)
+        signal->Range(po, filter_length - 1).AddVec(1.0, temp_pad);
+      else
+        signal->Range(po, signal_length - po).AddVec(1.0, temp_pad.Range(0, signal_length - po));
+    }
+  }
+}
+}
+
--- a/src/feat/signal.h
+++ b/src/feat/signal.h
@@ -1,0 +1,47 @@
+// feat/signal.h
+
+// Copyright 2015  Tom Ko
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+
+namespace kaldi {
+
+/*
+   This function implements a simple non-FFT-based convolution of two signals.
+   It is suggested to use the FFT-based convolution function which is more
+   efficient.
+*/
+void ConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal);
+
+/*
+   This function implements FFT-based convolution of two signals.
+   However this should be an inefficient version of BlockConvolveSignals()
+   as it processes the entire signal with a single FFT.
+*/
+void FFTbasedConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal);
+
+/*
+   This function implements FFT-based block convolution of two signals using 
+   overlap-add method. This is an efficient way to evaluate the discrete 
+   convolution of a long signal with a finite impulse response filter.
+*/
+void FFTbasedBlockConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal);
+
+}
+
--- a/src/featbin/Makefile
+++ b/src/featbin/Makefile
@@ -14,11 +14,11 @@ BINFILES = compute-mfcc-feats compute-plp-feats compute-fbank-feats \
     apply-cmvn-sliding compute-cmvn-stats-two-channel compute-kaldi-pitch-feats \
     process-kaldi-pitch-feats compare-feats wav-to-duration add-deltas-sdc \
     compute-and-process-kaldi-pitch-feats modify-cmvn-stats wav-copy \
-    append-vector-to-feats detect-sinusoids
+    wav-reverberate append-vector-to-feats detect-sinusoids
 
 OBJFILES = 
 
-TESTFILES =
+TESTFILES = 
 
 ADDLIBS = ../feat/kaldi-feat.a ../transform/kaldi-transform.a ../gmm/kaldi-gmm.a \
          ../thread/kaldi-thread.a ../tree/kaldi-tree.a ../matrix/kaldi-matrix.a \
--- a/src/featbin/wav-reverberate.cc
+++ b/src/featbin/wav-reverberate.cc
@@ -1,0 +1,255 @@
+// featbin/wav-reverberate.cc
+
+// Copyright 2015  Tom Ko
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/wave-reader.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
+/*
+   This function is to repeatedly concatenate signal1 by itself 
+   to match the length of signal2 and add the two signals together.
+*/
+void AddVectorsOfUnequalLength(const Vector<BaseFloat> &signal1, Vector<BaseFloat> *signal2) {
+  for (int32 po = 0; po < signal2->Dim(); po += signal1.Dim()) {
+    int32 block_length = signal1.Dim();
+    if (signal2->Dim() - po < block_length) block_length = signal2->Dim() - po;
+    signal2->Range(po, block_length).AddVec(1.0, signal1.Range(0, block_length));
+  }
+}
+
+BaseFloat MaxAbsolute(const Vector<BaseFloat> &vector) {
+  return std::max(std::abs(vector.Max()), std::abs(vector.Min()));
+}
+
+/* 
+   Early reverberation component of the signal is composed of reflections 
+   within 0.05 seconds of the direct path signal (assumed to be the peak of 
+   the room impulse response). This function returns the energy in 
+   this early reverberation component of the signal. 
+   The input parameters to this function are the room impulse response, the signal
+   and their sampling frequency respectively.
+*/
+BaseFloat ComputeEarlyReverbEnergy(const Vector<BaseFloat> &rir, const Vector<BaseFloat> &signal,
+                                   BaseFloat samp_freq) {
+  int32 peak_index = 0;
+  rir.Max(&peak_index);
+  KALDI_VLOG(1) << "peak index is " << peak_index;
+
+  const float sec_before_peak = 0.001;
+  const float sec_after_peak = 0.05;
+  int32 early_rir_start_index = peak_index - sec_before_peak * samp_freq;
+  int32 early_rir_end_index = peak_index + sec_after_peak * samp_freq;
+  if (early_rir_start_index < 0) early_rir_start_index = 0;
+  if (early_rir_end_index > rir.Dim()) early_rir_end_index = rir.Dim();
+
+  int32 duration = early_rir_end_index - early_rir_start_index;
+  Vector<BaseFloat> early_rir(rir.Range(early_rir_start_index, duration));
+  Vector<BaseFloat> early_reverb(signal);
+  FFTbasedBlockConvolveSignals(early_rir, &early_reverb);
+
+  // compute the energy
+  return VecVec(early_reverb, early_reverb) / early_reverb.Dim();
+}
+
+/*
+   This is the core function to do reverberation and noise addition
+   on the given signal. The noise will be scaled before the addition
+   to match the given signal-to-noise ratio (SNR) and it will also concatenate
+   itself repeatedly to match the length of the signal.
+   The input parameters to this function are the room impulse response,
+   the sampling frequency, the SNR(dB), the noise and the signal respectively.
+*/
+void DoReverberation(const Vector<BaseFloat> &rir, BaseFloat samp_freq,
+                        BaseFloat snr_db, Vector<BaseFloat> *noise,
+                        Vector<BaseFloat> *signal) {
+  if (noise->Dim()) {
+    float input_power = ComputeEarlyReverbEnergy(rir, *signal, samp_freq);
+    float noise_power = VecVec(*noise, *noise) / noise->Dim();
+    float scale_factor = sqrt(pow(10, -snr_db / 10) * input_power / noise_power);
+    noise->Scale(scale_factor);
+    KALDI_VLOG(1) << "Noise signal is being scaled with " << scale_factor
+                  << " to generate output with SNR " << snr_db << "db\n";
+  }
+
+  FFTbasedBlockConvolveSignals(rir, signal);
+
+  if (noise->Dim() > 0) {
+    AddVectorsOfUnequalLength(*noise, signal);
+  }
+}
+}
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+
+    const char *usage =
+        "Corrupts the wave files supplied via input pipe with the specified\n"
+        "room-impulse response (rir_matrix) and additive noise distortions\n"
+        "(specified by corresponding files).\n";
+
+    ParseOptions po(usage);
+    std::string noise_file;
+    BaseFloat snr_db = 20;
+    bool multi_channel_output = false;
+    int32 input_channel = 0;
+    int32 rir_channel = 0;
+    int32 noise_channel = 0;
+    bool normalize_output = true;
+    BaseFloat volume = 0;
+
+    po.Register("multi-channel-output", &multi_channel_output,
+                "Specifies if the output should be multi-channel or not");
+    po.Register("input-wave-channel", &input_channel,
+                "Specifies the channel to be used from input as only a "
+                "single channel will be used to generate reverberated output");
+    po.Register("rir-channel", &rir_channel,
+                "Specifies the channel of the room impulse response, "
+                "it will only be used when multi-channel-output is false");
+    po.Register("noise-channel", &noise_channel,
+                "Specifies the channel of the noise file, "
+                "it will only be used when multi-channel-output is false");
+    po.Register("noise-file", &noise_file,
+                "File with additive noise");
+    po.Register("snr-db", &snr_db,
+                "Desired SNR(dB) of the output");
+    po.Register("normalize-output", &normalize_output,
+                "If true, then after reverberating and "
+                "possibly adding noise, scale so that the signal "
+                "energy is the same as the original input signal.");
+    po.Register("volume", &volume,
+                "If nonzero, a scaling factor on the signal that is applied "
+                "after reverberating and possibly adding noise. "
+                "If you set this option to a nonzero value, it will be as"
+                "if you had also specified --normalize-output=false.");
+
+    po.Read(argc, argv);
+    if (po.NumArgs() != 3) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    if (multi_channel_output) {
+      if (rir_channel != 0 || noise_channel != 0)
+        KALDI_WARN << "options for --rir-channel and --noise-channel"
+                      "are ignored as --multi-channel-output is true.";
+    }
+
+    std::string input_wave_file = po.GetArg(1);
+    std::string rir_file = po.GetArg(2);
+    std::string output_wave_file = po.GetArg(3);
+
+    WaveData input_wave;
+    {
+      Input ki(input_wave_file);
+      input_wave.Read(ki.Stream());
+    }
+
+    const Matrix<BaseFloat> &input_matrix = input_wave.Data();
+    BaseFloat samp_freq_input = input_wave.SampFreq();
+    int32 num_samp_input = input_matrix.NumCols(),  // #samples in the input
+          num_input_channel = input_matrix.NumRows();  // #channels in the input
+    KALDI_VLOG(1) << "sampling frequency of input: " << samp_freq_input
+                  << " #samples: " << num_samp_input
+                  << " #channel: " << num_input_channel;
+    KALDI_ASSERT(input_channel < num_input_channel);
+
+    WaveData rir_wave;
+    {
+      Input ki(rir_file);
+      rir_wave.Read(ki.Stream());
+    }
+    const Matrix<BaseFloat> &rir_matrix = rir_wave.Data();
+    BaseFloat samp_freq_rir = rir_wave.SampFreq();
+    int32 num_samp_rir = rir_matrix.NumCols(),
+          num_rir_channel = rir_matrix.NumRows();
+    KALDI_VLOG(1) << "sampling frequency of rir: " << samp_freq_rir
+                  << " #samples: " << num_samp_rir
+                  << " #channel: " << num_rir_channel;
+    if (!multi_channel_output) {
+      KALDI_ASSERT(rir_channel < num_rir_channel);
+    }
+
+    Matrix<BaseFloat> noise_matrix;
+    if (!noise_file.empty()) {
+      WaveData noise_wave;
+      {
+        Input ki(noise_file);
+        noise_wave.Read(ki.Stream());
+      }
+      noise_matrix = noise_wave.Data();
+      BaseFloat samp_freq_noise = noise_wave.SampFreq();
+      int32 num_samp_noise = noise_matrix.NumCols(),
+            num_noise_channel = noise_matrix.NumRows();
+      KALDI_VLOG(1) << "sampling frequency of noise: " << samp_freq_noise
+                    << " #samples: " << num_samp_noise
+                    << " #channel: " << num_noise_channel;
+      if (multi_channel_output) {
+        KALDI_ASSERT(num_rir_channel == num_noise_channel);
+      } else {
+        KALDI_ASSERT(noise_channel < num_noise_channel);
+      }
+    }
+
+    int32 num_output_channels = (multi_channel_output ? num_rir_channel : 1);
+    Matrix<BaseFloat> out_matrix(num_output_channels, num_samp_input);
+
+    for (int32 output_channel = 0; output_channel < num_output_channels; output_channel++) {
+      Vector<BaseFloat> input(num_samp_input);
+      input.CopyRowFromMat(input_matrix, input_channel);
+      float power_before_reverb = VecVec(input, input) / input.Dim();
+
+      int32 this_rir_channel = (multi_channel_output ? output_channel : rir_channel);
+      Vector<BaseFloat> rir(num_samp_rir);
+      rir.CopyRowFromMat(rir_matrix, this_rir_channel);
+      rir.Scale(1.0 / (1 << 15));
+
+      Vector<BaseFloat> noise(0);
+      if (!noise_file.empty()) {
+        noise.Resize(noise_matrix.NumCols());
+        int32 this_noise_channel = (multi_channel_output ? output_channel : noise_channel);
+        noise.CopyRowFromMat(noise_matrix, this_noise_channel);
+      }
+
+      DoReverberation(rir, samp_freq_rir, snr_db, &noise, &input);
+
+      float power_after_reverb = VecVec(input, input) / input.Dim();
+
+      if (volume > 0)
+        input.Scale(volume);
+      else if (normalize_output)
+        input.Scale(sqrt(power_before_reverb / power_after_reverb));
+
+      out_matrix.CopyRowFromVec(input, output_channel);
+    }
+
+    WaveData out_wave(samp_freq_input, out_matrix);
+    Output ko(output_wave_file, false);
+    out_wave.Write(ko.Stream());
+
+    return 0;
+  } catch(const std::exception &e) {
+    std::cerr << e.what();
+    return -1;
+  }
+}
+
--- a/egs/aspire/s5/local/multi_condition/read_rir.py
+++ b/egs/aspire/s5/local/multi_condition/read_rir.py
@@ -13,14 +13,14 @@ def read_raw(input_filename, precision = np.float32):
 
 def wav_write(file_handle, fs, data):
   if str(data.dtype) in set(['float64', 'float32']):
-    data = (0.99 * data / np.max(np.abs(data))) * (2 ** 31)
-    data = data.astype('int32', copy = False)
-  elif str(data.dtype) == 'int32':
+    data = (0.99 * data / np.max(np.abs(data))) * (2 ** 15)
+    data = data.astype('int16', copy = False)
+  elif str(data.dtype) == 'int16':
     pass
   else:
     raise Exception('Not implemented for '+str(data.dtype))
   scipy.io.wavfile.write(file_handle, fs, data)
- 
+
 
 def usage():
   return """This is a python script to read impulse responses stored in custom formats. It handles AIR database."""
--- a/egs/aspire/s5/local/multi_condition/reverberate_wavs.py
+++ b/egs/aspire/s5/local/multi_condition/reverberate_wavs.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # Copyright 2014  Johns Hopkins University (Authors: Vijayaditya Peddinti).  Apache 2.0.
+#           2015  Tom Ko
 # script to generate multicondition training data / dev data / test data
 import argparse, glob, math, os, random, scipy.io.wavfile, sys
 
@@ -23,18 +24,9 @@ def return_nonempty_lines(lines):
 
   return new_lines
 
-def exists_wavfile(file_name):
-  return os.path.isfile(file_name)
-  try:
-    scipy.io.wavfile.read(file_name)
-    return True
-  except IOError:
-    return False
-
 if __name__ == "__main__":
   parser = argparse.ArgumentParser()
   parser.add_argument('--snrs', type=str, default = '20:10:0', help='snrs to be used for corruption')
-  parser.add_argument('--num-files-per-job', type=int, default = None, help='number of commands to be stored in each file')
   parser.add_argument('--check-output-exists', type = str, default = 'True', help = 'process file only if output file does not exist', choices = ['True', 'true', 'False', 'false'])
   parser.add_argument('--random-seed', type = int, default = 0, help = 'seed to be used in the randomization of impulses')
   parser.add_argument('wav_file_list', type=str, help='wav.scp file to corrupt')
@@ -75,49 +67,32 @@ if __name__ == "__main__":
         raise Exception('Unknown format of ' + file)
       impulse_noise_index.append([impulses_set, noises_list])
 
-  
-  if params.num_files_per_job is None:
-    lines_per_file = len(wav_files)
-  else:
-    lines_per_file = params.num_files_per_job
-  num_parts = int(math.ceil(len(wav_files)/ float(lines_per_file)))
-  indices_per_file = map(lambda x: xrange(lines_per_file * (x-1), lines_per_file * x), range(1, num_parts))
-  indices_per_file.append(xrange(lines_per_file * (num_parts-1), len(wav_files)))
- 
-  part_counter = 1
-  commands_file_base, ext = os.path.splitext(params.output_command_file)
-  for indices in indices_per_file:
-    command_list = []
-    for i in indices:
-      wav_file = " ".join(wav_files[i].split()[1:])
-      output_wav_file = wav_out_files[i]
-      impulse_file = impulses.next()
-      noise_file = ''
-      snr = ''
-      found_impulse = False
-      if add_noise:
-        for i in xrange(len(impulse_noise_index)):
-          if impulse_file in impulse_noise_index[i][0]:
-            noise_file = impulse_noise_index[i][1].next()
-            snr = snrs.next()
-            assert(len(wav_file.strip()) > 0)
-            assert(len(impulse_file.strip()) > 0)
-            assert(len(noise_file.strip()) > 0)
-            assert(len(snr.strip()) > 0)
-            assert(len(output_wav_file.strip()) > 0)
-            command_list.append("{0} --rir-file {1} --noise-file {2} --snr-db {3} - {4} \n".format(wav_file, impulse_file, noise_file, snr, output_wav_file))
-            found_impulse = True
-            break
-      if not found_impulse:
-        assert(len(wav_file.strip()) > 0)
-        assert(len(impulse_file.strip()) > 0)
-        assert(len(output_wav_file.strip()) > 0)
-        command_list.append("{0} --rir-file {1} - {2} \n".format(wav_file, impulse_file, output_wav_file))
-      if exists_wavfile(output_wav_file):
-        # we perform the check at this point to ensure replication of (wavfile, impulse, noise, snr) tuples across runs.
-        command_list.pop()  
-    file_handle = open("{0}.{1}{2}".format(commands_file_base, part_counter, ext), 'w')
-    part_counter += 1
-    file_handle.write("".join(command_list))
-    file_handle.close()
-  print num_parts
+  command_list = []
+  for i in range(len(wav_files)):
+    wav_file = " ".join(wav_files[i].split()[1:])
+    output_wav_file = wav_out_files[i]
+    impulse_file = impulses.next()
+    noise_file = ''
+    snr = ''
+    found_impulse = False
+    if add_noise:
+      for i in xrange(len(impulse_noise_index)):
+        if impulse_file in impulse_noise_index[i][0]:
+          noise_file = impulse_noise_index[i][1].next()
+          snr = snrs.next()
+          assert(len(wav_file.strip()) > 0)
+          assert(len(impulse_file.strip()) > 0)
+          assert(len(noise_file.strip()) > 0)
+          assert(len(snr.strip()) > 0)
+          assert(len(output_wav_file.strip()) > 0)
+          command_list.append("{4} {0} wav-reverberate --noise-file={2} --snr-db={3} - {1} - |\n".format(wav_file, impulse_file, noise_file, snr, output_wav_file))
+          found_impulse = True
+          break
+    if not found_impulse:
+      assert(len(wav_file.strip()) > 0)
+      assert(len(impulse_file.strip()) > 0)
+      assert(len(output_wav_file.strip()) > 0)
+      command_list.append("{2} {0} wav-reverberate - {1} - |\n".format(wav_file, impulse_file, output_wav_file))
+  file_handle = open(params.output_command_file, 'w')
+  file_handle.write("".join(command_list))
+  file_handle.close()
--- a/egs/aspire/s5/local/multi_condition/rirs/prep_aalto.sh
+++ b/egs/aspire/s5/local/multi_condition/rirs/prep_aalto.sh
@@ -7,6 +7,7 @@
 
 download=true
 sampling_rate=8k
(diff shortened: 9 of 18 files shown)


egs/aspire/s5/local/multi_condition/reverberate_data_dir.sh
```diff
@@ -72,12 +57,10 @@ for file in cmvn.scp feats.scp reco2file_and_channel; do
 done
 
 python local/multi_condition/get_reverberate_parameter_lists.py \
-  --snrs $snrs --num-files-per-job $num_files_per_job --random-seed $random_seed \
+  --snrs $snrs --random-seed $random_seed \
 $src_dir/wav.scp $log_dir/corrupted_${random_seed}.list $impnoise_dir \
-$log_dir/corrupt_wavs.${random_seed}.list > $log_dir/num_corruption_jobs || exit 1;
+$dest_dir/wav.scp || exit 1;
```
vijayaditya: remove the "wc" command. At least print a message to say what the number means.

egs/aspire/s5/local/multi_condition/get_reverberate_parameter_lists.py
```diff
@@ -75,17 +67,10 @@ def exists_wavfile(file_name):
...
-  if params.num_files_per_job is None:
-    lines_per_file = len(wav_files)
-  else:
-    lines_per_file = params.num_files_per_job
-  num_parts = int(math.ceil(len(wav_files)/ float(lines_per_file)))
-  indices_per_file = map(lambda x: xrange(lines_per_file * (x-1), lines_per_file * x), range(1, num_parts))
-  indices_per_file.append(xrange(lines_per_file * (num_parts-1), len(wav_files)))
+  lines_per_file = len(wav_files)
+  indices_per_file = map(lambda x: xrange(lines_per_file * (x-1), lines_per_file * x), range(1, 1))
+  indices_per_file.append(xrange(0, len(wav_files)))
```
vijayaditya: with the new changes using nested for loops with indices_per_file and indices does not make any sense. Use a single for loop with range(len(wav_files)).

egs/aspire/s5/local/multi_condition/reverberate_data_dir.sh
```diff
@@ -72,12 +57,10 @@ for file in cmvn.scp feats.scp reco2file_and_channel; do
 done
 
 python local/multi_condition/get_reverberate_parameter_lists.py \
```
vijayaditya: as you are directly generating the wav.scp rename this python script to reverberate_wavs.py

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
+// featbin/extract-segments.cc
```
vijayaditya: Please add a small unit test to confirm that the block convolution program generates the same results as the full signal convolution.

vijayaditya:
A side note:
We need to test out the speed of using wav-reverberate vs. extract-reverb-segments. A simple test for this would be to compare running
 wav-reverberate on the full signal followed by extract-segments multiple times.
vs
running wav-reverberate on random signals of the size of the segments (to emulate running extract-reverb-segments)

You can use sox to generate random signals.

src/featbin/wav-reverberate.cc
tomkocse: Do you want to add the unit test into the wav-reverberate.cc itself ?

vijayaditya: It would be difficult to run the unit test from the command line if it is in the same binary, so please add it to a new binary. (wav-reververate-test)

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+#include "util/common-utils.h"
+#include "feat/wave-reader.h"
+
+namespace kaldi {
+
+void FFTConvolution(Vector<BaseFloat> &x, Vector<BaseFloat> &h) {
+  int xdim = x.Dim();
+  int M = h.Dim();
+
+  KALDI_LOG << "M is " << M;
```
vijayaditya: Make the log output more descriptive.
e.g. "Length of the filter is " << M;

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  int N = 2;
```
vijayaditya: use more descriptive names for the variables. e.g. signal_length, filter_length, output_length.

If you are using the variables names to reflect corresponding quantities in an equation it is advisable to have the equations in the comments, to ensure that the code is easy to follow.

danpovey: Actually, that looks like a temporary logging statement that should be
removed entirely (programs should not produce spam on the output).
Also, the name is not Google compliant, should be FftConvolution, but also
function names should be verbs not nouns so it needs further change.  [I'm
not sure what it does so I can't say].  And non-const references as
function arguments are not allowed.

Dan

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+  while (N < 4 * M) {
+    N = N * 2;
+  }
+  KALDI_LOG << "Best N is " << N;
```
vijayaditya: Make the log output more descriptive.

src/featbin/wav-reverberate.cc
vijayaditya: Adding a verbosity argument to the binary and printing these statements when debugging level verbosity is required might be more desirable.

danpovey: The  --verbose argument is added by default.
Dan

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+    {
+      Input ki(source_file);
+      src_wave.Read(ki.Stream());
+    }
+    const Matrix<BaseFloat> &SRC = src_wave.Data();
+    BaseFloat samp_freq_src = src_wave.SampFreq();  // read sampling fequency
+    int32 num_samp_src = SRC.NumCols(),  // number of samples in recording
+          num_chan_src = SRC.NumRows();  // number of channels in recording
+
+    KALDI_LOG << "samp_freq of source is " << samp_freq_src
```
vijayaditya: change to something like

if (GetVerboseLevel() >= debug_verbose_level) {
      KALDI_LOG << "Sampling frequency of input wave file is " << ...
......
}

Please confirm what is the verbose level for debug mode and use that value for debug_verbose_level

danpovey: It's a gradation.  Zero is the default but above zero you can get
increasingly verbose.  You chould do

 KALDI_VLOG(1) << "samp_freq of source is " << samp_freq_src;


There is no need for GetVerboseLevel().

danpovey: ... but a lot of those messages looked not like well-thought-out
informational or debug messages but like temporarily things he put in to
debug the code.


Here:

  WaveData src_wave;
> +    {
> +      Input ki(source_file);
> +      src_wave.Read(ki.Stream());
> +    }

it should be

ReadKaldiObject(source_file, &src_wave);

And upper-case variable names like SRC are not allowed.

Dan

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+    }
+    const Matrix<BaseFloat> &RIR = rir_wave.Data();
+    BaseFloat samp_freq_rir = rir_wave.SampFreq();  // read sampling fequency
+    int32 num_samp_rir = RIR.NumCols(),  // number of samples in recording
+          num_chan_rir = RIR.NumRows();  // number of channels in recording
+    KALDI_LOG << "samp_freq of rir is " << samp_freq_rir
+               << " num_samp " << num_samp_rir
+               << " num_chan " << num_chan_rir;
+
+    Vector<BaseFloat> src(num_samp_src);
```
vijayaditya: New feature suggestion :
It would be better if you can support multi-channel outputs.

In the multi-channel output mode:
You assume you have single channel source file. If the user supplies multi-channel source, you give them an option to specify the channel to use (default to 0 as you do now).

You can work out the number of output channels from the number of channels in the RIR. 

Verify that the RIR and noise files have the same number of channels. We will assume that they are recorded in the same setup for now. 

This is not the case in the single channel mode, where we will allow RIRs and noise files with different number of channels but always select a particular channel specified by the user (defaulting to 0).

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+
+    Vector<BaseFloat> noise(0);
+    if (!noise_file.empty()) {
+      float peak = 0;
+      int peak_index = 0;
+      peak = rir.Max(&peak_index);
+      KALDI_LOG << "peak is " << peak;
+      KALDI_LOG << "peak_index is " << peak_index;
+
+      int early_start = peak_index - 0.001 * samp_freq_rir;
```
vijayaditya: early_start => early_rir_start_index
early_end => early_rir_end_index

declare constants to store the values 0.001 and 0.05, to ensure clarity

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+      int early_end = peak_index + 0.05 * samp_freq_rir;
+      if (early_start < 0) early_start = 0;
+      if (early_end > rir.Dim()) early_end = rir.Dim();
+
+      Vector<BaseFloat> early_rir(early_end - early_start);
+      SubVector<BaseFloat> subr(rir, early_start, early_end - early_start);
+      early_rir.CopyFromVec(subr);
+
+      Vector<BaseFloat> early_src(num_samp_src);
```
vijayaditya: rename subr.
rename early_src to early_reverb

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+      early_src.CopyFromVec(src);
+      KALDI_LOG << "Run early FFTConvolution ";
+      FFTConvolution(early_src, early_rir);
+
+      // compute the energy
+      for (int32 j = 0; j < early_src.Dim(); j++)
```
vijayaditya: use VecVec() operation to compute energy.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+        early_src(j) *= early_src(j);
+      float src_power = early_src.Sum() / early_src.Dim();
+
+      WaveData noise_wave;
+      {
+        Input ki(noise_file);
+        noise_wave.Read(ki.Stream());
+      }
+      const Matrix<BaseFloat> &NOISE = noise_wave.Data();
```
vijayaditya: variable names with caps not allowed. Rename NOISE accordingly.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+            num_chan_noise = NOISE.NumRows();
+      KALDI_LOG << "samp_freq of noise is " << samp_freq_noise
+                << " num_samp " << num_samp_noise
+                << " num_chan " << num_chan_noise;
+      noise.Resize(num_samp_noise);
+      noise.CopyRowFromMat(NOISE, 0);
+
+      // compute the noise energy
+      float n_power_sum = 0;
+      for (int32 j = 0; j < noise.Dim(); j++)
```
vijayaditya: Use VecVec() to compute energy

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+    } else {
+      for (int j = 0; j < xdim - po; j++)
+        x(po + j) = x_block_padded(j);
+      for (int j = 0; j < M-1 && j < xdim - po; j++)
+        x(po + j) += temp_pad(j);
+    }
+  }
+}
+
+void RepeatedlyAdd(Vector<BaseFloat> &x, Vector<BaseFloat> &h) {
```
vijayaditya: modify the signature to something like

void AddVectorsOfUnequalLength(const Vector<BaseFloat> &vector1, Vector<BaseFloat> *vector2 )

vijayaditya:
Break the code in main function into smaller parts. Move everything other than data reading and writing to separate functions.
Move early reverb energy estimation to separate function called
ComputeEarlyReverbEnergy(const Vector<BaseFloat> & room_impulse_response, const Vector<BaseFloat>& signal, BaseFloat * early_reverb_energy )

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    KALDI_LOG << "Run FFTConvolution ";
+    FFTConvolution(src, rir);
+
+    if (noise.Dim() > 0) {
+      RepeatedlyAdd(src, noise);
+    }
+
+    float src_max2 = src.Min();    // for checking the negative max
```
vijayaditya: move this code to a separate function called

MaxAbsolute(const Vector<BaseFloat> & vector, BaseFloat * max_value)

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,250 @@
...
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/wave-reader.h"
+
+namespace kaldi {
+
+void FFTConvolution(Vector<BaseFloat> &x, Vector<BaseFloat> &h) {
```
vijayaditya: Modify the signature to

BlockConvolveSignals(const Vector<BaseFloat> & signal1, Vector<BaseFloat>* signal2)

src/featbin/wav-reverberate.cc
tomkocse: I get an error:
error: no matching function for call to ‘kaldi::WaveData::Read(std::istream&, bool&)’
when i change to ReadKaldiObject(source_file, &src_wave);
Can you advise how can i resolve that?

kkm000:
I may be misunderstanding what the code does, but just want to point FWIW that `sox` supports both FIR convolution (`fir` effect) and adding noise (although the latter is a bit, ehm, [convoluted](http://linguistics.berkeley.edu/plab/guestwiki/index.php?title=Sox_in_phonetic_research#Add_noise_to_an_audio_file)).

src/featbin/wav-reverberate.cc
tomkocse: Then i will add two options for the user: --source-channel  and --rir-channel.
When --rir-channel is used, the multi-channel output mode will be on, otherwise is the ingle channel output mode
Is this fine?

vijayaditya:
The code computes the scaling coefficient for noise signal, this involves
computing the early reverberation component of the reverberated signal .
This makes a sox based implementation a bit circuitous.

Vijay

On Tue, Aug 18, 2015 at 5:45 AM, Kirill Katsnelson <notifications@github.com

src/featbin/wav-reverberate.cc
vijayaditya: Variable name source is a bit confusing in this context, please switch to input_wave
The newly needed options would be
--multi-channel-output (true|false)
--input-wave-channel (non-negative integers)
--rir-channel (will be used when multi-channel-output is false, else ignored with warning)
--noise-channel (will be used when multi-channel-output is false, else ignored with warning)

remember to check that the channel indices are non-negative.

tomkocse: Just to clarify,
when --multi-channel-output is true, the output wave will contain multiple channels and the number of channels equals the number of channels of the RIR file, correct?

By the way, if we allow users to select different channels in the RIR and noise file, can we skip verifying the RIR and noise file to have same number of channels? This provides users for having a greater flexibility in the combination of RIR and noise.

vijayaditya: >Just to clarify,
>when --multi-channel-output is true, the output wave will contain multiple channels and the number of channels equals the number of channels of the RIR file, correct?

Yes

>By the way, if we allow users to select different channels in the RIR and noise file, can we skip verifying the RIR and noise file to have same number of channels? This provides users for having a greater flexibility in the combination of RIR and noise.

Channel selection should only supported in single channel output mode. If they don't have the same number of channels the RIR and noise are probably not recorded in the same setup. In this case I don't know if it even makes sense to add the noise after reverberation. However lets leave this choice to the user for now. Do not check that the RIR and noise files have same number of channels in single channel mode.

danpovey:
Vijay, let me know when you think this is ready to merge.

tomkocse:
I have made the changes according to all your comments.
The multi channel output mode is added.
The test mode (with option --test-mode) is added by comparing block-based convolution and full signal convolution.

danpovey:
Thanks.  Vijay, please confirm when you think it's ready to commit, and
I'll quickly look over the code again before merging (if I find no issues).
Dan

vijayaditya:
I am going through it now, will add comments soon.

Vijay

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,331 @@
...
+  for (int po = 0; po < signal_length; po += block_length) {
+    // get a block of the signal
+    int process_length = block_length;
+    if (signal_length - po < process_length) process_length = signal_length - po;
+    signal_block_padded.SetZero();
+    signal_block_padded.Range(0, process_length).CopyFromVec(signal->Range(po, process_length));
+
+    srfft_->Compute(signal_block_padded.Data(), true);
+
+    for (int i = 0; i < fft_length / 2; i++) {
```
vijayaditya: Separate out the for loop into a function like ElementwiseProductOfFft. Same in the ConvolveSignals block above.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,331 @@
...
+
+    ParseOptions po(usage);
+    std::string noise_file;
+    BaseFloat snr_db = 20;
+    bool test_mode = false;
+    bool multi_channel_output = false;
+    int input_wave_channel = 0;
+    int rir_channel = 0;
+    int noise_channel = 0;
+
```
vijayaditya: Add more description for the optional arguments.

e.g.
multi-channel-output : true | false, specifies if the output should be multi-channel or not.

input-wave-channel : specifies the channel to be used from input as only a single channel will be used to generate reverberated output.
rir-channel : specifies the channel of the room impulse response, it will only be used when multi-channel-output is false
noise-channel : specifies the channel of the noise file, it will only be used when multi-channel-output is false

test-mode: true|false, if true the binary runs a unit test which confirms if the block-convolution and convolution outputs

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    Matrix<BaseFloat> out_matrix(num_chan_output, num_samp_input);
+
+    do {
+      Vector<BaseFloat> input(num_samp_input);
+      input.CopyRowFromMat(input_matrix, input_wave_channel);
+      Vector<BaseFloat> rir(num_samp_rir);
+      rir.CopyRowFromMat(rir_matrix, rir_chan);
+      rir.Scale(1 / static_cast<float>(2147483647));  // Assume 32 bits
```
vijayaditya: Why are you assuming 32bits ? You have no control on the pipeline which generates the RIR, so you should check if this is correct. Please see if the wavereader object can provide this information.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,331 @@
...
+    WaveData rir_wave;
+    {
+      Input ki(rir_file);
+      rir_wave.Read(ki.Stream());
+    }
+    const Matrix<BaseFloat> &rir_matrix = rir_wave.Data();
+    BaseFloat samp_freq_rir = rir_wave.SampFreq();  // read sampling fequency
+    int32 num_samp_rir = rir_matrix.NumCols(),  // number of samples in recording
+          num_chan_rir = rir_matrix.NumRows();  // number of channels in recording
+    KALDI_LOG << "samp_freq of rir is " << samp_freq_rir
```
vijayaditya: You should eliminate these logging statements or at least move them to the debug log. Same for noise and input wave files. 

You should reduce the size of the main function, it is 170 lines. You should keep it around 40. This makes it easier for reading. You could move everything other than file reading and corresponding asserts to new functions. e.g. create a new function for computing each channel output in the while loop @ line 273.
Move the code under testmode to a function like UnitTestBlockConvolution.

src/featbin/wav-reverberate.cc
danpovey: Generally speaking, option explanations should begin with a capital letter.  There is no reason to say true|false, because the boilerplate text that is printed out explains that it's boolean and mentions the default value.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,331 @@
...
+
+      for (int32 j = 0; j < input.Dim(); j++)
+        input(j) = (0.99 * input(j) / input_max) * pow(2, 15);
+
+      out_matrix.CopyRowFromVec(input, out_chan);
+      out_chan++;
+      rir_chan++;
+      noise_chan++;
+    } while (out_chan < num_chan_output);
+
```
danpovey: Above, the use of a do... while loop makes the code hard to read, as do the variables out_chan, rir_chan and noise_chan.
What would be clearer, I think, is something like this:
Firstly, if multi_channel_output is true, then check that the various other channel options were not set, and print an error message and die if so.
Then do something like

int32 num_output_channels = (multi_channel_output ? num_chan_rir : 1);
  for (int32 output_channel = 0; output_channel < num_output_channels;
       output_channel++) {
    int32 this_rir_channel = (multi_channel_output ? output_channel : rir_channel);
   // it's not good to use differently-spelled versions of the same thing as distinct variables, e.g. chan versus channel.

src/featbin/wav-reverberate.cc
danpovey: Vijay, does the absolute scaling of the rir even matter?  Is it supposed to be normalized somehow?  I am a bit confused.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,331 @@
...
+BaseFloat MaxAbsolute(const Vector<BaseFloat> &vector) {
+  float max = abs(vector.Max());
+  float max2 = abs(vector.Min());
+
+  if (max2 > max)
+    max = max2;
+
+  return max;
+}
+
```
danpovey: I would do just
return std::max(std::abs(vector.Max()), vector.Min())));

src/featbin/wav-reverberate.cc
vijayaditya: You are right it is no longer necessary. In the python script this was done as I was operating on different types of wav files (float, PCM, ...), so I would convert everything to floats, before convolution.

But Kaldi allows only PCM data, so we can just just directly use **rir**.

tomkocse: I think the scaling is still needed.
The elements of the RIR file are in very large scale.
This makes the signal still in the large scale after the convolution which result in a chance of getting this error:
ERROR (wav-reverberate:Write():wave-reader.cc:343) Wave file is out of range for 16-bit.

I now change the scaling to 
rir.Scale(1 / MaxAbsolute(rir));
to scale down the elements to be between [-1, 1 ]

tomkocse:
wav-reverberate.cc is revised again.

danpovey:
I wonder if it would make sense to normalize so that the energy is the same
as it was before, after reverberation?
Vijay, what do you think?
Dan

vijayaditya:
Scaling the entire signal, after reverberation and noise addition, would not affect the SNR of the signal.
So as long as we just scale the final signal before writing it we should be fine. But since this is operation might not be expected by the user, we should print out a warning. If printing a warning is considered to be verbose, we should at least point this out in the Usage() output.

danpovey:
One way would be to have an option to do it that's on by default?
I am a little confused as to what is the "proper" way to interpret the
scaling of signals, or of an RIR.  E.g. if there is an 8-bit PCM, is one
supposed to scale it by 256 before writing as 16-bit PCM?  For an RIR, are
we supposed to know how many bits were in the PCM?  I am wondering whether
some extension of the WaveReader class might be desirable, to make this
kind of stuff easier to deal with?
Dan

vijayaditya:
Scaling the input PCM data, with the maximum value of the x-bit PCM seems to be an acceptable way to do the intX to float conversion (widely used in Matlab codes I came across). Extending the wavereader would be desirable.

danpovey:
Hm.
The main thing I am concerned about is compatibility with the old, existing
code.
When we wrote the original MFCC and PLP extraction code we had in mind
compatibility with HTK, and HTK does not do that type of scaling; it uses
the raw pcm values.  So we can't have it that the WaveData/WaveHolder code
automatically do this for all waves on reading.
What I think would be acceptable is to modify the WaveData code so that if
it reads PCM data which is not 16 bits, it normalizes it to have the range
it would have if it were raw 16-bit PCM data.  Of course this should be
prominently documented, as should the fact that the scale of the data is
-2^15 to 2^15.  All of the recipes use 16-bit PCM, I believe, and I think
if people gave them 8-bit or 32-bit PCM and didn't do CMN, they would just
be getting nonsense results.  So in most cases this change would be
harmless.

Dan

vijayaditya:
OK. Starting a separate issue to track the WaveData change would be convenient.

@tomkocse can just assume that the data is 16bit PCM and scale the RIR accordingly. 

@tomkocse also remember to delete the file get_reverber*.py as it no longer being used.

kkm000:
Having all wave data normalized to BaseFloats [-1.0, 1.0] would make the signal path easier to understand. The colloquial "16-bit PCM" is a signed integer sample [-2^16, 2^16-1], "8-bit PCM" is an unsigned integer sample [0, 255] biased to 128 = 0V, and "32-bit PCM" is _usually_ an IEEE float [-1.0, 1.0]. I would avoid propagating all this mess through the signal path, and rather chop it away at the reader.

danpovey:
>> Having all wave data normalized to BaseFloats [-1.0, 1.0] would make the
signal path easier to understand. The colloquial "16-bit PCM" is a signed
integer sample [-2^16, 2^16-1], "8-bit PCM" is an unsigned integer sample
[0, 255] biased to 128 = 0V, and *"32-bit PCM" is **usually** an IEEE float
[-1.0, 1.0]*. I would avoid propagating all this mess through the signal
path, and rather chop it away at the reader.

I did not realize that 32-bit PCM is usually an IEEE float.  The current
reading code assumes that it is a 32-bit integer.  It's possible that we
have never encountered such data.  Are you sure about this?

Obviously if we could go back in time the way you are saying it would be
the best strategy, but I am concerned about back compatibility.  The
features would become incompatible unless we multiplied them by 2^16 during
feature extraction, and that is a big deal; it would break a lot of
setups.  Even if we compensated for this using CMN (which not all recipes
use!  e.g. online-nnet2), the dithering that happens in the feature code
would need to be changed to use 2^-15 as the scale on the dithering by
default instead of 1.0.
And existing code that other people might have written that reads features
in, would start behaving differently even for standard 16-bit features.
How about we have a parallel strategy: change WaveReader to normalize
everything to as if it were signed 16-bit integer (so max = 2^15), and also
create NormalizedWaveReader/NormalizedWaveHolder classes (maybe using
inheritance) that will normalize to [-1.0, 1.0].  (We can't use options
classes as there is no way to pass options into a Holder).  Note: if the
modified WaveReader encounters any non-16-bit data it should print a
warning, once [using a static variable to guard the warning], that the
behavior has changed.

By the way, it just occurred to me that if we normalized to [-1.0, 1.0]
it's unlikely that this would be the correct normalization for room impulse
responses, because a room impulse response that leaves energy unchanged
might be expected to sometimes have excursions outside the range[-1.0, 1.0].

Dan

danpovey:
BTW, I noticed this in the wav-reading code.  It looks like it does not
support the IEEE float type and will die if it sees it.
Dan


    // Support only KSDATAFORMAT_SUBTYPE_PCM for now. Interesting formats:


    //("00000001-0000-0010-8000-00aa00389b71", KSDATAFORMAT_SUBTYPE_PCM)


    //("00000003-0000-0010-8000-00aa00389b71",
KSDATAFORMAT_SUBTYPE_IEEE_FLOAT)

    //("00000006-0000-0010-8000-00aa00389b71", KSDATAFORMAT_SUBTYPE_ALAW)


    //("00000007-0000-0010-8000-00aa00389b71", KSDATAFORMAT_SUBTYPE_MULAW)


    if (guid1 != 0x00000001 || guid2 != 0x00100000 ||
        guid3 != 0xAA000080 || guid4 != 0x719B3800) {
      KALDI_ERR << "WaveData: unknown/unsupported WAVE_FORMAT_EXTENSIBLE
format.";
    }

kkm000:
I totally agree, as a compromise solution, we could always normalize to 2^15×[-1, 1]. As long as this is documented, it works.

I do not think that a discrete impulse response would overshoot 1.0. That would mean the filter is pumping energy into the response. Am I following you here?

danpovey:
Regarding the impulse response magnitude...  it probably depends whether we
measure the gain with respect to white noise or some other reference
signal; I'm not sure if there is a standard way.  Anyway, Vijay and Tom
will figure that out.

On Fri, Aug 21, 2015 at 1:35 PM, Kirill Katsnelson <notifications@github.com

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+  int filter_length = filter.Dim();
+
+  int fft_length = 2;
+  while (fft_length < signal_length + filter_length - 1) {
+    fft_length = fft_length * 2;
+  }
+  KALDI_VLOG(1) << "fft_length for full signal convolution is " << fft_length;
+
+  SplitRadixRealFft<BaseFloat> *srfft_;
+  srfft_ = new SplitRadixRealFft<BaseFloat>(fft_length);
```
kkm000: Remove the `_` at the end, not a member variable. 
Also, join these 2 lines.
I do not see where `srfft_` is deleted.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+  int signal_length = signal->Dim();
+  int filter_length = filter.Dim();
+
+  KALDI_VLOG(1) << "Length of the filter is " << filter_length;
+
+  int fft_length = 2;
+  while (fft_length < 4 * filter_length) {
+    fft_length = fft_length * 2;
+  }
+  KALDI_VLOG(1) << "Best FFT length is " << fft_length;
```
kkm000: If I were you, I'd separate this as RoundUpToPowerOf2() or like.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+
+  int block_length = fft_length - filter_length + 1;
+  KALDI_VLOG(1) << "Block size is " << block_length;
+  SplitRadixRealFft<BaseFloat> *srfft_;
+  srfft_ = new SplitRadixRealFft<BaseFloat>(fft_length);
+
+  Vector<float> filter_padded(fft_length);
+  filter_padded.Range(0, filter_length).CopyFromVec(filter);
+  srfft_->Compute(filter_padded.Data(), true);
```
kkm000: srttf to _, and it leaked again.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+
+  Vector<float> temp_pad(filter_length - 1);
+  temp_pad.SetZero();
+  Vector<float> signal_block_padded(fft_length);
+
+  for (int po = 0; po < signal_length; po += block_length) {
+    // get a block of the signal
+    int process_length = block_length;
+    if (signal_length - po < process_length) process_length = signal_length - po;
```
kkm000: To me, it would be a bit easier to read as 
```
if (process_length > signal_length - po) process_length = signal_length - po;
```
or, which arguably reads better,
```
int process_length = std::min(block_length, signal_length - po);
```

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+  }
+}
+
+BaseFloat MaxAbsolute(const Vector<BaseFloat> &vector) {
+  return std::max(std::abs(vector.Max()), std::abs(vector.Min()));
+}
+
+BaseFloat ComputeEarlyReverbEnergy(const Vector<BaseFloat> &rir, const Vector<BaseFloat> &signal,
+                                   BaseFloat samp_freq) {
+      int peak_index = 0;
```
kkm000: Indent ran away in whole function.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+
+      const float sec_before_peak = 0.001;
+      const float sec_after_peak = 0.05;
+      int early_rir_start_index = peak_index - sec_before_peak * samp_freq;
+      int early_rir_end_index = peak_index + sec_after_peak * samp_freq;
+      if (early_rir_start_index < 0) early_rir_start_index = 0;
+      if (early_rir_end_index > rir.Dim()) early_rir_end_index = rir.Dim();
+
+      int duration = early_rir_end_index - early_rir_start_index;
+      Vector<BaseFloat> early_rir(rir.Range(early_rir_start_index, duration));
```
kkm000: This is probably not a big deal because of a small size, but if `BlockConvolveSignals()` could take the first arg as a `VectorBase<>`, the matrix copy would not be necessary here, you could use the result of `.Range()` directly. Not sure if `VectorBase<>` works there, just a suggestion.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+    float scale_factor = sqrt(pow(10, -snr_db / 10) * input_power / noise_power);
+    noise->Scale(scale_factor);
+    KALDI_VLOG(1) << "scale_factor of channel: " << scale_factor << " for SNR db " << snr_db;
+  }
+
+  if (test_mode) {
+    UnitTestBlockConvolution(rir, signal);
+  } else {
+    BlockConvolveSignals(rir, signal);
+  }
```
kkm000: Second @vijayaditya comment, please separate into a library file and unit-test it separately from the utility. As is it will not be run by automated `make test` from src/. @dpovey, what would be the best library for signal-processing primitives? `src/feat`?

src/featbin/wav-reverberate.cc
kkm000: I think it is added it @danpovey's grandiose nnet3 merge couple days ago. Keep this change until  right after LGTM on all other points, since if you just pull master into your branch now the PR page will blow up to an unwieldy size.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+    Matrix<BaseFloat> out_matrix(num_output_channel, num_samp_input);
+
+    for (int32 output_channel = 0; output_channel < num_output_channel; output_channel++) {
+      Vector<BaseFloat> input(num_samp_input);
+      input.CopyRowFromMat(input_matrix, input_channel);
+
+      int32 this_rir_channel = (multi_channel_output ? output_channel : rir_channel);
+      Vector<BaseFloat> rir(num_samp_rir);
+      rir.CopyRowFromMat(rir_matrix, this_rir_channel);
+      rir.Scale(1 / MaxAbsolute(rir));
```
kkm000: Ok, this is now incorrect. The maximum sample value in a signal is pretty much a random number, and scaling to it makes little sense.

The scale multiplier should be 2^-15. Perhaps it makes sense to add a constant (into `wave-reader.h` I think?) to represent this value. @danpovey., what do you think the best way to handle this?

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,325 @@
...
+        noise.Resize(noise_matrix.NumCols());
+        int32 this_noise_channel = (multi_channel_output ? output_channel : noise_channel);
+        noise.CopyRowFromMat(noise_matrix, this_noise_channel);
+      }
+
+      DoReverberation(rir, &noise, &input, samp_freq_rir, snr_db, test_mode);
+
+      float input_max = MaxAbsolute(input);
+      for (int32 j = 0; j < input.Dim(); j++)
+        input(j) = (0.99 * input(j) / input_max) * pow(2, 15);
```
kkm000: This also messes with the amplitude pretty randomly. Normalizing to the max almost always means dropping enegry too low. Normally is DSP you just clamp overshooting samples. It might be a good idea to print a warning like "12345 samples (3% of total) were clamped, ~~deal with it~~ reduce volume".

vijayaditya: I agree. This was a temporary hack I had used in the python script to deal with overshooting values. The overshooting was inevitable due to noise addition. There were no pathological situations like generation of a low energy output signal, due to scaling with large max value, for the  audio database and (RIR, noise) database I had used. But scaling with <code>input_max</code> is definitely not advisable for a generic tool like <code>wav-reverberate</code>

@tomkocse  One way to avoiding a lot of overshooting samples is to determine the scaling coefficient from the entire database of reverberated and noised wav files. This would be outside the scope of <code>wav-reverberate</code>, when called by feature extraction code, as it would be operating on single audio files. So you could take this value as an input from the user (<code>--scaling-coefficient</code>). 

The user is still forced to determine the scaling coefficient through trial & error, depending on the number of overshoot warnings generated. A better solution would be to add a mode to the <code>wav-reverberate</code> binary, where given an input scp file (with several wav files and corresponding rir files, noise files and SNR values) you determine the scaling coefficient which reduces the overall overshoot to say less than 2%.

kkm000: @vijayaditya: I would even call the argument `--gain-db`, it is more transparent that "scaling" which can refer to anything numeric, and also expressed in dB. "If it's complaining too much of clipping, just drop it by 6dB". It does not have to be as precise as figuring out by scanning many files, I think. But this is indeed outside of this discussion's scope, as you noted.

vijayaditya: @kkm000 <code>--gain-db</code> sounds clearer than <code>--scaling-coefficient</code>.

I think doing the auto scaling to control overshoot, might be desirable for users, not interested in maintaining consistency across wave files.This could be enabled using <code>--reduce-clipping</code>.

I would also recommend adding a mode where <code>wav-reverberate</code> determines a global scaling coefficient for the given set of wav files, rir files, noise files and SNRs. This would help in ensuring better control on the signal processing pipeline and would help in maintaining consistency with some [standard reverberation scripts](http://reverb2014.dereverberation.com/download.html).

danpovey: Guys,
The solutions you are suggesting for the scaling problem are starting to sound extremely complicated and hard to implement.  I think what would be best is to just state that this program normalizes the output so that the overall signal energy after reverberation and adding noise will be the same as the signal input.  Then if you don't like the clipping, you scale down the input.   The code in wav-reverberate could, after scaling, clip the signal to the range [-1, 1], or have the WaveDataNormalized::Write function do this (and in either case, print out a warning).  Naturally, class WaveDataNormalized would have to document whether it expects the input to be clipped, or whether it does the clipping itself.
Regarding Kirill's question about where code of this nature should go-- yes, in src/feat/.  E.g. in src/feat/signal.h.
Do either of Vijay or Kirill have time to make the changes to the wave-reading code, and add the normalized versions?  That should probably be finalized before this pull request is finished.

src/featbin/wav-reverberate.cc
danpovey: You are right, it should be just
 SplitRadixRealFft<BaseFloat> *srfft(fft_length);

danpovey: You are right, it should just be
SplitRadixRealFft<BaseFloat> srfft(fft_length);
[in case "stack allocation" was not clear to Tom.]

src/featbin/wav-reverberate.cc
vijayaditya: On Aug 21, 2015 7:25 PM, "Daniel Povey" <notifications@github.com> wrote:
>
> In src/featbin/wav-reverberate.cc:
>
> > +      Vector<BaseFloat> rir(num_samp_rir);
> > +      rir.CopyRowFromMat(rir_matrix, this_rir_channel);
> > +      rir.Scale(1 / MaxAbsolute(rir));
> > +
> > +      Vector<BaseFloat> noise(0);
> > +      if (!noise_file.empty()) {
> > +        noise.Resize(noise_matrix.NumCols());
> > +        int32 this_noise_channel = (multi_channel_output ?
output_channel : noise_channel);
> > +        noise.CopyRowFromMat(noise_matrix, this_noise_channel);
> > +      }
> > +
> > +      DoReverberation(rir, &noise, &input, samp_freq_rir, snr_db,
test_mode);
> > +
> > +      float input_max = MaxAbsolute(input);
> > +      for (int32 j = 0; j < input.Dim(); j++)
> > +        input(j) = (0.99 * input(j) / input_max) * pow(2, 15);
>
> Guys,
> The solutions you are suggesting for the scaling problem are starting to
sound extremely complicated and hard to implement. I think what would be
best is to just state that this program normalizes the output so that the
overall signal energy after reverberation and adding noise will be the same
as the signal input. Then if you don't like the clipping, you scale down
the input. The code in wav-reverberate could, after scaling, clip the
signal to the range [-1, 1], or have the WaveDataNormalized::Write function
do this (and in either case, print out a warning). Naturally, class
WaveDataNormalized would have to document whether it expects the input to
be clipped, or whether it does the clipping itself.

OK. I am worried that after noise addition to have 0dB SNR, the output
signal+noise energy would be comparatively higher. Making its energy
similar to input might lead to a waveform with lower signal (not signal +
noise) energy than the input. This is the reason I suggest we just give the
user the option to specify a constant scaling factor. We can delegate the
responsibility of determining this factor to the user. The binary need not
provide an option to do this.

> Regarding Kirill's question about where code of this nature should go--
yes, in src/feat/. E.g. in src/feat/signal.h.
> Do either of Vijay or Kirill have time to make the changes to the
wave-reading code, and add the normalized versions? That should probably be
finalized before this pull request is finished.

I can do this. But I am not familiar with this module so I might take a day.

kkm000: Dan, I think I do not understand why there should be the "normalizing" wave reader. You suggested assuming 2^15×[-1, 1] as the normal range for digital audio in Kaldi. This is what we already get from linear PCM 16-bit files. Now, I do not know if there are any recipes working with non-16-bit audio, but I would be surprised if there were any: [this code](https://github.com/kaldi-asr/kaldi/blob/master/src/feat/wave-reader.cc#L275) is quite problematic for both 8 and 32 bit files. All 8-bit PCM files I came across in my life were unsigned int biased samples: 0x0 encodes -1.0V, 0x80 encodes 0V and 0xFF is for +1.0V. This is not what the code assumes, and I do not think that features would be sensible (it essentially reads signal as if transformed by discontinuous transformation V_out = V_in >= 0 ? V_in : 1.0 - V_in). I have seen errors quite less subtle go on unnoticed for years, but I do not think that cepsral coefficients would be botched. For 32 bit, I never seen a standard-headed (WAVEFORMATEX, not WAVEFORMATEXTENSIBLE) 32-bit file. I added very basic support for WAVEFORMATEXTENSIBLE PCM only back in April.

So in the end, it is very unlikely that either 8 or the 32 bit path was ever taken by any code. If you agree with this premise, then the wave reading/writing code is already normalizing; all that remains is to fix the 8 and 32 bit cases to agree with the internal representation of 2^15×[-1, 1] as the normal range.

kkm000: > fix the 8 and 32 bit cases to agree
Or drop them altogeter, because no one uses them anyway.

danpovey: I would be OK with simply dropping the 8 and 32-bit support and clearly
documenting that the WaveReader supports only 16-bit PCM and that the range
of the input/output is -2^15 to 2^15.  Probably we should also add support
for truncating samples that are too large or small, but of course print a
suitable warning message about truncated samples.

Dan


On Fri, Aug 21, 2015 at 5:52 PM, Kirill Katsnelson <notifications@github.com

danpovey: Guys, from this thread
https://sourceforge.net/p/kaldi/discussion/1355347/thread/4f07e084/?limit=25#1741
it looks like the 8-bit path of the code was used.
So I think the way forward right now could be to just modify the existing
8-bit and 32-bit code to scale the same as 16-bit.
This probably does make more sense than unnecessarily duplicating the code.

Dan

kkm000: I'll do the fixing then, if it is not very urgent. Seems that this PR will not be affected.

The term is "clip," not "truncate." Just FYI so it won't have inadvertently snuck into function names or comments.

What I want to do is try training a system on 8 bit samples (before fixing the code). Naively it seems that this should be a total wreck, but sometimes ML can extract information from an impossible mess (the incorrect sample decoding as it currently is does not lose information per se, as, amazingly, does not FFT). I just want to see what happens!

src/featbin/wav-reverberate.cc
tomkocse: It is possible to cast SubVector<> to Vector<> ?
As .Range() return SubVector<> type.

kkm000: No, as in both "generally no" and in "you probably do not want to do that." Instead, if you can use only the `VectorBase`'s methods in `BlockConvolveSignals()`, you can avoid the copy. If for any reason you cannot, then a copy is unavoidable.

src/featbin/wav-reverberate.cc
danpovey: Kirill, I don't think there's any point doing the 8-bit training.  Probably
no-one ever successfully used the 8-bit reading code.
I could not find any official Microsoft documentation of the fact that the
8-bit PCMs are unsigned, but according to https://en.wikipedia.org/wiki/WAV,
"There are some inconsistencies in the WAV format: for example, 8-bit data
is unsigned while 16-bit data is signed, and many chunks duplicate
information found in other chunks."
We already support only a subset of the WAV format, and I'd be happier to
go to an even smaller subset and require people to format their data to
16-bit linear PCM using sox or similar.  Let's just remove the 8-bit
support, because it looks like it never would have worked.  Also according
to http://wiki.multimedia.cx/?title=PCM
it looks like 32-bit PCM is very rare.  So I think we should just remove
the 32-bit support too, and declare that the WaveReader supports only
16-bit PCM and does not do any rescaling.
@vpeddinti, please note!
Dan



In src/featbin/wav-reverberate.cc

danpovey: BTW, another amusing comment on the WAV/RIFF specification (which might
explain some of the problems it has caused us): from
https://en.wikipedia.org/wiki/WAV,
"The RIFF specification attempts to be a formal specification, but its
formalism lacks the precision seen in other tagged formats."
I suppose this is an instance of the hacker mentality and disdain for
formal education that Microsoft had in its early years.




On Fri, Aug 21, 2015 at 10:28 PM, Kirill Katsnelson <

vijayaditya: @danpovey  OK, I stopped working on NormalizedWaveReader.

@tomkocse To summarize yesterday's discussion I think the TODOs are as follows ( @danpovey or @kkm000 please correct if necessary) : 

 1. Scale the RIRs, input wave and noise files with 1/2^15. Once the signal processing operations are done scale with 2^15 and call the wave writer.
 2. To avoid overshoot, compute the energy of the input wave file and scale the final reverberated+noised wav file to have this energy
 3. My suggestion : also have a constant scale option where you just scale the final reverberated+noised wav file with a user specified value.
 4. Move the unit-test to wav-reverberate-test.cc, to ensure this test is run by <code>make test</code>
 5. Remember to delete the python script get_reverber*.py as it is no longer used.

danpovey: >
> @danpovey <https://github.com/danpovey> OK, I stopped working on
> NormalizedWaveReader.
>
> @tomkocse <https://github.com/tomkocse> To summarize yesterday's
> discussion I think the TODOs are as follows ( @danpovey
> <https://github.com/danpovey> or @kkm000 <https://github.com/kkm000>
> please correct if necessary) :
>
>    1. Scale the RIRs, input wave and noise files with 1/2^15. Once the
>    signal processing operations are done scale with 2^15 and call the wave
>    writer.
>    2. To avoid overshoot, compute the energy of the input wave file and
>    scale the final reverberated+noised wav file to have this energy
>    3. My suggestion : also have a constant scale option where you just
>    scale the final reverberated+noised wav file with a user specified value.
>
> OK; but Vijay, please clarify whether this scale option is supposed to be
applied after the normalize-to-input-energy, or as an alternative.  Maybe
writing out the proposed usage message and option name would be better.
Vijay, will you be checking in the removal of 8-bit and 32-bit support in
the WaveReader code?

vijayaditya: <code> 
--normalize-output : true|false (default: true), if true the output is normalized to have the same energy as the input
</code>
<code>
--volume : float (default: 0) If not 0, this value would be multiplied with the output, for amplitude adjustment to avoid overshooting. Specifying a non-zero value also disables the normalize-output option.
</code>

@kkm000  I chose linear amplitude adjustment in place of dB scale adjustments as I thought that volume adjustment for avoiding overshooting would be easier in the linear scale. I changed the option to <code>--volume</code> for better clarity and consistency with <code>Sox</code>.

@danpovey I will check-in the modifications to WaveReader

danpovey: OK, let me edit these slightly to be less ambiguous.  Remember that boolean
options are automatically identified as boolean to the user by the
ParseOptions code, and so is the default value.

  opts.Register("normalize-output", &normalize_output, "If true, then after
reverberating and "
               "possibly adding noise, scale so that the signal energy is
the same as the original input "
               " signal.");
  opts.Register("volume", &volume, "If nonzero, a scaling factor on the
signal that is applied "
                "after reverberating and possibly adding noise.  If you set
this option to a nonzero value, "
               "it will be as if you had also specified
--normalize-output=false.");

I agree that a scaling factor is better than a decibel value-- decibels are
defined in a stupid way and I can never remember what it is.

Dan

kkm000:
@vijayaditya: There is no need to scale the signal at all, convolution is a linear operator. The impulse needs to be scaled to [-1; 1] though.

vijayaditya:
@kkm000 The RIR filter can increase the energy in the signal. Further we also add noise, so overshoots are common and scaling is essential.

kkm000:
@vijayaditya: What I mean is that it is not necessary to multiply the signal and noise files by 2^-15 before processing and then multiply it by 2^15 before output. This is number 1 in your list. 

Adjusting the gain post process is entirely another thing. There are multiple possible ways; you may adjust the filter gain to 1 at DC for example (for a FIR filter the gain at DC is simply the sum of the coefficients). The energy approach is entirely possible too; this is usually understood to equalize the perceived loudness. In both cases, you do not even need to pre-scale even the IR, as the equalization (of either DC gain or energy) will take care of the extra 2^15 multiplier.

Neither equalizaion will guarantee no clipping though.

kkm000:
@vijayaditya: and I can take care of the 8 and 32 but paths in WaveReader if you want. This change is not related to the code in this PR, so we can put that aside now.

vijayaditya:
On Sun, Aug 23, 2015 at 1:32 AM, Kirill Katsnelson <notifications@github.com
> wrote:

> @vijayaditya <https://github.com/vijayaditya>: What I mean is that it is
not necessary to multiply the signal and noise files by 2^-15 before
processing and then multiply it by 2^15 before output. This is number 1 in
your list.
Oh, OK. I suggested doing this just as a matter of convenience.

>Adjusting the gain post process is entirely another thing. There are
multiple possible ways; you may adjust the filter gain to 1 at DC for
example (for a FIR filter the gain at DC is simply the sum of the
coefficients). The energy approach is entirely possible too; this is
usually understood to equalize the perceived loudness. In both cases, you
do not even need to pre-scale even the IR, as the equalization (of either
DC gain or energy) will take care of the extra 2^15 multiplier. Neither
equalizaion will guarantee no clipping though.

vijayaditya:
Thanks, that would be helpful.

On Sun, Aug 23, 2015 at 1:50 AM, Kirill Katsnelson <notifications@github.com

tomkocse:
If i am not going to multiply the signal and noise files by 2^-15 before processing and then multiply it by 2^15 before output,
do I still need to take care of the overshooting by 
      float input_max = MaxAbsolute(input);
      input.Scale(pow(2, 15) * 0.99 / input_max);
before writing the output ?
Or should i leave it to WaveReader to handle it ?

tomkocse:
Should the --volume option be default 1 instead of default 0 ?

danpovey:
>If i am not going to multiply the signal and noise files by 2^-15 before
processing and then multiply it by 2^15 before output,
>do I still need to take care of the overshooting by
>float input_max = MaxAbsolute(input);
>input.Scale(pow(2, 15) * 0.99 / input_max);
> before writing the output ?

No, you don't have to do  that.

> Or should i leave it to WaveReader to handle it ?

Leave it to the WaveReader code.  However, @kkm000, please change the wave
writing code so that instead of just dying for out-of-range data, it clips
it and then warns how many samples were clipped.

> Should the --volume option be default 1 instead of default 0 ?

No, zero.  Look at the usage message I wrote-- it's only active if nonzero,
and if zero, the output volume is adjusted so that the signal energy is the
same as the input.

Dan

src/featbin/wav-reverberate-test.cc
```diff
@@ -0,0 +1,143 @@
+// featbin/extract-segments.cc
```
vijayaditya: Do not reimplement convolution and block-convolution. Include them using a header file.

Change the comment at the top of the file, it says 
>// featbin/extract-segments.cc

kkm000: @vijayaditya: perhaps it makes sense to put these DSP functions in one of the libraries, either `matrix/` (this is where we have `SplitRadixRealFft`) or `feat/`. @danpovey, what would you recommend? We do not run tests in `*bin/` directories.

danpovey: Either feat/ or matrix/ are suitable places to put convolution code in
principle,  but the bar is higher if you want to include them in matrix/.
That is, you'd need to think carefully about the interface and make sure
that the code handles all edge cases in a good way, does a reasonable thing
for all sizes of input, and is tested to a high standard.  It's your choice
what you want to do.
Dan

vijayaditya:
@tomkocse Could you decide whether you want to implement the convolution operations in feat/ or matrix/ ?
In case you decide on implementing it in feat/. You should create files .h and .cc files say feat/fft-convolution.{h,.cc}. Move the the convolve and block convolve methods to these files. You should also create a corresponding unit-test feat/fft-convolution-test.cc. Then modify wav-reverberate.cc to include the header file created above.

danpovey:
If in feat/, please put it in signal.{h,cc} because then we can put other
signal-level things there- I don't want to have a header per function.
Dan

tomkocse:
signal.{h,cc}  are added and i moved the unit test to signal-test.cc.
We don't run tests in featbin/ now.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/wave-reader.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
+void AddVectorsOfUnequalLength(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal) {
```
vijayaditya: Using the name filter for the first argument is confusing as you are performing convolution in this function. Better rename the arguments to signal1 and signal2

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+BaseFloat MaxAbsolute(const Vector<BaseFloat> &vector) {
+  return std::max(std::abs(vector.Max()), std::abs(vector.Min()));
+}
+
+BaseFloat ComputeEarlyReverbEnergy(const Vector<BaseFloat> &rir, const Vector<BaseFloat> &signal,
+                                   BaseFloat samp_freq) {
+  int peak_index = 0;
+  rir.Max(&peak_index);
+  KALDI_VLOG(1) << "peak index is " << peak_index;
+
```
vijayaditya: Add a brief text comment here describing what you are doing.

"Early reverberation component of the signal is composed of reflections within 0.05 seconds of the direct path signal (assumed to be the peak of the room impulse response). This function computes the energy in this early reverberation component of the signal."

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+void DoReverberation(const Vector<BaseFloat> &rir, Vector<BaseFloat> *noise,
+                        Vector<BaseFloat> *signal, BaseFloat samp_freq,
+                        BaseFloat snr_db) {
+  if (noise->Dim()) {
+    float input_power = ComputeEarlyReverbEnergy(rir, *signal, samp_freq);
+    float noise_power = VecVec(*noise, *noise) / noise->Dim();
+    float scale_factor = sqrt(pow(10, -snr_db / 10) * input_power / noise_power);
+    noise->Scale(scale_factor);
+    KALDI_VLOG(1) << "scale_factor of channel: " << scale_factor << " for SNR db " << snr_db;
```
vijayaditya: KALDI_VLOG(1) << "Noise signal is being scaled with the value " << scale_factor << " to generate output with SNR  " << snr_db <<  "db\n";

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+
+    po.Read(argc, argv);
+    if (po.NumArgs() != 3) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    if (multi_channel_output) {
+      if (rir_channel != 0 || noise_channel != 0)
+        KALDI_ERR << "Do not set --rir-channel or --noise-channel "
```
vijayaditya: It might be better to just generate a warning saying that the code will ignore rir_channel and noise_channel.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+
+    WaveData input_wave;
+    {
+      Input ki(input_wave_file);
+      input_wave.Read(ki.Stream());
+    }
+
+    const Matrix<BaseFloat> &input_matrix = input_wave.Data();
+    BaseFloat samp_freq_input = input_wave.SampFreq();  // read sampling fequency
+    int32 num_samp_input = input_matrix.NumCols(),  // number of samples in recording
```
vijayaditya: use consistent terminology. you are using "recording", "input_wave" to refer to the same thing

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+
+  int duration = early_rir_end_index - early_rir_start_index;
+  Vector<BaseFloat> early_rir(rir.Range(early_rir_start_index, duration));
+  Vector<BaseFloat> early_reverb(signal);
+  BlockConvolveSignals(early_rir, &early_reverb);
+
+  // compute the energy
+  return VecVec(early_reverb, early_reverb) / early_reverb.Dim();
+}
+
```
vijayaditya: [When defining a function, parameter order is: inputs, then outputs.](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Function_Parameter_Ordering)

Add a comment describing what the function does above it.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,234 @@
...
+                    << " #samples: " << num_samp_noise
+                    << " #channel: " << num_noise_channel;
+      if (multi_channel_output) {
+        KALDI_ASSERT(num_rir_channel == num_noise_channel);
+      } else {
+        KALDI_ASSERT(noise_channel < num_noise_channel);
+      }
+    }
+
+    int32 num_output_channel = (multi_channel_output ? num_rir_channel : 1);
```
vijayaditya: num_output_channel -> num_output_channels

src/feat/signal.h
```diff
@@ -0,0 +1,30 @@
...
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+
+namespace kaldi {
+
+void ConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal);
+
```
vijayaditya: Briefly describe what each function does.

src/feat/signal-test.cc
```diff
@@ -0,0 +1,52 @@
...
+
+static void InitRand(Vector<BaseFloat> *v) {
+  for (MatrixIndexT i = 0;i < v->Dim();i++)
+    (*v)(i) = RandGauss();
+}
+
+void UnitTestBlockConvolution() {
+  for (int i = 0; i < 10; i++) {
+    int signal_length = 4000000 + Rand() % 400000;
+    int rir_length = 10000 + Rand() % 1000;
```
vijayaditya: change rir to filter.

src/feat/signal-test.cc
```diff
@@ -0,0 +1,52 @@
...
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
+static void InitRand(Vector<BaseFloat> *v) {
```
vijayaditya: Just use [SetRandn()](http://kaldi-asr.org/doc/classkaldi_1_1VectorBase.html#a40c5a82d6d015bdcc0079b535b1e9a23)

src/feat/signal-test.cc
```diff
@@ -0,0 +1,52 @@
...
+    int signal_length = 4000000 + Rand() % 400000;
+    int rir_length = 10000 + Rand() % 1000;
+    Vector<BaseFloat> signal(signal_length);
+    Vector<BaseFloat> rir(rir_length);
+    InitRand(&signal);
+    InitRand(&rir);
+    Vector<BaseFloat> signal_test(signal);
+    ConvolveSignals(rir, &signal_test);
+    BlockConvolveSignals(rir, &signal);
+    AssertEqual(signal, signal_test, 0.001 * signal.Dim());
```
vijayaditya: tolerance value (0.001) should probably be smaller.

tomkocse:
Added comments to the code.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,243 @@
...
+  return std::max(std::abs(vector.Max()), std::abs(vector.Min()));
+}
+
+/* 
+   Early reverberation component of the signal is composed of reflections 
+   within 0.05 seconds of the direct path signal (assumed to be the peak of 
+   the room impulse response). This function returns the energy in 
+   this early reverberation component of the signal. 
+   The input parameters to this function are the room impulse response, the signal
+   and their sampling freqency respectively.
```
kkm000: s/freqency/frequency/

danpovey:
Vijay, is this ready for me to review?

vijayaditya:
Yes, its ready except for one suggested change to function signature.

Vijay

danpovey:
I'll wait.

tomkocse:
Function signature changed.

src/feat/signal.h
```diff
@@ -0,0 +1,40 @@
...
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+
+namespace kaldi {
+
+/*
+   This function implements convolution of two signals.
+   However this should be an inefficient version of BlockConvolveSignals()
+   as it process the entire signal with a single FFT.
```
vijayaditya: process -> processes

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,255 @@
...
+
+    po.Read(argc, argv);
+    if (po.NumArgs() != 3) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    if (multi_channel_output) {
+      if (rir_channel != 0 || noise_channel != 0)
+        KALDI_WARN << "options for --rir-channel and --noise-channel will be ignored"
```
vijayaditya: options for --rir-channel and --noise-channel ~~will be~~ are ignored ~~when~~ as --multi-channel-output is true."

danpovey:
Can you please clean up the commit history to have a single commit with a suitable message?
Most likely you can do
git rebase -i HEAD~10
and you'll see something like this:

pick 48dc4cc nnet3: fix to how objective function is normalized in nnet3-show-progress
pick f569d0c Fixing the segfault issue in lattice-oracle.cc: used lattice composition instead of GetOracleLattice(); also fixed various style issues that I noticed.
pick 1df55e7 Correcting minor bug in ASpIRE recipe, doesn't affect results
pick d1aceed fixing the install script to fail correctly

# Rebase 3e76aa5..85aa227 onto 3e76aa5
#
# Commands:
#  p, pick = use commit
#  r, reword = use commit, but edit the commit message
#  e, edit = use commit, but stop for amending
#  s, squash = use commit, but meld into previous commit
#  f, fixup = like "squash", but discard this commit's log message
#  x, exec = run command (the rest of the line) using shell
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
# However, if you remove everything, the rebase will be aborted.
#

and you edit the first line's commit message to the overall commit message you want for the whole thing, and the other ones from 'pick' to 'f' (or 'fix'), and their commit messages will be ignored.
e.g.

pick 48dc4cc nnet3: fix to how objective function is normalized in nnet3-show-progress
f f569d0c Fixing the segfault issue in lattice-oracle.cc: used lattice composition instead of GetOracleLattice(); also fixed various style issues that I noticed.
f 1df55e7 Correcting minor bug in ASpIRE recipe, doesn't affect results
f d1aceed fixing the install script to fail correctly

Then push it (you may need an option like --no-ff) to force the push because it's not fast forward).

src/feat/signal.cc
```diff
@@ -0,0 +1,117 @@
...
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
+void ElementwiseProductOfFft(const Vector<BaseFloat> &a, Vector<BaseFloat> *b) {
+  for (int i = 0; i < a.Dim() / 2; i++) {
```
danpovey: It's good practice to use a temporary
e.g.
int32 num_fft_bins = a.Dim() / 2;
and do i < num_fft_bins, to avoid the function call being repeated.
Also, inside the loop, please replace your code with:
    // do complex multiplication b *= a.
    ComplexMul(a(2*i), a(2*i + 1), &(b(2*i)), &(b(2*i + 1)));
which will be clearer.

src/feat/signal.cc
```diff
@@ -0,0 +1,117 @@
...
+    float tmp = (*b)(2*i);
+    (*b)(2*i) = tmp * a(2*i) - (*b)(2*i+1) * a(2*i+1);
+    (*b)(2*i+1) = tmp * a(2*i+1) + (*b)(2*i+1) * a(2*i);
+  }
+}
+
+int RoundUpToPowerOf2(int a) {
```
danpovey: Use the function RoundUpToNearestPowerOfTwo(), which already exists in base/kaldi-math.h.

src/feat/signal.cc
```diff
@@ -0,0 +1,117 @@
...
+  int roundup = 2;
+  while (roundup < a) {
+    roundup *= 2;
+  }
+  return roundup;
+}
+
+void ConvolveSignals(const Vector<BaseFloat> &filter, Vector<BaseFloat> *signal) {
+  int signal_length = signal->Dim();
```
danpovey: The standard integer type we normally use in Kaldi is int32, so please replace your ints with that.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,255 @@
...
+    po.Register("input-wave-channel", &input_channel,
+                "Specifies the channel to be used from input as only a "
+                "single channel will be used to generate reverberated output");
+    po.Register("rir-channel", &rir_channel,
+                "Specifies the channel of the room impulse response, "
+                "it will only be used when multi-channel-output is false");
+    po.Register("noise-channel", &noise_channel,
+                "Specifies the channel of the noise file, "
+                "it will only be used when multi-channel-output is false");
+    po.Register("noise-file", &noise_file,
```
danpovey: you could likely fit this on one line within 80 chars..

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,255 @@
...
+      KALDI_VLOG(1) << "sampling frequency of noise: " << samp_freq_noise
+                    << " #samples: " << num_samp_noise
+                    << " #channel: " << num_noise_channel;
+      if (multi_channel_output) {
+        KALDI_ASSERT(num_rir_channel == num_noise_channel);
+      } else {
+        KALDI_ASSERT(noise_channel < num_noise_channel);
+      }
+    }
+
```
danpovey: this whole function seems very long, but I'm not going to absolutely insist on shortening it.

src/featbin/wav-reverberate.cc
```diff
@@ -0,0 +1,255 @@
...
+    for (int32 output_channel = 0; output_channel < num_output_channels; output_channel++) {
+      Vector<BaseFloat> input(num_samp_input);
+      input.CopyRowFromMat(input_matrix, input_channel);
+      float power_before_reverb = VecVec(input, input) / input.Dim();
+
+      int32 this_rir_channel = (multi_channel_output ? output_channel : rir_channel);
+      Vector<BaseFloat> rir(num_samp_rir);
+      rir.CopyRowFromMat(rir_matrix, this_rir_channel);
+      rir.Scale(1 / pow(2, 15));
```
danpovey: replace the 1, 2 and 15 with 1.0, 2.0 and 15.0 please.  better to use literal floats if you intend them to be evaluated in a floating-point context, as you do here.  1.0 / (1 << 15) would perhaps be preferable, though, as no dependency on -lmath.

src/featbin/Makefile
```diff
@@ -13,12 +13,12 @@ BINFILES = compute-mfcc-feats compute-plp-feats compute-fbank-feats \
     interpolate-pitch copy-feats-to-htk copy-feats-to-sphinx extract-rows \
     apply-cmvn-sliding compute-cmvn-stats-two-channel compute-kaldi-pitch-feats \
     process-kaldi-pitch-feats compare-feats wav-to-duration add-deltas-sdc \
-    compute-and-process-kaldi-pitch-feats modify-cmvn-stats wav-copy \
+    compute-and-process-kaldi-pitch-feats modify-cmvn-stats wav-copy wav-reverberate \
```
danpovey: line is likely over 80 characters -> shorten

src/feat/signal-test.cc
```diff
@@ -0,0 +1,47 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "feat/signal.h"
+
+namespace kaldi {
+
```
danpovey: Please add in this file a simple non-FFT-based convolution method, and verify for short signals and filters that it behaves the same as your FFT-based convolution.  Your testing method does not test for bugs that would affect both of the FFT-based methods (e.g. problems with the sign, or a scalar factor that shouldn't be there).

tomkocse: I have added a non-FFT-based convolution method, it passes the test when i set the tolerance value to 0.0001 * signal.Dim() but it fails when i make it smaller to 0.00001 * signal.Dim().
is it acceptable?

danpovey: Seems OK.

tomkocse: "Your testing method does not test for bugs that would affect both of the FFT-based methods (e.g. problems with the sign, or a scalar factor that shouldn't be there)." <---
Can you elaborate this and explicitly tell what cases should be tested?

danpovey: I was just explaining why you should implement the simple convolution.  No
further action required.

tomkocse:
The test for non-FFT-based convolution is added.
Other minor issues are handled.
I just rebased the commit history and pushed with option --force.

danpovey:
OK, I'll merge now.  Thanks for putting up with the process!

PR #79, opened 2015-08-19
Added Travis CI integration scripts

kkm000:
Added Travis CI integration scripts.
Modified makefiles so that failed tests keep their log files around, so that the CI can capture them.

See https://travis-ci.org/workflow-demo-org/kaldi/ (and https://travis-ci.org/kaldi-asr/kaldi/ after merge and first push). New PRs should automatically show a small information icon as to whether the Travis build is in progress, or has failed or succeeded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@
 *.a
 *.lib
 
+# Test log files
+*.testlog
+
 # emacs saves
 [#]*[#]
 .[#]*
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
     packages:
+      - gdb
       - gcc-4.9
       - g++-4.9
       - gfortran-4.9
@@ -23,6 +24,7 @@ branches:
     - master
 
 before_install:
+  - cat /proc/sys/kernel/core_pattern
   - export XROOT=~/xroot
   - tools/extras/travis_install_bindeps.sh $XROOT
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+notifications:
+  email: false
+
+language: cpp
+
+os:
+  - linux
+
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - gcc-4.9
+      - g++-4.9
+      - gfortran-4.9
+      - liblapack-dev
+#      - libopenblas-dev
+#      - libblas-dev
+
+branches:
+  only:
+    - master
+
+before_install:
+  - export XROOT=~/xroot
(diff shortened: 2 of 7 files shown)


kkm000:
Ha! Already running a test!

kkm000:
Don't merge. The failure seems genuine bug log is missing.

kkm000:
travis-ci/travis-ci/#3754 prevents core files creation, and therefore gdb backtraces. Otherwise this works as designed. The failure in nnet-component-test is genuine, I am reproducing it locally. This is the first bug caught with the CI.

This PR is good to go by me. Please review.

danpovey:
What does "command" in the makefile do?  `which command` on clsp returns nothing.
I'd rather not have the Makefile rely on something obscure to work, if we can  avoid id.
Dan

jtrmal:
IIRC,  command  can be used in similar way to which.  Command should be
more correct,  as which is often hijacked by package managers.  I think
it's bash builtin.
Y.

jtrmal:
I forgot to say I didn't look at that makefile /script so I don't know the
context exactly.
Y.

danpovey:
BTW, I should mention that I decided now might be as good a time as any to
merge nnet3 back to master.  I can create a separate branch for ongoing
work such as CTC.  I am working on the merging right now.  Let me know if
there are any objections.

Dan

kkm000:
`command -v` is the POSIX standard built-in, and is perhaps the most portable way to see if a command is available, as an executable in the path, or (unlike `which`) as a built-in. 
```
sh-4.3$ command -v gdb; echo $?
/usr/bin/gdb
0
sh-4.3$ command -v gdbzzz; echo $?
1
sh-4.3$ command -v echo; echo $?
echo
0
```

You should have used `command -v command` instead of `which command` to see that `command` is indeed available as a command :-)

Another common use of `command` is unaliasing. This is common e. g. in bash completion scripts, as they run in the interactive shell.
```
kkm@yupana:~$ alias ls='echo aliased away'
kkm@yupana:~$ ls -l *.txt
aliased away -l winregrate.txt
kkm@yupana:~$ command ls -l *.txt
-rw-rw-r-- 1 kkm kkm 48876 May 13 21:22 winregrate.txt
```

danpovey:
OK thanks.
I'll merge this after I finish the nnet3 merge, though [assuming no-one
objects to that].  It's becoming a headache to have two diverging
development branches, and the nnet3 stuff itself is pretty stable now.
Dan


On Wed, Aug 19, 2015 at 3:26 PM, Kirill Katsnelson <notifications@github.com

kkm000:
Sure, no problem. I'll rebase if this one is reported unmergeable.

danpovey:
Merged after checking nothing was broken.

PR #81, opened 2015-08-19
LSTM training scripts added (untested)

vijayaditya:
Added scripts for doing LSTM training. Configs generated seem fine, but train scripts are untested. This is an intermediate commit.


vijayaditya:
Will recreate pull request after combining commits.

PR #84, opened 2015-08-20
LSTM changes complete (untested)

vijayaditya:
Added scripts for doing LSTM (stacked-LSTM or stacked LSTM-DNN) training. LSTM configs generated seem fine, but train scripts are untested. This is an intermediate commit.

This pull request also adds nnet3 TDNN scripts for AMI (decoding untested).


danpovey:
Merging even if it's partial work, as  it won't break anything.

Issue #88, opened 2015-08-24
Wave{Reader,Writer} changes


kkm000:
For tracking required changes (Discussed in #76)
- [ ] WaveWriter should not crash on out of range data, rather print a warning and continue (rather urgent)
- [ ] ~~Read and write 8 and 32 bit formats correctly.~~ drop support for 8 and 32 bits (see discussion). Scale all internal data to 2‐15×[-1,1] floats.
- [ ] Explain why this range in Docs

danpovey:
> For tracking required changes
> [ ] WaveWriter should not crash on out of range data, rather print a
> warning and continue (rather urgent)
>
Agreed.

> [ ] Read and write 8 and 32 bit formats correctly. Scale all internal data
> to 2‐15×[-1,1] floats.
>

My understanding of the discussion was that we would simply discontinue
support for 8-bit and 32-bit.  But if you are able to get it working
correctly in those cases, I don't object.


> [ ] Explain why this range in Docs
>

Yes- just do it above the class definition in a /** ... */ comment, and
above class member-function declarations in /// comments.  It's not just
about the range-- general documentation of those classes is lacking IIRC.

Dan

kkm000:
Re 8 and 32: you found one example in the wild of a Kaldi user asking abut a 8 bit file, so I thought you wanted to fix it. But I am all for dropping it; 8 and 32 bit files are in the "grey area" where one would play with sox first to make sure the format is chewable. And then at this point, why not use sox anyway.

So, drop?

danpovey:
We later determined that the 8-bit code was definitely broken, so likely
no-one was ever using it successfully.   And also that 32-bit PCM files are
very rare.  So my vote is to drop it.
Dan

kkm000:
@dpovey: ok to rename `wave-reader` => `wave-io`?

danpovey:
Yes, but there is something we need to do with the Makefile when we make
these types of changes, some kind of "ar d" command.  @jtrmal would know.
Dan

jtrmal:
Yes, the makefile in the src/feat dir might need a custom make rule  adding
the "ar -d wave-reader.o"
This is to prevent duplication of the objects.
BTW: this is just to reduce the people that the "kaldi won't build" after
updating -- but make clean && make depend would solve this as well
y.

kkm000:
@jtrmal: so you think it would **not** be ok just skip adding the `ar d` command? The immediate question then is how long should be maintained.

[...]

PR #91, opened 2015-08-24
#88 item 1: clip wave samples and print a warning.

kkm000:
`wave-reader` also sorely lacks unit tests. I had to write a skeleton of what can become one, but it is not ready for the prime time. I'll try to develop it into one.

### DIFF
--- a/src/feat/wave-reader.cc
+++ b/src/feat/wave-reader.cc
@@ -20,6 +20,7 @@
 // limitations under the License.
 
 #include <cstdio>
+#include <limits>
 #include <sstream>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


src/feat/wave-reader.h
```diff
@@ -52,10 +52,13 @@
 
 namespace kaldi {
 
+/// For historical reasons, we scale waveforms to the range
+/// (2^15-1)*[-1, 1], not the usual default DSP range [-1, 1].
+const BaseFloat kWave0dBSampleValue = 32768.0;
```
danpovey: Could you please just call this kWaveSampleMax ?  I think it would be less confusing

src/feat/wave-reader.cc
```diff
@@ -354,6 +361,10 @@ void WaveData::Write(std::ostream &os) const {
   }
   if (os.fail())
     KALDI_ERR << "Error writing wave data to stream.";
+  if (num_clipped > 0)
+    KALDI_WARN << "WARNING: clipped " << num_clipped
+               << " samples out of total " << num_chan * num_samp
+               << ". Reduce volume.";
 }
```
danpovey: Could you please change the punctuation from "Reduce volume." to "Reduce volume?".  Just to make it more of a suggestion than an imperative.

kkm000:
So if the fix is urgently required, please review and merge if ok. If not, I can get to the other items in the same PR but not immediately. Busy, busy, busy...

src/feat/wave-reader.cc
kkm000: sure, a sec.

kkm000:
2 comments done.

danpovey:
Thanks!
Merging.
It's nice to have a genuine ex-Googler who really understands the style.

PR #92, opened 2015-08-25
update AddCols

naxingyu:
updates:
* GPU AddCols and test code
* CPU AddCols
* nnet convolutional component call for AddCols
* nnet2 convolutional1d component call for AddCols


danpovey:
Thanks- I'll look at this to-morrow.

PR #93, opened 2015-08-25
Added nnet3_to_dot.py script to convert nnet3 models to dot graphs

vijayaditya:
Added a script to convert nnet3-am-info output to a dot graph. This can be useful for debugging new nnet3 architectures. Tested on TDNN and LSTM architectures. 

Corrected minor bugs in lstm recipe and nnet3/nnet-simple-component.cc.

**Todo:**

1. Add processing code for additional descriptors: Switch, FailOver and Round
2. Add dot options for better display of complex graphs
3. Add logic to detect repeated sub-graphs and
    1. use a single sub-graph to represent these (e.g. recognizing the same input descriptor is used for     several components)
    2.  use an atomic representation for this sub-graph (e.g. recognizing that the Affine transform + Non-linearity + normalization layer can be collapsed into a single operation)

**LSTM model**
lstm

**TDNN model**
tdnn

### DIFF
--- a/egs/ami/s5/local/nnet3/run_lstm.sh
+++ b/egs/ami/s5/local/nnet3/run_lstm.sh
@@ -15,6 +15,7 @@ use_sat_alignments=true
 affix=
 speed_perturb=true
 splice_indexes="-2,-1,0,1,2 0"
+common_egs_dir=
 
 . cmd.sh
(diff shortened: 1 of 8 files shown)


danpovey:
Thanks!  Merging.

PR #95, opened 2015-08-26
add delta for ConvNet recipe

naxingyu:
This option was part of the recently created ConvNet recipe. I forgot to add it to that commit.


danpovey:
Accepting it because it's part of an existing recipe, but doing it like this is extremely against the spirit of the nnet3 tools.  It would have been better to implement it as part of the network itself, using a component with similar code to SpliceComponent.  However, until the convnet recipes become more popular it would probably be a waste of time to do this.

naxingyu:
Agreed. It should be implemented as a DeltaComponent or something. I am also trying implementing the ParallelComponent in nnet2, because I heard (from Baidu) that convolutional networks worked when it was used in a "sub-band" mode.

danpovey:
I think it would be much easier to do this type of thing (e.g.
ParallelComponent) in the nnet3 framework, where no code extensions would
be needed.
Dan

PR #96, opened 2015-08-26
initial commit for CUDA interface change

naxingyu:
@danpovey This commit is to clarify my understandings of the requirements. It doesn't compile yet. Please drop in any comments to keep it on the right track.

### DIFF
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -131,8 +131,8 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
 
 template<typename T>
 void CuArray<T>::CopyToHost(T *dst) const {
-  if (dst == NULL) return;
   if (dim_ == 0) return;
+  KALDI_ASSERT(dst != NULL);
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) { 
     Timer tim;
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -119,29 +119,29 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) { 
     Timer tim;
-    CU_SAFE_CALL(cudaMemcpy(&dst->front(), Data(), dim_*sizeof(T), cudaMemcpyDeviceToHost));
+    CU_SAFE_CALL(cudaMemcpy(&dst->front(), Data(), dim_ * sizeof(T), cudaMemcpyDeviceToHost));
     CuDevice::Instantiate().AccuProfile("CuArray::CopyToVecD2H", tim.Elapsed());
   } else
 #endif
   {
-    memcpy(&dst->front(), data_, dim_*sizeof(T));
+    memcpy(&dst->front(), data_, dim_ * sizeof(T));
   }
 }
 
 
 template<typename T>
-void CuArray<T>::CopyToVec(T *dst) const {
+void CuArray<T>::CopyToHost(T *dst) const {
   if (dst == NULL) return;
   if (dim_ == 0) return;
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) { 
     Timer tim;
-    CU_SAFE_CALL(cudaMemcpy(dst, Data(), dim_*sizeof(T), cudaMemcpyDeviceToHost));
+    CU_SAFE_CALL(cudaMemcpy(dst, Data(), dim_ * sizeof(T), cudaMemcpyDeviceToHost));
     CuDevice::Instantiate().AccuProfile("CuArray::CopyToVecD2H", tim.Elapsed());
   } else
 #endif
   {
-    memcpy(dst, data_, dim_*sizeof(T));
+    memcpy(dst, data_, dim_ * sizeof(T));
   }
 }
 
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -131,9 +131,7 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
 
 template<typename T>
 void CuArray<T>::CopyToVec(T *dst) const {
-  if (dst == NULL) {
-    dst = (T *)malloc(dim_);
-  }
+  if (dst == NULL) return;
   if (dim_ == 0) return;
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) { 
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -130,6 +130,25 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
 
 
 template<typename T>
+void CuArray<T>::CopyToVec(T *dst) const {
+  if (dst == NULL) {
+    dst = (T *)malloc(dim_);
+  }
+  if (dim_ == 0) return;
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) { 
+    Timer tim;
+    CU_SAFE_CALL(cudaMemcpy(dst, Data(), dim_*sizeof(T), cudaMemcpyDeviceToHost));
+    CuDevice::Instantiate().AccuProfile("CuArray::CopyToVecD2H", tim.Elapsed());
+  } else
+#endif
+  {
+    memcpy(dst, data_, dim_*sizeof(T));
+  }
+}
+
+ 
+template<typename T>
 void CuArray<T>::SetZero() {
   if (dim_ == 0) return;
 #if HAVE_CUDA == 1
--- a/src/cudamatrix/cu-array-test.cc
+++ b/src/cudamatrix/cu-array-test.cc
@@ -74,7 +74,7 @@ static void UnitTestCuArray() {
       std::vector<T> vec2;
       cu_vec.CopyToVec(&vec2);
       T *vec22 = new T[vec.size()];
-      cu_vec.CopyToVec(vec22);
+      cu_vec.CopyToHost(vec22);
     }
 
     { // test assignment operator from CuArray.
--- a/src/cudamatrix/cu-array-test.cc
+++ b/src/cudamatrix/cu-array-test.cc
@@ -73,6 +73,8 @@ static void UnitTestCuArray() {
       CuArray<T> cu_vec(vec);
       std::vector<T> vec2;
       cu_vec.CopyToVec(&vec2);
+      T *vec22 = new T[vec.size()];
+      cu_vec.CopyToVec(vec22);
     }
 
     { // test assignment operator from CuArray.
--- a/src/cudamatrix/cu-array-test.cc
+++ b/src/cudamatrix/cu-array-test.cc
@@ -75,6 +75,7 @@ static void UnitTestCuArray() {
       cu_vec.CopyToVec(&vec2);
       T *vec22 = new T[vec.size()];
       cu_vec.CopyToHost(vec22);
+      delete[] vec22;
     }
 
     { // test assignment operator from CuArray.
--- a/src/cudamatrix/cu-array.h
+++ b/src/cudamatrix/cu-array.h
@@ -87,6 +87,7 @@ class CuArray {
   /// is done via memcpy.  So be very careful calling this function if your
   /// objects are more than plain structs.
   void CopyToVec(std::vector<T> *dst) const;
+  void CopyToVec(T *dst) const;
 
   /// Sets the memory for the object to zero, via memset.  You should verify
   /// that this makes sense for type T.
--- a/src/cudamatrix/cu-array.h
+++ b/src/cudamatrix/cu-array.h
@@ -87,7 +87,11 @@ class CuArray {
   /// is done via memcpy.  So be very careful calling this function if your
   /// objects are more than plain structs.
   void CopyToVec(std::vector<T> *dst) const;
-  void CopyToVec(T *dst) const;
+
+  /// Version of the above function that copies contents to a host array.
+  /// This function requires *dst to be allocated before calling. The allocated
+  /// size should be dim_ * sizeof(T)
+  void CopyToHost(T *dst) const;
 
   /// Sets the memory for the object to zero, via memset.  You should verify
   /// that this makes sense for type T.
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -1116,7 +1116,7 @@ template<typename Real>
 __global__
 static void _cuda_matrix_add_indexed_values(MatrixDim dim, Real alpha,
                                             const Int32Pair* indices, const Real* x,
-					    int s, Real* data) {
+                                            int s, Real* data) {
   int i = blockIdx.x * blockDim.x + threadIdx.x;
   if (i >= s)
     return;
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -1114,6 +1114,35 @@ static void _cuda_matrix_add_elements(Real *data, MatrixDim dim, Real alpha, Mat
 
 template<typename Real>
 __global__
+static void _cuda_matrix_add_indexed_values(Real *data, MatrixDim dim, Real alpha, const Int32Pair* indices, Real* x, int s) { 
+  int i = blockIdx.x * blockDim.x + threadIdx.x;
+  //int i = threadIdx.x;
+  if (i >= s)
+    return;
+  int data_i = indices[i].first * dim.stride + indices[i].second;
+  data[data_i] += alpha * x[i];
+
+  //int size = s / CU1DBLOCK; //the least size in a loop (later part)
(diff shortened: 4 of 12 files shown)


danpovey:
I am flying to Baltimore today, I may not have a good connection to review
this.

danpovey:
Hm, I just sent an email but it doesn't seem to have sent properly (bad
connection),
Your implementation of lookup here
template<typename Real>+void CuMatrixBase<Real>::Lookup(const
CuArray<Int32Pair> &indices,+ Real *output) const {
does not look right because you need to back off to non-CUDA if CUDA is not
compiled in or is not available.

The way you malloc'ed the array in another Lookup fiunction was not right
because (a) it should be new[], not malloc, and (b) output was passed by
value so the malloc'ed array would simply be lost.  My interface implies
that 'output' already is allocated and has the correct size (of course, you
should document that).
Dan

naxingyu:
There are two options for the new AddElements.
1) Assign indexes and values to MatrixElement array and use current cuda kernel (_cuda_matrix_add_elements);
2) Create a new cuda kernel accepting indexes and values (_cuda_matrix_add_indexed_values).
Which do you prefer?

danpovey:
Please write a new kernel.  I want this to be efficient.
Dan

naxingyu:
It compiles now. Unit tests passed.

src/cudamatrix/cu-array.h
```diff
@@ -87,6 +87,7 @@ class CuArray {
   /// is done via memcpy.  So be very careful calling this function if your
   /// objects are more than plain structs.
   void CopyToVec(std::vector<T> *dst) const;
+  void CopyToVec(T *dst) const;
```
danpovey: Please add a comment, e.g.
/// Version of the above that copies contents to a host array.
/// 'dst' must already be correctly sized.

src/cudamatrix/cu-array-inl.h
```diff
@@ -130,6 +130,23 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
...
+  if (dim_ == 0) return;
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) { 
+    Timer tim;
+    CU_SAFE_CALL(cudaMemcpy(dst, Data(), dim_*sizeof(T), cudaMemcpyDeviceToHost));
+    CuDevice::Instantiate().AccuProfile("CuArray::CopyToVecD2H", tim.Elapsed());
+  } else
+#endif
+  {
+    memcpy(dst, data_, dim_*sizeof(T));
```
danpovey: space around operator *

src/cudamatrix/cu-kernels.cu
```diff
@@ -1114,6 +1114,35 @@ static void _cuda_matrix_add_elements(Real *data, MatrixDim dim, Real alpha, Mat
 
 template<typename Real>
 __global__
+static void _cuda_matrix_add_indexed_values(Real *data, MatrixDim dim, Real alpha, const Int32Pair* indices, Real* x, int s) { 
+  int i = blockIdx.x * blockDim.x + threadIdx.x;
```
danpovey: You are not changing 'x' so x should be const.
It's preferred even in CUDA code that non-const arguments should be at the end, so unless there is a good reason, 'data' should be the last argument.
Some of the old code may not have been 100% reviewed and may not be fully compliant with these rules.  You can fix the old code too while you're about it, if you have time.

src/cudamatrix/cu-kernels.cu
```diff
@@ -2399,6 +2428,10 @@ void cudaF_matrix_add_elements(dim3 Gr, dim3 Bl, float *data, MatrixDim dim, flo
   _cuda_matrix_add_elements<<<Gr, Bl>>>(data, dim, alpha, x, s); 
 }
 
+void cudaF_matrix_add_indexed_values(dim3 Gr, dim3 Bl, float *data, MatrixDim dim, float alpha, const Int32Pair* indices, float* x, int s) {
```
danpovey: x should be const.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2528,8 +2528,45 @@ void CuMatrixBase<Real>::AddElements(Real alpha,
 }
 
 template<typename Real>
+void CuMatrixBase<Real>::AddElements(Real alpha, const CuArray<Int32Pair> &indexes,
+				     const Real *input) {
+  KALDI_ASSERT(input != NULL);
```
danpovey: Before checking that input != NULL, you might want to do
if (indexes.Dim() == 0)
  return;
We generally prefer all these routines to just return and do nothing for empty input, and having a NULL pointer would be fine if indexes.size() were empty.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2528,8 +2528,45 @@ void CuMatrixBase<Real>::AddElements(Real alpha,
...
+  // Checks the dimension.
+  MatrixIndexT num_rows = this->num_rows_, num_cols = this->num_cols_;
+  std::vector<Int32Pair> cpu_indexes;
+  indexes.CopyToVec(&cpu_indexes);
+  for (int32 i = 0; i < indexes.Dim(); ++i) {
```
danpovey: Please don't copy the indexes to CPU just for the sake of a check.  The whole point of giving a CuArray as argument to the function was to avoid this type of operation.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2528,8 +2528,45 @@ void CuMatrixBase<Real>::AddElements(Real alpha,
...
+    KALDI_ASSERT(cpu_indexes[i].first < num_rows && cpu_indexes[i].first >= 0 &&
+                 cpu_indexes[i].second < num_cols && cpu_indexes[i].second >= 0);
+  }
+
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) {
+    void *addr = CuDevice::Instantiate().Malloc(indexes.Dim() * sizeof(Real));
```
danpovey: 'addr' is not a great name, something like cu_array might be more suitable.  However, in situations like these I prefer you to do something like
  Vector<Real> tmp_vec(indexes.Dim(), kUndefined);
  and then use tmp_vec.Data().   This automatically takes care of error handling and deallocation.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2540,33 +2577,54 @@ void CuMatrixBase<Real>::Lookup(const std::vector<Int32Pair> &indices,
   // Checks the pointer.
   KALDI_ASSERT(output != NULL);
 
-  // Resizes the output vector.
-  output->resize(indices.size());
+#if HAVE_CUDA == 1
```
danpovey: This block inside the #if is wasteful in the case where CUDA is compiled in but not being used (e.g. no GPU available or not configured to use GPU), because it involves wasteful copying.  Inside the #if...#endif block you should have
} else
and then after the #endif, the next block can be inside braces {..}.  Make sure you understand why this works.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2540,33 +2577,54 @@ void CuMatrixBase<Real>::Lookup(const std::vector<Int32Pair> &indices,
...
+    output[i] = (*this)(indices[i].first, indices[i].second);
+  }
+#endif
+}
+
+template<typename Real>
+void CuMatrixBase<Real>::Lookup(const CuArray<Int32Pair> &indices,
+				Real *output) const {
+  int32 num_elements = indices.Dim();
+  // Checks the dimension.
```
danpovey: Please remove the checks that involve copying the whole vector to CPU.  On the GPU we typically just don't check things like this.

src/cudamatrix/cu-matrix.h
```diff
@@ -513,11 +513,22 @@ class CuMatrixBase {
...
-  // This function resizes the output to indexes.size(), and for each element of
-  // "indexes" it interprets it as a (row, column) index into *this, and puts
-  // (*this)(row, column) into the corresponding element of "output".
+  // For each i, with indexes[i] = (j, k), does (*this)(j, k) += input[i].
+  // Requires, but does not check, that the vector of indexes does not contrain
+  // repeated elements, 'input' is the start of an array of length equal
+  // to indexes.Dim().
+  void AddElements(Real alpha, const CuArray<Int32Pair> &indexes, const Real *input);
+
+  // This function implies the output is allocated and the size is equal to
```
danpovey: changes 'implies' to 'requires'.  References to argument names should be inside single quotes (this is a new rule), so e.g. output should be 'output'.  Please make very clear that 'output' is a host vector (i.e. on the CPU).

src/cudamatrix/cu-array-inl.h
```diff
@@ -130,6 +130,23 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
 
 
 template<typename T>
+void CuArray<T>::CopyToVec(T *dst) const {
```
danpovey: I just realized that CopyToVec is not a very good name because 'dst' is not a Vector, and we need to clarify that 'dst' is on the CPU.  Please change this name to CopyToHost.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2540,33 +2577,54 @@ void CuMatrixBase<Real>::Lookup(const std::vector<Int32Pair> &indices,
...
-    cuda_output.CopyToVec(output);
     
+    cuda_output.CopyToVec(output);
     CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
   } else
 #endif
   {
-    for (int32 i = 0; i < indices.size(); i++) {
-      (*output)[i] = (*this)(indices[i].first, indices[i].second);
+    for (int32 i = 0; i < num_elements; i++) {
```
danpovey: You don't need to copy to cpu_indices in order to access the indices on CPU, if CUDA is not compiled in or is not being used.  In that case, the CuArray Data() pointer will point to a CPU pointer, which you can index directly.

src/nnet2/train-nnet-ensemble.cc
```diff
@@ -81,18 +82,24 @@ void NnetEnsembleTrainer::TrainOneMinibatch() {
...
   for (int32 i = 0; i < nnet_ensemble_.size(); i++) {  
     CuMatrix<BaseFloat> tmp_deriv(post_mat[i]);
     post_mat[i].ApplyLog();
     std::vector<BaseFloat> log_post_correct;
-    post_mat[i].Lookup(sv_labels_ind, &log_post_correct);
+    BaseFloat *p_log_post_correct = new BaseFloat[sv_labels_ind.size()];
+    post_mat[i].Lookup(sv_labels_ind, p_log_post_correct);
+    for (int32 j = 0; j < sv_labels_ind.size(); j++)
+      log_post_correct.push_back(p_log_post_correct[j]);
+    delete[] p_log_post_correct;
```
danpovey: There was no need to change this code because the Lookup function that takes a std::vector<BaseFloat> still exists.  Also, if you wanted to use the one that takes the BaseFloat* pointer it would have been much more efficient to correctly size log_post_correct first, and then give the function the pointer &(log_post_correct[0]).

naxingyu:
Updated. I plan to open another PR for the "non-const position" issue, with a thorough check, including other issues, such as 'indexes' and 'indices'.

src/cudamatrix/cu-array-inl.h
```diff
@@ -119,17 +119,34 @@ void CuArray<T>::CopyToVec(std::vector<T> *dst) const {
...
   {
-    memcpy(&dst->front(), data_, dim_*sizeof(T));
+    memcpy(&dst->front(), data_, dim_ * sizeof(T));
   }
 }
 
 
 template<typename T>
+void CuArray<T>::CopyToHost(T *dst) const {
+  if (dst == NULL) return;
```
danpovey: You should be asserting that dst != NULL instead of just returning if it's NULL -- but do this after the statement
if (dim_ == 0) return;

src/cudamatrix/cu-matrix.cc
```diff
@@ -2528,8 +2528,40 @@ void CuMatrixBase<Real>::AddElements(Real alpha,
...
+  if (indexes.Dim() == 0) return;
+  KALDI_ASSERT(input != NULL);
+  
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) {
+    CuVector<Real> tmp_vec(indexes.Dim(), kUndefined);
+    CU_SAFE_CALL(cudaMemcpy(tmp_vec.Data(), input, indexes.Dim() * sizeof(Real),
+                            cudaMemcpyHostToDevice));
+
+    Timer tim;
```
danpovey: Please move this line "Timer tim;" to the top of the code block; it should measure everything we do in the function.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2540,33 +2572,53 @@ void CuMatrixBase<Real>::Lookup(const std::vector<Int32Pair> &indices,
   // Checks the pointer.
   KALDI_ASSERT(output != NULL);
```
danpovey: Please write
if (indices.size() == 0) return;
before this line.  output may be NULL if indices.size() is zero.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2540,33 +2572,53 @@ void CuMatrixBase<Real>::Lookup(const std::vector<Int32Pair> &indices,
...
+    }
+  }
+}
+
+template<typename Real>
+void CuMatrixBase<Real>::Lookup(const CuArray<Int32Pair> &indices,
+				Real *output) const {
+  int32 num_elements = indices.Dim();
+  if (num_elements == 0) return;
+  // Checks the pointer.
```
danpovey: remove redundant comment

src/cudamatrix/cu-matrix.cc
```diff
@@ -2540,33 +2572,53 @@ void CuMatrixBase<Real>::Lookup(const std::vector<Int32Pair> &indices,
...
+    cuda_output.CopyToHost(output);
     CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
   } else
 #endif
   {
-    for (int32 i = 0; i < indices.size(); i++) {
-      (*output)[i] = (*this)(indices[i].first, indices[i].second);
+    MatrixIndexT num_rows = this->num_rows_, num_cols = this->num_cols_;
+    for (int32 i = 0; i < num_elements; i++) {
+      Int32Pair index = indices.Data()[i];
```
danpovey: Here and in similar code elsewhere, it might be best to avoid the function call to Data() each time, by declaring a pointer outside the loop.

danpovey:
Thanks.
I pointed out a few small issues- after you fix those I may merge.
Dan

naxingyu:
Thanks. The fixes are commited.

src/cudamatrix/cu-array-test.cc
```diff
@@ -73,6 +73,8 @@ static void UnitTestCuArray() {
       CuArray<T> cu_vec(vec);
       std::vector<T> vec2;
       cu_vec.CopyToVec(&vec2);
+      T *vec22 = new T[vec.size()];
+      cu_vec.CopyToHost(vec22);
     }
```
danpovey: Please fix memory leak in test code.  (we might run valgrind on it at some point to check for memory leaks).

src/cudamatrix/cu-kernels.cu
```diff
@@ -1114,6 +1114,19 @@ static void _cuda_matrix_add_elements(Real *data, MatrixDim dim, Real alpha, Mat
 
 template<typename Real>
 __global__
+static void _cuda_matrix_add_indexed_values(MatrixDim dim, Real alpha,
+                                            const Int32Pair* indices, const Real* x,
+					    int s, Real* data) {
```
danpovey: indent this line one more char

naxingyu:
Done. I also checked the indentation in other files, replaced all the tabs inserted by emacs.

danpovey:
OK, I'll have a look.  In my .emacs I have
(setq-default tab-width 4)
(setq-default fill-column 80)
(setq-default indent-tabs-mode `nil)
as well as loading the standard google C++ style file like this:
(load-file "~/.google-c-style.el")
(add-hook 'c-mode-common-hook 'google-set-c-style)
(add-hook 'c-mode-common-hook 'google-make-newline-indent)

naxingyu:
Thanks for that!

在 2015/8/30 8:05, Daniel Povey 写道:

danpovey:
No, thanks for the code.
If you have time you could add the nnet3-copy command (there is an issue
about this).  It shouldn't take very long.
There are other TODOs for nnet3 there also, but they are mostly harder.
Dan

naxingyu:
Yes, I saw that. I'll have a look.

在 2015/8/30 8:19, Daniel Povey 写道:

naxingyu:
There are 'indexes' and 'indices' in many code, including in this one. What I could find in literature is that ''indices is generally preferred in mathematical, financial, and technical contexts, while indexes is relatively common in general usage." (http://grammarist.com/usage/indexes-indices/). Maybe it's not a big deal.... It's not affecting anything. Should it be taken care of?

danpovey:
We wouldn't want to change any code for this reason.
Dan

danpovey:
Actually, I would consider a pull request that changes variables names, but from 'indices' to 'indexes'.  It could be done by a script.  'indexes' is becoming more common, and is less confusing to non-native speakers of English.  In a paper I would use 'indices', but in code it seems pretentious to me.

Dan

PR #99, opened 2015-08-27
Update configure file to use MKL and TBB

kronos-cm:
Changes made in the src/configure file so that it supports the newest MKL version 11.3 with TBB as thread support instead of openMP

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -211,16 +211,18 @@ function linux_configure_mkl_libraries {
     for file in $threaded_libs; do
       local libfile=$mkllibdir/lib$file.so
       check_exists $libfile
-    done 
-    #echo  "-L$mkllibdir -Wl,-rpath=$mkllibdir -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lm"
-    linkline="-L$mkllibdir -Wl,-rpath=$mkllibdir  -lmkl_intel_lp64 -lmkl_core -lmkl_tbb_thread -ltbb -lstdc++ -lpthr$
+    done
+    if [ ! -f $mkllibdir/libmkl_tbb_thread.* ]; then
+      linkline "-L$mkllibdir -Wl,-rpath=$mkllibdir -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lm"
+    else
+      linkline="-L$mkllibdir -Wl,-rpath=$mkllibdir  -lmkl_intel_lp64 -lmkl_core -lmkl_tbb_thread -ltbb -lstdc++ -lpthread -lm"
+    fi
     echo $linkline
   elif  ! $static && ! $threaded ; then
     for file in $sequential_libs; do
       local libfile=$mkllibdir/lib$file.so
       check_exists $libfile
-    done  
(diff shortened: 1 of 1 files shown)


jtrmal:
Thanks.  The changes you've made, however, mean that the configure won't generate the correct linking line for earlier MKL, right?
Also I see the travis build failed -- I will have to look at the reason why.

src/configure
```diff
@@ -212,13 +212,17 @@ function linux_configure_mkl_libraries {
       local libfile=$mkllibdir/lib$file.so
       check_exists $libfile
     done 
-    echo  "-L$mkllibdir -Wl,-rpath=$mkllibdir -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lm"
+    #echo  "-L$mkllibdir -Wl,-rpath=$mkllibdir -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lm"
+    linkline="-L$mkllibdir -Wl,-rpath=$mkllibdir  -lmkl_intel_lp64 -lmkl_core -lmkl_tbb_thread -ltbb -lstdc++ -lpthr$
```
kkm000: Looks like the line was cut and the end mangled. It fails but way below because of the quoting: https://travis-ci.org/kaldi-asr/kaldi#L1370

kkm000:
@jtrmal: correct, the version I have (11.2.2) does not have `libmkl_tbb_thread`. And I do not even know where to get the new version. Mine came bunded with Compozer XE 2015 Update 2, and the latest is the Update 3 that comes with MKL 11.2.3. @kronos-cm, how to get this version? Is it a beta?

src/configure
```diff
@@ -279,7 +283,7 @@ function linux_configure_omplibdir {
   for libext in $libexts; do
     echo "Testing $libext from [$libexts] " >&2
 
-    check_library $mkllibdir "libiomp5" $libext \
+    check_library $mkllibdir "libiomp5" $libext || check_library $mkllibdir "libmkl_tbb_thread" $libext  \
```
jtrmal: Even this looks suspicious (because of the order of || and && -- I think the mixing of the operators is generally discouraged by  bash docs)

jtrmal:
Ok, this patch will obviously need much more work. @kronos-cm when you let us know how to get to the  11.3 MKL, I will have a look at it.

kronos-cm:
Hi, yes I will correct it so that it will either accept libiomp5 or tbb. I will also think on a better way to avoid the mixing operators. Maybe a switch?

kronos-cm:
@kkm000, I got an email that the new version was available. I downloaded it from https://registrationcenter.intel.com/en/products/
together with MPI and TBB, and no, even if the linelink advisor still says is a beta version, it is since Tuesday an official release.

kkm000:
I figured the new MKL comes with the new Parallels studio just released. I am downloading it at the moment.

src/configure
kronos-cm: Thanks fro noticing

jtrmal: BTW, looking at the link line advisor, the 11.3 still does support intel
OMP (iomp)...
y.

kronos-cm: You are right, but it does not install libiomp5 by default on mkl. At least it didn't in my case

jtrmal:
Closing this, as the other PR provides more general functionality.

PR #101, opened 2015-08-28
Mkl configure fix

jtrmal:
@kronos-cm Can you please check if this does the right thing for you?
You can do something like
curl https://github.com/kaldi-asr/kaldi/pull/#PR.diff | patch -N 0

Then add "--mkl-threading=tbb" to your configure command line

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -13,6 +13,8 @@
 # ./configure --shared  ## shared libraries.
 # ./configure --mkl-root=/opt/intel/mkl
 # ./configure --mkl-root=/opt/intel/mkl --threaded-math=yes
+# ./configure --mkl-root=/opt/intel/mkl --threaded-math=yes --mkl-threading=tbb
+#        This is for MKL 11.3 -- which does not seem  to provide Intel OMP libs
 # ./configure  --openblas-root=../tools/OpenBLAS/install  # before doing
 #        # this, cd to ../tools and type "make openblas".  Note:
 #        # this is not working correctly on all platforms, do "make test"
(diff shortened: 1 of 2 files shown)


kronos-cm:
Hi @jtrmal it works like a charm, thanks 
Only warning was, albeit unrelated was that the num-splice variable might be unused in the nnet-convnet bin

kkm000:
I, Like Yenda, thought Intel installation sets up short name links . I have (XE 2015 update 2)
```
kkm@yupana:/opt/intel$ realpath /opt/intel/mkl/lib
/opt/intel/composer_xe_2015.2.164/mkl/lib
```
I have not installed the 2016 yet. Will see.

Let me pull this change and see how it works with my version.

kkm000:
LGTM. Configured with 

`./configure --mathlib=MKL --use-cuda=yes --mkl-root=/opt/intel/mkl`

Still compiling, but I am seeing. `-L/opt/intel/mkl/lib/intel64 -Wl,-rpath=/opt/intel/mkl/lib/intel64 -lmkl_intel_lp64  -lmkl_core  -lmkl_sequential` in compiler command lines which looks right.

kkm000:
Sorry, did not realize you are talking about libomp5.It is not part of MKL though, and rather that of the compiler runtime (analog of libc). 

You need to source

`. /opt/intel/bin/iccvars.sh`

or for a specific version 

`. /opt/intel/composer_xe_2015.2.164/bin/iccvars.sh`

jtrmal:
Merging this -- I don't want to go into requiring the whole compiler support. Not sure how is it now, but it used to be possible to install only MKL. There is still a switch --omp-libdir that should allow to include additional library directory.

PR #102, opened 2015-08-29
Test

keli78:
Remove unnecessary null pointers in functions: DeleteTestFst, Fbank, and LangModelFst.

### DIFF
--- a/src/feat/feature-fbank.cc
+++ b/src/feat/feature-fbank.cc
@@ -43,7 +43,6 @@ Fbank::~Fbank() {
       iter != mel_banks_.end();
       ++iter)
     delete iter->second;
-  if (srfft_ != NULL)
     delete srfft_;
 }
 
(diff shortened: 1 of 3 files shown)


danpovey:
You need to clean up the commit history a bit.  Vimal or Vijay should be able to help, but others may be good at git too.  I recommend to either rebase, or apply a patch to a freshly created branch.

keli78:
I cleaned up the commit history.

danpovey:
OK, thanks, will merge.

danpovey:
BTW, you could look through kaldi for similar unnecessary checks prior to
deletes.

The following shell script will identify candidates for this:

for x in */*.{h,cc}; do perl -e '$ret = 1; while(<>) { if
(($last_line=~m/if/ || m/if/) && m/delete/) { print "$last_line $_"; $ret =
0; } $last_line = $_; } exit $ret; ' $x && echo $x; done

Get someone to explain to you the script if anyone is around.

Dan

PR #103, opened 2015-08-29
Added GradientClipperComponent for gradient control 

vijayaditya:
Added GradientClipperComponent for gradient explosion control in recurrent  networks.  Modified LSTM recipe to use the new component.

Recipe has not been tested and still needs refinement. This is an intermediate commit, for review.


egs/ami/s5/run_sdm_lstm.sh
```diff
@@ -168,13 +168,26 @@ fi
...
   local/nnet3/run_lstm.sh \
     --mic $mic \
-    --train-stage -5 \
+    --train-stage -1 \
     --speed-perturb true \
     --stage 7 \
-    --common-egs-dir exp/sdm1/nnet3/lstm_sp/egs \
+    --affix lc22 \
+    --norm-based-clipping false \
+    --clipping-threshold 20\
```
danpovey: don't you mean $clipping_threshold instead of 20?

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -147,7 +149,7 @@ if [ $# != 4 ]; then
   
   exit 1;
 fi
-
+set -x
```
danpovey: Why are you doing this set -x?  My understanding is that it exports bash variables without using the 'export' command.  I don't think silently exporting variables is a good idea.

vijayaditya: This is for debugging

"set -x: set this flag to trace the execution of all commands. This will make the interpreter print each command right before it is executed, so it will aid you in knowing what is happening at any point in time."

This will be removed when the recipe is finalized.

egs/ami/s5/run_sdm_lstm.sh
vijayaditya: this is the top level (run.sh level) file  so I don't take arguments from command line and directly set them in the script.

src/nnet3/nnet-simple-component.h
```diff
@@ -666,6 +666,55 @@ class NoOpComponent: public NonlinearComponent {
...
+class GradientClipperComponent: public Component {
+ public:
+  GradientClipperComponent(int32 dim, BaseFloat clipping_threshold, bool norm_based_clipping) {
+    Init(dim, clipping_threshold, norm_based_clipping);
+  }
+  GradientClipperComponent(): dim_(0), clipping_threshold_(-1), norm_based_clipping_(false) { }
+  void Init(int32 dim, BaseFloat clipping_threshold, bool norm_based_clipping);
+  virtual std::string Type() const { return "GradientClipperComponent"; }
+  virtual int32 Properties() const {
+    return kSimpleComponent|kLinearInInput|kPropagateInPlace;
```
danpovey: Your backprop code will work fine in-place, so there is no reason not to add |kBackpropInPlace.

src/nnet3/nnet-simple-component.cc
```diff
@@ -295,6 +295,96 @@ void NoOpComponent::Backprop(const std::string &debug_info,
...
+  if (clipping_threshold_ > 0) {
+    if (norm_based_clipping_) {
+      BaseFloat deriv_norm = in_deriv->FrobeniusNorm();
+      in_deriv->Scale(clipping_threshold_ / deriv_norm);
+    } else {
+      in_deriv->ApplyCeiling(clipping_threshold_);
+      in_deriv->ApplyFloor(-1 * clipping_threshold_);
+    }
+    if (GetVerboseLevel() >= 2) {
+      BaseFloat norm_before_clipping = out_deriv.FrobeniusNorm(),
```
danpovey: this code would have to be changed if you allow backprop in place (but anyway you are redundantly computing the FrobeniusNorma()).

danpovey:
is this ready for review?

vijayaditya:
I still have to add stats storing code for the GradientClipper component.

Vijay

src/nnet3/nnet-simple-component.cc
```diff
@@ -295,6 +295,110 @@ void NoOpComponent::Backprop(const std::string &debug_info,
...
+      // each row in the derivative matrix, which corresponds to one sample in
+      // the mini-batch, is scaled to have a max-norm of clipping_threshold_
+      BaseFloat clipping_threshold_2 = pow(clipping_threshold_, 2);
+      CuVector<BaseFloat> deriv_norm(in_deriv->NumRows());
+      deriv_norm.AddDiagMat2(1.0 / in_deriv->NumCols(),
+                             *in_deriv, kNoTrans, 0.0);
+      deriv_norm.ApplyFloor(clipping_threshold_2);
+      deriv_norm.Scale(1.0 / clipping_threshold_2);
+      deriv_norm.ApplyPow(-0.5);
+      in_deriv->MulRowsVec(deriv_norm);
```
danpovey: If you make this a bit more like the code I sketched by email, here, it will be a bit more efficient:
   CuVector<BaseFloat> clipping_scales(in_deriv->NumRows());
    clipping_scales.AddDiagMat2(pow(clipping_threshold_, -2), *in_deriv, kNoTrans, 0.0);
    // now clipping_scales contains the squared (norm of each row divided by clipping_threshold)
    int32 num_not_scaled = clipping_scales.ApplyFloor(1.0); 
    // now clipping_scales contains min(1, squared-(norm/clipping_threshold))
    if (num_not_scaled != clipping_scales.Dim()) {
      clipping_scales.ApplyPow(-0.5); 
      // now clipping_scales contains max(1, clipping_threshold/vector_norm)
      in_deriv->MulRowsVec(clipping_scales);
    }
I notice you use a scale of 1.0 / in_deriv->NumCols().  This way the clipping threshold corresponds not to a vector norm but to a vector norm per element.  I think using a vector norm might make more sense (i.e. skipping the 1.0/in_deriv->NumCols() factor), because for higher-dimensional features, the derivative per element might be smaller because each element matters less.  Also, using an overall vector norm is just simpler to explain.
Also, you can use the num_not_scaled value to keep track of what proportion of vectors were scaled, for stats.  You just need to store two scalars.  Please implement ZeroStats() to set them to zero.  I'll be changing the documentation for ZeroStats() to indicate that it's not just for components that implement StoreStats().

vijayaditya: On Mon, Aug 31, 2015 at 4:51 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/nnet3/nnet-simple-component.cc
> <https://github.com/kaldi-asr/kaldi/pull/103#discussion_r38357863>:
>
> > +
> > +  if (GetVerboseLevel() >= 2) {
> > +    norm_before_clipping = out_deriv.FrobeniusNorm();
> > +  }
> > +  if (clipping_threshold_ > 0) {
> > +    if (norm_based_clipping_) {
> > +      // each row in the derivative matrix, which corresponds to one sample in
> > +      // the mini-batch, is scaled to have a max-norm of clipping_threshold_
> > +      BaseFloat clipping_threshold_2 = pow(clipping_threshold_, 2);
> > +      CuVector<BaseFloat> deriv_norm(in_deriv->NumRows());
> > +      deriv_norm.AddDiagMat2(1.0 / in_deriv->NumCols(),
> > +                             *in_deriv, kNoTrans, 0.0);
> > +      deriv_norm.ApplyFloor(clipping_threshold_2);
> > +      deriv_norm.Scale(1.0 / clipping_threshold_2);
> > +      deriv_norm.ApplyPow(-0.5);
> > +      in_deriv->MulRowsVec(deriv_norm);
>
> If you make this a bit more like the code I sketched by email, here, it
> will be a bit more efficient:
> CuVector clipping_scales(in_deriv->NumRows());
> clipping_scales.AddDiagMat2(pow(clipping_threshold_, -2), *in_deriv,
> kNoTrans, 0.0);
> // now clipping_scales contains the squared (norm of each row divided by
> clipping_threshold)
> int32 num_not_scaled = clipping_scales.ApplyFloor(1.0);
> // now clipping_scales contains min(1, squared-(norm/clipping_threshold))
> if (num_not_scaled != clipping_scales.Dim()) {
> clipping_scales.ApplyPow(-0.5);
> // now clipping_scales contains max(1, clipping_threshold/vector_norm)
> in_deriv->MulRowsVec(clipping_scales);
> }
>
OK, modifying.


> I notice you use a scale of 1.0 / in_deriv->NumCols(). This way the
> clipping threshold corresponds not to a vector norm but to a vector norm
> per element. I think using a vector norm might make more sense (i.e.
> skipping the 1.0/in_deriv->NumCols() factor), because for
> higher-dimensional features, the derivative per element might be smaller
> because each element matters less. Also, using an overall vector norm is
> just simpler to explain.
>
OK.



> Also, you can use the num_not_scaled value to keep track of what
> proportion of vectors were scaled, for stats. You just need to store two
> scalars. Please implement ZeroStats() to set them to zero. I'll be changing
> the documentation for ZeroStats() to indicate that it's not just for
> components that implement StoreStats().
>
OK, I thought ZeroStats() call would not be made for normal components and
converted GradientClipper to a NonlinearComponent (in local repo). Do I
have to handle printing of these stats (using KALDI_VLOG statements) or do
is there a common interface I can use to get this accomplished ?

danpovey: I recommend that you put something in the Info() call regarding printing the stats, so you can get it by running nnet3-info.  This will get printed in the "progress" logs.
You could also modify the nnet-training code so that instead of 
  if (config.store_component_stats && config.zero_component_stats)
    ZeroComponentStats(nnet);

it says
  if (config.zero_component_stats)
    ZeroComponentStats(nnet);
so the stats will get zeroed before each iteration of training.
It would also make sense to change the way the Scale() and Add() functions are declared, by moving them to the Component base-class and giving them, there, a default implementation that does nothing.  This will allow you to simplify the calling code for Scale() and Add().  You can modify your GradientClipperComponent so that it overrides these functions, and uses them to scale/add the stats.

Also, please rename from GradientClipperComponent to ClipGradientComponent: the other components have "imperative" names like SumGroupComponent or NormalizeComponent, I'd rather be consistent with them.  Sorry for the inconvenience.


Dan

vijayaditya: On Mon, Aug 31, 2015 at 5:20 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/nnet3/nnet-simple-component.cc
> <https://github.com/kaldi-asr/kaldi/pull/103#discussion_r38361083>:
>
> > +
> > +  if (GetVerboseLevel() >= 2) {
> > +    norm_before_clipping = out_deriv.FrobeniusNorm();
> > +  }
> > +  if (clipping_threshold_ > 0) {
> > +    if (norm_based_clipping_) {
> > +      // each row in the derivative matrix, which corresponds to one sample in
> > +      // the mini-batch, is scaled to have a max-norm of clipping_threshold_
> > +      BaseFloat clipping_threshold_2 = pow(clipping_threshold_, 2);
> > +      CuVector<BaseFloat> deriv_norm(in_deriv->NumRows());
> > +      deriv_norm.AddDiagMat2(1.0 / in_deriv->NumCols(),
> > +                             *in_deriv, kNoTrans, 0.0);
> > +      deriv_norm.ApplyFloor(clipping_threshold_2);
> > +      deriv_norm.Scale(1.0 / clipping_threshold_2);
> > +      deriv_norm.ApplyPow(-0.5);
> > +      in_deriv->MulRowsVec(deriv_norm);
>
> I recommend that you put something in the Info() call regarding printing
> the stats, so you can get it by running nnet3-info. This will get printed
> in the "progress" logs.
>
OK.


> You could also modify the nnet-training code so that instead of
> if (config.store_component_stats && config.zero_component_stats)
> ZeroComponentStats(nnet);
>
> it says
> if (config.zero_component_stats)
> ZeroComponentStats(nnet);
> so the stats will get zeroed before each iteration of training.
>
OK.


> It would also make sense to change the way the Scale() and Add() functions
> are declared, by moving them to the Component base-class and giving them,
> there, a default implementation that does nothing. This will allow you to
> simplify the calling code for Scale() and Add(). You can modify your
> GradientClipperComponent so that it overrides these functions, and uses
> them to scale/add the stats.
>
OK.


> Also, please rename from GradientClipperComponent to
> ClipGradientComponent: the other components have "imperative" names like
> SumGroupComponent or NormalizeComponent, I'd rather be consistent with them.
>
OK.



> Sorry for the inconvenience.
>
No problem, this helps me understand the interface better.

vijayaditya: > It would also make sense to change the way the Scale() and Add() functions
 are declared, by moving them to the Component base-class and giving them,
 there, a default implementation that does nothing. This will allow you to
 simplify the calling code for Scale() and Add(). You can modify your
 GradientClipperComponent so that it overrides these functions, and uses
 them to scale/add the stats.

Moving the Add and Scale functions to the Component class would mean that UpdateableComponent would override these methods to perform operations on parameters.
I think it would be better if modify Add() and Scale() methods which modify stats to AddStats() and ScaleStats().

danpovey: That is the point.  They tend to be called only in model-averaging
operations, and you want everything to have it applied..  If someone
decides that they want to only apply those functions for updatable
components, they can just test the Properties() and apply it selectively.
Dan

danpovey: BTW, once you move those functions to the base-class, there is a bunch of
dynamic casting in calling code (with associated sanity checks) which is no
longer necessary and can be removed.
Dan

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -19,8 +19,10 @@ rand_prune=4.0 # Relates to a speedup we do for LDA.
 minibatch_size=512  # This default is suitable for GPU-based training.
                     # Set it to 128 for multi-threaded CPU-based training.
 
-samples_per_iter=400000 # each iteration of training, see this many samples
+frames_per_iter=80000 # each iteration of training, see this many samples
                         # per job.  This option is passed to get_egs.sh
+                        # this is 1/5 of that used for FF-DNN
+iters_per_archive=20  # number of iterations worth samples to store in an ark file
```
vijayaditya: Temporarily we will store egs for multiple iters in the same archive file. This is done to reduce the number of files created by get_egs.sh. get_egs.sh will later be modified to reduce the number of files created. This option will be removed then.

vijayaditya:
Made the suggested modifications.

egs/ami/s5/local/nnet3/run_lstm.sh
```diff
@@ -14,9 +14,29 @@ mic=ihm
...
 
+# lstm options
+splice_indexes="-2,-1,0,1,2"
+num_lstm_layers=1
+cell_dim=1024
+hidden_dim=1024
+recurrent_projection_dim=256
+non_recurrent_projection_dim=256
+truncated_sequence_width=20
+context_sensitive_chunk_width=20
```
danpovey: Wouldn't it be more consistent with the code to call this example_left_context or left_context?

danpovey:
OK, mostly OK, but I have a problem with your script changes relating to
the iters_per_archive.  Firstly, the way you are doing it is terrible for
the disk, as we discussed.  If others start doing it like this it's going
to be bad for our cluster.  Also, now that it turns out we can speed up
training by increasing the minibatch size (or setting
--measure-output-frames=false), we can make the number of archives
smaller.  So I don't think this change would even be needed.  Can you
please try to create a PR that doesn't have this change?  I'm sure using
git branches it's not super hard.

Dan

egs/ami/s5/local/nnet3/run_lstm.sh
vijayaditya: I am borrowing the terminology from Kai Chen, Zhi-Jie Yan, Qiang Huo "Training Deep Bidirectional LSTM Acoustic Model for LVCSR by a Context-Sensitive-Chunk BPTT Approach", Interspeech 2015

When users want to implement/use a feature based on paper, it would be easier to find if we use the same term as the paper to describe the option.

 Also left-context in the context of RNNs would not mean the same as left-context in FF-DNNs so I am bit wary of using the same name.

danpovey:
Ready to merge?

vijayaditya:
I am tuning the samples-per-iter value. Else ready to merge.

Vijay

egs/wsj/s5/steps/nnet3/lstm/make_configs.py
```diff
@@ -305,7 +331,7 @@ def ParseSpliceString(splice_indexes):
     if (num_hidden_layers < args.num_lstm_layers):
         sys.exit("--num-lstm-layers : number of lstm layers has to be greater than number of layers, decided based on splice-indexes")
 
-    left_context = left_context + args.bptt_truncation_width + args.context_sensitive_chunk_width
+    left_context = left_context + args.chunk_left_context
```
danpovey: This doesn't have to delay this merge, but I don't like the fact that chunk_left_context is provided to this script, because it doesn't have to do with the model per se.  The left-context and right-context of the model should be, IMO, just the intrinsic left and right context, and we can then add the chunk_left_context just before we pass the arg to get_egs.sh.  This will be more flexible in case we decide to do things like initially train on a certain (chunk-size, chunk-left-context) and then later in training, increase those values.

vijayaditya: It had to be provided to the make_config.py script as the same script also dumps the variables necessary for get_egs.sh. I will make necessary changes.

danpovey: OK.  I prefer to do the addition left_context = left_context +
args.chunk_left_context
in the lstm training script rather than that python script, to keep that
python script's job simple. Dan

vijayaditya: new make_configs.py just dumps model_left_context/model_right_context which
the minimum necessary context to generate one output label. Rest is done in
the shell script shown below.
I will push this commit when the lstm script is refined in a later pull
request.

<code>
. $dir/configs/vars || exit 1;
left_context=$((chunk_left_context + model_left_context))
right_context=$model_right_context
context_opts="--left-context=$left_context --right-context=$right_context"

! [ "$num_hidden_layers" -gt 0 ] && echo \
 "$0: Expected num_hidden_layers to be defined" && exit 1;

[ -z "$transform_dir" ] && transform_dir=$alidir
</code>

danpovey: ok thanks

PR #105, opened 2015-08-30
add nnet3-copy binary

naxingyu:
nnet3-copy added.

### DIFF
--- a/src/nnet3bin/Makefile
+++ b/src/nnet3bin/Makefile
@@ -10,7 +10,7 @@ BINFILES = nnet3-init nnet3-info nnet3-get-egs nnet3-copy-egs nnet3-subset-egs \
    nnet3-shuffle-egs nnet3-acc-lda-stats nnet3-merge-egs \
    nnet3-compute-from-egs nnet3-train nnet3-am-init nnet3-am-train-transitions \
    nnet3-am-adjust-priors nnet3-am-copy nnet3-compute-prob \
-   nnet3-average nnet3-am-info nnet3-combine nnet3-latgen-faster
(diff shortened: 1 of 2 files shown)


src/nnet3bin/Makefile
```diff
@@ -10,7 +10,7 @@ BINFILES = nnet3-init nnet3-info nnet3-get-egs nnet3-copy-egs nnet3-subset-egs \
    nnet3-shuffle-egs nnet3-acc-lda-stats nnet3-merge-egs \
    nnet3-compute-from-egs nnet3-train nnet3-am-init nnet3-am-train-transitions \
    nnet3-am-adjust-priors nnet3-am-copy nnet3-compute-prob \
-   nnet3-average nnet3-am-info nnet3-combine nnet3-latgen-faster
+   nnet3-average nnet3-am-info nnet3-combine nnet3-latgen-faster nnet3-copy
```
danpovey: We try to keep lines in the Makefile to 80 characters, you might want a line break here (with \)

src/nnet3bin/nnet3-copy.cc
```diff
@@ -0,0 +1,76 @@
+// nnet3bin/nnet3-copy.cc
+
+// Copyright 2012  Johns Hopkins University (author:  Daniel Povey)
```
danpovey: Please add something here like
//                            2015   Xingyu Na
after the first copyright line, to reflect that you worked on it.  (I imagine you are a student; this is normally what students should do if they are not doing this as part of some kind of employment).

naxingyu:
Done.

danpovey:
Thanks!
Merged.
Dan

PR #107, opened 2015-08-31
add nnet3-show-progress

naxingyu:


danpovey:
Thanks!
Merging.  
Next could you please try to add the nnet3-show-progress commands to the existing nnet3 training scripts?   Just do you best guess at them, you don't have to test it (note: you'll have to use nnet3-am-copy with --raw=true to convert to the raw-nnet form that this program requires).

naxingyu:
Yes. I'll have a look.

PR #108, opened 2015-08-31
update nnet3 training scripts for showing progress

naxingyu:
The mix_up condition is removed because I see mix_up is removed from the nnet3 scripts.


danpovey:
Looks good- thanks!
@vijayaditya, I'll merge and leave testing to you since you are working with these scripts.

vijayaditya:
OK. will test them.

PR #111, opened 2015-09-02
Fixing the segfault issue in lattice-oracle.cc

chenguoguo:
Used lattice composition instead of calling the function GetOracleLattice(); also fixed various style issues that I noticed.

@danijel3 perhaps you could have a try to see if this fixes your issue?


danpovey:
Have you done any testing to check it gives the same results as before?

chenguoguo:
Yes I tested on some sample lattice and the output looks correct. On that particular lattice I tested it wasn't giving the same output, but it was because the old methods was problematic -- it allowed a redundant path (with epsilon arc at the end of it), the best path was the same.

danpovey:
OK, merged.

PR #112, opened 2015-09-02
An attempt for updating to support cublas_v2

freewym:
Hi Dan,

Please take a look.

Yiming


src/cudamatrix/cu-vector.cc
```diff
@@ -42,7 +42,8 @@ namespace kaldi {
 
 
 template<typename Real>
-Real VecVec(const CuVectorBase<Real> &a,
+Real VecVec(cublasHandle_t handle,
```
danpovey: VecVec, like all "user-level functions", should not take the Handle argument.  Get it from the CuDevice.

src/cudamatrix/cu-vector.cc
```diff
@@ -51,9 +52,11 @@ Real VecVec(const CuVectorBase<Real> &a,
   if (CuDevice::Instantiate().Enabled()) {    
     Timer tim;
 
-    result = cublas_dot(a.Dim(), a.Data(), 1, b.Data(), 1);
+    cublasStatus_t stat;
+    stat = cublas_dot(handle, a.Dim(), a.Data(), 1, b.Data(), 1, &result);
 
-    CU_SAFE_CALL(cublasGetError());
+    //CU_SAFE_CALL(cublasGetError());
+    CU_SAFE_CALL(stat);
```
danpovey: CU_SAFE_CALL is not really designed to be called on a variable.  Instead, wrap the function call itself in CU_SAFE_CALL:
CU_SAFE_CALL(cublas_dot(handle, a.Dim(), a.Data(), 1, b.Data(), 1, &result));
Also, please remove rather than comment out deleted lines.

src/cudamatrix/cu-device.h
```diff
@@ -100,14 +103,15 @@ class CuDevice {
...
+
  private:
   CuDevice();
   CuDevice(CuDevice&); // Disallow.
   CuDevice &operator=(CuDevice&);  // Disallow.
 
+
   static CuDevice global_device_;
-  
+  static cublasHandle_t handle_;
```
danpovey: This should probably not be a static variable-- in a situation in future where we had multiple CuDevice objects it would make sense for each to have its own handle.  You can make it a regular member and acccess it as global_device_.handle_.

src/cudamatrix/cu-matrix.cc
```diff
@@ -1031,13 +1033,15 @@ void CuMatrixBase<Real>::SymAddMat2(
...
   if (CuDevice::Instantiate().Enabled()) {
     Timer tim;
-    char trans = (transA == kTrans ? 'N' : 'T');
+    cublasOperation_t trans = (transA == kTrans ? CUBLAS_OP_N : CUBLAS_OP_T);
     MatrixIndexT A_other_dim = (transA == kNoTrans ? A.num_cols_ : A.num_rows_);
-    
-    cublas_syrk('U', trans, num_rows_, A_other_dim, alpha, A.Data(),
+    cublasHandle_t handle = CuDevice::Instantiate().GetHandle();
+    cublasStatus_t stat;
+    stat=cublas_syrk(handle, CUBLAS_FILL_MODE_UPPER, trans, num_rows_, A_other_dim, alpha, A.Data(),
```
danpovey: Remember there is an 80-character length limits, you may have to split lines if you make them longer.
If you use emacs then check out my .emacs file, there is some relevant stuff in there e.g. calling google c style.

src/cudamatrix/cu-sp-matrix.h
```diff
@@ -46,7 +46,7 @@ class CuSpMatrix : public CuPackedMatrix<Real> {
   friend class CuRand<Real>;
 
   template<class R, class S>
-  friend R TraceSpSp(const CuSpMatrix<R> &A, const CuSpMatrix<S> &B);
+  friend R TraceSpSp(cublasHandle_t handle, const CuSpMatrix<R> &A, const CuSpMatrix<S> &B);
```
danpovey: This function should not take the handle argument.

src/cudamatrix/cu-sp-matrix.h
```diff
@@ -76,7 +76,7 @@ class CuSpMatrix : public CuPackedMatrix<Real> {
     CuPackedMatrix<Real>::Resize(nRows, resize_type);
   }
 
-  Real FrobeniusNorm() const { return sqrt(TraceSpSp(*this, *this)); }
+  Real FrobeniusNorm() const { cublasHandle_t handle = CuDevice::Instantiate().GetHandle(); return sqrt(TraceSpSp(handle, *this, *this)); }
```
danpovey: In situations where you use the handle only once, there is no need to declare a variable- you could just use the expression in the function call directly.  (in this case it shouldn't be needed at all once you remove it from TraceSpSp).

I suggest that you declare an inline function in cu-device.h as follows:

/// This function is declared as a more convenient way to get the CUDA device handle for use
/// in the CUBLAS v2 API, since we so frequently need to access it.
inline cublasHandle_t GetCublasHandle() {
  return CuDevice::Instantiate().GetHandle();
}
and then you can make things a bit shorter.

danpovey:
Thanks.
After you address my comments, once you push to freewym:master the pull
request will automatically get updated.
Generally speaking people use a named branch in their repo e.g.
freewym:cublas_v2_branch, for pull requests.  I guess it's not mandatory
though.
Dan

freewym:
Hi Dan,

Please check this update.

src/cudamatrix/cu-device.cc
```diff
@@ -559,20 +558,18 @@ void* CuDevice::Malloc(size_t size) {
...
 
-CuDevice::CuDevice(): active_gpu_id_(-1), verbose_(true)
-  { }
+CuDevice::CuDevice(): active_gpu_id_(-1), verbose_(true) {}
 
 
 CuDevice::~CuDevice() {
   if (Enabled()) {
-    cublasShutdown();
+    //cublasShutdown();
```
danpovey: please remove this comment

src/cudamatrix/cu-device.h
```diff
@@ -44,7 +45,8 @@ class CuDevice {
   ~CuDevice();
   static inline CuDevice& Instantiate() { return global_device_; }
 
-  // We provide functions Malloc, MallocPitch and Free which replace cudaMalloc,
+  inline cublasHandle_t GetHandle() { return handle_; }
+
```
danpovey: you removed a line of comment here, which should be restored.

src/cudamatrix/cu-device.h
```diff
@@ -100,14 +102,15 @@ class CuDevice {
...
  private:
   CuDevice();
   CuDevice(CuDevice&); // Disallow.
   CuDevice &operator=(CuDevice&);  // Disallow.
 
+
   static CuDevice global_device_;
-  
+  cublasHandle_t handle_;
```
danpovey: please add a blank line before the next comment

freewym: Sorry. Fixed.

danpovey: I am testing and looking to see if there is a speed difference (should not
be).  After that I may merge.
Dan

danpovey:
OK, I am happy with the content of your changes right now but I'd like you to clean up your commit history so there is just one commit, with a message like "changes to cudamatrix code to upgrade to the CuBLAS v2 API".  See if you can find someone proficient in git to help you do this.

PR #114, opened 2015-09-02
Check if priors job is done before exiting in steps/nnet2/train_discriminative2.sh

vimalmanohar:


jtrmal:
Simple "wait" at the end of the file wouldn't be enough?
y.

vimalmanohar:
Perhaps. I had some issue with using wait at the end. But it may not be a problem in this file.

jtrmal:
Ok.  Let me first,  before merging this,  apply the patch and test it.
Y.

jtrmal:
adding comment I talked with Vimal about (IRL), mainly so that @danpovey
can comment --
I was concerned that given the fact the priors adjustment modifies the
acoustic model, there might be interference between the adjustment itself
and the training process.
As it is run now, the adjustment runs on the background, so the next
iteration read the model before the model gets modified.
My concern about this is however that this does not have to be true in
cases where the training gets interrupted (for whatever reason) and
restarted, the already adjusted network will/can be read instead of the
unadjusted.
I'm not sure if that will matter or not -- but if it does indeed matter, we
should do something like (schematically)
cp 127.mdl 127.adj.mdl
#adjust priors on 127.mdl
ln -s 127.adj.mdl epoch4.mdl

so that the training process could be reproduced.
y.

danpovey:
Yes, it might affect the training process.  I don't see why the priors
adjustment had to run in the background in any case-- it seems like we
could just have it run in the foreground by removing the '&' and everything
would be much simpler.  Probably the time saved is small anyway.
Dan


adding comment I talked with Vimal about (IRL), mainly so that @danpovey

jtrmal:
I was proposing the copy+symlink so that the relationship of the epoch and
iteration would be kept -- that might be useful now and then (and also for
reporting if the adjustments helps or not).
I'd prefer, for technical reasons, to run it in serial. I think I was
mentioning it a few months back.
Dan, do you have any insight if it can (substantially) help (or harm) the
performance, if the adjustment will be done "in the loop", i.e. the next
model will be trained on the top of the adjusted network?

y.

danpovey:
It can definitely make a difference whether or not the training process
'sees' the prior adjustment.  My instinct is to let the training process
see the prior adjustment, but I'm not 100% sure, we'd need to test it.  The
current setup is really non-deterministic, but most likely the training
process is not seeing the prior adjustment if the GPU queue is lightly
loaded.

The way I would do it in case you *don't* want the training process to see
the prior adjustment is to use a differently named file, and to have the
thing backgrounded as it is now, and then wait before exiting.  In case you
*do* want the training process to see the prior adjustment, just do it
without the '&'.  But it needs to be experimentally tested.  Vimal, do you
have time?

Dan

vimalmanohar:
The time taken for adjusting priors about the same as 9 iterations of discriminative training. It is indended to be used only during decoding. So in order to make it equivalent to the earlier experiments done, Yenda's solution of copying it to a separate model would be better. 
I do not know how the prior adjustment will affect the training process. I can test this out.

danpovey:
OK- for now create a pull request for the separate-model solution (and wait
at the end).
Later after you do the experiments (which is no hurry) we can consider the
other issue.
Dan

Issue #90, opened 2015-08-24
Issues relating to RNNLM


elfring:
Would you like to add more error handling for return values from functions like the following?
- [fopen](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html "Open a stream.") ⇒ [TrainModelThread](https://github.com/kaldi-asr/kaldi/blob/1a59ece8f53e904d10cfce1758222a258e9a148e/tools/rnnlm-hs-0.1b/rnnlm.c#L410)
- [fscanf](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fscanf.html "Convert formatted input.") ⇒ [ReadVocab](https://github.com/kaldi-asr/kaldi/blob/1a59ece8f53e904d10cfce1758222a258e9a148e/tools/rnnlm-hs-0.1b/rnnlm.c#L333)

danpovey:
Ilya Edrenkin (cc'd) is responsible for this RNNLM training code; it is not
part of Kaldi proper.
But I don't see any reason why not to do it; if you submit a patch I would
likely approve it.

Dan

ilya-edrenkin:
Hi, 

I think it is a good idea.

Furthermore, @akhti has completely rewritten the rnnlm code and published it at https://github.com/yandex/faster-rnnlm
It is (almost) identical to rnnlm-hs-0.1b in terms of modeling capability, but also it:
1) is nearly 3 times faster;
2) supports NCE in addition to HS output;
3) greatly improved on style.
Probably we should replace the rnnlm in kaldi with this one since it is definitely better. Dan, what do you think about it?

jtrmal:
Ilya, how different are the command switches?
y.

ilya-edrenkin:
AFAIK Anton tried to make them 100% compatible (although there are some new switches, for example those that are used for NCE).

danpovey:
If someone had the time to make the changes and test the associated
scripts, and submit a pull request, that would be great.  Ilya?
Dan

jtrmal:
Ok, in that case, the first step would be to add
extras/install_faster-rnnlm.sh script (ideally downloading the sources
directly from github). This way we could compare the behavior/compatibility.
y.

jtrmal:
BTW, I cannot offer too much of my time -- I can review the script and help
with the issues you guys could run into, but not much else.
y.

ilya-edrenkin:
[...]

PR #116, opened 2015-09-02
Add faster-rnnlm rescoring to librispeech

akhti:
Add installation script for faster-rnnlm
Update rescoring results to compare HS and NCE

The pull request was 'announced' in issue #90


jtrmal:
Anton, thanks for the submission -- I think the install script should be
improved to support our "autoconfiguration" via the tools/env.sh -- have a
look at  extras/install_srilm.sh
Also, if you think the fasterrnnlm would be a drop-in replacement, feel
free to remove the original package from the repo and create symlink from
faster-rnnlm to rnnlm  (or something like that)
y.

akhti:
Do you mean that I should add rnnlm binary to the PATH? I believe this is reduntant as local/run_rnnlm.sh has an argument to exlicitly set prefered rnnlm version.

Btw, seems that "exit 1" is missing near line 9 in extras/install_srilm.sh .

jtrmal:
The purpose of the env.sh file is to get rid of  the (sometimes rather
convoluted) run-time detection of the "packages"/binaries and to provide
simple way how to set the necessary variables, not only path.
Remember that rnnlm can be used in other recipes as well and I would be
rather to provide the detection/configuration via the  this file than let
users to "invent" various way how to detect and handle various cases when
the binary exists/does not exists.
I'l look at the missing exit1, thanks
y.

akhti:
I agree that code duplication is evil. However, I do not like that installation script for one rnnlm implemantion makes itself default one. Recipe users may want to use different rnnlm implementation to compare their results. 

What do you think about the following solution. I will drop install_*rnnlm*.sh scripts and ad-hoc rnnlm installation within recipes. Instead I will write a single script that takes single argument rnnlm_version. If file tools/<rnnlm_version>/rnnlm exists, than nothing is done. Otherwise for some fixed number of options it will try to download and build the corresponding rnnlm toolkit and create executable (or symbolic link)  tools/<rnnlm_version>/rnnlm.

danpovey:
I think I am happy with the script as it is currently, just using a different name.  At least it's clear what's going on.  I would be happy to merge this pull request.  @jtrmal ?

Also, @chenguoguo @hainan-xv, please look into this and see if you think this has any implications for your RNNLM rescoring work.

jtrmal:
Well, I don't want to be obsessive about it but I think there is a certain
code style for recipes  (albeit not defined as precisely as for C++ code)
and we should be enforcing it, because it will be mostly us who will have
do the regular maintenance.
Yes, it makes sense to keep both installation scripts in there for now, but
in the long term I think we should get rid of the old  rnnlm code in favor
if the new one (as the new one seems supported better -- or that is my
hope). And I think that "magic" with the selection of the version of
rnnlm in the run_rnnlm.sh is really not nice. It only complicates things in
the long term.
So I don't see the reason why is there such a strong push against adding
RNNLMHS=$KALDI_ROOT/tools/rnnlm-hs
PATH=$PATH:$RNNLMHS
(or whatever) into tools/env.sh

y.

danpovey:
Well, I don't want to be obsessive about it but I think there is a certain
> code style for recipes (albeit not defined as precisely as for C++ code)
> and we should be enforcing it, because it will be mostly us who will have
> do the regular maintenance.
> Yes, it makes sense to keep both installation scripts in there for now, but
> in the long term I think we should get rid of the old rnnlm code in favor
> if the new one (as the new one seems supported better -- or that is my
> hope). And I think that "magic" with the selection of the version of
> rnnlm in the run_rnnlm.sh is really not nice. It only complicates things in
> the long term.
> So I don't see the reason why is there such a strong push against adding
> RNNLMHS=$KALDI_ROOT/tools/rnnlm-hs
> PATH=$PATH:$RNNLMHS
> (or whatever) into tools/env.sh
>

I think that the script should not be installing those things but should
instead be checking that it's installed and maybe prompting you to install
it and then dying; this is the normal pattern and it seems better from a
security standpoint.  However, it looks like the existing RNNLM rescoring
script, that's been there for a while, already takes the --rnnlm-ver
argument, so I think that should probably stay there for now unless someone
wants to redo all that stuff.  Actually I prefer to have it explicit in the
script which version of those tools is being used, because otherwise we
have a situation where the same script can produce different results
depending which installation scripts have been previously run.  This seems
like a kind of nondeterminism.  So I'm OK with the overall way the script
works.

Dan

jtrmal:
Ok then about the versioning, even though if there is such a strong call
for determinism, we should check out a specific version of the tool (as we
do for irstlm) and I believe it should be done other way.
Yes, the run_* script should only check if the given binary is present (and
not install it -- we had a discussion a few days back, that some user are
discouraged to have their on tools/ or perhaps whole kaldi directory) --
and for that either the RNNLMHS variable or you can even have RNNLMHSVER
set to something meaningful, as I assume there can be development in what
that package does. As for deciding of the version of rnnlm I'm okay with
keeping that switch there.
Again, I don't see why adding the environment variables into the env.sh is
such a big problem. I think I've spent enough time on maintaining the tools
directory and fixing the issues stemming from the fact that people invent
various "smart" things in their scripts how to detect if the given tool is
installed or not, to know this practice is helpful in a long term. I didn't
hear any arguments why the thing I'm proposing would be inadequate (or just
plain stupid).

But as I'm getting tired of this discussion, feel free to accept the PR and
I'll just create my personal list of things I will refuse to provide
help/support about.
y

danpovey:
I'd rather find a solution that everyone is happy with.
Yenda, I don't think it was absolutely clear to us what you were proposing
to do.  Were you planning to just have the call 'rnnlm' and rely on the
fact that it's in the path via env.sh?  Or are you now saying that you want
the env.sh to have some other variable (which would store what?  The
location of the 'bin' directory?)

In any case an issue with those approaches is that a person can
legitimately have different setups with the three different versions of the
RNNLM tools, e.g. different versions of the training script train RNNLMs
with different versions of the tools; I'm not sure that their training
command lines are even compatible.  When doing RNNLM rescoring, I believe
you have to rescore with the correct RNNLM tool that matches the model you
trained.  Certainly that is the case if we are comparing Mikolov's tool to
the Yandex ones; I don't know if the Yandex ones have the same model
format.  So your proposal might just cause crashes.
I know this whole situation is far from ideal.

If the newest RNNLM tool is always at least as good at Mikolov's one, then
one solution would be to change everything to just target that one tool,
but this would involve someone testing the old setups such as WSJ, which
might take some time.


Dan

chenguoguo:
I had a look at the scripts, and I think it's probably easier to have
rnnlm_ver variable in the script (as the script does now), so that we know
which tool we are using. IIRC Yandex's version and Mikolov's version have
different model format, although their command line arguments are mostly
the same. And Mikolov's version works better at least on the Aspire task I
tested.

Guoguo
--

danpovey:
BTW, if the 'faster' tool is truly a drop-in replacement for the rnnlm-hs
tool, and is better under all circumstances (and back-compatible in model
format) then we could just remove the existing rnnlm-hs-0.1b directory from
git, modify the extras/install_rnnlm_hs.sh script to install the 'faster'
version as rnnlm-faster, and modify all the scripts that currently use the
0.1b tool to use the -faster tool.

Dan

akhti:
That's true. faster-rnnlm supports everything that is supported by rnnlm-hs and can even read models trained with rnnlm-hs. Besides it supports NCE training that can outperform even Mikolov's models on some datasets.

However, that's could be wrong for other datasets. That's why I would like to have this --rnnlm_ver option.
Regarding installation anything from recipes, I agree that this approach could lead to various conflicts. I just followed current design, and could easily change this behaviour. Here and in wsj recipe.

jtrmal:
a) if I sound unwelcoming, I apologize -- I'm grateful for whatever small
submission.
b)  the idea of the configuration has may not settled enough. I wanted the
tools directory to have a mechanism that will allow the tools itself (or
more specifically the installation script) to have something like
  a1) assisted discovery mechanism (i.e. an egs script would be able to
tell easily if that tool is present and if the given binary is runable)
  a2) proper configuration of the environment -- this does not have look
like a big deal for C/C++ code, but actually is for python modules (for
example)

The a1) is realized by setting a variable of a particular name (such as I'm
suggesting RNNLMHS in this case, but I'm rather indifferent in what it
should be) and
the idea a2) is realized by setting the PATH, PYTHON_PATH (and possibly
others) by just adding/appending to them the right value.
The idea would be the tool would output the settings into that env file and
individual egs would source it either via path.sh (or perhaps in the
run_*.sh script)

I actually realized a week back (when we had the discussion about tools and
$KALDI_ROOT), that we could add a script that could be called for example
utils/verify_tool_installed.sh --binary "ngram"
tools/extras/install_srilm.sh
or
utils/verify_tool_installed.sh --env "SRILM" tools/extras/install_srilm.sh
(because I can see utility in both calls, and I still don't know which way
might be better)
which could verify either the env variable exists/is-set or the binary
existence could be verified using "which".
The other parameter might be voluntary -- and in that case the check could
print info about how the package should be installed.
The message should be something in the sense we print now:
The binary "ngram" could not be found. You should call the script
tools/extras/install_srilm.sh to install it.
(similar for the env. variable).

Version check could be also added, albeit that might be complicating things
overly -- for the run_rnnlm.sh i think
this would be more than ok:
if [ rnnlm_ver == "rnn-0.1b" ]; then
  utils/verify_tool_installed.sh --binary "rnnlm"
tools/extras/install_rnnlm.sh || exit 1
else
  utils/verify_tool_installed.sh --binary "rnnlm"
tools/extras/install_rnnlm-hs.sh || exit 1
fi

and that should be basically it.

(or whatever, I don't remember the script names and the version names).

I didn't mentioned the script, as coming with the
utils/verify_tool_installed.sh script seemed too much to ask someone from
the "outside", not because they wouldn't manage it, but simply because I
didn't want to come with extra burden for them -- there might be some
tweaking needed and figuring the right mechanism, I just didn't mention it
and primarily ask just for setting of those variables. And yes, now I
realize that there is a potential problem with a name clash (rnnlm binary
provided either by rnnlm or rnnlm-hs, which I didn't realize before, as
I've believed the rnnlm should be replaced by rnnlm-hs -- that was the
reason I was asking in the beginning if it's a drop in replacement
providing a super-set of functionality).
Does this explain better what I was aiming for?
y.

jtrmal:
Anton, we've talked about it with Dan outside and we decided that for this
PR it would be ok if you'd just modify the run_rnnlm.sh script to detect
the binary is available. If not, print something in the sense
"The rnnlm binary was not found. Please install the tool using
tools/extras/install_rnnml-hs.sh"
or something like that -- I will leave the exact wording to you (but it
should contain the hint about which of the two scripts to be run)
After that, I or Dan will merge the PR and we will deal with the rest of
the things separately.
Thanks for your work,
y.

ilya-edrenkin:
Anton, I think Dan is right -- there is no need to keep the old rnnlm-hs-0.1b in the repo as your version is clearly better and fully backward-compatible. I hope it will help keep things simpler and cleaner.

akhti:
It's me again!

I grepped for instal_*_rnnlm.sh calls and replaced all of them with checks and warning. Also, I've replaced rnnlm-hs with faster-rnnlm everywhere. Does it look OK?

danpovey:
Thanks!
@chenguoguo , would you mind looking over this?

chenguoguo:
Sure I'll go through it when I get home tonight

chenguoguo:
I went through all the commits, and I think I'm happy with the current patch. I have some comments below:

0. This is just for others' info, that rnnlm-hs has been removed from tools/ in this patch. I'm listing it here in case there's any objection from others.
1. In the bash scripts you added (tools/extras/check_for_rnnlm.sh and tools/extras/install_class_rnnlm.sh), you may want to change your shebang to "#!/bin/bash" so that they are consistent to other scripts.
2. In tools/extras/install_class_rnnlm.sh, you used the name "class rnnlm" for Mikolov's tool. Is "class rnnlm" the typical name for his tool? Perhaps it's install_mikolov_rnnlm.sh is more straightforward?
3. Here: egs/wsj/s5/local/run_rnnlm-hs_tri3b.sh, you may want to change the file name since you are no longer using rnnlm-hs

After you address those comments, if there's no more comments from Dan or Yenda, I think it's good to commit.

Guoguo

chenguoguo:
Sorry, accidently clicked on "Cloase and comment"... reopened it.

akhti:
Thanks for the comments!

2. Done
3. Done
4. I'm not sure about this. Actually, by default faster-rnnlm trains a model using HS (hierarchical softmax). So it seems quite natural to refer to them as rnnlm-hs, doesn't it?

chenguoguo:
Makes sense. I was hoping that this file name won't give people the impression that it's the old rnnlm-hs-0.1b tool, but I guess it's OK. The patch looks good to me then. If there's no comments from others, it's good to merge.

Guoguo

danpovey:
OK, I'm merging this because it's been so long.  @vdp, you might want to look over the Librispeech stuff and see if there is anything you want to clean up, but just do this as a separate pull request if you want to do that.

danpovey:
BTW, I should mention that Guoguo has been working on lattice RNNLM
rescoring.  This hasn't been merged yet because he was busy over the
summer, but it works.  However, it's with the old (Mikolov's) tool.
Hainan Xu will help with cleaning it up to check in.
Actually it just occurred to me that we might be able to develop it more
publicly and then get some help from the Yandex guys to support their tool
too.
Dan

Issue #69, opened 2015-08-11
nnet3- update CachingOptimizingCompiler to store a configurable number of computations.


danpovey:
Would require adding a hashing object for class ComputationRequest, so we can do fast lookup in a hash.
Idea is to cache the n most recently accessed Computations.  Probably a reasonable default is to cache around 20 computations.

danpovey:
@naxingyu , you might want to try this.
Make sure you understand hashing functions, hashing objects, equality-testing objects, and unordered_map first.  Grep around for unordered_map objects in the existing code.
Dan

naxingyu:
I'll see what I can do.

naxingyu:
My guess is
1) add a member to CachingOptimizingCompiler class,
     unordered_map<int32, ComputationRequest> request_cache_;
2) Reserve the size of request_cache_ to a configurable number (20 as 
default).
3) Each time Compile() is called, add it to the bucket.

Is it correct? And how would you like the key defined? I see we have 
StringHasher, VectorHasher, PairHasher and CindexHasher.

danpovey:
> My guess is
> 1) add a member to CachingOptimizingCompiler class,
> unordered_map<int32, ComputationRequest> request_cache_;
> 2) Reserve the size of request_cache_ to a configurable number (20 as
> default).

3) Each time Compile() is called, add it to the bucket.
>
> Is it correct? And how would you like the key defined? I see we have
> StringHasher, VectorHasher, PairHasher and CindexHasher.


This is not really right, it needs to be a map from something like
ComputationRequest to Computation (or possibly pointers).  Which requires a
hashing object.  And 'reserve' on the map is just an advisory thing to the
unordered_map code; it's unrelated to the core issue, which is only keeping
a number of Computations (the n most recent)... this might involve some
kind of priority queue, or at least storing something which says how
recently a particular computation was used (e.g. value in map could be
std::pair<int32,  Computation>, with the int32 being some kind of 't' value
keeping track of freshness.

Actually, a map may not even be necessary (or necessarily the most
efficient way of doing it)-- once we have the hash function of the
ComputationRequest, manually going through 20 hash functions of
ComputationRequests and comparing might not be that hard.  So simply a
priority queue may be the right way.  But I guess we are probing the limits
of your knowledge of C++.
Kirill, do you have time to either do this, or explain to Xingyu in more
detail how to do it?

Dan

danpovey:
> My guess is
[...]

PR #120, opened 2015-09-03
add nnet3 computation cache

naxingyu:
This is an attempt to resolve issue #69 . Please check if it make some sense.

### DIFF
--- a/src/nnet3/nnet-optimize.cc
+++ b/src/nnet3/nnet-optimize.cc
@@ -876,21 +876,23 @@ size_t ComputationRequestHasher::IoSpecificationToInt(const IoSpecification& spe
   return ans;
 }
 
-void CachingOptimizingCompiler::UpdateCache(const ComputationRequest &request,
-                                            NnetComputation &computation) {
+void CachingOptimizingCompiler::UpdateCache(const ComputationRequest *request,
+                                            NnetComputation *computation) {
   if (computation_cache_.size() == capacity_) {
     // full, locate the least-recently-accessed request
     const typename CacheType::iterator it
        = computation_cache_.find(access_queue_.front());
     KALDI_ASSERT(it != computation_cache_.end());
     // purge the least-recently-accessed request
+    delete it->first;
+    delete it->second.first;
     computation_cache_.erase(it);
     access_queue_.pop_front();
   }
   typename AqType::iterator ait
-    = access_queue_.insert(access_queue_.end(), &request);
-  computation_cache_.insert(std::make_pair(&request,
-                            std::make_pair(&computation, ait)));
+    = access_queue_.insert(access_queue_.end(), request);
+  computation_cache_.insert(std::make_pair(request,
+                            std::make_pair(computation, ait)));
 }
 
 void CachingOptimizingCompiler::UpdateAccessQueue(typename CacheType::iterator &cit) {
@@ -901,14 +903,28 @@ void CachingOptimizingCompiler::UpdateAccessQueue(typename CacheType::iterator &
                        cit->second.second);
 }
 
+CachingOptimizingCompiler::~CachingOptimizingCompiler() {
+  typename CacheType::const_iterator itr = computation_cache_.begin(),
+                                     end = computation_cache_.end();
+  for (; itr !=end; ++itr) {
+    delete itr->first;
+    delete itr->second.first;
+  }
+  access_queue_.clear();
+  computation_cache_.clear();
+}
+
 const NnetComputation* CachingOptimizingCompiler::Compile(
-    const ComputationRequest  &request) {
+    const ComputationRequest  &in_request) {
   NnetComputation *computation;
+  ComputationRequest *request;
   // find computation in the cache
-  typename CacheType::iterator cit = computation_cache_.find(&request);
+  typename CacheType::iterator cit = computation_cache_.find(&in_request);
   if (cit == computation_cache_.end()) {
     // if not found, compile and update cache
-    Compiler compiler(request, nnet_);
+    request = new ComputationRequest;
+    *request = in_request;
+    Compiler compiler(*request, nnet_);
     CompilerOptions opts;
     computation = new NnetComputation;
     compiler.CreateComputation(opts, computation);
@@ -916,7 +932,7 @@ const NnetComputation* CachingOptimizingCompiler::Compile(
     int32 verbose_level = 4;
     if (GetVerboseLevel() >= verbose_level) {
       std::ostringstream os1;
-      request.Print(os1);
+      request->Print(os1);
       KALDI_LOG << "Computation request is " << os1.str();
       std::ostringstream os2;
       computation->Print(os2, nnet_);
@@ -926,18 +942,18 @@ const NnetComputation* CachingOptimizingCompiler::Compile(
       CheckComputationOptions check_config;
       // we can do the rewrite check since it's before optimization.
       check_config.check_rewrite = true;
-      ComputationChecker checker(check_config, nnet_, request,
+      ComputationChecker checker(check_config, nnet_, *request,
                                  *computation);
       checker.Check();
     }
-    Optimize(opt_config_, nnet_, request, computation);
+    Optimize(opt_config_, nnet_, *request, computation);
     { // check the computation again.
       CheckComputationOptions check_config;
-      ComputationChecker checker(check_config, nnet_, request, *computation);
+      ComputationChecker checker(check_config, nnet_, *request, *computation);
       checker.Check();
     }
     computation->ComputeCudaIndexes();
-    UpdateCache(request, *computation);
+    UpdateCache(request, computation);
   } else {
     // if found, update access queue
     computation = cit->second.first;
--- a/src/nnet3/nnet-optimize.cc
+++ b/src/nnet3/nnet-optimize.cc
@@ -878,7 +878,7 @@ size_t ComputationRequestHasher::IoSpecificationToInt(const IoSpecification& spe
 
 void CachingOptimizingCompiler::UpdateCache(const ComputationRequest *request,
                                             NnetComputation *computation) {
-  if (computation_cache_.size() == capacity_) {
+  if (computation_cache_.size() == cache_capacity_) {
     // full, locate the least-recently-accessed request
     const typename CacheType::iterator it
        = computation_cache_.find(access_queue_.front());
@@ -910,8 +910,6 @@ CachingOptimizingCompiler::~CachingOptimizingCompiler() {
     delete itr->first;
     delete itr->second.first;
   }
-  access_queue_.clear();
-  computation_cache_.clear();
 }
 
 const NnetComputation* CachingOptimizingCompiler::Compile(
--- a/src/nnet3/nnet-optimize.cc
+++ b/src/nnet3/nnet-optimize.cc
@@ -845,6 +845,30 @@ void Optimize(const NnetOptimizeOptions &config,
 
 }
 
+void CachingOptimizingCompiler::UpdateCache() {
+  typename cache_type::iterator cit = computation_cache_.find(&request_);
+  if (cit == computation_cache_.end()) {
+    // not exist, insert
+    if (computation_cache_.size() == capacity_) {
+      // full, locate the least-recently-accessed request
+      const typename cache_type::iterator it
+         = computation_cache_.find(access_queue_.front());
+      KALDI_ASSERT(it != computation_cache_.end());
+      // purge the least-recently-accessed request
+      computation_cache_.erase(it);
+      access_queue_.pop_front();
+    }
+    typename aq_type::iterator ait
+      = access_queue_.insert(access_queue_.end(), &request_);
+    computation_cache_.insert(std::make_pair(&request_,
+                              std::make_pair(&computation_, ait)));
+  } else {
+    // exist, update access record by moving the accessed
+    // request to the end of the access queue
+    access_queue_.splice(access_queue_.end(), access_queue_, (*cit).second.second);
+  }
+}
+
 const NnetComputation* CachingOptimizingCompiler::Compile(
     const ComputationRequest  &request) {
   if (!(request == request_)) {
@@ -879,6 +903,7 @@ const NnetComputation* CachingOptimizingCompiler::Compile(
     }
     computation_.ComputeCudaIndexes();
   }
+  UpdateCache();
   return &computation_;
 }
 
--- a/src/nnet3/nnet-optimize.cc
+++ b/src/nnet3/nnet-optimize.cc
@@ -876,43 +876,42 @@ size_t ComputationRequestHasher::IoSpecificationToInt(const IoSpecification& spe
   return ans;
 }
 
-void CachingOptimizingCompiler::UpdateCache(bool insert_new_computation) {
-  if (insert_new_computation) {
-    // not exist, insert
-    if (computation_cache_.size() == capacity_) {
-      // full, locate the least-recently-accessed request
-      const typename CacheType::iterator it
-         = computation_cache_.find(access_queue_.front());
-      KALDI_ASSERT(it != computation_cache_.end());
-      // purge the least-recently-accessed request
-      computation_cache_.erase(it);
-      access_queue_.pop_front();
-    }
-    typename AqType::iterator ait
-      = access_queue_.insert(access_queue_.end(), &request_);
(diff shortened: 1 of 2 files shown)


src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,50 @@ void Optimize(const NnetOptimizeOptions &config,
               const ComputationRequest &request,
               NnetComputation *computation);
 
-
+struct ComputationRequestHasher {
+  size_t operator()(const ComputationRequest *cr) const {
```
danpovey: This hashing function is very insufficient.
The input.size() and outputs.size() will normally be 1, so everything will hash to the same value.  You should go right in and look at the individual indexes, plus the names of the inputs and outputs.

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,50 @@ void Optimize(const NnetOptimizeOptions &config,
...
+};
+ 
 /// This class enables you to do the compilation and optimization in one call,
 /// and also ensures that if the ComputationRequest is identical to the previous
 /// one, the compilation process is not repeated.
 class CachingOptimizingCompiler {
  public:
-  CachingOptimizingCompiler(const Nnet &nnet): nnet_(nnet) { }
+  CachingOptimizingCompiler(const Nnet &nnet,
+                            const int32 capacity = 20):
```
danpovey: You should make capacity part of the configuration class, and give it a more informative name, e.g. num-computations-cached.

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,50 @@ void Optimize(const NnetOptimizeOptions &config,
...
+                            const NnetOptimizeOptions &opt_config,
+                            const int32 capacity = 20):
+      nnet_(nnet), opt_config_(opt_config), capacity_(capacity) { }
 
   /// Does the compilation and returns a const pointer to
   /// the result, which is owned by this class, not the caller.
   /// It calls ComputeCudaIndexes() for you, because you wouldn't
   /// be able to do this on a const object.
   const NnetComputation* Compile(const ComputationRequest  &request);
+  void UpdateCache();
```
danpovey: This should not be public.

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,50 @@ void Optimize(const NnetOptimizeOptions &config,
...
  private:
   const Nnet &nnet_;
   NnetOptimizeOptions opt_config_;
   ComputationRequest request_;
   NnetComputation computation_;
+
+  typedef std::list<ComputationRequest*> aq_type;
+  typedef unordered_map<ComputationRequest*, std::pair<NnetComputation*,
+    typename aq_type::iterator>, ComputationRequestHasher> cache_type;
```
danpovey: The default == operator used here will just compare the pointer values.  That's not what you need.  You have to define an equality operator as an object, like you did for the hashing function.

vdp:
In addition to what Dan said, I think you will need to get rid of *request_* and *computation_* fields altogether. Unless I'm misunderstanding something, these are basically single-element cache, and we are trying to replace them with this more sophisticated data structure. As it is at the moment even if you improve the hashing and implement an identity operation for the request pointers you still will be storing pointers to the same object(in different hash map buckets), which doesn't make a lot of sense. You could probably allocate dynamically up to *capacity_* such objects and after that recycle the objects that you kick out from the *access_queue_* (maybe an even more descriptive name is possible for this list btw- how about eviction_q_ (or eviction_queue_)? I probably obsess too much over this, but as they say naming things is one of the only two hard problems in CS :)

danpovey:
What do you mean by "you still will be storing pointers to the same
object(in different hash map buckets)"?

vdp:
Update() is only storing pointers to request_ and computation_, so it seems to me it will end up storing many copies of pointers to the same objects. Or am I missing something?

naxingyu:
I update the code according to Dan's comments.
@vdp I need to read more nnet3 code to reply to your comments.

vdp:
As I've mentioned I also don't know the nnet3 code and I'm don't have much time to review, so take my comments with a grain of salt, and don't hesitate to point out if you feel something I say is incorrect.

Just a couple more things: you probably will need to change Compile() to actually use the cache; e.g. replace:

    if (!(request == request_)

with a check whether the object is already in the cache and so on. Also, if not too difficult, some sort of unit test might be helpful.

src/nnet3/nnet-optimize.h
```diff
@@ -44,7 +47,8 @@ struct NnetOptimizeOptions {
                          remove_assignments(true),
                          initialize_undefined(true),
                          move_sizing_commands(true),
-                         allocate_from_other(true) { }
+                         allocate_from_other(true),
+                         num_computations_cached(20) { }
```
danpovey: Sorry, now that I see this code I realize that I told you something wrong.  I though this was a class for the CachingOptimizingCompiler's options, but it was not.  Sorry.  You can revert that part of the code to the way it was, I think.

src/nnet3/nnet-optimize.h
```diff
@@ -76,18 +81,67 @@ void Optimize(const NnetOptimizeOptions &config,
               const ComputationRequest &request,
               NnetComputation *computation);
 
+// Hash function for ComputationRequest. It converts
+// ComputationRequest to hash code by looking at input
+// and output IoSpecifications vectors.
+struct ComputationRequestHasher {
```
danpovey: You should probably define these operators in the .cc file, not in the header.

src/nnet3/nnet-optimize.h
```diff
@@ -76,18 +81,67 @@ void Optimize(const NnetOptimizeOptions &config,
...
+    end = cr->outputs.end();
+    for (; itr != end; ++itr) {
+      ans += IoSpecificationToInt(*itr);
+    }
+    return ans;
+  }
+ private:
+  size_t IoSpecificationToInt(const IoSpecification& spec) const {
+    size_t ans, slen = spec.name.length();
+    const char *c = spec.name.c_str(), *send = c + slen;
```
danpovey: for this part of the code you can use the StringHasher() object, e.g.
StringHasher string_hasher;
size_t ans = string_hasher(spec.name);
BTW, your code was using an unitialized variable (ans).  You should run your code on valgrind to see if there are other instances of this.

src/nnet3/nnet-optimize.h
```diff
@@ -99,6 +153,28 @@ class CachingOptimizingCompiler {
...
+  // Compile() before returning computation. If the request is found in
+  // the cache, update the recently-accessed queue. Otherwise, insert
+  // the request to the end of the queue, and purge the least-recently-accessed
+  // request from the queue and the cache if the capacity is reached.
+  void UpdateCache();
+
+  // The access queue for keeping track of the freshness of computation.
+  // Most-recently-accessed computation is at the end, and
+  // least-recently-accessed computaiton is at the beginning.
+  typedef std::list<ComputationRequest*> aq_type;
```
danpovey: Typedefs should be upper-base e.g. AqType.

src/nnet3/nnet-optimize.cc
```diff
@@ -879,6 +903,7 @@ const NnetComputation* CachingOptimizingCompiler::Compile(
     }
     computation_.ComputeCudaIndexes();
   }
+  UpdateCache();
```
danpovey: The whole point of this exercise is to avoid redundantly compiling things we have compiled before- for speed.  Think about this more carefully.

naxingyu:
The code is restructured following Dan's comments and some of Vassil's advise.

src/nnet3/nnet-optimize.cc
```diff
@@ -878,6 +938,12 @@ const NnetComputation* CachingOptimizingCompiler::Compile(
       checker.Check();
     }
     computation_.ComputeCudaIndexes();
+    UpdateCache(true);
+  } else {
+    // if found, get computation and update access queue
+    request_ = *(cit->first);
+    computation_ = *(cit->second.first);
```
danpovey: There is no reason to have the variables 'computation_' and 'request_', they are just redundant copies.  Just return the pointer to the value in the map.

naxingyu: Sorry but I need to make this clear.
Should I remove member variables 'computation_' and 'request_' from the 
CachingOptimizingCompiler class?

在 2015/9/5 12:11, Daniel Povey 写道:

danpovey: Yes, remove them.

naxingyu:
Done.

danpovey:
There is a lack of clarity, when you pass pointer arguments around, about exactly where they are owned.  You should not pass a reference to a variable in a situation where you are passing ownership into the function- in that case, use a pointer, even if it is const.   
The purpose and function of UpdateCache() is unclear, as are its inputs and outputs.
Also, I have a feeling that there may be memory errors-- I don't see where the request is being allocated.
After you have addressed these issues, I'd like @vdp to look over the code more carefully and look for issues.  @vdp, do you think you will have time to do this?  Otherwise I'd like to find someone else whose C++ is a bit more advanced than yours, to check this.

src/nnet3/nnet-optimize.cc
naxingyu: BTW, should I delete the newed computations in a deconstructer.

danpovey: Yes- delete them there; and whenever you remove them from the map in the
UpdateCache function or whatever you decide to call it.  Get it to a state
where valgrind does not report any 'definitely lost'
Dan

naxingyu:
Deallocators are added but not tested in valgrind, yet. @vdp , do you have time to have a look? I need to add test code for cache in ```nnet-optimize-test.cc``` later. @danpovey , I see the optimizer is tested 20 times in the test code. Should I apply cache to this test? Or write a whole new test code for the cache?

vdp:
(was out of town for the weekend)
It seems to me that the allocations and deletions of request/computation objects are redundant. As suggested in a previous comment I think you can recycle the least-recently-used objects, as in the original code where the same request_ and computation_ objects are used repeatedly. I only skimmed through the code, but if there are serious issues the unit test and Valgrind should be able to catch them anyway...

danpovey:
I probably agree with Vassil - I have not checked the code, but if
something exists already you are supposed to just return the cached pointer
value.
Dan

naxingyu:
Hi Dan and Vassil,

I'd better explain what is going on here.

Yes, I did just return the cached pointer value if something exists.
It works like this. When Compile() is called using an input request, it 
try to lookup
the cache. If found, the pointer to the requested computation is returned,
and access queue is updated, nothing more. If not, the compilation runs,
and the pointer to the request and computation is cached. I thought that 
was the plan
for this cache.

Now, why are the allocations and deletions there? For the computations,
if it is used as in the original code, the pointers would point to a 
same block of memory
of computation. Therefore, I need to allocate a new memory to store the 
computation
and put the correct pointer into the cache. For the request, I checked 
how the
CachingOptimizingCompiler class is used in other code. Each time a 
compiler is initialized and
```compiler.Compile(request)``` is ran, a new request is used, whose 
life span is only
within the function that Compile() is called. Or least, the compiler 
doesn't know when the request is
destroyed on the outside. Therefore, memory is allocated to store the 
request and
computation when they are put into the cache, and is deleted when 
kicking them out.

Maybe my understanding of the whole caching was wrong at the first 
sight. Please correct me.
And I am truly sorry if this is holding the developing back. If it is, 
it's always ok to
close this and find someone else with better C++ skills and better 
understanding of
nnet3 to do this.

File: src/nnet3/nnet-optimize.h [hunk 11]
```diff
@@ -76,29 +77,73 @@ void Optimize(const NnetOptimizeOptions &config,
...
-  ComputationRequest request_;
-  NnetComputation computation_;
+
+  // The access queue for keeping track of the freshness of computation.
+  // Most-recently-accessed computation is at the end, and
+  // least-recently-accessed computaiton is at the beginning.
+  typedef std::list<const ComputationRequest*> AqType;
+  AqType access_queue_;
+
+  // Hash table for fast accssing a computation. Key is a pointer to
```
danpovey: accessing

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,73 @@ void Optimize(const NnetOptimizeOptions &config,
...
+  // ComputationRequest, and value is a pair of pointers to the Computation
+  // and its position in the access queue.
```
danpovey: Please replace this whole comment with the following; comments should contain information that is not already obvious from the code.
  // Map from computation-request to pair of (computation, and position in access_queue_).
  // Used for fast lookup of previously compiled computations.
  // All pointers are owned here.

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,73 @@ void Optimize(const NnetOptimizeOptions &config,
...
+
+  // This function updates the computation cache. It is called within
+  // Compile(). It insert the request to the end of the queue, and purge
+  // the least-recently-accessed request from the queue and the cache
+  // if the capacity is reached.
+  void UpdateCache(const ComputationRequest *request,
+                   NnetComputation *computation);
+  // This function updates the recently accessed queue.
+  void UpdateAccessQueue(typename CacheType::iterator &cit);
+  int32 capacity_;
```
danpovey: Please change capacity_ to cache_capacity_ and add just before it this comment:

  // This configuration value determines how many unique Computations
  // to cache in our most-recently-used cache.

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,73 @@ void Optimize(const NnetOptimizeOptions &config,
...
+  }
+};
+ 
 /// This class enables you to do the compilation and optimization in one call,
 /// and also ensures that if the ComputationRequest is identical to the previous
 /// one, the compilation process is not repeated.
 class CachingOptimizingCompiler {
  public:
-  CachingOptimizingCompiler(const Nnet &nnet): nnet_(nnet) { }
+ CachingOptimizingCompiler(const Nnet &nnet,
```
danpovey: Indentation level seems wrong, should be 2 characters.

src/nnet3/nnet-optimize.h
```diff
@@ -76,29 +77,73 @@ void Optimize(const NnetOptimizeOptions &config,
...
+  const NnetComputation* Compile(const ComputationRequest &request);
  private:
   const Nnet &nnet_;
   NnetOptimizeOptions opt_config_;
-  ComputationRequest request_;
-  NnetComputation computation_;
+
+  // The access queue for keeping track of the freshness of computation.
+  // Most-recently-accessed computation is at the end, and
+  // least-recently-accessed computaiton is at the beginning.
```
danpovey: Add to this comment the following:
  // Together with computation_cache_, this forms a most-recently-used (MRU)
  // cache for Computations, indexed by ComputationRequest.  Pointers
  // are owned in computation_cache_.

src/nnet3/nnet-optimize.cc
```diff
@@ -845,41 +845,121 @@ void Optimize(const NnetOptimizeOptions &config,
...
+}
+
+CachingOptimizingCompiler::~CachingOptimizingCompiler() {
+  typename CacheType::const_iterator itr = computation_cache_.begin(),
+                                     end = computation_cache_.end();
+  for (; itr !=end; ++itr) {
+    delete itr->first;
+    delete itr->second.first;
+  }
+  access_queue_.clear();
```
danpovey: I don't think these two clear() commands are really necessary because it's the destructor.  (I know there are instances where existing code does this).

danpovey:
OK, thanks.
I looked at your code and I think you are right and the code is OK.
I commented with various issues but they are all very cosmetic.
However, before I commit this I want to ask a student to test it and make
sure there is no degradation in speed.
(If there is, I might consider changing the hashing operator).
Dan

danpovey:
@pegahgh, could you please try downloading this patch, compiling the code, and comparing the training times for nnet-train?  My expectation is that it will be very slightly slower (it's solving an efficiency problem that doesn't apply to our current runs), but it should be a small difference.  
Dan

naxingyu:
Thanks. The comments are updated.

danpovey:
OK, very good.  It might be a couple of days before Pegah can test it
though.
Dan

danpovey:
The time taken did not seem to be appreciably affected (results from Pegah), so merging.

PR #121, opened 2015-09-03
changes to cudamatrix code to upgrade to the CuBLAS v2 API

freewym:
Hi Dan,

I was only able to squash my own commits (4) and some (~10) latest commits from kaldi-asr/kaldi into one commit because squash more (~3000 in total) commits would fail. I asked Guoguo and he said squash my own commits as one was fine and it was probably what you asked to do.  But I am not sure if it was fine. Anyway I have backed up the repository locally.

Yiming


danpovey:
Sorry, that wasn't what I meant.  I mean to squash only your own commits.
I think at this point you are in increasing danger of losing your work.
You should find someone more familiar with git to help you out with this-
e.g. Guoguo, or someone else like Vijay.
Dan

danpovey:
Actually, don't worry about this- I have a local copy of what you
previously committed and I can squash it myself.  I'm doing it now.

freewym:
Many thanks. I will still find someone to help me how to figure it out so that I will do it right in future.

PR #125, opened 2015-09-04
openSUSE

keli78:
Updated the installation script for installing Kaldi on openSUSE.

### DIFF
--- a/egs/wsj/s5/steps/nnet2/get_egs2.sh
+++ b/egs/wsj/s5/steps/nnet2/get_egs2.sh
@@ -30,6 +30,7 @@ frames_per_eg=8   # number of frames of labels per example.  more->less disk spa
                   # note: the script may reduce this if reduce_frames_per_eg is true.
 left_context=4    # amount of left-context per eg
 right_context=4   # amount of right-context per eg
+delta_order=      # delta feature order
 
 reduce_frames_per_eg=true  # If true, this script may reduce the frames_per_eg
                            # if there is only one archive and even with the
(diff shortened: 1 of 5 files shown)


danpovey:
Thanks.
Part of this problem was pre-existing, but all invocations of add_packages should have the same number of command line arguments (now 3).  These are the package names used for those utilities on the 3 OS variants.  This will involve a little research.  You can ask the others to help you, e.g. @vimalmanohar .  Each package manager has its own syntax and its own way of figuring out the package names.
Dan

keli78:
Thanks for your reply. I'm wondering whether you could explain a little bit
of "this problem".
Ke

On Fri, Sep 4, 2015 at 4:39 PM, Daniel Povey <notifications@github.com>
wrote:

> Thanks.
> Part of this problem was pre-existing, but all invocations of add_packages
> should have the same number of command line arguments (now 3). These are
> the package names used for those utilities on the 3 OS variants. This will
> involve a little research. You can ask the others to help you, e.g.
> @vimalmanohar <https://github.com/vimalmanohar> . Each package manager
> has its own syntax and its own way of figuring out the package names.
> Dan
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/125#issuecomment-137847113>.
>



-- 
Ke Li
Research Assistant
Dept. of Biomedical Engineering
Johns Hopkins University
Phone: 410-934-9802
Email: kli26@jhu.edu

danpovey:
What I mean is that previous to your changes, all invocations should have
had 2 arguments, but some had only 1 argument.

Now all should have 3.  Fixing this may involve some research on either
Debian or Red Hat.
BTW, please don't leave either of your OpenSuse instances running when you
are not using them.  You can 'stop' them using the interface.  My Google
cloud credits seem to have run out, so I am now paying for this.
Dan

danpovey:
I already merged this from a different pull request -> deleting.

PR #127, opened 2015-09-07
Added AddMatMatBatched code

freewym:
The batched AddMatMat function passed my test code.
 
Issues possibly needed to investigate:
1. Real ** cannot been converted to const Real ** when calling cublas_gemmBatched(). Currently I use const_cast at line 1761 in cu-matix.cc.
2. 0.5f and 0.0f cannot be automatically casted to type double as its counterpart in UnitTestCuMatrixAddMatMat(). Currently I use static_cast at line 1311 in cu-matrix-test.cc

valgrind shows there is 0 definite losts on both cu-matrix-test and cu-matrix-speed-test

The speed test shows the speed boosting of using GPU over CPU:
for CPU:
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 16, batchSize = 10, speed was 0.017399 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 32, batchSize = 10, speed was 0.0493568 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 64, batchSize = 10, speed was 0.0557112 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 128, batchSize = 10, speed was 0.0564673 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 256, batchSize = 10, speed was 0.0572283 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 512, batchSize = 10, speed was 0.0598009 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 1024, batchSize = 10, speed was 0.0606233 gigaflops.
for GPU:
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 16, batchSize = 10, speed was 0.0190346 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 32, batchSize = 10, speed was 0.312882 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 64, batchSize = 10, speed was 2.31153 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 128, batchSize = 10, speed was 10.5593 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 256, batchSize = 10, speed was 3.10337 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 512, batchSize = 10, speed was 28.5466 gigaflops.
LOG (TestCuMatrixMatMatBatched():cu-matrix-speed-test.cc:89) For CuMatrix::AddMatMatBatched<double>, for dim = 1024, batchSize = 10, speed was 40.4558 gigaflops.

### DIFF
--- a/src/cudamatrix/cu-matrix-speed-test.cc
+++ b/src/cudamatrix/cu-matrix-speed-test.cc
@@ -58,6 +58,38 @@ template<typename Real> void TestCuMatrixMatMat(int32 dim) {
             << dim << ", speed was " << gflops << " gigaflops.";
 }
 
+template<typename Real> void TestCuMatrixMatMatBatched(int32 dim, int32 batchCount) {
+  std::vector<CuMatrix<Real>* > a(batchCount), b(batchCount), c(batchCount);
+  std::vector<CuSubMatrix<Real>* > A, B, C;
+  
+  for (int32 i = 0; i < batchCount; i++) {
+    // first create a Matrix intance and then creat a SubMatrix instance from that
+    a[i] = new CuMatrix<Real>(dim, dim);
+    b[i] = new CuMatrix<Real>(dim, dim);
+    c[i] = new CuMatrix<Real>(dim, dim);
+    a[i]->SetRandn();
+    b[i]->SetRandn();
+    A.push_back(new CuSubMatrix<Real>(*(a[i]),0,a[i]->NumRows(),0,a[i]->NumCols()));
+    B.push_back(new CuSubMatrix<Real>(*(b[i]),0,b[i]->NumRows(),0,b[i]->NumCols()));
+    C.push_back(new CuSubMatrix<Real>(*(c[i]),0,c[i]->NumRows(),0,c[i]->NumCols()));
+  }
+  BaseFloat time_in_secs = 0.025;
+  Timer tim;
+  int32 iter = 0;
+  for (;tim.Elapsed() < time_in_secs; iter++) {
+    AddMatMatBatched(static_cast<Real>(1.0), C, A, kNoTrans, B, kNoTrans, static_cast<Real>(0.0));
+  }
+  for (int32 i = 0; i< batchCount; i++) {
+    delete a[i]; delete b[i]; delete c[i];
+    delete A[i]; delete B[i]; delete C[i];
+  }
+
+  BaseFloat fdim = dim;
+  BaseFloat gflops = (fdim * fdim * fdim * iter * batchCount) / (tim.Elapsed() * 1.0e+09);
+  KALDI_LOG << "For CuMatrix::AddMatMatBatched" << NameOf<Real>() << ", for dim = " << dim 
+	    << ", batchSize = " << batchCount << ", speed was " << gflops << " gigaflops.";
+}
+
 template<typename Real> void TestCuMatrixAddDiagVecMat(int32 dim, MatrixTransposeType trans) {
   BaseFloat time_in_secs = 0.015;
   CuMatrix<Real> M(dim, dim), N(dim, dim);
@@ -696,6 +728,8 @@ template<typename Real> void CudaMatrixSpeedTest() {
   int32 ns = sizes.size();
   for (int32 s = 0; s < ns; s++)
     TestCuMatrixMatMat<Real>(sizes[s]);
+  for (int32 s = 0; s < ns; s++)
+    TestCuMatrixMatMatBatched<Real>(sizes[s],10);
   for (int32 s = 0; s < ns; s++) {
     TestCuMatrixAddDiagVecMat<Real>(sizes[s], kNoTrans);
     TestCuMatrixAddDiagVecMat<Real>(sizes[s], kTrans);
--- a/src/cudamatrix/cu-matrix-speed-test.cc
+++ b/src/cudamatrix/cu-matrix-speed-test.cc
@@ -69,15 +69,19 @@ template<typename Real> void TestCuMatrixMatMatBatched(int32 dim, int32 batchCou
     c[i] = new CuMatrix<Real>(dim, dim);
     a[i]->SetRandn();
     b[i]->SetRandn();
-    A.push_back(new CuSubMatrix<Real>(*(a[i]),0,a[i]->NumRows(),0,a[i]->NumCols()));
-    B.push_back(new CuSubMatrix<Real>(*(b[i]),0,b[i]->NumRows(),0,b[i]->NumCols()));
-    C.push_back(new CuSubMatrix<Real>(*(c[i]),0,c[i]->NumRows(),0,c[i]->NumCols()));
+    A.push_back(new CuSubMatrix<Real>(*(a[i]), 0, a[i]->NumRows(), 0, 
+			    a[i]->NumCols()));
+    B.push_back(new CuSubMatrix<Real>(*(b[i]), 0, b[i]->NumRows(), 0, 
+			    b[i]->NumCols()));
+    C.push_back(new CuSubMatrix<Real>(*(c[i]), 0, c[i]->NumRows(), 0, 
+			    c[i]->NumCols()));
   }
   BaseFloat time_in_secs = 0.025;
   Timer tim;
   int32 iter = 0;
   for (;tim.Elapsed() < time_in_secs; iter++) {
-    AddMatMatBatched(static_cast<Real>(1.0), C, A, kNoTrans, B, kNoTrans, static_cast<Real>(0.0));
+    AddMatMatBatched(static_cast<Real>(1.0), C, A, kNoTrans, B, kNoTrans, 
+		    static_cast<Real>(0.0));
   }
   for (int32 i = 0; i< batchCount; i++) {
     delete a[i]; delete b[i]; delete c[i];
--- a/src/cudamatrix/cu-matrix.cc
+++ b/src/cudamatrix/cu-matrix.cc
@@ -1700,6 +1700,96 @@ double TraceMatMat(const CuMatrixBase<double> &A,
                    const CuMatrixBase<double> &B,
                    MatrixTransposeType trans);
 
+template<typename Real>
+void AddMatMatBatched(const Real alpha, std::vector<CuSubMatrix<Real>* > &C,
+		const std::vector<CuSubMatrix<Real>* > &A, MatrixTransposeType transA,
+		const std::vector<CuSubMatrix<Real>* > &B, MatrixTransposeType transB,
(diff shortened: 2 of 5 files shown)


src/cudamatrix/cu-matrix.cc
```diff
@@ -1700,6 +1700,93 @@ double TraceMatMat(const CuMatrixBase<double> &A,
...
+  if (CuDevice::Instantiate().Enabled()) {
+    // place all data (A[batchCount], B[batchCount], C[batchCount]) in memory together
+    // in order to make only one call of cudaMemcpy 	  
+    Real **DABC_arr = (Real**)CuDevice::Instantiate().Malloc(3*batchCount*sizeof(Real*));
+    Real **DA_arr = DABC_arr, **DB_arr = DABC_arr + batchCount, **DC_arr = DABC_arr + 2 * batchCount;
+    Real **HABC_arr = (Real**)malloc(3*batchCount*sizeof(Real*));
+    Real **HA_arr = HABC_arr, **HB_arr = HABC_arr + batchCount, **HC_arr = HABC_arr + 2 * batchCount; 
+    
+    for (int32 i = 0; i < batchCount; i++) {
+      memcpy(HA_arr+i, &(A[i]->data_), sizeof(Real*));
```
danpovey: Here, using memcpy is overkill.  Just do e.g.
Ha_arr[i] = A[i]->data_;
Also, although you'll still have to do a const_cast, it would be better to make DB_arr and DC_arr and HB_arr and HC_arr of type const Real**.  
When allocating HABC_arr, better to use the C++ new [] mechanism, e.g.
  Real **HABC_arr = new Real* [3*batchCount];
Also better to use lowercase variables as much as possible, and fewer abbreviations, per Google style guid e.g.  host_abc_array, and device_abc_array.

src/cudamatrix/cu-matrix.cc
```diff
@@ -1700,6 +1700,93 @@ double TraceMatMat(const CuMatrixBase<double> &A,
...
+			    (transA==kTrans? CUBLAS_OP_T:CUBLAS_OP_N), 
+			    m, n, k, alpha, const_cast<const Real**>(DB_arr), 
+			    B[0]->Stride(), const_cast<const Real**>(DA_arr), 
+			    A[0]->Stride(),beta, DC_arr, C[0]->Stride(), 
+			    batchCount));
+
+    CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
+
+    CuDevice::Instantiate().Free(DABC_arr);
+    free(HABC_arr);
```
danpovey: this would be delete [] host_abc_array;

src/cudamatrix/cu-matrix.cc
freewym: Fixed.

freewym:
I have run the test with cuda-memcheck and there is no leak.

One question about rebase again: after I squashed my own commits together locally, and tried to push it to my remote repository using 
git push origin HEAD -u

it gave
error: failed to push some refs to 'https://github.com/freewym/kaldi.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Merge the remote changes (e.g. 'git pull')
hint: before pushing again.

Is it because my local branch and the remote branch diverge after "squash"? If so, should I add "-f" when using push, or is there another way that I can do? 

Thanks,
Yiming

danpovey:
This is normal after doing squash because it is not a fast-forward push
(i.e. the history-tree is not an extension of the old history-tree).  Yes
you can use -f, but be careful that you still have all your changes intact.
Dan

freewym:
Thanks! Done.

src/cudamatrix/cu-matrix-speed-test.cc
```diff
@@ -58,6 +58,38 @@ template<typename Real> void TestCuMatrixMatMat(int32 dim) {
...
+  std::vector<CuSubMatrix<Real>* > A, B, C;
+  
+  for (int32 i = 0; i < batchCount; i++) {
+    // first create a Matrix intance and then creat a SubMatrix instance from that
+    a[i] = new CuMatrix<Real>(dim, dim);
+    b[i] = new CuMatrix<Real>(dim, dim);
+    c[i] = new CuMatrix<Real>(dim, dim);
+    a[i]->SetRandn();
+    b[i]->SetRandn();
+    A.push_back(new CuSubMatrix<Real>(*(a[i]),0,a[i]->NumRows(),0,a[i]->NumCols()));
```
danpovey: there should always be space after comma-- split into multiple lines if needed.

src/cudamatrix/cu-matrix.cc
```diff
@@ -1700,6 +1700,96 @@ double TraceMatMat(const CuMatrixBase<double> &A,
...
+    
+    for (int32 i = 0; i < batchCount; i++) {
+      host_a_array[i] = A[i]->data_;
+      host_b_array[i] = B[i]->data_;
+      host_c_array[i] = C[i]->data_;
+    }
+
+    CU_SAFE_CALL(cudaMemcpy(device_abc_array, host_abc_array, 3*batchCount*sizeof(Real*), cudaMemcpyHostToDevice));
+
+    Timer tim;
```
danpovey: you should probably move this timer to the start of the code block, line 1742.

freewym: Pushed a new commit.

danpovey: OK thanks, I'll merge after the checks finish.
You can move on to the other thing (regarding the batching for the
convolutional component).
Dan

danpovey:
Thanks!  Merged.

src/cudamatrix/cu-matrix.cc
freewym: Hi Dan,

I found 3 member functions in class Convolutional1dComponent having AddMatMat() in a loop: Propagate(), Backprop(), Update(), and have been working on them. I have 3 questions.

1. at line 3902 of nnet2/nnet-component.cc in function Propagate(), should the last argument of AddVecToRows() be 1.0 instead of 0.0?

2.  at line 4178 in function Update(), filters_grad is used to accumulate the sum. What I am doing right now is to compute each term of sum in parallel using AddMatMatBatched(), and then sum over the resulting matrices together to compute the final filters_grad. I was wondering if it is what I was supposed to do.

3. Could you please tell me how to test my code?

Thanks,
Yiming

danpovey: >
>
>
>    1.
>
>    at line 3902 of nnet2/nnet-component.cc in function Propagate(),
>    should the last argument of AddVecToRows() be 1.0 instead of 0.0?
>
> It doesn't matter, since that is the scale on something that's zero
anyway.

>
>    1.
>
>    at line 4178 in function Update(), filters_grad is used to accumulate
>    the sum. What I am doing right now is to compute each term of sum in
>    parallel using AddMatMatBatched(), and then sum over the resulting matrices
>    together to compute the final filters_grad. I was wondering if it is what I
>    was supposed to do.
>
> Firstly, make sure you allocate a single large matrix, with a large number
of rows (possibly better for alignment reasons), rather than separately
allocating a bunch of small matrices.
Rather than adding the small matrices together manually using repeated
calls to AddMatMat, please write a CuMatrix function as follows:

  /// This function is like AddMat (does *this += alpha * src), but handles
the case
  /// where src.NumRows() may be a multiple of this->NumRows(); in that case
  /// it divides 'src' into blocks of size equal to '*this' and adds each
of them (times alpha)
  /// to *this.
  void AddMatBlocks(Real alpha, const CuMatrixBase<Real> &src);

This can call a custom kernel- you can modify it from the one that AddMat
currently calls.  For now it's not handling transposing, or the case where
the num-cols is also a multiple, so it shouldn't be too hard to write.

You can then use this function to avoid the loop over AddMat.
Add some test code for AddMatBlocks in cu-matrix-test.cc, and speed test
for it in cu-matrix-speed-test.cc.

For the bias_grad, you don't have to do anything with a loop, you can just
compute it as follows using a single call:
 bias_grad.AddRowSumMat(1.0, out_deriv, 1.0);

Testing will automatically happen in the nnet-component-test.cc, you
shouldn't have to write anything extra.  But make your changes to the
forward and backward propagation one by one rather than at the same time,
and make sure the tests run-- because the test compare the forward and
backward passes and check the derivatives, it's easier to change things one
at a time so you know where the problem is.

Dan

danpovey: BTW, there doesn't need to be any fancy logarithmic reduction  in
AddMatBlocks-- just compute the sum using a simple for-loop.
Dan

freewym: Thanks. But one thing I am a bit confused: according to the code, the length of bias_grad is num_filters, but the numcols of out_derive is num_patches*num_filters. How to add the row sum of out_derive to bias_grad?

naxingyu: out_deriv is of (num_frames) x (num_filters * num_patches) . Perhaps you could reshape it to such a temp variable.
// This is a block-wise transpose of out_deriv
out_deriv_reshape(num_frames * num_patches, num_filters);
and do AddRowSumMat once for all.
Does it make sense?

danpovey: This kind of reshaping is not possible in general because of memory
alignment reasons-- in general, there is a gap between successive rows of a
matrix.
Dan

PR #131, opened 2015-09-09
Fixed steps/align_si.sh to honour the delta calculation options of the model

alumae:
Copy-paste from on steps/align_fmllr.sh


danpovey:
Thanks!

PR #138, opened 2015-09-14
Limit deriv times

danpovey:
This is a pull request for my changes to limit the times for which derivatives are computed in BPTT.  (It's not as simple as you might think, since the framework itself does the derivative computation for you).  I plan to merge once the checks finish, e.g. in 1 hour.
I have now debugged and tested this on Pegah's setup, and it's both faster (e.g. 1 minute 15 seconds -> 1 minute) and seems to give better objective function values.
The option that I added when testing this is --optimization.min-deriv-time=-7.  You may in general want to set this to the negative of the left-context of the network, or a bit more negative than that.  But the left-context of the network is a reasonable default.  You'll have to change the scripts to work with this.  Basically, any derivative quantities computed on times before that will be considered to be zero by the framework and won't be computed, and hence the model won't be updated for those times.
If you don't add any extra command line options the code behaves basically as before.
@vijayaditya  and @pegahgh please take note!  Also of course @naxingyu and @nichongjia.

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -213,8 +213,6 @@ void CuDevice::FinalizeActiveGpu() {
     KALDI_LOG << "The active GPU is [" << act_gpu_id << "]: " << name << "\t"
               << GetFreeMemory(&free_memory_at_startup_, NULL) << " version "
               << properties_.major << "." << properties_.minor;
-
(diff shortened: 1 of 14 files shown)


danpovey:
The first build failed in a test and I made some bug fixes.  Hope to check in this evening.

danpovey:
OK, after a few iterations all the tests pass and I am merging this.

PR #143, opened 2015-09-16
nnet3, LSTM scripts: add momentum, and remove --update-per-minibatch …

danpovey:
…(no longer needed due to optimization that has the same effect).  Caution: default=0.9, so it's on by default in the LSTM scripts.


danpovey:
I tested this on the command line with an existing setup, and it looks promising (slightly better likelihood than before, and no crash).  
I expect after this change, you will be able to increase the learning rate considerably without it crashing.

danpovey:
I changed this commit so momentum is no longer the default in the LSTM script (i.e. default is 0.0, the same as the baseline), and I'm merging.

PR #145, opened 2015-09-17
Adding nnet2 function NnetComputationChunked

david-ryan-snyder:
Adding nnet2 function NnetComputationChunked and updating associated unit test and binary. This version of NnetComputation divides the feature matrix into smaller chunks, and performs the posterior computation chunk by chunk. This makes the computation suitable for the GPU when the feature matrices are large.

### DIFF
--- a/src/nnet2/nnet-compute-test.cc
+++ b/src/nnet2/nnet-compute-test.cc
@@ -1,6 +1,7 @@
 // nnet2/nnet-compute-test.cc
 
 // Copyright 2014  Johns Hopkins University (author:  Daniel Povey)
+// Copyright 2015  David Snyder
 
 // See ../../COPYING for clarification regarding multiple authors
 //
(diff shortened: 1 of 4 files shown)


naxingyu:
Cool. We should have this in nnet3 too because the progress jobs are 
running reeeally slow on CPUs in lstm setups.

danpovey:
Actually this wouldn't help the nnet3 show-progress.  If speed is a problem
you can just make it run on GPU (i.e. change the queue options that it's
submitted with, and add --use-gpu=yes to the command line).  We don't do
this in order to not slow down the training.
Dan

naxingyu:
I tried --use-gpu=yes option but it results in out-of-memory crash.

danpovey:
It works fine for me.  Maybe your GPUs don't have much memory?
Dan

naxingyu:
I'm using Tesla K20m in exclusive mode... it's not a cutting-edge 
version but should be enough.

danpovey:
Same here.    What is the command line?
Dan

danpovey:
You might as well show the output too.
Dan

david-ryan-snyder:
Thanks for the merge. Should we also do something like this for the nnet3 equivalent?

danpovey:
The nnet3 computation (IIRC) is already done in chunks.
Dan


On Thu, Sep 17, 2015 at 5:27 PM, david-ryan-snyder <notifications@github.com

PR #149, opened 2015-09-17
Added --scale option to nnet3-am-copy, useful for shrinkage of parame…

vijayaditya:
…ter matrices

Removed minor bug in nnet3-train.cc

Note: Still testing usefulness of shrinkage in lstm recipe.


danpovey:
Looks like the test failure is in logistic-regression-test and unrelated to your changes- rather confusing since that program is quite simple.
But can you please change it in a separate pull request so that the default 'scale' option that causes nothing to happen, is 1.0 (in the script and program)?  It seems to me that this is easier to explain, as scaling by 1.0 naturally does nothing.

Issue #150, opened 2015-09-18
logistic-regression bug


danpovey:
@david-ryan-snyder, I found a bug in the logistic-regression code.
This was after a test failure in the logistic-regression-test.  It's not repeatable because the logistic-regression-test.cc does srand(time()),  but if you put a loop inside logistic-regression-test.cc you'll get an error in free().

If I put the loop in logistic-regression-test.cc and then run 
    valgrind --db-attach=yes logistic-regression-test

I get the following error:

==28965== Invalid write of size 4
==28965==    at 0x4501DE: kaldi::UnitTestTrain() (logistic-regression-test.cc:92)
==28965==    by 0x4505D1: main (logistic-regression-test.cc:128)
==28965==  Address 0x73d9b54 is 4 bytes after a block of size 17,680 alloc'd
==28965==    at 0x4C270FE: memalign (vg_replace_malloc.c:694)
==28965==    by 0x4C271A7: posix_memalign (vg_replace_malloc.c:835)
==28965==    by 0x46E5EE: kaldi::Matrix<float>::Init(int, int) (kaldi-matrix.cc:656)
==28965==    by 0x46E2F0: kaldi::Matrix<float>::Resize(int, int, kaldi::MatrixResizeType) (kaldi-matrix.cc:702)
==28965==    by 0x450932: kaldi::Matrix<float>::Matrix(int, int, kaldi::MatrixResizeType) (kaldi-matrix.h:744)
==28965==    by 0x4501B4: kaldi::UnitTestTrain() (logistic-regression-test.cc:90)
==28965==    by 0x4505D1: main (logistic-regression-test.cc:128)
==28965==
==28965==

0x00000000004501de in kaldi::UnitTestTrain () at logistic-regression-test.cc:92
92      xs_with_prior(i, n_xs) = 1.0;
(gdb) list
87
88    // Internally in LogisticRegression we add an additional element to
89    // the x vectors: a 1.0 which handles the prior.
90    Matrix<BaseFloat> xs_with_prior(n_xs, n_features + 1);
91    for (int32 i = 0; i < n_xs; i++) {
92      xs_with_prior(i, n_xs) = 1.0;
93    }
94    SubMatrix<BaseFloat> sub_xs(xs_with_prior, 0, n_xs, 0, n_features);
95    sub_xs.CopyFromMat(xs);
96
(gdb) p n_xs
$1 = 221

I think maybe you mean xs_with_prior(i, n_features) = 1.0, not n_xs.

[...]

PR #151, opened 2015-09-18
Fixing bug in ivector/logistic-regression-test.cc

david-ryan-snyder:
Also removing trailing whitespace from the logistic regression cc and h files.

This fix should close issue 150: https://github.com/kaldi-asr/kaldi/issues/150


danpovey:
Thanks- will wait till checks complete and then merge.
Dan


On Thu, Sep 17, 2015 at 9:26 PM, david-ryan-snyder <notifications@github.com

PR #152, opened 2015-09-18
Update of the recipe of Corpus of Spontaneous Japanese 

sw005320:
- Mainly bug fix and update the results


jtrmal:
Why did you change the /usr/bin/env perl to /usr/bin/perl? That will cause problems for people running the scripts with local (non-system) installation of perl.
y.

danpovey:
It would be better if you could avoid having the merge of master as a commit there.  In general it's better to rebase branches you're going to use as a commit.  In fact, if you rebase against the master right now it's possible that it will make the merge commit disappear.

sw005320:
OK.
I will do it again from the branch.
I will also fix the perl issue.
Maybe, I would do it by the new pull request

jtrmal:
Yes, you will have to open a new PR (as you will use a different branch).
Closing this PR for now.
y.

PR #153, opened 2015-09-18
Modified virtual functions Scale and Add 

keli78:
Modified virtual functions Scale and Add in the NnetComponent base and child classes in nnet3/


danpovey:
I think your Add() and Scale() functions will not work quite correctly if you leave your accumulators as int32.  Better to make them doubles.
Also the constructor of  NaturalGradientAffineComponent should be moved to the .cc file as it's no longer as short.

PR #154, opened 2015-09-18
added max-change related stats in NaturalGradientAffineComponent

xiaohui-zhang:


xiaohui-zhang:
ok fixed that

src/nnet3/nnet-simple-component.cc
```diff
@@ -1308,12 +1343,34 @@ void NaturalGradientAffineComponent::Update(
   precon_ones.CopyColFromMat(in_value_temp, in_value_temp.NumCols() - 1);
 
   BaseFloat local_lrate = scale * minibatch_scale * learning_rate_;
+  update_count_ += 1;
```
danpovey: make this say update_count_ += 1.0;
(and change any similar places)... always use floating-point constants in a floating-point context.
Also please rebase your changes so the fixes don't appear in the commit history.  rebase -i HEAD~3 and changing the last 2 lines from 'pick' to 'f' should do it (after you commit this fix).

xiaohui-zhang:
fixed floating-point constants. Thanks

danpovey:
cool.  I'll wait till checks complete and then merge.

PR #156, opened 2015-09-20
Add Convolutional1dComponent in nnet3 and its test code

freewym:
It passed the test code.

### DIFF
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -2160,7 +2160,7 @@ void Convolutional1dComponent::Propagate(const ComponentPrecomputedIndexes *inde
 				    filter_dim)));
     filter_params_batch.push_back(filter_params_elem);
 
-    tgt_batch[p]->AddVecToRows(1.0, bias_params_, 0.0); // add bias
+    tgt_batch[p]->AddVecToRows(1.0, bias_params_, 1.0); // add bias
   }
   
   // apply all filters
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -1930,6 +1930,592 @@ void NaturalGradientPerElementScaleComponent::Update(
   scales_.AddVec(max_change_scale, delta_scales);
 }
 
+Convolutional1dComponent::Convolutional1dComponent():
+    UpdatableComponent(),
+    patch_dim_(0), patch_step_(0), patch_stride_(0), is_gradient_(false) {}
+
+Convolutional1dComponent::Convolutional1dComponent(const Convolutional1dComponent &component):
+    UpdatableComponent(component),
+    filter_params_(component.filter_params_),
+    bias_params_(component.bias_params_),
+    is_gradient_(component.is_gradient_) {}
+
+Convolutional1dComponent::Convolutional1dComponent(const CuMatrixBase<BaseFloat> &filter_params,
+                                                   const CuVectorBase<BaseFloat> &bias_params,
+                                                   BaseFloat learning_rate):
+    UpdatableComponent(learning_rate),
+    filter_params_(filter_params),
+    bias_params_(bias_params) {
+  KALDI_ASSERT(filter_params.NumRows() == bias_params.Dim() &&
+               bias_params.Dim() != 0);
+  is_gradient_ = false;
+}
+
+// aquire input dim
+int32 Convolutional1dComponent::InputDim() const {
+  int32 filter_dim = filter_params_.NumCols();
+  int32 num_splice = filter_dim / patch_dim_;
+  return patch_stride_ * num_splice;
+}
+
+// aquire output dim
+int32 Convolutional1dComponent::OutputDim() const {
+  int32 num_filters = filter_params_.NumRows();
+  int32 num_patches = 1 + (patch_stride_ - patch_dim_) / patch_step_;
+  return num_patches * num_filters;
+}
+
+// initialize the component using hyperparameters
+void Convolutional1dComponent::Init(BaseFloat learning_rate,
+                                    int32 input_dim, int32 output_dim,
+                                    int32 patch_dim, int32 patch_step, int32 patch_stride,
+                                    BaseFloat param_stddev, BaseFloat bias_stddev) {
+  UpdatableComponent::Init(learning_rate);
+  patch_dim_ = patch_dim;
+  patch_step_ = patch_step;
+  patch_stride_ = patch_stride;
+  int32 num_splice = input_dim / patch_stride;
+  int32 filter_dim = num_splice * patch_dim;
+  int32 num_patches = 1 + (patch_stride - patch_dim) / patch_step;
+  int32 num_filters = output_dim / num_patches;
+  KALDI_ASSERT(input_dim % patch_stride == 0);
+  KALDI_ASSERT((patch_stride - patch_dim) % patch_step == 0);
+  KALDI_ASSERT(output_dim % num_patches == 0);
+
+  filter_params_.Resize(num_filters, filter_dim);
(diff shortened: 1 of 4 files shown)


src/nnet3/nnet-test-utils.cc
```diff
@@ -804,6 +804,18 @@ static void GenerateRandomComponentConfig(std::string *component_type,
       os << "input-dim=" << input_dim << " output-dim=" << output_dim;
       break;
     }
+    case 17: {
+      *component_type = "Convolutional1dComponent";
+      int32 patch_stride = 10, patch_step = 1, patch_dim = 4;
```
danpovey: I think it would be better to test a variety of values here- this may be masking bugs.

src/nnet3/nnet-simple-component.h
```diff
@@ -891,6 +891,131 @@ class NaturalGradientPerElementScaleComponent: public PerElementScaleComponent {
...
+
+  virtual int32 InputDim() const;
+  virtual int32 OutputDim() const;
+
+  virtual std::string Info() const;
+  virtual void InitFromConfig(ConfigLine *cfl);
+  virtual std::string Type() const { return "Convolutional1dComponent"; }
+  virtual int32 Properties() const {
+    return kSimpleComponent|kUpdatableComponent|kBackpropNeedsInput;
+	   // |kBackpropAdds;
```
danpovey: Remove this comment.

src/nnet3/nnet-simple-component.h
```diff
@@ -891,6 +891,131 @@ class NaturalGradientPerElementScaleComponent: public PerElementScaleComponent {
...
+  }
+
+  virtual void Propagate(const ComponentPrecomputedIndexes *indexes,
+                 const CuMatrixBase<BaseFloat> &in,
```
danpovey: fix indentation

src/nnet3/nnet-simple-component.cc
```diff
@@ -1930,6 +1930,594 @@ void NaturalGradientPerElementScaleComponent::Update(
...
+  }
+
+  if (in_deriv) {
+    std::vector<std::vector<int32> > reversed_column_map;
+    ReverseIndexes(column_map, InputDim(), &reversed_column_map);
+    std::vector<std::vector<int32> > rearranged_column_map;
+    RearrangeIndexes(reversed_column_map, &rearranged_column_map);
+    for (int32 p = 0; p < rearranged_column_map.size(); p++) {
+      CuArray<int32> cu_cols(rearranged_column_map[p]);
+      in_deriv->AddCols(patches_deriv, cu_cols);
```
danpovey: I think maybe this should be CopyCols.  Actually I'm very surprised that the tests passed because this component seems to add to its input and never zero it, yet the properties says that it just sets its input. TestSimpleComponentPropagateProperties() should have failed.  [see output_deriv2].

danpovey: OK, I see now that AddCols is appropriate, because you're calling it multiple times.  But I don't see that you ever zero in_deriv.  Ooh.. I see now.

freewym: It's at the first line of Backprop when doing resizing

src/nnet3/nnet-simple-component.cc
```diff
@@ -1930,6 +1930,594 @@ void NaturalGradientPerElementScaleComponent::Update(
...
+// back propagation function
+void Convolutional1dComponent::Backprop(const std::string &debug_info,
+                                        const ComponentPrecomputedIndexes *indexes,
+                                        const CuMatrixBase<BaseFloat> &in_value,
+                                        const CuMatrixBase<BaseFloat> &out_value,
+                                        const CuMatrixBase<BaseFloat> &out_deriv,
+                                        Component *to_update_in,
+                                        CuMatrixBase<BaseFloat> *in_deriv) const {
+  if (in_deriv)
+    static_cast<CuMatrix<BaseFloat>* >(in_deriv)->Resize(out_deriv.NumRows(), InputDim());
```
danpovey: Ouch!!  You should never static_cast CuMatrixBase to CuMatrix like this!!!  How could this code possibly work if it really pointed to a sub-matrix?  And it will, if it's ever used in a real neural net.    I hope you didn't do this in any previous code.
The code requires that in_deriv already have the correct size.  If you want to zero it you can call SetZero().
But actually, don't call SetZero().  The framework is designed so that if it's more natural to add to the derivative, you just set the flag kBackpropAdds, and it's then the responsibility of the calling code to zero the derivative (if needed).  That way, if the derivative needs to be added to something else, you may be able to avoid redundant copies of matrices.  So add the flag kBackpropAdds, and remove this statement.

freewym: Sorry about that. I did the static_cast because  Backprop() of CNN in nnet2 calls Resize() at the beginning. I will fix it.

freewym:
I fixed the those bugs. Also I removed the call of Resize() at the beginnig of Convolutional1dComponent::Backprop() in nnet2 and changed a line in its nnet-component-test.cc accordingly. Please let me know if there are any problems.
.
Yiming

src/nnet3/nnet-test-utils.cc
```diff
@@ -804,6 +804,24 @@ static void GenerateRandomComponentConfig(std::string *component_type,
       os << "input-dim=" << input_dim << " output-dim=" << output_dim;
       break;
     }
+    case 17: {
+      *component_type = "Convolutional1dComponent";
+      int32 patch_stride = 10 + Rand() % 50, patch_step = 1 + Rand() % 4, 
+	    patch_dim = 4 + Rand () % 5;
+
+      // decrease patch_stride until (patch_stride - patch_dim) % patch_step == 0
+      while ((patch_stride - patch_dim) % patch_step != 0)
```
freewym: Sorry. no while needed. will replace it with a single expression.

danpovey: In nnet2, in_deriv is supposed to be resized by the Backprop routine, that
it why it is typed CuMatrix not CuMatrixBase.  In nnet2, it's different.

freewym: Restored.

danpovey: OK.  Please rebase your changes so that only the initial commit remains.
Dan

freewym: Done. One issue: I left the previous 2 commits (by david and vijay) in the interactive editor when using rebase, so currently there are 3 commits in this pull request. But actually I didn't modify anything in their commits. I am not sure if it is fine for you to merge only mine. If not, I can create a new pull request. Sorry about it.
Yiming

danpovey: You must have done something wrong-- those shouldn't appear there.  See if
you can find someone to help you fix it.  To-morrow is fine.
Meanwhile, you could also try to add a test example in nnet-test-utils.cc
that tests the convolutional component.
Dan

freewym: I fixed that. Now there is only one commit of mine. 
By a test example, did you mean a NN with Convolutional1dComponent, rather than only testing a component? Could you please refer me to a similar example in nnet-test-utils.cc or somewhere else?
Thanks,
Yiming

danpovey: See GenerateConfigSequence and the functions it calls.

danpovey: You can add that to this pull request.

danpovey:
Thanks!  Merging.
I believe in typical convolutional setups there is some kind pooling component (e.g. max-pooling) after each convolutional component.  Perhaps you could look into this, and port any max-pooling code to nnet3 (if necessary), and modify the test config accordingly?

PR #157, opened 2015-09-20
test code for ApplyCeiling

keli78:
Add the test code for ApplyCeiling


danpovey:
The content is fine, but your commit history is strange.  To-morrow, ask someone to help you sort it out so there is just one commit.  E.g. you could generate a patch and apply it to a new copy of the master.

PR #158, opened 2015-09-21
Added gaurd in feat/signal.h, removed bug in NaturalGradientAffineCom…

vijayaditya:
…ponent::Scale

NGAffineComponent Bug:
The NGAffineComponent overrode the Scale() method but did call the parent Scale method. So the linear_params_ and bias_params_ were not being scaled.

@naxingyu @xiaohui-zhang @pegahgh This changes requires that all shrinkage experiments be re-run.

### DIFF
--- a/src/nnet3bin/nnet3-am-copy.cc
+++ b/src/nnet3bin/nnet3-am-copy.cc
@@ -48,7 +48,7 @@ int main(int argc, char *argv[]) {
         raw = false;
     BaseFloat learning_rate = -1;
     std::string set_raw_nnet = "";
-    BaseFloat scale = 0.0;
+    BaseFloat scale = 1.0;
 
     ParseOptions po(usage);
     po.Register("binary", &binary_write, "Write output in binary mode");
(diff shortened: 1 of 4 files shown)


naxingyu:
Got it.

src/nnet3bin/nnet3-am-copy.cc
```diff
@@ -89,18 +89,17 @@ int main(int argc, char *argv[]) {
...
       am_nnet.SetNnet(nnet);
     }
-    
+
     if (learning_rate >= 0)
       SetLearningRate(learning_rate, &(am_nnet.GetNnet()));
 
-    if (scale != 0.0)
-      ScaleNnet(scale, &(am_nnet.GetNnet()));
+    ScaleNnet(scale, &(am_nnet.GetNnet()));
```
danpovey: Mostly for clarity, also for efficiency, would be good to have
if (scale != 1.0)
here.  Will merge anyway though.

danpovey:
BTW, since it turns out we weren't even shrinking the most important components, but shrinking was still having a big effect, probably only the less aggressive shrinking scales will make sense (e.g. 0.95, and maybe even 0.98).  I hope the thing will be implemented soon where the shrinking scale varies with the learning rate.

PR #164, opened 2015-09-21
Avoid crash if wave remainer is not specified in caller

nshmyrev:
API allows to skip wave_remainder parameter in the call, while on some cases it can cause crash.


danpovey:
Thanks!

PR #167, opened 2015-09-21
Add the test code for ApplyCeiling and remove unnecessary null pointer checks 

keli78:
Add the test code for ApplyCeiling
Remove unnecessary null pointer checks


danpovey:
I'm very confused by this commit as the 'files changed' includes all those null-pointer checks that have already been committed.  Something has gone wrong with the history again.  Make sure you do 'git fetch' on the upstream before rebasing.
Dan

PR #169, opened 2015-09-21
Csj egs

sw005320:
In the CSJ recipe
- Fixed minor bugs
- Tuned some parameters, and updated the results
- Fixed "#!/usr/bin/env perl" issues


danpovey:
@jtrmal, I don't have time to check this right now, so I'll wait for you to merge this.

jtrmal:
Thanks for the update. I'm ok with merge.
BTW -- it's generally suggested (in the google style guide) to avoid spaces
at the end of the lines -- such as you have in the path.sh on line
+#export LC_ALL=C

y.

danpovey:
To fix spaces at the ends of lines, you can put in your .emacs
(add-hook 'write-file-hooks 'delete-trailing-whitespace)

sw005320:
OK
I will do it!
(Maybe I could not do all of them)

jtrmal:
OK, merging. It still would be great if you could fix the white spaces.
Thanks for maintaining the recipe!
y.

PR #171, opened 2015-09-22
Add Maxpooling in nnet3 and its test code, modify the test example fo…

freewym:
…r Cnn to add a maxpooling layer

All the tests are passed


danpovey:
Thanks!

PR #174, opened 2015-09-23
Removed AddNnet(delta_nnet_) call in NnetTrainer destructor. This was…

vijayaditya:
… critical for making momentum work.


danpovey:
would you mind removing the if-statement?
recently I had Ke remove all the if's before deletes, as C++ does that for
you.
Dan

vijayaditya:
done

PR #180, opened 2015-09-25
added adaptive shrinking in lstm recipe

vijayaditya:
This PR is not ready for merge, it has been created to discuss the adaptive shrinkage schedule.

### DIFF
--- a/egs/ami/s5/local/nnet3/run_lstm.sh
+++ b/egs/ami/s5/local/nnet3/run_lstm.sh
@@ -1,10 +1,18 @@
 #!/bin/bash
 
+# Copyright 2015  Johns Hopkins University (Author: Daniel Povey).
+#           2015  Vijayaditya Peddinti
+#           2015  Xingyu Na
+#           2015  Pegah Ghahrmani
+# Apache 2.0.
+
+
 # this is a basic lstm script
+# LSTM script runs for more epochs than the TDNN script
+# and each epoch takes twice the time
 
 # At this script level we don't support not running on GPU, as it would be painfully slow.
 # If you want to run without GPU you'd have to call lstm/train.sh with --gpu false
-set -e
 
 stage=0
 train_stage=-10
@@ -13,6 +21,9 @@ mic=ihm
 use_sat_alignments=true
 affix=
 speed_perturb=true
+common_egs_dir=
+
+# LSTM options
 splice_indexes="-2,-1,0,1,2 0 0"
 label_delay=5
 num_lstm_layers=3
@@ -21,27 +32,31 @@ hidden_dim=1024
 recurrent_projection_dim=256
 non_recurrent_projection_dim=256
 chunk_width=20
-chunk_left_context=20
+chunk_left_context=40
 clipping_threshold=30.0
 norm_based_clipping=true
-common_egs_dir=
 
 # natural gradient options
 ng_per_element_scale_options=
 ng_affine_options=
-num_epochs=5
+num_epochs=10
 
 # training options
 initial_effective_lrate=0.0003
 final_effective_lrate=0.00003
 num_jobs_initial=2
 num_jobs_final=12
-momentum=0.0
-shrink=0.99
+momentum=0.5
+adaptive_shrink=true
+shrink=0.98
 num_chunk_per_minibatch=100
 num_bptt_steps=20
 samples_per_iter=20000
 remove_egs=true
+
+#decode options
+extra_left_context=
+frames_per_chunk=
 
 # End configuration section.
 
@@ -103,6 +118,7 @@ if [ $stage -le 8 ]; then
     --cmvn-opts "--norm-means=false --norm-vars=false" \
     --initial-effective-lrate $initial_effective_lrate --final-effective-lrate $final_effective_lrate \
     --momentum $momentum \
+    --adaptive-shrink "$adaptive_shrink" \
     --shrink $shrink \
     --cmd "$decode_cmd" \
     --num-lstm-layers $num_lstm_layers \
@@ -123,15 +139,20 @@ if [ $stage -le 8 ]; then
 fi
 
 if [ $stage -le 9 ]; then
-  # this version of the decoding treats each utterance separately
-  # without carrying forward speaker information.
+  if [ -z $extra_left_context ]; then
+    extra_left_context=$chunk_left_context
+  fi
+  if [ -z $frames_per_chunk ]; then
+    frames_per_chunk=$chunk_width
+  fi
   for decode_set in dev eval; do
       (
       num_jobs=`cat data/$mic/${decode_set}_hires/utt2spk|cut -d' ' -f2|sort -u|wc -l`
       decode_dir=${dir}/decode_${decode_set}
 
       steps/nnet3/lstm/decode.sh --nj 250 --cmd "$decode_cmd" \
-          --extra-left-context $chunk_left_context  \
+          --extra-left-context $extra_left_context  \
+          --frames-per-chunk "$frames_per_chunk" \
           --online-ivector-dir exp/$mic/nnet3/ivectors_${decode_set} \
          $graph_dir data/$mic/${decode_set}_hires $decode_dir || exit 1;
       ) &
--- a/egs/wsj/s5/steps/nnet3/lstm/train.sh
+++ b/egs/wsj/s5/steps/nnet3/lstm/train.sh
@@ -14,19 +14,20 @@
 
 # Begin configuration section.
 cmd=run.pl
-num_epochs=5      # Number of epochs of training;
+num_epochs=10      # Number of epochs of training;
                    # the number of iterations is worked out from this.
 initial_effective_lrate=0.0003
 final_effective_lrate=0.00003
+adaptive_shrink=true     # if true, the shrink value is increased to 1 with a rate
+                         # inversely proportional to the effective_learning_rate
 shrink=0.99  # this parameter would be used to scale the parameter matrices
-rand_prune=4.0 # Relates to a speedup we do for LDA.
 num_chunk_per_minibatch=100  # number of sequences to be processed in parallel every mini-batch
 
 samples_per_iter=20000 # each iteration of training, see this many samples
                        # per job.  This option is passed to get_egs.sh
                        # this is substantially smaller than of that used for FF-DNN
                        # as multiple output labels are processed in each sample
-num_jobs_initial=1  # Number of neural net jobs to run in parallel at the start of training
+num_jobs_initial=1 # Number of neural net jobs to run in parallel at the start of training
 num_jobs_final=8   # Number of neural net jobs to run in parallel at the end of training
 prior_subset_size=20000 # 20k samples per job, for computing priors.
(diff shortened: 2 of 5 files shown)


egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -451,6 +453,12 @@ while [ $x -lt $num_iters ]; do
   ilr=$initial_effective_lrate; flr=$final_effective_lrate; np=$num_archives_processed; nt=$num_archives_to_process;
   this_learning_rate=$(perl -e "print (($x + 1 >= $num_iters ? $flr : $ilr*exp($np*log($flr/$ilr)/$nt))*$this_num_jobs);");
 
+  if [ "$adaptive_shrink" == "true" ]; then
+    this_shrink=$(perl -e "print $shrink + ($flr*$this_num_jobs/$this_learning_rate) * (1 - $shrink)")
```
danpovey: I'm not convinced this is right.  We want (1-shrink) to vary proportional to this_learning_rate.   I initially intended that the specified shrinkage be the initial one.
So 
  this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / ($num_jobs_initial * $ilr)) * (1 - $shrink));")

danpovey: Actually, after thinking about this some more, I think it might be better to make it
this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr)) * (1 - $shrink));")
This makes the shrink setting independent of the initial num-jobs, so that you could change the initial num-jobs without materially affecting the shrinkage rate.

vijayaditya: this_learning_rate is a function of the num_initial_jobs, so to make this_shrink independent of num_initial_jobs we actually need

this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / ($num_jobs_initial * $ilr)) * (1 - $shrink));")

danpovey: The way you are doing it would make the initial shrinking rate independent of the num-jobs-initial, that's true, but it would make the final shrinking rate highly dependent on num-jobs-initial too.  So, for instance, between num-jobs-initial=1 and num-jobs-initial=2 there would be a factor of 2 difference in the final-probs.  I think it's more appropriate to do it the way I proposed, as in
 this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr)) * (1 - $shrink));").
because this way, as we change the initial-num-jobs, the shrinking rate increases, which is actually appropriate because the initial actual learning rate is also increasing.  [this is in line with l2 regularization interpretation].

vijayaditya: In addition to the rate of change, I am also worried about the initial shrink value. This, I believe, would affect learning to a greater extent. Further the shrink value is not going to reach 1 towards the end of the training in the above equation. We actually wanted $shrink to reach 1 as this would ensure we finally are optimizing the actual objective function.

So though we do not have an l2 regularization interpretation, how about keeping the original shrink value formula based on the effective_learning_rate rather than actual_learning_rate 
<code>
 this_shrink=$(perl -e "print $shrink + ($flr/($this_learning_rate/$this_num_jobs)) * (1 - $shrink)")
</code>

vijayaditya: I take back my statement that initial shrink value is more important than rate.. I was looking at the results for various shrink values (See below plot) I think rate would be more important. So I implementing your formula..

shrink_vs_cvaccuracy

danpovey: I think both my formula
 this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr) * (1 - $shrink));")
which has an l2 regularization interpretation, and your formula
 this_shrink=$(perl -e "print $shrink + ($flr/($this_learning_rate/$this_num_jobs)) * (1 - $shrink)") 
which allows the shrinkage value to approach 1 (and might be suitable for larger data sets)
should be experimented with.  But I think your formula could be clarified by modifying the existing script to set a variable this_effective_lrate first, and then set a variable this_learning_rate afterwards; you can then use $this_effective_lrate in your formula instead of that ratio.

danpovey: That graph could be interpreted as telling us that 0.95 shrinkage was too
high at the end, but 0.95, 0.98 and 0.99 were all in a reasonable range at
the beginning.  So it could be also be interpreted as saying that it's
important to aggressively decrease the shrinkage towards the end.  Anyway,
up to you.

vijayaditya: OK, I will add both these till we have results on different databases. Looking at the above plot I think I will make your formula the default one.

vijayaditya:
ready for merge.. current default option is no adaptive_shrinkage which reproduces the results with the existing lstm recipe.

egs/ami/s5/local/nnet3/run_lstm.sh
```diff
@@ -102,6 +103,7 @@ if [ $stage -le 8 ]; then
     --cmvn-opts "--norm-means=false --norm-vars=false" \
     --initial-effective-lrate $initial_effective_lrate --final-effective-lrate $final_effective_lrate \
     --momentum $momentum \
+    --adaptive-shrink $adaptive_shrink \
```
danpovey: You need to have "$adaptive_shrink" or this won't parse right.

danpovey:
Could you please do some experiments before this is merged?  I don't want
to fill up the script with all kinds of options.  Let's just figure out
which schedule to use and then just make that the default.
Dan

vijayaditya:
I am testing out the two types of shrinkage schedules. We will have some
results in 10 hours.

Vijay

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -449,9 +459,21 @@ while [ $x -lt $num_iters ]; do
...
 
   ilr=$initial_effective_lrate; flr=$final_effective_lrate; np=$num_archives_processed; nt=$num_archives_to_process;
-  this_learning_rate=$(perl -e "print (($x + 1 >= $num_iters ? $flr : $ilr*exp($np*log($flr/$ilr)/$nt))*$this_num_jobs);");
+  this_effective_learning_rate=$(perl -e "print ($x + 1 >= $num_iters ? $flr : $ilr*exp($np*log($flr/$ilr)/$nt));");
+  this_learning_rate=$(perl -e "print ($this_effective_learning_rate*$this_num_jobs);");
+  if [ z $adaptive_shrink ]; then
+    this_shrink=$shrink
+  else
+    if [ "$adaptive_shrink" == "type1" ]; then
+      this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr) * (1 - $shrink));")
```
vijayaditya: this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr) * (1 - $shrink));")
which depends on the actual learning rate leads to zigzag and varying shrink values even for $shrink=0.95 and 0.98. Both the below training runs diverge

shrink_vs_cvaccuracy_iteration

vijayaditya: red corresponds to $shrink=0.98 and blue 0.95

vijayaditya: (rewriting the comment as it was hidden due to an updated pull request)

this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr) * (1 - $shrink));")
which depends on the actual learning rate leads to zigzag and highly varying shrink values even for $shrink=0.95 and 0.98. Both the below training runs diverge

red corresponds to $shrink=0.98 and blue 0.95
Plot1 shows shrink values
and plot2 cv accuracies.


shrink_vs_cvaccuracy_iteration

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -449,9 +459,21 @@ while [ $x -lt $num_iters ]; do
...
 
   ilr=$initial_effective_lrate; flr=$final_effective_lrate; np=$num_archives_processed; nt=$num_archives_to_process;
-  this_learning_rate=$(perl -e "print (($x + 1 >= $num_iters ? $flr : $ilr*exp($np*log($flr/$ilr)/$nt))*$this_num_jobs);");
+  this_effective_learning_rate=$(perl -e "print ($x + 1 >= $num_iters ? $flr : $ilr*exp($np*log($flr/$ilr)/$nt));");
+  this_learning_rate=$(perl -e "print ($this_effective_learning_rate*$this_num_jobs);");
+  if [ -z "$adaptive_shrink" ]; then
+    this_shrink=$shrink
+  else
+    if [ "$adaptive_shrink" == "type1" ]; then
+      this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr) * (1 - $shrink));")
```
vijayaditya: (rewriting the comment as it was hidden due to an updated pull request)

this_shrink=$(perl -e "print 1.0 - (($this_learning_rate / $ilr) * (1 - $shrink));")
which depends on the actual learning rate leads to zigzag and highly varying shrink values even for $shrink=0.95 and 0.98. Both the below training runs diverge

red corresponds to $shrink=0.98 and blue 0.95
Plot1 shows shrink values
and plot2 cv accuracies.

shrink_vs_cvaccuracy_iteration

vijayaditya:
Formula : <code> this_shrink=$(perl -e "print $shrink + ($flr/$this_effective_learning_rate) * (1 - $shrink)")</code>

Even with adaptive shrinkage the CV accuracies are no better than constant shrinkage of 0.99. The WERs are worse.

%WER 47.6   :  $shrink 0.95 + adaptive
%WER 47.1   :  $shrinkage 0.98 + adaptive
training diverges   :  $shrink 0.99 + adaptive
%WER 46.8 : $shrink 0.99 + constant 

(In the below plot 0.99 corresponds to the constant shrinkage run)
cv_shrink

danpovey:
On Sun, Sep 27, 2015 at 6:16 AM, Vijayaditya Peddinti <

> Formula : this_shrink=$(perl -e "print $shrink +
> ($flr/$this_effective_learning_rate) * (1 - $shrink)")
>
> Even with adaptive shrinkage the CV accuracies are no better than constant
> shrinkage of 0.99. The WERs are worse.
>
> %WER 47.6 : $shrink 0.95 + adaptive
> %WER 47.1 : $shrinkage 0.98 + adaptive
> training diverges : $shrink 0.99 + adaptive
>
Please implement that thing where you discard models from 'failed' jobs.  I
think the fact that shrink 0.99 + adaptive failed is likely just a fluke,
nothing systematic.  The values that I expect to work well are 0.99 and
0.995.  I think the 0.95 and 0.98 ones are just too strongly shrunk.
BTW, it would be nice to see (later) plots of the avg-deriv.  I'm hoping
that it will be more constant for the ones with adaptive shrinking.

Dan

vijayaditya:
Ok, will do.

Vijay

PR #185, opened 2015-09-29
run.pl and queue.pl CLI compatibility fixes 

jtrmal:
The for-cycle semantics bothers me -- I don't exactly understand the motivation behind that.
The biggest concern is, why is it repeated 3-times? Shouldn't be 2-times (tops) enough?

Secondly: this is a valid command line:
./run.pl --max-jobs-run 4 --opt2 5 JOB=1:16 --opt3 4 --max-jobs-run 14 JOB=1:16 --opt4 5   JOB=1:16   ./log/JOB.log true

we can override JOB two-times (i.e. the total number of JOB specification is 3) -- again, has this any utility, or is it just sloppiness?

y.

### DIFF
--- a/egs/wsj/s5/utils/queue.pl
+++ b/egs/wsj/s5/utils/queue.pl
@@ -18,7 +18,7 @@ use Getopt::Long;
 
 # The script now supports configuring the queue system using a config file
 # (default in conf/queue.conf; but can be passed specified with --config option)
-# and a set of command line options. 
+# and a set of command line options.
 # The current script handles:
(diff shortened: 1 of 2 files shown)


danpovey:
I think you are right that twice would be enough.  I was concerned about  option1 JOB=1:10 option2, but twice will cover this.
Do you want to change that before merging?

jtrmal:
Yep,  I'll  change  it. In both?
Y.

danpovey:
sure...

PR #187, opened 2015-09-30
add nnet3/lstm recipe for swbd

naxingyu:


danpovey:
thanks-  you might want to add results at some point.  e.g. in the RESULTS file or at the end of that file.

PR #192, opened 2015-09-30
Jsalt import ffs

vesis84:
A big pull request integrating code from JSALT workshop (FFS team),
- i-vector support for nnet1 (script changes),
- updating ami scripts (data preparation, scoring, results),
- updating rm nnet1 examples,

This is not all the code from the workshop, the other code is better to commit as new-add,
after it gets ready.

I will run the 'ami' recipe and check everything is fine.

Dan: Not sure if 'egs/ami/s5/conf/queue.conf' should be included (it contains some macros I am not using).

I know it is a lot of stuff, but still, can you please look at it? AFAIK it should not contain arguable changes, but still it will be better if another pair of eyes looks at it. 

Thanks!
K.


danpovey:
@jtrmal, would you mind doing the review for this?

jtrmal:
ok. will look. y.

danpovey:
I don't see any problem with the code here, but @vijayaditya, could you please see if there is anything here which might interfere with your experiments on the AMI recipe?
Dan

jtrmal:
I went through all of the PR.
I noticed one minor compiler compatibitlity issue:
in the tests, there are lines such as
const CuMatrixBase<BaseFloat>& __attribute__((__unused__)) m3 = r.Value();
That will not compile under Visual Studio and I'm not sure about clang --
it's GCC extension.
It could be written in "compatible" way, for example
const CuMatrixBase<BaseFloat>& m3 = r.Value();
static_cast<const void>(m3);
Perhaps these two statements could be combined together  -- I didn't test
it. The void cast is C++ standard way how to force compiler to ignore
unused values
Nevertheless, it would be better to address this in a similar way instead
of adding a #ifdef
Other than that I didn't find any problem
y.

vesis84:
Yes, Yenda is right, the portable way of suppressing 'unused variable' warning is better. Already added to this PR. Thanks!

egs/ami/s5/conf/queue.conf
```diff
@@ -0,0 +1,11 @@
+# Origin at : http://wiki.clsp.jhu.edu/view/Ws15_AWS_Kluster_Rules
```
vijayaditya: Please rename this file to queue_jsalt.conf

egs/ami/s5/local/score_asclite.sh
```diff
@@ -82,11 +82,34 @@ if [ $stage -le 2 ]; then
...
+        -h rt-stt -g $data/glm -r $dir/ascore_LMWT/stm $dir/ascore_LMWT/${oname}.ctm || exit 1
+    # Compress some scoring outputs : alignment info and graphs,
+    echo -n "compressing asclite outputs "
+    for LMWT in $(seq $min_lmwt $max_lmwt); do
+      ascore=$dir/ascore_${LMWT}
+      gzip -f $ascore/${oname}.ctm.filt.aligninfo.csv
+      cp $ascore/${oname}.ctm.filt.alignments/index.html $ascore/${oname}.ctm.filt.overlap.html
+      tar -C $ascore -czf $ascore/${oname}.ctm.filt.alignments.tar.gz ${oname}.ctm.filt.alignments
+      rm -r $ascore/${oname}.ctm.filt.alignments
+      echo -n "$LMWT "
```
vijayaditya: It was quite confusing to see numbers printed without an explanation of what these numbers are.
modify the echo to 
<code>echo -n "LMWT:$LMWT"  </code>

egs/ami/s5/run_prepare_shared.sh
```diff
@@ -3,41 +3,15 @@
...
-# To run this script you need SRILM,
 
 # Path to Fisher transcripts LM interpolation (if not defined only AMI transcript LM is built),
-FISHER_TRANS=`pwd`/eddie_data/lm/data/fisher/part1 # Edinburgh, [DEFAULT]
-# Path where AMI gets downloaded (or where locally available),
-AMI_DIR=$PWD/DOWNLOAD/amicorpus # [DEFAULT]
+#FISHER_TRANS=`pwd`/eddie_data/lm/data/fisher/part1 # Edinburgh,
+#FISHER_TRANS=/export/corpora4/ami/fisher_trans/part1 # JHU,
+#FISHER_TRANS=/mnt/matylda2/data/FISHER/fe_03_p1_tran # BUT,
+FISHER_TRANS=/export/ws15-ffs-data/corpora/LDC/LDC2004T19/fe_03_p1_tran # JSALT2015 workshop, cluster AWS-EC2,
```
vijayaditya: The cluster specific data locations are being overwritten by the commit !!

vesis84: Hi, I don't understand what is the problem here... Sure the locations are different. AMI_DIR is no longer used in run_prepare_shared.sh and the FISHER_TRANS in Edinburg is replaced by JSALT default. Well I am not sure what should be the default for FISHER_TRANS, it could be BUT location. if nobody objects :)

vijayaditya: If you look at the old file it had the cluster specific locations for these transcripts.. could you add that the new one..

<code>

-# We can make setup specific to the 'domain' where the cluster is,

-case "$(hostname -d)" in		
-  fit.vutbr.cz) # BUT cluster,		
-    FISHER_TRANS=/mnt/matylda2/data/FISHER/fe_03_p1_tran		
-  ;;		
-  clsp.jhu.edu) # JHU cluster		
-   FISHER_TRANS=/export/corpora4/ami/fisher_trans/part1		
-  ;;		
-  *) echo "Using defaults locations,"		
-  ;;		
-esac

</code>

egs/ami/s5/run_sdm.sh
```diff
@@ -10,17 +10,16 @@ mic=sdm$micid
...
+set -euxo pipefail
 
 # Path where AMI gets downloaded (or where locally available):
-[ ! -r conf/ami_dir ] && echo "Please, run 'run_prepare_shared.sh' first!" && exit 1
-AMI_DIR=$(cat conf/ami_dir)
+AMI_DIR=$PWD/wav_db # Default,
+#AMI_DIR=/mnt/scratch05/iveselyk/KALDI_AMI_WAV # BUT,
+#AMI_DIR=/export/ws15-ffs-data/corpora/ami # JSALT2015 workshop, cluster AWS-EC2,
 
+[ ! -r data/local/lm/final_lm ] && echo "Please, run 'run_prepare_shared.sh' first!" && exit 1
```
vijayaditya: This comment corresponds to stage 0:
The common scenario on any cluster is that the data would have been downloaded once and reused many times. So in most cases there would be no <code>data/local/downloads/wget_sdm.sh</code> even though there is data in the AMI_DIR. It is better to check the AMI_DIR has the required data, before starting the download. I think this check can be best performed in the <code>local/ami_download.sh</code> script. (See my comment there)

egs/ami/s5/run_sdm.sh
```diff
@@ -10,17 +10,16 @@ mic=sdm$micid
...
-set -o pipefail
-set -x
+# Set bash to 'debug' mode, it prints the commands (option '-x') and exits on : 
+# -e 'error', -u 'undefined variable', -o pipefail 'error in pipeline',
+set -euxo pipefail
 
 # Path where AMI gets downloaded (or where locally available):
-[ ! -r conf/ami_dir ] && echo "Please, run 'run_prepare_shared.sh' first!" && exit 1
-AMI_DIR=$(cat conf/ami_dir)
+AMI_DIR=$PWD/wav_db # Default,
```
vijayaditya: case "$(hostname -d)" in
  fit.vutbr.cz) # BUT cluster,
    AMI_DIR=$(mktemp -d $(find /mnt/scratch*/$USER -maxdepth 0)/kaldi_ami_data_XXXXXX)
  ;;
  clsp.jhu.edu) # JHU cluster
   AMI_DIR=/export/corpora4/ami/amicorpus
  ;;
  *) echo "Using defaults locations,"
  ;;
esac

egs/ami/s5/run_sdm.sh
```diff
@@ -10,17 +10,16 @@ mic=sdm$micid
 stage=0
```
vijayaditya: better make this stage=1 (see comment below for details).

egs/ami/s5/run_sdm.sh
vijayaditya: stage 1 missing.. renumber please

egs/ami/s5/run_prepare_shared.sh
vesis84: Yes I know, but this is not applicable to the case when no domain is set like in the 'AWS' cluster. AFAIK, manually commenting/uncommenting or eventually adding a path is a better solution. On the other hand. Both approaches can be combined...

egs/ami/s5/run_sdm.sh
vesis84: re-download of wavs is not an issue, it gets avoided 'wget --continue', re-download of transcripts is IMHO affordable

egs/ami/s5/run_sdm.sh
vesis84: Unfortunately the domain-based locations are not applicable to the case when no domain is set like in the 'AWS' cluster. AFAIK, manually commenting/uncommenting or eventually adding a path is a simpler solution.

vijayaditya: It is convenient to have push_button scripts. I was suggesting something like this where you have a default AMI_DIR and then update if the domain is known.

<pre><code>
AMI_DIR=$PWD/wav_db # Default,
case "$(hostname -d)" in
fit.vutbr.cz) # BUT cluster,
AMI_DIR=/mnt/scratch05/iveselyk/KALDI_AMI_WAV
;;
clsp.jhu.edu) # JHU cluster
AMI_DIR=/export/corpora4/ami/amicorpus
;;
*) echo "Using defaults location for AMI_DIR,"
;;
esac

</code></pre>

danpovey:
@jtrmal  and @vijayaditya, so do you think this can be merged now?

jtrmal:
I'm OK with merge
Y.

danpovey:
I just talked with Vijay.
I don't think I want to merge this as-is.  The current script will give people a strong temptation to just run it as-is and it will attempt to download the corpus from the server.   This (a) takes many hours, and (b) hogs the network bandwidth of our cluster.  I don't like scripts whose default behavior is to download big things instead of finding them locally, as inertia is very strong and people will tend to use them in this way.
I'm agnostic on the actual mechanism used for this.  I would be fine with people having to uncomment something in the run.sh, as long as the default behavior is to just fail (rather than to attempt to download something).
I think it may be hard to co-ordinate a solution in a short time though.  To avoid further discussion, I am almost tempted to just ask Vijay to implement his solution on top of Karel's pull request.   Karel, would you be upset if we did this?
It may not be ideal, and it may be a little complex, but I'm thinking partly with my sysadmin hat on here, and I just don't want our network to be flooded with unnecessary download requests.

danpovey:
OK, Karel has updated the pull request.
@vijayaditya, could you please check (or get someone to check) that it works OK for us?  Hopefully we can merge this to-morrow.

vijayaditya:
Good to merge !

PR #194, opened 2015-09-30
revise OpenBLAS git clone to more general HTTPS protocol

Hsinfu:
Using HTTPS protocol to clone the OpenBLAS repository works for the computer behind a ssh firewall. Here is the reference https://help.github.com/articles/which-remote-url-should-i-use/


danpovey:
Thanks!

PR #199, opened 2015-10-02
Added support for nnet3 raw computation without acoustic model

vimalmanohar:
1) Created nnet3-get-egs-dense-targets to create egs that continuous valued targets.
2) Separated the nnet computation part from the DecodableAmNnetSimple class to create NnetSimpleComputer, and DecodableAmNnetSimple is made a child of NnetSimpleComputer -- Not sure if this part follows the code standards

### DIFF
--- a/src/nnet3bin/nnet3-get-egs-dense-targets.cc
+++ b/src/nnet3bin/nnet3-get-egs-dense-targets.cc
@@ -1,0 +1,256 @@
+// nnet3bin/nnet3-get-egs-dense-targets.cc
+
+// Copyright 2012-2015  Johns Hopkins University (author:  Daniel Povey)
+//                2014  Vimal Manohar
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


vimalmanohar:
I haven't tested this out completely, but wanted to check if this is an appropriate change to do.

src/nnet3bin/nnet3-get-egs-dense-targets.cc
```diff
@@ -0,0 +1,247 @@
...
+        "Essentially this is a format change from features and posteriors\n"
+        "into a special frame-by-frame format.  This program handles the\n"
+        "common case where you have some input features, possibly some\n"
+        "iVectors, and one set of labels.  If people in future want to\n"
+        "do different things they may have to extend this program or create\n"
+        "different versions of it for different tasks (the egs format is quite\n"
+        "general)\n"
+        "\n"
+        "Usage:  nnet3-get-egs [options] <features-rspecifier> "
+        "<pdf-post-rspecifier> <egs-out>\n"
```
danpovey: fix usage message

danpovey:
It does look plausible.
Dan

PR #201, opened 2015-10-04
Added steps/nnet3/align.sh and corresponding binaries

vijayaditya:
Modifies src/nnet3/nnet-am-decodable-simple.h to accept CuMatrixBase in place of MatrixBase and CuVectorBase in place of VectorBase.


danpovey:
Vijay-- it was actually intentional to pass in the features in non-GPU memory.
This is so that it will work for very long files without using up the GPU memory-- for each chunk that we separately process, we transfer the required features in from CPU to GPU memory and process them there.  It just means that we do the CPU->GPU transfer in chunks rather than all at once.

Dan

vijayaditya:
OK. I will change it back.

vijayaditya:
removed the changes to the interface.. the new commit just adds the binary and shell script.

danpovey:
thanks, I'll wait till the checks complete.

PR #203, opened 2015-10-04
update documentation: adding examples of using user-specific language…

chenguoguo:
… model in online decoding

### DIFF
--- a/src/doc/online_decoding.dox
+++ b/src/doc/online_decoding.dox
@@ -380,25 +380,23 @@ vocabulary, for example, adding out-of-vocabulary words such as person names
 to the existing vocabulary. In this section we will explain how this can be
 done.
 
-We first have to create a pronunciation lexicon, either by adding more words to
-the existing pronunciation lexicon, or by introducing a completely different
-lexicon. In the first case, we can use grapheme-to-phoneme (G2P) conversion
-to generate pronunciations for new words automatically. We can train the G2P
-model using the recognizer's existing pronunciation lexicon, apply the trained
-G2P model to the new words to generate pronunciations, and add those new lexical
-entries to the existing pronunciation lexicon. The commonly used G2P tools are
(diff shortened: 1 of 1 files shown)


src/doc/online_decoding.dox
```diff
@@ -313,7 +313,63 @@ phones as appropriate, e.g. <code>--ivector-silence-weighting.silence-phones=1:2
...
+model with SRILM, and how to incorporate this language model to the existing
+online-nnet2 models.
+
+We first have to build an ARPA format language model with SRILM. Note that SRILM
+comes with a lot of training options, and we assume it's the user's
+responsibility to figure out what is the best setting for their own application.
+Suppose "train.txt" is our language model training corpus (e.g., training
+data transcriptions), and "wordlist" is our vocabulary (e.g., recognizer's
+vocabulary), we can build a 3gram Kneser-Ney language model using the following
+SRILM command
```
danpovey: Guoguo, I think we need to cover the case where the vocabulary might be different, as this will almost always be the case.  Of course this means preparing the dict directory as a modified version of the dict directory from the original build, in a compatible way (at least explain what this means, and maybe mention g2p sequitur without giving further details at this point)-- and then calling prepare_lang.sh with the --phone-symbol-table option set to the old directory's phones.txt to make sure it's compatible.
Actually it might make sense to put this in a separate subsection.
Please be more explicit that 'wordlist' is all the words in words.txt (except, I guess <eps> and #0)... I guess at this point you could point down to the separate subsection talking about the case where it's not that exact list of words.

chenguoguo: Yes that makes sense. i'll add one more section. will have to be offline
for couple of hours, i'll do when i get back.

Guoguo

chenguoguo: I added one more section for the vocabulary and updated the pull request.

Guoguo

danpovey:
Thanks- I merged it, and I'm compiling now.

PR #204, opened 2015-10-05
make codes comply with the Google style

keli78:
make source and header files in /base comply with the Google style

### DIFF
--- a/src/base/kaldi-error.cc
+++ b/src/base/kaldi-error.cc
@@ -90,7 +90,8 @@ inline void KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(const char *to_append,
   }
 }
 #else  // defined(HAVE_CXXABI_H) && defined(HAVE_EXECINFO_H)
-#define KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(to_append, ans) ans += to_append
+#define KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(to_append, ans) \
+  ans->append(to_append)
 #endif  // defined(HAVE_CXXABI_H) && defined(HAVE_EXECINFO_H)
 
 #ifdef HAVE_EXECINFO_H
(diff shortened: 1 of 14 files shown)


danpovey:
These changes look good generally.
However, I think if we enforce the Google style-guide rule against using-directives e.g. 'using namespace foo;' then all of our main routines - especially in binaries - will get very ugly.
I think we should make an exception to the 'using namespace x;' rule when that namespace is kaldi.  I'm OK with enforcing it when that namespace is anything else, e.g. 'fst'.  Could you please update your pull request to 
  (a) modify cpplint.py to make an exception for namespace kaldi,
  (b) revert those changes to main() routines in test code, to keep using 'using namespace kaldi;'
No hurry.
Thanks..
Dan

danpovey:
... actually there may be situations when even 'using namespace fst;' makes sense, e.g. in test routines in directory fstext/.  We'll just ignore the warnings from cpplint.py in those cases.

src/base/kaldi-error.cc
```diff
@@ -81,14 +83,14 @@ inline void KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(std::string &ans,
...
+    ans->append(demangled_name);
     free(demangled_name);
   } else {
-    ans += to_append;  // add the original string.
+    ans->append(to_append);  // add the original string.
   }
 }
 #else  // defined(HAVE_CXXABI_H) && defined(HAVE_EXECINFO_H)
-#define KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(ans, to_append) ans += to_append
+#define KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(to_append, ans) ans += to_append
```
danpovey: This is not right because ans is now a pointer... it would just be incrementing the pointer.  You should do *ans += to_append or (better) ans->append(to_append)

PR #208, opened 2015-10-05
fixed the bug where fstdeterminizestar sometimes computes the wrong w…

hainan-xv:
…eight under log-semiring with lots of epsilons

### DIFF
--- a/src/fstext/determinize-star-inl.h
+++ b/src/fstext/determinize-star-inl.h
@@ -1,6 +1,7 @@
 // fstext/determinize-star-inl.h
 
 // Copyright 2009-2011  Microsoft Corporation;  Jan Silovsky
+//           2015 Hainan Xu
 
 // See ../../COPYING for clarification regarding multiple authors
 //
@@ -468,46 +469,112 @@ template<class Arc> class DeterminizerStar {
   // Called by ProcessSubset.
   // Has no side effects except on the repository.
 
+  struct EpsilonClosureInfo {
+    EpsilonClosureInfo() {}
+
+    EpsilonClosureInfo(Element e, Weight w, bool i, int d = 0) :
+      element(e), weight_to_process(w), in_queue(i), depth(d) {}
+
+    // the weight in the Element struct is the total current weight
+    // that has been processed already
+    Element element;
+
+    // this stores the weight that we haven't processed (propagated)
+    Weight weight_to_process;
+
+    // whether "this" struct is in the queue
+    // we store the info here so that we don't have to look it up every time
+    bool in_queue;
+
+    // depth is the max number of epsilons (so far) from the original element set
+    // e.g. if the original set is (s1, s2)
+    // s1 -eps-> s4
+    // s2 -eps-> s3 -eps-> s4
+    // then the depth for s4 would eventually be 2
+
+    // to avoid duplicate of computation, we use a priority_queue for storing
+    // the state_id's to process, and always pick the one with the smallest
+    // depth so far first
+    //
+    // this would be similar to topo-sort the states before processing;
+    // (not quite identical since we don't compute the definitive depth first,
+    // but gradually update the depth while running
+    // it seems to speed up the process very significantly already
+    int depth;
+  };
+
+  // this is the struct that is put into the priority_queue. 
+  // The comparison operator makes sure that the queue.top() returns
+  // the one with the least depth
+  struct IdAndDepth {
+    IdAndDepth (InputStateId i, int d):
+      id(i), depth(d) {}
+    InputStateId id;
+    int depth;
+    bool operator <(const IdAndDepth& other) const {
+      return this->depth > other.depth;  // we want the lowest depth on top
+    }
+  };
+
   void EpsilonClosure(const vector<Element> &input_subset,
                       vector<Element> *output_subset) {
     // input_subset must have only one example of each StateId.
 
-    std::map<InputStateId, Element> cur_subset;
-    typedef typename std::map<InputStateId, Element>::iterator MapIter;
+    std::map<InputStateId, EpsilonClosureInfo> cur_subset;
+    priority_queue<IdAndDepth> queue;  // queue of things to be processed.
+
+    typedef typename std::map<InputStateId, EpsilonClosureInfo>::iterator MapIter;
     {
       MapIter iter = cur_subset.end();
       for (size_t i = 0; i < input_subset.size(); i++) {
-        std::pair<const InputStateId, Element> pr(input_subset[i].state, input_subset[i]);
+        queue.push(IdAndDepth(input_subset[i].state, 0));
+
+        // the weight has not been processed yet,
+        // so put all of them in the "weight_to_process"
+        EpsilonClosureInfo info(input_subset[i],
+                                input_subset[i].weight,
+                                true,
+                                0);
+        info.element.weight = Weight::Zero(); // clear the weight
+
+        std::pair<const InputStateId, EpsilonClosureInfo>
+                                 pr(input_subset[i].state, info);
         iter = cur_subset.insert(iter, pr);
         // By providing iterator where we inserted last one, we make insertion more efficient since
         // input subset was already in sorted order.
       }
     }
     // find whether input fst is known to be sorted in input label.
-    bool sorted = ((ifst_->Properties(kILabelSorted, false) & kILabelSorted) != 0);
+    bool sorted =
+            ((ifst_->Properties(kILabelSorted, false) & kILabelSorted) != 0);
 
-    vector<Element> queue(input_subset);  // queue of things to be processed.
-    bool replaced_elems = false; // relates to an optimization, see below.
     int counter = 0; // relates to max-states option, used for test.
     while (queue.size() != 0) {
-      Element elem = queue.back();
-      queue.pop_back();
-      // The next if-statement is a kind of optimization.  It's to prevent us
-      // unnecessarily repeating the processing of a state.  "cur_subset" always
-      // contains only one Element with a particular state.  The issue is that
-      // whenever we modify the Element corresponding to that state in "cur_subset",
-      // both the new (optimal) and old (less-optimal) Element will still be in
-      // "queue".  The next if-statement stops us from wasting compute by
-      // processing the old Element.
+      InputStateId id = queue.top().id;
+      int this_depth = queue.top().depth;
+      EpsilonClosureInfo &info = cur_subset[id];
+      Element &elem = info.element;
+      Weight unprocessed_weight = info.weight_to_process;
 
-      if (replaced_elems && cur_subset[elem.state] != elem)
+      KALDI_ASSERT(this_depth == info.depth);
+
+      elem.weight = Plus(elem.weight, unprocessed_weight);
+      info.weight_to_process = Weight::Zero();
+
+      info.in_queue = false;
+      queue.pop();
+
+      if (unprocessed_weight == Weight::Zero()) {
         continue;
+      }
 
       if (max_states_ > 0 && counter++ > max_states_) {
         std::cerr << "Determinization aborted since looped more than "
                   << max_states_ << " times during epsilon closure.\n";
         throw std::runtime_error("looped more than max-states times in determinization");
       }
+
+      // now we are going to propagate the "unprocessed_weight"
       for (ArcIterator<Fst<Arc> > aiter(*ifst_, elem.state);
            !aiter.Done(); aiter.Next()) {
         const Arc &arc = aiter.Value();
@@ -519,7 +586,10 @@ template<class Arc> class DeterminizerStar {
         if (arc.ilabel == 0) {  // Epsilon transition.
           Element next_elem;
           next_elem.state = arc.nextstate;
-          next_elem.weight = Times(elem.weight, arc.weight);
+          next_elem.weight = Weight::Zero();
+          Weight next_unprocessed_weight
+                         = Times(unprocessed_weight, arc.weight);
+
           // now must append strings
           if (arc.olabel == 0)
             next_elem.string = elem.string;
@@ -530,19 +600,24 @@ template<class Arc> class DeterminizerStar {
               seq.push_back(arc.olabel);
             next_elem.string = repository_.IdOfSeq(seq);
           }
-          typename std::map<InputStateId, Element>::iterator
+          typename std::map<InputStateId, EpsilonClosureInfo>::iterator
               iter = cur_subset.find(next_elem.state);
           if (iter == cur_subset.end()) {
             // was no such StateId: insert and add to queue.
-            cur_subset[next_elem.state] = next_elem;
-            queue.push_back(next_elem);
+            cur_subset[next_elem.state] =
+                         EpsilonClosureInfo(next_elem,
+                                            next_unprocessed_weight,
+                                            true,
+                                            this_depth + 1
+                                            );
+            queue.push(IdAndDepth(next_elem.state, this_depth + 1));
           } else {  // one is already there.  Add weights.
-            if (iter->second.string != next_elem.string) {
+            if (iter->second.element.string != next_elem.string) {
               std::cerr << "DeterminizerStar: FST was not functional -> not determinizable\n";
               { // Print some debugging information.  Can be helpful to debug
                 // the inputs when FSTs are mysteriously non-functional.
                 vector<Label> tmp_seq;
-                repository_.SeqOfId(iter->second.string, &tmp_seq);
+                repository_.SeqOfId(iter->second.element.string, &tmp_seq);
                 std::cerr << "First string: ";
                 for (size_t i = 0; i < tmp_seq.size(); i++) std::cerr << tmp_seq[i] << " ";
                 std::cerr << "\nSecond string: ";
@@ -552,11 +627,32 @@ template<class Arc> class DeterminizerStar {
               }
               throw std::runtime_error("Non-functional FST: cannot determinize.\n");
             }
-            Weight weight = Plus(iter->second.weight, next_elem.weight);
-            if (! ApproxEqual(weight, iter->second.weight, delta_)) {  // add extra part of weight to queue.
-              iter->second.weight = weight; // Update weight in map.
-              queue.push_back(next_elem);
-              replaced_elems = true;
+            if (iter->second.in_queue) {
+              // since in queue, no check since we're propagating them anyway
+              iter->second.weight_to_process = Plus(iter->second.weight_to_process, next_unprocessed_weight);
+              // probably should take it out of the queue, update the depth
+              // and put it back.
+              // but it seems it's fast enough already ;-)
+            }
+            else {
+              // is in the subset but not in queue (we know its unprcessed_weight == Zero())
+              Weight weight = Plus(iter->second.element.weight, next_unprocessed_weight);
+              if (! ApproxEqual(weight, iter->second.element.weight, delta_)) {
+                // add extra part of weight to queue.
+                iter->second.weight_to_process = next_unprocessed_weight;
+                iter->second.in_queue = true;
+                iter->second.depth = max(iter->second.depth, this_depth + 1);
+                queue.push(IdAndDepth(next_elem.state, iter->second.depth));
+              }
+              else {
+                iter->second.weight_to_process = Weight::Zero();
+
+                // we replace the weight
+                // but not add this node to the queue
+                iter->second.element.weight = weight;
+                iter->second.in_queue = false;
+                iter->second.depth = max(iter->second.depth, this_depth + 1);
+              }
             }
           }
         }
@@ -568,7 +664,10 @@ template<class Arc> class DeterminizerStar {
       output_subset->clear();
       output_subset->reserve(cur_subset.size());
       MapIter iter = cur_subset.begin(), end = cur_subset.end();
-      for (; iter != end; ++iter) output_subset->push_back(iter->second);
+      for (; iter != end; ++iter) {
+        KALDI_ASSERT(iter->second.weight_to_process == Weight::Zero());
+        output_subset->push_back(iter->second.element);
+      }
     }
   }
 
--- a/src/fstext/determinize-star-inl.h
+++ b/src/fstext/determinize-star-inl.h
@@ -564,10 +564,6 @@ template<class Arc> class DeterminizerStar {
       info.in_queue = false;
       queue.pop();
 
-      if (unprocessed_weight == Weight::Zero()) {
-        continue;
-      }
-
       if (max_states_ > 0 && counter++ > max_states_) {
         std::cerr << "Determinization aborted since looped more than "
                   << max_states_ << " times during epsilon closure.\n";
@@ -629,29 +625,33 @@ template<class Arc> class DeterminizerStar {
             }
             if (iter->second.in_queue) {
               // since in queue, no check since we're propagating them anyway
-              iter->second.weight_to_process = Plus(iter->second.weight_to_process, next_unprocessed_weight);
+              iter->second.weight_to_process =
+                  Plus(iter->second.weight_to_process, next_unprocessed_weight);
               // probably should take it out of the queue, update the depth
               // and put it back.
               // but it seems it's fast enough already ;-)
-            }
-            else {
-              // is in the subset but not in queue (we know its unprcessed_weight == Zero())
-              Weight weight = Plus(iter->second.element.weight, next_unprocessed_weight);
+            } else {
+              // is in the subset but not in queue
+              iter->second.weight_to_process = Plus(iter->second.weight_to_process,
+                                                    next_unprocessed_weight);
+
+              // this is because the code in "else" below: the 
+              // iter->second.weight_to_process might not be Zero()
+              Weight weight = Plus(iter->second.element.weight,
+                                   iter->second.weight_to_process);
+
+              // What is done below is, we propagate the weight (by adding them
+              // to the queue only when the change is big enough;
+              // otherwise we just store the weight, until before returning
+              // we add the element.weight and weight_to_process together
+              iter->second.depth = max(iter->second.depth, this_depth + 1);
               if (! ApproxEqual(weight, iter->second.element.weight, delta_)) {
                 // add extra part of weight to queue.
-                iter->second.weight_to_process = next_unprocessed_weight;
                 iter->second.in_queue = true;
-                iter->second.depth = max(iter->second.depth, this_depth + 1);
                 queue.push(IdAndDepth(next_elem.state, iter->second.depth));
-              }
-              else {
-                iter->second.weight_to_process = Weight::Zero();
-
-                // we replace the weight
-                // but not add this node to the queue
-                iter->second.element.weight = weight;
+              } else {
+                // we update the weight_to_process, but not add it to the queue
                 iter->second.in_queue = false;
-                iter->second.depth = max(iter->second.depth, this_depth + 1);
               }
             }
           }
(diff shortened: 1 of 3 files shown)


src/fstext/determinize-star-inl.h
```diff
@@ -468,46 +469,112 @@ template<class Arc> class DeterminizerStar {
...
+
+      KALDI_ASSERT(this_depth == info.depth);
+
+      elem.weight = Plus(elem.weight, unprocessed_weight);
+      info.weight_to_process = Weight::Zero();
+
+      info.in_queue = false;
+      queue.pop();
+
+      if (unprocessed_weight == Weight::Zero()) {
```
danpovey: Should something even be in the queue if it has zero weight?

src/fstext/determinize-star-inl.h
```diff
@@ -552,11 +627,32 @@ template<class Arc> class DeterminizerStar {
...
               throw std::runtime_error("Non-functional FST: cannot determinize.\n");
             }
-            Weight weight = Plus(iter->second.weight, next_elem.weight);
-            if (! ApproxEqual(weight, iter->second.weight, delta_)) {  // add extra part of weight to queue.
-              iter->second.weight = weight; // Update weight in map.
-              queue.push_back(next_elem);
-              replaced_elems = true;
+            if (iter->second.in_queue) {
+              // since in queue, no check since we're propagating them anyway
+              iter->second.weight_to_process = Plus(iter->second.weight_to_process, next_unprocessed_weight);
```
danpovey: line too long

src/fstext/determinize-star-inl.h
```diff
@@ -552,11 +627,32 @@ template<class Arc> class DeterminizerStar {
...
+              // probably should take it out of the queue, update the depth
+              // and put it back.
+              // but it seems it's fast enough already ;-)
+            }
```
danpovey: asymmetrical braces -- better have } else {

src/fstext/determinize-star-inl.h
```diff
@@ -552,11 +627,32 @@ template<class Arc> class DeterminizerStar {
...
+            else {
+              // is in the subset but not in queue (we know its unprcessed_weight == Zero())
```
danpovey: The logic here is what I intended.  You are effectively discarding any weight that does not meet the delta threshold.
What would be better is to *always* accumulate, in iter->second.weight_to_process, the extra weight, but to use the delta threshold to decide whether to put it in the queue (if it isn't already there).

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,221 @@ template<class Arc> class DeterminizerStar {
...
+        typename std::map<InputStateId, EpsilonClosureInfo>::iterator
+            iter = cur_subset->find(next_elem.state);
+        if (iter == cur_subset->end()) {
+          // was no such StateId: insert and add to queue.
+          (*cur_subset)[next_elem.state] =
+                       EpsilonClosureInfo(next_elem,
+                                          next_unprocessed_weight,
+                                          true,
+                                          this_depth + 1
+                                          );
```
danpovey: no extra newline please.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,221 @@ template<class Arc> class DeterminizerStar {
...
     // input_subset must have only one example of each StateId.
 
-    std::map<InputStateId, Element> cur_subset;
-    typedef typename std::map<InputStateId, Element>::iterator MapIter;
+    std::map<InputStateId, EpsilonClosureInfo> cur_subset;
+
+    typedef typename std::map<InputStateId, EpsilonClosureInfo>::iterator MapIter;
     {
       MapIter iter = cur_subset.end();
       for (size_t i = 0; i < input_subset.size(); i++) {
```
danpovey: It's never a good idea to call size() as part of a loop condition.  The compiler keeps having to compute it.  Instead, create a variable storing that size and compare against that.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,221 @@ template<class Arc> class DeterminizerStar {
...
-        std::pair<const InputStateId, Element> pr(input_subset[i].state, input_subset[i]);
+        queue.push(IdAndDepth(input_subset[i].state, 0));
```
danpovey: I think you can avoid actually adding these things to the queue here-- which is probably where most of the time is being taken.  Why don't you remove the 'queue.push' statement from this loop, and then after this loop, have another loop from 0 to input_subset.size(), where you set
      InputStateId id = input_subset[i].state
      int this_depth = 0;
This will process things in the same order as your queue, because you'll be processing all the 0-depth things first anyway.

hainan-xv: Done.

On Thu, Oct 8, 2015 at 9:18 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/fstext/determinize-star-inl.h
> <https://github.com/kaldi-asr/kaldi/pull/208#discussion_r41589135>:
>
> >      {
> >        MapIter iter = cur_subset.end();
> >        for (size_t i = 0; i < input_subset.size(); i++) {
> > -        std::pair<const InputStateId, Element> pr(input_subset[i].state, input_subset[i]);
> > +        queue.push(IdAndDepth(input_subset[i].state, 0));
>
> I think you can avoid actually adding these things to the queue here--
> which is probably where most of the time is being taken. Why don't you
> remove the 'queue.push' statement from this loop, and then after this loop,
> have another loop from 0 to input_subset.size(), where you set
> InputStateId id = input_subset[i].state
> int this_depth = 0;
> This will process things in the same order as your queue, because you'll
> be processing all the 0-depth things first anyway.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/208/files#r41589135>.
>



-- 
- Hainan

danpovey: Thanks- let us know how this affects the speed.
It should be a big effect.
Dan

hainan-xv: Actually I did the size() -> size and taking the 0th level nodes out of the
queue last night already... maybe the I forgot to push it in then.

It was about 10s faster than before, but still 10s slower than baseline.

On Thu, Oct 8, 2015 at 9:37 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/fstext/determinize-star-inl.h
> <https://github.com/kaldi-asr/kaldi/pull/208#discussion_r41590019>:
>
> >      {
> >        MapIter iter = cur_subset.end();
> >        for (size_t i = 0; i < input_subset.size(); i++) {
> > -        std::pair<const InputStateId, Element> pr(input_subset[i].state, input_subset[i]);
> > +        queue.push(IdAndDepth(input_subset[i].state, 0));
>
> Thanks- let us know how this affects the speed. It should be a big effect.
> Dan
> … <#1504a3f21ab92d3a_>
> On Thu, Oct 8, 2015 at 9:36 PM, hainan-xv <notifications@github.com>
> wrote: In src/fstext/determinize-star-inl.h <
> https://github.com/kaldi-asr/kaldi/pull/208#discussion_r41589978>: > { >
> MapIter iter = cur_subset.end(); > for (size_t i = 0; i <
> input_subset.size(); i++) { > - std::pair<const InputStateId, Element>
> pr(input_subset[i].state, input_subset[i]); > +
> queue.push(IdAndDepth(input_subset[i].state, 0)); Done. …
> <#1504a3e373cbfcd4_> On Thu, Oct 8, 2015 at 9:18 PM, Daniel Povey <
> notifications@github.com> wrote: In src/fstext/determinize-star-inl.h <
> https://github.com/kaldi-asr/kaldi/pull/208#discussion_r41589135>: > { >
> MapIter iter = cur_subset.end(); > for (size_t i = 0; i <
> input_subset.size(); i++) { > - std::pair<const InputStateId, Element>
> pr(input_subset[i].state, input_subset[i]); > +
> queue.push(IdAndDepth(input_subset[i].state, 0)); I think you can avoid
> actually adding these things to the queue here-- which is probably where
> most of the time is being taken. Why don't you remove the 'queue.push'
> statement from this loop, and then after this loop, have another loop from
> 0 to input_subset.size(), where you set InputStateId id =
> input_subset[i].state int this_depth = 0; This will process things in the
> same order as your queue, because you'll be processing all the 0-depth
> things first anyway. — Reply to this email directly or view it on GitHub <
> https://github.com/kaldi-asr/kaldi/pull/208/files#r41589135>. -- - Hainan
> — Reply to this email directly or view it on GitHub <
> https://github.com/kaldi-asr/kaldi/pull/208/files#r41589978>.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/208/files#r41590019>.
>



-- 
- Hainan

danpovey: OK, I may merge it as-is then.  But not right this second, I have to do
something.
Dan

src/fstext/determinize-star-inl.h
danpovey: this comment still stands- don't give ); its own line.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,232 @@ template<class Arc> class DeterminizerStar {
...
+                std::cerr << tmp_seq[i] << " ";
+              std::cerr << "\n";
+            }
+            throw std::runtime_error("Non-functional FST: cannot determinize.\n");
+          }
+          if (iter->second.in_queue) {
+            // since in queue, no check since we're propagating them anyway
+            iter->second.weight_to_process =
+                Plus(iter->second.weight_to_process, next_unprocessed_weight);
+            // probably should take it out of the queue, update the depth
```
danpovey: I would update the depth in the info (i.e. *iter) but not in  the queue, and remove the assertion that the depths match.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,232 @@ template<class Arc> class DeterminizerStar {
...
-
-      if (replaced_elems && cur_subset[elem.state] != elem)
-        continue;
+      InputStateId id = queue.top().id;
+      int this_depth = queue.top().depth;
+      EpsilonClosureInfo &info = cur_subset[id];
+      Element &elem = info.element;
+      Weight unprocessed_weight = info.weight_to_process;
+
+      KALDI_ASSERT(this_depth == info.depth);
```
danpovey: I'd remove this assertion.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,232 @@ template<class Arc> class DeterminizerStar {
...
-      // whenever we modify the Element corresponding to that state in "cur_subset",
-      // both the new (optimal) and old (less-optimal) Element will still be in
-      // "queue".  The next if-statement stops us from wasting compute by
-      // processing the old Element.
-
-      if (replaced_elems && cur_subset[elem.state] != elem)
-        continue;
+      InputStateId id = queue.top().id;
+      int this_depth = queue.top().depth;
+      EpsilonClosureInfo &info = cur_subset[id];
```
danpovey: Please make this const if you don't plan to change it.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,232 @@ template<class Arc> class DeterminizerStar {
...
+    // now we are going to propagate the "unprocessed_weight"
+    for (ArcIterator<Fst<Arc> > aiter(*ifst_, elem.state);
+         !aiter.Done(); aiter.Next()) {
+      const Arc &arc = aiter.Value();
+      if (sorted && arc.ilabel > 0) {
+        break;
+        // Break from the loop: due to sorting there will be no
+        // more transitions with epsilons as input labels.
+      }
+      if (arc.ilabel == 0) {  // Epsilon transition.
```
danpovey: You could make the indentation level more reasonable by changing this to
if (arc.ilabel != 0)
  continue;  // We are only processing epsilons here.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,232 @@ template<class Arc> class DeterminizerStar {
...
+  // Sub-routine that we call in EpsilonClosure()
+  // To avoid the function getting too long, we make it a separate function
+  // It takes the current "unprocessed_weight" and propagate it to the 
+  // states accessible from elem.state by an epsilon arc
+  // and add the results to cur_subset.
+  void ExpandOneElement(const Element &elem,
+                        bool sorted,
+                        Weight unprocessed_weight,
+                        std::map<InputStateId, EpsilonClosureInfo> *cur_subset,
+                        int this_depth) {
```
danpovey: Make this function const if possible.

src/fstext/determinize-star-inl.h
hainan-xv: I am not sure how to do that... It seems to do that we need to pass to the priority_queue a comparison class; if we rely on the information in "info", it'll make the comparison class not "well defined", since it will keep changing?

src/fstext/determinize-star-inl.h
hainan-xv: Actually the code will change it later 

line 680 "elem.weight = Plus(elem.weight, unprocessed_weight);"

Should I not use the reference here, but use the original "cur_subset[id].element.weight = ..." instead?

src/fstext/determinize-star-inl.h
danpovey: The key in queue will contain outdated information, which won't matter, and it won't matter that it's not consistent with the information in the "info".  It will just give you a more accurate value next time it gets entered into the queue.

src/fstext/determinize-star-inl.h
danpovey: no problem, leave it as-is.

src/fstext/determinize-star-inl.h
hainan-xv: Hmm I think I already fixed this in the last commit.

danpovey: you must not have pushed it.  it's showing up at https://github.com/kaldi-asr/kaldi/pull/208/files.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,221 @@ template<class Arc> class DeterminizerStar {
...
-              iter->second.weight = weight; // Update weight in map.
-              queue.push_back(next_elem);
-              replaced_elems = true;
-            }
-          }
-        }
+        throw std::runtime_error("looped more than max-states times"
+           " in determinization");
       }
+
```
danpovey: You need to be a bit careful about iterator invalidation here.  I think you are passing to ExpandOneElement a reference to 'elem' that could be invalidated inside ExpandOneElement.
Also, in general, you should avoid an iterator being invalidated while it is still in scope, or you should at least have a comment to explain what's going on.

src/fstext/determinize-star-inl.h
hainan-xv: It is not a const function since it changes the class variable "queue'.

I am actually not very sure about this design - the reason I did that is I feel if I use a local variable for queue, then it is going to call resize() over and over again each time we compute the EpsilonClosure for any set of "Element"s. Hopefully by having it as a class variable, it calls resize() much less often; but it seems a bad design from an OO point of view. 
(I tried reserve() but it didn't work on a queue)

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,221 @@ template<class Arc> class DeterminizerStar {
   // Called by ProcessSubset.
   // Has no side effects except on the repository.
 
+  struct EpsilonClosureInfo {
+    EpsilonClosureInfo() {}
+
+    EpsilonClosureInfo(Element e, Weight w, bool i, int d = 0) :
```
danpovey: Better to make this constructor take const Element &e, const Weight &w: avoid constructor calls.

src/fstext/determinize-star-inl.h
```diff
@@ -468,99 +469,221 @@ template<class Arc> class DeterminizerStar {
...
+        // now must append strings
+        if (arc.olabel == 0)
+          next_elem.string = elem.string;
+        else {
+          vector<Label> seq;
+          repository_.SeqOfId(elem.string, &seq);
+          if (arc.olabel != 0)
+            seq.push_back(arc.olabel);
+          next_elem.string = repository_.IdOfSeq(seq);
+        }
```
danpovey: Hainan - about the problem we were trying to solve with the 'in_queue' variable... I think I may have a way to solve this.
If you take the rest of this function starting from this line (and this function is anyway too long), and make it a function, e.g. AddOneElement(const Element &element, int32 depth);
then you could call that same function when iterating over input_subset [i.e. AddOneElement(input_subset[i], 0)].
Although this would have an O(n log n) lookup, you already have such a thing when creating the initial map, so there is no loss-- this way avoid having to have two such lookups.

src/fstext/determinize-star-inl.h
danpovey: OK.
You should add an underscore to queue's name.  That's why I assumed it was
a local variable.

src/fstext/determinize-star-inl.h
hainan-xv: Just implemented this - it seems it's way slower than before: around 40 seconds, while the initial implementation is about 20s slower. Also now I am using const & in the constructor for EpsilonClosureInfo() which should bring some speed up...

It seems to me this algorithm is doing the exact same thing as our 1st implementation; except that in the first implementation we could utilize the fact that the StateId's are sorted so the insertion to the set is linear, while here it is n log n. I just pushed the changes.

danpovey: Oh - disappointing.
Make sure you address that iterator invalidation problem I pointed out, if
you haven't already.

I realized after I commented that there would be a penalty in that method
because it first does a find() and then uses operator [] to set the
element, and both are n log n.
I have a couple of ideas for how you could address this:
(A) [written before idea (B)]
Anyway, having a separate function like this is probably a better way to
structure it, and we can do further optimizations, and then go back to the
old way of doing it if it doesn't work (but maybe keep the separate
function, inlined, for use just once).
My idea is to optimize that [newly created] method is this:
Instead of 'find' you can use function lower_bound to find the first
element which is <= the given element.  Then if the contents are the thing
you want, you found it; otherwise you can use operator -- on the returned
iterator (assuming it doesn't return the map's begin()), and use the result
as a 'hint' in the version of 'insert' that takes the hint.  This will
avoid having two n log n operations.

Idea (B) [I think this is better]:
Instead of a map from InputStateId to EpsilonClosureInfo, you can have a
vector of EpsilonClosureInfo, and separately store a map from InputStateId
to the index into the EpsilonClosureInfo vector.  The map from InputStateId
to the index can be a std::vector<int32> that is allocated just once, in
the constructor.  You will never 'clean up' this vector: the approach will
be that you do a direct lookup, then test 'is this index valid' (i.e. is it
in-range and if so does it point to what we asked for).  If not, you treat
it as empty.
The queue could then be a queue of pairs (depths, indexes) (i.e. store the
indexes into the EpsilonClosureInfo vector), which means no map lookup is
required.
The vector of EpsilonClosureInfo could also be a class member variable, to
avoid allocation and deallocation.


Dan

danpovey: BTW, since we're looking at this, there is another optimization that would
save a lot of time in some common cases:
We often determinize graphs that are epsilon-free or almost epsilon-free.
You could store a count of how many times the epsilon-propagation in a
subset did not actually find any epsilons, and if most of the time you're
not finding epsilons, you could activate a code path that would just check
all the input states for epsilons, and if none are found, return the input
states unchanged.  If you do this it might be better to store a weighted
count, so the total of [all num-subset-elements], and [num-subset-elements
that did have epsilons], to guard against the situation where the subsets
that have epsilons are a different size from those that don't.
Dan

src/fstext/determinize-star-inl.h
hainan-xv: I am not sure I could see the problem here... the elem is a reference to a sub-structure of a member of the cur_subset, and we never delete elements from cur_subset, and we don't have an iterator for the cur_subset either. I guess it's safe to assume that the existing members in the cur_subset wouldn't change its location in memory?

danpovey: cur_subset could be added to inside the funiiton.
But actually, you may be right. According to http://stackoverflow.com/questions/6438086/iterator-invalidation-rules, for insertions into maps, iterators to other elements are not affected.

hainan-xv: I guess I should write some comment saying this is OK

danpovey: makes sense.

jtrmal:
Also Hainan, the travis build fails -- see the error below
When creatig PR (with the assumption Dan is going to merge it), please make sure it compiles cleanly and the "make test"  runs ok as well

determinize-star-test.cc: In instantiation of ‘void fst::TestDeterminizeGeneral() [with Arc = fst::ArcTpl<fst::TropicalWeightTpl<float> >]’:
determinize-star-test.cc:602:46:   required from here
determinize-star-test.cc:48:65: error: no matching function for call to ‘DeterminizeStar(fst::VectorFst<fst::ArcTpl<fst::TropicalWeightTpl<float> > >&, fst::VectorFst<fst::ArcTpl<fst::TropicalWeightTpl<float> > >*, const float&, NULL, int&)’
       DeterminizeStar<Arc>(*fst, &ofst, kDelta, NULL, max_states);
                                                                 ^
determinize-star-test.cc:48:65: note: candidates are:
In file included from ../fstext/determinize-star.h:116:0,
                 from determinize-star-test.cc:22:
../fstext/determinize-star-inl.h:632:6: note: template<class F> bool fst::DeterminizeStar(F&, fst::MutableFst<typename I::Arc>*, float, bool*, int, bool)
 bool DeterminizeStar(F &ifst, MutableFst<typename F::Arc> *ofst,
      ^
../fstext/determinize-star-inl.h:632:6: note:   template argument deduction/substitution failed:
../fstext/determinize-star-inl.h: In substitution of ‘template<class F> bool fst::DeterminizeStar(F&, fst::MutableFst<typename I::Arc>*, float, bool*, int, bool) [with F = fst::ArcTpl<fst::TropicalWeightTpl<float> >]’:
determinize-star-test.cc:48:65:   required from ‘void fst::TestDeterminizeGeneral() [with Arc = fst::ArcTpl<fst::TropicalWeightTpl<float> >]’
determinize-star-test.cc:602:46:   required from here
../fstext/determinize-star-inl.h:632:6: error: no type named ‘Arc’ in ‘class fst::ArcTpl<fst::TropicalWeightTpl<float> >’
determinize-star-test.cc: In instantiation of ‘void fst::TestDeterminizeGeneral() [with Arc = fst::ArcTpl<fst::TropicalWeightTpl<float> >]’:
determinize-star-test.cc:602:46:   required from here
../fstext/determinize-star-inl.h:646:6: note: template<class F> bool fst::DeterminizeStar(F&, fst::MutableFst<fst::GallicArc<typename F::Arc> >*, float, bool*, int, bool)
 bool DeterminizeStar(F &ifst,
      ^
../fstext/determinize-star-inl.h:646:6: note:   template argument deduction/substitution failed:
../fstext/determinize-star-inl.h: In substitution of ‘template<class F> bool fst::DeterminizeStar(F&, fst::MutableFst<fst::GallicArc<typename F::Arc> >*, float, bool*, int, bool) [with F = fst::ArcTpl<fst::TropicalWeightTpl<float> >]’:
determinize-star-test.cc:48:65:   required from ‘void fst::TestDeterminizeGeneral() [with Arc = fst::ArcTpl<fst::TropicalWeightTpl<float> >]’
determinize-star-test.cc:602:46:   required from here
../fstext/determinize-star-inl.h:646:6: error: no type named ‘Arc’ in ‘class fst::ArcTpl<fst::TropicalWeightTpl<float> >’
In file included from determinize-star-test.cc:20:0:
../base/kaldi-math.h:126:17: warning: ‘kaldi::kLogZeroBaseFloat’ defined but not used [-Wunused-variable]
 const BaseFloat kLogZeroBaseFloat = -std::numeric_limits<BaseFloat>::infinity();
                 ^
make[1]: *** [determinize-star-test] Error 1

hainan-xv:
Fixed.

Thanks

On Tue, Oct 20, 2015 at 3:05 PM, jtrmal <notifications@github.com> wrote:

> Also Hainan, the travis build fails -- see the error below
> When creatig PR (with the assumption Dan is going to merge it), please
> make sure it compiles cleanly and the "make test" runs ok as well
>
> determinize-star-test.cc: In instantiation of ‘void
> fst::TestDeterminizeGeneral() [with Arc =
> fst::ArcTplfst::TropicalWeightTpl<float >]’:
> determinize-star-test.cc:602:46: required from here
> determinize-star-test.cc:48:65: error: no matching function for call to
> ‘DeterminizeStar(fst::VectorFstfst::ArcTpl<fst::TropicalWeightTpl<float >
> >&, fst::VectorFstfst::ArcTpl<fst::TropicalWeightTpl<float > >
> *, const float&, NULL, int&)’ DeterminizeStar(*fst, &ofst, kDelta, NULL,
> max_states);
> ^
> determinize-star-test.cc:48:65: note: candidates are:
> In file included from ../fstext/determinize-star.h:116:0,
> from determinize-star-test.cc:22:
> ../fstext/determinize-star-inl.h:632:6: note: template bool
> fst::DeterminizeStar(F&, fst::MutableFst*, float, bool*, int, bool)
> bool DeterminizeStar(F &ifst, MutableFst
>
>
> *ofst, ^ ../fstext/determinize-star-inl.h:632:6: note: template argument
> deduction/substitution failed: ../fstext/determinize-star-inl.h: In
> substitution of ‘template bool fst::DeterminizeStar(F&, fst::MutableFst*,
> float, bool
>
>
>
>
>
> *, int, bool) [with F = fst::ArcTplfst::TropicalWeightTpl<float >]’:
> determinize-star-test.cc:48:65: required from ‘void
> fst::TestDeterminizeGeneral() [with Arc =
> fst::ArcTplfst::TropicalWeightTpl<float >]’
> determinize-star-test.cc:602:46: required from here
> ../fstext/determinize-star-inl.h:632:6: error: no type named ‘Arc’ in
> ‘class fst::ArcTplfst::TropicalWeightTpl<float >’ determinize-star-test.cc:
> In instantiation of ‘void fst::TestDeterminizeGeneral() [with Arc =
> fst::ArcTplfst::TropicalWeightTpl<float >]’:
> determinize-star-test.cc:602:46: required from here
> ../fstext/determinize-star-inl.h:646:6: note: template bool
> fst::DeterminizeStar(F&, fst::MutableFst >*, float, bool
>
>
>
> *, int, bool) bool DeterminizeStar(F &ifst, ^
> ../fstext/determinize-star-inl.h:646:6: note: template argument
> deduction/substitution failed: ../fstext/determinize-star-inl.h: In
> substitution of ‘template bool fst::DeterminizeStar(F&, fst::MutableFst >*,
> float, bool
>
>
>
>
>
>
>
> *, int, bool) [with F = fst::ArcTplfst::TropicalWeightTpl<float >]’:
> determinize-star-test.cc:48:65: required from ‘void
> fst::TestDeterminizeGeneral() [with Arc =
> fst::ArcTplfst::TropicalWeightTpl<float >]’
> determinize-star-test.cc:602:46: required from here
> ../fstext/determinize-star-inl.h:646:6: error: no type named ‘Arc’ in
> ‘class fst::ArcTplfst::TropicalWeightTpl<float >’ In file included from
> determinize-star-test.cc:20:0: ../base/kaldi-math.h:126:17: warning:
> ‘kaldi::kLogZeroBaseFloat’ defined but not used [-Wunused-variable] const
> BaseFloat kLogZeroBaseFloat = -std::numeric_limits::infinity(); ^ make[1]:
> *** [determinize-star-test] Error 1
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/208#issuecomment-149669509>.
>



-- 
- Hainan

danpovey:
Closing this old version of the pull request.

PR #210, opened 2015-10-05
Fixed strict-aliasing issue in const ARPA related code; Also fixed so…

chenguoguo:
…me style issue that I saw in the related code

### DIFF
--- a/src/lm/Makefile
+++ b/src/lm/Makefile
@@ -1,6 +1,7 @@
 EXTRA_CXXFLAGS = -Wno-sign-compare
 
 all:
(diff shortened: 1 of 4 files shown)


danpovey:
... sorry, see latest email.
It looks like it would be better after all to change the code not turn off the warning, as the compiler can generate wrong code or take a performance hit due to not enough optimization.

chenguoguo:
Agreed. I turned to the union solution because of that. The other option is
to use char, but that requires more code changing. Let's use the union
solution then.

Guoguo

PR #211, opened 2015-10-05
Fixed strict-aliasing issue using union in const ARPA related code; A…

chenguoguo:
…lso fixed some style issue that I saw in the related code. Tested this on WSJ and I was able to generate the same binary G.carpa as before.

### DIFF
--- a/src/lm/const-arpa-lm.cc
+++ b/src/lm/const-arpa-lm.cc
@@ -17,7 +17,10 @@
 // See the Apache 2 License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
+#include <limits>
 #include <sstream>
+#include <utility>
 
 #include "lm/const-arpa-lm.h"
 #include "util/stl-utils.h"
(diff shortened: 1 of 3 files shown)


src/lm/const-arpa-lm.cc
```diff
@@ -937,8 +944,10 @@ void ConstArpaLm::WriteArpaRecurse(int32* lm_state,
   // Inserts the current LmState to <output>.
   ArpaLine arpa_line;
   arpa_line.words = seq;
-  arpa_line.logprob = *reinterpret_cast<float*>(lm_state);
-  arpa_line.backoff_logprob = *reinterpret_cast<float*>(lm_state + 1);
+  Int32AndFloat logprob_i(*lm_state);
+  arpa_line.logprob = logprob_i.f;
```
danpovey: I don't see why a temporary is needed...  just use lm_state->f.

chenguoguo: lm_state has type of int32*. Or perhaps you mean casting lm_state to
Int32AndFloat using something like ((union Int32AndFloat *)lm_state)->f?
That may still cause strict aliasing problem?

Guoguo

danpovey: OK, I see.  I had assumed that you had redefined lm_state to be that union.
Probably the compiler can optimize this fine..
did you see whether it affects speed?  [I don't expect it will, but just
wonder, if it's easy to test.]


Dan

chenguoguo:
My reply somehow didn't get posted here, not sure if you received.

"On WSJ it doesn't seem to have an impact on the speed."

danpovey:
OK thanks, merging then.

Issue #207, opened 2015-10-05
CTC tuning issue


danpovey:
@vijayaditya  - you said you were going to set up CTC for Switchboard - there is something that needs to be done first.
In the version of CTC that I've implemented - I'm calling it CCTC - there is a phone language model involved.  We don't want this phone language model to have too many history-states.  The program
ctc-init-transition-model
has an option --state-count-cutoff2plus (default: 200) that can be modified to ensure this.  [note: I haven't tuned any of its options... setting --ngram-order to 3 (default: 2) might help.]
With more data, we'll tend to get more history-states, and this will get unmanageable (we probably already have too many in the WSJ setup, at 2000).  Note: the output-layer size equals num-tree-leaves plus num-history-states.
The way I think this should be set up is to aim for a target number of history-states (e.g. 500 or 1000).  However, implementing this in code would be a pain.  What I think would be easiest- for now, at least- would be to set off a number of ctc-init-transition-model jobs with different cutoffs, and to select the one that has the closest number of history-states to the target.  You might just have to have a variable like cutoff_array="100 150 200 300 400 600 800 1000 1500 2000 3000" and try all those values [just make them separate jobs], and then
target_history_states=500
or whatever.  Note: you can run ctc-transition-model-info to see the number of history states.  You might have to write a python script to select the closest one.

Of course, in the short term it would be easiest to just tune the cutoff manually.

Dan

danpovey:
Mistake: default --ngram-order is 3, not 2 (which should have been obvious given that I was getting num-history-states more than num-phones).  Setting it to 4 might help.

PR #214, opened 2015-10-06
Added basic lstm-ctc script. LSTM and LSTM-CTC recipes for SWBD

vijayaditya:
This PR doesn't yet address #207 .  It has been created for collaboration with other members making similar efforts.

The recipe is still being run, so it will be updated as bugs are found.

To checkout this pull request locally see https://help.github.com/articles/checking-out-pull-requests-locally/

### DIFF
--- a/egs/swbd/s5c/local/ctc/run_lstm.sh
+++ b/egs/swbd/s5c/local/ctc/run_lstm.sh
@@ -1,23 +1,7 @@
 #!/bin/bash
 
-# Copyright 2015  Johns Hopkins University (Author: Daniel Povey).
-#           2015  Vijayaditya Peddinti
-#           2015  Xingyu Na
-#           2015  Pegah Ghahrmani
-# Apache 2.0.
-
-
-# this is a basic lstm script
-# LSTM script runs for more epochs than the TDNN script
-# and each epoch takes twice the time
-
-# At this script level we don't support not running on GPU, as it would be painfully slow.
-# If you want to run without GPU you'd have to call lstm/train.sh with --gpu false
-
 stage=0
 train_stage=-10
-has_fisher=true
-use_sat_alignments=true
 affix=
 speed_perturb=true
 common_egs_dir=
@@ -30,7 +14,7 @@ cell_dim=1024
 hidden_dim=1024
 recurrent_projection_dim=256
 non_recurrent_projection_dim=256
-chunk_width=20
+chunk_width=50
 chunk_left_context=40
 clipping_threshold=30.0
 norm_based_clipping=true
@@ -38,13 +22,13 @@ norm_based_clipping=true
 # natural gradient options
 ng_per_element_scale_options=
 ng_affine_options=
-num_epochs=4
+num_epochs=10
 
 # training options
 initial_effective_lrate=0.0003
 final_effective_lrate=0.00003
-num_jobs_initial=3
-num_jobs_final=15
+num_jobs_initial=2
+num_jobs_final=12
 momentum=0.5
 adaptive_shrink=true
 shrink=0.98
@@ -57,6 +41,9 @@ remove_egs=true
 extra_left_context=
 frames_per_chunk=
 
+# ctc options
+treedir=exp/ctc/tri5b_tree
+
 # End configuration section.
 
 echo "$0 $@"  # Print the command line for logging
@@ -64,6 +51,7 @@ echo "$0 $@"  # Print the command line for logging
 . cmd.sh
 . ./path.sh
 . ./utils/parse_options.sh
+
 
 if ! cuda-compiled; then
   cat <<EOF && exit 1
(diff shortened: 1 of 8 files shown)


danpovey:
I'll merge this now and check it later.

egs/swbd/s5c/local/nnet3/run_lstm.sh
```diff
@@ -19,36 +31,43 @@ hidden_dim=1024
 recurrent_projection_dim=256
 non_recurrent_projection_dim=256
 chunk_width=20
-chunk_left_context=20
-clipping_threshold=5.0
+chunk_left_context=40
```
danpovey: we're using chunk-left-context=40 now?

vijayaditya: Yes this was slightly better.

Vijay

danpovey: OK.  It might be worthwhile looking again at increasing the frames-per-eg, then (e.g. to 30?), and maybe reducing the minibatch size or learning rate slightly if you worry about stability (at least, until @xiaohui-zhang  checks in his changes to the max-change code).  It just seems to me that having the left-context much more than the frames-per-eg is a bit wasteful.

vijayaditya: This was a temporary thing until the state preservation feature was
implemented.

But I will try out chunk-width of 30 and left context of 20. With momentum
the instabilities were not seen. So I will just try this setting without
modifying chunks-per-minibatch or learning rate.

gmkim90:
Thanks to share the script !  This is exact script what I was looking for. 

About 4months ago, I implemented my own CTC for nnet2, and test it with DNN(non-recurrent)-CTC 
but for almost all utterances, p(blank symbol) >> p(non-blank symbol) during whole utterance. My conjecture is that CTC requires RNN or very deep(so that it can utilize long enough context) DNN to work.

danpovey:
In Google, during decoding they scale blank probabilities down by a factor
of 10 (i.e. multiply by 0.1).  In my scripts, after tuning that scale, I
had to use a less aggressive value of 0.4 (i.e. close to 1/2), but what I
implemented is not vanilla CTC, but an extension that I'm calling CCTC
(described in the link), so I was expecting to have to do less aggressive
scaling of the blank probability.
Dan

gmkim90:
For google's work, you are referring this paper : "Fast and Accurate RNN AM for Speech Recognition"(http://arxiv.org/pdf/1507.06947.pdf), right?

danpovey:
Yes.  It says "In decoding, we only scale the blank label posterior by a
constant scalar decided by cross-validation on a held-out set.".  The paper
doesn't seem to mention it, but I think that constant scalar is either 1/9
or 1/10.  I either heard this from one of the authors, or from some other
paper.
Dan

egs/swbd/s5c/local/nnet3/run_lstm.sh
vijayaditya: %WER 44.1  VD-LSTM with chunk-width 30 and chunk-left-context 30 (Total training time is 17:04:14)
%WER 43.8  VD-LSTM with chunk-width 20 and chunk-left-context 40 (Total training time is 18:43:35)

Not much benefit in terms of speed and slightly worse results. 

I ensured that the frames-per-iter is the same in both cases by using 
13334 chunks per iteration x 30 chunk-width
20000 chunks per iteration x 20 chunk-width

VD-LSTM : Variable delay LSTM

gmkim90:
I have read Google's recent papers dealing with CCTC and running script of ctc_swbd branch. (download from https://github.com/vijayaditya/kaldi/tree/ctc_swbd)

I have one question : which version of cctc does this code use? I wonder current swbd script use all the newly introduced techniques in KALDI CCTC (of hidden documentation at http://kaldi-asr.org/doc2/ctc.html) uploaded by manager. Specifically, does this code use feature listed below?
 - frame decimation
 - left-context phone unit
 - phone level LM
 - CTC probability normalization
 - LM correction factor at decoding

danpovey:
>
> I have read Google's recent papers dealing with CCTC and running script of
> ctc_swbd branch. (download from
> https://github.com/vijayaditya/kaldi/tree/ctc_swbd)
>
The Google papers were about CTC.  CCTC is an acronym I coined to describe
what I actually implemented, which is a kind of extension, including the
things that you describe below, and as described in
http://kaldi-asr.org/doc2/ctc.htm <http://kaldi-asr.org/doc2/ctc.html>.
All of those things are implemented in the code in the ctc branch.

Bear in mind that it is still being developed- while it's running OK with
regular DNNs, and we haven't got the CTC+LSTM working yet (we're still
tuning the scripts).
Dan

I have one question : which version of cctc does this code use? I wonder

gmkim90:
Thank you for reply.

As far as I understand, ctc branch provide script for wsj (and may use all of the features in homepage documentation) and ctc_swbd provide script for swbd by @vijayaditya. 
I just wonder whether ctc_swbd use same features so that I can directly compare ctc_swbd's script and documentation in homepage. I am checking them by myself now. 

Anyway, thanks! It is really interesting script&code to analyze.

PR #216, opened 2015-10-07
Make join_suffix.py python3 compatible

jzombi:
join_suffix.py works only with python2, this change makes it compatible with both python2 and python3

(tested with python 2.7.10 and 3.5)

### DIFF
--- a/egs/tedlium/s5/local/join_suffix.py
+++ b/egs/tedlium/s5/local/join_suffix.py
@@ -5,9 +5,10 @@
 
 
 import sys
+from codecs import open
 
 words = set()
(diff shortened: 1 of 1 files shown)


jtrmal:
Thanks for the patch. As a rule of thumb, we do not aim to be python3
compatible (I don't even know if it's always possible), so do not be
surprised if you find more problems (especially in the DNN training). Feel
free to submit a patch for those as well.
Merging.
yenda

jzombi:
Thanks. I'll try to fix at least those that I encounter.

2015-10-07 13:15 GMT+02:00 jtrmal <notifications@github.com>:

PR #241, opened 2015-10-13
Added ctc+tdnn recipe for SWBD

vijayaditya:
Not for merge !


danpovey:
This looks OK to me.  You could start running it now.

naxingyu:
My run on final lattice score gives:
ERROR (lattice-align-words:ExpectToken():io-funcs.cc:201) Expected token "<TransitionModel>", got instead "<CctcTransitionModel>".

danpovey:
lattice-align-words is for regular transition models, which the CTC model
lacks.   Try to skip that part of the pipe, it's only a small effect on the
scoring.  You might have to add an option to the scoring script.
Dan

PR #244, opened 2015-10-14
SRE10 example: Various improvements, mostly cosmetic, to the DNN-based speaker recognition system

david-ryan-snyder:
* Removed the requirement that --num_components (of the DNN) be passed to sid/init_full_ubm_from_dnn.sh. It is now parsed from nnet-am-info.

*  Modified src/fgmmbin/fgmm-global-init-from-accs.cc to first floor the covariance eigenvalues prior to inverting. 

* Fixed bugs in optional parts of DNN training recipe.

* Added a README to egs/sre10/v2/

### DIFF
--- a/egs/sre08/v1/sid/init_full_ubm_from_dnn.sh
+++ b/egs/sre08/v1/sid/init_full_ubm_from_dnn.sh
@@ -13,7 +13,6 @@ cmd="run.pl"
 stage=-2
 delta_window=3
 delta_order=2
-num_components=5297
 # End configuration section.
 
 echo "$0 $@"  # Print the command line for logging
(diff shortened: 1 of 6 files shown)


danpovey:
I just merged this but then noticed that the flooring code was not right.  I won't revert it though because it wouldn't break anything.  Here
+      // Floor variance Eigenvalues.
+      BaseFloat floored = std::max(
+          static_cast<BaseFloat>(gmm_opts.variance_floor),
+          static_cast<BaseFloat>(covar.MaxAbsEig() / gmm_opts.max_condition));
+      if (floored) {
+        tot_floored += floored;
+        gauss_floored++;
+      }
+      covar.InvertDouble();
it looks like you're not actually flooring it.  You're computing a floor value (call this floor instead of 'floored'), but then you need to do
  int32 floored = covar.ApplyFloor(floor);

PR #245, opened 2015-10-14
Fixing bug in fgmm-global-init-from-accs.cc

david-ryan-snyder:
Fixing bug in fgmm-global-init-from-accs.cc that was just introduced in #244.


danpovey:
Thanks!

PR #246, opened 2015-10-14
1509

xiaohui-zhang:


src/nnet3/nnet-training.h
```diff
@@ -155,6 +155,11 @@ class NnetTrainer {
   unordered_map<std::string, ObjectiveFunctionInfo, StringHasher> objf_info_;
 };
 
+class NnetPerturbedTrainer : public NnetTrainer {
```
danpovey: did you include this by accident?

src/nnet3/nnet-training.h
```diff
@@ -135,7 +135,7 @@ class NnetTrainer {
   bool PrintTotalStats() const;
 
   ~NnetTrainer();
- private:
+ protected:
```
danpovey: and this?

xiaohui-zhang: added that by accident.. ignore this pull request plz..

PR #248, opened 2015-10-14
removed the per-component max-change (gradient clipping) mechanism.

xiaohui-zhang:
We don't need the old max-change mechanism for nnet3 neural network training, since we've switched to a standard Frobenius norm based max-change mechanism done in nnet3/nnet-training.cc


xiaohui-zhang:
ready to merge

xiaohui-zhang:
ready to merge

PR #252, opened 2015-10-14
Cleaning up rnnlm related stuff in WSJ; also adding rnnlm lattice res…

chenguoguo:
…coring and updating RESULTS

Tested on my old models and updated the results.


danpovey:
Thanks!
Merging.

PR #254, opened 2015-10-15
compute-wer : adding verbose logprints of sentence alignment stats

vesis84:
- the per-sentence logprints go into the logfile (the overall WER is at stdout).

### DIFF
--- a/src/bin/compute-wer.cc
+++ b/src/bin/compute-wer.cc
@@ -139,6 +139,12 @@ int main(int argc, char *argv[]) {
         num_ins += ins;
         num_del += del;
         num_sub += sub;
+      
+        if (kaldi::g_kaldi_verbose_level >= 1) { 
+          std::cerr << key << " wer " << 100.0*(sub+del+ins)/ref_sent.size() 
+                    << " ref " << ref_sent.size() << " ins " << ins
+                    << " del " << del << " sub " << sub << std::endl;
+        }
 
         if (detailed_stats) {
           const int32 eps = -1;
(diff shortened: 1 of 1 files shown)


danpovey:
Karel, I'd like @jtrmal to look at this because he has looked at the scoring tools most recently and I know there are capabilities for more verbose output already, as used by steps/score_kaldi.sh (IIRC).
Also I'm not super hot on the use of std::cerr for what seems to be not true logging output, but a kind of output that seems possibly designed for further computer processing-- I think there are other, better ways to get this output already, which @jtrmal might be able to comment on
Do  you have a use case for this?

danpovey:
Karel, it looks to me like the program 'compute-wer' already supports detailed output similar to (actually, almost identical to) what you are adding, and this can be enabled by adding one extra argument to the program.  Closing this pull request.

danpovey:
OK, I see now that the detailed stats that it outputs are not the same as what you are outputting; they are word-level, not utterance-level stats.
However, if you want to add something like this, I think it should have a design that's more consistent with the current setup.  For example, add another optional argument giving the filename to print the per-utterance stats to.

jtrmal:
Karel,
If you look at the scoring script in egs/rm (and perhaps swbd and wsj)  you
will see  how to get sentence level stats and speaker level stats.
I'm flying today,  so if there is some extra feature  you need,  we can
talk about it on Fr.
Y.

vesis84:
Hi,
yes, I have noticed that there is the support for saving the alignment stats. But it was not obvious to me how to get them per-sentence. So I thought the verbose print could be a good fit, as it is not active by default.
While looking into the tool, it seemed to me that it is getting a bit too messy for a relatively simple tool. Perhaps we could load and compare the int vectors as strings to avoid the big 'if' which splits the case of text or integer inputs. Sure, it will be a little slower, but I think speed is not an issue here...
Does it sound good to you?
Cheers!
K.

vesis84:
Good, thanks! Found the scoring scripts for getting the stats. Thanks Yenda! K.

danpovey:
I agree that loading the int vectors as strings would be a welcome
simplification.  We could change the usage message for the "--text" option
to state that it now has no effect.
Dan

PR #255, opened 2015-10-15
add ctc option for scoring

naxingyu:
Add --use-ctc option for scoring scripts.


danpovey:
Vijay, I just saw this thread-- why don't you take his patch, fix it
yourself and make a new pull request?
Dan

vijayaditya:
PR #259 submitted including the changes from this PR. Output from nnet3-ctc-info is used to decide if the model is CTC model, rather than providing an option in the scripts.

naxingyu:
Got it. Thanks. I'll close this one.

PR #256, opened 2015-10-15
make bidirectional-LSTM config file, add nnet3 BLSTM and LSTM test sc…

freewym:
…ripts for wsj along with their results.

Hi Vijay,

I can also send you the config file generated by the BLSTM make_configs.py. Let me know if you would like to take a look at it.

The current conflict is probably due to the RESULTS file, that was edited after my most recent pull from master. Will resolve it in the end.

Thanks,
Yiming

### DIFF
--- a/egs/wsj/s5/RESULTS
+++ b/egs/wsj/s5/RESULTS
@@ -281,3 +281,18 @@ for x in exp/nnet2_online/nnet_ms_a_online/decode_*rnnlm-hs.nce20.h400.voc40k; d
 %WER 2.91 [ 164 / 5643, 24 ins, 9 del, 131 sub ] exp/nnet2_online/nnet_ms_a_online/decode_bd_tgpr_eval92_rnnlm-hs.nce20.h400.voc40k/wer_10_0.0
 %WER 7.83 [ 645 / 8234, 159 ins, 50 del, 436 sub ] exp/nnet2_online/nnet_ms_a_online/decode_tgpr_dev93_rnnlm-hs.nce20.h400.voc40k/wer_11_0.0
 %WER 5.40 [ 305 / 5643, 77 ins, 16 del, 212 sub ] exp/nnet2_online/nnet_ms_a_online/decode_tgpr_eval92_rnnlm-hs.nce20.h400.voc40k/wer_13_1.0
+
+# LSTM results:
+for x in exp/nnet3/lstm_ld5/decode_*; do grep WER $x/wer_* | utils/best_wer.sh ; done
+%WER 7.07 [ 582 / 8234, 61 ins, 92 del, 429 sub ] exp/nnet3/lstm_ld5/decode_bd_tgpr_dev93/wer_11_0.0
+%WER 4.27 [ 241 / 5643, 19 ins, 27 del, 195 sub ] exp/nnet3/lstm_ld5/decode_bd_tgpr_eval92/wer_11_1.0
+%WER 9.50 [ 782 / 8234, 155 ins, 76 del, 551 sub ] exp/nnet3/lstm_ld5/decode_tgpr_dev93/wer_11_0.0
+%WER 6.84 [ 386 / 5643, 87 ins, 23 del, 276 sub ] exp/nnet3/lstm_ld5/decode_tgpr_eval92/wer_11_1.0
+
+# BLSTM results
+for x in exp/nnet3/blstm_ld5/decode_*; do grep WER $x/wer_* | utils/best_wer.sh ; done
+%WER 6.79 [ 559 / 8234, 55 ins, 89 del, 415 sub ] exp/nnet3/blstm_ld5/decode_bd_tgpr_dev93/wer_11_0.0
+%WER 4.38 [ 247 / 5643, 18 ins, 29 del, 200 sub ] exp/nnet3/blstm_ld5/decode_bd_tgpr_eval92/wer_12_1.0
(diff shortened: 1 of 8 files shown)


egs/wsj/s5/local/nnet3/run_lstm.sh
```diff
@@ -0,0 +1,127 @@
...
+
+num_epochs=10
+
+# training options
+initial_effective_lrate=0.0003
+final_effective_lrate=0.00003
+num_jobs_initial=2
+num_jobs_final=12
+momentum=0.5
+shrink=0.98
```
vijayaditya: Check the latest swbd lstm recipe for new default settings. The shrink value is 0.99 in the new defaults, as we are using conditional shrinking, based on avg-derivative values at the sigmoid non-linearities.

egs/wsj/s5/RESULTS
```diff
@@ -277,4 +277,16 @@ for x in exp/nnet2_online/nnet_ms_a_online/decode_*; do grep WER $x/wer_* | util
 %WER 6.68 [ 377 / 5643, 102 ins, 13 del, 262 sub ] exp/nnet2_online/nnet_ms_a_online/decode_tgpr_eval92_utt/wer_10
 %WER 6.56 [ 370 / 5643, 100 ins, 12 del, 258 sub ] exp/nnet2_online/nnet_ms_a_online/decode_tgpr_eval92_utt_offline/wer_10
 
-
+# LSTM results:
```
vijayaditya: These results are not comparable to the nnet2_online(old tdnn) results above as we are using model-discarding in the new TDNN recipe. Could you run the TDNN recipe with the new steps/nnet3/train_tdnn.sh and add the results here.

egs/wsj/s5/local/nnet3/run_lstm.sh
```diff
@@ -0,0 +1,127 @@
...
+cell_dim=1024
+hidden_dim=1024
+recurrent_projection_dim=256
+non_recurrent_projection_dim=256
+chunk_width=20
+chunk_left_context=40
+clipping_threshold=30.0
+norm_based_clipping=true
+
+num_epochs=10
```
vijayaditya: We normally use 8 epochs for TDNN training with WSJ data. From other recipes we can see that LSTM is converging slowly compared to the TDNN recipe. I would suggest increasing the number of epochs for LSTM and BLSTM. Try 16. Plot the training and CV accuracies from the compute_prob_*.logs and check if the accuracies have saturated.

egs/wsj/s5/local/nnet3/run_lstm.sh
freewym: I checked the latest swbd lstm recipe in the master branch, but the defalut shrink value is still 0.98?

danpovey: Guys, some comments on this--
 - If convergence is slow, you could just increase the learning rates,
assuming you are using all the changes in the current master (e.g. relating
to how shrinkage is applied, and the max-change).  We don't have divergence
problems now.
 - I don't understand why you need a separate set of scripts for BLSTMs.
Isn't it just a question of putting some positive values in the
--lstm-delay option, and adding an --extra-right-context option to the
script?

vijayaditya: This was more of an exercise for Yiming to understand the make_configs.py scripts. Once the recipe is finalized we will just make the BLSTM an option in LSTM scripts.

naxingyu: I'm trying defining a bidirectional layer with (-1, +1) other than 
defining two single directional layers with (-1) and (1).
It requires the gate component-node to sum three components, such as:
```
component-node name=Lstm1_if1 component=Lstm1_W_i-xr 
input=Append(L0_lda, IfDefined(Offset(Lstm1_r_t, -1)))
component-node name=Lstm1_if2 component=Lstm1_w_ic 
input=IfDefined(Offset(Lstm1_c_t, -1))
component-node name=Lstm1_ib1 component=Lstm1_W_i-xr 
input=Append(L0_lda, IfDefined(Offset(Lstm1_r_t, 1)))
component-node name=Lstm1_ib2 component=Lstm1_w_ic 
input=IfDefined(Offset(Lstm1_c_t, 1))
component-node name=Lstm1_i_t component=Lstm1_i input=Sum(Lstm1_if1, 
Lstm1_if2, Lstm1_ib1, Lstm1_ib2)
```
This is not like the one in Alex (ASRU2013).
Or, if I use the current lstm maker, with ```--lstm-delay "-1 1 -1 1"``` 
and some extra right context, I still need to pass the output of the 
first layer to the input of the third layer, right?

Xingyu

danpovey: Xingyu, I think a bidirectional layer like that won't be definable, as it
will compile to a cyclic graph, and the tool will just reject it.  Unless
I'm missing something, there is no way to interpret something like that as
a forward type of computation.

naxingyu: I see. You are right, it'll compile to a cyclic graph.
Then I should go for the --lstm-delay "-1 1 -1 1" plan.
In that way, lstm2 should take input from "input" other than "lstm1", so 
that lstm1 and lstm2 make a bidirectional layer.
Then, lstm3 takes input from both lstm1 and lstm2, such as
```
component-node name=Lstm3_i1 component=Lstm3_W_i-xr 
input=Append(Lstm1_rp_t, Lstm2_rp_t, IfDefined(Offset(Lstm3_r_t, -1)))
```
Does it make sense?

danpovey: Yes.

PR #257, opened 2015-10-15
featbin/extract-feature-segments.cc: float to int conversion fix and a new option

alumae:
Fixing float to int32 conversion, and adding an option to trim start times by window_length-window_period.

This allows the lengths of extracted feature segments to match with the lengths of feature vectors that are calculated from already segmented audio.

### DIFF
--- a/src/featbin/extract-feature-segments.cc
+++ b/src/featbin/extract-feature-segments.cc
@@ -45,6 +45,8 @@ int main(int argc, char *argv[]) {
     BaseFloat min_segment_length = 0.1,  // Minimum segment length in seconds.
         max_overshoot = 0.0;  // max time by which last segment can overshoot
     BaseFloat samp_freq = 100;  // feature sampling frequency (assuming 10ms window shift)
+    int32 frame_length = 25;
+    bool trim_starts = false;
 
     // Register the options
     po.Register("min-segment-length", &min_segment_length,
@@ -54,6 +56,14 @@ int main(int argc, char *argv[]) {
     po.Register("max-overshoot", &max_overshoot,
                 "End segments overshooting by less (in seconds) are truncated,"
                 " else rejected.");
+    po.Register("frame-length", &frame_length,
+                "Frame length in ms, used when --trim-starts=true");
+    po.Register("trim-starts", &trim_starts,
+                "Trim segment start by (frame_length - frame_shift)/frame_shift samples,"
+                " where frame_shift = 1000/frame_rate "
+                " except for segment starts at the beginning of a file."
+                " This way the segment lengths will match with feature vector lengths "
+                " calculated from already segmented audio");
 
     // OPTION PARSING ...
     // parse options  (+filling the registered variables)
@@ -63,7 +73,6 @@ int main(int argc, char *argv[]) {
       po.PrintUsage();
       exit(1);
     }
-
 
     std::string rspecifier = po.GetArg(1); // get script file/feature archive
     std::string segments_rxfilename = po.GetArg(2);// get segment file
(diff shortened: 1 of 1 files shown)


src/featbin/extract-feature-segments.cc
```diff
@@ -45,6 +45,8 @@ int main(int argc, char *argv[]) {
     BaseFloat min_segment_length = 0.1,  // Minimum segment length in seconds.
         max_overshoot = 0.0;  // max time by which last segment can overshoot
     BaseFloat samp_freq = 100;  // feature sampling frequency (assuming 10ms window shift)
```
danpovey: This is an 'orphan' program, it is used nowhere in the scripts.  Therefore we can make non-back-compatible changes.
The samp-freq option should be changed to --frame-shift -- obviously this changes its meaning -- and your --trim-starts option should be changed to --snip-edges, both to match with things like compute-mfcc-feats.  (note: using --snip-edges=false when dumping features will make your life easier, but we can support the option here too).  And try to make their documentation compatible with that in compute-mfcc-feats.

src/featbin/extract-feature-segments.cc
```diff
@@ -133,10 +149,13 @@ int main(int argc, char *argv[]) {
       const Matrix<BaseFloat> &feats = feat_reader.Value(utterance);
       int32 num_samp = feats.NumRows(), // total number of samples present in wav data
           num_chan = feats.NumCols(); // total number of channels present in wav file
-
       // Convert start & end times of the segment to corresponding sample number
-      int32 start_samp = static_cast<int32>(start * samp_freq);
-      int32 end_samp = static_cast<int32>(end * samp_freq);
+      int32 start_samp = static_cast<int32>(round((start * samp_freq)));
```
danpovey: 'round' behaves differently from static_cast, as it rounds to nearest.  Are you 100% sure this is correct?
Have you done any testing of this?

danpovey: Also, please co-ordinate carefully with the NumFrames() function.  If you can actually make use of that function directly, it would be ideal, as it would be ideal-- you could create a temporary object of type FrameExtractionOptions inside the code.  (you could just take the sampling rate to be 16k as it wouldn't affect the results if you assume an exact number of samples go into the frame shift and frame length).

src/featbin/extract-feature-segments.cc
alumae: Thanks, I didn't know about the --snip-edges option in compute-mfcc-feats. This actually explains why the extract-feature-segments program seems to have worked before without the need for such option.

src/featbin/extract-feature-segments.cc
alumae: Yes, round() fixes a problem that was caused by static_cast to sometimes cause an incorrect number of frames to be extracted. 

Will look at the NumFrames() function.

alumae:
I made the suggested changes. I couldn't use NumFrames() easily because I also need to fiddle with the start frame.

src/featbin/extract-feature-segments.cc
```diff
@@ -76,7 +89,12 @@ int main(int argc, char *argv[]) {
     Input ki(segments_rxfilename); // no binary argment: never binary.
 
     int32 num_lines = 0, num_success = 0;
-    
+
+    int32 snip_length = 0;
+    if (snip_edges) {
+      snip_length =  static_cast<int32>(ceil(1.0 * (frame_length - frame_shift) / frame_shift));
```
danpovey: Please be mindful of the 80-char limit.  you have to split to multiple lines.

danpovey:
@ngoel17, I think you've used 'extract-segments' in the past, do you have any comments on this?
And @alumae, can you describe how you've tested this to make sure it gives the correct segment sizes?

alumae:
The testing procedure: I have a dataset with 'segments' file. Features A are extracted he usual way, i.e., from segmented audio.

Then, I received different features, extracted from the same dataset, but from unsegmented audio. Used the extract-feature-segments to extract feature segments (features B) and noticed the lengths of A don't match with lengths in B (there was a 2 frame difference, *sometimes* 3). After the fix in the PR they match 100%.

danpovey:
Thanks!  Merging.

PR #262, opened 2015-10-16
Numeric sort on the probability columns.

francisr:
It is necessary because normalising the probability can change the order.


jtrmal:
Thank you for the patch. I have actually noticed this as well, but didn't have time to fix it yet.

Issue #223, opened 2015-10-09
tools/extras/check_dependencies.sh libtool issues


dwks:
One more issue: running "make" from a clean install in tools/ on Debian 8.2 fails because check_dependencies thinks I need the binary "libtool". It seems Debian now has "libtoolize" in the libtool package which generates local "libtool"s (https://github.com/zeromq/libzmq/issues/1385), and this is used by e.g. openfst's configure script. The full build works correctly if I check for "libtoolize" instead of "libtool" in tools/extras/check_dependencies.sh.

Otherwise, if I run "make" three separate times (or make -k) the build mostly completes, although it does not make the openfst symlink, so I have to make it myself or configure kaldi with --fst-root pointing at the right place. Note: there is also a Debian package "libtool-bin" which provides a system-wide "libtool" executable, and installing that will make the build succeed, although I don't think the executable is actually used because local "libtool"s are still created.

Maybe it would be best to test for either libtool or libtoolize. I'm not very familiar with autotools unfortunately.

Last comment: when building tools/ in parallel with make -j N, the check_dependencies is run at the same time as other configure scripts etc, which makes its output interleaved and difficult to spot. This will run check_dependencies first even if multiple jobs were specified:
```
diff --git a/tools/Makefile b/tools/Makefile
index cb3a35b..0c63b13 100644
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -34,6 +34,9 @@ all: check_required_programs sph2pipe atlas sclite openfst
 check_required_programs:
        extras/check_dependencies.sh

+# make sure check_required_programs runs before anything else:
+sph2pipe atlas sclite openfst sctk: | check_required_programs
+
 clean: openfst_cleaned sclite_cleaned

 openfst_cleaned:
@@ -72,7 +75,7 @@ openfst-$(OPENFST_VERSION)/lib: | openfst-$(OPENFST_VERSION)/Makefile

 # Add the -O flag to CXXFLAGS on cygwin as it can fix the compilation error   
 # "file too big".
-openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION)/.patched
+openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION)/.patched | check_required_programs
 ifeq ($(OSTYPE),cygwin)
[...]

PR #264, opened 2015-10-16
fixing issues with readlink -f under MacOSX (readlink there does not …

jtrmal:
fixing issues with readlink -f under MacOSX (readlink there does not accept -f, as reported in #223


jtrmal:
@danpovey I did check on linux (if it still does what is expected there).
Untested on MacOSX (but it should work, I tested if `readlink --invalid-switch $PWD || pwd` returns what is expected

PR #265, opened 2015-10-17
Adding a support for the PBS 

jtrmal:
Based on SGE's queue.pl


danpovey:
Merging this, since even if something is wrong with it, nothing will be broken.

PR #268, opened 2015-10-17
Allow nnet-am-average to skip any layer(s), not just the last one

jtrmal:
For the sake of simplicity, I index all components, not just the UpdatableComponents descendants.


jtrmal:
Merging. Was OK'd by Dan.

PR #273, opened 2015-10-19
Move parsing fucntions for opts to protected to allow sub-classing of…

matthewaylett:
… parse opts and flexibility for adding default parameters


matthewaylett:
Very small change to test how I get small supporting changes into repo.

In Idlak we need to allow sevral laevels of configuration files, my making these functions protected we can sub class parse options and retain the parsing code as it is.

matthewaylett:
Sorry, pushing the wrong buttons I think...

PR #274, opened 2015-10-19
matrix-scale, adding the tool (similar to vector-scale but for matrices)

vesis84:


danpovey:
The program "matrix-copy" already supports the --scale option.

PR #278, opened 2015-10-20
Improvements in configuring and installation of kaldi

dbdean:
Please accept my pull request for some improvements to kaldi.

This pull request adds some code to src/configure to search LD_LIBRARY_PATH for libatlas.a, and adds and 'install' target to src/Makefile that puts all the bins and the openfst libs in $PREFIX/{bin,lib} so that kaldi can be installed outside of the compile source tree.

Thanks.


src/configure
```diff
@@ -27,7 +27,7 @@ CONFIGURE_VERSION=2
 
 function rel2abs {
   if [ ! -z "$1" ]; then
-    local retval=`cd $1 2>/dev/null && pwd || exit 1`
```
jtrmal: this is intentionally written in this way -- MacOSX's readlink  does not accept -f switch

dbdean: Ok, I made the change because it failed for directories which don't yet exist, which I needed to specify an install prefix using --prefix.

src/configure
```diff
@@ -628,7 +632,8 @@ function linux_configure_static {
   if $threaded_atlas; then pt=pt; else pt=""; fi
 
   if [ -z $ATLASLIBDIR ]; then # Note: it'll pick up the last one below.
-    for dir in /usr{,/local}/lib{64,}{,/atlas,/atlas-sse2,/atlas-sse3} \
```
jtrmal: I don't think this is correct change -- LD_LIBRARY_PATH is a variable that should be used by linker. For linker during compilation, LIBRARY_PATH should be consulted.

src/configure
jtrmal: OK, then I guess the correct solution would be either test existence of the directory (and return 1, if it does not exist)

dbdean: The intention was to pass a directory to configure using --prefix=PREFIX, but that the directory may not already exist. I used the existing read_dirname function so as not to duplicate functionality, but discovered that it doesn't return the correct path if it doesn't already exist (which readlink -f) does.

Perhaps it would be ok if the prefix had to already exist, and then we don't have to make any changes to this function.

src/configure
```diff
@@ -82,6 +82,7 @@ static_fst=false
 use_cuda=true
 dynamic_kaldi=false
 mkl_threading=sequential
+prefix=/opt/kaldi
```
dbdean: Just noticed this should be INSTALLPREFIX to match the rest of my new code

src/configure
jtrmal: OK -- just print some informative message (or you can try create the
directory and call the read_dirname again)?
y.

src/configure
dbdean: I've changed it to LIBRARY_PATH in the latest commit

dbdean:
@jtrmal, I've made further changes to this pull request based upon your comments

src/Makefile
```diff
@@ -140,6 +141,13 @@ $(SUBDIRS) : mklibdir
 $(EXT_SUBDIRS) : mklibdir
 	$(MAKE) -C $@
 
+install: all
+	mkdir -p $(INSTALLPREFIX)/{bin,lib}
+	# copying binaries to $(INSTALLPREFIX)/bin
+	find $(SUBDIRS_BIN) -type f -executable -exec cp -v {} $(INSTALLPREFIX)/bin/ ';'
+	# copying libs to $(INSTALLPREFIX)/lib
+	cp -av ../tools/openfst/lib/* $(INSTALLPREFIX)/lib/
+
```
jtrmal: Is there a particular reason why you install the openfst libst, but not binaries?
I think some of them are used by the scripts in the egs/
I'm fairly indecisive what would be the correct thing to do, especially given the fact that the openfst we are using is patched.

dbdean: jtrmal, at this stage, I only installed the kaldi binaries and the libraries they depended upon (I used ldd to check all of them). If openfst binaries or other dependencies should be installed here, they can be added too, but I wasn't going to do so until our researchers complained that it wasn't working.

I think having a should-be working install target is a good start, and it can be improved later if something is missing.

My intention was that this should install to somewhere like /opt/kaldi rather than /usr so that it doesn't interfere with packages installed by the system and apt-get.

danpovey: Kaldi doesn't currently have a notion of 'install'.  If we were to do it
we'd have to think quite carefully about it.  Let's leave this till later.
Dan

In src/Makefile

posophe:
Thanks for your work, but could you please consider adding $BUILDROOT support ? Otherwise, it won't be usable by maintainers.

jtrmal:
Can you please explain what that option should do?
Kaldi is not exactly intended to be installed or distributed as a package,
so unless what you ask for will be more or less trivial thing, I don't
think anyone will have the time to implement it.
y.

posophe:
I meant DESTDIR, sorry.

Well, it's relatively common. It's just an option to define an additional directory where Kaldi will be installed. Concretely, make DESTDIR=$BUILDROOT install will install Kaldi in
$BUILDROOT/$PREFIX. The build systems like OBS use chroot as build environment so it's obviously not possible to install a soft in /usr directly.

jtrmal:
I don't think it viable -- look, patching OpenFST is done in order to
modify it's behavior in certain ways.  The fact it compiles with 1.5 does
not mean the OpenFST now behaves the way we need (and it certainly does not
behave the way we need in 1.3.4 nor 1.4.1 -- that is the reason we patch
it).
So I don't think it's reasonable to provide a way how the users/system
packagers could mess up the system libraries (be overriding the "correct"
openfst libs by ours).

Perhaps, the only sensible way how to deal with this is to patch the
OpenFST build system as well, to create libraries with different names, say
"libopenfst-kaldi" or whatever, so that these patched libraries will not
collide with the "correct" openfst.

@danpovey any other comments on this?
y.

posophe:
The only correct way is to submit your changes/fixes to openfst upstream

jtrmal:
@danpovey did it and for some reason it didn't make it through.

danpovey:
They don't object to my changes, it's just that they haven't had the time
to review them.  (it's been a couple years, so I'm not holding my breath).

Kaldi doesn't support the notion of building it elsewhere than the code
location, and I don't think it would be useful to actual Kaldi users-
scripts would have to be changed and it would be a hassle.  Sorry.

Dan

danpovey:
OK, I think I see where all this is coming from.  You are the sysadmin at QUT who was asked by Ahilan to install Kaldi on the HPC cluster, and you're trying to do it using some kind of notion of 'package'.
Now, I assume the HPC cluster has some kind of user data-space that's exported as some kind of directory.  What I think would make more sense is for him to get an interactive session on the HPC cluster (so he can see all the libraries that the cluster provides), and within that session, to run through the standard Kaldi installation process in his data-space.

Kaldi isn't really designed to be installed at the system level because it's intended that the users may frequently modify code and update their repository, which in a multi-user system would not work well if installed in system space.

Dan

naxingyu:
I just report my case here. We have a cluster of tens of stations, NFS 
and GE enabled. I compiled Kaldi on a directory like /software/kaldi/ 
and every authenticated user in the cluster set their KALDI_ROOT to this 
dir. What I do-- sort of maintaining-- is recompiling the libraries and 
binaries once every month to keep it up-to-date. We have around a dozen 
users and it seems working okay, for average users, not developers.

Xingyu

danpovey:
Closing this as we won't be making any changes to the build process in the near future.

Issue #89, opened 2015-08-24
Remove unnecessary null pointer checks


elfring:
[Extra null pointer checks are not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for NULL before delete p?") like the following.
* [DeleteTestFst](https://github.com/kaldi-asr/kaldi/blob/b347cf82d304ad7f36499a073455662a589baf50/src/fstext/deterministic-fst-test.cc#L111 "DeleteTestFst function")
* [Fbank](https://github.com/kaldi-asr/kaldi/blob/36cad857c7e5c1c7ee00866ce96e2afdc4823a8b/src/feat/feature-fbank.cc#L41 "Destructor for Fbank")
* [LangModelFst](https://github.com/kaldi-asr/kaldi/blob/b347cf82d304ad7f36499a073455662a589baf50/src/lm/kaldi-lm.h#L74 "Destructor for LangModelFst")

danpovey:
Agreed.  I'll leave it as an open issue for now, since a lot of people would be able to submit a fix for this.

jtrmal:
Resolved, closing...

elfring:
Thanks for your small source code improvements.

elfring:
Were there a few update candidates left over?
* [LatticeStringRepository::Destroy](https://github.com/kaldi-asr/kaldi/blob/85a3dd5f0b71e419abf1169a26b759bfc423a543/src/fstext/determinize-lattice-inl.h#L171 "LatticeStringRepository::Destroy function")
* [OnlineFeaturePipeline::SetTransform](https://github.com/kaldi-asr/kaldi/blob/85a3dd5f0b71e419abf1169a26b759bfc423a543/src/online2/online-feature-pipeline.cc#L223 "OnlineFeaturePipeline::SetTransform function")
* [TableMatcherImpl::SetState](https://github.com/kaldi-asr/kaldi/blob/85a3dd5f0b71e419abf1169a26b759bfc423a543/src/fstext/table-matcher.h#L103 "TableMatcherImpl::SetState function")

elfring:
How do you think about to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/readability-delete-null-pointer.html "Deletion of unnecessary null pointer checks")?

PR #281, opened 2015-10-20
Removing null-pointer checks as mentioned in #89

jtrmal:
I used the following shell code to find the offenders

find .  \( -name "*.h" -o -name "*.cc" \) | xargs grep -P  ".*if *\((.*)\).*delete.*\1.*"

might not cover all cases (but this is a low-importance subject anyways, I believe)


jtrmal:
I actually forgot to compile, so let's wait if all checks finish
y.

danpovey:
Thanks!

PR #283, opened 2015-10-21
WIP: ConvolutionComponent added to nnet3, to do 2D convolution

vijayaditya:
This is not ready for merge!! Update() method has not yet been implemented.

This PR has been added as I need some clarifications:

Clarifications needed:
1. Is Resize(input_dim, output_dim) an interface method which every component has to implement ? The resize of convolution component requires more info i.e., input_x_dim, input_y_dim and input_z_dim, filt_x_dim and filt_y_dim (also corresponding steps filt_x_step and filt_y_step if they are not compatible with the newly specified sizes). The output dim is determined based on the above specified parameters. Can I write a Resize method with this new signature ?
2. Is Update() an interface method or can I change its signature too ? I create an out_deriv_patches matrix in the Backprop method which can be reused for the Update() method so I would like to just pass this to the Update() after modifying its signature ?


danpovey:
Resize() is not an interface method, no.  I think it was added to AffineComponent for a special purpose, probably relating to 'mixing up'.  It's not important.
Update() is not an interface method- it's an internal method that we've sometimes found it convenient to add.  [this was to enable some code sharing in the backprop for the natural gradient and non-natural-gradient versions of affine components].

kkm000:
>This is not ready for merge!!

I think it might be a good idea to create a convention for work in progress PRs, for example, add `WIP:` to the PR title. When ready to go, remove the WIP prefix. This way maintainers would know this is not to merge. If @danpovey and @jtrmal agree, I'd document that in the contribution guide. (FWIW, GitLab enforces this by not showing the Accept button if the word WIP is in the title. I sent a feature request to GitHub a few months ago).

danpovey:
OK, sure.

vijayaditya:
#297 has been submitted in place of this PR.

PR #284, opened 2015-10-22
Add an option to discard any partial minibatches of uneven size that …

freewym:
…are encountered at the end in nnet3-merge-egs

### DIFF
--- a/src/nnet3bin/nnet3-merge-egs.cc
+++ b/src/nnet3bin/nnet3-merge-egs.cc
@@ -62,6 +62,7 @@ int main(int argc, char *argv[]) {
     bool compress = false;
     int32 minibatch_size = 512;
     bool measure_output_frames = true;
+    bool discard_partial_minibatches = false;
 
     ParseOptions po(usage);
     po.Register("minibatch-size", &minibatch_size, "Target size of minibatches "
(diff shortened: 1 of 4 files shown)


src/nnet3bin/nnet3-merge-egs.cc
```diff
@@ -105,8 +109,8 @@ int main(int argc, char *argv[]) {
       // Do Next() now, so we can test example_reader.Done() below .
       example_reader.Next();
       num_read++;
-      
-      if (minibatch_ready || (example_reader.Done() && !examples.empty())) {
+
+      if (minibatch_ready || (!discard_partial_minibatches && (example_reader.Done() && !examples.empty()))) {
```
danpovey: please split this line.

danpovey: Also, you can modify the nnet3 training scripts (TDNN and LSTM) to add the flag --discard-partial-minibatches=true to nnet3-train.  We don't need to make this configurable-- it won't appreciably affect the training, and it will increase speed.

freewym: Done.

danpovey:
Thanks!
Merging.

PR #290, opened 2015-10-23
WIP: Added function that returns sum of the elements raised to power p.

matthewaylett:
Appears to compile although there seems to be a separate linkiing error with kws-functions.cc:148: first defined here making it hard to be sure.

### DIFF
--- a/src/matrix/kaldi-vector.h
+++ b/src/matrix/kaldi-vector.h
@@ -310,6 +310,10 @@ class VectorBase {
   /// Returns sum of the elements
   Real Sum() const;
 
+  /// Returns sum of the elements raised to power p. For p>0, this is the same
+  /// as Norm(p) raised to the p-th power.
+  Real SumPower(Real p) const;
+
   /// Returns sum of the logs of the elements.  More efficient than
   /// just taking log of each.  Will return NaN if any elements are
   /// negative.
--- a/src/matrix/kaldi-vector.cc
+++ b/src/matrix/kaldi-vector.cc
@@ -696,6 +696,32 @@ Real VectorBase<Real>::Sum() const {
 }
 
 template<typename Real>
+Real VectorBase<Real>::SumPower(Real p) const {
+  Real sum = 0.0;
+  if (p == 0.0) {
+    for (MatrixIndexT i = 0; i < dim_; i++)
+      if (data_[i] != 0.0) sum += 1.0;
+  } else if (p == 1.0) {
+    for (MatrixIndexT i = 0; i < dim_; i++)
+      sum += std::abs(data_[i]);
+  } else if (p == 2.0) {
+    for (MatrixIndexT i = 0; i < dim_; i++)
+      sum += data_[i] * data_[i];
+  } else {
+    Real tmp;
+    for (MatrixIndexT i = 0; i < dim_; i++) {
+      tmp = pow(std::abs(data_[i]), p);
+      if (tmp == HUGE_VAL) {  // HUGE_VAL is what pow returns on error.
+        KALDI_ERR << "Could not raise element " << i << "to power " << p
+                  << ": returned value = " << tmp;
+      }
+      sum += tmp;
+    }
+  }
+  return sum;
+}
+
+template<typename Real>
 Real VectorBase<Real>::SumLog() const {
   double sum_log = 0.0;
   double prod = 1.0;
--- a/src/matrix/matrix-lib-test.cc
+++ b/src/matrix/matrix-lib-test.cc
@@ -625,6 +625,7 @@ static void UnitTestSimpleForVec() {  // testing some simple operaters on vector
     V2.CopyFromVec(V1);
     AssertEqual(V1.Norm(p), V2.Norm(p));
     AssertEqual(sqrt(V1.Sum()), V.Norm(2.0));
+    AssertEqual(V1.Sum(), V.SumPower(2.0));
   }
 
   for (MatrixIndexT i = 0; i < 5; i++) {


matthewaylett:
This was added  by Arnab to support the aperiodic extraction code.

jtrmal:
When you mentiong the linking error, what do you mean by that? Is it
somewhere in the travis log?
Or on your own machine?
I don't think we should merge this until we figure out what's going on.
y.

matthewaylett:
Hi

Sorry I'm being clumsy about this... I get this error with or without by
change (I updated o golden). So I rather dumbly assumed others had the same
problem. I don't see how this can be connected to adding the function in
kaldi-vector.cc (but that doesn't mean it isn't) and I agree don't merge if
you aren't getting this same problem.

I had a look but its not that obvious waht the problem is.

Best

M.



make[1]: Entering directory `/afs/
inf.ed.ac.uk/group/project/kaldi/kaldi_idlak/src/bin'
g++ -rdynamic -Wl,-rpath=/afs/
inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/lib
align-mapped.o ../lm/kaldi-lm.a ../decoder/kaldi-decoder.a
../lat/kaldi-lat.a ../hmm/kaldi-hmm.a ../transform/kaldi-transform.a
../gmm/kaldi-gmm.a ../tree/kaldi-tree.a ../matrix/kaldi-matrix.a
../util/kaldi-util.a ../base/kaldi-base.a ../thread/kaldi-thread.a
../fstext/kaldi-fstext.a   -L/afs/
inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/lib -lfst
/usr/lib64/atlas/liblapack.so /usr/lib64/atlas/libcblas.so
/usr/lib64/atlas/libatlas.so /usr/lib64/atlas/libf77blas.so -lm -lpthread
-ldl -o align-mapped
../lat/kaldi-lat.a(lattice-functions.o): In function
`kaldi::ComputeCompactLatticeAlphas(fst::VectorFst<fst::ArcTpl<fst::CompactLatticeWeightTpl<fst::LatticeWeightTpl<float>,
int> > > const&, std::vector<double, std::allocator<double> >*)':
/afs/
inf.ed.ac.uk/group/project/kaldi/kaldi_idlak/src/lat/lattice-functions.cc:109:
multiple definition of
`kaldi::ComputeCompactLatticeAlphas(fst::VectorFst<fst::ArcTpl<fst::CompactLatticeWeightTpl<fst::LatticeWeightTpl<float>,
int> > > const&, std::vector<double, std::allocator<double> >*)'
../lat/kaldi-lat.a(kws-functions.o):/afs/
inf.ed.ac.uk/group/project/kaldi/kaldi/src/lat/kws-functions.cc:110: first
defined here
../lat/kaldi-lat.a(lattice-functions.o): In function
`kaldi::ComputeCompactLatticeBetas(fst::VectorFst<fst::ArcTpl<fst::CompactLatticeWeightTpl<fst::LatticeWeightTpl<float>,
int> > > const&, std::vector<double, std::allocator<double> >*)':
/afs/
inf.ed.ac.uk/group/project/kaldi/kaldi_idlak/src/lat/lattice-functions.cc:147:
multiple definition of
`kaldi::ComputeCompactLatticeBetas(fst::VectorFst<fst::ArcTpl<fst::CompactLatticeWeightTpl<fst::LatticeWeightTpl<float>,
int> > > const&, std::vector<double, std::allocator<double> >*)'
../lat/kaldi-lat.a(kws-functions.o):/afs/
inf.ed.ac.uk/group/project/kaldi/kaldi/src/lat/kws-functions.cc:148: first
defined here
collect2: ld returned 1 exit status
make[1]: *** [align-mapped] Error 1
make[1]: Leaving directory `/afs/
inf.ed.ac.uk/group/project/kaldi/kaldi_idlak/src/bin'
make: *** [bin] Error 2

jtrmal:
OK, I see the problem now. There was a change around the time the
sourceforge broke down when we moved some of the functions (notably
ComputeCompactLatticeAlphas
and ComputeCompactLatticeBetas) from kws-functions.cc into the
lattice-functions.cc
Depending on how careful you were when merging in upstream/golden, you
might have ended up with the functions defined in both files. The quickfix
would be to remove the functions from kws-functions.cc and kws-functions.h
The proper fix would be to make the merge again, correctly,  but that might
be quite difficult thing depending on how much local modifications you have
and how different your history is now.
y.

matthewaylett:
Hi

Thanks for the info. This sounds like a good explanation.

I actually had old .o files lieing around. Although I would have thought
dependencies should ahve recopiled them. Anyway. I did make clean and the
problem went away.

However I'm getting something else wierd now.

make clean
make

is fine. But if I do make again:



g++ -msse -msse2 -Wall -I.. -pthread -DKALDI_DOUBLEPRECISION=0
-DHAVE_POSIX_MEMALIGN -Wno-sign-compare -Wno-unused-local-typedefs
-Winit-self -DHAVE_EXECINFO_H=1 -rdynamic -DHAVE_CXXABI_H -DHAVE_ATLAS
-I/afs/inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/ATLAS/include
-I/afs/inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/include
-Wno-sign-compare -g    -c -o nnet3-compute.o nnet3-compute.cc
nnet3-compute.cc: In function ‘int main(int, char**)’:
nnet3-compute.cc:102: error: reference to ‘int64’ is ambiguous
/afs/
inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/include/fst/types.h:31:
error: candidates are: typedef int64_t int64
../base/kaldi-types.h:48: error:                 typedef int64_t
kaldi::int64
nnet3-compute.cc:102: error: reference to ‘int64’ is ambiguous
/afs/
inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/include/fst/types.h:31:
error: candidates are: typedef int64_t int64
../base/kaldi-types.h:48: error:                 typedef int64_t
kaldi::int64
nnet3-compute.cc:102: error: expected ‘;’ before ‘frame_count’
nnet3-compute.cc:153: error: ‘frame_count’ was not declared in this scope
nnet3-compute.cc:160: error: ‘frame_count’ was not declared in this scope
../base/kaldi-math.h: At global scope:
../base/kaldi-math.h:126: warning: ‘kaldi::kLogZeroBaseFloat’ defined but
not used
cc1plus: warning: unrecognized command line option
"-Wno-unused-local-typedefs"

I'm guessing this is also due to my update.

So when you say:

on how careful you were when merging in upstream/golden

What does care mean in this context? Is just doing

git fetch golden

Not the way to update?

I'm away home now. I can probbaly fix this problem but I'd rather just redo
my repo. Its not got anything in it yet except the small fixes I pushed up
to my fork.

I can clone my fork again and check this compiles next week.

I don't think any of these problems is connected with the small change in
kaldi-vectors and I'd like to get that into the core so I don't have to
change it again. Following this most of my changes don't really impact key
code in kaldi. Perhaps someone can update and check?

at least I am begining to understand git so some progress.

v best

Matthew

danpovey:
I am resolving that int64 issue  by pushing something to master- it needed
a typedef.  (It depends on your compiler, that's why we didn't see it
before.)
Dan

src/matrix/kaldi-vector.h
```diff
@@ -310,6 +310,10 @@ class VectorBase {
   /// Returns sum of the elements
   Real Sum() const;
 
+  /// Returns sum of the elements raised to power p. For p>0, this is the same
+  /// as Norm(p) raised to the p-th power.
+  Real SumPower(Real p) const;
```
danpovey: The documentation says that it's the same as Norm(p)^p (except for p == 0, where it seems to be the same).
That being said, I don't see why it's needed.  Can't you just replace its invocation wth Norm(p)^p?

matthewaylett:
Hi

I will test this alternative. ASAP.

M.

matthewaylett:
Hi

Its the same except for floating point error (I believe), and thus not quite the same.
For the aperiodic energy stuff (where its used to calculate an error function myabe not an issue but for calcualting standard ceps (added to signal-process.cc by Arnab) and used for:

BaseFloat total_energy = signal_window.SumPower(2);

I think I'd be happier with Arnabs implementation. Tricky because I didn't make these changes and still need to run tests on this dsp code.

Main reason to get these changes out of the way is that they are so core they keep colliding with other changes being made in nnet etc.

Your call Dan, if you like I can replace with powf(v.Norm(p), p) but I worry underlying dsp stuff might then be slightly "wrong".

Best

M.

danpovey:
I don't want to add this function because it's redundant and I'm concerned
about bloat.


> Its the same except for floating point error (I believe), and thus not
> quite the same.
> For the aperiodic energy stuff (where its used to calculate an error
> function myabe not an issue but for calcualting standard ceps (added to
> signal-process.cc by Arnab) and used for:
>
> BaseFloat total_energy = signal_window.SumPower(2);
>

In this case you could simply call VecVec(signal_window, signal_window) and
get the same answer.
In other cases (if different powers are used) you could use that expression
you mentioned.

Dan

matthewaylett:
Hi

All sounds very sensible so had a go at this but got a slightly wierd
result.

It seems:;

cblas_Xdot(adim, a.Data(), 1, b.Data(), 1);

being called in VecVec

Also does not produce the same results as

sum += vec[i] * vec[i] used in the SumPower function.

e.g.
76224520.000000 76224496.000000
20225596.000000 20225580.000000
etc.

In fact powf(v.Norm(), 2.0) is closer (but also different).

Any idea why this might be?

Really sorry about this, I wish I'd got Arnab to deal with all of this
before he was consumed by Apple... Its not even my code :-(

v best

Matthew

danpovey:
Check that your BLAS installation is working-- does 'make test' in matrix/
work?

Dan

matthewaylett:
Hi

Yep seems to work:

Running matrix-lib-test ...... SUCCESS
Running kaldi-gpsr-test ...... SUCCESS
Running sparse-matrix-test ...... SUCCESS

Heres the linker line in case it looks odd:

g++ -rdynamic -Wl,-rpath=/afs/
inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/lib
 sparse-matrix-test.o kaldi-matrix.a ../base/kaldi-base.a   -L/afs/
inf.ed.ac.uk/user/m/matthewa/kaldi/kaldi_idlak/tools/openfst/lib -lfst
/usr/lib64/atlas/liblapack.so /usr/lib64/atlas/libcblas.so
/usr/lib64/atlas/libatlas.so /usr/lib64/atlas/libf77blas.so -lm -lpthread
-ldl -o sparse-matrix-test

Anything else I should check?

Best

M.

danpovey:
I don't know what to suggest- you'll have to do a little debugging.   This
isn't rocket science-- the sum-of-squares of a vector should be easy to
compute manually and  you can figure out where the bug is.
Dan

matthewaylett:
Hi

Okay. Now I'm really puzzled and I think I must be doing something so dumb
I just can't see it.

I altered the UnitTestDotprod in matrix-lib-test.cc to check a vector dot
product between a vector and itself.

This passed.

But if you call the test earlier it fails. Or at least it does on my
machine.

My only thought is that maybe a random seed is causing a problem? But I
just don't see why this should happen or what the implications are.
Valgrind shows nothing odd.

I've attached my edited matrix-lib-test.cc - it has a modified
UnitTestDotprod function that is called twice - first at the top of the
unit test and then where it was always called.

I wondered if someone might run this to establish if either:
1. I'm crazy
2. I've done something so stupid I can't see it.
3. There is a really odd pernicious bug.

Best

Matthew

danpovey:
Your attachment was stripped out by github.  Maybe paste the relevant parts of it into the conversation?  That would be more convenient for me.
You could also run it in valgrind- this will inform you of memory errors, which are usually the cause of unrepeatable things.
Dan

matthewaylett:
Hi

As I said previously, nothing from valgrind, I did that as soon as I got
this problem.

Here is the modified unit test:

template<typename Real> static void UnitTestDotprod() {
  for (MatrixIndexT iter = 0;iter < 5;iter++) {
    MatrixIndexT dimM = 200 + Rand()%100;
    Vector<Real> v(dimM), w(dimM);

    InitRand(&v);
    InitRand(&w);
    Vector<double> wd(w);

    Real f = VecVec(w, v), f2 = VecVec(wd, v), f3 = VecVec(v, wd);
    Real f4 = VecVec(w, w);
    Real sum = 0.0, sum2 = 0.0;
    for (MatrixIndexT i = 0;i < dimM;i++) sum += v(i)*w(i);
    for (MatrixIndexT i = 0;i < dimM;i++) sum2 += w(i)*w(i);
    KALDI_ASSERT(std::abs(f-sum) < 0.0001);
    KALDI_ASSERT(std::abs(f2-sum) < 0.0001);
    KALDI_ASSERT(std::abs(f3-sum) < 0.0001);
    KALDI_ASSERT(std::abs(f4-sum2) < 0.0001);
  }
}


Then just added:

  UnitTestDotprod<Real>();

at the top of MatrixUnitTest (so its running twice at different points).

So as you say, not exactly rocket science. I'm bemused. I'm very much
expecting no one else will get the problem but just in case :-).

One explanation is that the function is running Rand() twice so doing
different calculations. Maybe the for loop might be losing some accuracy
and getting it first time is just unlucky (Rand generating a v small number
perhaps). Much more likley to underflow in a squared operation. Is cblas_Xdot
doing the actual calculation in a higher precision? But if so the unit test
can fail randomly in general and we'd need to change the sums to doubles I
think.

To be honest, I can just use a for loop to do this instead for now (or
accept a loss/gain in precision). So no need to accept the SumPower change
anyway. So I'll wind that back anyway.

Best

M.








M.

danpovey:
You are computing the dot-prod of 2 separate vectors:
InitRand(&v);
InitRand(&w);
... VecVec(v, w);

You should be doing
VecVec(v, v).
Dan

matthewaylett:
Real f4 = VecVec(w, w);
...
for (MatrixIndexT i = 0;i < dimM;i++) sum2 += w(i)*w(i);
...
KALDI_ASSERT(std::abs(f4-sum2) < 0.0001);


Are my additions.

M.

danpovey:
That threshold is quite small-- I can imagine that in single precision it might sometimes fail (but not by much-- you could easily find out in gdb).  Multiply the threshold by 10 and it would probably succeed.

matthewaylett:
Hi

Yep I think its caused by precision

If I change sum2 to a double and cast the vals before muliplying it passes.

cblas_Xdot  must be doing the calculation in a higher precision despite
being passed floats. So strictly speaking the unit test should take that
into account.

Anyway, mystery seems solved.

As I said, I'll wind back the SumPower change.

v best

M.

danpovey:
OK.  I'll check in changes to the thresholds in that test to the master, next time I check in.
Meanwhile I'll close this issue.

PR #296, opened 2015-10-26
fix nnet2 convnet input spec

naxingyu:
Input of convolutional1dcomponent can be rearranged if rearrange_input_ is set true. It is required when appending multiple convolutional1dcomponent, as pointed out by @vijayaditya in [this discussion](https://groups.google.com/d/msgid/kaldi-developers/CANdDGsHqJBKkn%2BjFqwJDLTmBVDoAK3c3XH67jQ3HpR5Z1MmJSg%40mail.gmail.com?utm_medium=email&utm_source=footer).
Documents are updated.

### DIFF
--- a/egs/wsj/s5/steps/nnet2/train_convnet_accel2.sh
+++ b/egs/wsj/s5/steps/nnet2/train_convnet_accel2.sh
@@ -4,6 +4,7 @@
 #                2013  Xiaohui Zhang
 #                2013  Guoguo Chen
 #                2014  Vimal Manohar
+#                2015  Xingyu Na
 # Apache 2.0.
 
 # train_convnet_accel2.sh is modified from train_pnorm_accel2.sh. It propotypes
@@ -84,6 +85,7 @@ patch_dim1=7          # dim of convolutional kernel in the first layer
 pool_size=3           # size of pooling after the first convolutional layer
 num_filters2=256      # number of filters in the second convolutional layer
 patch_dim2=4          # dim of convolutional kernel in the second layer
+patch_step2=1         # patch step of the second convolutional layer
 
 mix_up=0 # Number of components to mix up to (should be > #tree leaves, if
         # specified.)
@@ -262,10 +264,8 @@ if [ $stage -le -2 ]; then
   tot_input_dim=$[$feat_dim*$tot_splice]
   num_patch1=$[1+($feat_dim-$patch_dim1)/$patch_step1]
   num_pool=$[$num_patch1/$pool_size]
-  patch_dim2=$[$patch_dim2*$num_filters1]
-  patch_step2=$num_filters1
-  patch_stride2=$[$num_pool*$num_filters1]   # same as pool outputs
-  num_patch2=$[1+($num_pool*$num_filters1-$patch_dim2)/$patch_step2]
+  patch_stride2=$num_pool
+  num_patch2=$[1+($patch_stride2-$patch_dim2)/$patch_step2]
   conv_out_dim1=$[$num_filters1*$num_patch1] # 128 x (36 - 7 + 1)
   pool_out_dim=$[$num_filters1*$num_pool]
   conv_out_dim2=$[$num_filters2*$num_patch2]
@@ -284,7 +284,7 @@ SoftmaxComponent dim=$num_leaves
 EOF
   
   cat >$dir/replace.1.config <<EOF
-Convolutional1dComponent input-dim=$pool_out_dim output-dim=$conv_out_dim2 learning-rate=$initial_lrate param-stddev=$stddev bias-stddev=$bias_stddev patch-dim=$patch_dim2 patch-step=$patch_step2 patch-stride=$patch_stride2
+Convolutional1dComponent input-dim=$pool_out_dim output-dim=$conv_out_dim2 learning-rate=$initial_lrate param-stddev=$stddev bias-stddev=$bias_stddev patch-dim=$patch_dim2 patch-step=$patch_step2 patch-stride=$patch_stride2 rearrange-input=true
 NormalizeComponent dim=$conv_out_dim2
 AffineComponentPreconditionedOnline input-dim=$conv_out_dim2 output-dim=$num_leaves $online_preconditioning_opts learning-rate=$initial_lrate param-stddev=0 bias-stddev=0
 SoftmaxComponent dim=$num_leaves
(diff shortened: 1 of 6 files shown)


danpovey:
@vijayaditya, could you please check this?

vijayaditya:
Checking..

egs/wsj/s5/steps/nnet2/train_convnet_accel2.sh
```diff
@@ -284,7 +284,7 @@ SoftmaxComponent dim=$num_leaves
 EOF
   
   cat >$dir/replace.1.config <<EOF
-Convolutional1dComponent input-dim=$pool_out_dim output-dim=$conv_out_dim2 learning-rate=$initial_lrate param-stddev=$stddev bias-stddev=$bias_stddev patch-dim=$patch_dim2 patch-step=$patch_step2 patch-stride=$patch_stride2
+Convolutional1dComponent input-dim=$pool_out_dim output-dim=$conv_out_dim2 learning-rate=$initial_lrate param-stddev=$stddev bias-stddev=$bias_stddev patch-dim=$patch_dim2 patch-step=$patch_step2 patch-stride=$patch_stride2 rearrange-input=true
```
vijayaditya: You might want to rethink what is the default vectorization of the input. Every convolution component other than the first convolution component in the network has cft vectorized input. You might want to make this the default vectorization and rearrange-input when fct input is used.

src/nnet2/nnet-component-test.cc
```diff
@@ -376,19 +376,19 @@ void UnitTestConvolutional1dComponent() {
     if (Rand() % 2 == 0) {
       component.Init(learning_rate, input_dim, output_dim,
                      patch_dim, patch_step, patch_stride,
-                     param_stddev, bias_stddev);
+                     param_stddev, bias_stddev, false);
```
vijayaditya: test for both rearrange_input true and false.

src/nnet2/nnet-component.cc
```diff
@@ -3890,9 +3894,14 @@ void Convolutional1dComponent::Propagate(const ChunkInfo &in_info,
 
   // build-up a column selection map
   for (int32 p = 0, index = 0; p < num_patches; p++) {
+    int32 fstride = p * patch_step_;
```
vijayaditya: change variable names p, s and d to something more understandable..

src/nnet2/nnet-component.cc
```diff
@@ -4110,6 +4124,8 @@ void Convolutional1dComponent::Read(std::istream &is, bool binary) {
   ReadBasicType(is, binary, &patch_step_);
   ExpectOneOrTwoTokens(is, binary, ostr_beg.str(), "<PatchStride>");
   ReadBasicType(is, binary, &patch_stride_);
+  ExpectOneOrTwoTokens(is, binary, ostr_beg.str(), "<RearrangeInput>");
```
vijayaditya: I would recommend using variable names which do not require the user to read the code to understand their usage. I would recommend using a variable which specifies in detail what is the default vectorization order and when users would have to rearrange_input.

I tried to solve this issue by using input_vectorization variable in ConvolutionComponent being written for nnet3. (see #283, @[line](https://github.com/vijayaditya/kaldi/blob/05b00b8834d3c45baed4461dc923109da994c3cb/src/nnet3/nnet-simple-component.h#L922) ) The documentation for this component is not complete, but I basically tried to explain that zyx is the default vectorization order and users can specify that they are using yzx input. Based on this info the component decides if it has to rearrange the input or not.

danpovey: Vijay, it's not clear to me that this explanation can be done in the
variable very easily... even if you talk about zyx and yzx, it would take
some explanation to express what this really means (it's certainly not
clear to me).  So for me it would be find to leave the variable name as it
is, and put this explanation somewhere in the documentation.
Unless it's the case that we will most likely later want to support other
kinds of rearrangement and having a boolean variable is too limiting.
Dan

vijayaditya: I don't think we will have more than 2 types of vectorization, so the boolean would be sufficient. (Append descriptor will only for zyx or yzx vectorization)

danpovey:
Note that the component-test failed.

naxingyu:
Thanks for checking. I correct the bug and squashed the commits. Now it 
compiles.
1) About the nomenclature. It is quite hard to make the options clear 
using current "patch" thing. I'm thinking of a total renaming of the 
variables, using "filter-featmap" nomenclature (hope it doesn't confuse 
with the column_maps).
2) About the axes spanning. I think fct is more natural than cft, and it 
is of the same spanning with splicing. That's why I set it as default. 
And in my case, changing the option from "rearrange-input" to 
"appended-convolution" might make more sense for users who do not want 
to read code.
What do you think?

Xingyu

danpovey:
Please be careful with back-compatibility.  I don't want to break people's
existing experiments (e.g. make the code unable to read existing models)
without a good reason.
Dan

naxingyu:
Done. Please check.

src/nnet2/nnet-component.cc
```diff
@@ -3744,11 +3749,12 @@ void Convolutional1dComponent::Init(BaseFloat learning_rate,
 // initialize the component using predefined matrix file
 void Convolutional1dComponent::Init(BaseFloat learning_rate,
                                     int32 patch_dim, int32 patch_step, int32 patch_stride,
-                                    std::string matrix_filename) {
+                                    std::string matrix_filename, bool appended_conv) {
```
danpovey: watch 80-char limit

src/nnet2/nnet-component.h
```diff
@@ -1784,6 +1784,7 @@ class Convolutional1dComponent: public UpdatableComponent {
   const Convolutional1dComponent &operator = (const Convolutional1dComponent &other); // Disallow.
   CuMatrix<BaseFloat> filter_params_;
   CuVector<BaseFloat> bias_params_;
+  bool appended_conv_;
```
danpovey: This should be documented with a comment at this point.

naxingyu: Do you mean the used of appended_conv_? It is documented at Propogate().

Xingyu

danpovey: Yes, but if someone looks for that documentation they would never find it--
at least you need to state by the declaration of the variable, where
someone would find the documentation.

naxingyu: I see.. Will update this.

naxingyu:
Done style and document fix.

danpovey:
Thanks.  Once the checks complete, I'll merge.

PR #297, opened 2015-10-27
WIP: CNN2D component based on CNN1D component

vijayaditya:
After speed comparisons between AddMatMat and AddMatMatBatched calls  (see PR #283 ) it turns out that AddMatMatBatched is faster. So I have just created a new ConvolutionComponent which uses AddMatMatBatched call. This is similar to Convolutional1dComponent, in the basic logic of the Propagate, Backprop and Update methods. 
If this PR is found acceptable I will delete the Convolutional1dComponent. 

TODO: Improve the documentation --will be done in few hours.
             Remove Convolutional1dComponent

@naxingyu , @freewym  let me know your comments.

### DIFF
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -20,6 +20,7 @@
 
 #include <iterator>
 #include <sstream>
+#include <algorithm>
 #include "nnet3/nnet-simple-component.h"
 #include "nnet3/nnet-parse.h"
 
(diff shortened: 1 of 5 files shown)


danpovey:
So I guess the advantage of the ConvolutionComponent is that it supports 2d
convolution?
Is it similar in speed to the Convoluational1dComponent?

I would probably want to keep the Convolutional1dComponent around for a
while for back compatibility, and just say that it's deprecated.

Dan

vijayaditya:
Also a major gripe with the previous component was the way in which hyper-parameters were defined.
The hyper-parameters are completely redefined,  these are

<pre><code>
 // there is no filt_z_dim_ as it is the same input_z_dim_
  int32 input_x_dim_;
  int32 input_y_dim_;
  int32 input_z_dim_;
  int32 filt_x_dim_;
  int32 filt_y_dim_;
  int32 filt_x_step_;
  int32 filt_y_step_;
  TensorVectorizationType input_vectorization_;
</code></pre>

In addition there are minor speed optimizations (not tested).

src/nnet3/nnet-simple-component.cc
```diff
@@ -2676,6 +3324,84 @@ std::string MaxpoolingComponent::Info() const {
...
+                                const CuMatrixBase<BaseFloat> &, // out_value,
+                                const CuMatrixBase<BaseFloat> &out_deriv,
+                                Component *to_update,
+                                CuMatrixBase<BaseFloat> *in_deriv) const  {
+  // computing the reverse column_map
+  std::vector<int32> reverse_column_map(column_map_.Dim()),
+                     column_map(column_map_.Dim());
+  column_map_.CopyToVec(&column_map);
+
+  for (int32 i = 0; i < column_map.size(); i++)  {
```
naxingyu: Defining a map_size avoids calling size every time.

src/nnet3/nnet-simple-component.h
```diff
@@ -746,6 +746,55 @@ class ClipGradientComponent: public Component {
 
 };
 
+// PermuteComponent shuffles the rows in the input, according to the
+// specification.
```
naxingyu: Should it be "shuffles the cols"?

vijayaditya:
@danpovey  forgot to answer your question about speed comparison with Convolutional1dComponent. The new component is equivalent to Convolutional1dComponent for all matched cases.

src/nnet3/nnet-simple-component.h
vijayaditya: @naxingyu yep.. will modify

src/nnet3/nnet-simple-component.cc
danpovey: agreed-- the use of a local variable to store the size is better than calling size() each time.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1956,6 +1957,652 @@ void NaturalGradientPerElementScaleComponent::Update(
...
+  cfl->GetValue("input-vectorization-order", &input_vectorization_order);
+  if (input_vectorization_order.compare("zyx") == 0) {
+    input_vectorization = kZyx;
+  } else if (input_vectorization_order.compare("yzx") == 0) {
+    input_vectorization = kYzx;
+  } else {
+    KALDI_ERR << "Unknown or unsupported input vectorization order "
+              << input_vectorization_order
+              << " accepted candidates are 'yzx' and 'zyx'";
+  }
```
naxingyu: I think the term "xyz" is a bit confusing for user. How would one figure out whether the input of an intermediate component is 'zyx' or 'yzx', without reading code?

src/nnet3/nnet-simple-component.cc
```diff
@@ -1956,6 +1957,652 @@ void NaturalGradientPerElementScaleComponent::Update(
...
+              column_map.at(index) = ZyxVectorIndex(x_step * filt_x_step_ + x,
+                                                 y_step * filt_y_step_ + y, z,
+                                                 input_x_dim_, input_y_dim_,
+                                                 input_z_dim_);
+            } else if (input_vectorization_ == kYzx)  {
+              column_map.at(index) = YzxVectorIndex(x_step * filt_x_step_ + x,
+                                                 y_step * filt_y_step_ + y, z,
+                                                 input_x_dim_, input_y_dim_,
+                                                 input_z_dim_);
+            }
```
naxingyu: ```x_step * filt_x_step_``` and ```y_step * filt_y_step_``` can be computed outside of the tri-loops and reused.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1956,6 +1957,652 @@ void NaturalGradientPerElementScaleComponent::Update(
...
+              vector_index = ZyxVectorIndex(x_step * filt_x_step_ + x,
+                                            y_step * filt_y_step_ + y, z,
+                                            input_x_dim_, input_y_dim_,
+                                            input_z_dim_);
+            } else if (input_vectorization_ == kYzx)  {
+              vector_index = YzxVectorIndex(x_step * filt_x_step_ + x,
+                                            y_step * filt_y_step_ + y, z,
+                                            input_x_dim_, input_y_dim_,
+                                            input_z_dim_);
+            }
```
naxingyu: ```x_step * filt_x_step_``` and ```y_step * filt_y_step_``` can be reused.

danpovey: The compiler can discover repeated expressions like that; what concerns me
more here is the repeated use of class variables like filt_x_step_, since
the compiler can't always know that they are unchanged- especially if other
functions are called in between- and it may repeatedly re-load them. It
would be better to have a local copy like
int32 filt_x_step = filt_x_step_;
Dan

vijayaditya: OK.. modifying according to suggestions.

vijayaditya:
@danpovey added description in the header file for ConvolutionComponent. If thats OK, the PR is ready for merge.

danpovey:
OK- will merge after checks complete.

vijayaditya:
checks done

jtrmal:
Merging...

src/nnet3/nnet-simple-component.cc
vijayaditya: In the component description it is specified that the output of the convolution component is always zyx. I also specify that yzx mode has been added due to the binaries like add-deltas which output "yz" vectorized data. Let me know if more information is essential and also what that information is.

The output of any other intermediate component always has input_z_dim_ = 1, at least for the existing components in nnet3 framework. So specifying the vectorization order as zyx or yzx does not matter. 

When I say other components I mean components which are not elementwise non-linearities, as elementwise non-linearities preserve their input vectorization order.

src/nnet3/nnet-simple-component.h
```diff
@@ -880,6 +929,218 @@ class NaturalGradientPerElementScaleComponent: public PerElementScaleComponent {
...
+ * Input : A matrix where each row is a  vectorized 3D-tensor.
+ *        The 3D tensor has dimensions
+ *        x: (e.g. time)
+ *        y: (e.g. frequency)
+ *        z: (e.g. channels like features/delta/delta-delta)
+ *
+ *        The component supports input vectorizations of type zyx and yzx.
+ *        The default vectorization type is zyx.
+ *        e.g. for input vectorization of type zyx the input is vectorized by
+ *        spanning axes z, y and x of the tensor in that order.
```
danpovey: this is not really clear to me-- if you explain it in terms of which dimension has the largest stride, it might be clearer.
E.g. in zyx vectorization, does z have a stride of one, or does x?

vijayaditya: OK will do, in zyx z has stride 1.

--Vijay

PR #304, opened 2015-10-29
fix nnet3 lstm exp dir naming

naxingyu:
Previously, setting --label-delay 0 results in 'lstm_sp_ld0', after this fix, change to 'lstm_sp', so that '$use_delay' is actually doing its job.

### DIFF
--- a/egs/swbd/s5c/local/nnet3/run_lstm.sh
+++ b/egs/swbd/s5c/local/nnet3/run_lstm.sh
@@ -64,15 +64,13 @@ where "nvcc" is installed.
 EOF
 fi
 
-use_delay=false
-if [ $label_delay -gt 0 ]; then use_delay=true; fi
-
 suffix=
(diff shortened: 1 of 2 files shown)


egs/swbd/s5c/local/nnet3/run_lstm.sh
```diff
@@ -64,7 +64,7 @@ where "nvcc" is installed.
 EOF
 fi
 
-use_delay=false
+use_delay=
 if [ $label_delay -gt 0 ]; then use_delay=true; fi
```
danpovey: It seems like the use_delay variable is not really necessary at all now.
Couldn't you do something like
if [ $label_delay -gt 0 ]; then dir=${dir}_ld$label_delay; fi
... then there would be no need for the use_delay variable which IMO just confuses things.

naxingyu: right... done.

PR #305, opened 2015-10-29
priors_fix: Fixed issue when creating priors egs 

vimalmanohar:
Fixed issue when creating priors egs in get_egs_discrimiative2.sh

### DIFF
--- a/egs/wsj/s5/steps/nnet2/get_egs_discriminative2.sh
+++ b/egs/wsj/s5/steps/nnet2/get_egs_discriminative2.sh
@@ -86,23 +86,21 @@ nj_ali=$(cat $alidir/num_jobs) || exit 1;
 sdata=$data/split$nj
 utils/split_data.sh $data $nj
 
-
-
-
 if [ $nj_ali -eq $nj ]; then
   ali_rspecifier="ark,s,cs:gunzip -c $alidir/ali.JOB.gz |"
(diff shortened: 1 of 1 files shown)


danpovey:
This setup is rather wasteful because you copy the transforms and alignments to disk, but you then use them only once.  There is no point dumping something if you only need it once.  In both cases you can just give it an archive containing everything, and it will take only what it needs.

vimalmanohar:
I have modified it. If it is ok, I can make a clean patch.

danpovey:
Look OK.  Sure, make a clean patch.

vimalmanohar:
Made a clean patch.

danpovey:
Thanks.  @vijayaditya, could you please do a quick pass of checking before I merge this..

vijayaditya:
OK

egs/wsj/s5/steps/nnet2/get_egs_discriminative2.sh
```diff
@@ -256,39 +255,23 @@ if [ $stage -le 10 ]; then
 priors_egs_list=
 for y in `seq $num_archives_priors`; do
```
vijayaditya: Why is num_archives_priors still being used ? I thought the entire point was to just have one job for prior computation.

danpovey: It's OK-- creating the egs for the priors is fast but doing the computation
using them is not that fast.  So I didn't propose to have just one archive
of priors.
Dan

vijayaditya:
OK, in that case looks fine to merge.

PR #307, opened 2015-11-01
Add PerElementOffsetComponent to nnet3

hhadian:
I'm not sure if everything is logically correct. Particularly CollapseWithNext  and parameter updates.

### DIFF
--- a/src/nnet3/nnet-component-test.cc
+++ b/src/nnet3/nnet-component-test.cc
@@ -54,6 +54,42 @@ void TestNnetComponentAddScale(Component *c) {
   delete c3;
 }
 
+void TestNnetComponentVectorizeUnVectorize(Component *c) {
+  UpdatableComponent *uc = dynamic_cast<UpdatableComponent*>(c);
+  if((uc==NULL) || (uc->NumParameters() == 0))
+    return;
+  UpdatableComponent *uc2 = dynamic_cast<UpdatableComponent*>(uc->Copy());
+  uc2->SetZero(false);
+  Vector<BaseFloat> params(uc2->NumParameters());
+  uc2->Vectorize(&params);
+  KALDI_ASSERT(params.Min()==0.0 && params.Sum()==0.0);
+  uc->Vectorize(&params);
+  uc2->UnVectorize(params);
+  KALDI_ASSERT(uc2->Info() == uc->Info());
+  KALDI_ASSERT(uc2->DotProduct(*uc2) == uc->DotProduct(*uc));
+  Vector<BaseFloat> params2(uc2->NumParameters());
+  uc2->Vectorize(&params2);
+  for(int i = 0; i < params.Dim(); i++)
+    KALDI_ASSERT(params(i) == params2(i));
+  delete uc2;
+}
+
+void TestNnetComponentUpdatableFlag(Component *c) {
+  UpdatableComponent *uc = dynamic_cast<UpdatableComponent*>(c);
+  if(uc==NULL)
+    return;
+  if(!(uc->Properties() & kUpdatableComponent)){
+    KALDI_ASSERT(uc->NumParameters() == 0);
+    KALDI_ASSERT(uc->DotProduct(*uc) == 0);
+    UpdatableComponent *uc2 = dynamic_cast<UpdatableComponent*>(uc->Copy());
+    uc2->Scale(7.0);
+    uc2->Add(3.0, *uc);
+    KALDI_ASSERT(uc2->Info() == uc->Info());
+    uc->SetZero(false);
+    KALDI_ASSERT(uc2->Info() == uc->Info());
+    delete uc2;
+  }
+}
 
 // tests the properties kPropagateAdds, kBackpropAdds,
 // kBackpropNeedsInput, kBackpropNeedsOutput.
@@ -271,6 +307,8 @@ void UnitTestNnetComponent() {
     TestNnetComponentIo(c);
     TestNnetComponentCopy(c);
     TestNnetComponentAddScale(c);
+    TestNnetComponentVectorizeUnVectorize(c);
+    TestNnetComponentUpdatableFlag(c);
     TestSimpleComponentPropagateProperties(*c);
     if (!TestSimpleComponentDataDerivative(*c, 1.0e-04) &&
         !TestSimpleComponentDataDerivative(*c, 1.0e-03) &&
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -795,18 +795,6 @@ Component *AffineComponent::CollapseWithNext(
   return ans;
 }
 
-Component *AffineComponent::CollapseWithNext(
-    const PerElementScaleComponent &next_component) const {
-  KALDI_ASSERT(this->OutputDim() == next_component.InputDim());
-  AffineComponent *ans =
-      dynamic_cast<AffineComponent*>(this->Copy());
-  KALDI_ASSERT(ans != NULL);
-  ans->linear_params_.MulRowsVec(next_component.scales_);
-  ans->bias_params_.MulElements(next_component.scales_);
-
-  return ans;
-}
-
 Component *AffineComponent::CollapseWithPrevious(
     const FixedAffineComponent &prev_component) const {
   // If at least one was non-updatable, make the whole non-updatable.
@@ -827,11 +815,6 @@ Component *AffineComponent::CollapseWithPrevious(
 
 void PerElementScaleComponent::Scale(BaseFloat scale) {
   scales_.Scale(scale);
-}
-
-void PerElementScaleComponent::Resize(int32 dim) {
-  KALDI_ASSERT(dim > 0);
-  scales_.Resize(dim);
 }
 
 void PerElementScaleComponent::Add(BaseFloat alpha,
@@ -856,17 +839,21 @@ void PerElementScaleComponent::SetZero(bool treat_as_gradient) {
 }
 
 void PerElementScaleComponent::PerturbParams(BaseFloat stddev) {
-  CuVector<BaseFloat> temp_scales(scales_);
+  CuVector<BaseFloat> temp_scales(scales_.Dim(), kUndefined);
   temp_scales.SetRandn();
   scales_.AddVec(stddev, temp_scales);
 }
 
 std::string PerElementScaleComponent::Info() const {
   std::stringstream stream;
+  BaseFloat scales_mean = scales_.Sum() / scales_.Dim();
   BaseFloat scales_stddev = std::sqrt(VecVec(scales_, scales_) /
                               scales_.Dim());
   stream << Type() << ", input-dim=" << InputDim()
          << ", output-dim=" << OutputDim()
+         << ", scales-min=" << scales_.Min()
+         << ", scales-max=" << scales_.Max()
+         << ", scales-mean=" << scales_mean
          << ", scales-stddev=" << scales_stddev
          << ", learning-rate=" << LearningRate();
   return stream.str();
@@ -891,37 +878,35 @@ void PerElementScaleComponent::Init(
     BaseFloat learning_rate, int32 dim,
     BaseFloat param_mean, BaseFloat param_stddev) {
   UpdatableComponent::Init(learning_rate);
-  scales_.Resize(dim);
   KALDI_ASSERT(dim > 0 && param_stddev >= 0.0);
+  scales_.Resize(dim);
   scales_.SetRandn();
   scales_.Scale(param_stddev);
   scales_.Add(param_mean);
 }
 
 void PerElementScaleComponent::Init(BaseFloat learning_rate,
-                                    std::string matrix_filename) {
+                                    std::string vector_filename) {
   UpdatableComponent::Init(learning_rate);
-  CuMatrix<BaseFloat> mat;
-  ReadKaldiObject(matrix_filename, &mat); // will abort on failure.
-  KALDI_ASSERT(mat.NumCols() == 1);
-  int32 dim = mat.NumRows();
-  scales_.Resize(dim);
-  scales_.CopyColFromMat(mat, 0);
+  CuVector<BaseFloat> vec;
+  ReadKaldiObject(vector_filename, &vec); // will abort on failure.
+  scales_.Resize(vec.Dim());
+  scales_.CopyFromVec(vec);
 }
 
 void PerElementScaleComponent::InitFromConfig(ConfigLine *cfl) {
-  bool ok = true;
   BaseFloat learning_rate = learning_rate_;
-  std::string matrix_filename;
+  std::string vector_filename;
   int32 dim = -1;
   cfl->GetValue("learning-rate", &learning_rate); // optional.
-  if (cfl->GetValue("matrix", &matrix_filename)) {
-    Init(learning_rate, matrix_filename);
+  if (cfl->GetValue("vector", &vector_filename)) {
+    Init(learning_rate, vector_filename);
     if (cfl->GetValue("dim", &dim))
       KALDI_ASSERT(dim == InputDim() &&
-                   "input-dim mismatch vs. matrix.");
+                   "input-dim mismatch vs. vector.");
   } else {
-    ok = ok && cfl->GetValue("dim", &dim);
+    if(!cfl->GetValue("dim", &dim))
+      KALDI_ERR << "'dim' not provided in the config line.";
     BaseFloat param_mean = 1.0, param_stddev = 0.0;
     cfl->GetValue("param-mean", &param_mean);
     cfl->GetValue("param-stddev", &param_stddev);
@@ -930,8 +915,6 @@ void PerElementScaleComponent::InitFromConfig(ConfigLine *cfl) {
   if (cfl->HasUnusedValues())
     KALDI_ERR << "Could not process these elements in initializer: "
               << cfl->UnusedValues();
-  if (!ok)
-    KALDI_ERR << "Bad initializer " << cfl->WholeLine();
 }
 
 void PerElementScaleComponent::Propagate(
@@ -1010,6 +993,176 @@ void PerElementScaleComponent::UnVectorize(
     const VectorBase<BaseFloat> &params) {
   scales_.CopyFromVec(params);
 }
+
+void PerElementOffsetComponent::Scale(BaseFloat scale) {
+  offsets_.Scale(scale);
+}
+
+
+void PerElementOffsetComponent::Add(BaseFloat alpha,
+                                   const Component &other_in) {
+  const PerElementOffsetComponent *other =
+      dynamic_cast<const PerElementOffsetComponent*>(&other_in);
+  KALDI_ASSERT(other != NULL);
+  offsets_.AddVec(alpha, other->offsets_);
+}
+
+PerElementOffsetComponent::PerElementOffsetComponent(
+    const PerElementOffsetComponent &component):
+    UpdatableComponent(component),
+    offsets_(component.offsets_) { }
+
+void PerElementOffsetComponent::SetZero(bool treat_as_gradient) {
+  if (treat_as_gradient) {
+    SetLearningRate(1.0);
+    is_gradient_ = true;
+  }
+  offsets_.SetZero();
+}
+
+void PerElementOffsetComponent::PerturbParams(BaseFloat stddev) {
+  CuVector<BaseFloat> temp_offsets(offsets_.Dim(), kUndefined);
+  temp_offsets.SetRandn();
+  offsets_.AddVec(stddev, temp_offsets);
+}
+
+std::string PerElementOffsetComponent::Info() const {
+  std::stringstream stream;
+  BaseFloat offsets_mean = offsets_.Sum() / offsets_.Dim();
+  BaseFloat offsets_stddev = std::sqrt(VecVec(offsets_, offsets_) /
+                              offsets_.Dim());
+  stream << Type() << ", input-dim=" << InputDim()
+         << ", output-dim=" << OutputDim()
+         << ", offsets-min=" << offsets_.Min()
+         << ", offsets-max=" << offsets_.Max()
+         << ", offsets-mean=" << offsets_mean
+         << ", offsets-stddev=" << offsets_stddev
+         << ", learning-rate=" << LearningRate();
+  return stream.str();
+}
+
+Component* PerElementOffsetComponent::Copy() const {
+  PerElementOffsetComponent *ans = new PerElementOffsetComponent();
+  ans->learning_rate_ = learning_rate_;
+  ans->offsets_ = offsets_;
+  ans->is_gradient_ = is_gradient_;
+  return ans;
+}
+
+BaseFloat PerElementOffsetComponent::DotProduct(
+    const UpdatableComponent &other_in) const {
+  const PerElementOffsetComponent *other =
+      dynamic_cast<const PerElementOffsetComponent*>(&other_in);
+  return VecVec(offsets_, other->offsets_);
+}
+
+void PerElementOffsetComponent::Init(
+    BaseFloat learning_rate, int32 dim,
+    BaseFloat param_mean, BaseFloat param_stddev) {
+  UpdatableComponent::Init(learning_rate);
+  KALDI_ASSERT(dim > 0 && param_stddev >= 0.0);
+  offsets_.Resize(dim);
+  offsets_.SetRandn();
+  offsets_.Scale(param_stddev);
+  offsets_.Add(param_mean);
+}
+
+void PerElementOffsetComponent::Init(BaseFloat learning_rate,
+                                    std::string vector_filename) {
+  UpdatableComponent::Init(learning_rate);
+  CuVector<BaseFloat> vec;
+  ReadKaldiObject(vector_filename, &vec); // will abort on failure.
+  offsets_.Resize(vec.Dim());
+  offsets_.CopyFromVec(vec);
+}
+
+void PerElementOffsetComponent::InitFromConfig(ConfigLine *cfl) {
+  BaseFloat learning_rate = learning_rate_;
+  std::string vector_filename;
+  int32 dim = -1;
+  cfl->GetValue("learning-rate", &learning_rate); // optional.
+  if (cfl->GetValue("vector", &vector_filename)) {
+    Init(learning_rate, vector_filename);
+    if (cfl->GetValue("dim", &dim))
+      KALDI_ASSERT(dim == InputDim() &&
+                   "input-dim mismatch vs. vector.");
+  } else {
+    if(!cfl->GetValue("dim", &dim))
+      KALDI_ERR << "'dim' not provided in the config line.";
+    BaseFloat param_mean = 0.0, param_stddev = 0.0;
+    cfl->GetValue("param-mean", &param_mean);
+    cfl->GetValue("param-stddev", &param_stddev);
+    Init(learning_rate, dim, param_mean, param_stddev);
+  }
+  if (cfl->HasUnusedValues())
+    KALDI_ERR << "Could not process these elements in initializer: "
+              << cfl->UnusedValues();
+}
+
+void PerElementOffsetComponent::Propagate(
+    const ComponentPrecomputedIndexes *indexes,
+    const CuMatrixBase<BaseFloat> &in,
+    CuMatrixBase<BaseFloat> *out) const {
+  out->CopyFromMat(in);
+  out->AddVecToRows(1.0, offsets_);
+}
+
+void PerElementOffsetComponent::Backprop(
+    const std::string &debug_info,
+    const ComponentPrecomputedIndexes *indexes,
+    const CuMatrixBase<BaseFloat> &, // in_value
+    const CuMatrixBase<BaseFloat> &, // out_value
+    const CuMatrixBase<BaseFloat> &out_deriv,
+    Component *to_update_in,
+    CuMatrixBase<BaseFloat> *in_deriv) const {
+  PerElementOffsetComponent *to_update =
+      dynamic_cast<PerElementOffsetComponent*>(to_update_in);
+
+  if (in_deriv) {
+    // Propagate the derivative back to the input.
+    in_deriv->CopyFromMat(out_deriv);
+  }
+
+  if (to_update != NULL)
+    to_update->offsets_.AddRowSumMat(to_update->learning_rate_, out_deriv);
+}
+
+void PerElementOffsetComponent::Read(std::istream &is, bool binary) {
+  // might not see the begin marker part because of how ReadNew() works.
+  ExpectOneOrTwoTokens(is, binary, "<PerElementOffsetComponent>", "<LearningRate>");
+  ReadBasicType(is, binary, &learning_rate_);
+  ExpectToken(is, binary, "<Offsets>");
+  offsets_.Read(is, binary);
+  ExpectToken(is, binary, "<IsGradient>");
+  ReadBasicType(is, binary, &is_gradient_);
+  ExpectToken(is, binary, "</PerElementOffsetComponent>");
+}
+
+void PerElementOffsetComponent::Write(std::ostream &os, bool binary) const {
+  WriteToken(os, binary, "<PerElementOffsetComponent>");
+  WriteToken(os, binary, "<LearningRate>");
+  WriteBasicType(os, binary, learning_rate_);
+  WriteToken(os, binary, "<Offsets>");
+  offsets_.Write(os, binary);
+  WriteToken(os, binary, "<IsGradient>");
+  WriteBasicType(os, binary, is_gradient_);
+  WriteToken(os, binary, "</PerElementOffsetComponent>");
+}
+
+int32 PerElementOffsetComponent::NumParameters() const {
+  return InputDim();
+}
+
+void PerElementOffsetComponent::Vectorize(VectorBase<BaseFloat> *params) const {
+  params->CopyFromVec(offsets_);
+}
+
+void PerElementOffsetComponent::UnVectorize(
+    const VectorBase<BaseFloat> &params) {
+  offsets_.CopyFromVec(params);
+}
+
+
 
 NaturalGradientAffineComponent::NaturalGradientAffineComponent(): max_change_per_sample_(0.0),
   update_count_(0.0), active_scaling_count_(0.0), max_change_scale_stats_(0.0) { }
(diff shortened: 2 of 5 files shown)


danpovey:
Thanks.  Things like CollapseWithNext are not needed- at least for now.
That's not part of the interface.
@vpeddinti, could you please check this?

Dan

danpovey:
I mean @vijayaditya

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+      dynamic_cast<const PerElementOffsetComponent*>(&other_in);
+  return VecVec(offsets_, other->offsets_);
+}
+
+void PerElementOffsetComponent::Init(
+    BaseFloat learning_rate, int32 dim,
+    BaseFloat param_mean, BaseFloat param_stddev) {
+  UpdatableComponent::Init(learning_rate);
+  offsets_.Resize(dim);
+  KALDI_ASSERT(dim > 0 && param_stddev >= 0.0);
```
vijayaditya: place this assert before the resize method.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+  int32 dim = -1;
+  cfl->GetValue("learning-rate", &learning_rate); // optional.
+  if (cfl->GetValue("matrix", &matrix_filename)) {
+    Init(learning_rate, matrix_filename);
+    if (cfl->GetValue("dim", &dim))
+      KALDI_ASSERT(dim == InputDim() &&
+                   "input-dim mismatch vs. matrix.");
+  } else {
+    ok = ok && cfl->GetValue("dim", &dim);
+    BaseFloat param_mean = 1.0, param_stddev = 0.0;
```
vijayaditya: I think the default initializers should be zero mean and unit std-dev

danpovey: Actually, I'd make the defaults zero mean and default stddev...
the reason to make stddevs nonzero is usually to ensure nonzero gradients
in multilayer networks, but that's not an issue for this type of component.
Dan

danpovey: sorry I mean zero mean and zero stddev

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+  if (!ok)
+    KALDI_ERR << "Bad initializer " << cfl->WholeLine();
+}
+
+void PerElementOffsetComponent::Propagate(
+    const ComponentPrecomputedIndexes *indexes,
+    const CuMatrixBase<BaseFloat> &in,
+    CuMatrixBase<BaseFloat> *out) const {
+  
+  out->CopyRowsFromVec(offsets_);
```
vijayaditya: You have to add offsets_ to out, not overwrite the matrix. IIRC there is unit test which tests this. Did you code pass all the unit tests ?

vijayaditya: Also you can directly use the AddRowsVec() method

danpovey: To test it you need to modify GenerateRandomComponentConfig() in
nnet-test-utils.cc.  Make sure you modify the RandInt statement to include
the integer in the case statement you add.
Dan

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+  if (in_deriv) {
+    // Propagate the derivative back to the input.
+    in_deriv->CopyFromMat(out_deriv);
+  }
+
+  if (to_update != NULL) {
+    // Next update the model (must do this 2nd so the derivatives we propagate
+    // are accurate, in case this == to_update_in.)
+    if (to_update->is_gradient_)
+      to_update->UpdateSimple(in_value, out_deriv);
```
vijayaditya: your UpdateSimple doesn't need the in_value. You can just pass the out_deriv value.

danpovey: This would affect the flags returned in Properties() too.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+    else  // the call below is to a virtual function that may be re-implemented
+      to_update->Update(debug_info, in_value, out_deriv);  // by child classes.
```
vijayaditya: You don't need a separate Update method, as this is not a natural_gradient component.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+  }
+}
+
+void PerElementOffsetComponent::Read(std::istream &is, bool binary) {
+  // might not see the begin marker part because of how ReadNew() works.
+  ExpectOneOrTwoTokens(is, binary, "<PerElementOffsetComponent>", "<LearningRate>");
+  ReadBasicType(is, binary, &learning_rate_);
+  ExpectToken(is, binary, "<Params>");
```
vijayaditya: I think it might be preferable to just use the name Offsets in place of Params.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1022,196 @@ void PerElementScaleComponent::UnVectorize(
...
+  ExpectToken(is, binary, "<IsGradient>");
+  ReadBasicType(is, binary, &is_gradient_);
+  ExpectToken(is, binary, "</PerElementOffsetComponent>");
+}
+
+void PerElementOffsetComponent::Write(std::ostream &os, bool binary) const {
+  WriteToken(os, binary, "<PerElementOffsetComponent>");
+  WriteToken(os, binary, "<LearningRate>");
+  WriteBasicType(os, binary, learning_rate_);
+  WriteToken(os, binary, "<Params>");
```
vijayaditya: Change the token name here too.

src/nnet3/nnet-simple-component.h
```diff
@@ -873,6 +875,83 @@ class PerElementScaleComponent: public UpdatableComponent {
...
+    return kSimpleComponent|kUpdatableComponent|
+        kLinearInParameters|kPropagateInPlace;
+  }
+
+  virtual void Propagate(const ComponentPrecomputedIndexes *indexes,
+                         const CuMatrixBase<BaseFloat> &in,
+                         CuMatrixBase<BaseFloat> *out) const;
+  virtual void Backprop(const std::string &debug_info,
+                        const ComponentPrecomputedIndexes *indexes,
+                        const CuMatrixBase<BaseFloat> &in_value,
```
vijayaditya: change 
<code>const CuMatrixBase<BaseFloat> &in_value</code>
to
<code>const CuMatrixBase<BaseFloat> &</code> as it will be clear to readers that we won't use in_value

src/nnet3/nnet-simple-component.cc
vijayaditya: @danpovey  could you give some information on kPropagateAdds and kBackpropAdds ? Is it better to set them or unset them ?

src/nnet3/nnet-simple-component.h
```diff
@@ -873,6 +875,83 @@ class PerElementScaleComponent: public UpdatableComponent {
...
+ public:
+  virtual int32 InputDim() const { return offsets_.Dim(); }
+  virtual int32 OutputDim() const { return offsets_.Dim(); }
+
+  virtual std::string Info() const;
+  virtual void InitFromConfig(ConfigLine *cfl);
+
+  PerElementOffsetComponent() { } // use Init to really initialize.
+  virtual std::string Type() const { return "PerElementOffsetComponent"; }
+  virtual int32 Properties() const {
```
vijayaditya: You can also add kBackpropInPlace. This component is not kLinearInParameters.

danpovey: Regarding kPropagateAdds and kBackpropAdds: it's a question of what is most
convenient for the propagate and backprop code, i.e. what will avoid extra
operations.  For instance, in the affine component we use kPropagateAdds
and kBackpropAdds because AddMatMat adds to its output (and a zero
coefficient doesn't stop this, because it propagates infinities).
Dan

src/nnet3/nnet-test-utils.cc
```diff
@@ -793,7 +793,7 @@ void ComputeExampleComputationRequestSimple(
 
 static void GenerateRandomComponentConfig(std::string *component_type,
                                           std::string *config) {
-  int32 n = RandInt(0, 20);
+  int32 n = RandInt(0, 21);
```
vijayaditya: I think you might not have run
<code>make nnet-component-test </code>

Please run that and ensure all tests are successful.

src/nnet3/nnet-simple-component.cc
hhadian: Ok. This problem is also present in PerElementScaleComponent::Init.

danpovey: Thanks-- then fix that too in your pull request.

src/nnet3/nnet-simple-component.cc
hhadian: Yes of course all tests were passed successfully (and I had modified the RandInt  line)
I overwrite it but then add matrix 'in' to it. Isn't that right?

danpovey: Yes, that's probably right.

src/nnet3/nnet-simple-component.cc
hhadian: That's right. I'll remove the if statement altogether.

src/nnet3/nnet-simple-component.cc
hhadian: What to pass in place of in_value? I mean wrt kaldi coding standards. Is UpdateSimple(out_deriv, out_deriv) ok?

danpovey: There would be no UpdateSimple function- you should rewrite the function to
do any update in-place.  If you were to keep UpdateSimple (and don't do
this), then you would remove the extra argument.
For the Propagate and Backprop functions, you generally pass an empty
matrix for unused inputs.
Dan

src/nnet3/nnet-simple-component.h
hhadian: Yes this component is affine in both parameters and inputs but the kLinearInParameters bit is also set for AffineComponent which is not linear in all parameters (i.e. in bias parameters). Is that right?

danpovey: The kLinearInParameters flag means that if you scale the parameters by
alpha, you scale the output values by alpha.  This is true of the affine
component, but not of the offset-component.
Dan

hhadian: Ok. Thank you.

src/nnet3/nnet-test-utils.cc
hhadian: No I ran all the tests successfully. I tested again now and it was successful.

src/nnet3/nnet-test-utils.cc
```diff
@@ -984,6 +984,12 @@ static void GenerateRandomComponentConfig(std::string *component_type,
       os << "new-column-order=" << buffer.str();
       break;
     }
+    case 21: {
+      *component_type = "PerElementOffsetComponent";
+      os << "dim=" << RandInt(1, 100)
+         << " learning-rate=" << learning_rate << " param-mean=0.0 param-stddev=1.0";
```
hhadian: So I should also change the initialization that I have added here to "param-mean=0.0 param-stddev=0.0" ? 
Wouldn't that initialize offsets to all zero ? Is that ok for the test ?

src/nnet3/nnet-simple-component.cc
hhadian: Ok. Thank you. Should I keep the Update method or delete that too (currently it just calls UpdateSimple )?

src/nnet3/nnet-test-utils.cc
vijayaditya: It is better to randomly select param-mean and param-stddev.

src/nnet3/nnet-simple-component.cc
hhadian: And 1 more thing: if I rewrite the function Backprop to do the update in-place then the function can not be const anymore. Is that all right ?

vijayaditya: You would be updating the *to_update* component not the* this *component.
So the function can be const. Also remember to take the learning_Rate from
the *to_update * component.

--Vijay

hhadian: Ok. How about keeping the Update method? Should I delete it?

src/nnet3/nnet-test-utils.cc
hhadian: In what range ?
So it is better to randomly select these for the scale-component too ?

vijayaditya: I guess you can just have two cases where the both mean and stddev are 0 and one where mean is 0 and stddev is 1. Randomly select one of these two cases.

src/nnet3/nnet-simple-component.cc
vijayaditya: Yes.

src/nnet3/nnet-test-utils.cc
danpovey: On Mon, Nov 2, 2015 at 1:14 AM, hhadian <notifications@github.com> wrote:

> In src/nnet3/nnet-test-utils.cc
> <https://github.com/kaldi-asr/kaldi/pull/307#discussion_r43599487>:
>
> > @@ -984,6 +984,12 @@ static void GenerateRandomComponentConfig(std::string *component_type,
> >        os << "new-column-order=" << buffer.str();
> >        break;
> >      }
> > +    case 21: {
> > +      *component_type = "PerElementOffsetComponent";
> > +      os << "dim=" << RandInt(1, 100)
> > +         << " learning-rate=" << learning_rate << " param-mean=0.0 param-stddev=1.0";
>
> So I should also change the initialization that I have added here to
> "param-mean=0.0 param-stddev=0.0" ?
> Wouldn't that initialize offsets to all zero ? Is that ok for the test ?
>


Probably best to have nonzero parameters for the test.

hhadian: I changed it to randomly select one of the two cases. Is it better to revert it to always "param-mean=0.0 param-stddev=1.0" ?
I also did the rest of modifications.

danpovey: For testing code, random is good.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1012,185 @@ void PerElementScaleComponent::UnVectorize(
   scales_.CopyFromVec(params);
 }
 
+void PerElementOffsetComponent::Scale(BaseFloat scale) {
+  offsets_.Scale(scale);
+}
+
+void PerElementOffsetComponent::Resize(int32 dim) {
```
danpovey: We don't actually need the Resize function as it's not part of the interface.  It was likely needed in the place you got the code from, for a special purpose.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1012,185 @@ void PerElementScaleComponent::UnVectorize(
...
+void PerElementOffsetComponent::SetZero(bool treat_as_gradient) {
+  if (treat_as_gradient) {
+    SetLearningRate(1.0);
+    is_gradient_ = true;
+  }
+  offsets_.SetZero();
+}
+
+void PerElementOffsetComponent::PerturbParams(BaseFloat stddev) {
+  CuVector<BaseFloat> temp_offsets(offsets_);
```
danpovey: Just for good practice, I'd initialize this with 
  (offsets_.Dim(), kUndefined)
instead of
  (offsets_)

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1012,185 @@ void PerElementScaleComponent::UnVectorize(
...
+  std::stringstream stream;
+  BaseFloat offsets_mean = offsets_.Sum() / offsets_.Dim();
+  BaseFloat offsets_stddev = std::sqrt(VecVec(offsets_, offsets_) /
+                              offsets_.Dim()) - (offsets_mean * offsets_mean);
+  stream << Type() << ", input-dim=" << InputDim()
+         << ", output-dim=" << OutputDim()
+		 << ", offsets-mean=" << offsets_mean
+         << ", offsets-stddev=" << offsets_stddev
+         << ", learning-rate=" << LearningRate();
+  return stream.str();
```
danpovey: I think for both this and the scale component, it would also be useful to print out the min and max values.  The exact names are not important (this is mostly for humans to read).

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +1012,185 @@ void PerElementScaleComponent::UnVectorize(
...
+  UpdatableComponent::Init(learning_rate);
+  KALDI_ASSERT(dim > 0 && param_stddev >= 0.0);
+  offsets_.Resize(dim);
+  offsets_.SetRandn();
+  offsets_.Scale(param_stddev);
+  offsets_.Add(param_mean);
+}
+
+void PerElementOffsetComponent::Init(BaseFloat learning_rate,
+                                    std::string matrix_filename) {
```
danpovey: If we were to initialize this from a file, we'd store it as a vector, not as a matrix.  So this should be a vector, not a matrix-- and the names should reflect that.  The scale-component has the same issue.
Most likely we made it a matrix for some script reason when converting models from nnet1, but I don't intend to support nnet1 -> nnet3 conversion so this won't be an issue.  This feature isn't used right now, in either component.

src/nnet3/nnet-simple-component.h
```diff
@@ -309,6 +310,7 @@ class AffineComponent: public UpdatableComponent {
   Component *CollapseWithNext(const FixedAffineComponent &next) const;
```
danpovey: These CollapseWithNext functions are not being used right now, as the calling code does not exist.  So leaving them out is fine-- it looks like they are not implemented anyway.  You can also remove any similar declarations in the PerElementScaleComponent that have not been implemented.

src/nnet3/nnet-simple-component.h
```diff
@@ -873,6 +875,68 @@ class PerElementScaleComponent: public UpdatableComponent {
...
+  virtual int32 OutputDim() const { return offsets_.Dim(); }
+
+  virtual std::string Info() const;
+  virtual void InitFromConfig(ConfigLine *cfl);
+
+  PerElementOffsetComponent() { } // use Init to really initialize.
+  virtual std::string Type() const { return "PerElementOffsetComponent"; }
+  virtual int32 Properties() const {
+    return kSimpleComponent|kUpdatableComponent|
+           kBackpropInPlace|kPropagateInPlace;
```
danpovey: Please remove the kPropagateInPlace flag, as your propagation code would not work correctly if &in == out.
I just realized that we are not testing the kPropagateInPlace flag; I am adding testing code for this.

src/nnet3/nnet-test-utils.cc
danpovey: BTW, it can be helpful to do something to the code which should cause the tests to fail (e.g. scale the output by some constant factor in the propagate function, which should cause the derivative test to fail), and verify that it does fail.  This can catch situations where for some reason the testing code was never being invoked.

src/nnet3/nnet-simple-component.h
vijayaditya: @danpovey 
A clean-up of the unused code from Affine and other components, to ensure that the unused methods are not copied into newer classes, might be necessary.

Also there is no test for the Vectorize/Unvectorize methods. The Convolution*Components don't implement these methods and there was never a failed test.

danpovey: @hhadian, could you please also modify nnet-component-test.cc to add a test
for the Vectorize and Unvectorize() ?

This could be done using similar code to the code in
TestNnetComponentAddScale.
This could also test SetZero().
You could copy a component using Copy() and then set it to zero (use
treat_as_gradient == false), and then if ParameterDim() != 0, test that the
vectorized parameters are in fact zero.  Then vectorize the parameters of
the original component, call UnVectorize on the new component using the
resulting parameter vector, and make sure that the new component's
parameters are the same as the old ones (e.g. by comparing the info strings
and checking that the DotProduct calls return the same value).

You should also test that if the component does not say it is updatable
(kUpdatable == false), then the ParameterDim() is zero, and that Add(),
Scale(), and SetZero() are all no-ops, and that DotProduct returns zero...
again, you can use the you could also use the info-string test to check
this.  We are testing that non-updatable components behave as if they don't
have any parameters from the point of view of those functions:  this is
necessary to check that components that are 'optionally updatable', which
they may be in future, have consistent behavior if they are set as not
updatable.

Dan

hhadian: Yes. I'll do it today and include it in the pull request.

src/nnet3/nnet-simple-component.cc
hhadian: So I'll delete it.

src/nnet3/nnet-simple-component.cc
hhadian: That's definitely more efficient. I'll fix it in both scale- and offset- components.

src/nnet3/nnet-simple-component.cc
hhadian: Ok. I'll add it for both components.
I have made a mistake here. I must move (offsets_mean * offsets_mean) inside the sqrt.
Also in PerElementScaleComponent::Info the mean^2 term is not subtracted in stddev computation. Is it assumed to be zero ?
And finally is it not better to use sample std (i.e. /(N-1) instead of /N) ?

src/nnet3/nnet-simple-component.cc
hhadian: It was a question for me. I'll fix it.

src/nnet3/nnet-simple-component.h
hhadian: I could rewrite the propagate function to work fine even if &in and out point to the same memory. Something like this:
out->CopyFromMat(in);
out->AddVecToRows(1.0, offsets_);
I didn't know AddVecToRows existed. Is this fine?

src/nnet3/nnet-simple-component.h
hhadian: About the second part, if a component does not have the kUpdatable flag set, then probably it doesn't inherit from UpdatableComponent and so SetZero(), NumParameters(), and DotProduct() might not be available. 
However I can check that if a component does not have kUpdatable flag and is an instance of UpdatableComponent then these methods should do nothing. 
Also I believe these should in a separate function something like TestNnetComponentUpdatability. Is that right?

src/nnet3/nnet-simple-component.cc
danpovey: When we print out the stddevs we always actually print out the root-mean-squares... it's kind of misnamed.  I'll consider changing the name to rms, but later.

src/nnet3/nnet-simple-component.h
danpovey: Yes, rewriting like that would make sense.

src/nnet3/nnet-test-utils.cc
hhadian: Do you mean temporarily to test the testing codes ?

danpovey: Yes-- not to check in.

src/nnet3/nnet-simple-component.h
hhadian: About the second part, I'm not sure how to check Scale(), SetZero(), and Add(). This is the code I have come up with:
void TestNnetComponentVectorizeUnVectorize(Component *c) {
  UpdatableComponent *uc = dynamic_cast<UpdatableComponent*>(c);
  if((uc==NULL) || (uc->NumParameters() == 0))
    return;
  Vector<BaseFloat> params;
  UpdatableComponent *uc2 = dynamic_cast<UpdatableComponent*>(uc->Copy());
  uc2->SetZero(false);
  uc2->Vectorize(&params);
  KALDI_ASSERT(params.Min()==0.0 && params.Sum()==0.0);
  uc->Vectorize(&params);
  uc2->UnVectorize(params);
  KALDI_ASSERT(uc2->Info() == uc->Info());
  KALDI_ASSERT(uc2->DotProduct(*uc2) == uc->DotProduct(*uc));
  delete uc2;
}
void TestNnetComponentUpdatableFlag(Component *c) {
  UpdatableComponent *uc = dynamic_cast<UpdatableComponent*>(c);
  if(uc==NULL)
    return;
  if(!(uc->Properties() & kUpdatableComponent)){
    KALDI_ASSERT(uc->NumParameters() == 0);
    KALDI_ASSERT(uc->DotProduct(*uc) == 0);
    UpdatableComponent *uc2 = dynamic_cast<UpdatableComponent*>(uc->Copy());
    uc2->Scale(7.0);
    uc2->Add(3.0, *uc);
    KALDI_ASSERT(uc2->Info() == uc->Info());
    uc->SetZero(false);
    KALDI_ASSERT(uc2->Info() == uc->Info());
    delete uc2;
  }
}

Also this test caused Convolutional1dComponent to fail in Vectorize test because it does not implement it. Should I fix that too?

danpovey: That looks right, from a quick look.
Yes you should fix that.

hhadian: I took another look and Convolutional1dComponent  seems to be deprecated.

vijayaditya: Convolution1dComponent might be deprecated, but ConvolutionComponent also has the same issue (see #308 ). It would be great if you rectify this problem.

hhadian: Sure I will fix it right away.

hhadian: Currently the Vectorize functions do not automatically resize the out vector (*params). Isn't it better to automatically resize it to NumParameters() ?

danpovey: It writes to a VectorBase, which is not resizable... there are situations
where it's writing to a sub-vector of a larger vector, which is why we
designe it that way.
Dan

danpovey: And please fix the Convolutional1dComponent even if it's deprecated.. we
don't want tests to fail.

hhadian: Yes, of course. I'll fix that too.

src/nnet3/nnet-simple-component.cc
hhadian: Can scales in scale-component be negative? If yes then I think min and max would be more useful if they are in terms of magnitude; otherwise we might miss near-zero scales.

danpovey: There is nothing to stop that, but it's probably unlikely in practice.  You
could have min, max and min-abs.

src/nnet3/nnet-test-utils.cc
hhadian: I scaled the output and the derivative test failed.

danpovey: OK thanks.
So you've addressed all the issues we raised?

hhadian: Yes. Except abs-min which I assumed unnecessary based on your comment (and also there was no such function already implemented).
Also I did not remove CollapseWithNext() and Resize() from ScaleComponent because I was not sure if you wanted them to be deleted too.

vijayaditya: @dpovey
May be it would be better to just delete these methods from all the
components.

--Vijay

danpovey: Definitly remove the CollapseWithNext() and Resize() functions from this
component.
Regarding removing them from other components-- there may be a time when we
need them, so let's just leave the code there for now.
Dan

hhadian: I just removed them.

danpovey: @vijayaditya, do you have time to do a last pass of checking this commit?

vijayaditya: Yeah, I can do it.

src/nnet3/nnet-component-test.cc
```diff
@@ -54,6 +54,38 @@ void TestNnetComponentAddScale(Component *c) {
...
+  UpdatableComponent *uc2 = dynamic_cast<UpdatableComponent*>(uc->Copy());
+  uc2->SetZero(false);
+  Vector<BaseFloat> params(uc2->NumParameters());
+  uc2->Vectorize(&params);
+  KALDI_ASSERT(params.Min()==0.0 && params.Sum()==0.0);
+  uc->Vectorize(&params);
+  uc2->UnVectorize(params);
+  KALDI_ASSERT(uc2->Info() == uc->Info());
+  KALDI_ASSERT(uc2->DotProduct(*uc2) == uc->DotProduct(*uc));
+  delete uc2;
```
vijayaditya: This test wouldn't fail if there is a permutation of rows of the parameter matrix.

danpovey: true, this is a possible error that we need to worry about in that code .
maybe this could be solved by doing 'vectorize' one more time and checking
the result is the same as the first time.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +996,181 @@ void PerElementScaleComponent::UnVectorize(
...
+  if (cfl->GetValue("vector", &vector_filename)) {
+    Init(learning_rate, vector_filename);
+    if (cfl->GetValue("dim", &dim))
+      KALDI_ASSERT(dim == InputDim() &&
+                   "input-dim mismatch vs. vector.");
+  } else {
+    ok = ok && cfl->GetValue("dim", &dim);
+    BaseFloat param_mean = 0.0, param_stddev = 0.0;
+    cfl->GetValue("param-mean", &param_mean);
+    cfl->GetValue("param-stddev", &param_stddev);
```
vijayaditya: You haven't tested if ok is true here. The Init method will throw an error saying dim < 0 if there is no dim in the config string. This will be very confusing to the user.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1011,6 +996,181 @@ void PerElementScaleComponent::UnVectorize(
...
+  PerElementOffsetComponent *to_update =
+      dynamic_cast<PerElementOffsetComponent*>(to_update_in);
+
+  if (in_deriv) {
+    // Propagate the derivative back to the input.
+    in_deriv->CopyFromMat(out_deriv);
+  }
+
+  if (to_update != NULL) {
+    // Next update the model (must do this 2nd so the derivatives we propagate
```
vijayaditya: this comment is not relevant here. Delete it.

src/nnet3/nnet-simple-component.cc
```diff
@@ -2635,6 +2795,19 @@ int32 ConvolutionComponent::NumParameters() const {
   return (filter_params_.NumCols() + 1) * filter_params_.NumRows();
 }
 
+void ConvolutionComponent::Vectorize(VectorBase<BaseFloat> *params) const {
```
vijayaditya: I don't exactly know where Vectorize and Unvectorize are going to be used. But isn't it important to even add the hyper-parameters like filter and input dimensions to the params ?

danpovey: The only situations where it makes sense to compare these vectors is where
the nnets have the same structure, so no, don't include those things in the
params.  They are supposed to be just the parameters, and in some
situations we might scale them or average them, which doesn't make sense
for integer values.
Dan

vijayaditya:
Looks fine, good to merge.

vijayaditya:
Forgot to mention that the Vectorization unit test has to be modified to be more rigorous.

src/nnet3/nnet-simple-component.cc
hhadian: That is true. Actually I assumed the codes for scales-component to be already verified and mistake-free, and that's why I never checked them after copying. I will fix it for both components.

src/nnet3/nnet-component-test.cc
hhadian: I added another vectorize and checked elements of the params vector 1-by-1.  I left the two asserts (on Info() and DotProduct()) to stay as tests for Info() and DotProduct().

vijayaditya: you probably forgot to push the commit.

hhadian: Actually I was disconnected before pushing and I couldn't connect for ~1 hour.
Unfortunately my connection to CLSP grid is very unreliable and slow at some times. I'm not sure why.

danpovey: Probably it's to do with your own connection.  You might find it easier to
do your editing on a local machine, and only push your changes to CLSP when
you want to test something.  That's what I do when my connection is flaky.
Dan

hhadian: Yes the problem is from my side. 
Thank you that's a good idea. I'll cerainly try it when my connection is slow.

danpovey:
If this is ready to merge-- Hossein, could you please clean up the history using e.g. 'git rebase -i HEAD~10' 
.. then edit
pick dc607c1 Adds PerElementOffsetComponent
pick efdf774 Adds PerElementOffsetComponent to NewComponentOfType
pick 1b03814 Add PerElementOffsetComponent to GenerateRandomComponentConfig
pick c5e9276 minor KALDI_ASSERT fix
pick b3e85f4 Apply notes from pull request discussion
pick 3c352ff Minor code improvements regarding offsets- and scales- components
pick 5181725 Add Vectorize/Unvectorize to Convolution*Component
pick e4ec4d5 Add component tests for Vectorize/Unvectorize & kUpdatable
pick e5f9182 Remove CollapseWithNext and Resize from PerElementScaleComponent
pick 2482180 Minor code-improvements

by replacing all but the first 'pick' with 's' (for 'squash') ?

then push.
I have backed up your repo in case somethign goes wrong.

hhadian:
I did a rebase but then my push was rejected. I did some search and it seems I have to --force it.
This was the error:
To https://github.com/hhadian/kaldi.git
 ! [rejected]        master -> master (non-fast-forward)
error: failed to push some refs to 'https://github.com/hhadian/kaldi.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Merge the remote changes (e.g. 'git pull')
hint: before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

Should I force it now?

danpovey:
yes, force it

danpovey:
Thanks!  Merging.

PR #310, opened 2015-11-02
add bi-directional lstm recipe

naxingyu:


danpovey:
Thanks.  @vijayaditya  and @freewym, could you please have a look at this?

@naxingyu, I was expecting that you would want to check in example scripts (there isn't an example script even for HKUST), and more results-- I know that you have results on other setups.

egs/wsj/s5/steps/nnet3/lstm/make_configs.py
```diff
@@ -170,14 +172,30 @@ def ParseSpliceString(splice_indexes, label_delay=None):
...
+                                             lstm_delay = lstm_delay[i])
+            prev_layer_output_b = nodes.AddLstmLayer(config_lines, "Lstm{0}b".format(i+1), prev_layer_output, args.cell_dim,
+                                             args.recurrent_projection_dim, args.non_recurrent_projection_dim,
+                                             args.clipping_threshold, args.norm_based_clipping,
+                                             args.ng_per_element_scale_options, args.ng_affine_options,
+                                             lstm_delay = -lstm_delay[i])
+            prev_layer_output['descriptor'] = "Append(" + prev_layer_output_f['descriptor'] + ", " + prev_layer_output_b['descriptor'] + ")"
+            prev_layer_output['dimension'] = prev_layer_output_f['dimension'] + prev_layer_output_b['dimension']
+            nodes.AddFinalLayer(config_lines, prev_layer_output, args.num_targets, args.ng_affine_options, args.label_delay)
+            prev_layer_output['descriptor'] = prev_layer_output_f['descriptor'] + ", " + prev_layer_output_b['descriptor']
```
freewym: Does Append() need to been added for input of the latter hidden layer?

vijayaditya:
@freewym and @naxingyu looking at both your patches I can see that the only major change is in the make_configs.py file
@freewym 's patch allows for a more generic set of architectures. I would recommend using @freewym 's patch. @naxingyu  is that OK with you ?

danpovey:
Regardless of this, it would still be great to be able to see the example scripts and results from @naxingyu -- this will help us to configure things.

naxingyu:
Thanks for the comments.
@danpovey I added a --bi-directional option to the hkust/lstm recipe and made its default 'true' so currently that would serve as an example script. And the updated result of hkust came from this script.
@freewym You are right, I didn't think about that.
@vijayaditya I checked Yiming's patch. Yes, that is more genetic, and is more easily configured by just modifying the lstm-delay string. I'm OK with any decision you made.

danpovey:
You had results on a few different setups-- is there anything we should
know about how you configured this?  E.g. do you have some rule for
converting the configuration from LSTM to BLSTM (in terms of the
dimensions, etc.)?
Dan

naxingyu:
The results were on hkust and swbd. I didn't change any configuration 
from LSTM but made it bi-directional, i.e. the setups hasn't been 
specifically tuned for BLSTM.
The only tuning I tried was reducing the recurrent/non-recurrent 
dimensions but didn't work out as good as the current setup.

Xingyu

freewym:
Thanks. I am also trying to see if only reducing the recurrent/non-recurrent dimensions can improve the performance on WSJ.

naxingyu:
I'm closing this PR if it is not going to be merged.

danpovey:
OK.  Thanks again-- sorry for the change of plan.

PR #311, opened 2015-11-02
changed the default max-param-change in lstm training script from 1 to 2

xiaohui-zhang:


danpovey:
I am committing this, but I'd like to warn everyone who is doing LSTM experiments to be careful!
You might have to add --max-change 1 to example scripts if you want to ensure backward compatibility.  However, I expect that this setup will usually be better than what was before.

PR #313, opened 2015-11-05
fix the gst-plugin build script to work with clang

rikrd:
These changes to the gst-plugin Makefile allow it to be built using clang.

### DIFF
--- a/src/gst-plugin/Makefile
+++ b/src/gst-plugin/Makefile
@@ -9,7 +9,7 @@ EXTRA_CXXFLAGS += -Wno-sign-compare -I ../../tools/portaudio/install/include
 EXTRA_CXXFLAGS += $(shell pkg-config --cflags gstreamer-1.0)
 EXTRA_CXXFLAGS += $(shell pkg-config --cflags glib-2.0)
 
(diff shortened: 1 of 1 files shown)


danpovey:
Have you tested that it still works with gcc?

rikrd:
No, haven't tried it. I can try it next week.
Doesn't the automatic build do that?

danpovey:
The automatic build only tests the core Kaldi stuff, not the extensions
like the gst plugin.

danpovey:
Any update on this?  Have you tested with gcc?

danpovey:
@alumae, what do you think we should do about this?

rikrd:
I had completely forgot about this.

I've just tried with gcc right now, and built it from a clean checkout of
this branch just fine.

As I side note, I did bump into this:
http://sourceforge.net/p/kaldi/discussion/1355348/thread/32cd7922/
But it is unrelated to this pull request.

On Wed, Nov 25, 2015 at 8:18 PM, Daniel Povey <notifications@github.com>
wrote:

> @alumae <https://github.com/alumae>, what do you think we should do about
> this?
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/313#issuecomment-159721229>.
>



-- 
ricard
http://twitter.com/ricardmp
http://www.ricardmarxer.com
http://www.caligraft.com

danpovey:
Merging- thanks!

PR #314, opened 2015-11-06
Make sparse-matrix-test to compile with gcc -O2

akreal:


danpovey:
Thank you, Rasputin!  I will commit when the checks complete.

akreal:
You are welcome! :-)

PR #323, opened 2015-11-11
Fixed data prep in SWBD, to ensure CV data does not overlap with trai…

vijayaditya:
…n in speed perturbed case

Modifies utils/copy_data_dir.sh

### DIFF
--- a/egs/swbd/s5c/local/nnet3/run_ivector_common.sh
+++ b/egs/swbd/s5c/local/nnet3/run_ivector_common.sh
@@ -21,21 +21,23 @@ if [ "$speed_perturb" == "true" ]; then
     for datadir in train_nodup; do
       utils/perturb_data_dir_speed.sh 0.9 data/${datadir} data/temp1
       utils/perturb_data_dir_speed.sh 1.1 data/${datadir} data/temp2
-      utils/combine_data.sh --extra-files utt2uniq data/${datadir}_tmp data/temp1 data/temp2
+      utils/combine_data.sh data/${datadir}_tmp data/temp1 data/temp2
+      utils/validate_data_dir.sh --no-feats data/${datadir}_tmp
       rm -r data/temp1 data/temp2
-
+      
       mfccdir=mfcc_perturbed
       steps/make_mfcc.sh --cmd "$train_cmd" --nj 50 \
         data/${datadir}_tmp exp/make_mfcc/${datadir}_tmp $mfccdir || exit 1;
       steps/compute_cmvn_stats.sh data/${datadir}_tmp exp/make_mfcc/${datadir}_tmp $mfccdir || exit 1;
       utils/fix_data_dir.sh data/${datadir}_tmp
-
+      
       utils/copy_data_dir.sh --spk-prefix sp1.0- --utt-prefix sp1.0- data/${datadir} data/temp0
       utils/combine_data.sh data/${datadir}_sp data/${datadir}_tmp data/temp0
+      utils/fix_data_dir.sh data/${datadir}_sp
       rm -r data/temp0 data/${datadir}_tmp
     done
   fi
-
+  
   if [ $stage -le 2 ] && [ "$generate_alignments" == "true" ]; then
     #obtain the alignment of the perturbed data
     steps/align_fmllr.sh --nj 100 --cmd "$train_cmd" \
--- a/egs/wsj/s5/utils/combine_data.sh
+++ b/egs/wsj/s5/utils/combine_data.sh
@@ -39,6 +39,32 @@ for dir in $*; do
   fi
 done
 
+# W.r.t. utt2uniq file the script has different behavior compared to other files
+# it is not compulsary for it to exist in src directories, but if it exists in 
+# even one it should exist in all. We will create the files where necessary
+has_utt2uniq=false
+for in_dir in $*; do
+  if [ -f $in_dir/utt2uniq ]; then
+    has_utt2uniq=true
+    break
+  fi
+done
+
+if $has_utt2uniq; then
+  # we are going to create an utt2uniq file in the destdir
+  for in_dir in $*; do
+    if [ ! -f $in_dir/utt2uniq ]; then
+      # we assume that utt2uniq is a one to one mapping
+      cat $in_dir/utt2spk | awk '{printf("%s %s\n", $1, $1);}' 
+    else
+      cat $in_dir/utt2uniq
+    fi
+  done | sort -k1 > $dest/utt2uniq
+  echo "$0: combined utt2uniq"
+fi
+# some of the old scripts might provide utt2uniq as an extrafile, so just remove it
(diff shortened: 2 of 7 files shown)


danpovey:
I don't think this is the right solution-- having copy_data_dir.sh automatically create utt2uniq is a bit messy, as it adds that files everywhere even if it's not needed.

I think a better solution would be to modify combine_data.sh so that if any of the source directories have an utt2uniq file, it combines the utt2uniq files, but in that case if any of the source directories do *not* have the utt2uniq files, it will use a one-to-one mapping as the utt2uniq file, so each utterance maps to itself.  I believe this will do the right thing in our existing setups.

vijayaditya:
copy_data_dir.sh provides the options to add prefixes/suffixes. Prefix option is actually used in the SWBD recipe. The utterance id would be modified in this case and when combined_data.sh works on this copied directory without the utt2uniq file, it won't know that the new utterance id in the copied directory corresponds to a unique utterance id which all utt2uniq files in other directories point to.

vijayaditya:
Rather than relying on the user to ensure that they do not modify the utterance ids unknowingly, I think it might be better to always preserve the utterance ids using utt2uniq files in all setups.

danpovey:
Perhaps we can have a convention that if there is no  'utt2uniq' file it
means no prefixes/suffixes were ever added, so utt2uniq is effectively a
one-to-one map, but any operations that add prefixes/suffixes should create
an utt2uniq file.  Then things like copy_data_dir.sh and
combine_data_dir.sh and validate_data_dir.sh should respect the utt2uniq
file automatically.
Dan

danpovey:
Even if we put the utt2uniq file everywhere, we would still have to assume
that no utt2uniq file means it's a one-to-one map, for back-compatibility
reasons.  So why don't you implement it my way for now-- and any tests with
this setup will automatically make sure the back-compatibility code works--
and then later on if we decide, we can have things like copy_data_dir.sh
automatically create the utt2uniq file.

vijayaditya:
ok.. making the modifications.

vijayaditya:
done.. modified combine_data.sh to assume utt2uniq is a one-to-one mapping when it is not present. It creates this file in the directories that don't have it.

egs/swbd/s5c/local/nnet3/run_ivector_common.sh
```diff
@@ -31,7 +31,7 @@ if [ "$speed_perturb" == "true" ]; then
       utils/fix_data_dir.sh data/${datadir}_tmp
 
       utils/copy_data_dir.sh --spk-prefix sp1.0- --utt-prefix sp1.0- data/${datadir} data/temp0
-      utils/combine_data.sh data/${datadir}_sp data/${datadir}_tmp data/temp0
+      utils/combine_data.sh --extra-files utt2uniq data/${datadir}_sp data/${datadir}_tmp data/temp0
```
danpovey: there should be no need for the '--extra-files' argument here.  it should happen automatically.

egs/wsj/s5/utils/copy_data_dir.sh
```diff
@@ -57,6 +57,14 @@ mkdir -p $destdir
 cat $srcdir/utt2spk | awk -v p=$utt_prefix -v s=$utt_suffix '{printf("%s %s%s%s\n", $1, p, $1, s);}' > $destdir/utt_map
 cat $srcdir/spk2utt | awk -v p=$spk_prefix -v s=$spk_suffix '{printf("%s %s%s%s\n", $1, p, $1, s);}' > $destdir/spk_map
 
+if [ ! -f $srcdir/utt2uniq ]; then
+  if [[ ! -z $utt_prefix  ||  ! -z $utt_suffix ]]; then
```
vijayaditya: modified to create an utt2uniq file only if there is a prefix or suffix.

egs/swbd/s5c/local/nnet3/run_ivector_common.sh
danpovey: ... well, it should happen automatically if any source directory has the utt2uniq file.
Otherwise there is no need, if we follow the convention that no utt2uniq file means the utterance-id itself is the utt2uniq file, because having no utt2uniq file gives us the correct mapping.

egs/wsj/s5/utils/combine_data.sh
```diff
@@ -39,6 +39,21 @@ for dir in $*; do
...
+# it is not compulsary for it to exist in all directories
+# when it does not exist we have to create it
+# check if utt2uniq is in the extra_files
+for file in $extra_files; do
+  if [ "$file" == "utt2uniq" ]; then
+    for in_dir in $*; do
+      if [ ! -f $in_dir/utt2uniq ]; then
+        # we assume that utt2uniq is a one to one mapping
+        cat $in_dir/utt2spk | awk '{printf("%s %s\n", $1, $1);}' > $in_dir/utt2uniq
+      fi
```
danpovey: I don't think creating the utt2uniq files in the source directories is the right solution-- you should have the conditional at the time you combine them-- if you do combine them.
Conceptually, the source directories are 'const'.  bash doesn't have 'const', but in our heads they are const.

vijayaditya:
modifications done

egs/wsj/s5/utils/combine_data.sh
```diff
@@ -39,6 +39,26 @@ for dir in $*; do
   fi
 done
 
+# W.r.t. utt2uniq file the script has different behavior compared to other files
+# it is not compulsary for it to exist in src directories, but if it exists in 
+# even one it should exist in all. We will create the files where necessary
+num_utt2uniq_files=$(find $* -name "utt2uniq"|wc -l)
```
danpovey: this is dangerous due to .backup directories.  You should iterate instead, e.g. for d in $*; do [ -f $x/utt2uniq ] && echo yes; done | wc -l

vijayaditya:
rectified that..

danpovey:
Thanks.
Because this type of change could cause a lot of problems if there is a bug, I'd rather have you and Tom use the code, and wait a few days before merging... if that's OK.

vijayaditya:
Sure. I tested it on SWBD. Tom might be able to test it on ASpIRE and
Yiming might be able to test it on WSJ.

I just want to inform people (some cc'ed) using SWBD nnet3 recipes that
that the nnet3 systems will perform better on CV than train data during
training.

--Vijay

danpovey:
Small-scale testing, like WSJ or even RM, should be enough.

danpovey:
Also, please update validate_data_dir.sh: you can add utt2uniq to the list currently including 'vad.scp utt2lang'.

vijayaditya:
Done.

danpovey:
@vijayaditya, do you think this is safe to merge?  I think this problem you are fixing here is causing issues for others too.

vijayaditya:
Had to modify fix_data_dir.sh to correct utt2uniq file.  The changes work on SWBD and do not break other perturbed setup (tedlium).

Ready to merge.

PR #326, opened 2015-11-13
Curowsparsematrix class

naxingyu:


danpovey:
OK, I will merge this so progress can continue.  I'll check it later, or ask someone else to.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -37,6 +37,88 @@
...
+
+  /// Copy to CPU-based matrix. We will add the transpose option later when it
+  /// is necessary.
+  template <typename OtherReal>
+  void CopyToSmat(SparseMatrix<OtherReal> *smat) const;
+
+  /// Swap with CPU-based matrix.
+  void Swap(SparseMatrix<Real> *smat);
+
+  /// Swap with possibly-CPU-based matrix.
```
danpovey: this should probably say 'possibly-GPU-based matrix'

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -37,6 +37,88 @@
...
+ private:
+  // This member is only used if we did not compile for the GPU, or if the GPU
+  // is not enabled.  It needs to be first because we reinterpret_cast this
+  std::vector<SparseVector<Real> > cpu_rows_;
+
+  MatrixIndexT num_rows_;
+  MatrixIndexT num_cols_;
+  MatrixIndexT elements_per_row_;
+  CuArray<RowElement<Real> > data_; // pairs(column-index, value)
+  // dim = num_rows_ * elements_per_row_.
```
danpovey: you should document here that if any row has fewer than 'elements_per_row' elements, we pad with elements that have -1 as the row index.

src/cudamatrix/cu-sparse-matrix.cc
```diff
@@ -42,6 +42,205 @@
...
+  {
+    num_rows_ = smat->NumRows();
+    num_cols_ = smat->NumCols();
+    Mat().Swap(smat);
+  }
+}
+
+template <typename Real>
+void CuRowSparseMatrix<Real>::Swap(CuRowSparseMatrix<Real> *smat) {
+#if HAVE_CUDA == 1
```
danpovey: This Swap() implementation is more complicated than it needs to be.
Just do a shallow swap of each element, i.e.:
cpu_rows_.swap(other->cpu_rows);
std::swap(num_rows_, other->num_rows_);
std::swap(num_cols_, other->num_cols_);
std::swap(elements_per_row_, other->elements_per_row_);
data_.Swap(&(other->data_));

Even if a data-member isn't being used, swapping emepty members is OK.

naxingyu: I don't see a Swap implementation in CuArray. Should I add it?
Xingyu

danpovey: Yes.

PR #328, opened 2015-11-13
Refactoring run_dnn.sh

mrtdoulaty:
Added few variables to hold some common paths and avoid repeating them in the following lines

### DIFF
--- a/egs/wsj/s5/local/nnet/run_dnn.sh
+++ b/egs/wsj/s5/local/nnet/run_dnn.sh
@@ -3,16 +3,16 @@
 # Copyright 2012-2014  Brno University of Technology (Author: Karel Vesely)
 # Apache 2.0
 
-# This example script trains a DNN on top of fMLLR features. 
+# This example script trains a DNN on top of fMLLR features.
 # The training is done in 3 stages,
 #
 # 1) RBM pre-training:
-#    in this unsupervised stage we train stack of RBMs, 
+#    in this unsupervised stage we train stack of RBMs,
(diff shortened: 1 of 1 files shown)


danpovey:
I'll let Karel decide as this is an nnet1 script.
Personally I prefer my own example scripts to be fairly concrete and not use too many variables, so if it was my own script I would not necessarily accept this patch.

mrtdoulaty:
The main reason to add extra variables (which generally I don't like to have many of them as well) is that when you want to use a different location for a particular experiment (e.g. it's different from the main one), then you have to change the same path in more than a single location in the script and that can cause some further issues. But setting them in only one location makes it easier to change.

danpovey:
Personally I would rather just have people replace them manually if they
need to.  This script is very old and I'm not sure that it even represents
the state of the art in nnet1-- let alone Kaldi neural nets in general.  So
only if something is a bug, or like a bug, I would tend to fix it in these
scripts.
Dan

vesis84:
Hi, frankly, I don't really like the proposed change. First it would cause an inconsistency with scripts for the other databases (RM,SWBD). Second, I believe the script should be easy to read and ready to do 'quick copy and modify'. For the manual changes it is excellent to use the regular expression matching, which finds all the locations of some string, say a directory name. In vim you can simply press '*' while having cursor on a word. And lastly, for tuning something in the middle of the script, it is more convenient to be able to redefine the output directory nearby, which is as it is now... Best, Karel.

PR #315, opened 2015-11-06
make codes in src/util/ comply with Google style

keli78:
make codes in src/util comply with google C++ style

PR #329, opened 2015-11-13
Fix clang compiler errors.

dogancan:
It seems these were inadvertently introduced while making the code in util/ comply with the Google style (#315).


danpovey:
Thanks!  Merging.

PR #331, opened 2015-11-16
Stft toeplitz

hknerdgn:
This branch includes short-time Fourier transform (STFT) and inverse STFT transform implementations for kaldi made during Jelinek 2015 workshop. The branch also includes channel adaptation code for linearly filtering one signal in time-domain to make it closer to another signal. The branch includes unit tests for STFT, inverse STFT and channel adaptation.

Hakan Erdogan


danpovey:
Hm.
There are a lot of issues with this commit-- if you want anything of this
committed, we're going to have to go bit by bit, in quite small chunks.
Firstly, with all the fourier transform stuff, you seem to have implemented
it as a non-fast fourier transform, without realizing that there is already
code for fast fourier transforms in Kaldi (which is how we are able to
compute MFCCs).  The only thing that isn't implemented is a binary that
makes the fourier transform available from the command line (we didn't need
one since we generally use mel bins).  To implement that would involve much
less code than what you have added, and anyway I'd probably want to wait
till it was needed in some example script.
The channel adaptation stuff is potentially interesting but it's hard to
know what is going on there as it's not really documented (e.g. the
ChannelAdapt function doesn't have any documentation and the variable names
are single-letter).
Another issue is that all of this code is 'orphan code'.  What I mean is
that it is not used anywhere in the scripts.  In general we don't check in
orphan code like this.  If you could show a use-case for some of it, e.g.
the channel adaptation stuff, and supplied an example script for one of the
setups, that could be interesting.  For example, you might be able to work
with @vijayaditya and Tom Ko (@tomko?) to demonstrate that the the channel
adaptation is useful for the Aspire setup, where we know there is a channel
mismatch as the Singapore people got improvement from simulating the
telephone channel response.
It could be that some of this stuff is useful because it's used in some
setup you had in the workshop, that ended up working somehow.  But I'd want
to consider it together with those scripts, and not checking in this first
and then the scripts.

Dan

hknerdgn:
Dan,

I am using the FFT code within kaldi actually. It may look a bit crowded due to options handling and things like that. It may be the case that I have more options than needed :). We can submit all the chime3 experiments together with this commit as well, where these things are used. But I thought we can do it separately since these things may be useful elsewhere.

Basically, the use case for STFT and ISTFT is for speech enhancement experiments. If you would like to run a mask-based enhancement system within kaldi, it would be useful. Another use case could be when you want to generate speech data from spectral domain data such as HMM-based synthesis, voice conversion, bandwidth extension and some other similar algorithms.

The use case for channel adaptation involves having a target of speech enhancement algorithm when the target is the signal at a close talking microphone and the noisy source is obtained from a distant talking one. We would want to adapt the cleaner target to the channel of the input such that it could be used as a target for a mask-based enhancement algorithm. Another use case may be when you want to evaluate signal level accuracy of speech denoising. In fact, if we want to implement metrics like SDR, SIR and SAR, we would need this functionality.

I agree that the Toeplitz solver can be cleaned up (get rid of single letter variables and things like that) and I can work on it a bit more.

Let me know how you would like to proceed. I can work with Vijay on these if you like.

danpovey:
I think before we can proceed we should talk about a specific use-case and how we will demonstrate its usefulness in an example script.  I don't want to add anything just because it *might* be useful in future- that causes bloat.  If you have something that you think will improve performance in some scenario, whether in an existing example script or a new one, then let's talk about that.  But let's take this offline for now-- think about it and email me and Vijay; I'll close this pull request until we can pin this down.

PR #336, opened 2015-11-17
WIP: Add SparseLinearComponent class

hhadian:
Some notes:
1. In CuRowSparseMatrix::NumRows() : we should either always return Mat().NumRows() when no-cuda instead of num_rows_,  or otherwise we have to make sure everywhere Mat() is modified we update num_rows_  too (also for num_cols_).
2. I'm not sure if stddev is meaningful for a SparseLinearComponent. It might be good to show zero-percentage in Info(). If it is good, I will add it.
3. Note: CopyFromMat/ToMat for sparse-matrix is actually LoadFromMat/ToMat because they resize automatically. Should I change this ?
4. I'm not sure about line 81 of nnet-component-test.cc : if(!(uc->Properties() & kUpdatableComponent) || (uc->NumParameters() == 0))
Shouldn't all UpdatableComponent instances have the kUpdatableComponent flag set ?


danpovey:
> Some notes:
> 1. In CuRowSparseMatrix::NumRows() : we should either always return
> Mat().NumRows() when no-cuda instead of num_rows_, or otherwise we have to
> make sure everywhere Mat() is modified we update num_rows_ too (also for
> num_cols_).
>
Sure.

> 2. I'm not sure if stddev is meaningful for a SparseLinearComponent. It
> might be good to show zero-percentage in Info(). If it is good, I will add
> it.
>
Sure.

> 3. Note: CopyFromMat/ToMat for sparse-matrix is actually LoadFromMat/ToMat
> because they resize automatically. Should I change this ?
>
I think it's OK in this case to just document that it changes the size.

> 4. I'm not sure about line 81 of nnet-component-test.cc :
> if(!(uc->Properties() & kUpdatableComponent) || (uc->NumParameters() == 0))
> Shouldn't all UpdatableComponent instances have the kUpdatableComponent
> flag set ?
>
No, it's possible (at least, in the future) to inherit from
UpdatableComponent and not be in updatable mode, and the flag won't be
set.  The flag is the accurate one.


Dan

hhadian:
Thank you. Also to test CopyFromMat/ToMat of SparseMatrix I added a test function (TestNnetComponentScaleAddDotProduct) to make sure all these functions work properly for SparseLinearComponent because they were not tested elsewhere.

hhadian:
Yes, there was conflict regarding NumRows(), NumCols(), and NumElements(). You can see how I resolved it in https://github.com/hhadian/kaldi/commit/87418bc76393ab4215b755084c114f082447d62b
Briefly: 
1. I put "num_rows_ = smat.num_rows_;     num_cols_ = smat.num_cols_;" out of the #if-block in the assignment operator, because they are used in AddMatSmat asserts and this caused problem after copying. I also changed NumRows(), NumCols(), and NumElements() to return Mat().NumRows() and Mat().NumCols() and Mat().NumElements() when in CPU-mode to make sure they are always right.
2. I removed the two template function specializations at line 58 of cu-sparse-matrix.cc and added a class template specialization at the end of the file for all functions.

That's all. If I mistakenly changed a line, please let me know to fix it.

PR #338, opened 2015-11-17
WIP: modify WaveData::Read(std::istream &is) function to deal with data re…

Minhua722:
modify WaveData::Read(std::istream &is) function to deal with data read in stream mode when we don't know its exact size in advance.

If data are read in stream mode, we will get riff_chunk_size, data_chunk_size either 0 or max of Uint32. The function is modified to deal with this case separately.


danpovey:
@kkm000 , can you please comment on this?
Vijay will also check.

src/feat/wave-reader.cc
```diff
@@ -227,49 +227,73 @@ void WaveData::Read(std::istream &is) {
...
+  std::vector<char> chunk_data_vec;
   char *data_ptr = &(chunk_data_vec[0]);
-  if (num_bytes_read == 0 && num_bytes_read != data_chunk_size) {
-    KALDI_ERR << "WaveData: failed to read data chunk (read no bytes)";
-  } else if (num_bytes_read != data_chunk_size) {
-    KALDI_ASSERT(num_bytes_read < data_chunk_size);
-    KALDI_WARN << "Read fewer bytes than specified in the header: "
-               << num_bytes_read << " < " << data_chunk_size;
+  uint32 num_bytes_read = 0;
+  if (!((riff_chunk_size == 0 && data_chunk_size == 0)
```
vijayaditya: rather performing the same check as above again, just use a bool like <code>is_stream_mode </code> which is set the first time you detect possible stream mode.

src/feat/wave-reader.cc
```diff
@@ -227,49 +227,73 @@ void WaveData::Read(std::istream &is) {
...
+                   << "so that we don't know exact size in advance";
+    else
+        KALDI_ERR << "Expected " << riff_chunk_size << " bytes in RIFF chunk, but "
+                  << "after first data block there will be " << riff_chunk_read
+                  << " + " << data_chunk_size << " bytes "
+                  << "(we do not support reading multiple data chunks).";
   }
 
+  std::vector<char> chunk_data_vec;
   char *data_ptr = &(chunk_data_vec[0]);
```
vijayaditya: I don't think this is a robust to vector resizes. See http://stackoverflow.com/questions/2447392/does-stdvector-change-its-address-how-to-avoid

src/feat/wave-reader.cc
```diff
@@ -227,49 +227,73 @@ void WaveData::Read(std::istream &is) {
...
-
-  std::vector<char> chunk_data_vec(num_bytes_read);
-  uint32 data_address = 0;
-  for (int i = 0; i < data_pointer_vec.size(); i++) {
-    memcpy(&(chunk_data_vec[data_address]), data_pointer_vec[i],
-           data_size_vec[i]);
-    delete[] data_pointer_vec[i];
-    data_address += data_size_vec[i];
+    if ((riff_chunk_size == 0 && data_chunk_size == 0) 
+         || (riff_chunk_size == 4294967295 && data_chunk_size == 4294967295))
```
kkm000: I would rather use `std::numeric_limits<uint32>::max()` here

src/feat/wave-reader.cc
```diff
@@ -227,49 +227,73 @@ void WaveData::Read(std::istream &is) {
...
+        KALDI_WARN << "Read in " << riff_chunk_size << " bytes in RIFF chunk, "
+                   << "Read in " << riff_chunk_read << " bytes in data chunk."
```
kkm000: Add a space at the end of this and the next string constant.

src/feat/wave-reader.cc
```diff
@@ -227,49 +227,73 @@ void WaveData::Read(std::istream &is) {
...
-    KALDI_ASSERT(num_bytes_read < data_chunk_size);
-    KALDI_WARN << "Read fewer bytes than specified in the header: "
-               << num_bytes_read << " < " << data_chunk_size;
+  uint32 num_bytes_read = 0;
+  if (!((riff_chunk_size == 0 && data_chunk_size == 0)
+     || (riff_chunk_size == 4294967295 && data_chunk_size == 4294967295))) {
+     std::vector<char*> data_pointer_vec;
+     std::vector<int> data_size_vec;
+     num_bytes_read = 0;
+     for (int32 remain_chunk_size = data_chunk_size; remain_chunk_size > 0;
```
kkm000: maybe use `std::vector<char>` or `std::stringstream` to accumulate the data bytes, and let it handle the resizes? The code would be simpler to read. AFAIK, `std::vector` is handling resizes quite reasonably.

src/feat/wave-reader.cc
```diff
@@ -227,49 +227,73 @@ void WaveData::Read(std::istream &is) {
...
+     }
+  } else {
+    // when data are read in stream model, we don't know size in advance
+    // data_chunk_size and riff_chunk_size (Uint32) could be 
+    // either ox00000000 or oxFFFFFFFF
+    // we need to read in the stream differently in this case
+    chunk_data_vec.clear();
+    char this_data_byte;
+    num_bytes_read = 0;
+    // read it byte by byte since we don't know the exact size
```
kkm000: Using a buffer with `is.read` (and `is.gcount()` to query the number of bytes actually read) is usually more efficient. `std::back_inserter` is your friend here (as it is in the above loop I commented on).

danpovey: Kirill, some code snippets might be helpful here.  Even I'm not sure what
you mean about std::back_inserter (not familiar with it.)


On Wed, Nov 18, 2015 at 1:09 AM, Kirill Katsnelson <notifications@github.com

src/feat/wave-reader.cc
Minhua722: Is it acceptable if I put "char *data_ptr = &(chunk_data_vec[0]);" after chunk_data_vec has been fully processed, since the original code uses this pointer for the following data_ read.

src/feat/wave-reader.cc
Minhua722: I added "is_stream_mode" in the new commit

src/feat/wave-reader.cc
Minhua722: Thanks, I used this in the new commit

src/feat/wave-reader.cc
Minhua722: Sorry, I can't quite understand which lines you comment on. The data bytes are now accumulated using std::vector<char> chunk_data_vec; (based on std::vector<char*> data_pointer_vec; and std::vector<int> data_size_vec;) when is_stream_mode == false

src/feat/wave-reader.cc
kkm000: @danpovey: I have been unable to answer immediately, sorry. The idea is to read sizable chunks from the stream, and push them into the `chunk_data_vec`:

```C++
while (is.good()) {
  char buffer[256];
  is.read(buffer, sizeof buffer);
  std::copy(buffer, buffer + is.gcount(), std::back_inserter(chunk_data_vec));
}
```

`std::back_inserter` is a convenience function to create a back insert iterator, the iterator type that effectively calls `vector.push_back(x)` for each `*iterator++ = x` operation.

Minhua722: @kkm000 Thanks! I've applied them in the new commit.

kkm000:
@Minhua722: This module has no tests. I started writing them but never actually got anywhere far on that. Do you think you would be able to create a few test cases? All it should take is put some known data into different wave files and then read it and verify it is read correctly. There is already the `test_data` suborder for these

The machinery being added is quite complex and has some edge cases. A real danger for code like this (and any signal processing code in general) is that quite serious bugs in it can go undiscovered for years, as ML algorithms would often show only a very slight decrease in performance on quite mangled data. I do not know how much time do you have on hand, but adding tests would be a very big improvement for this part of kaldi. I am keeping #88 open to keep that on my radar.

src/feat/wave-reader.cc
```diff
@@ -221,55 +221,88 @@ void WaveData::Read(std::istream &is) {
...
-  }
-
-  std::vector<char> chunk_data_vec(num_bytes_read);
-  uint32 data_address = 0;
-  for (int i = 0; i < data_pointer_vec.size(); i++) {
-    memcpy(&(chunk_data_vec[data_address]), data_pointer_vec[i],
-           data_size_vec[i]);
-    delete[] data_pointer_vec[i];
-    data_address += data_size_vec[i];
+  if (!is_stream_mode) {
```
kkm000: Can you maybe rewrite this branch as well? I understand it is old code, but it is extremely ugly. No need to allocate a kerzillion of small buffers, read into them, then merge into a large one. Just read all into its final location in one single read! You should actually get 3 lines of code instead of 40 some. Something to the effect of

```C++
chunk_data_vec.resize(data_chunk_size);
is.read(&chunk_data_vec.first(), data_chunk_size);
uint32 num_bytes_read = is.gcount();
```

This is actually all it should take to read the chunk!

src/feat/wave-reader.cc
```diff
@@ -221,55 +221,88 @@ void WaveData::Read(std::istream &is) {
...
+     }
+  } else {
+    // when data are read in stream model, we don't know size in advance
+    // data_chunk_size and riff_chunk_size (Uint32) could be 
+    // either ox00000000 or oxFFFFFFFF
+    // we need to read in the stream differently in this case
+    chunk_data_vec.clear();
+    num_bytes_read = 0;
+    while (is.good()) {
+      char buffer[256];
```
kkm000: I think you shouild better use s/256/kBlockSize/ -- the constant is already defined.

Minhua722:
@kkm000 I've rewritten the original wave read code. 

For the test module staff, I'm not sure if I understand the problem properly and I've sent you another email separately.

Minhua722:
I've added my current understanding of the test staff

kkm000:
@Minhua722: It might be really unsimple to properly test this streaming code. The idea I have in mind is to prepare a couple files (representing egde cases, like finite stream and infinite stream) of known simple contents and compare the result. You are on the right way, but I do not think the matlab ASCII file is the right data set to compare. But in any case, if you are not comfortable doing that, do not worry if you decide to drop it on the floor; I'll pick it up! :)

danpovey:
One way to do this might be to have a function like this:

void CreateWaveTestCase(std::ostream &os,
                                      WaveData *data);

that will create some random wave  data and write it in a randomized way
(i.e. possibly with zero or infinite or inaccurate header and possibly
multiple data chunks), and for reference will write the data that the
stream is supposed to represent.
Then the wave-reading code could read that stream and make sure the
resulting object is the same  as it is supposed to be.

Dan

Minhua722:
Sorry. I'm still confused of what I should do... 
@kkm000 Maybe I could work on it after you push an initial version.

danpovey:
Kirill, how about I wait until the test code is done before merging this
pull request.. I don't think it's urgent.

kkm000:
@danpovey, @Minhua722: No problem, I'll pick it up from this point.

kkm000:
I am marking this original PR WIP since I am pulling in this branch to work on it.

kkm000:
@danpovey: I have been working on the tests tonight, and wanted to come up with a test scaffolding to easily code different structure of wave files we support (and there is a lot of edge cases already). What I am thinking of is a simple syntax to express the prototype file structure right in code, and then allowing tests with different combinations of byte ordering and streaming/non-streaming encoding produced from this prototype. Essentially I am thinking along the lines of

```C++
TestWaveData wave_file(
  Chunk("fmt ",  L(1), S(2), S(3)),
  Chunk("data", S(4), S(5), S(6)));
istream& stm = wave_file.RenderStream(byte_order_flag, infinite_length_flag, ...);
```
`L` and `S` are for a "long" (32-bit) and "short" data words, and the combinations of flags would provide multiple different streams to feed `WaveReader`.

This all is cute and nice. C++ however is not the easiest language to express these things (ah, Lisp would ve very natural!). My first attempt at a preprocessor abuse did not end up well: I came up with a horrific combo of macros and classes to make a nicer syntax possible. I do not really like it. My working implementation of the exact syntax above uses some C++11 features, and they are disabled in the compilation by default.

What do you think of this--would you be ok with enabling -std=c++0x for tests only? They are not part of the Kaldi main library code anyway. This should be fine with gcc 4.7+, and MSVC and Intel are fine with that too, if anyone ever runs tests under these at all except me. I do not really know right now how to express these test cases with binary streams concisely without the new language features.

danpovey:
Why don't you just #ifdef the tests?  Kald may still get built on systems
without c++11.

Dan


On Sat, Nov 28, 2015 at 3:59 AM, Kirill Katsnelson <notifications@github.com

kkm000:
Actually, I think it makes sense to do both. Even the latest gcc 4.9 does not enable C++11 features without the switch. AFAIK, clang and msvc are free of this idiosyncrasy.

danpovey:
I don't think it makes sense to activate flags for the test programs that
are not being used in the main program-- it's a mismatch.  And this isn't
really a good enough reason to modify the Makefiles.  You can just #ifdef
the test and test yourself that it runs correctly by running with c++0x in
your local environment- the test won't run by default.
Dan




On Sat, Nov 28, 2015 at 4:54 PM, Kirill Katsnelson <notifications@github.com

kkm000:
Well, I am frankly torn on this. I also do not entirely appreciate the idea of enabling the new language features even for the tests, as this opens a whole fresh can of worms. In addition, having to run the tests manually is a sure way to have it abandoned. On the other hand, we'll have enabled them on Travis in any case, so this may be of a lesser concern.

I think I should look more into the preprocessor + template solution not requiring C++11. Maybe I'll come up with something more elegant that what I had. It is not going to be typesafe and more prone to runtime errors in defining the test, but we should not really care, I believe, as long as the tests cases are defined correctly with the limitations in mind.

What do you think is better approach of the two?

And I am implicitly dismissing creating tests with wave files defined with just `char data[] = { 0x... };` or checking in binary files. That would be really ugly,

danpovey:
It's up to you-- it's fine with me if these particular tests get executed
only by you for the time being.  We can just run them again if we ever
change that code again.  I don't think it's super critical.
Dan


On Sat, Nov 28, 2015 at 8:46 PM, Kirill Katsnelson <notifications@github.com

kkm000:
SGTM, thanks. Ceteris paribus, I'll stick to the least ugly approach.

danpovey:
@kkm000, did you actually finish this?  I though this pull request was waiting on something from you, but I may be wrong.

danpovey:
@kkm000, are you still going to finish this pull request, or should I find someone else to take ownership?

kkm000:
@danpovey: I'll do it, it's on my list. Not the first priority. I think I should find some time in February.

danpovey:
Kirill, remember this is on your TODO list!

kkm000:
I do! I'll get to it soon.

danpovey:
@kkm000, any update on this?  Can I just ask someone else to look at this?

danpovey:
@kkm000, what do you think should be done about this?  Will you do it eventually?

giuliopaci:
Hi everybody, is there any update on this? Now that [latest release of ffmpeg](http://ffmpeg.org/download.html#release_3.1)  supports sphere files (with mulaw/Alaw/shorten encodings), being able to replace sph2pipe with ffmpeg without workarounds seems very appealing.

danpovey:
I think there isn't an update.  I'll try to think of someone to take it
over from Kirill, since he doesn't seem to have time.

However, this PR is orthogonal to the question of replacing sph2pipe with
ffmpeg.  I'd rather wait a while to do that, until the latest release of
ffmpeg has propagated to all the repositories of the standard distributions.

Dan

danpovey:
@LvHang, can you please look into this PR, resolve the conflicts and update it, check the code for obvious bugs, and try to test that it doesn't break reading normal wav files?  Kirill has been saying that he would write test code but I think we should accept that that isn't going to happen.

LvHang:
Ok, get it

Daniel Povey <notifications@github.com>于2017年4月13日 周四19:57写道：

kkm000:
Sorry. Let me get back on it right now. I'll give an update later today.

danpovey:
Thanks!

LvHang:
So， that means kkm000 will write the test code and I will fix the conflicts, right?

danpovey:
He may just finish the whole thing himself, but  if there isn't progress
within a day or two I may ask you to do do it yourself.

kkm000:
Continued in #1546, hence closing.

PR #335, opened 2015-11-17
Fix compilation issues with gcc4.4.7

francisr:
When I try to compile and run the tests with gcc4.4.7 on CentOS I get this error:
```
nnet-component-test.cc: In function ‘int main()’:  
nnet-component-test.cc:350: error: reference to ‘int32’ is ambiguous  
kaldi/tools/openfst/include/fst/types.h:30: error: candidates are: typedef int32_t int32  
../base/kaldi-types.h:47: error:                 typedef int32_t kaldi::int32
nnet-component-test.cc:350: error: reference to ‘int32’ is ambiguous
kaldi/tools/openfst/include/fst/types.h:30: error: candidates are: typedef int32_t int32
../base/kaldi-types.h:47: error:                 typedef int32_t kaldi::int32
nnet-component-test.cc:350: error: expected ‘;’ before ‘loop’
nnet-component-test.cc:350: error: ‘loop’ was not declared in this scope
```
This patch fixes it.

PR #340, opened 2015-11-18
Fixes to unit tests under Windows

kkm000:
Looks like I hit the same issue as the recent PR #335 (got a rebase conflict with it!), only with the  VS 2013 compiler. I think my fix is more correct (I avoid running tests twice when CUDA is not compiled), but I can drop this part if there any objections. The parenthesis in `posterior-test.cc` is a real deal though.


danpovey:
Thanks..
I prefer to keep those CUDA things as they were, a lot of things have that pattern and I'd prefer it to be consistent.  Running the tests twice is fine-- they are randomized anyway
Dan

kkm000:
@danpovey: Sure, killed other changes.

danpovey:
Thanks- merged.

Issue #341, opened 2015-11-18
Modification to NormalizeComponent


danpovey:
I need NormalizeComponent to be extended.. it relates to the 'chain' branch that I'm working on in my own repo.
Instead of rescaling so that the output rms value is 1, I'd like this to be configurable, so that you can set
target-rms=0.25
in the config (or whatever value you want).  Obviously this would have to be tested- just extending the function GenerateRandomComponentConfig() should be enough.
@pegahgh, could you please do this this afternoon?  I need it for something.
Dan

pegahgh:
Sure! I will modify the component!

PR #342, opened 2015-11-18
Modification to NormalizeComponent #341 by adding target_rms option

pegahgh:


src/nnet3/nnet-simple-component.cc
```diff
@@ -186,7 +187,111 @@ const BaseFloat NormalizeComponent::kNormFloor = pow(2.0, -66);
...
+  std::string tok; // TODO: remove back-compatibility code.
+  ReadToken(is, binary, &tok);
+
+  // read target_rms_ if it is available.
+  if (tok == "<TargetRms>") {
+    ReadBasicType(is, binary, &target_rms_);
+    ReadToken(is, binary, &tok);
+  }
+
+  if (tok == "<ValueSum>") {
```
danpovey: This branch of the if-statement can be deleted now.

src/nnet3/nnet-simple-component.cc
```diff
@@ -186,7 +187,111 @@ const BaseFloat NormalizeComponent::kNormFloor = pow(2.0, -66);
...
+  WriteBasicType(os, binary, count_);
+  WriteToken(os, binary, ostr_end.str());
+}
+
+std::string NormalizeComponent::Info() const {
+  std::stringstream stream;
+  KALDI_ASSERT(InputDim() == OutputDim());  // always the case
+  stream << Type() << ", dim=" << dim_ 
+         << ", target_rms=" << target_rms_;
+
```
danpovey: If most of this is just a duplication of NonlinearComponent::Info(), why don't you do
stream << NonlinearComponent::Info();
to get the common parts, and avoid code duplication?

src/nnet3/nnet-simple-component.cc
```diff
@@ -186,7 +187,111 @@ const BaseFloat NormalizeComponent::kNormFloor = pow(2.0, -66);
...
+    value_avg.Scale(1.0 / count_);
+    stream << ", value-avg=" << SummarizeVector(value_avg);
+    Vector<double> deriv_avg_dbl(deriv_sum_);
+    Vector<BaseFloat> deriv_avg(deriv_avg_dbl);
+    deriv_avg.Scale(1.0 / count_);
+    stream << ", deriv-avg=" << SummarizeVector(deriv_avg);
+  }
+  return stream.str();
+}
+// The output y_i = target_rms * x_i / rms(x), so rms(y) = target_rms,
```
danpovey: Pegah, I think it turns out I was mistaken about this function setting the rms to a certain value, it sets the vector norm to that value.  So this makes this comment wrong, and (sorry) it also makes the choice of variable name wrong.  It should be target-norm instead of target-rms.

danpovey:
Pegah, I directly merged this into my 'chain' branch and it'll get merged when I eventually merge that into trunk, so I'll close this pull request without merging.
(BTW I made some changes for efficiency, see /home/dpovey/kaldi-chain/src/nnet3/).

PR #348, opened 2015-11-20
modify add elements kernel

zhichaowang:
modify add elements with multiple blocks.

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -1121,25 +1121,10 @@ static void _cuda_comp_obj_deriv(MatrixElement<Real> *x, int s, const Real* z, M
 template<typename Real>
 __global__
 static void _cuda_matrix_add_elements(Real *data, MatrixDim dim, Real alpha, MatrixElement<Real>* x, int s) {
-  int i = threadIdx.x;
-  if (i >= s)
-    return;
-  int size = s / CU1DBLOCK; //the least size in a loop (later part)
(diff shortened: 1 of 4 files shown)


src/cudamatrix/cu-kernels.cu
```diff
@@ -1121,25 +1121,10 @@ static void _cuda_comp_obj_deriv(MatrixElement<Real> *x, int s, const Real* z, M
 template<typename Real>
 __global__
 static void _cuda_matrix_add_elements(Real *data, MatrixDim dim, Real alpha, MatrixElement<Real>* x, int s) {
```
danpovey: please use 'num_elements' as a variable name instead of 's'... it's clearer.

src/cudamatrix/cu-kernels.cu
```diff
@@ -1121,25 +1121,10 @@ static void _cuda_comp_obj_deriv(MatrixElement<Real> *x, int s, const Real* z, M
...
-    loop_start = threshold + i*size;
-    loop_end = threshold + (i+1)*size;
-  }
-  for(int j = loop_start; j < loop_end; j++) {
-    *(data + x[j].row * dim.stride + x[j].column) += alpha * x[j].weight;
-  }
+    int i = blockIdx.x * blockDim.x + threadIdx.x;
+    if (i >= s)
+        return; 
+    *(data + x[i].row * dim.stride + x[i].column) += alpha * x[i].weight;
```
danpovey: I know this was the same as the old code, but indexing, like
data[.........]
would be clearer.

PR #349, opened 2015-11-20
make_tdnn_configs.py: num_targets check and error message dealt with feat_dim instead of num_targets

jfainberg:


danpovey:
Thanks!

PR #354, opened 2015-11-22
WIP Music/Speech Discrimination and GMM-based VAD

david-ryan-snyder:
**Music/Speech Discrimination**
* Example is in egs/bn_music_speech 
* Currently works just at a segment level, to classify a segment as music or speech. The sid/music_id.sh script is similar to the sid/gender_id.sh.
* Uses MUSAN for training, Broadcast News (https://catalog.ldc.upenn.edu/LDC98S71) for testing
* Currently, the example does *not* demonstrate anything at the frame level

**compute-vad-from-frame-likes**
The proposed binary is a frame-level VAD that uses archives of frame-level log-likelihoods as input. The input most likely comes from fgmm-global-get-frame-likes. The binary takes as input 2 or more log-likelihood archives. The frame-level labels are based on the input index for which the frame is most likely. For example, suppose the input is as follows:

`compute-vad-from-frame-likes scp:noise_likes.scp scp:speech_likes.scp scp:music_likes.scp ark,t:vad.txt`

Then the output in vad.txt might be of the form:
`utt1 [ 0 0 0 0 0 0 1 1 1 1 1 1 ... 2 2 2 2 2 2 0 0 0 0 ]`

Where "0" corresponds to the noise_likes.scp, "1" to the speech_likes.scp, and "2" to the music_likes. However, if a map file is passed to the option --map with the lines
`0 0`
`1 1`
`2 0 `

Then the "2" would get mapped to "0," and the line in vad.txt would become: 
`utt1 [ 0 0 0 0 0 0 1 1 1 1 1 1 ... 0 0 0 0 0 0 0 0 0 0 ]`

Also, a similar table supports weighting the likelihoods by a log-prior. Again, the indices correspond to the order of the likelihoods in the input. E.g.,
`0 -1.386`
`1 -0.693`
`2 -1.386`

**merge-vads**
If you are using a nonspeech GMM that models noises but not silence, then it is better to rely on the energy-based VAD for silence detection. Therefore, it is necessary to combine the output of multiple frame-based VADs. The merge-vads binary combines *two* sets of VAD decisions. By default, the rule for combining two VADs is as follows. If either of the decisions is 0 (nonspeech) then the output decision is also 0. If both of the decisions are 1 (speech) then the output is 1. However, the default behavior can be modified by providing a table to the argument --map. For example:
` 0 0 0 `
` 0 1 0 `
` 1 0 0 `
` 1 1 1 `
` 0 2 2 `
` 1 2 2 `
Where the first column corresponds to a label in the first VAD input, the second column a label in the second VAD input, and the third is the desired output label. In this example, 2 might represent music, which may only be modeled in the second VAD input.


egs/bn_music_speech/v1/cmd.sh
```diff
@@ -0,0 +1,28 @@
...
+#export decodebig_cmd="queue.pl -q all.q@@blade -l ram_free=4G,mem_free=4G"
+#export cuda_cmd="queue.pl -q long.q@@pco203 -l gpu=1"
+#export cuda_cmd="queue.pl -q long.q@pcspeech-gpu"
+#export mkgraph_cmd="queue.pl -q all.q@@servers -l ram_free=4G,mem_free=4G"
+
+#c) run it locally...
+#export train_cmd=run.pl
+#export decode_cmd=run.pl
+export cuda_cmd=run.pl
+#export mkgraph_cmd=run.pl
```
danpovey: we don't use cuda_command any more so you can remove that-- any any other crud in comments.

egs/bn_music_speech/v1/conf/merge_vad.conf
```diff
@@ -0,0 +1,6 @@
+0 0 0
```
danpovey: I wouldn't call this merge_vad.conf '.conf' since it isn't the type of file that's read by the --config=xxx option to binaries.
If you could support comment lines in this, having a comment at the top of the file explaining what it means would be nice.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,186 @@
...
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "matrix/kaldi-matrix.h"
+#include "util/stl-utils.h"
+
+using namespace kaldi;
+using kaldi::int32;
+
+void PrepareMap(const std::string &map_rxfilename, int32 num_classes,
```
danpovey: Some documentation would be nice.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,186 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "matrix/kaldi-matrix.h"
+#include "util/stl-utils.h"
+
+using namespace kaldi;
```
danpovey: I think this is disallowed by the Google style guide.  Generally what I do is just put the functions being defined in namespace Kaldi.  Or you could have the using-directive inside the function if really needed.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,186 @@
...
+
+    po.Register("map", &map_rxfilename, "Table that defines the frame-level"
+    " labels.  For each row, the first field is the zero-based index of the"
+    " input likelihood archive and the second field is the associated"
+    " integer label.");
+
+    po.Register("priors", &priors_rxfilename, "Table that specifies"
+    " log-priors for each class. For each row, the first field is the"
+    " zero-based index of the input likelihood archive and the second"
+    " field is the log-prior.");
```
danpovey: It seems to me that you would always have a file like
0 something
1 something
2 something
and therefore the 0 1 and 2 are redundant.  Also, wouldn't it be easier to just specify something like
  --priors=0.1,0.8,0.1
i.e. a comma-separated vector (colon-separation is now deprecated), which you can parse using SplitStringToFloats to a vector of float?  This will avoid the hassle of a file.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,186 @@
...
+        int32 max_indx = 0;
+        BaseFloat max_post = likes[0](i) + priors[0];
+        for (int32 j = 0; j < num_classes; j++) {
+          BaseFloat other_post = likes[j](i) + priors[j];
+          if (other_post > max_post) {
+            max_indx = j;
+            max_post = other_post;
+          }
+        }
+        if (map.find(max_indx) == map.end())
```
danpovey: it's better to get into the habit of using the map<...>::const_iterator here, as in:
map<...>::const_iterator iter = map.find(max_indx);
if (iter == map.end()) {
...
} else {
   // use iter->second.
}

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,186 @@
...
+          KALDI_ERR << "Missing label " << max_indx  << " in map";
+        vad_result(i) = map[max_indx];
+      }
+      if (vad_result.Dim() > 0) {
+        vad_writer.Write(utt, vad_result);
+        num_done++;
+      }
```
danpovey: Is there a situation where vad_results.Dim() will be zero?  If so, do you need to increase num_err?

src/ivectorbin/merge-vads.cc
```diff
@@ -0,0 +1,131 @@
...
+}
+
+int main(int argc, char *argv[]) {
+  try {
+    const char *usage =
+      "This program merges two Vector archives of frame-level voice\n"
+      "activity decisions. By default, the program assumes that the\n"
+      "input vectors consist of floats that are 0.0 if a frame is judged\n"
+      "as nonspeech and 1.0 if it is considered speech. The default\n"
+      "produces an output Vector that is 1.0 if both inputs are 1.0\n"
```
danpovey: float vectors with this interpretation are referred to elsewhere as 'weights', as in 'ali-to-weights'.  You might want to use that terminology here, and refer to that program and whatever programs you actually use in conjunction with this program, in the "See also:" message.  The See also: is highly encouraged, and can contain related programs or programs that have a similar purpose.

danpovey: also you may want to put this program in the See also: of closely related programs (doesn't have to be symmetric).

egs/bn_music_speech/v1/run.sh
```diff
@@ -0,0 +1,88 @@
...
+
+sid/music_id.sh --cmd "$train_cmd" --nj 40 \
+  exp/full_ubm_music exp/full_ubm_speech \
+  data/bn exp/bn_music_speech
+sid/music_id.sh --cmd "$train_cmd" --nj 40 \
+  exp/full_ubm_noise exp/full_ubm_speech \
+  data/bn exp/bn_noise_speech
+
+printf "EER using GMMs trained on music and speech"
+compute-eer <(python local/print_scores.py exp/bn_music_speech/ratio)
```
danpovey: we generally just invoke python scripts directly, relying on the shebang to identify it as a python script.

src/ivectorbin/merge-vads.cc
david-ryan-snyder: I think the interpretation of these vectors is **not** the same as weights. This is more like the output of compute-vad (in that binary, floats are either 0.0 or 1.0, there is no logical interpretation of 0.5). Even though it is stored as a float, it is basically handled as an integer.

In order to make these binaries compatible with compute-vad, I'm also storing the VAD decisions of compute-vad-from-likes and merge-vads as archives of float Vectors. But now that I think about it, I don't really see any advantage to doing it this way. There's currently no case where we use a frame-level VAD where it makes sense to have nonintegral VAD decisions. Would you mind if we change compute-vad and select-voiced-frames to work with Int32Vectors instead? Of course, I would also modify the new VAD programs in the same way.

danpovey: I think it's better to leave them as floats, because then they are
compatible with the output of ali-to-weights (which can be configured to
output 0.0 and 1.0).
Dan


On Fri, Dec 4, 2015 at 12:34 PM, david-ryan-snyder <notifications@github.com

david-ryan-snyder: Do you mean post-to-weights or maybe ali-to-post? I'm not seeing a program by that name.

danpovey: yes you're right, ali-to-post  | post-to-weights.

david-ryan-snyder:
Hi Dan,

Thank you for reviewing the commit. I believe that I've addressed each of your comments.

Best,
David

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,212 @@
...
+    for (int32 i = 0; i < num_classes; i++)
+      priors->push_back(log(1.0/num_classes)); // Uniform priors
+  } else {
+    SplitStringToFloats(priors_str, ",", false, priors);
+    for (int32 i = 0; i < priors->size(); i++)
+      (*priors)[i] = log((*priors)[i]);
+  }
+
+  if (priors->size() != num_classes)
+    KALDI_ERR << "Prior table has " << priors->size() << " classes.  "
```
danpovey: Please reword the error message to reflect how the priors are now passed.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,212 @@
...
+  typedef kaldi::int32 int32;
+  try {
+    const char *usage =
+      "This program computes frame-level voice activity decisions from a\n"
+      "set of input frame-level log-likelihoods.  Frames are assigned\n"
+      "labels according to the class for which the log-likelihood\n"
+      "(optionally weighted by a prior) is maximal.  The class labels are\n"
+      "determined by the order of inputs on the command line.  See options\n"
+      "for more details.\n"
+      "See also: fgmm-global-get-frame-likes, compute-vad, merge-vads\n"
```
danpovey: "see also" is typically after the "e.g.:" line, i.e. right at the bottom.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,212 @@
...
+      "\n"
+      "Usage: compute-vad-from-frame-likes [options] <likes-rspecifier-1>\n"
+      "    ... <likes-rspecifier-n> <vad-wspecifier>\n"
+      "e.g.: compute-vad-from-frame-likes --map=label_map.txt\n"
+      "    scp:likes1.scp scp:likes2.scp ark:vad.ark\n";
```
danpovey: It might be a good idea to mention explicitly that fgmm-global-get-frame-likes might be where you'd get the frame likelihoods from.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,212 @@
...
+
+    ParseOptions po(usage);
+    std::string map_rxfilename;
+    std::string priors_str;
+
+    po.Register("map", &map_rxfilename, "Table that defines the frame-level"
```
danpovey: Odd indentation, normally the "'s would be aligned.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,212 @@
...
+        continue;
+      }
+      for (int32 i = 0; i < num_classes - 1; i++) {
+        if (!readers[i]->HasKey(utt)) {
+          KALDI_WARN << "No vector for utterance " << utt;
+          num_err++;
+          continue;
+        }
+        Vector<BaseFloat> other_like(readers[i]->Value(utt));
+        if (like_dim != other_like.Dim()) {
```
danpovey: Generally we'd print a warning and increase num_err if something like this happens.

david-ryan-snyder:
OK, those comments have been addressed.

src/ivectorbin/compute-vad-from-frame-likes.cc
```diff
@@ -0,0 +1,215 @@
...
+    std::vector<RandomAccessBaseFloatVectorReader *> readers;
+    std::string vad_wspecifier = po.GetArg(po.NumArgs());
+    BaseFloatVectorWriter vad_writer(vad_wspecifier);
+
+    for (int32 i = 2; i < po.NumArgs(); i++) {
+      RandomAccessBaseFloatVectorReader *reader
+        = new RandomAccessBaseFloatVectorReader(po.GetArg(i));
+      readers.push_back(reader);
+    }
+
```
danpovey: This can be done later, but I think it would be quite easy here to make it a Viterbi computation which would avoid very short runs of one or another class.  You could just specify a penalty for switching, and interpret that a log-probability for transitioning between different classes.  The computation would not be that complicated to code.  But it won't change the interfaces, so this can be done later.

src/ivectorbin/merge-vads.cc
```diff
@@ -0,0 +1,163 @@
...
+      Vector<BaseFloat> vad2(second_vad_reader.Value(utt));
+      if (vad1.Dim() != vad2.Dim()) {
+        KALDI_WARN << "VAD length mismatch for utterance " << utt;
+        num_err++;
+        continue;
+      }
+      Vector<BaseFloat> vad_result(vad1.Dim());
+      for (int32 i = 0; i < vad1.Dim(); i++) {
+        std::pair<int32, int32> key(static_cast<int32>(vad1(i)),
+          static_cast<int32>(vad2(i)));
```
danpovey: it might make sense to add a warning if you detect non-integer inputs. can be done later though.

danpovey:
OK, merging.

engrfarazscut:
i am facing this problem.plz if someone can help me....
screenshot from 2019-01-10 11-55-00

david-ryan-snyder:
This error seems unrelated to this pull request, and it probably should've be asked on Kaldi help (http://kaldi-asr.org/forums.html), not here.

Anyway, the error message indicates that you gave the wrong input to compute-eer. If you run the binary without any arguments, it will print a usage message telling you what the input should look like.

engrfarazscut:
ok thanks.....

PR #363, opened 2015-11-25
fix to const-arpa-lm code to be able to handle larger models.

danpovey:


kkm000:
LGTM

danpovey:
Thanks...
actually I want to hold off on this, because I noticed another problem - the ConstArpaLm has an unnecessarily big on-disk format-- and we can fix it at the same time.  Guoguo will work on this.
Dan

PR #367, opened 2015-11-28
WIP: documented ConstArpaLm stuff in const-arpa-lm.h; fixed I/O bug (…

chenguoguo:
…was using 25% more of what's needed on the disk); changed lm_states_size_ to int64

I've done some test with old/new models (including valgrind). Will ask Andre to test as well.


danpovey:
Thanks!  I'll wait till we hear from Andre, as this isn't super urgent.

src/lm/const-arpa-lm.cc
```diff
@@ -697,16 +712,100 @@ void ConstArpaLm::Read(std::istream &is, bool binary) {
...
+  KALDI_ASSERT(bos_symbol_ < num_words_ && bos_symbol_ > 0);
+  KALDI_ASSERT(eos_symbol_ < num_words_ && eos_symbol_ > 0);
+  KALDI_ASSERT(unk_symbol_ < num_words_ &&
+               (unk_symbol_ > 0 || unk_symbol_ == -1));
+  lm_states_end_ = lm_states_ + lm_states_size_ - 1;
+  memory_assigned_ = true;
+  initialized_ = true;
+}
+
+void ConstArpaLm::ReadInternalDeprecated(std::istream &is, bool binary) {
```
danpovey: maybe call this ReadInternalOldFormat() instead.

src/base/io-funcs.h
```diff
@@ -182,6 +182,17 @@ template<class T> inline void WriteIntegerVector(std::ostream &os, bool binary,
 template<class T> inline void ReadIntegerVector(std::istream &is, bool binary,
                                                 std::vector<T> *v);
 
+/// Function for writing array of integer types.
```
danpovey: Please just put these templates in const-arpa-lm.cc for now until we need them for something else later.  They are confusingly similar to ReadIntegerVector and WriteIntegerVector and I won't want people to assume these functions are the first choice for writing this type of thing.

src/lm/const-arpa-lm.h
```diff
@@ -29,6 +29,172 @@
...
+       each LmState).
+    4. In ConstArpaLmBuilder, create a memory block for all the LmStates (after
+       sorting and updating the address). This includes <lm_state_> that stores
+       all the LmStates in an int32 array, <unigram_states_> that keeps the
+       address of unigram LmStates, <overflow_buffer_> that keeps the address
+       of LmState whose address differs too much from the parent address. See
+       above how we handle the leaf case.
+    5. With the information in step 4, create the class ConstArpaLm.
+*/
+
```
danpovey: Thanks for the documentation, it looks very good.

chenguoguo:
I updated the code according to the comments.

src/lm/const-arpa-lm.cc
```diff
@@ -29,6 +29,110 @@
 
 namespace kaldi {
 
+// Function for writing array of integer types.
+template<class T> inline void WriteIntegerArray(std::ostream &os, bool binary,
```
danpovey: It occurred to me that this function is not really needed.  Just do the equivalent of
os.write(reinterpret_cast<const char *>(array),      sizeof(T) * array_sz);
and something similar for reading.
of course check os.good() afterward.

danpovey:
Thanks.  Have you tested that it works?

chenguoguo:
Yes, I've tested with my models. We can ask Andre to run it again just in
case.

Guoguo

danpovey:
It's OK, I'll merge.

PR #369, opened 2015-11-29
fix a bug where utterance_id is missing for Mikolov's rnnlm rescoring…

hainan-xv:
…; fix a bug in rnnlmrescore.sh where old lmscore is not removed under certain conditions


danpovey:
Thanks!

PR #370, opened 2015-11-30
Partial lattice from lattice-align-words-lexicon not sorted.

francisr:


danpovey:
In general there is not supposed to be a guarantee that lattices output from Kaldi programs are topologically sorted- if needed we normally sort them after reading them.  But I'll merge anyway, since it would be hard to do the sorting in that Perl script.

PR #377, opened 2015-12-03
Adding the data combining script and program.

tomkocse:
The processing speed of the script steps/cleanup/combine_short_segments.sh is very slow.
The for loop that keeps calling utils/filter_scp.pl for each speaker is one of the parts.
The whole script is done at bash level. It will be appreciated if anyone can advise how to rewrite to speed up the script.

### DIFF
--- a/egs/wsj/s5/steps/cleanup/combine_short_segments.sh
+++ b/egs/wsj/s5/steps/cleanup/combine_short_segments.sh
@@ -7,7 +7,7 @@
 frame_shift=10
 # End configuration section.
 
-#echo "$0 $@"
+echo "$0 $@"
 
 [ -f ./path.sh ] && . ./path.sh
 . parse_options.sh || exit 1;
@@ -25,6 +25,7 @@ if [ $# -ne 3 ]; then
   exit 1;
 fi
 
+#This script works in the unit of frames
 frame_per_sec=$(( 1000/$frame_shift ))
 min_seg_len=$(echo $1*$frame_per_sec | bc)
 input_dir=$2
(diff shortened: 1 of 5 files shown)


egs/wsj/s5/steps/cleanup/combine_short_segments.sh
```diff
@@ -0,0 +1,165 @@
...
+
+echo "$0 $@"
+
+[ -f ./path.sh ] && . ./path.sh
+. parse_options.sh || exit 1;
+
+if [ $# -ne 3 ]; then
+  echo "This script concatenates segments with length shorter than the specified length in seconds."
+  echo ""
+  echo "Usage: steps/cleanup/combine_short_segments.sh [options] <minimum-length> <data-in> <data-out>"
```
vijayaditya: Please change  <code> <data-in> <data-out></code>  to <code> <input-data-dir> <output-data-dir></code>

egs/wsj/s5/steps/cleanup/combine_short_segments.sh
```diff
@@ -0,0 +1,165 @@
...
+    $text = join(" ", @col[1..@col-1]);
+    $utt2text{$col[0]} = $text;
+  }
+
+  foreach $seg (sort keys %uttlist) {
+    if ($utt2len{$seg} < $min_seg_len) {
+      $localmin = inf;
+      $target = $seg;
+      @utts = split(" ",$spk2utt{$utt2spk{$seg}});
+      foreach $seg2 (@utts) {
```
vijayaditya: I don't think it is good to combine non-contiguous segments. This will affect things like ivector estimation. Ensure that the seg2 is next to seg. Keep combining segments until you reach minimum length.

egs/wsj/s5/steps/cleanup/combine_short_segments.sh
```diff
@@ -0,0 +1,165 @@
...
+    $uttlist{$utt_id} = $utt_id;
+    $utt2feat{$utt_id} = $feat;
+  }
+  while (<LI>) {
+    chomp;
+    @col = split;
+    @col == 2 || die "Error: bad line $_\n";
+    ($utt_id, $len) = @col;
+    $utt2len{$utt_id} = $len;
+    $utt2item{$utt_id} = 1;
```
vijayaditya: Please describe in comments what is the use of utt2item

egs/wsj/s5/steps/cleanup/combine_short_segments.sh
```diff
@@ -0,0 +1,165 @@
...
+    } elsif ($utt2item{$seg} == 1) {
+      print UO "$uttlist{$seg} $utt2spk{$seg}\n";
+      print FO "$uttlist{$seg} $utt2feat{$seg}\n" ;
+      print TO "$uttlist{$seg} $utt2text{$seg}\n";
+    }
+  }
+
+' $input_dir/utt2spk \
+$input_dir/spk2utt \
+$input_dir/feats.scp \
```
vijayaditya: reorder the arguments to ensure that all the inputs are before the outputs

egs/wsj/s5/steps/cleanup/combine_short_segments.sh
danpovey: OK- originally I had in mind that it's OK to combine non-contiguous segments, but I see your point about the iVectors, so fair enough.

src/featbin/concat-feats.cc
```diff
@@ -0,0 +1,96 @@
...
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    using namespace std;
+    
+    const char *usage =
+        "Concatenate feature files (assuming they have the same dimensions)\n"
+        "Usage: concat-feats <in-rxfilename1> <in-rxfilename2> [<in-rxfilename3> ...] <out-wxfilename>\n"
+        " e.g. concat-feats mfcc/foo.ark:12343 mfcc/foo.ark:56789 -\n"
+        "See also: copy-feats, copy-matrix, append-vector-to-feats, paste-feats\n";
```
danpovey: Please make sure that concat-feats appears as "See also:" in the usage message of paste-feats and copy-feats.

egs/wsj/s5/steps/cleanup/combine_short_segments.sh
```diff
@@ -0,0 +1,191 @@
...
+  [ ! -f $input_dir/$f ] && echo "$0: no such file $input_dir/$f" && exit 1;
+done
+
+export LC_ALL=C
+
+# This is the main function, to look for segments with length shorter than the specified length
+# and concatenates them with other segments to make sure every combined segments are 
+# with enough length. Here, after the search of the under-length segment, it looks for
+# another segment where the combined length is the closest to the specified length.
+
```
danpovey: I don't think it makes sense to have a function that's only called once.  In C/C++ it would be good coding practice, but it's not really traditional in bash.

egs/wsj/s5/steps/cleanup/combine_short_segments.sh
```diff
@@ -0,0 +1,191 @@
...
+if [ -f $input_dir/segments ]; then
+  awk -v l=$frame_per_sec '{
+    len=($4 - $3) * l;
+    if (len > 0) {
+      printf("%s %d \n", $1, len);
+    }
+  }' $input_dir/segments >$output_dir/feats.length
+else
+  feat-to-len --print-args=false scp:$input_dir/feats.scp ark,t:$output_dir/feats.length
```
danpovey: feat-to-len outputs the length in frames.  You'll have to convert to the length in seconds.
You should probably have a command-line option called frame_shift, default 0.01, to make it possible to correctly handle features with non-default frame shift.

tomkocse: I suppose to work out the script in the length of frames, so i convert the length in seconds to the length in frames at the beginning of the script. The  frame_shift option is used equivalently.

danpovey: oh I see.

tomkocse: Anyway i change the script to work in the length in seconds now.

danpovey:
Have you tested the script with both segments and no-segments input?
Dan

tomkocse:
Yes, i tested in both cases.
Just please note that the length of the utterances will be 0.02 sec (2 frames) less when they are computed from feat-to-len ( when no segments input) due to the sliding window in feature extraction.
So the result will be slightly different.

danpovey:
Merging- thanks.

PR #380, opened 2015-12-04
WIP: perturbed training that might help LSTM/cwRNN training (experimental codes. don't merge yet)

xiaohui-zhang:


egs/wsj/s5/steps/nnet3/lstm/train_perturbed.sh
```diff
@@ -0,0 +1,732 @@
+#!/bin/bash
```
vijayaditya: It would be better to just make perturbed training an option in the original lstm script.

PR #382, opened 2015-12-06
Fix a lattice determinization bug.

vdp:
I'm not that familiar with this code yet, but this seems like an obvious omission..
I've only performed a toy test on 2 utterances, and the (non-compressed) lattice sizes are as follows:
>   old lat.ark: 323482
>   new lat.ark: 209510

The run time for a debug(-O0 -g) version of the nnet3-latgen-faster didn't change noticeably- there was a 2-3% difference, but that could be due to e.g. slightly different load on my laptop, while testing.

### DIFF
--- a/src/lat/determinize-lattice-pruned.cc
+++ b/src/lat/determinize-lattice-pruned.cc
@@ -538,6 +538,7 @@ template<class Weight, class IntType> class LatticeDeterminizerPruned {
                    << forward_cost << ", "
                    << state.forward_cost;
       }
+      return state_id;
     }
     OutputStateId state_id = static_cast<OutputStateId>(output_states_.size());
     OutputState *new_state = new OutputState(subset, forward_cost);


vdp:
By the way I've also tried lattice rescoring before and after the fix and the results seem to match.

danpovey:
Agreed, it looks like that was a bug, leading to
not-as-minimal-as-they-could-be lattices.  It could have been there for a
long time, and it may exist in other determinization code- search for
functions of the same name elsewhere.  I'll merge.

Dan

vdp:
The only other method with the same name belongs to the LatticeDeterminizer class(in fstext/determinize-lattice-inl.h), but there the cached "minimal" state IDs are handled correctly. I've also searched for (sub)strings from the comment and the warning message in the previously buggy "if" block, to see if it was copy/pasted to/from somewhere else, but this doesn't appear to be the case.

By the way, the previously mentioned rescoring was done using the following commands:

> lattice-scale --acoustic-scale=-1 --lm-scale=-1 ark:"$lat" ark:- | \
>  lattice-compose ark:- "fstproject --project_output=true $oldG |" ark:- | \
>  lattice-determinize ark:- ark:- | \
>  lattice-scale --acoustic-scale=-1 --lm-scale=-1 ark:- ark:- | \
>  lattice-compose --phi-label=$phi ark:- "$newG" ark:- | \
>  lattice-determinize ark:- ark:$outdir/lat.ark || exit 1;

a.k.a. "mode 3" in lmrescore.sh.
I noticed that the resulting lattice for the old, non-optimal, input is only slightly bigger than the result for the smaller input. I guess maybe not really surprising, because that may have been the case whether lattice-determinize has a bug(de-minimizing the good input) or not(somehow minimizing the bad input?)[I'm not really sure though, because I haven't looked yet how all these tools are implemented]. 
At least that's what I thought. However, today I decided to run the rescoring again but with both $oldG and $newG set to the pruned trigram, used during decoding, and I find the result somewhat baffling. Sizes of the output(i.e. rescored) lattices in this case are:

> for old(non-optimal) input: 339182
> for new input: 315334

i.e. they are closer to the size of the "buggy" input than the optimal one. Is this normal? lattice-determinize is using LatticeDeterminizer class to do the determinization, and we've already determined it doesn't have this bug. Is this due to e.g. not pruning?

danpovey:
OK, I'm not 100% sure that I understand all this, but whatever it is, I
wouldn't worry about it too much.
BTW, lattice-determinize has a --minimize option that should, if set to
true, mean that the bug has no effect, since the redundant states would be
combined later.
Dan

vdp:
Well, I guess the TL;DR of my overly lengthy post(sorry, didn't have the time to formulate it properly yesterday) is that I wonder why the size of the lattice seem to "explode" during rescoring, especially given that we are rescoring with the same G we use during decoding. Is this a sign that there are other "bugs" in lattice-determinize that cause similar not-as-minimal-as-it-could-be output, or is it normal? 
For example, I notice, that the subset hash function (class SubsetKey) is position dependent, and in the same time methods like EpsilonClosure() can return subset vectors in arbitrary order(I think), due to iteration over unordered_map. I didn't actually analyse in detail how the vectors are processed after that, and whether their order is "normalized" somewhere along the pipeline, but just for the heck of it I added a std::sort(subset) call as a last statement in EpsilonClosure(). This seems to reduce the size of the rescored lattices by 3-5%, at least for my toy test.
Anyways, this is not something I'd like to spend much time at the moment either- leaving this note just as clarification and possibly for future reference..

danpovey:
Oh.
The EpsilonClosure() is supposed to return subset vectors in sorted order.
I think what happened is that  when @hainan-xv fixed a different issue in
EpsilonClosure, he must have forgotten to sort them.  The tests didn't
catch this because the result is still deterministic, just not as minimal
is it could be (and there's no natural test for that).

You'll see a comment at the end of EpsilonClosure()
   // copy cur_subset to subset.


      // sorted order is automatic.
and this was true when the map type was std::map, but when it was changed
to unordered_map it's no longer true.  Could you please submit a pull
request for this issue, and clean up the comment mismatch too?

Dan

PR #385, opened 2015-12-08
WIP: Confidence calibration (ami, ihm)

vesis84:
Adding example of confidence calibration (ami, ihm). 
The proposed implementation is not deeply integrated into the existing code of scoring,
everything is created in a new directory 'exp/tri4a_mmi_b0.1', including 2 runs of scoring (dev, eval).
The present form is intended to be used in a 'copy and modify' fashion to adapt it to the new databases. Certainly, the comments are welcome...


egs/ami/s5/local/calibrate_ctm.sh
```diff
@@ -0,0 +1,101 @@
...
+# Prepare training data for calibration model,
+utils/scoring/prepare_calibration_data.py \
+  --conf-targets $caldir/dev_train_targets.ark --conf-feats $caldir/dev_train_feats.ark \
+  --segments $dev/segments --reco2file-and-channel $dev/reco2file_and_channel \
+  $caldir/dev.ctm $dev_latdepth $arpa_gz $lang/words.txt
+
+# Train the calibration model,
+logistic-regression-train --binary=false ark:$caldir/dev_train_feats.ark \
+  ark:$caldir/dev_train_targets.ark $caldir/calibration.mdl
+
```
danpovey: I wonder if a 'default' model could be stored in text form in a configuration-type file somewhere, for use when there no calibration data.  This could just be taken from random setup, such as this one.
That would make this setup usable even when there is only one test set.

danpovey:
Karel,
I notice that the setup has what I regard as an unpleasantly deep dependency on the sclite scoring software.  Remember that sclite is not a part of Kaldi-- we only rely on it for scoring when the traditional testing setup requires it.  It would be better if it could be done in a different way so as not to make use of sclite until the actual scoring phase.  Obviously this would be more work.  Would you have time to rework it in this way?  If not I'll have to think what to do (i.e. accept this for now, or ask someone to try to get a cleaner setup).
Dan

danpovey:
Karel, let us know what it is you need that you are getting from sclite- maybe there is an alternative way to get the same thing.

vesis84:
Hi Dan, commenting on the two ideas.
- default calibration model : for this I'd probably remove some of the input features for the logistic regression, these might behave differently on different langauges. Maybe I'd leave there just the logit of the MBR posterior, but it should be tested how well it generalizes.
- replacing sclite : what is needed is per-word alignment to the ground truth. This can be either obtained in 'prf' from sclite, or it is possible to use yenda's 'align-text'. The alternative pipeline could be like this: a) create ctm with confidences with segment-based timing, b) convert it to 'tra' c) align with 'text' d) augment the ctm with the alignment. The issue is, for some databases we have 'stm' reference, while for other we have 'text' reference. So AFAIK there should be 2 scripts to support both.
- yes, I can create an example of that, it would be easier for me to do it. What is the preferred database? The 'wsj'? Or something with higher WER?
Best!
K.

vesis84:
In the kaldi based scoring that is based on 'text' file, the hypothesis is always stored without the confidence. Currently the confidences were present only in the 'sclite' scoring, which is the reason why it was implemented that way...

danpovey:
Regarding default calibration model- actually I think maybe we can forget
about this.  Once we have a calibration mechanism that can accept 'text'
files, users can just use a subset of training data.

Regarding needing two different pipelines for stm versus 'text'
supervision: can't we just convert stm to text?  If not, we'll always have
available the option to hold out a subset of training data- or use the
training data itself; it wont be ideal but it would give us *something*.
I'd rather just have a single pipeline if at all possible.

I didn't understand all of the details of what you were saying, but what I
think would be best would be a pipeline without any dependencies at all on
sclite.  It's a big package to have as a requirement, and very hard to
debug (and not always easy to install).

Regarding what setup to use an example-- WSJ is fine with me; in fact it's
preferred, as it's the place where most new features get demonstrated first.

Dan

vesis84:
Okay, we can convert the 'stm' to 'text', but by doing this we also have to import the VAD from the 'stm'. And I usually use my own VAD (at least in Babel) that is different from the 'stm' VAD. In Babel I have seen 'stm' files where segments contained lots of silences and some noise, which was increasing the insertion errors. 

Calibrating on the training data is also not ideal, the held out 'dev' like on switchboard is a good solution.

And as I have said, until now the per-word confidences were present only in the 'ctm' files created by the 'lattice-to-ctm-conf', and these were always scored by sclite. If we want to support per-word confidences in kaldi-scoring, we need to modify the scoring scipts and add intermediate step lattice->*ctm*->tra.

I'll try this concept with the switchboard 'dev' held-out from the training...
K.

danpovey:
>
> Okay, we can convert the 'stm' to 'text', but by doing this we also have
> to import the VAD from the 'stm'. And I usually use my own VAD (at least in
> Babel) that is different from the 'stm' VAD. In Babel I have seen 'stm'
> files where segments contained lots of silences and some noise, which was
> increasing the insertion errors.
>
If this is only going to be used for the calibration, it probably wouldn't
make much of a difference.

> And as I have said, until now the per-word confidences were present only
> in the 'ctm' files created by the 'lattice-to-ctm-conf', and these were
> always scored by sclite. If we want to support per-word confidences in
> kaldi-scoring, we need to modify the scoring scipts and add intermediate
> step lattice->*ctm*->tra.
>
I'd say for now we don't have to modify the Kaldi scoring pipeline at all.
We can still score as normal from the lattice output -- at least for now.
If we want a mechanism, in the future, to somehow score the confidences
within the Kaldi scoring pipeline, we can introduce that.  What we are
introducing right now is a separate optional pathway by which you can get
per-word confidences- it would be a separate script, independent of the
scoring pipeline.  We can produce these in ctm format [although
lattice-alignment should probably be optional, as for many purposes we
won't care about the times.].  An option to produce per-frame confidences
might be nice too, if that could be done, as those are useful for various
other purposes.  People sometimes want per-word confidences for various
reasons, including downstream processing in a bigger task, or for
unsupervised training.  Typically for these purposes, integrating too
tightly with the scoring pipeline will be undesirable.

It might be necessary to reorganize the sclite scoring scripts slightly
(or, better, produce an alternative version of them) to be able to work
directly from the ctm produced by your confidence mechanism.  I think it
would be most ideal to have your confidence-producing script write out the
'intermediate' ctm's, i.e. the ones that are relative to the segments,
rather than the ones converted by reco2file_and_channel, since the
'intermediate' ctm's are the ones that are most natural to process within
Kaldi (since they are relative to the utterances themselves), and these
will be more useful for other purposes, such as unsupervised training.

If the time comes when we want to be able to evaluate how good the per-word
confidences are in a setup where there is no sclite, we could modify the
Kaldi scoring pipeline to accept confidences somehow... this will require
some thought, and I'd rather not have to decide this right now.

Dan

I'll try this concept with the switchboard 'dev' held-out from the

vesis84:
That is true, but having another segmentation means also another feature extraction and lattice generation, which may be an unnecessary complication, id what we want to do is only to refine the confidence scores... Now I am getting confused, what should be the next step to modify this pull request?

vesis84:
Should I do the parsing from the yenda's 'align-text' as an alternative to parsing the 'sclite' output? Or is it fine to use the sclite output as it is now?

danpovey:
Actually I don't fully understand all the issues involved.
Basically what I want is a pipeline that, if needed can be run from start
to finish without any involvement of sclite tools-- to the point where you
have the confidences.  And then we can use sclite, if we want, to see how
good the confidences are.
I'd say, if you are able to do it (and I would really appreciate it), do
something of this nature in the way that you feel is best, and then we can
look at the details.  There is no point pinning it down further at this
stage as there are many decisions you'll have to make.

Regarding 'align-text'-- parsing that would be fine, but if it's more
convenient to write a new tool in C++ (e.g.  one that would read the
reference text and lattices), that could also be a reasonable option.  It
doesn't have to be all scripts.
This could be an important feature.  It's something that many people have
requested, as it's quite useful in applications.

danpovey:
... actually, from a certain point of view, having the central parts of
this in code (at least, the bit required in test-time for a particular
utterance) would be more convenient.
Imagine that someone is building an IVR application and they want to get
the confidences in C++.  This is a very plausible scenario.  Of course to
make it configurable so you can use the features that you want, a config
class would be needed.
Expect there to be a little bit of back and forth on this.

vesis84:
Hi, I have changed the code, so that 'sclite' is not used in the confidence calibration code. There is an example for 'swbd,ami,tedlium'. For AMI, the NCE is 0.002 worse than when 'sclite' was used. This is probably because of the 'glm' mapping, which allows to evaluate some as correct, even if they are different from the reference. This changes the supervision of the calibration, but the difference is quite small...
@Dan: I think it is ready for a next round of reivew.

egs/wsj/s5/utils/scoring/append_prf_to_ctm.py
```diff
@@ -0,0 +1,74 @@
+#!/bin/env python
+
+# Copyright 2015  Brno University of Technology (author: Karel Vesely)
+# Apache 2.0
+
+import sys
+import numpy as np
+
```
danpovey: Karel, you won't like to hear this, but the numpy dependency is problematic.
numpy is not very easy to install, and tends to cause version-mismatch problems.  And it seems unnecessary here as you're only using it for parsing, not for any mathematical stuff.
I'd be OK with keeping the numpy dependency only if we downgrade this to a setup-specific script in 'local'/.

egs/wsj/s5/utils/scoring/prepare_calibration_data.py
```diff
@@ -0,0 +1,151 @@
+#!/bin/env python
```
danpovey: I'm not crazy about the location as utils/scoring/.  The way I'm thinking of this confidence calibration is as a utility that doesn't intrinsically relate to scoring, while 'scoring' refers specifically to the process of getting scores from output (not of producing that output).

egs/tedlium/s5/local/confidence_calibration.sh
```diff
@@ -0,0 +1,55 @@
...
+
+. utils/parse_options.sh
+set -euxo pipefail
+
+# Derived options,
+dev_caldir=$dev_latdir/confidence_$lmwt
+eval_caldir=$eval_latdir/confidence_$lmwt
+
+###### Train the calibration,
+steps/conf_train_calibration.sh --cmd "$decode_cmd" --lmwt $lmwt \
```
danpovey: There should probably be a subdirectory here, e.g. steps/conf/.
I am moving towards using subdirectories, as steps/ is getting over-full.
Within sub-directories, it's OK to put in steps/, scripts that would traditionally have gone in utils/ (i.e. scripts that are utility-style scripts that don't take the "$cmd" option).

egs/tedlium/s5/local/confidence_calibration.sh
```diff
@@ -0,0 +1,55 @@
...
+# Scoring tools,
+hubscr=$KALDI_ROOT/tools/sctk/bin/hubscr.pl 
+hubdir=`dirname $hubscr`
+
+# Inputs,
+ctm=$eval_caldir/ctm_calibrated
+stm=$eval_data/stm
+glm=$eval_data/glm
+
+# Normalizng CTM, just like in 'local/score_sclite.sh',
```
danpovey: Karel, all this normalization-- is it 100% necessary in order to get good results?  This applies to that 'grep-pattern' thing too.  It introduces what I see as a too-tight connection between the confidence calibration and the scoring.
It seems to me that a more general method would be to take the N most common words in the calibration data and give them their own features.

egs/wsj/s5/utils/scoring/prepare_calibration_data.py
```diff
@@ -0,0 +1,151 @@
...
+  # Load the depths (ASCII), fill the buffer with the depths,
+  with open(depths_file,'r') as f:
+    for l in f:
+      utt,d = l.split(' ',1)
+      d = np.array(d.split(),dtype='int')
+      frame_begin = int(np.rint(100*segments[utt][1]))
+      depths[reco2file_and_channel[segments[utt][0]]][frame_begin:frame_begin+len(d)] = d
+
+# Load the unigram probabilities in 10log from ARPA,
+import gzip, re
```
danpovey: In order to make it more general, could we let the input to this script just be a file containing the unigram probabilities in some very simple format, and put on the caller the responsibility to get this from the arpa?  That way it will be more adaptable to non-arpa-format language models.

egs/tedlium/s5/local/confidence_calibration.sh
vesis84: Hi, in tedlium the 'glm' is a dummy file, but I believe the scoring should be the same as in local/score.sh, so we can compare the scores with normal setup. The 'grep-pattern' is now removed, but I keep filtering words like '[...],<...>,%...,...-' because these are never used for normal words and it was improving the NCE on swbd by 0.004.

vesis84:
Next revision, where Dan's suggestions are implemented, except to rewrite everything without numpy. Numpy is used in : steps/conf/append_eval_to_ctm.py, steps/conf/convert_ctm_to_tra.py and steps/conf/prepare_calibration_data.py. 
Especially this last script would be difficult to rewrite without numpy. I believe for most linux distributions, numpy is available as a packacge. Or another option for individual use is the Enthought Canopy. Since we started using it at BUT, there are no problems with python anymore...

egs/wsj/s5/steps/conf/train_calibration.sh
```diff
@@ -0,0 +1,122 @@
...
+    --conf-targets $dir/train_targets.ark --conf-feats $dir/train_feats.ark \
+    $dir/ctm_aligned $latdepth $unigrams $dir/word_categories
+fi
+
+# Filter the targets,
+if [ $stage -le 4 ]; then
+  # Removing words with format: [...], <...>, %..., ...-,
+  grep -i -v -E "\^(\[.*\]|<.*>|%[^,]*|[^,]*-)," $dir/train_targets.ark >$dir/train_targets_filt.ark
+fi
+
```
danpovey: Kare, it's problematic to do filtering of this type in anything but a script in local/.  This makes sense for English perhaps, given normal transcription conventions, but what about, say, Buckwalterized Arabic?  Some of these characters are a valid letter in that encoding.  All scripts are supposed to be language-independent and in general we don't do anything that you couldn't do in principle from the integerized form of the file, in any of the generic scripts.   In fact, using anything other than the integer identifiers in generic scripts  (except for scoring) is somewhere between highly discouraged to forbidden, because with encoding issues and locales you can run into a lot of unexpected issues.  I wonder if it would be possible to move the parts of this that need to look at words.txt to an outer level, so you'd have the user supply a file of the form

word-identifier1  feature1 feature2 .. featureN  
word-identifier2  feature1 feature2 .. featureN

and this way the script itself would never have to either read the actual words, or make language-specific assumptions?  This way you could add a feature for partial-words and the like, in a way that makes sense for the corpus, and add a feature for the length (and you could decide yourself whether you want to count the bytes or the Unicode characters), and so on.  This would also mean less work to do in prepare_calibration_data.py.

egs/wsj/s5/utils/lattice_depth_per_frame.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# Copyright 2015  Brno University of Technology (Author: Karel Vesely)
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+# Extract lattice-depth for each frame.
```
danpovey: this script lattice_depth_per_frame.sh should probably be in steps/conf/ unless and until we need it for other purposes.  It's not much more than a wrapper for the C++ program-- and in fact for something as simple as this, unless you use it a lot I'd tend to just include the lines of script directly rather than invoking a separate script.

danpovey:
> Next revision, where Dan's suggestions are implemented, except to rewrite
> everything without numpy. Numpy is used in :
> steps/conf/append_eval_to_ctm.py, steps/conf/convert_ctm_to_tra.py and
> steps/conf/prepare_calibration_data.py.
> Especially this last script would be difficult to rewrite without numpy. I
> believe for most linux distributions, numpy is available as a packacge. Or
> another option for individual use is the Enthought Canopy. Since we started
> using it at BUT, there are no problems with python anymore...
>

Karel, I'm sorry but I'm not going to accept the dependency on numpy.
You're right that numpy is available as a package but it's a very big
package, and it will increase the install time for Kaldi on systems that
didn't previously have it.  It also tends to have its own dependencies, and
if people are trying to play around with their BLAS installation to improve
the speed, there is a lot of potential for bad interactions with the numpy
installation.  If you don't want to do the work to remove the dependency, I
can always find someone else to do it after the other issues are sorted
out, as it's just a mechanical process.

egs/tedlium/s5/local/confidence_calibration.sh
```diff
@@ -0,0 +1,66 @@
...
+set -euxo pipefail
+
+# Derived options,
+dev_caldir=$dev_latdir/confidence_$lmwt
+eval_caldir=$eval_latdir/confidence_$lmwt
+
+###### Data preparation,
+# Prepare filtering for excluding data from train-set,
+word_filter=$(mktemp)
+awk '{ keep_the_word = $1 !~ /^(\[.*\]|<.*>|%.*|!.*|-.*|.*-)$/; print $0, keep_the_word }' \
```
danpovey: a comment that this prints out either "1 some-word" for words to keep, or "0 some-word" for words to reject such as <cough>, would be nice.

vesis84: Fixed

egs/wsj/s5/steps/conf/train_calibration.sh
```diff
@@ -0,0 +1,120 @@
...
+word_min_count=10 # Minimum word-count for single-word category,
+normalizer=0.0025 # L2 regularization constant,
+stage=0
+# end configuration section.
+
+[ -f ./path.sh ] && . ./path.sh
+. parse_options.sh || exit 1;
+
+if [ $# -ne 7 ]; then
+  echo "Usage: $0 [opts] <data-dir> <lang-dir|graph-dir> <word-filter> <word-length> <unigrams> <decode-dir> <calibration-dir>"
```
danpovey: some documentation of the format of the inputs would be nice, since some of them are non-standard.
However, I think it would be possible to make the input format more general and also easier to work with.
Couldn't the input format be just a file containing:
integer-word-id  feature1 feature2 ... featureN?
That way you wouldn't be filtering out words like <cough> or partial words, you'd just give them their own feature (which presumably would have a negative coefficient), and that would produce a better model for cases where you ended up wanting to predict them.
(BTW, the program 'paste' should be useful here).

danpovey:
Also, sorry for so many requested changes-- if this wasn't an important feature I wouldn't care this much.

egs/wsj/s5/steps/conf/train_calibration.sh
```diff
@@ -0,0 +1,120 @@
...
+  align-text --special-symbol="<eps>" ark:$data/text ark:- ark,t:- | \
+  utils/scoring/wer_per_utt_details.pl --special-symbol "<eps>" \
+  >$dir/align_text 
+  # Append alignment to ctm,
+  steps/conf/append_eval_to_ctm.py $dir/align_text $dir/ctm $dir/ctm_aligned
+fi
+
+# Prepare word-categories (based on wotd frequencies in 'ctm'),
+steps/conf/prepare_word_categories.py --min-count $word_min_count $lang/words.txt $dir/ctm $dir/word_categories
+
```
danpovey: At some point in the future we may want to be able to do domain transfer of the learned confidence model.  This might require us to generate features on a different words.txt (of course, there will probably be some overlap).  The user-provided features could easily be transferred, but for these word_categories, would be there be a way to transfer them?  E.g. keeping the same category for words that are there, but specifying one category as a default category for unseen-in-dev words (presumably the same as for low-count words)?  I'm not saying this needs to be implemented, just bear in mind that we may want to implement it in the future.

danpovey: ... actually, getting these word_categories on the training data instead of the dev data might make this problem simply go away-- that way, you'd require the user to run some script with data/train/text as input, and produce the word-category features, which would just become another field in the user-supplied word feature file.  This would simplify things quite a bit.  Then if all the feature generation is under the user's control, they would presumably have to figure out how to apply it to a mismatching words.txt.

egs/wsj/s5/steps/conf/train_calibration.sh
vesis84: Well, IMHO it is more `readable' to have one feature per file, while the file-name `says' what it is. For example the filtering flag is not even an input feature for logistic regression. Then, some features don't depend only on the word itself, for example : 'frame_length / word_length' needs the word duration, which can't be accommodated by generic look-up table for 'some word-specific features'. On the other hand it increases the complexity of the 'loclal/' script, which needs to prepare more complicated input. I see that possibility to freely add some word-specific features is interesting, but more interesting would be to be able to add some 'dynamic' features no the word-identities. However this implies a modification in the 'steps/conf/prepare_calibration_data.py', which is the only place where information from various sources can be freely combined. So if we adopt the format:
integer-word-id feature1 feature2 ... featureN
we need to implicitly hard-wire the meaning of the columns, which can cause confusion if somebody later on modifies the code. Do you still want to adopt it? I think it is not a good idea, but still I can change it that way...

danpovey: >
>
> > +# (- categorical distribution of 'lang/words.txt', DISABLED)
> > +
> > +# begin configuration section.
> > +cmd=
> > +lmwt=12
> > +decode_mbr=true
> > +word_min_count=10 # Minimum word-count for single-word category,
> > +normalizer=0.0025 # L2 regularization constant,
> > +stage=0
> > +# end configuration section.
> > +
> > +[ -f ./path.sh ] && . ./path.sh
> > +. parse_options.sh || exit 1;
> > +
> > +if [ $# -ne 7 ]; then
> > +  echo "Usage: $0 [opts] <data-dir> <lang-dir|graph-dir> <word-filter> <word-length> <unigrams> <decode-dir> <calibration-dir>"
>
> Well, IMHO it is more readable' to have one feature per file, while the
> file-namesays' what it is. For example the filtering flag is not even an
> input feature for logistic regression.
>
I was suggesting to make the filtering flag an input feature for logistic
regression instead of using it as a filter-- it might give as good results
as what you are doing, and it would simplify the setup.

> Then, some features don't depend only on the word itself, for example :
> 'frame_length / word_length' needs the word duration, which can't be
> accommodated by generic look-up table for 'some word-specific features'.
>
OK, there would be some features which would be added by the
confidence-calibration script itself, rather than the user, and you'd have
to treat the word-length specially.  But there are some features that could
be treated as generic features (e.g. unigrams, and the feature that was
previously the filtering feature, which would just be taken as a hint that
some words are not regular words).  So I'm saying, we could lump all of
those into one category rather than giving them specific names- this would
make the script more generic.

> On the other hand it increases the complexity of the 'loclal/' script,
> which needs to prepare more complicated input. I see that possibility to
> freely add some word-specific features is interesting, but more interesting
> would be to be able to add some 'dynamic' features no the word-identities.
> However this implies a modification in the
> 'steps/conf/prepare_calibration_data.py', which is the only place where
> information from various sources can be freely combined. So if we adopt the
> format:
> integer-word-id feature1 feature2 ... featureN
> we need to implicitly hard-wire the meaning of the columns, which can
> cause confusion if somebody later on modifies the code. Do you still want
> to adopt it? I think it is not a good idea, but still I can change it that
> way...
>
OK, so we might have to hard-code the word length as the first feature, or
give it a separate file.  Are there any other difficulties with what I
proposed?

vesis84: No, I'll try it out, now it's clearer... Thanks!

egs/wsj/s5/steps/conf/train_calibration.sh
vesis84: Yes, the 'unseen words' can go to the category of 'infrequent words'. I can add there a try-hook for 'KeyError' to use the default category '0', this is trivial. The split into categories is saved along with calibration model, so new lattices with new 'words.txt' should recover from that. But for for new words, there are other features to be added (lenght, exclude flag). And maybe in such case, it is better to retrain the calibration. It is a small model... Also, the 'word_categories' are giving only a small improvement, and can be dropped. I keep it in the model, because the various small improvements add up in the end. 

For the 'dev' vs. 'train' for getting the categories. The 'dev' contrains correct words, insertions and subsititutions, while the 'train' contains the reference. And the 'dev' corresponds to data-point counts in the training data of logistic regression, so it can better identify low-representend words in calibration training, compared to case when 'data/train/text' was used. In the Microsoft confidence paper they were using application-specific 'dev' set too...

vesis84: The whole pipeline could be simplified, if we keep only the logit-posterior. But the NCE would be worse of course...

vesis84: When I replaced the filtering be binary feature, 0 ... normal words, 1... special words (that are filtered out now), the NCE on swbd eval got worse : 0.250 -> 0.248. It seems that the data filtering is what the model likes more...

danpovey: Regarding whether to get the word categories from 'dev' or train:
  - How about putting it in the outer level, user-controlled script, and let them run the script that gets them?  That way, in the normal case they can get them on the dev set; but they can choose which set to get them from?  In general I think it's nicer to have as many of the word-specific features as possible come from the user rather than be hard-coded in the script that does the confidence calibration.

danpovey: Regarding 'are there applications where confidences of partial words or non-lexical events are used?':  there might be.  For example, for transcript filtering for semi-supervised training you might need this.  Just having the model throw those words away doesn't seem like the most ideal thing.  And then if you did want confidences for those words, you'd have to give the script an empty list of them, which seems less than ideal.  I'm also concerned that you may be over-tuning on a single test set.  Personally I would not be comfortable to add a 'special' feature to the script (like this filtering) unless I had seen an effect on a couple of different setups.  If you plan to publish this, that would also be nice, as reviewers will likely balk at testing on just one setup.  Perhaps you could figure out a way to evaluate the accuracy of this on some setup that's scored with Kaldi scoring scripts?  That will provide a mechanism for other people to tune this, when they don't have sclite scoring set up.  It shouldn't be that hard to figure out the alignment of the 'text' transcript with the reference using the program Yenda wrote, and then correlate that with the confidences listed in the ctm.  [I would print out the average log-prob of {correct | false} as well as the NCE, as it corresponds more closely to the objective being trained.]  Of course if it does seem to be true in general, that filtering is important, I won't object to it, but if it's good on Switchboard and makes it worse or doesn't matter on, say, Librispeech, I'd say we should leave it out.

danpovey: I thought of a way to avoid having an excessive amount of complexity in the local/ script.  How about having a separate script in steps/conf/ whose job it is to prepare the word-specific features?   We could add language-specific options later as needed, but initially it would be configured to work well for English.
Dan

vesis84: Hi Dan, I waited some time to cool down and think it over. Okay, so how about creating a feature file for the words with the format:
word(str) id(int) filter(int) category(int) word-len(int) log-p-unigram(float) ... [extra feats]

The first 5 columns will have fixed order, while any number of columns can be added.
The first two columns are the 'words.txt'.

The filtering gets disabled by putting '1' everywhere. As it is now, the filtering means that words are excluded only from the training. The confidences get computed for all the words, including the partial words etc... I did check the NCE on 3 databases: switchboard, tedlium and ami/ihm. The filtering helped a little on switchboard, on other databases it had no effect. So I suppose it won't be harmful to other databases either.

The category will be encoded as an integer, 0 means the default category for infrequent an unseen words. The counts will be computed from file format same as 'text'. Ideally the categories should be computed from the 'ctm' that is used for calibration training, which does not exist in 'local/' before the 'steps/conf/train_calibration.sh' is called. So how about optionally passing the 'text' into the 'steps/conf/train_calibration.sh' to replace the 'ctm' by the 'train/text'? The alternative is to produce it outside, but then we will need to add dependency on the normal scoring (sometimes kaldi sometimes, sclite, probably you won't like this)...
So we'll better exclude that column from the 'external' word features.

Word lenght is the number of characters in the word.

And the log-p-unigram is the first column of the section that can have arbitrary number of features that get forwarded to the logistic regression input.

I don't plan to publish a paper about confidence calibration, the implementation follows standard techniques. Concerning the alternative way to compute the NCE, it might not be necessary at this moment. I'd like to finish this up to continue the work on the thesis. I primarily created the implementation for my personal use in the thesis, and as a courtesy I wanted to share it into kaldi. But I don't think I should spend too much time on it...

danpovey: >
> In egs/wsj/s5/steps/conf/train_calibration.sh
> <https://github.com/kaldi-asr/kaldi/pull/385#discussion_r48456960>:
>
> > +  rm $dir/*.ctm; set -x
> > +fi
> > +
> > +# Get evaluation of the 'ctm' using the 'text' reference,
> > +if [ $stage -le 1 ]; then
> > +  steps/conf/convert_ctm_to_tra.py $dir/ctm - | \
> > +  align-text --special-symbol="<eps>" ark:$data/text ark:- ark,t:- | \
> > +  utils/scoring/wer_per_utt_details.pl --special-symbol "<eps>" \
> > +  >$dir/align_text
> > +  # Append alignment to ctm,
> > +  steps/conf/append_eval_to_ctm.py $dir/align_text $dir/ctm $dir/ctm_aligned
> > +fi
> > +
> > +# Prepare word-categories (based on wotd frequencies in 'ctm'),
> > +steps/conf/prepare_word_categories.py --min-count $word_min_count $lang/words.txt $dir/ctm $dir/word_categories
> > +
>
> Hi Dan, I waited some time to cool down and think it over. Okay, so how
> about creating a feature file for the words with the format:
> word(str) id(int) filter(int) category(int) word-len(int)
> log-p-unigram(float) ... [extra feats]
>
> The first 5 columns will have fixed order, while any number of columns can
> be added.
> The first two columns are the 'words.txt'.
>
> The filtering gets disabled by putting '1' everywhere. As it is now, the
> filtering means that words are excluded only from the training. The
> confidences get computed for all the words, including the partial words
> etc... I did check the NCE on 3 databases: switchboard, tedlium and
> ami/ihm. The filtering helped a little on switchboard, on other databases
> it had no effect. So I suppose it won't be harmful to other databases
> either.
>

OK, fine.

> The category will be encoded as an integer, 0 means the default category
> for infrequent an unseen words. The counts will be computed from file
> format same as 'text'. Ideally the categories should be computed from the
> 'ctm' that is used for calibration training, which does not exist in
> 'local/' before the 'steps/conf/train_calibration.sh' is called. So how
> about optionally passing the 'text' into the
> 'steps/conf/train_calibration.sh' to replace the 'ctm' by the 'train/text'?
> The alternative is to produce it outside, but then we will need to add
> dependency on the normal scoring (sometimes kaldi sometimes, sclite,
> probably you won't like this)...
> So we'll better exclude that column from the 'external' word features.
>

Sure, makes sense.

> Word lenght is the number of characters in the word.
>
OK.  BTW, it occurred to me that the number of phones in a word might also
be a relevant measure of length- in particular for computing your
normalized length- and that if the model is available, the number of phones
in a word is available directly from the alignment (the function
SplitToPhones()).  This might involve some C++ coding though.  I'll leave
it up to you whether you want to pursue this

> And the log-p-unigram is the first column of the section that can have
> arbitrary number of features that get forwarded to the logistic regression
> input.
>
> I don't plan to publish a paper about confidence calibration, the
> implementation follows standard techniques. Concerning the alternative way
> to compute the NCE, it might not be necessary at this moment. I'd like to
> finish this up to continue the work on the thesis. I primarily created the
> implementation for my personal use in the thesis, and as a courtesy I
> wanted to share it into kaldi. But I don't think I should spend too much
> time on it...
>

OK, fine.
Dan

vesis84: Okay, I'll replace the number of characters by number of phones. This can be parsed dirtily and easily from the 'lang/phones/align_lexicon.txt' In 'local/' of course. Supposing the phone counts in pronunciation variants don't change much, we can take the last one. It gave a tiny 0.250 -> 0.251 NCE improvement on switchboard. On tedlium an ami/ihm it did nothing.

danpovey: OK, cool.

Dan

vesis84:
confidence: implementing next suggestions from Dan,
- replacing character-length of words by number of phonemes in a word,
- implementing word-specific feature file,
  - format: "word word-id filter  word-phone-length other_feats'
- replacing words with word-id's in 'steps/conf/prepare_calibration_data.py',

danpovey:
Thanks!
I won't do another pass of detailed review.  Do you consider it ready to merge?

vesis84:
Yes, it is ready for merging. Thanks!

PR #387, opened 2015-12-08
Modified nnet3-combine to accept single models

vijayaditya:


src/nnet3bin/nnet3-combine.cc
```diff
@@ -85,23 +85,28 @@ int main(int argc, char *argv[]) {
     
     
     int32 num_nnets = po.NumArgs() - 2;
-    NnetCombiner combiner(combine_config, num_nnets, egs, nnet);
+    if (num_nnets > 1)  {
```
danpovey: I wonder if this should be
if (num_nnets > 1 || !combine_config.enforce_sum_to_one) {  
because there could actually be some meaningful change even for one nnet, if that config is set to false.

vijayaditya:
@danpovey  corrected and tested.

danpovey:
Thanks!  Merging.

PR #388, opened 2015-12-09
Small fix to lattice determinization

vdp:
As requested by Dan in #382 

Dan, I think that in general you are trying to enforce 80 character limit, so you may find some lines a bit longish, but there are much longer lines in these files anyway, so I don't think this is a big deal (plus one could argue that in today's world, where the buying power of the YouTube fans leads to omnipresent abominations like 16:9 screens, the vertical real estate is at least as valuable).
Also, I'm not sure that the final block scopes are needed, but I wanted to make as few changes as possible.

### DIFF
--- a/src/fstext/determinize-lattice-inl.h
+++ b/src/fstext/determinize-lattice-inl.h
@@ -547,6 +547,10 @@ template<class Weight, class IntType> class LatticeDeterminizer {
       return (state != other.state || string != other.string ||
               weight != other.weight);
     }
+    // This operator is only intended to support sorting in EpsilonClosure()
+    bool operator < (const Element &other) const {
+      return state < other.state;
+    }
   };
 
   // Arcs in the format we temporarily create in this class (a representation, essentially of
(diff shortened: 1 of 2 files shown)


src/fstext/determinize-lattice-inl.h
```diff
@@ -547,6 +547,11 @@ template<class Weight, class IntType> class LatticeDeterminizer {
       return (state != other.state || string != other.string ||
               weight != other.weight);
     }
+    // This operator is only intended to support sorting in the function
```
danpovey: Can you please implement an operator < instead?  That's the operator the sorting looks for by default, and in that case you wouldn't have to provide std::greater<Element>() to the function.  Also, this way we'll get the same order that the old code would have produced, which doesn't really matter but which is maybe less surprising.

vdp:
@danpovey OK, I did it this way in order to reuse the already existing ">" operator in LatticeDeterminizerPruned, and added this operator LatticeDeterminizer for consistency.. Please find the corrected version (not tested yet, will recompile and test in a bit).

danpovey:
Looks fine but I'll wait till you say it's tested.  Thanks.

vdp:
Seems to work OK..

PR #389, opened 2015-12-09
Bug fixes in music-speech example and binaries

david-ryan-snyder:


danpovey:
David, I think you may have forgotten to add the link to the 'sid' directory to git.

PR #397, opened 2015-12-14
some updates for SparseLinearComponent

hhadian:
1. Minor fixes regarding assertions in non-updatable mode and keeping transposed version in sync.
2. Fix a bug in SparseMatrix::CopyFromMat from previous commit. 
3. Add FrobeniusNorm() for CuRowSparseMatrix and add uncentered-variance to SparseLinearComponent's Info().


danpovey:
Accepting without review, will review whole branch if and  when we merge to trunk.

PR #399, opened 2015-12-15
Multitree

hainan-xv:
add the code and scripts related to trees;
slightly changed the sparse matrix code 
 - add support for >65535 num_rows
 - change the init function so that it reads directly from a SparseMatrix file instead of MatrixBase


danpovey:
Merging without review for now-- will add comments later if I have time.

PR #402, opened 2015-12-16
Added an option to the nnet3 configs so that the final log softmax layer is selective.

tomkocse:


danpovey:
Thanks!
Merging.
BTW, I think you mean "optional" instead of "selective".  (Perhaps that's an Indian-subcontinent usage of that word?).

PR #403, opened 2015-12-16
music-speech: Adding symbolic link to the sid directory in egs/bn_music_speech/v1

david-ryan-snyder:
This directory is necessary to get the example to run.


jtrmal:
Merging.

PR #404, opened 2015-12-17
Make the cstr scripts for wsj use dict suffix

phsmit:
The scripts for the cstr style wsj directory layout did not take in
account the dict_suffix. The script is now brought up-to-date with
wsj_extend_dict.sh

(this is my first PR here. Please let me know if I have forgotten something or should do something different)


danpovey:
This seems to make sense, and since it won't break anything in the normal setting I'll merge it right away.
Thanks, and welcome!

PR #405, opened 2015-12-19
Minor cosmetic changes in lattice determinization code

vdp:
Change a few lines in EpsilonClosure(), that look a bit more hairier, than they need to be (likely related to the change from std::map to std::unordered_map, mentioned by Dan in #382).


danpovey:
Thanks!  Merging.

PR #406, opened 2015-12-19
WIP : Added AMI recipe for using clean alignments, minor bug correction in …

vijayaditya:
…IHM recipe

This commit is WIP as it modifies the local/nnet3/run_ivector_common.sh. It add options to train with clean alignments which complicates this significantly. So some feedback on understandability of the script is welcome.

### DIFF
--- a/egs/ami/s5/RESULTS_sdm
+++ b/egs/ami/s5/RESULTS_sdm
@@ -73,6 +73,7 @@ for x in exp/$mic/nnet2_online/*/decode*; do [ -d $x ] && [[ $x =~ "$1" ]] && gr
 
 #TDNN
 #Total training time is 5:19:19
+# local/nnet3/run_tdnn.sh --mic sdm1
 %WER 46.1 | 15377 94333 | 59.1 25.8 15.0 5.2 46.1 65.6 | -24.026 | exp/sdm1/nnet3/tdnn_sp/decode_dev/ascore_11/dev_hires_o4.ctm.filt.sys
 %WER 50.9 | 13867 89975 | 53.6 27.9 18.5 4.5 50.9 70.1 | -23.332 | exp/sdm1/nnet3/tdnn_sp/decode_eval/ascore_11/eval_hires_o4.ctm.filt.sys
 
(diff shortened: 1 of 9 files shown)


danpovey:
It's great that you seem to be getting so much better results.

However, there are some things in there that I don't really like:
  - you source that 'set_variable_names.sh' script.  That is potentially quite confusing because when people are searching for where a particular variable is set, they'll typically look within the script itself.
  - you say something about resetting the 'mic' variable.  Resetting variables in scripts is potentially very confusing.

In general in this script I think there is an overuse of variables.  Having too many variables can be quite confusing in example scripts.  It would be better, IMO, to just change the scripts to use the best setup with parallel data, if that works better.  Or if for some reason you need the old setup still, then rely on copy-and-modify.  In my example scripts I tend to rely a lot more on copy-and-modify than on variables, because in general you only end up caring about one setting, and it's easier to just have a straightforward script with very few variables (mainly just --stage and --train-stage) that just calls the setup that you want.

vijayaditya:
OK, I will simplify the scripts by making speed perturbation and SAT alignments default. I will just create a copy of the script for using SDM alignments.

vijayaditya:
I have made speed_perturbation non-optional. 
I decided to keep the option <code>use_sat_alignments</code> as all the systems in RESULTS file have been built without sat alignments. I will eliminate this once I have rerun these systems with sat alignments.

I have reduced branching in the <code>local/nnet3/run_ivector_common.sh</code> script by adding two additional scripts <code>local/nnet3/prepare_parallel_perturbed_alignments.sh</code> and <code>local/nnet3/prepare_perturbed_alignments.sh</code>. 

Let me know if further simplification is necessary.

egs/ami/s5/RESULTS_sdm
```diff
@@ -83,14 +84,25 @@ for x in exp/$mic/nnet2_online/*/decode*; do [ -d $x ] && [[ $x =~ "$1" ]] && gr
...
 %WER 47.9 | 14766 89956 | 56.0 25.7 18.2 3.9 47.9 64.1 | -23.997 | exp/sdm1/nnet3/lstm_sp_ld5/decode_eval/ascore_10/eval_hires_o4.ctm.filt.sys
 
 # BLSTM
-# local/nnet3/run_blstm.sh
+# local/nnet3/run_blstm.sh --mic sdm1
 %WER 42.8 | 14948 94501 | 62.2 25.2 12.6 5.1 42.8 65.8 | -24.499 | exp/sdm1/nnet3/lstm_sp_bidirectional_ld0/decode_dev/ascore_10/dev_hires_o4.ctm.filt.sys
 %WER 46.1 | 13760 89981 | 57.8 25.9 16.3 3.9 46.1 68.2 | -24.143 | exp/sdm1/nnet3/lstm_sp_bidirectional_ld0/decode_eval/ascore_10/eval_hires_o4.ctm.filt.sys
 
+# BLSTM + clean alignments
+#  local/nnet3/run_blstm.sh --mic $mic \
```
danpovey: what is $mic here?

egs/ami/s5/local/nnet3/prepare_parallel_datadirs.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
```
danpovey: at the top of scripts like this, a couple of lines saying what it's for would be nice.

egs/ami/s5/local/nnet3/prepare_parallel_perturbed_alignments.sh
```diff
@@ -0,0 +1,77 @@
+mic=sdm1
```
danpovey: same herei please add a comment at the top saying what this script does.

egs/ami/s5/local/nnet3/run_lstm.sh
```diff
@@ -1,4 +1,3 @@
-#!/bin/bash
 
 # Copyright 2015  Johns Hopkins University (Author: Daniel Povey).
 #           2015  Vijayaditya Peddinti
```
danpovey: it would probably make sense to mention that this is also called from other scripts in local/nnet3/, such as run_blstm.sh.

egs/ami/s5/local/nnet3/prepare_parallel_perturbed_alignments.sh
vijayaditya: Sorry that was an oversight.. I will add the comments

vijayaditya:
Completed the specified changes.

danpovey:
OK, merging.  Thanks!

PR #412, opened 2015-12-23
fix a bug, avoiding ambiguous reference

tzyll:
ambiguous reference "int64" between base/kaldi-types.h and ../tools/openfst/include/fst/types.h

### DIFF
--- a/egs/wsj/s5/steps/nnet3/ctc/make_denlats.sh
+++ b/egs/wsj/s5/steps/nnet3/ctc/make_denlats.sh
@@ -1,0 +1,167 @@
+#!/bin/bash
+
+# Copyright 2012-2015  Johns Hopkins University (Author: Daniel Povey).
+# Apache 2.0.
+
+# This script does decoding with a nnet+ctc neural-net.  If the neural net was
+# built on top of fMLLR transforms from a conventional system, you should
+# provide the --transform-dir option.
+
+# Begin configuration section.
+stage=1
+transform_dir=    # dir to find fMLLR transforms.
+nj=4 # number of decoding jobs.  If --transform-dir set, must match that number!
+acwt=1.0  # Just a default value, used for adaptation and beam-pruning..
+blank_scale=0.4
+lattice_acoustic_scale=10.0  # This is kind of a hack; it's used for
+                             # compatibility with existing scoring scripts,
+                             # since we normally search the LM-scale in integer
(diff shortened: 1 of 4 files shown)


danpovey:
Thanks.  Merging, but be aware that I don't have plans to do much with this branch in future.  The 'chain' models are performing better than the CTC models, and are simpler and easier to motivate, so that's what I intend to use going forward.  (Of course, it wouldn't be hard to just merge all this to master, but that would mean that we have to support it, which I don't want to do).

tzyll:
Our experiments showed that with larger data (e.g.1400h), CTC does well, even better than DNN baseline without MPE, So we want to put MPE on CTC to see its potential.

danpovey:
Makes sense.  BTW, you might get less gains with MPE from this version of
CTC than Google did.  The reason would take a while to explain- basically,
certain  mismatches due the use of context-dependent labels in Google's
setup don't exist in the setup I implemented because of the way we
normalize the probabilities.

What would be interesting for me is if you would test the 'chain' models
(chain branch; and see egs/swbd/s5/local/chain/run_tdnn_2y.sh for the
top-level example script) on your larger setup.  I haven't had a chance to
do this yet.  I'm confident the results will be even better than those CTC
numbers.
Dan

tzyll:
Thanks a lot. I will test 'chain', then share the results on at least 1400h Chinese speech.

Yours,
Yuan

----- 原始邮件 -----
发件人: "Daniel Povey" <notifications@github.com>
收件人: "kaldi-asr/kaldi" <kaldi@noreply.github.com>
抄送: "Zhiyuan Tang" <tangzy@cslt.riit.tsinghua.edu.cn>
发送时间: 星期五, 2015年 12 月 25日 上午 5:25:40
主题: Re: [kaldi] fix a bug, avoiding ambiguous reference (#412)

Makes sense. BTW, you might get less gains with MPE from this version of 
CTC than Google did. The reason would take a while to explain- basically, 
certain mismatches due the use of context-dependent labels in Google's 
setup don't exist in the setup I implemented because of the way we 
normalize the probabilities. 

What would be interesting for me is if you would test the 'chain' models 
(chain branch; and see egs/swbd/s5/local/chain/run_tdnn_2y.sh for the 
top-level example script) on your larger setup. I haven't had a chance to 
do this yet. I'm confident the results will be even better than those CTC 
numbers. 
Dan

Issue #413, opened 2015-12-23
SumReduceComponent


danpovey:
This relates to the 'chain' branch.  This branch will be merged back to master in the next month or two most likely (but certainly will eventually).
There is a component called SumReduceComponent that interprets its input as a sequence of blocks, and sums the blocks.  I want it to have an option to output the block sums instead, so that output[0] = input[0] + input[1] + input[2] ... instead of output[0] = input[0] + input[100] + ... .
We don't need to worry about back compatibility just now because that isn't used in any current recipe.
The block sizes are determined by the input and output dims.

One of the problems is that I don't have a good way of explaining the difference in a catchy way that can be expressed in a boolean option.  See if you can think of something.  
Also doing the summation efficiently is a little tricky.  For now I think the easiest way to do it, without involving writing new CUDA kernels, is to do it in the same way as is done in the SumGroupComponent, which will involve storing an indexes_ and a reverse_indexes_ member (treat them as derived variables, i.e. never read to/from disk).  In this case, the SumReduceComponent acts like a special, restricted case of the SumGroupComponent.

danpovey:
created for @galv

galv:
Is it necessary to add this functionality to the existing class SumReduceComponent, or could a new class be created? I am sure it can be done this way, but it feels wrong that most methods will begin with a switch on the boolean argument to decide which functionality to perform. If we want more options in the future, once this code is "released" in recipes, it will make serialization more of a pain to keep adding functionalities to the same class.

Otherwise, I believe "over-elements" would be a suitable boolean option name, as you are asking for a reduction over the elements of each block, instead of a reduction over the blocks.

danpovey:
That makes sense- you can do it as a separate component.

galv:
Another question. I want to be clear on the definition of a block. Sorry, but I cannot find much about what a block is from the code through grep. block matrices show up too often.

Is this is a correct statement about blocks? A vector of blocks is a vector of groups, where each group has the same number of elements. (This statement ignores the rows of the matrix, but my understanding is that the row dimension is the batch dimension anyway.)

If so, I agree with you that this is a special case of "SumGroupComponent". Is there a reason why you did not propose simply subclassing SumGroupComponent instead of reimplementing its functionality? I would otherwise make this design decision myself, but you were oddly specific (which I appreciate) in the original issue. It's a serious question, since I do not know the grand picture why you want this particular functionality.

danpovey:
[...]

PR #419, opened 2015-12-26
Add special case to SimpleComponent's InitFromConfig for easy

galv:
creation when all groups are of the same size.

Addresses issue #413.

I chose to implement the functionality by providing a new initialization to SumGroupComponent. As a result, serialization is not as efficient as it could be. If this is undesirable, I will redo this.


danpovey:
Looks great-- thanks!  Merging.

PR #421, opened 2015-12-28
New option for nnet1-to-raw-nnet

tal1974:
New option for nnet1-to-raw-nnet. I asked here https://groups.google.com/forum/#!topic/kaldi-help/NSmygntxkgE. May I change this code now?

### DIFF
--- a/src/nnet2bin/nnet1-to-raw-nnet.cc
+++ b/src/nnet2bin/nnet1-to-raw-nnet.cc
@@ -28,17 +28,17 @@
 #include "nnet2/nnet-component.h"
 
 namespace kaldi {
-  bool affinecompprecond = false;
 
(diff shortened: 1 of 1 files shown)


src/nnet2bin/nnet1-to-raw-nnet.cc
```diff
@@ -28,6 +28,7 @@
 #include "nnet2/nnet-component.h"
 
 namespace kaldi {
+  bool affinecompprecond = false;
```
danpovey: this should be declared in the main() and passed into the function, and should have a name like use_preconditioned_affine_component.

tal1974:
I corrected, as you wrote.

danpovey:
Thanks!  Merging.

PR #423, opened 2015-12-29
add RepeatedAffineComponent to chain branch

zhichaowang:
add RepeatedAffineComponent to chain branch

### DIFF
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -975,6 +975,337 @@ Component *AffineComponent::CollapseWithPrevious(
   return ans;
 }
 
+//RepeatedAffineComponent::RepeatedAffineComponent() {}
+
+RepeatedAffineComponent::RepeatedAffineComponent(const RepeatedAffineComponent & component) :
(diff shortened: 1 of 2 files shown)


danpovey:
OK, thanks.  The Copy() function needs to be implemented, but I'll do that myself.
Merging now.

danpovey:
I just pushed a change to finish this and fix problems, do
git diff 0430d56..cbf511e
to see the diff.
Issues include:
 - memory leaks
 - backprop code with race condition [multiple simultaneous writes to the same matrix].
 - the backprop function not being 'const' (note: it's the parameters of the component in 'to_update' that need to be changed.)
 - adding test example in nnet-test-utils.cc and making sure the tests pass.
 - efficiency improvements (avoid loops calling CUDA kernels)
 - remove  a couple functions that are not part of the standard interface, and were not needed.
 - add the component to NewComponentOfType() in nnet-component-itf.cc

naxingyu:
Thanks for the fix.
X.

在 2015/12/29 12:50, Daniel Povey 写道:

PR #427, opened 2015-12-31
Block affine component

galv:
Note the commit message regarding a change to nnet-test-utils.cc

There is no actual change in natural-gradient-online-test.cc. I simply accidentally committed a change, and couldn't figure out how to remove it via rebasing, so I had to change it again back to its previous state in the second commit.

### DIFF
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -2,6 +2,7 @@
 
 // Copyright      2015  Johns Hopkins University (author: Daniel Povey)
 //                2015  Guoguo Chen
+//                2015  Daniel Galvez
 
 // See ../../COPYING for clarification regarding multiple authors
 //
@@ -1297,8 +1298,6 @@ BlockAffineComponent::BlockAffineComponent(const BlockAffineComponent &other) :
   bias_params_(other.bias_params_),
   num_blocks_(other.num_blocks_) {}
 
-//TODO: Copy constructor, and just-params constructors
-// At least, RepeatedAffineComponent has these.
 Component* BlockAffineComponent::Copy() const {
   BlockAffineComponent *ans = new BlockAffineComponent(*this);
   return ans;
@@ -1327,12 +1326,11 @@ std::string BlockAffineComponent::Info() const {
 void BlockAffineComponent::Init(BaseFloat learning_rate, int32 input_dim,
                                 int32 output_dim, int32 num_blocks,
                                 BaseFloat param_stddev, BaseFloat bias_stddev) {
-  KALDI_ASSERT(input_dim > 0 && output_dim > 0 && num_blocks > 1);
+  KALDI_ASSERT(input_dim > 0 && output_dim > 0 && num_blocks >= 1);
   KALDI_ASSERT(output_dim % num_blocks == 0 && input_dim % num_blocks == 0);
   const int32 num_columns_per_block = input_dim / num_blocks;
 
   UpdatableComponent::Init(learning_rate);
-  // TODO: What about is_gradient_???
 
   linear_params_.Resize(output_dim, num_columns_per_block);
   bias_params_.Resize(output_dim);
@@ -1346,13 +1344,6 @@ void BlockAffineComponent::Init(BaseFloat learning_rate, int32 input_dim,
 }
 
 void BlockAffineComponent::InitFromConfig(ConfigLine *cfl) {
-  // TODO: Make this documentation cleaner and viewable from doxygen.
-  // num_blocks
-  // input dim - num cols of block diagonal matrix.
-  // output dim - num rows of block diagonal matrix.
-  // Rows of each block is output dim / num_blocks
-  // bias dimension should be output dim
-
   int32 input_dim, output_dim, num_blocks;
   if(!cfl->GetValue("input-dim", &input_dim) ||
      !cfl->GetValue("output-dim", &output_dim) ||
@@ -1416,10 +1407,6 @@ void BlockAffineComponent::Backprop(const std::string &debug_info,
                                     CuMatrixBase<BaseFloat> *in_deriv) const {
   BlockAffineComponent *to_update = dynamic_cast<BlockAffineComponent*>(to_update_in);
 
-  // my own sanity check.
-  // TODO: Remove these after testing.
-  KALDI_ASSERT(out_deriv.NumCols() == OutputDim());
-
   const int32 num_rows_in_block = linear_params_.NumRows() / num_blocks_;
   const int32 num_cols_in_block = linear_params_.NumCols();
 
@@ -1428,10 +1415,6 @@ void BlockAffineComponent::Backprop(const std::string &debug_info,
   // If we wanted to add with coefficient 0.0 we'd need to zero the
   // in_deriv, in case of infinities.
   if (in_deriv) {
-    // my own sanity check.
-    // TODO: Remove these after testing.
-    KALDI_ASSERT(in_deriv->NumCols() == InputDim());
-
     std::vector<CuSubMatrix<BaseFloat> *> in_deriv_batch, out_deriv_batch, linear_params_batch;
 
     for(int block_counter = 0; block_counter < num_blocks_; block_counter++) {
@@ -1502,7 +1485,6 @@ void BlockAffineComponent::Backprop(const std::string &debug_info,
     } // end linear params update
 
     { // bias update
-      // be fishly about this. I didn't think this through fully.
       to_update->bias_params_.AddRowSumMat(to_update->learning_rate_,
                                            out_deriv, 1.0);
     } // end bias update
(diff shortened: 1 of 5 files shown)


src/nnet3/nnet-simple-component.h
```diff
@@ -454,6 +455,76 @@ class RepeatedAffineComponent: public UpdatableComponent {
...
+  //   N
+  //   O ] but we actually treat it as the matrix:
+  // [ M 0 0
+  //   0 N 0
+  //   0 0 O ]
+  CuMatrix<BaseFloat> linear_params_;
+  CuVector<BaseFloat> bias_params_;
+  int32 num_blocks_;
+ private:
+  const RepeatedAffineComponent &operator = (const RepeatedAffineComponent &other); // Disallow.
```
danpovey: should be BlockAffineComponent not RepeatedAffineComopnent

src/nnet3/nnet-test-utils.cc
```diff
@@ -1026,12 +1026,21 @@ static void GenerateRandomComponentConfig(std::string *component_type,
     case 24: {
       *component_type = "RepeatedAffineComponent";
       int32 num_repeats = RandInt(1, 50),
-          input_dim = num_repeats * RandInt(1, 15),
-          output_dim = num_repeats * RandInt(1, 15);
```
danpovey: was there a reason this crashed with 1?

danpovey:
Thanks!  Apart from those very small issues I pointed out, this looks great.
Dan

src/nnet3/nnet-test-utils.cc
galv: The numerical routines (i.e., Propagate and Backprop) did not crash when input_dim=1 and output_dim=1. Rather, the crash in ./nnet-component-test was because predicted_obj_change was not approximately equal to measured_objf_change in TestSimpleComponentDataDerivative().

Obviously, the test works for other (larger) input parameters. Simply removing a test case that doesn't work seems a little bad, but I think the test is likely to be a false positive since the tests work for other tested parameters.

Of course I also don't really understand the tests as well as I should. (I don't understand why the output_deriv-output dot product is a valid objective function.)

danpovey: It prints out the measured and predicted objf changes- how different were
they?  Were they at least usually the same sign?   If they were just a bit
out of the tolerance it's OK to disable that test case, but if they were
completely wrong, it could indicate a problem e.g. with the derivatives for
the bias.
Dan

galv: I'll look into the details of what exactly triggers the failure and get
some numbers soon.

On Thu, Dec 31, 2015 at 3:35 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/nnet3/nnet-test-utils.cc
> <https://github.com/kaldi-asr/kaldi/pull/427#discussion_r48670460>:
>
> > @@ -1026,12 +1026,21 @@ static void GenerateRandomComponentConfig(std::string *component_type,
> >      case 24: {
> >        *component_type = "RepeatedAffineComponent";
> >        int32 num_repeats = RandInt(1, 50),
> > -          input_dim = num_repeats * RandInt(1, 15),
> > -          output_dim = num_repeats * RandInt(1, 15);
>
> It prints out the measured and predicted objf changes- how different were
> they? Were they at least usually the same sign? If they were just a bit out
> of the tolerance it's OK to disable that test case, but if they were
> completely wrong, it could indicate a problem e.g. with the derivatives for
> the bias. Dan
> … <#151f9c09c98656f2_>
> On Thu, Dec 31, 2015 at 9:48 AM, Daniel Galvez <notifications@github.com>
> wrote: In src/nnet3/nnet-test-utils.cc <#427 (comment)
> <https://github.com/kaldi-asr/kaldi/pull/427#discussion_r48664776>>: > @@
> -1026,12 +1026,21 @@ static void GenerateRandomComponentConfig(std::string
> *component_type, > case 24: { > *component_type =
> "RepeatedAffineComponent"; > int32 num_repeats = RandInt(1, 50), > -
> input_dim = num_repeats * RandInt(1, 15), > - output_dim = num_repeats *
> RandInt(1, 15); The numerical routines (i.e., Propagate and Backprop) did
> not crash when input_dim=1 and output_dim=1. Rather, the crash in
> ./nnet-component-test was because predicted_obj_change was not
> approximately equal to measured_objf_change in
> TestSimpleComponentDataDerivative(). Obviously, the test works for other
> (larger) input parameters. Simply removing a test case that doesn't work
> seems a little bad, but I think the test is likely to be a false positive
> since the tests work for other tested parameters. Of course I also don't
> really understand the tests as well as I should. (I don't understand why
> the output_deriv-output dot product is a valid objective function.) — Reply
> to this email directly or view it on GitHub <
> https://github.com/kaldi-asr/kaldi/pull/427/files#r48664776>.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/427/files#r48670460>.
>



-- 
Daniel Galvez
danieltgalvez.com

src/nnet3/nnet-simple-component.cc
```diff
@@ -1291,6 +1292,293 @@ void RepeatedAffineComponent::UnVectorize(const VectorBase<BaseFloat> &params) {
...
+
+    { // linear params update
+
+      std::vector<CuSubMatrix<BaseFloat> *> in_value_batch,
+        out_deriv_batch, linear_params_deriv_batch;
+
+      // this constructor zeros out the entire matrix. Unfortunately, we need
+      // this for AddMatMatBatched even though beta is 0 because it is possible
+      // there may be NaN in the matrix's space, and 0 * NaN is NaN.
+      CuMatrix<BaseFloat> linear_params_deriv(linear_params_.NumRows(),
```
danpovey: I don't think the temporary is necessary here.  It was needed in the RepeatedAffineComponent to avoid a race condition, but here you can directly add to linear_params_.  The learning rate is normally large enough that any roundoff error is negligible, and anyway we do it like this elsewhere.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1291,6 +1292,293 @@ void RepeatedAffineComponent::UnVectorize(const VectorBase<BaseFloat> &params) {
...
+     !cfl->GetValue("output-dim", &output_dim) ||
+     !cfl->GetValue("num-blocks", &num_blocks) || cfl->HasUnusedValues()) {
+    KALDI_ERR << "Invalid initializer for layer of type "
+              << Type() << ": \"" << cfl->WholeLine() << "\"";
+  }
+
+  // optional parameters
+  BaseFloat learning_rate = learning_rate_;
+  cfl->GetValue("learning-rate", &learning_rate);
+  BaseFloat param_stddev = 1.0 / std::sqrt(input_dim),
```
danpovey: Please make the default param_stddev 1.0 / std::sqrt(input_dim / num_blocks), and also do the same
in the RepeatedAffineComponent.  There's a reason for this (it's natural if you follow the logic in the original paper that introduced this rule, to this case).  I already fixed it in the branch where I'm using it, but not upstream.  It does matter.

src/nnet3/nnet-simple-component.cc
galv: I think you're absolutely right here. I can't believe I didn't realize this myself.

src/nnet3/nnet-simple-component.cc
galv: I assume you're talking about Glorot intialization. If so, I agree with these changes.

src/nnet3/nnet-test-utils.cc
galv: Okay, I have confirmed that the tests fail only RepeatedAffineComponent (NOT BlockAffineComponent. That looks fine.), when input_dim / num_repeats and output_dim / num_repeats are equal to 1. Even then, it is not guaranteed to fail, however. It can also fail for other small values.

I suspect the reason has to do with striding. I'm worried your hack to view the vector as a matrix assumes that does not take into account the effect of extra padding used between rows for the sake of alignment. A padding issue exactly like the kind of thing that would come up only with very small dimensions. I don't know CUDA, but the [documentation](http://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g32bd7a39135594788a542ae72217775c) mentions that padding is used in allocations.

I'll test this hypothesis soon, hopefully tomorrow.

danpovey: I think I see what might be the problem.  The following code:
      int32 size_as_vector =
          (linear_params_.NumRows() - 1) * linear_params_.Stride() +
          linear_params_.NumCols(),
          stride_as_matrix = linear_params_.NumRows() *
linear_params_.Stride();
where it says linear_params_.Stride(), needs to refer to
linear_params_deriv_repeated.Stride().
They are usually but not always the same.
Dan

galv:
I've integrated all of your changes and rebased against chain.

I noticed that in the chain branch, you only partially made the striding fix you proposed. You replaced only one instance of linear_params_.Stride() with linear_params_deriv_repeated.Stride(). I made the change in both places, and have thoroughly tested to make sure the gradient checks pass for the case that input-dim= 1* num_repeats_ and output-dim = 1 * num_repeats_, and for small repeated blocks in general.

The previous error must have been due to accessing illegal memory. Sometimes when the downhill test failed, it reported that the new objective function was nan or infinity.

danpovey:
Thanks a lot!   Merging.

PR #431, opened 2016-01-03
Multitree

hainan-xv:


src/tree/event-map.cc
```diff
@@ -94,6 +94,11 @@ bool ConstantEventMap::IsSameTree (const EventMap* other) const {
   return true;
 }
 
+void ConstantEventMap::ExpandTree(const std::vector<key_yesset> &questions, int* next) {
```
danpovey: here, key_yesset appears to be a type; why is it not written LikeThis?

danpovey:
Please in future avoid commits without a commit message (or '..').  In future we can rebase this.  All of this branch is a work in progress, so I'll merge it to make collaboration, but before anything in this branch gets merged to master we'll have to do a thorough review.

PR #433, opened 2016-01-05
WIP: added I/O interfaces for NnetComputation and ComputationRequest

xiaohui-zhang:
@danpovey Dan can you review my PR? Some details may need to be improved. Tests were added into nnet-compute-test.cc as you mentioned.


src/nnet3/nnet-computation.cc
```diff
@@ -413,6 +608,210 @@ void NnetComputation::Print(std::ostream &os, const Nnet &nnet) const {
...
+//  WriteToken(os, binary, "<ComponentPrecomputedIndexes>");
+//  for (int32 c = 0; c < component_precomputed_indexes.size(); c++) {
+//    if (component_precomputed_indexes[c] != NULL) 
+//      component_precomputed_indexes[c]->Write(os, binary);
+//  }
+
+  WriteToken(os, binary, "<NumIndexes>");
+  WriteBasicType(os, binary, indexes.size());
+  WriteToken(os, binary, "<Indexes>");
+  for (size_t c = 0; c < indexes.size(); c++) {
```
danpovey: You should use WriteIntegerVector here- much more efficient.

src/nnet3/nnet-computation.cc
```diff
@@ -413,6 +608,210 @@ void NnetComputation::Print(std::ostream &os, const Nnet &nnet) const {
...
+      WriteBasicType(os, binary, indexes[c][i]);
+  }
+
+  WriteToken(os, binary, "<NumIndexesMulti>");
+  WriteBasicType(os, binary, indexes_multi.size());
+  WriteToken(os, binary, "<IndexesMulti>");
+  for (size_t c = 0; c < indexes_multi.size(); c++) {
+    WriteBasicType(os, binary, indexes_multi[c].size());
+    for (size_t i = 0; i < indexes_multi[c].size(); i++) {
+      WriteBasicType(os, binary, indexes_multi[c][i].first);
```
danpovey: You could add a function WriteIntegerPairVector modeled after WriteIntegerVector (assume the types are the same).

src/nnet3/nnet-computation.cc
```diff
@@ -413,6 +608,210 @@ void NnetComputation::Print(std::ostream &os, const Nnet &nnet) const {
...
+    WriteBasicType(os, binary, iter->second.first);
+    WriteBasicType(os, binary, iter->second.second);
+  }
+
+  WriteToken(os, binary, "<NumCommands>");
+  WriteBasicType(os, binary, commands.size());
+  WriteToken(os, binary, "<Commands>");
+  for (size_t c = 0; c < commands.size(); c++) {
+    commands[c].Write(os, binary);
+  }
```
danpovey: statements like
if (!binary) os << std::endl;
may help readability.

src/nnet3/nnet-computation.cc
```diff
@@ -413,6 +608,210 @@ void NnetComputation::Print(std::ostream &os, const Nnet &nnet) const {
...
+
+  WriteToken(os, binary, "<NeedModelDerivative>");
+  WriteBasicType(os, binary, need_model_derivative);
+}
```
danpovey: Don't forget an opening and closing tag <NnetComputation> and </NnetComputation>

xiaohui-zhang:
updated.

xiaohui-zhang:
done

src/base/io-funcs-inl.h
```diff
@@ -87,6 +88,111 @@ template<class T> inline void ReadBasicType(std::istream &is,
...
+  } else {
+    // focus here is on prettiness of text form rather than
+    // efficiency of reading-in.
+    // reading-in is dominated by low-level operations anyway:
+    // for efficiency use binary.
+    os << "[ ";
+    typename std::vector<std::pair<T, T> >::const_iterator iter = v.begin(), end = v.end();
+    for (; iter != end; ++iter) {
+      if (sizeof(T) == 1)
+        os << static_cast<int16>(iter->first) << ","
```
danpovey: it's probably a better practice to use single quotes for one-character strings in this context, not that it matters much.

src/base/io-funcs-test.cc
```diff
@@ -45,6 +45,16 @@ void UnitTestIo(bool binary) {
     std::vector<char> vec3;
     for (size_t i = 0; i < 10; i++) vec3.push_back(Rand()%100);
     WriteIntegerVector(outfile, binary, vec3);
+    std::vector<std::pair<int32, int32> > vec4;
```
danpovey: best not to leave the sizes fixed at 10 but make them random.  you might hit a special case somewhere (e.g. size = 1).

src/nnet3/nnet-common.cc
```diff
@@ -151,7 +152,123 @@ void ReadIndexVector(std::istream &is, bool binary,
...
+    } else {  // handle the general case less efficiently.
+      os.put(127);
+      WriteBasicType(os, binary, node_index);
+      WriteBasicType(os, binary, index.n);
+      WriteBasicType(os, binary, index.t);
+      WriteBasicType(os, binary, index.x);
+    }
+  } else {
+    Index last_index = vec[i-1].second;
+    if (index.n == last_index.n && index.x == last_index.x &&
```
danpovey: You are not testing for whether the node_index is the same.  And I don't see where your reading code reads in the node index in the same branch.  It makes me wonder what went wrong in your testing- were you not handling or checking the case where node indexes differ?

src/nnet3/nnet-compute-test.cc
```diff
@@ -27,6 +28,51 @@
...
+    
+  std::ostringstream os2, os3;
+  request->Write(os2, binary);
+  request2.Write(os3, binary);
+  
+  if (binary) {
+    KALDI_LOG << os2.str();
+    KALDI_LOG << original_output;
+    KALDI_ASSERT(os2.str() == original_output);
+    KALDI_ASSERT(os3.str() == original_output);
```
danpovey: since the object has, I think, an operator ==, why not use that as an additional test here?  Do the same in any similar situations.

src/nnet3/nnet-common-test.cc
```diff
@@ -58,6 +59,40 @@ void UnitTestIndexIo() {
   }
 }
 
+void UnitTestCindexIo() {
+  std::vector<Cindex> cindexes(RandInt(0, 10));
+
+  for (int32 i = 0; i < cindexes.size(); i++) {
+    if (i == 0 || RandInt(0, 1) == 0) {
```
danpovey: I don't see where  you are setting cindexes[i].first.

xiaohui-zhang:
done

danpovey:
There is a build failure and it looks like you have unresolved merge failures.

xiaohui-zhang:
done

src/base/io-funcs-inl.h
```diff
@@ -87,6 +88,111 @@ template<class T> inline void ReadBasicType(std::istream &is,
...
+  // Compile time assertion that this is not called with a wrong type.
+  KALDI_ASSERT_IS_INTEGER_TYPE(T);
+  if (binary) {
+    char sz = sizeof(T);  // this is currently just a check.
+    os.write(&sz, 1);
+    int32 vecsz = static_cast<int32>(v.size());
+    KALDI_ASSERT((size_t)vecsz == v.size());
+    os.write(reinterpret_cast<const char *>(&vecsz), sizeof(vecsz));
+    if (vecsz != 0) {
+      os.write(reinterpret_cast<const char *>(&(v[0])), sizeof(T)*vecsz*2);
```
danpovey: need space around operator*.

src/base/io-funcs-inl.h
```diff
@@ -87,6 +88,111 @@ template<class T> inline void ReadBasicType(std::istream &is,
...
+    }
+  } else {
+    // focus here is on prettiness of text form rather than
+    // efficiency of reading-in.
+    // reading-in is dominated by low-level operations anyway:
+    // for efficiency use binary.
+    os << "[ ";
+    typename std::vector<std::pair<T, T> >::const_iterator iter = v.begin(), end = v.end();
```
danpovey: probably past 80-char limit

xiaohui-zhang:
fixed

danpovey:
Samuel, I think it's best if you go ahead and add script-level support for this and we merge the code and script changes all together.  The idea is to speed up the nnet3 training by getting rid of the compilation stage.  You'd add a --read-cache option and a --write-cache option to NnetTrainerOptions.  Failure to open the cache file would produce a warning, not an error.  On each iteration, all jobs would have the option --read-computation-cache=dir/cache.$x, and job zero would have the option --write-computation-cache=dir/cache.$[$x+1]
and after each iteration you'd do rm $dir/cache.$x.
On mix-up iters you'd have to remove the cache prior to running the training (and to suppress the warning, probably remove the option) as the cached computation wouldn't be valid.

Dan

src/nnet3/nnet-common.cc
```diff
@@ -151,7 +152,175 @@ void ReadIndexVector(std::istream &is, bool binary,
...
+      signed char c = index.t - last_index.t;
+      os.put(c);
+    } else {  // handle the general case less efficiently.
+      os.put(127);
+      WriteBasicType(os, binary, index.n);
+      WriteBasicType(os, binary, index.t);
+      WriteBasicType(os, binary, index.x);
+    }
+  }
+  if (i == vec.size())
```
danpovey: These two lines are not needed and are confusing.

src/nnet3/nnet-common.cc
```diff
@@ -151,7 +152,175 @@ void ReadIndexVector(std::istream &is, bool binary,
...
+      if (c != 127)
+        KALDI_ERR << "Unexpected character " << c
+                  << " encountered while reading Cindex vector.";
+      ReadBasicType(is, binary, &(index.n));
+      ReadBasicType(is, binary, &(index.t));
+      ReadBasicType(is, binary, &(index.x));
+    }
+  } else {
+    Index &last_index = (*vec)[i-1].second;
+    if (std::abs(int(c)) < 125) {
```
danpovey: The way you are writing these things is very inefficient in the normal case- it gives almost no compression.
Remember in the normal computation, indexes are first sorted on t and then n.  So indexes differing only in n is actually the normal case (and they'll normally differ only by 1).  However, it's possible that some components may sort their input differently and sort instead on n first, so indexes differing only in n should also be supported to be written efficiently.  I suggest to allocate two different ranges of the character for each case (differing only in n vs. differing only in t).

danpovey:
Something that will be needed for this to work correctly is to add virtual Read and Write functions to the interface of ComponentPrecomputedIndexes, and a function ReadNew() for reading before the type is known.  This will later generate some problems on merging with the chain branch, where I have implemented a non-simple component; I'll fix it then.  For now just implement ReadNew() as a stub that fails.

danpovey:
closing as already merged to chain.

PR #434, opened 2016-01-05
Moving CuSparseMatrix/SparseMatrix out of CuVector/Vector

chenguoguo:
I named the function CopyToVec() in CuSparseMatrix to be consistent with that function in SparseMatrix. Tests passed.

### DIFF
--- a/src/cudamatrix/cu-sparse-matrix.cc
+++ b/src/cudamatrix/cu-sparse-matrix.cc
@@ -58,7 +58,7 @@ Real CuSparseMatrix<Real>::Sum() const {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) {
     CuVector<Real> sum_vec(this->NumElements(), kUndefined);
-    this->CopyToVec(&sum_vec);
+    this->CopyElementsToVec(&sum_vec);
     return sum_vec.Sum();
   } else
 #endif
@@ -72,7 +72,7 @@ Real CuSparseMatrix<Real>::FrobeniusNorm() const {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) {
     CuVector<Real> element_vec(this->NumElements(), kUndefined);
-    this->CopyToVec(&element_vec);
+    this->CopyElementsToVec(&element_vec);
     return element_vec.Norm(2);
   } else
 #endif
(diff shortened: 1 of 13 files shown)


danpovey:
OK, that change is good, but we can go further, and get rid of a bunch of unnecessary code.
Firstly, we don't need the OtherReal version of CopyToVec.  And (sorry), I think both of them should be changed to be called CopyElementsToVec, it's less ambiguous.
But also the special CUDA kernel is not needed.  You can use sizeof(MatrixElement<Real>) / sizeof(Real) as a fake stride, and &(elements[0].weight) as the start of the array, and make use of cublas_copy.  
I noticed that the functions CopyColFromVec and CopyColFromMat also use an unnecessary CUDA kernel-- they can also make use of cublas_copy.

chenguoguo:
I just pushed a commit, message below:

1. Changed CopyToVec in CuSparseMatrix/SparseMatrix to CopyElementsToVec,
and removed OtherReal version. Note that I didn't remove the OtherReal
version of CopyElementsToVec in SparseVector as that function is used by
other function;
2. Replaced unnecessary CUDA kernels used in CopyElementsToVec,
CopyColFromVec and CopyColFromMat with cublas_copy. Note,
cuda_copy_col_from_mat_df and cuda_copy_col_from_mat_fd were not replaced.

src/cudamatrix/cu-sparse-matrix.cc
```diff
@@ -57,7 +57,8 @@ template <typename Real>
 Real CuSparseMatrix<Real>::Sum() const {
 #if HAVE_CUDA == 1
```
danpovey: Please rebase against the latest master and test.  there should be lines here:
  if (NumElements() == 0)
    return 0.0;
Please test by logging into a GPU machine and going 'make test' in src/ - this will test more thoroughly.

src/cudamatrix/cu-sparse-matrix.cc
```diff
@@ -70,7 +71,8 @@ template <typename Real>
 Real CuSparseMatrix<Real>::FrobeniusNorm() const {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) {
-    CuVector<Real> element_vec(*this);
+    CuVector<Real> element_vec(this->NumElements(), kUndefined);
+    this->CopyElementsToVec(&element_vec);
     return element_vec.Norm(2);
```
danpovey: This code is not right, you need account for the fact that the matrix's size is different from the number of nonzero elements.  Otherwise it's not really the frobenius norm.

chenguoguo: We are computing sqrt(sum_{i=1}^N (x_i^2)) as the Frobenius norm right? Why
would the number of elements matter? Perhaps I misunderstood
CuVector.Norm()?

Guoguo

danpovey: oh sorry, I forgot the definition.

chenguoguo: OK I merged master and unit tests passed.

Guoguo

danpovey:
Thanks.  In the longer term it would be better to create versions of cublas_copy with our own kernels, that work for different types, which is more generic than cuda_copy_col_from_mat_fd  and the like.  Could you please create an issue to track this?  I'll merge this now.

src/cudamatrix/cu-sparse-matrix.cc
```diff
@@ -202,6 +206,27 @@ void CuSparseMatrix<double>::CopyToSmat(SparseMatrix<float> *smat) const;
 template
 void CuSparseMatrix<double>::CopyToSmat(SparseMatrix<double> *smat) const;
 
+template <typename Real>
+void CuSparseMatrix<Real>::CopyElementsToVec(CuVector<Real> *vec) const {
```
danpovey: Oops.  I just noticed, this should be CuVectorBase, not CuVector.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -95,6 +95,10 @@ class CuSparseMatrix {
   template <typename OtherReal>
   void CopyToSmat(SparseMatrix<OtherReal> *smat) const;
 
+  /// Copy elements to CuVector. It is the caller's responsibility to resize
+  /// <*vec>.
+  void CopyElementsToVec(CuVector<Real> *vec) const;
```
danpovey: should be CuVectorBase

chenguoguo: That should be changed yes, I fixed that.

Also created an issue for cublas_copy.

Guoguo

danpovey:
I'll wait till tests finish before merging.  Thanks.

PR #437, opened 2016-01-06
fixed inline const to enum in pitch-functions.* to avoid some compila…

pegahgh:
…tion issue using visual studio


danpovey:
Thanks!  Merging.  I looked into this issue a long time ago and Visual Studio is definitely against the C++ standard, but it's easy to fix their issue here, so we'll fix it.

Issue #441, opened 2016-01-08
Speedup in RepeatedAffineComponent


danpovey:
@galv, if you have time you can do this too [in the 'chain' branch].

It looks like the function AddMatMatBatched, and the underlying sgemmBatched, is not as optimized as it could be.  This makes the RepeatedAffineComponent slower than it should be, which is affecting the speed of some of my experiments.

There is a way we can optimize this not too painfully, by turning everything into a single matrix multiply.   This will make the code a lot simpler too.  Basically it will require two temporary matrices, one for the input and one for the output [or one for the input-derivative and one for the output-derivative, in backprop].  These matrices will have a lot more rows than the original matrices-- the number of rows will be greater by a factor of the group-size.  To avoid wasting too much memory due to the padding of the stride, you might want to experiment with making them both sub-matrices of a single larger matrix with the same number of rows, but a larger number of columns- or at least make it easy in the code to try both, since experimentation may be easier for me.  [Note: if the absolute amount of memory becomes too large we can reduce with the 'max-rows-process' option of CompositeComponent].

There is a function called Tensor3dCopy, which is defined in ctc/cctc-tombstone.cc in the following commit on my personal repo (e.g. accessible if you are currently on the jesus branch), which you will find useful in doing the data copying.  You can access this as follows:
git checkout c83118672dc5e1ddbdece094fcb5b1fe5422d78d~1 -b tensor3dcopy

This can be moved to cu-matrix.{h,cc}.
There is some CUDA code you'll need for this, which is located in cctc-kernels.cu [can be moved to cu-kernels.h], and a declaration located in cctc-kernels-ansi.h [move to cu-kernels-ansi.h] and in cctc-kernels.h [move to cu-kernels.h].

galv:
I've seen a trick like this used in implementing LSTMs actually, for calculating each gate all at once, in a single gemm call. I also read the CUBLAS docs on the batched gemm call; I agree it's wrong for this component. I had no idea how large the dimensions would be. Sorry. Anyway:

First of all, so we're on the same page:
```
group-size == input_dim / _num_repeats
```
right? Here, **input_dim** is the input feature dimension to the component itself. **output_dim**, which I introduce later similar denote the output feature dimension of this component (the hidden feature dimension). **N** is the number of elements in a batch. In the following, "*" denotes multiplication, while "X" denotes what we typically mean when we say a matrix M X N (M-by-N).

[...]

PR #443, opened 2016-01-11
Fast repeated affine component

galv:
For #441

Natural Gradient is not implemented. Tell me if that is needed; I'll have to take some time to understand the method.

The name "fast-mode" is a little optimistic. I did not test the relative speed of this yet. I've been distracted from building the chain model for TEDLIUM, and figured you (Dan) would want this ASAP.

### DIFF
--- a/src/cudamatrix/cu-kernels-ansi.h
+++ b/src/cudamatrix/cu-kernels-ansi.h
@@ -95,12 +95,6 @@ void cudaF_add_vec_to_cols(dim3 Gr, dim3 Bl, float alpha, const float *col, floa
 void cudaF_add_vec_to_rows(dim3 Gr, dim3 Bl, float alpha, const float *row, float beta, float *dst, MatrixDim d);
 void cudaF_add_mat_diag_vec(dim3 Gr, dim3 Bl, float alpha, float *mat, MatrixDim mat_dim, const float *mat2, int mat2_row_stride, int mat2_col_stride, const float *vec, float beta);
 void cudaF_add_mat_mat_elements(dim3 Gr, dim3 Bl, float *data, const float *srcA_data, const float *srcB_data, MatrixDim dim, int srcA_stride, int srcB_stride, float alpha, float beta);
-/// 3 tensors are packed into matrices.
-void cudaF_rearrange_3d_tensor(dim3 Gr, dim3 Bl, int32_cuda xdim,
-                               int32_cuda xstride_in, int32_cuda ystride_in,
-                               int32_cuda zstride_in, int32_cuda xstride_out,
-                               int32_cuda ystride_out, int32_cuda zstride_out,
-                               const float *src, float *dst);
 
 /*
  * CuVector
@@ -184,7 +178,14 @@ void cudaF_matrix_lookup(dim3 Gr, dim3 Bl, const float *data, MatrixDim dim,
 void cudaF_equal_element_mask(dim3 Gr, dim3 Bl, const float *mat1,
                               const float *mat2, float *mask, MatrixDim mat1_dim,
                               int mat2_stride, int mask_stride);
-  
+/*
+ * 3-tensor. Both CuVector and CuMatrix can be interpretted as 3-tensors.
+ */
+void cudaF_rearrange_3d_tensor(dim3 Gr, dim3 Bl, int32_cuda xdim,
+                               int32_cuda xstride_in, int32_cuda ystride_in,
+                               int32_cuda zstride_in, int32_cuda xstride_out,
+                               int32_cuda ystride_out, int32_cuda zstride_out,
+                               const float *src, float *dst);
 /*********************************************************
  * double CUDA kernel calls
  */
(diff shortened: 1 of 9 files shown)


danpovey:
Thanks a lot!  Will review later today or to-morrow.  I'll do the natural
gradient stuff if needed.
Dan

danpovey:
I think you forgot to add the fast_mode_ to Read() and Write().  Also make sure they are copied in the copy-constructor (needed in Copy()).

galv:
The copy constructor does in fact copy fast_mode_. See: [this line](https://github.com/galv/kaldi/commit/71ac1300b285a7a7fb484a6ec3b02e5972ab3232#diff-9dce7a91d342f07a96b9d469c8f075b1R941).

You're right on Read() and Write(), though. Sorry I let this slip by.

galv:
Also: WriteUpdatableCommon() and ReadUpdatableCommon() both handle serializing is_gradient_, but you can see all the simple components' serialization methods reading and writing is_gradient_ themselves. Eck. Is it like this for backwards compatibility with previously serialized models?

danpovey:
Thanks- you'll need back-compatibility code for Read().  Please make
fast-mode default to true; this will make my life easier in testing the
speed.
Dan

danpovey:
BTW,  the function ReadToken will be used in the back-compatibility code.

galv:
Great, thanks. I was about to say that ExpectToken would always print an error message even i the accompanying exception was caught.

galv:
I've pushed a commit for Read and Write. The current test code makes it nontrivial to add a test for the backwards compatibility mode (which *does* set fast_mode_ to true), but I have thought about it for a while and am fairly positive it works.

danpovey:
OK thanks.  I'll test the speed before I merge.

danpovey:
@galv, it may be a while before I merge this, if at all... testing revealed that it was very flow.  I changed the order of the x, y and z indexes in CopyTensor3d and this helped, but it was still slower than the baseline, with too much time being taken in CopyTensor3d.  This is going to have to wait until I can look at it again.  I'll leave it up though, until I decide what to do about it.

galv:
Unfortunately, I doubt anything else can be done at this point. CUBLAS
really can't be optimized further obviously, and Copy3dTensor must exist
because of the existence of padding in CuMatrix objects.

On Tue, Jan 12, 2016 at 11:00 PM, Daniel Povey <notifications@github.com>
wrote:

> @galv <https://github.com/galv>, it may be a while before I merge this,
> if at all... testing revealed that it was very flow. I changed the order of
> the x, y and z indexes in CopyTensor3d and this helped, but it was still
> slower than the baseline, with too much time being taken in CopyTensor3d.
> This is going to have to wait until I can look at it again. I'll leave it
> up though, until I decide what to do about it.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/443#issuecomment-171156141>.
>



-- 
Daniel Galvez
danieltgalvez.com

danpovey:
BTW, just for the record, it looks like AddMat took about 35% less time in my configuration than AddMatMatBatched- or maybe even a bit faster than that.  I'm considering adding a flag that says that a component prefers its input and output matrices to be allocated stride == dim.  I have to think about this.

danpovey:
@galv -- I have a new plan, and if you want to help with this, you can:
The new plan is to modify that component to detect when the input and output matrices' stride is equal to the num-rows, and if so to activate a different, faster code path.  Note that this is not hard to do also in the natural gradient update-- the actual natural gradient part of the update is segregated (see the block that says if (!is_gradient_))...  
Other code would have to be written to set and interpret flags to make sure the stride has no gaps, and I would do that.

galv:
I believe you mean "stride is equal to the **num-cols**".

Anyway, I am aware that this has some complications, but I can certainly redo this as you suggest. However, I can't work on it until after the weekend.

I took the time to read about striding issues in CUDA, and I think what you propose will in fact finally allow us to get a speed up. My understanding is that accessing un-aligned memory is not an issue on modern nvidia GPUs. See the figure under "Misaligned Data Accesses" [here](http://devblogs.nvidia.com/parallelforall/how-access-global-memory-efficiently-cuda-c-kernels/).

galv:
Are you still interested in this speedup? I can plan out what needs to be done, but I would have to ask a few questions. e.g., what the best way to allocate a CuMatrix without padding would be (e.g., new constructor, or optional boolean argument to existing constructor).

danpovey:
@dgalvez, if you are going to add the option to  initialize  matrix with
stride==num-cols, please do so ASAP as a separate commit.  I need this for
something else.

galv:
I haven't worked on this because I was of the impression that allocating the matrix with stride==num-cols would require modifying the CuMatrix code (in order to expose a new constructor which uses cudaMalloc instead of cudaMallocPitch). I'm not familiar with the code; it'll be a few days.

If you know of an easier way to implement this, then let me know.

danpovey:
It's OK, I'll do it myself, it will be quick for me.  (to add the
constructor)

galv:
okay. Thank you.

danpovey:
I'm closing this pull request as I've done it myself (i'm testing it, it is a little faster; it's mixed in with some other stuff and I'll merge it soon).

galv:
+1

On Sat, Feb 6, 2016 at 12:05 AM, Daniel Povey <notifications@github.com>
wrote:

> Closed #443 <https://github.com/kaldi-asr/kaldi/pull/443>.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/443#event-541577417>.
>



-- 
Daniel Galvez
danieltgalvez.com

PR #444, opened 2016-01-11
Windows line endings

Timmmm:
There's an issue with github if you try to download the repository as a zip (rather than cloning it with git), then it runs `git archive` on the repository.

Unfortunately they run in on a Linux machine, and git changes all the line endings in text files to their native format (LF on Linux). Patch requires CRLF endings on Windows so it fails.

The change to `.gitattributes` stops git mangling any patch files. The change to the .patch file makes the line endings CRLF. Other files may require changing to CRLF.

There's no really nice way to fix this in general, since when `git archive` is run it doesn't know if the target is Windows or Linux, and Windows `patch` always requires CRLF, whereas Linux `patch` requires the patch file to match whatever it is patching. I'd suggest keeping separate patch files for Windows and Linux (which there is already in this case fortunately).


jtrmal:
Timmmm, first of thanks for the PR.
@kkm000 , can you please verify the suggested changes? The priority should be preventing the linux checkout from breaking -- if there would be some issue, I'd rather keep windows checkouts/archives broken.

jtrmal:
Actually, I reviewed the patch again and I think it's good as it it (as each of the commits is fairly limited in scope and well isolated).
y.

danpovey:
I really don't like the idea of changing to CRLF endings in any of the patch files to keep Windows happy.  Remember 95% or more of use of Kaldi is on UNIX.   I'd prefer to modify the Windows scripts, e.g. to call unix2dos on those files (if that program exists on Windows).

Timmmm:
Apparently there is [a way to convert line endings](http://stackoverflow.com/a/27844521/265521) on the Windows command line, but it seems to me that it would be nice to avoid adding an extra step to the installation process.

Fortunately in this case the patch I changed (`openfstwin-1.3.4.patch`) is Windows-specific so it should never be used on Unix. All the other patches should still have Unix line endings.

danpovey:
Oh, you're right.  I didn't notice that there was 'win' in the filename.
Yenda, if you don't see any problems with this you can merge it.
Dan

jtrmal:
OK, merging.

PR #447, opened 2016-01-12
Changes from pegah for tensormultiply in matrix/cudamatrix

vijayaditya:
This is the tensor multiply operation implemented by @pegahgh . This will be used for some new components in nnet3.

### DIFF
--- a/src/cudamatrix/cu-kernels-ansi.h
+++ b/src/cudamatrix/cu-kernels-ansi.h
@@ -4,6 +4,7 @@
 //                2013  Johns Hopkins University (author: Daniel Povey)
 //                2013  Hainan Xu    
 //                2013  Xiaohui Zhang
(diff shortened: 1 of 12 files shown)


danpovey:
I don't want to merge this until i'm convinced that we need it.  What
exactly do you want to do with this, and are you sure that AddMatMatBatched
is not the more suitable solution?

vijayaditya:
Using an AddMatMatBatched call would be easier, but it would require a permutation of the columns as this component will be used on spliced output (i.e., output of Append descriptor).

danpovey:
You could use a temporary inside the propagate/backprop code.  I'd rather
rely on the CUBLAS internals than custom ones, as they'll improve them in
future.  You should at least compare the speed, before we add all that code
to the repo.
Dan

vijayaditya:
OK.

PR #448, opened 2016-01-13
Windows docs

Timmmm:
I've reformatted the Windows INSTALL file so it is markdown (nicely rendered by Github):

https://github.com/Timmmm/kaldi/blob/windows_docs/windows/INSTALL.md

I had to rename it to .md so Github will render it. I also added a note about using --enable-openblas.


jtrmal:
Thanks. The issue I see with the rendering is that it does not show the alternative paths properly -- in the original readme, openblas and MKL are alternatives -- there is no distinction of this kind in the rendered markdown 
y

danpovey:
Yenda, please merge this (or not) as you see fit...

Timmmm:
I agree about the alternatives. To be honest I don't think it was that clear in the original, but anyway - give me a few days and I'll improve this patch so it is explicit.

Timmmm:
Hopefully this is ok. Thanks for merging my other pull requests by the way!

danpovey:
@jtrmal, please look into this one to..

jtrmal:
That looks fine. Thanks. Merging. y.

Issue #422, opened 2015-12-29
Rewrite ARPA-FST compiler


kkm000:
Discussion: https://groups.google.com/d/topic/kaldi-developers/m_xloj88tsk/discussion

danpovey:
Just a note: Guoguo wrote some arpa-reading code in the code that creates the 'const-arpa' object.  That code is maybe not perfect, but it's a lot better than the old arpa2fst code.  Some of this could be re-used, and regardless, we should try to keep them in sync as much as possible.

kkm000:
I think I'll do two PRs for this, to make reviewing easier: one factoring out the ARPA file parser into a separate object from const-arpa, and another with the new compiler using the parser, and tests.

danpovey:
OK, cool.

On Mon, Jan 11, 2016 at 4:22 PM, Kirill Katsnelson <notifications@github.com

kkm000:
@danpovey: What do you think I should keep of kadli-lm and kaldi-lm-table? There is a class `LangModelFst` that does not seem completed to me. The intention was to make it read and write grammars in different formats, namely ARPA, some textual representation of WFST and something else defined in an enum but never mentioned in the code. Do you want to keep that, or just care about the code needed to compile ARPA file to an FST, i. e. what the arpa2fst does, and this is it?

Also, I understand IRST LM support is not needed any more, correct?

danpovey:
No, that stuff is very old and there is no plan to continue that- we have
other mechanisms now.  Just implement arpa2fst.
dan


On Mon, Jan 11, 2016 at 8:44 PM, Kirill Katsnelson <notifications@github.com

kkm000:
Another question. It seems trivially easy at this point to extend arpa2fst to optionally take `words.txt` and thus eliminate a separate `utils/find_arpa_oovs.pl` pass (and not call remove_oovs.pl in the next pass), and also essentially do `utils/eps2disambig.pl | utils/s2eps.pl` during the FST construction. Do you think it is a good idea to add options to do just that? I mean, instead of 

```
gunzip -c $lm | utils/find_arpa_oovs.pl $out_dir/words.txt \
  > $out_dir/oovs_${lm_base}.txt

gunzip -c $lm \
  | egrep -v '<s> <s>|</s> <s>|</s> </s>' \
  | arpa2fst - | fstprint \
  | utils/remove_oovs.pl $out_dir/oovs_${lm_base}.txt \
[...]

PR #458, opened 2016-01-19
Separated ARPA parsing from const LM construction

kkm000:
This is part 1 of #422.

There are a couple oddities I noted  First, Guoguo's parser handles the case when the count of ngram of a given order is 0 (e. g. "ngram 2=0" straight in ARPA header). To me, it looks like an error in file at best, as probabilities of longer n-grams are naturally zeros then, contrary to what the file would indicate. Also, later the const-ARPA builder ASSERTs that a shorter prefix n-gram is seen (e. g. if looking at "a b c", assert that "a b" was seen and recorded).

Which way of the two it should go? I think that the second behavior is correct, only the message needs to be better that the generic ASSERT on a cryptic condition.

Also, while at it, a question: why does not Kaldi use GoogleTest/GoogleMock for testing? Just never came to it, or is there a stronger reason?


danpovey:
> This is part 1 of #422 <https://github.com/kaldi-asr/kaldi/issues/422>.
>
> There are a couple oddities I noted First, Guoguo's parser handles the
> case when the count of ngram of a given order is 0 (e. g. "ngram 2=0"
> straight in ARPA header). To me, it looks like an error in file at best, as
> probabilities of longer n-grams are naturally zeros then, contrary to what
> the file would indicate.
>

But what if the n-gram order is the last one or later n-gram orders also
have zero counts of ngram?  Perhaps it could happen from some sources that
all n-grams of the last order or orders would be pruned away.


> Also, later the const-ARPA builder ASSERTs that a shorter prefix n-gram is
> seen (e. g. if looking at "a b c", assert that "a b" was seen and recorded).
>

This is required because of the need to assign a backoff weight, i.e. if
there is a history "a b" it should have a backoff weight recorded which the
ARPA stores with n-gram "a b".

> Which way of the two it should go? I think that the second behavior is
correct, only the message needs to be better that the generic ASSERT on a
cryptic condition.

Sure, there should be a better error message.

> Also, while at it, a question: why does not Kaldi use
> GoogleTest/GoogleMock for testing? Just never came to it, or is there a
> stronger reason?
>
I'm not sure that I have ever heard of these.  We didn't have any actual
Google employees on the team.


Dan

kkm000:
I never seen such zeroed-out highest orders in the wild. I was getting smaller order models from SRILM, but it never added literal 0 counts for higher orders, just simply did not print them. This would not be right as it would technically allowed specifying backoff-weights with highest order n-grams. What I mean is e. g. if we have

```
\data\
ngram 1=3
ngram 2=3

\1-grams:
. . . .
\2-grams:
-1.5	x y -5
```

the parser will correctly complain about the -5 token being the backoff weight on a highest-orer ngram. But if we also added `ngram 3=0` to the header, what would that mean? Will 3 be the maximum order, and the token for back-off weight legal? It's a dark-gray area, given the absence of formal spec for ARPA files. I'd rather think that zero counts in files smell fishy--unless you really know that such files exist.

> I'm not sure that I have ever heard of these.

https://github.com/google/googletest is a very nice and simple testing framework. It was released as open source quite a while ago (although I have not been exposed to it before Google). Writing mocks, just what I had to do to test the parser, is much easier with it. I'll keep it on the list of possible improvements to Kaldi, but with a lower priority.

danpovey:
>
> I never seen such zeroed-out highest orders in the wild. I was getting
> smaller order models from SRILM, but it never added literal 0 counts for
> higher orders, just simply did not print them. This would not be right as
> it would technically allowed specifying backoff-weights with highest order
> n-grams. What I mean is e. g. if we have
>
> \data\
> ngram 1=3
> ngram 2=3
>
> \1-grams:
> . . . .
> \2-grams:
> -1.5    x y -5
>
> the parser will correctly complain about the -5 token being the backoff
> weight on a highest-orer ngram. But if we also added ngram 3=0 to the
> header, what would that mean? Will 3 be the maximum order, and the token
> for back-off weight legal? It's a dark-gray area, given the absence of
> formal spec for ARPA files. I'd rather think that zero counts in files
> smell fishy--unless you really know that such files exist.
>

I don't know for sure that they exist but the specification doesn't say and
it seems like they could possibly happen in weird situations, so I'd rather
any message just be a warning.
Dan

kkm000:
I think then this is how I should do it then:

1. If the highest order (or a few highest) have zero counts, delete them (i. e. if a 3-gram model has "ngram 3=0", treat is as a 2-gram model) and print a warning.
2. If there is a 0-count in between, return it from the parser. (???)
2a) const-arpa converter is already guaranteed to die on such a file, since it needs the prefix model to exist.
2b) ARPA-to-FST (the compile code I am going to tackle next) ignores "a b c" if "a b" not found. I should add  a warning as well in this case; currently I am discarding it silently. A 0-count level will effectively cause all higher levels to be ignored.

Regarding 2)--is there a way to sensibly interpret e. g.

```
ngram 1=10
ngram 2=0
ngram 3=20
```
?

danpovey:
I think you are over-thinking this.  Just print a warning any time you see
something of the form ngram n=0 and leave it at that.  we likely won't see
this case, and there's no point adding too many extra lines of code to
handle it.
Dan

danpovey:
... the point is that those other checks will already fail for invalid
models of this type, so there is no point adding further checks at this
level.

kkm000:
Makes sense, thanks.

kkm000:
Added the warning for 0 n-grams, and added a better error message if parent model was not found.

src/lm/arpa-file-parser.h
```diff
@@ -0,0 +1,123 @@
...
+  /// Constructs the parser with the given settings and optional symbol table.
+  /// If symbol table is provided, then the file should contain text n-grams,
+  /// and the words are mapped to symbols through it. bos_symbol and
+  /// eos_symbol in the settings structure must be valid symbols in the table,
+  /// and so must be unk_symbol if provided. The table is not owned by the
+  /// parser, but may be augmented, if oov_handling is set to kAddToSymbols.
+  /// If symbol table is a null pointer, the file should contain integer
+  /// symbol values, and oov_handling has no effect. bos_symbol and eos_symbol
+  /// must be valid symbols still.
+  ArpaFileParser(ArpaParseSettings settings, fst::SymbolTable* symbols);
```
danpovey: We generally call options classes ArpaParseOptions etc. --better to be consistent.

src/lm/arpa-file-parser.h
```diff
@@ -0,0 +1,123 @@
...
+ protected:
+  /// Override called before reading starts. This is the point to prepare
+  /// any state in the derived class.
+  virtual void ReadStarted() { }
+
+  /// Override function called to signal that ARPA header with the expected
+  /// number of n-grams has been read, and ngram_counts() is now valid.
+  virtual void HeaderAvailable() { }
+
+  /// Pure override that must beimplemented to process current n-gram. The
```
danpovey: beimplemented -> be implemented

src/lm/arpa-file-parser.h
```diff
@@ -0,0 +1,123 @@
...
+*/
+struct NGram {
+  NGram() : logprob(0.0), backoff(0.0) { }
+  std::vector<int32> words;  ///< Symbols in LTR order.
+  float logprob;             ///< Log-prob of the n-gram.
+  float backoff;             ///< log-backoff weight of the n-gram.
+};
+
+/**
+    ArpaFileParser is an abstract base class for ARPA LM file conversion.
```
danpovey: It would be nice to point to a place where this is used for an example of how to use it.

src/lm/arpa-file-parser.h
```diff
@@ -0,0 +1,123 @@
...
+  /// Pure override that must beimplemented to process current n-gram. The
+  /// n-grams are sent in the file order, which guarantees that all
+  /// (k-1)-grams are processed before the first k-gram is.
+  virtual void ConsumeNGram(const NGram&) = 0;
+
+  /// Override function called after the last n-gram has been consumed.
+  virtual void ReadComplete() { }
+
+  /// Read-only access to symbol table.
+  fst::SymbolTable* symbols() const { return symbols_; }
```
danpovey: Kirill-- I'm afraid we deviate slightly from the Google style guide in that accessor members are written LikeThis().  [motivation: the interface reveals too much about the implementation by using this naming convention and makes refactoring harder.]  Also should this returned pointer be const?

danpovey:
@chenguoguo, could you please help test this, and make sure that your const-arpa-lm code still has the same behavior after applying his patch?

kkm000:
> make sure that your const-arpa-lm code still has the same behavior after applying his patch?

Oh, by the way, I should have mentioned this. I verified the generated `.carpa` file is bitwise same as before, down to the precision of log base conversion. For the test, I replaced [the use of a constant](https://github.com/kaldi-asr/kaldi/pull/458/files#diff-10f840cccab787d139b2b4314192fcdaR174} with `Log(10.0f)` which was what the original code had, and built the large unpruned 3-gram file in librispeech example, ensuring it was same by `cmp`ing it to the one generated by the original recipe and code.

With the literal constant, the file is (bitwise) different but perhaps in the least significant mantissa bit.

kkm000:
@chenguoguo: ...but if you want to double-check (which would never hurt indeed, thanks!), you may pull the pre-merged pull request into your local repo from GitHub with this trick:

```
$ git fetch origin +refs/pull/458/merge:
$ git checkout -qf FETCH_HEAD
```
where `origin` is the name of the main Kaldi remote.

danpovey:
OK, so if you're saying you've tested it yourself this way, I can merge
without Guoguo checking.  Let me know when I should check the code again.



On Fri, Jan 22, 2016 at 1:35 AM, Kirill Katsnelson <notifications@github.com

chenguoguo:
I quickly scanned the code change and I think it's generally good. I didn't
test it since Kirill has already done the testing.

Guogo

kkm000:
@chenguoguo, @danpovey: Thanks for the review! Suggested changes done, PTAL.

danpovey:
Very nice testing code.  Thanks!  Merging.

PR #460, opened 2016-01-20
Implement the 3d maxpooling component.

tomkocse:


danpovey:
obviously vijay will do the review of this.

src/nnet3/nnet-simple-component.cc
```diff
@@ -4261,6 +4254,256 @@ std::string MaxpoolingComponent::Info() const {
...
+      // get output buffer of the pool
+      int32 q = qy + qx * num_pools_y;
+      CuSubMatrix<BaseFloat> pool(patches_zyx.ColRange(q * input_z_dim_, input_z_dim_));
+      pool.Set(-1e20); // reset a large negative value
+      int32 offset_x = qx * pool_x_step_;
+      int32 offset_y = qy * pool_y_step_;
+      for (int32 px = offset_x; px < (pool_x_size_ + offset_x); px++) {
+        for (int32 py = offset_y; py < (pool_y_size_ + offset_y); py++) {
+          int32 p = py + px * input_y_dim_;
+          pool.Max(in.ColRange(p * input_z_dim_, input_z_dim_));
```
danpovey: er... you seem to have a very tight inner loop here: the operation nested is inside 3 loops!  I have a hard time believing this could be efficient on a GPU.

tomkocse: What about doing the max pooling along x and y axes separately ? 
That can reduce the number of nested loops.

danpovey: I don't really understand the details of this.  I'll let Vijay comment.
What you have done may be fine as a temporary measure to test whether this
even works.
Dan

galv: I believe what Dan is trying to say is that you shouldn't be calling pool.Max in a loop like this. You don't exploit the power of the GPU by serially calling CUDA kernels. (At least, I assume pool.Max calls a CUDA kernel.)

This problem is trivially parallelizable in that each output can be computed independently of all of the others. (Each output element is the maximum of a different patch of the input 3-tensor.) You may have to write a new kernel. Without getting too deep into this, torch's cuda neural net implementation has a [3d pooling component](https://github.com/torch/cunn/blob/master/VolumetricMaxPooling.cu). Perhaps it can help you, though the fact that Kaldi has to vectorize its 3-tensors is bound to cause adapting that code to have complications.

tomkocse: Solving the problem in a GPU parallelizable manner requires reshaping and packing the input into different patches by CPU. Will that be more expensive in time ?
Please correct me if i am having a wrong concept on this.

naxingyu: That's how it was implemented in nnet/nnet2. It is less expensive than 
looped calls of the same kernel.

在 2016/1/20 14:15, tomkocse 写道:

vijayaditya: Converting the input matrix to patches is already done in convolution
component. You have a look at this. BTW the conversion to patches is done
on the gpu.

tomkocse: OK, i will rewrite the propagate and backpropagate functions.

danpovey:
BTW, I want to  make a comment just for the future, that implementing convolution using a simple component may not be the ideal way to do it.  IMO it makes more sense to do it using a general component, and for the time axis and possibly also the 'extra' axis (deltas, etc.) use the different rows of the input, via the t and x dimensions.

danpovey:
Also, I think the convolutional stuff should be moved to nnet-convolutional-component.{h,cc}.  nnet-simple-component.{h,cc} is getting too long.

vijayaditya:
Good to know that there is possibility of preserving time axis through the
convolution component. It was a concern for me before. I will do the
conversion.

Vijay

danpovey:
It won't be trivial to do this with a GeneralComponent- you'll have to
create the PrecomputedIndexes to keep track of various things.  I suggest
you wait a few days as I am working on some other instances of
GeneralComponent and once you see what I'm doing it will be easier to see
how to do it.

vijayaditya:
OK.

Vijay

vijayaditya:
please squash your commits.

tomkocse:
The recent two commits are merged.

src/nnet3/nnet-component-itf.cc
```diff
@@ -96,6 +96,8 @@ Component* Component::NewComponentOfType(const std::string &component_type) {
     ans = new ConvolutionComponent();
   } else if (component_type == "MaxpoolingComponent") {
     ans = new MaxpoolingComponent();
+  } else if (component_type == "Maxpooling3dComponent") {
```
vijayaditya: @danpovey  I don't know of any recipes which are using the MaxpoolingComponent in nnet3. We are already printing deprecation warning for Convolution1DComponent. How about removing both Convolution1DComponent and MaxpoolingComponent in this commit ? I think it might be better to use the name MaxpoolingComponent instead of Maxpooling3dComponent for the new component.

danpovey: OK, fine.
Is maxpooling3d efficient for the (possibly more common) 1d and 2d cases?
Dan

vijayaditya: Still going through the new commit, will suggest changes if it is not optimal for 1d/2d cases.

src/nnet3/nnet-simple-component.cc
```diff
@@ -4261,6 +4254,273 @@ std::string MaxpoolingComponent::Info() const {
...
+// aquire output dim
+int32 Maxpooling3dComponent::OutputDim() const {
+  int32 num_pools_x = 1 + (input_x_dim_ - pool_x_size_) / pool_x_step_;
+  int32 num_pools_y = 1 + (input_y_dim_ - pool_y_size_) / pool_y_step_;
+  int32 num_pools_z = 1 + (input_z_dim_ - pool_z_size_) / pool_z_step_;
+  return num_pools_x * num_pools_y * num_pools_z;
+}
+
+// initialize the component using hyperparameters
+void Maxpooling3dComponent::Init(int32 input_x_dim, int32 input_y_dim, int32 input_z_dim,
```
vijayaditya: 80 character limit violated

src/nnet3/nnet-simple-component.cc
```diff
@@ -4261,6 +4254,273 @@ std::string MaxpoolingComponent::Info() const {
...
+    }
+  }
+  CuArray<int32> cu_cols(column_map);
+  patches->CopyCols(in, cu_cols);
+}
+
+/*
+  3d max pooling propagate function
+  It is assumed that each row of the input matrix
+  is a vectorized 3D-tensor in zxy format.
```
vijayaditya: is a vectorized 3D-tensor of TensorVectorizationType zxy (see ConvolutionComponent description in nnet-simple-component.h for more information).

src/nnet3/nnet-simple-component.cc
```diff
@@ -4261,6 +4254,273 @@ std::string MaxpoolingComponent::Info() const {
...
+  int32 num_pools_y = 1 + (input_y_dim_ - pool_y_size_) / pool_y_step_;
+  int32 num_pools_z = 1 + (input_z_dim_ - pool_z_size_) / pool_z_step_;
+
+  int32 num_pools = num_pools_x * num_pools_y * num_pools_z;
+  int32 pool_size = pool_x_size_ * pool_y_size_ * pool_z_size_;
+  CuMatrix<BaseFloat> patches(num_frames, num_pools * pool_size, kUndefined);
+  InputToInputPatches(in, &patches);
+
+  out->Set(-1e20); // reset a large negative value
+  for (int32 q = 0; q < pool_size; q++)
```
vijayaditya: @tomkocse  could you check the time spent in the for loop. A MaxBatched(), similar to AddMatMatBatched() might be useful. There is no such kernel right now.

src/nnet3/nnet-simple-component.cc
```diff
@@ -4261,6 +4254,273 @@ std::string MaxpoolingComponent::Info() const {
...
+                                   const CuMatrixBase<BaseFloat> &out_value,
+                                   const CuMatrixBase<BaseFloat> &out_deriv,
+                                   Component *, // to_update,
+                                   CuMatrixBase<BaseFloat> *in_deriv) const {
+  int32 num_frames = in_value.NumRows();
+  int32 num_pools_x = 1 + (input_x_dim_ - pool_x_size_) / pool_x_step_;
+  int32 num_pools_y = 1 + (input_y_dim_ - pool_y_size_) / pool_y_step_;
+  int32 num_pools_z = 1 + (input_z_dim_ - pool_z_size_) / pool_z_step_;
+
+  int32 num_pools = num_pools_x * num_pools_y * num_pools_z;
```
vijayaditya: All the computation in this function should be inside the <code>if (in_deriv) {}</code> block.

danpovey: .. I suppose doing
if (!in_deriv)
  return;
would accomplish the same without increasing the indentation.

src/nnet3/nnet-simple-component.cc
```diff
@@ -4261,6 +4254,273 @@ std::string MaxpoolingComponent::Info() const {
...
+  int32 pool_size = pool_x_size_ * pool_y_size_ * pool_z_size_;
+  CuMatrix<BaseFloat> in_value_patches(num_frames, num_pools * pool_size, kUndefined);
+  CuMatrix<BaseFloat> in_deriv_patches(num_frames, num_pools * pool_size, kSetZero);
+  InputToInputPatches(in_value, &in_value_patches);
+
+  for (int32 q = 0; q < pool_size; q++) {
+    // zero-out mask
+    CuMatrix<BaseFloat> mask;
+    out_value.EqualElementMask(in_value_patches.ColRange(q * num_pools, num_pools), &mask);
+    mask.MulElements(out_deriv);
```
vijayaditya: I think maintaining the in_deriv_patches is waste of memory. Most of it is zero, except for the column corresponding to the max_element. Directly copy it derivative to the in_deriv matrix.

src/nnet3/nnet-simple-component.h
```diff
@@ -1643,6 +1643,109 @@ class MaxpoolingComponent: public Component {
 };
 
 /**
+ * Maxpooling3dComponent implements 3d-maxpooling.
+ * This component is designed to be used after a ConvolutionComponent
+ * so that the input matrix is propagated from a 2d/3d convolutional layer.
```
vijayaditya: there is only one type of convolution layer, this does pseudo 3d convolution (i.e., there is no hopping along the 3rd axis).

src/nnet3/nnet-simple-component.h
```diff
@@ -1643,6 +1643,109 @@ class MaxpoolingComponent: public Component {
...
+                           pool_x_step_(0), pool_y_step_(0), pool_z_step_(0) { }
+  virtual int32 InputDim() const;
+  virtual int32 OutputDim() const;
+
+  virtual std::string Info() const;
+  virtual void InitFromConfig(ConfigLine *cfl);
+  virtual std::string Type() const { return "Maxpooling3dComponent"; }
+  virtual int32 Properties() const {
+    return kSimpleComponent|kBackpropNeedsInput|kBackpropNeedsOutput|
+	    kBackpropAdds;
```
vijayaditya: @danpovey it is not clear to me when kBackpropAdds and kPropagateAdds bits have to be set. Is it sufficient to choose some value for these bits and honor that in the prop and backprop codes or should the decision of setting these bits be based on other information ?

src/nnet3/nnet-component-itf.cc
vijayaditya: @danpovey  I think the component is going to perform similarly for 1d, 2d and 3d inputs. There could be performance optimizations if we chose to have non-overlapping maxpooling windows. We can add additional code blocks which exploit the lack of overlap if there are noticeable delays.

src/nnet3/nnet-simple-component.cc
tomkocse: Can you advise how to  check the time between two check points in kaldi ?
It seems to me that a function similar to AddMatMatBatched() doesn't match the usage in max pooling as
the elements within the same pool still have to be compared one by one,
can you evaluate how it could be useful ?
I will first commit the changes to other comments.

vijayaditya: See the [line]( https://github.com/kaldi-asr/kaldi/blob/master/src/base/timer-test.cc#L33) for timing code.

src/nnet3/nnet-simple-component.h
danpovey: Vigay: regarding when to set kBackpropAdds and kPropagateAdds- it's based
on what's most convenient and efficient in the backprop and propagate
code.  For instance, if the propagate code uses AddMatMat, you should set
kPropagateAdds because you'd need a separate statement to zero it (and
there might be situations where you are actually adding to something).
Prefer addition in situations where you would otherwise have to zero the
matrix.
Dan

tomkocse: For a MaxpoolingComponent with parameters 
input-x-dim = 11, input-y-dim = 5, input-z-dim = 12, 
pool-x-size = 11, pool-y-size = 1, pool-z-size = 6, 
pool-x-step = 9, pool-y-step = 1, pool-z-step = 6

The time elapsed is 0.000340939 secs

danpovey: I don't think that number is very meaningful without additional info, e.g.
minibatch size.  May be best to work it out in gigaflops and compare it
with operations like matrix-multiply.
Dan

danpovey:
Actually, forget about this-- I think I will do this myself.
The convolutional code needs a rewrite, I think, for better documentation and more natural use of CUDA.

vijayaditya:
When you mention the rewrite are you talking about rewriting the
convolution component as a non-simple component ?

--Vijay

danpovey:
I'm thinking about that too.

Dan

danpovey:
Guys, I changed my mind about implementing the convolutional stuff myself.
I realized that my plan was not as advantageous as I realized, and the amount of work was more than I had imagined.
So you can go ahead with this.
However, I don't think the way you have coded the maxpooling and its backprop is very ideal at all.  It would be better to do it in a CUDA kernel.

tomkocse:
@danpovey , are you worrying the case when the block size is large but the number of pool is small ?
I admit that the way i code is assuming small block size and large number of pool.

danpovey:
There just seem to be a lot of individual CUDA calls.

danpovey:
I'll merge this now.  If efficiency becomes a problem we can add a kernel that does max stuff without having to be called in a loop.

PR #461, opened 2016-01-21
Added PerDimWeightedAverage component for chain models in SWBD, works…

vijayaditya:
… as good as jesus (tdnn_4f) model

### DIFF
--- a/egs/swbd/s5c/local/chain/run_tdnn_4q.sh
+++ b/egs/swbd/s5c/local/chain/run_tdnn_4q.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+
+# this is based on Dan's tdnn_2o script
+# it has a different splicing configuration
+# it uses the PerDimensionWeightedAverage pooling in place of the Jesus layer
+
+# it relies on new steps/nnet3/chain/train_tdnn_b.sh script which accepts more 
+# parameters are calls new config generator steps/nnet3/tdnn/make_configs.py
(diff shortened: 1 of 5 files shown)


vijayaditya:
I have created a new config generator for TDNNs similar to those available for LSTM. This will not affect any existing TDDNN/Jesus scripts.

danpovey:
Why do you need a separate train_tdnn.sh script for this?  Can't you use the same pattern as for the jesus-configs, to create the configs?

danpovey:
It's the duplication of code that I'm not crazy about, and the name of the
script train_tdnn_b.sh.  Why couldn't you just slightly extend the existing
train_tdnn.sh script?
BTW, did the low-pass filtering help?
Dan

vijayaditya:
the only thing different in the new train_tdnn.sh script is the make_configs.py which is being called. This new steps/nnet3/tdnn/make_configs.py script makes use of components defined in steps/nnet3/components.py

vijayaditya:
Low-pass filtering didn't help but I would like to keep it as it was just overtraining a lot and I hope it will improve after l2 regularization. Will start this experiment tomorrow.

danpovey:
OK.
Better to just make changes to train_tdnn.sh than to duplicate the script.
Dan

vijayaditya:
OK, will do.

vijayaditya:
just realized I am using a tracking branch for pushing this commit. so I will close this and recreate a pull request.

PR #464, opened 2016-01-22
thchs30 recipe

wangdong99:
The thchs30 recipe for the free Chinese database thchs-30 (http://www.openslr.org/18/).


egs/thchs30/s5/path.sh
```diff
@@ -0,0 +1,18 @@
+#Change the following settings to match your computing env,
+#so that appropriate (CPU/GPU) Kaldi commands are visible
+
+export KALDI_ROOT=/work3/zxw/kaldi-2016-1-21
```
jtrmal: It would be perhaps generally more useful to have the KALDI_ROOT set up in the same way as in other recipes, i.e. three levels up?

egs/thchs30/s5/path.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ -z `which nvidia-smi 2>/dev/null` ];then
+#if nvidia_smi; then
+    src=src_nocuda
+    echo "using no-cuda kaldi src: $KALDI_ROOT src"
+else
+    src=src_cuda
+    echo "using cuda kaldi src: $KALDI_ROOT src"
+fi
+
```
jtrmal: The src_nocuda/src_cuda will be perhaps confusing for other people than you (as the default directory is just "src", so no matter what path will be chosen (depending on existence nvidia-smi), it will always generate an incorrect $PATH
Also, ad the "which"  command -- you might want possibly consider using "command -v nvidia-smi >/dev/null 2>&1", which is POSIX defined -- "which" is sometimes hijacked by packagers (on ubuntu, for example) and that might lead to confusion. But I leave this on you and Dan.

egs/thchs30/s5/local/insert_blank.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/python2.6
```
jtrmal: I suggest using only shebang
#!/usr/bin/env python

because I believe your program should be able to run in pythin 2.7 as well.

egs/thchs30/s5/local/dae/add-noise-mod.py
```diff
@@ -0,0 +1,157 @@
+#!/nfs/disk/perm/linux-x86_64/bin/python2
+###!/usr/bin/python
```
jtrmal: I suggest using only shebang
#!/usr/bin/env python

Everything else complicates running the scripts elsewhere. The situation is complicated enough with Py2.x and Py3

wangdong99:
Thanks to Yenda's nice suggestions. Most of them have been  accepted.

danpovey:
@keli78, before I merge this could you please try to run this recipe on our system and see if any problems show up?
Dan

wangdong99:
sure. please send us the info.

danpovey:
I notice that you checked in some wav files with your pull request.
We don't check in binary files like that in to Kaldi's code repo.  They should be made part of the download from openslr- a separate archive if you want.  Prepare something for Yenda to upload.
Dan

egs/thchs30/s5/cmd.sh
```diff
@@ -0,0 +1,35 @@
...
+#export cuda_cmd="queue.pl -q long.q@@pco203 -l gpu=1"
+#export cuda_cmd="queue.pl -q long.q@pcspeech-gpu"
+#export mkgraph_cmd="queue.pl -q all.q@@servers -l ram_free=4G,mem_free=4G"
+
+#c) CSLT cluster options
+
+export train_cmd="queue.pl -q cpu.q"
+export decode_cmd="queue.pl -q cpu.q"
+export cuda_cmd="queue.pl -q cuda.q@grid-11"
+
```
danpovey: It would be better to use the 'new-style' options here.  train_cmd and decode_cmd should just be queue.pl, and cuda_cmd would be queue.pl --gpu 1, and locally you could create a conf/queue.conf to explain how to translate these into options to GridEngine.
See http://kaldi-asr.org/doc/queue.html and search for conf/queue.conf

egs/thchs30/s5/local/dae/run_dae.sh
```diff
@@ -0,0 +1,154 @@
...
+   sigma0=10 #some random in SNR
+   seed=32
+   verbose=0
+
+   addnoise_opt="--noise-level $noise_level --sigma0 $sigma0 --seed $seed --verbose $verbose --noise-prior $noise_prior --noise-src $noise_scp --wavdir NULL"
+
+   rm -rf data/dae/train && mkdir -p data/dae/train || exit 1
+   cp -L data/fbank/train/{wav.scp,spk2utt,utt2spk,text} data/dae/train || exit 1
+   local/dae/make_fbank.sh --nj $nj --cmd "$train_cmd"  \
+   --addnoise-opt """$addnoise_opt"""  \
```
danpovey: you should indent lines that appear after backslash to make clear that they are part of the previous command.

egs/thchs30/s5/local/dae/run_dae.sh
```diff
@@ -0,0 +1,154 @@
...
+   if [ $stdtest = true ]; then
+     #download noisy wav if use the standard test data
+     echo "using standard test data"
+     if [ $dwntest = true ];then
+       echo "downloading the noisy test data from Tsinghua Univ..."
+       echo "this may be slow for some connections, and is not very stable."
+       echo "you may want to try alternative mirror sites."
+       echo "check http://data.cslt.org/thchs30/challenges/asr.html"
+       (
+        mkdir -p wav && cd wav && \
```
danpovey: don't you want the openslr.org location to be the default one?  For people outside China that will definitely be more convenient.

egs/thchs30/s5/local/dae/run_dae.sh
```diff
@@ -0,0 +1,154 @@
...
+            ;;
+         cafe)
+            noise_prior="0.0,0.0,0.0,10.0" 
+            ;;
+       esac
+
+       addnoise_opt="--noise-level $noise_level --sigma0 $sigma0 --seed $seed --verbose $verbose --noise-prior $noise_prior --noise-src $noise_scp --wavdir NULL"
+       rm -rf data/dae/test/0db/$x && mkdir -p data/dae/test/0db/$x && \
+       cp -L data/fbank/test/{wav.scp,spk2utt,utt2spk,text} data/dae/test/0db/$x || exit 1
+       local/dae/make_fbank.sh --nj $nj --cmd "$train_cmd"  \
```
danpovey: I have a problem with the way you have integrated your addition of noise into the 'make_fbank.sh'.  It makes it impossible to do anything that needs non-fbank features.  And these recipes that you have used, nnet1 recipes based on filterbanks, are very old recipes; we have a lot of other things now that will probably work better.  Instead you should do this all at the wav.scp level, so you have a data directory that has commands inside the wav.scp to add the noise.  Or if that's not efficient, then dump noise-added data to disk and put their locations in the wav.scp.  Then it's a standard data directory and people can dump any kinds of features they want.

danpovey:
Actually, @keli78, wait before testing it.  There are various other issues we have to resolve first.

egs/thchs30/s5/run.sh
```diff
@@ -0,0 +1,112 @@
...
+#produce MFCC features 
+rm -rf data/mfcc && mkdir -p data/mfcc &&  cp -R data/{train,dev,test,test.ph} data/mfcc || exit 1;
+for x in train dev test; do
+   #make  mfcc 
+   steps/make_mfcc.sh --nj $n --cmd "$train_cmd" data/mfcc/$x exp/make_mfcc/$x mfcc/$x || exit 1;
+   #compute cmvn
+   steps/compute_cmvn_stats.sh data/mfcc/$x exp/mfcc_cmvn/$x mfcc/$x/_cmvn || exit 1;
+done
+#copy feats and cmvn to test.ph, avoid duplicated mfcc & cmvn 
+cp data/mfcc/test/feats.scp data/mfcc/test.ph && cp data/mfcc/test/cmvn.scp data/mfcc/test.ph || exit 1;
```
danpovey: we don't normally put dots in directory names (like ph), and it's not clear what ph means here.

egs/thchs30/s5/run.sh
```diff
@@ -0,0 +1,112 @@
...
+
+
+# prepare language stuff
+# build a large lexicon that invovles words in both the training and decoding. 
+# The large lexioon is mainly for reasonable lattice generation in discriminative training
+(
+  echo "make word graph ..."
+  cd $H; mkdir -p data/{dict,lang,graph} && \
+  cp local/dict/{extra_questions.txt,nonsilence_phones.txt,optional_silence.txt,silence_phones.txt} data/dict && \
+  cat local/dict/lexicon.txt $thchs/lm_word/lexcion.txt |grep -v '<s>'|grep -v '</s>' |sort -u > data/dict/lexicon.txt || exit 1;
```
danpovey: please add spaces around the pipe symbol, for readability.

egs/thchs30/s5/local/score.sh
```diff
@@ -0,0 +1,67 @@
...
+mkdir -p $dir/scoring/log
+
+cat $data/text | sed 's:#::g' | sed 's:<SPOKEN_NOISE>::g'|PYTHONIOENCODING=utf-8 local/insert_blank.py | tee >$dir/scoring/test_filt.txt
+
+$cmd LMWT=$min_lmwt:$max_lmwt $dir/scoring/log/best_path.LMWT.log \
+  lattice-scale --inv-acoustic-scale=LMWT "ark:gunzip -c $dir/lat.*.gz|" ark:- \| \
+  lattice-add-penalty --word-ins-penalty=$word_ins_penalty ark:- ark:- \| \
+  lattice-best-path --word-symbol-table=$symtab \
+    ark:- ark,t:$dir/scoring/LMWT.tra || exit 1;
+
```
danpovey: you can remove the $reverse stuff, it's not used any more.

egs/thchs30/s5/local/dict/extra_questions.txt
```diff
@@ -0,0 +1,7 @@
+sil
+a1 ai1 an1 ang1 ao1 e1 ei1 en1 eng1 i1 ia1 ian1 iang1 iao1 ie1 in1 ing1 iong1 iu1 ix1 iy1 o1 ong1 ou1 u1 ua1 uai1 uan1 uang1 ueng1 ui1 un1 uo1 v1 van1 ve1 vn1
```
danpovey: The vocabulary-related things like the lexicon and extra_questions.txt and so on should be a download from openslr.org--- or better, just make the lexicon the download and make everything else derived from it.  If we added all the lexicons of the various recipes to the Kaldi repository it would get huge very fast.

egs/thchs30/s5/local/dae/train.sh
```diff
@@ -0,0 +1,437 @@
+#!/bin/bash
+
```
danpovey: This doesn't looks like a script that really belongs in local/.  What changes did you make to Karel's standard script in steps/?  Can't you just do it via options to the standard script?

egs/thchs30/s5/path.sh
danpovey: yes, as Yenda says, the normal setting is something like KALDI_ROOT=../../..
Bear in mind that the point with all these scripts is to make it easy for it to work on *someone else's* system.  Whether it works easily on your system is a very secondary consideration-- for you, you know how your system works and you can easily change things.  Someone who downloads your recipe doesn't need extra hassle.

egs/thchs30/s5/path.sh
danpovey: I agree with Yenda- you should remove all of this.
Things that are specific to a person's local system should in general never appear in scripts like these, unless they are things like directory names that you could never have a suitable default value form.

egs/thchs30/s5/local/dae/add-noise-mod.py
```diff
@@ -0,0 +1,153 @@
...
+import logging
+import wave
+import math
+import struct
+import sys
+import os
+
+try:
+  import pyximport; pyximport.install()
+  from util import *
```
jtrmal: Not sure if the line "from util import *" should be here, protected by the try/except as well

egs/thchs30/s5/local/dae/add-noise-mod.py
```diff
@@ -0,0 +1,153 @@
...
+
+def num_samples(mat):
+    return len(mat)
+
+def scp(scp_filename):
+    with open(scp_filename) as f:
+        for l in f:
+            yield tuple(l.strip().split())
+
+def wave_header(sample_array, sample_rate):
```
jtrmal: standard python library contains a module called "wave" which could be used instead of wav handling code here. Would be possibly more portable but I don't particularly care if you change it or not.

egs/thchs30/README.txt
```diff
@@ -0,0 +1,4 @@
+THCHS30 is an open Chinese speech database published by Center for Speech and Language Technology (CSLT) at Tsinghua University.
+
+The origional recording was conducted in 2002 by Dong Wang, supervised by Prof. Xiaoyan Zhu, at the Key State Lab of Intelligence and System, Department of Computer Science, Tsinghua Universeity, and the original name was 'TCMSD', standing for 'Tsinghua Continuous Mandarin Speech Database'. The publication after 13 years has been initiated by Dr. Dong Wang and was supported by Prof. Xiaoyan Zhu. We hope to provide a toy database for new researchers in the field of speech recognition. Therefore, the database is totally free to academic users.
+
```
jtrmal: Perhaps add a link to the your website and/or openslr

egs/thchs30/s5/local/dae/add-noise-mod.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
```
jtrmal: Add a copyright notice and author names, please. This should be probably done for all the other scripts you've written as well.

egs/thchs30/s5/local/score.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+# Copyright 2012  Johns Hopkins University (Author: Daniel Povey)
+# Apache 2.0
+
```
jtrmal: This while script should perhaps be replaced by a link to steps/score_kaldi.sh
which does the same thing

egs/thchs30/s5/local/dae/add-noise-mod.py
wangdong99: util is imported only when pyximport is successful, to provide faster add-noise.

wangdong99:
two main changes: (1) cmd.sh, uses the new mechanism for queue setting (2)local/score.sh, link to steps/score_kaldi.sh, and change insert_blank.py to wer_output_filter.

jtrmal:
Perfect :)
y.

egs/thchs30/s5/run.sh
```diff
@@ -0,0 +1,113 @@
...
+thchs=/nfs/public/materials/data/thchs30-openslr
+#corpus and trans directory
+#you can obtain the database by uncommting the following lines
+#( cd $thchs
+#     echo "downloading THCHS30 at $PWD ..."
+#     wget http://www.openslr.org/resources/18/data_thchs30.tgz
+#     wget http://www.openslr.org/resources/18/resource.tgz
+#     tar xvf data_thchs30.tgz && tar xvf resource.tgz 
+#)
+
```
jtrmal: The previous commented lines are not completely correct (after uncommenting) -- the $thchs variable will be either unset or set to come directory that won't probably be available anywhere else than your system. I'd say it would make more sense to write:

thchs=/nfs/public/materials/data/thchs30-openslr
if [ ! -d $thchs ] ; then
  thchs=./thchs30-openslr
  [ -d $thchs ] || mkdir $thchs
  cd $thchs
  echo "downloading THCHS30 at $PWD ..."
  wget http://www.openslr.org/resources/18/data_thchs30.tgz
  wget http://www.openslr.org/resources/18/resource.tgz
  tar xvf data_thchs30.tgz && tar xvf resource.tgz 
fi

eventually you can add some more checks to avoid the downloading and untaring the data each time it's  the recipe is run.

danpovey: Yes, definitely.  Please see how it's done in the Librispeech recipe for an
example.  You don't have to be as careful e.g. with the md5 checksums, but
at least check that it hasn't been downloaded already.
Dan

egs/thchs30/s5/local/dae/add-noise-mod.py
jtrmal: OK, now I understand -- I was wrongly assuming util is a python system package. Perhaps rename it to less generic, e.g. tsch30_util (or something like that)?

egs/thchs30/s5/run.sh
jtrmal: BTW, I just recalled I was adding two new files to the openslr. When you download only these two, are you sure you won't miss the third one? I think those were noises?

wangdong99: Agree. the Librispeech way is really good. I duplicated the script and made some modifications to support thchs30. Another change is that we now only use the standard noisy test data, so all the data packages are downloaded together.

egs/thchs30/s5/local/dae/add-noise-mod.py
wangdong99: Yes, agree. Changed to thchs30_util.

wangdong99:
two changes: (1) download all the data packages altogether (2) use standard noisy test data only.

egs/thchs30/s5/local/dae/run_dae.sh
```diff
@@ -0,0 +1,149 @@
...
+   $train_cmd JOB=1:$nj exp/dae/gendata/add_noise_dev.JOB.log \
+     local/dae/add-noise-mod.py --noise-level $noise_level \
+       --sigma0 $sigma0 --seed $seed --verbose $verbose \
+       --noise-prior $noise_prior --noise-src $noise_scp \
+       --wav-src exp/dae/gendata/dev_split_JOB.scp --wavdir $wavdir \
+       || exit 1
+   steps/make_fbank.sh --nj $nj --cmd "$train_cmd"  \
+     data/dae/dev/0db exp/dae/gendata fbank/dae/dev/0db || exit 1
+   steps/compute_cmvn_stats.sh data/dae/dev/0db exp/dae/cmvn \
+     fbank/dae/dev/0db/_cmvn || exit 1
```
danpovey: I just noticed this-- please remove all directory extensions of the type
x/y/_cmvn
(This is unnecessary and odd-looking).   Leave them as just x/y/.

wangdong99: ok. done.

danpovey:
are you sure you removed them?

wangdong99:
Dan, did you still find them in any place?

danpovey:
OK, I think they're gone now.  Merging.  Thanks!

wangdong99:
Thanks!

naxingyu:
From what I can tell, the data is named as "spk_utt.wav".
Current recipe seems not using the speaker info, resulting in one-to-one 
spk2utt/utt2spk mappings.
Should that be used? SAT+fmllr will make more sense then...

Xingyu

在 2016/2/4 11:01, wangdong99 写道:

wangdong99:
Yes, using true speaker information probably leads to better SAT model. The current solution is just temporary, since some wav files were not labelled  so well in terms of speakers. Checking them now..

PR #470, opened 2016-01-27
Fix compilation issues when using KALDI_DOUBLEPRECISION=1

funous:
Fix some compilation issues by specifying templates instead of letting the compiler guess and replacing hardcoded floats in signal processing by BaseFloat typedef

When dealing with rnnlm, floats are used everywhere, no matter what the precision of the rest of Kaldi is (it was already like that in most of the rnnlm code, with a few exceptions)

### DIFF
--- a/src/nnet2/nnet-component.cc
+++ b/src/nnet2/nnet-component.cc
@@ -3933,7 +3933,7 @@ void Convolutional1dComponent::Propagate(const ChunkInfo &in_info,
   }
   
   // apply all filters
-  AddMatMatBatched(1.0f, tgt_batch, patch_batch, kNoTrans, filter_params_batch,
+  AddMatMatBatched<BaseFloat>(1.0f, tgt_batch, patch_batch, kNoTrans, filter_params_batch,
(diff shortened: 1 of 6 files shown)


src/nnet2/nnet-component.cc
```diff
@@ -3933,7 +3933,7 @@ void Convolutional1dComponent::Propagate(const ChunkInfo &in_info,
   }
   
   // apply all filters
-  AddMatMatBatched(1.0f, tgt_batch, patch_batch, kNoTrans, filter_params_batch,
+  AddMatMatBatched<BaseFloat>(1.0f, tgt_batch, patch_batch, kNoTrans, filter_params_batch,
```
danpovey: Since we're at it, could you please replace the 1.0f in these AddMatMatBatched calls with 1.0?
1.0f seems to have been an attempt to call the <float> version of the template, but isn't needed and is confusing.

funous:
Ok, I'll do it.

danpovey:
Thanks!  Merging.

PR #474, opened 2016-02-01
Update cu-device.cc

tal1974:
windows VS2013 CUDA 7.0 or 7.5 function cuMemGetInfo has 2 argumets


jtrmal:
THanks -- while we are at it, can you please have a look at line 450, where the call is made again and fix it? Actuall, if I read the code correctly, that statement will never get active -- in that case you can remove it.

tal1974:
Yes, 450 line is dead. I remove it, but can check this code on my mashine.

tal1974:
I mean I checked for case _MSC_VER. Maybe someone else can check code for case  "not _MSC_VER"?

danpovey:
I'm pretty sure if there is any problem with the code it's with _MSC_VER,
since things are used frequently on UNIX.

Dan

jtrmal:
Thanks, merging!

PR #478, opened 2016-02-02
INSTALL.md

tal1974:
CUDA 7.0, 7.5 doesn't support vs2015. CUDA 8.0 isn't released yet.

### DIFF
--- a/windows/INSTALL.md
+++ b/windows/INSTALL.md
@@ -148,7 +148,7 @@ for their processors. Unfortunately it isn't free. OpenBLAS is free alternative 
 
          (kaldi)/tools$ generate_solution.pl --vsver vs2015 --enable-openblas
 
-    Another example, for a build using OpenBLAS, VS 2013 and CUDA support:
+    Another example, for OpenBLAS, VS 2013 and CUDA support:
 
          (kaldi)/tools$ generate_solution.pl --vsver vs2013 --enable-cuda --enable-openblas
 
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks- Yenda will check this.

jtrmal:
Thanks, I will look into it. My feeling is that  I was able to compile the the combination (cuda 7.0 + vs2015)  without any problems.
The probable reason why it's said it's unsupported is that it does not have a proper VS2015 integration. (but we do not care about that, as we generate the sln files "by hand").
But it's been a while, so I might be not rem,embering it correctly

tal1974:
it’s interesting. i'll try to check

tal1974:
I installed VS2015. First of all I had to copy some CUDA 7.0 files from C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\BuildCustomizations to \V140\BuildCustomizations. But after this I couldn’t build kaldi (I tried kaldi-cudamatrix.lib) because an error occurred:
1>c:\program files\nvidia gpu computing toolkit\cuda\v7.0\include\host_config.h(119): fatal error C1189: #error:  -- unsupported Microsoft Visual Studio version! Only the versions 2010, 2012, and 2013 are supported!

jtrmal:
Hm, as it seems NVIDIA is really determined in the fact that cuda 7.0 is
not supported under vs2015 :)
THhnks for trying this -- I'll merge your PR and will perhaps experiment
with this later when I will have more time.
Thanks a lot! Your help is appreciated.
y.

PR #479, opened 2016-02-02
Make CWD check less strict

mbait:
It might be that `configure` is still invoked from 'src' directory, yet the
directory name itself isn't 'src'. While a real procedure of comparing two
filesystem paths to be equal is quite complex, this small trick does exactly
what we want to do - check that `configure` is invoked from the same directory
where it is.


danpovey:
Thanks!  Merging.

PR #482, opened 2016-02-03
NormalizeComponent with add-log-std option 

pegahgh:


src/nnet3/nnet-simple-component.h
```diff
@@ -157,6 +161,9 @@ class NormalizeComponent: public NonlinearComponent {
   // about 0.7e-20.  We need a value that's exactly representable in
   // float and whose inverse square root is also exactly representable
   // in float (hence, an even power of two).
+
+  bool add_log_std_; // If true, log(max(epsi, sqrt(row_in^T row_in / D)))
```
danpovey: you should call it add_log_stddev_, not add_log_std_.
Also, " is added as node to output layer." is not the right terminology as used in nnet3, its an extra dimension in the output layer.

src/nnet3/nnet-simple-component.cc
```diff
@@ -275,6 +278,12 @@ void NormalizeComponent::Read(std::istream &is, bool binary) {
     ReadBasicType(is, binary, &target_rms_);
     ReadToken(is, binary, &tok);
   }
+  //  Read add_log_std_ token, if it sets. 
+  if (tok == "<AddLogSum>") {
```
danpovey: Please make the token name consistent with the variable name.

pegahgh:
changed option names to add-log-stddev.

src/nnet3/nnet-component-itf.cc
```diff
@@ -251,7 +251,7 @@ void NonlinearComponent::ZeroStats() {
 
 std::string NonlinearComponent::Info() const {
   std::stringstream stream;
-  KALDI_ASSERT(InputDim() == OutputDim());  // always the case
+  //KALDI_ASSERT(InputDim() == OutputDim());  // always the case
```
danpovey: if you don't need the assert statement then delete it.

src/nnet3/nnet-simple-component.cc
```diff
@@ -392,6 +430,19 @@ void NormalizeComponent::Backprop(const std::string &debug_info,
   in_deriv->AddDiagVecMat(-1.0 / d_scaled,
                           dot_products, in_value,
                           kNoTrans, 1.0);
+
+  if (add_log_stddev_) {
```
danpovey: You wouldn't need to use a temporary matrix if you computed this before the other part of the input derivative.

pegahgh:
fixed!

src/nnet3/nnet-component-itf.cc
```diff
@@ -251,7 +251,6 @@ void NonlinearComponent::ZeroStats() {
 
 std::string NonlinearComponent::Info() const {
   std::stringstream stream;
-  KALDI_ASSERT(InputDim() == OutputDim());  // always the case
   stream << Type() << ", dim=" << InputDim();
```
danpovey: You should extend the Info(), so that it's clear from the info string whether you have this extra feature.
Probably if the feature is included, it should say input-dim=xxx, output-dim=xxx, include-log-stddev=true

src/nnet3/nnet-simple-component.cc
```diff
@@ -243,23 +243,26 @@ const BaseFloat NormalizeComponent::kNormFloor = pow(2.0, -66);
 // This component modifies the vector of activations by scaling it so that the
 // root-mean-square equals 1.0.  It's important that its square root
 // be exactly representable in float.
-void NormalizeComponent::Init(int32 dim, BaseFloat target_rms) {
+void NormalizeComponent::Init(int32 dim, BaseFloat target_rms, bool add_log_stddev) {
```
danpovey: watch the 80-char limit

pegahgh:
Fixed!

src/nnet3/nnet-component-itf.cc
danpovey: pegah, you still haven't done this (fix the info string)

pegahgh: I had extended the NormalizeComponent::Info() to contain input-output dim if add-log-stddev is true.
I modified NonlinearComponent::Info() . Is this what you asked for?

danpovey:
I see- looks right.

PR #483, opened 2016-02-04
Avoid ValueError by splitting on any number of spaces

jfainberg:
Removing the argument from split() ensures that we get the desired result also when the input splice_indexes has multiple spaces separating the fields instead of a single space. E.g.:
`splice_indexes="-4,-3,-2,-1,0,1,2,3,4`  0  -2,2  0  -4,4 0"
`splice_indexes.split()`
`['-4,-3,-2,-1,0,1,2,3,4',` '0', '-2,2', '0', '-4,4', '0']
`splice_indexes.split("` ")
`['-4,-3,-2,-1,0,1,2,3,4',` '', '0', '', '-2,2', '', '0', '', '-4,4', '0']
Where the latter gives me a ValueError (and is the default in train_tdnn.sh).

### DIFF
--- a/egs/wsj/s5/steps/nnet3/make_tdnn_configs.py
+++ b/egs/wsj/s5/steps/nnet3/make_tdnn_configs.py
@@ -70,7 +70,7 @@ else:
 splice_array = []
 left_context = 0
 right_context = 0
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks!  Merging.

PR #486, opened 2016-02-05
WIP: CNN recipes

tomkocse:


egs/swbd/s5c/local/nnet3/run_cldnn.sh
```diff
@@ -0,0 +1,153 @@
...
+
+# Copyright 2015  Johns Hopkins University (Author: Daniel Povey).
+#           2015  Vijayaditya Peddinti
+#           2015  Xingyu Na
+#           2015  Pegah Ghahrmani
+#           2016  Tom Ko
+# Apache 2.0.
+
+
+# this is a basic lstm script
```
vijayaditya: change the description for the new script. Describe in detail the architecture used. Described how ivectors are used. Describe the CNN config in detail.

egs/swbd/s5c/local/nnet3/run_cldnn.sh
```diff
@@ -0,0 +1,153 @@
...
+
+stage=9
+train_stage=-10
+has_fisher=true
+affix=
+speed_perturb=true
+common_egs_dir=
+
+# CNN options
+cnn_indexes="3,8,1,1,256,1,3,1,1,3,1"
```
vijayaditya: explain in detail the what these hyper-parameters are. For complicated components like I would suggest using python dict format.  e.g.
<pre><code>
cnn_parameters='{
'layer1': {
                 'filter-x-dim' : 3,
                 'filter-y-dim' : 8,
                 'filter-x-step' : 1,
                 'filter-y-step' : 1,
                 'num-filters' : 256
     }
}'

You can directly parse this string in python using eval.

naxingyu:
Is there an update on the RESULTS file?

vijayaditya:
@xingyu
This is work in progress submitted for intermediate review. The results
will be updated later.

egs/swbd/s5c/local/nnet3/run_cldnn.sh
```diff
@@ -0,0 +1,153 @@
...
+cnn_affine_dim=256
```
vijayaditya: What is cnn_affine_dim ? Explain here or use the above specified python dict format.

egs/swbd/s5c/local/nnet3/run_cnn.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# Copyright 2015  Johns Hopkins University (Author: Daniel Povey).
+#           2015  Vijayaditya Peddinti
+#           2016  Tom Ko
+# Apache 2.0.
+
+# this is the CNN+TDNN system, built in nnet3;
+# this system differs from the standard TDNN system
+# in replacing the LDA layer with CNN 
+# and using fbank features to the CNN input
```
vijayaditya: Also there is a change in how ivectors are used. Better name the script run_cnn_tdnn.sh

egs/wsj/s5/steps/nnet3/cldnn/make_configs.py
```diff
@@ -0,0 +1,288 @@
...
+    init_config_lines = copy.deepcopy(config_lines)
+    init_config_lines['components'].insert(0, '# Config file for initializing neural network prior to')
+    init_config_lines['components'].insert(0, '# preconditioning matrix computation')
+    nodes.AddOutputNode(init_config_lines, prev_layer_output)
+    config_files[args.config_dir + '/init.config'] = init_config_lines
+
+    if args.ivector_dim > 0:
+      iv_layer_output = {'descriptor':  'ReplaceIndex(ivector, t, 0)',
+                         'dimension': args.ivector_dim}
+      iv_layer_output = nodes.AddAffineLayer(config_lines, "IV", iv_layer_output, args.ivector_dim, "")
```
vijayaditya: Use a more informative name than IV, use something like Affine_ivector.

egs/swbd/s5c/local/nnet3/run_cldnn.sh
danpovey: Vijay, I'm  not so crazy about that dict format.

egs/wsj/s5/steps/nnet3/cldnn/make_configs.py
```diff
@@ -0,0 +1,288 @@
...
+
+    # Add the init config lines for estimating the preconditioning matrices
+    init_config_lines = copy.deepcopy(config_lines)
+    init_config_lines['components'].insert(0, '# Config file for initializing neural network prior to')
+    init_config_lines['components'].insert(0, '# preconditioning matrix computation')
+    nodes.AddOutputNode(init_config_lines, prev_layer_output)
+    config_files[args.config_dir + '/init.config'] = init_config_lines
+
+    if args.ivector_dim > 0:
+      iv_layer_output = {'descriptor':  'ReplaceIndex(ivector, t, 0)',
```
vijayaditya: sometime later you might want to add the changes from @freewym 's PR for using multiple ivectors. (https://github.com/kaldi-asr/kaldi/pull/468)

egs/swbd/s5c/local/nnet3/run_cldnn.sh
vijayaditya: I am worried that the simple string format will introduce a lot of user
errors.
It will force users to specify default parameters even when not required
and it is a bit uninformative.
The python dict format is well documented so I hope it will be better than
defining our own string formats.

--Vijay

egs/wsj/s5/steps/nnet3/cldnn/train.sh
```diff
@@ -0,0 +1,721 @@
+#!/bin/bash
+
```
vijayaditya: @danpovey there is redundancy in steps/nnet3/*/train.sh scripts as we add more architectures. Some of these scripts differ in nothing but the config generators.
e.g. cldnn/train.sh and lstm/train.sh would be very similar. cnn/train.sh tdnn/train.sh would be very similar.
 I think we should look at a better way to pass the nnet3 architecture parameters. I am thinking about passing all the make_config.py parameters using a config file so that steps/nnet3/*/train.sh script can just invoke the config generator as
<code>
steps/nnet3/$model/make_configs.py $config_file
</code>

Any suggestions ?

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -34,6 +34,75 @@ def AddLdaLayer(config_lines, name, input, lda_file):
     return {'descriptor':  '{0}_lda'.format(name),
             'dimension': input['dimension']}
 
+def AddCnnLayer(config_lines, name, input, cnn_indexes=[0]):
+    components = config_lines['components']
+    component_nodes = config_lines['component-nodes']
+    conv_input_x_dim = input['3d-dim'][0];
+    conv_input_y_dim = input['3d-dim'][1];
+    conv_input_z_dim = input['3d-dim'][2];
```
vijayaditya: Please use smaller functions like AddConvolutionLayer and AddMaxPoolLayer.

egs/wsj/s5/steps/nnet3/make_cnn_configs.py
```diff
@@ -0,0 +1,342 @@
...
+        print('component-node name=final-log-softmax component=final-log-softmax '
+              'input=final-fixed-scale', file=f)
+    else:
+        print('component-node name=final-log-softmax component=final-log-softmax '
+              'input=final-affine', file=f)
+
+    print('output-node name=output input=final-log-softmax', file=f)
+    f.close()
+
+# component name=nonlin1 type=PnormComponent input-dim=$pnorm_input_dim output-dim=$pnorm_output_dim
```
vijayaditya: Are these comments relevant to this python script ?

egs/swbd/s5c/local/nnet3/run_cldnn.sh
danpovey: Regarding this python dict format, etc.-- I don't like it as it's not
compatible with the way we pass most options.  If things have names then
they can be regular options in the script, like anything else.  I agree
that a lot of things are being passed in in this string, but maybe all of
them don't need to be configurable at this level.  Using the dict format is
getting uncomfortably close to some kind of weird metaprogramming.

egs/wsj/s5/steps/nnet3/cldnn/train.sh
danpovey: I agree that we need a better way to deal with the commonalities in the
nnet3 training scripts.  E.g. making the config-generation a separate call
than the main network training.  But this is a big project and I don't
think it should be mixed up with the current task of checking in a CNN
recipe.

Dan

vijayaditya: OK, I will submit a separate PR with a some suggested changes.

egs/wsj/s5/steps/nnet3/cldnn/make_configs.py
```diff
@@ -0,0 +1,288 @@
...
+     
+    list = [('Offset(input, {0})'.format(n) if n != 0 else 'input') for n in splice_indexes[0]]
+    splice_descriptor = "Append({0})".format(", ".join(list))
+    cnn_input_dim = len(splice_indexes[0]) * args.feat_dim
+    prev_layer_output = {'descriptor':  splice_descriptor,
+                         'dimension': cnn_input_dim,
+                         '3d-dim': [len(splice_indexes[0]), args.feat_dim, 1],
+                         'vectorization': 'yzx'}
+
+    num_cnn_layers = len(cnn_array)
```
vijayaditya: In standard CNN recipes, the input has filter-bank coefficients, deltas and double-deltas. This gives appreciable gains. You could compute the deltas and double deltas using fixed convolution layers. You can supply the output of this fixed convolution component to the first CNN layer. The fixed convolution layer can have filters which compute static, delta and double delta features.

tomkocse: Can I create deltas and double-deltas by add-deltas?

vijayaditya: Yes. You could also compute them inside the network.

Vijay

danpovey: it's definitely preferable to create the deltas and delta-deltas inside the network, because otherwise the egs get huge, and the training jobs will kill the disk.
(or you could try just giving the first convolutional layer extra temporal context, and make the time-dimension of the filter the same as the temporal context, so it could learn it itself).

danpovey:
Why is the name 'cldnn' used?  What does it mean?  Is it a known acronym?

vijayaditya:
Its an acronym from this paper CONVOLUTIONAL, LONG SHORT-TERM MEMORY, FULLY
CONNECTED DEEP NEURAL NETWORKS
<http://static.googleusercontent.com/media/research.google.com/en//pubs/archive/43455.pdf>
.

Vijay

13265170340:
What does has_fisher=true mean?

PR #487, opened 2016-02-05
Added a tool to compute the confidence interval of WER

nsmartinez:
Just a slight modification of compute-wer to estimate the confidence interval of WER, so different system can be compared more reliably. It is based on the paper:

http://www-i6.informatik.rwth-aachen.de/PostScript/InterneArbeiten/Bisani_BootstrapEstimatesForConfidenceIntervalsInASRPerformanceEvaluation_ICASSP_2004.pdf


src/bin/compute-confidence-interval-wer.cc
```diff
@@ -0,0 +1,138 @@
...
+      word_errs = LevenshteinEditDistance(ref_sent, hyp_sent, &num_ins, &num_del, &num_sub);
+      edit_word_per_hyp.push_back(std::pair<int32, int32>(word_errs, num_words));
+
+      num_sent++;
+      sent_errs += (ref_sent != hyp_sent);
+    }
+
+    //Extract WER for a number of replications of the same size as the hypothesis extracted
+    BaseFloat wer_accum = 0.0, wer_mult_accum = 0.0;
+    for ( int32 i = 0; i <= replications; ++i ) {
```
danpovey: whitespace is wrong, should be
  for (int32 i = 0; i <= replications; i++) {

src/bin/compute-confidence-interval-wer.cc
```diff
@@ -0,0 +1,138 @@
...
+int main(int argc, char *argv[]) {
+  using namespace kaldi;
+  typedef kaldi::int32 int32;
+
+  try {
+    const char *usage =
+        "Compute a bootstrapping of WER to extract the 95\% confidence interval\n"
+        "Takes two transcription files, in integer or text format,\n"
+        "and outputs overall WER statistics to standard output along with its confidence interval"
+        "supposing that WER of sentences subsamples follow the normal distribution.\n"
```
danpovey: Rather than saying supposing that WER of sentences subsamples follow the normal distribution, say
"using the 'bootstrap' method of Bisani and Ney\n"

Don't say 'Takes two transcription files, in integer or text format'
because it implies you are comparing the accuracy of two independent transcriptions.
Say
 'Takes a reference and a transcription file, in integer or text format.\n'

It would be better if you also implemented the two-system comparison method that they described in that paper-- you could implement it in the same binary, so if you give an extra argument it does the two-system one.
I suggest to call the binary compute-wer-confidence, so that it's more easily discoverable once you know about compute-wer.

danpovey:
@nsmartinez, did you see my comments?  Are you OK with doing what I said?

nsmartinez:
Sorry for the delay! I was travelling. Yes I see the comments and I am ok
with all the modifications you proposed. I think I can get them done during
this week.

2016-02-07 21:33 GMT+00:00 Daniel Povey <notifications@github.com>:

danpovey:
ok, thanks.

nsmartinez:
I have added the new functionality but I have not yet fully tested it, I hope I do it tomorrow. Also I had some doubts about code organisation, as some code is still repeated.

src/bin/compute-wer-confidence.cc
```diff
@@ -0,0 +1,245 @@
...
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "util/parse-options.h"
+#include "tree/context-dep.h"
+#include "util/edit-distance.h"
+#include "base/kaldi-math.h"
+
+namespace kaldi {
+
+void GetEditionsSingleHyp( const std::string &hyp_rspecifier,
```
danpovey: I think you mean edits, not editions.
the code duplication is OK with me.

src/bin/compute-wer-confidence.cc
```diff
@@ -0,0 +1,245 @@
...
+
+}
+
+void GetBootstrapWERInterval(const std::vector<std::pair<int32, int32> > & edit_word_per_hyp,
+        int32 replications, 
+        BaseFloat *mean, BaseFloat *interval) {
+    BaseFloat wer_accum = 0.0, wer_mult_accum = 0.0;
+
+    for (int32 i = 0; i <= replications; ++i) {
+
```
danpovey: too many blank lines... the normal Kaldi code doesn't look like this.

src/bin/compute-wer-confidence.cc
```diff
@@ -0,0 +1,245 @@
...
+        "Compute a bootstrapping of WER to extract the 95\% confidence interval\n"
+        "Take a reference and a transcription file, in integer or text format,\n"
+        "and outputs overall WER statistics to standard output along with its confidence interval\n"
+        "using the bootstrap methos of Bisani and Ney\n"
+        "If a second transcription file corresponding to the same reference is provided, a\n"
+        "bootstrap comparison of the two transcription is performed to estimate the probability\n"
+        "of improvement.\n"
+        "\n"
+        "Usage: compute-wer-confidence [options] <ref-rspecifier> <hyp-rspecifier> [<hyp2-rspecifier>]\n"
+        "E.g.: compute-wer-confidence --replications=10000 --mode=present ark:data/train/text ark:hyp_text\n"
```
danpovey: I'd remove 'replications' from the example comnand line.

danpovey: also, have another example command line for the 2-hyp case, like this:
E.g.: cmdline 1
or:  cmdline 2

danpovey:
We need to make some script-level changes so that the confidence intervals are computed in the normal Kaldi scoring (score_kaldi.sh), and add a script to compare two decoding directories (e.g. score_kaldi_compare.sh, to run on the directories previously scored by score_kaldi.sh .. I'd suggest to just compare the two best LMWT+ins-pen combinations found by the individual scripts).
If you're not familiar with bash scripting, maybe @jtrmal or me can do it.

nsmartinez:
Thanks for the suggestions I will include them. I am familiar with bash
script so I will include try to implement the script you suggest so it's
easy to use in recipes.

2016-02-10 18:14 GMT+00:00 Daniel Povey <notifications@github.com>:

jtrmal:
Cool, thanks. y.

nsmartinez:
I included the bash script and modify the tool to fit your comments, are you ok these changes?

danpovey:
@jtrmal, do you have time  to look at this?

jtrmal:
I will have a look. y.

danpovey:
@jtrmal, don't forget this!  I know you are traveling..

jtrmal:
Yeah, I apologize, I shouldn't have forgotten about this. I'll get to this
ASAP. y.

src/bin/compute-wer-confidence.cc
```diff
@@ -0,0 +1,247 @@
...
+      for (int32 j = 0; j <= edit_word_per_hyp.size(); ++j) {
+        int32 random_pos = kaldi::RandInt(0, edit_word_per_hyp.size());
+        word_errs += edit_word_per_hyp[random_pos].first - 
+                        edit_word_per_hyp2[random_pos].first;
+        }
+      if(word_errs > 0)
+        ++improv_accum;
+    }
+    // Compute mean WER and std WER
+    *p_improv = static_cast<BaseFloat>(improv_accum) / (replications-1);
```
jtrmal: This does not seem right -- shouldn't the p_improv be straight ratio between improv_accum and replication? I.e. not "replications-1"
Update: I looked into the paper and it seems the paper is indeed using "improv_accum/replications"

src/bin/compute-wer-confidence.cc
```diff
@@ -0,0 +1,247 @@
...
+    RandomAccessTokenVectorReader hyp_reader2(hyp_rspecifier2);
+    int32 num_words = 0, word_errs = 0, 
+            num_ins = 0, num_del = 0, num_sub = 0;
+    
+    // Main loop, store WER stats per hyp,
+    for (; !ref_reader.Done(); ref_reader.Next()) {
+      std::string key = ref_reader.Key();
+      const std::vector<std::string> &ref_sent = ref_reader.Value();
+      std::vector<std::string> hyp_sent, hyp_sent2;
+      if (!hyp_reader.HasKey(key) && !hyp_reader2.HasKey(key)) {
```
jtrmal: This makes me a little bit unsure. If you use compute-wer after post-processing the outputs, the utterances might actually disappear (or just be completely empty). I wonder if you'd be willing to improve the code in similar way the original compute-wer is -- it has two or three modes how it deals with missing utterance.

jtrmal:
I think the code is in a good shape. Nicolas, thank you for your work.

my comments:
I'd prefer to call the binary compute-wer-bootci (I saw bootci used several
times to denote the "boostrap conf intervals", so I think it might be more
self explaining -- compute-wer-confidence might perhaps cause incorrect
association with the confidences of the individual words or something to
that extent.
I'd also welcome extension of the steps/scoring_kaldi.sh to produce the
confidence intervals by default for the best system (for example a file in
the details/ directory).

There were two particular  minor issues which I think are not exactly right
--please review if I'm right or not -- I might be just mistaken.

y

nsmartinez:
Ok, I will rename it and correct the (replications-1) issue. Regarding the different scoring modes, I didn't implement them because there were tricky cases, e.g. samples exists in one and do not exists in the other, but I could consider them void (as you suggest), so I will implement it. Last, I will also modify score_kaldi to include the intervals.

nsmartinez:
Just updated the branch, tell if you have any problem/suggestion

jtrmal:
Nicolas, thanks!
Did you verify it runs OK on some recipe? Let me know, please -- I think
it's ready to merge, but I don't have the time to run an experiment and I
don't want to commit something that breaks the script.
Thank you!
y.

nsmartinez:
I tested it in an already performed experiment, and it worked. I think it
should work also if you run the experiment from scratch.

jtrmal:
Thanks. Merging.
y.

PR #490, opened 2016-02-08
WIP : Python version of train scripts

vijayaditya:
This is a very preliminary draft of the python training script. This PR has been created for review of basic structure of the code. There might be several errors. 

The script exploits python multi-line string enclosed by """" """" to avoid escaping the " and '  characters in the queue.pl and run.pl commands being run.

### DIFF
--- a/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
+++ b/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
@@ -1,0 +1,618 @@
+import subprocess
+import logging
+import math
+import re
+import time
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter('%(asctime)s [%(filename)s:%(lineno)s - %(funcName)s - %(levelname)s ] %(message)s')
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+
+
+def CheckIfCudaCompiled():
+    p = subprocess.Popen("cuda-compiled")
+    p.communicate()
+    if p.returncode == 1:
+        return False
+    else:
+        return True
+
+def RunKaldiCommand(command, wait = True):
+    """ Runs commands frequently seen in Kaldi scripts. These are usually a
+        sequence of commands connected by pipes, so we use shell=True """
+    #logger.info("Running the command\n{0}".format(command))
+    p = subprocess.Popen(command, shell = True,
+                         stdout = subprocess.PIPE,
+                         stderr = subprocess.PIPE)
+
+    if wait:
+        [stdout, stderr] = p.communicate()
+        if p.returncode is not 0:
+            raise Exception("There was an error while running the command {0}\n".format(command)+"-"*10+"\n"+stderr)
+        return stdout, stderr
+    else:
+        return p
+
+def GetSuccessfulModels(num_models, log_file_pattern, difference_threshold=1.0):
+    assert(num_models > 0)
+
+    parse_regex = re.compile("LOG .* Overall average objective function for 'output' is ([0-9e.\-+]+) over ([0-9e.\-+]+) frames")
+    objf = []
+    for i in range(num_models):
+        model_num = i + 1
+        logfile = re.sub('%', str(model_num), log_file_pattern)
+        lines = open(logfile, 'r').readlines()
+        this_objf = -100000
+        for line_num in range(1, len(lines) + 1):
+            # we search from the end as this would result in
+            # lesser number of regex searches. Python regex is slow !
+            mat_obj = parse_regex.search(lines[-1*line_num])
+            if mat_obj is not None:
+                this_objf = float(mat_obj.groups()[0])
+                break;
+        objf.append(this_objf);
+    max_index = objf.index(max(objf))
+    accepted_models = []
+    for i in range(num_models):
+        if (objf[max_index] - objf[i]) <= difference_threshold:
+            accepted_models.append(i+1)
+
+    if len(accepted_models) != num_models:
+        logger.warn("Only {0}/{1} of the models have been accepted for averaging, based on log files {2}.".format(len(accepted_models), num_models, log_file_pattern))
+    
+    return [accepted_models, max_index+1]
+
+def GetNumberOfLeaves(alidir):
+    [stdout, stderr] = RunKaldiCommand("tree-info {0}/tree 2>/dev/null | grep num-pdfs".format(alidir))
+    parts = stdout.split()
+    assert(parts[0] == "num-pdfs")
+    num_leaves = int(parts[1])
+    if num_leaves == 0:
+        raise Exception("Number of leaves is 0")
+    return num_leaves
+
+def GetNumberOfJobs(alidir):
+    try:
+        num_jobs = int(open('{0}/num_jobs'.format(alidir), 'r').readline().strip())
+    except IOError, ValueError:
+        raise Exception('Exception while reading the number of alignment jobs')
+    return num_jobs
+def GetIvectorDim(ivector_dir = None):
+    if ivector_dir is None:
+        return 0
+    [stdout_val, stderr_val] = RunKaldiCommand("feat-to-dim --print-args=false scp:{dir}/ivector_online.scp -".format(dir = ivector_dir))
+    ivector_dim = int(stdout_val)
+    return ivector_dim
+
+def GetFeatDim(feat_dir):
+    [stdout_val, stderr_val] = RunKaldiCommand("feat-to-dim --print-args=false scp:{data}/feats.scp -".format(data = feat_dir))
+    feat_dim = int(stdout_val)
+    return feat_dim
+
+def ReadKaldiMatrix(matrix_file):
+    try:
+        lines = map(lambda x: x.split(), open(matrix_file).readlines())
+        first_field = lines[0][0]
+        last_field = lines[-1][-1]
+        lines[0] = lines[0][1:]
+        lines[-1] = lines[-1][:-1]
+        if not (first_field == "[" and last_field == "]"):
+            raise Exception("Kaldi matrix file has incorrect format, only text format matrix files can be read by this script")
+        for i in range(len(lines)):
+            lines[i] = map(lambda x: int(x), lines[i])
+        return lines 
+    except IOError:
+        raise Exception("Error while reading the kaldi matrix file {0}".format(matrix_file))
+
+def WriteKaldiMatrix(output_file, matrix):
+    # matrix is a list of lists
+    file = open(output_file, 'w')
+    file.write("[ ")
+    row_len = len(matrix[0])
+    num_rows = len(matrix)
+    for row_index in range(len(matrix)):
+        if row_len != len(matrix[row_index]):
+            raise Exception("All the rows of a matrix are expected to have the same length")
+        file.write(" ".join(map(lambda x: str(x), matrix[row_index])))
+        if row_index != num_rows - 1:
+            file.write("\n")
+    file.write(" ]")
+    file.close() 
+
+import shutil
+def CopyEgsPropertiesToExpDir(egs_dir, dir):
+    try:
+        for file in ['cmvn_opts', 'splice_opts', 'final.mat']:
+            file_name = '{dir}/{file}'.format(dir = egs_dir, file = file)
+            if os.path.isfile(file_name):
+                shutil.copy2(file_name, dir)
+    except IOError:
+        raise Exception("Error while trying to copy egs property files to {dir}".format(dir = dir))
+
+def SplitData(data, num_jobs):
+   RunKaldiCommand("utils/split_data.sh {data} {num_jobs}".format(data = data,
+                                                                  num_jobs = num_jobs)) 
+
+def ParseModelConfigVarsFile(var_file):
+    try:
+        var_file_handle = open(var_file, 'r')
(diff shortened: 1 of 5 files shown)


egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,1134 @@
...
+
+    if wait:
+        [stdout, stderr] = p.communicate()
+        if p.returncode is not 0:
+            raise Exception("There was an error while running the command {0}".format(command)
+        return stdout, stderr
+    else:
+        return p
+
+def GetSuccessfulModes(num_models, log_file_pattern):
```
danpovey: i think you mean models, not modes

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,1134 @@
...
+
+def ForceSymlink(file1, file2):
+    try:
+        os.symlink(file1, file2)
+    except OSError, e:
+        if e.errno == errno.EEXIST:
+            os.remove(file2)
+            os.symlink(file1, file2)
+
+def ComputePresoftmaxPriorScale(dir, alidir, num_jobs, run_opts,
```
danpovey: I'm not sure that we're using this prior-scale thing in the latest example scripts-- IIRC in the end we found it wasn't helpful (?)

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,1134 @@
...
+    for iter in range(num_iters):
+        if (args.exit_stage is not None) and (iter == args.exit_stage):
+            logger.info("Exiting early due to --exit-stage {0}".format(iter))
+            return
+        current_num_jobs = int(0.5 + args.num_jobs_initial + (args.num_jobs_final - args.num_jobs_initial) * iter / num_iters)
+        
+        if iter in realign_iters:
+            prev_egs_dir=cur_egs_dir
+            cur_egs_dir="{0}/egs_{1}".format(args.dir, "iter"+str(iter))
+            Realign(args.dir, iter, args.alidir, args.lang,
```
danpovey: this should probably be inside a check for 'stage'.  If it isn't in the baseline script it would be a bug.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,1134 @@
...
+        logger.warning("""
+    Without using a GPU this will be very slow.  nnet3 does not yet support multiple threads.""")
+        
+        run_opts.train_queue_opt = ""
+        run_opts.parallel_train_opts = "--use-gpu=no"
+        run_opts.combine_queue_opt = ""
+        run_opts.prior_gpu_opt = "--use-gpu=no"
+        run_opts.prior_queue_opt = ""
+    
+    if args.run_in == "cluster":
```
danpovey: Vijay, this destroys all the power of the $cmd variable, just having two options like that, and it wouldn't be compatible with existing scripts.  It's not a rule that $cmd can only be one of two options-- may people use their own scripts.

vijayaditya: OK. I will just change it to cmd, as in current script.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,1132 @@
...
+    combine_queue_opt
+    prior_gpu_opt
+    prior_queue_opt
+    parallel_train_opts
+    
+    def __init__():
+        pass
+
+
+if __name__ == "__main__":
```
vdp: Just a couple, mostly stylistic/cosmetic, suggestions. I'm by no means a Python guru, so I don't claim that what I propose is necessarily better than the current code..

Alternatively you might use:

```
def main():
  <actual top-level code here>

if __name__ == '__main__':
  main()
```

That way the variables in the body of main() will be local. As it is now, *args* and *run_opts*, for example, are in fact visible in all functions. This is usually is not a problem in my experience, but once in a while may cause some head-scratching.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,1132 @@
...
+    # we add compulsary arguments as named arguments for readability
+    parser = argparse.ArgumentParser(description="""
```
vdp: In my Python code I prefer to put the command line parsing code in a separate function and place it at the top, for easier reading. In fact as I think of this now, it might be better to place the main() code at the top as well(perhaps after option parsing), especially when the script is relatively long as in this case, because this is the logical place from which to start reading the code.
By the way, and I'm sure you know this already, you could add **formatter_class=argparse.ArgumentDefaultsHelpFormatter** as a parameter to ArgumentParser's constructor- will print out the default value for the options, when you call the script with "--help".

vijayaditya: Thanks for the suggestion I will do this.

egs/wsj/s5/steps/nnet3/train_rnn.py
vijayaditya: Thanks a lot for pointing this out. I didn't know that definitions in __main__ were global.

vijayaditya:
I have removed most of the bugs from the script. Training seems to run fine. I will be polishing the script further. But I think there will not be major changes and hence the script is ready for review.

I decided to create a kaldi_lib.py with training steps which are shared across feed-forward DNN and RNN training. Though this violates the initial goal of having a self-contained script it helps users identify steps which are unique to FF-DNN and RNN, as they will be coded in train_dnn.py and train_rnn.py.
I think this will help us better maintain the scripts in the future.

I have moved the make_configs.py script from train script to the local/nnet3/run_lstm.sh scripts, to ensure we can reuse the train_rnn.py for different RNN architectures. I will be doing the same for train_dnn.py too.

I am not too interested in changing decode.sh script as it already small and we will not gain a lot in terms of readability.

I am working on lstm chain script now, for which I will reuse a lot of steps from kaldi_lib.py

danpovey:
maybe nnet3_train_lib.py would be a better name than kaldi_lib.py

vijayaditya:
Yes that is better, modifying the name.

vijayaditya:
Tested everything including the realignment pipeline. Realignment pipeline does not work due to issues raised in #495 . There are no more changes from my side, please review. 

Note : Experiment to reproduce results is running.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,657 @@
...
+                        help="If the derivative averages are below this threshold we scale the parameter matrices with the shrink-value. It is less than 0.25 for sigmoid non-linearities.")
+
+    # RNN specific trainer options
+    parser.add_argument("--trainer.rnn.num-chunk-per-minibatch", type=int, dest='num_chunk_per_minibatch',
+                        default=100,
+                        help="Number of sequences to be processed in parallel every minibatch" )
+    parser.add_argument("--trainer.rnn.num-bptt-steps", type=int, dest='num_bptt_steps',
+                        default=None,
+                        help="The number of time steps to back-propagate from the last label in the chunk. By default it is same as the chunk-width." )
+    parser.add_argument("--trainer.rnn.shrink", type=float, dest='shrink',
```
danpovey: just a note on shrinking, although it won't affect this pull request for now.
I am working in ~dpovey/kaldi-chain/ on a more elegant mechanism called 'self-repairing nonlinearities'.  It's an option in the nonlinearities (ReLU, sigmoid, tanh) where once the average value or derivative stats (as appropriate) get out of the 'good' range, it adds a small term to the derivative to try to force it back to a good range.  Hopefully this will work better than the shrinking mechanism- certainly it's more general.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,657 @@
+#!/usr/bin/env python
+
```
danpovey: There needs to be a copyright notice at the top of these files (just include your own name).

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -0,0 +1,622 @@
...
+    print num_iters_combine
+    for iter in range(num_iters - num_iters_combine + 1, num_iters + 1):
+      model_file = '{0}/{1}.mdl'.format(dir, iter)
+      if not os.path.exists(model_file):
+          raise Exception('Model file {0} missing'.format(model_file))
+      raw_model_strings.append('"nnet3-am-copy --raw=true {0} -|"'.format(model_file))
+
+    combine_num_chunk_per_minibatch = int(1024.0/(chunk_width))
+    
+   # Below, we use --use-gpu=no to disable nnet3-combine-fast from using a GPU,
```
danpovey: this comment is out of date, remove it.

danpovey: ... also from the baseline bash script.

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -0,0 +1,622 @@
...
+                 remove_egs = True):
+    try:
+        if remove_egs:
+            RemoveEgs(egs_dir)
+
+        for iter in range(num_iters):
+            print("deleting iter {0}".format(iter))
+            RemoveModel(nnet_dir, iter, num_iters, 1,
+                        preserve_model_interval)
+    except (IOError, OSError) as err:
```
danpovey: i see no need for try-except here as there seem to be no expected errors-- or perhaps I'm missing something.  anyway it's just re-throwing the error.

vijayaditya: I am catching the exception just to add some logging info, nothing more than that.

danpovey:
How well tested is this?

vijayaditya:
The experiment where I try to recreate the results of the train.sh script
is near  completion.

Vijay

vijayaditya:
Tested end-to-end.

Added new reporting capability, where the training script sends periodic
update emails, if asked to, about the progress of the experiment or any error which
kills the experiment.

vijayaditya:
any other comments/suggestions ?

danpovey:
Regarding the periodic update emails-- bear in mind that not all clusters
will allow you to send email from the command line, and configuring email
is becoming increasingly hard due to spam-fighting efforts.  We don't want
to get into a situation where people say "why doesn't it work" and we have
to explain to them how to install postfix and work around local firewall
policies.  Perhaps wording it right will be enough?

Dan

vijayaditya:
OK, I will add a disclaimer saying this is experimental and needs local
expertise.

Also, when there is an error in sending the email the logger just reports
inability to send the mail and continues with the training.

--Vijay

vijayaditya:
added a disclaimer near the email option.

<pre><code>
+                        help=""" Email-id to report about the progress of the experiment.
 +                              NOTE: This is a very experimental feature. It assumes
 +                              the machine on which the script is being run can send
 +                              emails from command line via. mail program. The
 +                              Kaldi mailing list will not support this feature.
 +                              It might require local expertise to setup. """)
</code></pre>

danpovey:
OK, but not sure that "experimental feature" is the right wording, maybe
remove that sentence.

vijayaditya:
Done.

danpovey:
ok but not ready to merge yet- will look again when I get a chance

vijayaditya:
Could I get some comments on this PR. @tomkocse and @freewym  are close to completing their CNN and GRU recipes and it will be preferable if we ensure that there is no rework to later translate these recipes to use the python scripts.

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -0,0 +1,618 @@
...
+
+def ForceSymlink(file1, file2):
+    try:
+        os.symlink(file1, file2)
+    except OSError, e:
+        if e.errno == errno.EEXIST:
+            os.remove(file2)
+            os.symlink(file1, file2)
+
+def ComputePresoftmaxPriorScale(dir, alidir, num_jobs, run_opts,
```
danpovey: check if we're using this in any recipes- if not we could remove it.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -0,0 +1,703 @@
...
+    try:
+        Train(args, run_opts)
+    except Exception as e:
+        if args.email is not None:
+            message = "Training session for experiment {dir} died due to an error.".format(dir = args.dir)
+            sendMail(message, message, args.email)
+        traceback.print_exc()
+        raise e
+         
+def sendMail(message, subject, email_id):
```
danpovey: is there a reason this is not SendMail?

danpovey:
these issues are small so I'll merge this.

danpovey:
oops, I see there are conflicts, I can't merge.

egs/wsj/s5/steps/nnet3/train_rnn.py
vijayaditya: Changing..

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
vijayaditya: We are using it in the TDNN recipes. This file is shared by both train_dnn.py and train_rnn.py

vijayaditya:
resolving the conflicts.

vijayaditya:
resolved all conflicts.

PR #491, opened 2016-02-08
added I/O interfaces for NnetComputation and ComputationRequest.

xiaohui-zhang:
(only added the script level changes for egs/wsj/s5/steps/nnet3/lstm/train.sh for now)

### DIFF
--- a/egs/wsj/s5/steps/nnet3/lstm/train.sh
+++ b/egs/wsj/s5/steps/nnet3/lstm/train.sh
@@ -560,10 +560,12 @@ while [ $x -lt $num_iters ]; do
       cur_num_hidden_layers=$[1+$x/$add_layers_period]
       config=$dir/configs/layer$cur_num_hidden_layers.config
       raw="nnet3-am-copy --raw=true --learning-rate=$this_learning_rate $dir/$x.mdl - | nnet3-init --srand=$x - $config - |"
+      cache_io_opts="--read-cache=$dir/cache.$x"
     else
       do_average=true
       if [ $x -eq 0 ]; then do_average=false; fi # on iteration 0, pick the best, don't average.
       raw="nnet3-am-copy --raw=true --learning-rate=$this_learning_rate $dir/$x.mdl -|"
+      cache_io_opts=""
     fi
     if $do_average; then
       this_num_chunk_per_minibatch=$num_chunk_per_minibatch
@@ -594,9 +596,7 @@ while [ $x -lt $num_iters ]; do
                                                # the other indexes from.
         archive=$[($k%$num_archives)+1]; # work out the 1-based archive index.
         if [ $n -eq 1 ]; then
-          cache_io_opts="--read-cache=$dir/cache.$x --write-cache=$dir/cache.$[$x+1]" # opts for cache (storing pairs of nnet-computations and compute-requests) I/O
-        else 
-          cache_io_opts="--read-cache=$dir/cache.$x"
+          cache_io_opts=$cache_io_opts" --write-cache=$dir/cache.$[$x+1]" # opts for cache (storing pairs of nnet-computations and compute-requests) I/O
         fi
         $cmd $train_queue_opt $dir/log/train.$x.$n.log \
           nnet3-train $parallel_train_opts $cache_io_opts --print-interval=10 --momentum=$momentum \
--- a/egs/wsj/s5/steps/nnet3/lstm/train.sh
+++ b/egs/wsj/s5/steps/nnet3/lstm/train.sh
@@ -593,8 +593,13 @@ while [ $x -lt $num_iters ]; do
         k=$[$num_archives_processed + $n - 1]; # k is a zero-based index that we will derive
                                                # the other indexes from.
         archive=$[($k%$num_archives)+1]; # work out the 1-based archive index.
+        if [ $n -eq 1 ]; then
+          cache_io_opts="--read-cache=$dir/cache.$x --write-cache=$dir/cache.$[$x+1]" # opts for cache (storing pairs of nnet-computations and compute-requests) I/O
+        else 
+          cache_io_opts="--read-cache=$dir/cache.$x"
+        fi
         $cmd $train_queue_opt $dir/log/train.$x.$n.log \
-          nnet3-train $parallel_train_opts --print-interval=10 --momentum=$momentum \
+          nnet3-train $parallel_train_opts $cache_io_opts --print-interval=10 --momentum=$momentum \
           --max-param-change=$max_param_change \
           --optimization.min-deriv-time=$min_deriv_time "$raw" \
           "ark:nnet3-copy-egs $context_opts ark:$cur_egs_dir/egs.$archive.ark ark:- | nnet3-shuffle-egs --buffer-size=$shuffle_buffer_size --srand=$x ark:- ark:-| nnet3-merge-egs --minibatch-size=$this_num_chunk_per_minibatch --measure-output-frames=false --discard-partial-minibatches=true ark:- ark:- |" \
@@ -643,6 +648,7 @@ while [ $x -lt $num_iters ]; do
   fi
   x=$[$x+1]
   num_archives_processed=$[$num_archives_processed+$this_num_jobs]
+  rm $dir/cache.$x
 done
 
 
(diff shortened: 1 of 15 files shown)


danpovey:
Thanks.  Have you done any testing of this?  Was there any impact at all on
speed?

Dan

xiaohui-zhang:
I've tested it on LSTM training to make sure it worked. Haven't observed the change in speed yet. will do.

src/nnet3/nnet-optimize.cc
```diff
@@ -432,6 +433,134 @@ void CachingOptimizingCompiler::UpdateCache(const ComputationRequest *request,
                             std::make_pair(computation, ait)));
 }
 
+void CachingOptimizingCompiler::ReadCache(std::istream &is, bool binary) {
+  // we won't read cached computations if any optimize option has been changed.
+  bool read_cache = true;
+  ExpectToken(is, binary, "<NnetOptimizeOptions>");
```
danpovey: It would be more straightforward to give a Read and Write function and an == operator to the options class.

src/nnet3/nnet-optimize.cc
```diff
@@ -432,6 +433,134 @@ void CachingOptimizingCompiler::UpdateCache(const ComputationRequest *request,
...
+  WriteToken(os, binary, "<AllocateFromOther>");
+  WriteBasicType(os, binary, opt_config_.allocate_from_other);
+  WriteToken(os, binary, "<MinDerivTime>");
+  WriteBasicType(os, binary, opt_config_.min_deriv_time);
+  WriteToken(os, binary, "<MaxDerivTime>");
+  WriteBasicType(os, binary, opt_config_.max_deriv_time);
+  WriteToken(os, binary, "</NnetOptimizeOptions>");
+
+  WriteToken(os, binary, "<ComputationCacheSize>");
+  WriteBasicType(os, binary, computation_cache_.size());
```
danpovey: It's dangerous to write types of machine-dependent type directly to disk like this.  Use a temporary int32.

src/nnet3/nnet-optimize.cc
```diff
@@ -23,6 +24,83 @@
...
+  ExpectToken(is, binary, "<MaxDerivTime>");
+  ReadBasicType(is, binary, &max_deriv_time);
+  ExpectToken(is, binary, "</NnetOptimizeOptions>");
+}
+
+void NnetOptimizeOptions::Write(std::ostream &os, bool binary) const {
+  WriteToken(os, binary, "<NnetOptimizeOptions>");
+  WriteToken(os, binary, "<Optimize>");
+  WriteBasicType(os, binary, optimize);
+  WriteToken(os, binary, "<Consolidate_model_update>");
```
danpovey: why the underscores?

src/nnet3/nnet-optimize.cc
```diff
@@ -428,10 +504,46 @@ void CachingOptimizingCompiler::UpdateCache(const ComputationRequest *request,
     access_queue_.pop_front();
   }
   AqType::iterator ait = access_queue_.insert(access_queue_.end(), request);
+  KALDI_LOG << "inserting new pair";
```
danpovey: remove this line

danpovey:
Thanks!   Merging.

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -593,8 +595,11 @@ while [ $x -lt $num_iters ]; do
         k=$[$num_archives_processed + $n - 1]; # k is a zero-based index that we will derive
                                                # the other indexes from.
         archive=$[($k%$num_archives)+1]; # work out the 1-based archive index.
+        if [ $n -eq 1 ]; then
+          cache_io_opts=$cache_io_opts" --write-cache=$dir/cache.$[$x+1]" # opts for cache (storing pairs of nnet-computations and compute-requests) I/O
```
danpovey: I just realized there is a bug here, as the same cache_io_opts variable will be used for n > 1.

danpovey: Also please put $cache_io_opts inside the quotes.

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -643,6 +648,7 @@ while [ $x -lt $num_iters ]; do
   fi
   x=$[$x+1]
   num_archives_processed=$[$num_archives_processed+$this_num_jobs]
+  rm $dir/cache.$x
```
danpovey: please do this with 2>/dev/null

egs/wsj/s5/steps/nnet3/lstm/train.sh
xiaohui-zhang: sure. Actually I defined cache_io_opts earlier in the script so that problem shouldn't happen. But I should move the comment upward.

danpovey: I think you're mistaken-- think about the loop over n.  You're making all
the jobs write to the cache.

egs/wsj/s5/steps/nnet3/lstm/train.sh
danpovey: this is a bug too, that you are removing $dir/cache.$x after incrementing x.  It means you delete the one you are going to need.
Did you ever test this script?

egs/wsj/s5/steps/nnet3/lstm/train.sh
```diff
@@ -560,10 +560,12 @@ while [ $x -lt $num_iters ]; do
       cur_num_hidden_layers=$[1+$x/$add_layers_period]
       config=$dir/configs/layer$cur_num_hidden_layers.config
       raw="nnet3-am-copy --raw=true --learning-rate=$this_learning_rate $dir/$x.mdl - | nnet3-init --srand=$x - $config - |"
+      cache_io_opts="--read-cache=$dir/cache.$x"
```
danpovey: You should be setting this cache_io_opts to "" and the other one to this value- you have switched them.

xiaohui-zhang: sorry my bad.. I confused this script and another script I tested, which
excludes these bugs... will update soon.

On Wed, Feb 10, 2016 at 8:30 PM, Daniel Povey <notifications@github.com>
wrote:

> In egs/wsj/s5/steps/nnet3/lstm/train.sh
> <https://github.com/kaldi-asr/kaldi/pull/491#discussion_r52554998>:
>
> > @@ -560,10 +560,12 @@ while [ $x -lt $num_iters ]; do
> >        cur_num_hidden_layers=$[1+$x/$add_layers_period]
> >        config=$dir/configs/layer$cur_num_hidden_layers.config
> >        raw="nnet3-am-copy --raw=true --learning-rate=$this_learning_rate $dir/$x.mdl - | nnet3-init --srand=$x - $config - |"
> > +      cache_io_opts="--read-cache=$dir/cache.$x"
>
> You should be setting this cache_io_opts to "" and the other one to this
> value- you have switched them.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/491/files#r52554998>.
>



-- 
Xiao-hui

src/nnet3/nnet-computation.cc
```diff
@@ -414,6 +648,209 @@ void NnetComputation::Print(std::ostream &os, const Nnet &nnet) const {
...
+  KALDI_ASSERT(num_input_output_info >= 0);
+  input_output_info.clear();
+  ExpectToken(is, binary, "<InputOutputInfo>");
+  for (size_t c = 0; c < num_input_output_info; c++) {
+    int32 key;
+    std::pair<int32, int32> val;
+    ReadBasicType(is, binary, &key);
+    ReadBasicType(is, binary, &(val.first));
+    ReadBasicType(is, binary, &(val.second));
+    input_output_info.insert(std::make_pair<int32, std::pair<int32, int32> >(key, val));
```
gramonov: I noticed that this would break compilation in C++11, since `key` and `val` are l-values. In C++11 this becomes
`std::make_pair(key, val);`

Thanks,

danpovey: I think there are other issues with make_pair, we've tried to avoid it.
Samuel, you should be using the following, which invokes the constructor:

std::pair<int32, std::pair<int32, int32> >(key, val)

xiaohui-zhang: Thanks. Have committed a fix. Dan, can you remind me what else issues does make_pair have? I'll avoid using it in the future.

danpovey: It tends to cause compilation errors with some compiler versions, I forget
the exact reasons but we always use std::pair constructor.

PR #496, opened 2016-02-12
Consider user-defined environment variables

mbait:
Put kaldi-defined compiler and linker flags before user-defined flags so the
former could be overriden.

### DIFF
--- a/src/makefiles/linux_atlas.mk
+++ b/src/makefiles/linux_atlas.mk
@@ -13,7 +13,7 @@ $(error ATLASLIBS not defined.)
 endif
 
 
-CXXFLAGS = -msse -msse2 -Wall -I.. \
+KALDI_CXXFLAGS = -msse -msse2 -Wall -I.. \
 	   -pthread \
       -DKALDI_DOUBLEPRECISION=0 -DHAVE_POSIX_MEMALIGN \
       -Wno-sign-compare -Wno-unused-local-typedefs -Winit-self \
@@ -24,11 +24,12 @@ CXXFLAGS = -msse -msse2 -Wall -I.. \
       -g # -O0 -DKALDI_PARANOID 
 
 ifeq ($(KALDI_FLAVOR), dynamic)
-CXXFLAGS += -fPIC
+KALDI_CXXFLAGS += -fPIC
 endif
 
-LDFLAGS = -rdynamic $(OPENFSTLDFLAGS)
-LDLIBS = $(EXTRA_LDLIBS) $(OPENFSTLIBS) $(ATLASLIBS) -lm -lpthread -ldl
+CXXFLAGS := $(KALDI_CXXFLAGS) $(CXXFLAGS)
+LDFLAGS := -rdynamic $(OPENFSTLDFLAGS) $(LDFLAGS)
+LDLIBS := $(EXTRA_LDLIBS) $(OPENFSTLIBS) $(ATLASLIBS) -lm -lpthread -ldl $(LDLIBS)
 CC = g++
 CXX = g++
 AR = ar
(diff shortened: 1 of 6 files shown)


danpovey:
I'd like to think about this for a while rather than merge it right away.

src/makefiles/linux_atlas.mk
```diff
@@ -24,11 +24,12 @@ CXXFLAGS = -msse -msse2 -Wall -I.. \
       -g # -O0 -DKALDI_PARANOID 
 
 ifeq ($(KALDI_FLAVOR), dynamic)
-CXXFLAGS += -fPIC
+KALDI_CXXFLAGS += -fPIC
 endif
 
-LDFLAGS = -rdynamic $(OPENFSTLDFLAGS)
-LDLIBS = $(EXTRA_LDLIBS) $(OPENFSTLIBS) $(ATLASLIBS) -lm -lpthread -ldl
+CXXFLAGS := $(KALDI_CXXFLAGS) $(CXXFLAGS)
```
danpovey: if we're going to do this I think it's ugly to  have EXTRA_CXXFLAGS be  included in KALDI_CXXFLAGS-- it would be clearer, IMO, to do
CXXFLAGS := $(KALDI_CXXFLAGS) $(EXTRA_CXXFLAGS) $(CXXFLAGS)
and remove EXTRA_CXXFLAGS from KALDI_CXXFLAGS.

src/makefiles/linux_cuda.mk
```diff
@@ -2,7 +2,8 @@
 CUDA_INCLUDE= -I$(CUDATKDIR)/include
 CUDA_FLAGS = -g -Xcompiler -fPIC --verbose --machine 32 -DHAVE_CUDA
 
-CXXFLAGS += -DHAVE_CUDA -I$(CUDATKDIR)/include 
-LDFLAGS += -L$(CUDATKDIR)/lib -Wl,-rpath=$(CUDATKDIR)/lib
-LDLIBS += -lcublas -lcudart #LDLIBS : The libs are loaded later than static libs in implicit rule
+CXXFLAGS := -DHAVE_CUDA -I$(CUDATKDIR)/include $(CXXFLAGS)
```
danpovey: is it really necessary for these types of flags, to have them added at the beginning of CXXFLAGS?  There's no way you'd want to override these.

src/makefiles/linux_cuda.mk
```diff
@@ -2,7 +2,8 @@
...
+LDFLAGS := -L$(CUDATKDIR)/lib -Wl,-rpath=$(CUDATKDIR)/lib $(LDFLAGS)
```
danpovey: Again, I don't see why it's necessary to repeat this pattern for all types of flags, to add things at the end rather than the beginning
Plus, the way you're doing it with LDLIBS could possibly make a difference to the linking order which might make the difference between compiling and not compiling for some setups.
Actually now I see how extensive the changes are that you want to make, I'm starting to have second throughts about this pull request.  All this stuff is hard to test and hard to debug when there are failures because we don't have access to all platforms.

mbait:
I include `$(CXXFLAGS)` **after** the Kaldi-specified flags to be able to override the defaults.  But you're right that this is a bit slippery. An alternative to this way could be substituting a particular variable completely with the `?=` type of assignment. What do you think?

danpovey:
I think that is dangerous with these types of things because surely
CXXFLAGS and the like will tend to have defaults.
I'm moving toward the point of view that we should just leave this alone.
kaldi.mk was always intended to be human editable.  If you want to
configure it differently you are supposed to either edit kaldi.mk by hand,
or modify the configure script or the prototype makefiles if that's what
you want to do.  This is not very difficult to do.   Or you could add in
kaldi.mk,

CXXFLAGS += your_special_flags


Dan

jtrmal:
I didn;t manage to read the whole discussion, but my two cents are that
given the fact that we do not use any complex makefile generator (such as
the autotools or cmake or whatever) I'm not that sure if including some
additional env variables will bring any sufficient benefit -- you can
modify the makefiles directly by hand (or automatically) and I don't see
any significant conceptual difference.
y

danpovey:
Guys, to avoid wasting our time I'm going to close it.  I don't think this is going to get merged.  Sorry.

PR #497, opened 2016-02-12
xvector

david-ryan-snyder:
One deviation between this and the interface we had discussed earlier: ComputeXvectorObjfAndDeriv originally had a single output parameter for the deriv. I split this into two parts, one for the deriv with respect to the xvectors, and another for the S and b parameters. My reasoning is that the S and b parameters are global (so, one derivative vector), whereas the xvectors are dependent on the input.

### DIFF
--- a/src/ivector/xvector-test.cc
+++ b/src/ivector/xvector-test.cc
@@ -1,0 +1,131 @@
+// ivector/xvector-test.cc
+
+// Copyright 2016  Daniel Povey
+//                 David Snyder
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "ivector/xvector.h"
+#include "util/kaldi-io.h"
+#include "cudamatrix/cu-matrix-lib.h"
+
+namespace kaldi {
+
+bool TestXvectorExtractorDerivative(BaseFloat perturb_delta) {
+  int32 xvector_dim = RandInt(4, 30),
+        num_rows = 2 * RandInt(2, 10); // The number of rows must be even
+                                       // and greater than 2.
+  CuSpMatrix<BaseFloat> S(xvector_dim);
+  S.SetRandn();
+  // Necessary to keep the similarity scores from getting too large or small.
+  S.Scale(1.0e-01);
+  BaseFloat b = RandInt(-100, 100) / 10.0,
+            tot_weight,
+            tot_objf;
+  int32 S_dim = S.NumCols() * (S.NumCols() + 1) / 2;
+  CuMatrix<BaseFloat> xvector_pairs(num_rows, xvector_dim, kSetZero),
+                      deriv_xvector(num_rows, xvector_dim, kSetZero);
+  CuVector<BaseFloat> deriv_S_and_b(S_dim + 1, kSetZero);
+  xvector_pairs.SetRandn();
+  ComputeXvectorObjfAndDeriv(xvector_pairs, S, b, &deriv_xvector,
+    &deriv_S_and_b, &tot_objf, &tot_weight);
+  CuVector<BaseFloat> deriv_xvector_vec(xvector_dim);
+
+  // Sum over the derivatives for xvector input.
+  deriv_xvector_vec.AddRowSumMat(1.0, deriv_xvector, 0.0);
+  BaseFloat l2_xvector = 0,
+            l2_S = 0,
+            l2_b = 0;
+
+  // Compare the xvector derivatives calculated above with a numerical
+  // approximation.
+  for (int32 i = 0; i < xvector_dim; i++) {
+    CuMatrix<BaseFloat> xvector_pairs_p(xvector_pairs);
+    CuMatrix<BaseFloat> xvector_pairs_n(xvector_pairs);
+    for (int32 j = 0; j < num_rows; j++) {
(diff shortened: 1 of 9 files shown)


src/ivector/xvector.cc
```diff
@@ -0,0 +1,140 @@
...
+    KALDI_ASSERT(deriv_S_and_b == NULL);
+  else {
+    KALDI_ASSERT(deriv_xvector->NumCols() == xvector_dim);
+    KALDI_ASSERT(deriv_xvector->NumRows() == N);
+    KALDI_ASSERT(deriv_S_and_b->Dim() == S_dim + 1);
+  }
+
+  // Handle portion of the objf corresponding to pairs of xvectors
+  // from the same classes.
+  for (int32 i = 0; i < N/2; i++) {
```
danpovey: This looks plausible from the point of view of the math, but it's very against the spirit of CUDA programming to have loops like this.  The idea is to have all the parallel stuff happening on the GPU card, so you go to great lengths to do things in as few matrix operations as possible.   But I think it's best if you push on with the rest of this project and let me think about the best way to do this- I'll let you know when I've figured out an efficient way to do it.

src/ivector/xvector-test.cc
```diff
@@ -0,0 +1,131 @@
+// ivector/xvector-test.cc
+
+// Copyright 2016  Daniel Povey
```
danpovey: just have your own name on this file.
only include me in the header, but since I'm an employee of JHU you'd have to put
//  Copyright 2016    Johns Hopkins University  (Author: Daniel Povey)
                    2016    David Snyder

src/ivector/xvector.cc
```diff
@@ -0,0 +1,140 @@
...
+
+namespace kaldi {
+
+void ComputeXvectorObjfAndDeriv(
+    const CuMatrixBase<BaseFloat> &xvector_pairs,
+    const CuSpMatrix<BaseFloat> &S,
+    BaseFloat b, CuMatrixBase<BaseFloat> *deriv_xvector,
+    CuVector<BaseFloat> *deriv_S_and_b, BaseFloat *tot_objf,
+    BaseFloat *tot_weight) {
+
```
danpovey: OK, here's a plan for more efficient computation.
firstly, in the following I have not thought carefully about variable names.
you can if you want keep the old version for debugging purposes somewhere- e.g. put it in the -test.cc file.
- copy S to a temporary CuMatrix (since we don't have matrix-multiply with symmetric-packed on CUDA).
- compute the product P =  xvector_pairs * S. 
- compute two matrices: Q = xvector_pairs * xvector_pairs',
   and R = xvector_pairs * P'.
Note: if you want you can use SymAddMat2 in computing Q, since the args are identical, and then do CopyLowerToUpper- it might be slightly faster.
- compute T = Q - R + b, which is the matrix of scores (log-odds).

At this point, define a function to compute the objf from T and the derivs w.r.t. the elements of T.  For now let's ignore the fact that it's symmetric and compute things for both the upper and lower parts, and later divide by 2-- anyway this phase likely won't dominate.
Here is my draft of the documentation for this function:
/*  This function is used in computing the objective function and derivatives in 
    xvector training.
   @param [in] scores   'scores' is a symmetric matrix of scores which are to be
              interpreted as log-odds (according to the model) of pairs coming from the
              same class, so scores(i, j) is the model's log p(same/different) for elements i and j
              of the original minibatch of input.
   @param [out] objf_terms   objf_terms is a matrix of the same dimension as 'scores
               whose elements we will sum to get the objective function for this minibatch.  
              We assume that the data in 'scores' has been arranged in such
              a way that indexes (0, 1) are from the same class; (2, 3) are from the same class and
              so on, but the classes are otherwise unique.   This function computes the appropriate
              contributions to the objective function, as follows. 
                   if i == j:   objf_terms(i, j)== 0       # the same exact element is not scored
                   elsif i%2 == j%2:  objf_terms(i, j) = log(p(same)) = -log(1 + exp(-scores(i, j))
                   else:    objf_terms(i, j) = 1/scores.NumRows() * log(p(different))
                                                      = -1/scores.NumRows() * log(1+exp(scores(i,j))
    @param [out] objf_derivs    Element (i,j) of this matrix is the derivative of objf_terms(i,j)
                                         with respect to scores(i, j).  */
  void ComputeXvectorObjfFromScores(const CuMatrixBase<BaseFloat> &scores,
                                                                CuMatrixBase<BaseFloat> *objf_terms,
                                                                CuMatrixBase<BaseFloat> *objf_derivs);
 Note: this function will have its own CUDA kernel.  Pegah can help with this if you need.
 You can actually declare this function in cu-math.h and define it in cuda-math.cc.  
 Copy and modify the code for the Sigmoid kernel- but you'll have some if-statements
 and you'll need to add new function arguments.  Actually maybe changing the DiffSigmoid
 kernel and making one of the inputs an output would be easier, since that also will have
 3 matrix arguments.   Inside the .cc file, after the #endif, instead of backing off to
 another named function you can just put the non-CUDA version of the code right
 there in the function.  The files you'll have to edit for this are cu-math.{h,cc},
 cu-kernels-ansi.h, cu-kernels.h and (most crucially) cu-kernels.cu.  It will be 
 mostly copy-and-modify; search for diff_sigmoid.

----
OK, now let's return to ComputeXvectorObjfAndDeriv.  After you have computed the matrix
of objf_terms you can just sum it to compute the actual objective function (let 'tot_weight'
be S.NumRows() and fix the documentation).  And as for objf_derivs, you can use that
to start backpropagating through the bits of the computation that you just did, to compute
derivatives w.r.t xvector_pairs and S and b.  Again you'll have to be careful about the symmetry in
S; the function ScaleDiag in class PackedMatrix may come in useful here.

Your current code can all go in the -test.cc file.

src/ivector/xvector.cc
```diff
@@ -0,0 +1,140 @@
...
+    CuVector<BaseFloat> *deriv_v, CuVector<BaseFloat> *deriv_w,
+    CuVector<BaseFloat> *deriv_S_and_b) {
+  int32 d = is_same ? 1 : -1,
+        S_dim = S.NumCols() * (S.NumCols() + 1) / 2;
+  deriv_v->Resize(v.Dim(), kSetZero);
+  deriv_w->Resize(v.Dim(), kSetZero);
+  deriv_S_and_b->Resize(S_dim + 1, kSetZero);
+
+  // This scalar is common to the different derivatives.
+  BaseFloat deriv_coef = d * Exp(-1 * d * similarity_score)
```
danpovey: this can be simplified to d / (1 + Exp(d * similarity_score).

danpovey:
David, in the hope of accelerating work on this I'm creating a branch 'xvector' in my personal repo git@github.com:danpovey/kaldi.git.  Can you please submit a  pull request to there?  (I think personal repositories accept them too).  It's OK to include partial work.  I'm thinking of sketching out some of the scripts.

danpovey:
closing as moved to my own repo.

PR #499, opened 2016-02-14
If compiling with MSVS 2015, dont redefine snprintf

DeepHorizons:
As of Windows 10 and MSVS 2015, `snprintf` is no longer identical to `_snprintf`

This PR checks for MSVS 2015 or greater, will not redefine `snprintf` if so

### DIFF
--- a/src/base/kaldi-utils.cc
+++ b/src/base/kaldi-utils.cc
@@ -20,7 +20,9 @@
 #include <Synchapi.h>
 #elif defined(_WIN32) || defined(_MSC_VER) || defined(MINGW)
 #include <Windows.h>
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks!

jtrmal:
Thanks, merging.
y.

danpovey:
Does the preprocessor do lazy evaluation?
I'm just wondering what happens to the following when _MSC_VER is not
defined.
+#if defined(_MSC_VER) && _MSC_VER < 1900

jtrmal:
I believe at least the gcc/clang do. y.

DeepHorizons:
A quick test right now says that an undefined preprocessor variable wont crash (and evaluates to false)

PR #512, opened 2016-02-19
Do not overwrite user flags

mbait:


danpovey:
I think it's arguable whether or not this is is a good idea (as in the normal case that flag would not be set, and the user will wonder why we're doing +=), but if @jtrmal doesn't object I am OK to merge this because I don't see that it would cause a problem.

mbait:
Actually, I don't get why `EXTRA_CXXFLAGS` is overwritten here, because in kaldi.mk it's supposed to be user-defined, and thus adding that option `CXXFLAGS` would be more consistent. If you don't accept this patch, please suggest the proper way to pass the user flags.

danpovey:
In the individual makefiles, EXTRA_CXXFLAGS is set before including kaldi.mk.
It's a mechanism for the Makefiles in the individual directories to pass
directory-specific flags into the compilation process.
The way you are supposed to change the flags that Kaldi is compiled with,
is by editing kaldi.mk, or editing the prototype Makefile that produces it
if that's what you prefer.  kaldi.mk is very simple, it's not like a file
produced by automake.  It's supposed to be like a Makefile that a human
would write-- the configure script just helps you create that file for the
common cases.

Dan

danpovey:
Is there something about your workflow that means you can't just edit
kaldi.mk?

mbait:
> Is there something about your workflow that means you can't just edit
kaldi.mk?

Yes, I'm adding Kaldi to an automatic build system which calls 'configure'
and then 'make'. I could make a script to generate or adjust 'kaldi.mk',
but how do I know that it won't break after the next pull from the upstream?

I didn't notice that kaldi.mk is included after those variable assignments.
I think it's better to not accept my patch then.

On Thu, Feb 18, 2016 at 9:26 PM, Daniel Povey <notifications@github.com>
wrote:

> Is there something about your workflow that means you can't just edit
> kaldi.mk?
>
>
>
> On Fri, Feb 19, 2016 at 12:24 AM, Daniel Povey <dpovey@gmail.com> wrote:
>
> > In the individual makefiles, EXTRA_CXXFLAGS is set before including
> > kaldi.mk. It's a mechanism for the Makefiles in the individual
> > directories to pass directory-specific flags into the compilation
> process.
> > The way you are supposed to change the flags that Kaldi is compiled with,
> > is by editing kaldi.mk, or editing the prototype Makefile that produces
> > it if that's what you prefer. kaldi.mk is very simple, it's not like a
> > file produced by automake. It's supposed to be like a Makefile that a
> > human would write-- the configure script just helps you create that file
> > for the common cases.
> >
> > Dan
> >
> >
> > On Fri, Feb 19, 2016 at 12:19 AM, Alexander Solovets <
> > notifications@github.com> wrote:
> >
> >> Actually, I don't get why EXTRA_CXXFLAGS is overwritten here, because in
> >> kaldi.mk it's supposed to be user-defined, and thus adding that option
> >> CXXFLAGS would be more consistent. If you don't accept this patch,
> >> please suggest the proper way to pass the user flags.
> >>
> >> —
> >> Reply to this email directly or view it on GitHub
> >> <https://github.com/kaldi-asr/kaldi/pull/512#issuecomment-186055741>.
> >>
> >
> >
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/512#issuecomment-186058277>.
>



-- 
Sincerely, Alexander

danpovey:
Is it only the CXXFLAGS that you need to change, or are there others?
One possible solution is for you to, after "configure", do

echo "CXXFLAGS += USER_CXXFLAGS" >> kaldi.mk

and then set USER_CXXFLAGS from your Makefile.

We could certainly bear this usage pattern in mind so we would avoid doing
things with CXXFLAGS that would break this.

PR #513, opened 2016-02-21
Fix processing of phonemes when adding position dependency symbols:

OliverWalter:
- Revert to old behaviour (for x in cat `...`)
 - (while read x) does not split the line if it contains more than one phoneme
 - In case of stress marks UW UW0 UW1... the whole line woule be taken as one phoneme
 - This broke the WSJ recipe.... (and maybe others)
 - To avoid problems with phonemes containg * (or other symbols like '?') disable globbing by (set -f)


danpovey:
Oops.  I should have checked more carefully.

PR #514, opened 2016-02-22
Added plotting feature to the steps/nnet3/report scripts. Refactored …

vijayaditya:
…scripts.

Generates plots and tables. Plots are generated only if matplotlib is installed, else only tables are generated.

Non-linearity stats plots, accuracy plots, log-likelihood plots are generated. In addition to the logs corresponding to these plots the parameter difference tables are also generated.


danpovey:
merging since this won't break anything

PR #522, opened 2016-02-23
Add argument name while calling python functions

vimalmanohar:
@vijayaditya @danpovey @freewym 
I think all the python files must be changed so that calling script must add the argument name while passing the argument's value. This way reordering of arguments in the called function need not be propagated to the calling script.


vijayaditya:
Good idea we should probably have a set of guidelines to ensure
contributors would could have a reference in the future. I will try to
write this up.

--VIjay

danpovey:
I'd wait on writing it up for things to stabilize- not clear where to put
it anyway.

PR #525, opened 2016-02-25
WIP : Added python scripts for chain model training

vijayaditya:
I am trying to recreate the results for LSTM+chain using python scripts. I am not committing the original shell scripts used to generate the results shown in RESULTS as they were not updated with the latest options. 

I will add self-repair which was shown to improve LSTM+X-ent results to this python recipe before finalizing the commit.

This PR has been mainly written for initial review.

### DIFF
--- a/egs/swbd/s5c/local/chain/run_lstm_d.sh
+++ b/egs/swbd/s5c/local/chain/run_lstm_d.sh
@@ -1,0 +1,216 @@
+#!/bin/bash
+
+# based on run_tdnn_2o.sh
+
+set -e
+
+# configs for 'chain'
+stage=10
+train_stage=-10
+get_egs_stage=-10
+speed_perturb=true
+dir=exp/chain/lstm_d  # Note: _sp will get added to this if $speed_perturb == true.
+decode_iter=
+decode_dir_affix=
+
+# LSTM options
+splice_indexes="-2,-1,0,1,2 0 0"
+lstm_delay=" -3 -3 -3 "
+label_delay=5
+num_lstm_layers=3
+cell_dim=1024
+hidden_dim=1024
+recurrent_projection_dim=256
+non_recurrent_projection_dim=256
+
+# training options
+leftmost_questions_truncate=-1
+chunk_width=150
+chunk_left_context=40
+chunk_right_context=0
+xent_regularize=0.025
+
+# decode options
+extra_left_context=
+extra_right_context=
+frames_per_chunk=
+
+remove_egs=false
+common_egs_dir=
+
+affix=
+# End configuration section.
+echo "$0 $@"  # Print the command line for logging
+
+. cmd.sh
+. ./path.sh
+. ./utils/parse_options.sh
+
+if ! cuda-compiled; then
+  cat <<EOF && exit 1
+This script is intended to be used with GPUs but you have not compiled Kaldi with CUDA
+If you want to use GPUs (and have them), go to src/, and configure and make on a machine
+where "nvcc" is installed.
+EOF
+fi
+
+# The iVector-extraction and feature-dumping parts are the same as the standard
+# nnet3 setup, and you can skip them by setting "--stage 8" if you have already
+# run those things.
+
+suffix=
+if [ "$speed_perturb" == "true" ]; then
+  suffix=_sp
+fi
+
+dir=$dir${affix:+_$affix}
+if [ $label_delay -gt 0 ]; then dir=${dir}_ld$label_delay; fi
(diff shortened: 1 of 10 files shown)


egs/swbd/s5c/local/chain/run_lstm_d.sh
```diff
@@ -0,0 +1,221 @@
...
+    --command="$decode_cmd" \
+    --feat.cmvn-opts="$cmvn_opts" \
+    --chain.l2-regularize=$l2_regularize \
+    --chain.xent-regularize=$xent_regularize \
+    --chain.apply-deriv-weights=false \
+    --chain.lm-opts="--num-extra-lm-states=2000" \
+    --trainer.rnn.num-chunk-per-minibatch=$num_chunk_per_minibatch \
+    --trainer.max-param-change=$max_param_change \
+    --trainer.num-epochs=$num_epochs \
+    --trainer.optimization.shrink-value 1.0 \
```
danpovey: do you really want shrink-value = 1.0?

vijayaditya:
I added a modified version of local/chain/run_tdnn_6h.sh which shows how to use python training script.
I ensured that we can use same python chain training script for both RNN and FF-DNNs.

Running tests.

vijayaditya:
log-probabilities are progressing as desired. I have also tested all parts of the training code. This PR  is ready for merge after review.

danpovey:
ok thanks.

egs/swbd/s5c/local/chain/run_lstm_d.sh
```diff
@@ -0,0 +1,217 @@
...
+# LSTM options
+splice_indexes="-2,-1,0,1,2 0 0"
+lstm_delay=" -3 -3 -3 "
+label_delay=5
+num_lstm_layers=3
+cell_dim=1024
+hidden_dim=1024
+recurrent_projection_dim=256
+non_recurrent_projection_dim=256
+include_log_softmax=true
```
danpovey: not sure why this is set to true- would be better IMO to hardcode it to false rather than having a variable.  chain models do not need the final log-softmax and actually it's incompatible with l2-regularize

vijayaditya: Oops thats a bug. Corrected it.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -0,0 +1,707 @@
...
+    # transform.
+    if (args.stage <= -6):
+        logger.info("Creating phone language-model")
+        chain_lib.CreatePhoneLm(args.dir, args.tree_dir, run_opts, lm_opts = args.lm_opts)
+
+    if (args.stage <= -5):
+        logger.info("Creating denominator FST")
+        chain_lib.CreateDenominatorFst(args.dir, args.tree_dir, run_opts)
+
+    num_leaves = chain_lib.GetNumberOfLeaves(args.tree_dir)
```
danpovey: you don't seem to use the num_leaves -> could remove it and the function that gets it.

vijayaditya: I removed the GetNum* call. But I use the GetNum* function in make_configs.py scripts to get the number of leaves from the tree dir, so I am keeping the function.

danpovey:
Vijay, it looks like your scripts require numpy.
I think it would make sense to look at the script that checks requirements, and modify it to ensure that numpy is installed, and tell people how to install it.

vijayaditya:
Numpy's use is isolated to 2 functions. I will make the necessary
modifications to ensure that the scripts that don't use it run without
issues. I will also add the necessary installation instructions.

--Vijay

danpovey:
OK, thanks.
Where is it used?
We have to decide whether to make it a global requirement or have those
scripts issue instructions themselves.
Dan

danpovey:
... if it's just used for reporting you could just disable that
functionality and print a warning if there is no numpy.

vijayaditya:
Numpy and Scipy are used in TDNN+Pooling recipe, where I use this to design
a low-pass filter with required bandwidth. It is not used anywhere else in
the current recipes.

I am making changes to isolate this to one function in
nnet3/tdnn/make_configs.py. An error will be raised if the user will try to
run this recipe without numpy/scipy.

--Vijay

vijayaditya:
I have submitted a PR #531 which isolates the requirement of numpy/scipy to
one function in tdnn/make_configs.py. This is just required for one TDNN
recipe.

--Vijay

PR #526, opened 2016-02-25
Add CNN option to current TDNN recipe.

tomkocse:


egs/swbd/s5c/local/nnet3/run_tdnn.sh
```diff
@@ -33,6 +34,32 @@ suffix=
...
+# to FBANK in the network instead of directly using FBANK features
+# from the storage. This script uses MFCC features as its input
+# and were converted to FBANK features with an inverse of DCT matrix
+# at the first layer of the network.
+
+  cnn_layer="filt_x_dim=3 filt_y_dim=8 filt_x_step=1 filt_y_step=1 num_filters=256 pool_x_size=1 pool_y_size=3 pool_z_size=1 pool_x_step=1 pool_y_step=3 pool_z_step=1"
+  cnn_opts+=(--cnn-layer "$cnn_layer")
+  cnn_bottleneck_dim=256
+  cnn_opts+=(--cnn-bottleneck-dim $cnn_bottleneck_dim)
+  cepstral_lifter="22.0"
```
vijayaditya: You should probably determine this from some feature config.

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -88,6 +88,17 @@ def AddLdaLayer(config_lines, name, input, lda_file):
     return {'descriptor':  '{0}_lda'.format(name),
             'dimension': input['dimension']}
 
+def AddFixedAffineLayer(config_lines, name, input, lda_file):
```
vijayaditya: rename the last argument to parameter file. Could you also modify AddLdaLayer to call AddFixedAffineLayer.

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -87,6 +102,24 @@ def PrintConfig(file_name, config_lines):
     f.write("\n".join(config_lines['component-nodes']))
     f.close()
 
+def ParseCnnString(cnn_string):
+    cnn_array = []
+    cnn_list={}
+    for cl in range(0, len(cnn_string)):
+      split1 = cnn_string[cl].split(" "); 
+      if len(split1) != 11:
+        sys.exit("invalid --cnn-layer input, must contain 11 arguments for one CNN layer: "
```
vijayaditya: Right now you are not allowing specification of default arguments. You could actually define an argparser which has default arguments and parse the cnn config line. To do that you would have to specify the cnn arguments using "--". e.g.

<code>--cnn-layer="--filter-x-step=1 --filter-y-step=1"</code>

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -124,6 +157,49 @@ def ParseSpliceString(splice_indexes, label_delay=None):
             'num_hidden_layers':len(splice_array)
             }
 
+def ComputeLifterCoeffs(lifter, dim):
```
vijayaditya: You could probably move these 3 functions to steps/nnet3/nnet3_train_lib.py

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -217,7 +312,37 @@ def ParseSpliceString(splice_indexes, label_delay=None):
...
+      cnn_input_dim = len(splice_indexes[0]) * args.feat_dim
+      prev_layer_output = {'descriptor':  splice_descriptor,
+                           'dimension': cnn_input_dim,
+                           '3d-dim': [len(splice_indexes[0]), args.feat_dim, 1],
+                           'vectorization': 'yzx'}
+
+      for cl in range(0, num_cnn_layers):
+        prev_layer_output = AddCnnLayer(config_lines, "L{0}".format(cl), prev_layer_output, cnn_array[cl])
+
+      prev_layer_output = nodes.AddAffineLayer(config_lines, "cnn-bottleneck", prev_layer_output, args.cnn_bottleneck_dim, "")
```
vijayaditya: what happens if cnn_bottleneck_dim is not specified ?

egs/wsj/s5/steps/nnet3/tdnn/train.sh
```diff
@@ -119,6 +123,15 @@ if [ $# != 4 ]; then
   echo "                                                   # Frame indices used for each splice layer."
   echo "                                                   # Format : layer<hidden_layer_index>/<frame_indices>....layer<hidden_layer>/<frame_indices> "
   echo "                                                   # (note: we splice processed, typically 40-dimensional frames"
+  echo "  --cnn-layer <string|cnn-layer0/filt_x_dim=3 filt_y_dim=8 filt_x_step=1 filt_y_step=1 num_filters=256 pool_x_size=1 pool_y_size=3 pool_z_size=1 pool_x_step=1 pool_y_step=3 pool_z_step=1> "
```
vijayaditya: does shell script allow for multiple specification of --cnn-layer parameters ?

danpovey: I'd rather not allow the shell script to accept multiple versions of the
same parameter.  This will be hard to implement and will break the normal
shell-level interface.
Dan

vijayaditya: Ideally we would call make_configs.py from the local/* scripts so this
would not be necessary in the future. I plan to take care of these things
when I write the steps/nnet3/train_dnn.py.

Tom basically wants to push his CNN recipe, with the current single layer
CNN configuration as he wants to focus on other things. The single layer
CNN recipe will not need to accept multiple versions of the same parameter,
so Tom may be you could simplify your life and not worry about providing
the multiple CNN layer configs to steps/nnet3/tdnn/train.sh. However ensure
that all other parts of the script can handle multiple CNN layers.

--Vijay

--VIjay

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
tomkocse: args.cnn_bottleneck_dim is default to be 256 unless specified

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
tomkocse: Actually i am using --cnn-layer to let user to choose whether to use CNN or not, if this option is not used, arg.cnn_layer is None then LDA is used instead of CNN

egs/swbd/s5c/local/nnet3/run_tdnn.sh
tomkocse: @Vijay, could you please specify how i should change this ?

egs/wsj/s5/steps/nnet3/tdnn/train.sh
vijayaditya: @tomkocse I added train_dnn.py in PR #528 . Could you base your script on that PR. This will simplify things to a great extent.

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
vijayaditya: I think the default should be no bottle_neck.

egs/swbd/s5c/local/nnet3/run_tdnn.sh
vijayaditya: @danpovey Is there a way to figure out if there has been cepstral liftering and if yes the cepstral liftering coefficient ? I thought that we could determine this from conf/mfcc_hires.conf. However as this option could also be specified from the command line relying on just the conf file is not safe.

Could you let us know what could be the best way to get the cepstral lifter coefficient  ?

danpovey: In egs/swbd/s5c/local/nnet3/run_tdnn.sh
> <https://github.com/kaldi-asr/kaldi/pull/526#discussion_r54309055>:
>
> > +# As the dimension of the CNN output is usually large, we place
> > +# a linear layer at the output of CNN for dimension reduction
> > +# The ivectors are processed through a fully connected affine layer,
> > +# then concatenated with the CNN bottleneck output and
> > +# passed to the deeper part of the network.
> > +# Due to the data compression issue, it is better to convert MFCC
> > +# to FBANK in the network instead of directly using FBANK features
> > +# from the storage. This script uses MFCC features as its input
> > +# and were converted to FBANK features with an inverse of DCT matrix
> > +# at the first layer of the network.
> > +
> > +  cnn_layer="filt_x_dim=3 filt_y_dim=8 filt_x_step=1 filt_y_step=1 num_filters=256 pool_x_size=1 pool_y_size=3 pool_z_size=1 pool_x_step=1 pool_y_step=3 pool_z_step=1"
> > +  cnn_opts+=(--cnn-layer "$cnn_layer")
> > +  cnn_bottleneck_dim=256
> > +  cnn_opts+=(--cnn-bottleneck-dim $cnn_bottleneck_dim)
> > +  cepstral_lifter="22.0"
>
> @danpovey <https://github.com/danpovey> Is there a way to figure out if
> there has been cepstral liftering and if yes the cepstral liftering
> coefficient ? I thought that we could determine this from
> conf/mfcc_hires.conf. However as this option could also be specified from
> the command line relying on just the conf file is not safe.
>

Get it from the config file-- I never pass in options to compute-mfcc-feats
except with the config file, I don't think that's even supported at the
script level.
Dan

PR #527, opened 2016-02-25
Documenting the behavior of compute-atwv correctly, as suggested by V…

jtrmal:
…assil Panayotov and Raghav Menon


jtrmal:
@danpovey @vdp 
Please feel free to review -- I will postpone the merge until at least one of you approves :)

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,21 @@ int main(int argc, char *argv[]) {
 
     const char *usage = "Computes the Actual Term-Weighted Value and prints it."
         "\n"
-        "Usage: compute-atwv [options]  ref-rspecifier hyp-rspecifier [alignment csv]\n"
-        " e.g.: compute-atwv ark:ref.1 ark:hyp.1 ali.csv\n"
+        "Usage: compute-atwv [options] nof-trials  ref-rspecifier hyp-rspecifier [alignment csv]\n"
```
danpovey: traditionally it would look like this:
<number-of-trials>  <ref-rspecifier> <hyp-rspecifier> [alignment csv]\

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,21 @@ int main(int argc, char *argv[]) {
...
+        " e.g.: compute-atwv 32485.4 ark:ref.1 ark:hyp.1 ali.csv\n"
         "\n"
-        "where the alignment format is compatible with the alignment produced\n"
-        "using the F4DE tool -- you are responsible for mapping the utterance\n"
-        "identifiers and the term string to the correct ones - use the script\n"
-        "utils/int2sym.pl and the utterance/keyword maps\n";
+        "NOTES: \n"
```
danpovey: I think you may be assuming too much knowledge with these notes.  You have not explained the format of ref-rspecifier and hyp-rspecifier, e.g. you could show an example line from each.

src/kwsbin/compute-atwv.cc
jtrmal: You're right, even though "thanks to"  github eating up the angle braces including the content, I had hard time to figure out what you mean. Fixing...

jtrmal:
OK, I tried to address the comments. I don't want to spend the time on documenting the F4DE as I wouldn't be able to be terse enough and I think the docs are long enough already.
There is a bunch of examples in the F4DE distribution, perhaps even docs.

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,31 @@ int main(int argc, char *argv[]) {
...
+        "  a) the number of trials is usually equal to the size of the searched\n"
+        "     collection in seconds\n"
+        "  b  the ref-rspecifier/hyp-rspecifier are the kaldi IO specifiers for both\n"
+        "     the reference and the hypotheses (found hits), consecutively.\n"
```
danpovey: consecutively->respectively
(I think)

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,31 @@ int main(int argc, char *argv[]) {
...
+        "     The format is the same for both of them. Each line is of \n"
+        "     the following format\n"
+        "\n"
+        "     <KW-ID> <utterance-id> <start-time-frames> <end-time-frames> <score>\n\n"
```
danpovey: maybe start-frame and end-frame would be clearer

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,31 @@ int main(int argc, char *argv[]) {
...
+        "     e.g.:\n\n"
+        "     KW106-189 348 459 560 0.8\n"
+        "\n"
+        "  b) the alignment-csv file format is equivalent with the alignment file\n"
```
danpovey: equivalent with -> equivalent to

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,31 @@ int main(int argc, char *argv[]) {
 
     const char *usage = "Computes the Actual Term-Weighted Value and prints it."
         "\n"
-        "Usage: compute-atwv [options]  ref-rspecifier hyp-rspecifier [alignment csv]\n"
-        " e.g.: compute-atwv ark:ref.1 ark:hyp.1 ali.csv\n"
+        "Usage: compute-atwv [options] <nof-trials> <ref-rspecifier> <hyp-rspecifier> [alignment-csv]\n"
```
danpovey: need angle brackets around <alignment-csv> ... and is it wspecifier? rspecifier?  wxfilename?

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,31 @@ int main(int argc, char *argv[]) {
...
+        "\n"
+        "     <KW-ID> <utterance-id> <start-time-frames> <end-time-frames> <score>\n\n"
+        "     e.g.:\n\n"
+        "     KW106-189 348 459 560 0.8\n"
+        "\n"
+        "  b) the alignment-csv file format is equivalent with the alignment file\n"
+        "     produced using the F4DE tool. However, we do not set some fields\n"
+        "     and the utterance identifiers are numeric. You can use the script\n"
+        "     utils/int2sym.pl and the utterance/keyword maps to convert\n" 
+        "     the numerical ids into text\n"
```
danpovey: would be nice to have an example of a line of alignment-csv, since you've documented the line format of the other files.

src/kwsbin/compute-atwv.cc
```diff
@@ -37,13 +37,31 @@ int main(int argc, char *argv[]) {
...
+        "  c) the scores are expected to be probabilities. Please note that\n"
+        "     the output from the kws-search is in -log(probability).\n"
+        "  d) compute-atwv does not perform any score normalization (it's just\n"
+        "     for scoring purposes). Without the score normalization/calibration\n"
```
danpovey: remove "the" in "the score normalizaton"

src/kwsbin/compute-atwv.cc
jtrmal: <nof-trials> is just a number
[alignment-csv] is a normal file

src/kwsbin/compute-atwv.cc
jtrmal: That is the issue -- a single example won't do because the format is quite complex and I don't think it's the right idea to explain it here. It would fit into the doxygen docs but there is no chance I can spend a lot of time on that right now :/
Also, it's not a mandatory parameter and can be ommitted (I'll mention that explicitly).

jtrmal:
I included the fixes and tried to clarify the csv output use.

PR #534, opened 2016-02-28
Add chain library to nnet3bin dependencies

akreal:
I also added `chain` to all tools' dependencies in main Makefile. Though it is not required (`nnet3` pools it anyway), looks like it should be there.


danpovey:
agreed.

PR #537, opened 2016-02-29
Fix <unk> bug in TEDLIUM Cantab Research language model.

galv:
The Cantab LM used <unk> to represent out of vocabulary words.
However, the scripts supposed that <UNK> was used to represent
out of vocabulary words. For some strange reason, the Cantab LM gives
<unk> a pronunciation "AH NG K". The end result was that the models
considered unknown words to have the pronunciation "AH NG K" instead
of an unknown pronunciation, and never actually saw any unknown words,
degrading performance.

This bug was previously fixed for the pruned 3-gram language model,
but not for the unpruned 4-gram model used for rescoring.
The fix is to use <unk> as the OOV word entirely throughout the scripts.


danpovey:
It has a surprisingly big effect.  Thanks for your fix!  Merging.

PR #539, opened 2016-03-02
Raw nnet3 training

vimalmanohar:
Scripts for raw nnet3 training supporting quadratic objective function and dense targets.
Let me know if this looks ok. 
@vijayaditya @danpovey

### DIFF
--- a/egs/swbd/s5c/local/nnet3/run_raw_tdnn.sh
+++ b/egs/swbd/s5c/local/nnet3/run_raw_tdnn.sh
@@ -43,6 +43,7 @@ fi
 
 if [ -z "$data_dir" ]; then
   echo "\$data_dir is required"
+  exit 1
 fi
 
 if [ $stage -le 9 ]; then
--- a/egs/wsj/s5/steps/nnet3/components.py
+++ b/egs/wsj/s5/steps/nnet3/components.py
@@ -208,7 +208,10 @@ def AddFinalSigmoidLayer(config_lines, input, output_dim,
         ng_affine_options = " param-stddev=0 bias-stddev=0 ",
         label_delay=None,
         name_affix = None,
-        objective_type = "linear"):
+        objective_type = "quadratic"):
+    # Useful when you need the final outputs to be probabilities
+    # between 0 and 1.
+    # Usually used with an objective-type such as "quadratic"
     components = config_lines['components']
     component_nodes = config_lines['component-nodes']
 
--- a/egs/wsj/s5/steps/nnet3/components.py
+++ b/egs/wsj/s5/steps/nnet3/components.py
@@ -152,7 +152,17 @@ def AddSoftmaxLayer(config_lines, name, input):
             'dimension': input['dimension']}
 
 
-def AddOutputLayer(config_lines, input, label_delay=None, suffix=None):
+def AddSigmoidLayer(config_lines, name, input, self_repair_scale = None):
+    components = config_lines['components']
+    component_nodes = config_lines['component-nodes']
+
+    self_repair_string = "self-repair-scale={0:.10f}".format(self_repair_scale) if self_repair_scale is not None else ''
+    components.append("component name={0}_sigmoid type=SigmoidComponent dim={1}".format(name, input['dimension'], self_repair_string))
+    component_nodes.append("component-node name={0}_sigmoid component={0}_sigmoid input={1}".format(name, input['descriptor']))
+    return {'descriptor':  '{0}_sigmoid'.format(name),
+            'dimension': input['dimension']}
+
+def AddOutputLayer(config_lines, input, label_delay = None, suffix=None, objective_type = "linear"):
     components = config_lines['components']
     component_nodes = config_lines['component-nodes']
     name = 'output'
@@ -160,9 +170,9 @@ def AddOutputLayer(config_lines, input, label_delay=None, suffix=None):
         name = '{0}-{1}'.format(name, suffix)
 
     if label_delay is None:
-        component_nodes.append('output-node name={0} input={1}'.format(name, input['descriptor']))
+        component_nodes.append('output-node name={0} input={1} objective={2}'.format(name, input['descriptor'], objective_type))
     else:
-        component_nodes.append('output-node name={0} input=Offset({1},{2})'.format(name, input['descriptor'], label_delay))
+        component_nodes.append('output-node name={0} input=Offset({1},{2}) objective={3}'.format(name, input['descriptor'], label_delay, objective_type))
 
 def AddFinalLayer(config_lines, input, output_dim,
         ng_affine_options = " param-stddev=0 bias-stddev=0 ",
@@ -170,8 +180,8 @@ def AddFinalLayer(config_lines, input, output_dim,
         use_presoftmax_prior_scale = False,
         prior_scale_file = None,
         include_log_softmax = True,
-        name_affix = None):
-
+        name_affix = None,
+        objective_type = "linear"):
     components = config_lines['components']
     component_nodes = config_lines['component-nodes']
 
@@ -186,13 +196,34 @@ def AddFinalLayer(config_lines, input, output_dim,
     if include_log_softmax:
         if use_presoftmax_prior_scale :
             components.append('component name={0}-fixed-scale type=FixedScaleComponent scales={1}'.format(final_node_prefix, prior_scale_file))
-            component_nodes.append('component-node name=Final-fixed-scale component={0}-fixed-scale input={1}'.format(finale_node_prefix,
+            component_nodes.append('component-node name={0}-fixed-scale component={0}-fixed-scale input={1}'.format(final_node_prefix,
                 prev_layer_output['descriptor']))
             prev_layer_output['descriptor'] = "{0}-fixed-scale".format(final_node_prefix)
         prev_layer_output = AddSoftmaxLayer(config_lines, final_node_prefix, prev_layer_output)
     # we use the same name_affix as a prefix in for affine/scale nodes but as a
     # suffix for output node
-    AddOutputLayer(config_lines, prev_layer_output, label_delay, suffix = name_affix)
+    AddOutputLayer(config_lines, prev_layer_output, label_delay, suffix = name_affix, objective_type = objective_type)
+
+def AddFinalSigmoidLayer(config_lines, input, output_dim,
+        ng_affine_options = " param-stddev=0 bias-stddev=0 ",
+        label_delay=None,
+        name_affix = None,
+        objective_type = "linear"):
+    components = config_lines['components']
+    component_nodes = config_lines['component-nodes']
+
+    if name_affix is not None:
+        final_node_prefix = 'Final-' + str(name_affix)
+    else:
+        final_node_prefix = 'Final'
+
+    prev_layer_output = AddAffineLayer(config_lines,
+            final_node_prefix , input, output_dim,
+            ng_affine_options)
+    prev_layer_output = AddSigmoidLayer(config_lines, final_node_prefix, prev_layer_output)
+    AddOutputLayer(config_lines, prev_layer_output, label_delay, suffix = name_affix, objective_type = objective_type)
+
+
 
 def AddLstmLayer(config_lines,
                  name, input, cell_dim,
--- a/egs/wsj/s5/steps/nnet3/get_egs_targets.sh
+++ b/egs/wsj/s5/steps/nnet3/get_egs_targets.sh
@@ -1,0 +1,411 @@
+#!/bin/bash
+
+# Copyright 2012-2015 Johns Hopkins University (Author: Daniel Povey).  
+#           2015-2016 Vimal Manohar
+# Apache 2.0.
+
+# This script is similar to steps/nnet3/get_egs.sh but used
+# when getting general targets (not from alignment directory) for raw nnet 
+#
+# This script, which will generally be called from other neural-net training
+# scripts, extracts the training examples used to train the neural net (and also
+# the validation examples used for diagnostics), and puts them in separate archives.
+#
+# This script dumps egs with several frames of labels, controlled by the
+# frames_per_eg config variable (default: 8).  This takes many times less disk
+# space because typically we have 4 to 7 frames of context on the left and
+# right, and this ends up getting shared.  This is at the expense of slightly
+# higher disk I/O while training.
+
+
+# Begin configuration section.
+cmd=run.pl
+feat_type=raw       # set it to 'lda' to use LDA features.
+target_type=sparse  # dense to have dense targets, 
+                    # sparse to have posteriors targets
+num_targets=        # required for target-type=sparse with raw nnet
+frames_per_eg=8   # number of frames of labels per example.  more->less disk space and
+                  # less time preparing egs, but more I/O during training.
+                  # note: the script may reduce this if reduce_frames_per_eg is true.
+left_context=4    # amount of left-context per eg (i.e. extra frames of input features
+                  # not present in the output supervision).
+right_context=4   # amount of right-context per eg.
+valid_left_context=   # amount of left_context for validation egs, typically used in
+                      # recurrent architectures to ensure matched condition with
+                      # training egs
+valid_right_context=  # amount of right_context for validation egs
+compress=true   # set this to false to disable compression (e.g. if you want to see whether
+                # results are affected).
+
+reduce_frames_per_eg=true  # If true, this script may reduce the frames_per_eg
+                           # if there is only one archive and even with the
+                           # reduced frames_per_eg, the number of
+                           # samples_per_iter that would result is less than or
+                           # equal to the user-specified value.
+num_utts_subset=300     # number of utterances in validation and training
+                        # subsets used for shrinkage and diagnostics.
+num_valid_frames_combine=0 # #valid frames for combination weights at the very end.
+num_train_frames_combine=10000 # # train frames for the above.
+num_frames_diagnostic=4000 # number of frames for "compute_prob" jobs
+samples_per_iter=400000 # this is the target number of egs in each archive of egs
+                        # (prior to merging egs).  We probably should have called
+                        # it egs_per_iter. This is just a guideline; it will pick
+                        # a number that divides the number of samples in the
+                        # entire data.
+
+transform_dir=     
+
+stage=0
+nj=6         # This should be set to the maximum number of jobs you are
+             # comfortable to run in parallel; you can increase it if your disk
+             # speed is greater and you have more machines.
+online_ivector_dir=  # can be used if we are including speaker information as iVectors.
+cmvn_opts=  # can be used for specifying CMVN options, if feature type is not lda (if lda,
+            # it doesn't make sense to use different options than were used as input to the
+            # LDA transform).  This is used to turn off CMVN in the online-nnet experiments.
+
+echo "$0 $@"  # Print the command line for logging
+
+if [ -f path.sh ]; then . ./path.sh; fi
+. parse_options.sh || exit 1;
+
+if [ $# != 3 ]; then
+  echo "Usage: $0 [opts] <data> <targets-scp> <egs-dir>"
+  echo " e.g.: $0 data/train data/train/snr_targets.scp exp/tri4_nnet/egs"
+  echo ""
+  echo "Main options (for others, see top of script file)"
+  echo "  --config <config-file>                           # config file containing options"
+  echo "  --nj <nj>                                        # The maximum number of jobs you want to run in"
+  echo "                                                   # parallel (increase this only if you have good disk and"
+  echo "                                                   # network speed).  default=6"
+  echo "  --cmd (utils/run.pl;utils/queue.pl <queue opts>) # how to run jobs."
+  echo "  --samples-per-iter <#samples;400000>             # Target number of egs per archive (option is badly named)"
+  echo "  --feat-type <lda|raw>                            # (raw is the default).  The feature type you want"
+  echo "                                                   # to use as input to the neural net."
+  echo "  --frames-per-eg <frames;8>                       # number of frames per eg on disk"
+  echo "  --left-context <width;4>                         # Number of frames on left side to append for feature input"
+  echo "  --right-context <width;4>                        # Number of frames on right side to append for feature input"
+  echo "  --num-frames-diagnostic <#frames;4000>           # Number of frames used in computing (train,valid) diagnostics"
+  echo "  --num-valid-frames-combine <#frames;10000>       # Number of frames used in getting combination weights at the"
+  echo "                                                   # very end."
+  echo "  --stage <stage|0>                                # Used to run a partially-completed training process from somewhere in"
+  echo "                                                   # the middle."
+
+  exit 1;
+fi
+
+data=$1
+targets_scp=$2
+dir=$3
+
+# Check some files.
+[ ! -z "$online_ivector_dir" ] && \
+  extra_files="$online_ivector_dir/ivector_online.scp $online_ivector_dir/ivector_period"
+
+for f in $data/feats.scp $targets_scp $extra_files; do
+  [ ! -f $f ] && echo "$0: no such file $f" && exit 1;
+done
+
+sdata=$data/split$nj
+utils/split_data.sh $data $nj
+
+mkdir -p $dir/log $dir/info
+
+
+# Get list of validation utterances.
+awk '{print $1}' $data/utt2spk | utils/shuffle_list.pl | head -$num_utts_subset | sort \
+    > $dir/valid_uttlist || exit 1;
+
+if [ -f $data/utt2uniq ]; then  # this matters if you use data augmentation.
+  echo "File $data/utt2uniq exists, so augmenting valid_uttlist to"
+  echo "include all perturbed versions of the same 'real' utterances."
+  mv $dir/valid_uttlist $dir/valid_uttlist.tmp
+  utils/utt2spk_to_spk2utt.pl $data/utt2uniq > $dir/uniq2utt
+  cat $dir/valid_uttlist.tmp | utils/apply_map.pl $data/utt2uniq | \
+    sort | uniq | utils/apply_map.pl $dir/uniq2utt | \
+    awk '{for(n=1;n<=NF;n++) print $n;}' | sort  > $dir/valid_uttlist
+  rm $dir/uniq2utt $dir/valid_uttlist.tmp
+fi
+
+awk '{print $1}' $data/utt2spk | utils/filter_scp.pl --exclude $dir/valid_uttlist | \
+   utils/shuffle_list.pl | head -$num_utts_subset | sort > $dir/train_subset_uttlist || exit 1;
+
+if [ ! -z "$transform_dir" ] && [ -f $transform_dir/trans.1 ] && [ $feat_type != "raw" ]; then
+  echo "$0: using transforms from $transform_dir"
+  if [ $stage -le 0 ]; then
+    $cmd $dir/log/copy_transforms.log \
+      copy-feats "ark:cat $transform_dir/trans.* |" "ark,scp:$dir/trans.ark,$dir/trans.scp"
+  fi
+fi
+if [ -f $transform_dir/raw_trans.1 ] && [ $feat_type == "raw" ]; then
+  echo "$0: using raw transforms from $transform_dir"
+  if [ $stage -le 0 ]; then
+    $cmd $dir/log/copy_transforms.log \
+      copy-feats "ark:cat $transform_dir/raw_trans.* |" "ark,scp:$dir/trans.ark,$dir/trans.scp"
+  fi
+fi
+
+
+
+## Set up features.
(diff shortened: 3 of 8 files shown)


egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -152,7 +152,16 @@ def AddSoftmaxLayer(config_lines, name, input):
             'dimension': input['dimension']}
 
 
-def AddOutputLayer(config_lines, input, label_delay=None, suffix=None):
+def AddSigmoidLayer(config_lines, name, input):
```
vijayaditya: Add self_repair_scale option.

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -160,9 +169,9 @@ def AddOutputLayer(config_lines, input, label_delay=None, suffix=None):
         name = '{0}-{1}'.format(name, suffix)
 
     if label_delay is None:
-        component_nodes.append('output-node name={0} input={1}'.format(name, input['descriptor']))
+        component_nodes.append('output-node name={0} input={1} objective={2}'.format(name, input['descriptor'], objective_type))
     else:
-        component_nodes.append('output-node name={0} input=Offset({1},{2})'.format(name, input['descriptor'], label_delay))
+        component_nodes.append('output-node name={0} input=Offset({1},{2}) objective={3}'.format(name, input['descriptor'], label_delay, objective_type))
```
vijayaditya: use named arguments in format()

egs/wsj/s5/steps/nnet3/get_egs.sh
```diff
@@ -265,23 +293,70 @@ valid_egs_opts="--left-context=$valid_left_context --right-context=$valid_right_
 
 echo $left_context > $dir/info/left_context
 echo $right_context > $dir/info/right_context
-num_pdfs=$(tree-info --print-args=false $alidir/tree | grep num-pdfs | awk '{print $2}')
-if [ $stage -le 3 ]; then
-  echo "$0: Getting validation and training subset examples."
-  rm $dir/.error 2>/dev/null
-  echo "$0: ... extracting validation and training-subset alignments."
 
+if ! $raw_nnet; then
```
vijayaditya: the steps/nnet3/get_egs.sh is really complicated with the new changes. I would recommend using separate scripts.

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -43,8 +43,16 @@ def GetArgs():
     # General neural network options
     parser.add_argument("--splice-indexes", type=str, required = True,
                         help="Splice indexes at each layer, e.g. '-3,-2,-1,0,1,2,3'")
+    parser.add_argument("--skip-lda", type=str, action=nnet3_train_lib.StrToBoolAction,
+                        help="skip lda layer", default=False, choices = ["false", "true"])
     parser.add_argument("--include-log-softmax", type=str, action=nnet3_train_lib.StrToBoolAction,
                         help="add the final softmax layer ", default=True, choices = ["false", "true"])
+    parser.add_argument("--add-final-sigmoid", type=str, action=nnet3_train_lib.StrToBoolAction,
```
vijayaditya: group add-final-sigmoid and objective type options. Checkout how to groups arguments in argparse.

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -121,6 +129,10 @@ def CheckArgs(args):
         args.nonlin_input_dim = args.pnorm_input_dim
         args.nonlin_output_dim = args.pnorm_output_dim
 
+    if add_final_sigmoid and include_log_softmax:
```
vijayaditya: there is something called mutually_exclusive_group in argparse. Use that rather than manually performing the check.

egs/wsj/s5/steps/nnet3/components.py
vimalmanohar: The rest of components.py does not have named arguments. Should I modify all of it?

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -339,17 +355,25 @@ def MakeConfigs(config_dir, splice_indexes_string,
         prev_layer_output = nodes.AddAffRelNormLayer(config_lines, "Tdnn_{0}".format(i),
                                                     prev_layer_output, nonlin_output_dim, norm_target_rms = 1.0 if i < num_hidden_layers -1 else final_layer_normalize_target)
         # a final layer is added after each new layer as we are generating configs for layer-wise discriminative training
-        nodes.AddFinalLayer(config_lines, prev_layer_output, num_targets,
-                           use_presoftmax_prior_scale = use_presoftmax_prior_scale,
-                           prior_scale_file = prior_scale_file,
-                           include_log_softmax = include_log_softmax)
+
+        if not add_final_sigmoid:
```
vijayaditya: it feels strange to check for the not of a boolean. Could you switch the check to 
if add_final_sigmoid:

Also add a brief local comment about why this is done (i.e., for quadratic objectives)

egs/wsj/s5/steps/nnet3/get_egs.sh
vimalmanohar: Agreed. What about steps/nnet3/tdnn/train.sh? Should I create a script like steps/nnet3/tdnn/train_raw.sh?

egs/wsj/s5/steps/nnet3/tdnn/train.sh
```diff
@@ -5,7 +5,7 @@
 # Copyright 2012-2015  Johns Hopkins University (Author: Daniel Povey).
 #           2013  Xiaohui Zhang
 #           2013  Guoguo Chen
-#           2014  Vimal Manohar
+#           2014-2016  Vimal Manohar
```
vijayaditya: it would be great if you could move this code to steps/nnet3/train_dnn.py. This will help us use the same script for a variety of architectures.

egs/wsj/s5/steps/nnet3/get_egs.sh
vijayaditya: I was suggesting splitting the get_egs.sh script. Not the train script. If
you have time to modify the train script I would recommend adding the
changes to steps/nnet3/train_dnn.py.

egs/wsj/s5/steps/nnet3/components.py
vijayaditya: If you have the time yes.

egs/wsj/s5/steps/nnet3/get_egs.sh
vijayaditya: I reason I would be fine with adding the changes to the same python script is due to the extensive use of functions in python script. It is usually easier to read. If you do feel the python script is also getting complicated due to excessive branching you could consider splitting the train.py script. Its up to you how you want to do this.

vimalmanohar: Sure. Will look into this.

vimalmanohar:
I made most of the fixes suggested by @vijayaditya and fixed some bugs. It should be good for commit.
I reverted to not using the mutually exclusive group because the options are not really mutually exclusive (they only cannot both be true). 
I have left the suggestions to add this to train_dnn.py and to use named arguments in components.py to a future PR because they still require more time.

egs/wsj/s5/steps/nnet3/get_egs_raw_nnet.sh
```diff
@@ -0,0 +1,411 @@
+#!/bin/bash
```
danpovey: I don't like the name of this program, because there is no inherent connection with a 'raw' neural net- the get_egs program doesn't need to see the neural net.  maybe call it get_egs_targets.sh, as in this case you supply more general targets, not alignments.

egs/wsj/s5/steps/nnet3/tdnn/train.sh
danpovey: Vimal, I don't want to check in this code as-is.
You have made a huge number of changes to nnet3/tdnn/train.sh.  This creates problems for testing.
Obviously we're not going to re-run the million TDNN training setups that we have here and there in Kaldi.
And even apart from that, your changes obscure what was already a very complicated script.
I would possibly be OK with checking in a separate script whose purpose is to train DNNs for this type of setup.
Or you may be able to do it with Vijay's train_dnn.py, although again, I don't want a situation where you introduce
a ton of if statements into a program-- if there are a lot of changes, I'd prefer you to create a separate python version of the training script; this will mean we don't have to worry about introducing bugs into Vijay's working setup.
I don't have strong opinions about whether you check in a shell script or a python script, but I think it should be a separate script.  Right now your changes add to much complexity to the script.

danpovey: BTW, you might find it easier to have the top-level train.sh script not be responsible for generating the configs or the egs-- if you assume that the top-level run.sh calls the appropriate scripts beforehand, that would make your life easier.  I'm following this pattern in the xvector setup.

vimalmanohar: I can add the changes to a new script called nnet3/tdnn/train_raw.sh.

danpovey: that would be OK with me, but try to simplify it by cutting out unnecessary
code.

vimalmanohar:
I added a new script that deals only with raw nnet3 model and reads targets directly instead of ali-dir.

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -186,13 +196,34 @@ def AddFinalLayer(config_lines, input, output_dim,
...
+            component_nodes.append('component-node name={0}-fixed-scale component={0}-fixed-scale input={1}'.format(final_node_prefix,
                 prev_layer_output['descriptor']))
             prev_layer_output['descriptor'] = "{0}-fixed-scale".format(final_node_prefix)
         prev_layer_output = AddSoftmaxLayer(config_lines, final_node_prefix, prev_layer_output)
     # we use the same name_affix as a prefix in for affine/scale nodes but as a
     # suffix for output node
-    AddOutputLayer(config_lines, prev_layer_output, label_delay, suffix = name_affix)
+    AddOutputLayer(config_lines, prev_layer_output, label_delay, suffix = name_affix, objective_type = objective_type)
+
+def AddFinalSigmoidLayer(config_lines, input, output_dim,
```
vijayaditya: You should add a comment on what a FinalSigmoidLayer is and where it is intended to be used. I think the default objective_type for this function should be quadratic, as it will used primarily in that context.

egs/wsj/s5/steps/nnet3/get_egs_targets.sh
```diff
@@ -0,0 +1,411 @@
...
+    ;;
+  default)
+    echo "$0: Unknown --target-type $target_type. Choices are dense and sparse"
+    exit 1
+esac
+
+if [ $stage -le 3 ]; then
+  echo "$0: Getting validation and training subset examples."
+  rm -f $dir/.error 2>/dev/null
+  echo "$0: ... extracting validation and training-subset alignments."
```
vijayaditya: update the message

egs/wsj/s5/steps/nnet3/get_egs_targets.sh
```diff
@@ -0,0 +1,411 @@
...
+      file=$dir/egs_orig.$x.$y.ark
+      [ -L $file ] && rm $(readlink -f $file)
+      rm $file
+    done
+  done
+  if [ $archives_multiple -gt 1 ]; then
+    # there are some extra soft links that we should delete.
+    for f in $dir/egs.*.*.ark; do rm $f; done
+  fi
+  echo "$0: removing temporary alignments and transforms"
```
vijayaditya: update the message

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -121,6 +131,9 @@ def CheckArgs(args):
         args.nonlin_input_dim = args.pnorm_input_dim
         args.nonlin_output_dim = args.pnorm_output_dim
 
+    if args.add_final_sigmoid and args.include_log_softmax:
+        sys.exit("--include-log-softmax and --add-final-sigmoid cannot both be true.")
```
vijayaditya: Don't use <code>sys.exit</code> raise Exceptions.

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
...
+egs_dir=
+add_lda=false
+max_lda_jobs=10  # use no more than 10 jobs for the LDA accumulation.
+lda_opts=
+egs_opts=
+transform_dir=     # If supplied, this dir used instead of alidir to find transforms.
+cmvn_opts=  # will be passed to get_lda.sh and get_egs.sh, if supplied.
+            # only relevant for "raw" features, not lda.
+feat_type=raw  # or set to 'lda' to use LDA features.
+align_cmd=              # The cmd that is passed to steps/nnet2/align.sh
```
vijayaditya: are you going to use alignments ? If not this realign code should not be here. trim it.

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
...
+# End configuration section.
+
+trap 'for pid in $(jobs -pr); do kill -KILL $pid; done' INT QUIT TERM
+
+echo "$0 $@"  # Print the command line for logging
+
+if [ -f path.sh ]; then . ./path.sh; fi
+. parse_options.sh || exit 1;
+
+if [ $# != 3 ]; then
```
vijayaditya: update usage message to reflect your current options.

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
...
+else
+  ivector_dim=$(feat-to-dim scp:$online_ivector_dir/ivector_online.scp -) || exit 1;
+fi
+
+if [ $stage -le -5 ]; then
+  echo "$0: creating neural net configs";
+
+  config_opts=()
+
+  if [ ! -z "$relu_dim" ]; then
```
vijayaditya: As @danpovey pointed out it would make your life a lot easier if you remove the config generation stage from this script. I place it in the top level run.sh script like [this](https://github.com/kaldi-asr/kaldi/blob/master/egs/swbd/s5c/local/nnet3/run_tdnn.sh#L52).

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
...
+sdata=$data/split$nj
+utils/split_data.sh $data $nj
+
+mkdir -p $dir/log
+echo $nj > $dir/num_jobs
+
+
+# First work out the feature and iVector dimension, needed for tdnn config creation.
+case $feat_type in
+  raw) feat_dim=$(feat-to-dim --print-args=false scp:$data/feats.scp -) || \
```
vijayaditya: while you are at it you could drop support for lda features from this script to make it cleaner. We are not using this option in nnet3 scripts as we use a preconditioning matrix in the network.

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
...
+  # Appendix C.6 of http://arxiv.org/pdf/1410.7455v6.pdf; it's a scaled variant
+  # of an LDA transform but without dimensionality reduction.
+  $cmd $dir/log/get_transform.log \
+     nnet-get-feature-transform $lda_opts $dir/lda.mat $dir/lda_stats || exit 1;
+
+  ln -sf ../lda.mat $dir/configs/lda.mat
+fi
+
+
+if $include_log_softmax && ! $dense_targets && [ $stage -le -2 ]; then
```
vijayaditya: Is your code going into this branch given that you would not be using it with sparse targets any more ?

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
...
+  # different subsets will lead to different probs.
+  $cmd $dir/log/compute_prob_valid.final.log \
+    nnet3-compute-prob --compute-accuracy=$compute_accuracy $dir/final.raw \
+    "ark:nnet3-merge-egs ark:$cur_egs_dir/valid_diagnostic.egs ark:- |" &
+  $cmd $dir/log/compute_prob_train.final.log \
+    nnet3-compute-prob --compute-accuracy=$compute_accuracy $dir/final.raw \
+    "ark:nnet3-merge-egs ark:$cur_egs_dir/train_diagnostic.egs ark:- |" &
+fi
+
+if $include_log_softmax && [ $stage -le $[$num_iters+1] ]; then
```
vijayaditya: Is this branch still necessary ?

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
```diff
@@ -0,0 +1,636 @@
+#!/bin/bash
```
vijayaditya: I think the script should be named as <code>train_rawnnet_densetargets.sh</code>. Since there is another effort on raw features things can get pretty confusing in the future.

vijayaditya:
remember to edit the commit message to be more descriptive.

lifeiteng:
I don't think ` egs/wsj/s5/steps/nnet3/get_egs_targets.sh` have done all the job, if you use Sigmoid at last layer, you should also store the label info(dense target may be 1 or 0); 
when target value is 0, you will don't know to update which label's parameters(if target is for label 10, so the 0-9 Row parameters of last AffineTransform should not be updated), so you should also store the label info, maybe implement anothor method
```
  NnetIo(const std::string &name,
         int32 t_begin, const MatrixBase<BaseFloat> &feats, const Posterior &labels);
```

lifeiteng:
Also some changes in 
```
void NnetTrainer::ProcessOutputs(const NnetExample &eg,
                                 NnetComputer *computer) {
  std::vector<NnetIo>::const_iterator iter = eg.io.begin(),
      end = eg.io.end();
  for (; iter != end; ++iter) {
    const NnetIo &io = *iter;
    int32 node_index = nnet_->GetNodeIndex(io.name);
    KALDI_ASSERT(node_index >= 0);
    if (nnet_->IsOutputNode(node_index)) {
      ObjectiveType obj_type = nnet_->GetNode(node_index).u.objective_type;
      BaseFloat tot_weight, tot_objf;
      bool supply_deriv = true;
      // TODO use labels in NnetIo to Rewrite cross-entropy objective's derive (for sigmoid) in ComputeObjectiveFunction() 
      ComputeObjectiveFunction(io.features, obj_type, io.name,
                               supply_deriv, computer,
                               &tot_weight, &tot_objf);
      // TODO the accuracy's computation also should be changed.
      objf_info_[io.name].UpdateStats(io.name, config_.print_interval,
                                      num_minibatches_processed_++,
                                      tot_weight, tot_objf);
    }
  }
}
```

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
vimalmanohar: I think the raw features must be called raw waveform features or something. "raw" here is for raw nnet without transition model (HMM) and that semantic is already being used in all the nnet3 binaries with --raw option. This script also supports sparse targets (e.g. SAD, speech / silence / music / noise classification or any other classification task that does not involve a HMM). The include_log_softmax is there for the same reason.

egs/wsj/s5/steps/nnet3/tdnn/train_raw.sh
vimalmanohar: I see. It is possible to move it outside and comment out the config generation to show an example.

vimalmanohar:
@lifeiteng I am not sure what scenario you have in mind. This is the use case I have in mind:
The sigmoid is to be used as last layer if you want the output in each dimension to be like probabilities between zero and one. It is left to the user to decide the appropriate targets, which should be a vector with each element between 0 and 1. The objective function itself is independent of what is used in the network. Usually you can use quadratic for this but linear (which will be a cosine distance) might also ok.

vimalmanohar:
I have made changes based on the suggestions. I created an example run script in swbd/s5c

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -357,11 +385,23 @@ def MakeConfigs(config_dir, splice_indexes_string,
...
+    include_log_softmax_str = 'false'
+    if include_log_softmax:
+        include_log_softmax_str = 'true'
+
     # write the files used by other scripts like steps/nnet3/get_egs.sh
     f = open(config_dir + "/vars", "w")
     print('model_left_context=' + str(left_context), file=f)
     print('model_right_context=' + str(right_context), file=f)
     print('num_hidden_layers=' + str(num_hidden_layers), file=f)
+    print('num_targets=' + str(num_targets), file=f)
```
vimalmanohar: @vijayaditya 
I had to add extra options the vars file as they are required by the train script. It seems that the python library to parse the config file currently only supports integer options. Does it make sense to modify it to support generic options (string etc.)?

vijayaditya: which python library are you talking about ?

vimalmanohar: The function ParseModelConfigVarsFile in [this](https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py). I want to modify this to return a python dict, which maps variables to strings by default, but integers for specific variables like model_left_context. This might require modifying all the python nnet3 train scripts.

vijayaditya: I would recommend writing a new function tailored to your needs.

egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh
```diff
@@ -0,0 +1,626 @@
...
+max_param_change=2.0  # max param change per minibatch
+samples_per_iter=400000 # each iteration of training, see this many samples
+                        # per job.  This option is passed to get_egs.sh
+num_jobs_initial=1  # Number of neural net jobs to run in parallel at the start of training
+num_jobs_final=8   # Number of neural net jobs to run in parallel at the end of training
+prior_subset_size=20000 # 20k samples per job, for computing priors.
+num_jobs_compute_prior=10 # these are single-threaded, run on CPU.
+get_egs_stage=0    # can be used for rerunning after partial
+online_ivector_dir=
+presoftmax_prior_scale_power=-0.25
```
danpovey: I think you can delete anything related to presoftmax_prior_scale.

egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh
```diff
@@ -0,0 +1,626 @@
...
+
+if [ -z "$online_ivector_dir" ]; then
+  ivector_dim=0
+else
+  ivector_dim=$(feat-to-dim scp:$online_ivector_dir/ivector_online.scp -) || exit 1;
+fi
+
+
+###
+# The config creation is moved to upper-level run script.
```
danpovey: I don't see the need for this extended comment here.

egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh
```diff
@@ -0,0 +1,626 @@
...
+  num_archives_expanded=$[$num_archives*$frames_per_eg]
+fi
+
+[ $num_jobs_initial -gt $num_jobs_final ] && \
+  echo "$0: --initial-num-jobs cannot exceed --final-num-jobs" && exit 1;
+
+[ $num_jobs_final -gt $num_archives_expanded ] && \
+  echo "$0: --final-num-jobs cannot exceed #archives $num_archives_expanded." && exit 1;
+
+
```
danpovey: you should probably check that add_lda is defined at this point and die if not.  otherwise we'll get confusing errors if something went wrong in generating the configs, or a different script was used.

egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh
```diff
@@ -0,0 +1,626 @@
...
+     nnet-get-feature-transform $lda_opts $dir/lda.mat $dir/lda_stats || exit 1;
+
+  ln -sf ../lda.mat $dir/configs/lda.mat
+fi
+
+
+if $include_log_softmax && ! $dense_targets && [ $stage -le -2 ]; then
+  echo "$0: preparing initial vector for FixedScaleComponent before softmax"
+  echo "  ... using priors^$presoftmax_prior_scale_power and rescaling to average 1"
+
```
danpovey: we have stopped using this in recent scripts.  If you're not using it, please delete the code.

egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh
```diff
@@ -0,0 +1,626 @@
...
+  $cmd $dir/log/compute_prob_valid.final.log \
+    nnet3-compute-prob --compute-accuracy=$compute_accuracy $dir/final.raw \
+    "ark:nnet3-merge-egs ark:$egs_dir/valid_diagnostic.egs ark:- |" &
+  $cmd $dir/log/compute_prob_train.final.log \
+    nnet3-compute-prob --compute-accuracy=$compute_accuracy $dir/final.raw \
+    "ark:nnet3-merge-egs ark:$egs_dir/train_diagnostic.egs ark:- |" &
+fi
+
+if $include_log_softmax && [ $stage -le $[$num_iters+1] ]; then
+  echo "Getting average posterior for purposes of adjusting the priors."
```
danpovey: this whole block is pointless, I think, since we don't need the priors.  Unless you need the average posterior for some other purpose.

vimalmanohar: Yes, I need priors to get likelihoods for use in a HMM decoder for segmentation. I can add a description in the comments.

vimalmanohar:
I have made the suggested changes

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -357,11 +385,23 @@ def MakeConfigs(config_dir, splice_indexes_string,
     left_context += int(parsed_splice_output['left_context'])
     right_context += int(parsed_splice_output['right_context'])
 
+    add_lda_str = 'false'
```
danpovey: I'd do add_lda_str = ('true' if add_lda else 'false')

egs/swbd/s5c/local/nnet3/run_raw_tdnn.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
danpovey: typically these local/run*.sh script are 'complete', i.e. they contain all the required arguments to actually do something.

vimalmanohar: I understand, but this is meant to be an example for using the configs with the train script. I do not have a complete example yet as the SNR predictor or SAD requires more codes and scripts to be added.

danpovey: maybe explain that it's incomplete and give some idea how you'd use it.

vimalmanohar:
I move the example to aspire as an autoencoder mapping from reverberated to clean.

danpovey:
OK, merging.

lifeiteng:
@vimalmanohar 
From [Softmax Regression vs. k Binary Classifiers] (http://ufldl.stanford.edu/wiki/index.php/Softmax_Regression)
"Now, consider a computer vision example, where you're trying to classify images into three different classes, suppose your classes are indoor_scene, black_and_white_image, and image_has_people. it would be more appropriate to build three separate logistic regression classifiers."

We can make "three separate logistic regression classifiers" deeper, let them share hidden/recurrent layers.

If you have a dataset, each picture is labeled like
* "image has people, not indoor_scene" so you don't kwon wether it is black_and_white_image.

So you can make one positive sample for  image_has_people, one negative sample for indoor_scene, but you can't let it be positive or negative sample for black_and_white_image which means you should't update the separate logistic regression classifier's  parameters of class  black_and_white_image.

vimalmanohar:
@lifeiteng 
This can be treated as a standard multi-task problem. I think for this it is most sensible to have two / three separate output nodes in nnet3 configuration and skip specific output nodes from the egs when corresponding label is not defined.

PR #540, opened 2016-03-02
Chain signal perturb

pegahgh:


danpovey:
Thanks!
I don't want to review this until you have the whole system working, and
can (for example) test it in some kind of neural net training.  This way,
if it doesn't work, I can save time.  However, I hope it will work.
Dan

pegahgh:
Sure!
I will try to make whole setup working!

danpovey:
I think this isn't moving forward right now, so closing it.

PR #545, opened 2016-03-03
Use curl instead of wget

ceefour:
Assuming the user installs Git for Windows, they will already have `curl` and `unzip` but still no `wget`, so make it easy to follow the instructions.

Some enhancements.


jtrmal:
THanks, merging.

PR #548, opened 2016-03-04
added simple TDNN comparison run_tdnn_6v.sh for jesus models run_tdnn…

vijayaditya:
…_6h.sp

### DIFF
--- a/egs/swbd/s5c/local/chain/run_tdnn_6v.sh
+++ b/egs/swbd/s5c/local/chain/run_tdnn_6v.sh
@@ -1,0 +1,207 @@
+#!/bin/bash
+
+# this is same as v2 script but with xent-regularization
+# it has a different splicing configuration
+set -e
+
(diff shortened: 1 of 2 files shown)


vijayaditya:
this model has 9.8M parameters in chain part of network i.e., not counting parameters in xent only part of the nnet.

danpovey:
Thanks- can you show me where you ran this?

vijayaditya:
It is running in /export/a11/vpeddinti/tdnn_pool/egs/swbd/s5c/exp/chain/tdnn_v7_sp

Please note that I restarted the training from scratch due to some bugs in the config, and the restarted training is at iteration 33. So only logs till that iteration are valid.

danpovey:
OK.
BTW, I see this: 'denominator_scale': 1.0,
I think this config is no longer needed, it was for CTC.

PR #550, opened 2016-03-04
A few very minor fixes to Karel's confidence calibration scripts.

vdp:
Mostly, we make sure that the calls to the Python scripts are changed to
"python \<script.py>", because the scripts start with "#!/bin/env"
(instead "#!/usr/bin/env") which doesn't work in e.g. Ubuntu.
Also a couple of typos corrected in the Python scripts themselves.

@vesis84 : Karel I hope you don't mind the changes. I believe
they should work in your environment too.

Just FYI, the mean NCE value for the LibriSpeech's "dev-other"
subset(I've split it in two, and used the first half for training and the
second one for testing) is 0.302.
Looking at the Karel's numbers, that @danpovey forwarded to the
mailing list it looks like it's about the same as AMI and somewhat
lower than the value for TED-LIUM. That sounds about right to me, 
given that the "other" subsets tend to contain heavily accented speech.

### DIFF
--- a/egs/wsj/s5/steps/conf/append_eval_to_ctm.py
+++ b/egs/wsj/s5/steps/conf/append_eval_to_ctm.py
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/usr/bin/env python
 
 # Copyright 2015  Brno University of Technology (author: Karel Vesely)
 # Apache 2.0
--- a/egs/wsj/s5/steps/conf/append_prf_to_ctm.py
+++ b/egs/wsj/s5/steps/conf/append_prf_to_ctm.py
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/usr/bin/env python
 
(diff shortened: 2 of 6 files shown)


danpovey:
Vassil, I think the right fix for the shebang is to change the shebang to
 /usr/bin/env and call the script directly.  We are trying to enforce a
uniform shebang style for all python scripts, and I prefer calling them
without 'python'.

vdp:
Agreed, that this is indeed the cleaner way to fix this. 
I just wasn't sure how many systems might require the non-standard shebang line. Hopefully if someone(notably Karel) really needs to use this non-standard location for "env", it won't be hard for them to update their systems(e.g. by creating symlinks)..

Dan, please find the updated patch.

danpovey:
Thanks!

vesis84:
Hi Vassil, Dan, thanks for the fixes!
Yes the sha-bang, '#!/usr/bin/env python' works well for our system too.
Best regards!
Karel.

PR #563, opened 2016-03-07
Added lattice-determinize-non-compact

vimalmanohar:


danpovey:
Thanks, I'll look at this to-morrow

danpovey:
These are more code changes than I had anticipated.
Your semi-supervised training stuff, I'd rather not deal with right now-- I think we should compare with other approaches before committing it.
Your new discriminative-training wrapper-- it's not clear to me why this is needed?  What is the reason you can't use the old wrapping function, or maybe modify it slightly?

vimalmanohar:
This code supports using lattices as numerator instead of using alignment. Probably applicable to semi-supervised training.

vimalmanohar:
Most of the changes here are for semi-supervised training.

danpovey:
I thought it was the nnet3 discriminative training that we were checking in
(??)

vimalmanohar:
The codes were modified for work for both. The semi-supervised stuff can be added later. I am adding the other commits first.

danpovey:
It would have been better if you had kept these things separate.  I'm not
even sure that I want to check in the semi-supervised stuff.  Please start
work on a clean nnet3-discriminative-training setup that doesn't include a
bunch of unrelated changes!

vimalmanohar:
I have made the changes. I ran one iteration and the result matches the previous run. I will run a full experiment to check again.

danpovey:
ok thanks.
i won't check the PR till to-morrow.

src/latbin/lattice-determinize-non-compact.cc
```diff
@@ -0,0 +1,328 @@
...
+    typedef kaldi::int64 int64;
+    using fst::SymbolTable;
+    using fst::VectorFst;
+    using fst::StdArc;
+
+    const char *usage =
+        "lattice-determinize lattices (and apply a pruning beam)\n"
+        " (see http://kaldi.sourceforge.net/lattices.html for more explanation)\n"
+        "This version of the program retains the original "
+        "acoustic scores of arcs in the lattice. "
```
danpovey: mention that it writes the output in the non-compact format.

src/nnet3/discriminative-supervision.cc
```diff
@@ -0,0 +1,528 @@
...
+    } else {
+      output_supervision->resize(output_supervision->size() + 1);
+      output_supervision->back() = src;
+    }
+
+    output_supervision->back().Check();
+  }
+}
+
+void AppendLattice(Lattice *lat, const Lattice &src_lat) {
```
danpovey: there is a lot of lava-flow here.
state times are not needed, and your comment is not needed.
But why use this function at all?  Just call Concat and then TopSort.

vimalmanohar:
Removed that function

src/nnet3/discriminative-supervision.cc
```diff
@@ -0,0 +1,450 @@
...
+  supervision->weight = weight;
+  supervision->num_sequences = 1;
+  supervision->frames_per_sequence = num_ali.size();
+  supervision->num_ali = num_ali;
+  supervision->den_lat = den_lat;
+  TopSort(&(supervision->den_lat));
+
+  supervision->Check();
+
+  return true;
```
danpovey: this function always returns true, so I don't see a reason to have it as a separate function- why don't you make it a constructor?

src/nnet3/discriminative-supervision.cc
```diff
@@ -0,0 +1,450 @@
...
+// It helps us to more completely minimize the lattice.  Note: we
+// can't do this if the criterion is MPFE, because in that case the
+// objective function will be affected by the phone-identities being
+// different even if the pdf-ids are the same.
+void DiscriminativeSupervisionSplitter::CollapseTransitionIds(
+    const std::vector<int32> &state_times, Lattice *lat) const {
+  typedef Lattice::StateId StateId;
+  typedef Lattice::Arc Arc;
+
+  int32 num_frames = state_times.back();   // TODO: Check if this is always true
```
danpovey: it's always true since it's a non-compact lattice, and you've ensured that it's topologically sorted.

src/nnet3/discriminative-supervision.cc
vimalmanohar: I think I intended to return false if TopSort fails. I can check that and return false if it fails.

danpovey: It would never fail- would indicate a serious bug in the lattice.
I think it should be a class member anyway.  Initialize() function would
make more sense.

src/nnet3/discriminative-supervision.cc
```diff
@@ -0,0 +1,450 @@
...
+
+void DiscriminativeSupervisionSplitter::LatticeInfo::Check() const {
+  // Check if all the vectors are of size num_states
+  KALDI_ASSERT(state_times.size() == alpha_p.size() &&
+               state_times.size() == beta_p.size());
+
+  // Check that the states are ordered in increasing order of state_times.
+  // This must be true since the states are in breadth-first search order.
+  int32 t = 0;
+  for (std::vector<int32>::const_iterator it = state_times.begin();
```
danpovey: KALDI_ASSERT(IsSorted(state_times));

src/nnet3/discriminative-supervision.h
```diff
@@ -0,0 +1,247 @@
...
+  DiscriminativeSupervisionOptions supervision_config;
+  
+  SplitDiscriminativeSupervisionOptions() :
+    remove_output_symbols(false), collapse_transition_ids(false), 
+    remove_epsilons(false), determinize(false),
+    minimize(false) { }
+
+  void Register(OptionsItf *opts) {
+    opts->Register("collapse-transition-ids", &collapse_transition_ids,
+                   "Collapse transition ids");
```
danpovey: what does this mean???

src/nnet3/discriminative-supervision.cc
vimalmanohar: I can make it Initialize. This was how it is in nnet2. It was returning false when frames_per_sequence is 0 or when there is a mismatch between lattice and alignment num_frames.

src/nnet3/discriminative-supervision.h
```diff
@@ -0,0 +1,247 @@
...
+ 
+  DiscriminativeSupervisionSplitter(
+      const SplitDiscriminativeSupervisionOptions &config,
+      const TransitionModel &tmodel,
+      const DiscriminativeSupervision &supervision);
+
+  // A structure used to store the forward and backward scores 
+  // and state times of a lattice
+  struct LatticeInfo {
+    std::vector<double> alpha_p;
```
danpovey: google style guide forbids abbreviations.  I suggest to get rid of the _p suffixes.

danpovey: also document that they are log(alpha) and log(beta).

src/nnet3/discriminative-supervision.cc
danpovey: OK- you can make it Initialize and have it return bool.
but do check that the times match and return false if not.

src/nnet3/discriminative-supervision.h
```diff
@@ -0,0 +1,247 @@
...
+                             Lattice *lat) const;
+
+};
+
+/// This function appends a list of supervision objects to create what will
+/// usually be a single such object, but if the weights and num-frames are not
+/// all the same it will only append Supervision objects where successive ones
+/// have the same weight and num-frames, and if 'compactify' is true.  The
+/// normal use-case for this is when you are combining neural-net examples for
+/// training; appending them like this helps to simplify the decoding process.
```
danpovey: s/decoding/training/.  also in wherever you got this comment from originally.

src/nnet3/discriminative-supervision.h
vimalmanohar: This was because we had considered experimenting if we needed to also add initial and final costs of the objective, which would have been alpha_r and beta_r.

src/nnet3/discriminative-supervision.h
vimalmanohar: I'll add description.

src/nnet3/discriminative-supervision.h
danpovey: ok, well change it-- in any case that would have had to be documented.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+  // same number of rows as supervision.num_frames * supervision.num_sequences,
+  // and the same number of columns as tmodel.NumPdfs(); but the
+  // ordering of the rows of 'nnet_output' is not the same as the ordering of
+  // frames in paths in the 'supervision' object (which has all frames of the
+  // 1st sequence first, then the 2nd sequence, and so on).  Instead, the
+  // frames in 'nnet_output' are ordered as: first the first frame of each
+  // sequence, then the second frame of each sequence, and so on. 
+  // This is done to be similar to the setup in 'chain' training 
+  // even though this does not offer any computational advantages here
+  // as in the 'chain' case.
```
danpovey: You could also mention that this is the default ordering used internally in the nnet3 code, so it may have a slight efficiency advantage by avoiding reordering at the output.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+  DiscriminativeComputation(const DiscriminativeTrainingOptions &opts,
+      const TransitionModel &tmodel,
+      const CuVectorBase<BaseFloat> &log_priors,
+      const DiscriminativeSupervision &supervision,
+      const CuMatrixBase<BaseFloat> &nnet_output,
+      DiscriminativeTrainingStats *stats,
+      BaseFloat *l2_term,
+      CuMatrixBase<BaseFloat> *nnet_output_deriv,
+      CuMatrixBase<BaseFloat> *xent_output_deriv);
```
danpovey: I don't understand why you have a separate derivative matrix called 'xent_output_deriv'.
The reason this existed in the chain setup was that the nnet has two distinct outputs.
In discriminative training, there will presumably just be one output (?)  Or are you also trying to implement the same kind of xent smoothing as we support in the chain models, with a different output?

vimalmanohar: I need to conduct experiments on this. This smoothing is standard with sequence training, called F-smoothing or something. Microsoft paper says this helps.

danpovey: >
> > +  // frames in paths in the 'supervision' object (which has all frames of the
> > +  // 1st sequence first, then the 2nd sequence, and so on).  Instead, the
> > +  // frames in 'nnet_output' are ordered as: first the first frame of each
> > +  // sequence, then the second frame of each sequence, and so on.
> > +  // This is done to be similar to the setup in 'chain' training
> > +  // even though this does not offer any computational advantages here
> > +  // as in the 'chain' case.
> > +  DiscriminativeComputation(const DiscriminativeTrainingOptions &opts,
> > +      const TransitionModel &tmodel,
> > +      const CuVectorBase<BaseFloat> &log_priors,
> > +      const DiscriminativeSupervision &supervision,
> > +      const CuMatrixBase<BaseFloat> &nnet_output,
> > +      DiscriminativeTrainingStats *stats,
> > +      BaseFloat *l2_term,
> > +      CuMatrixBase<BaseFloat> *nnet_output_deriv,
> > +      CuMatrixBase<BaseFloat> *xent_output_deriv);
>
> I need to conduct experiments on this. This smoothing is standard with
> sequence training, called F-smoothing or something. Microsoft paper says
> this helps.
>
In my experience it doesn't help.
But even if you are going to do that, there is no need to have a separate
argument here, since there will be just one output of the network, and you
can just add to the derivative.
(in MMI, you can do it just by scaling the numerator as you add it).

vimalmanohar: This would be same as in the chain training. The chain objective gets replaced by the sequence objective. The xent remains the same as before. We would need a separate output that goes through a softmax.

danpovey: This would only make sense if the model was a chain model and it already
had the two outputs- otherwise, why not use the single output of the nnet,
which already has a softmax, for both purposes?  So you should probably
document that xent_output_deriv is only intended for use with chain models.

vimalmanohar: I remember decoding with the softmax part of the output, but the results were 10% absolute worse than chain part. I will try again to make sure I did correctly.

danpovey: I'm not sure why you said what you just said about "I remember decoding
with the softmax part of the output, but the results were 10% absolute
worse than chain part. I will try again to make sure I did correctly."  I
think you may have misinterpreted something I said.  This is not relevant.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+  
+  if (!SplitStringToIntegers(opts_.silence_phones_str, ":", false,
+                             &silence_phones_)) {
+    KALDI_ERR << "Bad value for --silence-phones option: "
+              << opts_.silence_phones_str;
+  }
+}
+
+void DiscriminativeComputation::Compute() {
+  if (opts_.criterion == "mmi" && opts_.boost != 0.0) {
```
danpovey: there is a 'soft' 40-line limit on functions.
you have blown *way* past the limit with this function.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+    double objf = tot_objf / tot_t_weighted;
+    KALDI_LOG << "Average modulus of SMBR gradients is " << avg_gradients 
+              << " per frame, over "
+              << tot_t_weighted << " frames";
+    KALDI_LOG << "SMBR objective function is " << objf
+              << " per frame, over " << tot_t_weighted << " frames.";
+  } 
+  
+  if (AccumulateGradients()) {
+    {
```
danpovey: double braces-- remove.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+      temp.Scale(1.0/tot_t_weighted);
+      if (print_avg_gradients) {
+        KALDI_LOG << "Vector of average gradients wrt output activations is: \n" << temp;
+      } else {
+        KALDI_VLOG(4) << "Vector of average gradients wrt output activations is: \n" << temp;
+      }
+    }
+  }
+  if (AccumulateOutput()) {
+    {
```
danpovey: same here.

src/nnet3/discriminative-training.h
```diff
@@ -0,0 +1,250 @@
...
+                                   debug_training_advanced(false) { }
+
+  void Register(OptionsItf *opts) {
+    opts->Register("criterion", &criterion, "Criterion, 'mmi'|'mpfe'|'smbr', "
+                   "determines the objective function to use.  Should match "
+                   "option used when we created the examples.");
+    opts->Register("acoustic-scale", &acoustic_scale, "Weighting factor to "
+                   "apply to acoustic likelihoods.");
+    opts->Register("drop-frames", &drop_frames, "For MMI, if true we drop frames "
+                   "with no overlap of num and den frames");
```
danpovey: s/frames/pdf-ids/.  Same in where you got this code from.

src/nnet3/discriminative-training.cc
vimalmanohar: I meant the xent output is probably not well trained enough in the case of
chain model. So it may not be good to do sequence training on top of it

On Mon, Mar 7, 2016, 19:50 Daniel Povey <notifications@github.com> wrote:

> In src/nnet3/discriminative-training.cc
> <https://github.com/kaldi-asr/kaldi/pull/563#discussion_r55300406>:
>
> > +  // frames in paths in the 'supervision' object (which has all frames of the
> > +  // 1st sequence first, then the 2nd sequence, and so on).  Instead, the
> > +  // frames in 'nnet_output' are ordered as: first the first frame of each
> > +  // sequence, then the second frame of each sequence, and so on.
> > +  // This is done to be similar to the setup in 'chain' training
> > +  // even though this does not offer any computational advantages here
> > +  // as in the 'chain' case.
> > +  DiscriminativeComputation(const DiscriminativeTrainingOptions &opts,
> > +      const TransitionModel &tmodel,
> > +      const CuVectorBase<BaseFloat> &log_priors,
> > +      const DiscriminativeSupervision &supervision,
> > +      const CuMatrixBase<BaseFloat> &nnet_output,
> > +      DiscriminativeTrainingStats *stats,
> > +      BaseFloat *l2_term,
> > +      CuMatrixBase<BaseFloat> *nnet_output_deriv,
> > +      CuMatrixBase<BaseFloat> *xent_output_deriv);
>
> I'm not sure why you said what you just said about "I remember decoding
> with the softmax part of the output, but the results were 10% absolute
> worse than chain part. I will try again to make sure I did correctly." I
> think you may have misinterpreted something I said. This is not relevant.
> … <#msg-f:1528192811736950802_>
> On Mon, Mar 7, 2016 at 7:47 PM, Vimal Manohar <notifications@github.com>
> wrote: In src/nnet3/discriminative-training.cc <#563 (comment)
> <https://github.com/kaldi-asr/kaldi/pull/563#discussion_r55300096>>: > +
> // frames in paths in the 'supervision' object (which has all frames of the
> > + // 1st sequence first, then the 2nd sequence, and so on). Instead, the
> > + // frames in 'nnet_output' are ordered as: first the first frame of
> each > + // sequence, then the second frame of each sequence, and so on. >
> + // This is done to be similar to the setup in 'chain' training > + //
> even though this does not offer any computational advantages here > + // as
> in the 'chain' case. > + DiscriminativeComputation(const
> DiscriminativeTrainingOptions &opts, > + const TransitionModel &tmodel, > +
> const CuVectorBase<BaseFloat> &log_priors, > + const
> DiscriminativeSupervision &supervision, > + const CuMatrixBase<BaseFloat>
> &nnet_output, > + DiscriminativeTrainingStats *stats, > + BaseFloat
> *l2_term, > + CuMatrixBase<BaseFloat> *nnet_output_deriv, > +
> CuMatrixBase<BaseFloat> *xent_output_deriv); I remember decoding with the
> softmax part of the output, but the results were 10% absolute worse than
> chain part. I will try again to make sure I did correctly. — Reply to this
> email directly or view it on GitHub <
> https://github.com/kaldi-asr/kaldi/pull/563/files#r55300096>.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/563/files#r55300406>.
>
-- 
Vimal Manohar
PhD Student
Electrical & Computer Engineering
Johns Hopkins University

vimalmanohar: I had assumed by "why not use the single output of the nnet, which already has a softmax, for both purposes", you meant for me to rename "output-xent" as "output" and use that for sequence training. This gives poor results (actually only 2% worse WER after tuning acwt and post-decode-acwt) than using "output".  So it is better to use "output". Hence  we need xent_output_deriv. We can even use this with CE models that don't have "output-xent" by modifying the nnet config prior to sequence training. This way we can use the same code for both chain and non-chain models.

danpovey: On Tue, Mar 8, 2016 at 1:10 AM, Vimal Manohar <notifications@github.com>
wrote:

> In src/nnet3/discriminative-training.cc
> <https://github.com/kaldi-asr/kaldi/pull/563#discussion_r55320043>:
>
> > +  // frames in paths in the 'supervision' object (which has all frames of the
> > +  // 1st sequence first, then the 2nd sequence, and so on).  Instead, the
> > +  // frames in 'nnet_output' are ordered as: first the first frame of each
> > +  // sequence, then the second frame of each sequence, and so on.
> > +  // This is done to be similar to the setup in 'chain' training
> > +  // even though this does not offer any computational advantages here
> > +  // as in the 'chain' case.
> > +  DiscriminativeComputation(const DiscriminativeTrainingOptions &opts,
> > +      const TransitionModel &tmodel,
> > +      const CuVectorBase<BaseFloat> &log_priors,
> > +      const DiscriminativeSupervision &supervision,
> > +      const CuMatrixBase<BaseFloat> &nnet_output,
> > +      DiscriminativeTrainingStats *stats,
> > +      BaseFloat *l2_term,
> > +      CuMatrixBase<BaseFloat> *nnet_output_deriv,
> > +      CuMatrixBase<BaseFloat> *xent_output_deriv);
>
> I had assumed by "why not use the single output of the nnet, which already
> has a softmax, for both purposes", you meant for me to rename "output-xent"
> as "output" and use that for sequence training.
>

I was talking about the non-chain case.


> This gives poor results (actually only 2% worse WER after tuning acwt and
> post-decode-acwt) than using "output". So it is better to use "output".
> Hence we need xent_output_deriv. We can even use this with CE models that
> don't have "output-xent" by modifying the nnet config prior to sequence
> training. This way we can use the same code for both chain and non-chain
> models.
>
That's an odd thing to do, in my opinion.  And since your experiments seem
to be finding that discriminative training does not even work for the chain
models, it's not clear that the chain models should be dominating the
design of this code.

Dan

vimalmanohar: I have only done one experiment on Switchboard. It would be best to try on other setups.

danpovey: OK, I can probably merge this before we finalize this issue in order for
things to be done before the interpseech deadline.
But there are some minor issues I pointed out that you need to address.

vimalmanohar:
I have addressed all the issues and squashed the commits.

src/latbin/lattice-determinize-non-compact.cc
```diff
@@ -0,0 +1,329 @@
...
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "util/stl-utils.h"
+#include "fstext/fstext-lib.h"
+#include "lat/kaldi-lattice.h"
+#include "lat/lattice-functions.h"
+#include "lat/push-lattice.h"
+#include "lat/minimize-lattice.h"
+
+#ifdef _MSC_VER
```
danpovey: you don't need any of this ifdef stuff, it's included in one of the standard Kaldi headers.
Please also remove it from wherever you got it from.

danpovey: i mean any of the declarations regarding unordered_map.

src/nnet3/discriminative-training.h
```diff
@@ -0,0 +1,244 @@
...
+  double tot_objf;      // for 'mmi', the (weighted) denominator likelihood; for
+                        // everything else, the objective function.
+  double tot_num_count; // total count of numerator posterior 
+  double tot_den_count; // total count of denominator posterior 
+  double tot_num_objf;  // for 'mmi', the (weighted) numerator likelihood; for
+                        // everything else 0
+
+  DiscriminativeTrainingStatsOptions config;
+
+  // Used to accumulates gradients when config.accumulate_gradients is true
```
danpovey: clarify that this means the average gradient w.r.t. the neural net output.

src/nnet3/discriminative-training.h
```diff
@@ -0,0 +1,244 @@
...
+                   "Number of pdfs");
+  }
+
+  DiscriminativeTrainingStatsOptions() :
+    accumulate_gradients(false), accumulate_output(false),
+    num_pdfs(0) { }
+};
+
+struct DiscriminativeTrainingStats {
+  double tot_t;          // total number of frames
```
danpovey: DiscriminativeTrainingStats is not a good name because it conjures up sufficient statistics, which would imply this is the main thing you're accumulating.  You could call it DiscriminativeObjectiveInfo.

src/nnet3/discriminative-training.h
```diff
@@ -0,0 +1,244 @@
...
+  void Add(const DiscriminativeTrainingStats &other);
+
+  // Returns the objective function value for the criterion
+  inline double TotalObjf(const std::string &criterion) const {
+    if (criterion == "mmi") return (tot_num_objf - tot_objf);
+    return tot_objf;
+  }
+
+  // Returns the weighted count
+  inline double TotalT() const {
```
danpovey: I think there is no need for this function.

src/nnet3/nnet-discriminative-diagnostics.h
```diff
@@ -0,0 +1,104 @@
...
+#include "nnet3/nnet-compute.h"
+#include "nnet3/nnet-optimize.h"
+#include "nnet3/nnet-discriminative-example.h"
+#include "nnet3/nnet-diagnostics.h"
+#include "nnet3/discriminative-training.h"
+
+namespace kaldi {
+namespace nnet3 {
+
+struct DiscriminativeObjectiveInfo {
```
danpovey: ok, you are already using the name DiscriminativeObjectiveInfo.
Why don't you just take all the members of DiscriminativeTrainingStats, and put them here?

danpovey: .. I was wondering why tot_l2_term is so special that it has to be put in another level of struct.

src/nnet3/nnet-discriminative-diagnostics.h
```diff
@@ -0,0 +1,104 @@
...
+  
+  ~NnetDiscriminativeComputeObjf();
+ private:
+  void ProcessOutputs(const NnetDiscriminativeExample &eg,
+                      NnetComputer *computer);
+
+  NnetComputeProbOptions nnet_config_;
+
+  discriminative::DiscriminativeTrainingOptions discriminative_training_config_;
+  discriminative::DiscriminativeTrainingStatsOptions discriminative_training_stats_config_;
```
danpovey: is it really necessary to have a separate options class for the stats?

src/nnet3/nnet-discriminative-example.cc
```diff
@@ -0,0 +1,419 @@
...
+    }
+  }
+  // check to see if something went wrong.
+  if (request->inputs.empty())
+    KALDI_ERR << "No inputs in computation request.";
+  if (request->outputs.empty())
+    KALDI_ERR << "No outputs in computation request.";
+}
+
+void ShiftDiscriminativeExampleTimes(int32 frame_shift,
```
danpovey: I'm not sure that this function makes sense to have.
The reason this was OK for the chain models is that their numerator graph gives them a bit of wiggle room on the time; but with denominator lattices, the times are fixed, and they might change if you shift the data by less than a frame.  
So if you wanted to cover all shifts of the data properly, you'd have to generate the lattices multiple different times.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+    }
+    CuArray<Int32Pair> cu_deriv_indexes(deriv_indexes);
+    xent_output_deriv_->AddElements(supervision_.weight, cu_deriv_indexes, deriv_data.data());
+  }
+
+
+  this_stats.tot_den_count += tot_den_post;
+  this_stats.tot_num_count += tot_num_post;
+
+  if (this_stats.AccumulateGradients())
```
danpovey: there is no need to have a function to tell you whether to accumulate these gradients.  just check the config value.  it will be easier if you have just one config class rather than two.

src/nnet3/discriminative-training.cc
```diff
@@ -0,0 +1,602 @@
...
+  }
+
+
+  this_stats.tot_den_count += tot_den_post;
+  this_stats.tot_num_count += tot_num_post;
+
+  if (this_stats.AccumulateGradients()) 
+    (this_stats.gradients).AddRowSumMat(1.0, CuMatrix<double>(*output_deriv_temp));
+  
+  if (this_stats.AccumulateOutput()) {
```
danpovey: .. and the same here.

src/nnet3/nnet-discriminative-training.h
```diff
@@ -0,0 +1,146 @@
...
+  }
+};
+
+// This struct is used in multiple nnet training classes for keeping
+// track of objective function values.
+// Also see struct AccuracyInfo, in nnet-diagnostics.h.
+struct DiscriminativeObjectiveFunctionInfo {
+  int32 current_phase;
+
+  double tot_aux_objf;
```
danpovey: I'm not sure if you changed this code since my last comments, or if this is different class?
What is tot_aux_objf needed for?  My suspicion is that either it is not needed, or you could give it a more specific name to indicate its exact function.
And I don't think we need a struct to wrap a struct here-- please just have one level of struct for this.

src/nnet3/nnet-discriminative-training.cc
```diff
@@ -0,0 +1,289 @@
...
+    PrintStatsForThisPhase(output_name, criterion, minibatches_per_phase);
+    current_phase = phase;
+    stats_this_phase.Reset();
+    tot_aux_objf_this_phase = 0.0;
+  }
+  stats_this_phase.tot_t_weighted += this_minibatch_weight;
+  stats_this_phase.tot_objf += this_minibatch_tot_objf;
+  tot_aux_objf_this_phase += this_minibatch_tot_aux_objf;
+
+  stats.Add(stats_this_phase);
```
danpovey: this looks like a bug- you seem to be adding stats_this_phase to stats each time, so you are counting the same minibatch multiple times.

src/nnet3/nnet-discriminative-training.cc
```diff
@@ -0,0 +1,289 @@
...
+    KALDI_ASSERT(phase == current_phase + 1); // or doesn't really make sense.
+    PrintStatsForThisPhase(output_name, criterion, minibatches_per_phase);
+    current_phase = phase;
+    stats_this_phase.Reset();
+    tot_aux_objf_this_phase = 0.0;
+  }
+  stats_this_phase.Add(this_minibatch_stats);
+  tot_aux_objf_this_phase += this_minibatch_tot_aux_objf;
+
+  stats.Add(stats_this_phase);
```
danpovey: ditto here- likely a bug

src/nnet3/nnet-discriminative-training.h
```diff
@@ -0,0 +1,146 @@
...
+  // control how frequently we print logging messages.
+  void UpdateStats(const std::string &output_name,
+                   const std::string &criterion,
+                   int32 minibatches_per_phase,
+                   int32 minibatch_counter,
+                   BaseFloat this_minibatch_weight,
+                   BaseFloat this_minibatch_tot_objf,
+                   BaseFloat this_minibatch_tot_aux_objf = 0.0);
+  
+  void UpdateStats(const std::string &output_name,
```
danpovey: why are two UpdateStats functions needed?

src/nnet3/nnet-discriminative-training.h
vimalmanohar: This is a different class. Its following the same type of structures as
used for CE and chain. There is an ObjectiveInfo in nnet-diagnostics and
ObjectiveFunctionInfo in nnet-training. ObjectiveFunctionInfo stores
current phase etc. additionally. The inner structure would be for storing
things that are common to both diagnostics and training.
aux_objf exists in CE and chain for using in regularization etc. even
though it is currently not in use I think.

On Tue, Mar 8, 2016 at 6:11 PM Daniel Povey <notifications@github.com>
wrote:

> In src/nnet3/nnet-discriminative-training.h
> <https://github.com/kaldi-asr/kaldi/pull/563#discussion_r55445700>:
>
> > +    nnet_config.Register(opts);
> > +    discriminative_training_config.Register(opts);
> > +    discriminative_training_stats_config.Register(opts);
> > +    opts->Register("apply-deriv-weights", &apply_deriv_weights,
> > +                   "If true, apply the per-frame derivative weights stored with "
> > +                   "the example (you'll normally want to leave this as true.");
> > +  }
> > +};
> > +
> > +// This struct is used in multiple nnet training classes for keeping
> > +// track of objective function values.
> > +// Also see struct AccuracyInfo, in nnet-diagnostics.h.
> > +struct DiscriminativeObjectiveFunctionInfo {
> > +  int32 current_phase;
> > +
> > +  double tot_aux_objf;
>
> I'm not sure if you changed this code since my last comments, or if this
> is different class?
> What is tot_aux_objf needed for? My suspicion is that either it is not
> needed, or you could give it a more specific name to indicate its exact
> function.
> And I don't think we need a struct to wrap a struct here-- please just
> have one level of struct for this.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/563/files#r55445700>.
>
-- 
Vimal Manohar
PhD Student
Electrical & Computer Engineering
Johns Hopkins University

src/nnet3/nnet-discriminative-training.h
```diff
@@ -0,0 +1,146 @@
...
+
+  NnetDiscriminativeTrainingOptions(): apply_deriv_weights(true) { }
+
+  void Register(OptionsItf *opts) {
+    nnet_config.Register(opts);
+    discriminative_training_config.Register(opts);
+    discriminative_training_stats_config.Register(opts);
+    opts->Register("apply-deriv-weights", &apply_deriv_weights,
+                   "If true, apply the per-frame derivative weights stored with "
+                   "the example (you'll normally want to leave this as true.");
```
danpovey: remove (you'll normally want to leave this as true

src/nnet3/nnet-discriminative-training.h
```diff
@@ -0,0 +1,146 @@
...
+#include "nnet3/discriminative-training.h"
+
+namespace kaldi {
+namespace nnet3 {
+
+struct NnetDiscriminativeTrainingOptions {
+  NnetTrainerOptions nnet_config;
+  discriminative::DiscriminativeTrainingOptions discriminative_training_config;
+  discriminative::DiscriminativeTrainingStatsOptions discriminative_training_stats_config;
+
```
danpovey: can you please rename DiscriminativeTrainingOptions to DiscriminativeOptions (the name is too long), and combine these two discriminative config classes into one?

src/nnet3/discriminative-training.h
```diff
@@ -0,0 +1,244 @@
...
+    accumulate_gradients(false), accumulate_output(false),
+    num_pdfs(0) { }
+};
+
+struct DiscriminativeTrainingStats {
+  double tot_t;          // total number of frames
+  double tot_t_weighted; // total number of frames times weight.
+  double tot_objf;      // for 'mmi', the (weighted) denominator likelihood; for
+                        // everything else, the objective function.
+  double tot_num_count; // total count of numerator posterior
```
danpovey: I don't think separate counts are needed for num and den-- I can't think of a scenario where they could possibly be different.

danpovey: actually it's OK- I see that they are computed separately.

src/nnet3/nnet-discriminative-training.h
vimalmanohar: The two UpdateStats are for sequence and xent. The objf for xent is not stored inside the structure, which can store additional stuff like averaged gradients, averaged output etc.

src/nnet3/discriminative-training.h
```diff
@@ -0,0 +1,244 @@
...
+struct DiscriminativeTrainingStats {
+  double tot_t;          // total number of frames
+  double tot_t_weighted; // total number of frames times weight.
+  double tot_objf;      // for 'mmi', the (weighted) denominator likelihood; for
+                        // everything else, the objective function.
+  double tot_num_count; // total count of numerator posterior 
+  double tot_den_count; // total count of denominator posterior 
+  double tot_num_objf;  // for 'mmi', the (weighted) numerator likelihood; for
+                        // everything else 0
+
```
danpovey: you could add a tot_l2 value here (but make clear whether it's normalized by tot_t or tot_t_weighted).  Then you wouldn't need class DiscriminativeObjectiveInfo, you could rename this class to DiscriminativeObjectiveInfo instead.

src/nnet3/nnet-discriminative-training.h
vimalmanohar: I can move the xent objf inside the same structure.

src/nnet3/nnet-discriminative-training.h
```diff
@@ -0,0 +1,146 @@
...
+// This struct is used in multiple nnet training classes for keeping
+// track of objective function values.
+// Also see struct AccuracyInfo, in nnet-diagnostics.h.
+struct DiscriminativeObjectiveFunctionInfo {
+  int32 current_phase;
+
+  double tot_aux_objf;
+  discriminative::DiscriminativeTrainingStats stats;
+  
+  double tot_aux_objf_this_phase;
```
danpovey: OK, you can keep this class, but remove the aux_objf stuff, it's not needed; you just need two members, stats and stats_this_phase; and please fix the bug I mentioned-- the easiest fix is to not use the Add() function bug add to the stats of both class-members each time UpdateStats is called.  And I strongly suspect that you only really need one UpdateStats function here.

src/nnet3/nnet-discriminative-training.h
danpovey: OK, move it inside.
And you shouldn't overload functions like this- you need to update the
names to something meaningful.
Dan

src/nnet3/discriminative-training.h
vimalmanohar: Yes, I did this and will be updated in the next commit. The DiscriminativeObjectiveFunctionInfo will still be there as it needs to store stats per minibatch.

src/nnet3/nnet-discriminative-example.cc
vimalmanohar: For chain models, I have sub-sampled alignments and lattices. So I need to shift input by 1 and 2 frames in the decode and align scripts to get the different shifts?
Also would this affect the chain + sMBR results?

danpovey: No, I am saying you should not shift without regenerating the lattices with
the data at the correct shift.  It would be easiest for now to do it using
shift-feats at the command line level, and just dump separate egs.
Dan

danpovey: Actually, I think it would make sense to compare 2 ways of doing the shifts:
  - (a), dumping the lattices with no shift and doing the shift at the egs
level the same way we do for regular chain training, with the
'ShiftDiscriminativeExample()' function or whatever it was called,
   - (b), a more exact method where we dump the lattices at all shifts.
This is just for doing due diligence on the chain+discriminative-training
setup.  I don't really think it's going to give us much improvement, and
it's not that urgent.
Dan

vimalmanohar: (a) is what I have results for. I can make the scripts to support (b) to
try later.

On Tue, Mar 8, 2016, 20:43 Daniel Povey <notifications@github.com> wrote:

> In src/nnet3/nnet-discriminative-example.cc
> <https://github.com/kaldi-asr/kaldi/pull/563#discussion_r55460891>:
>
> > +      // the IoSpecification for the -xent output is the same
> > +      // as for the regular output, except for its name which has
> > +      // the -xent suffix (and the has_deriv member may differ).
> > +      io_spec_xent = io_spec;
> > +      io_spec_xent.name = name + "-xent";
> > +      io_spec_xent.has_deriv = use_xent_derivative;
> > +    }
> > +  }
> > +  // check to see if something went wrong.
> > +  if (request->inputs.empty())
> > +    KALDI_ERR << "No inputs in computation request.";
> > +  if (request->outputs.empty())
> > +    KALDI_ERR << "No outputs in computation request.";
> > +}
> > +
> > +void ShiftDiscriminativeExampleTimes(int32 frame_shift,
>
> Actually, I think it would make sense to compare 2 ways of doing the
> shifts: - (a), dumping the lattices with no shift and doing the shift at
> the egs level the same way we do for regular chain training, with the
> 'ShiftDiscriminativeExample()' function or whatever it was called, - (b), a
> more exact method where we dump the lattices at all shifts. This is just
> for doing due diligence on the chain+discriminative-training setup. I don't
> really think it's going to give us much improvement, and it's not that
> urgent. Dan
> … <#msg-f:1528287055664639253_msg-f:1528286726484747763_>
> On Tue, Mar 8, 2016 at 8:32 PM, Daniel Povey <dpovey@gmail.com> wrote: >
> No, I am saying you should not shift without regenerating the lattices >
> with the data at the correct shift. It would be easiest for now to do it >
> using shift-feats at the command line level, and just dump separate egs. >
> Dan > > > On Tue, Mar 8, 2016 at 7:53 PM, Vimal Manohar <
> notifications@github.com> > wrote: > In
> src/nnet3/nnet-discriminative-example.cc > <#563 (comment)
> <https://github.com/kaldi-asr/kaldi/pull/563#discussion_r55456425>>: > >
> > + // the IoSpecification for the -xent output is the same > > + // as for
> the regular output, except for its name which has > > + // the -xent suffix
> (and the has_deriv member may differ). > > + io_spec_xent = io_spec; > > +
> io_spec_xent.name = name + "-xent"; > > + io_spec_xent.has_deriv =
> use_xent_derivative; > > + } > > + } > > + // check to see if something
> went wrong. > > + if (request->inputs.empty()) > > + KALDI_ERR << "No
> inputs in computation request."; > > + if (request->outputs.empty()) > > +
> KALDI_ERR << "No outputs in computation request."; > > +} > > + > > +void
> ShiftDiscriminativeExampleTimes(int32 frame_shift, > > For chain models, I
> have sub-sampled alignments and lattices. So I need > to shift input by 1
> and 2 frames in the decode and align scripts to get the > different shifts?
> > Also would this affect the chain + sMBR results? > > — > Reply to this
> email directly or view it on GitHub > <
> https://github.com/kaldi-asr/kaldi/pull/563/files#r55456425>. >
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/563/files#r55460891>.
>
-- 
Vimal Manohar
PhD Student
Electrical & Computer Engineering
Johns Hopkins University

danpovey:
Can you get a GPU and do 'make test' at the top level to make sure everything still compiles?

danpovey:
it's OK, I'll do it myself after I merge.

PR #567, opened 2016-03-09
specifying namespace std when using cerr in src/lm/kaldi-lmtable.cc. …

xiaohui-zhang:
…This only affects compilation with some versions of openfst.


danpovey:
Samuel, I know this is what I asked you to do, but I just realized they should be KALDI_LOG.  Could you change it to that please?

xiaohui-zhang:
sure!

Issue #569, opened 2016-03-09
nnet3_train_lib.ParseModelConfigVars failing due to changes made in raw_nnet commit


vijayaditya:
@vimalmanohar Could you ensure that your changes do not break the current recipes. You could modify the ParseModelConfigVars.

vimalmanohar:
I had asked about this function. This happened because the python and bash scripts parse the same file in different ways. I suggest we should modify the python function to ensure the config file is parsed the same as the bash sourcing. i.e. ignore extra variables that are defined in the config file.

vimalmanohar:
https://github.com/vimalmanohar/kaldi/blob/raw_nnet_python/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py#L249
This is my proposed change. We can change the train_tdnn.py script to work with this. It will be compatible with previously generated configs.

vijayaditya:
I remember asking you to write a new function to parse these config files. Ensure backward compatibility with whatever changes you propose.

PR #572, opened 2016-03-09
raw_fix: Modified python function parsing to be similar to bash sourcing

vimalmanohar:
#569 @vijayaditya

### DIFF
--- a/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
+++ b/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
@@ -175,25 +175,22 @@ def SplitData(data, num_jobs):
                                                                   num_jobs = num_jobs))
 
 def ParseModelConfigVarsFile(var_file):
+    variables = {}
     try:
         var_file_handle = open(var_file, 'r')
-        model_left_context = None
-        model_right_context = None
-        num_hidden_layers = None
         for line in var_file_handle:
             parts = line.split('=')
             field_name = parts[0].strip()
-            field_value = parts[1]
+            field_value = parts[1].strip()
             if field_name in ['model_left_context', 'left_context']:
-                model_left_context = int(field_value)
+                variables['model_left_context'] = int(field_value)
             elif field_name in ['model_right_context', 'right_context']:
-                model_right_context = int(field_value)
+                variables['model_right_context'] = int(field_value)
             elif field_name == 'num_hidden_layers':
-                num_hidden_layers = int(field_value)
-
-        if model_left_context is not None and model_right_context is not None and num_hidden_layers is not None:
-            return [model_left_context, model_right_context, num_hidden_layers]
-
+                variables['num_hidden_layers'] = int(field_value)
+            else:
+                variables[field_name] = field_value
+        return variables
     except ValueError:
         # we will throw an error at the end of the function so I will just pass
         pass
--- a/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
+++ b/egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
@@ -24,6 +24,14 @@ def SendMail(message, subject, email_id):
         logger.info(" Unable to send mail due to error:\n {error}".format(error = str(e)))
         pass
 
+def StrToBool(values):
+    if values == "true":
+        return True
+    elif values == "false":
+        return False
+    else
+        raise ValueError
+
 class StrToBoolAction(argparse.Action):
     """ A custom action to convert bools from shell format i.e., true/false
         to python format i.e., True/False """
@@ -180,6 +188,10 @@ def ParseModelConfigVarsFile(var_file):
         var_file_handle = open(var_file, 'r')
         for line in var_file_handle:
             parts = line.split('=')
+            if len(parts) == 0:
+                continue
+            if len(parts) != 2:
+                break
             field_name = parts[0].strip()
             field_value = parts[1].strip()
             if field_name in ['model_left_context', 'left_context']:
@@ -188,6 +200,8 @@ def ParseModelConfigVarsFile(var_file):
                 variables['model_right_context'] = int(field_value)
             elif field_name == 'num_hidden_layers':
                 variables['num_hidden_layers'] = int(field_value)
+            elif field_name == 'add_lda':
+                variables['add_lda'] = StrToBool(field_value)
             else:
                 variables[field_name] = field_value
         return variables
--- a/egs/wsj/s5/steps/nnet3/train_rnn.py
+++ b/egs/wsj/s5/steps/nnet3/train_rnn.py
@@ -500,7 +500,15 @@ def Train(args, run_opts):
     config_dir = '{0}/configs'.format(args.dir)
     var_file = '{0}/vars'.format(config_dir)
 
-    [model_left_context, model_right_context, num_hidden_layers] = ParseModelConfigVarsFile(var_file)
+    variables = ParseModelConfigVarsFile(var_file)
+    try:
+        model_left_context = variables['model_left_context']
+        model_right_context = variables['model_right_context']
+        num_hidden_layers = variables['num_hidden_layers']
+    except KeyError as e:
+        raise Exception("KeyError({0}): {1}. Variables need to be defined in {2}".format(
+            e.errno, e.strerror, '{0}/configs'.format(args.dir)))
+
     # Initialize as "raw" nnet, prior to training the LDA-like preconditioning
     # matrix.  This first config just does any initial splicing that we do;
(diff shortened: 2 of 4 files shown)


egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -175,25 +175,22 @@ def SplitData(data, num_jobs):
...
+                variables['model_right_context'] = int(field_value)
             elif field_name == 'num_hidden_layers':
-                num_hidden_layers = int(field_value)
-
-        if model_left_context is not None and model_right_context is not None and num_hidden_layers is not None:
-            return [model_left_context, model_right_context, num_hidden_layers]
-
+                variables['num_hidden_layers'] = int(field_value)
+            else:
+                variables[field_name] = field_value
```
vijayaditya: You could handle the booleans too.. if someone did

if add_lda:

it would evaluate to true if the add_lda was either "true" or "false".
better convert all booleans to python bools.

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -500,7 +500,15 @@ def Train(args, run_opts):
     config_dir = '{0}/configs'.format(args.dir)
     var_file = '{0}/vars'.format(config_dir)
```
vijayaditya: could you please also add the changes to steps/nnet3/chain/train.py

vimalmanohar: I already made changes in that.

vimalmanohar:
I have added add_lda option too.

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -175,25 +183,28 @@ def SplitData(data, num_jobs):
...
-                model_right_context = int(field_value)
+                variables['model_right_context'] = int(field_value)
             elif field_name == 'num_hidden_layers':
-                num_hidden_layers = int(field_value)
-
-        if model_left_context is not None and model_right_context is not None and num_hidden_layers is not None:
-            return [model_left_context, model_right_context, num_hidden_layers]
-
+                variables['num_hidden_layers'] = int(field_value)
+            elif field_name == 'add_lda':
```
vijayaditya: better do a more generic conversion. If field_value is true/false convert it to python bool

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -175,25 +183,28 @@ def SplitData(data, num_jobs):
                                                                   num_jobs = num_jobs))
 
 def ParseModelConfigVarsFile(var_file):
+    variables = {}
```
danpovey: guys- I don't see why you are adding this huge mechanism.  Didn't the old method work fine?  And didn't we only want to parse a few specific variables and ignore the rest?
Is there a specific projected use for this?

danpovey:
I don't think this is the right direction.
I previously made a fix to this function in one place, to move a conversion to 'int' inside the if-statements (after you had checked the variable name).  It was just a change to a handful of lines.  That worked.
You are adding tons of code and I don't see that it adds any benefit.  Unless you have a good argument, I don't think you should pursue this.

vijayaditya:
In his new raw_nnet scripts
<https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh#L166>
Vimal dumps additional info like objective_type in the <model>/config/vars
file. So we started supporting additional info in this file.

Also, as we moved config generation outside the training script, it would
be good to ensure that the config/vars is consistent with the arguments
provided to the training script. e.g. we could perform checks like
include-log-softmax is true and --l2-regularization is non-zero , so mostly
there is some error in the user script.

--Vijay

danpovey:
In his new raw_nnet scripts
> <
> https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/nnet3/tdnn/train_raw_nnet.sh#L166
> >
> Vimal dumps additional info like objective_type in the <model>/config/vars
> file. So we started supporting additional info in this file.
>
> Also, as we moved config generation outside the training script, it would
> be good to ensure that the config/vars is consistent with the arguments
> provided to the training script. e.g. we could perform checks like
> include-log-softmax is true and --l2-regularization is non-zero , so mostly
> there is some error in the user script.
>

I think it would be better to support a specific list of variable names,
extending the list as needed, and set any unset ones to 'None' if needed.
That way the calling code is much simpler.

Dan

danpovey:
e.g. instead of returning [left-context, right-context, num-layers] the
function would return
[left-context, right-context, num-layers, l2-regularize].
In your proposed code you are pushing the hassle onto the outer function.

danpovey:
Vijay, I see that this PR started because you asked Vimal to make sure his changed did not cause ParseModelConfigVars to crash.
I had already changed it to make sure it didn't crash, and it works fine.  I think I made the same changes in the chain lib.
Right now I think there is no reason to make any changes at all, unless there is a specific bug that you are trying to solve that I don't know about.

vijayaditya:
The script in the current state does not support parsing the add_lda option
and other variables being added by the config_generator (
https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/nnet3/tdnn/make_configs.py#L595).
So the script crashes when used in master branch. This was identified from
Tom's attempts to run CNN experiments with xent training.

You were probably looking at the chain branch where these changes can't be
seen, as I removed them temporarily to avoid some issues being caused by
conflicts between vimal's tdnn/make_configs.py  changes and my changes.

--Vijay

danpovey:
Can you show me how it crashes?

danpovey:
BTW, I fixed that crash previously by moving the conversion to 'int' inside
the if statements, and that fix is in master too.  So if it was crashing I
suspect it was due to out-of-date code.

vijayaditya:
OK. In that case this is not immediately needed. 

However we might still take advantage of the variables being dumped in the config/vars file to perform checks (include-logsoftmax and l2-regularize) and avoid stages (e.g. avoid PreconditioningMatrix computation if add_lda is false). 

So Vimal could you just modify the code to return multiple values ?
[left-context, right-context, num-layers,  add_lda, l2-regularize].

vimalmanohar:
This is actually infeasible and not back-compatible. What do you do if the configs do not have the options add_lda and l2_regularize? Also different dnn architectures may have different of variables. I suppose add_lda is used for raw nnet and CNN, but not the other architectures. l2_regularizer does not make sense in non-chain system.

vijayaditya:
hmm.. this discussion is just going around in circles.

As Dan suggested it is better to do the error checking in the function rather than the calling code, to keep the training scripts simple. As the chain/xent can decide which values they want they could call an appropriate parsing function, which parses the values and does the appropriate checks.

You could close this PR for now, I will just add necessary functions when we want them.

danpovey:
My proposal was to just leave the variables as None in that case.

Dan

vijayaditya:
I guess Vimal's concern is that we would have to change the function calls in all scripts whenever we want additional values in one script, so he was using the dictionary.

vimalmanohar:
If the issue is that the training script looks "complicated", the conversion from dictionary to array can be done by a separate function as Vijay suggested. This solves the issue of needing to change function in all the script when we add additional values. This is a recurrence of the previously solved issue. https://github.com/kaldi-asr/kaldi/pull/522

danpovey:
How about something like
[a,b,c] = ParseConfigFile(config_file, 'a', 'b', 'c')
and have it always just return strings, or None if not defined?  Then the
use can handle the conversion to int or bool or whatever?

Dan

vijayaditya:
I think it would be easier to return a dictionary, the user could specify the variables they need and the function could check for the existence of these variables.

<code>value_dict = ParseConfigFile(config_file, ['a','b','c'])</code> value_dict would have 'a','b' and 'c'.

danpovey:
If you return a dictionary then the user has to do lookup anyway so there
is no reason to pass the args.
OK, do it the way you wanted.  But call it something different.
Dan

vijayaditya:
The args being passed to ensure that the function checks for the existence of these variables.

vimalmanohar:
One approach would be to do something similar to reading Nnet3 component configs.

[a,b,c] = ParseVariables(ParseConfigFile(config_file)),
where the ParseConfigFile is common to all files and returns a dictionary.
ParseVariables reads specific things out of it and return an array. ParseVariables is defined in the train scripts and hence it can return the variables in their appropriate type. e.g. left_context as integer.

vijayaditya:
@vimalmanohar lets take this discussion offline. We might be spamming the list.

PR #573, opened 2016-03-09
Add options for score.sh to rescoring script.

galv:
We want to specify the language model weight range for decoding in
chain models, but previously lmrescore_const_arpa.sh did not allow
passing in arguments to score.sh. This is also necessary to specify
--decode-mbr false to score.sh for chain scoring in TEDLIUM.


galv:
You may want to modify your swbd setup to use this option. I noticed that your tdnn_6v chain setup uses language model weights from 5 to 20 instead of 8 to 12.

danpovey:
I may in future but it's not urgent--it doesn't take very long so it's just
a bit of wasted computation.

PR #579, opened 2016-03-11
Major rewrite of ARPA compiler

### LINKED ISSUE #422: Rewrite ARPA-FST compiler

### LINKED ISSUE #565: arpa2fst rewrite
rewrite arpa2fst to use @kkm000's new code.

kkm000:
A few things remain to be done still (mostly noted to myself)
 - Expand tests to the epsilon-substitution mode
 - Add the subst mode to the new arpa2fst
 - Carefully select the optimized compilation mode based on grammar size and symbol table size.

I am removing a few files and moving arpa2fst to src/lmbin. I remember there is a trick involved when removing library members, but I do not remember what was it  (@jtrmal, please remind me).

I mostly inherited the tests (however had to rewrote them, and hopefully made them more palatable).

This is still WIP, and I think the "old" bin/arpa2fst is currently broken by the removal of the old sources in this directory. But please comment on the code.

I will be mostly totally offline till Monday.

Closes #422 
Closes #565


kkm000:
This should be good to go (in the sense it's no longer WIP). Sorry the changeset is so large, but this is really one thing. I'll send usage changes in scripts in a separate PR (`arpa2fst` is still backward compatible, but invocation can be simplified greatly with the new `--read-symbol-table` and `--disambig-symbol` switches.

src/bin/Makefile
```diff
@@ -7,7 +7,7 @@ BINFILES = align-equal align-equal-compiled acc-tree-stats \
         show-alignments compile-questions cluster-phones \
         compute-wer compute-wer-bootci make-h-transducer \
```
danpovey: We need to make sure arpa2fst is removed from here.  You could do:
all:
      -rm arpa2fst 2>/dev/null

danpovey:
could you please run the test program and whatever tests you were doing on the regular arpa2fst program, in valgrind to check for memory leaks?

src/bin/Makefile
kkm000: Thanks, and also, IIRC, something needs to be done to `lm/kaldi-lm.a` due to removed source files there, but I do not remember the incantation. Do you?

kkm000:
> could you please run the test program and whatever tests you were doing on the regular arpa2fst program, in valgrind to check for memory leaks?

Tests check out fine (some `FstRegisterer` and other expected static leaks). I'll valgrind a 8M LM overnight.

src/bin/Makefile
danpovey: Actually I don't recall how we did that and can't remember (@jtrmal might
know), but in this case it's not important to do, as those symbols won't be
defined elsewhere and won't cause collisions.
Dan


On Tue, Mar 15, 2016 at 9:14 PM, Kirill Katsnelson <notifications@github.com

kkm000:
Added the `rm -f`. Also, no detectable leaks from valgrind. Test:
```
==30643== Command: ./arpa-lm-compiler-test
==30643==
==30643== HEAP SUMMARY:
==30643==     in use at exit: 76,940 bytes in 89 blocks
==30643==   total heap usage: 59,734 allocs, 59,645 frees, 3,009,461 bytes allocated
==30643==
==30643== LEAK SUMMARY:
==30643==    definitely lost: 0 bytes in 0 blocks
==30643==    indirectly lost: 0 bytes in 0 blocks
==30643==      possibly lost: 1,525 bytes in 39 blocks
==30643==    still reachable: 75,415 bytes in 50 blocks
```
arpa2fst with a sizable LM
```
==30683== Command: arpa2fst --disambig-symbol=#0 --read-symbol-table=data/lang/words.txt -
==30683==
==30683== HEAP SUMMARY:
==30683==     in use at exit: 76,950 bytes in 90 blocks
==30683==   total heap usage: 65,944,107 allocs, 65,944,017 frees, 2,142,127,069 bytes allocated
==30683==
==30683== LEAK SUMMARY:
==30683==    definitely lost: 0 bytes in 0 blocks
==30683==    indirectly lost: 0 bytes in 0 blocks
==30683==      possibly lost: 1,525 bytes in 39 blocks
==30683==    still reachable: 75,425 bytes in 51 blocks
==30683==         suppressed: 0 bytes in 0 blocks
```

danpovey:
Thanks!
How does the speed compare with the old arpa2fst?
I assume you verified that the output was the same in sum way?
And

On Tue, Mar 15, 2016 at 9:39 PM, Kirill Katsnelson <notifications@github.com

kkm000:
The speed improvement is not very great for the utility itself, but with the whole pipeline of fstprint/perl processing/fstcompile is very slow. Old script
```
real    2m42.743s
user    4m55.162s
```
new script (just arpa2fst | fstarcsort)
```
real    0m37.042s
user    0m43.567s
```
So, the real time speed-up is about ×4.5. Also, memory consumption is significantly (twice, IIRC) less.

As for testing, I did compare compositions of a few random sentences with the "old" and "new" G.fst on a large LM, and they were identical in structure and weight (in addition to coverage and expected weight testing in arpa-lm-compiler-test). Interesting, however, that `fstequivalent --random oldG.fst newG.fst` did not think the two FSTs were equivalent, and I could not yet trace why. Not sure if they should be, however.

danpovey:
I see you don't have any script changes committed.
If the recommended usage is different there should be script changes.
Dan

kkm000:
I was planning to commit them in a separate MR. This one is just blowing up out of proportion. As it is, arpa2fst is backward-compatible.

danpovey:
ok.
let me think whether to review/check this further before merging.

danpovey:
I am a little concerned that fstequivalent does not think the new and old ones are equivalent.
Does it print anything about why not?  Try higher verbose options-- maybe it will print the paths concerned and the weights concerned, so we'll know why not.

danpovey:
Oh, and when you said they were not equivalent-- is this with the old script, or the new script?

kkm000:
> I am a little concerned that fstequivalent does not think the new and old ones are equivalent.

I am not so much concerned but rather interested why. I have an idea though. I know I generate fewer epsilon transitions than the old code (in fact, none except the backoff transitions), and, since all these epsilons are replaced with #0's on the input side, the machines cannot be equivalent: for example, one accepts `aaron #0 jade` and the other would not accept this sentence, but would accept `aaron #0 #0 jade`. This is all the same for us, but makes them formally different. Since my generated sentence machines to be composed on the left have #0-generating self-loops, like 

```
0 1 0 aaron
0 0 0 #0
1 2 0 jade
1 1 0 #0
2 2 0 #0
2
```
they compose to produce identical results.

I'll try to confirm my hypothesis

And no, debug output from `fstequivalent` is not helpful.

> Oh, and when you said they were not equivalent-- is this with the old script, or the new script?

New.

src/bin/Makefile
jtrmal: I think I used something like
ar d name_of_old_file.o

I got that called every time by re-defining one of the compilation targets
(I think this from makefiles/default_rules.mk):
$(LIBFILE): $(OBJFILES)
in the local Makefile

but I don't know from the top of my head.
y.

On Tue, Mar 15, 2016 at 9:14 PM, Kirill Katsnelson <notifications@github.com

jtrmal: I didn't notice it's not needed anymore, sorry.
y.

danpovey: @xiaohui-zhang, could you please download and compile this patch somewhere
on our grid so I can have a look at it locally, and also run the script
that calls arpa2fst inside one of our standard recipes to help verify that
when called in the standard way, it produces the same output?
Dan

kkm000:
I think I am seeing an indication of a problem. Changing back to WIP just in case. I'll find time tonight to confirm.

kkm000:
There was a bug, and just pushed a fix for it. It is interesting that it required a big LM to manifest itself in a way that it is discoverable. The reason is that the truncated encoded history key becomes a poor hash value of itself, and the Tails() operation (take all words but the head) returns the same type of hash of the correct tails. In a small grammar, like those used in unit tests, collisions do not occur. In a medium sized one, they are to rare to manifest themselves easily. 

With the fix, the FSTs for the large LM reported equivalent in random testing, both "old" and "new" style scripting.
```
$ fstequivalent --random -npath=1000 temp/arpa2fst/old_G.fst temp/arpa2fst/new_G.fst ; echo $?
0
$ fstequivalent --random -npath=1000 temp/arpa2fst/old_G.fst temp/arpa2fst/mid_G.fst ; echo $?
0
```
(mid_G is the G built with the new tool but using old scripts).

But the real million dollar question here is how to unit-test such classes. The class is part of "ugly" implementation hidden in the .cc file, out of reach of unit tests, and the external API tests passed with flying colors on 10-line grammars. Catching this error with unit tests would be sweet and simple though.

kkm000:
I also sampled 10 random paths each from "old" and "new" G, discarding weights ("test sentences"), composed them on the left of all 3 old_, new_ and mid_ machines and verified that all 20 compositions had the same weight of the shortest path.

src/bin/Makefile
kkm000: @xiaohui-zhang, if you want pre-merged MR from GitHub, you can do
```
git fetch origin +refs/pull/579/merge:
git checkout FETCH_HEAD
```
(`origin` here stands for the remote name of the main Kaldi repo).

danpovey:
Cool.
I'll wait for Samuel to do his thing too, and I'll have a look at the usage
etc., and then hopeful y merge it.
Dan



On Wed, Mar 16, 2016 at 9:45 PM, Kirill Katsnelson <notifications@github.com

src/bin/Makefile
xiaohui-zhang: Thanks Krill! sorry for the delay. @danpovey , here is a set of comparison of outputs /export/a11/xzhang/babel/104-pashto-llp/data/lang/{G.fst, G.bk.fst} G.fst is the one produced by this MR. The weights differ a bit, only in the last few digits though. Here is the compiled binary: /export/a11/xzhang/kaldi/src/lmbin/arpa2fst

danpovey: Kirill, can you please remove the --natural-base option and hardcode the
default to true (including in the code)?  There was never a need for this.
Dan

kkm000: Sure. It also exists in the const-arpa compiler--do you want to remove it from that program too?

danpovey: Yes, please.


On Tue, Mar 22, 2016 at 5:50 PM, Kirill Katsnelson <notifications@github.com

kkm000: Done in https://github.com/kaldi-asr/kaldi/pull/579/commits/a6f3b82

src/lmbin/arpa2fst.cc
```diff
@@ -0,0 +1,129 @@
...
+#include "util/kaldi-io.h"
+#include "util/parse-options.h"
+
+
+int main(int argc, char *argv[]) {
+  using namespace kaldi;
+  try {
+    const char *usage  =
+        "Converts an ARPA format language model into a FST\n"
+        "Usage: arpa2fst [opts] (input_arpa|-)  [output_fst|-]\n";
```
danpovey: inconsistent use of square versus round brackets.  and <input-arpa> and <output-fst> should be written in angle brackets.  (I know the code was old).

kkm000: Great catch. Done, and with an example.

danpovey: I think you forgot to commit.

src/lmbin/arpa2fst.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "util/kaldi-io.h"
+#include "util/parse-options.h"
+
+
+int main(int argc, char *argv[]) {
+  using namespace kaldi;
+  try {
+    const char *usage  =
+        "Convert an ARPA format language model into an FST\n"
+        "Usage: arpa2fst [opts] <input_arpa> <output_fst>\n"
```
danpovey: OK, I see you made the change.

danpovey:
OK, merging.  Thanks!

kkm000:
I'll make another PR fixing the use in scripts.

Also I think I should add an option to ilabel-soft the FST and default it to be `true`. Piping to `fstarcsort` makes little for parallelism, as the machine will not be sorted until fully read, but doubles the total RAM required. `fst::ArcSort` sorts in-place.

danpovey:
Makes sense.

PR #581, opened 2016-03-11
More accurate language folder check for phonetic langs

nshmyrev:
Current script fails with error due to wdisambig


jtrmal:
@danpovey LGTM

danpovey:
Sorry, I thought we had committed this already.   Merging.

PR #583, opened 2016-03-12
ami+chain: Added blstm and tdnn recipes. Updated tdnn results.

vijayaditya:


danpovey:
On this setup, LSTMs seem to do strangely well.

vijayaditya:
The TDNN might need more left context as this is reverberation task.

--VIjay

PR #584, opened 2016-03-13
sequence: Bug fixes

vimalmanohar:


src/nnet3/discriminative-supervision.cc
```diff
@@ -434,9 +434,10 @@ void AppendSupervision(const std::vector<const DiscriminativeSupervision*> &inpu
       output_supervision->resize(output_supervision->size() + 1);
       output_supervision->back() = src;
     }
-
-    if (compactify)
-      fst::TopSort(&output_supervision->back().den_lat);
+  }
+
```
danpovey: this code is still not quite right.
In general it may produce multiple outputs, the array size is not guaranteed 1 and they all may need to be top-sorted.
In principle you could iterate over them, but that also is inefficient because there is no point doing TopSort and Check() on outputs that had just one corresponding input (i.e. were not really appended).

PR #589, opened 2016-03-14
[WIP] 3D-CNN; The convolution component is further generalized to sup…

tomkocse:
…port  actual 3D convolution. The CNN input/output is a tensor with 3 axes (time, frequency and feature-map). In normal CNNs we do "3-D" convolution, but always ensure that the 3-D filter has length equal to input 3-D tensor along feature map axis. However we could perform actual 3-D convolution by using a 3-D filter which has length less than input along feature map axis.
The problem now is that the training become very slow when we are using 4 convolution layers.
From the log we saw that it spend a lot of time on AddMatMatBatched() :100: 
[cudevice profile]
AddMatVec       28.8452s
MulElements     39.1796s
MulRowsVec      52.0823s
AddMatBlocks    52.1977s
ApplyHeaviside  52.4034s
AddDiagVecMat   53.8776s
AddDiagMatMat   64.9836s
CuMatrix::SetZero       73.9582s
CuMatrixBase::CopyFromMat(from other CuMatrixBase)      77.5472s
AddMatMat       112.539s
CuMatrix::Resize        127.168s
AddCols 161.896s
AddVecToRows    203.794s
CopyCols        230.77s
AddMatMatBatched        999.96s
Total GPU time: 2424.14s (may involve some double-counting)

We would like to see if there is any way to speed up the training of the convolution component.

### DIFF
--- a/egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
+++ b/egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
@@ -58,6 +58,9 @@ if [ "$use_cnn" == "true" ]; then
   cnn_bottleneck_dim=  # remove this option if you don't want to add the bottleneck affine
   cepstral_lifter=  # have to fill this in if you are not using the default lifter value in the production of MFCC
                     # Here we assume your are using the default lifter value (22.0)
+
+  # When --filt-z-dim is not specified it will be equal to input-z-dim
+  # and when --filt-z-step is not specified it will be equal to --filt-z-dim
   cnn_layer="--filt-x-dim=4 --filt-y-dim=4 --num-filters=1024"
   cnn_opts+=(--cnn.layer "$cnn_layer")
   cnn_layer="--filt-x-dim=2 --filt-y-dim=4 --filt-z-dim=16 --num-filters=1"
@@ -75,7 +78,7 @@ if [ "$use_cnn" == "true" ]; then
   cmvn_opts="--norm-means=true --norm-vars=false"
 fi
 
-dir=exp/nnet3/tdnn_testminibatch
+dir=exp/nnet3/tdnn
 dir=$dir${affix:+_$affix}
 dir=${dir}$suffix
 train_set=train_nodup$suffix
--- a/egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
+++ b/egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
@@ -1,0 +1,149 @@
+#!/bin/bash
+
+# this is the standard "tdnn" system, built in nnet3; it's what we use to
+# call multi-splice.
+
+. cmd.sh
+
+
+# At this script level we don't support not running on GPU, as it would be painfully slow.
+# If you want to run without GPU you'd have to call train_tdnn.sh with --gpu false,
+# --num-threads 16 and --minibatch-size 128.
+
+stage=9
+affix=
+train_stage=-10
+has_fisher=true
+speed_perturb=true
+use_cnn=true
+cmvn_opts="--norm-means=false --norm-vars=false"
+common_egs_dir=
+reporting_email=
+remove_egs=true
+
+. cmd.sh
+. ./path.sh
+. ./utils/parse_options.sh
+
+
+if ! cuda-compiled; then
+  cat <<EOF && exit 1
+This script is intended to be used with GPUs but you have not compiled Kaldi with CUDA
+If you want to use GPUs (and have them), go to src/, and configure and make on a machine
+where "nvcc" is installed.
+EOF
+fi
+
+suffix=
+if [ "$speed_perturb" == "true" ]; then
+  suffix=_sp
+fi
+
+cnn_opts=()
+if [ "$use_cnn" == "true" ]; then
+# When the cnn option is on, the LDA is replaced by
+# an CNN layer at the front of the network
+# and the recipe uses fbank features as the CNN input
+# As the dimension of the CNN output is usually large, we place
+# a linear layer at the output of CNN for dimension reduction
+# The ivectors are processed through a fully connected affine layer,
+# then concatenated with the CNN bottleneck output and
+# passed to the deeper part of the network.
+# Due to the data compression issue, it is better to convert MFCC
+# to FBANK in the network instead of directly using FBANK features
+# from the storage. This script uses MFCC features as its input
+# and were converted to FBANK features with an inverse of DCT matrix
+# at the first layer of the network.
+
+  cnn_bottleneck_dim=  # remove this option if you don't want to add the bottleneck affine
+  cepstral_lifter=  # have to fill this in if you are not using the default lifter value in the production of MFCC
+                    # Here we assume your are using the default lifter value (22.0)
+  cnn_layer="--filt-x-dim=4 --filt-y-dim=4 --num-filters=1024"
+  cnn_opts+=(--cnn.layer "$cnn_layer")
+  cnn_layer="--filt-x-dim=2 --filt-y-dim=4 --filt-z-dim=16 --num-filters=1"
+  cnn_opts+=(--cnn.layer "$cnn_layer")
+
+  cnn_layer="--filt-x-dim=1 --filt-y-dim=4 --num-filters=1024"
+  cnn_opts+=(--cnn.layer "$cnn_layer")
+  cnn_layer="--filt-x-dim=1 --filt-y-dim=4 --filt-z-dim=32 --num-filters=1"
+  cnn_opts+=(--cnn.layer "$cnn_layer")
+
+  [ ! -z "$cnn_bottleneck_dim" ] && cnn_opts+=(--cnn.bottleneck-dim $cnn_bottleneck_dim)
+  [ ! -z "$cepstral_lifter" ] && cnn_opts+=(--cnn.cepstral-lifter $cepstral_lifter)
+  affix=3dcnn
+  # We found that using CNN with mean-normalized features obtains better results
+  cmvn_opts="--norm-means=true --norm-vars=false"
+fi
+
+dir=exp/nnet3/tdnn_testminibatch
+dir=$dir${affix:+_$affix}
+dir=${dir}$suffix
+train_set=train_nodup$suffix
+ali_dir=exp/tri4_ali_nodup$suffix
+
+local/nnet3/run_ivector_common.sh --stage $stage \
+	--speed-perturb $speed_perturb || exit 1;
+
+if [ $stage -le 9 ]; then
+  echo "$0: creating neural net configs";
+
+  # create the config files for nnet initialization
+  python steps/nnet3/tdnn/make_configs.py  \
+    "${cnn_opts[@]}" \
+    --feat-dir data/${train_set}_hires \
+    --ivector-dir exp/nnet3/ivectors_${train_set} \
+    --ali-dir $ali_dir \
+    --relu-dim 1024 \
+    --splice-indexes "-2,-1,0,1,2 -1,2 -3,3 -7,2 0"  \
+    --use-presoftmax-prior-scale true \
+   $dir/configs || exit 1;
+fi
+
+
+if [ $stage -le 10 ]; then
+  if [[ $(hostname -f) == *.clsp.jhu.edu ]] && [ ! -d $dir/egs/storage ]; then
+    utils/create_split_dir.pl \
+     /export/b0{3,4,5,6}/$USER/kaldi-data/egs/swbd-$(date +'%m_%d_%H_%M')/s5/$dir/egs/storage $dir/egs/storage
+  fi
+
+  steps/nnet3/train_dnn.py --stage=$train_stage \
+    --cmd="$decode_cmd" \
+    --feat.online-ivector-dir exp/nnet3/ivectors_${train_set} \
+    --feat.cmvn-opts="$cmvn_opts" \
+    --trainer.num-epochs 2 \
+    --trainer.optimization.minibatch-size 512 \
+    --trainer.optimization.num-jobs-initial 3 \
+    --trainer.optimization.num-jobs-final 16 \
+    --trainer.optimization.initial-effective-lrate 0.0017 \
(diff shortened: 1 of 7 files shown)


vijayaditya:
@tomkocse did you run the nnet3 unit tests to confirm that the ConvolutionComponent works fine ?

egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
```diff
@@ -0,0 +1,149 @@
...
+ali_dir=exp/tri4_ali_nodup$suffix
+
+local/nnet3/run_ivector_common.sh --stage $stage \
+	--speed-perturb $speed_perturb || exit 1;
+
+if [ $stage -le 9 ]; then
+  echo "$0: creating neural net configs";
+
+  # create the config files for nnet initialization
+  python steps/nnet3/tdnn/make_configs.py  \
```
vijayaditya: To simplify the code add two calls to make_configs.py one for normal TDNN and one for CNN. In this case you can directly provide the --cnn-layer options to the make_configs.py script, rather than storing them in another variable cnn_opts.

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -128,15 +142,19 @@ def AddConvolutionLayer(config_lines, name, input,
...
-                       "input-x-dim={input_x_dim} input-y-dim={input_y_dim} input-z-dim={input_z_dim} "
-                       "filt-x-dim={filt_x_dim} filt-y-dim={filt_y_dim} "
-                       "filt-x-step={filt_x_step} filt-y-step={filt_y_step} "
-                       "input-vectorization-order={vector_order}".format(name = name,
-                       input_x_dim = input_x_dim, input_y_dim = input_y_dim, input_z_dim = input_z_dim,
-                       filt_x_dim = filt_x_dim, filt_y_dim = filt_y_dim,
-                       filt_x_step = filt_x_step, filt_y_step = filt_y_step,
-                       vector_order = input_vectorization))
+    # if filt_z_dim is zero we assume it equals to the input
+    if filt_z_dim == 0:
```
vijayaditya: if filt_z_dim is same as input_z_dim it should be None, not 0.

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -147,12 +165,14 @@ def AddConvolutionLayer(config_lines, name, input,
...
     num_x_steps = (1 + (input_x_dim - filt_x_dim) / filt_x_step)
     num_y_steps = (1 + (input_y_dim - filt_y_dim) / filt_y_step)
-    output_dim = num_x_steps * num_y_steps * num_filters;
+    num_z_steps = (1 + (input_z_dim - filt_z_dim) / filt_z_step)
+    output_dim = num_x_steps * num_y_steps * num_z_steps * num_filters;
     return {'descriptor':  '{0}_conv_t'.format(name),
             'dimension': output_dim,
-            '3d-dim': [num_x_steps, num_y_steps, num_filters],
+            '3d-dim': [num_x_steps, num_y_steps, num_z_steps * num_filters],
             'vectorization': 'zyx'}
```
vijayaditya: make it clear that the output is actually a vectorized 4d-tensor but we ignore this.

egs/wsj/s5/steps/nnet3/tdnn/make_configs.py
```diff
@@ -348,8 +353,10 @@ def ParseCnnString(cnn_param_string_list):
 
     cnn_parser.add_argument("--filt-x-dim", required=True, type=int)
     cnn_parser.add_argument("--filt-y-dim", required=True, type=int)
+    cnn_parser.add_argument("--filt-z-dim", type=int, default = 0)
```
vijayaditya: default is None.

src/nnet3/nnet-simple-component.cc
```diff
@@ -3778,10 +3828,14 @@ void ConvolutionComponent::Write(std::ostream &os, bool binary) const {
   WriteBasicType(os, binary, filt_x_dim_);
   WriteToken(os, binary, "<FiltYDim>");
   WriteBasicType(os, binary, filt_y_dim_);
+  WriteToken(os, binary, "<FiltZDim>");
+  WriteBasicType(os, binary, filt_z_dim_);
   WriteToken(os, binary, "<FiltXStep>");
   WriteBasicType(os, binary, filt_x_step_);
   WriteToken(os, binary, "<FiltYStep>");
   WriteBasicType(os, binary, filt_y_step_);
+  WriteToken(os, binary, "<FiltZStep>");
```
vijayaditya: This breaks compatibility with CNNs created using old binaries. You should make it clear in the commit message that the CNN models are no longer accessible using this new code.

src/nnet3/nnet-simple-component.cc
```diff
@@ -3724,7 +3769,8 @@ void ConvolutionComponent::Update(const std::string &debug_info,
 
 void ConvolutionComponent::SetZero(bool treat_as_gradient) {
   if (treat_as_gradient) {
-    SetActualLearningRate(1.0);
+    learning_rate_ = 1.0;  // don't call SetLearningRate, that would apply the
```
vijayaditya: Its not clear why you want to do this ?

src/nnet3/nnet-simple-component.h
```diff
@@ -1498,6 +1499,8 @@ class ConvolutionComponent: public UpdatableComponent {
 
   int32 filt_y_dim_;    // size of the filter along y-axis
 
+  int32 filt_z_dim_;    // size of the filter along z-axis
+
   // there is no filt_z_dim_ as it is always assumed to be
```
vijayaditya: remove this comment, as it no longer valid.

vijayaditya:
You will have to add new unit test cases for the 3D convolution case. 

You will have to ensure that the convolution component works fine for all the old unit-test cases.

For the speed up: Could you check if cuDNN is efficiently handling the 3D convolution case ? If yes we can attempt to directly use that library.

galv:
Regarding the slowness of AddMatMatBatched, can you say how large your row and column dimensions are for each patch? That is, what is the size of each patch? The NVIDIA documentation says that you can expected a speedup with [cublasSgemmBatched](http://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemmbatched), which is being called under AddMatMatBatched for 100x100 matrices. (Dan and I also encountered a bottleneck with AddMatMatBatched while creating RepeatedAffineComponent.) Some mental calculations tell me that if you use a 3D convolution kernel of size 3x3x3, then if your number of rows is around 300, you may have slow down.

Regarding a solution, I believe some implementations construct a Toeplitz matrix or do an FFT for specific sizes; it probably requires some tuning. cudnn is probably a reasonable bet. I'm actually personally interested in convolution implementation as I'm interested in implementing complex convolution. Let me know if I can help you out with something.

vijayaditya:
@galv What are the 100x100 matrices you are referring to, is this the size of each individual matrix in the AddMatMatBatched call ? If yes is this restriction on the size of both the left and right operands or just on result matrix ?

Typical input in this case is around 40x4x4096 and the kernel is 4x4x64. Along the third axis there is no overlap in the hops, so the number of patches is considerably small (37 * 1 * 64 = 2368). There will be around ~2400 matrix multiplications in the AddMatMatBatched call.

Another reason for the slow down here is that the size of the convolution inputs/outputs is really large. e.g. these matrices take up around 0.5GB of the device memory in the architecture being used here. We are forced to reduce the minibatch size / frames-per-iter to tackle this issue. @tomkocse says he had to reduce the minibatch size by 4 in the xent training setup. He tried using composite components, but reports that there was no appreciable memory demand reduction.

If you would like to help in speeding up the convolution code let me know. One of the first things to test out would be checking cuDNN's convolution implementation and checking if it is fast enough for our use.
We would also need to ensure that both the memory and computational complexity of the desired implementation is small.

egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
```diff
@@ -0,0 +1,149 @@
...
+# Due to the data compression issue, it is better to convert MFCC
+# to FBANK in the network instead of directly using FBANK features
+# from the storage. This script uses MFCC features as its input
+# and were converted to FBANK features with an inverse of DCT matrix
+# at the first layer of the network.
+
+  cnn_bottleneck_dim=  # remove this option if you don't want to add the bottleneck affine
+  cepstral_lifter=  # have to fill this in if you are not using the default lifter value in the production of MFCC
+                    # Here we assume your are using the default lifter value (22.0)
+  cnn_layer="--filt-x-dim=4 --filt-y-dim=4 --num-filters=1024"
```
vijayaditya: @tomkocse the filt-z-step should be equal to filt-z-dim in this experiment. If you do this the size of the matrices would be substantially reduced. Please test this with high priority.

tomkocse: @vijayaditya when the  filt-z-step is not specified it will be by default equal to filt-z-dim, so we are doing non-overlapping convolution on the z axis, i will add a comment to clarify this .

tomkocse:
@vijayaditya Yes i have run the unit test, sorry that i forget to include it in the commit, i will upload it now

egs/swbd/s5c/local/nnet3/run_tdnn_3dcnn.sh
```diff
@@ -0,0 +1,149 @@
...
+  cnn_opts+=(--cnn.layer "$cnn_layer")
+
+  [ ! -z "$cnn_bottleneck_dim" ] && cnn_opts+=(--cnn.bottleneck-dim $cnn_bottleneck_dim)
+  [ ! -z "$cepstral_lifter" ] && cnn_opts+=(--cnn.cepstral-lifter $cepstral_lifter)
+  affix=3dcnn
+  # We found that using CNN with mean-normalized features obtains better results
+  cmvn_opts="--norm-means=true --norm-vars=false"
+fi
+
+dir=exp/nnet3/tdnn_testminibatch
```
vijayaditya: use a more appropriate name, e.g. exp/nnet3/cnn3d_tdnn

galv:
@vijayaditya 

To your first question: yes. This is with respect to the size of each individual matrix. I am not sure about the second part, since the documentation doesn't say much, but I suspect the restriction is on both the input (operand) matrices and the output matrices. I wish I could give a good justification for this, but based on skimming the CUDA Handbook, I believe the overall point is that AddMatMatBatched provides a speed up only when your matrices are small.

I agree using/investigating cuDNN is the best first thing to do. Its convolution routines are supposed to be very well-tuned; apparently it can even choose what it does based on benchmarking on your device! I also believe there is an option to limit memory usage. I think the only other alternative has something to do with forming a Toeplitz matrix to represent the convolution. We will have to see.

I actually need to implement complex-valued convolution sometime soon for a class. So I don't mind learning more about this to try to speed things up.

But cuDNN requires registration to use. You do realize that this will make Kaldi less self-contained, I imagine.

vijayaditya:
There are ways to circumvent the registration issue.

By forming patches and using matrix matrix multiply as in [1] we are
basically doing the Toeplitz matrix style multiplication without actually
forming the Toeplitz matrix, which is a severe waste of memory. However
when I tested the speed of [1] with AddMatMatBatched, the latter was
faster. This is due to the increased number of memory copies in [1].
Further it requires more temporary matrices, which is definitely not
suitable in our current scenario.
([1]:Kumar Chellapilla, Sidd Puri, Patrice Simard. High Performance
Convolutional Neural Networks for Document Processing. Guy Lorette. *Tenth
International Workshop on Frontiers in Handwriting Recognition*, Oct 2006,
La Baule (France), Suvisoft, 2006.)

 FFT based multiplication is the next best thing but the cost of FFT is not
usually justified for speech data due to the dimensions of the data we use.
However FFT based convolution would definitely make sense for us now as we
are using filters in the range of 4096. But even in this case  people have
found implementation of custom FFT implementations to be faster than cuFFT.
See
Mathieu, Michael, Mikael Henaff, and Yann LeCun. "Fast training of
convolutional networks through FFTs." *arXiv preprint arXiv:1312.5851*
 (2013).

For speech specific implementations look at this paper in Interspeech. It
should be available somewhere on the CLSP cluster.

*Efficient GPU Implementation of Convolutional Neural Networks for Speech
Recognition*
 Ewout van den Berg 1, Daniel Brand 2, Rajesh Bordawekar 2, Leonid
Rachevsky 1, Bhuvana Ramabhadran 1; 1IBM Watson, USA; 2IBM T.J. Watson
Research Center, USA

Deep learning has enjoyed tremendous success in speech recognition in
recent years. Despite their widespread use, training large and complex
architectures remains very time consuming. A prime example of this are
convolutional neural networks (CNNs), which have provided state-of-the-art
results, but are also among the most computationally intensive networks to
train. In this paper, we study four different methods for GPU acceleration
of CNNs: a native implementation using cuBLAS, two implementations based on
NVIDIA’s recently released deep-learning cuDNN library, and an
implementation based on cuFFT. We analyze the performance of each of these
approaches on the forward operation, the gradient computation, and the
backward propagation. The overall best performance is obtained using the
custom native implementation, which was found to be up to *6.9 times faster
than cuDNN.* The paper concludes with results on the end-to-end training
speed of our CNN network on an LVCSR task.

--Vijay

galv:
I wasn't aware that it might not be as simple as dropping in cudnn and tuning the options. I'm still willing to work on investigating different implementations, though. How urgent is this?

vijayaditya:
This is a high-priority project. The CNN speed is a major bottleneck for
Tomko's experiments. I hope he will find time to go through the cuDNN
implementations.

Other implementations discussed are not immediately necessary. They are
more of a research project and the amount of effort we would want to spend
on CNN implementation depends on the results of Tomko's experiments.

--Vijay

galv:
Okay. cudnn is not something which should be assigned to me then (at least not in full) because I know at the very least I cannot get to it for t least the next 5 days. I'm still interested in alternative implementations of convolution personally because, as mentioned, I'm interested in implementing complex convolution for a project.

vijayaditya:
This PR is no longer relevant as we are completely rewriting the ConvolutionComponent. (See #604)

PR #592, opened 2016-03-15
sequence: caching for discriminative training + some bug fixes

vimalmanohar:


src/nnet3/nnet-discriminative-training.cc
```diff
@@ -123,16 +133,16 @@ void NnetDiscriminativeTrainer::ProcessOutputs(const NnetDiscriminativeExample &
...
       objf_info_[sup.name].stats.Reset();
     }
-
+    
     ComputeDiscriminativeObjfAndDeriv(opts_.discriminative_config, 
                                       tmodel_, log_priors_,
                                       sup.supervision, nnet_output,
-                                      &(objf_info_[sup.name].stats), 
+                                      &stats, 
                                       &nnet_output_deriv,
```
danpovey: vimal, it looks to me like with this change you are just throwing away the stats.

vimalmanohar: There's an UpdateStats down below (line 178) to update the objf_info_ object using the newly accumulated stats in the object "stats". This is what was originally intended.

PR #593, opened 2016-03-15
Relaxing the constraints on the written form of phones.

francisr:
The first constraint is not to have conflicts with disambiguation symbols.
The second one is because prepare_lang.sh could think that position dependent phones are used when they are not.


danpovey:
Thanks!
Merging.

PR #599, opened 2016-03-18
fixes #476

### LINKED ISSUE #476: Better error status handling for tables of archives
@xiaohui-zhang found an instance where a pipe of features read as a SequentialBaseFloatMatrixReader failed partway through due to an nfs error, leading to nonzero exit status of the program.  However, the program did not die because the error was not propagated.  It looks like we always ignore the exit-status of programs in Input [the Input::Close() function returns no value].  
The TODO is:
  - Let Input::Close() return the exit code as int32, in the case of a pipe [kPipeInput], and always zero otherwise.  Obviously document this.
  - Let the Close() function and destructor of class SequentialTableReaderArchiveImpl to check for an additional error condition:  If the state is kEof (meaning we reached the end of the archive) then it should call input_.Close() and check the status.  If the status is nonzero then if the the permissive (,p) option was used a warning should be printed; otherwise it should invoke KALDI_ERR.

@xiaohui-zhang, perhaps you could work on this?  This will need to be extensively tested prior to committing, so you'd have to come up with a satisfactory pull request and then use it personally for a while.

Note: to catch this type of error in all circumstances we might sometimes have to set -o pipefail.  But this will depend on the script.  We don't have to worry about this just yet.

xiaohui-zhang:


danpovey:
This looks good.  However, I'd like you to keep it in your personal repo for a little while to test it.  Please make sure it doesn't get deleted.

xiaohui-zhang:
sure!

On Mar 18, 2016 12:44 AM, "Daniel Povey" <notifications@github.com> wrote:
>
> This looks good. However, I'd like you to keep it in your personal repo
for a little while to test it. Please make sure it doesn't get deleted.

xiaohui-zhang:
@danpovey  The reading partial matrices thing is committed also.

danpovey:
Samuel, there is a failing check
https://travis-ci.org/kaldi-asr/kaldi/builds/117877799
and it points to this:

In file included from gmm-decode-faster-regtree-fmllr.cc:29:0:
../util/kaldi-table-inl.h: In instantiation of ‘const T&
kaldi::RandomAccessTableReaderScriptImpl<Holder>::Value(const string&)
[with Holder = kaldi::KaldiObjectHolder<kaldi::RegtreeFmllrDiagGmm>;
kaldi::RandomAccessTableReaderScriptImpl<Holder>::T =
kaldi::RegtreeFmllrDiagGmm; std::string = std::basic_string<char>]’:
gmm-decode-faster-regtree-fmllr.cc:288:1:   required from here
../transform/regtree-fmllr-diag-gmm.h:112:8: error: ‘void
kaldi::RegtreeFmllrDiagGmm::operator=(const kaldi::RegtreeFmllrDiagGmm&)’
is private
   void operator = (const RegtreeFmllrDiagGmm&);  // Disallow assignment
operator
        ^
In file included from ../util/kaldi-table.h:467:0,
                 from ../util/common-utils.h:27,
                 from gmm-decode-faster-regtree-fmllr.cc:25:
../util/kaldi-table-inl.h:1641:15: error: within this context
       object_ = holder_.Value();
               ^


... you shouldn't assume the assignment operator is defined; and you
shouldn't be doing a redundant copy in cases where the range operator is
not being used.  That's wasteful of both CPU and memory.
Dan

xiaohui-zhang:
Thanks @danpovey . fixed

xiaohui-zhang:
closing it as I intended to commit to chain

PR #602, opened 2016-03-19
WIP: Do not merge !!! : No subsampling config added

vijayaditya:
This PR has been created just for a review. This is not for merging.

@danpovey any other changes needed for no sub-sampling experiment ?


danpovey:
It looks reasonable.
But on second thoughts, you should probably run two versions-- that one, and one in which you modify the topology to assign different pdf-ids to the self-loop and the forward transition.  I found this to be better (IIRC) in the one-stat setting.
See ~dpovey/gen_topo6.py
Dan

vijayaditya:
OK will do.

Vijay

danpovey:
BTW, don't use b06 or b06 for striping, they are too busy- change the list.

PR #604, opened 2016-03-21
WIP: integration of cuDNN

freewym:
@tomkocse I am now reading the conv code in CNTK for the integration.


tools/extras/install_cudnn.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# https://developer.nvidia.com/rdp/cudnn-download
+
+if [ ! -f srilm.tgz ]; then
```
jtrmal: -f cudnn.tgz

jtrmal: Also, I think we should use the original name (cudnn-7.0-linux-x64-v4.0-prod.tgz).
The reason is that we are going to use it as a library (I assume), so the API version will be important.

freewym:
Thanks. Fixed.

tomkocse:
@freewym, thank you very much, cntk just provide example for 2d convolution component but what we need is 3d or 4d convolution. CUDNN provides N-dimension convolution routine, could you please use that in the integration ?

freewym:
@tomkocse Yes I use CNTK as a reference and want to get an idea of how it should look like.

vijayaditya:
@jtrmal Do you have the time to review this PR over its lifetime ?

jtrmal:
yeah, I'll try to keep an eye on it. Please bump if I won't be reacting.
y.

src/cudamatrix/cu-device.h
```diff
@@ -26,6 +26,7 @@
 #if HAVE_CUDA == 1
 
 #include <cublas_v2.h>
+#include <cudnn.h>
```
galv: You should wrap this, and I think all of the code that uses the cudnn context in cu-device.cc in this:
```
#ifdef HAVE_CUDNN
code goes here...
#endif
```

You could define the HAVE_CUDNN macro somewhere in src/configure by adding a line like "CXX_FLAGS += -DHAVE_CUDNN" to kaldi.mk

The point is that I don't think we can reasonably assume most users of Kaldi will want or need cudnn. So we have to have these ugly macros to support compiling both with and without cudnn.

danpovey: If the installation is easy (e.g. if they let us distribute it), we might
just make the decision that whenever we support CUDA, we also require CUDNN
support.  That would reduce the number of separate cases of compilation
that have to be tested.  At this point I think we haven't made the decision
to go with CUDNN at all, this is somewhat experimental.

Dan

galv: Installation is not easy. You need to register as a developer on NVIDIA's website, and you cannot redistribute the binary: "NVIDIA hereby grants Licensee the following limited, non-exclusive, **non-transferable**, non-sublicensable right to use the SOFTWARE"

danpovey: They told me privately that they would consider amending the license to
allow us to redistribute it.

src/cudamatrix/cu-device.cc
```diff
@@ -216,7 +216,10 @@ void CuDevice::FinalizeActiveGpu() {
     active_gpu_id_ = act_gpu_id; // CuDevice::Enabled() is true from now on
     // Initialize the CUBLAS
     CU_SAFE_CALL(cublasCreate(&handle_));
-
+  #if HAVE_CUNN == 1
```
galv: `HAVE_CUDNN`, not `HAVE_CUNN`

src/cudamatrix/cu-device.cc
```diff
@@ -216,7 +216,10 @@ void CuDevice::FinalizeActiveGpu() {
...
+    // Initialize CUDNN
+    CU_SAFE_CALL(cudnnCreate(&cudnn_));
```
galv: You'll need to place a macro like this in cu-common.h

#define CUDNN_SAFE_CALL(fun) \
{ \
  cudnnStatus_t ret; \
  if ((ret = (fun)) != CUDNN_STATUS_SUCCESS) { \
    KALDI_ERR << "cudnnStatus_t " << ret << " : \"" << cuddnnGetErrorString(ret) << "\" returned from '\
" << #fun << "'"; \
  } \
  cudaDeviceSynchronize(); \
}

and use it instead of CU_SAFE_CALL. The reason why is that cudnn's functions return a totally different type  (cudnnStatus_t) than CUDA itself (cudaError_t) to report errors. You can see the definition of cudaError_t here: http://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gf599e5b8b829ce7db0f5216928f6ecb6

Meanwhile, the definition of cudaError_t is totally different. You can see that in cudnn.h. We need this so users get accurate error messages.

src/cudamatrix/cudnn-convolution-test.cc
```diff
@@ -0,0 +1,120 @@
...
+  size_t nbDimsX = 5;
+  MatrixIndexT dimX[] = {100, 20, 300, 200, 100};
+  MatrixIndexT sizeX = 1;
+  for (int32 i = 0; i < nbDimsX; i++) {
+    sizeX *= dimX[i];
+  } 
+  MatrixIndexT strideX[] = {20 * 300 * 200 * 100, 300 * 200 * 100, 200 * 100, 100, 1};
+  cudnnTensorDescriptor_t xDesc;
+  cudnnConv.InitializeTensorDescriptor(nbDimsX, dimX, strideX, xDesc);
+  CuMatrix<Real> x(1, sizeX);
```
galv: I see what you're trying to do here. You're avoiding creating a matrix to represent the tensor since a matrix only has two dimensions anyway, and it uses cudaMallocPitched, which may or may not make the stride equal to the number of columns. Dan recently made a change so that you can ensure that stride equals the number of columns, via this constructor in cu-matrix.h:

```
  /// Constructor with memory initialisation                                                             
  CuMatrix(MatrixIndexT rows, MatrixIndexT cols,
           MatrixResizeType resize_type = kSetZero,
           MatrixStrideType stride_type = kDefaultStride) {
    Resize(rows, cols, resize_type, stride_type);
  }
```

where you set stride_type = kStrideEqualNumCols.

Since the nnet3 component interface uses matrices where the row dimensions represents the batch, you will want to depend on this constructor having been used to create the input and output batch matrices. We already do in RepeatedAffineComponent.

freewym: Thanks. Will change as you suggested.

Yiming

galv: I don't know if it's a necessary change. I think cudnn also accepts tensors where strides do not equal the number of elements in a dimension. I just wanted to let you know.

galv:
What I'm wondering about now is why you choose to create the CuDnnConvolution class. It looks like it abstracts away error handling and timing, but that's all. It's essentially a namespace. I get the impression that a lot of these functions could go directly into an nnet3 component directly or as private functions. I'm just curious what you may have intended with this separate class.

danpovey:
If it acts as a namespace, better to have a namespace IMO

galv:
I actually agree with this now, especially since I would like to use cudnn's convolution code to do complex valued convolution (which essentially involves four convolution calls on real matrices). This class prevents all the code I would like to reuse from being private in an nnet3 (real-valued) convolution component.

freewym:
I was thinking of providing an API similar to CuMatrix for nnet3 to use, and hide the error handling, etc. And also use template mechanism to  converting, for example, CUDNN_DATA_DOUBLE to double for the user, so that they can just use template class instantiation as they use CuMatrix.

galv:
I forgot about templating. This sounds good.

I'm done reading for the night. I might continue to look at things tomorrow, but don't wait on me if you're going to work more on this.

I think an important step will be adding unit tests that allow us to tune for setups we will be using, since CUDNN provides multiple implementations based on allowed workspace size, and so on. I would love to say I could do this, but I can't promise that yet.

galv:
Your current test case does not pass because it is running out of memory. On my GEForce 780 Ti, only about 500 MB of the 3000 MB total are free before the call `cudnnConv.FindBestConvolutionFwdAlgo`. This is not enough, which is why you get CUDNN_STATUS_ALLOC_FAILED returned from that call. If you don't declare any of the CuMatrix<> objects, you will have enough space for the call to succeed, though.

galv:
I began making a rough implementation of this today, since Yiming has been busy. There are a few troublesome details (e.g., CUDNN uses opaque types  that we cannot serialize directly), but I do not think it will be a large problem. We will need to test speeds. While 3D convolution can implement 2D convolution, I am not sure whether the 3D implementation will work well in the 2D case. I intend to speak to Yiming in person tomorrow.

I have two questions:

1. Currently  ConvolutionComponent supports yzx (read the docstring above ConvolutionComponenet in nnet-simple-component.h to understand what this means) vectorization of inputs. Do we need to maintain this support? We could presumably pipe the input in through another binary to transpose the dimensions to xyz vectorization instead.
2. @tomkocse, did you ever diagnose how much memory was being allocated on the GPU? I can find this myself, but I would first have to setup a switchboard directory.

vijayaditya:
Could you please provide an update on the status of this effort ?

freewym:
I think Daniel is working on implementing ConvolutionalComponent based on the API provided in this PR.

Yiming

galv:
Yes, I was working on that. Frankly, my code I have is long overdude for a pull request. There are important things that must be discussed anyhow.

I will try my best to get a pull request in tonight; there's an annoying piece of code I need to write first since C is a painful language. I will include important questions I need answers for in that pull request.

vijayaditya:
If you are busy with some thing more urgent this can wait. I just wanted to
ensure that we don't forget about this.

galv:
Well, whether or not I get to making the PR tonight, I will say that gradient checks do pass.

galv:
I spoke too soon. Looks like NaNs are cropping up.

galv:
I've fixed up an invalid memory access. I'll try to get a PR up tomorrow.

vijayaditya:
@freewym @galv @tomkocse Could you please give an update on this PR. What is the expected time of completion ?

galv:
The next step is me opening a new PR. I'll try my darn hardest to get it opened up today. This has been stewing too long. All I need to do is integrating timing code into the code I have now.

freewym:
work continued in #770

PR #606, opened 2016-03-21
chain/get_egs.h : minor changes to ensure that valid/train subsets ha…

vijayaditya:
…ve utts with length at least frames_per_eg


danpovey:
I see you closed it, but I'm not crazy about this because it's very inefficient.
Getting the durations- even once- is slow, but you are doing it twice.
I'd rather have you rely on utils/data/get_utt2dur.sh (which allows the durations to be cached), and utils/data/get_frame_shift.sh.

PR #607, opened 2016-03-21
chain/get_egs.h : minor changes to ensure that valid/train subsets ha…

vijayaditya:
…ve utts with length at least frames_per_eg


danpovey:
As before- this is quite inefficient.  Better to use  utils/data/get_frame_shift.sh and utils/data/get_utt2dur.sh, which in most cases will get the durations from the segments file (and if not, at any rate the durations will be cached and reused).
Dan

vijayaditya:
(both the PRs are exactly the same. I was unable to decide whether to merge to master or chain branches which led to this confusion.)

OK, I will use utils/data/get_utt2dur.sh

vijayaditya:
@danpovey done

danpovey:
Seems OK, but have you tested it?  We could just wait till you need to run something else and then merge it.

vijayaditya:
I tested it.

PR #610, opened 2016-03-23
bunch of KALDI_DOUBLEPRECISION=1 fixes

jtrmal:


src/chain/Makefile
```diff
@@ -40,8 +40,7 @@ ifeq ($(CUDA), true)
   #For toolkit older than 6.5, add the compute capability 1.0
   CUDA_VER_GT_6_5 := $(shell [ $(CUDA_VERSION) -ge 65 ] && echo true)
   ifneq ($(CUDA_VER_GT_6_5), true)
-    CUDA_ARCH += -gencode arch=compute_13,code=sm_13 \
-                 -gencode arch=compute_10,code=sm_10
+    CUDA_ARCH += -gencode arch=compute_13,code=sm_13
```
jtrmal: THe compXchg and compCAS are defined since compute 1.1 or 1.2 (not sure now), but definitely not 1.0

src/thread/Makefile
```diff
@@ -8,7 +8,7 @@ TESTFILES = kaldi-thread-test kaldi-task-sequence-test
 OBJFILES =  kaldi-thread.o kaldi-mutex.o kaldi-semaphore.o kaldi-barrier.o
 
 LIBNAME = kaldi-thread
-ADDLIBS = ../matrix/kaldi-matrix.a ../base/kaldi-base.a
+ADDLIBS = ../base/kaldi-base.a
```
jtrmal: It does not seem to need the kaldi-matrix (and was confusing make -j8).
make depend would probably fix the confusion, but this seemed better solution

src/makefiles/cygwin.mk
```diff
@@ -4,7 +4,8 @@ ifndef FSTROOT
 $(error FSTROOT not defined.)
 endif
 
-CXXFLAGS = -msse -msse2 -Wall -I.. -DKALDI_DOUBLEPRECISION=0  \
+DOUBLE_PRECISION = 0
```
jtrmal: I propose to do it this way, as the variable has to be propagated into CUDA_FLAGS, so that the variable KALDI_DOUBLEPRECISION is set correctly during compilation of kernels (in chain/)

danpovey:
looks good to me- merge yourself when checks done.

jtrmal:
Thanks -- merging as all tests finished OK both on floats and doubles.
y.

PR #611, opened 2016-03-23
chain: fixes #476 and enabled reading partial matrices with TableReader.

### LINKED ISSUE #476: Better error status handling for tables of archives

xiaohui-zhang:


src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,66 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "util/kaldi-holder.h"
+#include "matrix/kaldi-matrix.h"
+
+namespace kaldi {
+
+template bool ExtractObjectRange(const Matrix<double> &, const std::string &, Matrix<double> *);
```
danpovey: better if instantiation comes after definition.  Watch 80-char limit.

src/util/kaldi-holder.h
```diff
@@ -205,6 +206,24 @@ class HtkMatrixHolder;
...
+/// The generic version of this function just fails; we overload the template
+/// whenever we need it for a specific class.
+template <class T>
+bool ExtractObjectRange(const T &input, const std::string &range, T *output) {
+  KALDI_ERR << "Ranges not supported for objects of this type.";
+  return false;  
+}
+
+template <class Real>
+bool ExtractObjectRange(const Matrix<Real> &input, const std::string &range, Matrix<Real> *output);
```
danpovey: watch 80 char limit

src/util/kaldi-table-inl.h
```diff
@@ -302,6 +325,24 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
       if (!key_.empty() && !data_rxfilename_.empty()) {
         // Got a valid line.
         state_ = kHaveScpLine;
+        // Here we "peel off" the range_ specifer (if any, e.g. [1:2,2:10])
+        // from data_rxfilename_ (e.g. 1.ark:100).
+        if (data_rxfilename_.find_last_of("]") == data_rxfilename_.length() - 1) {
```
danpovey: if (data_rxfilename_.back() == ']') 
 is easier

src/util/kaldi-table-inl.h
```diff
@@ -302,6 +325,24 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+          std::vector<std::string> splits;
+          SplitStringToVector(data_rxfilename_, "[]", true, &splits);
```
danpovey: Better to spend a little more effort to parse it exactly- this will accept too many wrong things.  If the code gets too long you can declare a function for this.

src/util/kaldi-table-inl.h
xiaohui-zhang: hmm I tried this and it needs c++11..

src/util/kaldi-table-inl.h
```diff
@@ -302,6 +325,24 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
       if (!key_.empty() && !data_rxfilename_.empty()) {
```
danpovey: The use of data_rxfilename_ to store the [rxfilename+range] at one point and [rxfilename] at another is rather inelegant.  I think you should use a temporary local variable to store the rest-of-line,  determine whether or not it contains a range, and only then set the class member variables.
And I don't think the way you store the previous data_rxfilename_ is very good either.  The only time you care about this is to test if it's the same as the current one.  What would be better I think is to split the state up further: have states
kHaveScpLine, kObjectLoaded, kObjectLoadFailed, and then to only be used when you have a range: states kRangeExtracted, kRangeExtractionFailed.
Then inside NextScpLine, once you work out what the next data_rxfilename_ would be, if the previous state was one of kObjectLoaded or kRangeExtracted or kRangeExtractionFailed, you know you have the object, so if the rxfilename is the same you can just set the state to kObjectLoaded.
This is more in the spirit of the original design.
It may (or may not) make sense to have a separate function to extract the range.

src/util/kaldi-table-inl.h
danpovey: OK then you can do

data_rxfilename_[data_rxfilename_.size()-1]

src/util/kaldi-table-inl.h
```diff
@@ -316,17 +357,19 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
     }
   }
 
-
   Input script_input_;  // Input object for the .scp file
   Input data_input_;   // Input object for the entries in
   // the script file.
   Holder holder_;  // Holds the object.
+  T object_; // the object extracted from holder_.
```
danpovey: clarify that this is only used when the object range is specified.

src/util/kaldi-table-inl.h
```diff
@@ -1487,7 +1526,8 @@ class RandomAccessTableReaderScriptImpl:
  public:
   typedef typename Holder::T T;
 
-  RandomAccessTableReaderScriptImpl(): last_found_(0), state_(kUninitialized) {}
+  RandomAccessTableReaderScriptImpl(): range_(""),
+                                       data_rxfilename_last_(""), last_found_(0), state_(kUninitialized) {}
```
danpovey: watch 80-char limit

danpovey: you can move range_ to the next line to reduce the indent.

src/util/kaldi-table-inl.h
```diff
@@ -1487,7 +1526,8 @@ class RandomAccessTableReaderScriptImpl:
  public:
   typedef typename Holder::T T;
 
-  RandomAccessTableReaderScriptImpl(): last_found_(0), state_(kUninitialized) {}
+  RandomAccessTableReaderScriptImpl(): range_(""),
```
danpovey: no need to initialize strings to "", it happens automatically.

src/util/kaldi-table-inl.h
```diff
@@ -1549,6 +1589,8 @@ class RandomAccessTableReaderScriptImpl:
     last_found_ = 0;
```
danpovey: I can't see that last_found_ is ever used.

xiaohui-zhang: it's used in LookUpKey

xiaohui-zhang:
updated

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+  }
+  if (row_range.size() == 0) {
+    row_range.push_back(0);
+    row_range.push_back(input.NumRows() - 1);
+  }
+  if (col_range.size() == 0) {
+    col_range.push_back(0);
+    col_range.push_back(input.NumCols() - 1);
+  }
+  if (!(row_range.size() ==2 && col_range.size() == 2 &&
```
danpovey: space after ==

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+  if (!(row_range.size() ==2 && col_range.size() == 2 && 
+        row_range[0] >= 0 && row_range[0] < row_range[1] &&
+        row_range[1] < input.NumRows() && col_range[0] >=0 &&
+        col_range[0] < col_range[1] && col_range[1] < input.NumCols())) {
+    KALDI_ERR << "Invalid range specifier: " << range;
+    return false;
+  }
+  int32 row_offset = row_range[1] - row_range[0] + 1,
+        col_offset = col_range[1] - col_range[0] + 1;
+  output->Resize(row_offset, col_offset);
```
danpovey: add kUndefined.

src/util/kaldi-table-inl.h
```diff
@@ -248,49 +271,66 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
-    // If you don't want this exception to be thrown you can
-    // call Close() and check the status.
-    if (state_ == kLoadSucceeded)
-      holder_.Clear();
+    if (this->IsOpen() && !Close())
+      KALDI_ERR << "TableReader: reading script file failed: from scp "
+                      << PrintableRxfilename(script_rxfilename_);
   }
  private:
   bool LoadCurrent() {
```
danpovey: document the possible states at the end of this function, and the associated return statuses.

src/util/kaldi-table-inl.h
```diff
@@ -248,49 +271,66 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+        } else {  // holder_ will not contain data.
+          KALDI_WARN << "Failed to load object from "
+                     << PrintableRxfilename(data_rxfilename_);
+          state_ = kLoadFailed;
+          return false;
+        }
       }
     }
+    
+    // The needed object has been loaded last time. Here we just
```
danpovey: remove "last time", it's inaccurate.

src/util/kaldi-table-inl.h
```diff
@@ -298,10 +338,38 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
     }
     std::string line;
     if (getline(script_input_.Stream(), line)) {
-      SplitStringOnFirstSpace(line, &key_, &data_rxfilename_);
-      if (!key_.empty() && !data_rxfilename_.empty()) {
+      std::string data_rxfilename_next, rest;
+      // rest: a temp string which may contain the range specifier.
```
danpovey: this description of 'rest' is not very clear- you need to back up and describe what it is at the top level before getting into details.  Or just remove the comment.

src/util/kaldi-table-inl.h
```diff
@@ -298,10 +338,38 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+      SplitStringOnFirstSpace(line, &key_, &rest);
+      if (!key_.empty() && !rest.empty()) {
```
danpovey: This whole function is very hard to read now that you've added all this parsing code.  It would be better to break it out into a separate function.

src/util/kaldi-table-inl.h
```diff
@@ -298,10 +338,38 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+          if (data_rxfilename_ == data_rxfilename_next) {
+            state_ = kObjectLoaded;
+          } else {
+            holder_.Clear();
+          }
+        }
+        data_rxfilename_ = data_rxfilename_next;
       } else {
         // Got an invalid line.
         state_ = kError;  // we can't make sense of this
```
danpovey: Please make sure that if we go to the state kError, at some point an informative error message is printed.  It's not clear that that was the case in the previous code.

src/util/kaldi-table-inl.h
```diff
@@ -248,49 +271,66 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+          } // o.w. we'll go ahead to extract the partial object.
+        } else {  // holder_ will not contain data.
+          KALDI_WARN << "Failed to load object from "
+                     << PrintableRxfilename(data_rxfilename_);
+          state_ = kLoadFailed;
+          return false;
+        }
       }
     }
+
```
danpovey: I don't think you are correctly handling the case where the state is kObjectLoaded but there is no range specifier (e.g. if someone uses the same object twice in succession with no range).

src/util/kaldi-table-inl.h
danpovey: you still need to clarify this..

src/util/kaldi-table-inl.h
```diff
@@ -340,6 +410,12 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
     // or the user called FreeCurrent().. note,
     // if when called by user we are in this state,
     // it means the user called FreeCurrent().
+    kObjectLoaded, // Called NextScpLine() and found the next    yes         yes
+    // object we need was alreay loaded in the last line.
```
danpovey: I think you are not accurately describing kObjectLoaded.  It implies that we have an object in the holder, corresponding to data_rxfilename_ (IIRC), but no extracted range.  This doesn't necessarily mean the object was from the last line, it could be that we just called LoadObject() and no range was specified.

src/util/kaldi-table-inl.h
```diff
@@ -1576,17 +1653,26 @@ class RandomAccessTableReaderScriptImpl:
         KALDI_ERR << "Could not get item for key " << key
                   << ", rspecifier is " << rspecifier_ << "[to ignore this, "
                   << "add the p, (permissive) option to the rspecifier.";
+      
+      KALDI_ASSERT(state_ == kHaveObject);
```
danpovey: looks like you assert the same thing twice.

src/util/kaldi-table-inl.h
```diff
@@ -1549,6 +1624,8 @@ class RandomAccessTableReaderScriptImpl:
     last_found_ = 0;
     script_.clear();
     current_key_ = "";
+    range_ = "";
+    data_rxfilename_last_ = "";
```
danpovey: You shouldn't use these initializers-- I think they are a C++11 extension.  Anyway strings are automatically empty.
And do you need data_rxfilename_last_?  If so please explain why.

src/util/kaldi-table-inl.h
```diff
@@ -1576,17 +1653,26 @@ class RandomAccessTableReaderScriptImpl:
         KALDI_ERR << "Could not get item for key " << key
                   << ", rspecifier is " << rspecifier_ << "[to ignore this, "
                   << "add the p, (permissive) option to the rspecifier.";
+      
+      KALDI_ASSERT(state_ == kHaveObject);
       KALDI_ASSERT(state_ == kHaveObject && key == current_key_);
     }
-
     if (state_ == kHaveObject) {
       state_ = kGaveObject;
```
danpovey: It looks like there are still quite a few issues with this PR.  Can you please triple-check it next time?  Go through every line of your diff.

src/util/kaldi-holder.cc
xiaohui-zhang: what do you mean?

danpovey: I mean add kUndefined as the last arg to Resize.  It avoids zeroing
unnecessarily.

xiaohui-zhang: got it thanks

src/util/kaldi-table-inl.h
xiaohui-zhang: it's handled in ExtracObjectRange (returning holder_.Value() if range_.empty()).

danpovey: But that requires an unnecessary copy.  Objects can be large.

src/util/kaldi-table-inl.h
xiaohui-zhang: Did you mean LoadCurrent()? there is no LoadObject() here. Yep "last" is not accurately. I think in my implementation kObjectLoaded just describes: we already have an object in the holder, corresponding to data_rxfilename_ in the *next* line. It doesn't matter that the range was specified or not (in the current line or next line)

danpovey: Yes I think so.
Anyway make sure to describe it accurately.

src/util/kaldi-table-inl.h
xiaohui-zhang: it's not initialization. ok so I'll switch to clear(). The above line current_key_ = ""; was there before..

src/util/kaldi-table-inl.h
xiaohui-zhang: got it

danpovey: Plus that template will just crash except for class Matrix.  
You don't need to have it work for the empty string- it should just crash in that case.  If there is no range, that function shouldn't be called.

xiaohui-zhang: sure. 
yep I've made sure it crashes except for Matrix by overloading.

src/util/kaldi-table-inl.h
xiaohui-zhang: there are actually many cases of using this in this header file. Should I change all of them to clear()?

danpovey: No, leave them-- I mistakenly thought this was inside the constructor.

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "util/kaldi-holder.h"
+#include "matrix/kaldi-matrix.h"
+
+namespace kaldi {
+
+template<class Real>
```
danpovey: A comment here explaining what kinds of input it accepts would be nice, i.e. what kinds of ranges it supports.

danpovey: or by declaration.

src/util/kaldi-table-inl.h
```diff
@@ -38,6 +40,25 @@ namespace kaldi {
 /// \addtogroup table_impl_types
 /// @{
 
+// In SequentialTableReaderScriptImpl and RandomAccessTableReaderScriptImple,
+// We use this function to separate the input string "line"
+// into the range specifier (e.g. [1:2,2:10]) and the data_rxfilename
+// (e.g. 1.ark:100). It returns true if successful.
+static bool ExtractRangeSpecifier(const std::string &line,
```
danpovey: you shouldn't say 'static' in a header.
Probably best to just declare it here, and define it in kaldi-table.cc.

src/util/kaldi-table-inl.h
```diff
@@ -38,6 +40,25 @@ namespace kaldi {
 /// \addtogroup table_impl_types
 /// @{
 
+// In SequentialTableReaderScriptImpl and RandomAccessTableReaderScriptImple,
+// We use this function to separate the input string "line"
+// into the range specifier (e.g. [1:2,2:10]) and the data_rxfilename
```
danpovey: in this comment please give an example of the string 'line' itself, e.g. "1.ark:100[1:2,2:10]", and maybe other examples of inputs and outputs would be nice, e.g. without ranges or with pipes.
Although old code doesn't do this, best to use single quotes to surround variable names.

src/util/kaldi-table-inl.h
```diff
@@ -340,6 +441,13 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
     // or the user called FreeCurrent().. note,
     // if when called by user we are in this state,
     // it means the user called FreeCurrent().
+    kObjectLoaded, // Called NextScpLine() and found we already  yes         yes
```
danpovey: I don't think there is a need for a separate state kObjectLoaded because it behaves the same as kLoadSucceeded.  I.e. it means we have the object.  It complicates the analysis, IMO.

xiaohui-zhang: Since there is no function to test whether holder_ is empty, we need either a string like data_rxfilename_last_ (which I used before) or an extra state like this (which you asked me to add) to tell LoadCurrent whether to load the object from disk.

danpovey: There is already a member variable named data_rxfilename_, right?  Can't
you always ensure that if that variable is set and the state if one of the
states that indicates we have an object, then the holder_ contains the
object it points to?  That way when you do NextScpLine() or whatever it's
called, you can decide whether to set the state to kHaveScpLine, or
kLoadSucceeded, depending on whether the new rxfilename is the same as the
old.

Dan

danpovey: I guess what I'm saying is that the states kLoadSucceeded and kObjectLoaded are really the same state: it means you have the object corresponding to the data_rxfilename_ in the holder but you don't have any range object.  So I'm saying you should merge the states.  I may have mis-spoken before.

src/util/kaldi-table-inl.h
```diff
@@ -164,9 +187,11 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
   }
   const T &Value() {
     StateType orig_state = state_;
-    if (state_ == kHaveScpLine) LoadCurrent();  // Takes
-    // state_ to kLoadSucceeded or kLoadFailed.
-    if (state_ == kLoadFailed) {  // this can happen due to
+    if (state_ == kHaveScpLine || state_ == kObjectLoaded) {
```
danpovey: You could rename LoadCurrent() to EnsureObjectLoaded(), and call it here regardless of 'state_'.  Then you could make it the responsibility of EnsureObjectLoaded() to return (with return-status true) if the object was already loaded and nothing more was needed to do.

src/util/kaldi-table-inl.h
```diff
@@ -178,25 +203,41 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
       else  // orig_state_ was kLoadFailed, which only could have happened
         // if the user called FreeCurrent().
         KALDI_ERR << "You called Value() after FreeCurrent().";
-    } else if (state_ != kLoadSucceeded) {
+    } else if (state_ == kRangeExtractionFailed) {
+      KALDI_ERR << "Failed to load object from "
+                << PrintableRxfilename(data_rxfilename_) 
+                << " within range " << range_
```
danpovey: within -> with

danpovey: actually, it might be clearer to just print
data_rxfilename_[range]
i.e. to print square brackets.

src/util/kaldi-table-inl.h
xiaohui-zhang: ok that where the confusion happened. thanks for the explanation. fixed.

danpovey: Can you please go through the files and make sure that wherever I made a
comment, you've addressed it?  You seem only to have addressed some of my
latest comments.

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-holder.cc
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-holder.cc
danpovey: I don't see a comment either here or where it was declared.
And when I click on your email from github it says it can't find the commit.  Please send me an email telling me the git hash of your most recent commit, so I can figure out what went wrong.

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
xiaohui-zhang: solved!

src/util/kaldi-holder.cc
xiaohui-zhang: solved!

src/util/kaldi-holder.cc
xiaohui-zhang: solved!

src/util/kaldi-table-inl.h
```diff
@@ -38,6 +40,16 @@ namespace kaldi {
...
 /// @{
 
+// In SequentialTableReaderScriptImpl and RandomAccessTableReaderScriptImple,
+// We use this function to separate the input string 'line'
+// (e.g "1.ark:100[1:2,2:10]"), or (e.g. "sph2pipe -f wav -p -c 1 1.sph |")
+// into the data_rxfilename (e.g. "1.ark:100") and the optional range specifier
+// (valid cases are: row range e.g. [1:2], column range e.g. [:,2:10], or both
+// e.g. [1:2,2:10].) It returns true if successful. 
+bool ExtractRangeSpecifier(const std::string &line, 
+                                 std::string *data_rxfilename,
```
danpovey: move these two lines a few characters to the left to match normal style..

src/util/kaldi-table-inl.h
```diff
@@ -178,25 +191,40 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
   }
+
   void SwapHolder(Holder *other_holder) {
     // call Value() to ensure we have a value, and ignore its return value while
     // suppressing compiler warnings by casting to void.
     (void) Value();
-    if (state_ == kLoadSucceeded) {
+    if (state_ == kLoadSucceeded ||
+        state_ == kRangeExtracted || state_ == kRangeExtractionFailed) {
       holder_.Swap(other_holder);
```
danpovey: Now that I look at this I realize there is an incompatibility (my fault) between the design of this feature, and the design of the "bg" feature.  'other_holder' needs to get the same value that we would have returned from Value(), otherwise the combination of these two features won't work.  But this is not easy because there is no swap method in Holder that swaps with a raw object [and implementing such a thing in general would be hard, as we can't rely on Swap() functions always existing even for Kaldi objects].

After thinking about this a bit, I think the best way to resolve this is a slight redesign of this object-indexing feature.
Instead of a namespace function ExtractObjectRange, there will be such a function ExtractRange in all the Holder classes, and the type written to will be another Holder, and the 'object_' will instead be a holder object of type range_holder_.  In most cases you'll give it an implementation that just throws.

I think actually the easiest way to implement this in such a way that it has a functioning implementation for matrices, is to *keep* the existing ExtractObjectRange function that you've written, and have it called from the ExtractRange function of the KaldiObjectHolder.  (BTW, for other Holder types you will have to write an ExtractRange function that throws; make sure to compile at src/ level to be sure you haven't missed any.)

This redesign actually has another advantage.  Currently we have no way to clear the object (since operator = is not guaranteed to exist), so when we do NextScpLine() we have no way to free memory tied up in the range object.  Now that it's a Holder object, the memory can be freed by creating a temporary new Holder object, and swapping with the range_holder_.

xiaohui-zhang: what's "bg" feature? Otherwise it's pretty clear to me. will do as you
suggested.

On Sun, Mar 27, 2016 at 12:26 AM, Daniel Povey <notifications@github.com>
wrote:

> In src/util/kaldi-table-inl.h
> <https://github.com/kaldi-asr/kaldi/pull/611#discussion_r57523237>:
>
> >    void SwapHolder(Holder *other_holder) {
> >      // call Value() to ensure we have a value, and ignore its return value while
> >      // suppressing compiler warnings by casting to void.
> >      (void) Value();
> > -    if (state_ == kLoadSucceeded) {
> > +    if (state_ == kLoadSucceeded ||
> > +        state_ == kRangeExtracted || state_ == kRangeExtractionFailed) {
> >        holder_.Swap(other_holder);
>
> Now that I look at this I realize there is an incompatibility (my fault)
> between the design of this feature, and the design of the "bg" feature.
> 'other_holder' needs to get the same value that we would have returned from
> Value(), otherwise the combination of these two features won't work. But
> this is not easy because there is no swap method in Holder that swaps with
> a raw object [and implementing such a thing in general would be hard, as we
> can't rely on Swap() functions always existing even for Kaldi objects].
>
> After thinking about this a bit, I think the best way to resolve this is a
> slight redesign of this object-indexing feature.
> Instead of a namespace function ExtractObjectRange, there will be such a
> function ExtractRange in all the Holder classes, and the type written to
> will be another Holder, and the 'object_' will instead be a holder object
> of type range_holder_. In most cases you'll give it an implementation that
> just throws.
>
> I think actually the easiest way to implement this in such a way that it
> has a functioning implementation for matrices, is to *keep* the existing
> ExtractObjectRange function that you've written, and have it called from
> the ExtractRange function of the KaldiObjectHolder. (BTW, for other Holder
> types you will have to write an ExtractRange function that throws; make
> sure to compile at src/ level to be sure you haven't missed any.)
>
> This redesign actually has another advantage. Currently we have no way to
> clear the object (since operator = is not guaranteed to exist), so when we
> do NextScpLine() we have no way to free memory tied up in the range object.
> Now that it's a Holder object, the memory can be freed by creating a
> temporary new Holder object, and swapping with the range_holder_.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/kaldi-asr/kaldi/pull/611/files/d778589ed02590c92d5578281dd181f22280fa8b#r57523237>
>



-- 
Xiao-hui

danpovey: "bg" is where for sequential readers, if you add ",bg" to the rspecifier it
does the reading in a background thread (useful in processes that use GPUs,
to get fuller usage).

xiaohui-zhang: ok thanks

On Sun, Mar 27, 2016 at 12:55 AM, Daniel Povey <notifications@github.com>
wrote:

> In src/util/kaldi-table-inl.h
> <https://github.com/kaldi-asr/kaldi/pull/611#discussion_r57523436>:
>
> >    void SwapHolder(Holder *other_holder) {
> >      // call Value() to ensure we have a value, and ignore its return value while
> >      // suppressing compiler warnings by casting to void.
> >      (void) Value();
> > -    if (state_ == kLoadSucceeded) {
> > +    if (state_ == kLoadSucceeded ||
> > +        state_ == kRangeExtracted || state_ == kRangeExtractionFailed) {
> >        holder_.Swap(other_holder);
>
> "bg" is where for sequential readers, if you add ",bg" to the rspecifier
> it does the reading in a background thread (useful in processes that use
> GPUs, to get fuller usage).
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/kaldi-asr/kaldi/pull/611/files/d778589ed02590c92d5578281dd181f22280fa8b#r57523436>
>



-- 
Xiao-hui

src/util/kaldi-holder-inl.h
```diff
@@ -101,6 +102,12 @@ template<class KaldiType> class KaldiObjectHolder {
     std::swap(t_, other->t_);
   }
 
+  bool ExtractRange(KaldiObjectHolder<T> *other, std::string range) {
```
danpovey: It would be more efficient to make all of these 'range' arguments const std::string&.

src/util/kaldi-table-inl.h
```diff
@@ -178,20 +191,35 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+      return range_holder_.Value();
+    } else {
+      // Here state_ is kLoadSucceeded.
+      return holder_.Value();
+    }
   }
+
   void FreeCurrent() {
-    if (state_ == kLoadSucceeded) {
+    if (state_ == kLoadSucceeded ||
```
danpovey: Here, FreeCurrent() should only be called when we successfully have a value, so state kRangeExtractionFailed isn't expected in non-error operation.  Actually, if state_ == kLoadSucceeded [i.e. there is no range object], we should clear holder_, otherwise if state is kRangeExtracted then clear range_holder_, otherwise print a warning.

src/util/kaldi-table-inl.h
```diff
@@ -199,11 +227,15 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
     if (state_ == kLoadSucceeded) {
       holder_.Swap(other_holder);
       state_ = kLoadFailed;
+    } else if (state_ == kRangeExtracted) {
+      range_holder_.Swap(other_holder);
+      state_ = kLoadFailed;
```
danpovey: at this point I think the state should be set to kLoadSucceeded, indicating that we still have the base object (but no range).

src/util/kaldi-table-inl.h
```diff
@@ -245,52 +282,81 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
-                 << PrintableRxfilename(data_rxfilename_);
-      state_ = kLoadFailed;
-      return false;
-    } else {
-      if (holder_.Read(data_input_.Stream())) {
-        state_ = kLoadSucceeded;
-        return true;
-      } else {  // holder_ will not contain data.
-        KALDI_WARN << "Failed to load object from "
+    if (state_ != kLoadSucceeded) {
```
danpovey: Now that we are calling EnsureObjectLoaded() from Value() without first checking the state, we have to be prepared for it this function to be called twice.  If the state is kRangeExtracted or kRangeExtractionFailed, this function should just return without doing anything, because everything that needs to be done is already done.
And I think this if-statement (which I know is old code) could be reorganized for more clarity to say
  if (state_ == kHaveScpLine) { 
    ... do all this stuff
  } else if (state != kLoadSucceeded) {
     KALDI_ERR << "EnsureObjectLoaded() called at the wrong time.";
  }

src/hmm/posterior.h
```diff
@@ -77,6 +77,11 @@ class PosteriorHolder {
   void Swap(PosteriorHolder *other) {
     t_.swap(other->t_);
   }
+
+  bool ExtractRange(PosteriorHolder *other, const std::string &range) {
+    KALDI_ERR << "ExtracRange is not defined for this type of holder.";
```
danpovey: typo ExtracRange (also present in other files- check all).

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+bool ExtractObjectRange(const Matrix<Real> &input, const std::string &range,
+                        Matrix<Real> *output) {
+  if (range.empty()) {
+    KALDI_ERR << "Empty range specifier."; 
+    return false;
+  }
+  std::vector<std::string> splits;
+  SplitStringToVector(range, ",", true, &splits);
```
danpovey: this 'true' should probably be 'false'.

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+template<class Real>
+bool ExtractObjectRange(const Matrix<Real> &input, const std::string &range,
+                        Matrix<Real> *output) {
+  if (range.empty()) {
+    KALDI_ERR << "Empty range specifier."; 
+    return false;
+  }
+  std::vector<std::string> splits;
+  SplitStringToVector(range, ",", true, &splits);  
+  if (splits.size() > 2) {
```
danpovey: You should also to die if split.size() == 0 at this point.

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+    KALDI_ERR << "Invalid range specifier: " << range;
+    return false;
+  }
+  std::vector<int32> row_range, col_range;
+  SplitStringToIntegers(splits[0], ":", true, &row_range);
```
danpovey: these two 'true's should also be 'false'.

danpovey: ... but you should handle splits[0] == ":" and splits[1] == ":" as special cases.  [passing 'true' is a messy way to do it as it would treat : and :: the same]

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,74 @@
...
+    col_range.push_back(input.NumCols() - 1);
+  }
+  if (!(row_range.size() == 2 && col_range.size() == 2 && 
+        row_range[0] >= 0 && row_range[0] < row_range[1] &&
+        row_range[1] < input.NumRows() && col_range[0] >=0 &&
+        col_range[0] < col_range[1] && col_range[1] < input.NumCols())) {
+    KALDI_ERR << "Invalid range specifier: " << range;
+    return false;
+  }
+  int32 row_offset = row_range[1] - row_range[0] + 1,
```
danpovey: 'offset' is not the right name here, 'size' would be better.

src/util/kaldi-table-inl.h
```diff
@@ -1631,25 +1748,48 @@ class RandomAccessTableReaderScriptImpl:
...
-          } else {
-            KALDI_WARN << "Error reading object from "
-                "stream " << PrintableRxfilename(script_[key_pos].second);
-            state_ = kNotHaveObject;
+          data_rxfilename = script_[key_pos].second;
+          range_ = "";
+        }
+        // we only need to update holder_ if data_rxfilename is different
+        // from last time.
+        if (data_rxfilename != data_rxfilename_last_) {
```
danpovey: Here, there seems to be no reason to have data_rxfilename_last_ as a class-member variable.  You could just have it as a local variable inside the function.

danpovey: oh, I see.  The data_rxfilename_ is not stored.  The variable is needed, but I think it should be called data_rxfilename_, not data_rxfilename_last_, because it relates to the currently held object, not the one before.

src/util/kaldi-table-inl.h
```diff
@@ -1631,25 +1748,48 @@ class RandomAccessTableReaderScriptImpl:
...
+          if (!input_.Open(data_rxfilename)) {
+            KALDI_WARN << "Error opening stream "
+                       << PrintableRxfilename(data_rxfilename);
             return false;
+          } else {
+            data_rxfilename_last_ = data_rxfilename;
+            // Make sure holder empty.
```
danpovey: actually we don't have to clear the holder, it will happen anyway inside Read(), IIRC

src/util/kaldi-table.cc
```diff
@@ -22,6 +22,21 @@
 
 namespace kaldi {
 
+bool ExtractRangeSpecifier(const std::string &line, 
+                           std::string *data_rxfilename,
+                           std::string *range) {
+  if (line.empty() || line[line.size()-1] != ']') 
+    return false;
+  std::vector<std::string> splits;
+  SplitStringToVector(line, "[", true, &splits);
```
danpovey: this 'true' should be 'false'

src/util/kaldi-table-inl.h
danpovey: Also you have to make 100% sure that data_rxfilename_ always corresponds to the object stored in holder_.  This has to be true not just when ranges are used.  And in any circumstance under which the holder is cleared, if it exists you need to set data_rxfilename_  to empty-- make sure there is no code path in which the holder is cleared or overwritten but data_rxfilename_ is not cleared or overwritten.

src/util/kaldi-table-inl.h
```diff
@@ -1631,25 +1748,48 @@ class RandomAccessTableReaderScriptImpl:
...
-          if (holder_.Read(input_.Stream())) {
-            state_ = kHaveObject;
-            current_key_ = key;
-            return true;
-          } else {
-            KALDI_WARN << "Error reading object from "
-                "stream " << PrintableRxfilename(script_[key_pos].second);
-            state_ = kNotHaveObject;
+          data_rxfilename = script_[key_pos].second;
+          range_ = "";
```
danpovey: at this point you might as well do
  if (!range_.empty()) {
     range_ = "";
     range_holder_.Clear(); 
  }
to save memory.

src/util/kaldi-table-inl.h
danpovey: ... and document by the declaration, that data_rxfilename_ is the rxfilename that was used to read the object in holder_, and will be nonempty whenever holder_ contains an object-- and then make sure that this always holds.

xiaohui-zhang:
I've tested it with various error cases like  [1::3,1:2] [1:3,1:2]] [[1:3,1:2] [,1:2] [1,1:2] [1,,1:2] [,,1:2]. Also tested calling Value() 0,1,2 times after each call of Next() or HasKey().

src/util/kaldi-table-inl.h
```diff
@@ -1631,25 +1751,49 @@ class RandomAccessTableReaderScriptImpl:
...
+          } else {
+            data_rxfilename_ = data_rxfilename;
+            if (holder_.Read(input_.Stream())) {
+              state_ = kHaveObject;
+              current_key_ = key;
+              return true;
+            } else {
+              KALDI_WARN << "Error reading object from "
+                  "stream " << PrintableRxfilename(script_[key_pos].second);
+              state_ = kNotHaveObject;
```
danpovey: You need to check every code path and check that the invariances hold-- e.g. here, we have a code path where holder_contains no object, but data_rxfilename_ is nonempty.

src/util/kaldi-table-inl.h
```diff
@@ -1631,25 +1751,49 @@ class RandomAccessTableReaderScriptImpl:
...
+              return false;
+            }
           }
         }
```
danpovey: I think the code would be easier to understand if the 3 lines below were in an else statement.

src/util/kaldi-table-inl.h
```diff
@@ -298,11 +373,38 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+      // (e.g. 1.ark:100) and possibly the range_ specifer
+      // (e.g. [1:2,2:10]) from "rest".
+      std::string data_rxfilename_next, rest;
+      SplitStringOnFirstSpace(line, &key_, &rest);
+      if (!key_.empty() && !rest.empty()) {
         // Got a valid line.
         state_ = kHaveScpLine;
+        if (rest[rest.size()-1] == ']') {
+          if(!ExtractRangeSpecifier(rest, &data_rxfilename_next, &range_)) {
+            // Got an invalid line.
```
danpovey: in this code path nothing is printed to say what the error was.

danpovey:
Samuel, let me take over this pull request, I'll copy your branch.  We are not going to get down to a bug-free pull request within a reasonable time unless I do it personally.

xiaohui-zhang:
sure. sorry for the bugs.

danpovey:
Closing this one and reopening it as a different PR after my changes.

PR #612, opened 2016-03-23
WIP: Swbd xent cmp

galv:
Setup for deciding whether chain model effectiveness is due to neural network setup or cost function.

@vijayaditya 

Hopefully I understood you correctly in you wanted here. The changes seem almost a little too obvious (use local/nnet3/run_tdnn.sh's relu_dim and splice-indexes in a local/chain/run_tddn_6z.sh). Since both neural networks are the same then, they obvioulsy have the same number of parameters. If I misunderstood, tell me immediately.

Biggest hyperparameter difference is the learning rate between the two files I mentioned above.


vijayaditya:
Could you submit this PR with the file named as run_tdnn_6z.sh. Git would then generate a diff, which makes the comparison easy for me.

kkm000:
We are [using `WIP` to mark work in progress](http://kaldi.sourceforge.net/tutorial_git.html), so I've changed the title.

galv:
I submitted a new commit, so you can now see the diff of egs/swbd/s5c/local/chain/run_tdnn_6z.sh.

It's fairly obvious, so again, let me know if I misunderstood. I originally thought I'd have to get the same number of parameters between this chain model (using the xent splice indexes) and the old chain model, but it seems that is not what you wanted.

egs/swbd/s5c/local/chain/run_tdnn_6z.sh
```diff
@@ -55,7 +55,7 @@ final_layer_normalize_target=0.5
 num_jobs_initial=3
 num_jobs_final=16
 minibatch_size=128
-relu_dim=576
+relu_dim=1024
```
vijayaditya: Make this 768 to match the parameters. You should take into account the final affine transform size when calculating parameters.

egs/swbd/s5c/link_most.sh
```diff
@@ -0,0 +1,33 @@
...
+#  ctc
+  make_dbl3
+  make_hires
+  make_hires_dbl
+  make_hires_dbl2
+  make_hiresf
+  make_mfcc
+  mono
+  mono_ali
+#  nnet3
```
vijayaditya: You should also link the ivector extractor from the Dan's setup.

egs/swbd/s5c/local/chain/run_tdnn_6z.sh
```diff
@@ -55,7 +55,7 @@ final_layer_normalize_target=0.5
 num_jobs_initial=3
```
vijayaditya: You can reuse the lattices and trees from the previous experiments. You should be able to start this experiment from stage 12.

egs/swbd/s5c/local/chain/run_tdnn_6z.sh
galv: Do you mean the final affine transform represented by "0" at the end of the splice indexes? Actually, if you look at swbd/s5c/local/nnet3/run_tdnn.sh, you will see this 0 at the end of the splice indexes, as well [here](https://github.com/kaldi-asr/kaldi/blob/master/egs/swbd/s5c/local/nnet3/run_tdnn.sh#L57). Even though my understanding is that the point of the final affine component is for using cross-entropy training as a regularizer in chain models, this has snuck into the cross entropy model as well. (I haven't done a git blame to find out when this originally appeared.) Since the splice indexes are exactly the same, I believe relu_dim should be the same between the the cross entropy model and chain model in order to guarantee that they have the same number of parameters, as they will use the same neural network when we set the splice indexes the same. At the least, I convinced my self the relu_dim should be the same between the two models to guarantee the same number of parameters with an equation I came up with for number of parameters as a function of relu_dim.

Also, sorry about the delay. I know we need to get this out ASAP.

danpovey: Probably easiest just to set it to 768 for now- not sure how Vijay computed
it but likely correct.

vijayaditya: [6z.pdf](https://github.com/kaldi-asr/kaldi/files/187176/6z.pdf)

See the figure in the attached pdf. You can see that model 6z has a fork at the penultimate layer. There are 2 final layers one in the branch which goes to chain objfn and one in the branch which goes to xent objfn. You should count the parameters in the branch which goes to chain objn. You can read steps/nnet3/tdnn/make_configs.py to understand how the final " 0 " is interpreted by the python code to write out the model config.

We need to increase the relu_dim from 576 in 6z to 768 as the splicing indices in 6z splice 3 hidden outputs at the input of every tdnn layer while the splicing indices used in the local/nnet3/run_tdnn.sh splice only 2 hidden layer outputs at every tdnn layer.

galv: Ah, okay, I think I misinterpreted you.

Currently, I propose that 6z be changed to have:
```
--splice-indexes "-2,-1,0,1,2 -1,2 -3,3 -7,2 0" \ 
```
which mostly has splices of size 2 while it before had
```
--splice-indexes "-1,0,1 -1,0,1,2 -3,0,3 -3,0,3 -3,0,3 -6,-3,0 0" \
```
which mostly has splices of size 3.

So the purpose of changing relu_dim from 576 to 768 is so that the input dimension of each affine transform is (4/3)*2 = 8/3 (2 is for the 2 splices as input, 4/3 is 768/576) in the new setup, which is almost 3, for the 3 splices of input for the 6z setup.

But this analysis assumes that the new model has the same number of layers as the cross entropy model. The cross-entropy model has two fewer layers. I hope you agree with me that this is a problem.

I'm still a bit confused, so perhaps the best thing I can say is: look at my splice-indices again. Is this what you wanted? Please make sure. It has two fewer layers than the current chain models. I reduced the number of layers to make the number of layers the same as the cross entropy model.

vijayaditya: The goal of this experiment is to use the same model (to the extent possible) as the xent system and see the improvement due to chain objfn. So changing splice indexes in _6z.sh script to 
--splice-indexes "-2,-1,0,1,2 -1,2 -3,3 -7,2 0" and making relu-dim 768 should be sufficient. Ideally we would not even changed relu-dim from 1024 to 768 as we wanted to compare improvement from just changing the objfn. But we do know from previous experiments that models trained with chain objfn overfit, so they should be smaller. So we reduce the number of parameters, to match those of 6z system.

In case you are not convinced you can start another experiment with 
 --splice-indexes "-2,-1,0,1,2 -1,2 -3,3 -7,2 0" and relu-dim which will make the model the same as xent system (except for the number of leaves i.e., output-dimension of the network). We have free GPUs right now, so we can afford the second experiment.

galv: Yep, got it. I won't run the other experiment you proposed, as I get your point now. Sorry for misunderstanding.

I'm going to keep num_epochs=4 and learning_rate the same as in 6z. The cross entropy model has a higher learning rate, but I think it makes sense to make smaller steps for a discriminative model.

Will get the experiment going now. Will reopen this if I see problems.

vijayaditya: OK, keep all hyper-parameters the same as in 6z. Just change splice-indexes
and relu-dim.

galv: Yep. That is what I am doing.

PR #613, opened 2016-03-23
Fix InputDim and IvectorDim

jcsilva:


jtrmal:
@vijayaditya  Does this make sense? I can merge right away if it fixes a substantial issue.
y.

vijayaditya:
Good to merge, it is surprising we never found this issue till today.

jtrmal:
merging. Thanks Eduardo.

PR #614, opened 2016-03-24
sequence: added input frame shift to chain sequence training and removed frame shift during at example level during training

vimalmanohar:


danpovey:
OK.  You can put results and WER comparisons in the script file in future- it's a reasonable way to document what the results were.

PR #616, opened 2016-03-24
sequence: Added shift_feats.sh file missing from previous commit

vimalmanohar:
I wrote a script to wrap shift-feats binary, which was supposed to be in the previous commit. Should I commit this or wait for Samuel to finish the indexing-on-scp work?


danpovey:
I'll merge it, as you'd still need the wrapping script after Samuel's changes, you'd just change how it works internally.

danpovey:
Once we merge Samuel's stuff, you could modify your setup to only use
negative shifts- then there will be no question of how to duplicate the 1st
frame.

PR #618, opened 2016-03-25
steps/cleanup : Added a new version of the combine_short_segments script

vijayaditya:


danpovey:
Don't you need this in chain?  Should I merge from master to chain after this?

danpovey:
And wouldn't you need to also change all invocations of this script, since the suffix changed?

vijayaditya:
I didn't think this was ever used. Let me check.

danpovey:
OK, you can merge it yourself after you've checked, just bear in mind that
it may be necessary to merge changes from master to chain if you need it in
chain.

vijayaditya:
Modified one invocation of the script. Merging for now as this is not used much and it is necessary for running AMI experiments immediately.

PR #620, opened 2016-03-25
WIP : ami: bug fixes to chain recipe

vijayaditya:
The previous commit pushing ami chain recipe was untested. 
These are bug fixes for the AMI recipe. I will keep updating as I see more issues.

### DIFF
--- a/egs/ami/s5/local/chain/run_tdnn_ami_4.sh
+++ b/egs/ami/s5/local/chain/run_tdnn_ami_4.sh
@@ -162,11 +162,12 @@ fi
 
 # combining the segments in training data to have a minimum length of frames_per_eg + tolerance
 # this is critical stage in AMI (gives 1% absolute improvement)
+min_seg_len=$(python -c "print ($frames_per_eg+5)/100.0")
 if [ $stage -le 12 ]; then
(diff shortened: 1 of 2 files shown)


danpovey:
Vijay, after I looked at your run_tdnn_ami_4.sh script I was left a little confused.  The general principle with run.sh scripts is that they should be self-contained so that if you run them with no arguments you get the result.  In this case you seem to have run the script with various arguments, and there is no very clear indication in the script (a) what arguments you ran it with, or are supposed to run it with, and (b) what the results were in each case.  Also it's not clear how this script relates to the other 6z script in the same directory and how the results compare.
I'm OK to have these scripts take a small number of arguments if it's really necessary for some reason, but it needs to be very clearly stated at the top of the script, and the "recommended" configuration should be clearly singled out.  Also it's not the case that all possible arguments need to be exposed by these run.sh scripts (i.e. made settable by invoking the run.sh script with arguments).  The buck has to stop somewhere, and the run.sh script is where it stops, so if they are to be configurable, it should only be in a very minimal way as necessary (e.g. if a variable is used only once in the script, in general it shouldn't exist at all).

PR #624, opened 2016-03-27
WIP: Refactor the path.sh structure

jtrmal:
I modified only a handful of the recipes for review of the proposed concept


jtrmal:
@danpovey  shall I proceed with modifiing the rest of the path.sh files, or is here anything that should be changed?
I'm considering to test the presence of the file $KALDI_ROOT/src/path.sh prior sourcing, for example

danpovey:
Looks OK to me but make sure it works e.g. that you can run the commands.
Merge yourself.

danpovey:
Yes, go ahead- looks good.
You could try '|| exit 1' while sourcing it- that would work as a test.
Dan

jtrmal:
OK, thanks. I tried running rm recipe and it went past a few mono
iterations. After finishing the changes, I'll run the full recipe.
y.

danpovey:
Probably what you've done is enough testing- we'll get reports if something
fails
Dan

PR #629, opened 2016-03-29
Modified Global Phone recipe

BogdanvL:
I provided some modifications for GP recipe.

In RESULTS file one could find WER obtained with models trained with modified GP recipe.
Original GP recipe  includes WER rates for mono and tri1 models.

I will be happy to share my recipe with Kaldi community.


danpovey:
Thanks!
I think that since the old RESULTS had more languages, it would be a good
idea if you were to include the old RESULTS at the bottom of the results
file, with a suitable note.
Were there any substantial changes to the pipeline?  I notice you have done
some cosmetic cleanup.

BogdanvL:
I included old RESULTS at the bottom of the results file.

I provide some corrections to Arnab script.
I used state-of-the-art WSJ pipeline during developing of my recipe.

danpovey:
OK.
These days we would normally just skip SGMMs and go straight on to some kind of neural net training.  But obviously that requires GPUs; and there are a lot of options.  I'm merging your recipe.
Thanks!

BogdanvL:
OK. Thanks.

nnet models will be provided by Idiap later.
There is an ongoing projects which use multilingual ASR system.

Best wishes!

PR #630, opened 2016-03-30
configure, refactored selection of GPU architectures

vesis84:
- moved from src/cudamatrix/Makefile to src/configure
- also fixes bug from xie7132524@gmail.com

### DIFF
--- a/src/chain/Makefile
+++ b/src/chain/Makefile
@@ -2,7 +2,6 @@
 all:
 
 include ../kaldi.mk
-
 LDFLAGS += $(CUDA_LDFLAGS)
 LDLIBS += $(CUDA_LDLIBS)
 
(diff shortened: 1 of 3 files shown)


danpovey:
This seems like an improvement but I'd like to test it out to-morrow so not merging right away.

danpovey:
@hainan-xv, could you please test this out that it works for us?  Obviously run configure on a GPU machine.

hainan-xv:
Sure. I will do it by tonight.

On Wed, Mar 30, 2016 at 1:56 PM, Daniel Povey <notifications@github.com>
wrote:

> @hainan-xv <https://github.com/hainan-xv>, could you please test this out
> that it works for us? Obviously run configure on a GPU machine.
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/630#issuecomment-203552645>
>



-- 
- Hainan

hainan-xv:
Should I test the compilation and make test only, or should I also test it for some real data, like RM?

danpovey:
Just test compilation and make test

hainan-xv:
All tests passed on t01.

On Wed, Mar 30, 2016 at 5:22 PM, Daniel Povey <notifications@github.com>
wrote:

> Just test compilation and make test
>
> On Wed, Mar 30, 2016 at 5:20 PM, hainan-xv <notifications@github.com>
> wrote:
>
> > Should I test the compilation and make test only, or should I also test
> it
> > for some real data, like RM?
> >
> > —
> > You are receiving this because you commented.
> > Reply to this email directly or view it on GitHub
> > <https://github.com/kaldi-asr/kaldi/pull/630#issuecomment-203640321>
>
> >
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/630#issuecomment-203640821>
>



-- 
- Hainan

danpovey:
Karel, have you tested that it works as intended when you update an existing repository to have this without re-running configure?
I think it would be better if you change the error:
 $(error CUDA_ARCH is undefined, run 'src/configure')
to be a little clearer this resulted from a script change, unless people are confused why a previously working setup would now stop working.
How about:
 $(error The mechanism for configuring CUDA has changed, please re-run 'src/configure')

jtrmal:
To me, incrementing the value would be a simpler solution without introducing any new complexity in the scripts. I think people running in sync with github recognize the fact they might have to reconfigure and recompile. Reconfiguration is as easy as taking the second line from kaldi.mk and run it again.
But I don't care too much.
y.

vesis84:
Yes, increasing the value is a good solution. I suppose most of the researches use kaldi with CUDA extensions anyway, while for others the reconfigure is not needed, but causes no harm if done.
I'll update the constant in 'configure'.
K.

danpovey:
Karel, I just realized that we have another directory where we compile GPU stuff-- the chain directory.  Would you mind changing that also?

vesis84:
Okay, src/chain/Makefile is now updated.

danpovey:
OK, thanks.  Merging.

PR #631, opened 2016-03-30
moving the src/path.sh into tools/config/common_path.sh

jtrmal:
Tested by running rm
y.


danpovey:
This is not urgent so I'll wait a day or so in case anyone has a comment.

danpovey:
Neither the old src/path.sh nor  tools/config/common_path.sh are supposed
to be sourced directly: instead they are supposed to be sourced from the
path.sh in the experimental directory.  So the usage is the same.

Dan

danpovey:
Yenda, since tools/config is never going to be on the path, I see no reason not to name the file path.sh instead of common_path.sh.  And do you have some kind of plan for the directory `tools/config`?  It doesn't exist currently.
Dan

kkm000:
> Neither the old src/path.sh nor tools/config/common_path.sh are supposed to be sourced directly: instead they are supposed to be sourced from the path.sh in the experimental directory.

Yes, that's what I mean, Eval it there instead of sourcing in egs/xxx/path.sh. This way the path can be set up automatically without having to set KALDI_SOURCE up front, and at the same time break the requrement of the invoking shell being bash.

danpovey:
So the file (say) tools/common/path.sh would do something like
echo PATH=xyz:\$PATH
I suppose that's a possibility.  I'm happy for Yenda to do whatever he
prefers.
Dan


On Fri, Apr 1, 2016 at 12:49 AM, Kirill Katsnelson <notifications@github.com

kkm000:
Yes, exactly. I would write that as

`declare -p PATH | awk '{print "export",$3}'`

just to be on the safe side re quoting.

kkm000:
I mean in case the original PATH contained spaces or something else weird

kkm000:
Come think of it, safer

PATH="xyz:\$PATH"
declare -p PATH foo | awk '{$1="export"; $2=""; print}'

jtrmal:
@danpovey , yes I have plans with tools/config -- want to store the dynamically generated configs for the individual tools there. The current single file approach (tools/env.sh) is inconvenient in cases where the tool will get recompiled/reinstalled and the config has to be re-created (we usually give up and ask the user to delete the related parts in tools/env.sh). In these cases, simply overwriting specific file would be much easier. So for the time being, I'd keep the name as common_path.sh and use path.sh for example for a script that would source all the partial config scripts. Or I could make it more self-explaining and rename it to binaries_path.sh or whatever you will like.
@kkm000 I understand and I'm a little bit torn here. The issue I'm concerned about with doing $() is when we want to report _a_ failure (for whatever reason). In that case, we would have to handle the return codes and so on (which would make the egs path.sh quite complex). In case we won't do it,  the script will continue  as if nothing happened. And users do not usually care about errors until something crashes later on (or get 150% WER). Sometimes it could be hard to figure out the primary issue. I'm still opened to be convinced to go either way, though...

danpovey:
Yenda, I'm fine with  you doing it as in the PR.

kkm000:
@jtrmal: The script is only setting the PATH, for God's sake! Realistically, I doubt it is possible to come up with a PATH set in error such that an eg scenario runs end to end, completes without a single runtime error (so that all required kaldi tools, or whatever ran instead of them, have been found!!!), but get incorrect results, such as bad WER, at the end. You probably need to set PATH to point to another. broken Kaldi installation. I believe it is much more reasonable to assume that everything will crash and burn quickly if PATH is not modified due to an error 

This has hardly been a concern; of 1100+ places sourcing path.sh, only 85 check whether sourcing succeeded (and more than that check if it exists before sourcing it, and thus ignore it altogether if it's not).

I would not complicate the egs' path.sh either. It can be as simple as single line 
```
eval `../somewhere/common_path.sh`
```
`../somewhere/common_path.sh` can `echo false` to indicate error. If common_path.sh itself is missing... well, nothing is going to work anyway, and you get an error for free telling the user that common_path.sh is missing. 

If you like, I can send a PR showing how this would all look and work.

kkm000:
If you want it completely solid and POSIX-only, then this 2-liner path.sh
```
eval `../somewhere/common_path.sh`
kaldi-path-verify
```
where `kaldi-path-verify` runs a couple of kaldi commands with `--help 2>/dev/null` (does not have to print even a diagnostics, since an error would be indicated above already). Then this can be sourced in any POSIX shell as
```
. path.sh || exit
```
I am just trying to minimize the path.sh. Currently they seem to be a bit convoluted.

danpovey:
@jtrmal, don't forget to decide about this!

jtrmal:
Yeah,  I will look into it probably on Monday. Was busy with other stuff
and tired on overall.  Y.

@jtrmal <https://github.com/jtrmal>, don't forget to decide about this!

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/kaldi-asr/kaldi/pull/631#issuecomment-207834308>

jtrmal:
Merging it as it is for now. Will keep @kkm000's suggestions in mind for the next steps.
y.

PR #633, opened 2016-03-31
Update Makefile

rickychanhoyin:
Update for CUDA toolkit with version 7.0 and above 

[ gencode arch=compute_13,code=sm_13 
  gencode arch=compute_10,code=sm_10 ] are not supported anymore in cuda toolkit 7.0 and above

### DIFF
--- a/src/cudamatrix/Makefile
+++ b/src/cudamatrix/Makefile
@@ -48,14 +48,19 @@ ifeq ($(CUDA), true)
   ifeq ($(CUDA_VER_GT_6_0), true)
     CUDA_ARCH += -gencode arch=compute_50,code=sm_50
   endif
-  #For toolkit older than 6.5, add the compute capability 1.0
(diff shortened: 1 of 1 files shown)


jtrmal:
Ricky, this patch will conflict with #630
Frankly, I don't think compute_10 or compute_13 should ever be supported (at the cost of complicating things) -- devices supporting only those caps are memory limited and feature limited, so I'd prpopse just to remove both (irrespective of cuda version).

jtrmal:
BTW, that patch (#630) already includes your changes. 
Feel free to apply it and give it a run. It would be most welcome.
I will close this PR for duplicity. Thanks anyway!

rickychanhoyin:
I see. Thanks.

Ricky
2016/4/1 上午12:12 於 "jtrmal" <notifications@github.com> 寫道：

PR #635, opened 2016-04-01
WIP: Proposed: path.sh to automatically add src/*bin directories to PATH

kkm000:
Since the new `src/path.sh` is bash-only anyway, I modified it like this to add all `src/*bin` directories to the path automatically. This is not the prettiest code out there, but does the job. This uses only bash built-ins.

Thoughts?

[setting to WIP as it interferes with the ongoing changes #631]

### DIFF
--- a/src/path.sh
+++ b/src/path.sh
@@ -1,25 +1,9 @@
-# we are using BASH_SOURCE[0], because its set correctly even when the file
+# We are using BASH_SOURCE[0], because its set correctly even when the file
 # is sourced.
-# The formatting of the path export command is intentionally weird, because
-# this allows for easy diff'ing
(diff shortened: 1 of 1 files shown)


kkm000:
Yep--that's why I changed that to WIP

jtrmal:
I think it all boils down if we want that list (directories to be added to PATH) to be generated automatically or if we are OK with maintaining it by hand. Given the fact that there is not that many dirs (and they are not added/renamed often), I don't see keeping list of directories as much of a burden and IMO ads clarity to it.

danpovey:
I agree that it's better to do it by hand- who knows if people might rename
directories e.g. as backups, and not expect them to be automatically added
to the path.

kkm000:
Sounds good, I am dropping this PR then.

Issue #632, opened 2016-03-30
librispeech SPOKEN_NOISE vs UNK


danpovey:
In Librispeech, we have <SPOKEN_NOISE> in oov.txt instead of <UNK>.  It would be better to have <UNK> in oov.txt and no <SPOKEN_NOISE>, because some later script (well, the --remove-oov option to mkgraph) assume that the unk symbol in the LM has the same word-id as this.
Dan

vdp:
I can't seem to remember how exactly we are using SPOKEN_NOISE and UNK in the recipe.. Is this just a question of replacing SPOKEN_NOISE w/ UNK in lines 55 and 211 of run.sh? Also, I can't see the "remove-oov" option in the current version of utils/mkgraph.sh- perhaps this is related to the changes that were necessary for the chain paper?

vdp:
Nevermind about the "remove-oov" option- I see it's in the "chain" branch.

danpovey:
Yes I think it is as simple as that.  They have the same pronunciation so
it won't make any difference in general.  Would you mind doing a PR, to the
chain branch?
Dan

PR #637, opened 2016-04-01
Change the OOV word for LibriSpeech from SPOKEN_NOISE to UNK

vdp:
(Addresses issue #632)

Dan, BTW I think the scoring(local/score.sh) may not be entirely correct.. It tries to filter out NOISE and SPOKEN_NOISE(the script was probably copied from another recipe), but the OOV words from the reference transcripts are in fact not mapped to any such special token. Probably not a big deal because the OOV rates for LS are fairly low IIRC.


danpovey:
OK, thanks-merging.
I think the way the scoring works is OK- it's not normally considered right to do that type of mapping on the reference.

PR #639, opened 2016-04-01
bugfix in SRI language model formatting

mmmaat:
Hello,

This is a little bug fix on wsj/s5/utils/prepare_lm_sri.sh, please merge it it or have any comment if you are interested in.

### DIFF
--- a/egs/wsj/s5/utils/format_lm_sri.sh
+++ b/egs/wsj/s5/utils/format_lm_sri.sh
@@ -104,6 +104,27 @@ awk '{print $1}' $out_dir/words.txt > $tmpdir/voc || exit 1;
 change-lm-vocab -vocab $tmpdir/voc -lm $tmpdir/lm.gz -write-lm $tmpdir/out_lm \
   $srilm_opts || exit 1;
 
+# change-lm-vocab introduces error in out_lm when lm.gz has an order
+# lower than 3 (i.e. on unigrams and bigrams). An empty trigram (and
+# potentially bigram) is created in the lm file. This leads arpa2fst
+# to fail in the next step with a message "Zero ngram count in ngram
+# order [...] There is possibly a problem with the file". The
+# following code looks for empty ngrams (can be of any order) and
+# delete them from the file.
+empty_ngrams=$(sed -rn "s/^ngram +([[:digit:]]+)=0$/\1/p" $tmpdir/out_lm)
+if [ ! -z "$empty_ngrams" ]; then
+    # filter out empty data fields
+    for n in $empty_ngrams
+    do
+        sed "/${n}-grams:/d" $tmpdir/out_lm > $tmpdir/tmp_lm
+        mv $tmpdir/tmp_lm $tmpdir/out_lm
+    done
+
+    # filter out empty headers
+    sed '/^ngram.*=0$/d' $tmpdir/out_lm > $tmpdir/tmp_lm
+    mv $tmpdir/tmp_lm $tmpdir/out_lm
+fi
+
 arpa2fst $tmpdir/out_lm | fstprint \
   | utils/eps2disambig.pl | utils/s2eps.pl \
   | fstcompile --isymbols=$out_dir/words.txt --osymbols=$out_dir/words.txt \


egs/wsj/s5/utils/format_lm_sri.sh
```diff
@@ -104,6 +104,27 @@ awk '{print $1}' $out_dir/words.txt > $tmpdir/voc || exit 1;
 change-lm-vocab -vocab $tmpdir/voc -lm $tmpdir/lm.gz -write-lm $tmpdir/out_lm \
   $srilm_opts || exit 1;
 
+# change-lm-vocab introduces error in out_lm when lm.gz has an order
```
danpovey: This might have started happening when we updated the arpa2fst code.  @kkm000, I wonder if it would be easier to just change this error to a warning?

danpovey: I see that it *is* currently a warning:
```
+    if (ngram_counts_[cur_order - 1] == 0) {
+      KALDI_WARN << "Zero ngram count in ngram order " << cur_order
+                 << "(look for 'ngram " << cur_order << "=0' in the \\data\\ "
+                 << " section). There is possibly a problem with the file.";
+      continue;
+    }
```
@mmmaat, can you please be more specific on what the problem is that you are experiencing, e.g. how does arpa2fst fail?  Or this just to suppress a warning  message?

kkm000: @danpovey, I can remove the warning. Just never thought such a model could be legit, but if it is, maybe the warning should not be there at all. What do you think?

danpovey: I think the warning is valid-- after all, that configuration doesn't make
sense, it's effectively a bug in SRILM.  But I want to wait to hear whether
it was just a warning that was the issue, or whether there was later an
actual error.

kkm000: What I can do is ignore highest-order n-grams down until there is an order of non-zero cardinality. I think this is what actually happens here. This could be useful as it can potentially interfere with the selection of packed vs regular history keys. Packed keys support up to 4-grams, inclusively. If there is a 0 count 5-gram section, that will pessimize the compile.

danpovey: Sure, if it's easy enough.

danpovey: .. but probably still print a warning.  It's a feature, not a bug.

mmmaat: Actually I was wrong on the error message (which is you're right a warning). The error is generated latter in the code: 
ERROR (arpa2fst:Read():arpa-file-parser.cc:228) in line 3235: Invalid or unexpected directive line '\3-grams:', expected \end\.

Here is the full output of the format_lm_sri script:

> Converting '/home/mbernard/dev/abkhazia/test/language/s5/data/local/language/G.arpa.gz' to FST
> arpa2fst /tmp/kaldi.J7H6/out_lm
> LOG (arpa2fst:Read():arpa-file-parser.cc:90) Reading \data\ section.
> LOG (arpa2fst:Read():arpa-file-parser.cc:147) Reading \1-grams: section.
> LOG (arpa2fst:Read():arpa-file-parser.cc:147) Reading \2-grams: section.
> WARNING (arpa2fst:Read():arpa-file-parser.cc:134) Zero ngram count in ngram order 3(look for 'ngram 3=0' in the \data\  section). There is possibly a problem with the file.
> ERROR (arpa2fst:Read():arpa-file-parser.cc:228) in line 3235: Invalid or unexpected directive line '\3-grams:', expected \end\.
> ERROR (arpa2fst:Read():arpa-file-parser.cc:228) in line 3235: Invalid or unexpected directive line '\3-grams:', expected \end\.
> [stack trace: ]
> kaldi::KaldiGetStackTrace()
> kaldi::KaldiErrorMessage::~KaldiErrorMessage()
> kaldi::ArpaFileParser::Read(std::istream&, bool)
> arpa2fst(main+0x678) [0x40f298]
> /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf5) [0x7f68dd512b45]
> arpa2fst() [0x410282]
> ERROR: FstHeader::Read: Bad FST header: standard input
> fstisstochastic /tmp/tmprYKXjl/G.fst
> 0 0
> Succeeded in formatting LM '/home/mbernard/dev/abkhazia/test/language/s5/data/local/language/G.arpa.gz' -> '/tmp/tmprYKXjl/G.fst'

kkm000: @mmmaat: I think I understand where and why this happens, but it would be very helpful if you either attached the failing ARPA file (if it's not very large) to the ticket I opened to this issue, #643, or just posted the directives and ngram counts from the file, as below:
```
$ zcat lm_tgsmall.arpa.gz | grep '^\\\|^ngram'
\data\
ngram 1=200003
ngram 2=1016673
ngram 3=340026
\1-grams:
\2-grams:
\3-grams:
\end\
```

danpovey:
Closing as issue is resolved.

PR #642, opened 2016-04-02
SetLogHandler() to redirect Kaldi loggigng to custom function

kkm000:
Toolkits with libraries generating log messages often implement a way to intercept these messages by allowing the client to register a function to be called when a log message is produced. See examples in [gRPC](https://github.com/grpc/grpc/blob/02f5f8de36924b9fe55ccda800fa465b94bd181d/include/grpc/impl/codegen/log.h#L90) and [protobuf](https://github.com/google/protobuf/blob/b0f661181d10bddc08e380992590a1cdd92be92b/src/google/protobuf/stubs/logging.h#L206). This is a feature sorely missing in Kaldi when it is used as a library in customized clients.

Here's my hack to introduce the mechanism, `SetLogHandler()` API.

WIP because the feature is still lacking unit tests, but otherwise feature-complete, and feedback is welcome. I combined all KaldiLog*Message helper classes into a single class `MessageLogger` with an additional constructor argument for message severity. I also expressed `KaldiAssertFailure_()` via it. I think this is the Right Thing, since Kaldi programs relying on `IsKaldiError()` would not previously count assertion failures as errors; now they do.

### DIFF
--- a/src/base/kaldi-error.cc
+++ b/src/base/kaldi-error.cc
@@ -32,16 +32,16 @@
 #include "base/kaldi-error.h"
 
 namespace kaldi {
-int32 g_kaldi_verbose_level = 0;  // Just initialize this global variable.
+int32 g_kaldi_verbose_level = 0;
 const char *g_program_name = NULL;
+static LogHandler g_log_handler = NULL;
+
 
 // If the program name was set (g_program_name != ""), the function
 // GetProgramName returns the program name (without the path) followed by a
 // colon, e.g. "gmm-align:".  Otherwise it returns the empty string "".
 const char *GetProgramName() {
-  if (g_program_name == NULL) return "";
-  else
-    return g_program_name;
+  return g_program_name == NULL ? "" : g_program_name;
 }
 
 // Given a filename like "/a/b/c/d/e/f.cc",  GetShortFileName
@@ -57,7 +57,6 @@ const char *GetShortFileName(const char *filename) {
     return last_slash;
   }
 }
-
 
 #if defined(HAVE_CXXABI_H) && defined(HAVE_EXECINFO_H)
 // The function name looks like a macro: it's a macro if we don't have ccxxabi.h
@@ -79,7 +78,7 @@ inline void KALDI_APPEND_POSSIBLY_DEMANGLED_STRING(const char *to_append,
 
   char *demangled_name = abi::__cxa_demangle(stripped.c_str(), 0, 0, &status);
 
-  // if status != 0 it is an error (demangling failure),  but not all names seem
+  // if status != 0 it is an error (demangling failure), but not all names seem
   // to demangle, so we don't check it.
 
   if (demangled_name != NULL) {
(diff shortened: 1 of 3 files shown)


src/base/kaldi-error.cc
```diff
@@ -131,70 +130,64 @@ std::string KaldiGetStackTrace() {
 
 void KaldiAssertFailure_(const char *func, const char *file,
                          int32 line, const char *cond_str) {
-  std::ostringstream ss;
-  ss << "KALDI_ASSERT: at " << GetProgramName() << func << ':'
-     << GetShortFileName(file)
-     << ':' << line << ", failed: " << cond_str << '\n';
+  MessageLogger ml(LogMessageEnvelope::Error, func, file, line);
+  ml.stream() << "Asserion failed: " << cond_str;
```
danpovey: typo: asserion

src/base/kaldi-error.cc
```diff
@@ -131,70 +130,64 @@ std::string KaldiGetStackTrace() {
...
-                 << GetShortFileName(file) << ':' << line << ") ";
-}
-
-KaldiLogMessage::~KaldiLogMessage() {
-  std::string str = ss_.str();
-  while (!str.empty() && str[str.length() - 1] == '\n')
-    str.resize(str.length() - 1);
-  fprintf(stderr, "%s\n", str.c_str());
-}
+static void SendToLog(const LogMessageEnvelope& envelope,
```
danpovey: we tend to put the & to the right, next to the variable name, in situations like this...

src/base/kaldi-error.cc
```diff
@@ -131,70 +130,64 @@ std::string KaldiGetStackTrace() {
...
-KaldiErrorMessage::KaldiErrorMessage(const char *func, const char *file,
-                                     int32 line) {
-  this->stream() << "ERROR (" << GetProgramName() << func << "():"
-                 << GetShortFileName(file) << ':' << line << ") ";
+MessageLogger::MessageLogger(LogMessageEnvelope::Severity severity,
+                             const char *func, const char *file, int32 line) {
+  // Obviously, we assume the strings survive the destruction of this object.
+  envelope_.severity = severity;
+  envelope_.func = func;
+  envelope_.file = GetShortFileName(file);  // Pointer inside file.
```
danpovey: maybe single quotes around 'file'

src/base/kaldi-error.h
```diff
@@ -63,60 +66,50 @@ inline int32 GetVerboseLevel() { return g_kaldi_verbose_level; }
...
-class KaldiErrorMessage {
- public:
-  KaldiErrorMessage(const char *func, const char *file, int32 line);
+// Class MessageLogger is invoked from the KALDI_ERR, KALDI_WARN, KALDI_LOG and
+// KALDI_LOG macros. It formats the message, then either prints it to stderr or
+// passes to the log custom handler if provided, then, in case of the error,
+// throws an std::runtime_exception.
+//
+// TODO(kkm): Is this still true? C++ standard is very explicit on cerr safety.
+// Note: we avoid using cerr, due to problems with thread safety. fprintf is
```
danpovey: Could that be C++11-specific?  
It's possible that the problem we encountered was due to making multiple calls to the cerr, like
  cerr << a << b;
which could result in mixing between different threads' calls.  Anyway it's probably safe to leave it using stderr.

danpovey:
This seems reasonable but I'll leave it up for a few days to get comments, and I'll maybe use it myself a bit.

src/base/kaldi-error.h
kkm000: Ah, right, if I'm to believe [this answer](http://stackoverflow.com/a/14637716/1149924) which specifically says "unlike in C++03." I'll update this comment.

src/base/kaldi-error.cc
kkm000: Bummer! I always though it was the other way round. I think it makes sense to mention that in Kaldi coding style. I just checked the google style, and it's ambivalent on the space location with the `*` and `&`, only prohibiting spaces on both sides.

_Oops. I think I just used the word_ ambivalent _for the second time in two days..._

kkm000:
This is still WIP, tests lacking. The original tests were inadequate too, as 2 of my errors that I caught while developing this slipped by them (one was printing "LOG" instead of "WARNING" and another not throwing on error). I think I'll try to improve them while I am working on this

kkm000:
@danpovey: Yes, that's absolutely a great thing to leave that for other people to test and comment. Our unit testing is less than amazing.

danpovey:
@kkm000, is this still a WIP?
I don't think we need to be hung up on unit tests for the error reporting code- improvements requiring Google Test are a separate issue, IMO.  This kind of falls into the "who is to check the checkers?" category.

kkm000:
Not at all. I just stated with unit tests for this particular area because most of my initial coding errors slipped by the existing tests.

vdp:
I wonder if it could be useful in some circumstances, to use both a custom logger **and** the Kaldi's built-in logging..
If so, perhaps the prototype of LogHandler can be changed to:

```
typedef bool (*LogHandler)(const LogMessageEnvelope &envelope,
                           const char *message);
```

, where the meaning of the returned value is "use Kaldi logging too?"

Then we can make a slight change to *SendToLog()*

```
if (g_log_handler != NULL && !g_log_handler(envelope, message))
    return;
```

danpovey:
I don't see why someone couldn't just register a log handler that would do
the normal Kaldi stuff *and* their own stuff.
However, this brings up a question in my mind.  Kaldi is distributed as
source, not as binaries or as a library with headers to be linked against.
Therefore, is it really even necessary to have a function like this?
Surely it would be easier to just make it easy to change the log-handler
code directly.  Now, that might require a slight refactoring to make sure
the logging code is all in one place, e.g. it all passes through a single
function.

Dan


On Sun, Apr 10, 2016 at 12:59 AM, Vassil Panayotov <notifications@github.com

kkm000:
@danpovey: Maintaining changes may not be easy. Currently, I am trying to keep local changes to the minimum. In fact, I am compiling OpenFST without source changes by giving it a few fake trivial .h headers on include path, and selected decoding-time subset of Kaldi with substituted trivial headers to remove multithreading support and pthread dependency. I can only commend the overall layout and structure of Kaldi code--I thought it would be harder that it was. The only pain point was the new `bg` table reader, where the use of pthreads.h escaped `kaldi/thread`, but that does not obviously abstracts.

Tangentially, current logging is so full of copy/paste code with unnecessary variations that I see my change as positive for code readability and hygiene in any case.

@vdp: I believe I understand what you mean, but what would be the practical point?

danpovey:
On Sun, Apr 10, 2016 at 1:57 PM, Kirill Katsnelson <notifications@github.com
> wrote:

> @danpovey <https://github.com/danpovey>: Maintaining changes may not be
> easy. Currently, I am trying to keep local changes to the minimum. In fact,
> I am compiling OpenFST without source changes by giving it a few fake
> trivial .h headers on include path, and selected decoding-time subset of
> Kaldi with substituted trivial headers to remove multithreading support and
> pthread dependency. I can only commend the overall layout and structure of
> Kaldi code--I thought it would be harder that it was. The only pain point
> was the new bg table reader, where the use of pthreads.h escaped
> kaldi/thread, but that does not obviously abstracts.
>
It might be useful to add some kind of #ifdef to kaldi-table-reader-inl.h
to optionally disable the ,bg option, in case users want to get rid of the
pthread dependency.  I was thinking of doing that but didn't get around to
it.

> Tangentially, current logging is so full of copy/paste code with
> unnecessary variations that I see my change as positive for code
> readability and hygiene in any case.
>
Yes, your change is probably for the best.

@vdp <https://github.com/vdp>: I believe I understand what you mean, but
what would be the practical point?

Yes, I think the bool return status isn't really necessary, as someone
could just copy-paste the existing Kaldi logging code and add it to their
own code, to do both at once.

Dan

vdp:
One situation that I was thinking about was when you have, say, a custom logging that sends the messages over a network or whatever, but sometimes for debug purposes you want to see them printed in real time to stderr too. The intent was to avoid duplication of code(which is usually not the right thing to do), but yeah, I agree that in this particular case changing the API to accommodate this would probably be an overkill.

src/base/kaldi-error.h
```diff
@@ -63,60 +66,51 @@ inline int32 GetVerboseLevel() { return g_kaldi_verbose_level; }
...
-class KaldiWarnMessage {
- public:
-  inline std::ostream &stream() { return ss_; }
-  KaldiWarnMessage(const char *func, const char *file, int32 line);
-  ~KaldiWarnMessage();
- private:
-  std::ostringstream ss_;
+/// Log message severity and source location info.
+struct LogMessageEnvelope {
+  // Positive number cast to LogLevel is for verbose logging.
```
danpovey: do you mean LogLevel here in the comment?  maybe remove 'cast to LogLevel'?

kkm000: I removed this sloppy comment and updated the following one (LogLevel was the née of the enum Severity), plus couple other minor changes: https://github.com/kaldi-asr/kaldi/pull/642/commits/408c22d356e21c805be9856f6d4291f9b0eab3de

Squash?

danpovey: sure, squashing makes sense.

On Sun, Apr 17, 2016 at 3:43 PM, Kirill Katsnelson <notifications@github.com

kkm000: Done.

kkm000:
Thanks for the reviews!

PR #647, opened 2016-04-03
Range rebase2

danpovey:
see commit messages... this is about fixes and extensions to table-reader stuff, and supporting matrix ranges in [].

### DIFF
--- a/src/feat/wave-reader.h
+++ b/src/feat/wave-reader.h
@@ -103,6 +103,7 @@ class WaveData {
     data_.Swap(&(other->data_));
     std::swap(samp_freq_, other->samp_freq_);
   }
+
  private:
   static const uint32 kBlockSize = 1024 * 1024;  // Use 1M bytes.
   Matrix<BaseFloat> data_;
@@ -171,6 +172,11 @@ class WaveHolder {
     t_.Swap(&(other->t_));
   }
 
+  bool ExtractRange(WaveHolder *other, const std::string &range) {
+    KALDI_ERR << "ExtractRange is not defined for this type of holder.";
+    return false;
+  }
+
  private:
   T t_;
 };
@@ -215,6 +221,11 @@ class WaveInfoHolder {
 
   void Swap(WaveInfoHolder *other) {
     t_.Swap(&(other->t_));
+  }
+  
+  bool ExtractRange(WaveInfoHolder *other, const std::string &range) {
+    KALDI_ERR << "ExtractRange is not defined for this type of holder.";
+    return false;
   }
  private:
   T t_;
--- a/src/feat/wave-reader.h
+++ b/src/feat/wave-reader.h
@@ -172,7 +172,7 @@ class WaveHolder {
     t_.Swap(&(other->t_));
   }
 
-  bool ExtractRange(WaveHolder *other, const std::string &range) {
+  bool ExtractRange(const WaveHolder &other, const std::string &range) {
     KALDI_ERR << "ExtractRange is not defined for this type of holder.";
     return false;
   }
@@ -222,8 +222,8 @@ class WaveInfoHolder {
   void Swap(WaveInfoHolder *other) {
     t_.Swap(&(other->t_));
   }
-  
-  bool ExtractRange(WaveInfoHolder *other, const std::string &range) {
+
+  bool ExtractRange(const WaveInfoHolder &other, const std::string &range) {
     KALDI_ERR << "ExtractRange is not defined for this type of holder.";
     return false;
   }
--- a/src/util/kaldi-holder.cc
+++ b/src/util/kaldi-holder.cc
@@ -27,11 +27,11 @@ template<class Real>
 bool ExtractObjectRange(const Matrix<Real> &input, const std::string &range,
                         Matrix<Real> *output) {
   if (range.empty()) {
-    KALDI_ERR << "Empty range specifier."; 
+    KALDI_ERR << "Empty range specifier.";
     return false;
   }
   std::vector<std::string> splits;
-  SplitStringToVector(range, ",", false, &splits);  
+  SplitStringToVector(range, ",", false, &splits);
   if (!((splits.size() == 1 && !splits[0].empty()) ||
         (splits.size() == 2  && !splits[0].empty() && !splits[1].empty()))) {
     KALDI_ERR << "Invalid range specifier: " << range;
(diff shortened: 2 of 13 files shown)


danpovey:
I'm not sure if anyone has any comments on this?  This is an extension to the table-reading code so that it will support ranges of matrices, like an scp file with

 a432ab-14312-9876  /foo/bar/mfcc_1.ark:43123[0:59]

I have some misgivings about it though... it seems like a useful feature, but for instance, it breaks certain symmetries (like the second part of the line of an scp filename is always a valid rxfilename), and it requires extra functions in Holder objects.  Ill probably merge it though, as it does seem like it would be useful in certain circumstances, and people have asked about this feature.

src/feat/wave-reader.h
```diff
@@ -171,6 +172,11 @@ class WaveHolder {
     t_.Swap(&(other->t_));
   }
 
+  bool ExtractRange(const WaveHolder &other, const std::string &range) {
+    KALDI_ERR << "ExtractRange is not defined for this type of holder.";
```
kkm000: It seems that all holders except one or two need an exact clone of this function. I would rather use some SFINAE magic (like that in http://stackoverflow.com/a/264088/1149924) at call site, to detect if given holder type has the member. I can draft a patch for that if you think it would be less ugly than this.

src/util/kaldi-holder.cc
```diff
@@ -0,0 +1,79 @@
...
+                        Matrix<Real> *output) {
+  if (range.empty()) {
+    KALDI_ERR << "Empty range specifier.";
+    return false;
+  }
+  std::vector<std::string> splits;
+  SplitStringToVector(range, ",", false, &splits);
+  if (!((splits.size() == 1 && !splits[0].empty()) ||
+        (splits.size() == 2  && !splits[0].empty() && !splits[1].empty()))) {
+    KALDI_ERR << "Invalid range specifier: " << range;
```
kkm000: Does it make sense to support ranges like `[:10]` or `[10:]`, standing for all from beginning/to end respectively?

src/util/kaldi-table-inl.h
```diff
@@ -88,8 +101,10 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
 
   SequentialTableReaderScriptImpl(): state_(kUninitialized) { }
 
+  // you may call Open from states kUninitialized and kError.
```
kkm000: Do we require all comments be capitalized as proper sentences? Google style does, IIRC.

src/feat/wave-reader.h
danpovey: I had a look into it, and I think I'd rather leave it as is.  It feels like the use of hacks like this is working against the grain of the language, and would be hard to debug.

src/util/kaldi-holder.cc
danpovey: Probably, yes-- I'm happy to add them as and when they're needed though (busy with other stuff).

src/util/kaldi-table-inl.h
danpovey: I'll bear that in mind in future.   If I update this patch for something else I'll fix that.

src/util/kaldi-table-inl.h
```diff
@@ -245,104 +275,209 @@ template<class Holder>  class SequentialTableReaderScriptImpl:
...
+      // After extracting "key" from "line", we put the rest
+      // of "line" into "rest", and then extract data_rxfilename_
+      // (e.g. 1.ark:100) and possibly the range_ specifer
+      // (e.g. [1:2,2:10]) from "rest".
+      std::string data_rxfilename, rest;
+      SplitStringOnFirstSpace(line, &key_, &rest);
+      if (!key_.empty() && !rest.empty()) {
         // Got a valid line.
-        state_ = kHaveScpLine;
+        if (rest[rest.size()-1] == ']') {
```
kkm000: Use string::back()?

xiaohui-zhang: hmm it's a feature of c++11...

src/util/kaldi-table.cc
```diff
@@ -22,6 +22,21 @@
...
 namespace kaldi {
 
+bool ExtractRangeSpecifier(const std::string &line,
+                           std::string *data_rxfilename,
+                           std::string *range) {
+  if (line.empty() || line[line.size()-1] != ']')
+    KALDI_ERR << "ExtractRangeRspecifier called wrongly.";
+  std::vector<std::string> splits;
+  SplitStringToVector(line, "[", false, &splits);
+  if (splits.size() == 2 && !splits[0].empty() && !splits[1].empty()) {
```
kkm000: `splits[1].empty()` is tautologically `true` if evaluated, as the last character known to be `]` will be included into last split.
The condition as written allows for empty range spec `[]` -- is it legit?

src/util/kaldi-table.cc
```diff
@@ -22,6 +22,21 @@
 
 namespace kaldi {
 
+bool ExtractRangeSpecifier(const std::string &line,
+                           std::string *data_rxfilename,
+                           std::string *range) {
+  if (line.empty() || line[line.size()-1] != ']')
```
kkm000: Can use `line.back()` for readability.

src/util/kaldi-table-inl.h
kkm000: Oh, my bad. It is.

src/util/kaldi-table.cc
danpovey: I'm changing to size() > 1.

danpovey:
just pushed fixes to a couple of those comments.

src/feat/wave-reader.h
kkm000: Up to you, not trying to convince you at all; I just want to comment on the "against the grain" part. C++11 makes such constructs easier to implement, with library metafunctions like std::enable_if (many borrowed from boost, naturally).

Having just written a massively multithreaded service with non-homogenous interacting threads in pure C++11, I would say that it feels different from the C++ as we knew it before, to the point where it feels like a wholly different language. Personally I liked it, as it certainly makes leaking memory harder and its library multithreading is piece of cake, but not everybody is so amazed. But all in all it seems that the "grain" is shifting its direction these days, as fully compatible C++ compilers become widely available.

kkm000:
LGTM

danpovey:
OK thanks.
I'll merge soon- want to wait just in case there are more comments.

Issue #640, opened 2016-04-01
use-log10 option


danpovey:
I notice that the use-log10 option still exists in the internals of the ARPA-related code.  This shouldn't be there- it was never really needed as ARPAs are always log-10.

kkm000:
Ok, will do.

PR #634, opened 2016-04-01
Updated a few scripts to use new arpa2fst options

kkm000:
I converted a few scripts by hand: LibriSpeech because they are easy to test here (everything is downloaded and set up), WSJ because they are the WSJ, and AMI because they start with the 'A', for the lack of a better explanation. Of these, I ran LibriSpeech to convert and confirmed `utils/validate_lang.pl` succeed. Others are untested (most I cannot test anyway).

There are 39 more occurrences, but they formatted differently so scripting the changes is not easy.(`find -name '*.sh' | sort | xargs grep --color -nP 'arpa2fst -(?!-)'` to find).

Is the goal to convert them all? I can do them by hand a few at a time when I get bored of everything else.

PR #649, opened 2016-04-04
Improved ARPA compilers diagnostics.

kkm000:
- Option `--max-arpa-warnings` to control number of warnings, default 30.
- Much improved error message with both source line number and the line itself.

This addresses the warning flood from arpa2fst noted in https://github.com/kaldi-asr/kaldi/pull/634#issuecomment-204861154

This does not remove the log10 option per #640, to keep the PR smaller. That's coming next.

### DIFF
--- a/src/lm/arpa-file-parser.cc
+++ b/src/lm/arpa-file-parser.cc
@@ -31,7 +31,8 @@ namespace kaldi {
 
 ArpaFileParser::ArpaFileParser(ArpaParseOptions options,
                                fst::SymbolTable* symbols)
-    : options_(options), symbols_(symbols), line_number_(0) {
+    : options_(options), symbols_(symbols),
+      line_number_(0), warning_count_(0) {
 }
 
 ArpaFileParser::~ArpaFileParser() {
@@ -70,49 +71,51 @@ void ArpaFileParser::Read(std::istream &is, bool binary) {
 
   ngram_counts_.clear();
   line_number_ = 0;
+  warning_count_ = 0;
+  current_line_.clear();
 
-#define PARSE_ERR (KALDI_ERR << "in line " << line_number_ << ": ")
+#define PARSE_ERR (KALDI_ERR << LineReference() << ": ")
 
   // Give derived class an opportunity to prepare its state.
   ReadStarted();
 
-  std::string line;
-
   // Processes "\data\" section.
   bool keyword_found = false;
-  while (++line_number_, getline(is, line) && !is.eof()) {
-    if (line.empty()) continue;
+  while (++line_number_, getline(is, current_line_) && !is.eof()) {
+    if (current_line_.empty()) continue;
 
     // Continue skipping lines until the \data\ marker alone on a line is found.
     if (!keyword_found) {
-      if (line == "\\data\\") {
+      if (current_line_ == "\\data\\") {
         KALDI_LOG << "Reading \\data\\ section.";
         keyword_found = true;
       }
       continue;
     }
 
-    if (line[0] == '\\') break;
+    if (current_line_[0] == '\\') break;
 
     // Enters "\data\" section, and looks for patterns like "ngram 1=1000",
     // which means there are 1000 unigrams.
-    std::size_t equal_symbol_pos = line.find("=");
+    std::size_t equal_symbol_pos = current_line_.find("=");
     if (equal_symbol_pos != std::string::npos)
-      line.replace(equal_symbol_pos, 1, " = ");  // Inserts spaces around "="
+      // Guaranteed spaces around the "=".
+      current_line_.replace(equal_symbol_pos, 1, " = ");
     std::vector<std::string> col;
-    SplitStringToVector(line, " \t", true, &col);
+    SplitStringToVector(current_line_, " \t", true, &col);
     if (col.size() == 4 && col[0] == "ngram" && col[2] == "=") {
       int32 order, ngram_count = 0;
       if (!ConvertStringToInteger(col[1], &order) ||
           !ConvertStringToInteger(col[3], &ngram_count)) {
-        PARSE_ERR << "Cannot parse ngram count '" << line << "'.";
+        PARSE_ERR << "cannot parse ngram count";
       }
       if (ngram_counts_.size() <= order) {
         ngram_counts_.resize(order);
       }
       ngram_counts_[order - 1] = ngram_count;
     } else {
-      KALDI_WARN << "Uninterpretable line in \\data\\ section: " << line;
+      KALDI_WARN << LineReference()
+                 << ": uninterpretable line in \\data\\ section";
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/lm/arpa-file-parser.h
```diff
@@ -43,13 +44,17 @@ struct ArpaParseOptions {
 
   ArpaParseOptions()
       : bos_symbol(-1), eos_symbol(-1), unk_symbol(-1),
-        oov_handling(kRaiseError), use_log10(false) { }
+        oov_handling(kRaiseError), use_log10(false),
```
danpovey: now might be a good time to remove the use_log10 option.

src/lm/arpa-file-parser.h
```diff
@@ -86,9 +91,35 @@ class ArpaFileParser {
   void Read(std::istream &is, bool binary);
 
   /// Parser options.
-  const ArpaParseOptions& Options() const { return options_; }
+  const ArpaParseOptions &Options() const { return options_; }
```
danpovey: I should have been more clear-- for some reason the way most of the code is, that in function return types it's to the left, but to the right in declarations, e.g.
const ArpaParseOptions& Options(const ArpaParseOptions &foo);
Not that it really matters, but that's the way the code generally looks.

src/lm/arpa-file-parser.h
```diff
@@ -100,24 +131,28 @@ class ArpaFileParser {
   /// Pure override that must be implemented to process current n-gram. The
   /// n-grams are sent in the file order, which guarantees that all
   /// (k-1)-grams are processed before the first k-gram is.
-  virtual void ConsumeNGram(const NGram&) = 0;
```
danpovey: BTW, I think the Google  style guide forbids consecutive capitals (unless I misunderstand, ???), so it would be
ConsumeNgram.

src/lm/arpa-file-parser.h
kkm000: I'll make a separate PR. It will affect *a lot* of tests. I used it to simplify testing (read -1.23, expect -1.23 in the NGram structure). It seems a small change, but it is big in the number of lines, and I do not want to mix the to, to keep them reviewable.

src/lm/arpa-file-parser.h
kkm000: It actually does--from my experience it is far easier to spot errors in uniform code. I'll change  affected lines in this patch.

src/lm/arpa-file-parser.h
```diff
@@ -86,9 +91,35 @@ class ArpaFileParser {
...
-  const ArpaParseOptions& Options() const { return options_; }
+  const ArpaParseOptions &Options() const { return options_; }
 
  protected:
+  /// A status value returned from ConsumeNGram() overload.
+  class Status {
+   public:
+    /// Return the warninig status to print a warninig and continue.
+    static Status Warning(const std::string& message) {
+      return Status(kWarninig, message); }
```
danpovey: Warninig is misspelled.  
But actually I think I have a bigger problem with this code-- I think writing a special class for error-handling just for parsing ARPA files is just too much.  It feels like meta-programming.  Better to keep things simple.
If something is an error, then use KALDI_ERR.
if something is OK, then do nothing.
If something is a warning, then print a warning, and you can limit the number of warnings using a static namespace-scope variable declared inside the .cc file where this code lives, e.g. set it to 10.  That's how we normally rate-limit warnings.  Kaldi is distributed as source, so in the unlikely event that someone cares about the warning-limit enough to want to change it, they can modify the source.

src/lm/arpa-file-parser.cc
```diff
@@ -29,9 +29,19 @@
 
 namespace kaldi {
 
+void ArpaParseOptions::Register(OptionsItf *opts) {
```
danpovey: We normally define Register inline in the header because it's effectively part of the documentation.
A comment mentioning why not all the members are registered would be good.

src/lm/arpa-file-parser.h
kkm000: Well, I did not do that for the love of classes. The diagnostic was really poor: I could print a line number in ARPA file, but if that comes out of pipeline, good luck finding it. So I need to communicate back from ConsumeNGram what it decided to do with it: accept, accept with a warning or die fatally. Formatting is done in the common class, and is actually useful:for locating the problem:
```
line 21 [-1.5       a β -3.2] skipped: word 'β' not in symbol table.
line 24 [-0.03333        <s> b b b] skipped: no parent (n-1)-gram exists
```
Rate limiting *by itself* is not a problem, the counter is a private member of the ARPA parser class. The question is *what* to count. I would really hate to increment a global variable defined in one file from two others every time I print KALDI_WARN. It's probably even worse.

danpovey: Or perhaps there's some other reason why you're doing this, like giving the warning more context because ProcessNGram (or whatever) doesn't know the line number.  I don't think a huge amount of context (like line numbers) is not really necessary because it's not hard to work out from the ARPA file, and these errors are rare occurrences.  I'd be happier you just using KALDI_ERR and KALDI_WARN (guarded with a counter), called directly where it's needed, rather than this HandleError thing.  That makes the code much easier to debug as the error happens where it originated and all the variables are on the stack; and people familiar with the Kaldi coding style will understand what's going on.

src/lm/arpa-file-parser.h
kkm000: It's [not quite precise](https://google.github.io/styleguide/cppguide.html#Type_Names) on that:

> Type names start with a capital letter and have a capital letter for each new word, with no underscores.

So, strictly speaking, this depends on whether you treat "n-gram" as one word or two. :)

danpovey: OK, it's fine then.

On Mon, Apr 4, 2016 at 12:02 AM, Kirill Katsnelson <notifications@github.com

src/lm/arpa-file-parser.h
danpovey: The global variable is specific to this of ARPA-reading code, which is only
accessed from this one program.  Or having it in the options class or in
the class itself would be fine with me- whatever you prefer.

If you really need the line number, you can make it an additional argument
to the function where it's needed.

That way all the errors and warnings can be printed exactly where they
originate.  I'm not going to be persuaded easily on this issue.


On Sun, Apr 3, 2016 at 11:59 PM, Kirill Katsnelson <notifications@github.com

kkm000: The origin is an issue, I agree. Let me think what I can do to have both good diags and local reporting.

kkm000:
I have rehacked diagnostics. Now the warnings and errors are printed locally, and counted and formatted using two utility methods in the base class:
```c++
  /// Inside ConsumeNGram(), returns a formatted reference to the line being
  /// compiled, to print out as part of diagnostics.
  std::string LineReference() const;

  /// Increments warning count, and returns true if a warning should be
  /// printed or false if the count has exceeded the set maximum.
  bool ShouldWarn();
```
Dan's other comments addressed in 2de0f71.

danpovey:
OK, looks good.   I assume you tested it at least somewhat?

danpovey:
Also, it would be good if you could squash the commits.

kkm000:
Yes, I ran arpa2fst with different warning limits. Tests I wrote for the parser are good, but they do not test stderr output. I really think we should use gTest, it can do even that.

Commits sqaished. Do you feel the log looks neater this way? I have mixed feelings about rewriting history, just wondering what do you think.

danpovey:
OK with me; merge yourself when checks complete and if you tested it somehow.

danpovey:
I heard complaints (well, one complaint) about the kaldi commit log being
hard to read due to too many small commits.  That's why I'm asking people
to squash them.

danpovey:
Actually, merging now.

kkm000:
Thanks for the review, by the way. Compared with my original code, the end result is really more straightforward and lightweight.

danpovey:
cool...

On Mon, Apr 4, 2016 at 11:21 PM, Kirill Katsnelson <notifications@github.com

PR #651, opened 2016-04-04
Saving the aligned ctm, with the wav id.

francisr:
Write the wav id into the aligned CTM, because otherwise there isn't an easy way to know which part is related to which file. 
Save the aligned CTM in the tmp folder of the new data dir, when making the segmentation data dir.

### DIFF
--- a/egs/wsj/s5/steps/cleanup/create_segments_from_ctm.pl
+++ b/egs/wsj/s5/steps/cleanup/create_segments_from_ctm.pl
@@ -325,7 +325,7 @@ sub ProcessWav {
   # Save the aligned CTM if needed
   if(defined($ACT)){
     for (my $i=0; $i<=$#aligned_ctm; $i++) {
-      print $ACT "$aligned_ctm[$i][0] $aligned_ctm[$i][1] ";
+      print $ACT "$wav_id $aligned_ctm[$i][0] $aligned_ctm[$i][1] ";
       print $ACT "$aligned_ctm[$i][2] $aligned_ctm[$i][3]\n";
     }
   }
--- a/egs/wsj/s5/steps/cleanup/make_segmentation_data_dir.sh
+++ b/egs/wsj/s5/steps/cleanup/make_segmentation_data_dir.sh
@@ -164,6 +164,7 @@ steps/cleanup/create_segments_from_ctm.pl \
   --min-sil-length $min_sil_length \
   --separator $separator --special-symbol $special_symbol \
   --wer-cutoff $wer_cutoff \
+  --aligned-ctm-filename $new_data_dir/tmp/aligned_ctm \
   $new_data_dir/tmp/ctm $new_data_dir/tmp/aligned.txt \
   $new_data_dir/segments $new_data_dir/text
 


egs/wsj/s5/steps/cleanup/create_segments_from_ctm.pl
```diff
@@ -325,7 +325,7 @@ sub ProcessWav {
   # Save the aligned CTM if needed
   if(defined($ACT)){
     for (my $i=0; $i<=$#aligned_ctm; $i++) {
```
danpovey: I can't figure out what change is being made here, and this is a deficiency in the original code.  At a place like this it's nice to have an example, like
# print "my_recording1 0.1 0.5 some_word\n"
The CTM format has a well-defined number of fields.  It's not true that this code can be correct, and the old code could have been correct.    Perhaps you and the original author (@vimalmanohar ?) can clarify.

vimalmanohar: @chenguoguo or @ngoel17 are the original authors. aligned_ctm might not have been intended to be in CTM format at all. The CTM format, I think, is <wav-id> <channel-id> <start> <end> <word> <confidence>. The new code here does not put the information in that format. <channel-id> is missing and the ordering of the columns is incorrect.

francisr: It currently prints "<word> <start> <duration> <error>", where error is 1 if the word was right in the decoded ctm and 0 otherwise (or the opposite, can't remember).   
I guess it would be nicer to have this a correctly formatted CTM. I just needed the wav id, so I didn't bother making it right.

danpovey: If it's called CTM it should be a correctly formatted CTM.
It seems the script requires the input to be one of those Kaldi-style CTMs
where instead of "wav-id channel" it has "recording-id 1", or "recording-id
A" (they're equivalent).  This should be documented at the top of the
script.   It can just print out another Kaldi-style ctm, with "recording_id
1" at the start, but the wav_id in the script should be changed to say
recording_id, because that's what it really is.  Remi, could you modify the
pull request like this?

Cc'ing Nagendra in case he was relying on this output for something.

Dan

danpovey:
Closing this - see PR 655.

egs/wsj/s5/steps/cleanup/create_segments_from_ctm.pl
chenguoguo: Sorry just saw this. That part of the code was added by Nagendra and he
originally just wanted to dump some temporary file instead of a real CTM I
think. And yes the script uses Kaldi format CTM which uses recording-id
instead of wav-id (it creates a segment file which uses recording-id's). I
can make a pull request to change the wav_id to recording_id if that's not
already done.

Guoguo

danpovey: Actually on reviewing the calling script I think it may have been using a
real wav-id with a channel.   See my more recent PR for my proposed fix.
I realized after writing the email that there are really only two types of
CTMs in Kaldi: one with utterance-id (and channel = 1), and one with the
wav-id (aka file) and channel.  And I think it's the latter that this one
uses.

Dan

chenguoguo: I recall a segment file has the following fields (at least in some old
scripts?):

"utterance-id recording-id start-time end-time"

In this script, I was creating segment file using:

171   print $SO "$seg_id $wav_id $seg_start $seg_end\n";

So I probably messed it up in the first place. I had a look at the calling
scripts, In wsj/s5/local/run_segmentation.sh, steps/get_ctm.sh is called
with --use-segments=true (default value), and it uses a
"fake" reco2file_and_channel that I create in this
script steps/cleanup/split_long_utterance.sh using the following lines:

# CAVEAT: We are not dealing with channels here. Each channel should have a
# unique file name in wav.scp.
paste -d ' ' <(cut -d ' ' -f 1 $output_dir/wav.scp) \
  <(cut -d ' ' -f 1 $output_dir/wav.scp) | awk '{print $1" "$2" A";}' \
  > $output_dir/reco2file_and_channel

Should we fix this?

Guoguo

PR #652, opened 2016-04-04
Adds gmm-sum-hlda command line tool.

ahhentz:
The gmm-sum-hlda command line tool is used to combine HLDA accumulators
to another accumulator. It is useful in training using a compute
cluster, to avoid copying num_nodes * cpus_per_node by combining the
node's accumulators locally. This way, only num_nodes accumulators need
to be copied.


danpovey:
I agree that this could be useful in some circumstances, but currently none
of our recipes use HLDA, so I don't want to commit this as it would be
"orphan code", not used in any script.  Sorry.
Best to keep it in your own fork.
Dan

PR #654, opened 2016-04-04
Api.ai pretrained chain model example scripts

realill:
New eg on how to decode files using kaldi and pretrained Api.ai model


egs/apiai_decode/s5/README
```diff
@@ -0,0 +1,33 @@
+Example scripts on how to use a pre-trained chain model and kaldi base code to recognize any number of wav files.
+
+IMPORTANT: wav files must be in 16000 S16_LE format.
+
+Pretrained models were released by Api.ai under Creative Commons Attribution-ShareAlike 4.0 International Public License. For more details, visit https://github.com/api-ai/api-ai-english-asr-model
+
+Usage:
+- Ensure kaldi is compiled and this scripts are inside kaldi/egs/<subfolder/ directory
```
danpovey: remove <, maybe? or add >

egs/apiai_decode/s5/download-model.sh
```diff
@@ -0,0 +1,24 @@
...
+if [ ! -d exp ]; then
+	mkdir exp;
+fi;
+
+if [ -d exp/api.ai-model ]; then
+	echo "Found existing model, removing";
+	rm -rf exp/api.ai-model/
+fi
+
+mv api.ai-kaldi-asr-model exp/api.ai-model/ || ( echo "Unable to move model to exp/" && exit 1 )
```
danpovey: i think the trailing slash here is confusing.

egs/apiai_decode/s5/local/score.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
+# Copyright 2012-2014  Johns Hopkins University (Author: Daniel Povey, Yenda Trmal)
+# Apache 2.0
```
danpovey: It would be more straightforward, I think, to make local/score.sh a soft link to (IIRC) ../steps/score_kaldi.sh.  See if that works for you.

egs/apiai_decode/s5/download-model.sh
realill: Which one of two trailing slashes? :) exp/?

danpovey: after exp/api.ai-model/.
IMO, trailing slash is normally used to emphasize when you are moving
something into an already existing directory.

realill: I will make changes based on your review. I also need to remove  local/download-model.sh, which is duplicate file.

egs/apiai_decode/s5/path.sh
```diff
@@ -0,0 +1,4 @@
+export KALDI_ROOT=`pwd`/../../..
```
danpovey: path.sh should look something like this now:
```
export KALDI_ROOT=`pwd`/../../..
[ -f $KALDI_ROOT/tools/env.sh ] && . $KALDI_ROOT/tools/env.sh
export PATH=$PWD/utils/:$KALDI_ROOT/tools/openfst/bin:$PWD:$PATH
[ ! -f $KALDI_ROOT/src/path.sh ] && echo >&2 "The standard file $KALDI_ROOT/src/path.sh is not present -> Exit!" && exit 1
. $KALDI_ROOT/src/path.sh
export LC_ALL=C
```

realill: Changed it.

egs/apiai_decode/s5/README
```diff
@@ -0,0 +1,33 @@
+Example scripts on how to use a pre-trained chain model and kaldi base code to recognize any number of wav files.
```
danpovey: mention the language (English), and say what type of data it is optimized for if you can.

egs/apiai_decode/s5/download-model.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Downlaods Api.ai chain model into exp/api.ai-model/ (will replace one if exists)
+
+DOWNLOAD_URL="https://api.ai/downloads/api.ai-kaldi-asr-model.zip"
+
```
danpovey: I'm a little concerned that this URL you have chosen gives you no capability for increasing version numbers (e.g. after retraining) or adding more languages.  Don't you want to structure it in such a way that it's trivial to extend to newly created models, and other languages?

egs/apiai_decode/s5/recognize-wav.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+# Copyright 2016 Api.ai (Author: Ilya Platonov)
+# Apache 2.0
+
+# This script demonstrates kaldi decoding using pretrained model. It will decode list of wav files.
+#
+# IMPORTANT: wav files must be in 16000 S16_LE format.
```
danpovey: i think we would normally describe this as 16kHz, 16 bit little-endian.

egs/apiai_decode/s5/download-model.sh
realill: If we will go into versioning, then https://api.ai/downloads/api.ai-kaldi-asr-model.zip will be latest version (link), and other versions will have numbers at the end.

Also we will track everything here https://github.com/api-ai/api-ai-english-asr-model

danpovey: OK, well eventually it will make sense to update the README to mention
this, but I guess we can do that later on.

realill: Interesting idea is to add alternative models into this script. Like Librispeech. Maybe having some models repo of some sort, if kaldi will be getting more popular. Just random thought.

danpovey:
OK, I think it's just about ready to merge, but could you please squash the commits?

danpovey:
OK, merging.   Thanks.  You can mention it on the list if you want.

PR #656, opened 2016-04-05
Kill arpa2fst use_log10 option (#640)

### LINKED ISSUE #640: use-log10 option

kkm000:
Closes #640.


danpovey:
Thanks!  Merging.

PR #660, opened 2016-04-05
ami+chain: bug-fix in scoring, improves results for chain models; ad…

vijayaditya:
…ript) which improves results.

I also started using sub-shells to generate ctms in parallel for various LMWTs.

@vesis84 you might want to review this commit, as this branch will be later merged into master.

### DIFF
--- a/egs/ami/s5/local/score.sh
+++ b/egs/ami/s5/local/score.sh
@@ -13,6 +13,7 @@ cmd=run.pl
 stage=0
 min_lmwt=9 # unused,
 max_lmwt=15 # unused,
+iter=final
 asclite=true
 #end configuration section.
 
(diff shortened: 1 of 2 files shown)


danpovey:
Looks straightforward -> merging

vesis84:
Hi, a quick question. Why do you handle the "frame-shift" inside the scoring script? Now you have to add this into all the 'local/score.sh' scripts. Wouldn't it be better to handle frame-shift on the input of the decoder in the 'steps/nnetX/decode.sh'? (in nnet1 it is done like that)

vesis84:
The subshells look okay, even though it is not consistent with the other scoring scripts... Definitely you must have had a good reason to do it that way...

danpovey:
The frame-shift means the number of milliseconds between one frame and the next-- it's needed because the output of the decoder is at 1/3 the frame rate, and the scoring script needs to know about it because it affects the times in ctm generation and sclite is sensitive to the times.

vesis84:
Aha, I see, so it is something different from the "delayed targets in LSTM training". Thanks for explaining.

PR #666, opened 2016-04-06
Online nnet3 code

realill:


danpovey:
Thanks.
I notice there are no example scripts.  For the nnet2 setup we have scripts to create the config files, etc., and of course example scripts for specific setups.  Do you have any of those?
Dan

egs/wsj/s5/steps/online/nnet3/decode.sh
```diff
@@ -0,0 +1,158 @@
...
+
+echo "$0 $@"  # Print the command line for logging
+
+[ -f ./path.sh ] && . ./path.sh; # source the path.
+. parse_options.sh || exit 1;
+
+if [ $# != 3 ]; then
+   echo "Usage: $0 [options] <graph-dir> <data-dir> <decode-dir>"
+   echo "... where <decode-dir> is assumed to be a sub-directory of the directory"
+   echo " where the models are, as prepared by steps/online/nnet2/prepare_online_decoding.sh"
```
danpovey: check for references to nnet2.  and try to make sure the nnet3 versions of the scripts exist.

realill:
The only example I have is here in Online section 
https://github.com/kaldi-asr/kaldi/blob/chain/egs/apiai_decode/s5/README

I changed decode.sh a bit, now it has online_config option, which by default is MODEL_DIR/online.conf . This is what MODEL_DIR/conf/online_nnet2_decoding.conf used to be

danpovey:
OK.  In order to support iVectors, we should make a script to prepare the
config file, similar to what we do in nnet2.
@xiaohui-zhang, do you have time to port to prepare-online-decoding stuff
from nnet2 to nnet3, on top of this PR, and try to add an example script?

egs/wsj/s5/steps/online/nnet3/decode.sh
realill: I can copy/paste nnet2 script, maybe change it a bit to produce online.conf instead of conf/online_nnet2_decoding.conf. I can not test it on ivectors.

Still maybe nnet2/nnet3 script should be just one script.

Also I always did not like how   steps/online/nnet2/prepare_online_decoding.sh works because of it producing absolute paths. You can not move model and reuse in other place after that.

There should be some sort of standardization on how models are bundled, so they can be reused from anywhere.

danpovey: > In egs/wsj/s5/steps/online/nnet3/decode.sh
> <https://github.com/kaldi-asr/kaldi/pull/666#discussion_r58906478>:
>
> > +max_state_duration=40 # This only has an effect if you are doing silence
> > +  # weighting.  This default is probably reasonable.  transition-ids repeated
> > +  # more than this many times in an alignment are treated as silence.
> > +iter=final
> > +online_config=
> > +# End configuration section.
> > +
> > +echo "$0 $@"  # Print the command line for logging
> > +
> > +[ -f ./path.sh ] && . ./path.sh; # source the path.
> > +. parse_options.sh || exit 1;
> > +
> > +if [ $# != 3 ]; then
> > +   echo "Usage: $0 [options] <graph-dir> <data-dir> <decode-dir>"
> > +   echo "... where <decode-dir> is assumed to be a sub-directory of the directory"
> > +   echo " where the models are, as prepared by steps/online/nnet2/prepare_online_decoding.sh"
>
> I can copy/paste nnet2 script, maybe change it a bit to produce
> online.conf instead of conf/online_nnet2_decoding.conf. I can not test it
> on ivectors.
>
OK, we can test that.

> Still maybe nnet2/nnet3 script should be just one script.
>
I prefer the nnet3 stuff to be self-contained without unnecessary
dependency on the nnet2 directory.

> Also I always did not like how
> steps/online/nnet2/prepare_online_decoding.sh works because of it producing
> absolute paths. You can not move model and reuse in other place after that.
>
> There should be some sort of standardization on how models are bundled, so
> they can be reused from anywhere.
>
You're right, it's is not ideal, but let's mostly copy it for now-- this
probably isn't the time for a big rewrite.

Dan

realill: Ok I will do this script. I am a bit busy today, so maybe tomorrow.

I want to point out that I reused OnlineNnet2FeaturePipeline directly in this code, which probably should be changed if you want to keep self-containment. Also if I remember this correctly you mention that you want to change features by using several ivectors ...

danpovey: Thanks.

We can leave it using the OnlineNnet2FeaturePipeline for now- it would be a
lot of code to duplicate.

realill: Is it ok to use steps/score_kaldi.sh in decode.sh instead of local/score.sh ?

danpovey: No, because in some setups it's required to score with NIST's sclite tool.
If the --iter option is not required (e.g. if it will be 'final'), just
don't pass in the option.
Dan

realill:
Adding prepare_online_decoding.sh script, also changing apiai_decode to be in sync with changes.

danpovey:
@vimalmanohar, I am quite busy and reviewing this will take me a long time... do you have time to do an initial pass of review?  What I am looking for is:
  - obvious bugs and stylistic/policy-type issues where it's not compliant
  - functionality: does it have basically the same functionality as the old nnet2 version, including iVector support.
  - completeness: are all the scripts that are needed to run this, present?  
We'll also need an example script on one of the standard corpora- even RM would be enough, it doesn't matter, this is to verify the code works.  

@pegahgh, could you please work on an example script for this?  It can be on any setup; it could be based on some nnet2 script.  Actually, there are two separate types of example script that we need for this: a non-chain setup, and a chain script.  Both can be with TDNNs, because BLSTM decoding raises a bunch of complicated issues that I don't want to deal with, and this framework may not really be sufficient- we'd have to do some coding to properly support BLSTMs.  You could put it at the end of some existing training scripts, as a separate final stage.

vimalmanohar:
Ok.  I will do an initial review tonight.

src/nnet3/online-nnet3-decodable.cc
```diff
@@ -0,0 +1,210 @@
+// nnet3/online-nnet3-decodable.cc
```
vimalmanohar: This file must be renamed to something like online-nnet3-decodable-simple.cc

src/nnet3/online-nnet3-decodable.cc
```diff
@@ -0,0 +1,210 @@
...
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "nnet3/online-nnet3-decodable.h"
+#include "nnet3/nnet-utils.h"
+
+namespace kaldi {
+namespace nnet3 {
+
+DecodableNnet3Online::DecodableNnet3Online(
```
vimalmanohar: To be similar to the code in https://github.com/api-ai/kaldi/blob/chain/src/nnet3/nnet-am-decodable-simple.cc, this class needs to be separated into a base class (e.g. OnlineNnet3DecodableBase) with only nnet and priors objects but not am_nnet and trans_model, and a child (e.g. OnlineNnet3DecodableAmNnetSimple) that has the am_nnet and trans_model objects. This might be required to create an nnet3 version of https://github.com/api-ai/kaldi/blob/75f8b027a2e6b6b679d26d0b4b667cea631ee6d1/src/online2bin/online2-wav-nnet2-am-compute.cc.

danpovey: Actually I don't think that structure really necessary in this case, unless
there is some compelling reason for it (which I can't think of now).

BTW, the reason for Vimal's previous comment, regarding the "-simple" in
the name, is that we expect a certain structure of nnet, with inputs
"input" and possibly "ivector", and an output called "output".  This should
be checked using the IsSimpleNnet() function.
Dan

pegahgh:
Sure, I will do that.

src/nnet3/online-nnet3-decodable.cc
realill: I am all for code reuse, and I think NnetDecodableBase could be reused in both offline and online version as a base class, but it will probably require refactoring all three classes. DoNnetComputation method pretty much copy-paste. Still I am not in position to do this kind of refactoring.

So required changes for now are: change a name to online-nnet3-decodable-simple and add IsSimpleNnet() check. I will wait for more requirements and proceed with changes.

src/nnet3/online-nnet3-decodable.cc
```diff
@@ -0,0 +1,210 @@
...
+    input_frame_begin = subsampled_frame * subsample;
+  int32 max_possible_input_frame_end = features_ready /* - ( features_ready - right_context_) % subsample */;
+  if (input_finished && opts_.pad_input)
+    max_possible_input_frame_end += right_context_;
+  int32 input_frame_end = std::min<int32>(max_possible_input_frame_end,
+                                          input_frame_begin +
+                                          left_context_ + right_context_ +
+                                          opts_.max_nnet_batch_size);
+  KALDI_ASSERT(input_frame_end > input_frame_begin);
+  Matrix<BaseFloat> features(input_frame_end - input_frame_begin,
```
realill: Also my general concern about decoders is that, they tend not to reuse input/output buffers. Maybe this is point for optimization here. @danpovey what do you think?

src/online2bin/online2-wav-nnet3-latgen-faster.cc
```diff
@@ -0,0 +1,288 @@
...
+    OnlineNnet2FeaturePipelineInfo feature_info(feature_config);
+
+    if (!online) {
+      feature_info.ivector_extractor_info.use_most_recent_ivector = true;
+      feature_info.ivector_extractor_info.greedy_ivector_extractor = true;
+      chunk_length_secs = -1.0;
+    }
+    
+    TransitionModel trans_model;
+    nnet3::AmNnetSimple nnet;
```
vimalmanohar: For consistency with rest of nnet3 code, this variable and its derivatives could be renamed am_nnet.

src/nnet3/online-nnet3-decodable.h
```diff
@@ -0,0 +1,152 @@
...
+    opts->Register("pad-input", &pad_input,
+                   "If true, pad acoustic features with required acoustic context "
+                   "past edges of file.");
+    opts->Register("max-nnet-batch-size", &max_nnet_batch_size,
+                   "Maximum batch size we use in neural-network decodable object, "
+                   "in cases where we are not constrained by currently available "
+                   "frames (this will rarely make a difference)");
+
+    opts->Register("frame-subsampling-factor", &frame_subsampling_factor,
+                       "Required if the frame-rate of the output (e.g. in 'chain' "
```
vimalmanohar: You could remove the extra indentation.

src/nnet3/online-nnet3-decodable.cc
danpovey: Regarding re-use of buffers, etc.-- the input features are very small
relative to the size of the computation itself, and I don't like to
optimize things until it's proven that they are significant.

vimalmanohar:
I have reviewed the changes. I think it looks fine other than what I already pointed out.

src/nnet3/online-nnet3-decodable.cc
```diff
@@ -0,0 +1,210 @@
...
+
+int32 DecodableNnet3Online::NumFramesReady() const {
+  int32 features_ready = features_->NumFramesReady();
+  if (features_ready == 0)
+    return 0;
+  bool input_finished = features_->IsLastFrame(features_ready - 1);
+  if (opts_.pad_input) {
+    // normal case... we'll pad with duplicates of first + last frame to get the
+    // required left and right context.
+    if (input_finished) return subsampling(features_ready);
```
realill: What about subsampling method name not strarting with capital S? :)

danpovey: Yes, it should start with a capital.  Also the name is not very
explanatory.  Should be called something like NumSubsampledFrames() -- or
just use an inline expression if it's simple.

realill: I created it because I was trying to use (num + frame_subsampling_factor  - 1)/(  frame_subsampling_factor) formula at first. It should stay, it makes code more readable. I will rename it.

realill:
I updated code.

There are few places where AmNnetSimple instances are called nnet or model, I can rename all of them into am_nnet

danpovey:
yes, rename all of them.

realill:
Done with renaming and other stuff.

danpovey:
Thanks.  Before merging I want to wait on what @pegahgh is doing, i.e. getting some example scripts.  My hope is to merge both the PRs at around the same time.  If there are any issues in this code I am hoping that Pegah's work will expose them.

pegahgh:
Hi Dan
I did online decoding on swbd and I got huge wer degradation in nnet3
online setup(wer is abound 40%). I am still looking to find out to find out
the problem.

realill:
@pegahgh what kind of models did you use? chain/ivectors?

danpovey:
What type of models- was it tdnn or BLSTM?  This is not expected to work
for BLSTM.
Run with higher verbose level and make sure it's giving reasonable objf
improvement for the iVectors.  I assume the setup has iVectors.
Dan

pegahgh:
I used chain+ivectors.

pegahgh:
I used chain + tdnn(6v) + ivector.

danpovey:
Vijay may be able to help debug if he is there.

realill:
@pegahgh  Try out apiai_decode scripts model with something like this
online2-wav-nnet3-latgen-faster --frame-subsampling-factor=3   --minimize=false --max-active=7000 --min-active=200 --beam=15.0 --lattice-beam=8.0   --acoustic-scale=1.0   --word-symbol-table=exp/api.ai-model/words.txt   --mfcc-config=exp/api.ai-model/mfcc.conf   exp/api.ai-model/final.mdl exp/api.ai-model/HCLG.fst   ark:data/test-corpus/spk2utt   scp:data/test-corpus/wav.scp   ark:out.lat

pegahgh:
I think I found the problem. I will rerun the decoding.

pegahgh:
I tested online nnet3 setup with 2 systems.
1. no ivector system => The offline and online decoding performance are the
same.

2. with ivector system => The online system performance degraded a lot (30%
degradation!). I checked the ivector extraction parameter and extractor and
it is the same as offline setup. I think there should be a bug in
online-nnet3-decodable-simple.* code.

danpovey:
Pegah, please run with a higher verbose level and see what objf
improvements it's reporting; it should be in the range 5 to 10.
@vimalmanohar, can you please see if you can find any bug in the iVector
related code?  In particular check that it's getting the iVector from a
suitable frame.
Dan

realill:
ivectors support was added by @jcsilva maybe he can find the problem.

pegahgh:
I ran it with higher verbose and the obj improvement range for i-vector is
7 to 13.

danpovey:
OK, then you and vimal check the code and see if it's getting the iVector
from a suitable frame (e.g. not always 0) and see that in  multi-utterance
decoding it's suitable carrying the adaptation-state forward, as in the
nnet2 code.

src/nnet3/online-nnet3-decodable-simple.cc
```diff
@@ -0,0 +1,210 @@
...
+  // MFCCs in the left chunk
+  SubMatrix<BaseFloat> mfcc_mat = features.ColRange(0,mfcc_dim);
+  
+  Vector<BaseFloat> input_ivector;
+  if(ivector_dim != -1){
+    // iVectors in the right chunk
+    SubMatrix<BaseFloat> ivector_mat = features.ColRange(mfcc_dim,ivector_dim);
+    // Get last ivector... not sure if GetCurrentIvector is needed in the online context
+    // I think it should work fine just getting the last row for testing
+    input_ivector = ivector_mat.Row(ivector_mat.NumRows() - 1);
```
vimalmanohar: @danpovey @pegahgh 
Could this be the issue? May be we should use the ivector for the current frame instead of the last one. We will check this out.

vimalmanohar:
The adaptation-state stuff is here too. The feature pipeline used here is
that same as that in online nnet2 setup.

On Tue, Apr 12, 2016 at 5:49 PM Daniel Povey <notifications@github.com>
wrote:

> OK, then you and vimal check the code and see if it's getting the iVector
> from a suitable frame (e.g. not always 0) and see that in multi-utterance
> decoding it's suitable carrying the adaptation-state forward, as in the
> nnet2 code.
>
>
> On Tue, Apr 12, 2016 at 2:47 PM, pegahgh <notifications@github.com> wrote:
>
> > I ran it with higher verbose and the obj improvement range for i-vector
> is
> > 7 to 13.
> >
> > On Tue, Apr 12, 2016 at 5:05 PM, Daniel Povey <notifications@github.com>
> > wrote:
> >
> > > Pegah, please run with a higher verbose level and see what objf
> > > improvements it's reporting; it should be in the range 5 to 10.
> > > @vimalmanohar, can you please see if you can find any bug in the
> iVector
> > > related code? In particular check that it's getting the iVector from a
> > > suitable frame.
> > > Dan
> > >
> > >
> > >
> > > On Tue, Apr 12, 2016 at 2:02 PM, pegahgh <notifications@github.com>
> > wrote:
> > >
> > > > I tested online nnet3 setup with 2 systems.
> > > > 1. no ivector system => The offline and online decoding performance
> are
> > > the
> > > > same.
> > > >
> > > > 2. with ivector system => The online system performance degraded a
> lot
> > > (30%
> > > > degradation!). I checked the ivector extraction parameter and
> extractor
> > > and
> > > > it is the same as offline setup. I think there should be a bug in
> > > > online-nnet3-decodable-simple.* code.
> > > >
> > > >
> > > >
> > > > On Tue, Apr 12, 2016 at 1:20 PM, Pegah Ghahrmani <pegahgh@gmail.com>
> > > > wrote:
> > > >
> > > > > I think I found the problem. I will rerun the decoding.
> > > > >
> > > > >
> > > > > On Tue, Apr 12, 2016 at 1:08 PM, Ilya Platonov <
> > > notifications@github.com
> > > > >
> > > > > wrote:
> > > > >
> > > > >> @pegahgh <https://github.com/pegahgh> Try out apiai_decode
> scripts
> > > > model
> > > > >> with something like this
> > > > >> online2-wav-nnet3-latgen-faster --frame-subsampling-factor=3
> > > > >> --minimize=false --max-active=7000 --min-active=200 --beam=15.0
> > > > >> --lattice-beam=8.0 --acoustic-scale=1.0
> > > > >> --word-symbol-table=exp/api.ai-model/words.txt
> > > > >> --mfcc-config=exp/api.ai-model/mfcc.conf
> exp/api.ai-model/final.mdl
> > > > >> exp/api.ai-model/HCLG.fst ark:data/test-corpus/spk2utt
> > > > >> scp:data/test-corpus/wav.scp ark:out.lat
> > > > >>
> > > > >> —
> > > > >> You are receiving this because you were mentioned.
> > > > >> Reply to this email directly or view it on GitHub
> > > > >> <
> https://github.com/kaldi-asr/kaldi/pull/666#issuecomment-209010538
> > >
> > > > >>
> > > > >
> > > > >
> > > >
> > > > —
> > > > You are receiving this because you were mentioned.
> > > > Reply to this email directly or view it on GitHub
> > > > <https://github.com/kaldi-asr/kaldi/pull/666#issuecomment-209100545>
> > >
> > > >
> > >
> > > —
> > > You are receiving this because you were mentioned.
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/kaldi-asr/kaldi/pull/666#issuecomment-209101687>
> > >
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <https://github.com/kaldi-asr/kaldi/pull/666#issuecomment-209116682>
> >
>
> —
> <https://github.com/kaldi-asr/kaldi/pull/666#issuecomment-209117022>
>
-- 
Vimal Manohar
PhD Student
Electrical & Computer Engineering
Johns Hopkins University

danpovey:
Getting it for the most recently available frame will tend to improve
results, not degrade them.
And I think it's what is done in the nnet2 decoding framework.

danpovey:
BTW, there are missing spaces after commas, e.g. in

 features.ColRange(mfcc_dim,ivector_dim);

src/nnet3/online-nnet3-decodable-simple.cc
```diff
@@ -0,0 +1,210 @@
...
+  // Maybe a new "nnet3 feature pipeline"?
+  int32 mfcc_dim = am_nnet_.GetNnet().InputDim("input");
+  int32 ivector_dim = am_nnet_.GetNnet().InputDim("ivector");
+  // MFCCs in the left chunk
+  SubMatrix<BaseFloat> mfcc_mat = features.ColRange(0,mfcc_dim);
+  
+  Vector<BaseFloat> input_ivector;
+  if(ivector_dim != -1){
+    // iVectors in the right chunk
+    SubMatrix<BaseFloat> ivector_mat = features.ColRange(mfcc_dim,ivector_dim);
```
danpovey: this code should check that the dim of the features equals mfcc_dim + ivector_dim- this might catch certain model/feature mismatches.

src/nnet3/online-nnet3-decodable-simple.cc
```diff
@@ -0,0 +1,210 @@
...
+
+  int32 num_subsampled_frames = NumSubsampledFrames(input_frame_end - input_frame_begin -
+          left_context_ - right_context_);
+  // I'm not checking if the input feature vector is ok.
+  // It should be done, but I'm not sure if it is the best place. 
+  // Maybe a new "nnet3 feature pipeline"?
+  int32 mfcc_dim = am_nnet_.GetNnet().InputDim("input");
+  int32 ivector_dim = am_nnet_.GetNnet().InputDim("ivector");
+  // MFCCs in the left chunk
+  SubMatrix<BaseFloat> mfcc_mat = features.ColRange(0,mfcc_dim);
```
danpovey: this should not use the = syntax but the constructor () syntax because if it were assignment, the code would not work.
 (probably C++ automatically makes it construction but let's be explicit).

src/nnet3/online-nnet3-decodable-simple.cc
danpovey: Same here- use constructor () syntax.

src/nnet3/online-nnet3-decodable-simple.cc
```diff
@@ -0,0 +1,210 @@
...
+  computer.GetOutputDestructive("output", &cu_output);
+  // subtract log-prior (divide by prior)
+  if (log_priors_.Dim() != 0)
+    cu_output.AddVecToRows(-1.0, log_priors_);
+  // apply the acoustic scale
+  cu_output.Scale(opts_.acoustic_scale);
+  scaled_loglikes_.Resize(0, 0);
+  // the following statement just swaps the pointers if we're not using a GPU.
+  cu_output.Swap(&scaled_loglikes_);
+//  current_log_post_subsampled_offset_ = output_t_start / subsample;
```
danpovey: remove this comment.

egs/wsj/s5/steps/online/nnet3/decode.sh
```diff
@@ -0,0 +1,154 @@
...
+fi
+
+if [ $stage -le 0 ]; then
+  $cmd $parallel_opts JOB=1:$nj $dir/log/decode.JOB.log \
+    $decoder $opts $silence_weighting_opts --do-endpointing=$do_endpointing $frame_subsampling_opt \
+     --config=$online_config \
+     --max-active=$max_active --beam=$beam --lattice-beam=$lattice_beam \
+     --acoustic-scale=$acwt --word-symbol-table=$graphdir/words.txt \
+     $srcdir/${iter}.mdl $graphdir/HCLG.fst $spk2utt_rspecifier "$wav_rspecifier" \
+      $lat_wspecifier || exit 1;
```
danpovey: lat_wspecifier needs to be in double quotes here.

danpovey:
Pegah, I looked at where you were testing this, and it looks like the non-online decoding is even worse:

pwd
/export/a09/pegahgh/kaldi-jesus-git/egs/swbd/s5c
a03:s5c: grep Sum exp/chain_online/tdnn_6v_sp/decode_eval2000_sw1_tg/score*/*ys | utils/best_wer.sh 
%WER 36.1 | 1831 21395 | 65.4 13.1 21.5 1.5 36.1 66.5 | exp/chain_online/tdnn_6v_sp/decode_eval2000_sw1_tg/score_9_0.5/eval2000_hires.ctm.swbd.filt.sys
a03:s5c: grep Sum exp/chain_online/tdnn_6v_sp_online/decode_eval2000_sw1_tg/score*/*ys | utils/best_wer.sh 
tdnn_6v_sp_online/     tdnn_6v_sp_online_old/ 
a03:s5c: grep Sum exp/chain_online/tdnn_6v_sp_online/decode_eval2000_sw1_tg/score*/*ys | utils/best_wer.sh 
%WER 35.4 | 1831 21395 | 66.0 11.9 22.1 1.4 35.4 66.8 | exp/chain_online/tdnn_6v_sp_online/decode_eval2000_sw1_tg/score_10_0.0/eval2000_hires.ctm.swbd.filt.sys

So it could be an unrelated issue, like you forgot the --self-loop-scale 1.0 scale when building the graph, or you used a bad model, or some other mismatch.

Be careful that the frame_subsampling_factor  gets copied over into the online directory when you do prepare_online_features.sh-- this affects the scoring.  @realill, if you take care of these fixes it would be great, as you control this branch.

pegahgh:
Dan
I told you that exp/chain_online/tdnn_6v_sp uses online ivector extracted
using onlin2-wav-dump-features and select-feats. The offline results is in
exp/chain_online/tdnn_6v_sp/old_decodes

danpovey:
Also @pegahgh, I see that you made various fixes/changes to the online-decoding script.  Can you please communicate them to Ilya somehow?

danpovey:
So where is the command that you used to dump those iVectors?

realill:
@danpovey finalize requirements and I will do this changes. I am switching to other projects right now, so it may take few days.

jcsilva:
When we tested this code with iVectors, we have not seen that huge degradation in WER.

We ran it again today, using the following syntax for online recognition :

online2-wav-nnet3-latgen-faster --verbose=3 --online=true --ivector-extraction-config=exp/online/nnet_online_alt_mixed_sp/conf/ivector_extractor.conf --max-active=7000 --beam=15.0 --lattice-beam=8.0 --acoustic-scale=1.0 --frame-subsampling-factor=$(cat exp/chain/tdnn_6z_sp/frame_subsampling_factor) --feature-type=mfcc --mfcc-config=exp/online/nnet_online_alt_mixed_sp/conf/mfcc.conf exp/chain/tdnn_6z_sp/final.mdl exp/chain/tdnn_6z_sp/graph_smallest/HCLG.fst ark:data/data8k_hires/spk2utt scp:data/data8k_hires/wav.scp ark:- 2>online.log | lattice-scale --acoustic-scale=10 ark:- ark:- | gzip -c > online/lat.1.gz; echo 1 > exp/online/num_jobs

And the reported online WER was 20.14%.

We also tried offline recognition:

online2-wav-nnet3-latgen-faster --verbose=3 --online=false --ivector-extraction-config=exp/online/nnet_online_alt_mixed_sp/conf/ivector_extractor.conf --max-active=7000 --beam=15.0 --lattice-beam=8.0 --acoustic-scale=1.0 --frame-subsampling-factor=$(cat exp/chain/tdnn_6z_sp/frame_subsampling_factor) --feature-type=mfcc --mfcc-config=exp/online/nnet_online_alt_mixed_sp/conf/mfcc.conf exp/chain/tdnn_6z_sp/final.mdl exp/chain/tdnn_6z_sp/graph_smallest/HCLG.fst ark:data/data8k_hires/spk2utt scp:data/data8k_hires/wav.scp ark:- 2>offline.log | lattice-scale --acoustic-scale=10 ark:- ark:- | gzip -c > offline/lat.1.gz; echo 1 > offline/num_jobs

The offline WER was 19.57%.

pegahgh:
The bug fixed and the the online setup with ivector works as same as online
setup with ivector.

realill:
Ok, I will go through @danpovey code review notes, to finish this up :+1:

realill:
I merged @pegahgh changes are merged and I did @danpovey review changes. I retested it on api.ai mode. 

Hopefully it is good to go now :)

egs/swbd/s5c/conf/decode_online.config
```diff
@@ -0,0 +1,2 @@
+beam=15.0 # beam for decoding.  Was 13.0 in the scripts.
```
danpovey: It looks like you may have copied this from somewhere like WSJ.  This beam is too large- try setting it to 11.0 which is the one used in conf/decode.config. @pegahgh, please re-test the online decoding, this may affect the WERs.

egs/swbd/s5c/local/chain/run_tdnn_6v.sh
```diff
@@ -223,5 +223,54 @@ if [ $stage -le 14 ]; then
...
+  done
+fi
+
+if [ $stage -le 17 ]; then
+  iter_opts=
+  if [ ! -z $decode_iter ]; then
+    iter_opts=" --iter $decode_iter "
+  fi
+  for decode_set in train_dev eval2000; do
+      (
```
danpovey: I think you can comment out this whole section in this file, because others probably won't want or need to run it, and just make a note that this was used to verify that online and offline decoding gave the same results.  Please make sure the results are recorded somewhere-- they can be in this file, for instance.

egs/wsj/s5/steps/online/nnet3/decode.sh
```diff
@@ -0,0 +1,154 @@
...
+# Copyright 2014  Johns Hopkins University (Author: Daniel Povey)
+#           2016  Api.ai (Author: Ilya Platonov)
+# Apache 2.0
+
+# Begin configuration section.  
+stage=0
+nj=4
+cmd=run.pl
+max_active=7000
+threaded=false
```
danpovey: please remove this option and the modify_ivector_config option, as they are not relevant any more.

realill: Done, I assume egs/swbd/s5c/local/chain/run_tdnn_6v.sh should be done by @pegahgh , since I am bit out of context here.

pegahgh: Done!

realill: Merged.

src/online2/online-nnet3-decoding.h
```diff
@@ -0,0 +1,132 @@
...
+
+
+#ifndef KALDI_ONLINE2_ONLINE_NNET3_DECODING_H_
+#define KALDI_ONLINE2_ONLINE_NNET3_DECODING_H_
+
+#include <string>
+#include <vector>
+#include <deque>
+
+#include "../nnet3/online-nnet3-decodable-simple.h"
```
danpovey: There should be no ../ here.  Make sure that this mistake doesn't happen elsewere.

src/online2/online-nnet3-decoding.h
```diff
@@ -0,0 +1,132 @@
+// online2/online-nnet3-decoding.h
+
+// Copyright 2014  Johns Hopkins University (author: Daniel Povey)
```
danpovey: This file is in the wrong directory-- it should be in nnet3, as it relates to nnet3.  Having nnet3/ depend on nnet2/ is undesirable, but having nnet2/ depend on nnet3/ is really weird.

realill: online-nnet3-decoding.h is in online2 because online-nnet2-decoding.h is online2 as well.

If I move this, then nnet3 will depend on online2, is this ok?

danpovey: oh sorry I misunderstood.  I thought it was in nnet2.

src/online2/online-nnet3-decoding.h
realill: FYI: 
$ grep "\.\." -R . | grep "#include"
./fstext/fstext-utils.h:#include "../base/kaldi-common.h" // for error reporting macros.
./fstext/fstext-utils.h:#include "../util/text-utils.h" // for SplitStringToVector

danpovey: thanks- fixing that.

danpovey:
OK, merging this.  Thanks, Ilya and Pegah!   @pegahgh, can you please re-test one of your decodings after I merge to make sure it's still working?

danpovey:
and @pegahgh, note that this is merged to the chain branch... I may merge to master after you test agaih.

xiaosdawn:
Hi, all. If I have a relatively large acoustic model trained from standard materials, say broadcast news recordings, and a small model trained from recording of specific fields, i.e. telephone speech or even speech control for machines. Is there a way to combine and adapt them to generate a new model? If so, is there a recipe for me to start with? or someone could give me a start guide?
Any advises or ideas will be appreciate! Thanks.

danpovey:
No, you can't combine trained models; you'd have to train a model on a
mixture of the data.

PR #667, opened 2016-04-07
[WIP] Simulated RIR reverberation in SWB

tomkocse:


egs/swbd/s5c/local/reverberate/IR_stats.m
```diff
@@ -0,0 +1,348 @@
+function [rt,drr,cte,cfs,edt] = IR_stats(x, fs, varargin)
+%function [rt,drr,cte,cfs,edt] = IR_stats(filename, varargin)
```
vijayaditya: We can't include this code. Ask the users to specify this file location and die if you cant find it.

egs/swbd/s5c/local/reverberate/prep_sim_rirs.sh
```diff
@@ -0,0 +1,85 @@
...
+  echo "  $0 [options] <output-dir>"
+  echo "e.g.:"
+  echo " $0  data/impulses_noises"
+  exit 1;
+fi
+
+output_dir=$1
+mkdir -p $output_dir
+
+cat >./genrir.m <<EOF
```
vijayaditya: Not a good idea to have such huge inline scripts. There is a lot of scope for bugs. Save this file separately.

egs/swbd/s5c/local/reverberate/prep_sim_rirs.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+# Copyright 2016  Tom Ko
+# Apache 2.0
+# script to simulate room impulse responses (RIRs)
+
+sampling_rate=8000
```
vijayaditya: If this code depends on external matlab codes.. check for their existence and die if not found.
Give the user an option to specify the location of the required files.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
+# Copyright 2016  Tom Ko
+# Apache 2.0
+# script to generate reverberated data
+
+# we're using python 3.x style print but want it to work in python 2.x,
+from __future__ import print_function
+import argparse, glob, math, os, random, sys, warnings, copy, imp, ast
+
+nodes = imp.load_source('', 'steps/nnet3/components.py')
```
vijayaditya: imp.load_source('components', 'steps/nnet3/components.py')

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+  return new_lines
+
+
+def GetArgs():
+    # we add compulsary arguments as named arguments for readability
+    parser = argparse.ArgumentParser(description="Writes config files and variables "
+                                                 "for TDNNs creation and training",
+                                     epilog="See steps/nnet3/tdnn/train.sh for example.")
+
+    parser.add_argument("--rir-list", type=str, default = None,
```
vijayaditya: Provide format of rir-ist file

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+                        help="RIR information file")
+    parser.add_argument("--noise-list", type=str, default = None,
+                        help="Noise information file")
+    parser.add_argument("--num-replications", type=int, dest = "num_replica", default = 1,
+                        help="Number of replicate to generated for the data")
+    parser.add_argument("input_dir",
```
vijayaditya: I usually prefer specifying even compulsary arguments as named arguments. You can use required=True option to make named arguments compulsory.

vijayaditya:
BTW please move these files to a separate recipe. swbd_reverb. The train, dev and test sets are different for this recipe. There is no sense in storing these under the swbd recipe. If you want you could softlink required files from the swbd recipe.

vijayaditya:
@vimalmanohar As you would be one of the main users of these scripts in your VAD recipe, it would be very helpful if you are involved in this review. This is a still a WIP !!

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+      new_lines.append(line.strip())
+
+  return new_lines
+
+
+def GetArgs():
+    # we add compulsary arguments as named arguments for readability
+    parser = argparse.ArgumentParser(description="Writes config files and variables "
+                                                 "for TDNNs creation and training",
+                                     epilog="See steps/nnet3/tdnn/train.sh for example.")
```
vijayaditya: Add help formatter to the argparse constructor.

e.g.
<pre><code>
3     # we add compulsary arguments as named arguments for readability
 34     parser = argparse.ArgumentParser(description="""
 35     Trains a feed forward DNN acoustic model using the cross-entropy objective.
 36     DNNs include simple DNNs, TDNNs and CNNs.
 37     """,
 38     formatter_class=argparse.ArgumentDefaultsHelpFormatter)

</code></pre>

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+        raise Exception(args.rir_list + "not found")
+    
+    if args.noise_list is not None:
+        if not os.path.isfile(args.noise_list):
+            raise Exception(args.noise_list + "not found")
+
+    return args
+
+
+def Corrupt_wav(input_dir, output_dir, rir_list, num_replica, prefix):
```
vijayaditya: CorruptWav

vijayaditya:
Function naming convention in python scripts is mixed. Just stick to camelcase. (https://www.python.org/dev/peps/pep-0008/)

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+            split1[0] = prefix + str(i) + "_" + split1[0]
+            split1[1] = prefix + str(i) + "_" + split1[1]
+            print(" ".join(split1), file=f)
+    f.close()
+
+
+def Main():
+    args = GetArgs()
+
+    snr_array = [20 , 15, 10 , 5, 0]
```
vijayaditya: I would suggest moving this code from Main to another function.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+    if args.noise_list is not None:
+        Corrupt_wav_noise(args.input_dir, args.output_dir, args.rir_list, args.noise_list, snr_array, args.num_replica, "rvb")
+    else:
+        Corrupt_wav(args.input_dir, args.output_dir, args.rir_list, args.num_replica, "rvb")
+    Replicate_file2(args.input_dir + "/utt2spk", args.output_dir + "/utt2spk", args.num_replica, "rvb")
+    train_lib.RunKaldiCommand("utils/utt2spk_to_spk2utt.pl <{output_dir}/utt2spk >{output_dir}/spk2utt"
+                    .format(output_dir = args.output_dir))
+
+    if os.path.isfile(args.input_dir + "/text"):
+        Replicate_file1(args.input_dir + "/text", args.output_dir + "/text", args.num_replica, "rvb")
```
vijayaditya: Choosing better names for the methods Replicate* would make the code more readable.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+    if os.path.isfile(args.input_dir + "/segments"):
+        Replicate_file2(args.input_dir + "/segments", args.output_dir + "/segments", args.num_replica, "rvb")
+    if os.path.isfile(args.input_dir + "/reco2file_and_channel"):
+        Replicate_file2(args.input_dir + "/reco2file_and_channel", args.output_dir + "/reco2file_and_channel", args.num_replica, "rvb")
+ 
+    train_lib.RunKaldiCommand("utils/validate_data_dir.sh --no-feats {output_dir}"
```
vijayaditya: following the 80 line limit just for this line of code is unnecessary.

egs/swbd/s5c/local/reverberate/rir_generator.cpp
```diff
@@ -0,0 +1,459 @@
+/*
```
vijayaditya: Please do not  include external c++ code like this. I think it would be better to add this to external tools folder in kaldi/tools/extra, add the appropriate **download** and installation script. Check for the binary in the local script and if it is not found print a user message asking the user to run the installation.

egs/swbd/s5c/local/reverberate/prep_sim_rirs.sh
danpovey: Are you talking about whether matlab itself is installed?
Definitely there should be a check for that.
I'm very unhappy with having a dependency on matlab at all-- many people can't afford it.    But I suppose it won't be very easy to get rid of this dependency.  Even ensuring octave compatibility would be better than relying on matlab itself.

egs/swbd/s5c/local/reverberate/run_ivector_common.sh
```diff
@@ -0,0 +1,43 @@
...
+num_data_reps=3
+ali_dir=exp/tri4_ali_nodup_sp
+
+set -e
+. cmd.sh
+. ./path.sh
+. ./utils/parse_options.sh
+
+mkdir -p exp/nnet3_reverberate
+if [ $stage -le 0 ]; then
```
vijayaditya: We will assume that the simulated RIRs are available on openslr.org, so just download from there.
We will have to write a separate script to prepare these RIRs and upload to openslr.org, but that need not be part of the main recipe scripts. It can be an supporting script in the same recipe folder.

egs/swbd/s5c/local/reverberate/prep_sim_rirs.sh
danpovey: An option is to generate a 'sufficient' number of RIRs ourselves and put them in a Kaldi-readable format on openslr.org, and then just have the script that generated them provided for information.

egs/swbd/s5c/local/reverberate/run_ivector_common.sh
```diff
@@ -0,0 +1,43 @@
...
+snrs="20:10:15:5:0"
+num_data_reps=3
+ali_dir=exp/tri4_ali_nodup_sp
+
+set -e
+. cmd.sh
+. ./path.sh
+. ./utils/parse_options.sh
+
+mkdir -p exp/nnet3_reverberate
```
vijayaditya: when you move these files to swbd_reverb, just make the experiment directory exp/nnet3. Also you can directly start working on the chain recipe. We need not spend time on the xent versions of the recipe.

egs/swbd/s5c/local/reverberate/prep_sim_rirs.sh
vijayaditya: This is a temporary dependency. If these scripts are found to be useful we
will translate the necessary scripts to python (of course with a
scipy/numpy dependency).

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
vijayaditya: snrs should be user specified.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+            raise Exception(args.noise_list + "not found")
+
+    return args
+
+
+def Corrupt_wav(input_dir, output_dir, rir_list, num_replica, prefix):
+    rirs = list_cyclic_iterator(return_nonempty_lines(open(rir_list).readlines()), random_seed = 1)
+    wav_files = open(input_dir + "/wav.scp", 'r').readlines()
+    command_list = []
+    for i in range(num_replica):
```
vijayaditya: You would also have to support the specification of other options of wav-reverberate (e.g. --multi-channel)

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+    snrs = list_cyclic_iterator(snr_array)
+    wav_files = open(input_dir + "/wav.scp", 'r').readlines()
+    command_list = []
+    for i in range(num_replica):
+        for j in range(len(wav_files)):
+            split1 = wav_files[j].split()
+            wav_pipe = " ".join(split1[1:])
+            output_wav_id = prefix + str(i) + "_" + split1[0]
+            rir = rirs.next().split()
+            rir_file_location = rir[-1]
```
vijayaditya: provide the format of rir in comments.

egs/swbd/s5c/local/reverberate/IR_stats.m
danpovey: I don't see any point distributing scripts that have a dependency on code that we can't distribute.  This would only work on our grid -> pointless to check in, surely.

danpovey: I see this file seems to be available online at
http://www.mathworks.com/matlabcentral/fileexchange/42566-impulse-response-acoustic-information-calculator/content/IR_stats.m
This could be a download, I guess- or a patched download if needed.  Anyway, at that location, it says the code is covered by the BSD license, which implies that we can redistribute it.

vijayaditya: Hmm.. in that case we will just have to rely on the openslr.org for the
RIRs and noise. The scripts which prepare these simulated RIRs for upload
to openslr.org, will not be checked in.

vijayaditya: This script is not central to the recipe. It would just be necessary for
generating the metadata for files being uploaded to openslr.org.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+            noise = noises.next().split()
+            noise_file_location = noise[-1]
+            snr = snrs.next()
+            command_list.append("{4} {0} wav-reverberate --noise-file={2} --snr-db={3} - {1} - |\n".format(wav_pipe, rir_file_location, noise_file_location, snr, output_wav_id))
```
vijayaditya: This function would be a lot more complicated as you would have to specify multiple point source noise files and corresponding RIRs; and when choosing an isotropic noise you would have to make sure it is from the same room and corresponds to the same microphone position.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
```diff
@@ -0,0 +1,165 @@
...
+            rir = rirs.next().split()
+            rir_file_location = rir[-1]
+            command_list.append("{2} {0} wav-reverberate - {1} - |\n".format(wav_pipe, rir_file_location, output_wav_id))
+
+    file_handle = open(output_dir + "/wav.scp", 'w')
+    file_handle.write("".join(command_list))
+    file_handle.close()
+
+
+def Corrupt_wav_noise(input_dir, output_dir, rir_list, noise_list, snr_array, num_replica, prefix):
```
vijayaditya: User should be able to specify an option which specifies whether point sources have to be added. Also @vimalmanohar requires foreground and background noises, it might be better if @vimalmanohar decides this function signature.

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
vimalmanohar: I think we would have to decide how these noises and rirs are chosen. e.g. Is a isotropic noise always chosen if it exists? How many point source noises or rirs are chosen for each file?
We should be choosing a foreground noise and background noise separately. Foreground noise is usually short non-stationary noise like door knocks, taps, siren etc. It would not make sense to add these in the same way as the stationary background noise, which I believe is just added over the entire duration of the file. The way I was adding them is that wav-reverberate takes a bunch of several different foreground noises and randomly adds them to the signal at different times.

vimalmanohar: It would be better to rename wav-reverberate as wav-corrupt since it adds noises as well.

vimalmanohar: From Vijay's example:
wav-reverberate --multi-channel-output=false \
             --rir-channel 1 \
              --point-source=noise1.wav --rir=rir1.wav --snr-db=10\
              --point-source=noise2.wav --rir=rir2.wav --snr-db=5 \
              --isotropic-noise=noise3.wav --snr-db=20 \
              --normalize-output=true \
               speech_rir.wav speech.wav 
The binary must support all these options. The lines for randomly picking noise, rir and snr must be modified to decide the number of different rir and noises (both foreground and background) and then decide the snr level for each noise.

vijayaditya:
@tomko Please try to modify the same PR as it will help people follow the
discussion. Do not create new PRs unless absolutely necessary. Even in this
case refer to the old PR.

tomkocse:
This PR was closed because the RIR simulation part should be excluded from this PR.
For the new PR about reverberate_data_dir.py, please refer to https://github.com/kaldi-asr/kaldi/pull/706

egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
danpovey: On Wed, Apr 6, 2016 at 8:55 PM, Vimal Manohar <notifications@github.com>
wrote:

> In egs/swbd/s5c/local/reverberate/reverberate_data_dir.py
> <https://github.com/kaldi-asr/kaldi/pull/667#discussion_r58816432>:
>
> > +    rirs = list_cyclic_iterator(return_nonempty_lines(open(rir_list).readlines()), random_seed = 1)
> > +    noises = list_cyclic_iterator(return_nonempty_lines(open(noise_list).readlines()), random_seed = 1)
> > +    snrs = list_cyclic_iterator(snr_array)
> > +    wav_files = open(input_dir + "/wav.scp", 'r').readlines()
> > +    command_list = []
> > +    for i in range(num_replica):
> > +        for j in range(len(wav_files)):
> > +            split1 = wav_files[j].split()
> > +            wav_pipe = " ".join(split1[1:])
> > +            output_wav_id = prefix + str(i) + "_" + split1[0]
> > +            rir = rirs.next().split()
> > +            rir_file_location = rir[-1]
> > +            noise = noises.next().split()
> > +            noise_file_location = noise[-1]
> > +            snr = snrs.next()
> > +            command_list.append("{4} {0} wav-reverberate --noise-file={2} --snr-db={3} - {1} - |\n".format(wav_pipe, rir_file_location, noise_file_location, snr, output_wav_id))
>
> From Vijay's example:
> wav-reverberate --multi-channel-output=false \
> --rir-channel 1 \
> --point-source=noise1.wav --rir=rir1.wav --snr-db=10\
> --point-source=noise2.wav --rir=rir2.wav --snr-db=5 \
> --isotropic-noise=noise3.wav --snr-db=20 \
> --normalize-output=true \
> speech_rir.wav speech.wav
> The binary must support all these options. The lines for randomly picking
> noise, rir and snr must be modified to decide the number of different rir
> and noises (both foreground and background) and then decide the snr level
> for each noise.
>

Vimal, I think the discussion of the usage moved on quite bit since that
example.  Kaldi options parsing does not support multiple instances of the
same option, or position-dependent meaning of options.  We came up with an
alternative usage that differs in a bunch of respects.. I don't recall
right now where the thread is.

Dan

vijayaditya: We need someone to pick-up wav-reverberate.cc implementation. I will create a new issue describing the command-line arguments in detail. I will do this sometime tomorrow.

PR #669, opened 2016-04-07
don't raise exception when delete cache file

lifeiteng:


egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -493,6 +493,11 @@ def TrainOneIteration(dir, iter, egs_dir,
     elif os.stat(new_model).st_size == 0:
         raise Exception("{0} has size 0. Something went wrong in iteration {1}".format(new_model, iter))
 
+    try:
+        os.remove("{dir}/cache.{iter}".format(dir = dir, iter = iter))
+    except OSError:
+        pass
```
vdp: nitpick: IMO silencing exceptions like it's done here, goes against their intended use(reporting unexpected events). Personally I tend to prefer something like:
```
if os.path.exists(<file>):
    os.remove(<file>)

**Comment by danpovey** [comment]
I modified the PR, taking into account Vassil's comments, and pushed the changes directly.  So I'm deleting the PR.

**Comment by vdp** [comment]
In fact [some people](http://stackoverflow.com/questions/10840533/most-pythonic-way-to-delete-a-file-which-may-not-exist) feel that Feiteng's original code is more "Pythonic". For me the only valid argument they have, is that the try/catch approach is safer when you have another process that can delete the file between the calls to os.path.exists() and os.remove(). This is probably not a big concern in our case, and even if it was, the "correct" way to do it would be to define a function like in the second part of the accepted answer:

```
def silentremove(filename):
    try:
        os.remove(filename)
    except OSError as e: # this would be "except OSError, e:" before Python 2.6
        if e.errno != errno.ENOENT: # errno.ENOENT = no such file or directory
            raise # re-raise exception if a different error occured
```

PR #670, opened 2016-04-07
fixed uninitialized variable warnings in steps/cleanup/create_segments_from_ctm.pl

navneeth1990:
Creates segments and text file from the input ctm and aligned text. Outputs an optional aligned_ctm file in proper ctm format (if --aligned-ctm-filename flag is set). The final column of align_ctm file is an error count and not confidence scores. It is useful for debugging


danpovey:
Apparently this PR has conflicts.  Please resolve!

PR #674, opened 2016-04-08
Introduce googletest

kkm000:
- Download and compile googletest from tools/Makefile.
- Compile tests with its includes and the library
- Test kaldi-error more realistically (real message output), as an example of use.

### DIFF
--- a/src/base/kaldi-error-test.cc
+++ b/src/base/kaldi-error-test.cc
@@ -17,37 +17,63 @@
 // See the Apache 2 License for the specific language governing permissions and
 // limitations under the License.
 
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
 
 #include "base/kaldi-common.h"
 
-// testing that we get the stack trace.
 namespace kaldi {
+namespace {
 
-void MyFunction2() {
-  KALDI_ERR << "Ignore this error";
+using namespace ::testing;
+
+TEST(DefaultLoggingTest, Exceptions) {
+  EXPECT_NO_THROW(KALDI_LOG << "Test");
+  EXPECT_NO_THROW(KALDI_WARN << "Test");
+  EXPECT_ANY_THROW(KALDI_ERR << "Test");
+  EXPECT_ANY_THROW(KALDI_ASSERT(4 + 2 == 42));
 }
 
-void MyFunction1() {
-  MyFunction2();
+TEST(DefaultLoggingTest, WarninigStderrPrefix) {
+  testing::internal::CaptureStderr();
+  KALDI_WARN << "Twas brillig";
+  std::string stderr_log = testing::internal::GetCapturedStderr();
+  EXPECT_THAT(stderr_log, StartsWith("WARNING "));
+  EXPECT_THAT(stderr_log, HasSubstr(" Twas brillig"));
 }
 
-void UnitTestError() {
-  {
-    std::cerr << "Ignore next error:\n";
-    MyFunction1();
-  }
-}
-
-
-}  // end namespace kaldi.
-
-int main() {
-  kaldi::g_program_name = "/foo/bar/kaldi-error-test";
+TEST(DefaultLoggingTest, ErrorStderrPrefix) {
+  testing::internal::CaptureStderr();
   try {
-    kaldi::UnitTestError();
-    KALDI_ASSERT(0);  // should not happen.
-  } catch(std::runtime_error &r) {
-    std::cout << "UnitTestError: the error we generated was: " << r.what();
+    KALDI_ERR << "Twas brillig";
+  } catch (...) {
   }
+  std::string stderr_log = testing::internal::GetCapturedStderr();
+  EXPECT_THAT(stderr_log, StartsWith("ERROR "));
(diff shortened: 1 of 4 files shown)


danpovey:
Looks interesting.  Test is failing now.  Would likely require changes to configure script (don't we have to link with it?) and update to configure script version.  We might want to wait for a quiet period to do this.

kkm000:
I do not think any configure script changes are necessary. To me, the changes in `default_rules,mk` seem to cover it. I was careful not to change the way non-test Kaldi sources are built.

I just fixed the error.. Travis script failed on `make -j googletest` when it tried to download googletest source and also compile it at the same time. I just added the missing dependency.

As for when is the best time to introduce it, indeed up to you. We can test very concisely with it, more things (such as errors) and mocking is easy (like the classes I had to write to test ARPA parser are actually mocks). But then it becomes an integral required part of Kaldi testing system. From my point of view, it's pretty safe, as [googletest's own compiler requirements](https://github.com/google/googletest#user-content-linux-requirements) are pretty low, and it can probably be compiled on any system capable of compiling Kaldi.

vdp:
A few years ago I was looking for a unit test library for a project of mine and decided to go with [Catch](https://github.com/philsquared/Catch). It worked fine for me, but that was a small project.. The good thing about Catch is that it's header-only, i.e. no compilation is required. It only requires a C++98 compliant compiler too.

Kirill, do you know how it compares feature-wise with Google Test (Catch's feature list can be seen [here](https://github.com/philsquared/Catch/blob/master/docs/why-catch.md))?

kkm000:
@vdp:I was missing Google Test (or a better test framework if you want) doing my latest two larger changes: the ARPA compiler and the logging intercept. For the ARPA compiler, I needed to implement a mock, and gMock comes with Google Test. For the logging, two errors slipped by the (practically non-existent) unit tests, both in the log output. Google Test can intercept stdout and stderr into a string.I do not see either of these in Catch (yet?). To the Catch's defence, it is not a mocking framework, and gMock supports integrating into other test frameworks besides gTest, but, since it is already integrated into gTest, that's an advantage, IMO.

I like its BDD style test harnesses, that's expressive, and English test naming.

As for header-only or not, I think this is irrelevant for a big project like Kaldi. We compile OpenFST and ATLAS (the latter I have not an idea why we do, but it takes longer than Kaldi to build).

kkm000:
@jtrmal: Can you please look at this PR when you are not as busy? Not urgent at all.

vdp:
@kkm000: I see, thanks!
I'm not familiar with the internals of the existing unit tests in Kaldi, so Dan(and possibly others) are in much better position to assess the pros and cons of the introduction of specialized unit testing/mocking framework(s). In general though, if there is a substantial margin for improvement, IMO it should be seriosly considered. I mean, Kaldi continues to grow fast, so if the robustness and expressivity of the testing code can be improved, while making it easier to write new tests(e.g. by using mocks) in the same time, then it sounds like a pretty good deal..

jtrmal:
Will look, sorry -- this PR slipped under my radar.
y.

kkm000:
@jtrmal
> this PR slipped under my radar.

(_makes a radar chirp sound_)

Seriously, I am especially asking for the feedback on the way I am [changing the makefile](https://github.com/kaldi-asr/kaldi/pull/674/files#diff-62df187a2edfcbc8ba192646a494c483R73) -- namely, such that Google Test is available only to tests. An alternative would be to add the includes and libraries to configure, as @danpovey suggested above, but that would make it available to all code. I am basing my decision on two assumptions:
 * The test framework should be available only to test. If non-test code has `if (testing) { ... }`, it's wrong.
 * Google Test is prerequisite of Kaldi testing (No gTest = no testing, i. e. no option to configure without Google Test)
Do these assumptions look reasonable?

danpovey:
I'm still not entirely sure how I feel about this.
If the primary advantage of google test is that it makes it easier to test Kaldi's logging capabilities, I'm not sure that it warrants the additional dependency.

kkm000:
I am sure it makes testing many Kaldi capabilities easier. I cannot be quantitative here, but I really wanted mocking when implementing something-mock-like in the ARPA parser test. It has type parametrized tests (write test once then test with BaseFloat being float or double), value parametrized tests, it has much better diagnostics (Expectation failed, expected value:.. actual value...), it does not crash on first expectation failure if you do not want it to (ASSERT and EXPECT are separate facilities). It makes it much more convenient to write small test cases, testing one aspect of code each (Kaldi approach is untenable -- if you have 50 tests in a file, make sure they all are actually called from main()). Well, I'll stop here.

jtrmal:
I'm ok with that.
y.

On Mon, Apr 18, 2016 at 3:22 PM, Kirill Katsnelson <notifications@github.com

danpovey:
Let's leave it up for a while though- I'm still not certain about taking on
the extra dependency, and about how useful it will be for anyone except
Kirill.
Dan

danpovey:
Closing this-- I don't lean towards including it.  If there are strong feelings that we need it (except from Kirill, obviously), I'll reconsider.

Issue #671, opened 2016-04-07
get_utt2dur.sh does not seems to work properly for speed-perturbed data


freewym:
Hi,

I want to obtain the total duration for a subset. So I tried to use utils/data/get_utt2dur.sh to get the duration of each utterance. It works fine on non-speed-perturbed data. But as to speed-perturb data, the duration for speed-perturbed utterances seem not correct.

E.g., In /home/ywang/kaldi-chain2/egs/librispeech/s5/data/train_960_sp/utt2dur, all utterances prefixed with sp0.9- or sp1.1- have the same duration 67108.73, which does not look correct, while durations of  utterances prefixed with sp1.0-, which correspond to the non-perturbed ones, look reasonable.

I also tried on swbd/s5c/data/train_nodup_sp. get_utt2dur.sh works fine when the file segments is present. But if I removed segments, the script will compute the duration from wav.scp and there also seems to have the same problem for speed-perturbed utterances.

danpovey:
Please sort this out with Vijay.  It's possible that the script that
perturbed the data did not correctly modify the utt2dur, and you're using a
cached version of the utt2dur file.  Getting it from the wav.scp should
work fine- it will go through the speed-perturbing before it's read.
Dan

vijayaditya:
The issue occurs to issue in wav-to-duration.cc. The wav_data.Duration() returns the wrong duration as wav_data_ has the wrong num_cols_.
There seems to be an issue with reading file-length from streaming wav-data.
<code>sox WARN wav: Length in output .wav header will be wrong since can't seek to fix it</code>

danpovey:
Oh.  I think it's probably a bug in sox where, when it runs the
speed-modification command, it outputs the original (un-speed-perturbed)
length.  Sometimes it feels like sox's authors aren't even trying to make
it correct.
We could work around it by ensuring that the script that does the speed
perturbing create the utt2dur files itself- even if they were not present
in the source directory.
Dan

vijayaditya:
OK, this will be a modification in speed-perturbation code then.  @tomkocse / @freewym would you have time to fix this. 

To do:
If there is no segments file in the data directory, before perturbing the data generate an utt2dur file. Create the utt2dur file for perturbed data by modifying the non-perturbed data's utt2dur file.

danpovey:
To avoid problems in the future, it might also be a good idea to add an
option to wav-to-duration like --read-entire-file, which would force it to
[...]

PR #678, opened 2016-04-10
WIP: fix for utt2dur when applied on speed-pertubed data

freewym:
In order to fix the issue raised in #671 

class WaveInfoHolder and class WaveHolder do not have relationship in terms of inheritance. So currently I did two separate instantiations of SequentialTableReader in the "if  else " branches in featbin/wav-to-duration.cc, which duplicate some code. Please let me know if there are better solutions.

Yiming

### DIFF
--- a/egs/wsj/s5/utils/data/get_utt2dur.sh
+++ b/egs/wsj/s5/utils/data/get_utt2dur.sh
@@ -70,7 +70,7 @@ else
     fi
 
     read_entire_file=false
-    if [ `cat $data/wav.scp | sed -n '/sox.*speed/p' | wc -l` -gt 0 ]; then
+    if cat $data/wav.scp | grep -q 'sox.*speed'; then
       read_entire_file=true
       echo "$0: reading from the entire wav file to fix the problem caused by sox commands with speed perturbation. It is going to be slow."
     fi
--- a/egs/wsj/s5/utils/data/get_utt2dur.sh
+++ b/egs/wsj/s5/utils/data/get_utt2dur.sh
@@ -10,8 +10,6 @@
 # first tries interrogating the headers, and if this fails, it reads the wave
 # files in entirely.)
 
-read_entire_file=false
-
 . utils/parse_options.sh
 . ./path.sh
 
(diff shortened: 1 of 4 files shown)


egs/wsj/s5/utils/data/get_utt2dur.sh
```diff
@@ -68,7 +70,10 @@ else
       echo  "$0: wav-to-duration is not on your path"
       exit 1;
     fi
```
danpovey: I think at this point, if read_entire_file is false, you should check whether the wav.scp has any of the offending sox commands that tend to create this problem, and if so, set it to true and print a message saying why.

freewym: Add code to check if there are any sox command with keyword "speed"; if so, set read_entire_file=true.

egs/wsj/s5/utils/data/get_utt2dur.sh
```diff
@@ -68,7 +68,14 @@ else
...
-    if ! wav-to-duration scp:$data/wav.scp ark,t:$data/utt2dur 2>&1 | grep -v 'nonzero return status'; then
+
+    read_entire_file=false
+    if [ `cat $data/wav.scp | sed -n '/sox.*speed/p' | wc -l` -gt 0 ]; then
```
danpovey: I think a grep or grep -E command with output piped to /dev/null would be more elegant-- grep returns true if it matched at least one line.

drTonyRobinson: On 11/04/16 21:16, Daniel Povey wrote:
>
> In egs/wsj/s5/utils/data/get_utt2dur.sh 
> <https://github.com/kaldi-asr/kaldi/pull/678#discussion_r59273232>:
>
> > @@ -68,7 +68,14 @@ else
> >        echo  "$0: wav-to-duration is not on your path"
> >        exit 1;
> >      fi
> > -    if ! wav-to-duration scp:$data/wav.scp ark,t:$data/utt2dur 2>&1 | grep -v 'nonzero return status'; then
> > +
> > +    read_entire_file=false
> > +    if [ `cat $data/wav.scp | sed -n '/sox.*speed/p' | wc -l` -gt 0 ]; then
>
> I think a grep or grep -E command with output piped to /dev/null would 
> be more elegant-- grep returns true if it matched at least one line.
>

instead of stuffing the output to /dev/null you might want to use -q

        -q, --quiet, --silent
               Quiet; do not write anything to standard output.  Exit 
immediately with zero status if any match is found, even if  an error  was
               detected.  Also see the -s or --no-messages option. (-q 
is specified by POSIX.)

Tony

-- 
Speechmatics is a trading name of Cantab Research Limited
We are hiring: www.speechmatics.com/careers 
<https://www.speechmatics.com/careers>
Dr A J Robinson, Founder, Cantab Research Ltd
Phone direct: 01223 794096, office: 01223 794497
Company reg no GB 05697423, VAT reg no 925606030
51 Canterbury Street, Cambridge, CB4 3QG, UK

freewym: Thanks. Fixed.

danpovey:
Merging.  Thanks!

PR #681, opened 2016-04-11
Update arpa2fst invocations in individual egs/*/local scripts

### LINKED ISSUE #665: Complete arpa2fst invocations conversion
...started is #634.

### LINKED ISSUE #680: Sorting of G.fst in babel recipes
https://github.com/kaldi-asr/kaldi/blob/master/egs/babel/s5b/local/arpa2G_syllables.sh#L47 does not arcsort the G.fst (but the respective scripts in both babel/s5 and babel/s5c do).

https://github.com/kaldi-asr/kaldi/blob/master/egs/babel/s5b/local/arpa2G.sh#L108 olabel-sorts the G.fst (as does one in s5c).

Is this not on purpose? Is sorting irrelevant in this recipe?

kkm000:
Changed the was `arpa2fst` invoked in all `egs/*/local` scripts. This excludes scripts in common `tools` and `utils` directories linked to `wsj/s5`. I split the changes because the scripts in this PR have narrower scope of possible breakage. 

I think just a careful (but really _careful_) eyeballing would be sufficient. There are two variables to verify, namely references to `words.txt` and `G.fst` in each changed case. I would appreciate if anyone examines this set with a fresh set of eyes.

The only mildly non-trivial change is in `babel/s5*/local/arpa2G_syllables.sh`. 

Add `?w=1` to the end of the URL to suppress whitespace only change in  GitHub's diff.

Closes #665 
Closes #680

### DIFF
--- a/egs/aspire/s5/local/fisher_create_test_lang.sh
+++ b/egs/aspire/s5/local/fisher_create_test_lang.sh
@@ -1,4 +1,4 @@
-#!/bin/bash 
+#!/bin/bash
 #
 
 if [ -f path.sh ]; then . path.sh; fi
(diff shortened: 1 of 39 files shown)


danpovey:
This seems reasonable, but I want to wait a bit more, e.g. a week, to see if any problems arise from the changes we already made.  For instance, that guy on the list was having problems with lmrescore.sh, and I haven't yet got certainty on what the issue was and whether it might be arpa2fst related.

kkm000:
SGTM. Makes total sense.

PR #682, opened 2016-04-11
Update arpa2fst invocations in common (wsj/{utils,steps}) files

kkm000:
Only 3 files, but changes are not quite as trivial as in egs scripts.

`steps/make_phone_graph.sh` and `utils/format_lm_sri.sh` are both invoked from `swbd/s5{b,c}`, could be tested by this recipe.

`egs/wsj/s5/utils/reverse_lm.sh` is called only from aurora4/s5 and sprakbanken/s5.

### DIFF
--- a/egs/wsj/s5/steps/make_phone_graph.sh
+++ b/egs/wsj/s5/steps/make_phone_graph.sh
@@ -4,7 +4,7 @@
 
 # Copyright 2013  Johns Hopkins University (Author: Daniel Povey).  Apache 2.0.
 
-# This script makes a phone-based LM, without smoothing to unigram, that 
+# This script makes a phone-based LM, without smoothing to unigram, that
 # is to be used for segmentation, and uses that together with a model to
 # make a decoding graph.
 # Uses SRILM.
@@ -46,7 +46,7 @@ done
 loc=`which ngram-count`;
 if [ -z $loc ]; then
   if uname -a | grep 64 >/dev/null; then # some kind of 64 bit...
-    sdir=`pwd`/../../../tools/srilm/bin/i686-m64 
+    sdir=`pwd`/../../../tools/srilm/bin/i686-m64
   else
     sdir=`pwd`/../../../tools/srilm/bin/i686
   fi
@@ -92,17 +92,14 @@ fi
 
 if [ $stage -le 3 ]; then
   echo "$0: creating G_phones.fst from ARPA"
-  gunzip -c $dir/phone_graph/arpa_noug.gz | arpa2fst - - | fstprint | \
-    utils/eps2disambig.pl | utils/s2eps.pl | \
-    awk '{if (NF < 5 || $5 < 100.0) { print; }}' | \
-    fstcompile --isymbols=$lang/phones.txt --osymbols=$lang/phones.txt \
-       --keep_isymbols=false --keep_osymbols=false | \
-    fstconnect | \
-    fstrmepsilon > $dir/phone_graph/G_phones.fst
-   fstisstochastic $dir/phone_graph/G_phones.fst  || echo "[info]: G_phones not stochastic."
+  gunzip -c $dir/phone_graph/arpa_noug.gz | \
+    arpa2fst --disambig-symbol=#0 --read-symbol-table=$lang/phones.txt - - | \
+    fstprint | awk '{if (NF < 5 || $5 < 100.0) { print; }}' | fstcompile | \
+    fstconnect > $dir/phone_graph/G_phones.fst
+  fstisstochastic $dir/phone_graph/G_phones.fst || echo "[info]: G_phones not stochastic."
 fi
 
-  
+
 if [ $stage -le 4 ]; then
   echo "$0: creating CLG."
 
@@ -118,7 +115,7 @@ if [ $stage -le 5 ]; then
   echo "$0: creating Ha.fst"
   make-h-transducer --disambig-syms-out=$dir/phone_graph/disambig_tid.int \
     --transition-scale=$tscale $dir/phone_graph/ilabels_${N}_${P} $dir/tree $dir/final.mdl \
-       > $dir/phone_graph/Ha.fst 
+       > $dir/phone_graph/Ha.fst
 fi
 
 if [ $stage -le 6 ]; then
@@ -135,7 +132,7 @@ if [ $stage -le 7 ]; then
     $dir/final.mdl < $dir/phone_graph/HCLGa.fst > $dir/phone_graph/HCLG.fst || exit 1;
 
   if [ $tscale == 1.0 -a $loopscale == 1.0 ]; then
-    # No point doing this test if transition-scale not 1, as it is bound to fail. 
+    # No point doing this test if transition-scale not 1, as it is bound to fail.
     fstisstochastic $dir/phone_graph/HCLG.fst || echo "[info]: final HCLG is not stochastic."
   fi
 
--- a/egs/wsj/s5/utils/format_lm_sri.sh
+++ b/egs/wsj/s5/utils/format_lm_sri.sh
@@ -85,30 +85,15 @@ mkdir -p $out_dir
 cp -r $lang_dir/* $out_dir || exit 1;
 
 lm_base=$(basename $lm '.gz')
-gunzip -c $lm | utils/find_arpa_oovs.pl $out_dir/words.txt \
-  > $out_dir/oovs_${lm_base}.txt || exit 1;
-
-# Removing all "illegal" combinations of <s> and </s>, which are supposed to
-# occur only at being/end of utt.  These can cause determinization failures
-# of CLG [ends up being epsilon cycles].
-gunzip -c $lm \
-  | egrep -v '<s> <s>|</s> <s>|</s> </s>' \
-  | gzip -c > $tmpdir/lm.gz || exit 1;
-
 awk '{print $1}' $out_dir/words.txt > $tmpdir/voc || exit 1;
 
 # Change the LM vocabulary to be the intersection of the current LM vocabulary
 # and the set of words in the pronunciation lexicon. This also renormalizes the
 # LM by recomputing the backoff weights, and remove those ngrams whose
 # probabilities are lower than the backed-off estimates.
-change-lm-vocab -vocab $tmpdir/voc -lm $tmpdir/lm.gz -write-lm $tmpdir/out_lm \
-  $srilm_opts || exit 1;
-
-arpa2fst $tmpdir/out_lm | fstprint \
-  | utils/eps2disambig.pl | utils/s2eps.pl \
-  | fstcompile --isymbols=$out_dir/words.txt --osymbols=$out_dir/words.txt \
-    --keep_isymbols=false --keep_osymbols=false \
-  | fstrmepsilon | fstarcsort --sort_type=ilabel > $out_dir/G.fst || exit 1;
+change-lm-vocab -vocab $tmpdir/voc -lm $lm -write-lm - $srilm_opts | \
+  arpa2fst --disambig-symbol=#0 \
+           --read-symbol-table=$out_dir/words.txt - $out_dir/G.fst || exit 1
 
 fstisstochastic $out_dir/G.fst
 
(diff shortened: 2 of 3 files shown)


egs/wsj/s5/steps/make_phone_graph.sh
```diff
@@ -92,17 +92,14 @@ fi
...
-    utils/eps2disambig.pl | utils/s2eps.pl | \
-    awk '{if (NF < 5 || $5 < 100.0) { print; }}' | \
-    fstcompile --isymbols=$lang/phones.txt --osymbols=$lang/phones.txt \
-       --keep_isymbols=false --keep_osymbols=false | \
-    fstconnect | \
-    fstrmepsilon > $dir/phone_graph/G_phones.fst
-   fstisstochastic $dir/phone_graph/G_phones.fst  || echo "[info]: G_phones not stochastic."
+  gunzip -c $dir/phone_graph/arpa_noug.gz | \
+    arpa2fst --disambig-symbol=#0 --read-symbol-table=$lang/phones.txt - - | \
+    fstprint | awk '{if (NF < 5 || $5 < 100.0) { print; }}' | fstcompile | \
```
danpovey: I just noticed an issue in the arpa2fst usage:
Typo in neither:
  --keep-symbols              : Store symbol table with FST. Forced true if symbol tables are neiter read or written (bool, default = false)

I think the usage could also do with a little more text explaining the effect of the various symbol table options, and what the choices are.  I don't think the behavior is very obvious or intuitive from the current usage message.

egs/wsj/s5/utils/format_lm_sri.sh
```diff
@@ -85,30 +85,15 @@ mkdir -p $out_dir
...
-  $srilm_opts || exit 1;
-
-arpa2fst $tmpdir/out_lm | fstprint \
-  | utils/eps2disambig.pl | utils/s2eps.pl \
-  | fstcompile --isymbols=$out_dir/words.txt --osymbols=$out_dir/words.txt \
-    --keep_isymbols=false --keep_osymbols=false \
-  | fstrmepsilon | fstarcsort --sort_type=ilabel > $out_dir/G.fst || exit 1;
+change-lm-vocab -vocab $tmpdir/voc -lm $lm -write-lm - $srilm_opts | \
+  arpa2fst --disambig-symbol=#0 \
+           --read-symbol-table=$out_dir/words.txt - $out_dir/G.fst || exit 1
```
danpovey: have you done any kind of testing of this script?  I don't want to break half the recipes.

egs/wsj/s5/utils/reverse_lm.sh
```diff
@@ -38,25 +38,13 @@ mkdir -p $outdir
 for f in phones.txt words.txt L.fst L_disambig.fst phones/; do
   cp -r $langdir/$f $outdir
```
danpovey: I think if you are going to change the reverse_lm.sh script like this you should try to run that stuff- it's in the WSJ example.  Otherwise leave it alone- it's never really used these days.  For instance, I'm not sure if the forbidden combinations of <s> and </s> would interact badly with the reverse_arpa script.

egs/wsj/s5/steps/make_phone_graph.sh
kkm000: How long a usage line is acceptable?

Would this statement be more clear:

> Store symbol table with FST. Symbols are always stored if symbol tables are neither read or written otherwise

This complication is intended to keep arpa2fst backward-compatible. The old behavior of just `arpa2fst input | fstprint` is to print the FST with symbols. I could have set the default to true always, but that would be a wholesale inconvenience in the new invocation pattern that would have to carry --keep_symbols=false over every call. Hence I took the tradeoff of a more complex default. Arguably, this makes some intuitive sense: if an ARPA file is read, but no words.txt is provided and the symbols are not saved to a separate file, force save the symbols to FST, otherwise they will be lost and the FST will be useless. In another program I would just print an error message if an invalid combination of switches is provided, but in this case it does not cut it.

egs/wsj/s5/utils/format_lm_sri.sh
kkm000: No I did not, and neither do I. Is there a way someone would run one of `swbd/s5{b,c}` on your end? This would cover 2 out of 3 scripts in this PR. I have no data.

egs/wsj/s5/utils/reverse_lm.sh
kkm000: I think I should leave the `<s>` and `</s>` filters. I am less keen on leaving the whole script alone, as that would hold the last references to `utils/eps2disambig.pl`, `utils/s2eps.pl`, and `utils/remove_oovs.pl`, all of which are not really useful. Same thing--no WSJ data here though. :(

kkm000: I can perhaps run an old/new comparison on _an_ ARPA model.

egs/wsj/s5/steps/make_phone_graph.sh
danpovey: On Mon, Apr 11, 2016 at 2:19 PM, Kirill Katsnelson <notifications@github.com
> wrote:

> In egs/wsj/s5/steps/make_phone_graph.sh
> <https://github.com/kaldi-asr/kaldi/pull/682#discussion_r59283535>:
>
> > @@ -92,17 +92,14 @@ fi
> >
> >  if [ $stage -le 3 ]; then
> >    echo "$0: creating G_phones.fst from ARPA"
> > -  gunzip -c $dir/phone_graph/arpa_noug.gz | arpa2fst - - | fstprint | \
> > -    utils/eps2disambig.pl | utils/s2eps.pl | \
> > -    awk '{if (NF < 5 || $5 < 100.0) { print; }}' | \
> > -    fstcompile --isymbols=$lang/phones.txt --osymbols=$lang/phones.txt \
> > -       --keep_isymbols=false --keep_osymbols=false | \
> > -    fstconnect | \
> > -    fstrmepsilon > $dir/phone_graph/G_phones.fst
> > -   fstisstochastic $dir/phone_graph/G_phones.fst  || echo "[info]: G_phones not stochastic."
> > +  gunzip -c $dir/phone_graph/arpa_noug.gz | \
> > +    arpa2fst --disambig-symbol=#0 --read-symbol-table=$lang/phones.txt - - | \
> > +    fstprint | awk '{if (NF < 5 || $5 < 100.0) { print; }}' | fstcompile | \
>
> How long a usage line is acceptable?
>
> Would this statement be more clear:
>
> Store symbol table with FST. Symbols are always stored if symbol tables
> are neither read or written otherwise
>
>
I wasn't talking about the options-registering string, but about the usage
message (after "usage:") for the tool as a whole.  And example command
lines (after the usage, after "e.g.:"-- you could have 'new' and 'old'
example command lines, and explain that in the 'new' one, the symbol tables
are not output, and explain the differences in the output.  Since the
behavior of this tool is much more complex than most Kaldi tools, it
deserves more explanation in the usage message.  E.g. would it be clear to
you yourself in 10 years what the program was doing?

egs/wsj/s5/utils/format_lm_sri.sh
danpovey: @xiaohui-zhang, could you please run enough of the Switchboard s5c recipe with this PR, to verify that it doesn't break?  E.g. maybe up to the tri2 stage; make sure that at least a couple of these scripts have been invoked and we've tested the output.  Or if you can find an existing copy (e.g. in my /home/dpovey/kaldi-chain/egs/swbd/s5c), you could try to just run the individual scripts affected in the PR, and see how similar the output is.
Dan

egs/wsj/s5/utils/reverse_lm.sh
danpovey: ok, all that sounds reasonable- try to test it on some ARPA file, and change all except the BOS/EOS filters.

egs/wsj/s5/utils/format_lm_sri.sh
kkm000: Thanks!

egs/wsj/s5/utils/reverse_lm.sh
kkm000: The reverse_lm script is used only in aurora4/s5 and sprakbanken/s5. The latter is free, so I can probably pull a real grammar. This should be thorough enough.

Another interesting question is should we try to implement and maintain unit tests for the utility scripts (and executable tools by extension). Sounds nasty, but a quick search reveals a few bash unit-testing frameworks. The implementation is for another day indeed, as there is just one of me, unfortunately. But as a point to be in--does it make sense?

danpovey: >
> In egs/wsj/s5/utils/reverse_lm.sh
> <https://github.com/kaldi-asr/kaldi/pull/682#discussion_r59294521>:
>
> > @@ -38,25 +38,13 @@ mkdir -p $outdir
> >  for f in phones.txt words.txt L.fst L_disambig.fst phones/; do
> >    cp -r $langdir/$f $outdir
>
> The reverse_lm script is used only in aurora4/s5 and sprakbanken/s5. The
> latter is free, so I can probably pull a real grammar. This should be
> thorough enough.
>
> Another interesting question is should we try to implement and maintain
> unit tests for the utility scripts (and executable tools by extension).
> Sounds nasty, but a quick search reveals a few bash unit-testing
> frameworks. The implementation is for another day indeed, as there is just
> one of me, unfortunately. But as a point to be in--does it make sense?
>
A few of the utilities have comments at the end indicating how they were
tested.  Regarding unit-testing the scripts I'm not sure how many of them
would be amenable to unit testing.  Also the line where Kaldi ends and the
rest of the world begins can be a little fuzzy, and the range of
environments where the scripts are supposed to work is not always well
defined.  But at least seeing what it would look like would be interesting.
Dan

egs/wsj/s5/utils/format_lm_sri.sh
xiaohui-zhang: Hi Dan I've tested on swbd. The decoding results till tri2 makes zero difference. It only involves utils/format_lm_sri.sh till tri2 stage though.  steps/make_phone_graph.sh is involved in resegmentation.  Shall I go ahead test it?

danpovey: Yes it would be good if you could test the resegmentation script too.  Bear
in mind that it could be broken for unrelated reasons.
Dan

egs/wsj/s5/steps/make_phone_graph.sh
```diff
@@ -92,17 +92,14 @@ fi
...
+    fstconnect > $dir/phone_graph/G_phones.fst
```
danpovey: fstrmepsilon was in the old command line.  I think this may have been important but I forget the reason.

kkm000: It was in every pipeline because of utils/s2eps.pl. The new invocation of arpa2fst with --disambig-symbol produces inherently epsilon-free machines. `fstconnect` does not add arcs, so `fstrmepsilon` is not needed.

danpovey: OK thanks.

On Wed, Apr 13, 2016 at 2:48 PM, Kirill Katsnelson <notifications@github.com

kkm000:
Just confirmed the `reverse_lm` script produced an apparently identical G.fst with a relatively small ARPA model (the G has ~200K arcs).
```
$ fstequivalent --random --npath=2000 data/lang.reverse.orig/G.fst data/lang.reverse/G.fst; echo $?
0
```

danpovey:
OK thanks.

On Wed, Apr 13, 2016 at 3:19 PM, Kirill Katsnelson <notifications@github.com

egs/wsj/s5/utils/format_lm_sri.sh
xiaohui-zhang: The new make_phone_graph.sh results in 0.1% degradation of the tri4a_reseg system:
%WER 22.8 | 1831 21395 | 79.8 13.7 6.4 2.7 22.8 62.2 | exp.test/tri4a_reseg/decode_eval2000_sw1_tg/score_16_1.0/eval2000.ctm.swbd.filt.sys

Baseline:
%WER 22.7 | 1831 21395 | 80.2 13.8 6.0 2.9 22.7 62.4 | exp/tri4a_reseg/decode_eval2000_sw1_tg/score_14_0.0/eval2000.ctm.swbd.filt.sys

danpovey: Thanks for testing... my suspicion is that that small difference is due to
random noise though.
Dan

PR #683, opened 2016-04-11
utils/mkgraph.sh finds the information about the context in the tree.

francisr:
The options --mono and --quinphone are still there and can lead to exit 1 if there is a mismatch with the tree used.

### DIFF
--- a/egs/wsj/s5/utils/mkgraph.sh
+++ b/egs/wsj/s5/utils/mkgraph.sh
@@ -13,9 +13,8 @@
 # (this is compiled from this repository using Doxygen,
 # the source for this part is in src/doc/graph_recipe_test.dox)
 
+set -o pipefail
 
-N=3
-P=1
 tscale=1.0
 loopscale=0.1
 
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/mkgraph.sh
```diff
@@ -58,6 +56,14 @@ for f in $required; do
   [ ! -f $f ] && echo "mkgraph.sh: expected $f to exist" && exit 1;
 done
 
+N=`tree-info $tree | grep "context-width" | cut -d' ' -f2`
```
danpovey: You should check for exit status here e.g. by adding || exit 1, in case the path is bad or Kaldi is not compiled.
I prefer $() over `` for new code.

francisr:
I've updated the PR, is it fine for you?

danpovey:
Nearly.  But you need to use curly braces not parentheses-- your code would not exit because it's just exiting from a subshell into the main shell script.

danpovey:
And please squash the commits when you're done- you can do it with
git rebase -i HEAD~3
and change the last two commits to 's' (squash).

francisr:
Done. I had to use set -o pipefail, because `cut` exits with 0 if the input is empty.

danpovey:
Thanks!   Merging.

PR #684, opened 2016-04-11
WIP: Included ReLU component in the nnet1 version of DNN.

creatorscan:
Included the ReLU compoent in the nnet1 version

### DIFF
--- a/src/cudamatrix/cu-matrix.cc
+++ b/src/cudamatrix/cu-matrix.cc
@@ -1439,6 +1439,53 @@ void CuMatrixBase<Real>::DiffTanh(const CuMatrixBase<Real> &value,
   }
 }
 
+
+template<typename Real>
+void CuMatrixBase<Real>::ReLU(const CuMatrixBase<Real> &src) {
+  KALDI_ASSERT(SameDim(*this, src));
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) {
+    Timer tim;
+
+    dim3 dimBlock(CU2DBLOCK, CU2DBLOCK);
+    dim3 dimGrid(n_blocks(src.NumCols(), CU2DBLOCK), n_blocks(src.NumRows(), CU2DBLOCK));
+
+    cuda_tanh(dimGrid, dimBlock, this->data_, src.data_, this->Dim(), src.Stride());
+    CU_SAFE_CALL(cudaGetLastError());
+
+    CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
+  } else
+#endif
+  {
+    Mat().ReLU(src.Mat());
+  }
+}
+
+
+
+template<typename Real> // Ein -> diff, Y -> value
+void CuMatrixBase<Real>::DiffReLU(const CuMatrixBase<Real> &value,
+                                  const CuMatrixBase<Real> &diff) {
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) {
+    Timer tim;
+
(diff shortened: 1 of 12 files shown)


vesis84:
Hi, many thanks for the code!
Maybe the ReLu code could be more generic. It could support the 'leaky relu' and 'updatable slopes'. 
The slopes for the positive and negative activations could be defined by trainable constants, while each neuron has 2 values, plus there could be a switch to average those within the 'layer'. By default the relu parameters will be excluded from updating by 'UpdatableComponent::learn_rate_coef=0.0', or trained with lower speed using small positive value. It may also be handy to have a mechanism to switch of the updates of 'negative slopes'.

So the ReLu component could be composed of: the 2 CuVectors with slopes 'positive_slope_' 'negative_slope_', and the 3 constants: <LearnRateCoef> 0.0, <SlopePerNeuron> T, <UpdateNegative> T.

Do you have some more ideas how to make it even more generic?
For the moment I'll keep the PR open, the update is done by pushing
into your source branch of this PR.
Cheers!
Karel.

creatorscan:
I was working on few of these tweaks for relu. So I will update it by this weekend.

src/cudamatrix/cu-matrix.cc
```diff
@@ -1439,6 +1439,53 @@ void CuMatrixBase<Real>::DiffTanh(const CuMatrixBase<Real> &value,
...
+void CuMatrixBase<Real>::ReLU(const CuMatrixBase<Real> &src) {
+  KALDI_ASSERT(SameDim(*this, src));
+#if HAVE_CUDA == 1
+  if (CuDevice::Instantiate().Enabled()) {
+    Timer tim;
+
+    dim3 dimBlock(CU2DBLOCK, CU2DBLOCK);
+    dim3 dimGrid(n_blocks(src.NumCols(), CU2DBLOCK), n_blocks(src.NumRows(), CU2DBLOCK));
+
+    cuda_tanh(dimGrid, dimBlock, this->data_, src.data_, this->Dim(), src.Stride());
```
danpovey: i think you mean relu not tanh here- the same in the derivative code.  Did you test this?
Also, there are already existing low-level functions that would allow you to implement relu, there is no need to change the cudamatrix/ library.  Look at how it's implemented in nnet2 and nnet3.

danpovey:
It's odd that this branch would have conflicts-- perhaps you were working from an older Kaldi version and need to rebase your changes.

src/cudamatrix/cu-matrix.cc
creatorscan: Thank you Dan for pointing out the mistake. I fixed it. And, as you mentioned I also have implemented as in nnet2 using ApplyFloor and ApplyHeaviside in the nnet-activation.h itself. I found the later version to be bit slower comparatively.

mallidi:
Hi Karel/Karthick,

Is there any recipe which uses ReLu for nnet1 ? 

Harish

creatorscan:
Yes Harish, we have implemented Parametric ReLU. Hope we will merge it 
to Kaldi within 2 days.

danpovey:
Leave time for review!  2 days seems a bit quick.  I think in your WIP pull
request you had some kernels which were redudant.
Also, this is just advice, but I think you should make sure there is a way
to do regular ReLU.  People may want to use that, e.g. as a baseline.
Personally I tried things like parametric-ReLU in the past but did not get
anything from it.

vesis84:
Hi, yes, we closed this pull request. There will be a new PR with the parametric ReLUs, where the kernels will not be redundant because there will be 2 trainable parameters per neuron. The basic ReLU will be supported by the parametric ReLU with specific preset...

PR #686, opened 2016-04-11
WIP : chain: Added new steps/nnet3/chain/get_egs.py script to support extra…

vijayaditya:
…cting training egs with multiple chunk-widths (one chunk-width per archive). This was essential to avoid BLSTMs from exploiting the fixed chunkwidths to improve the chain objfn.

The current script just dumps egs with multiple chunk-widths, but with a single left/right context. I think it might be better to vary chunk-widths randomly during training. This choice was made as it was becoming difficult to find sufficient data for randomization when splitting data for different chunk_widths and contexts.  I will be adding changes to steps/nnet3/chain/train.py to add context perturbation during training.

Being tested; adding changes to fix minor bugs.


egs/wsj/s5/steps/nnet3/chain/get_egs.py
```diff
@@ -0,0 +1,727 @@
...
+
+    # The examples will go round-robin to egs_list.  Note: we omit the
+    # 'normalization.fst' argument while creating temporary egs: the phase of egs
+    # preparation that involves the normalization FST is quite CPU-intensive and
+    # it's more convenient to do it later, in the 'shuffle' stage.  Otherwise to
+    # make it efficient we need to use a large 'nj', like 40, and in that case
+    # there can be too many small files to deal with, because the total number of
+    # files is the product of 'nj' by 'num_archives_intermediate', which might be
+    # quite large.
+    num_frames = data_lib.GetNumFrames(feat_dir, utts)
```
vijayaditya: Unlike the shell script the num_frames, required to compute num_archives, is computed after excluding utts in valid_uttlist; so we will have smaller number of archives.

danpovey: It might be easier to have the chunk-width and left/right-context varying
along a single axis, e.g. proportional to each other-- you could define the
min-chunk-width, max-chunk-width, min-left-context, max-left-context, etc.
and have a num-sizes variable determining how many distinct sizes to dump.
 [if the {min,max} values are not distinct you could set this automatically
to 1].  To make the usage easier, there could be command line options like
--chunk-width that would automatically set the min and max to the same
values.
I have not looked at the script in detail- this is just my first impression
based on reading your comments.


Dan

vijayaditya: I think it would be better to just dump all egs with max-left-context/max-right-context and just perturb these during training. This might reduce the sensitivity of the model to the contexts.

danpovey: My feeling is that as long as there is some kind of variety present, it's
not going to be able to learn where the edges are-- having it very finely
randomized is not needed.  The reason why I prefer having this localized to
the get_egs script is that it helps keep the training script simple-- and
anyway we have to deal with multiple chunk lengths in the get_egs script,
so why not deal with the multiple contexts there too?
Dan

vijayaditya: OK, I will modify the left and right contexts to ensure that the total
number of input frames is always constant.

--Vijay

danpovey: > > +def GenerateTrainingExamplesFromUtts(dir, lat_dir, chain_dir, feat_dir,
>
> > +                                     utts, chunk_width,
> > +                                     train_feats_string, train_egs_opts_string,
> > +                                     supervision_opts, ivector_opts,
> > +                                     num_jobs, max_shuffle_jobs_run,
> > +                                     frames_per_iter, cmd, only_shuffle):
> > +
> > +    # The examples will go round-robin to egs_list.  Note: we omit the
> > +    # 'normalization.fst' argument while creating temporary egs: the phase of egs
> > +    # preparation that involves the normalization FST is quite CPU-intensive and
> > +    # it's more convenient to do it later, in the 'shuffle' stage.  Otherwise to
> > +    # make it efficient we need to use a large 'nj', like 40, and in that case
> > +    # there can be too many small files to deal with, because the total number of
> > +    # files is the product of 'nj' by 'num_archives_intermediate', which might be
> > +    # quite large.
> > +    num_frames = data_lib.GetNumFrames(feat_dir, utts)
>
> OK, I will modify the left and right contexts to ensure that the total
> number of input frames is always constant.
>


Er- not sure what you mean by this exactly.  What I was proposing was that
as you increase the chunk size, you also increase the left-context,
proportionally.  It wouldn't make sense to always leave the total
num-frames in the example the same-- what about models that didn't even
need extra left-context?  Anyway, this would generate very excessive
context sizes if the range of chunk sizes was quite large.

Dan

danpovey: ... actually scrub the 'proportionally' thing.
My proposal was that the user specifies the
{min-max}-{chunk-size,extra-{left,right}-context}, and the number of
distinct sizes (say --num-sizes), and you just vary them along the same
axis, so that the smallest size will have the
min-{chunk-size,extra-{left,right}-context} and the largeset size would
have the max-{chunk-size,extra-{left,right}-context}
Dan

vijayaditya: Using a small chunk-*-context for small chunks does not seem right to me. Shorter chunk-sizes/chunk-*-contexts can lead to worse pseudo-likelihood estimates, so I don't think reducing them both simultaneously is a good idea.

If we were to choose perturbation during training, the only script which needs to be modified is steps/nnet3/chain/train.py and even in this the change would be local to one function. So I think it would be worth the 1 hour effort, as we would not have to make these decisions about the dependencies between chunk-*-context and chunk-widths.

danpovey: On Mon, Apr 11, 2016 at 1:21 PM, Vijayaditya Peddinti <

> In egs/wsj/s5/steps/nnet3/chain/get_egs.py
> <https://github.com/kaldi-asr/kaldi/pull/686#discussion_r59274117>:
>
> > +def GenerateTrainingExamplesFromUtts(dir, lat_dir, chain_dir, feat_dir,
> > +                                     utts, chunk_width,
> > +                                     train_feats_string, train_egs_opts_string,
> > +                                     supervision_opts, ivector_opts,
> > +                                     num_jobs, max_shuffle_jobs_run,
> > +                                     frames_per_iter, cmd, only_shuffle):
> > +
> > +    # The examples will go round-robin to egs_list.  Note: we omit the
> > +    # 'normalization.fst' argument while creating temporary egs: the phase of egs
> > +    # preparation that involves the normalization FST is quite CPU-intensive and
> > +    # it's more convenient to do it later, in the 'shuffle' stage.  Otherwise to
> > +    # make it efficient we need to use a large 'nj', like 40, and in that case
> > +    # there can be too many small files to deal with, because the total number of
> > +    # files is the product of 'nj' by 'num_archives_intermediate', which might be
> > +    # quite large.
> > +    num_frames = data_lib.GetNumFrames(feat_dir, utts)
>
> Using a small chunk-*-context for small chunks does not seem right to me.
> Shorter chunk-sizes/chunk-*-contexts can lead to worse pseudo-likelihood
> estimates, so I don't think reducing them both simultaneously is a good
> idea.
>
Having them go the opposite way would be fine with me-- just don't enforce
that they sum to the same value, which seems a bit arbitrary.

> If we were to choose perturbation during training, the only script which
> needs to be modified is steps/nnet3/chain/train.py and even in this the
> change would be local to one function. So I think it would be worth the 1
> hour effort, as we would not have to make these decisions about the
>
dependencies between chunk-*-context and chunk-widths.
>
My concern is not so much about the effort per se, but about the fact that
any future changes to the training script would have to work around that.
Whereas we are already dealing with this issue in the get_egs script.  Plus
doing it in the training script is less efficient as we do unnecessary disk
I/O.

Dan

vijayaditya: OK, got it.

OK, in that case I will take the <code>--{min,max}-{left,right}-context</code> values from the user and randomly sample from these for each archive. Right now I am allowing the user to specify each chunk-width manually. So to specify three chunk-widths they would basically specify three <code>--chunk-width</code> options

vijayaditya:
The new directory structure to support multiple chunk-widths would be egs_dir/cw$chunk_width. The  ark files from these chunk_width specific directories will be shuffled and soft-linked in egs_dir. There is one additional change which breaks backward compatibility. I will be using a single file egs_dir/info/utt_partition_info which has information about all the chunk_width specific partitions. This file would have one line for each partition in the format
`--egs-dir <> --chunk-width <> --num-frames <> --egs-per-archive <> --num-archives <> --min-left-context <> --max-left-context <> --min-right-context <> --max-right-context <> `

The validation/train_subset egs will be dumped with default context of mean({min,max}-chunk-left-context).

For some time I will maintain backward compatibility to allow egs written by old code to be readable by the new training code.

egs/wsj/s5/steps/nnet3/chain/get_egs.py
danpovey: On Mon, Apr 11, 2016 at 1:32 PM, Vijayaditya Peddinti <

> In egs/wsj/s5/steps/nnet3/chain/get_egs.py
> <https://github.com/kaldi-asr/kaldi/pull/686#discussion_r59275930>:
>
> > +def GenerateTrainingExamplesFromUtts(dir, lat_dir, chain_dir, feat_dir,
> > +                                     utts, chunk_width,
> > +                                     train_feats_string, train_egs_opts_string,
> > +                                     supervision_opts, ivector_opts,
> > +                                     num_jobs, max_shuffle_jobs_run,
> > +                                     frames_per_iter, cmd, only_shuffle):
> > +
> > +    # The examples will go round-robin to egs_list.  Note: we omit the
> > +    # 'normalization.fst' argument while creating temporary egs: the phase of egs
> > +    # preparation that involves the normalization FST is quite CPU-intensive and
> > +    # it's more convenient to do it later, in the 'shuffle' stage.  Otherwise to
> > +    # make it efficient we need to use a large 'nj', like 40, and in that case
> > +    # there can be too many small files to deal with, because the total number of
> > +    # files is the product of 'nj' by 'num_archives_intermediate', which might be
> > +    # quite large.
> > +    num_frames = data_lib.GetNumFrames(feat_dir, utts)
>
> OK, got it.
>
> OK, in that case I will take the --{min,max}-{left,right}-context values
> from the user and randomly sample from these for each archive. Right now I
> am allowing the user to specify each chunk-width manually. So to specify
> three chunk-widths they would basically specify three --chunk-width
> options
>

I think specifying the same option multiple times is very susceptible to
misinterpretation, and against the general Kaldi pattern of key=value
specification of configuration values.  Why not specify the min and max,
like for the context, and have a num-sizes variable with a reasonable
default, like 5, which wouldn't normally have to be modified?  Then if the
user just specifies --frames-per-chunk, have it set both the min and max.

vijayaditya: How about just having a variable like
<code>--chunk-widths "100 125 150 175"</code> ?

danpovey: In egs/wsj/s5/steps/nnet3/chain/get_egs.py
> <https://github.com/kaldi-asr/kaldi/pull/686#discussion_r59287723>:
>
> > +def GenerateTrainingExamplesFromUtts(dir, lat_dir, chain_dir, feat_dir,
> > +                                     utts, chunk_width,
> > +                                     train_feats_string, train_egs_opts_string,
> > +                                     supervision_opts, ivector_opts,
> > +                                     num_jobs, max_shuffle_jobs_run,
> > +                                     frames_per_iter, cmd, only_shuffle):
> > +
> > +    # The examples will go round-robin to egs_list.  Note: we omit the
> > +    # 'normalization.fst' argument while creating temporary egs: the phase of egs
> > +    # preparation that involves the normalization FST is quite CPU-intensive and
> > +    # it's more convenient to do it later, in the 'shuffle' stage.  Otherwise to
> > +    # make it efficient we need to use a large 'nj', like 40, and in that case
> > +    # there can be too many small files to deal with, because the total number of
> > +    # files is the product of 'nj' by 'num_archives_intermediate', which might be
> > +    # quite large.
> > +    num_frames = data_lib.GetNumFrames(feat_dir, utts)
>
> How about just having a variable like
> --chunk-widths "100 125 150 175" ?
>
> It would be OK with me, I guess (although less symmetric with the
 extra-{left,right}-context options).
Dan

danpovey:
> The new directory structure to support multiple chunk-widths would be
> egs_dir/cw$chunk_width. The ark files from these chunk_width specific
> directories will be shuffled and soft-linked in egs_dir. There is one
> additional change which breaks backward compatibility. I will be using a
> single file egs_dir/info/utt_partition_info which has information about all
> the chunk_width specific partitions. This file would have one line for each
> partition in the format
> --egs-dir <> --chunk-width <> --num-frames <> --egs-per-archive <>
> --num-archives <> --min-left-context <> --max-left-context <>
> --min-right-context <> --max-right-context <>
>
Fine, I guess.  The format of the info/ dir may not make as much sense with
the varying chunk widths etc.  But see which parts of that info/ dir are
actually being used by the training script.  It could be that most of that
information is not even really needed, and other parts of the information
could be presented in a more invariant way (e.g. frames-per-archive instead
of egs-per-archive).... most of this information only needs to be
approximately correct anyway.


Dan

The validation/train_subset egs will be dumped with default context of

vijayaditya:
The assignment of utterances to utterance partitions corresponding to different chunk-widths is based on the following logic

<pre><code>
        # get the partition preference order
        # get distances different chunk boundaries
        distances = []
        additional_cost = 10
        for cw in chunk_widths:
            distance_from_left_boundary = width % cw
            distance_from_right_boundary = cw - distance_from_left_boundary
            distance_from_left_boundary = distance_from_left_boundary + additional_cost
            distances.append([cw, min(distance_from_left_boundary, distance_from_right_boundary)])
        distances = sorted(distances, key = lambda x:x[1])
        utt_partition_preference[utt_name] = map(lambda x: x[0], distances)
</code></pre>

With this logic not all chunk_widths might get equal number of frames. I will later decide if I want to ensure equal number of frames. e.g. in AMI-SDM

<pre><code>
2016-04-11 21:53:11,444 [get_egs.py:451 - GetTrainUttPartitions - INFO ] Generating egs with 3 different chunk width(s):
 chunk-width 100 -- num-frames 33510456
 chunk-width 125 -- num-frames 16547557
 chunk-width 150 -- num-frames 5520765
</code></pre>

danpovey:
I think it's OK if they get different numbers of frames as long as they're
not too different.  However, I think it would be better if your normalized
your preference as "relative" error, i.e. relative to the chunk-width
itself- or maybe even to the chunk-width squared.  That would reduce the
current strong preference for shorter chunk-widths.  We don't want all the
chunk widths to be the same length.
Dan

vijayaditya:
screen shot 2016-04-14 at 3 41 28 pm

Using different chunk-widths seems to be very helpful in preventing overfitting. However the WERs did not improve. I still have to figure out the correct left/right context to use when decoding with the new models.

I am motivated to check if context perturbation during training is actually helpful, so I will run an experiment with that and add it to the PR only in case it provides significant improvements. I will isolate this perturbation code to the maximum extent possible to ensure that it does not come in the way of future changes to the training code.

vijayaditya:
BTW the training session ran for smaller number of epochs as we have lesser number of arks. Previously `num_archives ` was being computed based on all frames in the training data directory and not the number of frames remaining after removal of validation utterances.

danpovey:
Very strange.  Make sure that you're properly holding out the validation
data.
Dan

vijayaditya:
My bad, there was a bug in script which led to 70% overlap between train and valid utts.

vijayaditya:
There has been an increase in WER when using larger number of chunk-widths. Investigation in progress.

danpovey:
Were the train and valid probs similar to before?

vijayaditya:
The train probabilities were similar to using a single chunk width, but the
validation probabilities were worse.

Vijay

danpovey:
Doing the same with TDNNs might help to narrow down the cause.
Be on the lookout for bugs that cause you to not train on the entire
dataset, or not properly randomize.  E.g. linking the final egs to the
wrong source egs, or having overlapping subsets.

Dan

vijayaditya:
OK.

vijayaditya:
Status update : I have restarted work on this PR.

vijayaditya:
This PR was submitted w.r.t. chain branch which is no longer updated. I am resubmitting the PR w.r.t. master branch.

PR #688, opened 2016-04-12
Updated arpa2fst usage message

kkm000:
xref #682


src/lmbin/arpa2fst.cc
```diff
@@ -31,7 +31,11 @@ int main(int argc, char *argv[]) {
         "Convert an ARPA format language model into an FST\n"
         "Usage: arpa2fst [opts] <input_arpa> <output_fst>\n"
         " e.g.: arpa2fst --disambig-symbol=#0 --read-symbol-table="
-        "data/lang/words.txt lm/input.arpa G.fst\n";
+        "data/lang/words.txt lm/input.arpa G.fst\n\n"
+        "Note: When called without switches, the output G.fst will contain an "
+        "embedded\n" "symbol table. This is compatible with the way our "
+        "original tool worked.\n";
```
danpovey: I'd call it "a previous version of arpa2fst" for more specificity.  And the linebreak in the middle of the c++ line looks odd- put it at the ends of lines for compatibility with other code, it's OK if the resulting messages are  a little compressed.

kkm000: Done (and squashed).

PR #694, opened 2016-04-12
smbr: Avoid extra epochs if frame shift is not used during training

vimalmanohar:


vimalmanohar:
Relates to the comment https://github.com/kaldi-asr/kaldi/commit/6a9ab33f5c68304e3ee5e300cfcfc715f5411bb5#commitcomment-17067942 #

PR #696, opened 2016-04-13
pipe added

tzyll:


danpovey:
Thanks- that does look like a bug.

kkm000:
Thanks, it was indeed a bug!

Issue #383, opened 2015-12-07
speaker-id: Bug inivector/voice-activity-detection.cc


david-ryan-snyder:
On the Kaldi ASR forums, lzhu.qs@gmail.com pointed out that there may be a bug in ivector/voice-activity-detection.cc line 54. `if (log_energy_data[t] > energy_threshold)`  should probably be changed to `if (log_energy_data[t2] > energy_threshold)`.

Matthew (@mmaciej2), this VAD is used in certain speaker ID and language ID scripts. Could you see how this change affects the examples in egs/sre08 and egs/sre10? By fixing this, some of the settings in the VAD are going to work differently, especially vad_frames_context (see http://kaldi.sourceforge.net/structkaldi_1_1VadEnergyOptions.html). If you get a degradation initially, you'll want to vary vad_frames_context, until it works as well or better than the original, bugged version. I recommend trying this (and possibly tuning the config parameters) first on egs/sre10/v1, and then using what works on the other examples.

kkm000:
I assume that was all to the fix? Closing, let me know if incorrect.

david-ryan-snyder:
Yes, this should be closed. Thanks!

PR #697, opened 2016-04-13
vadfix: fixed bug in voice activity detection context, changed default context from 5 to 0

mmaciej2:
This pull request fixes issue #383.

Experimentation with different vad contexts on sre10 showed that a larger context did not seem to help. A vad context of 0 is equivalent to the bugged version. None of the results should change as a result of this fix.


david-ryan-snyder:
To be clear, the bugged version of the code prevented the vad-frames-context option from working properly--it was as if vad-frames-context=0. However, after fixing the bug, it seemed like a vad-frames-context of 0 was better after all, at least for SRE10. Because of that, my suggestion to Matthew was to set the new default value of vad-frames-context to 0. This way, none of the results were affected in any of the examples.

danpovey:
Thanks!  Merging.

PR #702, opened 2016-04-15
Improvements to configure

giuliopaci:
This pull request currently includes two commits affecting src/configure.

The first one adds three options (--speex-root, --speex-libdir, --speex-includedir) that can be used to locate speex library. In my setup speex lib are in /usr/lib/x86_64-linux-gnu, while include files are in /usr/include:
```./configure --speex-libdir=/usr/lib/x86_64-linux-gnu --speex-root=/usr``` lets me use the system library instead of a custom one.

I second one remove all bashisms in configure script. In my system the default shell is dash (or at least I would like to switch back to dash again, as I temporary changed to bash). One way to solve the issue is, as it was done for the configure script, to properly set to /bin/bash the shebang; another way is to remove bash specific code. The benefits of supporting dash, rather then bash, are portability and speed, so, when possible, I think it is better to try to have scripts that work with dash.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #
 # This configure script is hand-generated, not auto-generated.
 # It creates the file kaldi.mk, which is %included by the Makefiles
@@ -25,23 +25,23 @@
 #I.e. after each change that affects the kaldi.mk or the build system as whole
 CONFIGURE_VERSION=3
 
-function rel2abs {
+rel2abs() {
   if [ ! -z "$1" ]; then
     local retval=`cd $1 2>/dev/null && pwd || exit 1`
     echo $retval
   fi
 }
 
-function read_dirname {
+read_dirname() {
   local dir_name=`expr "X$1" : '[^=]*=\(.*\)'`;
   local retval=`rel2abs $dir_name`
   [ -z $retval ] && echo "Bad option '$1': no such directory" && exit 1;
   echo $retval
 }
 
-function is_set {
+is_set() {
   local myvar=${1:-notset}
-  if [ "$myvar" == "notset" ]; then
+  if [ "$myvar" = "notset" ]; then
     return 1
   else
     return 0
@@ -68,7 +68,7 @@ unset CLAPACKROOT
 unset OPENBLASROOT
 unset MKLLIBDIR
 
-function usage {
+usage() {
   echo 'Usage: ./configure [--static|--shared] [--threaded-atlas={yes|no}] [--atlas-root=ATLASROOT] [--fst-root=FSTROOT]
   [--openblas-root=OPENBLASROOOT] [--clapack-root=CLAPACKROOT] [--mkl-root=MKLROOT] [--mkl-libdir=MKLLIBDIR] [--speex-root=SPEEXROOT] [--speex-libdir=SPEEXLIBDIR] [--speex-includedir=SPEEXINCLUDEDIR]
   [--omp-libdir=OMPDIR] [--static-fst={yes|no}] [--static-math={yes|no}] [--threaded-math={yes|no}] [--mathlib=ATLAS|MKL|CLAPACK|OPENBLAS]
@@ -191,20 +191,20 @@ done
 # code that may use a GPU.  Touching this file is a way to force this.
 touch cudamatrix/cu-common.h 2>/dev/null
 
-function failure {
+failure() {
   echo "***configure failed: $* ***" >&2
   if [ -f kaldi.mk ]; then rm kaldi.mk; fi
   exit 1;
 }
 
-function check_exists {
+check_exists() {
   if [ ! -f $1 ]; then failure "$1 not found."; fi
 }
 
-function check_for_bad_gcc {
-  if which gcc >&/dev/null; then  # gcc is on the path
+check_for_bad_gcc() {
+  if which gcc >/dev/null 2>&1; then  # gcc is on the path
     gcc_version=$(gcc -dumpspecs 2>&1 | grep -A1 -F '*version:' | grep -v version)
-    if [ "$gcc_version" == "4.8.2" ]; then
+    if [ "$gcc_version" = "4.8.2" ]; then
       echo "*** WARNING: your version of gcc seems to be 4.8.2. ***"
       echo "*** This version of gcc has a bug in nth_element ***"
       echo "*** in its implementation of the standard library ***"
@@ -214,7 +214,7 @@ function check_for_bad_gcc {
   fi
 }
 
-function check_for_slow_expf {
+check_for_slow_expf() {
   cd probe
   rm -f exp-test
   make -f Makefile.slow_expf 1>/dev/null
@@ -228,7 +228,7 @@ function check_for_slow_expf {
 }
 
 
-function exit_success {
+exit_success() {
   check_for_bad_gcc;
   check_for_slow_expf;
   echo "SUCCESS"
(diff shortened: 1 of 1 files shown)


danpovey:
I'll consider adding the speex options separately if you make a separate PR, but there is no point removing the bashisms from this script and I don't want the hassle of testing their removal.
This script does not depend on bash being the default shell in /bin/sh, it only depends on bash being installed and available at /bin/bash.  Since almost every script in Kaldi requires bash, there is no point installing Kaldi at all if you don't have bash installed on the system.

giuliopaci:
I just opened a new pull request, as you requested.

Will you consider any change that removes bashisms in future? If not, I will avoid removing them and just change to /bin/bash the shebang when I will find one.

danpovey:
We should always have /bin/bash as the shebang- let me know if in any case
it is not /bin/bash.
There is no point removing bashisms, for many reasons.
Firstly, making the shebang /bin/sh and using bash-or-dash shell scripts is
not workable because /bin/sh could easily point to csh.  In fact, Kaldi
scripts will work with /bin/sh pointing to any widely used shell *except*
dash, due to brace expansion used in some scripts.
Secondly, this is pointless unless done everywhere.  There are I think over
100,000 lines of shell scripts.  I don't think you or anyone else is
volunteering to convert them all and test them all.

Dan

jtrmal:
Also, why is it you mind bash being used? I don't see any reason why bash
couldn't be used as a shell interpret?
y.

giuliopaci:
I never said that bash could not be used. I just said that one should not be forced to have bash as the system /bin/sh.
Properly setting /bin/bash whenever I found a bashism, is a viable solution to me as it solves my issue.

For compatibility reason (and for the fact that if a bourne-shell script can be run under dash it is much faster and requires less resources) I prefer to remove bashisms, but this is just a matter of taste.

jtrmal:
OK, I understand now. Thanks.
Some of the scripts actually need bash set as sh, because  of the advanced
glob/filenames expansion. Dash does not support that specific feature (and
rewriting it would not be exactly trivial) -- that is the reason why we
basically enforce bash as sh and why we do not care about bashisms.
y.

danpovey:
It's not the scripts that have /bin/bash as the shebang that care about
whether /bin/sh points to bash.
It's expressions like  'gunzip -c foo/bar.{1,2,3,4,5,}.gz |' passed into
programs and evaluated using the system shell.  Removing those by having
the script generate an extremely long string itself and passing it into the
program, would fix that.  There aren't many of them, but generating the
long string would be ugly and in some case is possibly in danger of
reaching some kind of command-line length limit.

Dan

giuliopaci:
I know the differences among those two shells very well. And this is not the most complicated to solve. For example if you have long pipes and you want something similar to the pipefail option in posix shells, you are going to have hard time replacing them.

This explanation makes very clear why bash is needed as the system shell as well.

An option could be to explicitly invoke a bash shell in the command, although it requires a little bit of escaping in the general case: 'bash -c "gunzip -c foo/bar.{1,2,3,4,5,}.gz" |'

danpovey:
Hm.
Bear in mind that installing Kaldi is not really a common thing, it's an
upstream developement tool.  If someone creates a speech recognition
application for Debian (for instance) using Kaldi, they should be packaging
a small subset of the Kaldi code as needed for runtime.  If they require
all of Kaldi, they aren't developing their application right.
Dan

giuliopaci:
In this case I am the upstream of myself, as I am developing models for the purpose of forced alignment for myself. The tool that I need should be able to perform forced alignment of:
- a large set of files from the same speaker (e.g., TTS corpus);
- a single long file spoken by a single speaker (e.g., audiobooks);
- a single short file from any speaker.
The tool should be able to cope with mismatches among the audio and textual data. This is the final application in my case. I guess most of kaldi is needed in this case, especially for the first task, that is very similar to what is done in any of egs' subdirectory.

As I am both the upstream and the final user, I am obviously directly affected by the issue of both worlds.

So, while I agree that this issue affects others (and I am concerned about them, because I still have not abandoned the idea to have kaldi packaged for Debian), it also affects me directly now, as I am forced to use bash as the default shell on all the systems where I want to install kaldi (meaning that I need root access to all of those systems and have to accept the performance penalty that comes with it, which is not big while using kaldi, but can be significant for other tasks where scripts usage is preponderant).

giuliopaci:
Anyway, how are those code snippets executed?

danpovey:
Using popen.
Here is an example:

steps/nnet2/relabel_egs.sh:alignments=`eval echo $alidir/ali.{$(seq -s ','
$num_jobs_align)}.gz`

that gets passed into a program.

The reason we don't just give it $alidir/ali.*.gz is that they needed to be
provided in numerical order, not dictionary-sorting order.

Dan

giuliopaci:
Clear. Would you evaluate the inclusion of a popen replacement that grants /bin/bash usage, if I write/find one?
I see two possible solutions: a popen reimplementation on top of fork/execl/dup2 (thus requiring a Posix system to be compiled) or a wrapper of popen that runs bash and properly escapes the command.

danpovey:
That seems like an overly heavyweight fix.  There are only a handful of
scripts where we have this issue.  If we can somehow find them, it might
not be that hard to replace the problematic commands with something based
on bash -c "...".
Dan

giuliopaci:
If you will accept this other solution, then I will try to spot them and implement this solution.

If there are just a few cases, I agree that wrapping/reimplementing popen is not worth of it.

danpovey:
I seem to remember that there may be one or two additional scripts where
there is subscripting of bash arrays- that may have to be looked at also.
At least some of the scripts that have this issue will explicitly test for
dash
Dan

danpovey:
Sorry, scratch the thing about subscripting of bash arrays.  This would
have been done via queue.pl, and the bash arrays would not make it through
the perl layer since they are not standard environment variables, so that
situation must not exist, I must have misremembered.

giuliopaci:
I have a couple of questions:
1) in one of your previous comments you mentioned the command line length limit: did you ever experienced it? I am asking because, sooner or later, the arguments will be expanded anyway. However I know that system() on Windows has 1024 command line length limit, while commands executed by bash may indeed have longer command lines;
2) in the example that you provided I do not see the bash requirement... Am I right? Essentially what is passed to gunzip is the final expanded list, evaluated by bash through eval+echo, right?

danpovey:
> I have a couple of questions:
> 1) in one of your previous comments you mentioned the command line length
> limit: did you ever experienced it? I am asking because, sooner or later,
> the arguments will be expanded anyway. However I know that system() on
> Windows has 1024 command line length limit, while commands executed by bash
> may indeed have longer command lines;
>
I don't know for sure.  But there may be cases where there are multiple
such instances of gunzip on a single command line, so we might reach the
limit even if the individual things are below it.  The scripts don't run on
Windows anyway.

> 2) in the example that you provided I do not see the bash requirement...
> Am I right? Essentially what is passed to gunzip is the final expanded
> list, evaluated by bash through eval+echo, right?
>
The braces in the example are inside quotes and would not be expanded until
they are passed to popen by the kaldi executable.

Dan

giuliopaci:
They are not inside quotes, they are inside backticks, so the expression is evaluated before assignment to the variable.

danpovey:
OK, that may not have been a good example then.  But I think there are
other examples where the scripts don't work with dash.
Dan

giuliopaci:
Perfect. I will investigate.
I just wanted to be sure that I did not misunderstand that piece of code.

I guess the right keyword to look at is "ark", right?

danpovey:
Yes, but "ark" is everywhere.
Let me take care of this- it will be hard for you to figure out which
scripts depend on bash.  I hope to do it in a week or tow.

giuliopaci:
Ok, let me know if there is anything I can do to help. :-)

PR #704, opened 2016-04-15
Singular warning fixes

kkm000:
Heeding a few compile warnings, each occurring only once in the whole codebase. No rocket surgery here, just fixing some annoyances.

### DIFF
--- a/src/fstext/context-fst-inl.h
+++ b/src/fstext/context-fst-inl.h
@@ -88,8 +88,8 @@ template<class Arc, class LabelT>
 ContextFstImpl<Arc, LabelT>::ContextFstImpl(const ContextFstImpl &other):
     phone_syms_(other.phone_syms_),
     disambig_syms_(other.disambig_syms_) {
-  std::cerr << "ContextFst copying not yet supported [not hard, but would have to test.]";
-  exit(1);
+  KALDI_ERR << "ContextFst copying not yet supported "
+            << "[not hard, but would have to test.]";
 }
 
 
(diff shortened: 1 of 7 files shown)


src/fstext/context-fst-inl.h
```diff
@@ -310,9 +310,9 @@ bool ContextFstImpl<Arc, LabelT>::CreateArc(StateId s,
     // the output arcs, just 0.
     return CreatePhoneOrEpsArc(s, nextstate, olabel, phoneseq, oarc);
   } else {
-    std::cerr << "ContextFst: CreateArc, invalid olabel supplied [confusion about phone list or disambig symbols?]: "<<(olabel);
-    exit(1);
- }
+    KALDI_ERR << "ContextFst: CreateArc, invalid olabel supplied [confusion "
```
danpovey: better leave it as cerr... most of that code in fstext/ uses cerr (the aim was to make that independent of base/ for the most part)... it doesn't matter really but more consistent.

kkm000: Ah, I see, these are OpenFST extensions. Maybe use their logging then?

kkm000: Amusing, the warning I tried to fix was unreachable code at this line:
```c++
  return false;  // won't get here.  suppress compiler error.
```
Not all compilers are created equally smart. Oh well...

danpovey: oh- if it's doing exit(1) you could maybe remove the return(false).  possibly it was previously KALDI_ERR.

kkm000: I am sorry for pulling you into a discussion with a change so insignificant, did not really expect that. I just went back into the file to revert to cerr, and found that the same file uses `KADI_WARN`, `KALDI_ASSERT`, and also C `assert()`. It won't be consistent in this sense either way, KALDI_ERR or `std::cerr` (but at the very least, Kaldi-consistent with KALDI_ERR). But, in general, I think that if you treat fstext as (at least in the ideal) Kaldi-independent piece, it would be better off using OpenFST's LOG(FATAL/ERROR/...) macros than raw `cerr`. They default to stderr, but allow to intercept logging by substituting custom `fst/log.h` (as I do). If you agree, I'd be happy to make a thorough clean changeset to make logging consistent with OpenFST logging macros in this library.

danpovey: I think if we're doing a wholesale change, using KALDI_ERR and KALDI_WARN
in all those places would be better.

Dan

kkm000: Ok, sounds good. I'll push to this PR unless you think a separate one is better.

danpovey: fine.

src/fstext/pre-determinize-inl.h
```diff
@@ -547,9 +547,9 @@ void PreDeterminize(MutableFst<Arc> *fst,
     for (typename map<pair<StateId, ArcId>, size_t>::iterator m_iter = m_map.begin();
         m_iter != m_map.end();
         ++m_iter) {
-      pair<StateId, ArcId> pr = m_iter->first;
-      StateId state = pr.first;
-      ArcId arcpos = pr.second;
+      pair<StateId, ArcId> m_pr = m_iter->first;
```
danpovey: Kirill, since this m_pr is only used in the following two lines, I think it would better just to remove the variable entirely.
And then you can leave the other variable with its old name.

kkm000: Like `StateId state = m_iter->first.first; ArcId arcpos = m_iter->first.second;`? Looks wonderfully ugly to me. Just to make sure you really really want _that_.

kkm000:
As a side note on my latest assault on compiler warnings, it appears that OpenFST is sloppy in places on integral truncation. They freely pass e. g. `ssize_t` to [`Heap::Update(int, ..)`](https://github.com/danpovey/openfst/blob/151/openfst/src/include/fst/heap.h#L59), even as of 1.5.1.

src/fstext/pre-determinize-inl.h
danpovey: Yes please.

kkm000: Done.

danpovey:
Thanks.  Merging.

PR #705, opened 2016-04-17
Add table of examples (#469)

### LINKED ISSUE #469: table of egs
would like to take Fred Richardson's table, see
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kaldi-developers/hUM60Njtu5g/978NyJoDAQAJ
and make it part of the documentation.  need someone familiar with doxygen to make a pull request.

guoguo12:
I converted Fred Richardson's table (posted [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kaldi-developers/hUM60Njtu5g/978NyJoDAQAJ)) into an HTML table and made a new documentation page for it.

I didn't do any proofreading/fact-checking, but I'm willing to.

This resolves #469.

### DIFF
--- a/src/doc/examples.dox
+++ b/src/doc/examples.dox
@@ -24,335 +24,337 @@
 
  <table><tr><th>Name</th><th>BW</th><th>Lang</th><th>Train Domain</th><th>Train Hours</th><th>Train Speakers</th><th>License and Availability</th><th>Year Released</th><th>Speech Style</th><th>Test Domain</th><th>Kaldi Aprox Perf</th><th>LM Data</th><th>Lexicon</th></tr>
 <tr>
-  <td>AMI</td>
-  <td>16k</td>
-  <td>English<br>(+non-native)</td>
-  <td>Microphone: head-mike,<br>single and multiple<br>distance mikes</td>
-  <td>100</td>
-  <td>123 M<br>66 F</td>
-  <td>Free /<br>Download<br>http://groups.inf.ed.ac.uk/ami/corpus/</td>
-  <td>2014</td>
-  <td>Meeting room</td>
-  <td>Same as train<br>no overlap(?)</td>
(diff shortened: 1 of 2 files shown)


src/doc/examples.dox
```diff
@@ -0,0 +1,362 @@
+// doc/examples.dox
+
+// Copyright 2016 Fred Richardson
```
danpovey: you could put your name here too.

danpovey:
Thanks..
I'm a little concerned about this being hard to maintain-- the format makes it very hard to understand what field of the table each line of the .dox file corresponds to, so for someone to add a new example to the table - or even to edit an existing one- would be quite hard.  Can you see if you can figure out a way to make it easier?

guoguo12:
Added some comments to make it easier to read.

danpovey:
Thanks.
I think there should be a field for the directory, and there should be a preamble that explains that once you check out the Kaldi repository, the scripts can be found in subdirectories of egs/.  In future, in addition to just the WER we should probably be specifying what type of model it was  (e.g. an nnet1 xent+SMBR model; an nnet2 TDNN; an nnet3 chain+BLSTM+sMBR model, etc.  But this may be hard for you to do as it's time consuming, so for now just add a placeholder field for it.
Dan

guoguo12:
Added a brief preamble and a "Model Type" field.

I think it's pretty clear which subdirectories of `egs/` correspond to which rows in the table. Are there any that are particularly ambiguous?

danpovey:
I compiled this so it appears on the site right now.  @kkm000, would you mind commenting on this?

danpovey:
I'm merging this- I may reword it slightly or move it around, but it's a good start.

guoguo12:
Thanks! Let me know if I can help with anything.

PR #707, opened 2016-04-17
Fixed one option in nnet/run_lstm.sh

ialmajai:


danpovey:
Thanks- merging since it looks straightforward.

PR #708, opened 2016-04-17
force to generate utt2dur for every srcdir while calling combine_data.sh

freewym:
It was trying to fix a problem when the first srcdir of combine_data.sh contains utt2dur but not every other srcdir else. 

For example, when librispeech's run.sh was trying to combine train_clean_100(containing utt2dur) and train_clean_360(not containing utt2dur), the current script will result in a "combined" utt2dur file with entries only from train_clean_100.


danpovey:
I don't think this is the right solution-- it is against the 'const-ness' of the inputs to that script.
What would be better, IMO, is to have the combine_data.sh script check, not just for utt2dur but for all those scripts, that the inputs that it is combining all exist, and if not, to die.

Something like:
set -o pipefail
  ( for f in $*; do cat $f/$file || exit 1; done ) | sort -k1 > $dest/$file || exit 1;
set +o pipefail
(this will print a suitable error message naturally).

Then be a bit more careful in the calling scripts, that call combine_data.sh, to make sure everything has the utt2dur file.  For example, if you have a data perturbation script that doesn't copy the utt2dur, make sure it copies it- or regenerates it, as appropriate.

freewym:
Thanks! Fixed.

danpovey:
OK, thanks; merging.  If this generates additional errors we'll fix them as they arise.

PR #710, opened 2016-04-18
Subset on utt2spk when feats.scp is not necessary.

francisr:
There is a check for feats.scp later if it's needed.


danpovey:
Seems to make sense- merging.

PR #711, opened 2016-04-18
nnet1: creating tool 'nnet-set-learnrate', some structural changes in I/O,

vesis84:
- @Dan, @Yenda : this is a somewhat bigger change in 'nnet1', so I felt it appropriate to integrate it via a pull request. You can go through the code if you like, while it might not be necessary as it concerns the nnet1 code only.

- new tool 'nnet-set-learnrate', sets the scalar to individual components or whole Nnet.
- learn_rate_coef_, bias_learn_rate_coef_ is now part of interface UpdatableComponent, each component responsible to store it.
- many components are now reading <Tokens> in arbitrary order at ReadData(...), this will be good for backward-compatibility with old models when new <Tokens> are added.
- (planning to better follow the Google coding standard, this is not part of this pull-request)

### DIFF
--- a/src/nnet/nnet-activation.h
+++ b/src/nnet/nnet-activation.h
@@ -74,12 +74,11 @@ class BlockSoftmax : public Component {
     // parse config
     std::string token,
       dims_str;
-    while (!is.eof()) {
+    while (is >> std::ws, !is.eof()) {
       ReadToken(is, false, &token); 
       /**/ if (token == "<BlockDims>") is >> dims_str;
       else KALDI_ERR << "Unknown token " << token << ", a typo in config?"
                      << " (BlockDims)";
-      is >> std::ws; // eat-up whitespace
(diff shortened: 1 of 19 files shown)


danpovey:
I think it would be good if you ensure that at least for new code, you fix some of the most egregious style violations: for example space around operators such as <=; 80-char length limit.  Everyone else follows it.

vesis84:
I will focus on the style in the next pull request. I created this one as separate, to finish the functional changes. Fixing the style is a bigger project which will require changes on ~2500 code lines, and I did not want to mix it in the same branch.

kkm000:
> vesis84 merged commit 398e3dd into kaldi-asr:master 6 minutes ago
> 1 check was pending 

Ouch. It's halfway through. Everyone keep fingers on both hands crossed!

PR #712, opened 2016-04-19
haihua's nnet1 multilingual training recipe

haihua:
1 multilingual training code  nnet-train-frmshuff-mling is added (see nnet-mnet.h for details)
2 multilingual training script run-multilingual.sh is added
2.1 If you want to try kaldi nnet1 multilingual training recipe, please use --steps 1,2,3,4 for dnn, and use --steps 5,6,7 for bottle-neck;  and --steps 1,3,7,8 for my multilingual training recipe.  Please type the command and see the help information

### DIFF
--- a/egs/rm/s5/local/nnet/run-multilingual.sh
+++ b/egs/rm/s5/local/nnet/run-multilingual.sh
@@ -1,0 +1,393 @@
+#!/bin/bash
+
+# Copyright 2015  University of Illinois (Author: Amit Das)
+# Copyright 2012-2015  Brno University of Technology (Author: Karel Vesely)
+
+# Apache 2.0
+
+# This example script trains Multi-lingual DNN with <BlockSoftmax> output, using FBANK features.
+# The network is trained on multiple languages simultaneously, creating a separate softmax layer
+# per language while sharing hidden layers across all languages.
+# The script supports arbitrary number of languages.
+
+. path.sh
+. cmd.sh
+
+echo
+echo "## LOG: $0 $@"
+echo
+# begin options
+cmd=run.pl
+nj=40
+steps=
+validating_rate=0.1
+pretrain_rate=0.2
+learn_rate=0.008
+train_tool=nnet-train-frmshuff-mling
+train_tool_opts="--minibatch-size=2048 --randomizer-size=32768 --randomizer-seed=777"
+train_opts="--nn-depth 6 --hid-dim 2048 --splice 10"
+pretrain_cmd="steps/nnet/pretrain_dbn.sh --feat-type traps  --copy_feats_tmproot /local/hhx502"
+dnn_nnet_init=
+train_cmd="steps/nnet/train.sh --copy_feats_tmproot /local/hhx502"
+cmvn_opts="--norm-means=true --norm-vars=true"
+delta_opts="--delta-order=2"
+
+bn1_dim=80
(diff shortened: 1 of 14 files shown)


danpovey:
Urr...
Obviously this is a lot of work- and thanks- but I have reservations about various aspects of it.
I'm not sure how I feel about this as a project generally- you have implemented multilingual training on top of nnet1, with adding quite a lot of code, when it's more natural to do it in nnet3 and I'm not sure how much demand there will be for nnet1-based multilingual training given that nnet3 supports parallelization.  And there seems to be a lot of code.
You use 'LHUC' in the code .. not sure what this means or if it's an acronym (generally acronyms are discouraged, and this one isn't explained that I can see).
The way you use 'stages' is , IMO, rather odd and against the general principle or pattern of the stage option.
The code doesn't seem generally very well documented or self-explanatory-- e.g. you add class ParallelOneNnet with no explanation why the class is needed.. I'm sure other code you added is similarly undocumented, although I just happened to see this class.
I don't like 'mling' as an abbreviation- generally contractions like this are frowned upon, although we have 'multilang' elsewhere in the code, in nnet2 so that would be preferable.

haihua:
Yes, it is really confusing, due to  incomplete and unclear explanation.

1) for multilingual training, there is only one file concerned with it. It is nnet-mnet.h
2) For LHUC, it is concerned with DNN speaker adaptation (learning hidden unit contribution, by Pawel, Edinburgh). The code is finished, but I need  a test. I submit it only because nnet-nnet.cc file contains it, and if it is working, I will follow up.
3) for parallel net training (ParallelOneNnet), it is concerned with ivector based DNN adaptation code ( Yiajie's recipe, CMU), using sub-network to normalize ivector. I call it parallel net, because we can attach many parallel sub-networks to train the main network. The code  works, but is quite unclear.
4）for net3, right now, I have no idea how to follow it.
5) For Kaldi convention, I will obey.
Anyway, only  file nnet-mnet.h is for nnet1 multilingual training. Others are  irrelevant.

danpovey:
best to submit separate issues/techniques in separate PRs.
Dan

vesis84:
@Dan, I asked Haihua to send the code as PR, because he has faster multi-lingual training and I needed to take a look at it. It was not meant to be merged into the 'master' at this moment.
@Haihua, Dan is right about the split of PR per 'technique'. In general it is good to develop incrementally several independent things in parallel branches. So that one 'feature' can be merged at the time into the main branch. 
Dan is right about the missing in-line documentation and that it does not follow the 'Kaldi standards'. I will take some time to look at the code and think what to do next. There should be some relatively easy way how to re-use some parts of the code to extend the functionality of the main branch.
Thanks you and all the best!
Karel

jtrmal:
Also, please add "WIP" into name of the PR, in case it's not meant for
merging right away.
y.

vesis84:
Hi @haihua, I was thinking of the faster multi-lingual code. Unfortunately your code is designed very differently and it does not seem to be generic. It would be hard to support the multilingual NN with auxiliary multi-task targets for example.

It would be better to reuse as much of the existing code as possible. We could for example extend the '<ParallelComponent>'  by a vector of bool flags for propagation / backpropagation. This could be set from inside the 'nnet-train-perutt' according to the targets (makes sense also from mmi / smbr binaries).  

Would it be useful for you to extend my code if I send some instructions? 

Maybe we could create a branch to work on it, if you are interested. The first step would be to modify the multi-lingual example to use the '<ParallelComponent>'. Then we would introduce the flags...

All the best!
Karel.

PR #713, opened 2016-04-19
fix hkust locale

naxingyu:
remove locale dependent code


danpovey:
Thanks!  Merging without really checking it, as I assume you've tested it, plus the baseline code wasn't working reliably.

naxingyu:
Yes, I've tested it and finished the mono training. The result 
approximates the reference.

在 2016/4/20 0:53, Daniel Povey 写道:

danpovey:
Thanks!

PR #721, opened 2016-04-20
Add relative path option to ParseOptions

khers:
Using relative paths in config files today gives some odd behavior.  It
seems to attempt to resolve a relative path based on the $CWD of the
executable being run, but this does not always work.  This behavior is
not ideal when moving config files between experiments because it means
that all config files used must be updated for any paths that may have
changed.  This patch introduces a new field in the ParseOptions class
that allows the user to specify the "anchor" directory that will be
used to resolve any relative paths encountered.

Signed-off-by: Eric B Munson <eric@cobaltspeech.com>

### DIFF
--- a/src/util/parse-options.cc
+++ b/src/util/parse-options.cc
@@ -28,6 +28,11 @@
 #include <cstdlib>
 #include <cassert>
 #include <cstring>
+#include <climits>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
 
 #include "util/parse-options.h"
 #include "util/text-utils.h"
@@ -38,7 +43,8 @@ namespace kaldi {
 
 ParseOptions::ParseOptions(const std::string &prefix,
                            OptionsItf *other):
-    print_args_(false), help_(false), usage_(""), argc_(0), argv_(NULL) {
+    print_args_(false), help_(false), anchor_dir_(""), usage_(""),
+    argc_(0), argv_(NULL) {
   ParseOptions *po = dynamic_cast<ParseOptions*>(other);
   if (po != NULL && po->other_parser_ != NULL) {
     // we get here if this constructor is used twice, recursively.
@@ -464,11 +470,27 @@ void ParseOptions::PrintConfig(std::ostream &os) {
   os << '\n';
 }
 
+std::string ParseOptions::ResolvePath(const std::string &relative) {
+  if (anchor_dir_ == "")
+        return relative;
+
+  struct stat info;
+  if (::stat(relative.c_str(), &info) == 0)
+    return relative;
+
+  char tmp[PATH_MAX];
+  std::string cat = anchor_dir_ + "/" + relative;
+  ::realpath(cat.c_str(), tmp);
+  return std::string(tmp);
+}
+
 
 void ParseOptions::ReadConfigFile(const std::string &filename) {
-  std::ifstream is(filename.c_str(), std::ifstream::in);
+  std::string full_path = ResolvePath(filename);
+
+  std::ifstream is(full_path.c_str(), std::ifstream::in);
   if (!is.good()) {
-    KALDI_ERR << "Cannot open config file: " << filename;
+    KALDI_ERR << "Cannot open config file: " << full_path;
   }
 
   std::string line, key, value;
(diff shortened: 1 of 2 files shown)


danpovey:
It's good that you guys are willing to commit stuff.
However, I don't like this patch because it's confusing and a partial solution.
Firstly, it's expected that any relative paths should be relative to the current directory, and anything that changes this is potentially confusing (better to just modify the paths in the config files via script-level changes).
Secondly, this solution is only a very partial solution because you're just affecting how it resolves paths in a very specific instance-- the instance of where someone passes in --config=filename.
That is only about 1% of how people use filenames in config variables, and most of them you can't automatically detect because they are just strings.

In general neither relative nor absolute pathnames are a blanket solution -- absolute pathnames don't work if you want to move things, and relative pathnames don't work if you want to move some things but not other things.

khers:
> Firstly, it's expected that any relative paths should be relative to the current directory, and anything that changes this is potentially confusing (better to just modify the paths in the config files via script-level changes).

Do you mean that it would be better for a script to allow a user to specify what $CWD should be when is is run?  This doesn't meet the way we are using Kaldi.  What we'd prefer is way to specify a relative path in a config file that is loaded via --config= and to have the executable specify what the root for any relative path resolution should be without altering $CWD.  I picked the approach here because it did not impact the way ParseOptions is used today at all, but still allowed us to specify the anchor.

> Secondly, this solution is only a very partial solution because you're just affecting how it resolves paths in a very specific instance-- the instance of where someone passes in --config=filename.
That is only about 1% of how people use filenames in config variables, and most of them you can't automatically detect because they are just strings.

I want to make sure I understand what you are saying here.  This change currently only resolves relative paths when specifying a config file, but doesn't work for things like global-cmvn-stats in apply-cmvn-online for instance.

If so, I see what you mean.  We'd like these files to be "resolvable" as well and if you are open to taking a change which allows one to specify the anchor, I am happy to do the wider work of changing each of these options.

> In general neither relative nor absolute pathnames are a blanket solution -- absolute pathnames don't work if you want to move things, and relative pathnames don't work if you want to move some things but not other things.

I agree, but the way relative paths are handled today does not work for us so I was hoping to introduce a little more flexibility without breaking existing behavior.

danpovey:
> Firstly, it's expected that any relative paths should be relative to the
> current directory, and anything that changes this is potentially confusing
> (better to just modify the paths in the config files via script-level
> changes).
>
> Do you mean that it would be better for a script to allow a user to
> specify what $CWD should be when is is run? This doesn't meet the way we
> are using Kaldi. What we'd prefer is way to specify a relative path in a
> config file that is loaded via --config= and to have the executable specify
> what the root for any relative path resolution should be without altering
> $CWD. I picked the approach here because it did not impact the way
> ParseOptions is used today at all, but still allowed us to specify the
> anchor.
>
There is a way to specify what the $CWD is, it's the command 'cd'.  If you
want *some* paths to be as determined by the $CWD and others not, then you
could use absolute pathnames.

Secondly, this solution is only a very partial solution because you're just
> affecting how it resolves paths in a very specific instance-- the instance
> of where someone passes in --config=filename.
> That is only about 1% of how people use filenames in config variables, and
> most of them you can't automatically detect because they are just strings.
>
> I want to make sure I understand what you are saying here. This change
> currently only resolves relative paths when specifying a config file, but
> doesn't work for things like global-cmvn-stats in apply-cmvn-online for
> instance.
>
> If so, I see what you mean. We'd like these files to be "resolvable" as
> well and if you are open to taking a change which allows one to specify the
> anchor, I am happy to do the wider work of changing each of these options.
>
This gets very nasty very quickly, if you really think about the
consequences.  What you seem to want is for Kaldi to have *two* concepts of
a 'current directory', with one to be used for some purposes and one for
others.  That's pretty weird.  We wouldn't do this- sorry.

Dan

khers:
> There is a way to specify what the $CWD is, it's the command 'cd'.

Which only works if you want to or are able to change $CWD.

>  If you want *some* paths to be as determined by the $CWD and others not, then you could use absolute pathnames.

There are platforms on which you will likely not know the absolute path until runtime, if ever (Andriod for example)  this is a non-starter for those platforms.

> What you seem to want is for Kaldi to have *two* concepts of a 'current directory', with one to be used for some purposes and one for others

I never suggested that.  What I want is the ability to clarify when I want to use $CWD and when I don't.  That isn't two concepts of current directory, that is using $CWD unless told otherwise.

danpovey:
If you know the directory where the config files live, then surely you
could modify those config files to have absolute pathnames in them.
Going down the path you are asking for seems like it would introduce
debugging nightmares and greatly detract from the obviousness of Kaldi's
behavior.
Config files can contain filenames, as strings (without a tag saying this
is a filename) and those filenames can be passed into arbitrary code.  To
make only filenames in configs relative, we'd have to change code
throughout Kaldi to add a flag saying "this string came from a config
file", like the "dirty bit" in Perl.  Even if it was practical to do this,
I wouldn't allow it.


Dan

PR #726, opened 2016-04-21
Enable cross compile

khers:
The current build system does not allow a user to cross compile.  To get there we need to make some changes to the way the platform specific files under src/makefiles/*.mk are used.  At the end of this series, the user can effectively cross compile for any x86_[32|64]  or arm with the appropriate tool chain setup.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -28,6 +28,14 @@
 #        # built for the target install to /opt/cross/armv8hf and that the
 #        # armv8-rpi3-linux-gnueabihf toolchain is available in your path
 # ./configure --static --fst-root=/opt/cross/armv8hf --atlas-root=/opt/cross/armv8hf --host=armv8-rpi3-linux-gnueabihf
+#        # Cross compile for Android on arm
+#        # The only difference here is the addtion of the the --android-includes
+#        # flag because the toolchains produced by the Android NDK don't always
+#        # include the C++ stdlib headers in the normal cross compile include
+#        # path
+# ./configure --static --openblas-root=/opt/cross/arm-linux-androideabi \
+# --fst-root=/opt/cross/arm-linux-androideabi --host=arm-linux-androideabi \
+# --fst-version=1.4.1 --android-includes=/opt/cross/arm-linux-androideabi/sysroot/usr/include
 
 #This should be incremented after every significant change of the configure script
 #I.e. after each change that affects the kaldi.mk or the build system as whole
@@ -82,7 +90,7 @@ function usage {
   [--openblas-root=OPENBLASROOOT] [--clapack-root=CLAPACKROOT] [--mkl-root=MKLROOT] [--mkl-libdir=MKLLIBDIR]
   [--omp-libdir=OMPDIR] [--static-fst={yes|no}] [--static-math={yes|no}] [--threaded-math={yes|no}] [--mathlib=ATLAS|MKL|CLAPACK|OPENBLAS]
   [--use-cuda={yes|no}] [--cudatk-dir=CUDATKDIR][--mkl-threading=sequential|iomp|tbb|gomp] [--fst-version=VERSION]
-  [--host=HOST]';
+  [--host=HOST] [--android-includes=ANDROID_INC_DIR]';
 }
 
 threaded_atlas=false #  By default, use the un-threaded version of ATLAS.
@@ -96,6 +104,7 @@ mkl_threading=sequential
 # switch.  TARGET_ARCH will be the first value in HOST if set, and `uname -m` otherwise
 HOST=""
 TARGET_ARCH=""
+android=false
 
 cmd_line="$0 $@"  # Save the command line to include in kaldi.mk
 
@@ -204,6 +213,15 @@ do
     # this script will infer the target architecture from the specified triple.
     HOST=`expr "X$1" : '[^=]*=\(.*\)'`;
     shift ;;
+  --android-includes=*)
+    threaded_math=false;
+    static_math=true;
+    static_fst=true;
+    dynamic_kaldi=false;
+    MATHLIB='OPENBLAS';
+    android=true;
+    ANDROIDINC=`read_dirname $1`;
+    shift;;
   *)  echo "Unknown argument: $1, exiting"; usage; exit 1 ;;
   esac
 done
@@ -985,6 +1003,42 @@ fi
 
 # Most of the OS-specific steps below will append to kaldi.mk
 echo "Doing OS specific configurations ..."
+
+if $android ; then
+  OPENFSTLIBS="$FSTROOT/lib/libfst.a"
+  echo "OPENFSTLIBS = $OPENFSTLIBS" >> kaldi.mk
+
+  if [ -z $ANDROIDINC ] ;  then
+    failure "--android-includes must be specified for android builds"
+  fi
+
+  if [ -z $HOST ] ; then
+    failure "HOST must be specified for android builds"
+  fi
+
+  OPENBLASROOT=`rel2abs "$OPENBLASROOT"`
+  if [ -z "$OPENBLASROOT" ]; then
+    failure "Must specify the location of OPENBLAS with --openblas-root option (and it must exist)"
+  fi
+  if [ ! -f $OPENBLASROOT/lib/libopenblas.a ]; then
+    failure "Expected to find the file $OPENBLASROOT/lib/libopenblas.a"
+  fi
+  echo "Your math library seems to be OpenBLAS.  Configuring appropriately."
+
+  OPENBLASLIBS="$OPENBLASROOT/lib/libopenblas.a $OPENBLASROOT/lib/libclapack.a $OPENBLASROOT/lib/liblapack.a $OPENBLASROOT/lib/libblas.a $OPENBLASROOT/lib/libf2c.a"
+  echo "OPENBLASROOT = $OPENBLASROOT" >> kaldi.mk
+  echo "OPENBLASLIBS = $OPENBLASLIBS" >> kaldi.mk
+  echo "ANDROIDINC = $ANDROIDINC" >> kaldi.mk
+
+  cat makefiles/android_openblas.mk >> kaldi.mk
+
+  add_cross_tools
+
+  echo "Successfully configured OpenBLAS from $OPENBLASROOT."
+  echo "Configuration succeeded for platform Android"
+  exit_success
+fi
+
 
 # Check for Darwin at first, because we later call uname -o (for Cygwin)
 # which crashes on Darwin. Also the linear algebra libraries on Macs are
--- a/src/configure
+++ b/src/configure
@@ -22,10 +22,16 @@
 # ./configure --atlas-root=../tools/ATLAS/build
 # ./configure --use-cuda=no   # disable CUDA detection (will build cpu-only
 #                             # version of kaldi even on CUDA-enabled machine
+#        # Cross compile for armv8hf, this assumes that you have openfst built
+#        # with the armv8-rpi3-linux-gnueabihf toolchain and installed to
+#        # /opt/cross/armv8hf.  It also assumes that you have an ATLAS library
+#        # built for the target install to /opt/cross/armv8hf and that the
+#        # armv8-rpi3-linux-gnueabihf toolchain is available in your path
+# ./configure --static --fst-root=/opt/cross/armv8hf --atlas-root=/opt/cross/armv8hf --host=armv8-rpi3-linux-gnueabihf
 
 #This should be incremented after every significant change of the configure script
 #I.e. after each change that affects the kaldi.mk or the build system as whole
-CONFIGURE_VERSION=4
+CONFIGURE_VERSION=5
 
 function rel2abs {
   if [ ! -z "$1" ]; then
@@ -69,12 +75,14 @@ unset MKLROOT
 unset CLAPACKROOT
 unset OPENBLASROOT
 unset MKLLIBDIR
+unset HOST
 
 function usage {
   echo 'Usage: ./configure [--static|--shared] [--threaded-atlas={yes|no}] [--atlas-root=ATLASROOT] [--fst-root=FSTROOT]
   [--openblas-root=OPENBLASROOOT] [--clapack-root=CLAPACKROOT] [--mkl-root=MKLROOT] [--mkl-libdir=MKLLIBDIR]
   [--omp-libdir=OMPDIR] [--static-fst={yes|no}] [--static-math={yes|no}] [--threaded-math={yes|no}] [--mathlib=ATLAS|MKL|CLAPACK|OPENBLAS]
-  [--use-cuda={yes|no}] [--cudatk-dir=CUDATKDIR][--mkl-threading=sequential|iomp|tbb|gomp] [--fst-version=VERSION]';
+  [--use-cuda={yes|no}] [--cudatk-dir=CUDATKDIR][--mkl-threading=sequential|iomp|tbb|gomp] [--fst-version=VERSION]
+  [--host=HOST]';
 }
 
 threaded_atlas=false #  By default, use the un-threaded version of ATLAS.
@@ -84,6 +92,10 @@ static_fst=false
 use_cuda=true
 dynamic_kaldi=false
 mkl_threading=sequential
+# HOST and TARGET_ARCH are used when cross compiling, the user will specify HOST via the --host
+# switch.  TARGET_ARCH will be the first value in HOST if set, and `uname -m` otherwise
+HOST=""
+TARGET_ARCH=""
 
 cmd_line="$0 $@"  # Save the command line to include in kaldi.mk
 
@@ -187,14 +199,34 @@ do
   --fst-version=*)
     OPENFST_VER=`expr "X$1" : '[^=]*=\(.*\)'`;
     shift;;
+  --host=*)
+    # This expects the same format of host "triple" as autotools based projects
+    # this script will infer the target architecture from the specified triple.
+    HOST=`expr "X$1" : '[^=]*=\(.*\)'`;
+    shift ;;
   *)  echo "Unknown argument: $1, exiting"; usage; exit 1 ;;
   esac
 done
+
 
 # the idea here is that if you change the configuration options from using
 # CUDA to not using it, or vice versa, we want to recompile all parts of the
 # code that may use a GPU.  Touching this file is a way to force this.
 touch cudamatrix/cu-common.h 2>/dev/null
+
+
+function add_cross_tools {
+  # If the $HOST variable is set, we need to tell make to use the specified tools
+  if [ ! -z "$HOST" ]; then
+    echo '# A host triple was specified, we need to prepend all the tools with it' >> kaldi.mk
+    echo "HOST = $HOST" >> kaldi.mk
+    echo 'CC := $(HOST)-$(CC)' >> kaldi.mk
+    echo 'CXX := $(HOST)-$(CXX)' >> kaldi.mk
+    echo 'AR := $(HOST)-$(AR)' >> kaldi.mk
+    echo 'AS := $(HOST)-$(AS)' >> kaldi.mk
+    echo 'RANLIB := $(HOST)-$(RANLIB)' >> kaldi.mk
+  fi
+}
 
 function failure {
   echo "***configure failed: $* ***" >&2
@@ -207,8 +239,13 @@ function check_exists {
 }
 
 function check_for_bad_gcc {
-  if which gcc >&/dev/null; then  # gcc is on the path
-    gcc_version=$(gcc -dumpspecs 2>&1 | grep -A1 -F '*version:' | grep -v version)
+  if [ -z "$HOST" ] ; then
+    compiler="gcc"
+  else
+    compiler="$HOST-gcc"
+  fi
+  if which $compiler >&/dev/null; then  # gcc is on the path
+    gcc_version=$($compiler -dumpspecs 2>&1 | grep -A1 -F '*version:' | grep -v version)
     if [ "$gcc_version" == "4.8.2" ] || [ "$gcc_version" == "4.8.1" ]; then
       echo "*** WARNING: your version of gcc seems to be 4.8.1 or 4.8.2. ***"
       echo "*** These versions of gcc has a bug in nth_element ***"
@@ -221,16 +258,19 @@ function check_for_bad_gcc {
 }
 
 function check_for_slow_expf {
-  cd probe
-  rm -f exp-test
-  make -f Makefile.slow_expf 1>/dev/null
-  ./exp-test
-  if [ $? -eq 1 ]; then
-      echo "*** WARNING: expf() seems to be slower than exp() on your machine. This is a known bug in old versions of glibc. Please consider updating glibc. ***"
-      echo "*** Kaldi will be configured to use exp() instead of expf() in base/kaldi-math.h Exp() routine for single-precision floats. ***"
-      echo "CXXFLAGS += -DKALDI_NO_EXPF" >> ../kaldi.mk
+  # We cannot run this test if we are cross compiling.
+  if [[ "$TARGET_ARCH" == "`uname -m`" ]] ; then
+    cd probe
+    rm -f exp-test
+    make -f Makefile.slow_expf 1>/dev/null
+    ./exp-test
+    if [ $? -eq 1 ]; then
+        echo "*** WARNING: expf() seems to be slower than exp() on your machine. This is a known bug in old versions of glibc. Please consider updating glibc. ***"
+        echo "*** Kaldi will be configured to use exp() instead of expf() in base/kaldi-math.h Exp() routine for single-precision floats. ***"
+        echo "CXXFLAGS += -DKALDI_NO_EXPF" >> ../kaldi.mk
+    fi
+    cd ..
   fi
-  cd ..
 }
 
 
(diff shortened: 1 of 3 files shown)


danpovey:
You seem to have made a lot of changes to the build setup, affecting all platforms.
Our current setup supports many different platforms in many different configurations, and it is tested and generally works (however, we don't have an automated way to test this-- it's mostly based on user feedback).

It's very unlikely, without extensive testing, that such a large change would work in all the platforms and configurations.  Generally speaking it's better to make small incremental changes when it comes to the build setup-- changes that ideally only affect the use-case you are trying to support.

khers:
The first commit can be dropped, I just thought it might be useful to discard support for long unsupported OSX versions.  And if commit 2 needs to leave the darwin setup alone, that change is easy as well, but there is no clean way to get to commit 3 without modularizing the linux*.mk files.  Would making these changes to only the linux setup be more acceptable?

danpovey:
If there are Makefiles then *we* likely support those OS versions, even if
Apple does not.

See if you can support cross compilation in a way that has as minimal as
possible an effect on the configuration system.  The big change that you
made would be very hard to test adequately.
The way you were setting ARCH to a particular value in the 'configure'
script did not look like it would be back-compatible to all the ways we
currently configure and use Kaldi.


Dan

danpovey:
I think a better way to do this might be to create a .mk file specifically for linux->arm cross compilation, and add some flag specifically to ask it to use that .mk file (e.g. --arm-cross-compile=yes).
It needs to be clear from looking at your patch that it is not going affect any setup not using that flag, because changes to the configuration setup are very hard to test adequately.
Dan

danpovey:
@khers, is there any chance of you reworking this PR?  I haven't heard from you.

khers:
I have other things that need attention immediately.  I will come back to it when I free up a little.  Do you want to see this request closed until then?

danpovey:
No, we can leave it open.

danpovey:
@ngoel17, if you have time and have an ARM platform you can play around with, we need someone to extract parts of this pull request at some point.
There are aspects that we don't want to merge (removing old OS X versions), and aspects that we might be able to keep if properly tested (e.g. the reorganization of makefiles... I saw a typo about CXXFLAG [should be CXXFLAGS].).  And the most useful aspect that we do want is the ARM support.

aisch:
@danpovey i needed t build kaldi for arm (i'm not cross compiling tho) but needed to manuall remove e.g. `-msse -msse2 ` and added `-ftree-vectorize -mfloat-abi=hard -mfpu=neon` after which i was able to build and use `biglib` fine on arm. could we add at least those changes from this pr that allow native builds for linux arm?

danpovey:
I am fine to merge a PR that will enable ARM support but I don't have time to actually create a clean PR that would be merge-able without breaking other stuff.  I was asking Nagendra to do this but he is probably too busy also.  If you could do it this would be great.  I am happy to provide comments to point you in the right direction if you start working on it.

aisch:
@danpovey did you have a preference on duplicating config? e.g. in `src/makefiles/` we can just add files like `linux_atlas_armv7.mk` which are just cat'ed into `kaldi.mk` based on `uname -m` as was done for `cuda_64bit.mk`. or we can try sharing common stuff as was done in this pr (e.g. `_base_`). whichever you think is safer to merge.

danpovey:
I'm OK in principle with changing the build system to share common stuff,
as long as the resulting build system isn't too complicated, but it would
have to be tested at least on the common platforms, and I'd like to have it
such that it's fairly easy to demonstrate that for the other platforms, the
result will be the same as the current system.

danpovey:
@dogancan, you might want to see if there is anything useful in this PR that can be used in your current changes... but I'm not saying you have to do this-- we can leave it as a separate issue.  I don't want it to become a blocker for your other work.

khers:
I was finally able to come back around to this set.  It now builds on the work done for ARM and PPC native builds and then makes a few small changes to enable building for Android.

danpovey:
Cool-- thanks!
@dogancan, do you have a chance to look into this PR?  I'm OK with merging
stuff to support other platforms as long as  we can be fairly confident it
won't break our existing stuff.

src/configure
```diff
@@ -184,6 +189,29 @@ do
...
+    # the specified host triple and at a '-'
+    HOST=`expr "X$1" : '[^=]*=\(.*\)'`-;
+    shift ;;
+  --fst-version=*)
+    OPENFST_VER=`expr "X$1" : '[^=]*=\(.*\)'`;
+    shift;;
+  --target-arch=*)
+    TARGET_ARCH=`expr "X$1" : '[^=]*=\(.*\)'`;
+    shift ;;
+  --android)
```
dogancan: I see why you need a flag like this but I don't like that we will have a special flag for android cross-compilation. I mean I think we should have a more general way of specifying that we are cross compiling (something like a "--target=*" flag) and then check what the target platform is and do the things necessary for that platform. What do you think?

src/configure
```diff
@@ -896,6 +945,41 @@ fi
...
+  fi
+  if [ ! -f $OPENBLASROOT/lib/libopenblas.a ]; then
+    failure "Expected to find the file $OPENBLASROOT/lib/libopenblas.a"
+  fi
+  echo "Your math library seems to be OpenBLAS.  Configuring appropriately."
+
+  OPENBLASLIBS="$OPENBLASROOT/lib/libopenblas.a $OPENBLASROOT/lib/libclapack.a $OPENBLASROOT/lib/liblapack.a $OPENBLASROOT/lib/libblas.a $OPENBLASROOT/lib/libf2c.a"
+  echo "OPENBLASROOT = $OPENBLASROOT" >> kaldi.mk
+  echo "OPENBLASLIBS = $OPENBLASLIBS" >> kaldi.mk
+  echo "HOST = $HOST" >> kaldi.mk
```
dogancan: This was already added to kaldi.mk above.

src/configure
```diff
@@ -896,6 +945,41 @@ fi
...
 
+if $android ; then
+  OPENFSTLIBS="$FSTROOT/lib/libfst.a"
+  echo "OPENFSTLIBS = $OPENFSTLIBS" >> kaldi.mk
+
+  if [ -z $TOOLCHAININC ] ;  then
+    failure "--toolchain-includes must be specified for android builds"
+  fi
+
+  if [ -z $HOST ] ; then
```
dogancan: What happens if we are cross compiling for a platform other than android and host is not set?

src/configure
```diff
@@ -184,6 +189,29 @@ do
...
+  --android)
+    threaded_math=false;
+    static_math=true;
+    static_fst=true;
+    dynamic_kaldi=false;
+    MATHLIB='OPENBLAS';
+    TARGET_ARCH='arm';
+    android=true;
+    shift;;
+  --toolchain-includes=*)
```
dogancan: As it is this option name does not signal it has to do with android. Maybe we can call this "--android-incdir" so it is more consistent with the other option names.

src/configure
```diff
@@ -896,6 +945,41 @@ fi
 # Most of the OS-specific steps below will append to kaldi.mk
 echo "Doing OS specific configurations ..."
 
+if $android ; then
```
dogancan: What happens if we are on Mac OS or cygwin and "--android" flag is given?

src/configure
```diff
@@ -184,6 +189,29 @@ do
   --cudatk-dir=*)
     CUDATKDIR=`read_dirname $1`;
     shift ;; #CUDA is used in src/cudamatrix and src/nnet{,bin} only
+  --host=*)
+    # We are going to prepend this to compilation tool invocations, so we need to take
+    # the specified host triple and at a '-'
+    HOST=`expr "X$1" : '[^=]*=\(.*\)'`-;
+    shift ;;
+  --fst-version=*)
```
khers: From the commit log for that change:
    When cross compiling, I typically have the built openfst libraries
    installed to an architecture specific directories (without the openfst
    Makefiles).  This switch allows me to specify the openfst version
    without having to keep the Makefile installed beside the libraries.

So I have my android toolchain installed to `/opt/cross/arm-linux-androideabi`.  I have my openfst libraries installed in `/opt/cross/arm-linux-androideabi/lib` and `/opt/cross/arm-linux-androideabi/include` but the Makefile for openfst is not in the root directory.  This switch allows me to specify the openfst version without having to hand copy the Makefile into place with every openfst change (or on any new build host).

src/configure
```diff
@@ -184,6 +189,29 @@ do
   --cudatk-dir=*)
     CUDATKDIR=`read_dirname $1`;
     shift ;; #CUDA is used in src/cudamatrix and src/nnet{,bin} only
+  --host=*)
```
khers: I use the configure switches to allow my jenkins instance to build for all our supported architectures in one place.  Hand editting the kadli.mk file defeats this (or at least requires that all my builds have sed calls to replace all the tool definitions not just gcc, but ld, ar, and ranlib).

I want to note that you do not have to set the HOST variable, it is used in a way that when building natively, leaving it empty gives the current behavior.

src/configure
khers: I am fine wrapping this into the target flag, I had just hoped to have target mean architecture, not operating system.

src/configure
khers: sounds good

src/configure
khers: Do you mean we are building on a linux x86_64 system, but have set --target-arch=arm?

I hadn't thought to try that, but it seems to configure "successfully" but the build immediately fails when you try and compile because it passes arm specific flags to an x86_64 compiler.

I will update the configure script to fail is --target-arch is set to a non-native value but --host is not set.

src/configure
khers: Thanks, I will remove it

src/configure
khers: I don't have a mac or a windows machine to try it on, but I suspect it will work fine if you have the proper android toolchain available

src/configure
khers: How about triggering this off of the presence of the --android-incdir flag?

danpovey:
[this is regarding your directory locations for OpenFst]...
Eric, I don't think it makes sense to check in things that are specific to
the way you personally manage your builds.
The basic build pattern so far is that the openfst is in tools/openfst/,
and for a particular Kaldi directory, we're just building one
architecture.  People can always check out a separate copy of Kaldi if they
want to build a different architecture.
However, I haven't thought about the way this interacts with
cross-compilation, e.g. compiling on Linux for ARM, and I'm not sure how
all this interacts with OpenFst and whether its configure script supports
cross-compilation.
In any case, if possible I think we should stick with the pattern where one
Kaldi directory is used for building for one architecture, and OpenFst is
always located in the same place.  Having to prepend things to directory
names is complicated and I don't think we gain anything from it.
It's possible that there's something I don't understand here, though.

src/configure
dogancan: I see why I was confused about about this variable. You are using the host variable as a general prefix for the compiler tools -- not necessarily to specify the host architecture, right? From what I can tell the host variable represents target architecture when cross-compiling.

I don't like adding this variable to kaldi.mk when we are not cross-compiling. The kaldi.mk file is intended to be human-readable, so we are trying to keep it as simple and straightforward as possible. One way we can enable this behavior is to override the default compiler set in src/makefiles/*.mk when cross-compiling. I think overriding compiler tools via environment variables during the configure call will address this need in a more general way. So the configure call would be something like

`CXX=... ./configure ...`

We were already planning to add some partial support for this in #1249. For the moment we are exposing only CXX but we can also expose AR, AS and RANLIB if need be.

@danpovey Maybe we should do these changes after we merge #1249 if the plan is to merge it soon. On the other hand, it might be better to make these changes prior to C++11 requirement so people who prefer to stay on the 5.0 branch can also cross compile.

src/configure
dogancan: I like triggering this off of the presence of the --android-incdir flag.

src/configure
dogancan: My bad, I haven't thought to check the commit messages. @danpovey I am ok with this change. It seems like a valid use case and it won't break existing stuff.

danpovey:
BTW, for ARM compilation, it looks like there all kinds of flags regarding what kind of FPU the device has, and other options, which can very significantly affect the runtime speed:
  http://clang.llvm.org/docs/CrossCompilation.html

khers:
> Eric, I don't think it makes sense to check in things that are specific to the way you personally manage your builds. The basic build pattern so far is that the openfst is in tools/openfst/, and for a particular Kaldi directory, we're just building one architecture. 

This isn't specific to the way I personally mange a build, it supports doing things the way the *NIX systems have been working for quite a long time.  I am confused about wanting to keep the openfst libraries in tools/openfst only because the configure script already has the --fst-root switch, which allows you to point it at any directory for finding openfst.   I don't understand the final comment here, do you mean that you only intend to support building for x86?  Or that you only support building for one architecture at a time?

> People can always check out a separate copy of Kaldi if they want to build a different architecture. 

Are you suggesting that I should have to keep 3 copies of the Kaldi source on my local machine (and move code changes between them) in order to test changes on x86, ARM, and Android?

> However, I haven't thought about the way this interacts with cross-compilation, e.g. compiling on Linux for ARM, and I'm not sure how all this interacts with OpenFst and whether its configure script supports cross-compilation. 

OpenFST uses autotools, so it supports cross-compilation.

> In any case, if possible I think we should stick with the pattern where one Kaldi directory is used for building for one architecture, and OpenFst is always located in the same place. Having to prepend things to directory names is complicated and I don't think we gain anything from it. It's possible that there's something I don't understand here, though.

All of this is still supported, I am not taking anything away.  I want to try and be really clear here, if they way you currently configure kaldi is simply `configure --static --mathlib=ATLAS`, that will still work after these changes.  So I guess I don't understand the comment.

src/configure
```diff
@@ -184,6 +189,29 @@ do
...
+    # We are going to prepend this to compilation tool invocations, so we need to take
+    # the specified host triple and at a '-'
```
khers: What obscure bash syntax are you referring to?  I will add an example to the comment and see if that helps.

src/configure
```diff
@@ -433,6 +475,11 @@ function configure_cuda {
     if [ ! -f $CUDATKDIR/bin/nvcc ]; then
       failure "Cannnot find nvcc in CUDATKDIR=$CUDATKDIR"
     fi
+
+    if [[ "$TARGET_ARCH" != "`uname -m`" ]] ; then
+      failue "Cannot cross compile with CUDA support"
```
khers: Thanks, will fix for V2

khers:
> I'm a little confused why you're using HOST as the general compiler prefix... it seems that variable has to do more with the target than the host in general.
In http://bill.station51.net/index.php?post/2010/05/08/HOWTO%3A-Setting-up-an-ARM-toolchain,
they seem to be using a variable named CROSS_COMPILE for this:
export CROSS_COMPILE=arm-none-linux-gnueabi-
or maybe COMPILER_PREFIX would be more transparent, since it clearly indicates what it is.

I don't know who wrote that blog post, or what project he is referring to, but I modeled the options after what someone familiar with autotools would expect.  From the output of `configure --help` in openfst:

```
System types:
  --build=BUILD     configure for building on BUILD [guessed]
  --host=HOST       cross-compile to build programs to run on HOST [BUILD]
```

The --host switch controls where the program is supposed to run.

> With the flags --host and --target-arch, can you explain what those complicated bash expressions are doing? Are they really just adding a dash at the end, or is it something more complicated, and couldn't that just have been done by the user when supplying the flag? That would make the script much more self-explanatory.

The code for extracting the values passed in with the switch is directly lifted from options already in the configure script:

```
  --mathlib=*)
    MATHLIB=`expr "X$1" : '[^=]*=\(.*\)'`;
    shift ;;
```

The --target-arch and --fst-version flags are a direct lift of this.  The - character following the backticks in the --host option is a bare string.  Would it be more helpful if I parsed the option and then concatenated the '-' afterwards?  I added the '-' to the supplied argument to match the way autotools works (and to avoid confusing folks that are used to that.

> Are you trying to take just part of the --target option? Because normally, from what I can tell, in cross-compilation, 'target' is a triple. Deviating from this would IMO be confusing.

autotools expects something like `arm-linux-androideabi` as the argument to --host, so we could extract the value of target-arch from the host triple.  I was opting for more explicit here, but if it is preferable to drop the --target-arch flag and get the value from the host triple, I can do that.

khers:
> BTW, for ARM compilation, it looks like there all kinds of flags regarding what kind of FPU the device has, and other options, which can very significantly affect the runtime speed:
http://clang.llvm.org/docs/CrossCompilation.html

My needs for ARM match the ones from the earlier contributor that merged support for building ARM, so I want to reuse the options he put in as default.  I don't see any reason to add configurations for supporting software floating point or soft-fp at this point.

src/configure
khers: Will make that change for V2

src/configure
khers: If you want to wait on this set until #1249 merges, I can rebase it and use that instead.  But doing it with the --host flag allows the user to specify the host once, instead of 5 times (for all 5 tools used).

danpovey:
OK, I see now that the bash `expr` stuff was used elsewhere in the
script, for parsing --option-name=value arguments.

I had been trying read online about cross-compilation and had got a
bit confused about the meaning of 'host' and 'target'... some of the
explanations seem to be geared towards building compilers.

For the --host thing, I think your comment might be clearer with an
example, e.g.
# if the user specifies --host=arm-linux-androideabi, the
# HOST variable is set to "arm-linux-andoideabi-"

I think it would be helpful if you could illustrate example commands
including the --host flag, with comments at the top of the script,
you'll see there are a bunch of examples there.  I would err on
the side of adding more, rather than fewer, examples, and
plenty of explanation.
It may be unclear to users (actually, it's unclear to me at this point)
how the --host and --target-arch options are related.  You mentioned
that it may be possible to work out the --target-arch from the host...
IMO this might be preferable if it reduces confusion.   But I don't have
experience with cross-compilation.  @dogancan
may have comments too...

dogancan:
I agree that having both --host and --target-arch options is confusing and we should derive the target architecture from the host triple.

One final suggestion I have is to prefix HOST to compiler binaries by overwriting the values for CXX, CC, AR, AS and RANLIB when cross-compiling instead of altering src/makefiles/*.mk. I mean we will have a check in configure which overwrites these variables when host is set. The kaldi.mk file will look like the following
```
# These are the defaults defined in makefiles/*.mk
CXX = g++
CC = g++
...

# This part will be printed only when we are cross compiling
# As a bonus we do not need to add a "-" at the end of host triple :)
CXX = $(HOST)-$(CXX)
CC = $(HOST)-$(CC)
...
```

I don't think there is any need to wait for #1249 after these changes.

danpovey:
I think your proposal makes sense.

vdp:
While I haven't spent time to study the changes in any detail, I just want to say that having more flexibility in terms of where the things are installed would be great. In fact one of the very few things I don't like about Kaldi, is that the current build system doesn't(AFAIK) allow building stuff under a directory other than src/. As Eric has pointed out having a different source directory for each build is cumbersome when you make modifications. What I think would be good to have is an option like "--build-root", that would allow to have different versions of the binaries, such as:

     <some-dir>
             |
             --  x86
             |      |
             |      debug-build (say compiled with -O0 -g)
             |      |       |
             |      |       -- featbin
             |      |       |
             |      |       -- latbin
             |      |       |
             |      |       -- <etc> 
             |      |
             |      optimized-build
             |      |
             |      |
             |      my-fancy-feature-build
             |      |
             |      <etc>
             |
             -- arm32
             |
             -- arm64


I know this is likely outside the scope of this PR, but I noticed the exchange about OpenFst, and thought that this may be worth considering in future..

danpovey:
I'm open to doing such things, although I can't see myself having time to
personally figure out how to do it anytime soon.

khers:
>I agree that having both --host and --target-arch options is confusing and we should derive the target architecture from the host triple.
>
>One final suggestion I have is to prefix HOST to compiler binaries by overwriting the values for CXX, CC, AR, AS and RANLIB when cross-compiling instead of altering src/makefiles/*.mk. I mean we will have a check in configure which overwrites these variables when host is set. The kaldi.mk file will look like the following

Awesome, I will make the suggested changes and update this pull request.

khers:
>I'm open to doing such things, although I can't see myself having time to personally figure out how to do it anytime soon.
 @vdp 
FWIW, I carry a bit of a hack that adds a make install target which would allow a user to install the resulting binaries (and libraries) but it is not in any condition for review.  If there is interest in that work, I can clean it up and post a pull request.

I also have a few changes that I would like to propose for the way intermediate libraries are built.

vdp:
@khers 
I personally think that something like this would be nice to have. I guess this hack you are referring to is just walking through the src subdirectories and is copiying binary stuff to a user-specified location, and not a full blown solution that is managing all the dependencies between sources, object files, and libs/exes? The latter would make possible to avoid full recompilation in some(most?) cases(e.g. when you switch architectures), but I suspect could be too complicated to implement (I don't have experience with this myself)..

khers:
Just checking in on this series, did I catch all the changes that were requested?

src/configure
```diff
@@ -514,8 +593,13 @@ function linux_configure_speex {
...
 
 function fix_cxx_flag {
-  CXXCOMPILER=`grep "CXX = " kaldi.mk | awk '{print $3}'`
-  if [ $CXXCOMPILER=="g++" ]; then
+  USINGGXX=`grep -c "CXX = g++" kaldi.mk`
+  if [ $USINGGXX -ge 1 ]; then
+    if [ -z "$HOST" ] ; then
+      CXXCOMPILER="g++"
+    else
+      printf -v CXXCOMPILER "%s-g++" $HOST
```
danpovey: I think doing
CXXCOMPILER="$HOST-g++"
would be more readable, I had to look this up.

src/configure
```diff
@@ -84,6 +100,9 @@ static_fst=false
 use_cuda=true
 dynamic_kaldi=false
 mkl_threading=sequential
+HOST=""
+TARGET_ARCH=""
```
danpovey: Can you please add a comment above these, saying something like the following (edited for accuracy):
```
# HOST and TARGET_ARCH are only set on the command line in case of cross-compilation.
# If HOST is not set, TARGET_ARCH will be set to `uname -m` inside the script.
```

src/configure
```diff
@@ -558,7 +642,12 @@ function linux_check_static {
   if [ -f $dir/libatlas.a ]; then # candidate...
     # Note: on the next line, the variable assignment
     # LANG=en_US should apply just to the program called on that line.
-    if LANG=en_US gcc -o test_linking test_linking.cc -u ATL_flushcache $dir/libatlas.a 2>&1 | grep -i "incompatible" >/dev/null; then
+    if [ -z "$HOST" ] ; then
+      compiler="gcc"
+    else
+      printf -v compiler "%s-gcc" $HOST
```
danpovey: also please remove the printf -v here.

src/configure
```diff
@@ -251,14 +310,29 @@ function check_library {
...
 is_set $MKLROOT && echo "Force-configuring KALDI to use MKL"&& export MATHLIB="MKL"
 is_set $CLAPACKROOT && echo "Force-configuring KALDI to use CLAPACK"&& export MATHLIB="CLAPACK"
 is_set $OPENBLASROOT && echo "Force-configuring KALDI to use OPENBLAS"&& export MATHLIB="OPENBLAS"
 
+
+# If HOST is specified, parse the TARGET_ARCH, otherwise use uname -m
+if [[ "$HOST" == "" ]] ; then
+  TARGET_ARCH="`uname -m`"
+else
+  IFS='-' read -ra PARTS <<< "$HOST"
```
danpovey: please indicate what's going on here using an example, this code is quite complicated.

dogancan:
One thing I noticed is that we now have an almost completely separate code branch in the configure script for android cross-compilation. This is in contrast to how we handle the the cross compilation configuration for armv8hf and ppc64le. While I think I see the reasons why Eric chose to do it this way, I am also wondering if we can do this in a more principled way.

Just to be clear, I'm happy with the current state of this PR. I'm bringing this up just to get your opinions.

danpovey:
I actually prefer to keep things as separate as possible, it makes it
easier to verify that other configurations won't get broken when a change
is made.

dogancan:
Agreed. I think if we had separate branches for all target platforms with shared routines for finding math libraries and stuff that would be ideal. Anyway, I'll think about this after we merge the shortcut branch since we made significant changes to the configure script as part of the C++11 upgrade.

danpovey:
Do you think it's worth testing whether merging this patch into shortcut
would cause conflicts?  We could merge directly into shortcut, would avoid
hassles later.

dogancan:
Yes this will definitely cause merge conflicts. It will be hard for Eric to resolve the conflicts so it's probably best if I merge this PR locally and open a new PR against the shortcut branch. If there are no outstanding issues other than the last 2 comments regarding documentation, I can address those while resolving the conflicts.

danpovey:
I'll wait for Eric to address those small comments then merge it to master.

khers:
I think this last iteration covers all the requests.

danpovey:
did you push the changes?  I don't see them

danpovey:
it's  ok, they are there ('commits' tab shows  weird stuff).

danpovey:
I would have liked to wait for the travis build to finish but I think you are doing something odd when you commit (dates are wrong, and commit doesn't show up), which may be confusing it.
@keli78, could you please check out this PR and just make sure it compiles on our grid?

keli78:
Sure. Will test it soon.

keli78:
I test it on one of the b machines (in src/, ./configure --shared, make depend, make) and it compiles successfully.

danpovey:
Thanks everyone!  Merged.

dogancan:
@khers I am working on porting these changes to the shortcut branch. I would really appreciate if you can answer the following questions.

1. I noticed that the android openblas makefile (`src/makefiles/android_openblas.mk`) is adding the `-DHAVE_CLAPACK` flag to `CXXFLAGS`. Is this intentional? Since this setup is using openblas, I don't think having this flag is right. When that flag is set, Kaldi does not include the header files that come with openblas.

2. I also noticed that the default compiler for the android build is set to `clang++`. Is this a hard requirement or your preference? Does the same makefile work with `g++`?

danpovey:
Definitely any OPENBLAS build should define -DHAVE_OPENBLAS.

dogancan:
It defines both at the moment.

khers:
@dogancan 

1. That is a mistake on my part and is probably left over from when I initially tried getting things to work for android.  Would you like a PR for a fix to that?

2. It is a requirement, clang is the only supported compiler for native development on android and has been for a while.  g++ is still available last time I checked, but it is forzen at gcc-4.9 and will not be updated.  The makefile should work with the g++ available for android, but I didn't test it as that compiler is deprecated.

dogancan:
@khers Thanks. No need for a PR but do you mind testing if your android build works without the `-DHAVE_CLAPACK` flag? If `-DHAVE_CLAPACK` is specified along with `-DHAVE_OPENBLAS`, `src/matrix/kaldi-blas.h` includes clapack headers instead of openblas headers.

khers:
@dogancan When I remove `-DHAVE_CLAPACK` everything continues to work.

PR #727, opened 2016-04-21
Iban recipe

jtrmal:
Adding first version of the recipe.
I just noticed I accidently added more versions of the multisplice recipe, that will be cleaned up.
Still WIP but any comment welcome on the basic infrastructure and anything before nnet training


jtrmal:
@vijayaditya @danpovey  I think it's ready for review
@sarahjuan FYI 
I will have to prepare openslr upload, yet, i.e. the recipe is not runnable anywhere else.

vijayaditya:
started review

egs/iban/s5/cmd.sh
```diff
@@ -0,0 +1,3 @@
+export train_cmd="run.pl --max-jobs-run 32"
+export decode_cmd="run.pl --max-jobs-run 32"
+export cuda_cmd="run.pl --max-jobs-run 4"
```
vijayaditya: I don't think the AWS g2 instances have multiple GPUs per node.

danpovey: We shouldn't be using cuda_cmd in new recipes at all.  That was an nnet1
thing.

Dan

jtrmal: g2.8xlarge
<https://www.google.com/search?client=safari&rls=en&q=g2.8xlarge&ie=UTF-8&oe=UTF-8>
does
(4)
the workshop expenses are calculated assuming that one.
y.

egs/iban/s5/local/online/run_nnet2_multisplice.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+
```
vijayaditya: I think it might be better to move to nnet3. We are no longer "supporting" this version.

jtrmal:
removed cuda_cmd (wasn't used anyways). Removed online/nnet2 recipe, added nnet3.
New recipe is ~ 1% abs worse than the nnet2 recipe. THe splice indices must be tuned/converted from nnet2 recipe.

danpovey:
How do the train and valid objfs compare in the nnet2 versus nnet3 recipes?
And are you using the same num-epochs, learning rates, splicing etc., in
the nnet3 recipe?
Dan

jtrmal:
The splicing is is taking from WSJ (I didn't find the nnet3 splicing in RM,
from which I took the nnet2 splicing).
The format is different, so I hve to look into docs, how it should be
converted. The same thing for the #hidden layers (which is worked out from
the splicing description in nnet3).
y.

danpovey:
I suspect instead of
 "-4,-3,-2,-1,0,1,2,3,4 0 -2,2 0 -4,4 0"
something like
"-2,-1,0,1,2 -3,1 -5,3 0 0"
which is I think what you have in the nnet2 recipe,
would work better.
Dan

jtrmal:
Yes, that's what I'm running now. Will have results in a 30 mins or so
BTW, Are the hires mfcc features always helping? That was another
substantial change I did -- the previous recipe used normal mfccs only. Not
sure if this could be causing sparsity issues (either during ivecs or
during nnet training)
y.

danpovey:
We pretty much use them everywhere except RM, and I think they do help.
I think on so little data it might make sense to do the speed-perturbing.
It's hard to say without looking at the train/valid probs whether the
problem is undertraining or overtraining.
Dan

danpovey:
BTW, when you do the speed perturbing you can reduce the num-epochs, so it
doesn't have to take substantially longer.

vijayaditya:
For the tutorial sake reducing the number of epochs might be desirable. But
Tom usually saw better performance when he ran the perturbed setups for
larger number of epochs.

--Vijay

egs/iban/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+if [ $stage -le 4 ]; then
+  steps/online/nnet2/train_diag_ubm.sh --cmd "$train_cmd" --nj 16  --num-frames 200000 \
+     data/train_hires 256 exp/nnet3/tri5b exp/nnet3/diag_ubm
+fi
+
+if [ $stage -le 5 ]; then
+  # even though $nj is just 10, each job uses multiple processes and threads.
+ bash -x steps/online/nnet2/train_ivector_extractor.sh --cmd "$train_cmd" \
```
vijayaditya: If debugging is complete you could remove `bash -x`

egs/iban/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -0,0 +1,71 @@
...
+    --nj 10 --num-processes 1 --num-threads 2  --ivector-dim 50\
```
vijayaditya: why the reduced dimension of ivectors ? I haven't seen an nnet2/nnet3 experiments which measures the impact of ivector dimension. If you want to reduce parameters in the model you could probably reduce the splicing config.

egs/iban/s5/local/nnet3/run_tdnn.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# this is the standard "tdnn" system, built in nnet3; it's what we use to
```
vijayaditya: I recommend developers to add the name of the recipe/script from which the current script was adapted. This helps understand some choices in case we are trying to tune things in the future.

egs/iban/s5/local/nnet3/run_tdnn.sh
```diff
@@ -0,0 +1,70 @@
...
+
+
+# At this script level we don't support not running on GPU, as it would be painfully slow.
+# If you want to run without GPU you'd have to call train_tdnn.sh with --gpu false,
+# --num-threads 16 and --minibatch-size 128.
+
+stage=0
+train_stage=-10
+dir=exp/nnet3/nnet_tdnn_a
+. cmd.sh
```
vijayaditya: . ./cmd.sh

egs/iban/s5/local/nnet3/run_tdnn.sh
```diff
@@ -0,0 +1,70 @@
...
+
+if [ $stage -le 8 ]; then
+  #  --splice-indexes "-4,-3,-2,-1,0,1,2,3,4  0  -2,2  0  -4,4 0" \
+  #  --num-hidden-layers 5 \
+
+  steps/nnet3/train_tdnn.sh --stage $train_stage \
+    --num-jobs-initial 2 --num-jobs-final 4 \
+    --splice-indexes "-4,-3,-2,-1,0,1,2,3,4  0  -2,2  0  -4,4 0" \
+    --num-epochs 7 \
+    --add-layers-period 1 \
```
vijayaditya: I haven't actually measured how much the layer-wise discriminative training helps and how long it should be done. But until you test the impact I would recommend `--add-layers-period 2`. If there aren't sufficient iterations for this you could increase the number of epochs.

jtrmal:
the 7 epochs translate to 16 iteration or so, so it's quite fast (and I
couldn't go much lower than that because of the layers being added after
each epoch).

vijayaditya:
Sorry I was actually responding to @danpovey 's suggestion of doing speed perturbation and reducing number of epochs (e.g. 3-fold perturbation and reduce epochs to 3). I would recommend speed perturbation and but without reducing the number of epochs. 

In the last message I was saying that reducing the number of epochs after perturbation might be desirable for the tutorial, but since the amount of data is small you might actually see a benefit by increasing the epochs when the tutorial is done.

jtrmal:
I see. My original thinking was not introduce the speed perturbation at
all, given the fact that we are trying to make an tutorial scripts (at
least for now), so tried to make everyithing quite basic. For more advanced
techniques we have other corpora.
The only thing I'd like to have is "basic" DNN recipe that is actually
working better than SGMM :)
The nnet1 actually works better (but it includes sequence training, I
belive) and the nnet2 was ~0.5% better than the sgmm.


BTW, the splicing "-2,-1,0,1,2 -3,1 -5,3 0 0" (adapted from the previous
recipe) yielded significantly worse results (23 % and 20 % after rescoring).
y.

danpovey:
Speed perturbation is an easy win because no-one is forcing you to run it
for longer than you would normally- you don't have to increase the
num-epochs proportionally.

We do have sequence training working for nnet3 now.  It works for nnet2
also.  However it doesn't always give very big improvements in either
case.  The recipes in nnet2 and nnet3 are more tuned and optimized for
larger datasets, while nnet1 is tuned and optimized for smaller datasets
and uses techniques like pretraining and early stopping that are
particularly helpful for smaller datasets.  I think without adding speed
perturbation we're not really taking advantage of the capabilities of nnet3
(e.g. parallelization).

Dan

vesis84:
Wow :) Dan, thanks for the suggestion :) I should give a try to the speed perturbation definitely :D :D :D

danpovey:
Sorry Karel, I believe nnet3 is the future.  But it's good that you are
keeping me on my toes.

jtrmal:
So using speed pert. I was finally able to get system working at least a little bit better than sgmm.
Not being able to keep on tweaking that -- got much more other things to address for the workshop. 
@vijayaditya , it should be ready for the guinea-pig run.

danpovey:
What is the root dir for this system?

jtrmal:
It's not at the JHU cluster.

danpovey:
Oh; I wanted to see the compute_prob_*.final.log in some of the key nnet
builds, to see whether it's overtraining or undertraining that's the
problem.

vijayaditya:
@jtrmal  OK, I will run this from scratch on the JHU cluster.

jtrmal:
@danpovey,  ok to merge?  I think it would be easier for Vijay to run (and
eventually submit fixes)
Y.

jtrmal:
Squashed the commits and will merge in a couple of minutes.

PR #733, opened 2016-04-27
nnet3: fix to ConstantFunctionComponent

david-ryan-snyder:
Fix to ConstantFunctionComponent so that it reads and writes its learning rate. This fixes a bug which prevented it from being updated via nnet3-copy during training.

NOTE: This fix will break any existing models that use ConstantFunctionComponent. However, I don't think this component is widely used outside of a few experimental branches.

### DIFF
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -2141,10 +2141,35 @@ void ConstantFunctionComponent::Backprop(
 }
 
 void ConstantFunctionComponent::Read(std::istream &is, bool binary) {
-  ReadUpdatableCommon(is, binary);  // Read the opening tag and learning rate
-  ExpectOneOrTwoTokens(is, binary, "<ConstantFunctionComponent>",
-                       "<InputDim>");
-  ReadBasicType(is, binary, &input_dim_);
+  std::string token;
+  ReadToken(is, binary, &token);
+  if (token == "<ConstantFunctionComponent>") {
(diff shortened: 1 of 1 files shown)


src/nnet3/nnet-simple-component.cc
```diff
@@ -2140,6 +2141,7 @@ void ConstantFunctionComponent::Backprop(
 }
 
 void ConstantFunctionComponent::Read(std::istream &is, bool binary) {
+  ReadUpdatableCommon(is, binary);  // Read the opening tag and learning rate
   ExpectOneOrTwoTokens(is, binary, "<ConstantFunctionComponent>",
```
danpovey: This should just be ExpectToken, because by this point you've already read the opening tag.

danpovey:
Are you sure you don't want to work on the reading code to make it back compatible?  This would save you and maybe others a bunch of hassle.  You can replace ReadUpdatableCommon() with a modified version of its code.

david-ryan-snyder:
I didn't think anythin uses this component, outside of the xvector branch.

danpovey:
but might Pegah be using it.  And do you really want to invalidate all the
models you have on disk?

On Wed, Apr 27, 2016 at 5:06 PM, david-ryan-snyder <notifications@github.com

david-ryan-snyder:
OK, no problem.

david-ryan-snyder:
I added the backwards compatibility, and verified that it works for the old models.

PR #737, opened 2016-04-29
Cpplint style

vesis84:
- Refactoring 'nnet1' C++ code to make it closer to 'cpplint' style,
- Slight changes in 'Nnet' interface 
  - now it's safer manipulation of components, 
  - refactored GetWeights, SetWeights, GetGradient - from 'pull' to 'push',
- Small extensions of 'cudamatrix' (mostly CuArray),

@danpovey : can you please peek at the 'cudamatrix' code?


jtrmal:
@vesis84 , the travis build failed.
```
../cudamatrix/cu-array-inl.h: In member function ‘T kaldi::CuArray<T>::Min() const’:
../cudamatrix/cu-array-inl.h:262:12: error: ‘min_element’ is not a member of ‘std’
   T ans = *std::min_element(tmp.begin(), tmp.end());
```
min_element is AFAIK C++17 feature. Use just std::min (you might need to include `<algorithm>` for it).
From the code, you seem to look for the value itself, not the iterator, so std::min is actually more suitable.

jtrmal:
Sorry, I take that back. Still half asleep. std::min wouldn't help. min_element is the right call.
Not sure why it fails, perhaps it really needs just the `<algorithm>` included?
y.

vesis84:
Okay, thanks! The Travis passed now...

danpovey:
I'm  commenting via email because the diff is very big and seems to be
crashing github's interface.

It's good that you're cleaning up the style.

Regarding Min() and Max(), you need to define their behavior for empty
vectors.  The current code would segfault.

It's a little lazy to define Add() in the header.  Better to declare the
template override in the header and define the function in the .cc file
(better for compilation time, and this is the general policy).  You can
lose the "inline" keyword.

vesis84:
Hi,
I just commited the change. Now the header contains declaration of Add(),
the cu-array-inl.h contains templated method and declaration of the 
specialized method.
The implementation of speciazlized method is in cu-array.cc file (new file).

Also for Min(), Max(), there is a new assert checking the dimension is >0.

Commit:
https://github.com/kaldi-asr/kaldi/pull/737/commits/b07db1754e572dab3773138964a1638fa6cdcfc2

Waiting for the Travis now...

K.

danpovey:
OK.  Karel, please don't merge this yourself... I want to wait till it's fully reviewed.
I notice a lot of files have lines like this:
-    std::cerr << e.what();
     return -1;
   }
 }
perhaps you did this from some kind of script?  Those statements are supposed to be there.

Please document by the declaration of Min() and Max() that they will crash when called with the empty vector.  This differs from the behavior of Min() and Max() for class Vector, that will return -inf and +inf in those cases.

vesis84:
Hi, okay, I'll wait. The comments on Min() Max() are added.

std::cerr << e.what(); return -1; } } <<= This line was redundant, the error message is already printed in the destructor of class 'MessageLogger' instantiated be 'KALDI_ERR' macro (base/kaldi-error.cc:168). The command would print it 2nd time, which I intended to avoid. It was polluting the logfiles, making them very confusing.

danpovey:
Regarding
std::cerr << e.what();
this is a general policy in all Kaldi programs; if it is to be changed it
should be changed everywhere and not just in the nnet1 scripts.  But simply
removing it is not the right answer though, because if you get an
allocation failure, that is the only place that 'std::bad_alloc' would be
printed; without this line, that would just lead to a truncated log.  So
please put it back; we'll think separately about whether there is a way to
make this less confusing.
Dan

vesis84:
Okay, I'll put those lines back tomorrow... Need to look at it with clear head...

vesis84:
Hi, I added back the 'std::cerr << e.what()' for the exceptions in the nnet1 binaries.
I am looking into the error messages. I'll cover that in a separate WIP PR...

danpovey:
Thanks.
Replying by email rather than on specific lines, as this PR is crashing
github for me.

In error messages from shell scripts, the program name usually goes at the
start, not the end of the line, so it would be good to be consitent here,
e.g.
  echo "$0: [your error message]"

Dan

danpovey:
I just noticed that there are still changes to the end-of-program thing:

-  } catch(const std::exception& e) {
-    std::cerr << e.what() << '\n';
+  } catch(const std::exception &e) {
+    std::cerr << e.what();
     return -1;
   }

vesis84:
I. Script messages: '$0: message' This is now done!
II. Ends of programs: Yes, there are minor changes, these make consistent ends for all the nnet1 binaries... No need to worry about these...

danpovey:
Hm.  I just noticed that there seems to be a lot of inconsistency about
this in Kaldi programs in general.  Later we can consider what to do about
this.  I'm giving comments on other aspects of the PR by email as github
interface is broken (previously I only glanced at the PR).  I had to
manually create a diff and look at it.

Here:

egs/wsj/s5/steps/nnet/train.sh
+echo "$0: Successfuly finished! '$dir'"

please don't use exclamation mark.  It could cause an error if history
expansion is enabled for non-interactive shells (depends on .bashrc).

For the function cudaI32_set_const, please name it cuda_int32_set_const.
Actually the lower-case function names are not really right according to
the Google style, but if we're going to do it for CUDA kernels, let's be
consistent.

Here:

+/// Wrapper for reading,
+template<typename T>
+void ReadIntegerVector(std::istream& in, bool binary, CuArray<T>* vec);
+
+/// Wrapper for writing,
+template<typename T>
+void WriteIntegerVector(std::ostream& out, bool binary, const CuArray<T>&
vec);
+


please just make them Read() and Write() functions.  The only reason for
those functions existing for vector, is that we can't add members to a
class that we don't own.

Here:

-  Component* Copy() const { return new Softmax(*this); }
-  ComponentType GetType() const { return kSoftmax; }
+  Component* Copy() const {
+    return new Softmax(*this);
+  }
+
+  ComponentType GetType() const {
+    return kSoftmax;
+  }

actually the general approach is to put functions on a single line if they
fit in 80 chars
(and if there is only one command), so better to leave it as it was.  [I'd
be surprised
if cpplint.py would have complained about the original code..]  This issue
is repeated many times in different places.

I notice in many places you are moving the : to the previous line:
-  Tanh(int32 dim_in, int32 dim_out)
-    : Component(dim_in, dim_out)
+  Tanh(int32 dim_in, int32 dim_out) :
+    Component(dim_in, dim_out)
   { }

It would be better to avoid the space before the :, so make it
 Tanh(int32 dim_in, int32 dim_out):
because that's the way most of the rest of the code does it.  Not a big
deal.

And have the { } on he same line as the previous initializer, which is the
way most of the rest of the code does it, so:
  Tanh(int32 dim_in, int32 dim_out):
    Component(dim_in, dim_out)  { }

That's probably all.

vesis84:
Hi, I did integrate the feedback. Everything is changed, except the position of the empty blocks of constructors. It is better visible if it is on the new line, and it is not against the cpplint style.

danpovey:
It is against the style, see this:

"Parting Words

Use common sense and BE CONSISTENT.

If you are editing code, take a few minutes to look at the code around you
and determine its style. If they use spaces around their if clauses, you
should, too. If their comments have little boxes of stars around them, make
your comments have little boxes of stars around them too.

The point of having style guidelines is to have a common vocabulary of
coding so people can concentrate on what you are saying, rather than on how
you are saying it. We present global style rules here so people know the
vocabulary. But local style is also important. If code you add to a file
looks drastically different from the existing code around it, the
discontinuity throws readers out of their rhythm when they go to read it.
Try to avoid this."

The rest of the Kaldi code puts the { } on the same line, so yours should
too.  But I'm not willing to put further energy into arguing with you about
this, so I'll just merge the changes as-is.

PR #738, opened 2016-04-29
Fix an apparent typo in nnet3::ComputeNnetComputationEpochs()

vdp:
I was reading some nnet3 sources and while I'm not familiar with this code(yet!) it seems to me that the first argument to [ComputeTopSortOrder()](https://github.com/kaldi-asr/kaldi/blob/f5c1515b643b0badda7a3ca28ca8e9117032b5f4/src/nnet3/nnet-graph.cc#L286) should be scc_graph_transpose. I don't know yet how this fits into the whole computation steps compilation story - I'd guess it probably doesn't matter - but I think we should either apply this patch or remove the call to ComputeGraphTranspose(), along with the relevant comments.

### DIFF
--- a/src/nnet3/nnet-graph-test.cc
+++ b/src/nnet3/nnet-graph-test.cc
@@ -135,14 +135,14 @@ void BuildTestTopSortOrder(std::vector<int32> *node_to_order) {
 
   // The topological sorting order of the above SCC graph is as follows (from
   // our particular algorithm):
-  // 0 --> 3
-  // 1 --> 2
-  // 2 --> 1
-  // 3 --> 0
-  (*node_to_order)[0] = 3;
-  (*node_to_order)[1] = 2;
-  (*node_to_order)[2] = 1;
-  (*node_to_order)[3] = 0;
+  // 0 --> 0
+  // 1 --> 1
+  // 2 --> 2
+  // 3 --> 3
+  (*node_to_order)[0] = 0;
+  (*node_to_order)[1] = 1;
+  (*node_to_order)[2] = 2;
+  (*node_to_order)[3] = 3;
 }
 
 void UnitTestComputeGraphTranspose() {
--- a/src/nnet3/nnet-graph.cc
+++ b/src/nnet3/nnet-graph.cc
@@ -230,18 +230,18 @@ void ComputeTopSortOrder(const std::vector<std::vector<int32> > &graph,
   std::vector<bool> cycle_detector(graph.size(), false);
   std::vector<bool> is_visited(graph.size(), false);
 
-  std::vector<int32> reversed_orders;
+  std::vector<int32> orders;
   for(int32 i = 0; i < graph.size(); ++i) {
     if (!is_visited[i]) {
       ComputeTopSortOrderRecursive(i, graph, &cycle_detector,
-                                   &is_visited, &reversed_orders);
+                                   &is_visited, &orders);
     }
   }
 
-  KALDI_ASSERT(node_to_order->size() == reversed_orders.size());
-  for (int32 i = 0; i < reversed_orders.size(); ++i) {
-    KALDI_ASSERT(reversed_orders[i] >= 0 && reversed_orders[i] < graph.size());
-    (*node_to_order)[reversed_orders[i]] = graph.size() - i - 1;
+  KALDI_ASSERT(node_to_order->size() == orders.size());
+  for (int32 i = 0; i < orders.size(); ++i) {
+    KALDI_ASSERT(orders[i] >= 0 && orders[i] < graph.size());
+    (*node_to_order)[orders[i]] = i;
   }
 }
 
@@ -283,7 +283,7 @@ void ComputeNnetComputationEpochs(const Nnet &nnet,
   ComputeGraphTranspose(scc_graph, &scc_graph_transpose);
 
   std::vector<int32> scc_node_to_epoch;
-  ComputeTopSortOrder(scc_graph, &scc_node_to_epoch);
+  ComputeTopSortOrder(scc_graph_transpose, &scc_node_to_epoch);
   if (GetVerboseLevel() >= 6) {
     std::ostringstream os;
     for (int32 i = 0; i < scc_node_to_epoch.size(); i++)
--- a/src/nnet3/nnet-graph.h
+++ b/src/nnet3/nnet-graph.h
@@ -45,8 +45,8 @@ std::string PrintGraphToString(const std::vector<std::vector<int32> > &graph);
 /// where graph->size() == nnet.NumNodes().  For each node-index
 /// n, the vector in (*graph)[n] will contain a list of all the nodes
 /// that have a direct dependency on node n (in order to compute them).
-/// For instance, if n is an input node, (*graph)[n] will be the empty
-/// list because it won't depend on anything.
+/// For instance, if n is the output node, (*graph)[n] will be the empty
+/// list because no other node will depend on it.
 void NnetToDirectedGraph(const Nnet &nnet,
                          std::vector<std::vector<int32> > *graph);
 


vdp:
Hmm, it seems something is not quite right with this change- causes some tests to fail..

danpovey:
Vassil, that was a bug, but it looks like it is counterbalanced by a corresponding bug in ComputeTopSortOrder, in which the order is purposely reversed, see nnet-graph.cc:244.  Testing code is also reversed, see BuildTestTopSortOrder in nnet-graph-test.cc.  Could you please try also fixing those other things and seeing if the build succeeds?

Dan

vdp:
I updated the patch and all the tests now pass, **BUT** after staring a bit more at that code, it seems to me that the original version was correct after all. The comment for ComputeTopSortOrder() says:

    /// Given an acyclic graph (where each std::vector<int32> is a list of
    /// destination-nodes of arcs coming from the current node), compute a
    /// topological ordering of the graph nodes.

Given the expected input format, the DFS traversal based implementation and under the usual notion of topo-sorting, I think the reversal of the node order in the original code was correct(i.e. not just a counter-balancing "hack").

I think we should leave the original code as it is, and just remove the transposing code

    std::vector<std::vector<int32> > scc_graph_transpose;
    // compute transpose because we actually want the reverse of the topological
    // order so inputs come first and then outputs.
    ComputeGraphTranspose(scc_graph, &scc_graph_transpose);
 
as it's unnecessary. 
(and while we are at this we can probably fix the last 2 lines of the comments to NnetToDirectedGraph() as in the current patch, because I think they are inconsistent both with the rest of the comment and with the actual implementation)

danpovey:
Makes sense.

danpovey:
Oh, wait, no, that wasn't right.
The directed graph we're dealing with is a graph of *dependencies*, so the
output nodes have lines going to previous nodes.
The top-sort code was not right.  In DFS-based traversal you'd start with
the topologically first nodes.
So I think what I originally said was right.
Dan

vdp:
The point is that I think that in the original code the input to ComputeTopSortOrder() is not in "dependencies" but **depends_on_this**(to use the terminology of the GraphBuilder) format, and that's exactly what the function expects according to the comments.
Let me try to spell out the steps in ComputeNnetComputationEpochs() one by one and see where our interpretations of the code diverge. For concreteness let's concentrate on the first nodes of a neural net. Say node 0 is input, node 1 is ivector and node 2 is LDA transform, or rather its input descriptor.

* NnetToDirectedGraph(nnet, &graph) - after execution "graph" encodes the *depends_on_this* information for the NetworkNodes, so for our example graph[0] and graph[1] both contain { 2 }

* FindSccs(graph, &sccs) - after completion "sccs" holds the SCCs for "graph". A potential source of confusion for someone who's tracing the execution in a debugger could be that the SCCs are numbered in reverse- i.e. the output goes to SCC 0 and input goes to SCC (#nodes - 1)

* MakeSccGraph(graph, sccs, &scc_graph) - "scc_graph" encodes the *depends_on_this* information on SCCs, so for our example `scc_graph[sccs.size()-1]`(input SCC) and `scc_graph[sccs.size()-2]`(ivector SCC) both contain (sccs.size() - 3) wich is the ID for the SCC for the LDA component's input descriptor.

The next step(in effect) in the original code is ComputeTopSortOrder(), that takes this *depends_on_this* info(which is what it expects by design), appends SCC IDs to a vector using DFS traversal(so the output node, with SCC id 0 becomes the first element of the vector) and then reverses that order so that `scc_node_to_epoch[sccs.size()-1]` gets value of 0, meaning that that the input element is assigned the first possible epoch.

TL;DR: the original code is correct.

At least that's my understanding..

danpovey:
>
>
>
>
> The next step(in effect) in the original code is ComputeTopSortOrder(),
> that takes this *depends_on_this* info(which is what it expects by
> design), appends SCC IDs to a vector using DFS traversal(so the output
> node, with SCC id 0 becomes the first element of the vector) and then
> reverses that order so that scc_node_to_epoch[sccs.size()-1] gets value
> of 0, meaning that that the input element is assigned the first possible
> epoch.
>
I don't accept this reasoning because a function should always behave
according to its documentation and to the expectations arising from the
function name, not in a way that's dictated by the context in which it's
called.  ComputeTopSortOrder() violates this principle.  What I want to do
here is to remove the code in ComputeTopSortOrder that reverses the order,
and fix that line where you originally found a bug.  If you won't do this
then I will do it.

Dan

vdp:
The patch was updated yesterday. I think you are not reading my messages very carefully :)

danpovey:
Oh... I looked at the code and it's what I wanted, and looks right, so yes, maybe I misinterpreted something you said.  Merging.

vdp:
For the sake of clarifying this, I believe our misunderstanding lies in the interpretation of ComputeTopSortOrder's comment that I pasted in one of my previous messages:

    /// Given an acyclic graph (where each std::vector<int32> is a list of
    /// destination-nodes of arcs coming from the current node), compute a
    /// topological ordering of the graph nodes.

My understanding was that it expects "forward"(e.g. from the dependency to the dependent node), and the original test code seems to suggest this might have been your original intention as well. Now we've just changed it to work with "dependencies" instead. In fact that comment is ambiguous.

BTW, I think the word "phase" should be replaced by "epoch" in ComputeDependenciesSubset().

danpovey:
My intention also is for it to work with "forward" dependencies.  There is
no notion of backward dependencies in graph theory anyway.  That's just a
reversed graph.  So if you are correct, then it's not what I want.  But you
may not be correct.

The testing code is very unclear-- in order to understand it I'd have to
read 100 lines of code.   What it really needs is a test that's independent
and doesn't depend on the SCC stuff.  If you write an independent test it
will clarify what the code is really doing.  It may just be a question of
variables misnamed in the code that make it look like it gives a reversed
order.  It would be good if you can figure it out.

Regarding phase vs. epoch, I'll fix that myself.

Dan

danpovey:
BTW, it could be that the problem you discovered (with the _reversed
variable not being used) was the original problem, but when we noticed the
bug it caused, we didn't see that problem and instead assumed that the
top-sort was giving reversed output, so we wrote a wrapper for the original
top-sort function that reversed the output, and renamed the old TopSort
function by adding Reversed.
Of course the root of all this is inadequate testing code.

Dan

vdp:
Well, I am fairly certain that after the "fix", if we pass the graph represented by "forward" links we compute the reverse of the topological order, but for the sake of demonstration I prepared a [simple unit test](https://github.com/vdp/kaldi/commit/e6b3ff193e26dfd3ddc745ca0dcfd757800ed439#diff-cb0317b389ca5fafd520ee18917bc142R199) (in case the link doesn't work for some reason it's the last commit in the "fix-compute-epochs" branch in my clone of Kaldi).
Predictably the result of the execution is:

    ERROR (UnitTestComputeTopSortOrder2():nnet-graph-test.cc:229) Assertion failed: AssertVectorEqual(ref_node_to_order, computed_node_to_order) 
    && "Hmm, maybe the original code was correct after all :)"
    Stack trace is:
    kaldi::KaldiGetStackTrace()
    kaldi::KaldiAssertFailure_(char const*, char const*, int, char const*)
    kaldi::nnet3::UnitTestComputeTopSortOrder2()
    ./nnet-graph-test(main+0x25) [0x407016]
    /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf5) [0x7ff244ac1ec5]
    ./nnet-graph-test() [0x405939]

etc..

The actual node-to-order it computes for this test is:

    4 2 1 3 0

With the old version the test passes. Q.E.D.

danpovey:
OK, I see the issue now.  I had thought that NnetToDirectedGraph() was a
graph of dependencies, but it's a graph of reversed dependencies (and
actually its documentation states this).  This must have been a
miscommunication between me an Guoguo, but actually the way it's done is
fine as long as it's documentedd.
So actually there is no need for the 'ComputeGraphTranspose()', and what
you wanted to do to the code was correct.  It would be good if you could
prepare a patch to do it the way you had wanted, and also include your new
test code.

Dan

vdp:
Hallelujah :)

PR #754 - CI tests still running, but I've tested it locally and don't expect trouble.

PR #741, opened 2016-05-02
Update fisher swbd tdnn smbr

naxingyu:


naxingyu:
BTW, is the chain branch deprecated?

在 2016/5/3 2:29, Daniel Povey 写道:

danpovey:
I think it will be easier for the time being to make all PRs relative to
the master.  Right now I think the master is ahead of the chain branch.

Dan

PR #742, opened 2016-05-02
WIP: AN4 recipe

guoguo12:
This is a recipe for [AN4](http://www.speech.cs.cmu.edu/databases/an4/), a tiny (~1 hr, 130-word lexicon) but free corpus produced by CMU/Sphinx. The steps/config in the recipe are largely based on those for RM.

This is my first recipe, so any and all feedback is appreciated. Thanks!


jtrmal:
Thank you for this.
At this stage, I have only a few high-level comments.
a) add licensing terms and your name to any of the scripts you've written
b) extend the run.sh so it downloads the corpus in case it's not found (unless there is some reason that prevents it)
c) I don't think assuming sph2pipe is in the given directory is correct -- simply test if it's on the $PATH and  use `which` (or `command -v sph2pipe`  which is reportedly a preferred call)
y.

danpovey:
On Mon, May 2, 2016 at 12:09 PM, jtrmal <notifications@github.com> wrote:

> Thank you for this.
> At this stage, I have only a few high-level comments.
> a) add licensing terms and your name to any of the scripts you've written
> b) extend the run.sh so it downloads the corpus in case it's not found
> (unless there is some reason that prevents it)
> c) I don't think assuming sph2pipe is in the given directory is correct --
> simply test if it's on the $PATH and use which (or command -v sph2pipe
> which is reportedly a preferred call)
>

He may have copied the sph2pipe stuff from WSJ or similar.  I'm OK with the
way he has done it it actually-- there's no chance it will ever change as
it's not being developed further.  Plus it wouldn't be on the path unless
he explicitly put that directory on the path.  I checked my env.sh and it's
not there, and anyway he isn't sourcing env.sh.

Dan

egs/an4/s5/run.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+
+. cmd.sh
+. path.sh
+
+stage=1
+nj=1
+an4_root='/home/allen/data/an4'
```
danpovey: Since this data can easily be downloaded from the web, you should set up the script to download it automatically.  Look at Librispeech for an example - or "yesno", which is simpler and easier.  Does it have a stable URL somewhere?

egs/an4/s5/run.sh
```diff
@@ -0,0 +1,112 @@
...
+        exp/tri2/graph data/test exp/tri2/decode
+fi
+
+# align tri2
+if [ $stage -le 9 ]; then
+    steps/align_si.sh --nj $nj --cmd "$train_cmd" \
+        data/train data/lang exp/tri2 exp/tri2_ali
+fi
+
+# do MMI on top of LDA+MLLT and decode
```
danpovey: I'm not sure if it's really necessary to include MMI training, since the corpus is so tiny.
I would have thought that fMLLR would give more improvement, and is certainly easier to do.

egs/an4/s5/run.sh
guoguo12: Yep, it has a stable URL. I modified the download script for Librispeech.

egs/an4/s5/run.sh
guoguo12: Ah, interesting! I'll try fMLLR and get back to you.

guoguo12:
Thank you both for the feedback. I've added licensing terms and a script that downloads the corpus. I'll try fMLLR and report back.

guoguo12:
Sorry for the delay. (It's finals week for me!)

I replaced the MMI step with fMLLR, which brings the WER down to 5.95%.

Previous to c13c578, my `utt2spk` file actually listed each utterance as having its own unique speaker. But when I added in the correct speaker identities, the WER after the tri2 step (LDA+MLLT) went up. Does that seem reasonable? Why would that happen?

danpovey:
On Sun, May 8, 2016 at 5:59 AM, Allen Guo <notifications@github.com> wrote:

> Sorry for the delay. (It's finals week for me!)
>
> I replaced the MMI step with fMLLR, which brings the WER down to 5.95%.
>
> Previous to c13c578
> <https://github.com/kaldi-asr/kaldi/commit/c13c578cabbc55a27b0ba58f0edafdf9d0436a84>,
> my utt2spk file actually listed each utterance as having its own unique
> speaker. But when I added in the correct speaker identities, the WER after
> the tri2 step (LDA+MLLT) went up. Does that seem reasonable? Why would that
> happen?
>

It would affect the cepstral mean normalization (doing it per-speaker
instead of per utterance).
I think it's better to keep the speaker info anyway though.  It will
probably improve the performance after fMLLR.
Dan


>

guoguo12:
Got it. Anything else I should look into before this PR is merged?

danpovey:
no that I can think of, but when you think it's ready I'll have another
look.

guoguo12:
Okay, I think it's ready!

danpovey:
I'd prefer if you rebase to have just one commit- keep the history cleaner.

guoguo12:
Done!

danpovey:
OK, merging.  Thanks!

PR #744, opened 2016-05-02
Bug fixes to the Iban recipe

vijayaditya:
I will keep fixing the bugs in this recipe as I run different stages.

### DIFF
--- a/egs/iban/s5/local/prepare_lm.sh
+++ b/egs/iban/s5/local/prepare_lm.sh
@@ -10,7 +10,7 @@ set -e -o pipefail
 
 local/train_lms_srilm.sh --train-text data/train/text data/ data/srilm
 
-nl -nrz -w10  iban/LM/iban-bp-2012.txt | sort -R > data/local/external_text
+nl -nrz -w10  corpus/LM/iban-bp-2012.txt | sort -R > data/local/external_text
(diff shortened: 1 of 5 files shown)


egs/iban/s5/local/prepare_lm.sh
```diff
@@ -10,7 +10,7 @@ set -e -o pipefail
 
 local/train_lms_srilm.sh --train-text data/train/text data/ data/srilm
 
-nl -nrz -w10  iban/LM/iban-bp-2012.txt | sort -R > data/local/external_text
+nl -nrz -w10  corpus/LM/iban-bp-2012.txt | sort -R > data/local/external_text
```
jtrmal: This should probably be $corpus or something like that

vijayaditya: the database is being downloaded to corpus directory. Using a variable for
$corpus is not necessary.

jtrmal: Yes, that's true but I don't want to have the script to assume the location
-- for example at the CLSP, the location of the data would be somewhere in
/exports/corpora (as it is with tedlium or whatever).
But I see the whole script wasn't built with assumption of that, so I will
fix it later... Let's just leave it now as it is.
y.

jtrmal:
OK< merging

PR #752, opened 2016-05-03
Update to ComposeCompactLatticeDeterministic to fix a bug regarding t…

danpovey:
…he olabel (which shouldn't matter in our normal use-cases), and make final-probs application more efficient.


danpovey:
@david-ryan-snyder  and @hhadian, please take note of this upcoming change (should not affect anything).

hhadian:
I will test it with duration-modeling codes.

PR #756, opened 2016-05-04
Fix inversion of input and output of open2.

francisr:
Open2 doc: http://perldoc.perl.org/IPC/Open2.html
The first parameter is the output of the command, and the second one is its input. Previously it would never get into the while loop.
However now the error message about cyclicity triggers even on a G.fst that works fine for HCLG creation.

### DIFF
--- a/egs/wsj/s5/utils/lang/check_g_properties.pl
+++ b/egs/wsj/s5/utils/lang/check_g_properties.pl
@@ -56,14 +56,14 @@ while (<G>) {
       print "$0: validating $lang: error: line $_ in G.fst contains forbidden symbol <s> or </s>\n";
       exit(1);
     } elsif ($is_disambig{$A[2]}) {
-      print O $_;
+      print I $_;
       if ($A[3] != 0) {
         chop;
         print "$0: validating $lang: error: line $_ in G.fst has disambig on input but no epsilon on output\n";
         exit(1);
       }
     } elsif ($A[2] == 0) {
-      print O $_;
+      print I $_;
       $has_epsilons = 1;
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks..
So does this patch fix the issue that you found?  Your patch isn't causing any other problems, right?  I wasn't 100% sure about your commit message, what it meant.

francisr:
This fixes the fact we never got into the while loop. But now it seems to always detect that G.fst is cyclic with the LMs I used, so there must be something wrong somewhere.

egs/wsj/s5/utils/lang/check_g_properties.pl
```diff
@@ -73,8 +73,8 @@
   }
 }
 
-close(O);  # tell 'fstcompile | fstinfo' pipeline that its input is done.
-while (<I>) {
+close(I);  # tell 'fstcompile | fstinfo' pipeline that its input is done.
+while (<O>) {
   if (m/cyclic\s+/) {
```
danpovey: OK, I think at this point it should say m/cyclic\s+y/
See if that makes it behave as expected.

francisr: Right, that is obvious in hindsight...
I've updated the pull request.

jtrmal: Remi, does the check pass for you now?
y.

francisr: Yes it does, I've tried with two different LMs.

danpovey:
Thanks- merging.

PR #759, opened 2016-05-04
Small fixes

giuliopaci:
While reading scripts I found a few small errors.
The fixes are not related to each other, but they are trivial, so I hope a single pull request is ok.

### DIFF
--- a/tools/extras/install_irstlm.sh
+++ b/tools/extras/install_irstlm.sh
@@ -55,7 +55,7 @@ fi
   [ ! -z ${IRSTLM} ] && \
     echo >&2 "IRSTLM config is already in env.sh" && exit
 
-  wd=`readlink -f $wd 2>/dev/null || pwd`
+  wd=`pwd -P`
 
   echo "export IRSTLM=$wd/irstlm"
   echo "export PATH=\${PATH}:\${IRSTLM}/bin"
--- a/src/configure
+++ b/src/configure
@@ -953,7 +953,7 @@ if [ "`uname`" == "Linux" ]; then
       See makefiles/linux_64_mkl.mk to manually configure for other platforms."
     fi
 
-    if  is_set "$MKLROOT" -a ! is_set "$MKLLIBDIR"; then
+    if  ( is_set "$MKLROOT" && ! is_set "$MKLLIBDIR" ); then
       echo -n "Configuring MKL library directory: "
       MKLLIBDIR=`linux_configure_mkllibdir $MKLROOT`
       if [ $? -ne 0 ]; then


tools/extras/install_irstlm.sh
```diff
@@ -55,6 +55,7 @@ fi
   [ ! -z ${IRSTLM} ] && \
     echo >&2 "IRSTLM config is already in env.sh" && exit
 
+  wd=`pwd`
```
danpovey: why is this line added, isn't it overridden by the next line?
the '|| pwd' is about BSD compatibility.

tools/extras/install_srilm.sh
```diff
@@ -24,7 +24,7 @@ tar -xvzf ../srilm.tgz
 # set the SRILM variable in the top-level Makefile to this directory.
 cp Makefile tmpf
 
-cat tmpf | awk -v pwd=`pwd` '/SRILM =/{printf("SRILM = %s\n", pwd); next;} {print;}' \
+cat tmpf | gawk -v pwd=`pwd` '/SRILM =/{printf("SRILM = %s\n", pwd); next;} {print;}' \
```
danpovey: Kaldi requires awk to be installed anyway, so using awk is fine and actually recommended/preferred for Kaldi scripts...
we check for gawk because SRILM does not work without it.

tools/extras/install_irstlm.sh
giuliopaci: This line is necessary because otherwise ```readlink -f $wd``` makes no sense and always fail.

tools/extras/install_irstlm.sh
```diff
@@ -55,6 +55,7 @@ fi
...
   wd=`readlink -f $wd 2>/dev/null || pwd`
```
danpovey: oh I see.  you could probably replace $wd in this expression with $PWD, and remove the extra line.

tools/extras/install_srilm.sh
giuliopaci: Then I am going to fix another line in this script. Wait for the PR to be updated before merging.

giuliopaci: Sorry, I missed your last line. Then I am going to drop this commit.

tools/extras/install_irstlm.sh
giuliopaci: Reading http://www.unix.com/man-page/posix/1posix/pwd/ probably the desired behaviour is just one line:
```
wd=`pwd -P`
```

or, for maximum compatibility with already used systems:

```
wd=`pwd -P 2>/dev/null || pwd`
```

This will avoid the unportable ```readlink -f```, provide a portable solution and will be equivalent to or better than (from the performance point of view, that is completely irrelevant in this case) ```readlink -f $PWD``` in bash where ```pwd``` is a builtin.

danpovey: I'm OK with wd=`pwd -P`; it works on my mac.

giuliopaci:
I just updated the pull request.

danpovey:
Are you sure you pushed to your repo?  I dont see the w=`pwd -P`

giuliopaci:
Indeed I pushed before noticing that I was still in the middle of a rebase, due to an error. Now it should be ok.

danpovey:
Thanks-- merging.

PR #769, opened 2016-05-09
Cudnn 3D Convolution Component

galv:
Note: Unfortunately the tests won't run right now. CUDNN works only on the GPU, and I haven't thought of a good way to change the current test code to handle GPU-only components. But trust me when I say that all tests do pass (I made unpushed changes to the test code to check).

Here are a few talking points:

- Currently, CUDNN's convolution is much more full-featured than the one provided by Kaldi. The lack of a CPU implementation of things like upscaling will be a problem in the future, since I believe we like to do inference on the CPU.
- No attempt was made to provide the same interface as the current ConvolutionComponent. This was done partially for ease of implementation.
- @danpovey  should look at commit aa34c2e . It may not be the best way to fix the bug that it fixes.
- I don't currently allow CUDNN to select the best possible convolution algorithm itself.
- Only zyx vectorization is supported right now. CUDNN has primitives to convert yzx to zyx if necessary. I admit I'm not sure why Kaldi needs yzx configuration anyway.

Let me know if anything is unclear. I probably know CUDNN's interface better than anyone else here now.


galv:
Continuation of #604

danpovey:
I just created a branch 'cudnn' as a copy of master.  Please make this PR to that branch.
This doesn't replace detailed review-- which I don't have time for right now.

galv:
Trust me, I don't expect this to get into the main code base as it is ;D

PR #772, opened 2016-05-09
Adding nnet3/chain support to steps/cleanup/find_bad_utts_nnet.sh, no…

realill:
Adding nnet3/chain support for this script, so it now supports both nnet2 and nnet3. Probably nnet1.


danpovey:
OK.  Normally we'd have different scripts for the different types of nnet, but this seems fairly harmless, so I'm merging it.

PR #775, opened 2016-05-12
nnet1: updating the MMI / sMBR scripts,

vesis84:
- fixing bug when i-vectors are used with 'steps/nnet/train_mpe.sh',
- cosmetic changes,


vesis84:
(just script change, no need to wait for travis)

PR #779, opened 2016-05-15
Remove old memory alignment flags

### LINKED ISSUE #740: Get rid of HAVE_POSIX_MEMALIGN and HAVE_MEMALIGN
I want the code in kaldi-utils.h regarding memory alignment to be cleaned up.
The setups we support can be divided into two, in terms of memory alignment
(1) windows -> use aligned_malloc
(2) unix systems [Linux and Darwin] -> use posix_memalign

So we don't need the compilation flags HAVE_POSIX_MEMALIGN or HAVE_MEMALIGN any more (they were already legacy code, from proto-Kaldi).  These flags should be removed from the prototype Makefiles.

Anyone can help with this.

guoguo12:
As far as I can tell, this resolves #740. Tested on Ubuntu. Should be tested on Windows.

### DIFF
--- a/src/base/kaldi-utils.h
+++ b/src/base/kaldi-utils.h
@@ -46,25 +46,14 @@
 #endif
 #endif
 
-#ifdef HAVE_POSIX_MEMALIGN
-#  define KALDI_MEMALIGN(align, size, pp_orig) \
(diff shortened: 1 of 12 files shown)


danpovey:
Thanks!  Merging.

PR #644, opened 2016-04-02
Tedlium chain and nnet3 scripts

galv:

PR #783, opened 2016-05-16
Tedlium chain pr 3

galv:
Rebased version of #644

### DIFF
--- a/egs/tedlium/s5/local/chain/README
+++ b/egs/tedlium/s5/local/chain/README
@@ -1,0 +1,16 @@
+These are the instructions to reproduce the TEDLIUM models described in
+"Purely sequence-trained neural networks for ASR based on lattice-free
+MMI", by Povey et al.
+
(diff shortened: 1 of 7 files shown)


vijayaditya:
As these scripts are from #644, which was well reviewed, I just performed a cursory check. Changes are OK. Assuming the RESULTS have been updated I am performing the merge.

galv:
Alas, I never actually put these into the RESULTS file.

By the way, I don't think you should have squashed these two commits together. The first commit fixes a bug affecting all TEDLIUM recipes.

vijayaditya:
The commit message reports both the changes I think it is OK to squash.

Don't worry about the RESULTS file you can fix that in your next commit, it
is fine as you have the results in the individual scripts.

--Vijay

galv:
But now someone will have to dig through through a rather big commit to find the first change. Anyway, it's done.

PR #794, opened 2016-05-20
Add missing dependencies to Makefiles

akreal:


danpovey:
Hm yes, that could explain some build failures we've seen.
I'll merge when the checks complete.

PR #800, opened 2016-05-21
yesno recipe data prep: paths to absolute + misc. 

jfainberg:
Proposed fix for issue #457 . Tried to make the recipe more similar to other typical Kaldi recipes (e.g. WSJ):

- Changed relative paths to absolute
- Added usage message if missing argument
- Exit clauses
- "Data preparation succeeded" message
- Removed superfluous scripts variable

Tested to create identical data directories as the original recipe (diff -ra ...), and a full run yields 0.00% WER.

Let me know if this is in line with what you thought for the issue. Happy to revise.

### DIFF
--- a/egs/yesno/s5/local/prepare_data.sh
+++ b/egs/yesno/s5/local/prepare_data.sh
@@ -12,9 +12,10 @@ train_base_name=train_yesno
 test_base_name=test_yesno
 waves_dir=$1
 
+ls -1 $waves_dir > data/local/waves_all.list
+
 cd data/local
 
-ls -1 ../../$waves_dir > waves_all.list
 ../../local/create_yesno_waves_test_train.pl waves_all.list waves.test waves.train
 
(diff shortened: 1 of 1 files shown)


danpovey:
I'd prefer it if you just fixed the issue in a more minimal way.  The issue is that the script doesn't work if waves_dir is an absolute pathname, because it's accessed as ../../$waves_dir.   There are too many bash variables in your fix.

jfainberg:
Thanks! That makes sense. Will modify this weekend.

jfainberg:
Rebased and squashed the two commits.

danpovey:
This is still not really ideal because they're still relative pathnames, as far as I can see.  The aim is to get absolute pathnames in the wav.scp.

danpovey:
Actually I'll merge anyway-- it should fix the problem, even if not 100% ideal.

PR #803, opened 2016-05-22
WIP: Shrink in-value of ClipGradientComponent toward some smaller value when clipping proportion exceeds some threshold

freewym:
It looks taking effective on BLSTM: the likelihood with self-repair get matched with one without self-repair. Will play around with some alternatives, e.g., on what basis to average magnitude of the gradients. etc.

Yiming

### DIFF
--- a/egs/swbd/s5c/local/nnet3/run_lstm.sh
+++ b/egs/swbd/s5c/local/nnet3/run_lstm.sh
@@ -99,7 +99,8 @@ if [ $stage -le 9 ]; then
     --recurrent-projection-dim $recurrent_projection_dim \
     --non-recurrent-projection-dim $non_recurrent_projection_dim \
     --label-delay $label_delay \
-    --self-repair-scale 0.00001 \
+    --self-repair-scale-nonlinearity 0.00001 \
+    --self-repair-scale-clipgradient 1.0 \
    $dir/configs || exit 1;
 
 fi
--- a/egs/wsj/s5/steps/nnet3/components.py
+++ b/egs/wsj/s5/steps/nnet3/components.py
@@ -104,6 +104,7 @@ def AddAffRelNormLayer(config_lines, name, input, output_dim, ng_affine_options 
     components = config_lines['components']
     component_nodes = config_lines['component-nodes']
 
+    # self_repair_scale is a constant scaling the self-repair vector computed in RectifiedLinearComponent
     self_repair_string = "self-repair-scale={0:.10f}".format(self_repair_scale) if self_repair_scale is not None else ''
     components.append("component name={0}_affine type=NaturalGradientAffineComponent input-dim={1} output-dim={2} {3}".format(name, input['dimension'], output_dim, ng_affine_options))
     components.append("component name={0}_relu type=RectifiedLinearComponent dim={1} {2}".format(name, output_dim, self_repair_string))
@@ -222,6 +223,7 @@ def AddSigmoidLayer(config_lines, name, input, self_repair_scale = None):
     components = config_lines['components']
     component_nodes = config_lines['component-nodes']
 
+    # self_repair_scale is a constant scaling the self-repair vector computed in SigmoidComponent
     self_repair_string = "self-repair-scale={0:.10f}".format(self_repair_scale) if self_repair_scale is not None else ''
     components.append("component name={0}_sigmoid type=SigmoidComponent dim={1}".format(name, input['dimension'], self_repair_string))
     component_nodes.append("component-node name={0}_sigmoid component={0}_sigmoid input={1}".format(name, input['descriptor']))
@@ -285,7 +287,8 @@ def AddLstmLayer(config_lines,
                  ng_per_element_scale_options = "",
                  ng_affine_options = "",
                  lstm_delay = -1,
-                 self_repair_scale = None):
+                 self_repair_scale_nonlinearity = None,
+                 self_repair_scale_clipgradient = None):
     assert(recurrent_projection_dim >= 0 and non_recurrent_projection_dim >= 0)
     components = config_lines['components']
     component_nodes = config_lines['component-nodes']
@@ -306,7 +309,11 @@ def AddLstmLayer(config_lines,
     else:
         add_non_recurrent_projection = True
 
-    self_repair_string = "self-repair-scale={0:.10f}".format(self_repair_scale) if self_repair_scale is not None else ''
+    # self_repair_scale_nonlinearity is a constant scaling the self-repair vector computed in derived classes of NonlinearComponent,
+    # i.e.,  SigmoidComponent, TanhComponent and RectifiedLinearComponent
+    self_repair_nonlinearity_string = "self-repair-scale={0:.10f}".format(self_repair_scale_nonlinearity) if self_repair_scale_nonlinearity is not None else ''
+    # self_repair_scale_clipgradient is a constant scaling the self-repair vector computed in ClipGradientComponent
+    self_repair_clipgradient_string = "self-repair-scale={0:.2f}".format(self_repair_scale_clipgradient) if self_repair_scale_clipgradient is not None else ''
     # Natural gradient per element scale parameters
     ng_per_element_scale_options += " param-mean=0.0 param-stddev=1.0 "
     # Parameter Definitions W*(* replaced by - to have valid names)
@@ -330,17 +337,17 @@ def AddLstmLayer(config_lines,
 
 
     components.append("# Defining the non-linearities")
-    components.append("component name={0}_i type=SigmoidComponent dim={1} {2}".format(name, cell_dim, self_repair_string))
-    components.append("component name={0}_f type=SigmoidComponent dim={1} {2}".format(name, cell_dim, self_repair_string))
-    components.append("component name={0}_o type=SigmoidComponent dim={1} {2}".format(name, cell_dim, self_repair_string))
-    components.append("component name={0}_g type=TanhComponent dim={1} {2}".format(name, cell_dim, self_repair_string))
-    components.append("component name={0}_h type=TanhComponent dim={1} {2}".format(name, cell_dim, self_repair_string))
+    components.append("component name={0}_i type=SigmoidComponent dim={1} {2}".format(name, cell_dim, self_repair_nonlinearity_string))
+    components.append("component name={0}_f type=SigmoidComponent dim={1} {2}".format(name, cell_dim, self_repair_nonlinearity_string))
+    components.append("component name={0}_o type=SigmoidComponent dim={1} {2}".format(name, cell_dim, self_repair_nonlinearity_string))
+    components.append("component name={0}_g type=TanhComponent dim={1} {2}".format(name, cell_dim, self_repair_nonlinearity_string))
+    components.append("component name={0}_h type=TanhComponent dim={1} {2}".format(name, cell_dim, self_repair_nonlinearity_string))
 
     components.append("# Defining the cell computations")
     components.append("component name={0}_c1 type=ElementwiseProductComponent input-dim={1} output-dim={2}".format(name, 2 * cell_dim, cell_dim))
     components.append("component name={0}_c2 type=ElementwiseProductComponent input-dim={1} output-dim={2}".format(name, 2 * cell_dim, cell_dim))
     components.append("component name={0}_m type=ElementwiseProductComponent input-dim={1} output-dim={2}".format(name, 2 * cell_dim, cell_dim))
-    components.append("component name={0}_c type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} ".format(name, cell_dim, clipping_threshold, norm_based_clipping))
+    components.append("component name={0}_c type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} {4}".format(name, cell_dim, clipping_threshold, norm_based_clipping, self_repair_clipgradient_string))
 
     # c1_t and c2_t defined below
     component_nodes.append("component-node name={0}_c_t component={0}_c input=Sum({0}_c1_t, {0}_c2_t)".format(name))
@@ -379,7 +386,7 @@ def AddLstmLayer(config_lines,
     if (add_recurrent_projection and add_non_recurrent_projection):
         components.append("# projection matrices : Wrm and Wpm")
         components.append("component name={0}_W-m type=NaturalGradientAffineComponent input-dim={1} output-dim={2} {3}".format(name, cell_dim, recurrent_projection_dim + non_recurrent_projection_dim, ng_affine_options))
-        components.append("component name={0}_r type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} ".format(name, recurrent_projection_dim, clipping_threshold, norm_based_clipping))
+        components.append("component name={0}_r type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} {4}".format(name, recurrent_projection_dim, clipping_threshold, norm_based_clipping, self_repair_clipgradient_string))
         component_nodes.append("# r_t and p_t")
         component_nodes.append("component-node name={0}_rp_t component={0}_W-m input={0}_m_t".format(name))
         component_nodes.append("dim-range-node name={0}_r_t_preclip input-node={0}_rp_t dim-offset=0 dim={1}".format(name, recurrent_projection_dim))
@@ -390,7 +397,7 @@ def AddLstmLayer(config_lines,
     elif add_recurrent_projection:
         components.append("# projection matrices : Wrm")
         components.append("component name={0}_Wrm type=NaturalGradientAffineComponent input-dim={1} output-dim={2} {3}".format(name, cell_dim, recurrent_projection_dim, ng_affine_options))
-        components.append("component name={0}_r type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} ".format(name, recurrent_projection_dim, clipping_threshold, norm_based_clipping))
+        components.append("component name={0}_r type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} {4}".format(name, recurrent_projection_dim, clipping_threshold, norm_based_clipping, self_repair_clipgradient_string))
         component_nodes.append("# r_t")
         component_nodes.append("component-node name={0}_r_t_preclip component={0}_Wrm input={0}_m_t".format(name))
         component_nodes.append("component-node name={0}_r_t component={0}_r input={0}_r_t_preclip".format(name))
@@ -398,7 +405,7 @@ def AddLstmLayer(config_lines,
         output_dim = recurrent_projection_dim
 
     else:
-        components.append("component name={0}_r type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} ".format(name, cell_dim, clipping_threshold, norm_based_clipping))
+        components.append("component name={0}_r type=ClipGradientComponent dim={1} clipping-threshold={2} norm-based-clipping={3} {4}".format(name, cell_dim, clipping_threshold, norm_based_clipping, self_repair_clipgradient_string))
         component_nodes.append("component-node name={0}_r_t component={0}_r input={0}_m_t".format(name))
         output_descriptor = '{0}_r_t'.format(name)
         output_dim = cell_dim
--- a/egs/wsj/s5/steps/nnet3/components.py
+++ b/egs/wsj/s5/steps/nnet3/components.py
@@ -424,18 +424,23 @@ def AddBLstmLayer(config_lines,
                   ng_per_element_scale_options = "",
                   ng_affine_options = "",
                   lstm_delay = [-1,1],
-                  self_repair_scale = None):
+                  self_repair_scale_nonlinearity = None,
+                  self_repair_scale_clipgradient = None):
     assert(len(lstm_delay) == 2 and lstm_delay[0] < 0 and lstm_delay[1] > 0)
     output_forward = AddLstmLayer(config_lines, "{0}_forward".format(name), input, cell_dim,
                                   recurrent_projection_dim, non_recurrent_projection_dim,
                                   clipping_threshold, norm_based_clipping,
                                   ng_per_element_scale_options, ng_affine_options,
-                                  lstm_delay = lstm_delay[0], self_repair_scale = self_repair_scale)
+                                  lstm_delay = lstm_delay[0],
+                                  self_repair_scale_nonlinearity = self_repair_scale_nonlinearity,
+                                  self_repair_scale_clipgradient = self_repair_scale_clipgradient)
     output_backward = AddLstmLayer(config_lines, "{0}_backward".format(name), input, cell_dim,
                                    recurrent_projection_dim, non_recurrent_projection_dim,
                                    clipping_threshold, norm_based_clipping,
                                    ng_per_element_scale_options, ng_affine_options,
-                                   lstm_delay = lstm_delay[1], self_repair_scale = self_repair_scale)
+                                   lstm_delay = lstm_delay[1],
+                                   self_repair_scale_nonlinearity = self_repair_scale_nonlinearity,
+                                   self_repair_scale_clipgradient = self_repair_scale_clipgradient)
     output_descriptor = 'Append({0}, {1})'.format(output_forward['descriptor'], output_backward['descriptor'])
     output_dim = output_forward['dimension'] + output_backward['dimension']
 
--- a/egs/wsj/s5/steps/nnet3/lstm/make_configs.py
+++ b/egs/wsj/s5/steps/nnet3/lstm/make_configs.py
@@ -74,8 +74,10 @@ def GetArgs():
                         help="use norm based clipping in ClipGradient components ", default=True, choices = ["false", "true"])
     parser.add_argument("--clipping-threshold", type=float,
                         help="clipping threshold used in ClipGradient components, if clipping-threshold=0 no clipping is done", default=30)
-    parser.add_argument("--self-repair-scale", type=float,
-                        help="A non-zero value activates the self-repair mechanism in the sigmoid and tanh non-linearities of the LSTM", default=None)
+    parser.add_argument("--self-repair-scale-nonlinearity", type=float,
+                        help="A non-zero value activates the self-repair mechanism in the sigmoid and tanh non-linearities of the LSTM", default=0.00001)
+    parser.add_argument("--self-repair-scale-clipgradient", type=float,
+                        help="A non-zero value activates the self-repair mechanism in the ClipGradient component of the LSTM", default=1.0)
 
     # Delay options
     parser.add_argument("--label-delay", type=int, default=None,
@@ -212,7 +214,8 @@ def MakeConfigs(config_dir, feat_dim, ivector_dim, num_targets,
                 num_lstm_layers, num_hidden_layers,
                 norm_based_clipping, clipping_threshold,
                 ng_per_element_scale_options, ng_affine_options,
-                label_delay, include_log_softmax, xent_regularize, self_repair_scale):
+                label_delay, include_log_softmax, xent_regularize,
+                self_repair_scale_nonlinearity, self_repair_scale_clipgradient):
 
     config_lines = {'components':[], 'component-nodes':[]}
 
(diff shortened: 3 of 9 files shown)


danpovey:
@vpeddinti, do you have time to check this?

The basic idea was, in the ClipGradient component, to detect when we're
clipping too much (e.g. >1% of the time, or whatever), and then activate
some self-repair code.  The idea of the self-repair is to add something to
the backprop-d gradients that forces the input to the ClipGradient
component to be close to some default value (which might be 0.0 or 0.5, or
some other value, depending on what the input to the ClipGradient component
is, e.g. sigmoid or tanh or affine or whatever).  The hope is that if we
force the inputs to be small, it will tend to make the parameters on that
path small also [at least, some of them], and this may help to tamp down
the divergence.  This is maybe not an ideal solution but this problem seems
to happen rarely, and I'm hoping that this code will be activated only in
unusual circumstances, and as long as we can re-stabilize things, we can go
back to the normal regime.

We need the "extra part" of the gradient to have a suitable magnitude.  The
problem is- if the gradients weren't exploding already, we wouldn't be in
this mess.  So a small fixed scale, like 10^-5 as we use in the normal
self-repair, won't be sufficient.  So the code needs to make the magnitude
of the "extra part" of the gradients be proportional to the magnitude of
the existing gradients.

Dan

danpovey:
Yiming, obviously I assume you'll be checking this with something that was
diverging, like a "bad" configuration of your GRUs.
Dan

freewym:
Yes I will check on GRU. The BLSTM also got diverged with self-repair; with this code the BLSTM has returned to normal.

danpovey:
OK, please email me and Vijay the log locations.

src/cudamatrix/cu-vector.cc
```diff
@@ -328,7 +328,7 @@ MatrixIndexT CuVectorBase<Real>::ApplyFloor(Real floor_val) {
     cuda_vec_apply_floor(dimGrid, dimBlock, data_, floor_val, count_vec.Data(), dim_);
     CU_SAFE_CALL(cudaGetLastError());
     num_floored = count_vec.Sum();
-    CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyFloor", tim.Elapsed());
+    CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
```
danpovey: Yiming, the reason we sometimes write e.g. CuVectorBase::ApplyFloor instead of __func__, is that when the same function exists in multiple classes (e.g. also in  CuMatrixBase), we otherwise wouldn't know which one was taking time.

vijayaditya:
OK will do.

Vijay

src/nnet3/nnet-simple-component.cc
```diff
@@ -588,34 +608,64 @@ std::string ClipGradientComponent::Info() const {
...
+  num_backproped_ = num_backproped;
 }
 
 void ClipGradientComponent::InitFromConfig(ConfigLine *cfl) {
   int32 dim = 0;
   bool ok = cfl->GetValue("dim", &dim);
   bool norm_based_clipping = false;
   BaseFloat clipping_threshold = 15.0;
+  BaseFloat self_repair_clipped_proportion_threshold = kUnsetValue;
+  BaseFloat self_repair_clipping_to_value = kUnsetValue;
```
danpovey: I think self_repair_clipping_to_value is badly named.  How about self_repair_target?

src/nnet3/nnet-simple-component.cc
```diff
@@ -588,34 +608,64 @@ std::string ClipGradientComponent::Info() const {
          << ", clipping-threshold=" << clipping_threshold_
          << ", clipped-proportion="
          << (count_ > 0 ? static_cast<BaseFloat>(num_clipped_)/count_ : 0);
+  if (self_repair_clipped_proportion_threshold_ != BaseFloat(kUnsetValue))
```
danpovey: There is no need to invoke kUnsetValue anywhere in this code.  Just have those things default to whatever default value you currently use in the code.  The reason for kUnsetValue in NonlinearComponent was that the same base-class was responsible for setting the parameters, but  there were different defaults in different child classes.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
       in_deriv->ApplyCeiling(clipping_threshold_);
       in_deriv->ApplyFloor(-1 * clipping_threshold_);
     }
+
+    to_update->num_backproped_ += 1;
```
danpovey: should be spelled num_backpropped_ (two p's)

src/nnet3/nnet-simple-component.h
```diff
@@ -992,18 +1025,33 @@ class ClipGradientComponent: public Component {
...
                               // done
 
+  // some configuration values relating to self-repairing.
+  BaseFloat self_repair_clipped_proportion_threshold_;
+  BaseFloat self_repair_clipping_to_value_;
+  BaseFloat self_repair_scale_;
+  std::string debug_info_; //debug
+  
+  // this function is called from Backprop code and only does something if the
+  // self-repair-scale config value is set.
```
danpovey: this is not enough documentation

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+    const std::string &debug_info,
+    const CuMatrixBase<BaseFloat> &in_value,
+    CuMatrixBase<BaseFloat> *in_deriv, ClipGradientComponent *to_update) const {
+  KALDI_ASSERT(to_update != NULL);
+  BaseFloat default_clipped_proportion_threshold = 0.01;
+  BaseFloat default_clipping_to_value = 0.0;
+
+  // we use this 'repair_probability' (hardcoded for now) to limit
+  // this code to running on about half of the minibatches.
+  BaseFloat repair_probability = 0.5;
```
danpovey: I think in this case it may not be a good idea to use the repair_probability, since the derivatives are in different directions in this case.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+  // sign_mat = sign(in_value), i.e.,
+  // An element in sign_mat is 1 if its corresponding element in in_value > 0,
+  // or -1 otherwise
+  CuMatrix<BaseFloat> sign_mat(in_value);
+  sign_mat.ApplyHeaviside();
+  sign_mat.Scale(2.0);
+  sign_mat.Add(-1.0);
+
+  // repair_mat =
+  // floor(abs(in_value) - clipping_to_value, 0) .* sign(in_value)
```
danpovey: This formula is not right.
It should be just: (clipping_to_value - in_value).
That's the derivative (actually, half the derivative) of the  (in_value - clipping_to_value)^2 [interpreted as a sum over that formula elementwise].  Your formula coincides for clipping_to_value == 0.0, except your formula seems to have the wrong sign.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+  CuMatrix<BaseFloat> repair_mat(in_value);
+  repair_mat.ApplyPowAbs(1.0);
+  repair_mat.Add(-clipping_to_value);
+  repair_mat.ApplyFloor(0.0);
+  repair_mat.MulElements(sign_mat);
+
+  // magnitude =
```
danpovey: I think it would be better to set the magnitude at the whole minibatch level (or to be more precise, at the whole matrix level) because that will tend to give higher magnitudes of self-repair to input vectors that have larger absolute value, which will tend to be the ones that are diverging.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+  CuVector<BaseFloat> scales(repair_mat.NumRows());
+  scales.AddDiagMat2(pow(magnitude, -2), repair_mat, kNoTrans, 0.0);
+  // here we call ApplyPowAbs() instead of ApplyPow() because
+  // we want the inverse of 0 to be 0, rather than NaN(or whatever else).
+  scales.ApplyPowAbs(-0.5);
+  // now each element in scales = magnitude / (norm of each row of repair_mat)
+
+  // each row of the matrix being added to in_deriv is scaled to have the norm
+  // (magnitude / repair_probability).
+  in_deriv->AddDiagVecMat(-1.0 / repair_probability,
```
danpovey: Oh, I see with this -1.0 you are fixing the bug I noticed above with the repair_mat having the wrong sign.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+  repair_mat.ApplyFloor(0.0);
+  repair_mat.MulElements(sign_mat);
+
+  // magnitude =
+  // self_repair_scale_ * clipped_proportion * average norm of in-deriv
+  CuVector<BaseFloat> in_deriv_norm_vec(in_deriv->NumRows());
+  in_deriv_norm_vec.AddDiagMat2(1.0, *in_deriv, kNoTrans, 0.0);
+  in_deriv_norm_vec.ApplyPow(0.5);
+  double in_deriv_norm_sum = in_deriv_norm_vec.Sum();
+  BaseFloat magnitude = self_repair_scale_ * clipped_proportion *
```
danpovey: I see that you're including the clipped_proportion as a factor in the magnitude.. which is fine, and it explains why the optimal self_repair_scale is more than one.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+ 
+  CuVector<BaseFloat> scales(repair_mat.NumRows());
+  scales.AddDiagMat2(pow(magnitude, -2), repair_mat, kNoTrans, 0.0);
+  // here we call ApplyPowAbs() instead of ApplyPow() because
+  // we want the inverse of 0 to be 0, rather than NaN(or whatever else).
+  scales.ApplyPowAbs(-0.5);
+  // now each element in scales = magnitude / (norm of each row of repair_mat)
+
+  // each row of the matrix being added to in_deriv is scaled to have the norm
+  // (magnitude / repair_probability).
```
danpovey: Please verify with a print statement that each row of the matrix really does have the norm (magnitude / repair_probability)... and then remove the print statement.  There could be a bug somewhere.

danpovey: ... of course if you change it so the magnitude is set at a whole-matrix level, you'd be checking something different.

src/nnet3/nnet-simple-component.cc
```diff
@@ -668,12 +718,99 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
+  repair_mat.MulElements(sign_mat);
+
+  // magnitude =
+  // self_repair_scale_ * clipped_proportion * average norm of in-deriv
+  CuVector<BaseFloat> in_deriv_norm_vec(in_deriv->NumRows());
+  in_deriv_norm_vec.AddDiagMat2(1.0, *in_deriv, kNoTrans, 0.0);
+  in_deriv_norm_vec.ApplyPow(0.5);
+  double in_deriv_norm_sum = in_deriv_norm_vec.Sum();
+  BaseFloat magnitude = self_repair_scale_ * clipped_proportion *
+                        (in_deriv_norm_sum / in_deriv_norm_vec.Dim());
```
freewym: Here in_deriv_norm_sum is the sum of the norm of each row of in_deriv, so the magnitude is already set at the whole matrix level (average of the norm of each row of the matrix in_deriv).

danpovey: Yes but you are then applying that magnitude to each individual row of the
matrix.  What I am proposing instead is that you just scale the matrix so
that on average the rows have a certain magnitude.  It would be easier to
not think about magnitudes per row, but magnitudes at a whole-matrix level.

freewym: I see.

src/nnet3/nnet-simple-component.cc
freewym: @danpovey I checked in a new commit with all changes you suggested except this one. I don't quite understand the use of repair_probability. According to your above comment, it seems relavant to the direction of derivatives. But in my understanding, the nonlinearity component's self-repair and self-repair here both try to add a term to in-deriv to force the input value towards 0, so an input with value < 0 and an input with value > 0 move in opposite directions in both cases.

danpovey: >
> @danpovey <https://github.com/danpovey> I checked in a new commit with
> all changes you suggested except this one. I don't quite understand the use
> of repair_probability. According to your above comment, it seems relavant
> to the direction of derivatives. But in my understanding, the nonlinearity
> component's self-repair and self-repair here both try to add a term to
> in-deriv to force the input value towards 0, so an input with value < 0 and
> an input with value > 0 move in opposite directions in both cases.
>

I guess you're right.  Maybe it's really no different.

However, after thinking about all this, I'm getting increasingly unhappy
with a certain aspect of it.
Doing the self-repair will increase the magnitudes of the gradients; and
we're measuring the magnitudes of the gradients to decide whether to do
self-repair.
My concern is that if, for a certain job, we start doing self-repair, it
can become self-perpetuating for the lifetime of that job, because the
self-repair would increase the magnitudes of the gradients and lead to more
clipping, hence more self-repair.
I'm not sure how to fix this.  But maybe it's not really a problem.  See
how consistent the amount of self-repair is between jobs.  If some do much
more than others, maybe it's a problem.

freewym: In my BLSTM experiment, the self-repairs were called consistently at least between jobs within the same training iteration:
train.3.1 1391 times out of 12180
train.3.2 1200 times out of 12180
train.3.3 1372 times out of 12180
train.5.1 1639 times out of 12400
train.5.2 1224 times out of 12180
train.5.3 1344 times out of 12180
train.6.3   600 times out of 12240
train.8.2   432 times out of 12180

Perhaps there are two opposite forces when self-repair: one reducing the magnitude of gradient caused by reducing the magnitudes of parameter values, and the other one increasing the magnitude of gradient caused by reducing the inputs to nonlinearity on the path. It seems the former one dominates in this specific experiment.

danpovey: On Thu, Jun 2, 2016 at 3:08 PM, Yiming Wang <notifications@github.com>
wrote:

> In src/nnet3/nnet-simple-component.cc
> <https://github.com/kaldi-asr/kaldi/pull/803#discussion_r65599941>:
>
> >    }
> >  }
> >
> > +void ClipGradientComponent::RepairGradients(
> > +    const std::string &debug_info,
> > +    const CuMatrixBase<BaseFloat> &in_value,
> > +    CuMatrixBase<BaseFloat> *in_deriv, ClipGradientComponent *to_update) const {
> > +  KALDI_ASSERT(to_update != NULL);
> > +  BaseFloat default_clipped_proportion_threshold = 0.01;
> > +  BaseFloat default_clipping_to_value = 0.0;
> > +
> > +  // we use this 'repair_probability' (hardcoded for now) to limit
> > +  // this code to running on about half of the minibatches.
> > +  BaseFloat repair_probability = 0.5;
>
> In my BLSTM experiment, the self-repairs were called consistently at least
> between jobs within the same training iteration:
> train.3.1 1391 times out of 12180
> train.3.2 1200 times out of 12180
> train.3.3 1372 times out of 12180
> train.5.1 1639 times out of 12400
> train.5.2 1224 times out of 12180
> train.5.3 1344 times out of 12180
> train.6.3 600 times out of 12240
> train.8.2 432 times out of 12180
>

On your iterations 6 and 8 it was called a lot of times on just one of the
iterations, and never on the others.


> Perhaps there are two opposite forces when self-repair: one reducing the
> magnitude of gradient caused by reducing the magnitudes of parameter
> values, and the other one increasing the magnitude of gradient caused by
> reducing the inputs to nonlinearity on the path. It seems the former one
> dominates in this specific experiment.
>
You're forgetting that the extra part of the gradient, that we add, can
itself get backpropped through many frames.  So that would tend to increase
the magnitudes of the gradients directly.  But if it seems to be working
maybe we shouldn't worry too much.

What I was thinking of as a potential fix, was to only apply the
self-repair, say, every other minibatch, and only use the minibatches where
we're *not* applying self-repair to measure the gradients.
But this would be quite tricky to code, as it has to be done on the whole
minibatch to be effective (not on each component's backprop call, as there
can be many of those per minibatch).
Another idea is, when doing the self-repair, after we add the new term, to
scale down the entire matrix to have the same norm as before, so that the
magnitude doesn't change.

BTW, I have a theory about why it only happens in the backward version of
the BLSTM.  The iVector is estimated in a 'forward' causal way, so it
always sees the past.  But generally the iVectors in the future are more
accurate, as they've seen more info.  So maybe it's trying to propagate the
iVectors from the future to the past, resulting in 'forget' gates not being
active, and that sets the stage for instability.

egs/wsj/s5/steps/nnet3/components.py
```diff
@@ -270,7 +270,8 @@ def AddLstmLayer(config_lines,
                  ng_per_element_scale_options = "",
                  ng_affine_options = "",
                  lstm_delay = -1,
-                 self_repair_scale = None):
+                 self_repair_scale = None,
```
vijayaditya: @freewym please add comments explaining the differences between `self_repair_scale` and `self_repair_scale_clipgradient` in all the methods in `steps/nnet3/components.py` which use these features.

egs/wsj/s5/steps/nnet3/lstm/make_configs.py
```diff
@@ -76,6 +76,8 @@ def GetArgs():
                         help="clipping threshold used in ClipGradient components, if clipping-threshold=0 no clipping is done", default=30)
     parser.add_argument("--self-repair-scale", type=float,
                         help="A non-zero value activates the self-repair mechanism in the sigmoid and tanh non-linearities of the LSTM", default=None)
+    parser.add_argument("--self-repair-scale-clipgradient", type=float,
```
vijayaditya: It might be better to modify the defaults for both the  `--self-repair-scale` and ``--self-repair-scale-clipgradient` arguments as you find both these helpful.

vijayaditya: These changes would have to be made both in `components.py` and all the calling scripts.

egs/swbd/s5c/local/nnet3/run_lstm.sh
```diff
@@ -99,6 +99,7 @@ if [ $stage -le 9 ]; then
     --non-recurrent-projection-dim $non_recurrent_projection_dim \
     --label-delay $label_delay \
     --self-repair-scale 0.00001 \
```
vijayaditya: I would recommend renaming `--self-repair-scale` as `--self-repair-scale-nonlinearity`.

src/cudamatrix/cu-vector.cc
```diff
@@ -353,7 +353,7 @@ MatrixIndexT CuVectorBase<Real>::ApplyCeiling(Real ceiling_val) {
     cuda_vec_apply_ceiling(dimGrid, dimBlock, data_, ceiling_val, count_vec.Data(), dim_);
     CU_SAFE_CALL(cudaGetLastError());
     num_ceiled = count_vec.Sum();
-    CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyFloor", tim.Elapsed());
+    CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyCeiling", tim.Elapsed());
```
vijayaditya: Add a note in your commit message saying that you fixed this minor bug.

src/nnet3/nnet-simple-component.h
```diff
@@ -992,6 +1025,25 @@ class ClipGradientComponent: public Component {
                               // else element-wise absolute value clipping is
                               // done
 
+  // some configuration values relating to self-repairing.
+  BaseFloat self_repair_clipped_proportion_threshold_;
```
vijayaditya: describe each of these variables in detail

vijayaditya:
Please update the results for at least one recipe using this new feature.

vijayaditya:
Please add a unit test for this component.

vijayaditya:
Completed this round of review.

freewym:
ClipGradientComponent::BackProp() requires its argument "Component *to_update" to be non NULL for storing some stats. But the unit test code would pass value "NULL" to this argument with probability 0.5. I guess this is the reason why there is no unit test for this component in nnet-test-utils.cc

danpovey:
OK- can you please fix this?  The right fix is to just not store the stats
if 'to_update' is NULL, and if necessary, not to do the self-repair in this
case.


Dan

freewym:
Sure

src/cudamatrix/cu-vector.cc
galv: To be specific, __func__ outputs only "ApplyFloor" here. It doesn't include
the class name. It's that C legacy.

On Sun, May 22, 2016 at 1:39 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/cudamatrix/cu-vector.cc
> <https://github.com/kaldi-asr/kaldi/pull/803#discussion_r64155921>:
>
> > @@ -328,7 +328,7 @@ MatrixIndexT CuVectorBase<Real>::ApplyFloor(Real floor_val) {
> >      cuda_vec_apply_floor(dimGrid, dimBlock, data_, floor_val, count_vec.Data(), dim_);
> >      CU_SAFE_CALL(cudaGetLastError());
> >      num_floored = count_vec.Sum();
> > -    CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyFloor", tim.Elapsed());
> > +    CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
>
> Yiming, the reason we sometimes write e.g. CuVectorBase::ApplyFloor
> instead of *func*, is that when the same function exists in multiple
> classes (e.g. also in CuMatrixBase), we otherwise wouldn't know which one
> was taking time.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/kaldi-asr/kaldi/pull/803/files/5020094e0f6a6a4a407c3924367bc608e83a7323#r64155921>
>



-- 
Daniel Galvez

danpovey: Daniel's right, that's not what we want because there is a similarly named
function in class CuMatrix that we want to distinguish.

freewym: I think Daniel was looking at an outdated commit which is not the current
changes.

Yiming

On Mon, Jul 11, 2016 at 6:40 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/cudamatrix/cu-vector.cc
> <https://github.com/kaldi-asr/kaldi/pull/803#discussion_r70350398>:
>
> > @@ -328,7 +328,7 @@ MatrixIndexT CuVectorBase<Real>::ApplyFloor(Real floor_val) {
> >      cuda_vec_apply_floor(dimGrid, dimBlock, data_, floor_val, count_vec.Data(), dim_);
> >      CU_SAFE_CALL(cudaGetLastError());
> >      num_floored = count_vec.Sum();
> > -    CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyFloor", tim.Elapsed());
> > +    CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
>
> Daniel's right, that's not what we want because there is a similarly named
> function in class CuMatrix that we want to distinguish. On Mon, Jul 11,
> 2016 at 3:38 PM, Daniel Galvez <notifications@github.com> wrote:
> … <#m_8952389131996847612_>
> In src/cudamatrix/cu-vector.cc <#803 (comment)
> <https://github.com/kaldi-asr/kaldi/pull/803#discussion_r70350264>>: > @@
> -328,7 +328,7 @@ MatrixIndexT CuVectorBase<Real>::ApplyFloor(Real
> floor_val) { > cuda_vec_apply_floor(dimGrid, dimBlock, data_, floor_val,
> count_vec.Data(), dim_); > CU_SAFE_CALL(cudaGetLastError()); > num_floored
> = count_vec.Sum(); > -
> CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyFloor",
> tim.Elapsed()); > + CuDevice::Instantiate().AccuProfile(__func__,
> tim.Elapsed()); To be specific, __func__ outputs only "ApplyFloor" here. It
> doesn't include the class name. It's that C legacy. …
> <#m_-7613624424334759022_> On Sun, May 22, 2016 at 1:39 PM, Daniel Povey
> ***@***.***> wrote: In src/cudamatrix/cu-vector.cc <#803
> <https://github.com/kaldi-asr/kaldi/pull/803> (comment) <#803 (comment)
> <https://github.com/kaldi-asr/kaldi/pull/803#discussion_r64155921>>>: >
> @@ -328,7 +328,7 @@ MatrixIndexT CuVectorBase<Real>::ApplyFloor(Real
> floor_val) { > cuda_vec_apply_floor(dimGrid, dimBlock, data_, floor_val,
> count_vec.Data(), dim_); > CU_SAFE_CALL(cudaGetLastError()); > num_floored
> = count_vec.Sum(); > -
> CuDevice::Instantiate().AccuProfile("CuVectorBase::ApplyFloor",
> tim.Elapsed()); > + CuDevice::Instantiate().AccuProfile(__func__,
> tim.Elapsed()); Yiming, the reason we sometimes write e.g.
> CuVectorBase::ApplyFloor instead of *func*, is that when the same function
> exists in multiple classes (e.g. also in CuMatrixBase), we otherwise
> wouldn't know which one was taking time. — You are receiving this because
> you are subscribed to this thread. Reply to this email directly or view it
> on GitHub <
> https://github.com/kaldi-asr/kaldi/pull/803/files/5020094e0f6a6a4a407c3924367bc608e83a7323#r64155921>
> -- Daniel Galvez — You are receiving this because you were mentioned. Reply
> to this email directly, view it on GitHub <
> https://github.com/kaldi-asr/kaldi/pull/803/files/5020094e0f6a6a4a407c3924367bc608e83a7323#r70350264>,
> or mute the thread <
> https://github.com/notifications/unsubscribe/ADJVuyOV5Q_GS1v6vUNCvBSL3VmTXfq-ks5qUsYAgaJpZM4IkDwb>
> .
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/kaldi-asr/kaldi/pull/803/files/5020094e0f6a6a4a407c3924367bc608e83a7323#r70350398>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ADWAktIeQO9WmZko4-gg1jIu03e_Y-jBks5qUsZSgaJpZM4IkDwb>
> .
>



-- 
Yiming Wang
Department of Computer Science
The Johns Hopkins University
3400 N. Charles St.
Baltimore, MD 21218

freewym:
I found another reason for not including ClipGradientComponent in the unit test: if in-deriv is clipped, it will not pass the numerical test for derivatives since predicted_objf_change and measured_objf_change will no longer be close to each other. It will pass the unit test only if there is actually no clipping, i.e., if I set clipping-threshold=0 in the test.

vijayaditya:
The component is expected to fail numerical gradient tests, so we are not
interested in that. The idea is to write a unit test which just ensures
that the component works as intended.

--Vijay

freewym:
So we just need to turn off the clipping to avoid test failure?

freewym:
Done.

vijayaditya:
Are you planning to include the component level gradient clipping in the same commit ? If so could you please post the latest results here.

danpovey:
I think we should make the reintroduction of component-level gradient
clipping (the per-component max-change) a separate commit, as it's kind of
an orthogonal issue.  Naturally if there is some reason to combine them
that will make your lives easier, that's different.

src/nnet3/nnet-simple-component.cc
```diff
@@ -659,21 +709,109 @@ void ClipGradientComponent::Backprop(const std::string &debug_info,
...
     }
+
+    if (to_update != NULL) {
+      to_update->num_backpropped_ += 1;
+      RepairGradients(debug_info, in_value, in_deriv, to_update);
+    }
   }
 }
 
+void ClipGradientComponent::RepairGradients(
```
vijayaditya: Please describe the logic for repair process in the comment.

src/nnet3/nnet-test-utils.cc
```diff
@@ -1205,6 +1205,14 @@ static void GenerateRandomComponentConfig(std::string *component_type,
          << " use-natural-gradient=" << std::boolalpha << use_natural_gradient;
       break;
     }
+    case 28: {
+      *component_type = "ClipGradientComponent";
+      os << "dim=" << RandInt(1, 50);
+      os << " clipping-threshold=" << RandInt(1, 50)
+         << " norm-based-clipping=" << (RandInt(0, 1) == 0 ? "false" : "true")
```
vijayaditya: please test all possible config lines.

vijayaditya:
Looks good I will merge when the new comments have been addressed.

freewym:
Just pushed the fix and rebase.

src/nnet3/nnet-test-utils.cc
```diff
@@ -1205,6 +1205,16 @@ static void GenerateRandomComponentConfig(std::string *component_type,
...
+         << " self-repair-scale=" << (RandInt(0, 1) == 0 ? 0 : RandInt(1, 50))
+         << " self-repair-clipped-proportion-threshold=" << RandUniform()
+         << " self-repair-target=" << RandUniform();
```
vijayaditya: randomly add the optional arguments.

freewym:
OK. Done.

vijayaditya:
I am assuming you ran all the unit tests.

PR #804, opened 2016-05-22
Sre nnet1

mallidi:


danpovey:
That was a very huge pull request!  What were/are you hoping to check in? 
Dan

PR #805, opened 2016-05-23
Add biphone chain recipes for swbd (B)LSTM setups

naxingyu:
Results at the beginning of scripts.


danpovey:
cool- I'll wait for Vijay to check and merge.

danpovey:
Merging myself.
If you have time, please see whether doing an extra pass of GMM-SAT training with a left-biphone tree helps.  I think it might make the alignments more "left-biphone-compatible".

naxingyu:
Sounds interesting…… I'll try.

PR #806, opened 2016-05-23
minor bugs have been fixed on nnet1 make_cnn_proto.py file

haihua:


danpovey:
@vesis84 should review this.

vesis84:
Hi, yes, it looks simpler now. Merging in! 

Also, would you please have a suggestion for a more standard 'naming' of what was called as 'patch' and 'pool' ? I have a feeling that the names could be more intuitive, but I am not sure what would be the best...

Issue #797, opened 2016-05-20
Make combine_data.sh able to combine data with and without segments files.


danpovey:
... probably by adding segments with "0 -1" as timings.

jfainberg:
Happy to do this unless somebody else is already working on it. Assuming utterance and recording IDs are the same for data directories that didn't have segments, so "uttid uttid 0 -1"? (So "awk '{ print $1" "$1" 0 -1" }' $f/utt2spk")

danpovey:
Yes they are the same.
Dan

danpovey:
I haven't looked at your PR yet, but I just realized that there is something else that needs to be done in order for these combined data dirs to later be usable by get_utt2dur.sh.  I don't want the -1's in the segments file because it will be hard to handle in get_utt2dur.sh, plus it could be potentially confusing for many scripts to have to deal with this. Instead, you should work out the durations by calling get_utt2dur.sh on the input directories that lack segments files, and use those to fill out "proper values" in the segments files.

I'm thinking of changing validate_data_dir.sh also, to disallow the -1's in the segments file.

jfainberg:
That makes sense, thanks! I'll make my PR a WIP and have a go at using get_utt2dur.sh over the next few days.

PR #808, opened 2016-05-23
WIP: combine_data.sh: combine data with and without segments

jfainberg:
See #797 . Requires first source directory to include a segments file (as before). Subsequent directories which do not include one will have one generated with the format: "uttid uttid 0 -1".

Could move the check for segments out of the inner for loop, but that would duplicate some code (cat, sort etc.). This is still pretty much instant on some large corpora. Not sure if there are more elegant ways of doing this.

Note that for a data directory which has uttid ≠ recid, the new segments will be pruned by fix_data_dir.sh as it compares wav.scp to the 2nd column in segments.

### DIFF
--- a/egs/wsj/s5/utils/combine_data.sh
+++ b/egs/wsj/s5/utils/combine_data.sh
@@ -71,24 +71,20 @@ fi
 extra_files=$(echo "$extra_files"|sed -e "s/utt2uniq//g")
 
 # segments are treated similarly to utt2uniq. If it exists in some, but not all
-# src directories, then we create segments files where necessary.
-# These are stored in the src directories.
+# src directories, then we generate segments where necessary.
 has_segments=false
 segments_absent_somewhere=false
 for in_dir in $*; do
   if [ -f $in_dir/segments ]; then
     has_segments=true
-  else
-    segments_absent_somewhere=true
+    break
   fi
 done
 
 if $has_segments; then
-  if $segments_absent_somewhere; then
-    echo "$0 [info]: will generate missing segments"
-  fi
   for in_dir in $*; do
     if [ ! -f $in_dir/segments ]; then
+      echo "$0 [info]: will generate missing segments for $in_dir" 1>&2
       utils/data/get_segments_for_data.sh $in_dir
     else
       cat $in_dir/segments
--- a/egs/wsj/s5/utils/combine_data.sh
+++ b/egs/wsj/s5/utils/combine_data.sh
@@ -9,7 +9,7 @@
 # about what these directories contain.
 
 # Begin configuration section.
-extra_files= # specify addtional files in 'src-data-dir' to merge, ex. "file1 file2 ..."
+extra_files= # specify additional files in 'src-data-dir' to merge, ex. "file1 file2 ..."
 skip_fix=false # skip the fix_data_dir.sh in the end
 # End configuration section.
 
@@ -20,7 +20,8 @@ if [ -f path.sh ]; then . ./path.sh; fi
 
 if [ $# -lt 2 ]; then
   echo "Usage: combine_data.sh [--extra-files 'file1 file2'] <dest-data-dir> <src-data-dir1> <src-data-dir2> ..."
-  echo "Note, files that don't appear in first source dir will not be added even if they appear in later ones."
+  echo "Note, files that don't appear in all source dirs will not be combined,"
+  echo "with the exception of utt2uniq and segments, which are created where necessary."
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


danpovey:
Apologies, but new plan.
I think a better way to do this would be as follows:
create a script
utils/data/add_segments_to_data.sh <in-data> <out-data>
that will add segments to a data-dir that has none.
This should not add -1's as the segment ends though, it should use
get_utt2dur.sh to get durations and use those.

Then if combine_data.sh notices that some but not all of the inputs have
segments, it will call that script first to add segments, and then recurse
with modified inputs (or just modify the variables).

combine_data.sh should also be modified so that if some but not all the
inputs have a reco2file_and_channel file, then the output will have no such
file (and a warning will be printed).  The same should be done for various
optional auxiliary data-types, such as utt2dur, and spk2gender.

Note: if there is no segments file, then the wav.scp is by definition
indexed by utterance.

It might be easiest to re-implement combine_data_dir.sh as a python
script-- e.g. to rewrite it as python, and have the .sh file just link to
the python script.   Up to you though-- if you're still up for this project.

Dan

jfainberg:
I think this sounds good. I'm very happy to keep going, though it might take me a few more days. Let me know if any parts are urgent.

Joachim

danpovey:
Thanks - not a hurry.

jfainberg:
combine_data.sh (in the current repo) just cat's the files together, but it runs utils/fix_data_dir.sh at the end if skip_fix is set to false (see beginning of combine_data.sh).

danpovey:
@jfainberg, are you planning to work on this?

jfainberg:
Yes, sorry I was more busy than I anticipated. Will work on it this week.

danpovey:
looks plausible.

jfainberg:
I will run lots of tests on the typical Kaldi recipes to make sure the output is identical to that of combine_segments.sh, and also see if I can simplify it further. Will report when it's ready for review.

Are there any circumstances in which we would want to combine files that do not occur in all source directories, yet we can not generate them (e.g. feats.scp)?

danpovey:
If files are only present in some directories then they shouldn't be
combined because the validation would fail (or fix_data_dir.pl would remove
those utterances/speakers that don't have that info).
Dan

jfainberg:
Will test today. Also, apologies, didn't mean to remove the copyright notice in combine_data.sh, will put it back!

jfainberg:
Ok, I think this is ready for review. Is it along the lines of what you were thinking? Happy to change / refactor.

I think the branch conflict is that combine_data.sh was changed after I pulled it.

danpovey:
Can you please resolve the conflict before I review?

jfainberg:
Rebased and squashed the commits.

danpovey:
Hm.
I just noticed that you have replaced combine_data.sh with a python script and made it call the python script.
This is potentially quite confusing.  Can you explain what aspects of combine_data.sh were hard for you to implement in shell-script?
This isn't about the relative merits of python versus bash, it's just that I don't want to convert such a long-standing script into python.  For example, people may have local change that this would cause a merge conflict for.  I'd rather keep the changes as small as possible.

jfainberg:
I think I thought some things were more elegant in Python, but I agree (and I've learnt a few things along the way). I've rolled back my changes and instead modified the original combine_data.sh. I've treated segments as with utt2uniq, with the exception that segments files are created in the src directories, whereas for utt2uniq they're piped directly to the destination directory. This way add_segments_to_data.sh works the same way as get_utt2dur.sh.

danpovey:
Actually-- and this is 100% my fault-- I realized there is a potential
problem with the design I outlined, where the src directories are modified
by adding the segments file.
Firstly, it breaks a little bit the notion that the input directories are
not changed by this script... we relax this for utt2dur, but utt2dur is a
derived file.
The potentially more serious problem is that by adding the segments file,
we may change very slightly (due to rounding effects) the length of
features generated, and this might cause other things to break.
How hard would it be to change this so the segments file is created on the
fly, e.g. rename add_segments_to_data.sh to get_segments_for_data.sh and
get it to put its output on the stdout?

Dan

jfainberg:
No worries at all, that makes sense. The change was quite simple (please see latest commit), the only thing to be careful about is anything else that echoes to stdout in get_segments_for_data.sh and recursively in get_utt2dur.sh (if called by get_segments_for_data.sh). I've removed any info messages in get_segments_for_data.sh and piped those from get_utt2dur.sh to /dev/null. 

To avoid this we could remove get_segments_for_data.sh and instead incorporate it inside combine_data.sh, as with utt2uniq. But it might be nice with get_segments_for_data.sh as a separate script for other uses.

Joachim

egs/wsj/s5/utils/data/get_segments_for_data.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "Usage: $0 [options] <datadir>"
+  echo "e.g.:"
+  echo " $0 data/train > data/train/segments"
+  exit 1
+fi
+
+data=$1
+
+if [ ! -f $data/utt2dur ]; then
+  utils/data/get_utt2dur.sh $data 2>/dev/null || exit 1;
```
danpovey: Actually stderr is not the problem here, it's stdout, in case any messages are echoed to stdout.
Instead of `2>/dev/null` you should be doing `1>&2`

egs/wsj/s5/utils/combine_data.sh
```diff
@@ -71,24 +71,20 @@ fi
 extra_files=$(echo "$extra_files"|sed -e "s/utt2uniq//g")
 
 # segments are treated similarly to utt2uniq. If it exists in some, but not all
-# src directories, then we create segments files where necessary.
-# These are stored in the src directories.
+# src directories, then we generate segments where necessary.
 has_segments=false
 segments_absent_somewhere=false
```
danpovey: we probably no longer need this variable.

egs/wsj/s5/utils/combine_data.sh
```diff
@@ -71,24 +71,20 @@ fi
...
 done
 
 if $has_segments; then
-  if $segments_absent_somewhere; then
-    echo "$0 [info]: will generate missing segments"
-  fi
   for in_dir in $*; do
     if [ ! -f $in_dir/segments ]; then
+      echo "$0 [info]: will generate missing segments for $in_dir" 1>&2
       utils/data/get_segments_for_data.sh $in_dir
```
danpovey: At this point you should also probably redirect the stdout to stderr by adding 1>&2.
The reason is that get_segments_for_data.sh may print out messages to stdout if path.sh, cmd.sh or parse_options.sh do so.

danpovey: oh sorry this doesn't make sense as we need the stdout.
Instead  you could go through parse_options.sh and make sure that all error messages go to stderr not stdout.

danpovey: actually that's a separate issue, so I'll merge this anyway.

danpovey: ...actually I'll wait till you remove
 segments_absent_somewhere=false

jfainberg: Thank you! And thanks for your patience on this one.

danpovey:
Thanks!  Merging.

PR #809, opened 2016-05-24
Bug fix in steps/train_sat_basis.sh

jcsilva:
Binaries gmm-basis-fmllr-accs-gpost and gmm-est-basis-fmllr-gpost were not receiving the spk2utt information. With this configuragion, these scripts broke when there were several utterances per speaker.

### DIFF
--- a/egs/wsj/s5/steps/train_sat_basis.sh
+++ b/egs/wsj/s5/steps/train_sat_basis.sh
@@ -90,7 +90,7 @@ esac
 ## Get initial fMLLR transforms (possibly from alignment dir)
 if [ -f $alidir/trans.1 ]; then
   echo "$0: Using transforms from $alidir"
-  feats="$sifeats transform-feats --utt2spk=ark:$sdata/JOB/utt2spk ark,s,cs:$alidir/trans.JOB ark:- ark:- |"
+  feats="$sifeats transform-feats ark,s,cs:$alidir/trans.JOB ark:- ark:- |"
   cur_trans_dir=$alidir
 else 
   if [ $stage -le -5 ]; then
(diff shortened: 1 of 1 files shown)


danpovey:
Hm.
I think it was intentional that this was done on a per-utterance basis.  If there were not many speakers, you wouldn't be able to estimate a good basis.  
I think the real error is in this line:
  feats="$sifeats transform-feats --utt2spk=ark:$sdata/JOB/utt2spk ark:$dir/trans.JOB ark:- ark:- |"
where the --utt2spk option should not be present.

jcsilva:
Well, when I made those changes, I thought that steps/train_sat_basis could be used even in situations where we would use train_sat (e.g. there were several speakers and enough utterances per speaker). And if it was desired training in a per-utterance way, it would suffice to create the spk2utt file properly.

But according to your comment, train_sat_basis.sh should be used in a specific situation (when there were not so much data for each speaker), right? In this case, --utt2spk should be suppressed from all transform-feats that exist in train_sat_basis.

danpovey:
> Well, when I made those changes, I thought that steps/train_sat_basis
> could be used even in situations where we would use train_sat (e.g. there
> were several speakers and enough utterances per speaker). And if it was
> desired training in a per-utterance way, it would suffice to create the
> spk2utt file properly.
>
Yes but people won't want to change the spk2utt file, typically.  Your
changes would change the behavior of existing example scripts- they'd have
to be changed or at least re-tested.

But according to your comment, train_sat_basis.sh should be used in a
> specific situation (when there were not so much data for each speaker),
> right? In this case, --utt2spk should be suppressed from all
> transform-feats that exist in train_sat_basis.
>
> That would be much easier.
Dan

jcsilva:
I agree. I have already updated my commit...

danpovey:
OK, merging.  Thanks!

danpovey:
Segments files with a 5th field that represents the channel are not supported at the script level, although they are supported in the 'extract-segments' binary.  Such a data-dir would fail validation.

PR #810, opened 2016-05-24
Multi basis

vesis84:
Hi Dan, there are some new 'nnet1' components and binaries for processing data which were handy to have in kaldi. Could you please take a look? All the best, Karel.


src/nnetbin/append-post-to-feats.cc
```diff
@@ -0,0 +1,145 @@
...
+  try {
+    using namespace kaldi::nnet1;
+    using namespace kaldi;
+    using namespace std;
+
+    const char *usage =
+        "Append posteriors to features\n"
+        "\n"
+        "Usage: append-post-to-feats <in-rspecifier1> <in-rspecifier2> <out-wspecifier>\n"
+        " or: append-post-to-feats <in-rxfilename1> <in-rxfilename2> <out-wxfilename>\n"
```
danpovey: Usage message should mention options as [options] and should have an example line showing --post-dim.

src/nnetbin/append-post-to-feats.cc
```diff
@@ -0,0 +1,145 @@
...
+        "See also: paste-feats, concat-feats, append-vector-to-feats\n";
+
+    ParseOptions po(usage);
+
+    bool binary = true;
+    po.Register("binary", &binary, "If true, output files in binary "
+                "(only relevant for single-file operation, i.e. no tables)");
+
+    int32 post_dim = 0;
+    po.Register("post_dim", &post_dim, "Dimensionality of the posteriors.");
```
danpovey: "post_dim" should be "post-dim".

src/nnetbin/post-to-feat.cc
```diff
@@ -0,0 +1,135 @@
...
+// limitations under the License.
+
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "matrix/kaldi-matrix.h"
+#include "nnet/nnet-utils.h"
+
+/*
+namespace kaldi {
```
danpovey: remove comment

src/nnetbin/post-to-feat.cc
```diff
@@ -0,0 +1,135 @@
...
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi::nnet1;
+    using namespace kaldi;
+    using namespace std;
+
+    const char *usage =
+        "Convert posteriors to features\n"
+        "\n"
+        "Usage: post-to-feat <in-rspecifier> <out-wspecifier>\n"
```
danpovey: same comments as 'append-post-to-feats' apply.

danpovey: there should be cross-referencing in "see also" with post-to-weights.  make sure it goes both ways.

src/bin/post-to-weights.cc
```diff
@@ -33,7 +33,8 @@ int main(int argc, char *argv[]) {
         "weight-silence-post, to get silence weights)\n"
         "See also: weight-silence-post, post-to-pdf-post, post-to-phone-post\n"
         "get-post-on-ali\n"
-        "Usage: post-to-weights <post-rspecifier> <weights-wspecifier>\n";
+        "Usage: post-to-weights <post-rspecifier> <weights-wspecifier>\n"
+        "See also: post-to-feat\n";
```
danpovey: please consolidate the "see also" messages into one group.

danpovey:
Karel, please move post-to-feats and append-post-to-feats to from nnetbin/ to bin/, and make sure they don't include anything from nnet/.  That's where someone would expect to find them (well, there or featbin/) if they were looking for such programs, so having them in nnetbin/ is confusing.

vesis84:
Hi Dan, okay, I can do that... But then I have to move this function from 'nnet/nnet-utils.h':

// converts Posterior to Matrix,
template <typename Real>
void PosteriorToMatrix(const Posterior &post, const int32 post_dim, Matrix<Real> *mat);

It would be logical to put it to 1st place 'hmm/posterior.h', and add there the dependency on 'matrix/kaldi-matrix.h'.
Or I could put it to 2nd place 'feat/feature-functions.h', but there it adds the dependency on 'hmm/posterior.h'.

Perhaps, the 2nd place is introducing dependency on smaller amount of code, while the 1st place is more logical. So I find it hard to decide which is better... Which one would you prefer?

danpovey:
> Hi Dan, okay, I can do that... But then I have to move this function from
> 'nnet/nnet-utils.h':
>
> // converts Posterior to Matrix,
> template
> void PosteriorToMatrix(const Posterior &post, const int32 post_dim, Matrix
> *mat);
>
> It would be logical to put it to 1st place 'hmm/posterior.h', and add
> there the dependency on 'matrix/kaldi-matrix.h'.
> Or I could put it to 2nd place 'feat/feature-functions.h', but there it
> adds the dependency on 'hmm/posterior.h'.
>
> Perhaps, the 2nd place is introducing dependency on smaller amount of
> code, while the 1st place is more logical. So I find it hard to decide
> which is better... Which one would you prefer?
>
I prefer it to go in hmm/posterior.h.  The hmm directory already seems to
depend on the matrix directory (based on looking at the Makefile), and I
only tend to mind new dependencies if they are new at the directory level.

danpovey:
Thanks..
sorry, one more thing- it would be more consistent with the other names of programs to rename post-to-feat to post-to-feats.  Obviously make sure all usage messages are up-to-date after that.  Then, rebasing would be helpful.

vesis84:
Okay, I'll do that. I was also thinking of consistency too, there is a binary 'append-post-to-feats' so it is logical to have 'post-to-feats' as you propose. 

Then there is another family of binares 'feat-to-{dim,len,post}', and perhaps it is not a big deal that we have a pair of conversion tools whose names are not 100% symmetric (post-to-feats and feat-to-post).

Okay, the commit will be ready in a few moments.

vesis84:
Wow, this is how it looks in the web-app? :+1: (The commits moved)

src/featbin/post-to-feats.cc
```diff
@@ -0,0 +1,107 @@
...
+        "See also: post-to-weights feat-to-post, append-vector-to-feats, append-post-to-feats\n";
+
+    ParseOptions po(usage);
+
+    bool binary = true;
+    po.Register("binary", &binary, "If true, output files in binary "
+                "(only relevant for single-file operation, i.e. no tables)");
+
+    int32 post_dim = 0;
+    po.Register("post_dim", &post_dim, "Dimensionality of the posteriors.");
```
danpovey: post_dim -> post-dim

danpovey:
OK thanks, but actually what I meant by rebase in this instance is to squash the commits, I'd rather have a clean history.
You can commit yourself after fixing post_dim->post-dim and squashing.

PR #814, opened 2016-05-26
added option to compress features while resaving in nnet/{train,pretrain_dbn}.sh

mallidi:
This reduces space requirement in /tmp folders


vesis84:
Hi, it looks good. We already consulted this by email. Merging!

PR #822, opened 2016-05-30
modify_lr_fix: Removing nnet3-modify-learning-rates as it is not requ…

vimalmanohar:
…equired


egs/wsj/s5/steps/nnet3/train_discriminative.sh
```diff
@@ -57,7 +57,7 @@ num_threads=16  # this is the default but you may want to change it, e.g. to 1 i
 
 cleanup=true
 keep_model_iters=1
-retroactive=false
+retroactive=false   # relates to modify-learning-rates [deprecated]
```
danpovey: you can actually remove any of these options that are not currently used in any example script.   I imagine things like --retroactive are never actually used.

vimalmanohar:
Removed the options that can be removed.

PR #823, opened 2016-05-31
Fixed pitch_functions to behave correctly with snip_edges=false

bpotard:
The snip_edges option had not been implemented correctly in the pitch tracker, and while it was generating the right number of ouput frames, the delay was not at all as expected.

I have therefore fixed snip_edges=false behaviour to have an expected feature delay of frame_shift/2 instead of being identical to the delay of snip_edges=true (which itself has a delay > frame_length / 2).

Added a unit test that ensure snip_edges=false behaves as expected, by calculating the actual delay between outputs only differing by snip_edge through cross-correlation.

Note that the feature delay of snip_edges=true (the default) is rather difficult to predict (and almost certainly not what people assume), which almost certainly makes the use of pitch features problematic at the moment.
I would suggest switching to snip_edges=false for new recipes.

### DIFF
--- a/src/feat/feature-window.cc
+++ b/src/feat/feature-window.cc
@@ -105,16 +105,20 @@ FeatureWindowFunction::FeatureWindowFunction(const FrameExtractionOptions &opts)
   int32 frame_length = opts.WindowSize();
   KALDI_ASSERT(frame_length > 0);
   window.Resize(frame_length);
+  double a = M_2PI / (frame_length-1);
   for (int32 i = 0; i < frame_length; i++) {
-    BaseFloat i_fl = static_cast<BaseFloat>(i);
+    double i_fl = static_cast<double>(i);
     if (opts.window_type == "hanning") {
-      window(i) = 0.5  - 0.5*cos(M_2PI * i_fl / (frame_length-1));
+      window(i) = 0.5  - 0.5*cos(a * i_fl);
     } else if (opts.window_type == "hamming") {
-      window(i) = 0.54 - 0.46*cos(M_2PI * i_fl / (frame_length-1));
+      window(i) = 0.54 - 0.46*cos(a * i_fl);
     } else if (opts.window_type == "povey") {  // like hamming but goes to zero at edges.
-      window(i) = pow(0.5 - 0.5*cos(M_2PI * i_fl / (frame_length-1)), 0.85);
+      window(i) = pow(0.5 - 0.5*cos(a * i_fl), 0.85);
     } else if (opts.window_type == "rectangular") {
       window(i) = 1.0;
+    } else if (opts.window_type == "blackman") {
+      window(i) = opts.blackman_coeff - 0.5*cos(a * i_fl) +
+        (0.5 - opts.blackman_coeff) * cos(2 * a * i_fl);
     } else {
       KALDI_ERR << "Invalid window type " << opts.window_type;
     }
--- a/src/feat/pitch-functions.cc
+++ b/src/feat/pitch-functions.cc
@@ -773,13 +773,22 @@ int32 OnlinePitchFeatureImpl::NumFramesAvailable(
   // of frames only if the input is not finished.
   if (!input_finished_)
     frame_length += nccf_last_lag_;
-  if (num_downsampled_samples < frame_length) return 0;
-  else
-    if (input_finished_ && !snip_edges) {
-      return (int32)(num_downsampled_samples * 1.0f / frame_shift + 0.5f);
+  if (num_downsampled_samples < frame_length) {
+    return 0;
+  } else {
+    if (!snip_edges) {
+      if (input_finished_) {
+        return static_cast<int32>(num_downsampled_samples * 1.0f /
+                                  frame_shift + 0.5f);
+      } else {
+        return static_cast<int32>((num_downsampled_samples - frame_length / 2) *
+                                   1.0f / frame_shift + 0.5f);
+      }
+    } else {
+      return static_cast<int32>((num_downsampled_samples - frame_length) /
+                                 frame_shift + 1);
     }
-    else
-      return ((num_downsampled_samples - frame_length) / frame_shift) + 1;
+  }
 }
 
 void OnlinePitchFeatureImpl::UpdateRemainder(
@@ -835,19 +844,35 @@ void OnlinePitchFeatureImpl::ExtractFrame(
   int32 offset = static_cast<int32>(sample_index -
                                     downsampled_samples_processed_);
 
+  // Treat edge cases first
+  if (sample_index < 0) {
+    // Part of the frame is before the beginning of the signal. This
+    // should only happen if opts_.snip_edges == false, when we are
+    // processing the first few frames of signal. In this case
+    // we pad with zeros.
+    KALDI_ASSERT(opts_.snip_edges == false);
+    int32 sub_frame_length = sample_index + full_frame_length;
+    int32 sub_frame_index = full_frame_length - sub_frame_length;
+    KALDI_ASSERT(sub_frame_length > 0 && sub_frame_index > 0);
+    window->SetZero();
+    SubVector<BaseFloat> sub_window(*window, sub_frame_index, sub_frame_length);
+    ExtractFrame(downsampled_wave_part, 0, &sub_window);
+    return;
+  }
+
   if (offset + full_frame_length > downsampled_wave_part.Dim()) {
     // Requested frame is past end of the signal.  This should only happen if
     // input_finished_ == true, when we're flushing out the last couple of
     // frames of signal.  In this case we pad with zeros.
     KALDI_ASSERT(input_finished_);
-    int32 new_full_frame_length = downsampled_wave_part.Dim() - offset;
-    KALDI_ASSERT(new_full_frame_length > 0);
+    int32 sub_frame_length = downsampled_wave_part.Dim() - offset;
+    KALDI_ASSERT(sub_frame_length > 0);
     window->SetZero();
-    SubVector<BaseFloat> sub_window(*window, 0, new_full_frame_length);
+    SubVector<BaseFloat> sub_window(*window, 0, sub_frame_length);
     ExtractFrame(downsampled_wave_part, sample_index, &sub_window);
     return;
   }
-  
+
   // "offset" is the offset of the start of the frame, into this
   // signal.
   if (offset >= 0) {
@@ -1076,7 +1101,17 @@ void OnlinePitchFeatureImpl::AcceptWaveform(
 
   for (int32 frame = start_frame; frame < end_frame; frame++) {
     // start_sample is index into the whole wave, not just this part.
-    int64 start_sample = static_cast<int64>(frame) * frame_shift;
+    int64 start_sample;
+    if (opts_.snip_edges) {
+      // Usual case: offset starts at 0
+      start_sample = static_cast<int64>(frame) * frame_shift;
+    } else { 
+      // When we are not snipping the edges, the first offsets may be
+      // negative. In this case we will pad with zeros, it should not impact
+      // the pitch tracker.
+      start_sample = 
+        static_cast<int64>((frame + 0.5) * frame_shift) - full_frame_length / 2;
+    }
     ExtractFrame(downsampled_wave, start_sample, &window);
     if (opts_.nccf_ballast_online) {
       // use only up to end of current frame to compute root-mean-square value.
--- a/src/feat/pitch-functions.h
+++ b/src/feat/pitch-functions.h
@@ -138,7 +138,8 @@ struct PitchExtractionOptions {
                    "file, if specified there)");
     opts->Register("frame-length", &frame_length_ms, "Frame length in "
                    "milliseconds");
-    opts->Register("frame-shift", &frame_shift_ms, "Frame shift in milliseconds");
+    opts->Register("frame-shift", &frame_shift_ms, "Frame shift in "
+                   "milliseconds");
     opts->Register("preemphasis-coefficient", &preemph_coeff,
                    "Coefficient for use in signal preemphasis (deprecated)");
     opts->Register("min-f0", &min_f0,
@@ -190,20 +191,21 @@ struct PitchExtractionOptions {
                    "if --frames-per-chunk > 0 and "
                    "--simulate-first-pass-online=true");
     opts->Register("snip-edges", &snip_edges, "If this is set to false, the "
-                   "incomplete frames near the ending edge won't be snipped, so "
-                   "that the number of frames is the file size divided by the "
-                   "frame-shift. This makes different types of features give the "
-                   "same number of frames.");
-
+                   "incomplete frames near the ending edge won't be snipped, "
+                   "so that the number of frames is the file size divided by "
+                   "the frame-shift. This makes different types of features "
+                   "give the same number of frames.");
   }
   /// Returns the window-size in samples, after resampling.  This is the
   /// "basic window size", not the full window size after extending by max-lag.
+  // Because of floating point representation, it is more reliable to divide
+  // by 1000 instead of multiplying by 0.001, but it is a bit slower.
   int32 NccfWindowSize() const {
-    return static_cast<int32>(resample_freq * 0.001 * frame_length_ms);
+    return static_cast<int32>(resample_freq * frame_length_ms / 1000.0);
   }
   /// Returns the window-shift in samples, after resampling.
   int32 NccfWindowShift() const {
-    return static_cast<int32>(resample_freq * 0.001 * frame_shift_ms);
+    return static_cast<int32>(resample_freq * frame_shift_ms / 1000.0);
   }
 };
 
@@ -224,12 +226,12 @@ struct ProcessPitchOptions {
 
   int32 delta_window;
   int32 delay;
-  
-  bool add_pov_feature;  
-  bool add_normalized_log_pitch;  
+
+  bool add_pov_feature;
+  bool add_normalized_log_pitch;
   bool add_delta_pitch;
   bool add_raw_log_pitch;
-  
+
   ProcessPitchOptions() :
       pitch_scale(2.0),
       pov_scale(2.0),
@@ -272,7 +274,8 @@ struct ProcessPitchOptions {
                    "Number of frames on each side of central frame, to use for "
                    "delta window.");
     opts->Register("delay", &delay,
-                   "Number of frames by which the pitch information is delayed.");
+                   "Number of frames by which the pitch information is "
+                   "delayed.");
     opts->Register("add-pov-feature", &add_pov_feature,
                    "If true, the warped NCCF is added to output features");
     opts->Register("add-normalized-log-pitch", &add_normalized_log_pitch,
@@ -301,7 +304,7 @@ class OnlinePitchFeature: public OnlineBaseFeature {
   virtual int32 Dim() const { return 2; /* (NCCF, pitch) */ }
 
   virtual int32 NumFramesReady() const;
-  
+
   virtual BaseFloat FrameShiftInSeconds() const;
 
   virtual bool IsLastFrame(int32 frame) const;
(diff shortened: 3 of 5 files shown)


danpovey:
please be more specific in the commit message title, e.g. "misc. changes"... changes to what?

src/feat/pitch-functions.cc
```diff
@@ -778,8 +778,9 @@ int32 OnlinePitchFeatureImpl::NumFramesAvailable(
     if (input_finished_ && !snip_edges) {
       return (int32)(num_downsampled_samples * 1.0f / frame_shift + 0.5f);
     }
-    else
-      return ((num_downsampled_samples - frame_length) / frame_shift) + 1;
+  // BP: handles snip_edges == false in a more graceful way
+    else if (!snip_edges) return (int32)((num_downsampled_samples - frame_length/2) * 1.0f / frame_shift + 0.5f);
```
danpovey: past the 80-char line length limit.
probably best to remove this comment.
you may need to use braces on the if-statement to stay within the line length limit.

src/feat/pitch-functions.cc
```diff
@@ -856,14 +857,21 @@ void OnlinePitchFeatureImpl::ExtractFrame(
   } else {
     // frame is partly in the remainder and partly in the new part.
     int32 remainder_offset = downsampled_signal_remainder_.Dim() + offset;
-    KALDI_ASSERT(remainder_offset >= 0);  // or we didn't keep enough remainder.
+    /*KALDI_ASSERT(remainder_offset >= 0);*/  // or we didn't keep enough remainder.
```
danpovey: better to remove the line than comment it out.

danpovey: I'd do 
```
KALDI_ASSERT(remainder_offset > 0 || opts_.snip_edges == false);
// if snip_edges == false, we may be attempting to process some waveform 
// at the start of the file that doesn't exist, and this leads to remainder_offset < 0.
```

src/feat/pitch-functions.cc
```diff
@@ -856,14 +857,21 @@ void OnlinePitchFeatureImpl::ExtractFrame(
...
     KALDI_ASSERT(offset + full_frame_length > 0);  // or we should have
                                                    // processed this frame last
                                                    // time.
 
     int32 old_length = -offset, new_length = offset + full_frame_length;
-    window->Range(0, old_length).CopyFromVec(
-        downsampled_signal_remainder_.Range(remainder_offset, old_length));
+
+    // BP: this looks to me like something has not been done properly
```
danpovey: BP: I assume you are commenting on the code that has been added just below. 
The code needs to be fixed instead of just commenting that it's not right.
I do agree that it hasn't been properly done-- the part at the beginning of the file needs to be copied if it's partly there.

src/feat/pitch-functions.cc
```diff
@@ -1077,6 +1085,8 @@ void OnlinePitchFeatureImpl::AcceptWaveform(
   for (int32 frame = start_frame; frame < end_frame; frame++) {
     // start_sample is index into the whole wave, not just this part.
     int64 start_sample = static_cast<int64>(frame) * frame_shift;
+    // BP: handles snip_edges == false in a more graceful way
+    if (!opts_.snip_edges) start_sample = static_cast<int64>((frame + 0.5) * frame_shift) - full_frame_length / 2;
```
danpovey: line too long.

src/feat/pitch-functions.cc
```diff
@@ -1077,6 +1085,8 @@ void OnlinePitchFeatureImpl::AcceptWaveform(
   for (int32 frame = start_frame; frame < end_frame; frame++) {
     // start_sample is index into the whole wave, not just this part.
     int64 start_sample = static_cast<int64>(frame) * frame_shift;
+    // BP: handles snip_edges == false in a more graceful way
```
danpovey: Clease don't have comments like this.  Comments should explain what the code does-- not how the code has been changed, which is meaningless without seeing the history.   E.g., more graceful than what?

src/feat/pitch-functions.h
```diff
@@ -198,12 +198,16 @@ struct PitchExtractionOptions {
   }
   /// Returns the window-size in samples, after resampling.  This is the
   /// "basic window size", not the full window size after extending by max-lag.
+  // B.P. fixed a bug here. Even though the result of the operation is supposed 
+  // to be an integer, you still need rounding before casting :-)
+  // NB: 0.001 is impossible to represent exactly on a float or double, so for
+  // this to work without rounding, you would need to divide by 1000 instead.
   int32 NccfWindowSize() const {
-    return static_cast<int32>(resample_freq * 0.001 * frame_length_ms);
+      return static_cast<int32>(round(resample_freq * 0.001f * frame_length_ms));
```
danpovey: you have added unnecessary extra indentation here.
Can you explain why you feel this was a bug?  By default, casting rounds down.  Why is rounding to closest more appropriate in your opinion?  Unless it was causing some kind of crash, best to leave it as it was.

bpotard: This one is quite obscur that probably deserves a detailed explanation. Hope you have the time :-)

For example, consider frame_length_ms=0.9, resample_freq=10000. I would expect the formula above to give me a WindowSize of "9", but it actually gives "8" on my Intel box, and probably something different on ARM CPUs. In practice, you will get potentially unpredictable results whenever one of the operand is not perfectly representable on a floating point (e.g. anything in decimal such as 0.1 or 0.001).

This has for example some impacts on the number of samples generated, that will not match what other featbin programs generate. The WindowShift formula is potentially even more problematic. And yes, it is a bug I ran into, and it was a pain to figure out where it came from.

The issue here is that rounding down with float in general is dangerous, as you can quickly get into some unexpected situations such as the one above. This make things hard to reproduce externally, as many programming languages with the same formula would actually give you 9 (as you would expect).

My reasoning is that I would expect it to give me the "right" integer when the formula is supposed to give you an "exact" integer value, but as it is written, the result heavily depends on the internal floating point representation of frame_length_ms or resample_freq. The result of that formula as it is written is "hardware floating-point" implementation dependent in many cases, which is *never* something desirable, especially if you want kaldi to give reproducible results.

Now, the question is how do you fix it; dividing by 1000 rather than multiplying by 0.001 actually fixes pretty much any situation that will occur on Intel CPUs, but that makes the fix rather obscure. Furthermore, rounding to the closest is safer and probably closer to what most people would expect anyway:
 - it fixes the border cases above
 - it doesn't change the behaviour for "normal" cases
 - it is probably faster that doing a floating point division
 - it is _probably_ - but that could be discussed-  theoretically more useful than rounding down.

And sorry about the indentation! My editor was not properly configured to generate 2 spaces indentations.

bpotard:
Sorry about the commit messages. I will try to be more careful in the future, as it is a bit difficult to modify commit messages afterwards.

Also, is there a way to restrict a pull request to a set of commits? It just appears that every new commits becomes part of the same pull request...

src/feat/pitch-functions.cc
bpotard: You are right. I have now refactored the code a bit and I think the latest commit make things more clear, as the two edge cases are now treated in a similar manner.

src/feat/pitch-functions.cc
bpotard: now fixed

src/feat/pitch-functions.cc
bpotard: Changed the comments, hopefully it made things clearer

src/feat/pitch-functions.cc
bpotard: OK, this is now fixed. Comments removed, line now fits the 80-char limit

src/feat/pitch-functions.cc
bpotard: This bit has now been restored to previous version as border case is now treated separately

src/feat/pitch-functions.cc
```diff
@@ -835,6 +840,21 @@ void OnlinePitchFeatureImpl::ExtractFrame(
...
+  if (sample_index < 0) {
+    // Part of the frame is before the beginning of the signal. This
+    // should only happen if opts_.snip_edges == false, when we are
+    // processing the first few frames of signal. In this case
+    // we pad with zeros.
+    KALDI_ASSERT(opts_.snip_edges == false);
+    int32 new_full_frame_length = sample_index + full_frame_length;
+    KALDI_ASSERT(new_full_frame_length > 0);
+    window->SetZero();
+    SubVector<BaseFloat> sub_window(*window, full_frame_length - new_full_frame_length, new_full_frame_length);
```
danpovey: line too long

src/feat/pitch-functions.cc
```diff
@@ -835,6 +840,21 @@ void OnlinePitchFeatureImpl::ExtractFrame(
...
                                     downsampled_samples_processed_);
 
+  // Treat edge cases first
+  if (sample_index < 0) {
+    // Part of the frame is before the beginning of the signal. This
+    // should only happen if opts_.snip_edges == false, when we are
+    // processing the first few frames of signal. In this case
+    // we pad with zeros.
+    KALDI_ASSERT(opts_.snip_edges == false);
+    int32 new_full_frame_length = sample_index + full_frame_length;
```
danpovey: would be clearer to call this partial_frame_length

src/feat/pitch-functions.cc
```diff
@@ -856,7 +876,7 @@ void OnlinePitchFeatureImpl::ExtractFrame(
   } else {
     // frame is partly in the remainder and partly in the new part.
     int32 remainder_offset = downsampled_signal_remainder_.Dim() + offset;
-    KALDI_ASSERT(remainder_offset >= 0);  // or we didn't keep enough remainder.
+    KALDI_ASSERT(remainder_offset >= 0); // or we didn't keep enough remainder.
```
danpovey: google style guide mandates two spaces before comment.

src/feat/pitch-functions.cc
```diff
@@ -1076,7 +1096,17 @@ void OnlinePitchFeatureImpl::AcceptWaveform(
 
   for (int32 frame = start_frame; frame < end_frame; frame++) {
     // start_sample is index into the whole wave, not just this part.
-    int64 start_sample = static_cast<int64>(frame) * frame_shift;
+    int64 start_sample;
+    if (opts_.snip_edges) {
+      // Usual case offset starts at 0
```
danpovey: comments should be properly punctuated and capitalized sentences (not always true in old code, but we're trying to do it going forward).

src/feat/pitch-functions.cc
```diff
@@ -774,12 +774,17 @@ int32 OnlinePitchFeatureImpl::NumFramesAvailable(
   if (!input_finished_)
     frame_length += nccf_last_lag_;
   if (num_downsampled_samples < frame_length) return 0;
-  else
+  else {
     if (input_finished_ && !snip_edges) {
       return (int32)(num_downsampled_samples * 1.0f / frame_shift + 0.5f);
     }
+    else if (!snip_edges) {
+      return (int32)(
```
danpovey: c-style cast not allowed per google style guide, use static_cast<int32>(...)

src/feat/pitch-functions.cc
```diff
@@ -774,12 +774,17 @@ int32 OnlinePitchFeatureImpl::NumFramesAvailable(
...
+        (num_downsampled_samples - frame_length/2) * 1.0f / frame_shift + 0.5f);
+    }
     else
```
danpovey: braces much match so put braces on else.  Also it should be
if {
   x;
} else {
  y;
}
i.e. no line break before else.  check rest of your code.

src/feat/mel-computations.h
```diff
@@ -53,9 +53,10 @@ struct MelBanksOptions {
   // Enables more exact compatibibility with HTK, for testing purposes.  Affects
   // mel-energy flooring and reproduces a bug in HTK.
   bool htk_mode;
+  bool use_bark_scale;
   explicit MelBanksOptions(int num_bins = 25)
       : num_bins(num_bins), low_freq(20), high_freq(0), vtln_low(100),
-        vtln_high(-500), debug_mel(false), htk_mode(false) {}
+        vtln_high(-500), debug_mel(false), htk_mode(false), use_bark_scale(false) {}
```
danpovey: check line length

src/feat/feature-functions.cc
```diff
@@ -26,30 +26,96 @@
...
+  power_spectrum->Scale(1.0/dim);
+}
+
+template void PowerSpecToRealCeps(VectorBase<float> *power_spectrum);
+template void PowerSpecToRealCeps(VectorBase<double> *power_spectrum);
 
-  // now we have in waveform, first half of complex spectrum
-  // it's stored as [real0, realN/2-1, real1, im1, real2, im2, ...]
+
+template<class Real>
```
danpovey: Best not to use abbreviations in name, per Google style guide (use PowerSpectrumToRealCepstrum).
I wonder if it's necessary to include these functions in Kaldi, or whether you can add it somewhere in Idlak.  Generally we don't like code that's not called anywhere in Kaldi.

src/feat/pitch-functions.h
```diff
@@ -198,12 +198,16 @@ struct PitchExtractionOptions {
   }
   /// Returns the window-size in samples, after resampling.  This is the
   /// "basic window size", not the full window size after extending by max-lag.
+  // B.P. fixed a bug here. Even though the result of the operation is supposed
```
danpovey: best to remove this comment.  comments shouldn't describe changes to code.

src/feat/feature-functions.cc
bpotard: The code that refer to these functions is coming soon! We plan to add a set of tools to do aperiodic excitation extraction, which work in the cepstrum domain. Along with the pitch, these features allow you recreate high quality speech excitation, which is mostly important for TTS, but may be of use for  ASR as well - I am not aware of research using aperiodic excitation as a feature, but I don't see why it would not help.
We could have placed these 2 functions into a separate feature-cepstrum.cc file, but as they are closer to utilities than feature extractors themselves, that probably would not make much sense...

danpovey: What I'd prefer is that you keep this code out of the Kaldi codebase and
put it somewhere inside Idlak-- at least until such time as we have
demonstrated its usefulness for some task that's supported by Kaldi.  The
Kaldi codebase is getting big enough that we never really insert code
that's not used somewhere inside Kaldi (i.e. at the script level).
Dan

bpotard: Hi Dan,
We do believe that ultimately this code could be useful for the kaldi community at large, but we also agree that at this point it would not be of direct use. We will try to keep this code in separate files and only in idlak for now. And I'll try to put the code to google standard.

I'll remove these from the pull request, only keep the fix for pitch (as I originally intended).
When I have time I'll try to run the wsj recipe with some of those "exotic" features, and if we get any interesting results, we'll push again at that point!

Thanks for all the hard work.
Blaise

danpovey: OK, cool.
BTW, after thinking a bit about the fix regarding the pitch, I think
dividing by 1000.0 instead of multiplying by 0.001 is the better solution
[and should of course be done everywhere we multiply by 0.001, but I can do
that].  By calling 'round', we just shift the nondeterminism to a different
place, i.e. the n + 0.5 boundaries.
Dan

src/feat/pitch-functions.cc
```diff
@@ -774,12 +774,20 @@ int32 OnlinePitchFeatureImpl::NumFramesAvailable(
   if (!input_finished_)
     frame_length += nccf_last_lag_;
   if (num_downsampled_samples < frame_length) return 0;
-  else
-    if (input_finished_ && !snip_edges) {
-      return (int32)(num_downsampled_samples * 1.0f / frame_shift + 0.5f);
+  else {
```
danpovey: please make sure the if-else braces are matched, and any braces with else are on the same line, e.g.
if (foo) {
  x;
} else {
  y;
}
you can use cpplint.py (which we recently moved into misc/maintenance/) to check this type of thing.

src/feat/feature-window.cc
```diff
@@ -105,6 +105,7 @@ FeatureWindowFunction::FeatureWindowFunction(const FrameExtractionOptions &opts)
   int32 frame_length = opts.WindowSize();
   KALDI_ASSERT(frame_length > 0);
   window.Resize(frame_length);
+  BaseFloat a = M_2PI / (frame_length - 1);
```
danpovey: might  be more accurate to make this a double.

src/feat/pitch-functions.cc
bpotard: Hi Dan,

You are right, it actually helps a lot to use cpplint; although since most of the existing code is "dirty" anyway, it is not always easy to figure out what to clean.

Anyway, I passed cpplint on all the code I modified, and now all the modified parts should hopefully be in an acceptable state. Thanks!

bpotard:
Hi Xingyu,
The pitch tracker in kaldi returns a continuous pitch + voicing probability, it never returns 0 for the pitch. That actually makes pitch modelling much easier as there is no U/V boundary effect!
There is some zero-padding during the calculation of the pitch, but only in the input wave data. 
Blaise

src/feat/pitch-functions.cc
```diff
@@ -835,19 +844,35 @@ void OnlinePitchFeatureImpl::ExtractFrame(
   int32 offset = static_cast<int32>(sample_index -
                                     downsampled_samples_processed_);
 
+  // Treat edge cases first
+  if (sample_index < 0) {
+    // Part of the frame is before the beginning of the signal. This
+    // should only happen if opts_.snip_edges == false, when we are
+    // processing the first few frames of signal. In this case
+    // we pad with zeros.
+    KALDI_ASSERT(opts_.snip_edges == false);
```
danpovey: I think you may not be handling the case when there is a very short signal, less than a frame, and the user terminates the input.  So consider that 'window' may be too short (on the right side) for a full frame.

danpovey: ... or did we resolve this issue already?  Let me know if so.

bpotard: Normally the case where we have a pathological number of samples should be handled correctly:
 1. It will first go through the "sample_index < 0" case, which only fills the beginning of the window with zeros, then call the Extract_frame function again with a sample_index of 0 and a reduced window.
 2. On the second call it will go through the offset + full_frame_length > downsampled_wave_part.Dim() branch, which will fill the end of the window with 0, then call the function again on a sub-window
 3. On the 3rd call It will go in the general case, filling up what's left of the window with only valid data.

The window on the 3rd call should have exactly the same length as the number of samples in the input; although with an insufficient number of samples I am not sure what the tracker will output.

I can add some test for pathological cases if you think that could be useful!

danpovey: OK, that sounds fine.
So I guess this is ready to merge.  You're sure this doesn't change results
with snip-edges=true (the default) ?

bpotard: I had tested before my initial commit on some tests file that the output were identical to the current git with snip-edges=true; I have no retested after the recent changes but I would not expect any change in behaviour.

danpovey: Can you please re-test?  This is a change that would break a lot of setups
in a way that would be hard to roll back, if it was wrong.

bpotard: Hi Dan,

I have just retested on 10000 wav files of various lengths with the default settings apart from the snip-edges option.
 - snip-edges=false modifies the output of every single file, but the number of output frames is identical
 - snip-edges=true does not modify the output of any file.

So I would say the change should be safe, unless some of the recipes were using snip-edges=false already; but even then I would suspect it would improve their results.

danpovey:
OK, thanks.  I think it's just about ready to merge.
But can you please use rebase to squash your commits into one?
git rebase -i HEAD~11
and change all but the first "pick" to "s" or "squash".
Dan

danpovey:
Thanks- will merge when automatic checks are done.

bpotard:
Ok, great!
I have done the rebase, and pushed the change with -f, so that should leave a single commit and a cleaner history in kaldi.

PR #829, opened 2016-06-03
Update validate_lang.pl

jarfo:
%wdisambig_words_hash has to be a global variable since it is used in a different block (Line 732)


danpovey:
OK... I did not realize that the scope of "my" in perl is the code-block (as opposed to function).

PR #836, opened 2016-06-10
For compatbility with HTK USEPOWER option have added --use-power opti…

edwinferretjp:
…on (default[=true] is same as current code) to filterbank generation code. To use add --use-power=[false|true] to feature config file.

### DIFF
--- a/src/feat/feature-fbank.h
+++ b/src/feat/feature-fbank.h
@@ -46,6 +46,7 @@ struct FbankOptions {
   bool raw_energy;  // If true, compute energy before preemphasis and windowing
   bool htk_compat;  // If true, put energy last (if using energy)
   bool use_log_fbank;  // if true (default), produce log-filterbank, else linear
+  bool use_power;  // if true (default), use power in filterbank analysis, else magnitude.
 
   FbankOptions(): mel_opts(23),
                  // defaults the #mel-banks to 23 for the FBANK computations.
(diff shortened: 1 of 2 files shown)


src/feat/feature-fbank.h
```diff
@@ -55,7 +56,8 @@ struct FbankOptions {
                  energy_floor(0.0),  // not in log scale: a small value e.g. 1.0e-10
                  raw_energy(true),
                  htk_compat(false),
-                 use_log_fbank(true) {}
+    use_log_fbank(true),
```
danpovey: there is odd indentation here.  Please make sure you are writing spaces not tabs.

src/feat/feature-fbank.h
```diff
@@ -71,6 +73,8 @@ struct FbankOptions {
                    "to change other parameters).");
     opts->Register("use-log-fbank", &use_log_fbank,
                    "If true, produce log-filterbank, else produce linear.");
+    opts->Register("use-power", &use_power,
+		   "If true, use power in filterbank analysis, else magnitude.");
```
danpovey: please check that this line doesn't  exceed 80 characters.  And replace tabs with spaces...

galv:
I see you pushed a new commit. When you make a fix and you want it to be seen, you need to manually comment here. We aren't notified of commits by themselves.

danpovey:
Thanks.. merging.

edwinferretjp:
Thanks. Noted for the future.

PR #840, opened 2016-06-12
thchs30-add-real-spkinfo

naxingyu:
Changed data prep script to add valid utt2spk and spk2utt files. Results on SAT are improved.


naxingyu:
@danpovey This PR causes failure at the DAE stage, because the local/dae/run_dae.sh requires per-utterance CMVN applied on both inputs and outputs. I can submit a temporary PR to fix the failure by using per-speaker CMVN. But when I tested it, the performance is significantly worse than the original script. Should I make that PR? Or switch to per-utterance CMVN mode in the run_dae.sh script?

danpovey:
I don't really follow this setup very closely.  Do what you think is best
and let me know-- but I don't want to merge something that will break part
of the existing recipe.

PR #850, opened 2016-06-16
Fix duplicate option registers

naxingyu:


src/nnet3bin/nnet3-align-compiled.cc
```diff
@@ -69,8 +69,6 @@ int main(int argc, char *argv[]) {
                 "yes|no|optional|wait, only has effect if compiled with CUDA");
     po.Register("transition-scale", &transition_scale,
                 "Transition-probability scale [relative to acoustics]");
-    po.Register("acoustic-scale", &acoustic_scale,
```
danpovey: please remove this variable also, it seems to be unused.

naxingyu: Right. Done and cleaned.


在 2016/6/17 1:53, Daniel Povey 写道:

danpovey:
Thanks, will merge when checks complete.

Issue #837, opened 2016-06-10
 kaldi/egs/wsj/s5/steps/cleanup/make_segmentation_graph.sh  & make_utterance_graph.sh


vince62s:
In these 2 scripts, the section below does not work (at least for me on Ubuntu)
what's in bold needs to be removed.

if [ $ngram_order -gt 1 ]; then
  **ngram_count=`which ngram-count`;**
  **if [ -z $ngram_count ]; then**
    if uname -a | grep 64 >/dev/null; then # some kind of 64 bit...
      sdir=`pwd`/../../../tools/srilm/bin/i686-m64
    else
      sdir=`pwd`/../../../tools/srilm/bin/i686
    fi
    if [ -f $sdir/ngram-count ]; then
      echo Using SRILM tools from $sdir
      export PATH=$PATH:$sdir
    else
      echo You appear to not have SRILM tools installed, either on your path,
      echo or installed in $sdir.  See tools/install_srilm.sh for installation
      echo instructions.
      exit 1
    fi
  **fi**
fi

jtrmal:
Removing it does not really make any sense, if you understand what it's supposed to do.
When you say "does not work", what that does mean?

vince62s:
I know what this is supposed to do but the line ngram_count=`which ngram-count`;
terminates and exit the script.
might be caused by the set -e at the beginning of the script which make the which exit with non zero status.

jtrmal:
Yes, that would be the cause for the termination.
changing the line to

ngram_count=`which ngram-count` || true

should fix that -- can you please test? It would be great if you could
create PR (if it works).
Thanks!
y.

vince62s:
fixed

vince62s:
actually I got another issue with make_utterance_graph.sh
log gives me this:
steps/cleanup/make_utterance_graph.sh --cleanup true --tscale 1.0 --loopscale 0.1 --ngram-order 2 --srilm-options -wbdiscount exp/tri3_mmi_b0.1/graph_source_split/split1/1/text data/lang exp/tri3_mmi_b0.1 exp/tri3_mmi_b0.1/graph_source_split/split1/1
steps/cleanup/make_utterance_graph.sh: processing utterance utt1.
awk: program limit exceeded: maximum number of fields size=32767
	FILENAME="-" FNR=1 NR=1
# Accounting: time=0 threads=1
# Ended (code 2) at Tue Jun 14 21:20:04 CEST 2016, elapsed time 0 seconds

[...]

PR #851, opened 2016-06-19
fix the order >=2 process of bigger files

vince62s:
related to issue https://github.com/kaldi-asr/kaldi/issues/837
this works for me with big audio file and one long utterance.

### DIFF
--- a/egs/wsj/s5/steps/cleanup/make_utterance_graph.sh
+++ b/egs/wsj/s5/steps/cleanup/make_utterance_graph.sh
@@ -89,7 +89,10 @@ P=`tree-info --print-args=false $model_dir/tree |\
 if [ -f $graph_dir/sub_graphs/HCLG.fsts.scp ]; then
   rm $graph_dir/sub_graphs/HCLG.fsts.scp
 fi
-while read line; do
+
+cat $text | utils/sym2int.pl --map-oov $oov -f 2- $lang/words.txt | \
+ utils/int2sym.pl -f 2- $lang/words.txt | \
+ while read line; do
   uttid=`echo $line | cut -d ' ' -f 1`
   words=`echo $line | cut -d ' ' -f 2-`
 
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/steps/cleanup/make_utterance_graph.sh
```diff
@@ -105,12 +105,13 @@ while read line; do
...
       fstarcsort --sort_type=ilabel > $wdir/G.fst || exit 1;
   else
-    echo $words | awk -v voc=$lang/words.txt -v oov="$oov_txt" '
-      BEGIN{ while((getline<voc)>0) { invoc[$1]=1; } } {
-      for (x=1;x<=NF;x++) {
-      if (invoc[$x]) { printf("%s ", $x); } else { printf("%s ", oov); } }
-      printf("\n"); }' > $wdir/text
-    ngram-count -text $wdir/text -order $ngram_order "$srilm_options" -lm - |\
+     echo $words > $wdir/text
+     split -b 200000 $wdir/text $wdir/xxx
```
danpovey: I don't think this is very ideal because it may split in the middle of a word, leading to some words in the original text not being present in the LM that's produced.  How about this:
```
echo $words | perl -ane '@A = split; for ($n=0;$n<@A;$n++) { print "$A[$n] "; if(($n+1)%2 == 0 || $n+1==@A) {print "\n";} }' > $wdir/text

```

danpovey: ... also, I see that you have moved the dealing-with-OOVs into the ngram-count process.
This would be OK, except it's extremely inefficient to deal with the whole word-list on each utterance each time, if the utterances are short.  The original script had this problem too, as the wordlist was read by awk- I did not check it sufficiently when it was originally committed.  
The while loop is done <$text.  A much-more-efficient way to do this would be to pipe into the while loop, a command involving sym2int.pl (with the appropriate options to map unknown words) and then into int2sym.pl.

danpovey:
I'd still rather do it properly-- if it's an easy fix, why not do it?  There are edge cases that could cause a crash in your setup, e.g. splitting inside a UTF character; generating banned words or disambiguation symbols; etc.

I don't think you understood what I was saying, regarding the second, point.  I'm talking about the loop in the outer-level, enclosing while loop, that's done `<$text.`  I'm talking about fixing the OOV problem at that level, by piping the text through int2sym.pl and then sym2int.pl, with appropriate options, to get rid of the OOVs there.

danpovey:
Looks good, but can you please rebase your commits so there is just one commit?
I'm assuming you've tested this and it works.

danpovey:
... it's something like
git rebase -i HEAD~4
and change the 4 commits from 'pick pick pick pick' to something like "pick s s s" where s means squash.

vince62s:
Just rebased, but I think I need to push again ?
yes tested it and works. I had to change your perl line though because it adds a linefeed every 2 words which srilm does not like of course. So I changed to a reasonable cut-off value 30000 words.

danpovey:
Yes you may have to push; you will need the --force option as it's not a fast-forward push.

Issue #854, opened 2016-06-20
BLSTM layer in steps/nnet3/components.py


vijayaditya:
It would be convenient to have an AddBlstmLayer() function similar to AddLstmLayer().
BLSTMs are right not supported in steps/nnet3/lstm/make_configs.py, which supports a linear LSTM+FullyConnectedLayer network. However for users who desire to add this node to more complicated graphs a function to add BLSTM layers would be very useful.

@freewym  This would just require a minor modification of code as you would move code from steps/nnet3/lstm/make_configs.py to steps/nnet3/components.py. Please let me know if you can pick  this up. 

priority : low
complexity : low

freewym:
OK. Will do it later.

PR #856, opened 2016-06-21
Add AddBLstmLayer function to components.py

freewym:
Solve issue #854 
Also removed the code in make_configs.py dealing with the invalid syntax of the form "Append(Append(xx,yy),zz)", because now it can be automatically corrected as "Append(xx,yy,zz)" when the config file is being parsed.


danpovey:
Vijay can merge this if he thinks it's OK.

vijayaditya:
Looks fine. 
@freewym I am assuming you tested it as you committed a minor fix. So I am merging.

PR #857, opened 2016-06-21
Add option to skip nnet3 decode diagnostics

guoguo12:


danpovey:
Why is this helpful?  Are the diagnostics slow?

guoguo12:
Sorry, I should have given some kind of rationale.

Basically, I've created a Docker container that contains the minimal Kaldi installation required to decode nnet3. This container doesn't have Python installed, and the diagnostics won't work without Python. There are a few ways around this, but I think the cleanest is to make the diagnostics skippable, since the diagnostics aren't crucial to decoding.

danpovey:
OK, merging since it seems harmless.

guoguo12:
Thanks!

PR #862, opened 2016-06-24
For compatibility with MS Visual Studio.

tal1974:
In the case of "1.0 / 0.0" compiler generates an error:
1>  kaldi-vector.cc
1>..\..\..\src\matrix\kaldi-vector.cc(523): error C2124: divide or mod by zero

If the constant 0.0 is replaced by a variable, then there is no error.

### DIFF
--- a/src/matrix/kaldi-vector.cc
+++ b/src/matrix/kaldi-vector.cc
@@ -508,6 +508,7 @@ template<typename Real>
 Real VectorBase<Real>::Norm(Real p) const {
   KALDI_ASSERT(p >= 0.0);
   Real sum = 0.0;
+  const Real zero = 0.0;
   if (p == 0.0) {
     for (MatrixIndexT i = 0; i < dim_; i++)
       if (data_[i] != 0.0) sum += 1.0;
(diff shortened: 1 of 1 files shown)


jtrmal:
I believe the "nice" way is to compare the number with
std::numeric_limits<Real>::infinity()
which is more self-explanatory.
y.

danpovey:
Yes- it looks like the intention in that code was to compare with infinity
(although I'm wondering whether that division 1.0 / 0.0 actually generates
NaN in reality).  So please modify the PR to use Yenda's solution.

tal1974:
I absolutely agree with you. Just could not find this variable by myself.

danpovey:
Can you please rebase so there is just one commit?
git rebase HEAD~2
and change the second "pick" to "squash".
you'll have to push with --force to your github repo.

danpovey:
No need to close the PR, you just have to modify the branch it points to.
 (you need to push --force since it's a non-fast-forward update).

tal1974:
Sorry. I don't use git on my PC. I edit file on github.com. Now everything is done as before. Apparently, something went wrong.

tal1974:
It seems to be fixed.

danpovey:
There are still 3 commits :-(

PR #865, opened 2016-06-26
Standard way to represent inf

kangshiyin:
Fix #864 and #862 on CUDA code. All GPU tests passed.

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -25,6 +25,8 @@
 // In this file is the CUDA code of the CUDA kernels, plus the ANSI-C wrappers
 
 #include <cfloat>
+#include <limits>
+#include <math_constants.h>
 #include "cudamatrix/cu-kernels-ansi.h"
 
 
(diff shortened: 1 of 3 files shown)


src/cudamatrix/cu-kernels.cu
```diff
@@ -2743,7 +2745,7 @@ void cudaF_group_spec_pnorm(dim3 Gr, dim3 Bl, float* y, const float* x,
   } else if (power == float(2)) {
     _group_transform_reduce<<<Gr, Bl>>>(y, x, d, src_stride, group_size,
         TransReduceOp<L2NORM, float>());
-  } else if (power == float(1.0 / 0.0)) {
+  } else if (power == std::numeric_limits<float>::infinity()) {
```
danpovey: Is there a reason you using numeric_limits here and CUDART_INF above?

kangshiyin:
infinity() is a host function that can not compile in device code, so they each have a definition of inf.  

https://devtalk.nvidia.com/default/topic/456829/how-to-assign-infinity-to-variables-in-cuda-code-/

PR #874, opened 2016-06-29
Online feature fbank fix

wonkyuml:
- fix a bug (unknown feature) which occurs when “—endpoint=true” with
use of fbank feature


danpovey:
This is OK but can you please rebase against the current master so that the merges don't appear in your commit?
Something else it would be nice if you can fix is, whenever we do '* 1.0e-03' in Kaldi code, we should be doing ' / 1000.0'.  This avoids a non-determinacy that can otherwise happen due to 1/10 not being exactly representable in IEEE floating point, which can cause problems on ARM processors.  
Dan

PR #878, opened 2016-07-03
cuedrnnlm integration patch

rickychanhoyin:


danpovey:
We generally don't like to include lots of external code as part of the Kaldi repository. 
I think when Hainan did this he was able to extract a small amount of code from cued-rnnlm-- just the amount that was necessary to get this to work.

rickychanhoyin:
This one in extracted and modified from 
http://mi.eng.cam.ac.uk/projects/cued-rnnlm/src.evaloncpu.tar.gz
It supports current cued-rnnlm v0.1 (and also support former version of it - what I notice is that cued-rnnlm v0.1 output model format has changed a bit compare to several months ago)

rickychanhoyin:
The functions printPPLInfo , calppl , calnbest in cued-rnnlm-lib.cc can be removed (and also macro.h which is a dummy file, and cued-main.cc if we only use it as lattice rescoring library)

PR #881, opened 2016-07-03
nnet3 lstm: hidden_dim was omitted from MakeConfigs()

jfainberg:
hidden_dim was read into make_configs.py but not passed on to MakeConfigs(). 

Would only cause an error if num hidden layers > num lstm layers, when adding AddAffRelNormLayer which takes hidden_dim as argument. The default settings in train.sh has both equal.


danpovey:
@vijayaditya, please check and merge.

vijayaditya:
@jfainberg Thanks for the fix.

@danpovey None of the current (B)LSTM recipes use FF-DNN layers after LSTMs so the current recipes are unaffected by this bug.

PR #887, opened 2016-07-08
linux_x86_64_mkl.mk now show respect to --static-fst=yes

scinart:
linux_x86_64_mkl.mk is hard-coded to link openfst dynamically.

It should use ``$(OPENFSTLDFLAGS)`` and ``$(OPENFSTLIBS)`` instead, just like in `linux_atlas.mk`, `linux_clapack.mk` and `linux_openblas.mk`


danpovey:
thanks- will merge when checks finish.

Issue #891, opened 2016-07-10
Reg. random seed control for nnet3 experiments


vijayaditya:
@freewym You recently wrote some code to change the random seed in nnet3 scripts (for model initialization and egs shuffling) for PR #803. Could you please submit a PR to have this as a feature in `steps/nnet3/train_rnn.py`, `steps/nnet3/train_dnn.py` and `steps/nnet3/chain/train.py`.

freewym:
ok

On Sunday, July 10, 2016, Vijayaditya Peddinti <notifications@github.com>
wrote:

> @freewym <https://github.com/freewym> You recently wrote some code to
> change the random seed in nnet3 scripts (for model initialization and egs
> shuffling) for PR #803 <https://github.com/kaldi-asr/kaldi/pull/803>.
> Could you please submit a PR to have this as a feature in
> steps/nnet3/train_rnn.py, steps/nnet3/train_dnn.py and
> steps/nnet3/chain/train.py.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/kaldi-asr/kaldi/issues/891>, or mute the thread
> <https://github.com/notifications/unsubscribe/ADWAkimw6vLiMzp_aFw-7wgxvZeihxDtks5qUTk3gaJpZM4JI40u>
> .
>


-- 
Sent from my iPhone

PR #892, opened 2016-07-10
add an option to set random seed for model initialization and egs shu…

freewym:
…ffling in training. Ref #891

### DIFF
--- a/egs/wsj/s5/steps/nnet3/chain/train.py
+++ b/egs/wsj/s5/steps/nnet3/chain/train.py
@@ -131,7 +131,7 @@ def GetArgs():
                         default = 0,
                         help="Sets the random seed for model initialization and egs shuffling. "
                         "Warning: This random seed does not control all aspects of this experiment. "
-                        "There might be other random seeds used other stages of the experiment "
+                        "There might be other random seeds used in other stages of the experiment "
                         "like data preparation (e.g. volume perturbation).")
     parser.add_argument("--trainer.num-epochs", type=int, dest='num_epochs',
                         default = 10,
--- a/egs/wsj/s5/steps/nnet3/chain/train.py
+++ b/egs/wsj/s5/steps/nnet3/chain/train.py
@@ -596,7 +596,7 @@ def Train(args, run_opts):
     else:
         egs_dir = args.egs_dir
 
-    [egs_left_context, egs_right_context, frames_per_eg, num_archives] = train_lib.VerifyEgsDir(egs_dir, feat_dim, ivector_dim, left_context, right_context, args.srand)
+    [egs_left_context, egs_right_context, frames_per_eg, num_archives] = train_lib.VerifyEgsDir(egs_dir, feat_dim, ivector_dim, left_context, right_context)
     assert(args.chunk_width == frames_per_eg)
     num_archives_expanded = num_archives * args.frame_subsampling_factor
 
--- a/egs/wsj/s5/steps/nnet3/chain/train.py
+++ b/egs/wsj/s5/steps/nnet3/chain/train.py
@@ -416,6 +416,19 @@ def TrainOneIteration(dir, iter, srand, egs_dir,
     # Use the egs dir from the previous iteration for the diagnostics
     logger.info("Training neural net (pass {0})".format(iter))
 
+    # check if different iterations use the same random seed
+    if os.path.exists('{0}/srand'.format(dir)):
+        try:
+            saved_srand = int(open('{0}/srand'.format(dir), 'r').readline().strip())
+        except IOError, ValueError:
+            raise Exception('Exception while reading the random seed for training')
+        if srand != saved_srand:
+            logger.warning("The random seed provided to this iteration (srand={0}) is different from the one saved last time (srand={1}). Using srand={0}.".format(srand, saved_srand))
+    else: 
+        f = open('{0}/srand'.format(dir), 'w')                              
+        f.write(str(srand))                                                      
+        f.close()
+
     chain_lib.ComputeTrainCvProbabilities(dir, iter, egs_dir,
             l2_regularize, xent_regularize, leaky_hmm_coefficient, run_opts)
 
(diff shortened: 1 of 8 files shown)


danpovey:
Vijay, merge this if you think it's OK.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -127,6 +127,9 @@ def GetArgs():
 
 
     # trainer options
+    parser.add_argument("--trainer.srand", type=int, dest='srand',
+                        default = 0,
+                        help="Sets random seed for model initialization and egs shuffling")
```
vijayaditya: ```Warning: This random seed does not control all aspects of this experiment. There might be other random seeds used other stages of the experiment like data preparation (e.g. volume perturbation).```

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -362,11 +365,11 @@ def TrainNewModels(dir, iter, num_jobs, num_archives_processed, num_archives,
   --print-interval=10 --momentum={momentum} \
   --max-param-change={max_param_change} \
    "{raw_model}" {dir}/den.fst \
-  "ark,bg:nnet3-chain-copy-egs --truncate-deriv-weights={trunc_deriv} --frame-shift={fr_shft} ark:{egs_dir}/cegs.{archive_index}.ark ark:- | nnet3-chain-shuffle-egs --buffer-size={shuffle_buffer_size} --srand={iter} ark:- ark:-| nnet3-chain-merge-egs --minibatch-size={num_chunk_per_minibatch} ark:- ark:- |" \
+  "ark,bg:nnet3-chain-copy-egs --truncate-deriv-weights={trunc_deriv} --frame-shift={fr_shft} ark:{egs_dir}/cegs.{archive_index}.ark ark:- | nnet3-chain-shuffle-egs --buffer-size={shuffle_buffer_size} --srand={srand} ark:- ark:-| nnet3-chain-merge-egs --minibatch-size={num_chunk_per_minibatch} ark:- ark:- |" \
```
vijayaditya: Let the seed be different for each iteration. You could use `iter+srand`.

vijayaditya:
You seem to have forgotten setting the random seed for egs shuffling jobs.

danpovey:
Obviously it will still need to be the case that the egs-shuffling seeds
differs from iter to iter (e.g. add to it).

egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -365,11 +368,11 @@ def TrainNewModels(dir, iter, num_jobs, num_archives_processed, num_archives,
   --print-interval=10 --momentum={momentum} \
   --max-param-change={max_param_change} \
   --optimization.min-deriv-time={min_deriv_time} "{raw_model}" \
-  "ark,bg:nnet3-copy-egs {context_opts} ark:{egs_dir}/egs.{archive_index}.ark ark:- | nnet3-shuffle-egs --buffer-size={shuffle_buffer_size} --srand={iter} ark:- ark:-| nnet3-merge-egs --minibatch-size={num_chunk_per_minibatch} --measure-output-frames=false --discard-partial-minibatches=true ark:- ark:- |" \
+  "ark,bg:nnet3-copy-egs {context_opts} ark:{egs_dir}/egs.{archive_index}.ark ark:- | nnet3-shuffle-egs --buffer-size={shuffle_buffer_size} --srand={srand} ark:- ark:-| nnet3-merge-egs --minibatch-size={num_chunk_per_minibatch} --measure-output-frames=false --discard-partial-minibatches=true ark:- ark:- |" \
```
freewym: rand seed  for shuffling is set here.

vijayaditya: I am talking about the seed used in steps/nnet3/get_egs.sh which is used while dumping the egs.

freewym: Fixed/

egs/wsj/s5/steps/nnet3/get_egs.sh
```diff
@@ -345,7 +346,7 @@ if [ $stage -le 5 ]; then
 
   if [ $archives_multiple == 1 ]; then # normal case.
     $cmd --max-jobs-run $nj JOB=1:$num_archives_intermediate $dir/log/shuffle.JOB.log \
-      nnet3-shuffle-egs --srand=JOB "ark:cat $egs_list|" ark:$dir/egs.JOB.ark  || exit 1;
+      nnet3-shuffle-egs --srand=$[JOB+$srand] "ark:cat $egs_list|" ark:$dir/egs.JOB.ark  || exit 1;
```
vijayaditya: I don't think command will work. bash would replace $[JOB+$srand] with the value of srand.

vijayaditya:
don't forget changes to `steps/nnet3/chain/get_egs.sh`

danpovey:
It *might* work if the first $ is escaped, like
 \$[JOB+$srand]

Dan

freewym:
Yes. escape of $ works.

vijayaditya:
Remember to store the random seed used for an experiment in the experiment directory (e.g. `exp/nnet3/tdnn`. You could also add checks 

1. to see if the random seed used for generating the egs is the same as the one used for model initialization and egs shuffling in the training script
2. to verify that the random seed being used when restarting the training from an intermediate stage is the same as the one stored in the experiment directory.

When either of these checks fail please print a warning and report which random seed is being used.

@danpovey  I was wondering if we should have a random seed associated with an experiment (which includes data-preparation, egs dumping and training). We could store this `random_seed` file in the s5/config directory. We could ensure all scripts in `local/` access this file and supply it to the appropriate `steps/*` scripts via command-line. It is a bit tricky to ensure that the same random seed is used consistently as we usually reuse feature files or egs across experiments. Any comments ?

freewym:
Added the random seed checks.

danpovey:
Vijay, regarding the random seed, I don't think storing it in s5/config is a good idea because typically the only reason you'd want to do this is if you were doing multiple runs with the same experiment, and they'd all be in the same s5 directory.  I think it's best to put it inside a script and pass it as appropriate to scripts that that script calls.  And I don't think it's necessary to go overboard with the random seed and use it for things feature generation.  This will make next to no difference and just provides a lot of clutter.  There are other things that can affect the results, including the hardware it's run on, and non-reproducible differences caused by non-determinism in CUBLAS.  In any case, it will probably be rare that we'll want to run with different random seeds.

Actually I would have been happy with just setting the random seed in the neural net initialization in nnet-init.  I suspect you get 90% of the difference from that.

PR #896, opened 2016-07-14
LID using DNN

s-mousmita:
Adding lre07/v2 script that uses DNN to initialize the UBM for language ID I-vector extractor.


danpovey:
@david-ryan-snyder, please review this.

danpovey:
@david-ryan-snyder, just a reminder in case you forgot about this (but I imagine you are busy).

david-ryan-snyder:
@danpovey , thanks for the reminder. I'll try to take a look at this in the next couple of days.

david-ryan-snyder:
@s-mousmita, I'm just now looking into this. Sorry for the delays.

I don't understand why there are changes to the AMI and ASPIRE recipes in the pull request. Is this something you talked to @danpovey about? As far as I'm aware, the changes should be isolated to egs/lre07/v2 . This makes me think something went wrong in your branch. Can you see if you can fix this?

danpovey:
Also you seem to have checked in quite a few large files in to local/.   These should be some kind of download, we don't like to bulk up the repository by including large files in the example scripts.

You can put it on openslr.org, as long as the data is free.

david-ryan-snyder:
@s-mousmita, could you check if we really need those files at all (the big ones in local/)? My suspicion is that they were generated during evaluation or some plotting scripts, and were accidentally included in the commit.

danpovey:
BTW, automatically generated files should never go in local/; it's for
scripts only.

s-mousmita:
@danpovey @david-ryan-snyder    Sorry, it was some mistake I did while pushing the changes. 
Anyway I have created a new branch and sending a new pull request. Please check. 

I have removed those large files from local. Actually those were script generated results which I forgot to remove. Sorry for that again.

Thanks
Mousmita

PR #909, opened 2016-07-21
add max_arcs and max_loop for DeterminizeLatticePhonePruned

rickychanhoyin:


jtrmal:
Yes, thanks... It wasn't completely necessary to create a new PR -- google
for "git rebase", but this is ok as well.
y.

rickychanhoyin:
ok, thanks for telling.

src/lat/determinize-lattice-pruned.h
```diff
@@ -155,11 +155,14 @@ struct DeterminizeLatticePhonePrunedOptions {
...
   bool minimize;
+  // extra parameters for speed 
+  int max_loop;
+  int max_arcs;
   DeterminizeLatticePhonePrunedOptions(): delta(kDelta),
                                           max_mem(50000000),
                                           phone_determinize(true),
                                           word_determinize(true),
-                                          minimize(false) {}
+                                          minimize(false) , max_loop(-1), max_arcs(-1) {}
```
danpovey: these should be on separate lines to match the other options.

But I don't fully understand why you need to add these options, since  you're not registering them.

danpovey: ... and let me know what problem you are trying to fix with this PR.

rickychanhoyin: I use it somewhere else in my recognizer which directly set the max_arcs and max_loops on the DeterminizeLatticePhonePrunedOptions struct . this improve the speed on DeterminizeLatticePhonePrunedWrapper

danpovey: In general I prefer to just set the max-mem in order to limit computation.
It's simpler and has much the same effect. That's why I didn't expose those
variables.  See if you can get the same effect from max_mem.
In any case I'd rather not merge this, as it has no effect in any Kaldi
program-- if you use git for your own repository you can keep the changes
locally and any merging will happen automatically.

PR #916, opened 2016-07-25
nnet3: Added minibatch size option for jobs which compute diagnostics…

vijayaditya:
…. This is essential as BLSTMs were consuming a lot of memory with the default minibatch sizes


egs/wsj/s5/steps/nnet3/train_rnn.py
```diff
@@ -182,6 +182,11 @@ def GetArgs():
     parser.add_argument("--trainer.optimization.shrink-threshold", type=float, dest='shrink_threshold',
                         default = 0.15,
                         help="If the derivative averages are below this threshold we scale the parameter matrices with the shrink-value. It is less than 0.25 for sigmoid non-linearities.")
+    parser.add_argument("--trainer.optimization.cv-minibatch-size", type=int, dest='cv_minibatch_size',
+            default = 256,
+            help="Size of the minibatch to be used in diagnostic jobs (use smaller value for BLSTMs to avoid mem-killer.pl")
```
danpovey: Let's say "to control memory usage" here, as mem_killer.pl is specific to the CLSP grid.

vijayaditya:
@danpovey I modified the help message.

Issue #920, opened 2016-07-25
Back-compatibility issue with 6d2fd23


vijayaditya:
This is less of an issue and more of a notice to people who are trying to use trained (B)LSTM models. Commit 6d2fd23dff92458cbf3e18a0b3e3dc729123330e breaks back-compatibility of ClipGradientComponent.

danpovey:
Wait - this doesn't mean it can't now read old models, does it?  We never
do that.

vijayaditya:
Old models cannot be read with the current code.

```
ERROR (nnet3-am-copy:ExpectToken():io-funcs.cc:201) Expected token "<SelfRepairClippedProportionThreshold>", got instead "<NumElementsClipped>".

[ Stack-Trace: ]
nnet3-am-copy() [0x74b5a5]
kaldi::MessageLogger::HandleMessage(kaldi::LogMessageEnvelope const&, char const*)
kaldi::MessageLogger::~MessageLogger()
kaldi::ExpectToken(std::istream&, bool, char const*)
kaldi::nnet3::ClipGradientComponent::Read(std::istream&, bool)
kaldi::nnet3::Component::ReadNew(std::istream&, bool)
kaldi::nnet3::Nnet::Read(std::istream&, bool)
kaldi::nnet3::AmNnetSimple::Read(std::istream&, bool)
main
__libc_start_main
nnet3-am-copy() [0x560999]
```

danpovey:
OK, I commented separately on the PR saying how to fix it.  There always
needs to be back-compatibility code for reading models.

PR #921, opened 2016-07-25
changes in ClipGradientComponent::Read() to ensure back-compatibility…

freewym:
… when reading old models (resolve issue #920)


danpovey:
Thanks.  It would be great if you could check that this code can read the older models... perhaps Vijay can point you to one that he had a problem with.

vijayaditya:
An old model is available here
/export/b13/vpeddinti/lsr/egs/aspire/s5_olddataaugment/exp/nnet3/lstm_bidirectional/1300.mdl

--Vijay

freewym:
It passed the test with the command:

nnet3-am-copy --raw=true --learning-rate=0.001 /export/b13/vpeddinti/lsr/egs/aspire/s5_olddataaugment/exp/nnet3/lstm_bidirectional/1300.mdl temp.raw

It also passed the test on the "new" model.

PR #929, opened 2016-07-28
Adding the underlying scripts for my refactored version of Vimal's da…

danpovey:
…ta-cleanup-and-resegmentation.  No example scrips yet-- will come soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,10 @@
 *.obj
 *.pyc
 
+# profiling files and core files
+*.nvprof
+core
+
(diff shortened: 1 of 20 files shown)


danpovey:
Doing this as a pull request since it's a big change.
It shouldn't break anything-- mostly new scripts.
However I have rewritten steps/cleanup/combine_short_segments.py and moved it to utils/data/combine_short_segments.sh.  This should not really affect anything but the output likely won't be identical to the old script.

PR #937, opened 2016-08-02
WIP : Added support for cost functions which do not require supervision

vijayaditya:


danpovey:
What kinds of things do you hope to be able to implement with this change?

It seems to me like it is doubly computing stuff-- the original objective functions, plus your no-supervision objective functions.

Also, this wouldn't work with the way the computations are compiled.

If you don't specify a supervision object for an output, which contains a list of Indexes to be evaluated, then the output node will never be evaluated for any indexes.

If you let me know what the aim is here, I may be able to suggest a way to do this.

vijayaditya:
*"If you don't specify a supervision object for an output, which contains a list of Indexes to be evaluated, then the output node will never be evaluated for any indexes."*
**Thanks for pointing this out !**

I will reimplement this using kQuadratic objfn and sparse inputs, as you suggested offline.

PR #945, opened 2016-08-03
Array job submission for train_ivector_extractor.sh

nsmartinez:
Now train_ivector_extractor.sh executes the acc estimation as an
array job when num_processes and num_threads are equal to one. The
problem is that when nj is high, a huge number of queue.pl are submitted
resulting a huge load in the exec machine.

This could be done also for the case that num_processes and num_threads
are distinct to one, but I would recommend doing it using an new script for sake of clarity.

### DIFF
--- a/egs/wsj/s5/steps/online/nnet2/train_ivector_extractor.sh
+++ b/egs/wsj/s5/steps/online/nnet2/train_ivector_extractor.sh
@@ -142,19 +142,26 @@ while [ $x -lt $num_iters ]; do
   if [ $stage -le $x ]; then
     rm $dir/.error 2>/dev/null
 
-    Args=() # bash array of training commands for 1:nj, that put accs to stdout.
-    for j in $(seq $nj_full); do
-      Args[$j]=`echo "ivector-extractor-acc-stats --num-threads=$num_threads $dir/$x.ie '$feats' 'ark,s,cs:gunzip -c $dir/post.JOB.gz|' -|" | sed s/JOB/$j/g`
-    done
+    if [ $num_processes -eq 1 ] && [ $num_threads -eq 1 ]; then #submit as an array
+        $cmd --num-threads 1 JOB=1:$nj_full $dir/log/acc.$x.JOB.log \
+            ivector-extractor-sum-accs --parallel=true \
+                "ivector-extractor-acc-stats --num-threads=1 $dir/$x.ie '$feats' 'ark,s,cs:gunzip -c $dir/post.JOB.gz|' -|"  $dir/acc.$x.JOB || touch $dir/.error
+    else
+         Args=() # bash array of training commands for 1:nj, that put accs to stdout.
+         for j in $(seq $nj_full); do
+           Args[$j]=`echo "ivector-extractor-acc-stats --num-threads=$num_threads $dir/$x.ie '$feats' 'ark,s,cs:gunzip -c $dir/post.JOB.gz|' -|" | sed s/JOB/$j/g`
+         done
 
-    echo "Accumulating stats (pass $x)"
-    for g in $(seq $nj); do
-      start=$[$num_processes*($g-1)+1]
-      $cmd --num-threads $[$num_threads*$num_processes] $dir/log/acc.$x.$g.log \
-        ivector-extractor-sum-accs --parallel=true "${Args[@]:$start:$num_processes}" \
-          $dir/acc.$x.$g || touch $dir/.error &
-    done
-    wait
+        echo "Accumulating stats (pass $x)"
+        for g in $(seq $nj); do
+            start=$[$num_processes*($g-1)+1]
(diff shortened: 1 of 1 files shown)


danpovey:
queue.pl is supposed to be a lightweight script, it uses very little memory
or CPU.   In what way is it using excessive resources?  Is it about NFS
traffic?  We recently made a change so that for the 'sync' files it makes a
subdirectory, and that reduces NFS traffic a lot.

Dan

nsmartinez:
I had this part of the script submit almost 2 thousand jobs and the machine went over 800 of load, just caused by the queue.pl . This causes also a lot of stress in the communication with SGE server, which causes multiple qstat failures.

I reviewed the code of queue.pl and as you said the lockfile/wait logic should avoid that, but still 2000 procs competing for the same lock file or qstat access could produce this high load or other problem. 

It also a problem of the decided $nj which should be capped out to a sensible number, but also, I think this is the only place of the training that this intensive use of queue.pl happens instead of using an array.

danpovey:
They are not competing for the same lock files but for different ones.
And they check qstat only quite rarely; I think a better fix would be to
change the frequency with which queue.pl checks qstat (which is not
necessary to its operation anyway, it's to help debug things).
Can you confirm that the queue.pl that you are using has the following code?

```

      *if* (($check_sge_job_ctr < 100 && ($check_sge_job_ctr++ % 10) == 0)
||

          ($check_sge_job_ctr >= 100 && ($check_sge_job_ctr++ % 50) == 0)) {

 ```

And does it handle the qstat failures gracefully?  I changed it recently so
that it should.  If not, let me know how it fails.

I don't like your change because it makes the script difficult to maintain
and test: any change would have to be made to both branches, and it's
unlikely that testing would uncover a bug in the branch that you're adding.

Also, I don't think the settings that you are running with really make
sense- unless you have a cluster consisting of single-core machines.  The
way you are running it will excessively stress the disk and waste memory.

Dan

nsmartinez:
Yes I am using that version. From what I can tell they all use the same lock. This process is submitted $nj times as:

```$cmd --num-threads $[$num_threads*$num_processes] $dir/log/acc.$x.$g.log \
        ivector-extractor-sum-accs --parallel=true "${Args[@]:$start:$num_processes}" \
          $dir/acc.$x.$g || touch $dir/.error &```

in queue.pl:

```my $dir = dirname($logfile);
my $base = basename($logfile);
my $qdir = "$dir/q";
....
system("touch $qdir/sync/.kick");```

That means that all processes are trying to touch $dir/log/q/sync/.kick, right?

qstat failures are correctly handled but it is still producing a huge load.

In the infrastructure I am using, all the experiment is ran in a huge RAID and workers just read/write the data using NFS, so I think that using a higher num_processes/threads wouldn't be different, but it might be worth trying as sometimes reading/writing cache in NFS is not handled as one might think.

You are right in terms of making testing harder. I could wrap this part of the script in another bash script that properly submit array jobs using the desired number of proc/threads.  I would be just, generate the proper list of files depending on the array and executing the ivector-extractor-sum-accs.

jtrmal:
Dan, I had issues with this as well on MARCC.
My overall feeling is against the normal kaldi scripts patterns i.e. it
might cause unexpected load patterns. So from my POV it should be changed
to change it to say two-stage accumulation to honor the --nj parameter
setting.
y.

nsmartinez:
As I said, I noticed that in case nj is high this part of the training spawns lots of queue.pl and that increment the load substantially. You can run a simple test doing:

```for i in `seq 1 2000`; do (./queue.pl foo.$i.log sleep 10 &); done```

In my machine this increments the load over 800.

In the experiment I ran this triggered some alarms as the load increment lasted a long period of time. However, there were no qstat or NFS problems. Sorry to have brought up the discussion about NFS, my only issue was with the load increment.

This "load problem" (if it is a problem) only affects the machine submitting the exp so it might not really make a difference as other worker are executing the orders. Also, as you say queue.pl is light weight and should not affect the other processes, even with thousands being run, but as I saw that the problem could be easily solved using an array I preferred to make a fix to not cause this load increment.

jtrmal:
Actually, when I'm thinking about it, my issue  was different -- my issue
was  that  $num_threads*$num_processes was too big for a single machine to
be able to accommodate it, so SLURM either refused to run the job (I don't
recall now if it crashed or just was queued indefinitely).
y,

danpovey:
@jtrmal- that's to be expected, I guess that queue didn't have machines
with enough threads.

Nicolas--
I've had people talk before about high GridEngine load, and it frustrates
me a little because nobody really knows that GridEngine load even means.
It's just a black box computation from unknown inputs.  If you were to say
that there was excessive NFS traffic, or swapping, or oversubscribed CPU,
or something like that, I would admit that that was a problem.

When I do GridEngine administration, firstly I don't often even look at the
load, but if I do, I only use it as a reason to do further investigation,
not as a problem in itself.

danpovey:
Closing this PR as I don't think I'll be merging it.

PR #948, opened 2016-08-04
fix bug when do blstm nnet-forward

yunlinchen:
@danpovey

### DIFF
--- a/src/nnet/nnet-blstm-projected-streams.h
+++ b/src/nnet/nnet-blstm-projected-streams.h
@@ -573,6 +573,11 @@ class BLstmProjectedStreams : public UpdatableComponent {
     int DEBUG = 0;
 
     int32 nstream_ = sequence_lengths_.size();
+    // if not set num_streams, set 1 stream
+    if (nstream_ == 0) {
+      nstream_ = 1;
+      sequence_lengths_.push_back(in.NumRows());
+    }
     KALDI_ASSERT(in.NumRows() % nstream_ == 0);
     int32 T = in.NumRows() / nstream_;
     int32 S = nstream_;


danpovey:
OK, I'm going to merge this because we haven't heard from Karel for quite a long time.
If he objects to it, we can revert it later.

yunlinchen:
@danpovey.  okay,  because when I do nnet-forward for blstm model, always core for floating point exception. I check it, found that num_streams=0 as denominator when do division

danpovey:
OK.  Right now no-one is properly maintaining the nnet1 blstm code.  The nnet3 code is under much more active development.

yunlinchen:
@danpovey okay, thank you for your warm apply.

lifeiteng:
Actually , this is not a bug, you can set stream before call nnet-forward to avoid exception.

vesis84:
Hi, recently I was working with CTC-BLSTM from EESEN. Now I am refactoring the BLSTM code, the update will be out in few days...

yunlinchen:
@lifeiteng yes, but in this single binary:nnet-forward, no argument to set this.

yunlinchen:
@vesis84 okay, thx, blstm in nnet1?

vesis84:
The code should not depend on setting the stream-number externally. The default should be 1, if not specified otherwise. There will be a new interface: MultistreamComponent derived from UpdatableComponent handling this systematically...

vesis84:
yes, nnet1 blstm

yunlinchen:
@vesis84 okay, for  I read lstm stream code,  it set 1 when numstream=0, when you push your code, i will try it

PR #950, opened 2016-08-05
nnet1: redesigning LSTM, BLSTM code,

vesis84:
Any volunteer for the review?

- introducing interface 'MultistreamComponent',
  - handles stream-lengths and stream-resets,
- rewritten most of training tools 'nnet-train-lstm-streams',
  'nnet-train-blstm-streams',
- introducing 'RecurrentComponent' with simple forward recurrency.
- the LSTM/BLSTM components have clipping presets we recently
  found helpful for BLSTM-CTC system.
- renaming tools and components (removing 'streams' from names)
- updating the scripts for generating lstm/blstm prototypes
- updating 'rm' lstm/blstm examples


src/nnetbin/nnet-train-multistream.cc
```diff
@@ -0,0 +1,433 @@
...
+      size_t frames_to_go = 0;
+      for (int32 s = 0; s < num_streams; s++) {
+        frames_to_go += feats_utt[s].NumRows();
+      }
+      if (frames_to_go == 0) break;
+
+      // number of frames we'll pack as the streams,
+      std::vector<int32> frame_num_utt;
+
+      // pack the parallel data,
```
danpovey: It seems to me like the code would be easier to follow if you broke some things out into functions.
This is way past the "soft limit" of 40 lines per function.

danpovey:
Karel, regarding back-compatibility, I'm not sure if it can read the old models-- if not, can you at least make sure it dies with an informative error message?

danpovey:
Also, notice that the build failed.

src/nnetbin/nnet-train-multistream.cc
vesis84: Hi, okay, I rewrote this place. I understand that the short functions are easier to understand than the 'script-like' code of the binary with many dependencies. In the same the 'script-like' code is easier to modify. As a compromise I used the brace-blocks with a comment in some places.

What is happens in the 'nnet-train-multistream' code is that first, whole sentences are filled into 'vector<Matrix<>>', from which it is sliced later to form the multi-stream mini-batch with interleaved layout of frames:
[feaRow1_stream1,
feaRow1_stream2,
...
feaRow1_streamN,
feaRow2_stream1,
feaRow2_stream2,
...
feaRow2_streamN
]

Similar thing can be found in EESEN, where the 'batch' is formed from whole utterances of roughly same length. The multi-stream training with whole sentences is implemented in 'nnet-train-multistream-perutt'.

src/nnetbin/nnet-train-multistream.cc
```diff
@@ -0,0 +1,461 @@
...
+              );
+            }
+          }
+        }
+      }
+
+      // pass the info about padding,
+      nnet.SetSeqLengths(frame_num_utt);
+      // Show the 'utt' lengths in the VLOG[2],
+      if (kaldi::g_kaldi_verbose_level >= 2) {
```
danpovey: You're supposed to check the verbose level with GetVerboseLevel().

src/nnetbin/nnet-train-multistream.cc
```diff
@@ -0,0 +1,461 @@
...
+        WriteIntegerVector(
+          KALDI_LOG << "frame_num_utt[" << frame_num_utt.size() << "]",
```
danpovey: Karel, I don't know what's supposed to be going on here.. I'm astonished that this would compile, but even if it does, it's not OK.  The macros like KALDI_LOG are supposed to just be invoked, there should be no assumption about what they return.  You may have to use std::ostringstream.

src/nnetbin/nnet-train-multistream.cc
```diff
@@ -0,0 +1,461 @@
...
+        double time_now = time.Elapsed();
+        KALDI_VLOG(1) << "After " << num_done << " utterances, "
+          << "(" << total_frames/360000.0 << "h), "
+          << "time elapsed = " << time_now / 60 << " min; "
+          << "processed " << total_frames / time_now << " frames per sec.";
+      }
+
+      // monitor the NN training (--verbose=2),
+      int32 F = 25000;
+      if (kaldi::g_kaldi_verbose_level >= 3) {
```
danpovey: again, use GetVerboseLevel()

egs/rm/s5/local/nnet/run_blstm.sh
```diff
@@ -4,16 +4,12 @@
...
-# The BLSTM code comes from Ni Chongjia (I2R), thanks!
+# The initial BLSTM code comes from Ni Chongjia (I2R), thanks!
 
-# TODO, this BLSTM code needs to solve a problem how to determine the 
-# history for the 'backward' recurrency. Currently is taken the state
-# on 1st frame of previous mini-batch (20 frames).
-#
-# A more sensible approach should be single-stream training, 
-# and per-utterance updates. But the results were worse.
+# We use multi-stream training, while the BPTT is done over whole
```
danpovey: shouldn't there be results added somewhere for this RM example?

src/nnetbin/nnet-train-multistream.cc
vesis84: Yea, that is a "nice" trick that came up to my mind recently :) 
The macro returns 'std::ostream' even after writing to it and finally the integer vector is appended.
Okay, I'll change that...

PR #953, opened 2016-08-05
Various updates to tedlium release 2 recipe; removing so-far-unused f…

danpovey:
…iles (will add them later if needed.)


danpovey:
Note to @vince62s-- I've removed files that we haven't used yet, we can add them from the old tedlium recipe as we test them.  It's not clear to me how your iwslt data preparation script fits into this whole thing, could you please think about how to clarify that?

vince62s:
I saw you removed the nnet3 and chain scripts too ....
for the iwslt data prep it was mostly a copy of the data prep script except for the glm part.
Ideally, this script should be inserted in the data prep steps, and then there could be an extra "test set" called "iwslt" which would give some kind of reference to compare our results versus the iwslt sessions.
The only reason why I did not insert it was the download part, because I had to google it to find it. did not find any "official" download.

danpovey:
The nnet3 and chain scripts can be added back selectively, with changes;
I'll do that.  But I prefer to start from a clean slate.

Regarding the IWSLT stuff, I don't have time to deal with that-- if you can
figure out a proposal to include it in a suitable way and make a PR for the
changes, that would be good.  We can mirror the data on OpenSLR if needed,
later on, but don't let that block you for now.

Dan

PR #955, opened 2016-08-07
WIP : modifications to paste_feats.sh

vijayaditya:
I will keep modifying paste_feats.sh as I see more necessary changes.

### DIFF
--- a/egs/wsj/s5/steps/paste_feats.sh
+++ b/egs/wsj/s5/steps/paste_feats.sh
@@ -54,8 +54,15 @@ name=`basename $data`
 
 # get list of source scp's for pasting
 data_src_args=
+cmvn_src_args=
+create_cmvn_scp=true
+
 for data_src in ${data_src_arr[@]}; do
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/steps/paste_feats.sh
```diff
@@ -69,12 +76,16 @@ $cmd JOB=1:$nj $logdir/append.JOB.log \
    copy-feats --compress=$compress ark:- \
     ark,scp:$ark_dir/pasted_$name.JOB.ark,$ark_dir/pasted_$name.JOB.scp || exit 1;
 
+# create the cmvn scp file
+if [ "$create_cmvn_scp" == "true" ]; then
+  paste-feats --length-tolerance=0 $cmvn_src_args ark,scp:$ark_dir/cmvn.ark,$data/cmvn.scp
```
danpovey: this wouldn't work right because the CMVN stats have an extra dimension for the counts.  I think it would be safer to just recompute the CMVN stats.

vijayaditya: OK.

PR #960, opened 2016-08-11
some fixes of tdnn+{xent,chain}+sMBR recipes on librispeech and add t…

freewym:
…heir results. Fix an option name bug in steps/nnet3/align.sh

### DIFF
--- a/egs/librispeech/s5/RESULTS
+++ b/egs/librispeech/s5/RESULTS
@@ -443,6 +443,92 @@
 %WER 15.10 [ 7904 / 52343, 874 ins, 1070 del, 5960 sub ] exp/nnet3/tdnn_sp/decode_test_other_tgmed/wer_13_0.0
 %WER 16.29 [ 8528 / 52343, 828 ins, 1320 del, 6380 sub ] exp/nnet3/tdnn_sp/decode_test_other_tgsmall/wer_14_0.0
 
+# Results with nnet3 tdnn+sMBR
+# local/nnet3/run_tdnn_discriminative.sh
+# a subset of the full list of results (using the acoustic model obtained at the end of the training):
+%WER 4.07 [ 2214 / 54402, 313 ins, 195 del, 1706 sub ] exp/nnet3/tdnn_sp_smbr/decode_dev_clean_fglarge_epoch4.adj/wer_14_1.0
+%WER 4.20 [ 2286 / 54402, 337 ins, 206 del, 1743 sub ] exp/nnet3/tdnn_sp_smbr/decode_dev_clean_tglarge_epoch4.adj/wer_17_0.5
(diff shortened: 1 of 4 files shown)


egs/wsj/s5/steps/nnet3/align.sh
```diff
@@ -124,7 +124,7 @@ ivector_opts=
 if [ ! -z "$online_ivector_dir" ]; then
   ivector_period=$(cat $online_ivector_dir/ivector_period) || exit 1;
   # note: subsample-feats, with negative n, will repeat each feature -n times.
-  ivector_opts="--online-ivectors=scp:$online_ivector_dir/ivector_online.scp --online-ivector_period=$ivector_period"
+  ivector_opts="--online-ivectors=scp:$online_ivector_dir/ivector_online.scp --online-ivector-period=$ivector_period"
```
vijayaditya: What was the difference in performance due to this bug ?

freewym: I didn't compare. I just found it accidently

vijayaditya: You should modify the commit message to highlight this bug, as it would be relevant to several recipes.

egs/librispeech/s5/RESULTS
```diff
@@ -443,6 +443,73 @@
 %WER 15.10 [ 7904 / 52343, 874 ins, 1070 del, 5960 sub ] exp/nnet3/tdnn_sp/decode_test_other_tgmed/wer_13_0.0
 %WER 16.29 [ 8528 / 52343, 828 ins, 1320 del, 6380 sub ] exp/nnet3/tdnn_sp/decode_test_other_tgsmall/wer_14_0.0
```
vijayaditya: This results file is really tough to read and compare. Would you be able to provide a brief set of important results right after you specify the command and present the detailed results after than ?

egs/wsj/s5/steps/nnet3/align.sh
freewym: Modifed.

vijayaditya: Modifying the way this option is specified does not change the alignments, so I guess the option parser in c++ is immune this mis-specification.

vijayaditya:
Please fix the spelling mistakes in your commit message.

egs/librispeech/s5/RESULTS
freewym: Added a subset of the results before the full list of results

vijayaditya:
You can add a line new line in the commit message. Also please change
"Fix" --> "Fixed".

In general you should have comments which follow [this guideline.](https://google.github.io/styleguide/cppguide.html#Function_Comments)
**comments should be descriptive ("Opens the file") rather than imperative ("Open the file")**

Issue #947, opened 2016-08-03
bug in get_ctm.sh


vince62s:
I think there is a bug in the recent change below
for instance if nj=1 then the second line in bold will become lats=.../lat.{1}.gz
instead of lat.1.gz

```
nj=$(cat $dir/num_jobs)
+  **lats=$(eval echo $dir/lat.{$(seq -s, $nj)}.gz)**
   if [ -f $lang/phones/word_boundary.int ]; then
     $cmd LMWT=$min_lmwt:$max_lmwt $dir/scoring/log/get_ctm.LMWT.log \
       set -o pipefail '&&' mkdir -p $dir/score_LMWT/ '&&' \
-      lattice-1best --lm-scale=LMWT "ark:gunzip -c $dir/lat.*.gz|" ark:- \| \
+      lattice-1best --lm-scale=LMWT "ark:gunzip -c $lats|" ark:- \| \
       lattice-align-words $lang/phones/word_boundary.int $model ark:- ark:- \| \
-      nbest-to-ctm --frame-shift=$frame_shift ark:- - \| \
+      nbest-to-ctm --frame-shift=$frame_shift --print-silence=$print_silence ark:- - \| \
       utils/int2sym.pl -f 5 $lang/words.txt \| \
       $filter_cmd '>' $dir/score_LMWT/$name.ctm || exit 1;
-  else
-    if [ ! -f $lang/phones/align_lexicon.int ]; then
-      echo "$0: neither $lang/phones/word_boundary.int nor $lang/phones/align_lexicon.int exists: cannot align."
-      exit 1;
-    fi
-
+  elif [ ! -f $lang/phones/align_lexicon.int ]; then
     $cmd LMWT=$min_lmwt:$max_lmwt $dir/scoring/log/get_ctm.LMWT.log \
       set -o pipefail '&&' mkdir -p $dir/score_LMWT/ '&&' \
-      lattice-1best --lm-scale=LMWT "ark:gunzip -c $dir/lat.*.gz|" ark:- \| \
+      lattice-1best --lm-scale=LMWT "ark:gunzip -c $lats|" ark:- \| \
       lattice-align-words-lexicon $lang/phones/align_lexicon.int $model ark:- ark:- \| \
-      nbest-to-ctm --frame-shift=$frame_shift ark:- - \| \
+      nbest-to-ctm --frame-shift=$frame_shift --print-silence=$print_silence ark:- - \| \
[...]

PR #963, opened 2016-08-12
fix the bug about seq (issue #947)

LvHang:
@danpovey @vince62s 
This pull request is about issue #947 
1.I think vince62s's problem is mainly caused by the difference between 'dash' and 'bash' when they try to  interpret `{}` , rather than ‘seq’.
For example, the command "lat.{1,2,3}.gz" will be replaced by "lat.1.gz lat.2.gz lat.3.gz" in bash, but nothing will be done in dash.
2.I submit a solution and I have tested them in both dash and bash. Meanwhile, I did some tests on Switchboard. I think everything goes smoothly.
  ```C++
 - lats=$(eval echo $dir/lat.{$(seq -s, $nj)}.gz)
 + for n in $(seq $nj); do
 +   lats=$(eval echo "$lats $dir/lat.$n.gz")
 + done
  ```
3.I still have a doubt. At the begining of "get_ctm.sh" file, there is a command "#!/bin/bash ". I think the script will be conducted in bash circumstance. Why vince62s's problem will happen? I'm not very clear about it. 

If the code has any questions, please contect me as soon as possible. I would appreciate it very much.
Hang

### DIFF
--- a/egs/wsj/s5/steps/nnet2/get_egs_discriminative2.sh
+++ b/egs/wsj/s5/steps/nnet2/get_egs_discriminative2.sh
@@ -80,10 +80,6 @@ mkdir -p $dir/log $dir/info || exit 1;
 nj=$(cat $denlatdir/num_jobs) || exit 1; # $nj is the number of
                                          # splits of the denlats and alignments.
 
-
-[ "$(readlink /bin/sh)" == dash ] && \
-  echo "This script won't work if /bin/sh points to dash.  make it point to bash." && exit 1
-
 nj_ali=$(cat $alidir/num_jobs) || exit 1;
 
 sdata=$data/split$nj
@@ -91,16 +87,16 @@ utils/split_data.sh $data $nj
 
 if [ $nj_ali -eq $nj ]; then
   ali_rspecifier="ark,s,cs:gunzip -c $alidir/ali.JOB.gz |"
-  all_ids=$(seq -s, $nj)
-  prior_ali_rspecifier="ark,s,cs:gunzip -c $alidir/ali.{$all_ids}.gz | copy-int-vector ark:- ark,t:- | utils/filter_scp.pl $dir/priors_uttlist | ali-to-pdf $alidir/final.mdl ark,t:- ark:- |"
+  alis=$(for n in $(seq $nj); do echo $alidir/ali.$n.gz; done)
+  prior_ali_rspecifier="ark,s,cs:gunzip -c $alis | copy-int-vector ark:- ark,t:- | utils/filter_scp.pl $dir/priors_uttlist | ali-to-pdf $alidir/final.mdl ark,t:- ark:- |"
 else
   ali_rspecifier="scp:$dir/ali.scp"
   prior_ali_rspecifier="ark,s,cs:utils/filter_scp.pl $dir/priors_uttlist $dir/ali.scp | ali-to-pdf $alidir/final.mdl scp:- ark:- |"
   if [ $stage -le 1 ]; then
     echo "$0: number of jobs in den-lats versus alignments differ: dumping them as single archive and index."
-    all_ids=$(seq -s, $nj_ali)
+    alis=$(for n in $(seq $nj_ali); do echo $alidir/ali.$n.gz; done)
     $cmd $dir/log/copy_alignments.log \
-      copy-int-vector "ark:gunzip -c $alidir/ali.{$all_ids}.gz|" \
+      copy-int-vector "ark:gunzip -c $alis|" \
       ark,scp:$dir/ali.ark,$dir/ali.scp || exit 1;
   fi
 fi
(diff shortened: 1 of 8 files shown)


vince62s:
I fully agree, I am pretty sure it was run in bash .....
and the error message was that lat.{1}.gz did not exist, as if it was unable to handle the { }
but there are many other instances where the array stuff { } is properly interpreted.

jtrmal:
It was discussed here or on the mailing list a couple of times. The issue
 that popen(3) and all the higher level functions based on this call use sh
to run the command. Depending on to what sh is linked to --either dash or
bash -- the {} expansion is done correctly or not.

quoting popen(3):
The *command* argument is a pointer to a null-terminated string containing
a shell command line. This command is passed to*/bin/sh* using the *-c* flag;
interpretation, if any, is performed by the shell.


y.

jtrmal:
So the differences in behavior you've seen basically boils down to the
question where the expansion is supposed to happen -- if in the script
directly, then it will work ok (as the bash was specified in the hashbang),
but if you submit the expansion as a part of the ark: specification, it
will depend on what sh is symlinked to.
y.

vince62s:
ok very clear, thanks.

danpovey:
There are two separate issues: the bash/dash issue, and the fact that it won't work if nj == 1.   They can both be fixed in the same way, but please note that we're not talking about fixing this in just one place, there are many places in the scripts where this needs to be fixed, you need to do a 'find'.
Also, you solution takes O(nj^2), which is too slow.
You should do
```
 lats=$(for n in $(seq $nj); do echo $dir/lat.$nj.gz; done)
```
And if any of these scripts check whether /bin/sh links to dash, remove that check.

Dan

LvHang:
@jtrmal 
Your explanation is very clear. Thank you very much.

@danpovey 
I searched all the 'seq -s' commands in kaldi. I fixed most of them with the similar "for loop" commands as your suggestion. And if any of these scripts check whether /bin/sh links to dash, I have removed it.
The remaining two are
1. egs/csj/s5/local/csj_make_trans/csj_autorun.sh:
```C++
"dvd" ) num=dvd  ; SDB=  ; WAV= ; disc=$num`seq -s " "$num 3 17` ;;
```
Because I think it is safe because it will work in the script.
2. egs/librispeech/s5/local/run_rnnlm.sh:
```C++
rnnlm_cmd="taskset -c $(seq -s, 0 $(( $num_threads - 1 )) ) $rnnlm_cmd"
```
Because I checked the usage and found that *taskset* needs a comma-separated list. Meanwhile, I think it is safe,too.

Thank you very much for your guidance.
Hang

egs/librispeech/s5/local/lm/train_lm.sh
```diff
@@ -47,7 +47,7 @@ split_prefix=$tmp_dir/split
 if [ "$stage" -le 1 ]; then
   mkdir -p $tmp_dir
   echo "Splitting into $normjobs parts, to allow for parallel processing ..."
-  split_files=$(eval "echo $split_prefix-{$(seq -s',' $normjobs)}")
+  split_files=$(for n in $(seq $normjobs); do echo $split_prefix-$n; done)
```
danpovey: I think it's better if you revert this change, that will avoid me having to verify that it's correct.

egs/librispeech/s5/local/prepare_dict.sh
```diff
@@ -72,7 +72,7 @@ if [ $stage -le 1 ]; then
   auto_lexicon_prefix="$g2p_dir/lexicon_autogen"
 
   mkdir -p $g2p_dir/log
-  auto_vocab_splits=$(eval "echo $auto_vocab_prefix.{$(seq -s',' $nj)}")
+  auto_vocab_splits=$(for n in $(seq $nj); do echo $auto_vocab_prefix.$n; done)
```
danpovey: also this... revert anything that doesn't have to do with brace expansion.  the bug we are trying to fix is with brace expansion.

egs/wsj/s5/steps/nnet2/get_egs_discriminative2.sh
```diff
@@ -91,16 +86,16 @@ utils/split_data.sh $data $nj
 
 if [ $nj_ali -eq $nj ]; then
   ali_rspecifier="ark,s,cs:gunzip -c $alidir/ali.JOB.gz |"
-  all_ids=$(seq -s, $nj)
-  prior_ali_rspecifier="ark,s,cs:gunzip -c $alidir/ali.{$all_ids}.gz | copy-int-vector ark:- ark,t:- | utils/filter_scp.pl $dir/priors_uttlist | ali-to-pdf $alidir/final.mdl ark,t:- ark:- |"
+  alis=$(for all_ids in $(seq $nj); do echo $alidir/ali.$all_ids.gz; done)
```
danpovey: this variable name all_ids is not suitable now that you changed the code.  n would be better.  please make the same change elsewhere if you did the same.

LvHang:
@danpovey 
I have already reverted the changes in "librispeech" folder.
I have already fixed the variable name in everywhere.
Please check wheather the changes meet your standards. 
Thank you for your time and your suggestion.

Hang

danpovey:
Looks OK but can you rebase so there is only one commit?
Something like
git rebase -i HEAD~5
and change all but the first 'pick' to 's' [for squash].
Save your work in a different branch first, in case of errors.

LvHang:
@danpovey 
The changes have been integrated. There is only one commit.
Please check it. 
Thank you for your time and your guidance.

Hang

PR #978, opened 2016-08-18
Make slurm.pl more resilient to random squeue failures

psmit:
In our cluster squeue sometimes fails for other reasons than that a job does not exist anymore. This commit retries squeue once if it fails.

### DIFF
--- a/egs/wsj/s5/utils/slurm.pl
+++ b/egs/wsj/s5/utils/slurm.pl
@@ -95,6 +95,12 @@ sub print_usage() {
   exit 1;
 }
 
+sub exec_command {
+  my $command = join ' ', @_;
+  # To get the actual exit value, shift right by eight bits.
+  ($_ = qx{$command 2>&1}, $? >> 8);
+}
+
 if (@ARGV < 2) {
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/slurm.pl
```diff
@@ -95,6 +95,12 @@ ()
   exit 1;
 }
 
+sub exec_command {
+  my $command = join ' ', @_;
+  # To get the actual exit value, shift right by eight bits.
+  ($_ = qx{$command 2>&1}, $? >> 8);
```
danpovey: This PR is mostly OK, but can you please rewrite the exec_command function to be easier to understand and to use backticks or "system" instead of the very obscure 'qx' operator?

senarvi: From `system()` it's impossible to get the output of the command, but `qx` and backticks are equivalent, so we should be able to use backticks.

psmit:
I have implemented the suggested changes.

danpovey:
Thanks!  Merging.

Issue #966, opened 2016-08-15
ASpIRE reproducibility issue


vijayaditya:
Even after correcting the bug  #889  @tomkocse has issues reproducing the results in the ASpIRE recipe. However the WER difference is just 2% abs (it was 11% abs with #889 ). Using the `data/impulses_noises` directory prepared by the old recipe and the latest data reverberation scripts he is able to reproduce the results. Thus the reproducibility issue has been isolated to the new `aspire/s5/local/multi_condition/prepare_impulses_noises.sh` script. 
I will be working on this issue this week.

PR #981, opened 2016-08-20
WIP : aspire: fixed bug in impulse/noise normalization script

vijayaditya:
This hopefully addresses the issue #966 . The bug was that I was scaling the integer wav data with max_integer_value while it was still in integer format. This results in the normalization coefficient being computed from wav data which was just a sequence of {-1,0,1}. Thus the normalization coefficient were incorrect and the wave data was not properly normalized. The new impulses/noises are very similar to the old impulses/noises (created before the bug was introduced).

Testing in progress. We usually get a reliable WER estimate by 100th iteration, which occurs within an hour of training.

### DIFF
--- a/egs/aspire/s5/local/multi_condition/normalize_wavs.py
+++ b/egs/aspire/s5/local/multi_condition/normalize_wavs.py
@@ -29,11 +29,15 @@ def get_normalization_coefficient(file_list, is_rir, additional_scaling):
         assert(rate == sampling_rate)
       else:
         sampling_rate = rate
+      # numpy does not check for numerical overflow in integer type
+      # and it is important to convert data to float before division to avoid
+      # rounding errors
+      data = data.astype(np.float64)
       data = data / dtype_max_value
       if is_rir:
         # just count the energy of the direct impulse response
         # this is treated as energy of signal from 0.001 seconds before impulse
-        # to 0.05 seconds after impulse. This is done as we do not want the 
+        # to 0.05 seconds after impulse. This is done as we do not want the
         # recording length to influence the scaling factor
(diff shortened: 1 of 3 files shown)


vijayaditya:
Reg. 
```bash
login2:~: x=$(for n in $(seq 5); do echo $n; done)
login2:~: echo $x
1 2 3 4 5
```
Yesterday my run failed as the expansion happened with a newline character. Let me check if there was something peculiar about the machine it was running on.

vijayaditya:
@danpovey The difference in expansion happens due to the string delimiters.

```bash
x=$(for n in $(seq 10); do echo str.$n; done)
echo $x
# str.1 str.2 str.3 str.4 str.5 str.6 str.7 str.8 str.9 str.10

echo "$x"
#str.1
#str.2
#str.3
#str.4
#str.5
#str.6
#str.7
#str.8
#str.9
#str.10

```

danpovey:
Oh.  that same bug exists in a few scripts then, it was introduced in PR
#963.  My fault.  I'll commit a fix right now.

danpovey:
OK, see PR #983

vijayaditya:
Fixing normalization bug does not improve the results; continuing investigation.

vijayaditya:
Restarting work on this issue. The problems seems a bit more involved than anticipated.

vijayaditya:
@tomkocse is preparing an upload for openslr.org including the *good* real RIRs which give us better results, so this is no longer a *critical* issue. I will resolve this after @tomkocse merges his swbd+reverb recipe, as the turnaround time for this new experiment is significantly smaller (5 hrs) compared to the ASpIRE recipe (5 days).

PR #983, opened 2016-08-20
Fix bug regarding newlines in strings, introduced in commit d90d3545f

danpovey:
Also small cosmetic changes.


danpovey:
Note: see conversation in https://github.com/kaldi-asr/kaldi/pull/981 regarding the reasons for this commit.

PR #985, opened 2016-08-21
Fix bug in log-softmax kernel [RE matrices with different strides]

danpovey:


danpovey:
@kangshiyin please note this...

danpovey:
Also @vijayaditya, there is a chance that this could explain the poor performance of chain models with BLSTMs-- since BLSTMs use a lot of memory, they will be more likely to trigger this bug.  It might sometimes cause silent errors instead of crashes.

vijayaditya:
Thanks for the pointer. I will check for these issues.

--Vijay

src/cudamatrix/cu-kernels.cu
```diff
@@ -2214,12 +2226,12 @@ static void _log_softmax_reduce(Real *y, const Real *x,
...
 
   // Log sum step 1: substracts max, and takes exponentials.
-  y[threadIdx.x + j * d.stride] = x[threadIdx.x + j * d.stride] - max;
+  y[threadIdx.x + j * d.stride] = x[threadIdx.x + j * src_stride] - max;
   aux[threadIdx.x] = exp(y[threadIdx.x + j * d.stride]);
   for (int i = 1; i < steps; ++i) {
     if (threadIdx.x + i * THREADS < d.cols) {
       y[threadIdx.x + i * THREADS + j * d.stride] =
-        x[threadIdx.x + i * THREADS + j * d.stride] - max;
+        x[threadIdx.x + i * THREADS + j * src_stride] - max;
```
kangshiyin: It looks good now - `y` with `d.stride` and `x` with `src_stride`. I'm also working on this function, and will make a PR later.

PR #990, opened 2016-08-22
Remove sp from fisher_swbd nnet3 recipe

naxingyu:
Seq. training is not applied yet.

### DIFF
--- a/egs/fisher_swbd/s5/RESULTS
+++ b/egs/fisher_swbd/s5/RESULTS
@@ -41,45 +41,40 @@ for x in exp/nnet2_online/nnet_ms_a_online/decode_eval2000*_fg; do grep Sum  $x/
 %WER 11.8 | 1831 21395 | 89.6 7.2 3.2 1.4 11.8 49.0 | exp/nnet2_online/nnet_ms_a_online/decode_eval2000_utt_offline_fsh_sw1_fg/score_11/eval2000.ctm.swbd.filt.sys
 
 # nnet3 result on eval2000
-# BLSTM ran for about 760 hours, command:
-# local/nnet3/run_lstm.sh --affix bidirectional --lstm-delay " [-1,1] [-2,2] [-3,3] " --label-delay 0 \
-#                         --cell-dim 1024 --recurrent-projection-dim 128 --non-recurrent-projection-dim 128 \
+# BLSTM ran for about 280 hours, command:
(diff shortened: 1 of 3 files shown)


danpovey:
Oh... I didn't think this was what you were going to do.  I thought the thing you were going to check in was the change to add data-cleanup [is what you meant by max-wer?] and combine-short-segments, while keeping the speed-perturbation [even if in your local setup you had not done the data-perturbation for reasons of speed].  The speed perturbation is always helpful and if it's slow you can use fewer epochs.

naxingyu:
We had a discussion in the multi_en PR that the benefit of SP is limited for large scale data. And I tried using fewer epochs on SPed data, but the performance was bad. So I'll still tune on non-SP data and keep the remote repo with SP, is that correct?

danpovey:
Oh OK, I had misunderstood the other conversation.  I'll probably merge this PR as-is (i.e. remove sp), but I'd first like to get Vijay's opinion on this.
Also the chain models are a bit more data-hungry than regular nnet3, so there is a greater chance that sp will help on the chain setup.

naxingyu:
Yes, maybe it is. But at the mean time, they are also epoch-hungry in 
such large scale setup, especially for the LSTM models. Removing SP 
could at least give us a pretty good result within reasonable amount of 
time. For instance, chain/blstm took 16 days to train on SP data, while 
it took only 5 days to train on non-SP data, with the cost of ~0.5% abs. 
WER loss.


在 2016/8/22 13:40, Daniel Povey 写道:

danpovey:
OK, but you could always have reduced the number of epochs considerably-
unless the egs-dumping was the limiting factor.
Dan

vijayaditya:
The egs dumping process for chain models in ASpIRE (~5400 hrs of data) takes ~24 hours
when using 12 jobs for get_egs.sh. As fisher_swbd with sp option has similar data size it might be a limiting factor. However given that fisher_swbd is the task on which many papers report state-of-the-art results we might not want to remove speed-perturbation completely from the scripts. This could be useful when we are trying to publish our final results. 

I would recommend using speed_perturb=false by default but keeping the code related to perturbation.

naxingyu:
I think I'll close this PR and leave it as it is. But I'll continue the 
fisher_swbd exps without SP on my end for the combine_short_segment 
stuff, and discriminative training. BTW, if I get the results of 
chain+smbr before IS, do we put it in the presentation?


在 2016/8/23 5:57, Vijayaditya Peddinti 写道:

danpovey:
OK.
not sure about the interspeech presentation-- maybe.

PR #992, opened 2016-08-22
WIP : Changes to steps/data/reverberate_data_dir.py : to accept multiple {rir,noise}_lists as input and to optimize reco2dur computation.

tomkocse:
…sox piped command; allowing multiple input of rir_list and noise_list

### DIFF
--- a/egs/wsj/s5/steps/data/reverberate_data_dir.py
+++ b/egs/wsj/s5/steps/data/reverberate_data_dir.py
@@ -14,21 +14,31 @@ def GetArgs():
     parser = argparse.ArgumentParser(description="Reverberate the data directory with an option "
                                                  "to add isotropic and point source noises. "
                                                  "Usage: reverberate_data_dir.py [options...] <in-data-dir> <out-data-dir> "
-                                                 "E.g. reverberate_data_dir.py --rir-list-file rir_list "
+                                                 "E.g. reverberate_data_dir.py --rir-set-parameters rir_list "
                                                  "--foreground-snrs 20:10:15:5:0 --background-snrs 20:10:15:5:0 "
                                                  "--noise-list-file noise_list --speech-rvb-probability 1 --num-replications 2 "
                                                  "--random-seed 1 data/train data/train_rvb",
                                      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 
-    parser.add_argument("--rir-list-file", type=str, required = True, 
-                        help="RIR information file, the format of the file is "
+    parser.add_argument("--rir-set-parameters", type=str, action='append', required = True, dest = "rir_set_para_array", 
+                        help="Specifies the parameters of an RIR set. "
+                        "Supports the specification of  mixture_weight and rir_list_file_name. The mixture weight is optional. "
+                        "The default mixture weight is the probability mass remaining after adding the mixture weights "
+                        "of all the RIR lists, uniformly divided among the RIR lists without mixture weights. "
+                        "E.g. --rir-set-parameters '0.3, rir_list' or 'rir_list' "
+                        "the format of the RIR list file is "
                         "--rir-id <string,required> --room-id <string,required> "
                         "--receiver-position-id <string,optional> --source-position-id <string,optional> "
                         "--rt-60 <float,optional> --drr <float, optional> location <rspecifier> "
                         "E.g. --rir-id 00001 --room-id 001 --receiver-position-id 001 --source-position-id 00001 "
                         "--rt60 0.58 --drr -4.885 data/impulses/Room001-00001.wav")
-    parser.add_argument("--noise-list-file", type=str, default = None,
-                        help="Noise information file, the format of the file is"
+    parser.add_argument("--noise-set-parameters", type=str, action='append', default = None, dest = "noise_set_para_array",
+                        help="Specifies the parameters of an noise set. "
+                        "Supports the specification of  mixture_weight and noise_list_file_name. The mixture weight is optional. "
+                        "The default mixture weight is the probability mass remaining after adding the mixture weights "
+                        "of all the noise lists, uniformly divided among the noise lists without mixture weights. "
+                        "E.g. --noise-set-parameters '0.3, noise_list' or 'noise_list' "
+                        "the format of the noise list file is "
                         "--noise-id <string,required> --noise-type <choices = {isotropic, point source},required> "
                         "--bg-fg-type <choices = {background, foreground}, default=background> "
                         "--room-linkage <str, specifies the room associated with the noise file. Required if isotropic> "
@@ -45,6 +55,12 @@ def GetArgs():
                         help="Probability of adding point-source noises, e.g. 0 <= p <= 1")
     parser.add_argument("--isotropic-noise-addition-probability", type=float, default = 1.0,
                         help="Probability of adding isotropic noises, e.g. 0 <= p <= 1")
+    parser.add_argument("--rir-smoothing-weight", type=float, default = 0.3,
+                        help="Smoothing weight for the RIR probabilties, e.g. 0 <= p <= 1. If p = 0, no smoothing will be done. "
+                        "The RIR distribution will be mixed with a uniform distribution according to the smoothing weight")
+    parser.add_argument("--noise-smoothing-weight", type=float, default = 0.3,
+                        help="Smoothing weight for the noise probabilties, e.g. 0 <= p <= 1. If p = 0, no smoothing will be done. "
+                        "The noise distribution will be mixed with a uniform distribution according to the smoothing weight")
     parser.add_argument("--max-noises-per-minute", type=int, default = 2,
                         help="This controls the maximum number of point-source noises that could be added to a recording according to its duration")
     parser.add_argument('--random-seed', type=int, default=0, help='seed to be used in the randomization of impulses and noises')
@@ -67,13 +83,7 @@ def CheckArgs(args):
         os.makedirs(args.output_dir)
 
     ## Check arguments.
-    if not os.path.isfile(args.rir_list_file):
-        raise Exception(args.rir_list_file + " not found")
     
-    if args.noise_list_file is not None:
-        if not os.path.isfile(args.noise_list_file):
-            raise Exception(args.noise_list_file + " not found")
-
     if args.num_replicas > 1 and args.prefix is None:
         args.prefix = "rvb"
         warnings.warn("--prefix is set to 'rvb' as --num-replications is larger than 1.")
@@ -89,6 +99,12 @@ def CheckArgs(args):
 
     if args.isotropic_noise_addition_probability < 0 or args.isotropic_noise_addition_probability > 1:
         raise Exception("--isotropic-noise-addition-probability must be between 0 and 1")
+    
+    if args.rir_smoothing_weight < 0 or args.rir_smoothing_weight > 1:
+        raise Exception("--rir-smoothing-weight must be between 0 and 1")
+
+    if args.noise_smoothing_weight < 0 or args.noise_smoothing_weight > 1:
+        raise Exception("--noise-smoothing-weight must be between 0 and 1")
 
     if args.max_noises_per_minute < 0:
         raise Exception("--max-noises-per-minute cannot be negative")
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -368,11 +370,15 @@ def CreateReverberatedCopy(input_dir,
                            max_noises_per_minute  # maximum number of point-source noises that can be added to a recording according to its duration
                            ):
     
+    wav_scp = ParseFileToDict(input_dir + "/wav.scp", value_processor = lambda x: " ".join(x))
     if not os.path.isfile(input_dir + "/reco2dur"):
         print("Getting the duration of the recordings...");
-        data_lib.RunKaldiCommand("wav-to-duration --read-entire-file=true scp:{0}/wav.scp ark,t:{0}/reco2dur".format(input_dir))
+        read_entire_file="false"
```
danpovey: You are neglecting certain possibilities here:
  - What if some values use sox and some don't?
  - what if sox uses harmless commands that don't affect the header (basically, commands other than --speed)?
Look at the script utils/data/get_utt2dur.sh for inspiration.
Also, dicts in python have a values() function which may be useful here.

danpovey:
Looks OK to me; Vijay, please merge it if you're OK with it.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -439,12 +451,15 @@ def ParseRirList(rir_list_file):
...
+
     for line in rir_lines:
         rir = rir_parser.parse_args(shlex.split(line))
         rir_list.append(rir)
 
-    return SmoothProbabilityDistribution(rir_list)
+    return SmoothProbabilityDistribution(rir_list, smoothing_weight)
 
 # This dunction checks if the inputs are approximately equal assuming they are floats.
 def almost_equal(value_1, value_2, accuracy = 10**-8):
```
vijayaditya: better use sys.float_info.epsilon

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -20,14 +20,14 @@ def GetArgs():
                                                  "--random-seed 1 data/train data/train_rvb",
                                      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 
-    parser.add_argument("--rir-list-file", type=str, required = True, 
+    parser.add_argument("--rir-list-file", type=str, action='append', required = True, dest = "rir_list_file_array",
```
vijayaditya: When you want to combine two different lists with different number of items it would be preferable to assign different probability masses to each of these lists. 

e.g. real RIR list might have 500 RIRs and the sim. RIR list might have 20000 RIRs. Even in this case we would like to sample from real RIR list with 0.5 probability and sim RIR with 0.5 probability as we might believe real RIRs are more helpful. 
or 
Imagine the case where we know that test data is more likely to occur from a particular type of rooms in this case we would want to sample more RIRs from similar rooms even in training data. Assuming we can separate these RIRs into a new list we would like to assign a higher probability mass to this RIRs in this list.

So I would recommend supporting assigning different probability masses to different lists.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -368,11 +375,16 @@ def CreateReverberatedCopy(input_dir,
                            max_noises_per_minute  # maximum number of point-source noises that can be added to a recording according to its duration
                            ):
     
+    wav_scp = ParseFileToDict(input_dir + "/wav.scp", value_processor = lambda x: " ".join(x))
     if not os.path.isfile(input_dir + "/reco2dur"):
         print("Getting the duration of the recordings...");
-        data_lib.RunKaldiCommand("wav-to-duration --read-entire-file=true scp:{0}/wav.scp ark,t:{0}/reco2dur".format(input_dir))
+        read_entire_file="false"
+        for value in wav_scp.values():
+            if "sox" in value and "speed" in value:
```
vijayaditya: Add a comment here saying that we might want to check for other sox commands which modify the header as we come across these cases in our data.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -20,14 +20,14 @@ def GetArgs():
...
-    parser.add_argument("--rir-list-file", type=str, required = True, 
+    parser.add_argument("--rir-list-file", type=str, action='append', required = True, dest = "rir_list_file_array", 
                         help="RIR information file, the format of the file is "
                         "--rir-id <string,required> --room-id <string,required> "
                         "--receiver-position-id <string,optional> --source-position-id <string,optional> "
                         "--rt-60 <float,optional> --drr <float, optional> location <rspecifier> "
                         "E.g. --rir-id 00001 --room-id 001 --receiver-position-id 001 --source-position-id 00001 "
                         "--rt60 0.58 --drr -4.885 data/impulses/Room001-00001.wav")
-    parser.add_argument("--noise-list-file", type=str, default = None,
+    parser.add_argument("--noise-list-file", type=str, action='append', default = None, dest = "noise_list_file_array",
```
vijayaditya: Once again I would recommend supporting assignment of different probability masses to different lists.

tomkocse: @vijayaditya What is the user interface you prefer, what about adding an option like --rir-mass-probabilities "0.3:0.7" ?

vijayaditya: I would suggest something like
--rir-list-parameters "0.7,<filrname>"
--rir-list-parameters "0.3,<filrname>"

If no value is provided for some lists you just divide the remaining
probability mass among them.

If you want you could change the option name to be more descriptive, but I
would prefer keeping the filename and probability value together to prevent
user errors when several lists are being given. I finally hope to use this
script with ~10 different RIR databases we have access to, so this would be
a convenient format in that case.

Vijay

Vijay

vijayaditya:
@tomkocse Could you please address the comments requesting few minor changes ? I would like to merge this soon.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -404,9 +417,21 @@ def CreateReverberatedCopy(input_dir,
 
 
 # This function smooths the probability distribution in the list
-def SmoothProbabilityDistribution(list, smoothing_weight=0.3):
+def SmoothProbabilityDistribution(list, smoothing_weight=0.0, target_sum=1.0):
     if len(list) > 0:
-      uniform_probability = 1 / float(len(list))
+      num_unspecified = 0
+      accumulate_prob = 0
```
vijayaditya: `accumulated_prob`

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -404,9 +417,21 @@ def CreateReverberatedCopy(input_dir,
...
+      for item in list:
+          if item.probability is None:
+              num_unspecified += 1
+          else:
+              accumulate_prob += item.probability
+
+      # Compute the probability for the items without specifying their probability
+      if num_unspecified > 0 and accumulate_prob < 1:
+          uniform_probability = (1 - accumulate_prob) / float(num_unspecified)
+      else:
```
vijayaditya: Add a check 
```python
else if num_unspecified > 1 and accumulate_prob == 1:
       raise Exception(.....
```

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -417,15 +442,38 @@ def SmoothProbabilityDistribution(list, smoothing_weight=0.3):
       # Normalize the probability
       sum_p = sum(item.probability for item in list)
       for item in list:
-          item.probability = item.probability / sum_p
+          item.probability = item.probability / sum_p * target_sum
```
vijayaditya: why is target sum configurable, shouldn't it always be one ?

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -417,15 +442,38 @@ def SmoothProbabilityDistribution(list, smoothing_weight=0.3):
...
 
     return list
 
+
+# This function parse the array of rir set parameter strings.
+# It will assign probabilities to those rir sets which don't have a probability
+# It will also check the existence of the rir list files.
+def ParseRirParameterArray(rir_set_para_array):
```
vijayaditya: `def ParseRirParameterString` as you are parsing the strings in the array.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -438,13 +486,15 @@ def ParseRirList(rir_list_file):
...
 
+    rir_set_list = ParseRirParameterArray(rir_set_para_array)
+
     rir_list = []
-    rir_lines = map(lambda x: x.strip(), open(rir_list_file))
-    for line in rir_lines:
-        rir = rir_parser.parse_args(shlex.split(line))
-        rir_list.append(rir)
+    for rir_set in rir_set_list:
+        local_rir_list = map(lambda x: rir_parser.parse_args(shlex.split(x.strip())),open(rir_set.filename))
```
vijayaditya: `local_rir_list` --> `current_rir_list`

vijayaditya:
@vimalmanohar You would probably be one of the frequent users of this script. Could you please review the script (not just this PR) and let us know if there are any use-cases of interest, that we left out.

tomkocse:
@vijay, the above comments are addressed

egs/wsj/s5/steps/data/reverberate_data_dir.py
vijayaditya: @tomkocse you forgot about this.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -417,15 +443,38 @@ def SmoothProbabilityDistribution(list, smoothing_weight=0.3):
...
-          item.probability = item.probability / sum_p
+          item.probability = item.probability / sum_p * target_sum
 
     return list
 
+
+# This function parse the array of rir set parameter strings.
+# It will assign probabilities to those rir sets which don't have a probability
+# It will also check the existence of the rir list files.
+def ParseSetParameterString(set_para_array):
```
vijayaditya: Sorry but I think you should change it to `def ParseSetParameterStrings`

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -492,24 +543,29 @@ def ParseNoiseList(noise_list_file):
...
+                    iso_noise_dict[noise.room_linkage].append(noise)
             else:
-                if noise.room_linkage not in iso_noise_dict:
-                    iso_noise_dict[noise.room_linkage] = []
-                iso_noise_dict[noise.room_linkage].append(noise)
-        else:
-            pointsource_noise_list.append(noise)
+                current_pointsource_noise_list.append(noise)
+
+        pointsource_noise_list += SmoothProbabilityDistribution(current_pointsource_noise_list, 0.0, noise_set.probability)
```
vijayaditya: Is there a reason to not to smooth the noise distribution ?

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -14,20 +14,25 @@ def GetArgs():
...
+                        "E.g. --rir-set-parameters '0.3, rir_list' or 'rir_list' "
+                        "the format of the RIR list file is "
                         "--rir-id <string,required> --room-id <string,required> "
                         "--receiver-position-id <string,optional> --source-position-id <string,optional> "
                         "--rt-60 <float,optional> --drr <float, optional> location <rspecifier> "
                         "E.g. --rir-id 00001 --room-id 001 --receiver-position-id 001 --source-position-id 00001 "
                         "--rt60 0.58 --drr -4.885 data/impulses/Room001-00001.wav")
-    parser.add_argument("--noise-list-file", type=str, default = None,
+    parser.add_argument("--noise-set-parameters", type=str, action='append', default = None, dest = "noise_set_para_array",
                         help="Noise information file, the format of the file is"
```
vijayaditya: Update the usage message.

egs/wsj/s5/steps/data/reverberate_data_dir.py
tomkocse: @vijayaditya The probabilities are normalized so that the sum is equal to the probability of the set.

tomkocse: E.g. the probability of that set is 0.5, then the sum of probabilities of that set should be equal to 0.5

vijayaditya:
Looks good. I will merge once you confirm it is tested.

tomkocse:
@vijay, the script is tested again and you can merge it

vijayaditya:
could you please rebase and squash your commits.

PR #993, opened 2016-08-22
quick fix for kaldi online decoder crash

fanskyer:
From my experiment, in very rare cases, kaldi online decoder will crash due to last_seen_input_frames were not previously set. I made this quick fix and it can deal with this issue and decoder no longer crashed.

### DIFF
--- a/src/nnet2/nnet-compute-online.cc
+++ b/src/nnet2/nnet-compute-online.cc
@@ -31,7 +31,7 @@ NnetOnlineComputer::NnetOnlineComputer(const Nnet &nnet, bool pad_input)
   data_.resize(nnet_.NumComponents() + 1);
   reusable_component_inputs_.resize(nnet_.NumComponents()+1);
 }
-void NnetOnlineComputer::getLastSeenInputFrameLength() {
+int NnetOnlineComputer::getLastSeenInputFrameLength() {
     return last_seen_input_frame_.Dim();
 }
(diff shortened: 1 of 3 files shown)


danpovey:
Hm, thanks for finding the problem, but I don't think this is quite the right fix... can you please let me know what the crash was exactly, with maybe a stack trace?

fanskyer:
Here is one crash log, attached below, em, this quick fix is not fixing the fundamental issue 


ERROR (CopyRowsFromVec():kaldi-matrix.cc:867) Wrong sized arguments

[ Stack-Trace: ]
kaldi::MessageLogger::HandleMessage(kaldi::LogMessageEnvelope const&,
char const*)
kaldi::MessageLogger::~MessageLogger()
kaldi::MatrixBase<float>::CopyRowsFromVec(kaldi::VectorBase<float> const&)
kaldi::nnet2::NnetOnlineComputer::Flush(kaldi::CuMatrix<float>*)
kaldi::SingleUtteranceNnet2DecoderThreaded::RunNnetEvaluationInternal()
kaldi::SingleUtteranceNnet2DecoderThreaded::RunNnetEvaluation(void*)

danpovey:
Can you please apply the patch from my PR #1000 and see if it fixes the problem?

fanskyer:
sure, will test it, thanks



-----Original Message-----

PR #994, opened 2016-08-22
Replacing the default lmwt ranges in scoring scripts [both local and …

danpovey:
…global] with more reasonable values [chain systems need values <10, and no system really needs lmwt=20].


danpovey:
[note: @vince62s pointed out that this is the reason LM rescoring wasn't working for him on some chain system; rather than overriding the default in chain scripts, I think it's better to have more reasonable default values everywhere, the default range was too far to the right.]

PR #897, opened 2016-07-15
Speed up PnormComponent::Backprop

kangshiyin:
Speed up by merging cuda kernels and creating fast path for p=1,2,inf. All GPU tests passed.

```
Benchmark result (gigaflops)       dim   new     old    speedup
CuMatrix::DiffGroupPnorm<float>,    16   0.019   0.009  2.11x
CuMatrix::DiffGroupPnorm<float>,    32   0.074   0.036  2.06x
CuMatrix::DiffGroupPnorm<float>,    64   0.297   0.142  2.10x
CuMatrix::DiffGroupPnorm<float>,   128   1.142   0.520  2.20x
CuMatrix::DiffGroupPnorm<float>,   256   3.442   1.553  2.22x
CuMatrix::DiffGroupPnorm<float>,   512   6.856   2.943  2.33x
CuMatrix::DiffGroupPnorm<float>,  1024  11.653   3.915  2.98x
CuMatrix::DiffGroupPnorm<float>,  2048  13.812   4.263  3.24x
CuMatrix::DiffGroupPnorm<float>,  4096  14.431   4.381  3.29x
CuMatrix::DiffGroupPnorm<double>,    16   0.019   0.009  2.17x
CuMatrix::DiffGroupPnorm<double>,    32   0.073   0.033  2.20x
CuMatrix::DiffGroupPnorm<double>,    64   0.296   0.133  2.22x
CuMatrix::DiffGroupPnorm<double>,   128   1.068   0.457  2.34x
CuMatrix::DiffGroupPnorm<double>,   256   2.999   1.159  2.59x
CuMatrix::DiffGroupPnorm<double>,   512   4.921   1.705  2.89x
CuMatrix::DiffGroupPnorm<double>,  1024   6.932   1.993  3.48x
CuMatrix::DiffGroupPnorm<double>,  2048   7.499   2.087  3.59x
CuMatrix::DiffGroupPnorm<double>,  4096   7.684   2.104  3.65x
```

PR #1003, opened 2016-08-24
Fix bugs with random stride mode

kangshiyin:
Fix random stride mode bug and typo mentioned in #997 

Fix stride bug in `_calc_group_max_deriv` detected by random stride mode.

Fix stride bug in `_calc_group_pnomr_deriv`. Random stride mode did not detect this but it uses the same code template as `_calc_group_max_deriv`. As `CuMatrixBase::GroupPnormDeriv` has been replaced by `CuMatrixBase::DiffGroupPnorm` in #897. This bug did not affect NNet3/2 since then.

All GPU tests have passed.

### DIFF
--- a/src/cudamatrix/cu-device.h
+++ b/src/cudamatrix/cu-device.h
@@ -56,7 +56,9 @@ class CuDevice {
 
   inline void* MallocPitch(size_t row_bytes, size_t num_rows, size_t *pitch) {
     if (random_stride_mode_) {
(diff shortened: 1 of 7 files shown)


danpovey:
@kangshiyin, there are tests that use CUDA in the chain/ directory as well, you should probably check those too.

kangshiyin:
Random stride mode does not guarantee that the bug can be detected.  I will take a look.

PR #1005, opened 2016-08-24
Extract wav - perturb_data_dir_speed.sh implementation

psmit:
See PR #1004 . This script wraps anything that was in the scp file in an extract-wav call. Care is been taken that quoting is done correctly.

### DIFF
--- a/egs/wsj/s5/utils/perturb_data_dir_speed.sh
+++ b/egs/wsj/s5/utils/perturb_data_dir_speed.sh
@@ -74,9 +74,11 @@ if [ -f $srcdir/segments ]; then
         '{printf("%s %s %.2f %.2f\n", $1, $2, $3/factor, $4/factor);}' >$destdir/segments
 
   utils/apply_map.pl -f 1 $destdir/reco_map <$srcdir/wav.scp | sed 's/| *$/ |/' | \
+    # Handle three cases of rxfilenames appropriately; "input piped command", "file offset" and "filename" 
     awk -v factor=$factor \
         '{wid=$1; $1=""; if ($NF=="|") {print wid $_ " sox -t wav - -t wav - speed " factor " |"}
-          else {print wid " sox -t wav" $_ " -t wav - speed " factor " |"}}' > $destdir/wav.scp
+          else if (match($0, /:[0-9]+$/)) {print wid " wav-copy" $_ " - | sox -t wav - -t wav - speed " factor " |" } 
+          else  {print wid " sox -t wav" $_ " -t wav - speed " factor " |"}}' > $destdir/wav.scp
   if [ -f $srcdir/reco2file_and_channel ]; then
     utils/apply_map.pl -f 1 $destdir/reco_map <$srcdir/reco2file_and_channel >$destdir/reco2file_and_channel
   fi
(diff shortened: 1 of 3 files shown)


danpovey:
Oh OK, I see from your post on kaldi-developers what the problem was.
You say:
<quote>
 am trying to run the tedlium s5_r2 chain scripts on my own data. A problem is however that I always copy my wav-files in a wav-archive with wav-copy, as that is preferred for the type of filesystem we have (lustre), compared to keeping single files around.

The problem is that this chain scripts do wav perturbation by changing the scp file to do "sox -t wav wav.ark:1029386 -t wav - speed 0.9 |". This will not work as sox can't read the wav from the ark file.

Is there any standard solution for this? I was looking around, but I could not find any command that extracts the wav-file to a pipe.
</quote>

Instead of adding the new binary extract-wav, it would be better if you modify the program wav-copy to accept rxfilenames/wxfilenames.  Please see 'copy-matrix' as a template of a program that accepts both single-file and table input and output.  

And rather than including this in all cases, it would be better if you detect that the input is an offset into an archive [it may involve a little research into regexp's in awk], and only in that case include wav-copy in the output.  Otherwise it's wasteful-- yours is a rather unusual use-case.

psmit:
I'm actually surprised that it is such an unusual case. Our system admins loathed our group for always having millions of files around and nowadays they are happier :)

I implemented the changes you suggested. Happy to get more feedback on it!

danpovey:
I'd say that says more about the deficiencies of Lustre than anything
else.  Wav files are normally large enough that the metadata-related
operations should take very little time compared to the actual file reading.

Dan

psmit:
True about that, it is one downside of having separate metadata servers such as Lustre has. Anyway, it is good that Kaldi supports both (in a very elegant way). 

If you know any more scripts that could have this issue (besides the perturb scripts), I'm happy to fix those as well.

danpovey:
I can't think of any right now..

src/featbin/wav-copy.cc
```diff
@@ -26,35 +27,69 @@ int main(int argc, char *argv[]) {
...
+        "     wav-copy wav.ark:123456 \n"
         "See also: wav-to-duration extract-segments\n";
-    
+
     ParseOptions po(usage);
 
     po.Read(argc, argv);
 
-    if (po.NumArgs() != 2) {
+    if (po.NumArgs() < 1 || po.NumArgs() > 2) {
```
danpovey: please don't make the 2nd arg optional- for Kaldi programs it's not traditional to have optional args like this, so it could be confusing.

egs/wsj/s5/utils/perturb_data_dir_speed.sh
```diff
@@ -76,7 +76,8 @@ if [ -f $srcdir/segments ]; then
   utils/apply_map.pl -f 1 $destdir/reco_map <$srcdir/wav.scp | sed 's/| *$/ |/' | \
     awk -v factor=$factor \
         '{wid=$1; $1=""; if ($NF=="|") {print wid $_ " sox -t wav - -t wav - speed " factor " |"}
-          else {print wid " sox -t wav" $_ " -t wav - speed " factor " |"}}' > $destdir/wav.scp
+          else if (match($0, /:[0-9]+$/)) {print wid " wav-copy" $_ " | sox -t wav - -t wav - speed " factor " |" }
```
danpovey: it might be a good to add a comment  a few lines above, explaining why there need to be 3 cases.

psmit: Ok, I'll do that

src/featbin/wav-copy.cc
psmit: I only made the second argument optional for the new use case. I thought there are actually quite a couple of kaldi programs that allow to leave out the last arguments (source `find . -name "*.cc" | xargs grep "NumArgs() >"`), or does it have a different meaning in those cases? At least the first on the list, bin/add-self-loops.cc, has the same behaviour that a left out last argument is assumed to be stdout?

I do see that the coding style is normally a small bit different, with calling po.GetArg(n) anyway, even if n doesn't exist.

danpovey: It's only older programs that interpret empty arguments as stdout.  It's
not a pattern we're following any more- making them explicit makes the
scripts easier to interpret.

psmit: Ok, I'll change this as well. Should I squash this complete PR into one new commit?

danpovey: Yes, please squash.

psmit:
I have tested this now in multiple scenarios (with/without segments, pipes,filenames,fileoffsets) and I believe the code is ready to be merged

danpovey:
Thanks!  Merging.

PR #1006, opened 2016-08-26
Improved debug stride mode

kangshiyin:
This is a follow up of #1003. A counter ensues that adjacent allocations have different pitches so that the stride bugs are more likely to be detected in this debug stride mode. The mode is enabled in many CUDA tests including chain, nnet2 and nnet3 except for `nnet2/nnet-component-test`, as `nnet2::DropoutComponent` can not pass the test with unknown reason.
All CUDA tests have passed after disabling debug stride mode in `nnet2/nnet-component-test`.

### DIFF
--- a/src/cudamatrix/cu-device.h
+++ b/src/cudamatrix/cu-device.h
@@ -55,11 +55,14 @@ class CuDevice {
   inline void* Malloc(size_t size) { return allocator_.Malloc(size); }
 
   inline void* MallocPitch(size_t row_bytes, size_t num_rows, size_t *pitch) {
-    if (random_stride_mode_) {
+    if (debug_stride_mode_) {
       // The pitch bucket size is hardware dependent.
       // It is 512 on K40c with CUDA 7.5
-      return allocator_.MallocPitch(row_bytes + 512 * (rand() & 1), num_rows,
-                                    pitch);
+      // "% 8" ensures that any 8 adjacent allocations have different pitches
+      // if their original pitches are same in the normal mode.
+      return allocator_.MallocPitch(
+          row_bytes + 512 * ((num_debug_stride_allocations_++) % 8), num_rows,
+          pitch);
     } else {
       return allocator_.MallocPitch(row_bytes, num_rows, pitch);
     }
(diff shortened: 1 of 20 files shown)


src/nnet2/nnet-component-test.cc
```diff
@@ -859,6 +859,8 @@ int main() {
 
   for (int32 loop = 0; loop < 2; loop++) {
 #if HAVE_CUDA == 1
+    //// Uncomment the following line to expose the bug in UnitTestDropoutComponent
+    //CuDevice::Instantiate().SetDebugStrideMode(true);
```
kangshiyin: Uncomment this line to expose the bug in `UnitTestDropoutComponent`.

src/cudamatrix/cu-device.h
```diff
@@ -55,11 +55,14 @@ class CuDevice {
...
-    if (random_stride_mode_) {
+    if (debug_stride_mode_) {
       // The pitch bucket size is hardware dependent.
       // It is 512 on K40c with CUDA 7.5
-      return allocator_.MallocPitch(row_bytes + 512 * (rand() & 1), num_rows,
-                                    pitch);
+      // "% 8" ensures that any 8 adjacent allocations have different pitches
+      // if their original pitches are same in the normal mode.
+      return allocator_.MallocPitch(
+          row_bytes + 512 * ((num_debug_stride_allocations_++) % 8), num_rows,
```
danpovey: The way you are doing it here will use a potentially very large amount of memory, because it's using on average 512 * 4 extra bytes per row, and what if it's a matrix with a lot of rows?  Does it have to be % 8?

src/nnet2/nnet-component-test.cc
danpovey: It looks like there is no test code for AddMatMatDivMat.  This is used in the dropout component; it could have a bug.

src/cudamatrix/cu-device.h
contributeyourvoice: On Fri, Aug 26, 2016 at 2:15 PM, Daniel Povey <notifications@github.com>
wrote:

> In src/cudamatrix/cu-device.h
> <https://github.com/kaldi-asr/kaldi/pull/1006#discussion_r76368734>:
>
> >        // The pitch bucket size is hardware dependent.
> >        // It is 512 on K40c with CUDA 7.5
> > -      return allocator_.MallocPitch(row_bytes + 512 * (rand() & 1), num_rows,
> > -                                    pitch);
> > +      // "% 8" ensures that any 8 adjacent allocations have different pitches
> > +      // if their original pitches are same in the normal mode.
> > +      return allocator_.MallocPitch(
> > +          row_bytes + 512 * ((num_debug_stride_allocations_++) % 8), num_rows,
>
> The way you are doing it here will use a potentially very large amount of
> memory, because it's using on average 512 * 4 extra bytes per row, and what
> if it's a matrix with a lot of rows? Does it have to be % 8?
>
Maybe  4, 5 is large enough, but the larger the better for detection.
Largest tests currently use a few hundreds rows, resulting in a few MB of
extra memory per matrix. I thought it is affordable.

src/nnet2/nnet-component-test.cc
kangshiyin: It's code looks ok but I can make a test later.

danpovey:
ok will merge when checks finish

PR #1008, opened 2016-08-27
Add unit test for CuMatrix::AddMatMatDivMat

kangshiyin:
All GPU tests have passed.

### DIFF
--- a/src/cudamatrix/cu-matrix-test.cc
+++ b/src/cudamatrix/cu-matrix-test.cc
@@ -1111,6 +1111,26 @@ template<typename Real> static void UnitTestCuMatrixAddMatMatElements() {
   KALDI_ASSERT(M.Sum() != 0.0);
 }
 
+template<typename Real> static void UnitTestCuMatrixAddMatMatDivMat() {
+  // M = a * b / c (by element; when c = 0, M = a)
+  MatrixIndexT dimM = 100 + Rand() % 255, dimN = 100 + Rand() % 255;
+  CuMatrix<Real> M(dimM, dimN), A(dimM, dimN), B(dimM, dimN), C(dimM, dimN);
+  CuMatrix<Real> ref(dimM, dimN);
+  M.SetRandn();
+  A.SetRandn();
+  B.SetRandn();
+  C.SetRandn();
+
+  M.AddMatMatDivMat(A,B,C);
+  ref.AddMatMatElements(1.0, A, B, 0.0);
+  ref.DivElements(C);
+  AssertEqual(M, ref);
+
+  C.SetZero();
+  M.AddMatMatDivMat(A,B,C);
+  AssertEqual(M, A);
+}
+
 template<typename Real>
 static void UnitTestCuMatrixDivRowsVec() {
   MatrixIndexT dimM = 1000, dimN = 5;
@@ -2642,6 +2662,7 @@ template<typename Real> void CudaMatrixUnitTest() {
   UnitTestCuMatrixAddDiagVecMat<Real>();
   UnitTestCuMatrixAddMatDiagVec<Real>();
   UnitTestCuMatrixAddMatMatElements<Real>();
+  UnitTestCuMatrixAddMatMatDivMat<Real>();
   UnitTestCuTanh<Real>();
   UnitTestCuCholesky<Real>();
   UnitTestCuDiffTanh<Real>();


galv:
Looks good to me.

Note it's not documented whether or not this function is safe to call when "this" is one of "a", "b", or "c", (in case some code requires accumulating this computation multiple times), though it seems like it should be given that the computation is element-wise. Up to you to decide if you want to dig into that.

src/cudamatrix/cu-matrix-test.cc
```diff
@@ -1111,6 +1111,26 @@ template<typename Real> static void UnitTestCuMatrixAddMatMatElements() {
...
+  MatrixIndexT dimM = 100 + Rand() % 255, dimN = 100 + Rand() % 255;
+  CuMatrix<Real> M(dimM, dimN), A(dimM, dimN), B(dimM, dimN), C(dimM, dimN);
+  CuMatrix<Real> ref(dimM, dimN);
+  M.SetRandn();
+  A.SetRandn();
+  B.SetRandn();
+  C.SetRandn();
+
+  M.AddMatMatDivMat(A,B,C);
+  ref.AddMatMatElements(1.0, A, B, 0.0);
```
danpovey: There is a slight theoretical issue here, that the function AddMatMatDivMat has special behavior if an element of c is exactly 0.  The prob. that this happens is very small, of course.  we could just ignore it, but put a comment...

danpovey:
BTW, I discovered the reason why the dropout component test was failing.
It has to do with the way the CuRand class works; the random numbers depend
on the stride, and calling srand(0) and then SeedGpu isn't enough to
guarantee the same numbers if the stride is different.
But this won't matter in real applications.

kangshiyin:
On Sun, Aug 28, 2016 at 2:19 AM, Daniel Galvez <notifications@github.com>
wrote:

> Looks good to me.
>
> Note it's not documented whether or not this function is safe to call when
> "this" is one of "a", "b", or "c", (in case some code requires accumulating
> this computation multiple times), though it seems like it should be given
> that the computation is element-wise. Up to you to decide if you want to
> dig into that.
>
It is safe. We can add a comment.

src/cudamatrix/cu-matrix-test.cc
kangshiyin: On Sun, Aug 28, 2016 at 2:53 AM, Daniel Povey <notifications@github.com>
wrote:

> In src/cudamatrix/cu-matrix-test.cc
> <https://github.com/kaldi-asr/kaldi/pull/1008#discussion_r76520755>:
>
> > @@ -1111,6 +1111,26 @@ template<typename Real> static void UnitTestCuMatrixAddMatMatElements() {
> >    KALDI_ASSERT(M.Sum() != 0.0);
> >  }
> >
> > +template<typename Real> static void UnitTestCuMatrixAddMatMatDivMat() {
> > +  // M = a * b / c (by element; when c = 0, M = a)
> > +  MatrixIndexT dimM = 100 + Rand() % 255, dimN = 100 + Rand() % 255;
> > +  CuMatrix<Real> M(dimM, dimN), A(dimM, dimN), B(dimM, dimN), C(dimM, dimN);
> > +  CuMatrix<Real> ref(dimM, dimN);
> > +  M.SetRandn();
> > +  A.SetRandn();
> > +  B.SetRandn();
> > +  C.SetRandn();
> > +
> > +  M.AddMatMatDivMat(A,B,C);
> > +  ref.AddMatMatElements(1.0, A, B, 0.0);
>
> There is a slight theoretical issue here, that the function
> AddMatMatDivMat has special behavior if an element of c is exactly 0. The
> prob. that this happens is very small, of course. we could just ignore it,
> but put a comment...
>
I think the comment about `c=0` is already there. Line 1115 and in
`cu-matrix.h` .

kangshiyin:
On Sun, Aug 28, 2016 at 3:22 AM, Daniel Povey <notifications@github.com>
wrote:

> BTW, I discovered the reason why the dropout component test was failing.
> It has to do with the way the CuRand class works; the random numbers depend
> on the stride, and calling srand(0) and then SeedGpu isn't enough to
> guarantee the same numbers if the stride is different.
> But this won't matter in real applications.
>
> I'm thinking that maybe we can replace the kernels in cu-rand with the
official cuRAND in CUDA, which does not need a long time to generate the
seed.

danpovey:
It looks like using cuRAND means linking against a new, differently-named
library.
I'm concerned that it might cause installation problems in some cases.

Dan

PR #1009, opened 2016-08-27
Fix oracle calculation for nnet3 scripts

danijel3:
nnet3 decode scripts use a different split to other (GMM) decode scripts: --per-utt and they append "utt" to their split subdirs. Oracle WER calculations use regular split dirs and therefore miscalculate results (there are many missing files in results).

Easy solution - simply create a copy of the oracle script in the nnet3 dir that has the same split as the decode.


danpovey:
Hm.
This is actually a problem, and it should have occurred to me when we made
the nnet3 scripts decode per utterance.  In fact, it's a problem for
anything that does per-utterance splitting and whose output might later be
needed by something that cares about the split.  This includes things like
lattice combination.    @vijayaditya, see if anything in BABEL might be
affected by this.
However, your solution isn't very nice as it doesn't address the underlying
issue, which is that the lattices defy the expectations of other scripts.
I am leaning towards just making the nnet3 scripts decode per speaker
again.  In any case, in recent example scripts (tedlium s5_r2 and ami s5b)
I have been splitting up the speakers into chunks of fixed size so none of
the speakers that appear in the spk2utt are very long.  This solves the
issue that we couldn't use a large num-jobs for nnet3 decoding,.


Dan

danijel3:
Yea, I agree this is a quick, dirty and ungly solution. I wasn't aware that other parts were affected as well. Just noticed it on an experiment I was running. Maybe there is something that could be added to the decode dir to hint at the split method used? Something like a split_options file? Just a thought...

vijayaditya:
I have used per-utterance split just in ASpIRE and Babel and both of these use sclite scoring of ctms at the recording level.

danpovey:
A lot of things would become more complex if we made the split_options part
of the directory format-- it would potentially have to be interpreted by
through many scripts.

The easiest solution is to revert the nnet3 decoding script so it splits
per speaker. Vijay, you could possibly use
utils/data/modify_speaker_info.sh to split speakers in ASPIRE and Babel
into smaller pieces in the data-prep stage.  (maybe not Babel though-- it's
a complicated setup and something might break).

vijayaditya:
OK, reverting to old code is not an issue for me. Creating pseudo speaker
is desirable for other reasons in ASpIRE.

--Vijay

danpovey:
OK. Danijel, would you mind creating new PR where you modify the nnet3
decoding script to take away the --per-utt option and related changes?
Dan

danijel3:
Sure. I'll modify test and create the PR in a sec.

There are 2 decode scripts, btw: one in nnet3, and another in nnet3/lstm

Also make_denlats.sh splits on --per-utts as well. Want me to fix it there too?

Just noticed a funny thing, the local/nnet3/run_lstm.sh script set nj for decode to equal the number of speakers in the test set. Wouldn't that pose as a problem for some datasets?

vijayaditya:
The decode script in nnet3/lstm is deprecated. We are not using it in any
of the recipes as nnet3/decode.sh subsumes its functionality.

--Vijay

danijel3:
It's used at the end of https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/local/nnet3/run_lstm.sh

Want me to remove it from there?

vijayaditya:
Yes please.

--Vijay

danijel3:
I added the changes. I also removed all the other occurances of "lstm/decode.sh" from egs. To me, it looks like it should be safe, but I made it a separate commit, just in case...

danpovey:
and the denlats script only uses --per-utt for sub-splits, it isn't a
problem.

danpovey:
Thanks, merging.

Issue #901, opened 2016-07-18
Error on speaker recognition examples


RobertoFont:
Hello,

I think I've found an error on sre scripts. When performing cosine scoring, for example

```
cat $trials | awk '{print $1, $2}' | \
 ivector-compute-dot-products - \
  scp:exp/ivectors_sre08_train_short2_female/spk_ivector.scp \
  scp:exp/ivectors_sre08_test_short3_female/ivector.scp \
   foo
```
speaker-level ivectors (spk_ivector.scp) are lenght-normalized, but utterance-level ivectors (ivector.scp) are not. This way, dot products are not actually cosine distance scores.

Test ivectors should be piped into ivector-normalize-length or ivector.scp replaced by spk_ivector.scp if each utterance corresponds to a different speaker as is usually the case in test set.

Let me know if I'm wrong.

Regards.

danpovey:
@david-ryan-snyder, did this slip by you?

david-ryan-snyder:
@danpovey, @RobertoFont,

Training some SRE08 models for this now. I'll apply the fix and update the inline comments with the results once that's done.

david-ryan-snyder:
Since the fix is in Kaldi now, you can close this, @danpovey.

@RobertoFont, FYI

PR #1016, opened 2016-09-02
Updates to SRE08 example

david-ryan-snyder:
Updates:
 - length-normalization is now applied to all test i-vectors in egs/sre08/v1/run.sh. This fixes the cosine scoring issue (https://github.com/kaldi-asr/kaldi/issues/901). I think the same problem exists in the PLDA scoring: the test i-vectors should also be length-normalized, since the speaker-level i-vectors are, so I added it there as well. Also updated the results. The condition 0 (pooled) results don't change too much.

- Updated egs/sre08/v1/README to refer to all datasets used

- Improved style of run.sh so that 2 spaces are used for indentation; removed trailing whitespace

### DIFF
--- a/egs/sre08/v1/README.txt
+++ b/egs/sre08/v1/README.txt
@@ -1,13 +1,14 @@
-
-
-
  Data required for system development (on top of the data for testing described
- in ../README.txt), is Fisher parts 1 and 2 (you could probably get by OK with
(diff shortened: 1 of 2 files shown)


david-ryan-snyder:
@danpovey, FYI when you have the time.

PR #1018, opened 2016-09-06
nnet1: minor cosmetic change,

vesis84:


danpovey:
FYI, we normally use:
` parser = argparse.ArgumentParser( ... formatter_class=argparse.ArgumentDefaultsHelpFormatter)`
so the defaults are printed naturally.  But merging.

vesis84:
Okay, thanks!

PR #1022, opened 2016-09-09
Storing phones.txt and add check_phones_compatible.sh

LvHang:
@danpovey 
I have fininshed the shell script-- utils/lang/check_phones_compatible.sh. Firstly, this script will check the arguments in different conditions. Then, This script can judge whether the two phone symbol tables are same or not (except for possible differences in disambiguation symbols). 
I have tested it with four different phones.txt. (swbd_phones.txt; swbd_phones_withoutdisam.txt; swbd_phones_fewdisam.txt and timit_phones.txt). All  branches of the script were tested. Everything goes smoothly.

After that, I traverse all the files in wsj/s5/steps 
and add the command
utils/lang/check_phones_compatible.sh $lang/phones.txt $dir/phones.txt if set -e was done,
or utils/lang/check_phones_compatible.sh $lang/phones.txt $dir/phones.txt || exit 1 otherwise.
if any scripts that take a lang directory and a model or alignment directory.

At the same time, I will copy the phones.txt to the new directory if the scripts create new directory.
The command looks like: cp $alidir/phones.txt $dir || exit 1;

I'm not sure if this can accord with your standard. Thank you for your advice and guidance.
Hang

### DIFF
--- a/egs/wsj/s5/utils/lang/check_phones_compatible.sh
+++ b/egs/wsj/s5/utils/lang/check_phones_compatible.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# Copyright 2016 Hang Lyu
+
+# Licensed udner the Apache License, Version 2.0 (the "Lincense");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, EITHER EXPRESS OF IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+# WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+# MERCHANTABLITY OR NON-INFRINGEMENT.
+# See the Apache 2 License for the specific language governing permissions and
+# limitations under the License.
+
+# This script exits with status zero if the phone symbols tables are the same
+# except for possible differences in disambiguation symbols (meaning that all
+# symbols except those beginning with a # are mapped to the same values).
+# Otherwise it prints a warning and exits with status 1.
+# For the sake of compatibility with other scripts that did not write the 
+# phones.txt to model directories, this script exits silently with status 0 
+# if one of the phone symbol tables does not exist.
+# For the sake of compatibility with other scripts that did not write the 
+# phones.txt to model directories, this script exits silently with status 0 
+# if one of the phone symbol tables does not exist.
+
+. utils/parse_options.sh || exit 1;
+
+if [ $# -ne 2 ]; then
+  echo "Usage: utils/lang/check_phones_compatible.sh <phones-symbol-table1> <phones-symbol-table2>"
+  echo "e.g.: utils/lang/check_phones_compatible.sh data/lang/phones.txt exp/tri3/phones.txt"
+  exit 1;
+fi
+
+table_first=$1
+table_second=$2
+
+# check the files exist or not 
+if [ ! -f $table_first ]; then
+  if [ ! -f $table_second ]; then
+    echo "$0: Error! Both of the two phones-symbol tables are absent."
+    echo "Please check your command"
+    exit 1;
+  else
+    #The phones-symbol-table1 is absent. The model directory maybe created by old script.
+    #For back compatibility, this script exits silently with status 0.
+    exit 0;
+  fi
+elif [ ! -f $table_second ]; then
+  #The phones-symbol-table2 is absent. The model directory maybe created by old script.
+  #For back compatibility, this script exits silently with status 0.
+  exit 0;
+fi
+
+#Check the two tables are same or not (except for possible difference in disambiguation symbols).
+if ! cmp -s <(grep -v "^#" $table_first) <(grep -v "^#" $table_second); then
+  echo "$0: phone symbol tables $table_first and $table_second are not compatible."
+  exit 1;
+fi
+
+exit 0;
--- a/egs/wsj/s5/steps/align_basis_fmllr.sh
+++ b/egs/wsj/s5/steps/align_basis_fmllr.sh
@@ -57,6 +57,9 @@ mkdir -p $dir/log
 echo $nj > $dir/num_jobs
 [[ -d $sdata && $data/feats.scp -ot $sdata ]] || split_data.sh $data $nj || exit 1;
 
+utils/lang/check_phones_compatible.sh $lang/phones.txt $srcdir/phones.txt || exit 1;
+cp $srcdir/phones.txt $dir || exit 1;
+
 cp $srcdir/{tree,final.mdl} $dir || exit 1;
 cp $srcdir/final.occs $dir;
 splice_opts=`cat $srcdir/splice_opts 2>/dev/null` # frame-splicing options.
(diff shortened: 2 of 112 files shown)


danpovey:
thanks, I will have to get to this next week, after Interspeech.

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# Copyright 2016 Johns Hopkins University
```
danpovey: you should put your own name here, and not  Johns Hopkins University.

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,68 @@
...
+  exit 1;
+fi
+
+table_first=$1
+table_second=$2
+
+# check the files exist or not 
+if [ ! -f $table_first ]; then
+  if [ ! -f $table_second ]; then
+    echo "Error! Both of the two phones-symbol tables are absent."
```
danpovey: messages printed by this command should start with `"$0:"`

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,68 @@
...
+  echo "The phones-symbol-table2 is absent. The model directory maybe created by old script."
+  echo "For back compatibility, this script exits silently with status 0."
+  exit 0;
+fi
+
+#Check the two tables are same or not (except for possible difference in disambiguation symbols).
+tmpdir=$(mktemp -d /tmp/kaldi.XXXX);
+trap 'rm -rf "$tmpdir"' EXIT
+
+grep -v "^#" $table_first | sort > $tmpdir/table1.txt
```
danpovey: no need to sort, they will already be sorted.

LvHang:
@danpovey 
Ok, I will fix it when I come back to lab. Maybe two days later. Now I'm apply for the visa in Beijing.
Hang

LvHang:
@danpovey 
I have fixed the utils/lang/check_phones_compatible.sh according to your suggestion.
Please check it. Thanks.
Hang

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,68 @@
...
+  echo "usage: utils/lang/check_phones_compatible.sh <phones-symbol-table1> <phones-symbol-table2>"
+  echo "e.g.: utils/lang/check_phones_compatible.sh data/lang/phones.txt exp/tri3/phones.txt"
+  exit 1;
+fi
+
+table_first=$1
+table_second=$2
+
+# check the files exist or not 
+if [ ! -f $table_first ]; then
```
danpovey: *all* messages printed by this script should start with "$0:".
Also, I think it's best if the script just continues silently unless both phone symbol tables exist.  Otherwise it will create too many warnings when people use older directories.

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,68 @@
...
+    exit 0;
+  fi
+elif [ ! -f $table_second ]; then
+  echo "The phones-symbol-table2 is absent. The model directory maybe created by old script."
+  echo "For back compatibility, this script exits silently with status 0."
+  exit 0;
+fi
+
+#Check the two tables are same or not (except for possible difference in disambiguation symbols).
+tmpdir=$(mktemp -d /tmp/kaldi.XXXX);
```
danpovey: You can simplify this script to avoid the use of tmpdir, by using process substitution.
You can do 
`cmp <(grep -v "^#" $table_first) <(grep -v "^#" $table_second)`

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,68 @@
...
+trap 'rm -rf "$tmpdir"' EXIT
+
+grep -v "^#" $table_first > $tmpdir/table1.txt
+grep -v "^#" $table_second > $tmpdir/table2.txt
+
+if ! cmp $tmpdir/table1.txt $tmpdir/table2.txt; then
+  echo "Error! The phone symbols tables are different.(except for possible differences in disambiguation symbols)"
```
danpovey: A better error message would be:
   `echo "$0: phone symbol tables $table_first and $table_second are not compatible."`

LvHang:
@danpovey 
Now, all messages printed by this script should start with "$0:" and the error message is updated.
When one of the  phone symbol tables is absent, the script will exits silently(no echo) with status 0.
tmpdir is not used.
Please check it. Thanks.
Hang

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,62 @@
...
+    exit 0;
+  fi
+elif [ ! -f $table_second ]; then
+  #The phones-symbol-table2 is absent. The model directory maybe created by old script.
+  #For back compatibility, this script exits silently with status 0.
+  exit 0;
+fi
+
+#Check the two tables are same or not (except for possible difference in disambiguation symbols).
+if ! (cmp -s <(grep -v "^#" $table_first) <(grep -v "^#" $table_second)) ; then
```
danpovey: there is no need for the parenthesis (subshell) around cmp.

egs/wsj/s5/utils/lang/check_phones_compatible.sh
```diff
@@ -0,0 +1,62 @@
...
+# phones.txt to model directories, this script exits silently with status 0 
+# if one of the phone symbol tables does not exist.
+# For the sake of compatibility with other scripts that did not write the 
+# phones.txt to model directories, this script exits silently with status 0 
+# if one of the phone symbol tables does not exist.
+
+. utils/parse_options.sh || exit 1;
+
+if [ $# -ne 2 ]; then
+  echo "usage: utils/lang/check_phones_compatible.sh <phones-symbol-table1> <phones-symbol-table2>"
```
danpovey: Capital U in "usage".

LvHang:
@danpovey 
I have removed the parenthesis around cmp and uppercase U in "usage".
Thank you very much for your patience guidance. I really need more practise.
Thanks
Hang

danpovey:
Thanks!  Merging.

vesis84:
Hi, sorry for complaining, but it seems like this pull request has caused a backward compatibility problem for older directories (models / alignments / denlats). This causes troubles to all the people, who already have some model and want to continue experimenting...
To make it backward compatible, the solution is to first check if the file 'phones.txt' exists, and only then do the comparison.
K.

danpovey:
The script actually does check first.  The only case it complains is when
*both* phones.txt are missing, which we don't expect because one always
comes from a lang/ directory (IIRC).
Can you be specific about the situation where this causes a problem?

mallidi:
@danpovey 
I faced an error while running steps/nnet/train.sh . This is due to the line number 128 
 cp $alidir/phones.txt $dir 
If $alidir is generated by models prior to merging phones.txt do not exist. Probably if we copy from $lang dir , it should fix (atleast for this script, didn't check other scripts)

danpovey:
Yes, it looks like a lot of commands were added like
`cp $alidir/phones.txt $dir || exit 1;`
and these will all fail if the old dir doesn't have those things.
@LvHang, can you please change all of these to copy it from the $lang
directory instead (making sure, of course, that the $lang directory
actually exists in the script)?
I reverted the patch so others don't get this problem.

Dan

PR #1023, opened 2016-09-10
Speed up LogSoftmaxComponent::Backprop

kangshiyin:
Speed up by merging multiple kernels into one.
All GPU tests have passed.

```
Benchmark (gflops)                      dim  old     new    speedup
CuMatrix::DiffLogSoftmaxPerRow<float>,   16  0.0022  0.0153 7.03x
CuMatrix::DiffLogSoftmaxPerRow<float>,   32  0.0087  0.0577 6.66x
CuMatrix::DiffLogSoftmaxPerRow<float>,   64  0.0353  0.2678 7.59x
CuMatrix::DiffLogSoftmaxPerRow<float>,  128  0.1347  0.8785 6.52x
CuMatrix::DiffLogSoftmaxPerRow<float>,  256  0.4920  2.8799 5.85x
CuMatrix::DiffLogSoftmaxPerRow<float>,  512  1.3416  6.2052 4.63x
CuMatrix::DiffLogSoftmaxPerRow<float>, 1024  2.4438 10.4197 4.26x
CuMatrix::DiffLogSoftmaxPerRow<double>,   16  0.0019  0.0140 7.22x
CuMatrix::DiffLogSoftmaxPerRow<double>,   32  0.0073  0.0573 7.83x
CuMatrix::DiffLogSoftmaxPerRow<double>,   64  0.0282  0.1971 6.98x
CuMatrix::DiffLogSoftmaxPerRow<double>,  128  0.1113  0.7518 6.75x
CuMatrix::DiffLogSoftmaxPerRow<double>,  256  0.3945  2.4320 6.16x
CuMatrix::DiffLogSoftmaxPerRow<double>,  512  0.9307  4.5303 4.87x
CuMatrix::DiffLogSoftmaxPerRow<double>, 1024  1.5232  5.4336 3.57x
```

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -2599,6 +2599,62 @@ static void _diff_softmax(Real* x, const MatrixDim dim, const Real* value,
   }
 }
 
+// Differentiate backward through the log softmax function.
+// "out_value" is the log softmax output. Does, for each row i,
+// in_deriv(i) =  out_deriv(i) - sum(out_deriv(i)) .* exp(out_value(i))
+// ???(i) is row-vector.
+// CUDA thread layout: 1 thread block (CU1DBLOCK == 256 threads) per matrix-row.
+template<typename Real>
+__global__
+static void _diff_log_softmax(const MatrixDim in_deriv_dim,
+                              const Real* out_value, const int out_value_stride,
+                              const Real* out_deriv, const int out_deriv_stride,
+                              Real* in_deriv) {
+
+  __shared__ Real ssum[CU1DBLOCK];
+  const int tid = threadIdx.x;
+  const int i = blockIdx.x;
+  const int out_value_start = i * out_value_stride;
+  const int out_deriv_start = i * out_deriv_stride;
+  const int in_deriv_start = i * in_deriv_dim.stride;
+
+  // Loop along the matrix row. Reduce to CU1DBLOCK elements per row.
+  Real tsum = Real(0);
+  for (int j = tid; j < in_deriv_dim.cols; j += CU1DBLOCK) {
+    tsum += out_deriv[out_deriv_start + j];
+  }
+  ssum[tid] = tsum;
+  __syncthreads();
+
+  // Tree reduce to 2x warpSize elements.
+# pragma unroll
+  for (int shift = CU1DBLOCK / 2; shift > warpSize; shift >>= 1) {
+    if (tid < shift) {
+      ssum[tid] += ssum[tid + shift];
+    }
+    __syncthreads();
+  }
+
+  // Warp reduce to 1 element. Threads implicitly synchronized within a warp.
+  if (tid < warpSize) {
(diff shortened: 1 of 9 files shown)


src/cudamatrix/cu-kernels.cu
```diff
@@ -2599,6 +2599,60 @@ static void _diff_softmax(Real* x, const MatrixDim dim, const Real* value,
...
 }
 
+// Differentiate backward through the log softmax function.
+// "value" is the log softmax output. Does, for each row i,
+// x(i) =  diff(i) - sum(diff(i)) .* exp(value(i))
+// ???(i) is row-vector.
+// CUDA thread layout: one thread block (256 threads) per matrix-row.
+template<typename Real>
+__global__
+static void _diff_log_softmax(Real* x, const MatrixDim dim, const Real* value,
```
galv: style comment: I dislike using the name "value" here, even though you document it as the output of log softmax forward propagation. I think "output_value" and "output_value_stride" would be much better names.

galv:
Looks good to me. I like that you separated out your work into separate commits, but they should probably be squashed before being merged into master.

galv:
@kangshiyin By the way, did you make sure to run the nnet3 and nnet2 tests to make sure that gradient checks passed?

src/cudamatrix/cu-kernels.cu
```diff
@@ -2552,6 +2554,11 @@ static void _diff_xent(const int32_cuda* vec_tgt, Real* mat_net_out,
   }
 }
 
+// Differentiate backward through the softmax function.  Here, "value" is the
+// softmax output. Does, for each row i,
+// x(i) =  diff(i) * diag(value(i)) - diff(i) * (value(i)^T * value(i))
+// xxxx(i) is row-vector; '*' and '-' are matrix operations.
+// CUDA thread layout: one thread block (256 threads) per matrix-row.
```
danpovey: "256" -> "CU1DBLOCK == 256"

src/cudamatrix/cu-kernels.cu
danpovey: We generally prefer the outputs of functions to be the final argument, as it's more consistent with the Google style... OK if this function was pre-existing.
Perhaps more suitable names would be:
 x -> input_deriv, or x_deriv
 value -> output, or y
 diff -> output_deriv, or y_deriv

kangshiyin: It is now following the naming in `nnet3` as `in_deriv`, `out_value` and `out_deriv`. The output param is put to the end, but we may need to change all other kernel functions later.

src/cudamatrix/cu-kernels.cu
kangshiyin: Done.

kangshiyin:
@galv Yes, all GPU test have passed.
I would prefer to use separate commits as it may help diagnosis the bugs if any. But I'm using `git rebase master` after the modification to keep the branch line short.

PR #1026, opened 2016-09-14
cudamatrix: replacing the random number generator,

vesis84:
- The CUDA kernels are replaced by calls to 'curand.h' library...
- All tests passed: cudamatrix, nnet{1,2,3}
- It would be good to merge it sooner than later, it was cherry-picked from another local branch, which I currently work with...


src/cudamatrix/cu-rand.h
```diff
@@ -1,6 +1,6 @@
-// cudamatrix/cu-rand.h
+// cudamatrix/cu-rand2.h
```
danpovey: i think this should not be changed

danpovey:
Have a look at why the checks failed.
I'll talk with Yenda about when to merge-- not sure.  We have to balance the inconvenience to you, vs. the inconvenience to everyone else.

PR #1028, opened 2016-09-15
Nnet1 updates

vesis84:
- cudamatrix: The CUDA kernels are replaced by calls to 'curand.h' library...
  - All tests passed: cudamatrix, nnet{1,2,3}
- nnet1: some dropout related scripting, improving sentence selection in 'multistream-perutt' training

- Replaces PR #1026 (got a bit scared of possible consequences from combining rebase and cherry-pick, so I better open the PR with the original branch)


danpovey:
@jtrmal, what do you think?  I'm thinking of asking Karel to separate out the parts that use the curand stuff, to be committed at some later date, and just commit the remainder.  I don't see that any speed benefit here will outweigh the inconvenience of forcing every single Kaldi user to rerun "configure".

jtrmal:
Well, I think curand should indeed be made as a separate PR
Karel, is this rand replacement bringing you some noticeable speedup
somewhere? I originally thought it should not have any effect on speed
basically, as rand is used only to initialize matrices but perhaps the
impact is larger?
y.

vesis84:
Hi, the curand was ~4x faster in generating random matrices of floats, compared to the old implementation.

There are several benefits of the new implementation:
- the initialization of state buffers is no longer bound to specific dimensions of the matrix, and its initialization is not done on CPU anymore (with old code, the change of dimension of target matrix was causing slow re-seeding of state buffers on CPU).
- it is based on library calls, so we no longer need to maintain the random number generator ourselves

This will affect the speed of components, where random numbers are used, for example the Dropout component. Please take a look at the code...

danpovey:
I'm leaning towards merging this in one go since it might be a while before we need to update the configure version number for any other reason; but we should test it on our grid a bit first, since there are a lot of changes.  @freewym, do you have time in one of your neural net experiments to pull these changes?  I'm not suggesting to redo an existing experiment: just to do whatever you're doing but with these changes, and see if there are any crashes or unexpected large changes in performance.

freewym:
OK. Do you basically want to test the c++ level changes work fine, with any one of nnet3 experiment?

danpovey:
Yes.

danpovey:
and run 'make test' also, in src/.

jtrmal:
I think running configure and probably make clean is not that terrible
thing -- the biggest annoyance for me used to be the right command line and
we store that in the header of kaldi.mk.
Perhaps we could make the configure script to drop a hint where to find the
previous configure line?
Of course, if we could merge two things that need to bump the configure
version number at the same time it would be better.
y.

freewym:
make test failed in /nnet2 (but passed in /nnet and /nnet3)

log file: /home/ywang/kaldi-chain/src/nnet2/nnet-component-test.testlog

danpovey:
Ignore that test failure, it doesn't seem to be related to this change (must be random); I'm separately changing that test code to be more accepting.

danpovey:
@jtrmal, that's a good idea, to give a hint of the exact configure line to use.

vesis84:
Yea, in this case, the ./configure version is increased, but it has no other change. What changed are the template files: src/makefiles/cuda_{32,64}bit.mk 

(Btw. I could not really decode the mechanism how the version checking works, but it is good that it does the job well)

freewym:
Everything looks fine when experimenting on tdnn. First 50 iters of BLSTM training also look normal.

Yiming

PR #1029, opened 2016-09-16
fixed the usage of lmrescore_const_arpa.sh

sih4sing5hong5:
It seems this script was copied from `steps/lmrescore.sh` :)


danpovey:
thanks!

PR #1032, opened 2016-09-17
Added Dropout component in netted to be used ephemeral connection.

pegahgh:


src/nnet3/nnet-simple-component.h
```diff
@@ -79,6 +79,56 @@ class PnormComponent: public Component {
...
+// This component randomly zeros dropout_proportion of the input 
+// and the derivatives are backpropagated through the nonzero inputs.
+// Typically this component used during training but not in test time.
+// The idea is described under the name Dropout, in the paper 
+// "Dropout: A Simple Way to Prevent Neural Networks from Overfitting".
+class DropoutComponent : public RandomComponent {
+ public:
+  void Init(int32 dim, BaseFloat dropout_proportion = 0.0);
+
+  DropoutComponent(int32 dim, BaseFloat dp = 0.0) { Init(dim, dp); }
```
danpovey: dp is a too-short name.

src/nnet3/nnet-simple-component.h
```diff
@@ -79,6 +79,56 @@ class PnormComponent: public Component {
...
+
+  DropoutComponent():dim_(0), dropout_proportion_(0.0) { }
```
danpovey: space after colon

danpovey:
You haven't added DropoutComponent to the components used in testing.  So it's not being tested.  It's not clear that the tests would even work without modification- it might be necessary to reset the random number generators for some of the comparisons.

danpovey:
Also we need a way to set the dropout-proportions via the command line.
You could add a function
SetDropoutProportion()
in nnet-nnet.h, and add an option --set-dropout-proportion to nnet3-copy.

pegahgh:
Done all changes. I added SetDropoutProportion, RemoveDropout and modified test to reset seed for dropout component test. I will commit soon!

danpovey:
Please push your changes, and also rebase to clean up the commit history- I prefer just one commit.

pegahgh:
I didn't have access to my computer. I will push clean commit in 4 hour.

danpovey:
@pegahgh, please go through all comments on this page and make sure you have addressed them-- same for the other PR you have.  There are outstanding comments where the issue is not fixed.

src/nnet3/nnet-simple-component.h
pegahgh: Done!

src/nnet3/nnet-simple-component.h
pegahgh: Done!

pegahgh:
@danpovey I reviewed all comments on this page and I found all the issues fixed. Maybe the issue is that I did rebase and the fixed version of codes don't appear in conversation!
Did you check the files changed directly or you just checked the conversation?

danpovey:
Oh-- I just checked the conversation, it seems the files were updated if I click on the right tab.
I think it used to be the case that github would say "danpovey commented on an outdated commit" if the lines of code that I commented on had changed, but it didn't say that.

src/nnet3/nnet-nnet.cc
```diff
@@ -784,5 +784,30 @@ std::string Nnet::Info() const {
...
+  // resets random-number generators for all random
+  // components.
+  for (int32 c = 0; c < NumComponents(); c++) {
+    RandomComponent *rc = dynamic_cast<RandomComponent*>(GetComponent(c));
+    if (rc != NULL)
+      rc->ResetGenerator();
+  }
+}
+
+void Nnet::RemoveDropout() {
```
danpovey: this function would not work.. nnet3 isn't just a linear sequence of components.  Just remove this function for now.

pegahgh:
Maybe the reason is that I commented on conversation after committing changes, so it is not out-dated!!

src/nnet3/nnet-component-itf.h
```diff
@@ -350,6 +350,16 @@ class Component {
 };
 
 
+class RandomComponent: public Component {
+ public:
+  // This function is required in testing code and in other places we need
+  // consistency in the random number generation (e.g. when optimizing
+  // validation-set performance), but check where else we call sRand().  You'll
```
danpovey: should be srand not sRand.  this appears elsewhere too, I think.

pegahgh:
@danpovey  I can easily test ephemeral connection using dropout on TDNN!
Probably it would be better idea, if Yiming check this connection on LSTM, since he is more expert in that area!

danpovey:
Before that can be done, I need to do some coding to make it possible to remove the dropout components.  I want to make this part of a more general framework for editing nnets, that will also be used in your multilingual stuff.  So I think it might be better to wait a few days.   But I guess if you want to start work on the scripting components, and you're not otherwise busy, it's OK with me.

src/nnet3/nnet-nnet.h
```diff
@@ -229,6 +230,13 @@ class Nnet {
 
   // Assignment operator
   Nnet& operator =(const Nnet &nnet);
+
+  void ResetGenerators(); // resets random-number generators for all
+  // random components.  You must also set sRand() for this to be
+  // effective.
+  
+  // Removes all dropout components from network, which can be used during test time.
+  void RemoveDropout();
```
danpovey: Can you please take out this `RemoveDropout()` function?
I have been thinking about it, and I think the best way to do this would be at the config level: a file "remove-ephemeral-connections.config" would be created that modifies the nodes that sum over the real and emphemeral connections by replacing expressions like `Add(real_node, ephmeral_node)` with `real_node`.
Then we can clean up the model by taking out orphan nodes and components; I have written the code for this and will commit it at some point in future.

pegahgh:
Removed RemoveDropout function.

src/nnet3/nnet-simple-component.cc
```diff
@@ -86,6 +86,85 @@ void PnormComponent::Write(std::ostream &os, bool binary) const {
...
+                                const CuMatrixBase<BaseFloat> &out_value,
+                                const CuMatrixBase<BaseFloat> &out_deriv,
+                                Component *to_update,
+                                CuMatrixBase<BaseFloat> *in_deriv) const {
+  KALDI_ASSERT(in_value.NumRows() == out_value.NumRows() &&
+               in_value.NumCols() == out_value.NumCols());
+
+  KALDI_ASSERT(in_value.NumRows() == out_deriv.NumRows() &&
+               in_value.NumCols() == out_deriv.NumCols());
+  in_deriv->AddMatMatDivMat(out_deriv, out_value, in_value);
```
danpovey: since this is adding, you should include kBackpropAdds in the Properties().
You should also add the kLinearInInput flag.

src/nnet3/nnet-simple-component.cc
```diff
@@ -143,7 +143,9 @@ void DropoutComponent::Backprop(const std::string &debug_info,
 
   KALDI_ASSERT(in_value.NumRows() == out_deriv.NumRows() &&
                in_value.NumCols() == out_deriv.NumCols());
-  in_deriv->AddMatMatDivMat(out_deriv, out_value, in_value);
+  CuMatrix<BaseFloat> in_deriv_inplace(in_deriv->NumRows(), in_deriv->NumCols());
```
danpovey: I'm not sure why you made these changes?  This shouldn't have been necessary.  Let me know if there was a test failure when you added the flag, the cause might be something else.

src/nnet3/nnet-simple-component.cc
pegahgh: @danpovey 
I added kBackpropAdds and kLinearInInput to Properties(). 
It seems AddMatMatDivMat method doesn't add *this to its output, which is against what we expect in most of AddMat* functions.
I didn't modify this function and I supported kBackpropAdds in backprop by having temporary matrix and adding it to in_deriv!

we may need to modify AddMatMatDivMat method to do *this += a * b / c !!

src/nnet3/nnet-simple-component.cc
pegahgh: I added kBackpropAdds and kLinearInInput to Properties(). 
It seems AddMatMatDivMat method doesn't add this to its output, which is against what we expect in most of AddMat functions.
I didn't modify this function and I supported kBackpropAdds in backprop by having temporary matrix and adding it to in_deriv!

we may need to modify AddMatMatDivMat method to do *this += a * b / c !!

danpovey:
OK, I see the problem.
The issue is that the name of the function AddMatMatDivMat is out of sync with what the function actually does: it doesn't add to the matrix, it sets it.  It should be renamed to SetMatMatDivMat.
And you can remove the flag kBackpropAdds, add the flag kBackpropInPlace, and revert the code changes in the Backprop function.

pegahgh:
Should I rename the function in this PR, or you plan to do this later?

danpovey:
rename it

src/cudamatrix/cu-matrix.cc
```diff
@@ -989,7 +989,7 @@ void CuMatrixBase<Real>::AddMatBlocks(Real alpha, const CuMatrixBase<Real> &A,
 /// dst = a * b / c (by element; when c = 0, dst = a)
 /// dst can be an alias of a, b or c safely and get expected result.
 template<typename Real>
-void CuMatrixBase<Real>::AddMatMatDivMat(const CuMatrixBase<Real> &A,
+void CuMatrixBase<Real>::SetMatMatDivMat(const CuMatrixBase<Real> &A,
```
danpovey: If we don't do this now we'll forget about it-- can you also please change the kernel names to remove the _add_ and make it _set_ ?

src/nnet3/nnet-simple-component.cc
```diff
@@ -86,6 +86,85 @@ void PnormComponent::Write(std::ostream &os, bool binary) const {
...
+}
+
+std::string DropoutComponent::Info() const {
+  std::ostringstream stream;
+  stream << Type() << ", dim = " << dim_ 
+         << ", dropout-proportion = " << dropout_proportion_;
+  return stream.str();
+}
+
+void DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
```
vijayaditya: In this implementation of drop-out you are using a different mask for each sample in the minibatch. I don't know if other versions of drop-out do this (can anyone clarify ?). If they don't you should probably highlight this in the comments. I think this version is better.

danpovey:
Karel's did.  I expect they probably normally do it that way but don't know
for sure.

src/nnet3/nnet-simple-component.cc
vesis84: Hi Vijay, I am not aware of a paper where would be mentioned that the same dropout mask should be used for all the samples within the mini-batch. Yes, I am assuming that all the samples are independent, having different topologies of NN, so each sample has a different mask. Has anybody tried having it the other way? Thanks!
K.

vijayaditya:
@pegahgh I was curious about the expected completion date for this component.

danpovey:
The only thing that has to be done is to rename some kernels.
(Also, to resolve some conflicts).

pegahgh:
oh! I will rename them right now!

pegahgh:
Done!

src/nnet3/nnet-nnet.h
```diff
@@ -247,6 +248,12 @@ class Nnet {
   // as it could ruin the graph structure if done carelessly.
   void RemoveSomeNodes(const std::vector<int32> &nodes_to_remove);
 
+=======
+  void ResetGenerators(); // resets random-number generators for all
+  // random components.  You must also set srand() for this to be
+  // effective.
+  
+>>>>>>> 05d2e30... added DropoutComponent to nnet3 for ephemeral connection.
```
danpovey: looks like you did not completely resolve these conflicts!

pegahgh: oh sorry! fixed!

src/nnet3/nnet-test-utils.cc
```diff
@@ -1219,6 +1219,15 @@ static void GenerateRandomComponentConfig(std::string *component_type,
         os << " self-repair-target=" << RandUniform();
       break;
     }
+<<<<<<< HEAD
+=======
```
danpovey: and here.. check for === via grep et.

danpovey:
Also, please make sure this compiles with GPU... if you had compiled, you wouldn't have missed these errors.

pegahgh:
fixed type and it compiled with GPU!

PR #1038, opened 2016-09-19
updating the fst patches

jtrmal:
I actually regenerated the patches/ changed the format from the old context-patch into the new(er) unified diff/patch.
I did it applying the old patch and generating the new format.
For 1.3.4, I included the text-io.cc compilation fix


danpovey:
Cool.  If you think this is sufficiently tested, then merge it.

danpovey:
@jtrmal, don't forget to merge this!

PR #1039, opened 2016-09-19
Check that the `which` command exists

aevernon:
check_dependencies.sh depends on `which`, but some distributions, such as those intended for use with Docker, do not include it. Check to see if is installed.

### DIFF
--- a/tools/extras/check_dependencies.sh
+++ b/tools/extras/check_dependencies.sh
@@ -12,6 +12,11 @@ function add_packages {
   opensuse_packages="$opensuse_packages $3";
 }
 
+if ! which -v >/dev/null 2>&1; then
+  echo "$0: which is not installed."
+  add_packages which which which
+fi
+
 if ! which g++ >&/dev/null; then
   echo "$0: g++ is not installed."
   add_packages gcc-c++ g++ gcc-c++


jtrmal:
perhaps  a better way would be to replace "which" by "command -v", which I
believe is bash builtin
y.

aevernon:
I'm not familiar with `command`.

Do you mean specifically in my commit, or do you mean in all of check_dependencies.sh?

aevernon:
Oh, I see what you mean. You mean run `g++ -v`, `make`, `bzip -v`, and so forth and look for a non-zero exit code instead of invoking the `which` command.

Yes, that approach is described at https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script/677212#677212.

jtrmal:
Actually, I think it is OK if you leave the PR as it is as quite a few
scripts is using the "which" command and those would fail no matter what we
do to the check_dependencies.sh, unless the other files are fixed as well.
y.

jtrmal:
no, the command/built-in is actually really called "command"

i.e. the test would be
if command -v which  >/dev/null 2>/dev/null;  then
  echo "which does not exist"
fi


but it does not really matters because "which" is used in all recipes, i.e.
it makes sense do it the way you did it
y.

aevernon:
Neat, I didn't know about the `command` command.

For anyone else who comes across this ticket, `command -v` returns 0 if it can find its argument as a builtin or in the PATH, and it returns 1 if cannot find it.

```bash
$ command -v cat                                                                                      
/usr/bin/cat
$ echo $?                                                                                             
0
$ command -v time                                                                                     
time
$ echo $?                                                                                             
0
$ command -v does_not_exist                                                                           
$ echo $?                                                                                             
1
```

danpovey:
`if ! which which 2>/dev/null; ... `

PR #1041, opened 2016-09-20
Fix data prep destination in librespeech

rajathkumarmp:


egs/librispeech/s5/run.sh
```diff
@@ -29,7 +29,7 @@ local/download_lm.sh $lm_url data/local/lm
 # format the data as Kaldi data directories
 for part in dev-clean test-clean dev-other test-other train-clean-100; do
   # use underscore-separated names in data directories.
-  local/data_prep.sh $data/LibriSpeech/$part data/$(echo $part | sed s/-/_/g)
+  local/data_prep.sh $data/LibriSpeech/$part $data/$(echo $part | sed s/-/_/g)
```
danpovey: actually that was not a bug.  but thanks.

rajathkumarmp: @danpovey 
while running the run.sh I get this error
```
utils/data/get_utt2dur.sh: segments file does not exist so getting durations from wave files
utils/data/get_utt2dur.sh: could not get utterance lengths from sphere-file headers, using wav-to-duration
utils/data/get_utt2dur.sh: wav-to-duration is not on your path
```

what am I doing wrong?

vijayaditya:
please check if this binary exists

$KALDI_ROOT/src/featbin/wav-to-duration

PR #1042, opened 2016-09-20
Avoid segfault in wav-reverberate

jfainberg:
Using --additive-signals without also setting --snrs and --start-times
produces a segmentation fault when calling AddNoise (snr_vector and
start_time_vector empty). Will check that these arguments are set and
that their length matches the number of added signals. Not sure 
whether the check for empty arguments is superfluous when we have 
the assertions later on.

Alternatively, could set more sensible default values for snrs and
start_times, e.g. 0 for all additive signals in both snrs and
start_times?

### DIFF
--- a/src/featbin/wav-reverberate.cc
+++ b/src/featbin/wav-reverberate.cc
@@ -184,7 +184,8 @@ int main(int argc, char *argv[]) {
                 "A comma separated list of additive signals. "
                 "They can be either filenames or piped commands. "
                 "E.g. --additive-signals='noise1.wav,noise2.wav' or "
-                "'sox noise1.wav - |,sox noise2.wav - |' ");
+                "'sox noise1.wav - |,sox noise2.wav - |'. "
+                "Requires --snrs and --start-times.");
     po.Register("snrs", &snrs,
                 "A comma separated list of SNRs(dB). "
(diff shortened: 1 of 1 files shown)


vijayaditya:
Looks good, I will merge when the checks pass. Please check the reasons for failure.

jfainberg:
Great! Though I'm confused by the build fail. If I'm understanding the travis build correctly, it looks like it failed building openfst, or didn't succeed downloading it? This commit hasn't touched the tools directory... Is there a way to re-run the build?

vijayaditya:
@jfainberg  I am unable to find the option to restart the build. So

vijayaditya:
@jfainberg I restarted the tests, lets wait and see.

jfainberg:
Thanks! Unfortunately it failed at the same place... Bit confusing why it would fail, but I'll try to debug...

jtrmal:
The issue was caused by one of our previous commits, where we added checks
for "which" into check_dependencies.sh
Investigating fix
y.

jtrmal:
Fingers crossed, will be fixed by #1043 (lets wait until the build finishes)

jtrmal:
Just re-opened the PR to restart travis checks.

jfainberg:
@jtrmal Thank you!

PR #1044, opened 2016-09-20
Fix some bugs in egs/hkust/s5/local/hkust_prepare_dict.sh

ling0322:
I found two problems in egs/hkust/s5/local/hkust_prepare_dict.sh:

- The first one is this script would fail at line 240, which have a character '2'. And after deleting this line, it works well.
- The second one is when generating lexicon-ch.txt (line 272), it didn't do any check. And when processing the hkust dataset, generated lexicon-ch.txt may contains some strange punctuation. These punctuation would cause the failure of 'utils/pinyin_map.pl' in line 276.


jtrmal:
Thanks for the patch!

naxingyu:
Ou, I introduced this bug. Thanks!

PR #1046, opened 2016-09-20
Configure change -- print cmdline of the last configure call

jtrmal:


danpovey:
Seems to make sense-- commit yourself once tests finish.

PR #1048, opened 2016-09-21
fix small issues in RNNLM rescoring scripts in WSJ

hainan-xv:


egs/wsj/s5/steps/rnnlmrescore.sh
```diff
@@ -71,6 +71,8 @@ for f in $rnndir/rnnlm $data/feats.scp $indir/lat.1.gz; do
 done
 
 nj=`cat $indir/num_jobs` || exit 1;
+cp $indir/num_jobs $dir/num_jobs
```
vince62s: @hainan-xv  you need to do this after the mkdir -p

PR #1050, opened 2016-09-21
Add self-loop-scale warning

### LINKED ISSUE #1049: Modify lmrescore.sh and maybe mkgraph.sh regarding self-loop-scale
The default self-loop-scale in some scripts is 0.1.  However, for chain models, we always set this to 1.0 via the command line, and forgetting to do this can lead to bad results.
It would be good if these scripts could check whether the file `frame_subsampling_factor` exists in the directory concerned, and if it does exist and the self-loop-scale is not 1.0, then print a warning.
@psmit, perhaps you could do this since you experienced this problem.

psmit:
When using mkgraph or doing rescoring with chain models a self-loop-scale
of 1.0 should be used. Show a warning when chain models are used, but the
self-loop-scale is not set to 1.0.

Resolves #1049

### DIFF
--- a/egs/wsj/s5/steps/lmrescore.sh
+++ b/egs/wsj/s5/steps/lmrescore.sh
@@ -98,6 +98,8 @@ case "$mode" in
      # grammar and transition weights.
     mdl=`dirname $indir`/final.mdl
     [ ! -f $mdl ] && echo No such model $mdl && exit 1;
+    [[ -f `dirname $indir`/frame_subsampling_factor && $self_loop_scale != 1.0 ]] && \
+      echo "$0: WARNING: chain models need '--self-loop-scale 1.0'";
     $cmd JOB=1:$nj $outdir/log/rescorelm.JOB.log \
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/steps/lmrescore.sh
```diff
@@ -98,6 +98,8 @@ case "$mode" in
      # grammar and transition weights.
     mdl=`dirname $indir`/final.mdl
     [ ! -f $mdl ] && echo No such model $mdl && exit 1;
+    [[ -f `dirname $indir`/frame_subsampling_factor && $self_loop_scale != 1.0 ]] && \
+      echo "WARNING: chain models need '--self-loop-scale 1.0'";
```
danpovey: please in echo statements, start them with 'echo "$0:' so we know which script produces the message.

psmit: Fixed

danpovey:
Thanks!  Merging.

PR #1051, opened 2016-09-21
removing direct links for subscription

jtrmal:
I'm proposing to remove the direct links for emailing the group and subscription.
So far, I get around 80 percent of the requests without reason filled out, despite the fact it's explicitly mentioned in bold on the kaldi-asr.org/forums.html.
I'm hoping that this percentage is this high just because there is some other way how to create the subscription request :(


jtrmal:
Merging myself without waiting for Travis (it won't break anything).
y.

Issue #802, opened 2016-05-21
[documentation] Wit Zelinsky tutorial- add to docs


danpovey:
We should put the Wit Zelinsky tutorial in the docs somehow.

danpovey:
More specifics on this: an updated version of Wit's tutorial is here
 https://groups.google.com/forum/#!msg/kaldi-help/tzyCwt7zgMQ/wvCLpVVpBgAJ
It is in Markdown.  The first thing we should try is whether we can just include it with a markdown filename extension, e.g.  src/doc/kaldi_for_dummies.md, and change MARKDOWN_SUPPORT = NO to YES in src/Doxyfile.  Something would have to be added to mainpage.dox to link to the page.  I vaguely recall that we set it to NO for a reason.  Please generate the HTML with both NO and YES and see how it differs.  Cd to src and do . doc/README to compile the documentation (if you're not me, transferring it to kaldi-asr.org will fail).

akreal:
Hi! I tried to generate the HTML with `MARKDOWN_SUPPORT = YES` and noticed that multiplication sign (`*`) in some equations gets interpreted as italics mark.

danpovey:
That must be why we disabled it.  I can't find a way to enable it only for
some files.  It might be necessary to format the tutorial as regular
doxygen.

PR #1055, opened 2016-09-23
Add Wit Zieliński tutorial

akreal:
As discussed in #802, it is formatted as regular doxygen.

### DIFF
--- a/src/doc/kaldi_for_dummies.dox
+++ b/src/doc/kaldi_for_dummies.dox
@@ -1,0 +1,619 @@
+// doc/kaldi_for_dummies.dox
+
+// Copyright 2016 Wit Zieliński  Pavel Denisov
+
+// See ../../COPYING for clarification regarding multiple authors
+//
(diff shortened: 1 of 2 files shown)


danpovey:
I'm going to merge this as-is and compile it and put it up at kaldi-asr.org-- please look over the result to see if there are any problems!

akreal:
It looks good on kaldi-asr.org.

wantee:
There is some problems with the tree view in the left frame on kaldi-asr.org. The subpages in mainpage are messed up with the modules in a single root.
I am not sure since when this problem appeared, it seems not related with the codes. Maybe you use a different version of doxygen, @danpovey?

danpovey:
Yes, I think our version of Doxygen increased when we upgraded our grid to
a new Debian version recently.  We have 1.8.8 now.  I have never fully
understood how to control the left frame of the tree view.  If you can
figure it out it would be nice.
Dan

PR #1061, opened 2016-09-26
fix issues caused by different interfaces of rnnlm and faster-rnnlm; …

hainan-xv:
…fix the problem of PPL computation using unnormalized probabilities

### DIFF
--- a/egs/wsj/s5/local/wsj_train_rnnlms.sh
+++ b/egs/wsj/s5/local/wsj_train_rnnlms.sh
@@ -111,20 +111,30 @@ echo "Training RNNLM (note: this uses a lot of memory! Run it on a big machine.)
 #  -hidden 100 -rand-seed 1 -debug 2 -class 100 -bptt 2 -bptt-block 20 \
 #  -direct-order 4 -direct 1000 -binary >& $dir/rnnlm1.log &
 
-$cmd $dir/rnnlm.log \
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/local/wsj_train_rnnlms.sh
```diff
@@ -111,20 +111,30 @@ echo "Training RNNLM (note: this uses a lot of memory! Run it on a big machine.)
...
+  $cmd $dir/rnnlm.log \
+     $KALDI_ROOT/tools/$rnnlm_ver/rnnlm -threads $threads -independent -train $dir/train -valid $dir/valid \
+     -rnnlm $dir/rnnlm -hidden $hidden -seed 1 -class $class -bptt $bptt -bptt-block $bptt_block \
+     $rnnlm_options -direct $direct || exit 1;
+else
+  $cmd $dir/rnnlm.log \
+     $KALDI_ROOT/tools/$rnnlm_ver/rnnlm -threads $threads -independent -train $dir/train -valid $dir/valid \
+     -rnnlm $dir/rnnlm -hidden $hidden -rand-seed 1 -debug 2 -class $class -bptt $bptt -bptt-block $bptt_block \
+     $rnnlm_options -direct $direct -binary || exit 1;
+fi
```
vince62s: @hainan-xv since you made the choice to split the 2 cases rnnlm.org / faster-rnnlm, I think it would be clearer if we retain only the valid options in each case:
faster-rnnlm: I think independent is useless, class is not used
rnnlm.org : threads is irrelevant

PR #771, opened 2016-05-09
WIP: Multi-database English LVCSR recipe

guoguo12:
See #699.

PR #1062, opened 2016-09-26
multi_en: removed --out-suffix from run_ivector_common.sh

sikoried:
Removed the `--out-suffix` option from the `{make_mfcc,compute_cmvn_stats}.sh` calls in`multi_en/local/nnet3/run_ivector_common.sh`, which I forgot to remove prior to the merge of https://github.com/kaldi-asr/kaldi/pull/771.

There seem to be a number of merge commits tacked onto this PR, I'm assuming they go away with squash-on-merge?


danpovey:
Yes I think they will go away when I squash-- we'll find out.  Thanks!

PR #1063, opened 2016-09-26
Pick random ivector during chain training

vimalmanohar:


egs/wsj/s5/steps/nnet3/chain/get_egs.sh
```diff
@@ -66,6 +66,7 @@ max_shuffle_jobs_run=50  # the shuffle jobs now include the nnet3-chain-normaliz
                          # without overloading the disks.
 srand=0     # rand seed for nnet3-chain-copy-egs and nnet3-chain-shuffle-egs
 online_ivector_dir=  # can be used if we are including speaker information as iVectors.
+pick_random_ivector=false
```
danpovey: I'd rather just make this behavior the default.  It seems extremely unlikely that the old behavior would ever be better.  I know it's a hassle because it makes results not 100% comparable to old results.  But it will avoid having an extra configuration to deal with in the future.

vimalmanohar: I made the change, although its not visible in here.

danpovey:
I actually was talking about hard-coding it at the code level.  I see no reason why we'd ever really want to make that configurable, since it will probably turn out to be consistently better (we will test to some extent, of course).

danpovey:
Seems reasonable.  Out of caution I'll wait till Yiming has finished testing the change on Tedlium.

danpovey:
OK, so this helps by about 0.2% on AMI (from Vimal's experiments) and makes no difference on tedlium (from Yiming's experiment).  Merging since it seems a priori unlikely to ever hurt, and empirically sometimes seems to help a little.

PR #1068, opened 2016-09-27
Fixing lmrescore

psmit:
A user notified me that my earlier commit breaks other modes of lmrescore.sh then the default. This commit fixes that.

-f mode of rm is "ignore nonexistent files and arguments, never prompt"

### DIFF
--- a/egs/wsj/s5/steps/lmrescore.sh
+++ b/egs/wsj/s5/steps/lmrescore.sh
@@ -115,7 +115,7 @@ case "$mode" in
     ;;
 esac
 
-rm $outdir/Ldet.fst 2>/dev/null
+rm $outdir/Ldet.fst 2>/dev/null || true
 
 if ! $skip_scoring ; then
   [ ! -x local/score.sh ] && \


egs/wsj/s5/steps/lmrescore.sh
```diff
@@ -115,7 +115,7 @@ case "$mode" in
     ;;
 esac
 
-rm $outdir/Ldet.fst 2>/dev/null
+rm -f $outdir/Ldet.fst
```
danpovey: In general we avoid the -f flag of 'rm', because people might purposely make a file un-writable to stop it being deleted, and the -f flag overrides that.  Can you please change this to
`rm $outdir/Ldet.fst 2>/dev/null || true`
?

psmit:
Ok, fixed that

egs/wsj/s5/steps/lmrescore.sh
psmit: Fixed

danpovey:
Thanks! Merging.

PR #1071, opened 2016-09-28
WIP: Integration of word classes (example corpus: wsj)

horndasch:
**Word class integration (in wsj recipe)**

There are new scripts in `egs/wsj/s5/local/wclass` to integrate word classes
into the Kaldi workflow. The main scripts are
- `replace_wclass_entries_with_labels.sh` (to create a training text for the
  language model with word class labels replacing words from that word class)
- `create_wclass_SLMs.sh` (to create word class-based sub-language models)
- `embed_wclass_SLMs.sh` (to embed the sub-language models in the overall
  language model using fstreplace)

The directory `egs/wsj/s5/local/wclass` also includes additional helper scripts and resources for a wsj example, namely
- `wclass_list.txt` (the list of word classes, in this case US_STATE and
  WEEKDAY)
- `wsj.classes` (the mapping of words to classes according to
  http://www.speech.sri.com/projects/srilm/manpages/classes-format.5.html)
- `wsj_classes_lexicon.txt` (some additions for the lexicon, mainly compounds
  like 'SOUTH-DAKOTA')

To show how it works for wsj the following (local) scripts have been
modified/added
- `run.sh` (`wclass_usage=1` will switch on the use of word classes)
- `local/wsj_word_class_prep.sh` (a new script preparing wclass data)
- `local/wsj_train_lms.sh` (now includes a call to
  `replace_wclass_entries_with_labels.sh`)
- `local/wsj_format_local_lms.sh` (this script makes calls to
  `create_wclass_SLMs.sh` and `embed_wclass_SLMs.sh`)

Due to the fact that some compound words (such as 'SOUTH-DAKOTA' etc.) were not in the lexicon, but were used in the US_STATE word class, these entries had to be added to the lexicon. As a consequence the following local 'wsj' scripts have been slightly modified
- `local/wsj_extend_dict.sh`
- `local/wsj_prepare_dict.sh`

Please note
- The implementation now makes use of the `--extra-word-disambig-syms`
  option which was added to `utils/prepare_lang.sh` (see #1113). As a
  consequence two scripts from the original implementation were not
  needed any more.
- Non-terminal symbols, which are used to replace word class entries
  in the LM training text, are not added to `words.txt` any more.
- Some script interfaces have been simplified.


egs/wsj/s5/local/wsj_format_local_lms.sh
```diff
@@ -63,6 +206,32 @@ gunzip -c $lm_srcdir_3g/lm_unpruned.gz | \
...
+
+if [ $CLASSES == 1 ]; then
+  G_FST=data/lang${lang_suffix}_test_bd_fg/G.fst
+  G_WCLASS_FST=data/lang${lang_suffix}_test_bd_fg/G_wclass.fst
+  G_NO_WCLASS_FST=data/lang${lang_suffix}_test_bd_fg/G_no_wclass.fst
+  G_TMP_FST=data/lang${lang_suffix}_test_bd_fg/G_tmp.fst
+
+  cp $G_FST $G_WCLASS_FST
+
+  # go through all word classes in 'word_class_list.txt'
```
danpovey: It seems like this loop could have been broken out somehow as a script that you call... or at least there could have been more structure here.

egs/wsj/s5/local/wsj_format_local_lms.sh
```diff
@@ -42,16 +42,159 @@ if [[ -z $bos || -z $eos ]]; then
...
+fi
+
+
 # Be careful: this time we dispense with the grep -v '<s> <s>' so this might
 # not work for LMs generated from all toolkits.
 gunzip -c $lm_srcdir_3g/lm_pr6.0.gz | \
   arpa2fst --disambig-symbol=#0 \
            --read-symbol-table=$lang/words.txt - data/lang${lang_suffix}_test_bd_tgpr/G.fst || exit 1;
+
+if [ $CLASSES == 1 ]; then
```
danpovey: There is a lot of inconsistency in capitalization.  We prefer shell variables to be lower case-- yours are quite inconsistent.  And shell and python scripts should be lower case and with underscores, not dashes.

egs/wsj/s5/utils/create_wdisambig_file.pl
```diff
@@ -0,0 +1,60 @@
...
+# get the option values
+&GetOptions(\%optctl,
+    "word_class_list=s",
+    "remove_class_prefix=s",
+    );
+
+# start the list of disambiguation symbols with #0 (empty word symbol)
+print "#0\n";
+
+if(defined($word_class_list) and $word_class_list ne "") {
```
danpovey: please put spaces after if and while.

egs/wsj/s5/local/wsj_word_class_prep.sh
```diff
@@ -0,0 +1,69 @@
...
+# the main purpose of the script is to copy / convert word class files
+# from an external source (${CORPUS_WORD_CLASS_DIR}) to the place in
+# the Kaldi directory structure (${KALDI_WORD_CLASS_DIR})
+
+export LC_ALL=C
+
+CORPUS_NAME=wsj
+
+if [ $# != 2 ]; then
+  echo "Usage: ./local/${CORPUS_NAME}_word_class_prep.sh /path/to/corpus/word/class/files new/kaldi/word/class/dir"
```
danpovey: In this usage comment or at the top of this script (or both) there should be a bit more explanation of what this script does, and what its input and output formats are.
There was the same issue with at least one of the perl scripts you added, that there seemed to be too little explanation of what the script was for and what it does.

egs/wsj/s5/run.sh
```diff
@@ -50,7 +61,7 @@ local/wsj_format_data.sh --lang-suffix "_nosp" || exit 1;
  # use local/cstr_wsj_extend_dict.sh --dict-suffix "_nosp" $corpus/wsj1/doc/ instead.
   (
    local/wsj_extend_dict.sh --dict-suffix "_nosp" $wsj1/13-32.1  && \
-   utils/prepare_lang.sh data/local/dict_nosp_larger \
+   utils/prepare_lang.sh $word_class_command_line_option data/local/dict_nosp_larger \
```
danpovey: word_class_command_line_option could just be renamed to word_class_opt.

egs/wsj/s5/utils/create_wdisambig_file.pl
danpovey: Also note: just because some of my comments are not very minor, does not necessarily mean I am OK with the general structure of this.  At this point it wasn't super clear to me, from looking at your patch, how it works overall, and as a rule I don't spend too much time figuring these things out because if it's not immediately obvious, then the code needs to be explained better or made more obvious.  I'll re-review this after you've made some clarifications and maybe broken some parts of it out into separate scripts (if you think it makes sense).
Think about how much effort it will take a user, from looking at what you've done, to understand how this works, and to implement the same idea in a different setup (say, switchboard).  Mixing things too deeply into the data-prep scripts isn't always a good idea because those are corpus dependent.  Now, the LM training is part of the data-prep so it's hard to separate from the local scripts, but if you could find a way to isolate your changes somehow so that it's easier to apply to other setups, it would make things easier.   E.g. have a step that happens just before building the ARPA LM, and then a later step where you make a modification to the FST after the ARPA is built.  Feel free to define a special directory structure, if it would be helpful, where you store info about the word classes.

egs/wsj/s5/utils/prepare_lang.sh
```diff
@@ -285,26 +320,25 @@ if "$silprob"; then
...
+# replaces the awk expression to create '$dir/words.txt'; the awk expression
+# started with "cat $tmpdir/lexiconp.txt | awk '{print $1}' | sort | uniq  | awk ' ..."
+#
+# The main (and hopefully only) difference is that the perl script also creates
+# entries for word-class-disambiguation symbols in words.txt if word classes
+# are used.
+#
+# The script also creates entries for the <s> and </s> symbols (beginning and
+# end of sentences) as the awk expression did.
+utils/create_words_symbol_mapping.pl -lexp_file $tmpdir/lexiconp.txt $word_class_command_line_option \
```
danpovey: for long options we use --lexp-file, not -lexp_file.

egs/wsj/s5/local/wsj_prepare_dict.sh
```diff
@@ -81,6 +81,13 @@ grep -v ';;;' $dir/cmudict/cmudict.0.7a | \
 (echo '!SIL SIL'; echo '<SPOKEN_NOISE> SPN'; echo '<UNK> SPN'; echo '<NOISE> NSN'; ) | \
  cat - $dir/lexicon1_raw_nosil.txt | sort | uniq > $dir/lexicon2_raw.txt || exit 1;
 
+# Axel Horndasch (axel.horndasch@fau.de):
+# This is probably the worst hack ever made; but it will help me test word classes
+# for Kaldi on the WSJ corpus; please forgive me, but some compounds from the word
+# class "US_STATE" like "WEST-VIRGINIA" were not in the lexicon ...
+cat $dir/lexicon2_raw.txt local/wclass/wsj_classes_lexicon.txt | sort | uniq > $dir/lexicon2_raw_tmp.txt
+mv $dir/lexicon2_raw_tmp.txt $dir/lexicon2_raw.txt
```
danpovey: We may revisit this after you rework other aspects of the scripts.  What I'd like is that rather than just getting it working, you come up with some scripts that are easy to understand and use, where the aspects relating to words classes are as cleanly as possible separated from other aspects (or, if mixed in, are are mixed in in as natural a way as possible).

danpovey:
I sent this a few days ago via email to github, but I don't see it appearing here, so posting it directly.

I've been doing some more thinking about this pull request.

I'm not convinced that doing this through SRILM's notion of 'word class' is the right direction.  For instance, it seems to have forced you to create compound words for things like "north dakota".  That's unnecessary from the point of view of Kaldi itself.  And also the tie-in with SRILM is not very good, as we prefer to be agnostic about the language modeling toolkit.

What people really want here is the ability to have classes that could be many words, like "date" or "place" or "contact", as well as "day-of-week", "number" and the like.  To that end, it would probably be more practical to build the grammar (by hand, if needed), identify such things in the LM training data and replace them with a token that will get expanded later to the sub-grammar, and then train the LM in a normal way with no explicit notion of sub-classes.  Then 'fstreplace' would probably be used to replace the sub-grammar wherever it occurs in the FST.   It would also be possible to automatically train the weights in the sub-grammar based on observations from the LM training text.

To minimize the size of the resulting FST there are a few things that could be done (I did them in the 'unk-model' thing I recently checked in)... if you're going to go in this direction we can talk about that later.

But I think my revised position is that what needs to be done is to rework this, maybe even write it, with this more general framework in mind and not relying on SRILM's word-classes feature.  A tie-in (or optional tie-in) with the JSGF would be nice too.
I don't know if you think you have time to work on this, since it's becoming much more than just a reworking of your current pull request?  It would definitely be quite a lot of work, and would take on the timescale of months, most likely, once we take into account all the back and forth.

horndasch:
Thanks Dan for all your comments. I am currently in the final stages of major revisions to my concept and hope to update the pull request in the next few days. The changes I have made are described to some extent in this [document](https://horndasch.ssl-secured-server.de/s/W8BdG7LHwciC9Bc).

danpovey:
OK, here is a way that we could make measurable progress...
The way you have changed prepare_lang.sh, I don't like because it introduces a too-tight coupling between prepare_lang.sh and the very specific way you are doing word classes.
What would be better would be to add to prepare_lang.sh an option  --extra-word-disambig-syms disambig_syms.txt
where that file contains, one per line, an "extra" word-level disambiguation symbol.  These symbols would all be required to start with the hash mark `#` and *not* be numeric (to avoid overlap with the automatically generated symbols).  Also we should disallow `#-1` (hash-mark then minus one), because that is used internally in some FST stuff.
Perhaps you could try doing this as a separate pull request.
Don't forget that it needs to be suitable documented via comments and the usage message, and you need to check that the provided file matches the requirements.  Also check whether any changes are required to validate_lang.pl (e.g. any assumptions that the disambig symbols be numeric).
This should be easy to test in isolation.

egs/wsj/s5/local/wsj_format_local_lms.sh
horndasch: This should now be fixed.

egs/wsj/s5/local/wsj_format_local_lms.sh
horndasch: I added two new scripts which do what you suggested:
1. local/wclass/create_wclass_SLMs.sh (which creates word-class sub-language models)
2. local/wclass/embed_wclass_SLMs.sh (which embeds word-class sub-language models in the overall language model G.fst)

egs/wsj/s5/utils/create_wdisambig_file.pl
horndasch: All scripts have been revised to make them self-explanatory or at least much easier to understand. For that reason I changed the name of some of the scripts.

egs/wsj/s5/local/wsj_word_class_prep.sh
horndasch: The script (which is now called egs/wsj/s5local/wsj_wclass_prep.sh) has been revised as well and so have the perl scripts you mentioned.

egs/wsj/s5/run.sh
horndasch: I chose to call the option/parameter --wclass-opt/$wclass_opt. Hopefully this is better.

egs/wsj/s5/utils/prepare_lang.sh
horndasch: The handling of command line options for all perl scripts has been changed to conform with the Kaldi (script) coding conventions. I tried to use 'egs/wsj/s5/utils/write_kwslist.pl' as a template where possible.

egs/wsj/s5/local/wsj_prepare_dict.sh
horndasch: The changes to 'egs/wsj/s5/local/wsj_prepare_dict.sh' and 'egs/wsj/s5/local/wsj_extend_dict.sh' are still in this commit. Please let me know if I should remove them (I might have to use other example word classes then; currently I use WEEKDAY and US_STATES).

horndasch:
Please note that this pull request does not work right now, because I rebased it on the latest Kaldi master with an updated prepare_lang.sh (which contains the changes introduced by #1113).
Using "--extra-word-disambig-syms" for prepare_lang.sh, some things can be done more elegantly (and less invasive regarding prepare_lang.sh) when integrating word classes in the proposed way. I will commit an update once I have revised the existing solution.

danpovey:
Just had another look at this.. the major specific problems from the first time seem to be resolved (but still, the perl scripts should be lower case with underscores, not dashes).
However, I'm still not 100% sure whether to merge it.  I have a concern about its usefulness (as a trade-off with its size), as it seems to address a very specific use-case (SRILM with class-LMs).  What I think would be better is to support something like Thrax grammars (or general OpenFst-based grammars), and to train a regular LM where some of the tokens happen to expand via the grammar.  I'd like to leave this PR up for a bit and see if anyone has any comments.

danpovey:
@ngoel17, and any others who have used grammars... can I ask for comment on what you think Kaldi really needs in terms of support for grammars?  E.g. better support for a particular framework like thrax-based grammars?

PR #1072, opened 2016-09-28
fixed "Storing phones.txt and add check_phones_compatible.sh"

LvHang:
fix the bug which was proposed in kaldi-asr/kaldi #1022 
@vesis84 @mallidi 
Sorry for the inconvenience I have brought. I have fixed the bug. Thank you for your feedback.
@danpovey 
1. I changed all the `cp` commands which use `$alidir or $srcdir` to `$lang`.
2. For the scripts which has target model directory, but they don't have lang directory.
   (e.g. steps/combin_ali_dirs.sh, steps/nnet2/convert_lda_to_raw.sh, etc)
   I check if the `$srcdir/phones.txt` exists at first, then copy it to the new directory.
Sorry for the mistake.
Hang

### DIFF
--- a/egs/wsj/s5/steps/nnet/train_mmi.sh
+++ b/egs/wsj/s5/steps/nnet/train_mmi.sh
@@ -68,6 +68,10 @@ if ! $skip_cuda_check; then cuda-compiled || { echo "Error, CUDA not compiled-in
 
 mkdir -p $dir/log
 
+utils/lang/check_phones_compatible.sh $lang/phones.txt $srcdir/phones.txt
+utils/lang/check_phones_compatible.sh $lang/phones.txt $alidir/phones.txt
+cp $lang/phones.txt $dir
+
 cp $alidir/{final.mdl,tree} $dir
 
 silphonelist=`cat $lang/phones/silence.csl`
--- a/egs/wsj/s5/steps/nnet/train_mpe.sh
+++ b/egs/wsj/s5/steps/nnet/train_mpe.sh
@@ -72,6 +72,10 @@ if ! $skip_cuda_check; then cuda-compiled || { echo "Error, CUDA not compiled-in
 
 mkdir -p $dir/log
 
+utils/lang/check_phones_compatible.sh $lang/phones.txt $srcdir/phones.txt
+utils/lang/check_phones_compatible.sh $lang/phones.txt $alidir/phones.txt
+cp $lang/phones.txt $dir
+
 cp $alidir/{final.mdl,tree} $dir
 
 silphonelist=`cat $lang/phones/silence.csl`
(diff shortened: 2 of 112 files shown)


egs/wsj/s5/steps/nnet/train_mmi.sh
```diff
@@ -68,6 +68,10 @@ if ! $skip_cuda_check; then cuda-compiled || { echo "Error, CUDA not compiled-in
 
 mkdir -p $dir/log
 
+utils/lang/check_phones_compatible.sh $lang/phones.txt $srcdir/phones.txt
+utils/lang/check_phones_compatible.sh $lang/phones.txt $alidir/phones.txt
```
danpovey: this line is repeated.
Also (this comment reflects some other scripts in steps/nnet/), if a script has set -e in it, you don't need to do "|| exit 1", and it's prettier if you remove it.

egs/wsj/s5/steps/nnet/train_mpe.sh
```diff
@@ -72,6 +72,10 @@ if ! $skip_cuda_check; then cuda-compiled || { echo "Error, CUDA not compiled-in
 
 mkdir -p $dir/log
 
+utils/lang/check_phones_compatible.sh $lang/phones.txt $srcdir/phones.txt
+utils/lang/check_phones_compatible.sh $lang/phones.txt $alidir/phones.txt
```
danpovey: this line is repeated too.

egs/wsj/s5/steps/nnet2/convert_lda_to_raw.sh
```diff
@@ -44,6 +44,9 @@ for f in $src/final.mdl $src/final.mat $src/splice_opts $src/cmvn_opts; do
   [ ! -f $f ] && echo "$0: expected file $f to exist" && exit 1
 done
 
+if [ -f $src/phones.txt ]; then
+  cp $src/phones.txt $dir || exit 1;
```
danpovey: instead of an if-statement it would be more compact to do just
`cp $src/phones.txt $dir 2>/dev/null`
the `2>/dev/null` suppresses the warning.  In this case, we don't need to add `|| true` at the end because no-one has done `set -e` in this script (so the bad exit status wouldn't cause the script to exit).

, and the `|| true` will prevent an error

egs/wsj/s5/steps/nnet/train_mmi.sh
LvHang: I think the lines is different. Both srcdir and alidir are the input directories. So I check both of them. Do I really need to delete a line? I will fix the "|| exit 1" bug , if a script has set -e in it.

danpovey:
oh sorry did not notice.

LvHang:
@danpovey 
I reviewed all the files.
1. If a script has "set -e" in it, I deleted the "exit 1" command.
2. I use `cp $src/phones.txt $dir 2>/dev/null`, instead of the if-statement.
3. If the target directory is denlats-dir(e.g. make_denlats.sh), I will not copy the phones.txt. Because I think it isn't a model/alignment dir.
If they have any questions, please let me have it. Thanks a lot.
Hang

danpovey:
Looks good-- thanks!  Merging.

PR #1076, opened 2016-09-30
New augmentation recipe for reverberant speech

tomkocse:


egs/aspire/s5/local/multi_condition/musan_prep_noise_list.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+# Copyright 2016  Tom Ko
+# Apache 2.0
+# script to generate noise_list for MUSAN corpus
```
vijayaditya: mention that the output of this list is going to be used by your reverberate script, and the users can check that script for details about the format.

vijayaditya: provide a link the musan corpus on openslr.org

egs/aspire/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -21,14 +21,29 @@ set -e
...
   # prepare the impulse responses
   local/multi_condition/prepare_impulses_noises.sh --log-dir exp/make_reverb/log \
     --db-string "$db_string" \
     --download-rirs $download_rirs \
     --RIR-home $RIR_home \
     data/impulses_noises || exit 1;
 
+  # Generate the rir_list and noise_list for the reverberate_data_dir.py to corrupt the data
+  # this script just assumes air rwcp rvb2014 databases
+  python local/multi_condition/aspire_prep_rir_noise_list.py data/impulses_noises data/impulses_noises/info
```
vijayaditya: given that we are downloading several databases now, you could probably move all these to a new directory like `s5/databases`, rather than storing everything in the `data/` directory.

egs/aspire/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -37,16 +52,21 @@ if [ $stage -le 1 ]; then
...
-        --snrs "$snrs" --log-dir exp/make_corrupted_wav \
-        data/${data_dir}  data/impulses_noises $cur_dest_dir
-      reverb_data_dirs+=" $cur_dest_dir"
-    done
-    utils/combine_data.sh --extra-files utt2uniq data/${data_dir}_rvb $reverb_data_dirs
-    rm -rf $reverb_data_dirs
+
+    python steps/data/reverberate_data_dir.py \
+      --prefix "rev" \
+      --rir-set-parameters "1.0, SIM_RIR/simulated_rir_mediumroom/rir_list" \
```
vijayaditya: I would recommend adding your swbd recipes where you have examples of combining several rir sets and noise sets. 

Also please do not modify the original reverberation recipe as we are maintaining it for comparison with our ASpIRE submission. These experiments are very difficult to repeat and I would prefer not modifying this recipe until you have results which compare real rir+isotopic noise, real rir + sim rir + isotropic noise + point source noise settings.

If you would like to commit this new data prep recipe right now, you could move data preparation to another script and add the option of just using simulated rirs in that script.

Also remember to add the results with new recipe.

vijayaditya: Actually I think writing an example script which can be easily copied into other recipes is better. So you could create `local/reverberate_data_dir.sh` script where you download all the necessary databases, combine the different lists with different weights and finally reverberate the data directory.

You can call this script in  `local/nnet3/run_ivector_common.sh`.

vijayaditya: Also in the interest of making scripts easy to copy, keep everything related to data augmentation in a sub-directory like `local/multi_condition/reverberate/`. Most of the scripts in the current multi_condition directory can be moved to this script.

@danpovey People trying out data reverberation in other recipes like Babel, were copying the rir, noise prep scripts from ASpIRE. This replication is not desirable, so can we find a home for these scripts in the `steps/` directory ?

danpovey:
How about utils/data/ since we have other augmentation scripts there?  If
it doesn't have to run on the queue?

vijayaditya:
We use the queue for running several download jobs (one for each database)
in parallel. So can he put these in steps/data_augmentation ?

--Vijay

danpovey:
Wait-- I thought you were going to package all this stuff on openslr.org so
there is just one download?  We can't rely on those sites to continue to be
up.

danpovey:
... also, I don't like to have anything that has data in it, be in either
steps/ or utils/.  I wonder if it's possible to define a format for a
directory that contains info about the reverberations (e.g. the impulse
responses or whatever), and have it in that format on openslr.org (and
possibly scripts in local/ in some egs directory showing how you created
that), and then have a script in steps/ or utils/ that just applies that?
None of the scripts in steps/ or utils/ currently have any URLs for data
downloads in them.

vijayaditya:
We have mirrored all the external databases in openslr.org, however these are still individual uploads.
(e.g. MUSAN, sim-rir, RWCP, AIR, ....). Would you prefer having a single upload on openslr, which combines all these databases after the processing ? 

@tomkocse Now that I think of it, it might actually be preferable; as we don't need to have the dependency on python's scipy for processing the individual databases. Could you create an upload for this directory with the processed data (including real rirs, sim rirs, musan noises, isotropic noises) so that someone can just download and use it just using your reverberate script.

tomkocse:
@vijay, I will upload the swbd part after we we feel fine to the Aspire part.

egs/aspire/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -21,29 +23,20 @@ set -e
...
-  python local/multi_condition/musan_prep_noise_list.py  --sampling-rate 16000  musan/noise/free-sound musan/noise/free-sound/noise_list
-
-  # Download the simulated RIRs
-  wget --no-check-certificate http://www.openslr.org/resources/26/sim_rir.zip
-  unzip sim_rir.zip
-fi
-
 if [ $stage -le 1 ]; then
+  if [ "$base_rirs" == "simulated" ]; then
+    # Download the package that includes the real RIRs, simulated RIRs, isotropic noises and point-source noises
```
vijayaditya: For the `else` condition do not prepare the impulses and noises from scratch, just download the entire rirs_noises.zip and use the lists corresponding to the real_rirs and isotropic_noises. All I wanted to ensure was that we could replicate the ASpIRE submission results, the scripts need not be the same.

Please provide the scripts, in comments, which is used to create the upload. Rather than deleting the scripts which create the which create the RIR and noise list files move them to the directory  `local/multi_condition/impulses_noises` (IIRC). Also add a note here that this process is affected by #966  and this is being resolved in #981 .

egs/aspire/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -3,14 +3,16 @@
...
+stage=1
+snrs="20:10:15:5:0"
 foreground_snrs="20:10:15:5:0"
 background_snrs="20:10:15:5:0"
 num_data_reps=3
 db_string="'air' 'rwcp' 'rvb2014'" # RIR dbs to be used in the experiment
                                       # only dbs used for ASpIRE submission system have been used here
 RIR_home=db/RIR_databases/ # parent directory of the RIR databases files
 download_rirs=true # download the RIR databases from the urls or assume they are present in the RIR_home directory
+base_rirs="simulated"
```
vijayaditya: directly provide the rir_set_parameters here for generating the new results and the add in comments the rir_set_parameters for generating the old results.

egs/aspire/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -52,21 +45,34 @@ if [ $stage -le 1 ]; then
...
+        --speech-rvb-probability 1 \
+        --pointsource-noise-addition-probability 1 \
+        --isotropic-noise-addition-probability 1 \
+        --rir-smoothing-weight 0 \
+        --num-replications $num_reps \
+        --max-noises-per-minute 1 \
+        --random-seed 1 \
+        data/${data_dir} data/${data_dir}_rvb
+    else
+      reverb_data_dirs=
```
vijayaditya: Once again, it is not necessary to keep the old scripts. Just provide the appropriate options for using the new scripts to generate the old results (at least approximately). I think nothing changed other than the fact that you are using new rir and noise sets.

vijayaditya:
As I said before preserve the scripts which create the rir/noise_list files. You can check in the scripts which create the rir_noises.tar.gz file in openslr.org later. Assuming this is tested I am merging this.

jtrmal:
That will be my omission -- I forgot to process the request to publish that
database.
y.

jtrmal:
The archive is there and the link itself is valid now -- I tested it (in
the browser only)
y.

wonkyuml:
great. it works now. 
Wonkyum

PR #1079, opened 2016-09-30
trying to fix some issues under macos

jtrmal:


danpovey:
merge yourself if you think it's Ok

jtrmal:
I tried to compile on linux and it went just fine, so I'm merging this.

PR #1086, opened 2016-10-02
WIP fixing the cantab lm links

jtrmal:
My understanding from the brief scanning of the scripts is that "s5" needs the full archive and "s5_r2" is ok with the *-partial archive.

### DIFF
--- a/egs/tedlium/s5/local/download_data.sh
+++ b/egs/tedlium/s5/local/download_data.sh
@@ -16,8 +16,8 @@ else
 fi
 # Language models (Cantab Research):
 if [ ! -d cantab-TEDLIUM ]; then
-    echo "Downloading \"http://cantabresearch.com/cantab-TEDLIUM.tar.bz2\". "
-    wget --no-verbose --output-document=- http://cantabresearch.com/cantab-TEDLIUM.tar.bz2 | bzcat | tar --extract --file=- || exit 1
+    echo "Downloading \"http://www.openslr.org/resources/27/cantab-TEDLIUM.tar.bz2\". "
+    wget --no-verbose --output-document=- http://www.openslr.org/resources/27/cantab-TEDLIUM.tar.bz2 | bzcat | tar --extract --file=- || exit 1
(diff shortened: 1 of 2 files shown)


jtrmal:
Actually, do not merge yet -- I think I have to fix the r2 recipe more.
y.

danpovey:
OK, when you're done @vince62s can help test, if he's still awake... he was going to test something else there anyway.

jtrmal:
@vince62s, can you please test the recipes? just the LM preparation is ok, I guess...
Thanks.
y.

vince62s:
I only see changes to the download_data part.
@wantee was supposed to make chages in the ted_train_lm.sh
I can review the full process when completed.

danpovey:
I also committed a change from @wantee separately (I overlooked that it was
WIP).  So it should be in the repo already.

vince62s:
then I guess to need to commit the linked WIP in Pocolm too.

danpovey:
That won't be necessary, as the ability to take the lm-dir argument was
already present.

jtrmal:
guys, merge at your discretion -- the reason I was suggesting we might want
to test it is that the *partial file used in s5_r2 does not contain all the
files -- so if the recipe references/uses more that the training text and
dict, it will not work out of the box after merigng this.
y.

danpovey:
I just pushed a fix for the wordlist bug (which was in a commit I
previously merged).

vince62s:
download is ok.
train_lm.py is not:
train_lm.py: error: unrecognized arguments: data/local/local_lm/data/wordlist_4_train-2_ted-1.pocolm
maybe the dash in the name is an issue.

vince62s:
ohh missing mkdir -p ?

danpovey:
make sure your pocolm is up to date.

vince62s:
it is already

danpovey:
OK, it was a different problem-- an argument with spaces had not been quoted,  it was an issue in a previous commit.  I merged Yenda's commit and pushed a fix for that problems; if I find more problems I'll fix them too.

vince62s:
it finished fine.

PR #1087, opened 2016-10-02
nnet3: added command execution time computation code

vijayaditya:


src/nnet3/nnet-compute.cc
```diff
@@ -99,8 +100,10 @@ void NnetComputer::DebugBeforeExecute(int32 command,
 
 
 void NnetComputer::DebugAfterExecute(int32 command,
-                                     const CommandDebugInfo &info) {
+                                     const CommandDebugInfo &info,
+                                     BaseFloat command_start_time) {
   std::ostringstream os;
+  BaseFloat run_time = (static_cast<BaseFloat>(clock()) - command_start_time) / CLOCKS_PER_SEC;
```
danpovey: please use util/timer.h, for platform-independence.

danpovey: sorry, use base/timer.h instead, it was moved.

vijayaditya:
made the suggested changes.

PR #1090, opened 2016-10-03
Nnet1 dropout ivec

vesis84:
- adding support to annealed dropout,
- created an example how to prepare kaldi i-vectors on fMLLR features (AMI,IHM)
- nice gains on AMI IHM (dev, eval): w/o i-vector 24.2 24.5, per-spk i-vector 23.2 22.8, (Dan's lattice-free MMI 22.4 22.4)

### DIFF
--- a/egs/wsj/s5/steps/nnet/train_scheduler.sh
+++ b/egs/wsj/s5/steps/nnet/train_scheduler.sh
@@ -22,7 +22,7 @@ feature_transform=
 max_iters=20
 min_iters=0 # keep training, disable weight rejection, start learn-rate halving as usual,
 keep_lr_iters=0 # fix learning rate for N initial epochs, disable weight rejection,
-dropout_iters= # Disable dropout after 'N' initial epochs,
+dropout_schedule= # Dropout schedule for N initial epochs, for example: 0.9,0.9,0.9,0.9,0.9,1.0
 start_halving_impr=0.01
 end_halving_impr=0.001
 halving_factor=0.5
@@ -31,11 +31,11 @@ halving_factor=0.5
 verbose=1
 frame_weights=
 utt_weights=
- 
+
 # End configuration.
 
 echo "$0 $@"  # Print the command line for logging
-[ -f path.sh ] && . ./path.sh; 
+[ -f path.sh ] && . ./path.sh;
 
 . parse_options.sh || exit 1;
 
@@ -59,6 +59,8 @@ dir=$6
 [ ! -d $dir ] && mkdir $dir
 [ ! -d $dir/log ] && mkdir $dir/log
 [ ! -d $dir/nnet ] && mkdir $dir/nnet
+
+retention_array=($(echo ${dropout_schedule} | tr ',' ' '))
 
 # Skip training
 [ -e $dir/final.nnet ] && echo "'$dir/final.nnet' exists, skipping training" && exit 0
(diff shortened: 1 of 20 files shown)


egs/wsj/s5/steps/nnet/train_scheduler.sh
```diff
@@ -22,7 +22,7 @@ feature_transform=
 max_iters=20
 min_iters=0 # keep training, disable weight rejection, start learn-rate halving as usual,
 keep_lr_iters=0 # fix learning rate for N initial epochs, disable weight rejection,
-dropout_iters= # Disable dropout after 'N' initial epochs,
+dropout_schedule= # Dropout schedule for N initial epochs, for example: 0.9,0.9,0.9,0.9,0.9,1.0
```
danpovey: Are these probabilities the probability of dropout, or the probability of *not* dropping out?
I think when people normally describe dropout, it's the probability of setting the feature to zero, e.g. see
https://pdfs.semanticscholar.org/c2d7/8722ebac92766f1154497d8424108d906ae3.pdf

danpovey: Perhaps if you renamed this to dropout_retention_schedule it would lead to less confusion?

vesis84: Dan, how do you have it in nnet2, nnet3? As the probability that neuron is dropped? It would be good to have it the same way, I will change it then... 
(Actually, I was already thinking about it and discussed it with Harish)

mallidi:
Hi Karel,
Sorry for complaining. 
dropout_schedule seems to be clear. But dropout_retention_schedule might be better, as you are converting dropout_schedule into retention_schedule. 

Harish

vesis84:
Hi Harish (@mallidi), 
in the last commit I changed the 'dropout-retention' to 'dropout-rate'. 

This is a more standard formulation as @danpovey pointed out. Given that the values in 'dropout_schedule' are the probabilities that the **neurons are dropped**, we can use original variable name 'dropout_schedule'. 

The c++ code knows how to read the older models with <DropoutRetention> instead of <DropoutRate>, so the backward compatibility is okay...

Is it okay for your needs?
K.

mallidi:
Sure @vesis84 Thanks a lot for the annealed dropout 

Harish.

vesis84:
You are welcome ;) 

What seemed to work well on 'ami-ihm' was dropout-rate 0.2 in 5 initial epochs, then switching it to 0.0 (no dropout). This starts the learning rate decay, as without dropout the cross-entropy immediately increases on 'cv' data (and massively decreases on 'tr' data)... 

This schedule was better than 0.5 0.4 0.3 0.2 0.1 0.0, and some other combinations. 
Harish, I'm sure you might be more lucky with the annealed version :)

In the implementation there is one detail, after applying the dropout mask the output is up-scaled by 1/(1-p_drop). While, the cross-validations are always without dropout (hard-coded in the training binaries). So the up-scaling does a good job, there doesn't seem to be a severe mismatch caused by disabling the dropout in the cross-validation step... [I tried to exponentiate the 1/(1-p_drop) to make it little larger / smaller, but this was causing a mismatch visible on 'cv' loss]

K.

vesis84:
@danpovey I am done with the changes, from my side it's ready.

Issue #504, opened 2016-02-16
speaker-id: In SRE10 examples, use run.pl for all binaries and use "|| exit 1" for errors


david-ryan-snyder:
The current way of doing things is hiding informative error messages. 

From Dan on the forums:

> ```
> 
> "  ivector-compute-plda ark:$plda_data_dir/spk2utt \
> 
>     "ark:ivector-normalize-length scp:${plda_ivec_dir}/ivector.scp  ark:- |" \
> 
>       $plda_ivec_dir/plda 2>$plda_ivec_dir/log/plda.log
> ```
> 
> David, if you used run.pl here and did it '|| exit 1', instead of just redirecting the stderr, we'd get more informative error messages."

danpovey:
@david-ryan-snyder, don't forget this..

david-ryan-snyder:
@danpovey we should be able to close this issue now

Issue #376, opened 2015-12-02
speaker-id: Provide alternative ivector-plda-scoring length-normalization 


david-ryan-snyder:
Previously, length-normalization in the PLDA TransformIvector method relied on the within-class and between-class covariance of the PLDA model. This was simplified to use the same method as ivector-normalize-length. See http://sourceforge.net/p/kaldi/code/4925 for the original change.

However, there are situations where the older version performs better (thanks to Sree Harsha Yella and Najim Dehak for noticing this). 

In this issue, the older length-normalization needs to return as the default method in the PLDA class. However, there sould to be an option to ivector-plda-scoring that allows us to choose the alternative method. 

(assignee should be david-ryan-snyder)

david-ryan-snyder:
@danpovey I think we can close this issue now.

PR #1093, opened 2016-10-04
Various SID Fixes

david-ryan-snyder:
This addresses several issues in the SID examples, mostly related to length normalization. Also includes some cosmetic improvements. 

## PLDA Length Normalization
* Addresses Issue #376 
* This restores the original PLDA length normalization as the default when the option `--normalize-length=true` is set in the binary `ivector-plda-scoring`
* However, the simple length normalization (which is currently used and is the same as `ivector-normalize-length`) works better for the SRE08 eval, so the option `--simple-length-normalization` is added to provide an option that retains that approach. When set to `true` the default length normalization is replaced with the simple alternative.
* _Using the original PLDA length normalization results in improved SRE10 results_  (search for "EER"). The SRE08 results are unchanged, because `--simple-length-normalization=true` is set in that example.

## Informative Error Messages
* This addresses Issue #504 
* The sre10 scoring now uses run.pl and || exit 1

## Cosmetic Changes
* Removing trailing whitespace
* Using 2 space indent in run.sh scripts, etc

### DIFF
--- a/egs/sre08/v1/run.sh
+++ b/egs/sre08/v1/run.sh
@@ -243,7 +243,7 @@ trials=data/sre08_trials/short2-short3-female.trials
 ivector-compute-plda ark:data/train_female/spk2utt \
   'ark:ivector-normalize-length scp:exp/ivectors_train_female/ivector.scp  ark:- |' \
     exp/ivectors_train_female/plda 2>exp/ivectors_train_female/log/plda.log
-ivector-plda-scoring --num-utts=ark:exp/ivectors_sre08_train_short2_female/num_utts.ark \
+ivector-plda-scoring --simple-length-normalization=true --num-utts=ark:exp/ivectors_sre08_train_short2_female/num_utts.ark \
    "ivector-copy-plda --smoothing=0.0 exp/ivectors_train_female/plda - |" \
    "ark:ivector-subtract-global-mean scp:exp/ivectors_sre08_train_short2_female/spk_ivector.scp ark:- |" \
    "ark:ivector-normalize-length scp:exp/ivectors_sre08_test_short3_female/ivector.scp ark:- | ivector-subtract-global-mean ark:- ark:- |" \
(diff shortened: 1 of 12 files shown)


egs/sre08/v1/run.sh
```diff
@@ -276,11 +276,11 @@ ivector-plda-scoring --num-utts=ark:exp/ivectors_sre08_train_short2_male/num_utt
 # first, female.
 trials=data/sre08_trials/short2-short3-female.trials
 cat exp/ivectors_sre08_train_short2_female/spk_ivector.scp exp/ivectors_sre08_test_short3_female/ivector.scp > female.scp
-ivector-plda-scoring --num-utts=ark:exp/ivectors_sre08_train_short2_female/num_utts.ark \
-   "ivector-adapt-plda $adapt_opts exp/ivectors_train_female/plda scp:female.scp -|" \
-   scp:exp/ivectors_sre08_train_short2_female/spk_ivector.scp \
-   "ark:ivector-normalize-length scp:exp/ivectors_sre08_test_short3_female/ivector.scp ark:- |" \
-   "cat '$trials' | awk '{print \$1, \$2}' |" foo; local/score_sre08.sh $trials foo
+ivector-plda-scoring --simple-length-normalization=true --num-utts=ark:exp/ivectors_sre08_train_short2_female/num_utts.ark \
```
david-ryan-snyder: Good point. I'll make a new PR for it.

egs/sre08/v1/sid/train_diag_ubm.sh
```diff
@@ -86,6 +85,7 @@ for f in $data/feats.scp $data/vad.scp; do
    [ ! -f $f ] && echo "$0: expecting file $f to exist" && exit 1
 done
 
+parallel_opts="-pe smp $num_threads"
```
danpovey: also, the "-pe smp N" format is not recommended now.  please use --num-threads N" instead.

david-ryan-snyder: OK. This old way shows up in other SID/LID scripts so I'll make a new issue to fix it.

Issue #1031, opened 2016-09-16
Modify TransitionModel for more compact chain-model graphs


danpovey:
This is a rather complex and also high-impact issue which I would really appreciate being able to offload.    @francisr, would you be able to help with this?

In the chain model, we use a topology where the 1st transition of the state has a different pdf-id to the others.  Because the TransitionModel assumes a "HMM-like" model where the emissions are associated with the state, this cannot be implemented in the most natural way with 1 emitting state, and we need to do it with 2 emitting states.  This doubles the size of the graphs, and I just noticed that the chain-model graphs are actually only slightly smaller than the baseline graphs.  This is disappointing and will likely nearly double the graph-searsh computation time, not to mention memory.

I have been thinking about the best way to solve this, and it involves changes in the TransitionModel object.  I don't want to be too general here, because nobody really plays with this stuff much.  The minimal change is that instead of each state being associated with one pdf-class (in the prototype) or pdf-id (in the specific state, after clustering), each state will have, in general, two pdf-classes/pdf-ids; one for "forward transitions" and one for the self-loop.  Note that due to reordering (which is always used), the self-loop actually gets taken after the "forward transitions".  Obviously if we were designing this from scratch it wouldn't be designed this way, but there are historical reasons for the design.
# Changes to HmmTopology

 In the prototype topology format, the pdf-class of a state is specified using `<PdfClass> 0` and the like.  This format will be extended to support saying something like `<PdfClass> 0 <SelfLoopPdfClass> 1`.  If not specified (and for back compatibility in reading), the SelfLoopPdfClass will default to the same value as the PdfClass.  It will be required that in struct HmmState, the pdf_class and self_loop_pdf_class variables will either both be kNoPdf, or neither be kNoPdf (no mixing).  Check that the validation scripts in utils/ support this changed format.  The function NumPdfClasses() will need to be updated.   Documentation will need to be adjusted in doc/.  Note: after this change, the 'reordering' option in graph construction will no longer produce identical graphs with true vs. false.  Some documentation changes will be required.  But we always use 'true' in practice so this is not an issue of practical importance.
# Changes to TransitionModel

[...]

PR #1105, opened 2016-10-10
Modify TransitionModel for more compact chain-model graphs

naxingyu:
Place holder for addressing #1031 . WIP log:
1. self_loop_pdf_class added to HmmState, done
2. self_loop_pdf added to Tuple in TransitionModel. done
3. another branch of ContextDependencyInterface::GetPdfInfo. ugly done
4. create test code for new structures. done
5. back compatability for all read code. done
6. normal HMM validation using RM. done
7. chain code modification. done
8. chain validation using RM. done
9. iterate 2nd version of GetPdfInfo. done
10. documents and comments. tbd...


danpovey:
@naxingyu, any progress on this?  Any intermediate work to push?

naxingyu:
Yes, I'll commit something today.

X.

naxingyu:
@danpovey It's still a heavy draft, lots of todos, tuple-related modification, testing, documentation, etc., but the basic idea is structured. The 2nd version of GetPdfInfo is implemented using brute force enumeration. Please check if I get the intension correct or not.

src/hmm/hmm-topology.cc
```diff
@@ -125,6 +132,8 @@ void HmmTopology::Read(std::istream &is, bool binary) {
       entries_[i].resize(thist_sz);
       for (int32 j = 0 ; j < thist_sz; j++) {
         ReadBasicType(is, binary, &(entries_[i][j].pdf_class));
+        ReadBasicType(is, binary, &(entries_[i][j].self_loop_pdf_class));
```
danpovey: remember the reading code needs to be back compatible to previously written models.

src/hmm/hmm-topology.cc
```diff
@@ -248,7 +261,7 @@ void HmmTopology::Check() {
         if (seen_transition.count(dst_state) != 0)
           KALDI_ERR << "HmmTopology::Check(), duplicate transition found.";
         if (dst_state == k) {  // self_loop...
-          KALDI_ASSERT(entries_[i][j].pdf_class != kNoPdf && "Nonemitting states cannot have self-loops.");
+          KALDI_ASSERT(entries_[i][j].self_loop_pdf_class != kNoPdf && "Nonemitting states cannot have self-loops.");
```
danpovey: line length.

src/hmm/hmm-topology.h
```diff
@@ -99,19 +99,28 @@ class HmmTopology {
     /// but may be different to enable us to hardwire sharing of state, and may be
     /// equal to \ref kNoPdf == -1 in order to specify nonemitting states (unusual).
     int32 pdf_class;
+    int32 self_loop_pdf_class;
```
danpovey: you should probably have a comment for this.

src/hmm/hmm-topology.h
```diff
@@ -99,19 +99,28 @@ class HmmTopology {
...
 
     /// A list of transitions, indexed by what we call a 'transition-index'.
     /// The first member of each pair is the index of the next HmmState, and the
     /// second is the default transition probability (before training).
     std::vector<std::pair<int32, BaseFloat> > transitions;
 
-    explicit HmmState(int32 p): pdf_class(p) { }
+    explicit HmmState(int32 p): pdf_class(p),self_loop_pdf_class(p) { }
+    explicit HmmState(int32 p, int32 self_p) {
```
danpovey: better to rename these to pdf_class and self_loop_pdf_class [self-constructor still works even though it's the same as the variable names.]

src/hmm/hmm-topology.h
```diff
@@ -99,19 +99,28 @@ class HmmTopology {
...
 
     bool operator == (const HmmState &other) const {
-      return (pdf_class == other.pdf_class && transitions == other.transitions);
+      return (pdf_class == other.pdf_class &&
+              self_loop_pdf_class == other.self_loop_pdf_class &&
+              transitions == other.transitions);
     }
 
-    HmmState(): pdf_class(-1) { }
+  HmmState(): pdf_class(-1),self_loop_pdf_class(-1) { }
```
danpovey: space after comma, and put 2 spaces before HmmState().

src/hmm/transition-model.cc
```diff
@@ -24,74 +24,122 @@
...
-  for (size_t i = 0; i < phones.size(); i++) {  // setting up to_hmm_state_list.
-    int32 phone = phones[i];
-    const HmmTopology::TopologyEntry &entry = topo_.TopologyForPhone(phone);
-    for (int32 j = 0; j < static_cast<int32>(entry.size()); j++) {  // for each state...
-      int32 pdf_class = entry[j].pdf_class;
-      if (pdf_class != kNoPdf) {
-        to_hmm_state_list[std::make_pair(phone, pdf_class)].push_back(j);
+
+  if (SelfLoopEqualsForward()) {
+    // this branch deals with when self_loop_pdf_class is always the same as the pdf_class
```
danpovey: you could comment that this is the case for normal models, but not for chain models.

src/hmm/transition-model.cc
```diff
@@ -24,74 +24,122 @@
...
+    }
 
-  // now triples_ is populated with all possible triples of (phone, hmm_state, pdf).
-  std::sort(triples_.begin(), triples_.end());  // sort to enable reverse lookup.
+    for (int32 i = 0; i < static_cast<int32>(pdf_info.size()); i++) {
+      int32 phone = phones[i];
+      for (int32 j = 0; j < static_cast<int32>(pdf_info[i].size()); j++) {
+	int32 pdf_class = pdf_class_pairs[i][j].first,
+          self_loop_pdf_class = pdf_class_pairs[i][j].second;
+        const std::vector<int32> &state_vec = to_hmm_state_list[phone][std::make_pair(pdf_class, self_loop_pdf_class)];
```
danpovey: line length

src/hmm/transition-model.cc
```diff
@@ -24,74 +24,122 @@
...
+        for (size_t k = 0; k < state_vec.size(); k++) {
+          int32 hmm_state = state_vec[k];
+          tuples_.push_back(Tuple(phone, hmm_state, pdf, pdf));
+        }
       }
     }
-  }
+  } else {
+    std::vector<std::vector<std::vector<std::pair<int32, int32> > > > pdf_info;
+    std::vector<std::vector<std::pair<int32, int32> > > pdf_class_pairs;
```
danpovey: need a comment saying what pdf_info represents.-- and of course you have to figure out pdf_class_pairs, and also explain what it is.

naxingyu:
@danpovey comments are resolved. And I have updated the task list [here](https://github.com/kaldi-asr/kaldi/pull/1105#issue-181968927).

danpovey:
Thanks!
I'll check it more thoroughly when it's done, or closer to done.

naxingyu:
While validating the gmm models, I found that this [line](https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/info/gmm_dir_info.pl#L45) only works in Perl 5.12 and later. Modifying it to
```
@ans = readdir $dh;
```
will work for most version of Perl.

naxingyu:
*@danpovey* comments are resolved.
I have updated the task list here
https://github.com/kaldi-asr/kaldi/pull/1105#issue-181968927
please see if it make sense.

danpovey:
Regarding readdir, do you have time to make a separate PR to fix that?  Or is the fix in this PR?
Check the Travis logs, the checks failed.  I'll look into this when it's all finished-- sorry, I'm just super busy right now.  I'm sure any problems will be minor.

naxingyu:
I'm running the rm recipe. Modified version created exactly same model, 
graph and alignments with the original version on mono and tri1 stages. 
Difference occur on tri2b model. Debugging...

X.

在 2016/10/21 10:29, Daniel Povey 写道:

danpovey:
As long as the WER is in the same ballpark it's OK.  None of this is
completely deterministic, esp. if you have machines with different OS
versions or hardware in your cluster.  [differences in BLAS, for instance.]

naxingyu:
I'll make a separate PR for readdir.
About the Travis failure, yes, that was a trial commit. Now there is this weird thing. Please check this when you are not busy. Here it is: We have this inline member [function](https://github.com/kaldi-asr/kaldi/blob/master/src/hmm/transition-model.h#L312). If defined in this way, it returns wrong `pdf_id`, even if the `tuples_` is storing the correct transition information, as I checked. This results in incorrect LDA accumulaters, leading to the tri2b flaw that I reported. Now if I define it as a non-inline member function, put the definition in .cc file, the whole toolkit compiles flawless and the RM results of all non-chain stages are correct, when running on my cluster. However, when I push the change, Travis throws this failure.
```
/home/travis/build/kaldi-asr/kaldi/src/transform/fmpe.cc:540: undefined reference to `kaldi::TransitionModel::TransitionIdToPdf(int) const'
collect2: error: ld returned 1 exit status
```

danpovey:
Probably it didn't work on travis because you left 'inline' in the declaration.
In any case, your original problem could have been fixed by doing 'make depend' and recompiling.  This kind of thing happens when a class size changes due to new members, and code is not recompiled.

naxingyu:
On 2016/10/26 10:47, Daniel Povey wrote:
>
> Probably it didn't work on travis because you left 'inline' in the 
> declaration.
>
No, I removed 'inline' in the declaration. Otherwise it won't work on my 
cluster.
>
> In any case, your original problem could have been fixed by doing 
> 'make depend' and recompiling. This kind of thing happens when a class 
> size changes due to new members, and code is not recompiled.
>
OK I'll try that. Thanks. Will keep you posted.

naxingyu:
I'm stuck in building chain tree for a few days. I started with such a topo.
```
<Topology>
<TopologyEntry>
<ForPhones>
1 2 3 4 5 6 7 8 9 10 11 12 13 ... 192 193
</ForPhones>
<State> 0 <PdfClass> 0 <SelfLoopPdfClass> 1 <Transition> 0 0.5 <Transition> 1 0.5 </State>
<State> 1 </State>
</TopologyEntry>
</Topology>
```
The mono.{mdl,tree} are correctly initialized with tuple-type transitions.
```
Transition-state 1: phone = sil hmm-state = 0 pdf = 0 self-loop-pdf = 1
 Transition-id = 1 p = 0.5 [self-loop]
 Transition-id = 2 p = 0.5 [0 -> 1]
Transition-state 2: phone = sil_B hmm-state = 0 pdf = 0 self-loop-pdf = 1
 Transition-id = 3 p = 0.5 [self-loop]
 Transition-id = 4 p = 0.5 [0 -> 1]
Transition-state 3: phone = sil_E hmm-state = 0 pdf = 0 self-loop-pdf = 1
 Transition-id = 5 p = 0.5 [self-loop]
 Transition-id = 6 p = 0.5 [0 -> 1]
Transition-state 4: phone = sil_I hmm-state = 0 pdf = 0 self-loop-pdf = 1
 Transition-id = 7 p = 0.5 [self-loop]
 Transition-id = 8 p = 0.5 [0 -> 1]
Transition-state 5: phone = sil_S hmm-state = 0 pdf = 0 self-loop-pdf = 1
 Transition-id = 9 p = 0.5 [self-loop]
 Transition-id = 10 p = 0.5 [0 -> 1]
Transition-state 6: phone = aa_B hmm-state = 0 pdf = 2 self-loop-pdf = 3
 Transition-id = 11 p = 0.5 [self-loop]
 Transition-id = 12 p = 0.5 [0 -> 1]
Transition-state 7: phone = aa_E hmm-state = 0 pdf = 2 self-loop-pdf = 3
 Transition-id = 13 p = 0.5 [self-loop]
...
```
And the converted alignments seems good. However, the output tree does not contain `SE -1` nodes, meaning that the pdf-class is never queried in the final tree. Therefore the new GetPdfInfo is pooling same vectors of pdf-ids for pdf-class and self-loop-pdf-class. Currently I'm tracking the BuildTree and mostly the EventMap code to see what went wrong.

danpovey:
The place where it's going wrong might be in acc-tree-stats.
Look at tree-accu.cc line 90 where it does:

        int32 pdf_class = trans_model.TransitionIdToPdfClass(
            split_alignment[i+info.central_position][j]);
        // pdf_class will normally by 0, 1 or 2 for 3-state HMM.


        std::pair<EventKeyType, EventValueType> pr(kPdfClass, pdf_class);
        evec_more.push_back(pr);

The problem might be in the function TransitionIdToPdfClass().
Remember that you can no longer look up the pdf-class via the
transition-state, you have to take into account whether it's a self loop.

Dan

naxingyu:
Nice. A condition of IsSelfLoop() before getting pdf_class works. :)

X.

naxingyu:
Now that I get pdf_ids and self_loop_pdf_ids separately for the central 
phone and the brute-force combination is making more transition states 
than necessary, I see what your proposition really mean...
That's smart. Thanks.

X.

naxingyu:
Suppose, for phone aa_B, I have 14 forward pdfs and 12 self-loop pdfs. In the old format, there will be about 26 transition states. In the new format, I need to find possible combinations in the contexts of this phone from the 168 candidates. A naive progressive algorithm is
```
for phones:
    query forward_pdfs
    if forward_pdf more than 1:
        for phones:
            add phone to right context
            query forward_pdfs
            if forward_pdf more than 1:
                for phones:
                    add phone to left context
                    query forward_pdf (singleton)
                    query self_loop_pdf (singleton)
                    push (forward_pdf, self_loop_pdf) pair to pdf_info
            else:
                query self_loop_pdfs using partial context
                push (forward_pdfs, self_loop_pdfs) pairs to pdf_info
    else:
        query self_loop_pdfs using partial (only central) context
        push (forward_pdfs, self_loop_pdfs) pairs to pdf_info
    sort and uniq pdf_info of this phone
```
Do you think that this could shrink the number of pdf pairs from 168 to nearly half of the old format, ie 13? I'm trying though...

naxingyu:
The example above ended with 53 transition states, compared with 26 transition states using the old topology. And that will double the size of the HMM fst. Is there something wrong with the algorithm?

danpovey:
A larger number of transition-states is expected.  It may not impact the
size of the graph as much as you think-- most states of the graph have
known left and right context as they are parts of linear chains, and having
more transition states does not necessarily blow the graph up
proportionally.

Regarding your algorithm-- I have a couple of issues with it: it treats the
forward and self-loop pdf differently; and it seems to make some
assumptions about the left and right context being symmetric.

I would use a recursive algorithm: something like the following
(I'm not adding all args to the function; assume the central-position P,
the phone list and the tree, etc., are known, or are class members.


// 'context' is the context-window of phones, of
// length N, with -1 for those positions where phones
// that are currently unknown, treated as wildcards; at least
// the central phone [position P] must be a real phone, i.e.
// not -1.
// This function inserts any allowed pairs (forward_pdf, self_loop_pdf)
// to the set "pairs".
void  EnumeratePairs(int32 self_loop_pdf_class,
   int32 forward_pdf_class,
   const std::vector<int32> &context,
   std::unordered_set<pair<int32,int32>, PairHasher> *pairs) {
   int32 position;
   // Choose 'position' as a phone position in 'context' that's currently
-1, and that is as close as possible to the central position P.
   std::vector<int32> new_context(context);
   for each phone {
      new_context[position] = phone;
      std::vector<int32> forward_pdfs, self_loop_pdfs;
      //  Query the tree twice using MultiMap to get the lists of possible
pdfs 'forward_pdfs' and 'self_loop_pdfs' [only the pdf-class will differ in
the 2 calls.]
       if(forward_pdfs.size() == 1 || self_loop_pdfs.size() == 1) {
         // trivially enumerate the pairs and insert them into 'pairs'.
       } else {
          // call EnumeratePairs with 'new_context'
       }
    }
 }

egs/wsj/s5/steps/nnet3/chain/build_tree.sh
```diff
@@ -24,7 +24,7 @@ cluster_thresh=-1  # for build-tree control final bottom-up clustering of leaves
 frame_subsampling_factor=1
 leftmost_questions_truncate=10
 tree_stats_opts=
-cluster_phones_opts=
+cluster_phones_opts="--pdf-class-list=0"
```
danpovey: Was it necessary for some reason to change this?
This will have the effect that for finding the phone classes, it just uses the 1st transition of the phone, rather than the self-loop.  it's not clear to me that this would be better.
Actually, I can see why this would have been necessary when you had that bug and the tree was not being built right, but this is probably not needed now.

egs/wsj/s5/steps/nnet3/chain/build_tree.sh
```diff
@@ -66,9 +66,6 @@ cp $alidir/splice_opts $dir 2>/dev/null # frame-splicing options.
 cp $alidir/cmvn_opts $dir 2>/dev/null # cmn/cmvn option.
 cp $alidir/delta_opts $dir 2>/dev/null # delta option.
 
-utils/lang/check_phones_compatible.sh $lang/phones.txt $alidir/phones.txt || exit 1;
```
danpovey: is there a reason you deleted these lines?  Perhaps something went wrong in merge?

egs/wsj/s5/steps/nnet3/chain/gen_topo.py
```diff
@@ -41,9 +41,9 @@
 # We make the transition-probs 0.5 so they normalize, to keep the code happy.
 # In fact, we always set the transition probability scale to 0.0 in the 'chain'
 # code, so they are never used.
-print("<State> 0 <PdfClass> 0 <Transition> 1 0.5 <Transition> 2 0.5 </State>")
-print("<State> 1 <PdfClass> 1 <Transition> 1 0.5 <Transition> 2 0.5 </State>")
-print("<State> 2 </State>")
+print("<State> 0 <PdfClass> 0 <SelfLoopPdfClass> 1 <Transition> 0 0.5 <Transition> 1 0.5 </State>")
+#print("<State> 1 <PdfClass> 1 <Transition> 1 0.5 <Transition> 2 0.5 </State>")
```
danpovey: you can remove this commented-out line.  But you might want to put a comment at the top saying that this script was modified around such-and-such a date, when the code was extended to support having a different pdf-class on the self loop.

src/hmm/hmm-topology.cc
```diff
@@ -177,11 +199,13 @@ void HmmTopology::Write(std::ostream &os, bool binary) const {
   } else {
     WriteIntegerVector(os, binary, phones_);
     WriteIntegerVector(os, binary, phone2idx_);
+    if (!is_hmm) WriteBasicType(os, binary, static_cast<int32>(-1));
```
danpovey: comment that this -1 is a signal that the object has the new, extended format with SelfLoopPdfClass.

src/hmm/transition-model.cc
```diff
@@ -24,74 +24,137 @@
...
-  // can correspond to.
-  for (size_t i = 0; i < phones.size(); i++) {  // setting up to_hmm_state_list.
-    int32 phone = phones[i];
-    const HmmTopology::TopologyEntry &entry = topo_.TopologyForPhone(phone);
-    for (int32 j = 0; j < static_cast<int32>(entry.size()); j++) {  // for each state...
-      int32 pdf_class = entry[j].pdf_class;
-      if (pdf_class != kNoPdf) {
-        to_hmm_state_list[std::make_pair(phone, pdf_class)].push_back(j);
+
+  if (IsHmm()) {
```
danpovey: This function is too long.. you should probably break it out into a couple of different functions at least.  If you don't need to export something to the header you can declare the function static.

src/tree/context-dep.cc
```diff
@@ -179,6 +179,50 @@ void ContextDependency::Read (std::istream &is, bool binary) {
 }
 
 void ContextDependency::GetPdfInfo(const std::vector<int32> &phones,
```
danpovey: put a newline after ( and move things to the left so that you don't go so much above 80 characters.

src/tree/context-dep.cc
```diff
@@ -179,6 +179,50 @@ void ContextDependency::Read (std::istream &is, bool binary) {
...
+
+      vec[0] = std::make_pair(static_cast<EventKeyType>(P_),
+                              static_cast<EventValueType>(phone));
+      vec[1] = std::make_pair(kPdfClass, static_cast<EventValueType>(self_loop_pdf_class));
+      std::sort(vec.begin(), vec.end());
+      std::vector<EventAnswerType> self_loop_pdfs;  // self-loop pdfs that can be at this pos as this phone.
+      to_pdf_->MultiMap(vec, &self_loop_pdfs);
+      SortAndUniq(&self_loop_pdfs);
+
+      // brute force enumeration
```
danpovey: Let's not use the term 'brute force' for this.. what I meant by brute force was, finding the exact list of pairs by enumerating all triphones (or n-phones, in general).  Let's call this 'overgeneration' of pairs.

danpovey:
@naxingyu, any progress with this?  I'm looking forward to finding out how much smaller the graph is (and hopefully there will be a reduction in decoding time as well, and maybe even small WER improvements due to better pruning).

naxingyu:
I'm testing the recursive algorithm and found that there might be a problem with it. The algorithm might not find all possible pdf pairs, e.g. the pairs that only appear at phone sequence borders. This will cause a convert-ali failure. For instance, the alignment of this utterance in the RM training set
```
ahh05_sr301_trn  [ 2 8 5 18 17 17 17 17 17 ] [ 11392 11391 11391 11391 11400 11399 11399 11399 11420 11419 ] [ 824 823 823 852 851 874 ] [ 10290 10289 10302 10328 10327 ] [ 5422 5462 5461 5461 5461 5461 5461 5461 5510 5509 5509 5509 5509 ] [ 11852 11858 11857 11880 11879 ] [ 3166 3165 3184 3192 3191 ] [ 1634 1640 1639 1746 ] [ 7670 7669 7692 7691 7691 7691 7691 7691 7746 7745 7745 7745 7745 ] [ 4750 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4798 4797 4797 4806 4805 4805 ] [ 7390 7389 7438 7466 7465 7465 7465 ] [ 346 345 376 375 382 ] [ 7756 7755 7796 7818 7817 ] [ 11160 11174 11173 11173 11173 11173 11194 11193 11193 11193 ] [ 3742 3741 3741 3782 3781 3781 3810 3809 3809 3809 3809 3809 3809 ] [ 9050 9049 9162 9161 9161 9161 9166 9165 9165 ] [ 6156 6155 6220 6219 6262 6261 ] [ 1870 1908 1907 2006 ] [ 9532 9531 9531 9531 9548 9547 9547 9600 9599 9599 9599 ] [ 8748 8747 8747 8747 8747 8747 8747 8747 8808 8807 8807 8807 8807 8854 8853 8853 ] [ 4748 4747 4747 4747 4747 4786 4785 4810 ] [ 3428 3427 3448 3447 3466 ] [ 5656 5655 5702 5701 5701 5701 5752 5751 ] [ 8130 8129 8134 8133 8138 8137 8137 8137 8137 8137 8137 ] [ 11850 11849 11849 11849 11862 11861 11861 11861 11878 11877 11877 11877 11877 11877 ] [ 2 1 1 1 8 18 ] [ 1396 1395 1395 1395 1395 1400 1399 1399 1399 1399 1399 1399 1399 1496 1495 1495 1495 1495 1495 1495 ] [ 11216 11222 11221 11258 11257 11257 ] [ 3170 3180 3192 3191 ] [ 1606 1644 1643 1643 1748 ] [ 6512 6511 6511 6511 6511 6532 6531 6531 6531 6546 6545 6545 6545 6545 6545 6545 6545 ] [ 9074 9073 9136 9135 9210 9209 9209 9209 9209 ] [ 10982 10981 10981 11016 11015 11038 11037 11037 ] [ 11930 11929 11929 11954 11953 11964 11963 11963 ] [ 4230 4229 4229 4229 4310 4309 4309 4340 4339 4339 4339 4339 ] [ 2 8 18 ] [ 5424 5423 5423 5474 5473 5473 5473 5473 5473 5473 5490 5489 5489 5489 ] [ 7756 7755 7755 7796 7818 7817 7817 ] [ 2468 2467 2486 2485 2520 2519 2519 2519 2519 ] [ 520 519 519 519 519 519 519 519 519 519 519 519 519 519 519 578 577 577 598 597 597 ] [ 9506 9505 9505 9552 9551 9616 ] [ 9422 9458 9457 9457 9457 9457 9496 9495 ] [ 10372 10371 10371 10406 10448 10447 ] [ 2772 2780 2779 2779 2792 ] [ 9054 9164 9163 9163 9163 9163 9163 9163 9163 9194 9193 ] [ 4748 4747 4747 4780 4779 4779 4779 4779 4779 4779 4779 4820 4819 4819 ] [ 10276 10300 10299 10299 10336 10335 10335 10335 10335 10335 10335 ]
ahh05_sr301_trn  sil                         w_B                                                             ah_I                        td_E                              ih_B                                                                 z_E                               dh_B                         ax_E                    n_B                                                                  ey_I                                                                                                    m_E                                    ae_B                    n_E                          v_B                                                             eh_I                                                                 r_I                                              iy_I                              ax_I                    s_E                                                        r_B                                                                                 ey_I                                        dx_I                         ih_I                                        ng_I                                                       z_E                                                                                     sil              ax_B                                                                                                    v_E                                     dh_B                    ax_E                         k_B                                                                                      r_I                                              uw_I                                                z_I                                                 er_E                                                            sil        ih_B                                                                      n_E                                    b_B                                              ae_I                                                                                    s_E                               s_B                                         td_I                                    ch_I                         r_I                                                        ey_I                                                                      td_E
```
does not end with sil. The final phone window is (76, 159, 0). The pdf pair is (216, 373). It only appear in this context. However, the recursive algorithm generate the pairs from (76, 159, 1) to (96, 159, 194), which doesn't include this one.

naxingyu:
I'm testing the recursive algorithm and found that there might be a problem with it. The algorithm might not find all possible pdf pairs, e.g. the pairs that only appear at phone sequence borders. This will cause a convert-ali failure. For instance, the alignment of this utterance in the RM training set
```
ahh05_sr301_trn  [ 2 8 5 18 17 17 17 17 17 ] [ 11392 11391 11391 11391 11400 11399 11399 11399 11420 11419 ] [ 824 823 823 852 851 874 ] [ 10290 10289 10302 10328 10327 ] [ 5422 5462 5461 5461 5461 5461 5461 5461 5510 5509 5509 5509 5509 ] [ 11852 11858 11857 11880 11879 ] [ 3166 3165 3184 3192 3191 ] [ 1634 1640 1639 1746 ] [ 7670 7669 7692 7691 7691 7691 7691 7691 7746 7745 7745 7745 7745 ] [ 4750 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4749 4798 4797 4797 4806 4805 4805 ] [ 7390 7389 7438 7466 7465 7465 7465 ] [ 346 345 376 375 382 ] [ 7756 7755 7796 7818 7817 ] [ 11160 11174 11173 11173 11173 11173 11194 11193 11193 11193 ] [ 3742 3741 3741 3782 3781 3781 3810 3809 3809 3809 3809 3809 3809 ] [ 9050 9049 9162 9161 9161 9161 9166 9165 9165 ] [ 6156 6155 6220 6219 6262 6261 ] [ 1870 1908 1907 2006 ] [ 9532 9531 9531 9531 9548 9547 9547 9600 9599 9599 9599 ] [ 8748 8747 8747 8747 8747 8747 8747 8747 8808 8807 8807 8807 8807 8854 8853 8853 ] [ 4748 4747 4747 4747 4747 4786 4785 4810 ] [ 3428 3427 3448 3447 3466 ] [ 5656 5655 5702 5701 5701 5701 5752 5751 ] [ 8130 8129 8134 8133 8138 8137 8137 8137 8137 8137 8137 ] [ 11850 11849 11849 11849 11862 11861 11861 11861 11878 11877 11877 11877 11877 11877 ] [ 2 1 1 1 8 18 ] [ 1396 1395 1395 1395 1395 1400 1399 1399 1399 1399 1399 1399 1399 1496 1495 1495 1495 1495 1495 1495 ] [ 11216 11222 11221 11258 11257 11257 ] [ 3170 3180 3192 3191 ] [ 1606 1644 1643 1643 1748 ] [ 6512 6511 6511 6511 6511 6532 6531 6531 6531 6546 6545 6545 6545 6545 6545 6545 6545 ] [ 9074 9073 9136 9135 9210 9209 9209 9209 9209 ] [ 10982 10981 10981 11016 11015 11038 11037 11037 ] [ 11930 11929 11929 11954 11953 11964 11963 11963 ] [ 4230 4229 4229 4229 4310 4309 4309 4340 4339 4339 4339 4339 ] [ 2 8 18 ] [ 5424 5423 5423 5474 5473 5473 5473 5473 5473 5473 5490 5489 5489 5489 ] [ 7756 7755 7755 7796 7818 7817 7817 ] [ 2468 2467 2486 2485 2520 2519 2519 2519 2519 ] [ 520 519 519 519 519 519 519 519 519 519 519 519 519 519 519 578 577 577 598 597 597 ] [ 9506 9505 9505 9552 9551 9616 ] [ 9422 9458 9457 9457 9457 9457 9496 9495 ] [ 10372 10371 10371 10406 10448 10447 ] [ 2772 2780 2779 2779 2792 ] [ 9054 9164 9163 9163 9163 9163 9163 9163 9163 9194 9193 ] [ 4748 4747 4747 4780 4779 4779 4779 4779 4779 4779 4779 4820 4819 4819 ] [ 10276 10300 10299 10299 10336 10335 10335 10335 10335 10335 10335 ]
ahh05_sr301_trn  sil                         w_B                                                             ah_I                        td_E                              ih_B                                                                 z_E                               dh_B                         ax_E                    n_B                                                                  ey_I                                                                                                    m_E                                    ae_B                    n_E                          v_B                                                             eh_I                                                                 r_I                                              iy_I                              ax_I                    s_E                                                        r_B                                                                                 ey_I                                        dx_I                         ih_I                                        ng_I                                                       z_E                                                                                     sil              ax_B                                                                                                    v_E                                     dh_B                    ax_E                         k_B                                                                                      r_I                                              uw_I                                                z_I                                                 er_E                                                            sil        ih_B                                                                      n_E                                    b_B                                              ae_I                                                                                    s_E                               s_B                                         td_I                                    ch_I                         r_I                                                        ey_I                                                                      td_E
```
does not end with sil. The final phone window is (76, 159, 0). The pdf pair is (216, 373). It only appear in this biphone context. However, the recursive algorithm only generate the pairs from all triphones, (76, 159, 1) to (96, 159, 194).

danpovey:
Oh yes, that was an oversight on my part.
You need to add 0 to the list of phones to be used in context positions
[but not in the central positions.]

naxingyu:
Old RM chain graph:
```
# of states                                       37641
# of arcs                                         110261
initial state                                     0
# of final states                                 565
# of input/output epsilons                        18921
# of input epsilons                               19029
# of output epsilons                              81285
# of accessible states                            37641
# of coaccessible states                          37641
# of connected states                             37641
# of connected components                         1
# of strongly conn components                     2744
```
New RM chain graph
```
# of states                                       16233
# of arcs                                         56734
initial state                                     0
# of final states                                 307
# of input/output epsilons                        7511
# of input epsilons                               7534
# of output epsilons                              40009
# of accessible states                            16233
# of coaccessible states                          16233
# of connected states                             16233
# of connected components                         1
# of strongly conn components                     1195
```

naxingyu:
@danpovey compile fail due to `No space left on device` error.

danpovey:
I'm testing a pull request that builds the Travis stuff with shared
libraries-- should I hope fix this problem.

danpovey:
BTW, great news about the size of the graph.
I assume this was the HCLG decoding graph.

naxingyu:
Yes, that was the HCLG decoding graph. However, the WER doesn't look so 
good.
%WER 2.87 [ 360 / 12533, 57 ins, 67 del, 236 sub ] 
exp/chain/tdnn_5f/decode/wer_3_1.0
%WER 3.16 [ 396 / 12533, 51 ins, 89 del, 256 sub ] 
exp_newtrans/chain/tdnn_5o/decode/wer_4_1.0
Seems that the silence arcs are under-weighted, giving a higher del/err 
and lower ins/err.
Should I change the self-loop-scale? Does it make a difference?

X.

danpovey:
Self-loop-scale and transition-scale should be set to 1.0 in all
chain-model graph creation.  The difference could plausibly be due to
noise.  Maybe test on a bigger setup?
Check that the objective values you get during training are similar to
before.

danpovey:
I think I know why the WER might be bad.
In chain-den-graph.cc there is the line
```
  bool reorder = false;
```
This needs to be changed to true, otherwise there is a train/test mismatch that would affect the results.

naxingyu:
I've already changed it in this build.

danpovey:
... and when you do that, check the effect on the chain den-graph size in states and arcs [and also compare it with the size before the code change.]  It may be necessary to change DenGraphMinimizeWrapper to try whether doing the reversed minimization *before* the regular minimization is better [in terms of number of states and arcs.]

danpovey:
I don't see it in the diff on github... make sure to push your changes.

src/hmm/hmm-topology.h
```diff
@@ -124,6 +138,7 @@ class HmmTopology {
   // Checks that the object is valid, and throw exception otherwise.
   void Check();
 
+  bool IsHmm() const;
```
danpovey: please document this, e.g.
/// Returns true if this HmmTopology is really 'hmm-like', i.e. the pdf-class on
/// the self-loops and forward transitions of all states are identical.  [note: in HMMs,
/// the densities are associated with the states.]  We have extended this to 
/// support 'non-hmm-like' topologies (where those pdf-classes are different),
/// in order to make for more compact decoding graphs in our so-called 'chain models'
/// (AKA lattice-free MMI), where we use 1-state topologies that have different pdf-classes
/// for the self-loop and the forward transition.  Note that we always use the 'reorder=true'
/// option so the 'forward transition' actually comes before the self-loop.

naxingyu:
Just committed. Please start another round of review. And if you think 
the code looks fine. I'll start testing on a larger set, e.g. swbd.

X.

danpovey:
I think it's unlikely that there will be any issues in the code that would
be serious enough to require re-testing -> test on the larger setup.
Be aware that I committed something today which may affect BLSTM results
(and those are always unstable anyway), so better to use TDNNs.  Of course,
make sure that if you compare with a baseline TDNN system, it was trained
after the per-component-max-change thing was merged (that was a few weeks
ago I think).

src/chain/chain-den-graph.cc
```diff
@@ -220,6 +220,30 @@ static void SortOnTransitionCount(fst::StdVectorFst *fst) {
...
+  KALDI_LOG << "Number of states and arcs in transition-id FST after "
+            << "removing any epsilons introduced by reversal is "
+            << fst->NumStates() << " and "
+            << NumArcs(*fst);
+  fst::PushSpecial(fst, fst::kDelta * 0.01);
+}
+  /*
+void DenGraphMinimizeWrapper(fst::StdVectorFst *fst) {
+  for (int32 i = 1; i <= 3; i++) {
     fst::PushSpecial(fst, fst::kDelta * 0.01);
```
danpovey: please remove the comment once you decide which setting is better.

src/hmm/hmm-utils.cc
```diff
@@ -49,8 +49,8 @@ fst::VectorFst<fst::StdArc> *GetHmmAsFst(
   const HmmTopology &topo = trans_model.GetTopo();
   const HmmTopology::TopologyEntry &entry  = topo.TopologyForPhone(phone);
 
-  // vector of the pdfs, indexed by pdf-class (pdf-classes must start from zero
-  // and be contiguous).
+  // vector of the pdfs, indexed by pdf-class and self-loop pdf-class
```
danpovey: you can revert this change in wording.  pdf-class and self-loop pdf-class are not different types of index: self-loop pdf-class is the pdf-class that is on the self-loop.

src/hmm/hmm-utils.cc
```diff
@@ -94,10 +94,15 @@ fst::VectorFst<fst::StdArc> *GetHmmAsFst(
        hmm_state < static_cast<int32>(entry.size());
        hmm_state++) {
     int32 pdf_class = entry[hmm_state].pdf_class, pdf;
-    if (pdf_class == kNoPdf) pdf = kNoPdf;  // nonemitting state.
-    else {
+    int32 self_loop_pdf_class = entry[hmm_state].pdf_class, self_loop_pdf;
```
danpovey: I think you mean entry[hmm_state].self_loop_pdf_class.  But I think you don't need this variable.
GetHmmAsFst does not create the self-loops.

src/itf/context-dep-itf.h
```diff
@@ -60,12 +60,36 @@ class ContextDependencyInterface {
...
+  virtual void GetPdfInfo(
+      const std::vector<int32> &phones,
+      const std::vector<std::vector<std::pair<int32, int32> > > &pdf_class_pairs,
+      std::vector<std::vector<std::vector<std::pair<int32, int32> > > > *pdf_info)
+      const = 0;
+
   /// GetPdfInfo returns a vector indexed by pdf-id, saying for each pdf which
   /// pairs of (phone, pdf-class) it can correspond to.  (Usually just one).
   /// c.f. hmm/hmm-topology.h for meaning of pdf-class.
-  virtual void GetPdfInfo(const std::vector<int32> &phones,  // list of phones
```
danpovey: Clarify that this is the old, simpler interface of GetPdfInfo(), and that this one can only be called if the HmmTopology object's IsHmm() function call returns true.  This should probably be above the other one- more readable that way.

src/tree/context-dep.cc
```diff
@@ -178,9 +178,105 @@ void ContextDependency::Read (std::istream &is, bool binary) {
...
+    // Choose 'position' as a phone position in 'context' that's currently
+    // 0, and that is as close as possible to the central position P.
+    int32 min_dist = N_ + 1;
+    for (int32 i = 0; i < N_; i++) {
+      int32 dist = (P_ - i > 0) ? (P_ - i) : (i - P_);
+      if (phone_window[i] == 0 && dist < min_dist) {
+        position = i;
+        min_dist = dist;
+      }
+    }
```
danpovey: assert that min_dist < N_.

src/tree/context-dep.cc
```diff
@@ -178,9 +178,105 @@ void ContextDependency::Read (std::istream &is, bool binary) {
...
+          phone_window[win_start] = 0;
+	else
+	  phone_window[win_start] = phone;
+      }
+      unordered_set<std::pair<int32, int32>, PairHasher<int32> > pairs;
+      EnumeratePairs(phones, self_loop_pdf_class, pdf_class, phone_window, &pairs);
+      unordered_set<std::pair<int32, int32>, PairHasher<int32> >::iterator iter = pairs.begin(),
+                           end = pairs.end();
+      for (; iter != end; ++iter)
+        (*pdf_info)[phone][j].push_back(*iter);
```
danpovey: Before exiting the function you might want to sort all of these output vectors.  That will make the order deterministic... might be useful in some circumstances for testing purposes.

src/tree/context-dep.h
```diff
@@ -20,6 +20,7 @@
 #ifndef KALDI_TREE_CONTEXT_DEP_H_
 #define KALDI_TREE_CONTEXT_DEP_H_
 
+#include <set>
```
danpovey: No need for this include, as I think you do not use 'set'.  If you want, you can include util/stl-utils.h, which includes unordered_set (with suitable ifdefs).

danpovey:
Also make sure that the objf-impr in the tree is about the same, and the
x-entropy objective function is about the same.. these will guaratee that
nothing weird is going on with the tree.

naxingyu:
The objf-impr in the build-tree is exactly the same. The final xent objf on valid of the old model is
```
Overall log-probability for 'output' is -0.0543164 + -0.00827959 = -0.062596 per frame, over 20000 frames.
Overall log-probability for 'output-xent' is -1.50852 per frame, over 20000 frames.
```
and the new one is
```
Overall log-probability for 'output' is -0.0652472 + -0.00641051 = -0.0716577 per frame, over 20000 frames.
Overall log-probability for 'output-xent' is -1.27734 per frame, over 20000 frames.
```

danpovey:
How recently did you run the baseline with the old model?  Perhaps it was
affected by the per-comopnent max-change, if you ran it a long time ago.

Also, please show me the same quantities on the training-data subset (and
maybe from one of the final training logs too), otherwise it's hard to
judge the meaning of those differences.

src/hmm/hmm-utils.cc
danpovey: OK, then it's a bug-- I think you meant
int32 self_loop_pdf_class = entry[hmm_state].**self_loop**_pdf_class

danpovey: ... I'm surprised that would not have led to crashes later on, e.g. failures of TupleToTransitionState().

danpovey:
... also see if there are any documentation changes needed, e.g. search in doc/ for 'pdf-class', 'pdf_class', and 'triple'.

naxingyu:
Yes I've fix this locally.

danpovey:
ok-- could that have been affecting the RM results?

naxingyu:
No, I fixed that before I got the results.

danpovey:
going to bed now, but please push your latest changes when you're done, and show more diagnostics on the RM setup (e.g. train-set xent objf).

naxingyu:
Done. I'll send you the detailed report by email.

danpovey:
I just noticed that the RM recipe seems rather outdated, it's maybe not the best starting point.
The tedlium s5_r2 recipe is more up-to-date, using that as a starting point might be a better idea.

src/hmm/transition-model.h
```diff
@@ -315,7 +323,7 @@ inline int32 TransitionModel::TransitionIdToPdf(int32 trans_id) const {
   KALDI_ASSERT(static_cast<size_t>(trans_id) < id2state_.size() &&
                "Likely graph/model mismatch (graph built from wrong model?)");
   int32 trans_state = id2state_[trans_id];
-  return triples_[trans_state-1].pdf;
```
danpovey: There is a problem in this function.  It needs to check whether it's a self-loop pdf, and possibly return self_loop_pdf.  BTW, I think we should consider changing the 'pdf' in the tuples and in the HMM-states to 'forward_pdf'.  This might be clearer.  What do you think?

danpovey: Regarding TransitionIdToPdf(), the speed is going to be a problem if there is going to be an extra if-statement in there, as this gets called very frequently during decoding.  It will be better if you introduce a new derived parameter called id2pdf_, mapping from transition-id directly to pdf, then it's just a lookup.

src/hmm/transition-model.h
```diff
@@ -141,19 +141,21 @@ class TransitionModel {
...
   int32 TransitionStateToPhone(int32 trans_state) const;
   int32 TransitionStateToHmmState(int32 trans_state) const;
   int32 TransitionStateToPdf(int32 trans_state) const;
+  int32 TransitionStateToSelfLoopPdf(int32 trans_state) const;
   int32 SelfLoopOf(int32 trans_state) const;  // returns the self-loop transition-id, or zero if
   // this state doesn't have a self-loop.
 
   inline int32 TransitionIdToPdf(int32 trans_id) const;
   int32 TransitionIdToPhone(int32 trans_id) const;
   int32 TransitionIdToPdfClass(int32 trans_id) const;
```
danpovey: Just noticed this (after debugging)... this function [TransitionIdToPdfClass] needs to check whether it's a self loop and return a value accordingly.  I.e. it should return the; and there should probably be no function TransitionIdToSelfLoopPdfClass [its existence makes things a bit unclear; and I doubt there would be a legitimate use for it.]

danpovey:
Yes, please, write the old format when possible.  This will prevent Kaldi
version-compatibility problems in the normal case.

src/hmm/transition-model.h
```diff
@@ -141,17 +141,20 @@ class TransitionModel {
...
-  int32 TripleToTransitionState(int32 phone, int32 hmm_state, int32 pdf) const;
+  int32 TupleToTransitionState(int32 phone, int32 hmm_state, int32 pdf, int32 self_loop_pdf) const;
   int32 PairToTransitionId(int32 trans_state, int32 trans_index) const;
   int32 TransitionIdToTransitionState(int32 trans_id) const;
   int32 TransitionIdToTransitionIndex(int32 trans_id) const;
   int32 TransitionStateToPhone(int32 trans_state) const;
   int32 TransitionStateToHmmState(int32 trans_state) const;
+  int32 TransitionStateToForwardPdfClass(int32 trans_state) const;
+  int32 TransitionStateToSelfLoopPdfClass(int32 trans_state) const;
   int32 TransitionStateToPdf(int32 trans_state) const;
```
danpovey: rename to TransitionStateToForwardPdf

src/hmm/transition-model.h
```diff
@@ -141,17 +141,20 @@ class TransitionModel {
...
+  int32 TransitionStateToSelfLoopPdf(int32 trans_state) const;
   int32 SelfLoopOf(int32 trans_state) const;  // returns the self-loop transition-id, or zero if
   // this state doesn't have a self-loop.
 
-  inline int32 TransitionIdToPdf(int32 trans_id) const;
+  int32 TransitionIdToPdf(int32 trans_id) const;
   int32 TransitionIdToPhone(int32 trans_id) const;
   int32 TransitionIdToPdfClass(int32 trans_id) const;
```
danpovey: this name TransitionIdToPdfClass stays the same, as does TransitionIdToPdf, because it could be forward or self-loop.

naxingyu:
I see what you mean. Thanks!

X.

在 2016/11/18 5:18, Daniel Povey 写道:

danpovey:
Have you tested that the old code can work with the transition-models produced by the new code for the standard GMM models (i.e. with normal topology)?

src/doc/hmm.dox
```diff
@@ -125,7 +130,8 @@ context dependent HMM state depend on the following four things (you could view
 them as a 4-tuple):
```
danpovey: It would be clearer to describe this as 5-tuple... would lead to less potential confusion with the 4-tuple describing the transition-state.

src/doc/hmm.dox
```diff
@@ -198,7 +204,7 @@ prototype HMM (as given in the HmmTopology object).
       from (transition-state, transition-index) to transition-id, and vice versa.
 
   There are also in the transition-modeling code reference to the following concepts:
-    - A triple means a triple (phone, hmm-state, pdf) which is mappable to and from a transition-state.
+    - A tuple means a tuple (phone, hmm-state, forward pdf, self-loop pdf) which is mappable to and from a transition-state.
```
danpovey: Change this ay "A tuple means [by default] a 4-tuple.. "

src/doc/hmm.dox
```diff
@@ -92,7 +92,12 @@ great loss.
 The pdf-class is a concept that relates to the HmmTopology object.  The
 HmmTopology object specifies a prototype HMM for each phone.  Each
 numbered state of a
-"prototype HMM" has a variable "pdf_class".  If two states have the same
+"prototype HMM" has two variables "forward_pdf_class" and "self_loop_pdf_class".
+The "self_loop_pdf_class" is a kind of pdf-class that is associated
+with self-loop transition. It is by default identical to "forward_pdf_class",
+but it can be used to define less-convectional HMM topologies
+where the pdfs on the self-loop and forward transitions are different.
```
danpovey: After "different", write the following sentence:
(The decision to allow the pdf-class on just the self-loop to be different, while not embracing a fully "arc-based" representation where the pdfs on all transitions in the HMM are potentially independent, was made as a compromise, to allow for compatibility with previous versions of Kaldi while supporting the topology used in our "chain models" AKA lattice-free MMI).

danpovey:
Please check that all example scripts are executable.

naxingyu:
I have tested that all example scripts are executable (well maybe not all of them but the mostly used ones are tested, such as mono/tri gmms, lda/mllt/sat training, sgmms, the mmi/fmmi/mpe training scripts, chain model). The old code can work with the new transition model. And here are the detailed results, on RM and SWBD respectively. (markdown table view on Github)

| topo | WER | Average RTF | decoding graph | den fst |
| ------|------|---------------|-----------------|----------|
| old(5g) |  2.86  | 0.065548  | 37641 states, 110261 arcs | 6659 states, 53024 arcs |
| new(5n) | 2.71 | 0.051859 | 16233 states, 56734 arcs | 2943 states, 21590 arcs |

| topo | WER (swbd) | Average RTF | decoding graph | den fst |
| ------|------|---------------|-----------------|----------|
| old(7f) | 10.6  | 0.517429  | 3898357 states, 9129187 arcs | 9045 states, 111243 arcs |
| new(7n) | 10.4 | 0.461376 | 1057614 states, 3962312 arcs | 8265 states, 109004 arcs |

So after this modification, for the chain models, the graph is smaller, decoding is faster and accuracy is slightly higher. For normal HMMs, the models and graphs remain unchanged. I'll make the document changes on Monday and try rebase as you suggested.

jtrmal:
I wonder what is the source of the improvement?
y.

jtrmal:
I mean the WER improvement. y.

danpovey:
We do expect a slight improvement in the beam search (e.g. there is a fixed
max-states and there are now half the umber of states).
Xingyu, by executable, I mean the UNIX file permissions; one of your
example scripts did not have the executable bit set.

naxingyu:
@danpovey rebase works. ready now.

src/tree/context-dep.h
```diff
@@ -99,16 +100,57 @@ class ContextDependency: public ContextDependencyInterface {
...
  private:
   int32 N_;  //
   int32 P_;
   EventMap *to_pdf_;  // owned here.
 
+  // 'context' is the context-window of phones, of
+  // length N, with -1 for those positions where phones 
+  // that are currently unknown, treated as wildcards; at least 
+  // the central phone [position P] must be a real phone, i.e. 
+  // not -1.
```
danpovey: I'm a little confused now about how you solved the issue about word-boundaries.
The comment says that -1 is used for positions where we have not yet expanded the phone.
However, the actual code seems to be using 0 for the "not-expanded"/wildcard phone.
But for the left and right context when we hit the edge of the file, the phone should be set to zero.  Not including zero as an option for positions other than position P_ would cause a problem.  But I don't see how this can happen, given that you are (in the code) using 0 for the wildcard.

src/tree/context-dep.cc
```diff
@@ -178,9 +178,107 @@ void ContextDependency::Read (std::istream &is, bool binary) {
...
+  for (size_t i = 0 ; i < phones.size(); i++) {
+    // loop over phones
+    int32 phone = phones[i];
+    (*pdf_info)[phone].resize(pdf_class_pairs[phone].size());
+    for (size_t j = 0; j < pdf_class_pairs[phone].size(); j++) {
+      // loop over pdf_class pairs
+      int32 pdf_class = pdf_class_pairs[phone][j].first,
+            self_loop_pdf_class = pdf_class_pairs[phone][j].second;
+      for (size_t win_start = 0; win_start < phone_window.size(); win_start++) {
+	if (win_start != P_)
```
danpovey: wrong indentation... and I think you should use -1 for the wildcard positions here, so you can use 0 for the context.  And then you should probably modify EnumeratePairs() so that when it loops over the phones, it also includes 0 [meaning: hitting the edge of the file] in the context.  I'm surprised this hasn't caused a crash... I thought you had been hit by this issue before and fixed it, so now I wonder what the fix was.

naxingyu:
I forgot to update this comment... Yes I chose to use 0 for wildcards so 
that pdfs on boundaries can be included by simply Map other than 
MultiMap. I'll update this comment.

X.

danpovey:
No, the code is not right... it will lead to a crash for some trees.  You
need to make the wildcard phone -1, and include 0 as one of the possible
phones when recurse in line 239.

src/tree/context-dep.cc
```diff
@@ -178,9 +178,107 @@ void ContextDependency::Read (std::istream &is, bool binary) {
...
+    int32 self_loop_pdf_class, int32 forward_pdf_class,
+    const std::vector<int32> &phone_window,
+    unordered_set<std::pair<int32, int32>, PairHasher<int32> > *pairs) const {
+  std::vector<int32> new_phone_window(phone_window);
+  EventType vec;
+
+  std::vector<EventAnswerType> forward_pdfs, self_loop_pdfs;
+
+  int32 forward_pdf, self_loop_pdf;
+  if (Compute(phone_window, forward_pdf_class, &forward_pdf) &&
```
danpovey: Oh I see what you meant, with this code here, that's what prevented it from crashing, and it could be right.
But still, I think it would be better if you did it as I said-- that would be clearer, and may prevent unnecessary recursion.

danpovey: ... and no need to re-test fully. It should lead to an identical transition-model, and you can do a 'diff' after rerunning the stage when you get the transition-model from the tree and the topology.

naxingyu:
Got it.

danpovey:
Also, after that, as one final check before I merge, can you please run the RM setup but removing the left-biphone-related options?  I want to make sure it doesn't crash for triphone.  No need to check anything in related to that-- just run it and make sure there are no crashes.

danpovey:
What happens specifically?

danpovey:
... BTW, you shouldn't include 0 for the P'th position, only for the other
positions.

naxingyu:
No I didn't include 0 for the P'th position. When I use a event vector 
such as
(-1,0),(0,76),(1,159),(2,0) to hit the edge, it return a vector of pdfs 
that looks like
(-1683789, 1, 79, 2589367, 167, ....)
But when I do it using (-1,0),(0,76),(1,159),(2,0), it returns
(1, 79, 167, ...) correctly.

naxingyu:
The RM setup is triphone. I've tested left-biphone system on swbd, as 
shown in the table above.

X.

danpovey:
Merging.  Thanks!!

PR #1113, opened 2016-10-12
"--extra-word-disambig-syms" for prepare_lang.sh

horndasch:
The new command line option '--extra-word-disambig-syms disambig_syms.txt'
for prepare_lang.sh adds the possibility to add extra word-level
disambiguation symbols to the list of automatic disambiguation symbols.

The extra disambiguation symbols can be used during the modification of
G.fst graphs, e.g. when embedding word classes (a word-class disambiguation
symbol can for example be transduced to <eps> when entering or leaving a
word-class sub-language model).

### DIFF
--- a/egs/wsj/s5/utils/prepare_lang.sh
+++ b/egs/wsj/s5/utils/prepare_lang.sh
@@ -151,9 +151,9 @@ fi
 
 # In case there are extra word-level disambiguation symbols we need
 # to make sure that all symbols in the provided file are valid.
-if [[ ! -z $extra_word_disambig_syms ]]; then
-  if ! utils/validate_disambig_sym_file.pl --extra-disambig-syms $extra_word_disambig_syms; then
-    echo "Validation of disambiguation file \"$extra_word_disambig_syms\" failed."
+if [ ! -z "$extra_word_disambig_syms" ]; then
+  if ! utils/lang/validate_disambig_sym_file.pl --allow-numeric "false" $extra_word_disambig_syms; then
+    echo "$0: Validation of disambiguation file \"$extra_word_disambig_syms\" failed."
     exit 1;
   fi
 fi
@@ -292,10 +292,9 @@ echo $ndisambig > $tmpdir/lex_ndisambig
 
 # In case there are extra word-level disambiguation symbols they also
 # need to be added to the list of phone-level disambiguation symbols.
-if [[ ! -z $extra_word_disambig_syms ]]; then
+if [ ! -z "$extra_word_disambig_syms" ]; then
   # We expect a file containing valid word-level disambiguation symbols.
-  # The regular expression for awk is just a paranoia filter (e.g. for empty lines).
-  cat $extra_word_disambig_syms | awk '/^#\w+$/ { print $1 }' >> $dir/phones/disambig.txt
+  cat $extra_word_disambig_syms | awk '{ print $1 }' >> $dir/phones/disambig.txt
 fi
 
 # Create phone symbol table.
@@ -359,17 +358,15 @@ cat $tmpdir/lexiconp.txt | awk '{print $1}' | sort | uniq  | awk '
 
 # In case there are extra word-level disambiguation symbols they also
 # need to be added to words.txt
-if [[ ! -z $extra_word_disambig_syms ]]; then
-  # Since words.txt already exist, we need to extract the current word count.
+if [ ! -z "$extra_word_disambig_syms" ]; then
+  # Since words.txt already exists, we need to extract the current word count.
   word_count=`tail -n 1 $dir/words.txt | awk '{ print $2 }'`
 
   # We expect a file containing valid word-level disambiguation symbols.
-  # The regular expression for awk is just a paranoia filter (e.g. for empty lines).
-  cat $extra_word_disambig_syms | awk -v WC=$word_count '
-  /^#\w+$/
-  {
-    printf("%s %d\n", $1, ++WC);
-  }' >> $dir/words.txt || exit 1;
+  # The list of symbols is attached to the current words.txt (including
+  # a numeric identifier for each symbol).
+  cat $extra_word_disambig_syms | \
+    awk -v WC=$word_count '{ printf("%s %d\n", $1, ++WC); }' >> $dir/words.txt || exit 1;
 fi
 
 # format of $dir/words.txt:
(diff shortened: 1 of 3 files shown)


horndasch:
Dan asked for this pull request (see also his comment regarding pull request #1071).

> ... What would be better would be to add to prepare_lang.sh an option --extra-word-disambig-syms disambig_syms.txt where that file contains, one per line, an "extra" word-level disambiguation symbol.

I have added code to do that to prepare_lang.sh

> These symbols would all be required to start with the hash mark # and not be numeric (to avoid overlap with the automatically generated symbols). Also we should disallow #-1 (hash-mark then minus one), because that is used internally in some FST stuff. Perhaps you could try doing this as a separate pull request.

#1113 :o)

> Don't forget that it needs to be suitable documented via comments and the usage message, and you need to check that the provided file matches the requirements.

The check, if the provided file containing the extra word-level disambiguation symbols meet the requirements, is done by the new perl script utils/validate_disambig_sym_file.pl

> Also check whether any changes are required to validate_lang.pl (e.g. any assumptions that the disambig symbols be numeric).

I looked into this and didn't find such an assumption ("that the disambig symbols should be numeric"). All disambiguation symbols are read into a hash and are excluded when the test word sequences are generated.

But this leads me to one more issue regarding the integration of word classes (at least the way I suggest it): Before embedding the word-class sub-language models, I add non-terminal symbols which are later on replaced with the sub-language models (e.g. C=WEEKDAY) using fstreplace. These non-terminal symbols are entries in words.txt, but there are no according entries in the lexicon. As a consequence they can make validate_lang.pl fail, in case they show up in the randomly generated word sequence.

Since it would be messy to somehow automatically deduct the non-terminal symbols (e.g. from the extra word-level disambiguation symbols), I would suggest to add yet another option to prepare_lang.sh (and validate_lang.pl), e.g. something like '--non-terminal-syms non_term_syms.txt'. The content of the file could be passed on to validate_lang.pl so the script does not generate these symbols when creating the test word sequence.

danpovey:
I'll review this soon, and it's probably OK, but I want to respond to your comment about wanting to include the class names in words.txt.
validate_lang.pl is supposed to be called with no arguments, like any of these 'validate' scripts, and it may be called from anywhere.
Including the nonterminal symbols in words.txt is not a good idea as it would cause validation to fail and there is no reason to include it in the words.txt (except for your convenience in creating your scripts).  Because the lang/ directory is the main line and your stuff is an obscure branch, your inconvenience counts for much less than any inconvenience in using the lang/ directory.  So-- don't include the nonterminals in words.txt.

egs/wsj/s5/utils/prepare_lang.sh
```diff
@@ -143,6 +149,15 @@ if [[ ! -z $phone_symbol_table ]]; then
       print "Phone appears in the lexicon but not in the provided phones.txt: "$x; exit 1; }}}' || exit 1;
 fi
 
+# In case there are extra word-level disambiguation symbols we need
+# to make sure that all symbols in the provided file are valid.
+if [[ ! -z $extra_word_disambig_syms ]]; then
+  if ! utils/validate_disambig_sym_file.pl --extra-disambig-syms $extra_word_disambig_syms; then
+    echo "Validation of disambiguation file \"$extra_word_disambig_syms\" failed."
```
danpovey: please start echo statements with `"$0:` so it's clear which program produced the errror.

egs/wsj/s5/utils/prepare_lang.sh
```diff
@@ -275,6 +290,14 @@ echo $ndisambig > $tmpdir/lex_ndisambig
 
 ( for n in `seq 0 $ndisambig`; do echo '#'$n; done ) >$dir/phones/disambig.txt
 
+# In case there are extra word-level disambiguation symbols they also
+# need to be added to the list of phone-level disambiguation symbols.
+if [[ ! -z $extra_word_disambig_syms ]]; then
```
danpovey: just for stylistic consistency, please use single brackets `[ ]` unless double brackets are required by some special construct.

egs/wsj/s5/utils/validate_disambig_sym_file.pl
```diff
@@ -0,0 +1,80 @@
...
+- must start with the hash mark '#',
+- must not contain any whitespace,
+- must not be equal to '#-1' (disallowed because it is used internally in some
+  FST stuff).
+
+In case the option '--extra-disambig-syms' is used with 'true', the symbols must
+also be non-numeric (to avoid overlap with the automatically generated symbols).
+
+Allowed options:
+  --extra-disambig-syms  : Indicate extra symbols  (boolean,  default = "false")
```
danpovey: I think it would be clearer and more consistent with the bash scripts to write this as follows (and change the name):
 ```   --allow-numeric    (true|false)    :   Default true.  If false, disallow numeric disambiguation
                                                                    symbols like #0, #1 and so on.
 ```

egs/wsj/s5/utils/prepare_lang.sh
```diff
@@ -334,6 +357,21 @@ cat $tmpdir/lexiconp.txt | awk '{print $1}' | sort | uniq  | awk '
...
+# In case there are extra word-level disambiguation symbols they also
+# need to be added to words.txt
+if [[ ! -z $extra_word_disambig_syms ]]; then
+  # Since words.txt already exist, we need to extract the current word count.
+  word_count=`tail -n 1 $dir/words.txt | awk '{ print $2 }'`
+
+  # We expect a file containing valid word-level disambiguation symbols.
+  # The regular expression for awk is just a paranoia filter (e.g. for empty lines).
+  cat $extra_word_disambig_syms | awk -v WC=$word_count '
+  /^#\w+$/
```
danpovey: I think it would be better instead of filtering out empty lines, to make sure that the validation script rejects files with empty lines.  Also, please use less vertical space in this awk script, you could do it like
```
 cat $extra_word_disambig_syms | \
    awk  { .. all one one line }  
```
which is more similar to existing scripts.

egs/wsj/s5/utils/validate_disambig_sym_file.pl
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env perl
```
danpovey: we are no longer putting stuff in the top-level utils/ directory-- please locate this in utils/lang.

egs/wsj/s5/utils/prepare_lang.sh
horndasch: The validation script already filtered out empty lines. I removed the (paranoia) regexp filter in the awk command and made the awk command a one-liner.

egs/wsj/s5/utils/validate_disambig_sym_file.pl
horndasch: I moved it ...

egs/wsj/s5/utils/validate_disambig_sym_file.pl
horndasch: Yes, the new option name is more intuitive. The changes are now checked in.

danpovey:
OK, I assume you have run it and it works as expected?

horndasch:
I basically tested
- if prepare_lang.sh stopped given invalid disambig_syms.txt files and
- if the files data/lang/words.txt, data/lang/phones/wdisambig.txt and data/lang/phones/disambig.txt were written correctly.

danpovey:
ok, merging-- thanks!

PR #1117, opened 2016-10-14
speed up the reduce_data_dir_by_reclist.sh

sih4sing5hong5:
```
$ man grep
       -w, --word-regexp
              Select only those lines containing matches  that  form
              whole  words.  The test is that the matching substring
              must either be  at  the  beginning  of  the  line,  or
              preceded   by   a   non-word   constituent  character.
              Similarly, it must be either at the end of the line or
              followed  by  a non-word constituent character.  Word-
              constituent characters are letters,  digits,  and  the
              underscore.
       -F, --fixed-strings
              Interpret  PATTERN  as  a  list  of   fixed   strings,
              separated  by newlines, any of which is to be matched.
              (-F is specified by POSIX.)
```
I had tested it manually.
Any test for these scripts?


danpovey:
I had not realized that this script exists, and it should probably be deleted-- it has not been updated like other scripts that manipulate data directories.  Was there some reason you needed it?
The easiest and most correct way to do this type of filtering is to filter out some of the recordings manually and then use utils/fix_data_dir.sh to remove things that depend on those recordings.
If you submit a PR to delete this script, I would accept that.

sih4sing5hong5:
There are some bad recordings in my corpus.
And I got bad performance, 85% WER.

I didn't want to modify `data/train` directly.
I checked recordings and made a list about them.
Filtering by the list before training the model.

It's OK to delete the script because it can be replaced by 
```
cat $reclist | awk '{print "^"$0}' > $reclist.tmp
cat $srcdir/utt2spk | grep -ewF -f $reclist.tmp > $destdir/utt2spk
rm -f $reclist.tmp
utils/fix_data_dir.sh $destdir
```

danpovey:
Thanks--  removing that script.

danpovey:
Already did this separately -> closing PR.

sih4sing5hong5:
The `egs/wsj/s5/utils/reduce_data_dir.sh` is like `reduce_data_dir_by_reclist.sh`
You can remove `reduce_data_dir.sh`, too.

danpovey:
OK, I will- -thanks.

Issue #1119, opened 2016-10-14
copy-matrix critical error


danijel3:
I have a simple example of a file that causes the matrix reading code in text mode to crash.

This is a file called test.ark:
```
test  [
  1 2 3 4 5
  1 2 3 4
  1 2 3 4 5 ]
```

It is parsed using this command:
`copy-matrix ark,t:test.ark ark:/dev/null`

This causes this error:
`*** Error in '/home/guest/kaldi/src/bin/copy-matrix': munmap_chunk(): invalid pointer: 0x0000000001027220 ***
`
(with backtrace and memory map below)

Ran it through GDB and the error is in the file kaldi-matrix.cc in line 1442

The cause of the error is obviously the uneven number of elements in each row. While most other parse errors are checked in that code, this one doesn't seem to be. The error seems to be caused by the reserve statement in line 1409.

If you ask me, the easiest way to fix this problem is to simply not add faulty lines to the data vector. If the "cur_row" doesn't have the same number of elements as the preceding line, it doesn't belong in the vector anyway...

danpovey:
If there are lines with different numbers of elements, the matrix reading
code should crash via KALDI_ERROR.  Can you please add a check for this and
make a pull request?
Dan

danijel3:
Sure, I'll give it a go, but I could use some hints to check if a similar error could exist in other parts of the code?

danpovey:
I doubt there would be a similar issue anywhere else.

danijel3:
Turned out the error was somewhat different than anticipated. It was caused by a double delete in two cases. It's fixed now...

danpovey:
You mean it's fixed in current Kaldi code, or you'll submit a pull request?

On Sat, Oct 15, 2016 at 12:18 PM, Danijel Koržinek <notifications@github.com

danijel3:
It's fixed in pull request #1121 referenced above.

PR #1121, opened 2016-10-15
Fixing deletion errors in text matrix parsing code

danijel3:
Fixing the errors mentioned in issue #1119


src/matrix/kaldi-matrix.cc
```diff
@@ -1391,12 +1392,13 @@ void Matrix<Real>::Read(std::istream & is, bool binary, bool add) {
...
               specific_error << "Matrix has inconsistent #cols: " << num_cols
                              << " vs." << data[i]->size() << " (processing row"
-                             << i;
+                             << i << ")";
               goto cleanup;
             }
             for (int32 j = 0; j < num_cols; j++)
               (*this)(i, j) = (*(data[i]))[j];
             delete data[i];
+            data[i]=NULL;
```
danijel3: Sure. Is it okay now?

danpovey:
yes but I'll wait for automatic checks to complete and then merge.  Thanks!

src/matrix/kaldi-matrix.cc
```diff
@@ -1437,9 +1439,11 @@ void Matrix<Real>::Read(std::istream & is, bool binary, bool add) {
     // Note, we never leave the while () loop before this
     // line (we return from it.)
  cleanup: // We only reach here in case of error in the while loop above.
-    delete cur_row;
+    if(cur_row != NULL)
```
vdp: Just FYI, the check for NULL is not necessary (http://stackoverflow.com/questions/4190703/is-it-safe-to-delete-a-null-pointer).

danijel3: Okay, you're right. I may be overcompensating a bit, cause I treat Kaldi as something that is supposed to outlive many changes in standards and compiler versions.

PR #1122, opened 2016-10-15
WIP: work in preparation for online computation in nnet3

danpovey:
This pull request is currently mostly a refactoring of some of the nnet3 code, to make it easier to implement online computation of things like LSTMs so that we don't forget the already-computed things between chunks.

Later on, I'll push to this branch the actual online computation.

This should give the same results as previous code (except I _hope_ speed of the nnet3 compilation stage might be very slightly improved).  Certainly it shouldn't be worse. 
However I have only tested it this to the extent of making sure the automated tests run.  I would appreciate it if someone could run some experiments with these changes. 
All the changes are internal.  The online format of 'cached computations' changes, but the scripts should be robust to this change even if you switch over partway through an experiment.


danpovey:
this is covered in the 'looped computation' PR.

Issue #1118, opened 2016-10-14
Document about converting arpa-lm into G.fst


sih4sing5hong5:
The doc of [online_decoding](https://github.com/kaldi-asr/kaldi/blob/35ebbbe11235a5766e009e46b127901dada897b7/src/doc/online_decoding.dox#L356-L365), [data_prep.dox](https://github.com/kaldi-asr/kaldi/blob/85a3dd5f0b71e419abf1169a26b759bfc423a543/src/doc/data_prep.dox#L820-L828) and [format_lm.sh#L42-L44](https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/utils/format_lm.sh#L42-L44) are different.
I was confused about them.
Should we use the same example?

ognjentodic:
Also, "Preparing the grammar G" section in http://kaldi-asr.org/doc/graph_recipe_test.html is different (missing utils/s2eps.pl call).

jtrmal:
Vassil's post is slightly outdated but still very useful for these kinds of
questions
http://vpanayotov.blogspot.com/2012/06/kaldi-decoding-graph-construction.html

danpovey:
Those parts of the documentation that you pointed out are outdated-- the arpa2fst command was rewritten so that it does some of those extra steps.
@kkm000, do you have time to submit a PR to update the documentation?  Or does anyone else on this thread have time?

sih4sing5hong5:
Which example is recommended?
I use the one in `online_encoding` now.
What I can do?

danpovey:
Actually it doesn't matter I think, the tool is back compatible, but I'd
recommend the one in format_lm.sh.

sih4sing5hong5:
I will copy the `format_lm.sh` one to others later.
```
gunzip -c $lm \
  | arpa2fst --disambig-symbol=#0 \
             --read-symbol-table=$out_dir/words.txt - $out_dir/G.fst
```

danpovey:
ok thanks.

sih4sing5hong5:
#1129 merged

PR #1129, opened 2016-10-20
WIP: Update docs about arpa2fst

sih4sing5hong5:
#1118 

The docs below was in `data_prep.dox`.
Where they should be moved into?
```
Here, the variable <DFN>\$test</DFN> would be something like <DFN>data/lang_test_tg</DFN>.
The most important of these commands is <DFN>arpa2fst</DFN> which is a Kaldi program.
This program turns the ARPA-format language model into a Weight Finite State Transducer
(actually, an acceptor).
The <DFN>grep</DFN> commands remove certain "invalid" n-grams from the language model, which
should not have been there in the first place.  The program <DFN>remove_oovs.pl</DFN> removes
N-grams containing words not in our vocabulary (which would cause a crash in <DFN>fstcompile</DFN>); <DFN>eps2disambig.pl</DFN> turns the
<DFN>\<eps\></DFN> (\f$\epsilon\f$) symbols on backoff arcs into a special symbol <DFN>#0</DFN> which
we use to make the grammar determinizable; see \ref graph_disambig.  If you don't know
what a "backoff arc" is, see the literature on backoff N-grams, for instance Goodman's "A bit of progress in
language modeling", and the paper by Mohri we referenced above.  The command <DFN>s2eps.pl</DFN> turns the
sentence-begin and sentence-end symbols <DFN>\<s\></DFN> and <DFN>\</s\></DFN> into
epsilon (<DFN>\<eps\></DFN>), which means "no symbol"; <DFN>fstcompile</DFN> is an OpenFst
command which turns the
FST from text form into OpenFst binary form; and <DFN>fstrmepsilon</DFN> is also
an OpenFst command, which removes the small number of <DFN>\<eps\></DFN> symbols from the
FST, which were converted from <DFN>\<s\></DFN> and <DFN>\</s\></DFN>.
```


danpovey:
I am compiling this version of the documentation and pushing it to the site.  It should be done in about 10 minutes.  Please check that it looks OK, then let me know and I can merge.

sih4sing5hong5:
I checked and added small adjusts.
It is okay to merge when tested pass.

danpovey:
OK thanks, merging. [the tests do not cover doxygen.]

PR #1130, opened 2016-10-20
Fix "last index out of range" exception

aevernon:
This skips blank lines in input files, for example, the first line of
lexicon3.txt.

### DIFF
--- a/egs/swbd/s5c/local/format_acronyms_dict.py
+++ b/egs/swbd/s5c/local/format_acronyms_dict.py
@@ -31,6 +31,10 @@ fin_Letter.close()
 
 for lex in fin_lex:
     items = lex.split()
+
+    if items == []:
+        continue
+
     word = items[0]
     lexicon = lex[len(items[0])+1:].strip()
     # find acronyms from words with only letters and '


danpovey:
I commented by replying, but my reply does not seem to have shown up.  Repeating myself:

For me there are no blank lines in lexicon3.txt.  I think the problem may be earlier and I want to find out why that line was blank, e.g. it may be a locale-dependent issue that we should fix.
Can you please trace back and figure out where it came from?
If you take out the 'sort -u' you can use the context to figure out which line in lexicon2.txt was blank, and then work backwards.

aevernon:
The root cause was that I was using mawk (the version of awk that ships with Debian 8) instead of gawk, which was causing the second awk expression below from swbd1_prepare_dict.sh to not skip blank lines:

```bash
awk 'BEGIN{getline}($0 !~ /^#/) {print}' \
  $dir/lexicon0.txt | sort | awk '($0 !~ /^[[:space:]]*$/) {print}' \
   > $dir/lexicon1.txt
```

I have run into other problems using mawk with Kaldi. How do you feel about updating tools/extras/check_dependencies.sh to explicitly require GNU awk?

danpovey:
I responded again by email but it was lost.  Re-posting.

Our general policy has been to update the scripts to ensure that they do not use gawk-specific features.  In this case, a much easier way to do that would be to do:

grep -v '^#' $dir/lexicon0.txt | awk 'NF>0' | sort > $dir/lexicon1.txt

Is there any chance you could make a PR for this and grep around to see if you can see similar-looking expressions elsewhere that could be updated?

danpovey:
Our general policy has been to update the scripts to ensure that they do
not use gawk-specific features.  In this case, a much easier way to do that
would be to do:

grep -v '^#' $dir/lexicon0.txt | awk 'NF>0' | sort > $dir/lexicon1.txt

Is there any chance you could make a PR for this and grep around to see if
you can see similar-looking expressions elsewhere that could be updated?

danpovey:
For me there are no blank lines in lexicon3.txt.  I think the problem may
be earlier and I want to find out why that line was blank, e.g. it may be a
locale-dependent issue that we should fix.
Can you please trace back and figure out where it came from?
If you take out the 'sort -u' you can use the context to figure out which
line in lexicon2.txt was blank, and then work backwards.

aevernon:
I've opened #1143 to track the scripts that need to be updated.

aevernon:
I noticed that [tools/extras/install_srilm.sh](https://github.com/kaldi-asr/kaldi/blob/master/tools/extras/install_srilm.sh#L25) requires GNU awk.

danpovey:
Yes but that is specific to SRILM and we can't get around it.  We prefer to
have Kaldi scripts themselves not rely on GNU awk peculiarities.

Issue #977, opened 2016-08-17
Improving nnet generalization to feature offsets


danpovey:
I have thought of an efficient way to enhance the generalization of our models to additive feature offsets by adding randomly chosen shifts to the training features.  @vijayaditya, perhaps you can work with someone on this.  Or @pegahgh, if you want to do this and can do it soon, you can do it, as I know you have been interested in similar things in the past.

This is something that we could do when dumping egs for neural net training, including chain training, where in effect we dump egs for [say] 5 different versions of each utterance, differing by random feature shifts... but we won't physically dump all of those egs; instead we store special data with the egs that allows the differently shifted versions to be generated on the fly during training.

I'll try to make it concrete below.

Firstly, we work out a covariance for the feature-space CMN offsets, by taking the centered covariance of the speaker or pseudo-speaker means, and use this times some constant (probably a bit less than one, like 0.5, or maybe 1] to generate random CMN offsets to be applied to the features.  [Or possibly the offsets could have non-random means, to subtract a little of the speaker's existing offset from the global mean, thus keeping the same speaker-level covariance as before the perturbation... we can experiment with this.]

Each speaker we will have a certain number of 'versions' of that speaker, differing in random CMN offsets [e.g. num_cmn_offsets=5].  [Note: this will probably be done after we've done the 'max2' stuff so they won't be 'real' speakers in general, but parts of speakers.]   Note: there is no point in making num_cmn_offsets larger than the number of epochs that we use in neural net training, as we wouldn't have time to cycle through all the offsets otherwise.  The cmn_offsets will be randomly generated at some point and put in an archive indexed by speaker or (duplicated) per utterance, in a matrix with num-rows == num_cmn_offsets and num-cols == feature-dim, e.g. 40.  I have more details below on how we can store this info.

When we dump the egs we will only physically dump one version of the eg.  However, we will abuse the egs format slightly by adding an extra NnetIo member called 'offsets', which contains the cmn_offsets for that utterance.  It will contain the above-mentioned matrix of offsets [dump it with compression]; and for the indexes, I suggest to have n=0, t=0, and x corresponding to each row of the matrix, so x=0, x=1, .. x=4 in the normal case... this is not critical at all, it just means that the Check() function won't complain.

[...]

PR #1131, opened 2016-10-20
Improving nnet generalization to feature offsets

pegahgh:
This PR is the first implementation for issue #977 
@vimalmanohar  I would be thankful if you go through first round of review for this PR.


danpovey:
@pegahgh, do you have any WERs with this?

src/nnet3bin/nnet3-get-egs.cc
```diff
@@ -199,6 +210,16 @@ int main(int argc, char *argv[]) {
           num_err++;
           continue;
         }
+        const Matrix<BaseFloat> *cmn_offsets = NULL;
+        if (!utt2cmn_offsets_rspecifier.empty()) {
+          if (offset_reader.HasKey(key)) {
```
vimalmanohar: !offset_reader

egs/wsj/s5/steps/nnet3/get_egs.sh
```diff
@@ -325,10 +325,15 @@ if [ $stage -le 4 ]; then
...
+  # if $data/offsets.scp is defined, it will be used in generating egs.
+  offsets=
+  if [ -f $data/offsets.scp ]; then
+    offsets="--utt2spk-offsets=scp:$sdata/JOB/offsets.scp"
+  fi
   echo "$0: Generating training examples on disk"
   # The examples will go round-robin to egs_list.
   $cmd JOB=1:$nj $dir/log/get_egs.JOB.log \
-    nnet3-get-egs --num-pdfs=$num_pdfs $ivector_opt $egs_opts --num-frames=$frames_per_eg "$feats" \
+    nnet3-get-egs --num-pdfs=$num_pdfs $ivector_opt $egs_opts --num-frames=$frames_per_eg "$offsets" "$feats" \
```
vimalmanohar: I think this should be $offsets without the quotes

egs/wsj/s5/steps/compute_offsets.sh
```diff
@@ -0,0 +1,53 @@
...
+
+for f in $required_files; do
+  if [ ! -f $f ]; then
+    echo "$0: no such file $f"
+    exit 1;
+  fi
+done
+
+$cmd $logdir/compute_rand_offset.log \ 
+  generate-random-cmn-offsets --config=$offsets_config ark:$data/spk2utt scp:$data/feats.scp ark:- | copy-feats --compress=$compress ark:- ark,scp:$offsetdir/offsets.ark,$data/offsets.scp || exit 1;
```
vimalmanohar: Try to stick to 80 chars per line.

vimalmanohar: You don't seem to have committed this binary.

src/nnet3/nnet-example.h
```diff
@@ -44,14 +44,17 @@ struct NnetIo {
...
-
+  
   /// This constructor creates NnetIo with name "name", indexes with n=0, x=0,
   /// and t values ranging from t_begin to t_begin + feats.NumRows() - 1, and
   /// the provided features.  t_begin should be the frame that feats.Row(0)
   /// represents.
   NnetIo(const std::string &name,
          int32 t_begin, const MatrixBase<BaseFloat> &feats);
 
+  NnetIo(const std::string &name,
```
vimalmanohar: Explain why this is needed and where it is used.

pegahgh:
Not yet! I will update you as soon as I get the results.

danpovey:
@pegahgh, any results on this?

pegahgh:
The preliminary results using cmn-offset-scale=0.1 results in 0.2% WER
degradation on SWBD.
Although I used small offset scale, the random offsets are larger than
MFCC. I will try to run new experiment using smaller offset scale.
e.g. MFCC

74.33336 -14.56764  -35.18937 -6.718647

corresponding offsets for these dimensions.

567.3954 -114.6091 -132.4369 -56.41733

The centered covariance of the speaker or pseudo-speaker means is

(1/num_spks) * sum_{i=0, num_spks} (m_i - g)^t (m_i - g)

Large speaker mean, m_i  from global mean g  results in larger covariance,
even if this speaker has few number of frames and it doesn't affect global
CMN.

It may be good idea to use global CMN to generate random offset.

The following figure is the i-vector mapped to 2dim space using PCA for
different offsets of single utterance.

src/featbin/generate-random-cmn-offsets.cc
```diff
@@ -0,0 +1,217 @@
...
+    double global_count = global_stats(0, feat_dim);
+    global_mean.Scale(1.0 / global_count);
+
+    // add term -2 * (sum_{i=0}^n m_i^T) g
+    spk_cov.AddVecVec(-2.0, spk_mean_sum, global_mean);
+    
+    // add term gT * g
+    spk_cov.AddVecVec(1.0, global_mean, global_mean);
+    // compute sigma as spk_cov^0.5 for matrix w.r.t its svd decomposition.
+    Matrix<double> sigma(feat_dim, feat_dim);
```
danpovey: This matrix seems to be initialized as zero.

src/featbin/generate-random-cmn-offsets.cc
```diff
@@ -0,0 +1,217 @@
...
+    sigma.Power(0.5);
```
danpovey: It's possible that adding an offset generated from the un-scaled sigma will turn out to be 'too strong', and it might be necessary to scale down sigma by some constant  (e.g. 0.5).

src/featbin/generate-random-cmn-offsets.cc
```diff
@@ -0,0 +1,217 @@
...
+      spk_mean_sum.AddVec(1.0, spk_mean);
+      global_stats.AddMat(1.0, stats);
+    }
+    // global_mean (g) is the global mean for all frames.
+    Vector<double> global_mean(global_stats.Row(0).Range(0, feat_dim));
+    double global_count = global_stats(0, feat_dim);
+    global_mean.Scale(1.0 / global_count);
+
+    // add term -2 * (sum_{i=0}^n m_i^T) g
+    spk_cov.AddVecVec(-2.0, spk_mean_sum, global_mean);
```
danpovey: At this point you should be dividing by the number of speakers.

pegahgh:
I did all of that.
Actually I didn't commit latest changes. I will commit them now.

pegahgh:
I will use zero-mean while generating random offsets and offset = C^0.5 * x
where x = N(0,1) and scaling sigma by some constant is equivalent to
scaling offsets using cmn-offset-scale.

src/featbin/generate-random-cmn-offsets.cc
```diff
@@ -0,0 +1,222 @@
...
+      spk_mean_sum.AddVec(1.0, spk_mean);
+      global_stats.AddMat(1.0, stats);
+    }
+    // global_mean (g) is the global mean for all frames.
+    Vector<double> global_mean(global_stats.Row(0).Range(0, feat_dim));
+
+    double global_count = global_stats(0, feat_dim);
+    global_mean.Scale(1.0 / global_count);
+
+    Vector<double> global_mean2(spk_mean_sum);
```
danpovey: not a good choice of variable name.

src/featbin/generate-random-cmn-offsets.cc
```diff
@@ -0,0 +1,222 @@
...
+    double global_count = global_stats(0, feat_dim);
+    global_mean.Scale(1.0 / global_count);
+
+    Vector<double> global_mean2(spk_mean_sum); 
+    global_mean2.Scale(1.0 / num_spks);
+    // add term -2 * (sum_{i=0}^n m_i^T) g
+    spk_cov.AddVecVec(-2.0, spk_mean_sum, global_mean2);
+    
+    // add term gT * g
+    spk_cov.AddVecVec(1.0, global_mean2, global_mean2);
```
danpovey: this 1.0 should be num_spks. 
This may be the cause of the high variance you observed.

danpovey: ... also, I think it would be best if you add some 'self-testing' code to this program.  E.g. take the covariance of the speakers after adding the CMN offsets, multiply it on each side by the inverse-stddev of the original speaker covariance, print out the eigenvalues, and make sure they are close to 1 (if it's been done in such a way that the speaker covariance is supposed to be the same).

pegahgh:
Thanks Dan.
I fixed the bugs and added self-testing code to this program.
The figure shows eigenvalues for C^-1 * offset_C, which seems to the
eigenvalues are close to 1. Each offsets for each speaker counted as
separate speaker in computing offset speaker covariance.

egs/wsj/s5/steps/nnet3/train_dnn.py
```diff
@@ -298,13 +299,17 @@ def TrainNewModels(dir, iter, srand, num_jobs, num_archives_processed, num_archi
                                                # the other indexes from.
         archive_index = (k % num_archives) + 1 # work out the 1-based archive index.
         frame = (k / num_archives) % frames_per_eg
+        offset_num = -1
+        if num_cmn_offsets > 0:
+          offset_num = (archive_index - 1) % num_cmn_offsets
```
pegahgh: Dan, I fixed this issue in my local repo as follows, but I didn't commit the change.
epoch_index = num_archives_processed / (num_archives * frames_per_eg)
offset_num = epoch_index % num_cmn_offsets

e.g. num_archives_to_process = num_epochs * frames_per_eg * num_archives =>
epoch_index = num_archives_processed / ( frames_per_eg * num_archives)

Isn't that what we expect?
Also I think we need to process the data and its perturbed versions randomly.
e.g. in ubm offset model, while adding offset in 2nd epoch, I got large log-likelihood degradation in the beginning of 2nd epoch.
Any idea?

danpovey:
That formula should be better was was checked in, but my formula might be
even better as it mixes together different offsets at the same time.  In
your case you train on non-offsetted data for the 1st epoch, and then
suddenly start training on the offsetted data.
But if the log-like degradation is very large (e.g. more than 0.1 or so),
it may indicate that too much variance is being added to the features, or
maybe it indicates some other problem.

pegahgh:
Dan, I tested generalized shift on TDNN+xent and frames_per_offset is 8,
where we have 3 shifts of data in Chain model. I can start testing that on
chain model.
Also I think we should use k instead on num_archives_processed in the
formula to be compatible with frame number, am I right?

The range for num_archives_processed / num_archives is [0, frames_per_eg *
num_epochs] and and based on proposed formula, the offset for fixed
archive_index = f, and fixed shift of data 's' in epoch epoch_index,
(epoch_index * frames_per_eg + s) % num_offsets = epoch_index *
(frames_per_eg % num_offsets + (s % num_offsets). To have different offset
for same shift in different epochs, frames_per_eg % num_offsets != 0, (if
num_offsets and frames_per_eg would be coprime is one of the cases)!!

For TDNN+xent, frames_per_eg=8, num_offsets=3 and num_epochs=3 should be
appropriate!

Do you agree?

danpovey:
You seem to be pointing out that my formula is problematic if the
num_offsets is not also coprime to frames_per_eg.  That seems valid.  (So
for regular xent TDNN training it would have to coprime to frames_per_eg,
and for chain training it would have to be coprime to num_shifts(=3)).  Or
use a different formula.  I did not understand the specific details of what
you were saying.  But anyway, try to come up with a formula that would
satisfy all the requirements, or ensure that the num-offsets is coprime to
whatever it needs to be coprime to.  This idea may not even work.

danpovey:
@pegahgh, is this PR a duplicate and should it be deleted?

pegahgh:
Yes, the new PR is #1323.

PR #1141, opened 2016-10-26
printed out stats for self-repair in nonlinearity (self-repaired-prop…

freewym:
…ortion)

Has been tested for back-compatibility.


danpovey:
Thanks, will merge when the checks finish.

PR #1142, opened 2016-10-26
more robust parsing of the alignment

jtrmal:
tested, didn't crash


jtrmal:
I tested it on one system (via steps/score_kaldi.sh) and it produced  the same stats exactly.
y.

Issue #1143, opened 2016-10-26
Don't use gawk-specific features in example scripts


aevernon:
As discussed in #1130, the awk expression that removes whitespace-only lines does not work with mawk (the version of awk that ships with Debian 8):

```bash
awk 'BEGIN{getline}($0 !~ /^#/) {print}' \
  $dir/lexicon0.txt | sort | awk '($0 !~ /^[[:space:]]*$/) {print}' \
   > $dir/lexicon1.txt || exit 1;
```

@danpovey suggested the following instead:

```bash
grep -v '^#' $dir/lexicon0.txt | awk 'NF>0' | sort > $dir/lexicon1.txt || exit 1;
```

The files below need to be updated:

- [x] egs/fisher_english/s5/local/fisher_prepare_dict.sh
- [x] egs/sre10/v1/local/dnn/fisher_prepare_dict.sh
- [x] egs/fisher_swbd/s5/local/fisher_swbd_prepare_dict.sh
- [x] egs/fisher_swbd/s5/local/fisher_prepare_dict.sh
- [x] egs/fisher_swbd/s5/local/swbd1_prepare_dict.sh
- [x] egs/swbd/s5/local/swbd1_prepare_dict.sh
- [x] egs/swbd/s5/local/swbd_p1_prepare_dict.sh
- [x] egs/swbd/s5b/local/swbd1_prepare_dict.sh
- [x] egs/swbd/s5c/local/swbd1_prepare_dict.sh
- [x] egs/aspire/s5/local/fisher_prepare_dict.sh

aevernon:
Fixed by 84c4738 and 001e56e.

danpovey:
Thanks.  I've asked someone to do this.

aevernon:
It looks like there are still 12- to 24-hour delays when commenting on issues by email because your comment just showed up. I've notified GitHub.

danpovey:
Addressed in #1147

PR #1147, opened 2016-10-27
Fix the gawk-specific feature problems in prepare dict scripts.

jyang126jhu:
Fix the gawk-specific feature problems in prepare dict scripts.
Described here: https://github.com/kaldi-asr/kaldi/issues/1143

### DIFF
--- a/egs/fisher_swbd/s5/local/fisher_swbd_prepare_dict.sh
+++ b/egs/fisher_swbd/s5/local/fisher_swbd_prepare_dict.sh
@@ -116,7 +116,7 @@ cp $srcdict $dir/lexicon0.txt || exit 1;
 patch <local/dict.patch $dir/lexicon0.txt || exit 1;
 
 #(2a) Dictionary preparation:
-# Pre-processing (Upper-case, remove comments)
+# Pre-processing (remove comments)
 grep -v '^#' $dir/lexicon0.txt | awk 'NF>0' | sort > $dir/lexicon1_swbd.txt || exit 1;
 
 cat $dir/lexicon1_swbd.txt | awk '{ for(n=2;n<=NF;n++){ phones[$n] = 1; }} END{for (p in phones) print p;}' | \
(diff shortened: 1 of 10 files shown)


egs/fisher_swbd/s5/local/fisher_swbd_prepare_dict.sh
```diff
@@ -117,10 +117,7 @@ patch <local/dict.patch $dir/lexicon0.txt || exit 1;
 
 #(2a) Dictionary preparation:
 # Pre-processing (Upper-case, remove comments)
```
danpovey: Plese remove the 'Upper-case' comment here since it's not being done.

egs/fisher_swbd/s5/local/swbd1_prepare_dict.sh
```diff
@@ -20,10 +20,7 @@ srcdict=$srcdir/swb_ms98_transcriptions/sw-ms98-dict.text
 
 #(2a) Dictionary preparation:
 # Pre-processing (Upper-case, remove comments)
-awk 'BEGIN{getline}($0 !~ /^#/) {$0=tolower($0); print}' \
-  $srcdict | sort | awk '($0 !~ /^[:space:]*$/) {print}' \
-   > $dir/lexicon1.txt || exit 1;
-
+grep -v '^#' $srcdict | tr '[a-z]' '[A-Z]' | awk 'NF>0' | sort > $dir/lexicon1.txt || exit 1;
```
danpovey: The comment said 'Upper-case' but the code was actually lower-casing.  So please lower-case instead, and change the comment.  See if there are similar issues elsewhere.

danpovey:
OK, but please change the comments so that they correspond to what is actually being done w.r.t. casing.  All comments say 'upper-case', but some do lower, some do nothing.

jyang126jhu:
Done!

danpovey:
Thanks!  Merging.

PR #1154, opened 2016-10-29
Fix get_ctm_edits.py to be not too greedy with id's

psmit:
It is not too pretty this way (and possibly wasteful), but the whole function would need reorganization to do this in a nicer way. (See also email thread "Problems with cleanup, can't align in get_ctm_edits.py")


danpovey:
Probably OK, but let me wait till we find out why the nbest-to-ctm thing was failing.

psmit:
@danpovey Updated to include the lattice-1best fix

danpovey:
Thanks!

Issue #1128, opened 2016-10-18
Spatial regularization component


danpovey:
Some of you will have seen the Microsoft ArXiv paper, in which they report better-than-human performance on Switchboard.
https://arxiv.org/pdf/1610.05256v1.pdf
The new aspect of their work which is easiest to directly use in our system is their idea of "spatial regularization".  
This is something that happens on the output of hidden layers (e.g. affine layers or BLSTM layers).
The output of the layers is viewed [arbitrarily] as a two-dimensional grid, for instance a BLSTM layer with an output dimension of 512 could be viewed as a 16 \* 32 grid.
A penalty term is added to the objective function, that penalizes a particular activation being different from the average of the adjacent activations.  The way this is implemented in practice is that the output is convolved with a 3 x 3 kernel where the center pixel is 1 and all the other pixels have a value of -1/8, and some constant times the sum-squared of the result is added to the objective function [and obviously we backprop from there.]  The convolution is "circular", meaning that at the edges of the image we wrap around to the opposite edge of the image.

I propose to implement this as SpatialRegularizationComponent-- a component that will just copy its input in the forward pass (like ClipGradientComponent), but in the backward pass will add something the derivative.  We won't be able to add the objf term to the overall objective function, as this is now how the framework works, but this doesn't matter, as what's important is the derivative-- stats can be stored in the component from which we can get the objf via the Info() string if we really want to.

We'll need GPU and CPU versions of the computation.  Obviously the GPU one is the more important for speed, given that that's how we always train.  I don't want to assume that the dimensions of the activations are easy to factor.  For this and other reasons, I would like to implement this circular convolution on a fixed dimension, of size 256, and I'd like the thread-blocks to have dimension 16 \* 16 = 256.  The circular convolution will operate separately on 256-dimensional blocks of the layer activations, wrapping around the edges of these blocks (and it should behave as if padded with zeros for uneven-size blocks, although the code shouldn't assume that the memory is physically padded with zeros).  Remember not to use the macros CU1DBLOCK and CU2DBLOCK in the GPU code, as this would affect the behavior of the code-- just hardcode it as 16 \* 16 blocks.  In the kernel, the code will be quite trivial- no logarithmic reduction- but will need __syncthreads().

[...]

PR #863, opened 2016-06-25
Speed up SoftmaxComponent::Backprop()

kangshiyin:
Speed up `SoftmaxComponent::Backprop()` by merging multiple CUDA kernels into one.

```
                                     dim  old     new    speedup
CuMatrix::DiffSoftmaxPerRow<float>,   16  0.0036  0.0166 4.66x
CuMatrix::DiffSoftmaxPerRow<float>,   32  0.0146  0.0679 4.66x
CuMatrix::DiffSoftmaxPerRow<float>,   64  0.0583  0.2474 4.24x
CuMatrix::DiffSoftmaxPerRow<float>,  128  0.2251  0.8984 3.99x
CuMatrix::DiffSoftmaxPerRow<float>,  256  0.8341  2.8901 3.46x
CuMatrix::DiffSoftmaxPerRow<float>,  512  1.9272  6.7216 3.49x
CuMatrix::DiffSoftmaxPerRow<float>, 1024  2.7828 10.4916 3.77x
CuMatrix::DiffSoftmaxPerRow<double>,   16  0.0026  0.0149 5.70x
CuMatrix::DiffSoftmaxPerRow<double>,   32  0.0121  0.0587 4.85x
CuMatrix::DiffSoftmaxPerRow<double>,   64  0.0528  0.2289 4.34x
CuMatrix::DiffSoftmaxPerRow<double>,  128  0.1757  0.7635 4.34x
CuMatrix::DiffSoftmaxPerRow<double>,  256  0.5835  2.4046 4.12x
CuMatrix::DiffSoftmaxPerRow<double>,  512  1.4246  4.5517 3.19x
CuMatrix::DiffSoftmaxPerRow<double>, 1024  1.9497  4.3642 2.24x
```

PR #1162, opened 2016-11-01
Add Spatial Regularization Deriv

kangshiyin:
WIP: Add Spatial Regularization DerivThis PR is working on the backprop of SpatialRegularizationComponent as described in #1128 .

Two things that I'm not sure about.

1. The paper mentioned that a 512-D layer output would be viewed as a 16x32 grid. What if the layer width is not 512, say 256 or 700? Should a layer with arbitrary width be partitioned into several 512-D sections?

2. What is the purpose for logarithmic reduction? It seems to me that after the 2-D convolution with the 3x3 kernel, the filtered `out_value` can be directly added to `out_deriv` and finish the whole thing. Do I miss something?

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -2630,6 +2630,312 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
 }
 
 
+/// This CUDA kernel implements the idea 'spatial smoothing' as described in
+/// the paper https://arxiv.org/pdf/1610.05256v1.pdf
+///
+/// It adds spatial regularization derivative to `out_deriv`.
+/// The regularization term is calculated from `out_value`.
+/// Each row of `out_value` is first viewed as a n*16 2D image,
+/// high-pass filtered by a specific 3x3 Filter kernel with 1 at the center
+/// and -1/8 elsewhere for twice, scaled with `-scale` and then added to
+/// `out_deriv`.
+/// out_deriv = out_deriv - scale * Filter (X) (Filter (X) out_value),
+/// where (x) denotes convolution operation,
+/// and Filter is the 3x3 filter kernel.
+/// If regularization_sumsq is not NULL, the partial sum of the squared
+/// regularization term will be calculated and stored to the vector
+/// `regularization_sumsq` with the arrangement of one element per thread block.
+/// The sum can be computed later with CuVectorBase::Sum().
+/// `out_deriv` has the same dimension as `out_value`, but possibly a different
+/// stride.
+///
+/// This CUDA kernel uses 2D grid. Each row of the grid handles one row of
+/// data matrices including `out_value` and `out_deriv`.
+/// Each thread block contains 256 threads.
+/// For the first convolution, the thread block handles a data area of 16x16.
+/// For the second convolution, the thread block handles a data area of 14x16.
+/// Thus we need (NumCols()+255-16*2)/(256-16*2) blocks per grid row to cover
+/// the matrix row dim.
+/// The thread block also has 2D, with a row width of 16, matching the width
+/// of the image mentioned above.
+/// The min NumCols() can be 17, so that n>=2 for the n*16 image.
+///
+/// In the circular convolution stage, each thread block handles a 14x16 area
+/// out of the whole n*16 image. So we need to cache 18x16 pixels when
+/// the size of the filter kernel is 3x3, and filter is applied twice.
+/// The extra pixels are 2x16 heading apron, and 2x16 trailing apron,
+/// as suggested in https://www.evl.uic.edu/sjames/cs525/final.html
+/// When the row dim of `out_value` is not a multiple of 16, 0s will be padded.
+///
+/// Here is an example on how a 40-D hidden activation is cached.
+/// It demonstrates everything except the multi-thread-block configuration.
+/// The cache is of the size (2+3+2)x16 :
+/// Leading apron:  17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32
+///                 33 34 35 36 37 38 39 40  0  0  0  0  0  0  0  0
+/// active data:     1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16
+///                 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32
+///                 33 34 35 36 37 38 39 40  0  0  0  0  0  0  0  0
+/// Trailing apron:  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16
+///                 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32
+template<typename Real>
+__global__
+static void _add_spatial_regularization_deriv(const Real* out_value,
+                                              const MatrixDim out_value_dim,
+                                              Real* out_deriv,
+                                              const int out_deriv_stride,
+                                              const Real scale,
+                                              Real* regularization_sumsq) {
+  const int kBlockSize = 256;
+  const int kSectWidth = 16;
+  const int kActiveSize = kBlockSize - 2 * kSectWidth;
+  __shared__ union {
+    Real buf[kBlockSize / kSectWidth + 2][kSectWidth];
+    Real sum[kBlockSize];
+  } smem;
+
+  const int i = blockIdx.y;
+  const int tid = threadIdx.y * blockDim.x + threadIdx.x;
+  const int j = blockIdx.x * kActiveSize + tid;
+  int active_rows = kActiveSize / kSectWidth;
+
+  // Cache global mem data `out_value` to shared mem
+  if (gridDim.x == 1) {
+    // out_value_dim.cols <= kActiveSize
+    Real val = Real(0);
+    if (j < out_value_dim.cols) {
+      val = out_value[i * out_value_dim.stride + j];
+    }
+    const int image_rows = (out_value_dim.cols + (kSectWidth - 1)) / kSectWidth;
+    active_rows = image_rows;
+    if (threadIdx.y < image_rows) {
+      if (threadIdx.y >= image_rows - 2) {
+        // Cache leading apron
+        smem.buf[threadIdx.y - (image_rows - 2)][threadIdx.x] = val;
+      }
+      // Cache active data
+      smem.buf[threadIdx.y + 2][threadIdx.x] = val;
+    }
+    if (threadIdx.y < 2) {
+      // Cache trailing apron
+      smem.buf[threadIdx.y + 2 + image_rows][threadIdx.x] = val;
+    }
+  } else if (gridDim.x == 2) {
+    // out_value_dim.cols <= 2*kActiveSize
+    if (blockIdx.x == 0) {
+      // First block.
+      // Cache leading apron
+      if (threadIdx.y < 2) {
+        const int image_rows = (out_value_dim.cols + (kSectWidth - 1))
+            / kSectWidth;
+        const int active_tid = out_value_dim.cols
+            - (image_rows - 2) * kSectWidth;
+        Real val = Real(0);
+        if (tid < active_tid) {
+          val = out_value[i * out_value_dim.stride
+              + (image_rows - 2) * kSectWidth + tid];
+        }
+        smem.buf[threadIdx.y][threadIdx.x] = val;
+      }
+      // Cache active data
+      if (tid < kActiveSize) {
+        smem.buf[threadIdx.y + 2][threadIdx.x] = out_value[i
+            * out_value_dim.stride + j];
+      }
+      // Cache trailing apron
+      const int last_block_rows = ((out_value_dim.cols - 1) % kActiveSize)
+          / kSectWidth + 1;
+      if (last_block_rows == 1) {
+        // first row of the trailing apron
+        Real val = Real(0);
+        if (j + kActiveSize < out_value_dim.cols) {
+          val = out_value[i * out_value_dim.stride + kActiveSize + j];
+        }
+        if (threadIdx.y < 1) {
+          smem.buf[kBlockSize / kSectWidth][threadIdx.x] = val;
+        }
+        // second row of the trailing apron
+        if (threadIdx.y < 1) {
+          smem.buf[kBlockSize / kSectWidth + 1][threadIdx.x] = out_value[i
+              * out_value_dim.stride + tid];
+        }
+      } else {
+        if (threadIdx.y < 2) {
+          Real val = Real(0);
+          if (j + kActiveSize < out_value_dim.cols) {
+            val = out_value[i * out_value_dim.stride + kActiveSize + j];
+          }
+          smem.buf[kBlockSize / kSectWidth + threadIdx.y][threadIdx.x] = val;
+        }
+      }
+    } else {
+      // Last block.
+      // Cache leading apron and active data
+      const int last_block_rows = ((out_value_dim.cols - 1) % kActiveSize)
+          / kSectWidth + 1;
+      active_rows = last_block_rows;
+      Real val = Real(0);
+      if (j < out_value_dim.cols + 2 * kSectWidth) {
+        val = out_value[i * out_value_dim.stride + j - 2 * kSectWidth];
+      }
+      if (threadIdx.y < last_block_rows + 2) {
+        smem.buf[threadIdx.y][threadIdx.x] = val;
+      }
+      // Cache trailing apron
+      if (threadIdx.y < 2) {
+        smem.buf[last_block_rows + 2 + threadIdx.y][threadIdx.x] = out_value[i
+            * out_value_dim.stride + tid];
+      }
+    }
+  } else {
+    // out_value_dim.cols > 2*kActiveSize
+    if (blockIdx.x == 0) {
+      // First block.
+      // Cache leading apron
+      if (threadIdx.y < 2) {
+        const int image_rows = (out_value_dim.cols + (kSectWidth - 1))
+            / kSectWidth;
+        const int active_tid = out_value_dim.cols
+            - (image_rows - 2) * kSectWidth;
+        Real val = Real(0);
+        if (tid < active_tid) {
+          val = out_value[i * out_value_dim.stride
+              + (image_rows - 2) * kSectWidth + tid];
+        }
+        smem.buf[threadIdx.y][threadIdx.x] = val;
+      }
+      // Cache active data and trailing apron
+      smem.buf[threadIdx.y + 2][threadIdx.x] = out_value[i
+          * out_value_dim.stride + j];
+    } else if (blockIdx.x == gridDim.x - 1) {
+      // Last block.
+      // Cache leading apron and active data
+      const int last_block_rows = ((out_value_dim.cols - 1) % kActiveSize)
+          / kSectWidth + 1;
+      active_rows = last_block_rows;
+      Real val = Real(0);
+      if (j < out_value_dim.cols + 2 * kSectWidth) {
+        val = out_value[i * out_value_dim.stride + j - 2 * kSectWidth];
+      }
+      if (threadIdx.y < last_block_rows + 2) {
+        smem.buf[threadIdx.y][threadIdx.x] = val;
+      }
+      // Cache trailing apron
+      if (threadIdx.y < 2) {
+        smem.buf[last_block_rows + 2 + threadIdx.y][threadIdx.x] = out_value[i
+            * out_value_dim.stride + tid];
+      }
+    } else if (blockIdx.x == gridDim.x - 2) {
+      // Second to last block.
+      // Cache leading apron and active data
+      smem.buf[threadIdx.y][threadIdx.x] = out_value[i * out_value_dim.stride
+          + j - 2 * kSectWidth];
+      // Cache trailing apron
+      const int last_block_rows = ((out_value_dim.cols - 1) % kActiveSize)
+          / kSectWidth + 1;
+      if (last_block_rows == 1) {
+        // first row of the trailing apron
+        Real val = Real(0);
+        if (j + kActiveSize < out_value_dim.cols) {
+          val = out_value[i * out_value_dim.stride + kActiveSize + j];
+        }
+        if (threadIdx.y < 1) {
+          smem.buf[kBlockSize / kSectWidth][threadIdx.x] = val;
+        }
+        // second row of the trailing apron
+        if (threadIdx.y < 1) {
+          smem.buf[kBlockSize / kSectWidth + 1][threadIdx.x] = out_value[i
+              * out_value_dim.stride + tid];
+        }
+      } else {
+        if (threadIdx.y < 2) {
+          Real val = Real(0);
+          if (j + kActiveSize < out_value_dim.cols) {
+            val = out_value[i * out_value_dim.stride + kActiveSize + j];
+          }
+          smem.buf[kBlockSize / kSectWidth + threadIdx.y][threadIdx.x] = val;
+        }
+      }
+    } else {
+      // Middle blocks.
+      // Cache leading apron
+      if (threadIdx.y < 2) {
+        smem.buf[threadIdx.y][threadIdx.x] = out_value[i * out_value_dim.stride
+            + j - 2 * kSectWidth];
+      }
+      // Cache active data and trailing apron
+      smem.buf[threadIdx.y + 2][threadIdx.x] = out_value[i
+          * out_value_dim.stride + j];
+    }
+  }
+  __syncthreads();
+
+  // Fist circular convolution on 16x16 area
+  const int x = threadIdx.x;
+  const int x_left = (threadIdx.x + (kSectWidth - 1)) % kSectWidth;
+  const int x_right = (threadIdx.x + 1) % kSectWidth;
+  const int y = threadIdx.y + 1;
+  const int y_up = y - 1;
+  const int y_down = y + 1;
+  Real conv1 = Real(0);
+  if (threadIdx.y < active_rows + 2) {
+    conv1 = smem.buf[y][x]
+        - Real(0.125)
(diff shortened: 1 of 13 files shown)


jtrmal:
My feeling is that you just wanna factor the dimension into a 2-D grid possibly as square as possible. I.e. for 700 
```
factor(700) -> [2 2 5 5 7]

[2*2*7 5*5 ] = 28 x 25 
```

jtrmal:
Ah, scratch that, the factorization does not make sense... Not sure where my thoughts were.

kangshiyin:
Why? I thought it was reasonable.

Maybe we don't need partitioning. We could always put all the outputs into
one single large enough grid. The size of the grid can be 16*n. So at most
15 zeros need to padded at the last column of the grid.

danpovey:
I'd be OK with 16 * n.
The reason I suggested blocks of 256 is that I thought there might be a
limitation on the size of thread blocks (you'd need to do __synchthreads
internally and it would be most convenient to have one thread per
element).  But up to you.
The logarithmic reduction I refer to is to compute the sum-sq value, the
penalization term, for diagnostic purposes.  But any way that works (and is
fast) is OK with me.

I don't want to do factoring as what if the number is prime or has very
unevenly sized factors... seems a hassle.

jtrmal:
Yes, makes sense actually. Somehow I managed to convince myself  that 28*25 is not 700.
But I think you are right -- any sufficiently sqquare matrix might work just as well

danpovey:
BTW, you might only have to do __synchtreads internally if it supports
in-place backprop.  Not sure of the details.  Supporting in-place backprop
would be nice because it would save memory.
Personally I'd prefer to do it on fixed-size blocks (e.g. of dim 256) and
support in-place backprop, than not be able to do in-place backprop.

Dan

kangshiyin:
In-place backprop should be Ok regardless the size of the grid.

Sq-sum on the filtered regularization probably takes shorter time than the
conv-scale-add operation if all of them are implemented in one CUDA kernel.

I will update the doc in the code and start the implementation.

2016年11月2日 4:03 AM，"Daniel Povey" <notifications@github.com>写道：

> BTW, you might only have to do __synchtreads internally if it supports
> in-place backprop. Not sure of the details. Supporting in-place backprop
> would be nice because it would save memory.
> Personally I'd prefer to do it on fixed-size blocks (e.g. of dim 256) and
> support in-place backprop, than not be able to do in-place backprop.
>
> Dan
>
>

danpovey:
>
> In-place backprop should be Ok regardless the size of the grid.
>
> Sq-sum on the filtered regularization probably takes shorter time than the
> conv-scale-add operation if all of them are implemented in one CUDA kernel.
>

But I'd rather not have them in the same kernel if possible because it's
wasteful to compute the sq-sum on every single minibatch (we can do it only
on a subset if it's in a separate kernel).  But if it's really very fast,
it's OK to have it in the same kernel.


>
> I will update the doc in the code and start the implementation.


OK.

Please create a solution that will work for arbitrary hidden-dim sizes, I
don't want something that will fail due to max-threads-per-block
limitations when the hidden-dim gets large.



>
>
> 2016年11月2日 4:03 AM，"Daniel Povey" <notifications@github.com>写道：
>
> > BTW, you might only have to do __synchtreads internally if it supports
> > in-place backprop. Not sure of the details. Supporting in-place backprop
> > would be nice because it would save memory.
> > Personally I'd prefer to do it on fixed-size blocks (e.g. of dim 256) and
> > support in-place backprop, than not be able to do in-place backprop.
> >
> > Dan
> >
> >

src/cudamatrix/cu-kernels.cu
```diff
@@ -2629,19 +2629,109 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
...
-/// Add spatial regularization term to `out_deriv`.
+/// This CUDA kernel implements the idea 'spatial smoothing' as described in
+/// the paper https://arxiv.org/pdf/1610.05256v1.pdf
+///
+/// It adds spatial regularization term to `out_deriv`.
 /// The regularization term is calculated from `out_value`.
-/// Each row of `out_value` is first viewed as a 16*n grid,
-/// convoluted with the 3x3 kernel, scaled with `scale`
+/// Each row of `out_value` is first viewed as a n*16 or n*32 2D image,
+/// high-pass filtered by a 3x3 kernel, scaled with `scale`
```
danpovey: a 3x3 kernel -> a specific 3x3 kernel with 1 at the center and -1/8 elsewhere.

danpovey: also, what do you mean by n*16 or n*32?  Your code seems to refer to 16.  What does this depend on?  And how about end effects?  I don't see from your code where the 32 comes in.
What is the dimension of regularization_sqsum?
Mention that out_deriv has the same dimension as out_value, but possibly a different stride.

src/cudamatrix/cu-kernels.cu
```diff
@@ -2629,19 +2629,109 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
...
+  } smem;
+
+  const int i = blockIdx.y;
+  const int tid = threadIdx.y * blockDim.x + threadIdx.x;
+  const int j = blockIdx.x * kBlockSize + tid;
+
+
+  // Cache global mem data `out_value` to shared mem
+  if (threadIdx.y == 0) {
+    // Leading apron
```
danpovey: 'apron' seems to be some obscure NVidia jargon.. please explain in simpler terms what is going on here.

src/cudamatrix/cu-kernels.cu
```diff
@@ -2629,19 +2629,109 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
...
+  if (j < out_value_dim.cols) {
+    val = out_value[i * out_value_dim.stride + j];
+  }
+  smem.buf[threadIdx.y + 1][threadIdx.x] = val;
+  __syncthreads();
+
+  // circular convolution and scaled adding to `out_deriv`
+  if (j < out_value_dim.cols) {
+    val -= 0.125
+        * (sbuf[(tid - kSectWidth - 1) & (kBlockSize - 1)]
```
danpovey: I assume this code is outdated and should be smem.buf?

kangshiyin:
I will reply the above comments later. Here is another question.
It seems the shape of the 2D image is not important, any combination of 1 central pixel and 8 surrounding pixels can fit the high-pass filter.
So can we do this in 1-D like
out[i]=in[i]-0.125*(in[i+1]+in[i+2]+...+in[i+8])
Will this be topologically same as the 2D filter?

src/cudamatrix/cu-kernels.cu
kangshiyin: Can not find a better word. I use example to explain it.

danpovey:
"So can we do this in 1-D like
out[i]=in[i]-0.125*(in[i+1]+in[i+2]+...+in[i+8])
Will this be topologically same as the 2D filter?"
No it won't be the same... later we can investigate other topologies like 1d and 3d, but for now let's stick with the 2d filter.

src/cudamatrix/cu-matrix.h
```diff
@@ -334,6 +334,17 @@ class CuMatrixBase {
   void DiffLogSoftmaxPerRow(const CuMatrixBase<Real> &out_value,
                             const CuMatrixBase<Real> &out_deriv);
 
+  /// Add spatial regularization term to *this (`out_deriv`).
+  /// The regularization term is calculated from `out_value`.
+  /// Each row of `out_value` is first viewed as a 16*n grid,
+  /// convoluted with the 3x3 kernel, scaled with `scale`
+  /// and then added to `out_deriv`.
```
danpovey: Also, please rename all 'sqsum' to 'sumsq'.

danpovey: I just noticed that for the derivative of the sqsum, you need to apply the high-pass filtering twice (and then multiply by -1.0 `*` scale).

kangshiyin:
You are right. It has to be twice. The implementation seems to be much more
complex. It could either be
1) loading 20x16 data and convolution with 5x5 kernel on 16x16 image, or
2) loading 20x16 data, convolution with 3x3 kernel on 18x16 image and then
with 3x3 kernel on 16x16 image.
Not sure which one is better.

danpovey:
@kangshiyin, thanks for your latest commits.  It looks like nnet-component-test failed.  Did it fail when you ran it locally?

kangshiyin:
I'm still debugging it... I think I can finish it this weekend.

danpovey:
@kangshiyin, I hope you'll have time to make progress on this soon!  I'm confident that this method is going to improve all of our systems, so it will be very exciting to get it working.  I also have some ideas that are dependent on getting this working.

src/cudamatrix/cu-kernels.cu
```diff
@@ -2630,6 +2630,204 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
...
+/// This CUDA kernel implements the idea 'spatial smoothing' as described in
+/// the paper https://arxiv.org/pdf/1610.05256v1.pdf
+///
+/// It adds spatial regularization derivative to `out_deriv`.
+/// The regularization term is calculated from `out_value`.
+/// Each row of `out_value` is first viewed as a n*16 2D image,
+/// high-pass filtered by a specific 3x3 Filter kernel with 1 at the center
+/// and -1/8 elsewhere for twice, scaled with `-scale` and then added to
+/// `out_deriv`.
+/// out_deriv = out_deriv - scale * Filter (X) (Filter (X) out_value),
```
kangshiyin: @danpovey Sorry for the delay. Filtering twice makes things more complicated. I just finish both the GPU and CPU implementation draft. I'm working on the testing code. Could you help check the backprop formula at this line. I'm currently not sure if it correct. Thanks!

src/cudamatrix/cu-kernels.cu
```diff
@@ -2630,6 +2630,204 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
...
+    Real sum[kBlockSize];
+  } smem;
+
+  const int i = blockIdx.y;
+  const int tid = threadIdx.y * blockDim.x + threadIdx.x;
+  const int j = blockIdx.x * kActiveSize + tid;
+
+  // Cache global mem data `out_value` to shared mem
+  // out_value_dim.cols <= kActiveSize
+  if (blockDim.x == 1) {
```
danpovey: I think you mean if gridDim.x == 1

danpovey:
Also, here is an idea for testing-- write a function that would transform the image in a reversible way that shouldn't affect the derivatives.  I'm thinking of column permutations that would correspond to a circular shift in the image.  E.g. for shift parameters x > 0 and y > 0, make a vector of integers i, where i[k] = (k%16 + x) % 16  +  (16 * (k/16 + y)) % num_cols
and call CopyCols() taking a CuArray with those integers [copying to a new matrix], to permute the columns.

[this only applies for matrices with num-cols that are multiples of 16].

Then you'd have to apply the inverse transform to the derivatives and compare them with the derivatives computed on the un-perturbed features.  You could use x_inv = (-x + 16)  % 16  and y_inv = (-y + (num_cols/16)) % (num_cols/16) as the "inverse" shifts.  [They would just be -x and -y if C++ had a the same definition of the '%' operator as python.]  When doing that comparison, use AssertEqual with the default tolerance, which is very small-- it should be exact.

kangshiyin:
Currently the code has passed `cu-matrix-test`, indicating the CPU and GPU version of AddSpatialRegularizationDeriv give identical results. But neither of them passes `nnet-component-test`.
I think the engineering problem in cuda kernel function is now solved, but we still have a mathematical problem on the derivative formulation of the spatial regularization as follows.
```
/// out_deriv = out_deriv - scale * Filter (X) (Filter (X) out_value),
/// where (X) denotes convolution operation,
/// and Filter is the 3x3 filter kernel.
```
I'm not sure if the formulation is mathematically correct. Could you give some help? Thanks!

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
 
 
 template<typename Real>
+void MatrixBase<Real>::AddSpatialRegularizationDeriv(
+    const MatrixBase<Real>& out_value, Real scale, Real* regularization_sumsq) {
+  const unsigned int kBlockSize = 256;
```
danpovey: please move this function to the 'else' in the cu-matrix function, or to a static non-member function defined in cu-matrix.cc; I don't like to bulk up the kaldi-matrix library with these neural-net-specific things.  At some point these may get moved out of the cu-matrix class.

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
...
+  int ext_cols = ((this->NumCols() + (kSectWidth - 1)) / kSectWidth)
+      * kSectWidth;
+  Matrix<Real> val_ext(this->NumRows(), ext_cols);
+  Matrix<Real> conv1(this->NumRows(), ext_cols);
+  Matrix<Real> conv2(this->NumRows(), ext_cols);
+
+  // add 0 padding to the end of the rows
+  for (int32 i = 0; i < this->NumRows(); i++) {
+    for (int32 j = 0; j < this->NumCols(); j++) {
+      val_ext(i, j) = out_value(i, j);
```
danpovey: you can do this copying with a submatrix, e.g.
val_ext.RowRange(0, out_value.NumCols()).CopyFromMat(out_value);

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
...
+  const int image_rows = ext_cols / kSectWidth;
+  for (int32 i = 0; i < this->NumRows(); i++) {
+    for (int32 j = 0; j < ext_cols; j++) {
+      int image_x = j % kSectWidth;
+      int image_x_left = (image_x + (kSectWidth - 1)) % kSectWidth;
+      int image_x_right = (image_x + 1) % kSectWidth;
+      int image_y = j / kSectWidth;
+      int image_y_up = (image_y + (image_rows - 1)) % image_rows;
+      int image_y_down = (image_y + 1) % image_rows;
+      conv1(i, j) = val_ext(i, j)
```
danpovey: there are a lot of calls to operator () here, which is expensive, so better to get a pointer to the row and index that for the ones in 'val_ext'.

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
...
+                  + val_ext(i, image_y_up * kSectWidth + image_x_right)
+                  + val_ext(i, image_y * kSectWidth + image_x_left)
+                  + val_ext(i, image_y * kSectWidth + image_x_right)
+                  + val_ext(i, image_y_down * kSectWidth + image_x_left)
+                  + val_ext(i, image_y_down * kSectWidth + image_x)
+                  + val_ext(i, image_y_down * kSectWidth + image_x_right));
+    }
+  }
+  if (regularization_sumsq) {
+    Real sum_squared = Real(0);
```
danpovey: you can do just `*regularization_sumsq = TraceMatMat(conv1, conv1, kTrans);`

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
...
+    for (int32 i = 0; i < this->NumRows(); i++) {
+      for (int32 j = 0; j < this->NumCols(); j++) {
+        sum_squared += conv1(i, j) * conv1(i, j);
+      }
+    }
+    *regularization_sumsq = sum_squared;
+  }
+  for (int32 i = 0; i < this->NumRows(); i++) {
```
danpovey: best to use a variable for this->NumRows() and this->NumCols() when used as a loop test... avoids function call.

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
...
+    const MatrixBase<Real>& out_value, Real scale, Real* regularization_sumsq) {
+  const unsigned int kBlockSize = 256;
+  const unsigned int kSectWidth = 16;
+  KALDI_ASSERT(SameDim(*this, out_value));
+
+  int ext_cols = ((this->NumCols() + (kSectWidth - 1)) / kSectWidth)
+      * kSectWidth;
+  Matrix<Real> val_ext(this->NumRows(), ext_cols);
+  Matrix<Real> conv1(this->NumRows(), ext_cols);
+  Matrix<Real> conv2(this->NumRows(), ext_cols);
```
danpovey: conv2 is not used.

src/matrix/kaldi-matrix.cc
```diff
@@ -2784,6 +2784,78 @@ void MatrixBase<Real>::DiffTanh(const MatrixBase<Real> &value,
...
+  }
+  for (int32 i = 0; i < this->NumRows(); i++) {
+    for (int32 j = 0; j < this->NumCols(); j++) {
+      int image_x = j % kSectWidth;
+      int image_x_left = (image_x + (kSectWidth - 1)) % kSectWidth;
+      int image_x_right = (image_x + 1) % kSectWidth;
+      int image_y = j / kSectWidth;
+      int image_y_up = (image_y + (image_rows - 1)) % image_rows;
+      int image_y_down = (image_y + 1) % image_rows;
+      (*this)(i, j) -= scale * conv1(i, j)
```
danpovey: Here, 'scale' should scale the whole expression, not just the 1st term.

src/nnet3/nnet-component-test.cc
```diff
@@ -483,6 +483,116 @@ bool TestSimpleComponentModelDerivative(const Component &c,
...
+        int32 jm1 = (j + num_rows - 1) % num_rows,  // the + num_rows is to
+                                                    // ensure it's positive so %
+                                                    // is really modulus.
+            jp1 = (j + 1) % num_rows,
+            km1 = (k + num_cols - 1) % num_cols,
+            kp1 = (k + 1) % num_cols;
+        highpass_image(j, k) = image(j, k) - 0.125 * (
+            image(jm1, km1) + image(jm1, k) + image(jm1, kp1) +
+            image(j, km1) + image(j, kp1) +
+            image(jp1, km1) + image(jp1, k) + image(jm1, kp1));
```
danpovey: [bug in my code here], this jm1 should be jp1.

src/nnet3/nnet-component-test.cc
```diff
@@ -483,6 +483,116 @@ bool TestSimpleComponentModelDerivative(const Component &c,
   }
 }
 
+// this is a simple reference implementation of the 'regularization_sumsq'
+// computed by CuMatrix::AddSpatialRegularizationDeriv.
+BaseFloat FilteredSpatialSumsq(const CuMatrix<BaseFloat> &output_data) {
+  Matrix<BaseFloat> data(output_data);
+  int32 num_rows = (data.NumRows() + 15) / 16,
```
danpovey: [another bug in my code], this data.NumRows() should be data.NumCols().

src/nnet3/nnet-simple-component.cc
```diff
@@ -633,6 +633,135 @@ void NoOpComponent::Backprop(const std::string &debug_info,
...
+    KALDI_ERR << "Invalid initializer for layer of type "
+              << Type() << ": \"" << cfl->WholeLine() << "\"";
+  Init(dim, regularization_scale);
+}
+
+
+void SpatialRegularizationComponent::Init(int32 dim, BaseFloat regularization_scale) {
+  KALDI_ASSERT(dim > 0 && regularization_scale >= 0.0);
+  dim_ = dim;
+  regularization_scale_ = regularization_scale_;
```
danpovey: [also from my code],
this should read
regularization_scale_ = regularization_scale;

kangshiyin:
All GPU tests have passed.

Speed test shows that without computing the regularization value, the speed is similar to `SoftmaxComponent::Backprop()` in #863 . If the regularization value is computed, the speed will reduce to 37~60% of the original speed.

```
Benchmark (gflops)                               dim w/o sum w/i sum speed up
CuMatrix::AddSpatialRegularizationDeriv<float>,   16  0.0189  0.0071 0.37x
CuMatrix::AddSpatialRegularizationDeriv<float>,   32  0.0791  0.0281 0.36x
CuMatrix::AddSpatialRegularizationDeriv<float>,   64  0.3064  0.1104 0.36x
CuMatrix::AddSpatialRegularizationDeriv<float>,  128  1.0813  0.4043 0.37x
CuMatrix::AddSpatialRegularizationDeriv<float>,  256  2.4897  1.1313 0.45x
CuMatrix::AddSpatialRegularizationDeriv<float>,  512  4.7202  2.5182 0.53x
CuMatrix::AddSpatialRegularizationDeriv<float>, 1024  6.7612  4.0563 0.60x
CuMatrix::AddSpatialRegularizationDeriv<double>,   16  0.0192  0.0069 0.36x
CuMatrix::AddSpatialRegularizationDeriv<double>,   32  0.0756  0.0272 0.36x
CuMatrix::AddSpatialRegularizationDeriv<double>,   64  0.2951  0.1076 0.36x
CuMatrix::AddSpatialRegularizationDeriv<double>,  128  1.0167  0.3865 0.38x
CuMatrix::AddSpatialRegularizationDeriv<double>,  256  2.3078  1.0710 0.46x
CuMatrix::AddSpatialRegularizationDeriv<double>,  512  4.0806  2.2244 0.55x
CuMatrix::AddSpatialRegularizationDeriv<double>, 1024  5.5427  3.4308 0.62x
```

kangshiyin:
It's strange that I can pass all GPU and CPU test locally, but the on-line test keeps failing even after I re-pushed the last commit...

danpovey:
Probably just random differences.  It can be a good time to run the test
with a larger number of repetitions the 1st time you check in the code, to
see that the thresholds are suitable in general.

After making the following changes it works for me even with changing 3 ->
100 repetitions:

```
-  BaseFloat perturb_delta = 1.0e-04;
+  BaseFloat perturb_delta = 5.0e-04;

-  int32 test_dim = 3;
+  int32 test_dim = 5;

-  BaseFloat threshold = 0.05;
+  BaseFloat threshold = 0.1;

```

danpovey:
.. but it failed with 1000 repetitions.  Change the test_dim from 5 -> 7
and

 num_rows = RandInt(1, 100); to  num_rows = RandInt(1, 50);

src/nnet3/nnet-simple-component.cc
```diff
@@ -633,6 +633,135 @@ void NoOpComponent::Backprop(const std::string &debug_info,
...
+    const ComponentPrecomputedIndexes *indexes,
+    const CuMatrixBase<BaseFloat> &, //in_value
+    const CuMatrixBase<BaseFloat> &out_value,
+    const CuMatrixBase<BaseFloat> &out_deriv,
+    Component *to_update_in,
+    CuMatrixBase<BaseFloat> *in_deriv) const {
+  if (in_deriv != &out_deriv)
+    in_deriv->CopyFromMat(out_deriv);
+  // accumulate stats only every other time.
+  bool compute_sumsq = (to_update_in != NULL && RandInt(0, 1) == 0);
```
danpovey: You can change this to RandInt(0, 3) == 0, since computing the sumsq is a little slow.  [this is only for diagnostics.]

src/cudamatrix/cu-matrix.h
```diff
@@ -334,6 +334,18 @@ class CuMatrixBase {
...
 
+  /// Add spatial regularization term to *this (`out_deriv`).
+  /// The regularization derivative is calculated from `out_value`.
+  /// Each row of `out_value` is first viewed as a 16*n grid,
+  /// convoluted with the 3x3 kernel twice, scaled with `-scale`
+  /// and then multiplied by `out_deriv`.
+  /// *this = *this - scale * Filter (X) (Filter (X) out_value)
+  /// If regularization_sumsq is not NULL, the sum of the squared
+  /// regularization term will be calculated and stored.
+  void AddSpatialRegularizationDeriv(const CuMatrixBase<Real>& out_value,
```
danpovey: Before we commit this, can you please change this to make it a non-member function and move it to cu-math.{h,cc}?  You can change the 'this' to be a pointer argument after 'scale', and call it 'deriv'.
I think in the past we have allowed too many 'special-purpose' functions to clutter the interface of class CuMatrix.  [e.g. the STL's 'string' class did this, and that is widely considered to have been a bad idea.]

kangshiyin:
It should be ok now. All gpu tests have passed.

danpovey:
Great!  Hopefully I'll merge to-morrow.  Soon we'll do some tests with
integrating it with scripts.

danpovey:
On second thoughts, I'd rather merge this once we've done the testing in actual networks.  @vijayaditya is going to talk with @GaofengCheng and explain how to do this in the xconfig framework.  Hopefully within the next week we'll get some results.

GaofengCheng:
@danpovey @vijayaditya I have read MS's paper, as their results show, spatial is benificial , I will try spatial component right now (TDNN+LSTM and BLSTM)

vijayaditya:
I would recommend starting with the most stable setup ie., the TDNN.

-Vijay

danpovey:
@GaofengCheng, when you work on this for xent systems, you might want to try a regularization-scale of 0.2.  There is a factor of 0.5 in my version of objective function (to make the derivative easier), and in https://arxiv.org/pdf/1610.05256v1.pdf their scaling factor was 0.1.  However, the dynamic range of the chain objective function is smaller, so for chain training you might want to try smaller values, like 0.1 or 0.05.

danpovey:
This didn't give us any improvements, so closing the PR.

PR #1169, opened 2016-11-03
Specify full path to 2000 HUB5 transcripts

aevernon:
The top-level directory inside LDC2002T43.tgz is 2000_hub5_eng_eval_tr, so this needs to be passed to eval2000_data_prep.sh. Otherwise the script will give the error, "Expecting directory /export/corpora2/LDC/LDC2002T43/reference to be present".


danpovey:
This may depend what was done when it was unpacked.  On our grid this command works.
I don't want to merge this because it would break our setup.  I'd be OK if you changed the inner script to add /2000_hub5_eng_eval_tr to the directory if that exists.


Dan

aevernon:
Closing in favor of #1172.

PR #1171, opened 2016-11-04
Cleanup segmentation perl fixes

alumae:
Fixed two issues in the re-segmentation scripts caused by inconsistencies with different Perl versions. The first one makes sure that `data/train_split/orig2utt` (or similar) file will be sorted. Otherwise, the `steps/cleanup/decode_segmentation.sh` script might fail later, complaining that the utterances are not sorted.

The other commit changes the line that gets perl list size with another syntax. The old line used experimental syntax that only worked in some Perl versions.


jtrmal:
Thanks, merging.
y.

PR #1172, opened 2016-11-04
Try harder to find 2000 HUB5 transcripts

aevernon:
Depending on how LDC2002T43.tgz is unpacked, the transcripts might be under `2000_hub5_eng_eval_tr`. As requested in #1169, we append it to `tdir` if that directory exists.


danpovey:
Thanks!  Merging.

PR #1174, opened 2016-11-04
adding CER scoring capability

jtrmal:
Working on one of the Mandarin chinese recipes.
Need CER scoring for these purposes.


jtrmal:
Tested on HKUST

egs/wsj/s5/steps/score_kaldi.sh
```diff
@@ -137,9 +137,89 @@ if [ $stage -le 1 ]; then
   fi
 fi
 
+if $cer && [ $stage -le 2 ] ; then
```
danpovey: Since this is a special case, I don't like having such a big block.
How about calling steps/scoring/score_kaldi_cer.sh, and moving steps/score_kaldi.sh to steps/scoring/ and making it a link?  [trying to move to subdirs in steps/]

danpovey: i mean move this block of code into a newly named script, but call it from this script.

jtrmal: My feeling was that would need a lot of careful checking to keep things between score_kaldi.sh and score_kaldi_cer.sh, as it relies on a couple of the variables being set and the files being created and so on. I think I'll close it (for now).

jtrmal:
@danpovey  how about this.
I tried to make each script working on its own, yet support combination of those two in an auxiliary local/score.sh, that would just forward the parameters.

Do not merge yet --even if you ACK this, as I want to test it, there might be typos.

danpovey:
Looks reasonable

jtrmal:
Tested on HKUST and it works, including the to proposed solution for combined CER and WER scoring (via the parameter forwarding).
y.

jtrmal:
I just remembered you want me to move the scripts somewhere -- let me know where, as I've already forgot :-/

danpovey:
steps/scoring/.

jtrmal:
OK, moved, symlinked, rebased&squashed

danpovey:
Thanks!  Merging.

PR #1182, opened 2016-11-11
Fixed OpenFST building on Windows/cygwin64

binghe:
In tools/Makefile, when building OpenFST, the following code was used to have a special configure command on Windows/Cygwin:

`ifeq ($(OSTYPE),cygwin)`

But I found it doesn't work. $(OSTYPE) is null. However $(OS) is "Windows_NT", which is useful here.

For CXXFLAGS, in additional to "-O", I added "-Wa,-mbig-obj" for safety purposes. Now OpenFST builds correctly on Windows/Cygwin by just typing "make" in "tools" directory.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -78,7 +78,11 @@ openfst-$(OPENFST_VERSION)/lib: | openfst-$(OPENFST_VERSION)/Makefile
 # Add the -O flag to CXXFLAGS on cygwin as it can fix the compilation error
 # "file too big".
 openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION)/.patched | check_required_programs
-ifeq ($(OS),Windows_NT)
+# Note: OSTYPE path is probably dead for latest cygwin64 (installed on 2016/11/11).
+ifeq ($(OSTYPE),cygwin)
+	cd openfst-$(OPENFST_VERSION)/; ./configure --prefix=`pwd` --enable-static --enable-shared --enable-far --enable-ngram-fsts CXX=$(CXX) CXXFLAGS="$(CXXFLAGS) -O -Wa,-mbig-obj" LDFLAGS="$(LDFLAGS)" LIBS="-ldl"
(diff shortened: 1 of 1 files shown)


tools/Makefile
```diff
@@ -78,8 +78,8 @@ openfst-$(OPENFST_VERSION)/lib: | openfst-$(OPENFST_VERSION)/Makefile
 # Add the -O flag to CXXFLAGS on cygwin as it can fix the compilation error
 # "file too big".
 openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION)/.patched | check_required_programs
-ifeq ($(OSTYPE),cygwin)
-	cd openfst-$(OPENFST_VERSION)/; ./configure --prefix=`pwd` --enable-static --enable-shared --enable-far --enable-ngram-fsts CXX=$(CXX) CXXFLAGS="$(CXXFLAGS) -O" LDFLAGS="$(LDFLAGS)" LIBS="-ldl"
+ifeq ($(OS),Windows_NT)
+	cd openfst-$(OPENFST_VERSION)/; ./configure --prefix=`pwd` --enable-static --enable-shared --enable-far --enable-ngram-fsts CXX=$(CXX) CXXFLAGS="$(CXXFLAGS) -O -Wa,-mbig-obj" LDFLAGS="$(LDFLAGS)" LIBS="-ldl"
```
jtrmal: I'm a little nervous that you remove the old code path. How about adding it as "else if"
y.

binghe: Well ... I think adding "else if" is acceptable (and more safe). I'm using latest Cygwin64, but maybe in some old versions of Cygwin OSTYPE is "Cygwin" but OS is not set.

jtrmal:
Exactly. Please add some comment that that  OSTYPE path is probably dead
for cygwin version whatever you have and windows whatever you have.
THanks!
y.

binghe:
I have updated the branch with old conditional branch kept for safety purposes. Confirmed on Windows 10 / cygwin64.

jtrmal:
Thanks! merging.

PR #1186, opened 2016-11-14
WIP/help needed: Looped computation

danpovey:
This pull request represents some work I have been doing to support decoding in nnet3 in a way that naturally allows re-use of previously computed things.  This is done right inside the nnet3 compilation process, so we don't have to mess around externally with additional outputs and inputs.

It's applicable to decoding with TDNNs and LSTMs (and should be particularly helpful for LSTMs, because it avoids the need to extend the chunk with left context).
I haven't yet implemented the "online" decoding with this.. that can be done later.
The point this is at, is that all the C++ code is implemented, and tested based on automated tests, but I haven't modified any scripts or run any of the examples.
This is not just about testing new code paths, but existing code could in theory crash, because I made a lot of changes to core nnet3 stuff.
Because I have made the 'looped' computation available by adding a new binary (nnet3-latgen-faster-looped), nnet3 decoding-script changes would be required to make use of this: probably an option --looped-computation [true|false] that would trigger a branch in the script.  Note: the options --extra-left-context and --extra-right-context are not present in nnet3-latgen-faster-looped.  There is a config --extra-left-context-initial (extra left context for the first chunk), but I hope it won't be necessary to use or won't make much difference.

I am hoping someone can help me test this and do the script-level changes.  @tomkocse?  @vijayaditya, do you have any suggestion about this?

### DIFF
--- a/src/chainbin/nnet3-chain-acc-lda-stats.cc
+++ b/src/chainbin/nnet3-chain-acc-lda-stats.cc
@@ -54,7 +54,7 @@ class NnetChainLdaStatsAccumulator {
     NnetComputer computer(options, computation, nnet_, NULL);
 
     computer.AcceptInputs(nnet_, eg.inputs);
-    computer.Forward();
+    computer.Run();
     const CuMatrixBase<BaseFloat> &nnet_output = computer.GetOutput("output");
     AccStatsFromOutput(eg, nnet_output);
   }
@@ -202,5 +202,3 @@ int main(int argc, char *argv[]) {
     return -1;
   }
 }
-
-
--- a/src/matrix/compressed-matrix.h
+++ b/src/matrix/compressed-matrix.h
@@ -47,7 +47,7 @@ class CompressedMatrix {
   CompressedMatrix(): data_(NULL) { }
 
   ~CompressedMatrix() { Clear(); }
-  
+
   template<typename Real>
   CompressedMatrix(const MatrixBase<Real> &mat): data_(NULL) { CopyFromMat(mat); }
 
@@ -73,7 +73,7 @@ class CompressedMatrix {
 
   template<typename Real>
   CompressedMatrix &operator = (const MatrixBase<Real> &mat); // assignment operator.
-  
+
   /// Copies contents to matrix.  Note: mat must have the correct size.
   /// kNoTrans case uses a temporary.
   template<typename Real>
@@ -81,7 +81,7 @@ class CompressedMatrix {
                  MatrixTransposeType trans = kNoTrans) const;
 
   void Write(std::ostream &os, bool binary) const;
-  
+
   void Read(std::istream &is, bool binary);
 
   /// Returns number of rows (or zero for emtpy matrix).
@@ -113,7 +113,7 @@ class CompressedMatrix {
   void Swap(CompressedMatrix *other) { std::swap(data_, other->data_); }
 
   void Clear();
-  
+
   friend class Matrix<float>;
   friend class Matrix<double>;
  private:
@@ -163,7 +163,7 @@ class CompressedMatrix {
   static inline float CharToFloat(float p0, float p25,
                                   float p75, float p100,
                                   unsigned char value);
-  
+
   void *data_; // first GlobalHeader, then PerColHeader (repeated), then
   // the byte data for each column (repeated).  Note: don't intersperse
   // the byte data with the PerColHeaders, because of alignment issues.
--- a/src/nnet3/Makefile
+++ b/src/nnet3/Makefile
@@ -28,7 +28,8 @@ OBJFILES = nnet-common.o nnet-compile.o nnet-component-itf.o \
   discriminative-supervision.o nnet-discriminative-example.o \
   nnet-discriminative-diagnostics.o \
   discriminative-training.o nnet-discriminative-training.o \
-  online-nnet3-decodable-simple.o nnet-compile-looped.o
+  online-nnet3-decodable-simple.o nnet-compile-looped.o \
+  decodable-simple-looped.o
 
 
 LIBNAME = kaldi-nnet3
--- a/src/nnet3/Makefile
+++ b/src/nnet3/Makefile
@@ -28,7 +28,7 @@ OBJFILES = nnet-common.o nnet-compile.o nnet-component-itf.o \
   discriminative-supervision.o nnet-discriminative-example.o \
   nnet-discriminative-diagnostics.o \
   discriminative-training.o nnet-discriminative-training.o \
-  online-nnet3-decodable-simple.o nnet-compile-online.o
+  online-nnet3-decodable-simple.o nnet-compile-looped.o
 
 
 LIBNAME = kaldi-nnet3
--- a/src/nnet3/Makefile
+++ b/src/nnet3/Makefile
@@ -28,7 +28,7 @@ OBJFILES = nnet-common.o nnet-compile.o nnet-component-itf.o \
   discriminative-supervision.o nnet-discriminative-example.o \
   nnet-discriminative-diagnostics.o \
   discriminative-training.o nnet-discriminative-training.o \
-  online-nnet3-decodable-simple.o
+  online-nnet3-decodable-simple.o nnet-compile-online.o
 
 
 LIBNAME = kaldi-nnet3
@@ -37,6 +37,6 @@ ADDLIBS = ../chain/kaldi-chain.a ../cudamatrix/kaldi-cudamatrix.a \
           ../lat/kaldi-lat.a ../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a \
           ../transform/kaldi-transform.a ../gmm/kaldi-gmm.a \
           ../tree/kaldi-tree.a ../util/kaldi-util.a ../thread/kaldi-thread.a \
-          ../matrix/kaldi-matrix.a ../base/kaldi-base.a 
+          ../matrix/kaldi-matrix.a ../base/kaldi-base.a
 
 include ../makefiles/default_rules.mk
--- a/src/nnet3/am-nnet-simple.h
+++ b/src/nnet3/am-nnet-simple.h
@@ -94,7 +94,7 @@ class AmNnetSimple {
   /// This function works out the left_context_ and right_context_ variables
   /// from the network (it's a rather complex calculation).  You should call
   /// this if you have structurally changed the nnet without calling SetNnet(),
-  /// e.g. using non-const GetNnet().  void SetContext();
+  /// e.g. using non-const GetNnet().
   void SetContext();
  private:
 
--- a/src/nnet3/decodable-simple-looped.cc
+++ b/src/nnet3/decodable-simple-looped.cc
@@ -1,0 +1,251 @@
+// nnet3/decodable-simple-looped.cc
+
+// Copyright      2016  Johns Hopkins University (author: Daniel Povey)
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "nnet3/decodable-simple-looped.h"
+#include "nnet3/nnet-utils.h"
+#include "nnet3/nnet-compile-looped.h"
+
+namespace kaldi {
+namespace nnet3 {
+
+
+DecodableNnetSimpleLoopedInfo::DecodableNnetSimpleLoopedInfo(
+    const NnetSimpleLoopedComputationOptions &opts,
+    Nnet *nnet):
+    opts_(opts), nnet_(*nnet) {
+  Init(opts, nnet);
+}
+
+DecodableNnetSimpleLoopedInfo::DecodableNnetSimpleLoopedInfo(
+    const NnetSimpleLoopedComputationOptions &opts,
+    const Vector<BaseFloat> &priors,
+    Nnet *nnet):
+    opts_(opts), nnet_(*nnet), log_priors_(priors) {
+  if (log_priors_.Dim() != 0)
+    log_priors_.ApplyLog();
+  Init(opts, nnet);
+}
+
+
+DecodableNnetSimpleLoopedInfo::DecodableNnetSimpleLoopedInfo(
+    const NnetSimpleLoopedComputationOptions &opts,
+    AmNnetSimple *am_nnet):
+    opts_(opts), nnet_(am_nnet->GetNnet()), log_priors_(am_nnet->Priors()) {
+  if (log_priors_.Dim() != 0)
+    log_priors_.ApplyLog();
+  Init(opts, &(am_nnet->GetNnet()));
+}
+
+
+void DecodableNnetSimpleLoopedInfo::Init(
+    const NnetSimpleLoopedComputationOptions &opts,
+    Nnet *nnet) {
+  opts.Check();
+  KALDI_ASSERT(IsSimpleNnet(*nnet));
+  has_ivectors_ = (nnet->InputDim("ivector") > 0);
+  int32 left_context, right_context;
+  ComputeSimpleNnetContext(*nnet, &left_context, &right_context);
+  frames_left_context_ = left_context + opts.extra_left_context_initial;
+  frames_right_context_ = right_context;
+  frames_per_chunk_ = GetChunkSize(*nnet, opts_.frame_subsampling_factor,
+                                   opts.frames_per_chunk);
+  output_dim_ = nnet->OutputDim("output");
+  KALDI_ASSERT(output_dim_ > 0);
+  // note, ivector_period is hardcoded to the same as frames_per_chunk_.
+  int32 ivector_period = frames_per_chunk_;
+  if (has_ivectors_)
+    ModifyNnetIvectorPeriod(ivector_period, nnet);
+
+  ComputationRequest request1, request2, request3;
+  int32 num_sequences = 1;  // we're processing one utterance at a time.
+  int32 extra_right_context = 0;
+  CreateLoopedComputationRequestSimple(*nnet, frames_per_chunk_,
+                                       opts_.frame_subsampling_factor,
+                                       ivector_period, opts.extra_left_context_initial,
+                                       extra_right_context,
+                                       num_sequences,
+                                       &request1, &request2, &request3);
+
+  CompileLooped(*nnet, opts_.optimize_config, request1, request2, request3,
+                &computation_);
+  computation_.ComputeCudaIndexes();
+  KALDI_LOG << "Computation is:";
+  computation_.Print(std::cerr, *nnet);
+}
+
+
+DecodableNnetSimpleLooped::DecodableNnetSimpleLooped(
+    const DecodableNnetSimpleLoopedInfo &info,
+    const MatrixBase<BaseFloat> &feats,
+    const VectorBase<BaseFloat> *ivector,
+    const MatrixBase<BaseFloat> *online_ivectors,
+    int32 online_ivector_period):
+    info_(info),
+    computer_(info_.opts_.compute_config, info_.computation_,
+              info_.nnet_, NULL),
+    feats_(feats),
+    ivector_(ivector), online_ivector_feats_(online_ivectors),
+    online_ivector_period_(online_ivector_period),
+    num_chunks_computed_(0),
+    current_log_post_subsampled_offset_(-1) {
+  num_subsampled_frames_ =
+      (feats_.NumRows() + info_.opts_.frame_subsampling_factor - 1) /
+      info_.opts_.frame_subsampling_factor;
+  KALDI_ASSERT(!(ivector != NULL && online_ivectors != NULL));
+  KALDI_ASSERT(!(online_ivectors != NULL && online_ivector_period <= 0 &&
+                 "You need to set the --online-ivector-period option!"));
+}
+
+
+void DecodableNnetSimpleLooped::GetOutputForFrame(
+    int32 subsampled_frame, VectorBase<BaseFloat> *output) {
+    KALDI_ASSERT(subsampled_frame >= current_log_post_subsampled_offset_ &&
+                 "Frames must be accessed in order.");
+    while (subsampled_frame >= current_log_post_subsampled_offset_ +
+                            current_log_post_.NumRows())
+      AdvanceChunk();
+    output->CopyFromVec(current_log_post_.Row(
+        subsampled_frame - current_log_post_subsampled_offset_));
+}
+
+int32 DecodableNnetSimpleLooped::GetIvectorDim() const {
+  if (ivector_ != NULL)
+    return ivector_->Dim();
+  else if (online_ivector_feats_ != NULL)
+    return online_ivector_feats_->NumCols();
+  else
+    return 0;
+}
+
+
+void DecodableNnetSimpleLooped::AdvanceChunk() {
+  int32 begin_input_frame, end_input_frame;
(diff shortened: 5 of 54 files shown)


vijayaditya:
I would recommend testing the following:
1. When the model is trained with a fixed left-context does decoding with an infinite left-context help ?
     (I think it might. During my aspire experiments I once trained a BLSTM with 0 left/right context, due to a script error. When decoding using an extra-{left,right}-context of 40 was important even if there was a mismatch with the training. Using 0 extra-{left,right}-context led to an increase WER by 10% rel. IIRC)

2. This is slightly related to this PR so I am mentioning it here:
Do we need to change the training strategy by using variable chunk-left-context during training ? @GaofengCheng 's dropout in recurrent connections would be useful here. But he would have to update the dropout component to use the same dropout mask value for the entire history vector rather than a separate mask value for each element of the history vector. Also as we want to ensure that we simulate infinite left context during training we might want to check if larger chunk-left-context values would help (they might not; previous experiments did not show any change in results when using contexts of 40,60 or 80).

Finally as @freewym implemented the context preserving LSTM, I think he might benefit by following this PR closely, as he can check what changed in this implementation and how is better/worse compared to his implementation. @freewym Also, let us know if you can guide @tomkocse in designing the tests.

I will add more comments when I go through the PR.

freewym:
Sure. I will look into this PR.

tomkocse:
I could help if a slight delay is tolerable .

danpovey:
OK.

GaofengCheng:
@vijayaditya do you mean I should update a dropout component (maybe we can name it ScaleDropoutComponent) like this:
    if scaledropout-rate=0.2 then a matrix like this is added: ` [[0.8,0.8...]...[0.8,0.8,...]]`
My experiments using the DropoutComponent accomplished by @pegahgh improves results on Chain BLSTM AMI SDM.
If the ScaleDropoutComponent is what you mean.
I think we can do experiments between the ScaleDropoutComponent and DropoutComponent

danpovey:
No, he means a form of dropout where instead of the dropout-mask matrix having random zeros or ones, it has rows that are randomly either zero or one [but the same for each row.]

tomkocse:
@Dan, I start to work on this PR. To make sure i am working in a right way, here are my procedures:
1. check out the nnet3 binaries in branch danpovey:looped_computation
2. use the TDNN+LSTM scripts to train the nnet model
3. test this nnet model with either [true|false] in the --looped-computation option
Is this fine ?

vijayaditya:
If I understand correctly you have to just test a change during decoding.
You need not train a new model. You can use the models available
at /export/b01/vpeddinti/xconfig/egs/swbd/s5c/exp/chain/tdnn_lstm_1a_ld5_sp.

--Vijay

danpovey:
Yes, the model is unchanged.
It's about using the new decoding program with -looped in the name.
If you have any C++-level crashes, no need to debug it yourself, just ask
me.

GaofengCheng:
@danpovey @vijayaditya I plan to compare the two dropout methods, my experiments show dropout by Pegah have already helped, but we should try more methods. I will update a new dropout in this week, @danpovey  renaming the dropout by Pegah DotDropout and naming the idea by Vijay RowDropout, do you think it is OK?

danpovey:
Conventional dropout is what we have implemented, so the name Dropout is
probably suitable.  Anyway, it would probably be easiest to implement this
as an option to the DropoutComponent, as it only really affects one line of
the code, where the mask is created.  Maybe an option 'dropout_per_frame'
(default = false).

Dan

tomkocse:
@danpovey i remove the use of option --extra-left-context-initial in he decoding and i get an error 
ERROR (nnet3-latgen-faster-looped:CreateComputation():nnet-compile.cc:58) Not all outputs were computable, cannot create computation.

tomkocse:
the original decoding script was using  --extra-left-context-initial=-1 but nnet3-latgen-faster-looped:Check() not allow to use this option with negative value

danpovey:
OK, show me the path please if it's on our grid, I'll debug it when I have
a chance.

tomkocse:
@danpovey 
my working directory is /export/a08/tomko/chain/kaldi-trunk/egs/swbd/s5c/
the logs are at exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_tg_looped/log/decode.*.log
the top level script is local/chain/multi_condition/run_tdnn_7f.sh
steps/nnet3/decode.sh was modified to support an option --looped-computation [true|false]

danpovey:
OK... I changed-user to you (since I have root) and I fixed the bugs in the
code-- the code can run now.
I notice that that location does not seem to be being tracked under git.
At least, the git tracking information is seriously out of date.  So I'm
assuming you have some non-git method of tracking your changes, maybe based
on rsync or something.  Obviously I couldn't commit the changes I made.
The code changes I made were all in src/nnet3/, I believe.

Dan

tomkocse:
I just decode again and here are the result
without looped:
%WER 9.8 | 1831 21395 | 91.3 5.9 2.8 1.1 9.8 43.3 | exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_fsh_fg/score_10_0.0/eval2000_hires.ctm.swbd.filt.sys
%WER 19.7 | 2628 21594 | 82.8 11.9 5.3 2.5 19.7 55.5 | exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_fsh_fg/score_9_0.0/eval2000_hires.ctm.callhm.filt.sys
%WER 14.8 | 4459 42989 | 86.9 8.7 4.4 1.7 14.8 50.5 | exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_fsh_fg/score_10_0.0/eval2000_hires.ctm.filt.sys

with looped:
%WER 9.8 | 1831 21395 | 91.3 5.9 2.8 1.1 9.8 43.2 | exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_fsh_fg_looped/score_10_0.0/eval2000_hires.ctm.swbd.filt.sys
%WER 19.7 | 2628 21594 | 82.8 11.9 5.3 2.5 19.7 55.6 | exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_fsh_fg_looped/score_9_0.0/eval2000_hires.ctm.callhm.filt.sys
%WER 14.8 | 4459 42989 | 86.9 8.7 4.4 1.7 14.8 50.5 | exp/chain/tdnn_7f_sp_rvb1/decode_eval2000_sw1_fsh_fg_looped/score_10_0.0/eval2000_hires.ctm.filt.sys

I am working at a non-git folder as I suppose I don't have to commit anything in this PR.
Can you obtain the changes you made and commit to your own branch ?

vijayaditya:
@Tom Did you manage to decode any RNN based recipes ? Unlike FF-DNNs we
expect to see change in results here. Also could you please paste the
real-time factors for the two TDNN decodes ?

--Vijay

tomkocse:
Here are the real time factors for 5 out of 50 decoding processes:
without looped:
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:246) Time taken 150.483s: real-time factor assuming 100 frames/sec is 0.448454
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:246) Time taken 164.609s: real-time factor assuming 100 frames/sec is 0.899796
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:246) Time taken 97.1057s: real-time factor assuming 100 frames/sec is 0.974077
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:246) Time taken 283.33s: real-time factor assuming 100 frames/sec is 0.820485
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:246) Time taken 239.113s: real-time factor assuming 100 frames/sec is 0.897536

with looped:
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:251) Time taken 312.587s: real-time factor assuming 100 frames/sec is 0.931537
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:251) Time taken 226.34s: real-time factor assuming 100 frames/sec is 1.23723
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:251) Time taken 66.7531s: real-time factor assuming 100 frames/sec is 0.669607
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:251) Time taken 405.598s: real-time factor assuming 100 frames/sec is 1.17456
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:251) Time taken 162s: real-time factor assuming 100 frames/sec is 0.608084

@vijayaditya I will use your TDNN+LSTM model to decode

vijayaditya:
I don't understand why the looped decode would run twice as long. It is
expected to be faster as we reuse the state of the network from the
previous chunk. Could you please do a controlled comparison i.e., decode a
set of utterances on the same machine using the two binaries.


--Vijay

tomkocse:
I decoded on the same machine and time taken for looped is still longer:
without looped:
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:246) Time taken 178.792s: real-time factor assuming 100 frames/sec is 0.532816
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:249) Done 137 utterances, failed for 0
LOG (nnet3-latgen-faster:main():nnet3-latgen-faster.cc:251) Overall log-likelihood per frame is 0.474043 over 33556 frames.
LOG (lattice-scale:main():lattice-scale.cc:90) Done 137 lattices.

looped:
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:251) Time taken 188.171s: real-time factor assuming 100 frames/sec is 0.560766
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:254) Done 137 utterances, failed for 0
LOG (nnet3-latgen-faster-looped:main():nnet3-latgen-faster-looped.cc:256) Overall log-likelihood per frame is 0.474045 over 33556 frames.
LOG (lattice-scale:main():lattice-scale.cc:90) Done 137 lattices.

danpovey:
Hm.  In order to find out why the speed is different you may have to use
gprof.  I think this requires adding the -pg flag to the CXXFLAGS in
kaldi.mk (but look it up).  You'll likely have to recompile all of Kaldi.

Tom: I'm afraid I don't know what changes I made because the way I would
find out is by using a git command.   However, they will all be in
nnet3/*.{h,cc}.  I assume that you copied all those files from a git
repository-- how about you copy them back and then figure it out?  I can't
easily do this myself because I don't know exactly what revision you had
copied from.



Dan

tomkocse:
@vijayaditya I got an error when i decode with your TDNN+LSTM model
ERROR (nnet3-latgen-faster-looped:RemoveRedundantConfigLines():nnet-nnet.cc:492) Name 'tdnn1.affine' is not allowable, in line: component-node name=tdnn1.affine component=tdnn1.affine input=lda

Do you know how I can resolve this error?

vijayaditya:
You are probably using an old code base which does not have the latest
xconfig commit.

Check if the looped computation PR is up to date.

Vijay

danpovey:
Merge changes from the 'master' branch and it should work.  There was a
code change.
But before you attempt a merge it would be better if you make sure my bug
fixes are correctly checked into the git repo-- I don't want you to lose
the changes.

Dan

tomkocse:
@vijayaditya Here are the result of looped decoding, but the result get worse:
without looped:
%WER 9.5 | 1831 21395 | 91.5 5.6 2.9 1.0 9.5 42.8 | exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_fsh_fg/score_11_1.0/eval2000_hires.ctm.swbd.filt.sys
%WER 19.0 | 2628 21594 | 83.7 11.7 4.6 2.7 19.0 54.6 | exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_fsh_fg/score_9_0.0/eval2000_hires.ctm.callhm.filt.sys
%WER 14.4 | 4459 42989 | 87.5 8.6 3.9 1.9 14.4 49.6 | exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_fsh_fg/score_10_0.0/eval2000_hires.ctm.filt.sys

looped:
%WER 9.7 | 1831 21395 | 91.4 5.3 3.3 1.1 9.7 43.0 | exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_fsh_fg_looped/score_12_1.0/eval2000_hires.ctm.swbd.filt.sys
%WER 21.3 | 2628 21594 | 80.8 12.4 6.8 2.2 21.3 59.9 | exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_fsh_fg_looped/score_10_0.0/eval2000_hires.ctm.callhm.filt.sys
%WER 15.6 | 4459 42989 | 86.0 8.9 5.1 1.7 15.6 52.9 | exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_fsh_fg_looped/score_11_0.0/eval2000_hires.ctm.filt.sys

danpovey:
Hm, interesting.
Have a look at the details scoring output (I forget which file, the one
that has the actual output, maybe the ctm) and see if the errors cluster at
a particular part of the sentence, e.g. the start.
Also please run the standard LSTM decoding with a longer chunk size, like
100 frames, and see if it's closer to the looped decoding.  It could be
that for some reason it doesn't like the effectively-infinite context.

Dan

danpovey:
Also, diff the ctms and see if there are any obvious effects.  Things to
look for:
  - A consistent shift in the times.
  - Changes that seem to cluster at multiples of the chunk boundaries (50)
  - Changes near the beginning or end of the sentence.
I'm asking about if there are any general repeating patterns.

vijayaditya:
Try training a TDNN-LSTM with longer chunk-left-context (try 120). This
might better help emulate infinite left context during training.

--Vijay

danpovey:
I'm not sure that I want to encourage this setting as it makes models too
expensive to train.  (maybe dropout would be better solution).
Let's just confirm that it was the problem.

vijayaditya:
This is just to verify if the WER increase is due to mismatch in training
and test settings. We can then think of better ways to alleviate this issue.

--Vijay

danpovey:
OK, makes sense I guess.

tomkocse:
I change --frames-per-chunk from 50 to 150 but the gap between standard decoding and looped decoding didn't get closer
Here I just post the WER of the whole eval2000 as it has larger difference
standard decoding, --frames-per-chunk = 50, WER: 14.4
standard decoding, --frames-per-chunk = 150, WER: 14.2
looped deocding, --frames-per-chunk = 50, WER: 15.6
looped deocding, --frames-per-chunk = 150, WER: 15.6

The --frames-per-chunk option didn't affect the WER in looped docoding, is that normal ? 

I looked at the diff between the ctm files output from  standard and looped decoding.
I didn't observe any general repeating pattern, at least by my obervation

danpovey:
Thanks..
With the looped decoding, the frames-per-chunk should not make much of a
difference.
With the standard decoding, we expect the WER to get worse if the
--frames-per-chunk is made shorter than the value used in training (150).
The question is, does it also get worse if we make it larger?   We expect
it to get worse with larger --frames-per-chunk if for some reason the LSTMs
don't like to get infinite history.  Try 500.  That should be worse too.

Dan

tomkocse:
Why does --frames-per-chunk is needed in training now ? Am i wrong in understand that every frames are trained separately before (i.e. --frames-per-chunk = 1) ?

I saw that in the current TDNN+LSTM setup, the LSTM component feed it back to itself with a delay of 3 frames, that means it will only have memories of the previous 3 frames, why do you say it will get infinite history ?

danpovey:
The --frames-per-chunk is the length of the egs in chain training.  If you
are doing non-chain training with LSTMs, there is also some similarly named
parameter, possibly --frames-per-chunk, that dictates the size of the egs.
With LSTMs you train on fairly big chunks.

The infinite history is because LSTMs are recurrent.

The --frames-per-chunk I was asking you to change was in the decoding
script, though.

tomkocse:
I just verified that using a too large -frames-per-chunk will lead to worse result:
standard decoding, --frames-per-chunk = 50, WER: 14.4
standard decoding, --frames-per-chunk = 150, WER: 14.2
standard decoding, --frames-per-chunk = 500, WER: 14.6

Do you have other suggestions in analysing the worse result coming with looped decoding ?

danpovey:
Try 1000 too, please.

tomkocse:
result updated:
standard decoding, --frames-per-chunk = 50, WER: 14.4
standard decoding, --frames-per-chunk = 150, WER: 14.2
standard decoding, --frames-per-chunk = 500, WER: 14.6
standard decoding, --frames-per-chunk = 1000, WER: 14.6
standard decoding, --frames-per-chunk = 2000, WER: 14.6

danpovey:
Can you please show us the full decode.sh command line in both cases, and
the full c++-level command line?

Also, you may want to play with the --extra-left-context-initial option.
E.g. try setting it to 40.

danpovey:
I mean play with the --extra-left-context-initial option in the looped case.

tomkocse:
for standard decoding:
       steps/nnet3/decode.sh --acwt 1.0 --post-decode-acwt 10.0 \
          --looped_computation false \
          --nj 50 --cmd "$decode_cmd" $iter_opts \
          --extra-left-context $extra_left_context  \
          --extra-right-context $extra_right_context  \
          --frames-per-chunk "$frames_per_chunk" \
          --online-ivector-dir $ivector_dir/ivectors_${decode_set}_vijay \
         $graph_dir data/${decode_set}_hires \
         $dir/decode_${decode_set}${decode_dir_affix:+_$decode_dir_affix}_${decode_suff} || exit 1;

nnet3-latgen-faster --online-ivectors=scp:exp/nnet3_rvb/ivectors_eval2000_vijay/ivector_online.scp --online-ivector-period=10 --frame-subsampling-factor=3 --frames-per-chunk=150 --extra-left-context=50 --extra-right-context=0 --extra-left-context-initial=-1 --extra-right-context-final=-1 --minimize=false --max-active=7000 --min-active=200 --beam=15.0 --lattice-beam=8.0 --acoustic-scale=1.0 --allow-partial=true --word-symbol-table=exp/chain/tdnn_lstm_1a_ld5_sp/graph_sw1_tg/words.txt exp/chain/tdnn_lstm_1a_ld5_sp/final.mdl exp/chain/tdnn_lstm_1a_ld5_sp/graph_sw1_tg/HCLG.fst 'ark,s,cs:apply-cmvn --norm-means=false --norm-vars=false --utt2spk=ark:data/eval2000_hires/split50/1/utt2spk scp:data/eval2000_hires/split50/1/cmvn.scp scp:data/eval2000_hires/split50/1/feats.scp ark:- |' 'ark:|lattice-scale --acoustic-scale=10.0 ark:- ark:- | gzip -c >exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_tg/lat.1.gz'


for looped case:
       steps/nnet3/decode.sh --acwt 1.0 --post-decode-acwt 10.0 \
          --looped_computation true \
          --nj 50 --cmd "$decode_cmd" $iter_opts \
          --frames-per-chunk "$frames_per_chunk" \
           --online-ivector-dir $ivector_dir/ivectors_${decode_set}_vijay \
         $graph_dir data/${decode_set}_hires \
         $dir/decode_${decode_set}${decode_dir_affix:+_$decode_dir_affix}_${decode_suff}_looped || exit 1;

nnet3-latgen-faster-looped --online-ivectors=scp:exp/nnet3_rvb/ivectors_eval2000_vijay/ivector_online.scp --online-ivector-period=10 --frame-subsampling-factor=3 --frames-per-chunk=150 --minimize=false --max-active=7000 --min-active=200 --beam=15.0 --lattice-beam=8.0 --acoustic-scale=1.0 --allow-partial=true --word-symbol-table=exp/chain/tdnn_lstm_1a_ld5_sp/graph_sw1_tg/words.txt exp/chain/tdnn_lstm_1a_ld5_sp/final.mdl exp/chain/tdnn_lstm_1a_ld5_sp/graph_sw1_tg/HCLG.fst 'ark,s,cs:apply-cmvn --norm-means=false --norm-vars=false --utt2spk=ark:data/eval2000_hires/split50/1/utt2spk scp:data/eval2000_hires/split50/1/cmvn.scp scp:data/eval2000_hires/split50/1/feats.scp ark:- |' 'ark:|lattice-scale --acoustic-scale=10.0 ark:- ark:- | gzip -c >exp/chain/tdnn_lstm_1a_ld5_sp/decode_eval2000_sw1_tg_looped/lat.1.gz'

tomkocse:
better WER for looped case with --extra-left-context-initial option
looped, --frames-per-chunk = 150, --extra-left-context-initial = -1 , WER: 16.9
looped, --frames-per-chunk = 150, --extra-left-context-initial = 40 , WER: 15.9

danpovey:
I'm a bit confused- wasn't it 15.6 before?  is this with different settings?

tomkocse:
oh sorry, i mixed up the trigram result, I re-post the result again. These are the fisher 4gram WER for the whole eval2000:
standard decoding, --frames-per-chunk = 150,--extra-left-context-initial = -1,  WER: 14.2
looped, --frames-per-chunk = 150, --extra-left-context-initial = -1 , WER: 15.6
looped, --frames-per-chunk = 150, --extra-left-context-initial = 40 , WER: 14.5

danpovey:
OK, cool.  So the remaining degradation we have to address is 14.2 ->
14.5.  Probably you can set --extra-left-context-initial=5 or something
like that and it will be as good or almost as good as with 40-- please try
that.
I am hoping that a combination of dropout, and possibly training on
variable-length chunks including longer chunks, will get rid of the
remainder of the problem.

@vimalmanohar is working on code to check in to support dropout schedules,
and I believe @GaofengCheng may still be doing experiments figuring out the
best dropout schedules.  Later we can test whether this gets rid of the
problem; in case it doesn't, I am working on a plan to dump variable-length
egs for chain training, including longer egs.

tomkocse:
updated result:
standard decoding, --frames-per-chunk = 150,--extra-left-context-initial = -1, WER: 14.2
looped, --frames-per-chunk = 150, --extra-left-context-initial = -1 , WER: 15.6
looped, --frames-per-chunk = 150, --extra-left-context-initial = 5 , WER: 15.5
looped, --frames-per-chunk = 150, --extra-left-context-initial = 40 , WER: 14.5

danpovey:
Interesting-- and slightly disappointing that it requires, during test
time, the full amount of repetitions of the initial frame that we gave it
during training (i.e. it learned to rely on that).  The reason why we pad
the initial frame so much during training is to ensure that all the egs
have the same structure, so that we don't incur extra costs during
compilation.  Otherwise there is no reason to do it.

What I am considering doing is to relax the requirement that all the egs
should have the same structure, e.g. to allow them to have different
numbers of frames and to be padded or not padded (but still with only a
limited number of options in each training run-- not completely
free-form).  Then, nnet3-merge-egs will be modified to be able to batch up
only egs that share the same structure, into minibatches.

This would still incur some extra costs in compilation, because instead of
one structure for the egs, there might be, say, 12, in a given run
(corresponding to different lengths of chunks, and whether or not the
chunks are padded on the left or on the right, or both; padding would not
be needed at file begin/end, but we might sometimes choose to do it anyway
so that the model learns to deal with both options).  Thanks to the
computation caching, I am hoping that in practice this will not be
excessive.  In addition, there is a change I am thinking of making that
should make the compilation extremely fast, as long as all the egs in a
minibatch share the same structure (factor of 20 so, probably), so
compilation time won't really be an issue.  I may make this other change
(to compilation speed) first as it's less disruptive.

vijayaditya:
I guess another solution would involve @gaofeng 's frame-level dropout; as
it varies the history available randomly.

--Vijay

danpovey:
I'm commenting on this PR just as a mechanism to make people aware of how I'm trying to address this mismatch in chunk-size...
What I think is the problem is that if you run an LSTM for longer than in training, you can reach parts of the activation space that you couldn't reach during training; because a term gets added to the 'c' each time, and the magnitude of the hidden activations in 'c' is bounded by the number of LSTM steps.
The way I'm addressing this is scaling the c_t by a value slightly less than one in the recurrence; since we already have a BackpropTruncationComponent there I'm doing it that way [and also doing it on m_t, just for convenience.]   This will mean that the values of c_t are now bounded, and when the sequence lengths increase there is a less dramatic change to the parts of activation space that you can reach.
This looks extremely promising based on the train and valid probs.  I hope it pans out.  Perhaps someone will remember seeing people do this elsewhere.

danpovey:
This method [recurrence scaled by a factor less than one] seems to be effective in making the looped computation as good as the chunk-based computation, and it doesn't seem to degrade the baseline WER results; if anything they are very slightly better [could be noise].  My best results so far are with a scale of 0.85.

danpovey:
this is in the shortcut branch; closing this separate PR.

Issue #1189, opened 2016-11-14
arpa2fst crash (does not strip whitespaces)


jtrmal:
```
LOG (arpa2fst:Read():arpa-file-parser.cc:90) Reading \data\ section.
LOG (arpa2fst:Read():arpa-file-parser.cc:145) Reading \1-grams: section.
ERROR (arpa2fst:Read():arpa-file-parser.cc:143) line 55203 [\2-grams: ]: invalid directive, expecting '\2-grams:'
```
the crash is caused by expecting '\2-grams:' and getting '\2-grams: '
I tested it and it works after removing that whitespace.
Will look into it

PR #1191, opened 2016-11-14
fixing arpa2fst to allow traling whitespace in the  headers

jtrmal:
Adresses #1189

### DIFF
--- a/src/lm/arpa-file-parser.cc
+++ b/src/lm/arpa-file-parser.cc
@@ -38,10 +38,8 @@ ArpaFileParser::ArpaFileParser(ArpaParseOptions options,
 ArpaFileParser::~ArpaFileParser() {
 }
 
-std::string rtrim(const std::string &str) {
-  std::string ret = str;
-  ret.erase(ret.find_last_not_of(" \n\r\t")+1);
-  return ret;
+void TrimTrailingWhitespace(std::string *str) {
+  str->erase(str->find_last_not_of(" \n\r\t") + 1);
 }
 
 void ArpaFileParser::Read(std::istream &is, bool binary) {
@@ -90,7 +88,7 @@ void ArpaFileParser::Read(std::istream &is, bool binary) {
   while (++line_number_, getline(is, current_line_) && !is.eof()) {
     if (current_line_.empty()) continue;
 
-    current_line_ = rtrim(current_line_);
+    TrimTrailingWhitespace(&current_line_);
 
     // Continue skipping lines until the \data\ marker alone on a line is found.
     if (!keyword_found) {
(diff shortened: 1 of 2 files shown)


src/lm/arpa-file-parser.cc
```diff
@@ -38,6 +38,12 @@ ArpaFileParser::ArpaFileParser(ArpaParseOptions options,
 ArpaFileParser::~ArpaFileParser() {
 }
 
+std::string rtrim(const std::string &str) {
```
danpovey: Can you please rename this to TrimTrailingWhitespace?

src/lm/arpa-file-parser.cc
```diff
@@ -147,6 +155,7 @@ void ArpaFileParser::Read(std::istream &is, bool binary) {
     int32 ngram_count = 0;
     while (++line_number_, getline(is, current_line_) && !is.eof()) {
       if (current_line_.empty()) continue;
+      current_line_ = rtrim(current_line_);
```
danpovey: Adding this line will incur a big performance penalty and I don't think it's necessary to solve the current issue.  The ARPA file format is not well defined, but I don't think a reasonable person would put leading space before the \1-grams:  marker.

src/lm/arpa-file-parser.cc
```diff
@@ -147,6 +155,7 @@ void ArpaFileParser::Read(std::istream &is, bool binary) {
...
       if (current_line_[0] == '\\') break;
```
danpovey: Since you are looking at this... this code doesn't seem to be checking that the n-gram orders are as expected, it's just checking that the first character is backslash.  That's not very nice, and would lead to unexpected behavior if someone, say, skipped an ngram order.

src/lm/arpa-file-parser.cc
```diff
@@ -38,6 +38,12 @@ ArpaFileParser::ArpaFileParser(ArpaParseOptions options,
 ArpaFileParser::~ArpaFileParser() {
 }
 
+std::string rtrim(const std::string &str) {
+  std::string ret = str;
+  ret.erase(ret.find_last_not_of(" \n\r\t")+1);
```
danpovey: space around '+'

src/lm/arpa-file-parser.cc
danpovey: ... actually, why not make it take a pointer to string and return void, and just use resize to remove the trailing whitespace?

jtrmal:
I thought about it just a minute ago.   I could also iterate from
backwards,  so it would be almost no-op for proper strings.
Y

jtrmal:
updated. how about this?

danpovey:
Looks good-- you can merge once tests finish.

Issue #1190, opened 2016-11-14
Chain,nnet3 : --chunk-{left,right}-context not honored when reusing egs


vijayaditya:
When using `--egs.dir` option the current chain recipe does not ensure that the {left,right}-context being used for training is same as the one used specified in `--egs.chunk-{left,right}-context`. This is because `nnet3-chain-copy-egs` did not have `--{left,right}-context` options when the script was written. However @GaofengCheng recently added support for this in 93f2f65 . So this option has to be supported even in the scripts.

vijayaditya:
The context options were also not being passed on to the ComputeCvProbabilities methods even in the normal frame level training scripts.

@vimalmanohar Check if there are similar issues in the sMBR training scripts.

vijayaditya:
Same issue with the get_post.combined.* jobs which are used for computing the average posterior. 

BTW I want to add that these bugs would not be a issue unless you were training RNN AMs and using egs from a different experiment which has egs with different contexts.

vijayaditya:
I am pushing a PR correcting this in a few minutes.

vince62s:
do you recall this ?
https://groups.google.com/forum/#!searchin/kaldi-help/egs%7Csort:relevance/kaldi-help/udYrYTEDycU/JuodilwSAAAJ

I am not sure to get exactly what the issue would be.

vijayaditya:
The nnet3 compiler decides the unfolding width of RNNs based on the left and right context available. So providing more context than specified means that your RNN history vector is not exactly what you wanted it to be. This can lead to context mismatch errors between training and decoding; and finally affect the performance. 

In our experimentation we found that the RNN performance is usually sensitive to the amount of context available, at least up to a certain context-width.

vijayaditya:
Addressed in #1194

PR #1194, opened 2016-11-14
WIP : Added changes to ensure that the correct left/right contexts are used…

vijayaditya:
… for computation in nnet3 setup. For more details see Issue #1190 

Testing in progress.

### DIFF
--- a/egs/swbd/s5c/local/chain/run_blstm.sh
+++ b/egs/swbd/s5c/local/chain/run_blstm.sh
@@ -1,0 +1,1 @@
+tuning/run_blstm_6i.sh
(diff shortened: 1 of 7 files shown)


danpovey:
OK, @vijayaditya, merge when you think it's tested.
@vimalmanohar, please notice this, it might require changes in your PR where you refactor the training libraries [and BTW, I'd like to know where we are with that.]

vijayaditya:
@freewym Could you please review this PR. I will merge after your signoff.

freewym:
Looks fine to me.

danpovey:
Vijay, are you still awake?  This PR was problematic, it is causing a crash for me.  I think it has to do with frame-shifting, more context is needed.  Or the limitation of context should be done after the frame shift.

danpovey:
OK, I have fixed this with commit ac1f932.   It changes nnet3-chain-copy-egs so that it applies the frame shift before limiting context.  
Note: the crash was with TDNNs.  With BLSTMs you wouldn't have seen a crash because of the extra context.

Issue #1089, opened 2016-10-03
Efficient implementation of LSTM


danpovey:
I have an idea for a more efficient implementation of LSTM in the nnet3 framework.
@kangshiyin, if you have time, you might be a good person to do the coding aspects this, since it involves CUDA.  But @vijayaditya will probably want to do the scripts, and he and others may also have comments on the idea more generally.

I am referring to the formulation of LSTMs in http://static.googleusercontent.com/media/research.google.com/en//pubs/archive/43905.pdf.  Although we are using LSTMP, it won't affect what goes below.
Suppose the cell dimension is C.  We can compute the following 4 C-dimensional quantities in a single affine operation (this is something Vijay has already done in something that was not checked in, I think):

```
   i_part = W_{ix} x_t + W_{im} m_{t-1} + b_i
   f_part = W_{fx} x_t + W_{fm} m_{t-1} + b_f
   c_part = W_{cx} x_t + W_{cm} m_{t-1} + b_c
   o_part = W_{cx} x_t + W_{om} m_{t-1} + b_o
```

We will create a component that has an input dimension of 5  C (consisting of all of the above quantities plus c_{t-1}), and outputs something of dimension 2 C, consisting of c_t and m_t.  The component will have parameters of dimension 3 C (consisting of the three diagonal weight matrices w_{ic}, w_{fc} and w_{oc}).  Writing the diagonal weight matrices in lower-case to emphasize their diagonal-ness and using `*` for elementwise product, the component will do the following computation... and note that some of the quantities it uses do not appear in the input or output, they are internal to the function:

```
 i_t = Sigmoid(i_part + w_{ic}*c_{t-1})
 f_t = Sigmoid(f_part + w_{fc}*c_{t-1})
 c_t = f_t*c_{t-1} + i_t * Tanh(c_part)
 o_t = Sigmoid(o_part + w_{oc}*c_t)
 m_t = o_t * Tanh(c_t)
# note: the outputs are just c_t and m_t.
```

A note on derivatives: it will be easiest for both the CUDA kernels and for the integration with the natural gradient code, if the derivatives w.r.t. the 3 C parameters are output into a matrix whose dimension is (the number of separate indexes for which this function is being applied) by (3C): that is, that the kernels will not be responsible for summing up the gradient across all elements of the minibatch, but just computing the parameter gradients as a matrix and outputting them into a temporary variable declared inside the Backprop function.

danpovey:
sorry still editing

[...]

PR #1201, opened 2016-11-21
Some partial changes (don't compile yet) towards a more efficient LSTM...

danpovey:
implementation.
This is for @kangshiyin to look at, as I'll be asking him to help with the GPU aspects of this.
This relates to Issue #1089


kangshiyin:
Thanks, I will start working on this.

src/cudamatrix/cu-math.cc
```diff
@@ -329,14 +330,11 @@ void BackpropLstmNonlinearity(const CuMatrixBase<BaseFloat> &input,
                  self_repair_sum_out != NULL && SameDim(params, *params_deriv) &&
                  value_sum_out->NumRows() == 5 &&
                  value_sum_out->NumCols() == cell_dim &&
-                 SameDim(* ...
-                         // HERE
+                 SameDim(*value_sum_out, *deriv_sum_out) &&
+                 self_repair_sum_out->NumRows() == 5 &&
+                 value_sum_out->NumCols() == cell_dim);
```
kangshiyin: Should it be `self_repair_sum_out->NumCols() == cell_dim` ?

danpovey:
You're right, will fix.

danpovey:
I just checked in some test code for the derivatives in the LSTM
backpropagation code.  The CPU version tests correctly (first time!).
However it does not test the self-repair.  I plan to check various
diagnostics in the actual training to check that the self-repair is acting
as it normally does.

If you can find some way to at least check that the CPU and GPU versions
are doing the same thing w.r.t. the self-repair and stats accumulation
code, it would be nice.

danpovey:
I'm concerned about the effects of roundoff if it's converted to
BaseFloat.  And other value-sum/deriv-sum quantities are done in double.
So it's best if you leave it at double, but the quantities that you compute
internally can be in BaseFloat (and only convert to double when you add to
the deriv_sum and value_sum).

kangshiyin:
ok, I will follow your type definitions.

danpovey:
Note, this will not be merged, @kangshiyin's PR #1206 is the version that will be merged.  But leaving this up for now as we might learn something from the automated checks.

danpovey:
This PR will be merged indirectly via PR #1229.

Issue #1203, opened 2016-11-21
 Shifting of chain egs happens before copy-egs so there can be insufficient context errors


vijayaditya:
In ac1f932 , shifting of chain-egs happens before copy-egs call. The python training script has to be modified to request egs with contexts which account for this.

danpovey:
Can you explain what the issue is in more detail?
You understand why I changed it, right?

vijayaditya:
Sorry one of my setups broke and I thought this was the issue. However the setup was previously working as the copy preceded the shift, while it shouldn't have actually worked due to insufficient context.

PR #1204, opened 2016-11-21
chain-train : corrected a minor bug where the chain/train.py where re…

vijayaditya:
…quired context values were incorrect.

Related to the incorrectly raised issue #1203 .


danpovey:
merge yourself ..

danpovey:
Not sure if this would impact vimal?  Let him know if so.  Also

vijayaditya:
Yes it does. @vimalmanohar please see this commit.

PR #1205, opened 2016-11-21
Added the basic TDNN+LSTM scripts

vijayaditya:
The TDNN+LSTM was found to be as good as the BLSTM architecture in previous experiments (not checked in). This PR has scripts to specify these models as xconfig scripts. 

I will add the results in a day.


danpovey:
This is great news as it means we can do online decoding with less latency.
@tomkocse, you could if you want use this setup to test the 'looped'
decoding PR.

vijayaditya:
The latency of the TDNN+LSTM models right now is 21 frames, less than half
of that of the BLSTMs (51 frames). Further the TDNN layer outputs can be
shared across neighboring chunks unlike the BLSTM. However the best
performing TDNN+LSTM model has the same number of parameters as the BLSTM.
Reducing the parameters to 75% or 50% has increased WER.

The current set of experiments being run are designed to find the optimal
combination of TDNN and LSTM layers.

--Vijay

vijayaditya:
I added the tdnn_7i recipe which has almost twice the number of parameters as our current best tdnn, tdnn_7h. It however does not give us better results. However the log-probabilities (both train and CV) look better. The `local/chain/run_tdnn.sh` script still points to tdnn_7h.sh

image

vijayaditya:
I added the results for the TDNN+LSTM setup. As expected they look better than the TDNN  and LSTM results. Further TDNN+LSTM without pretraining is working better than TDNN+LSTM with it. (Please note that we saw improvements due to removal of layer-wise pretraining in TDNNs but not in LSTMs).

```bash
#System                  lstm_6j   tdnn_7h    this(old) this(new)
#WER on train_dev(tg)      14.66     13.84     13.88     13.42
#WER on train_dev(fg)      13.42     12.84     12.99     12.42
#WER on eval2000(tg)        16.8      16.5      16.0      15.7
#WER on eval2000(fg)        15.4      14.8      14.5      14.2
#Final train prob     -0.0824531-0.0889771-0.0515623-0.0538088
#Final valid prob     -0.0989325 -0.113102-0.0784436-0.0800484
#Final train prob (xent)      -1.15506   -1.2533 -0.782815   -0.7603
#Final valid prob (xent)      -1.24364  -1.36743 -0.946914 -0.949909
```

BLSTM training is still running. I might have the results in a day. The below plots compares the log-likelihood values for the TDNN+LSTM and BLSTM setups.

image

danpovey:
OK with me.

vijayaditya:
ready for review.

danpovey:
Looks good to me, merge when you want.

PR #1206, opened 2016-11-22
WIP: Efficient implementation of LSTM layer

kangshiyin:
This relates to Issue #1089
GPU code for forward propagation has passed unit test and speed test. The speed is close to the upper bound, which is equal to reading `input` once and writing `output` once without any real computing work.


danpovey:
Amazing-- thanks!!
I'm working on the code for the backprop on CPU right now.
It's a lot more complicated, because it's backprop and because of
self-repair.  I hope to be able to check it into the branch on my PR
to-morrow, and you can then hopefully merge and implement the GPU version.

danpovey:
Shiyin, I have checked in a rough draft of the CPU backprop code for the
LSTM component.  To-morrow I'll be debugging this and writing test code
that numerically checks the derivatives.

Dan

danpovey:
@vijayaditya, it would be great if you could either implement the scripting-level support for LSTMs based on this, or direct someone like @GaofengCheng how to do it.  I'm a little hazy on, say, the interaction with projected LSTM.   
I'm not sure what to do regarding whether to just replace the existing LSTM layers with this one, or name it differently.  I suppose you'd want it named differently initially for testing purposes (?).
Be aware that while the aspects of it that don't touch on self-repair have been debugged via the automated tests, the self-repair stuff has not been debugged and could well contain errors.  What I was hoping was that we could compare the plots of things like avg-deriv and self-repair-proportion, or look at them in the logs, and find any problems that way.  However, that would require changes to your plot-generation code to be able to parse the Info() output of this component.

kangshiyin:
I've matched the CPU impl. All GPU tests have passed. The speed is ok when `cell_dim` >=1024, and there may be some space to speed up for smaller `cell_dim`.

danpovey:
Great, thanks!

I'll leave your PR up there for now without merging, because I'd like to
integrate it with script-level changes and merge them all together-- I'll
wait for Vijay or Gaofeng to respond.

src/cudamatrix/cu-kernels.cu
```diff
@@ -2630,6 +2630,467 @@ static void _diff_log_softmax(const MatrixDim in_deriv_dim,
...
+                     f_t = Sigmoid(f_part + w_{fc}*c_{t-1})
+                     c_t = f_t*c_{t-1} + i_t * Tanh(c_part)
+                     o_t = Sigmoid(o_part + w_{oc}*c_t)
+                     m_t = o_t * Tanh(c_t)
+
+We use 1D thread block with CU1DBLOCK threads.
+It works best when cell_dim is a multiple of CU1DBLOCK.
+We use 1d Grid. Each block is working on one row of the in and out matrices.
+*/
+template<typename Real>
```
danpovey: Just FYI, I believe the BaseFloat typedef has also been made available inside this part of the code.
So it's not really necessary to have it templated, you could just define it directly with BaseFloat.  But up to you.

src/cudamatrix/cu-math.h
```diff
@@ -78,8 +78,147 @@ void Group2norm(const CuMatrixBase<Real> &src,
...
+                              const CuMatrixBase<Real> &output_deriv,
+                              const CuMatrixBase<double> &deriv_sum_in,
+                              const CuVectorBase<Real> &self_repair_config,
+                              double count_in,
+                              CuMatrixBase<Real> *input_deriv,
+                              CuMatrixBase<Real> *params_deriv,
+                              CuMatrixBase<double> *value_sum_out,
+                              CuMatrixBase<double> *deriv_sum_out,
+                              CuMatrixBase<Real> *self_repair_sum_out);
+template<typename Real>
```
danpovey: Again, all we really need is the BaseFloat version of this, but it's not something I care about a lot.
You could comment here that this function is a version of BackpropLstmNonlinearity that only uses
the CPU even if a GPU is available, and it has been made available for testing purposes.

src/cudamatrix/cu-kernels.cu
kangshiyin: I think template is required so that we can run both `double` and `float` tests at the same time.

danpovey:
Regarding double and float tests-- nothing forces you to have the test code
templated on double and float.  It was that way because most of the other
code required that, but there is no need to follow that pattern for this
code.  E.g. the the top-level test function that's templated on Real could
just call a non-templated function, or you could call it directly from
main().

kangshiyin:
I see. Thx. I prefer to testing both:)

2016年11月25日 1:09 PM，"Daniel Povey" <notifications@github.com>写道：

danpovey:
Closing this PR as we will merge it indirectly, via PR #1229.

PR #1207, opened 2016-11-22
changes for GALE mandarin setup

jtrmal:
I did some changes in order to allow me to train on all data.
depends on @naxingyu and @danpovey if they decide merge or if a better strategy would be only cherry-pick some files or changes.
y.


danpovey:
I don't like the weird RESULTS filename.

jtrmal:
that's how these filenames have been always generated for gale setups -- see arabic gale also...
I wanted to provide some numbers for @naxingyu as a reference.
I don't think it matters that much, unless you wanna merge it -- I'd suggest waiting how @naxingyu  feels better solution for him.

danpovey:
I guess there's no hurry to merge, but do you see a reason why we would not
merge this?  Does it improve results?
I don't like the idea of having a separate convention for writing results
for particular setups.  I must have not reviewed the original commits for
that setup very well.  I'd rather just have it as a RESULTS file, and
remove any old results files.  People can figure out from the git log when
and who committed it, if they need.

jtrmal:
I can, of course, change it -- no issue in that. Maybe the original idea was that the RESULTS file won't be overwritten by people running the recipes and generating their own results and if the results should be used as a reference, they should be manually renamed to RESULTS -- in that case I'm at fault, I wasn't sure of the reasons.
y.

naxingyu:
Thank you Yenda for the updated numbers. I was curious why there wasn't 
a RESULT file, now it explains. There are several issues with the 
original gale_mandarin recipe. It details WER of "report" and 
"conversation" but the references don't exist. Maybe it was not 
finished. And it confuses "test" with dev. But I saw that Yenda has 
reshaped the data structure so it should be addressed by now. Some of 
the example scripts are out-dated, using unavailable options of the 
steps scripts. I'll see if there is anything I can do with this, but I 
think we should agree on a suitable data structure. What Yenda proposes 
seems reasonable.

Actually, just think out loud, there is no such thing as 'eval2000' and 
'rt03' for mandarin/chinese. All the test sets are just dev sets. Is 
there a possibility that we find a suitable chinese test set?

Xingyu

danpovey:
Xingyu, I'd say go ahead and do what you think is best, taking into account
Yenda's changes.  This recipe is not properly maintained by anyone else;
don't assume that whatever you see there makes sense.  When and if you
think I should merge Yenda's PR, let me know.  I don't have time to think
about this much.

naxingyu:
OK.

naxingyu:
@jtrmal I don't see where you use all data to train. It's still running on LDC2013S08 and LDC2013T20.

jtrmal:
Let me check,  maybe I didn't commit the changes to run.sh

jtrmal:
I commit the run.sh and path.sh. Also renamed the RESULTS file. Sorry about
that.
y.

egs/gale_mandarin/s5/local/gale_prep_dict.sh
```diff
@@ -54,23 +66,6 @@ gawk 'NR==FNR{words[$1]; next;} ($1 in words)' \
...
-  cd tools/g2p
-  echo '#include <cstdio>' >> Utility.hh # won't compile on my system w/o this "patch"
-  python setup.py build 
-  python setup.py install --prefix=.
-  cd ../..
-  if [ ! -f tools/g2p/lib/python${pyver}/site-packages/g2p.py ]; then
-    echo "Sequitur G2P is not found - installation failed?"
-    exit 1
-  fi
-fi
```
naxingyu: Why is this removed? g2p.py is used later.

jtrmal:
we have an installation script in tools/extras/
It should be checking for availability of g2p.py and printing helpful error
message in case the g2p is not found, though -- my bad
y.

egs/gale_mandarin/s5/local/gale_prep_dict.sh
naxingyu: There should be a helpful error message if g2p is not found.

egs/gale_mandarin/s5/run.sh
```diff
@@ -130,40 +148,42 @@ steps/align_fmllr.sh --nj $nJobs --cmd "$train_cmd" \
...
-  steps/decode_sgmm_rescore.sh --cmd "$decode_cmd" --iter $n --transform-dir exp/tri3b/decode data/lang_dev \
+done
+
+for n in 1 2 3 4; do
+  steps/decode_sgmm2_rescore.sh --cmd "$decode_cmd" --iter $n --transform-dir exp/tri3b/decode data/lang_test \
     data/dev exp/sgmm_5a/decode exp/sgmm_5a_mmi_onlyRescoreb0.1/decode$n
 done
 
+wait
 local/nnet/run_dnn.sh
```
naxingyu: in local/nnet/run_dnn.sh, there is a invalid option to steps/nnet/train.sh (--use-gpu-id) and a missing `done` near the end of the script.

egs/gale_mandarin/s5/local/split_wer_per_corpus.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
naxingyu: This new script is not called in run.sh. We may think of deprecate the original split_wer.sh and rename this one as split_wer.sh. That one was errornous anyway.

danpovey:
Xingyu, I don't think Yenda was planning to do much more work on this.  I
think what's best is if you merge changes from his PR as you like, fix it
in the way that you think best, and then make  your own pull request, whih
we can merge.

naxingyu:
OK.

在 2016/12/1 4:31, Daniel Povey 写道:

jtrmal:
This is being handled by #1253, so closing...

PR #1208, opened 2016-11-22
ivector compatibility checks

jtrmal:
@vijayaditya @danpovey since this is a quite pervasive change I don't assume we should merge right away, but I will need your feedback on the way it is approached.

I tried to build on Vijay's ivector id check. THe conceptual change is that I generated the id on the fly  for an extractor directory (or any directory that contains the final.ie), so that it can lie on an RO medium (i.e. sometimes we are symlinking other colleagues systems).
THat however does not span accross the chain extractor->ivectors->decode nor extractor->ivectors->egs->training, so I'm not sure if it's of any value, as it is encreasing a state-space of the combinations we have to handle (i.e. how to react if we can get id for one directory and not for another directory).

y.


egs/wsj/s5/steps/nnet2/get_ivector_id.sh
```diff
@@ -0,0 +1,39 @@
...
+ivecdir=$1
+
+if [ -f $ivecdir/final.ie.id ] ; then
+  cat $ivecdir/final.ie.id
+elif [ -f $ivecdir/final.ie ] ; then
+  # note the creation can fail in case the extractor directory
+  # is not read-only media or the user des not have access rights
+  md5sum $ivecdir/final.ie | awk '{print $1}' | tee $ivecdir/final.ie.id 2>/dev/null || true
+else
+  exit 66
```
danpovey: why not exit 1?

jtrmal:
I'm checking for that specific code in the python script, just to handle
that specific case
y.

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
```diff
@@ -57,8 +57,12 @@ def CheckIfCudaCompiled():
 
 
 class KaldiCommandException(Exception):
-    def __init__(self, command, err):
+    def __init__(self, command, err, stdout=None, code=None):
         Exception.__init__(self,"There was an error while running the command {0}\n".format(command)+"-"*10+"\n"+err)
+        self.stderr = err;
+        self.stdout = stdout;
+        self.command = command;
+        self.code = code
```
danpovey: The way I want to change the RunKaldiCommand stuff is to just join the subprocess stdout/stderr to the parent process's stdout/stderr.   [stdout should not matter.]
That way, any errors or warnings generated by queue.pl  will be printed immediately.

jtrmal: I can call that Popen with stdout=subprocess.STDOUT,stderr=subprocess.STDERR, which I think would do what you want, but I'm not sure of consequences -- it would be better if @vijayaditya or @vimalmanohar  would propose a viable solution, as they authored the lib

danpovey:
Yes, better if Vimal or Vijay do this upstream, in a separate commit.

egs/wsj/s5/steps/nnet3/nnet3_train_lib.py
jtrmal: I can help with implementing it, if Vijay is too busy with his thesis, but he should propose what would be the way to go.

danpovey:
I notice that there  is a WIP PR 967 that does the same thing... however, yours may be more advanced.  I don't have a lot of time to devote to this. @jtrmal, it would be best if you take responsibility for making sure something happens on this.  I have no fundamental objections to your PR.  Merge it if, after comparing with Vijay's, you think it's all OK.  But don't make those changes to the python-script error printing, we should do that separately.

danpovey:
@jtrmal says that after we merge Vimal's reworking of the Python scripts, he will rebase and re-test this pull request.  So that's why this is not being merged yet.

danpovey:
@jtrmal, do you want to rebase this against the 'shortcut' branch?  Then we don't have to test too thoroughly, it will get tested before release.

jtrmal:
ok, rebased onto shortcut.
Running some basic checks on RM  just to see if it behaves as expected. Do not merge into shortcut yet.
Noticed the python training library  (egs/wsj/s5/steps/nnet3/nnet3_train_lib.py) has disappeared -- was it renamed? Or perhaps there were some issues during rebase. WIll investigate tmrw.

danpovey:
it was moved and refactored, it's in steps/libs/nnet3/

jtrmal:
I still wanna play with it to see how it behaves when the scripts will get run on an existing system.

jtrmal:
closed as this was replaced by #1395

PR #1211, opened 2016-11-23
swbd: updated TDNN+LSTM recipe. Interleaving these layers provides go…

vijayaditya:
…od improvement.


danpovey:
This is pretty cool.

PR #1219, opened 2016-11-27
Lstm integration

danpovey:
This pull request involves just one commit after merging changes from PR #1206 and PR #1066.

It is a first step towards script-level integration of the new, 'fast' LSTM layer... I am also developing some xconfig-based scripts on the tedlium s5_r2 setup.

But @GaofengCheng, if you can get it done first, maybe based on this, I'll merge your version.


danpovey:
For everyone's info: this is going well.  It's (very approximately) twice faster than the old LSTM implementation, and the WER results are if anything slightly better than before.
I've tested the LSTM and am testing the LSTMP now; it looks good, based on objective values.

Note: to use the new 'fast' LSTM layers you have to use a different layer name, fast-lstm-layer or fast-lstmp-layer.

The only thing in this PR that is not completely back compatible is, I modified and simplified the way the shrinkage is computed, and made it less configurable.  Instead of choosing to do it on sigmoid versus tanh units, it computes a 'saturation value' (between zero and one, defined as 1.0 - (avg_deriv / max_possible_avg_deriv), where 0.0 is completely unsaturated, and 1.0 is completely saturated.  This is averaged over all (sigmoid or tanh) units, and it's computed by an external perl script get_saturation.pl, that parses the info output (ugly yes, but it was easier to code in perl and will be easier to test as an external script).  You specify the saturation value as --trainer.optimization.shrink-saturation-threshold (default: 0.4, which is equivalent to the old threshold of avg-deriv=0.15 for sigmoid units).  In practice the saturation values for the sigmoid and tanh units tend to be quite similar, so this new way of specifying the threshold should not affect shrinkage substantially.

@vijayaditya: if it's not inconvenient you might want to do your tuning experiments on top of this PR.  I hope to merge this PR within the next few days.  And note: this PR includes Vimal's PR and I'm keeping it up to date by including Vimal's changes.

vijayaditya:
OK. This stage of TDNN+LSTM experiments are completed to my satisfaction. I
will merge the latest recommended recipes tomorrow. I will further
fine-tune the recipe after re-running these recommended recipes using the
new layer.

--Vijay

danpovey:
Closing this as I have created a replacement PR #1229 with the same changes, in the Kaldi repository instead of my personal repo.

PR #1221, opened 2016-11-28
Add PowerPC64le in Kaldi build system

jrziviani:
This commit introduces the PowerPC64le in the Kaldi build system, accepting Atlas, OpenBLAS and CUDA.

A tipical example to build Kaldi in Power8 le is:

 1 - clone Kaldi
 2 - after downloading openfst, download the recent version of
     config.guess
 3 - build tools
 4 - clone the latest openblas:
     https://github.com/xianyi/OpenBLAS/tree/master
     chekout master branch, build with make and make install
 5 - build Kaldi in src/
     run ./configure --shared --mathlib=OPENBLAS \
         --openblas-root=/opt/OpenBLAS/
     make -j 50

Signed-off-by: Jose Ricardo Ziviani <joserz@linux.vnet.ibm.com>
Signed-off-by: Leonardo Augusto Guimaraes Garcia <lagarcia@linux.vnet.ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,3 +51,14 @@ Development pattern for contributors
    You can use the [Google's cpplint.py]
    (https://raw.githubusercontent.com/google/styleguide/gh-pages/cpplint/cpplint.py)
    to verify that your code is free of basic mistakes.
+
+Platform specific notes
+-----------------------
+
+PowerPC 64bits little-endian (ppc64le):
+- Kaldi is expected to work out of the box in RHEL >= 7 and Ubuntu >= 16.04 with
+  OpenBLAS, ATLAS, or CUDA.
+- CUDA drivers for ppc64le can be found at [https://developer.nvidia.com/cuda-downloads]
+  (https://developer.nvidia.com/cuda-downloads).
+- An [IBM Redbook] (https://www.redbooks.ibm.com/abstracts/redp5169.html) is
+  available as a guide to install and configure CUDA.
--- a/src/configure
+++ b/src/configure
@@ -462,6 +462,8 @@ function configure_cuda {
       else
         cat makefiles/cuda_64bit.mk >> kaldi.mk
       fi
+    elif [ "`uname -m`" == "ppc64le" ]; then
+      cat makefiles/cuda_ppc64le.mk >> kaldi.mk
     else
       cat makefiles/cuda_32bit.mk >> kaldi.mk
     fi
@@ -526,6 +528,8 @@ function linux_atlas_failure { # function we use when we couldn't find
    echo ATLASLIBS = [somewhere]/liblapack.a [somewhere]/libcblas.a [somewhere]/libatlas.a [somewhere]/libf77blas.a $ATLASLIBDIR >> kaldi.mk
    if [[ "`uname -m`" == arm* ]]; then
      cat makefiles/linux_atlas_arm.mk >> kaldi.mk
+   elif [[ "`uname -m`" == ppc64le ]]; then
+     cat makefiles/linux_atlas_ppc64le.mk >> kaldi.mk
    else
      cat makefiles/linux_atlas.mk >> kaldi.mk
    fi
@@ -581,6 +585,8 @@ function linux_configure_debian_ubuntu {
   echo ATLASLIBS = $ATLASLIBS >> kaldi.mk
   if [[ "`uname -m`" == arm* ]]; then
     cat makefiles/linux_atlas_arm.mk >> kaldi.mk
+   elif [[ "`uname -m`" == ppc64le ]]; then
+    cat makefiles/linux_atlas_ppc64le.mk >> kaldi.mk
   else
     cat makefiles/linux_atlas.mk >> kaldi.mk
   fi
@@ -604,6 +610,8 @@ function linux_configure_debian_ubuntu3 {
   echo ATLASLIBS = $ATLASLIBS >> kaldi.mk
   if [[ "`uname -m`" == arm* ]]; then
     cat makefiles/linux_atlas_arm.mk >> kaldi.mk
+  elif [[ "`uname -m`" == ppc64le ]]; then
+    cat makefiles/linux_atlas_ppc64le.mk >> kaldi.mk
   else
     cat makefiles/linux_atlas.mk >> kaldi.mk
   fi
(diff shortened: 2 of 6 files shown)


jtrmal:
I'm ok with merging this. I'd suggest providing a readme and/or an example script for the extra steps because they are not completely obvious for someone who would be starting with all of this. 
y.

danpovey:
Yes, we definitely need to put detailed instructions in the repo itself, or
to modify tools/Makefile so that it does the needed thing automatically.

It might be better to have the INSTALL files in / and src/ have sections
for different architectures, or  modify the scripts so the same
instructions just work.

What OS did you get this to work on?

Dan

jtrmal:
Also, "downloading the latest version of config.guess" is rather unhelpful -- can you provide more detailed info where to find it or which version worked for you and so on?
Can you please also elaborate on the OS version you've used to compile and run?

jrziviani:
Hello!

Actually the only differences are:

1) **OpenFST**: the version used by Kaldi comes with a config.guess that doesn't recognize ppc64le correctly. When running make for the first time it fails:
```
[snip]
checking dependency style of g++... gcc3
checking build system type... ./config.guess: unable to guess system type

This script, last modified 2012-02-10, has failed to recognize
the operating system you are using. It is advised that you
download the most up to date version of the config scripts from

  http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD
and
  http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD

If the version you run (./config.guess) is already up to date, please
send the following data and any information you think might be
pertinent to <config-patches@gnu.org> in order to provide the needed
information to handle your system.
[snip]
```

So I replaced the existing config.guess with the option suggested (http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD), re-run make and it went fine.

2) **OpenBLAS**: Recent versions have patches to fix/improve support on Power8, so we need it.

**Notes**:
- Atlas looks fine as is. 
- Os tested: Ubuntu but it is expected to work on Fedora and CentOS for ppc64le as well.
- Cuda: AFAIK it's Ubuntu only, requires external package from NVidia website.

3) **My next steps**: I'll change the download scripts to automate items 1 and 2 (only for ppc). Do you guys have any other suggestion/concern?

Thank you for reviewing the patch!

Ziviani

laggarcia:
CUDA on POWER8 is available for Ubuntu and RHEL.

laggarcia:
Is updating the OpenFST and OpenBLAS versions that are currently downloaded by kaldi something out of question? That would solve the differences between POWER8 and x86 builds and would probably not generate any drawbacks for x86.

jtrmal:
It;s not out of the question, but not completely straightforward, because
bunch of the internals of the OpenFST has changed, so the Kaldi fst
extension in src/fst/ and src/fstbin would have to be fixed/re-implemented.
Someone told me a couple of months back they tried, but it didn't work --
neither they nor I pursued it further, so no estimation how difficult it
would be.
y.

danpovey:
Kaldi already supports two OpenFst versions-- you can uncomment something
in the Makefile to use 1.4.1.

Regarding OpenBLAS, we just download the most recent one from github.

jtrmal:
To clarify...  there are three different OpenFst release versions we are
talking about:
openfst-1.3.4 -- supported in kaldi, last version that does not need C++11
support, not developed anymore  by the openfst community
openfst-1.4.y -- supported in kaldi, needs C++11,  not developed anymore
 by the openfst community (IIRC)
openfst-1.5.x -- unsupported in kaldi, needs C++11, actively developed and
supported by the openfst community

there is no particular reason why 1.5.x could not be supported, but the
internals are not backward compatible (IIRC), so it would need some
hacking, understanding of the algorithms and time.

You guys seemed to try the default version of openfst we are using in
kaldi, i.e. 1.3.4.
Dan was suggesting that you could use 1.4.y, which autotools might already
contain support for your architecture
y.

danpovey:
BTW, guys-- we do want to use your help to provide support for this
architecture, but we want you to to work with us to try a few things out,
and bounce some ideas back and forth on the best way to support this
setup.  Be prepared for some back and forth, and some more work, before
this is merged.

jrziviani:
@jtrmal 

openfst 1.4.1 still uses an old config.guess version. So after downloading openfst I will download and replace config.guess from "http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD", if it's a ppc64 system.

_**Important**: openfst 1.4.1 is missing, the build fails when I try to use it. I was able to test because I found this package here: http://pkgs.fedoraproject.org/repo/pkgs/openfst/openfst-1.4.1.tar.gz/ca8f1730b9b9b281e515611fa9ae23c0/  I'll open a bug._

@jtrmal, @danpovey and @laggarcia 

Yes! forget about OpenBLAS, it works out of the box. The only thing is openfst/config.guess.

@danpovey 

Sure! I appreciate your interest in this patch, you can suggest/comment as you wish. :)

Thank you

jtrmal:
Looks good to me. Do you mind waiting a bit? We might be including OpenFst 1.5.x support in a few days so I think it would be great if you could verify your patch works for that version as well.

Also, I wonder if we want some platform-specific notes (perhaps in the main README) even just to say that everything should work out of the box. @danpovey ?

danpovey:
Yes, I think platform-specific notes would be good...  @jtrmal, if you come
up with suggestions, we'll talk about it then.

jrziviani:
@jtrmal hello!

I added the section with a few comments. Please, feel free to suggest any correction/improvement.
About openfst 1.5, sure. Just tell me when it's ready and I'll check it.

jtrmal:
Thanks. I think this is perfect. There is a PR #1241  that adds support for 1.5.4. I think at this time we do not have a timeline to merge it -- we probably try to make the jump together with migrating to C++11 compiler. But you can try the patch and let us know if it compiled OK.

@danpovey How about merging now and asking @jrziviani to validate the  compilation after switching to 1.5.4 and C++11? Actually, I think it should be fairly painless.

danpovey:
Since @dogancan is getting into the details of the build system, I think he would be better placed to review this than me.
@dogancan, my main concern here is that this doesn't break the build for existing platforms.  For the new PowerPC architecture, as long as it's better than nothing it would be OK with me.

dogancan:
These seem like pretty innocuous changes. They should not affect the rest of the build system. The config.guess patch will be obsolote after the C++11 upgrade since we will only support OpenFst >= 1.5.3. We will also need to add "-std=c++11" flag to CXXFLAGS. I can do those changes later. LGTM for merge.

danpovey:
OK thanks, merging.

jtrmal:
Thaks to all of you, guys!

PR #1222, opened 2016-11-28
Pass -f (force) to rm, otherwise user is prompted

aevernon:


danpovey:
Won't merge this because your problem is in the options in your .bashrc or similar file-- you have 'kiddie' mode enabled, probably some option like
alias rm=’rm –I’
To work around this we'd have to change it everywhere-- anyway, -f has other bad qualities like overriding permissions.

PR #1224, opened 2016-11-28
changed the definition of deriv-truncate-margin option. Now the margi…

freewym:
…n is defined on top of the model contexts, which applies to more general network archs

### DIFF
--- a/egs/ami/s5/local/chain/run_blstm_ami_5.sh
+++ b/egs/ami/s5/local/chain/run_blstm_ami_5.sh
@@ -128,7 +128,7 @@ if [ $stage -le 17 ]; then
     --trainer.optimization.initial-effective-lrate 0.001 \
     --trainer.optimization.final-effective-lrate 0.0001 \
     --trainer.optimization.momentum 0.0 \
-    --trainer.deriv-truncate-margin 10 \
+    --trainer.deriv-truncate-margin 8 \
     --egs.stage $get_egs_stage \
     --egs.opts "--frames-overlap-per-eg 0" \
     --egs.chunk-width $chunk_width \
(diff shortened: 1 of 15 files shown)


freewym:
Since the model contexts include the input splicing (usually [-2,-1,0,1,2]), so now {min|max}-deriv-time are actually offset by 2. If we want to make them the same as before, we just need to set  deriv-truncate-margin=8.

freewym:
I changed the margin to 8. Now it is equivalent to the previous margin=10. I think We could change it in the future if we find a better margin.

danpovey:
OK, I guess this is pretty harmless.  Merging.

freewym:
@vimalmanohar you might also need to make a minor change according to the change made in train.py here.

danpovey:
I think it shows up as a merge conflict-- I know because I merged this
commit into my branch that has Vimal's changes. But it's straightforward to
resolve.

danpovey:
Hmm... the same option should probably appear in train_rnn.py and
train_raw_rnn.py, but it doesn't.  Vimal, you should probably take care of
that (if you think that options should be there, which I think makes sense.)

I was also a little confused by this code (from Vimal's branch).. I assume
it's for back compatibility?

    parser.add_argument("--trainer.num-chunk-per-minibatch",
                        "--trainer.rnn.num-chunk-per-minibatch",
                        type=int,
                        dest='num_chunk_per_minibatch', default=512,
                        help="Number of sequences to be processed in "
                        "parallel every minibatch")

freewym:
Right now train_rnn.py uses --trainer.rnn.num-bptt-steps to determine {min,max}-deriv-time, which assumes it network is "pure" recurrent;  if necessary I can add the option --trainer.deriv-truncate-margin and deprecate num-bptt-steps.

danpovey:
I'm concerned that any changes you make to currently-checked-in code will
clash with Vimal's PR.  Is it possible for you to make a patch against
Vimal's PR, and create a WIP PR that contains Vimal's PR plus your changes
on top of it?  Then Vimal or I can pull changes from that branch.

Also, Vimal, can you please merge master into your branch?  It will make my
life easier in figuring out what changes I have in my LSTM-integration
branch, versus your branch.  There may be some bug fixes from me that
should be included.

BTW, one option I am considering is to merge your changes along with my
fast-lstm changes, since I'm working on top of your PR.

freewym:
OK. + @vimalmanohar in case you didn't get noticed from here

PR #1230, opened 2016-11-30
fix assert; use fmax overloading that has been documented.

kangshiyin:
A follow up fix to #1228 
All GPU tests have passed.


danpovey:
Thanks!  Merging.

PR #1233, opened 2016-11-30
WIP: Support for Windows threads

nshmyrev:
This patch implements Kaldi threading with Windows threads. Better for deployment if you don't have pthread-win32 dependency. Uncomplete yet, but decoding works on Windows. Let me know if you are interested in this.

Thanks


src/base/kaldi-math.h
```diff
@@ -69,9 +70,9 @@
 #endif
 
 
-#define KALDI_ISNAN std::isnan
-#define KALDI_ISINF std::isinf
-#define KALDI_ISFINITE(x) std::isfinite(x)
+#define KALDI_ISNAN(x) _isnan(x)
+#define KALDI_ISINF(x) (!_finite(x))
+#define KALDI_ISFINITE(x) _finite(x)
```
jtrmal: I think _isnan() and _finite() are Windows SDK only, i.e. I  don't think this compile on Linux anymore 
are you sure std::isnan and std::isinf and std::isfinite are not on Windows? I was able to compile the solution a couple of months back... Perhaps your VS is old(er)?

src/feat/pitch-functions.cc
```diff
@@ -136,7 +136,7 @@ void ComputeNccf(const VectorBase<BaseFloat> &inner_prod,
                inner_prod.Dim() == nccf_vec->Dim());
   for (int32 lag = 0; lag < inner_prod.Dim(); lag++) {
     BaseFloat numerator = inner_prod(lag),
-        denominator = pow(norm_prod(lag) + nccf_ballast, 0.5),
+        denominator = pow((BaseFloat)(norm_prod(lag) + nccf_ballast), (BaseFloat)0.5),
         nccf;
```
jtrmal: all typecast in the C++ style please, i.e. static_cast<BaseFloat>(x)

src/thread/kaldi-barrier.cc
```diff
@@ -18,15 +18,15 @@
 // limitations under the License.
 
 
-#include <pthread.h>
```
jtrmal: Again, I think your change is breaking linux building...

src/nnet3/nnet-compile.cc
```diff
@@ -156,18 +156,13 @@ void Compiler::ComputeDerivNeeded(
...
       const Component *c = nnet_.GetComponent(node.u.component_index);
-      if (c->Properties() & kUpdatableComponent) {
-        const UpdatableComponent *u = dynamic_cast<const UpdatableComponent*>(c);
-        KALDI_ASSERT(u != NULL);
-        if (u->LearningRate() != 0)
-          (*deriv_needed)[step] = true;
-      }
+      if (c->Properties() & kUpdatableComponent)
+        (*deriv_needed)[step] = true;
     }
```
jtrmal: I'm not sure about this change? If it realy is a fix, then perhaps do it in a separate PR

src/thread/kaldi-barrier.h
```diff
@@ -41,12 +43,15 @@ class Barrier {
...
  private:
+
+#ifndef WIN32
   pthread_mutex_t     mutex_;     ///< Mutex which control access to barrier 
   pthread_cond_t      cv_;        ///< Conditional variable to make barrier wait
 
   int32                 threshold_; ///< size of thread-group
   int32                 counter_;   ///< number of threads we wait for
   int32                 cycle_;     ///< cycle flag to keep synchronized
+#endif
```
jtrmal: I might be mistaken, but this makes Barrier() operations effectively no-op

src/util/kaldi-table-inl.h
```diff
@@ -540,7 +540,7 @@ template<class Holder>  class SequentialTableReaderArchiveImpl:
     }
     std::istream &is = input_.Stream();
     is.clear();  // Clear any fail bits that may have been set... just in case
-    // this happened in the Read function.
```
jtrmal: Is this intentional change?

windows/generate_solution.pl
```diff
@@ -976,6 +978,12 @@ sub isValidProjectLine{
 }
 ##foreach my $f (@makefiles) { print STDERR "Adding $f\n"; }
 
+@makefiles = ();
+my @folders = ("decoder", "lat", "hmm", "nnet2", "tree", "feat", "fstext", "cudamatrix", "matrix", "thread", "util", "base");
```
jtrmal: Why are you limiting the number of libraries you'll be building?

nshmyrev:
> I think it might be cleaner to either create classes for Windows only, i.e. create MultiThreaderWin class and so on... Or put it into a separate file. Or both. But I think it might be easier for further maintenance than the code with too many ifdefs in it.

Thanks for review, Jan. I agree that ifdefs around the codebase are not great. On the other hand the threading things in Kaldi are not very complex. If most of the ifdefs would stay in kaldi-thread, would you be more willing to accept this?

jtrmal:
I understand but you can always disable/unload the parts of solution you
don't want to build. Even if I accept your premise that tests and binaries
are not worth the compilation time, this is a breaking change for people
who are using a different set of libraries.
I think it would be better to provide an option the script that would
generate solution with disabled/unloaded tests and disabled/unloaded
binaries in general (note that I don't know where the information about a
particular project has been unloaded is stored in the solution files)
y.

jtrmal:
Even in that case I'd be in favor of alternative solutions.
a) complete isolation of the windows code  -- this is not as harsh request
as it seems, because you can isolate pthread implementation in one
namespace, windows implementation in another and then just used ifdef to
make this completely transparent to the kaldi code...
b) as this is not a complete work, it might be less painful to use the STL
threading (that would be perhaps even more useful for kaldi community as
whole)
y,

nshmyrev:
Ok, thanks. Closing this one then, probably not worth the attention.

src/feat/wave-reader.cc
```diff
@@ -357,7 +357,7 @@ void WaveData::Write(std::ostream &os) const {
   int num_clipped = 0;
   for (int32 i = 0; i < num_samp; i++) {
     for (int32 j = 0; j < num_chan; j++) {
-      int32 elem = static_cast<int32>(trunc(data_ptr[j * stride + i]));
```
nshmyrev: VS2010 doesn't have trunc. Parts of this patch are exactly for VS2010.

nshmyrev: Microsoft is truly awful, both C++ and their new things like UWP or async/await. Don't even think about C++11 please.

danpovey:
I was actually thinking that we could resolve this more cleanly by
upgrading to c++11.  But I guess you're saying that they are not going to
support that soon...

danpovey:
Regarding 'trunc', we could actually change trunc to floor here.  It makes
more sense because by changing the sign of the rounding when you cross
zero, you introduce low-pass noise, like a step function on zero crossings.
It would  make very little difference to checked-in recipes.

jtrmal:
I think that relates only to VS2010 -- when I look at the feature charts,
VS 2012 and VS2015 should be supporting those features.
y.

jtrmal:
https://msdn.microsoft.com/en-us/library/hh567368.aspx

I don't know about the quality of the implementation, of course.
y.

danpovey:
Nickolay, we do want to support Windows better, and we do want your help.  
Can we start with the aspects of this PR that don't touch on threading?
E.g. have you do your typecasts in static_cast style, and do the trunc->floor change, which
is probably a good thing anyway?
I also don't want to lose track of the issue of how best to do the threading stuff in Windows.
Personally I wouldn't mind just biting the bullet and upgrading to C++11 (or at least a small subset of C++11).  Yenda thinks we could support C++11 as an alternative, via ifdef.
I also don't think we should be targeting such old compiler versions.

nshmyrev:
Thank you, sure, I'll update this to only static_casts then. I'll add an option in generate to skip tests too.

PR #1238, opened 2016-12-01
Add support for doing frame-subsampling with regular (non-chain) nnet3 models

hhadian:


src/hmm/hmm-utils.h
```diff
@@ -245,6 +245,11 @@ bool SplitToPhones(const TransitionModel &trans_model,
    @param subsample_factor [in] The frame subsampling factor... normally 1, but
                                 might be > 1 if we're converting to a reduced-frame-rate
                                 system.
+   @param repeat_frames [in]    Only relevant when subsample_factor != 1
+                                If true, repeat frames of alignment by
+                                'subsample_factor' after alignment
+                                conversion, to keep the alignment the same
+                                length as the input alignment.
```
danpovey: add after this:
```
  [note: we actually do this by interpolating 'subsample_factor' 
  separately generated alignments, to keep the phone boundaries
  the same as the input where possible.]
```

egs/swbd/s5c/local/nnet3/run_reduced_frame.sh
```diff
@@ -0,0 +1,144 @@
+#!/bin/bash
+
+# This scripts does nnet3-tdnn training and decoding
```
danpovey: As regards the naming of the scripts, we need to use a convention that will scale as we do more
experiments.  Let's use 'lfr' (for low frame-rate) as the extension, and you can create the file
`tuning/run_tdnn_lfr1a.sh`
and make a soft link
run_tdnn_lfr.sh -> tuning/run_tdnn_lfr1a.sh
Remember to document in a comment, which specific script in tuning/ it was copied from,
and if possible add in a comment the summary line generated from steps/info/nnet3_dir_info.pl (excessive line length is OK), and
from local/nnet3/compare_wer_general.sh or whatever we now call that script, to compare
it with the baseline... if you don't have the baseline locally, just compare it with itself and
manually copy the numbers from the baseline to fake it.

danpovey: It's also not ideal that this script uses the older config generation and not the xconfigs.  When you have a chance, work with vimal to get an xconfig-based version of this script; you can give it a different name, like 1a->1b.

egs/wsj/s5/steps/nnet3/chain/build_tree.sh
```diff
@@ -171,7 +172,7 @@ if [ $stage -le -1 ]; then
   # for other purposes.
   echo "$0: Converting alignments from $alidir to use current tree"
   $cmd JOB=1:$nj $dir/log/convert.JOB.log \
-    convert-ali --frame-subsampling-factor=$frame_subsampling_factor \
+    convert-ali --repeat-frames=$repeat_frames --frame-subsampling-factor=$frame_subsampling_factor \
```
danpovey: watch line length.

danpovey:
The build error was because of http://openfst.cs.nyu.edu/ being down... I just pushed a change to tools/Makefile so that it doesn't try so many times.

egs/wsj/s5/steps/nnet3/chain/build_tree.sh
```diff
@@ -25,6 +25,7 @@ frame_subsampling_factor=1
 leftmost_questions_truncate=10
 tree_stats_opts=
 cluster_phones_opts=
+repeat_frames=false
```
hhadian: Should I limit line length (to 80 chars) here too? Because it will get ugly if I do.

danpovey: you don't have to be strict about the 80 chars here... don't go much above 100 though.

egs/swbd/s5c/local/nnet3/run_reduced_frame.sh
```diff
@@ -0,0 +1,144 @@
...
+  # Create a version of the lang/ directory that has one state per phone in the
+  # topo file. [note, it really has two states.. the first one is only repeated
+  # once, the second one has zero or more repeats.]
+  rm -rf $lang
+  cp -r data/lang $lang
+  silphonelist=$(cat $lang/phones/silence.csl) || exit 1;
+  nonsilphonelist=$(cat $lang/phones/nonsilence.csl) || exit 1;
+  # Use our special topology... note that later on may have to tune this
+  # topology.
+  steps/nnet3/chain/gen_topo.py $nonsilphonelist $silphonelist >$lang/topo
```
danpovey: Regarding the topology, it's possible that a single-state topology where the first state is not 'special' would be better.  I.e. make the self-loop-pdf-class zero instead of 1.  You'd have to try that as a separate experiment.

egs/swbd/s5c/local/nnet3/run_reduced_frame.sh
```diff
@@ -0,0 +1,144 @@
...
+    --dir=$dir  || exit 1;
+
+fi
+
+graph_dir=$dir/graph_sw1_tg
+if [ $stage -le 13 ]; then
+  # Note: it might appear that this $lang directory is mismatched, and it is as
+  # far as the 'topo' is concerned, but this script doesn't read the 'topo' from
+  # the lang directory.
+  utils/mkgraph.sh --left-biphone --self-loop-scale 1.0 data/lang_sw1_tg $dir $graph_dir
```
danpovey: if you use triphone context then you would remove the --left-biphone flag here.

egs/swbd/s5c/local/nnet3/run_reduced_frame.sh
```diff
@@ -0,0 +1,144 @@
...
+  # topology.
+  steps/nnet3/chain/gen_topo.py $nonsilphonelist $silphonelist >$lang/topo
+fi
+
+if [ $stage -le 10 ]; then
+  # Build a tree using our new topology. This is the critically different
+  # step compared with other recipes.
+  steps/nnet3/chain/build_tree.sh --repeat-frames true --frame-subsampling-factor 3 \
+      --leftmost-questions-truncate $leftmost_questions_truncate \
+      --context-opts "--context-width=2 --central-position=1" \
```
hhadian: I will try it now

egs/swbd/s5c/local/nnet3/run_reduced_frame.sh
hhadian: OK

egs/swbd/s5c/local/nnet3/run_reduced_frame.sh
hhadian: While trying this non-special first state, I got this error in build-tree.sh:

> cluster-phones exp/nnet3/tdnn_lfr2a_tree_sp/treeacc data/lang_lfr2a/phones/sets.int exp/nnet3/tdnn_lfr2a_tree_sp/questions.int 
> WARNING (cluster-phones:AutomaticallyObtainQuestions():build-tree.cc:576) After filtering the tree statistics to retain only stats where pdf-class is in the set { 1 }, most of your stats disappeared: the size changed from 69459 to 0.  You might be using a nonstandard topology but forgot to modify the --pdf-class-list option (it defaults to { 1 } which is the central state in a 3-state left-to-right topology). E.g. a 1-state HMM topology would require the option --pdf-class-list=0.
> ERROR (cluster-phones:EnsureClusterableVectorNotNull():cluster-utils.cc:94) All stats are NULL.

Should I give the --pdf-class-list option to cluster-phones or something else is wrong?

hhadian: I added the following option to build_tree.sh:
`      --cluster-phones-opts "--pdf-class-list=0" `

danpovey:
That sounds right.

hhadian:
sMBR on top of LFR (tdnn_lfr1h) and regular (tdnn_3c):

Epoch1:
```
# System               tdnn_3c_sp_smbr  tdnn_lfr1h_sp_smbr
# WER on train_dev(tg)      15.63     16.39
# WER on train_dev(fg)      14.32     15.08
# WER on eval2000(tg)        18.6      18.6
# WER on eval2000(fg)        17.0      16.8
# Final train objf       0.927103  0.888498
# Final valid objf       0.889319  0.872577
```

Epoch2:
```
# System               tdnn_3c_sp_smbr  tdnn_lfr1h_sp_smbr
# WER on train_dev(tg)      15.65     16.21
# WER on train_dev(fg)      14.37     14.85
# WER on eval2000(tg)        18.6      18.4
# WER on eval2000(fg)        17.2      16.6
# Final train objf        0.91574  0.879541
# Final valid objf       0.888475  0.867044
```

Epoch3:
```
# System               tdnn_3c_sp_smbr  tdnn_lfr1h_sp_smbr
# WER on train_dev(tg)      15.81     16.13
# WER on train_dev(fg)      14.61     14.77
# WER on eval2000(tg)        18.7      18.4
# WER on eval2000(fg)        17.2      16.6
# Final train objf       0.927103  0.888498
# Final valid objf       0.889319  0.872577
```

Epoch4:
```
# System               tdnn_3c_sp_smbr  tdnn_lfr1h_sp_smbr
# WER on train_dev(tg)      15.90     16.15
# WER on train_dev(fg)      14.75     14.78
# WER on eval2000(tg)        19.1      18.4
# WER on eval2000(fg)        17.6      16.7
# Final train objf       0.927103  0.888498
# Final valid objf       0.889319  0.872577
```

Baseline (nnet3/tdnn without frame subsampling):
```
# System                  tdnn_3c
# WER on train_dev(tg)      17.37
# WER on train_dev(fg)      15.94
# WER on eval2000(tg)        20.0
# WER on eval2000(fg)        18.2
# Final train prob       -1.43781
# Final valid prob       -1.56895
```

danpovey:
That's great news!
So it looks like once we do the sMBR training, our non-chain LFR systems
are even better than our regular-frame-rate systems.  We should pursue this
because it will give an alternative to the 'chain' setup which might work
better for some datasets; and who knows, we may discover a better way to
train these systems.  E.g. we might find that it's better to initialize the
chain system from a conventionally-trained, or even
discriminatively-trained, LFR system.
Are you doing the version of the training where you compute the alignments
and denlats for all 3 shifts of the features, as in
egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh?
Vimal says that is helpful for chain systems (and presumably it will for
your LFR systems).  But of course it's on the slow side.
We'll have quite a bit of work to do all this with the newer versions of
the discriminative training scripts that I'm working on in the 'shortcut'
branch.

vijayaditya:
@Hossein could you suggest what would be a comparable chain model for the
lfr_1h setup, in terms of number of parameters and context of the model.

Also could you say what the tdnn-splicing is for this model ?

--Vijay

hhadian:
@danpovey No I am not using frame shifts. Do you want me to try it? Another interesting thing is that I had to change acoustic weight options to `--acwt 0.1 --post-decode-acwt 1.0` to get a reasonable best lm-scale for the sMBR/LFR setup. Without sMBR, we used `--acwt 0.33 --post-decode-acwt 3.0`

@vijayaditya Both lfr1h and tdnn_3c have the exact same xconfig. The main lines are:
```
  fixed-affine-layer name=lda input=Append(-2,-1,0,1,2,ReplaceIndex(ivector, t, 0)) affine-transform-file=$dir/configs/lda.mat
  relu-renorm-layer name=tdnn1 dim=1024
  relu-renorm-layer name=tdnn2 input=Append(-1,2) dim=1024
  relu-renorm-layer name=tdnn3 input=Append(-3,3) dim=1024
  relu-renorm-layer name=tdnn4 input=Append(-7,2) dim=1024
  relu-renorm-layer name=tdnn5 dim=1024
  output-layer name=output input=tdnn5 dim=$num_targets max-change=1.5 presoftmax-scale-file=$dir/configs/presoftmax_prior_scale.vec
```

danpovey:
We'll do the frame-shifted stuff later, you'll see there is an issue about
that.  No hurry on that; anyway, I want it done with my refactored and
simplified versions of the discriminative training scripts that I'm working
on in the 'shorcut' branch.

Regarding the acoustic scales-- I think a better solution might be to pass
in the appropriate acwt to the sMBR related scripts (i.e. a value of 0.33,
pass it in to the alignment, denlat-creation and discriminative-training
scripts).  Those all default to 0.1 which isn't really appropriate for the
LFR system.

Something else that just occurred to me is, about the self-loop-scale: we
normally have it approximately the same as the optimal acwt, i.e. 0.1 for
regular systems, 1.0 for chain.  You might find a value of 0.33 works
better for these LFR systems.  First try it for making the graph for
decoding the conventionally-trained LFR system, and if 0.33 is a good
value, we'll have to pass that in everywhere we'd pass it in for
discriminative training (i.e. alignment and denlat creation, and of course
whan making the graph).

Dan

vijayaditya:
@Hossein
Also when you are ultimately comparing these with chain systems, you might
want to keep in the mind the difference in splicing indices. Chain models
usually have {-3,0,3} type indices while xent systems have {-7,2} type
indices which are more sparse. This usually results in detrimental
performance in xent systems, but speeds up the frame level training. Just
for the sake of comparisons you might want to use the same type of splicing
indices in both the chain and xent models.

--Vijay

hhadian:
@danpovey OK. I will try the self-loop-scale too.
@vijayaditya That's interesting, I will try it.

vdp:
This is tangentially related at best, so apologies to everyone, but Dan's mention of of self-loop scale reminded me of something I was wondering about recently.. Couple of days ago, I was looking at *steps/nnet3/align.sh* and noticed the default there is:
    
    scale_opts="--transition-scale=1.0 --acoustic-scale=0.1 --self-loop-scale=0.1"

, which I think is meant to be used with "regular", non-chain models. I was wondering how this should be changed if used with chain- perhaps:

    scale_opts="--transition-scale=1.0 --acoustic-scale=1.0 --self-loop-scale=1.0"

Does this look correct?

Also, out of curiosity, do you know if the use of assymmetric left/right contexts(e.g. (-7,2)) in TDNN(including chain-trained) models has any effect on the alignments? My guess would be that, unless there is realignment during TDNN training(haven't studied the training scripts yet), the TDNN-based forced alignments won't be much different from the GMM-based ones?

danpovey:
>
> This is tangentially related at best, so apologies to everyone, but Dan's
> mention of of self-loop scale reminded me of something I was wondering
> about recently.. Couple of days ago, I was looking at
> *steps/nnet3/align.sh* and noticed the default there is:
>
> scale_opts="--transition-scale=1.0 --acoustic-scale=0.1 --self-loop-scale=0.1"
>
> , which I think is meant to be used with "regular", non-chain models. I
> was wondering how this should be changed if used with chain- perhaps:
>
> scale_opts="--transition-scale=1.0 --acoustic-scale=1.0 --self-loop-scale=1.0"
>
> Does this look correct?
>

That is correct.  I have some changes I am have not yet committed in the
'shortcut' branch that will make the script warn if it detects that you are
using a low-frame-rate system but you are using the default scale_opts.


> Also, out of curiosity, do you know if the use of assymmetric left/right
> contexts(e.g. (-7,2)) in TDNN(including chain-trained) models has any
> effect on the alignments? My guess would be that, unless there is
> realignment during TDNN training(haven't studied the training scripts yet),
> the TDNN-based forced alignments won't be much different from the GMM-based
> ones?
>

No, they shouldn't be shifted versus the GMM ones, since they are trained
on GMM-based alignments and not realigned during training [it would take
many of passes of realignment to shift them, anyway].  The goal in putting
asymmetric context was originally to decrease latency, but in fact the WER
seems also to be improved by having slightly more left than right context.

hhadian:
The value of 0.33 for self-loop-scale worked a bit better:
```
# System                   tdnn_lfr  tdnn_lfr_self-loop-scale-0.33
# WER on train_dev(tg)      16.86     17.00
# WER on train_dev(fg)      15.57     15.57
# WER on eval2000(tg)        19.5      19.3
# WER on eval2000(fg)        17.9      17.8
# Final train prob        -1.11497
# Final valid prob         -1.21105
```

danpovey:
Cool.
It might make a difference in the discriminative training too.
BTW, at some point you should aim to use the "new-style" discriminative
training scripts in the 'shortcut' branch that instead of calling
make_denlat.sh and get_egs_discriminative.sh, call just get_degs.sh.  There
is one in the swbd/s5c setup called local/nnet3/tuning/run_tdnn_d.sh.  The
self-loop-sclae option should be provided in several places: alignment,
mkgraph, and get_degs (at least, I may have forgotten others).
Dan

hhadian:
Yes definitely. Using the new scripts and measuring different runtimes (such as train/valid prob, prior computation, etc.) is in my task list. I will run the rest of my discriminative experiments using the shortcut branch.

hhadian:
Result of using splicing similar to chain (for comparison):
```
# System                   tdnn_c    tdnn_e  tdnn_lfr1a  tdnn_lfr1c
# WER on train_dev(tg)      17.37     16.75     17.25     17.10
# WER on train_dev(fg)      15.94     15.34     15.90     15.74
# WER on eval2000(tg)        20.0      19.5      20.1      19.2
# WER on eval2000(fg)        18.2      18.0      18.5      17.7
# Final train prob       -1.43781  -1.40491  -1.32434  -1.29898
# Final valid prob       -1.56895  -1.55255  -1.42206  -1.43117
```
These setups are all xconfig-based and all have 5 hidden layers. 
`tdnn_c` and `tdnn_lfr1a` use baseline splicing `"-2,-1,0,1,2 -1,2 -3,3 -7,2 0"`.
`tdnn_e` and `tdnn_lfr1c` use chain-like splicing `"-2,-1,0,1,2 -1,0,1 -1,0,1 -3,0,3 -3,0,3"`
Intersting note is that the chain-like spicing has been more helpful to the LFR setup than to the regular one.

danpovey:
That's great!   It looks like by this point the LFR system is giving
basically the same results as the regular-frame-rate system.
Is the PR in a situation where it's ready to merge?  Make sure that any
scripts that you have results for, have the results in a comment near the
top of the script, and please update the soft links to point to the current
script, e.g. to tdnn_e instead of tdnn_c; you can introduce soft links for
the LFR scripts too, e.g. run_tdnn_lfr.sh


Dan

vijayaditya:
On Fri, Jan 27, 2017 at 3:18 PM, Hossein Hadian <notifications@github.com>
wrote:

> Result of using splicing similar to chain (for comparison):
>
> # System                   tdnn_c    tdnn_e  tdnn_lfr1a  tdnn_lfr1c
> # WER on train_dev(tg)      17.37     16.75     17.25     17.10
> # WER on train_dev(fg)      15.94     15.34     15.90     15.74
> # WER on eval2000(tg)        20.0      19.5      20.1      19.2
> # WER on eval2000(fg)        18.2      18.0      18.5      17.7
> # Final train prob       -1.43781  -1.40491  -1.32434  -1.29898
> # Final valid prob       -1.56895  -1.55255  -1.42206  -1.43117
>
> These setups are all xconfig-based and all have 5 hidden layers.
> tdnn_c and tdnn_lfr1a use baseline splicing "-2,-1,0,1,2 -1,2 -3,3 -7,2 0"
> .
> tdnn_e and tdnn_lfr1c use chain-like splicing "-2,-1,0,1,2 -1,0,1 -1,0,1
> -3,0,3 -3,0,3"
>



>
> Intersting note is that the chain-like spicing has been more helpful to
> the LFR setup than to the regular one.
>
This is interesting. May be denser sampling @ intermediate layers is
important in low frame rate models (irrespective of the objective function
being used), especially @ the lower layers. Better confirm it is not some
other thing like the change in the context which is affecting the results.

hhadian:
> It looks like by this point the LFR system is giving
> basically the same results as the regular-frame-rate system.

Yes, however, if you remember after tuning num-leaves for the **regular** setup, ~0.3-0.4% improvement was achieved (the code was not merged as the new code required rebuilding the tree and re-alignment). Of course, even in that case, the results are still about the same. I do not have the xconfig-based results for the tuned-leaves setup (which used 20% more leaves) to show here.
I will set the soft links and push the latest changes.

> Better confirm it is not some
> other thing like the change in the context which is affecting the results.

What are you suggesting I should try? I previously tried context size change by changing the last layer splicing (from 0) to -4, 4 (I didn't realize the total context size should be a factor of 3) and there was 0.3-0.5% improvement.

danpovey:
This probably isn't what Vijay meant, but it might be helpful to shift the
final -3,0,3 to be -6,-3,0.  Sometimes having more left than right context
helps slightly.

hhadian:
OK, will try it.

vijayaditya:
Thanks for the clarification Dan; that was what I meant.

The contexts used in the new LFR models were symmetric and also smaller, so
I was recommending that Hossein match the asymmetric context used in the
original TDNN models, while using a denser set of splicing indexes.

baseline TDNN: -2,-1,0,1,2 -1,2 -3,3 -7,2 0: Total context [-13, 9]
current TDNN : -2,-1,0,1,2 -1,0,1 -1,0,1 -3,0,3 -3,0,3 : Total context
[-10, 10]

suggested TDNN : -2,-1,0,1,2 -1,0,1 -1,0,1 -3,0,3 -3,0,3 -3,0 : Total
context [-13, 10]
comparative baseline : -2,-1,0,1,2 -1,2 -3,3 -7,2 0 0: Total context [-13,
9]


It would be informative if you can try the two experiments suggested-tdnn
and comparative-baseline.

--Vijay

hhadian:
I have pushed the latest changes. The suggested experiments are running.

danpovey:
Travis build failed with compile error:
```
hmm-utils-test.cc:315:60: error: too few arguments to function ‘bool kaldi::ConvertAlignment(const kaldi::TransitionModel&, const kaldi::TransitionModel&, const kaldi::ContextDependencyInterface&, const std::vector<int>&, int32, bool, bool, const std::vector<int>*, std::vector<int>*)’
                                   NULL, &old_alignment_copy);
                                                            ^
```

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_lfr1a.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
+
+# _lfr1a is as _c, but is LFR (low frame rate): it uses triphone chain topology
```
danpovey: You should probably run an experiment with TDNN+LSTMs, too.

hhadian:
Oh! I didn't change the tester binaries. Fixed it now.

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_lfr1a.sh
hhadian: OK will do.

hhadian:
The results of trying LFR with TDNN+LSTM:
```
# System               tdnn_lstm_b   tdnn_lstm_lfr1b
# WER on train_dev(tg)      16.63     16.95
# WER on train_dev(fg)      15.30     15.74
# WER on eval2000(tg)        18.7      18.6
# WER on eval2000(fg)        17.1      17.0
# Final train prob      -0.869086 -0.724979
# Final valid prob      -0.948473 -0.778266
```

danpovey:
Cool.  Make sure to include the script in your PR.

vijayaditya:
@Hossein Could you summarize your observations from these experiments.

IIRC; You are just seeing performance comparable to normal frame rate
models in both TDNNs and LSTMs (or was it worse for LSTMs). Of course the
computation savings still make the LFR models very attractive.

Also a comparison with LF-MMI models, especially in terms of computation
costs and training times for comparable architectures, would be very
informative.

I guess the training times for LF-MMI models are still smaller as we do not
usually do the sequence training using word level LMs and also because the
neural network outputs are always computed for sequences*, as opposed to
the single frame outputs computed in xent training of LFR-TDNN models.

--Vijay

* As you might have already guessed the computation of contiguous outputs
results in savings due to sharing of intermediate convolution outputs.

hhadian:
@danpovey Will do it.
@vijayaditya Yes the WERs so far are almost the same, but the main thing that is appealing is the faster decoding time. The training time should be almost the same because we are effectively training on all frame-shifts (i.e. output frame rate is the same as input frame rate).

Another note is that if we do discriminative training on top of xent/TDNN models, then it seems LFR-models perform a little better. And the discriminative training is faster because alignment and denominator lattice generation is much faster.

I could do a comparison in terms of training time with chain models. I guess the xent models (whether LFR or non-LFR) are faster to train because the LF-MMI objective function is computationally more costly than xent objective function. I guess I didn't understand your point.

Also, I am running an LFR experiment on fisher_swbd to see if it outperforms regular models on large databases. I will share the results when ready.

danpovey:
Vimal's point is that when we train 'chain' models we do them on whole
chunks, not randomized frames, which is faster per frame trained-on
(although there is a tradeoff with how fast the SGD converges, which is why
we traditionally do frame-wise randomization whenever the objective
function and neural net topology allow it).

hhadian:
OK, thanks.
While double checking TDNN+LSTM script before including in the PR, I realized certain parameters (such as extra-left-context) were not correctly passed to decode.sh. I fixed and decoded again:
```
# System               tdnn_lstm_b tdnn_lstm_lfr1b
# WER on train_dev(tg)      14.95     15.01
# WER on train_dev(fg)      13.98     14.16
# WER on eval2000(tg)        17.2      17.1
# WER on eval2000(fg)        16.0      15.9
# Final train prob      -0.869086 -0.724979
# Final valid prob      -0.948473 -0.778266
```

danpovey:
Cool!
BTW, be careful when you upgrade to the latest "kaldi-5.1" type scripts,
i.e. with decay-time and setting --extra-left-context-initial=0 and
--extra-right-context-final=0, you should pass those both into train.py and
into decode.sh, in addition to the --left-context and --right-context.
Dan

hhadian:
I didn't know about these when running my first experiment against 'shortcut' branch (4 days ago). Now the results are ready (but without using decay and without setting extra-left-context-initial/final to 0):

```
# System               tdnn_lstm_b (against shortcut)
# WER on train_dev(tg)      14.25
# WER on train_dev(fg)      13.58
# WER on eval2000(tg)        16.7
# WER on eval2000(fg)        15.4
# Final train prob      -0.738717
# Final valid prob      -0.816738
```
note: this is not LFR.
Compared to the equivalent experiment against 'master' (cf. my previous comment), this has `0.4-0.5%` absolute improvement. What is that because of?

I will send a comparison of runtimes for different stuff soon (i.e. master vs shortcut).

danpovey:
The improvement could be just random noise.  Generally speaking I don't
really expect any change.  Also notice that the frames-per-chunk is now a
comma-separated list, not just a single number.  It may be easier if you
just take the current recommended script there (I think it's 1e, check the
soft link) and modify it.

hhadian:
OK will use 1e.

hhadian:
Runtimes for TDNN+LSTM on master vs shortcut:
```
Branch                  master     shortcut
probtrain-b10:          803.5      307.5
probtrain-a10:          1519.27    536.308
probvalid-b10:          831.538    320.6
probvalid-a10:          1565.39    532.533
train-b10:              651.083    630.25
train-b12:              381.556    336
average averaging time: 44.0632    46.3103

combine-time is 8967 on machine b07 against *master*
adjust-priors-time is 19 on machine b19 against *master*

combine-time is 9283 on machine b07 against *shortcut*
adjust-priors-time is 7 on machine a07 against *shortcut*
```

hhadian:
The results for the comparison @vijayaditya  suggested:
```
# System                   tdnn_f    tdnn_g tdnn_lfr1d tdnn_lfr1e
# WER on train_dev(tg)      17.12     16.19     17.54     16.43
# WER on train_dev(fg)      15.67     14.92     16.15     15.04
# WER on eval2000(tg)        19.5      19.0      19.9      18.9
# WER on eval2000(fg)        18.0      17.3      18.1      17.5
# Final train prob       -1.42253  -1.28323  -1.32467  -1.17596
# Final valid prob       -1.56084  -1.44347  -1.42538  -1.30078
```
`tdnn_f` and `tdnn_lfr1d` --> splicing: `-2,-1,0,1,2 -1,2 -3,3 -7,2 0 0`
`tdnn_g` and `tdnn_lfr1e` --> splicing: `-2,-1,0,1,2 -1,0,1 -1,0,1 -3,0,3 -3,0,3 -3,0 `
The average relative improvement for non-LFR is 4.2% while for LFR is 5.4%.

@danpovey, did you see my previous comment on runtimes? Are they as expected?

danpovey:
Yes, the differences in training time are as expected-- thanks.

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_lfr1b.sh
```diff
@@ -0,0 +1,163 @@
...
+# At this script level we don't support not running on GPU, as it would be painfully slow.
+# If you want to run without GPU you'd have to call train_tdnn.sh with --gpu false,
+# --num-threads 16 and --minibatch-size 128.
+
+# System                   tdnn_d  tdnn_lfr1a  tdnn_lfr1b
+# WER on train_dev(tg)      16.72     17.25     17.00
+# WER on train_dev(fg)      15.31     15.90     15.57
+# WER on eval2000(tg)        19.2      20.1      19.3
+# WER on eval2000(fg)        17.8      18.5      17.8
+# Final train prob       -1.22859  -1.32434  -1.11497
```
danpovey: you can create soft links like run_tdnn_lfr.sh to point to the current best setup.

src/hmm/hmm-utils.cc
```diff
@@ -772,18 +772,38 @@ static inline void ConvertAlignmentForPhone(
...
+
 /**
-   This function, called from ConvertAlignment(), works out suitable new lengths
-   of phones in the case where subsample_factor != 1.  The input vectors
+   This function, called from ConvertAlignmentInternal(), works out suitable new
+   lengths of phones in the case where subsample_factor != 1.  The input vectors
    'mapped_phones' and 'old_lengths' must be the same size-- the length of the
    phone sequence.  The 'topology' object and 'mapped_phones' are needed to
    work out the minimum length of each phone in the sequence.
    Returns true only if it could not assign lengths (because the topology was
```
danpovey: fix true->false
 [pre-existing error in the documentation]

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_lfr1b.sh
hhadian: Will do it now

hhadian:
> I notice there is nothing here about discriminative training of the low-frame-rate systems.
> However I'm OK to commit this first and deal with that later.

It will be better to push them in a separate commit, so I can incorporate the new version of shift_feats in the script too.

PR #1240, opened 2016-12-02
changed default value of zeroing-threshold in BackpropTruncationCompo…

freewym:
…nent to 15; updated the results on AMI


src/nnet3/nnet-general-component.cc
```diff
@@ -991,8 +991,8 @@ void BackpropTruncationComponent::Init(int32 dim,
 void BackpropTruncationComponent::InitFromConfig(ConfigLine *cfl) {
   int32 dim = 0;
   bool ok = cfl->GetValue("dim", &dim);
-  BaseFloat clipping_threshold = 15.0;
-  BaseFloat zeroing_threshold = 2.0;
+  BaseFloat clipping_threshold = 30.0;
+  BaseFloat zeroing_threshold = 15.0;
```
danpovey: Larger values of these quantities are more dangerous, i.e. more likely to lead to instability.
I don't think it's sufficient to just test this on one setup, because it's the potential for divergence that this is supposed to guard against.  Have you done any other tests?

danpovey:
Also, this PR would need to be on top of the 'fast_lstm' branch-- there are
other LSTM config-generation objects there that would have to be changed.
But I want you to test it in that setup.  And I'd be more comfortable with
smaller thresholds, like 5 and 15 or 5 and 20, instead of 20 and 30, if
there is no clear difference in results.  It's safer in situations where
divergence is a possibility.  The WER improvements you had in the RESULTS
file were rather unimpressive.

Dan

freewym:
This PR is already on top of fast_lstm.

The old WERs reported in RESULTS are obtained without zeroing (i.e. using ClipGradientComponent  as the comment said). The results of tuning zeroing-threshold on ihm are:
threshold  WER dev|eval
4.0              22.7|22.8
5.0              22.4|22.6
6.0              22.4|22.7
10.0            22.5|22.6
15.0            22.4|22.4 

I have not tune it on sdm1

After the fix of max-deriv-time, the gradient explosion did not happen even on the babel georgian multicondition data (which had the most severe problem before the fix): when I disabled the zeroing, the clipped-proportion is at most ~0.004.

I also tuned the zeroing-threshold on swbd blstm_6i:
threshold  WER fg|tg
5.0              14.5|15.8  
6.0              14.4|15.8
10.0            14.4|15.8
15.0            14.2|15.6

The reason I chose 15.0 as threshold rather than 5 or 10 is mainly based on the WER on swbd. Not sure how much variation there could be for different runs with the same settings

danpovey:
OK I will think about it.

freewym:
FYI, added the zeroed-proportion stats of the 1st layer at the last iteration, which shows how often zeroing was activated:

ami ihm
threshold   zeroed-prop
4.0              0.06
5.0              0.045
6.0              0.035
10.0            0.015
15.0            0.006

swbd
threshold         zeroed-prop
5.0                        0.018
6.0                        0.014
10.0                      0.003
15.0                      0.0007

danpovey:
OK, I'll merge this.

Issue #1235, opened 2016-11-30
Support for openfst 1.5.x


jtrmal:
It would be good if we could include support for the openfst-1.5.x
I had reports it does not work right away, but due time constraints this wasn't pursued further.
Can be related to #1234  also.
We should at least evaluate how hard it would be to include the support and what are the major hurdles in that.

danpovey:
@GaofengCheng, maybe you could look into this?

dogancan:
Hi @danpovey. I think I can help with this. Let me investigate what changes are needed to support openfst 1.5.

jtrmal:
Cool,  thanks a lot!
Y.

danpovey:
Thanks.  @GaofengCheng, don't bother with this.. .I know you have other stuff you're doing.

kkm000:
Past 1.5 as of #1342, done.

PR #1241, opened 2016-12-02
[for later, maybe]: Add support for OpenFst 1.5.4

dogancan:
Addresses #1235.

Successfully built and ran tests on Ubuntu 16.04.1 LTS and macOS Sierra 10.12.1 both with OpenFst-1.3.4 and OpenFst-1.5.4.

@danpovey @jtrmal This is ready for review.

### DIFF
--- a/src/bin/phones-to-prons.cc
+++ b/src/bin/phones-to-prons.cc
@@ -170,7 +170,7 @@ int main(int argc, char *argv[]) {
                    << "not reach end-state, or mismatched lexicon.)";
         if (g_kaldi_verbose_level >= 2) {
           KALDI_LOG << "phn2word FST is below:";
-#ifdef HAVE_OPENFST_GE_10400
+#if OPENFST_VER >= 10400
           fst::FstPrinter<StdArc> fstprinter(phn2word, NULL, NULL, NULL, false, true, "\t");
 #else
           fst::FstPrinter<StdArc> fstprinter(phn2word, NULL, NULL, NULL, false, true);
--- a/src/configure
+++ b/src/configure
@@ -960,20 +960,6 @@ if [ ! -f makefiles/common.mk ]; then
     failure makefiles/common.mk not found
 fi
 
-
-echo "Checking OpenFST library in $FSTROOT ..."
-if [ ! -f $FSTROOT/include/fst/fst.h  ]; then
-    failure "Could not find file $FSTROOT/include/fst/fst.h:
-    you may not have installed OpenFst.  See ../tools/INSTALL"
-fi
-echo Checking OpenFst library was patched.
-if ! grep "multiple repeated" $FSTROOT/include/fst/minimize.h >/dev/null; then
-    echo "**  ERROR  **"
-    echo "** $FSTROOT/include/fst/minimize.h seems not to be patched:"
-    echo "patch not applied?  FST tools will not work in our recipe."
-    exit 1;
-fi
-
 # back up the old one in case we modified it
 if [ -f kaldi.mk ]; then
   echo "Backing up kaldi.mk to kaldi.mk.bak"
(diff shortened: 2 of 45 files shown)


jtrmal:
I just skimmed through it -- looks good but I'll try to run one more, more careful, pass.
I have questions, though -- it seems that the Dan's patches are not needed anymore? I know some things were accepted upstream but I'm surprised that everything should work out of the box now :)

the other question is about the variables OPENFST_GE_10400 and OPENFST_GE_10500
Reading the patches,it seems that both variables will be defined for OpenFst 1.5.x?

jtrmal:
As the openfst site seems to be down again, I went on and created 1.5.4
mirror on openslr.org
y.

danpovey:
Yenda, he was defining the variable  OPENFST_GE_10500.
However, I think it might be easier to extend in future if we were to to put -DOPENFST_VER=1.5, and then have
```
#if  defined (OPENFST_VER) && OPENFST_VER >= 1.5
```
(but there may be something wrong about this).
BTW, Dogan, another outstanding question is whether to upgrade our threading to use C++11 threading instead of pthreads (pending comments from users about whether C++11 is a problem).  We might want your help with that too.

danpovey:
... Actually, if we're really going to do this right, it would probably be better to have the configure script create a file base/config.h that contains all the defines.  That would make it easier for people who want to use Kaldi as a library-- they won't have to worry about getting the flags correct.  But I don't know if we want to mix this up with this issue... depends how much energy Dogan has.

jtrmal:
Yes, but if I'm looking at the configure changeset right both OPENFST_GE_10500
and OPENFST_GE_10400 will get defined. And yes, I was asking because if I
did read it correctly, it might be a more tidy approach to have only
single OPENFST_VER
variable (set to an appropriate value).

dogancan:
You are right that both OPENFST_GE_10400 and OPENFST_GE_10500 will be defined for openfst-1.5.4 since 1.5.4 is larger than both 1.4.0 and 1.5.0. I agree that it would be cleaner to have a single OPENFST_VER flag that we compare in preprocessor conditionals such as OPENFST_VER > 10400. Let me do that change now and we can continue this discussion after.

dogancan:
I just pushed some changes. We now define a single integer OPENFST_VER and work out which code should be compiled based on that. I made the conditionals such that all code that needs to be compiled against OpenFst < 1.4 (default case) are inside else clauses. This way we do something reasonable even if OPENFST_VER is not defined, that is we assume we are compiling against OpenFst < 1.4.

danpovey:
Thanks.
There may be other code changes required-- IIRC the new minimization
doesn't naturally accept non-deterministic inputs, you have to give some
kind of flag.  This may require options to fstiminimize in the graph
building (assuming the flags are exposed to the command line), else there
would be a runtime crash.
I'll try to check it out and figure out specifically what the issue is.
Dan

dogancan:
You are right. We need to call fstminimize with "--allow_nondet". It's unfortunate that this flag defaults to false in openfst-1.5.4. Patching openfst to flip the default value seems to be the easiest solution. I can add that patch to the pull request if you like.

dogancan:
By the way, let me know if you need help with updating the threading to C++11. I will do my best to help out. Cheers..

danpovey:
Regarding allow_nondet... yes; they wanted it to be back compatible to the
old value.  But I just searched... it looks like we never call fstminimize
from scripts (we call our own binary fstminimizeencoded).  But calls from
code will need to be patched.
RE threads-- thanks, I will ask, soon.

jtrmal:
I'd be in favor of being able to use a stock openfst,  if it's possible,
 but I wonder if this would be causing problems for people downstream (in
companies) - in their proprietary codes.
Y.

danpovey:
@dogancan, when you get a chance, search for calls to Minimize() in the code and figure out the best way to pass the flag.  [only necessary if it doesn't directly follow Determinize().]  Creating a wrapper might be a way to do it (you could put it in fstext-utils.h), to avoid cluttering the code with ifdefs.  E.g. MinimizeNondet().

dogancan:
@danpovey, from what I can tell, there is just [one call](https://github.com/kaldi-asr/kaldi/blob/master/src/kwsbin/kws-search.cc#L202) to Minimize() that doesn't follow Determinize() or DeterminizeStar(). I don't think that one call needs to be modified since its input should be deterministic anyway. In fact I think we can remove that Minimize() call altogether. There is no need for minimization at that point in the search algorithm.

danpovey:
Looks like there are quite a few calls to AcceptorMinimize.

dogancan:
I see 3 calls to AcceptorMinimize() but I don't think we need to change them. It seems the new minimization code in openfst checks this allow_nondet flag only inside the Minimize() call.

danpovey:
Are you sure the minimize is not needed in the kws code?  Might it be
easier to just patch it than to test it?

dogancan:
I might be wrong if the index structure has changed since I last looked at it. I agree it's easier to patch than test, so I will just do that.

danpovey:
Thanks!
I won't merge this right away, some testing needs to be done and I'll think about the best way to do this.
I'll create a separate issue RE threading.

danpovey:
@dogancan, just to clarify: we're doing the upgrade of OpenFst as part of your C++11 PR, right?
Also, something we need to decide is whether to have two separate version-bumps for your C++11 PR and my 'shortcut-etc.'  PR, or to combine them into one version-bump and test them together.

dogancan:
Yes C++11 PR includes the OpenFst 1.6 upgrade. This PR adds support for OpenFst 1.5.4 but does not require C++11 support when building against OpenFst < 1.4 . I kept it open in case we would like to merge this to 5.0 branch after C++11 upgrade hits master.

dogancan:
@danpovey I was going to rebase this against the 5.0 branch but it seems like we don't have that branch yet.

danpovey:
I thought I had created that, but it seems I overlooked it.  Done now.

nshmyrev:
@jtrmal This can be closed now since it has been updated long time ago.

PR #1245, opened 2016-12-04
Create db dirrectory if it doesn't exists in TEDLIUM/s5 download script

akegalj:


vince62s:
We strongly suggest you switch to s5_r2 which gives better results.
but good point.

akegalj:
Yes, I just wanted to compare how models compare on my data set (a bit specific domain)

Thanks

Issue #1234, opened 2016-11-30
The upgrade to C++11


danpovey:
This is an issue for discussion.
Should we allow the use of C++11 features in Kaldi?  It seems like C++11 is old enough now that complier support should be universal.
The most obvious benefit is that we could use C++11 threads, which are easier to use than pthreads, and would make it easier to support Visual Studio (assuming their support for C++11 works OK).
I'm not saying we'd necessarily go all-in on things like "auto" and lambdas.
Also this change would make it more straightforward to keep up-to-date with the latest OpenFst.  It might even be possible to use an un-patched version of OpenFst, since the changes to minimization that we previously patched have now been accepted upstream.  [However, there is a change regarding threading, Incr() and Decr() that I think might still need to be patched, see openfst_gcc41up.patch.]

Obviously it would take a while to upgrade; we'd have to maintain a separate branch for a while.

vdp:
FWIW, I think it should be safe to migrate to C++11. For example Google's, TensorFlow which is obviously an important piece of their strategy of waging the Cloud Wars, and supports a bunch of platforms including (since recently) Windows. Unsurprisingly their code is very well written style-wise, IMHO, and could be used as a guide for Kaldi's potential C++11 transition. I don't know why you don't like "auto"- it will remove much of the line-noise that are the iterator definitions in older C++ versions.

Just my 2c- obviously the opinion of the people who are actively contributing at the moment should be given much more weight than mine.

dogancan:
I am on the same boat with @vdp. Even if we don't transition all of existing Kaldi codebase, a compiler with C++11 support is a perfectly reasonable requirement IMHO. We are currently going through hoops to support a number of OpenFst versions. If we require C++11, not only can we remove the support for old OpenFst versions, but also make things significantly easier for people who want to maintain Kaldi packages for Linux.

danpovey:
@dogancan, it looks like there is support for allowing Kaldi to require C++11.  So I think we'll go ahead with this.  It would be great if you could help, as I know you're much more familiar with C++11 than I am.

[...]

PR #1249, opened 2016-12-06
WIP: C++11 Upgrade

dogancan:
This PR is a work in progress. See #1234 #1235 for earlier discussion.


jtrmal:
yes, I think that would be a good place to start.
y.

dogancan:
@jtrmal, @danpovey: What packages should we recommend if the system default g++ does not have support for -std=c++11? Should we just print a warning and punt?

dogancan:
@jtrmal, @danpovey : I am planning to rework the Kaldi build system so that environment variables like CXX and CXXFLAGS are handled by src/configure rather than src/Makefile. I mean the Makefile will still respect these environment variables, just that the standard (or recommended) way of overriding these variables will be during the configure call. This way we will be able to tell what compiler and compiler flags people are intending to use during the configure call and act accordingly. Any objections?

danpovey:
Regarding handling CXX and CXXFLAGS by src/configure: I agree that this
makes sense.  We made the Makefile respect them because a user specifically
wanted this for some automatic-build scenario, but this shouldn't be the
normal way to specify them.

Regarding what to do when the g++ version is too old: I'd probably print an
error message and have check_depenencies.sh exit with nonzero status.
AFAIK it is usually hard to install a newer version of g++ than the system
default, and not screw up the system; and how would we know what version to
recommend?  Perhaps we could say in the error message that g++ supports
C++11 from such-and-such a version.

Dan

danpovey:
@dogancan, you're working on this, right?  No hurry.. just making sure it will get finished at some point.

dogancan:
Yes, I am still working on reorganizing and cleaning up the build system. I made extensive changes to the configure script and refactored makefiles/*.mk so that most of the build settings are shared across platforms. I'll probably push those changes within the next few days to get your feedback. I guess I'll start with the threading changes once these are out of the way.

danpovey:
OK.  Just FYI I rejected refactoring of the makefiles in the past due to
concerns about testability, but since you've done things carefully in the
past I would be inclined to accept it if you say it is safe.

dogancan:
@jrziviani Could you test the ppc64le build with this branch and see if config.guess still needs to be patched? The config.guess file distributed with OpenFst-1.5.4 is newer than the ones distributed with OpenFst < 1.5 but still older than the one we are downloading for ppc64le. As it is, this branch still downloads and patches config.guess on ppc64le. If the config.guess distributed with OpenFst-1.5.4 is good enough to build on ppc64le, we would like to remove the patch.  Note that you will need a relatively recent compiler with C++11 support (e.g. g++ >= 4.7) to build this branch.

dogancan:
@danpovey, @jtrmal I think we should do threading etc. changes with a separate pull request and limit this one to the overhaul of codebase and the build system. What do you think? If you agree with me on this, I think this pull request is more or less ready for review. I rebased the c++11 branch against the current master and synchronized the ppc64le specific makefiles with the rest of the build system. Since this pull request touches a large number of files and makes extensive changes to the configuration/build system, I think we should be cautious and make sure it does not cause any regressions on major platforms. I have tested things on Ubuntu 12.04 (Travis), Ubuntu 16.04 and macOS 10.12. It's probably a good idea to ask other people with unusual setups to try building against this branch before we pull the trigger.

@danpovey I reverted most of the changes to makefiles/*.mk since factoring out platform independent configuration flags makes kaldi.mk more complicated, hence harder to edit manually. Remaining changes are largely cosmetic.

The failing travis builds are due to the build timing out before tests are complete. From what I can tell the travis build takes approximately 15 minutes longer when we require the C++11 standard. I am not sure what is causing the slowdown but I would guess it has to do with the specific compiler or standard library building with. They might have resolved these regressions in recent versions of g++.

danpovey:
OK thanks.
Will try to look over this in the next couple of days.

src/fstext/table-matcher.h
```diff
@@ -301,14 +286,15 @@ class TableMatcher : public MatcherBase<typename F::Arc> {
...
 
   virtual void SetState_(StateId s) { impl_->SetState(s); }
   virtual bool Find_(Label label) { return impl_->Find(label); }
   virtual bool Done_() const { return impl_->Done(); }
   virtual const Arc& Value_() const { return impl_->Value(); }
   virtual void Next_() { impl_->Next(); }
-  DISALLOW_COPY_AND_ASSIGN(TableMatcher);
+
+  TableMatcher &operator=(const TableMatcher &) = delete;
 };
```
jtrmal: I guess it's OK this way but if I had been doing this, I'd change the DISALLOW_COPY_AND_ASSIGN

src/fstext/trivial-factor-weight.h
```diff
@@ -52,17 +52,8 @@
 // This has the advantage that it always works, for any input (also I just
 // prefer this approach).
 
-#ifdef _MSC_VER
 #include <unordered_map>
 using std::unordered_map;
```
jtrmal: there is a header ./util/stl-utils.h that would do the same.  Not sure what would be the prefered solution, though.

src/fstext/trivial-factor-weight.h
```diff
@@ -307,19 +292,18 @@ class TrivialFactorWeightFstImpl
 
   typedef unordered_map<Element, StateId, ElementKey, ElementEqual> ElementMap;
 
-  const Fst<A> *fst_;
+  std::unique_ptr<const Fst<A>> fst_;
   float delta_;
```
jtrmal: I see bellow this that you initialize unique_ptr using make_shared. Is that a correct way how to handle this? I know that make_unique exists only since C++14, but I'm not sure about interactions between assingments of unique_ptr and shared_ptr

src/fstext/table-matcher.h
dogancan: Thanks for the comments. I tried to follow the latest OpenFst conventions for stuff in fstext since along with this upgrade we will require OpenFst >= 1.5.3. From what I can tell, OpenFst largely stopped using this macro along with the upgrade to smart pointers.

src/fstext/trivial-factor-weight.h
dogancan: I am guessing there are other places in the code base with similar include guards which are no longer necessary since unordered map is in std namespace in all c++11 standard libraries. I'll update utils/stl-utils.h.

src/fstext/trivial-factor-weight.h
dogancan: Here make_shared initializes the underlying fst implementation not the unique pointer kept by that implementation.

danpovey:
We're going to have to figure out how to test this adequately.
One option is to create a c++11 branch in kaldi's official repository and ask people to merge with that version or check out that version, to test it, before we switch over the master to point to that.

I'm actually considering starting to use meaningful version numbers for Kaldi, and wonder whether this is a good time to start.  But I'll write something on the other PR about that.
We should go ahead trying to test this PR, regardless.

jrziviani:
@dogancan Hi! yes, openfst 1.5.4 looks good. I've got your branch and removed that config.guess patch for ppc. It compiles flawlessly.

Do you want me to send a patch reverting that change or you prefer to include it in your set?

Thank you!

dogancan:
@jrziviani Thanks. You don't need to do anything. I will take care of the rest.

danpovey:
I have asked @LvHang to test on a few different platforms, both from clean install and upgrade from master.

dogancan:
OpenFst-1.6.0 has just been released. I started with updating this branch to support 1.6.0. From a cursory look at compiler errors, I noticed that they removed the `DISALLOW_COPY_AND_ASSIGN` macro and introduced a new namespace `fst::internal`. It seems like supporting 1.5 along with 1.6 will require checking openfst version in a bunch of places in the codebase via preprocessor conditionals. Do we care whether we support OpenFst <1.6.0 moving forward?

danpovey:
I'm OK to just make the jump to 1.6.0 without going via 1.5.

dogancan:
I made the changes needed for OpenFst-1.6 support. I also removed dynamic libraries from the prerequisite list used for binaries. Addresses the problem of relinking all binaries whenever kaldi-base dynamic library is updated due to a version change. From now on dynamically linked binaries/libraries are relinked only when they are recompiled due to changes in the source files they depend on -- not when dynamic libraries they depend on are updated. No change to current behavior when building with static libraries.

dogancan:
Copying this from above to bring it up.

>The failing travis builds are due to the build timing out before tests are complete. From what I can tell the travis build takes approximately 15 minutes longer when we require the C++11 standard. I am not sure what is causing the slowdown but I would guess it has to do with the specific compiler or standard library we are building with. They might have resolved these regressions in recent versions of g++.

jtrmal:
@kkm000 Any ideas where to look at? Is it some limitation in the free travis service? Or is there some directive that specifies the time needed?  I can spend some time on it as I assume you are busy, but any starting hints would be great :)

danpovey:
@dogancan, it looks like some conflicts have accumulated.
Have you managed to get it to build fast enough for travis now?

dogancan:
The only change that helped was to increase the number of concurrent make jobs to 4 (from 3). It currently takes around 40 mins to build. Travis limit is 50 mins so it is building for now. If we change the travis configuration to use clang++, we will probably go down to 30 mins or so. This change requires changing the setup on travis.org though so I can't test it myself. I'll take care of the conflicts in a bit.

danpovey:
ok thanks.
I'm about to push some changes to make some of the slower tests faster,
should shave off a couple minutes.

danpovey:
closing this PR as it was merged to 'shortcut' branch.

PR #1251, opened 2016-12-06
minor bug fix in XconfigAffineLayer when initializing param-stddev

freewym:


danpovey:
Hm.  That was definitely a bug.  But before I merge this, can you please test one of the TDNN+LSTM setups to see whether the performance is affected?  It's possible that this somehow improves results by making it train faster at the beginning.  [It would be catastrophic if we didn't have the renormalize layers.]

freewym:
param-stddev in tdnn/lstm layer are correctly set I think, by not specifying it in xconfig, the c++ code would set it as 1/sqrt(input_dim). XconfigAffineLayer is not used in our current tdnn or lstm recipes. @vijayaditya please correct me if I was wrong.

danpovey:
Yes... I just wanted you to run the LSTM+TDNN setup because it does contain
regular 'basic' layers which would be affected by this bug, and it's our
current recommended/best setup.

freewym:
In tdnn+lstm setups we use XconfigBasicLayer, which is independent of XconfigAffineLayer. The bug was only in XconfigAffineLayer.

danpovey:
Oh.  Are we using XconfigAffineLayer in any recipes?

freewym:
I don't think so.

danpovey:
OK, merging then.

vijayaditya:
We are not using XconfigAffineLayer in any committed recipe (testing it on
newer architectures). It had been committed to the master for the sake of
completeness in the layer library.

--Vijay

PR #1256, opened 2016-12-08
tensorflow acoustic model training under kaldi environment

GaofengCheng:
this version support: 
1. one gpu card DNN training 
2. reading feature directly from kaldi (compress false) 
3. based on wsj , if you want to use on another dataset, you have to do some work

further things need to do:
1. better baseline
2. multi cards, distributed, better run on cluster
3. more model support


vijayaditya:
Thanks. I will review this over the next few days.


Vijay

danpovey:
Why was that directory called neuralNetworks?  is that a name required or
suggested by TensorFlow itself?  It goes against the normal directory
naming conventions we use.

vdp:
BTW what exactly is the purpose and the scope of this PR? Training and decoding w/ TF-trained acoustic models?

danpovey:
I think so... I haven't followed the details.

GaofengCheng:
@vdp yes, extending kaldi nnet with external library

vince62s:
@GaofengCheng at decoding time did you try with gpu or cpu ? if you tried on cpu only, was it reasonably fast ? I am asking because I did the rnnlm rescoring job on TF but, without gpu decoding, is too slow for now.

vijayaditya:
Yes, that is the final goal of this effort.

--Vijay

egs/wsj/s5/conf/wsj_tensorflow_kaldi.conf
```diff
@@ -0,0 +1,80 @@
+# Copyright 2016    Vincent Renkens
```
danpovey: need licence type (OK to say just Apache 2.0.)
I'd be happier if this file was in some subdirectory of local/, as you may need more versions of this file as you tune things.  The conf/ directory is more for "global" configuration files, not experiment-specific ones.

egs/wsj/s5/steps/decode_tensorflow.sh
```diff
@@ -0,0 +1,82 @@
...
+minimize=false
+# End configuration section.
+
+echo "$0 $@"  # Print the command line for logging
+
+[ -f ./path.sh ] && . ./path.sh; # source the path.
+. parse_options.sh || exit 1;
+
+if [ $# -ne 3 ]; then
+  echo "Usage: $0 [options] <graph-dir> <data-dir> <decode-dir>"
```
danpovey: this `<data-dir>` doesn't seem to be anything like the normal data-dir... seems to contain loglikes.  You haven't reworked the usage message of this script to explain what it is or to rename it.

egs/wsj/s5/steps/libs/tensorflow/kaldi_tensorflow/kaldiInterface.py
```diff
@@ -0,0 +1,85 @@
...
+	feats_shuffledfile = open(featdir + '/feats_shuffled.scp', 'w')
+	feats_shuffledfile.writelines(feats)
+
+##read the alignment file generated by kaldi
+#
+#@param filename path to alignment file
+#
+#@return a dictionary containing the alignments with the utterance IDs as keys      
+def read_alignments(filename):
+	print("alifile_name: "+str(filename))
```
danpovey: this file seems to have an indent-level of 8, not 4 as we use for Kaldi code (and I think as required in google style). Make sure you are using spaces not tabs.  I mentioned in another PR what your emacs config should look like.... for vim setup, ask Vimal.

egs/wsj/s5/steps/libs/tensorflow/tensorflow_nnet/nnetactivations.py
```diff
@@ -0,0 +1,133 @@
+# Copyright 2016    Vincent Renkens
```
danpovey: you'll have to check into the copyrights for these files... hopefully it's something compatible with Apache 2.0, or MIT license.  If not, he may agree to release under Apache 2.0.

egs/wsj/s5/tensor_2048_relu.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2016    Vincent Renkens
```
danpovey: Please don't put files in the top-level experimental directory.
They should be in local/tensorflow/ or something like that.

danpovey: Also, is this the top-level example script?  I'm surprised there isn't a top-level shell script, that would make sure everything is in place... e.g. I saw something about needing un-compressed features.  Did you have to modify any of the checked-in shell scripts or run anything separate, to ensure the features on disk were un-compressed?

danpovey:
@GaofengCheng, any updates on this?  Do you have any results, and how do they compare with Kaldi-based nnet training?

egs/wsj/s5/steps/libs/tensorflow/kaldi_tensorflow/kaldiInterface.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2016    Vincent Renkens
```
danpovey: please use underscores in names, e.g. kaldi_interface.py.  But I'm not crazy about this name, because it could be confused with some kind of generic python to C++ interface.  Maybe examples.py or example_utils.py or something like that?

GaofengCheng:
@danpovey  Povey, I'm working on this and got comparable results compared with kaldi acoustic (relu DNN),  after some fuctions added( opts to choose whether applying cmvn, learning rate store etc...) and fix your comment issues, I'll update, sorry for delay.....

danpovey:
Interesting.  When you say "relu DNN", are you talking about TDNNs, or an
older recipe.
BTW, the reason we don't normally enable CMN is so that the models can be
used for online decoding.  It does give an improvement even with
Kaldi-based training.

GaofengCheng:
@danpovey just fully connected NN, here is the results (all based on WSJ):
image
But I'm not sure about one thing, the feature I use is `train_si284_hires`, I find the kaldi training do not CMN, because for online decoding ? Someone told me CMN is done during `run_ivector_common.sh`, so there is no need to CMN again.

danpovey:
OK.
I think that is not the best Kaldi result, the best one is actually a TDNN
```
for x in exp/nnet3/nnet_tdnn_a/decode_*; do grep WER $x/wer_* |
utils/best_wer.sh ; done
%WER 7.19 [ 592 / 8234, 51 ins, 109 del, 432 sub ]
exp/nnet3/nnet_tdnn_a/decode_bd_tgpr_dev93/wer_13_0.5
%WER 3.93 [ 222 / 5643, 23 ins, 20 del, 179 sub ]
exp/nnet3/nnet_tdnn_a/decode_bd_tgpr_eval92/wer_10_1.0
%WER 9.78 [ 805 / 8234, 167 ins, 72 del, 566 sub ]
exp/nnet3/nnet_tdnn_a/decode_tgpr_dev93/wer_10_0.0
%WER 6.40 [ 361 / 5643, 87 ins, 16 del, 258 sub ]
exp/nnet3/nnet_tdnn_a/decode_tgpr_eval92/wer_10_1.0
```
We don't keep the WSJ recipe very updated as there is not much data.
No, the CMN stats are computed in the ivector stuff but they are not
actually used.  See this in local/nnet3/run_tdnn.sh:
```
    --cmvn-opts "--norm-means=false --norm-vars=false" \
```
Of course the feedforward DNN is probably far from the best Tensorflow
model anyway, but it's a good start.  I don't think you have to do too much
tuning-- there is value in having just a basic setup checked in.

GaofengCheng:
@danpovey  Povey, hope to get your suggestions about my next stage work on this PR (adding more models like LSTM or ....? ) ?

danpovey:
That would be good but I wonder whether we could aim to commit something
first (-> focus on cleaning it up).
I am also hoping that @vijayaditya may have time to look over it and give
some input.
How long does the TensorFlow training take, and how does it compare with
the Kaldi baseline?

GaofengCheng:
@danpovey Tensorflow (with batch norm, learning rate auto adjsut) takes less epochs to converge on my setup, but speed for each epoch is no obvious faster than kaldi
image
wait for Vijay's suggestion

danpovey:
Are you saying that the time per epoch is the same as Kaldi even though the
Kaldi one is using multiple GPUs?

GaofengCheng:
sorry for confusion povey, one gpu means one gpu we see using `nvidia-smi`(one gpu core, not card....).... when I compare speed between kaldi and tensor, I calculate each epoch time for tensorflow roughly by `total_training_time/(epoch_num)` ; time for Kaldi roughly by `average_gpu_num*total_training_time/(epoch_num)` ,`average_gpu_num=(ini_job+final_job)/2`

jtrmal:
How long does kaldi epoch takes vs  tensorflow epoch takes? Or how long it
takes? I'm curious about wall-clock time, not the accumulated CPU/GPU time.
Thanks,
y.

GaofengCheng:
@jtrmal Does `wall-clock time` means total training time? If so, on WSJ, 1024 6layer relu DNN, the tensor training with one gpu core takes about 12.5 hours(6 epoch). IIRC, kaldi takes  1hours and 20 minutes (init2, final14) for 8 epochs.
I'm trying to use tensorlayer (https://github.com/zsdonghao/tensorlayer) to replace the neural network in this PR. If done, neural network will be enriched.

jtrmal:
yes, that was the info I was looking for. thanks a lot!
y.

francisr:
Is this still under development?

danpovey:
It's not being developed with the highest priority, but I think it's not
completely abandoned.  If you check out his PR on this and try to get it to
work, maybe you can help with this.

GaofengCheng:
@francisr As I have many things to do (English exam and other affairs), the dev will be very slow for the recent several months . It'll be better if you can help, the PR could run directly under the WSJ dir.

PR #1269, opened 2016-12-15
Various nnet3 changes [shortcut compilation; looped computation; LSTM changes; variable chunk-size,minibatch-size]

danpovey:
This pull request is for some changes in the core nnet3 compilation process that should substantially speed up the compilation process for computation requests with certain types of regular structure (in practice, this covers all computation requests we currently use for training).

I am doing this on top of the looped-computation PR; this is to avoid merge conflicts, because some of the same parts of the code are touched (although the looped-computation PR covers a separate issue).

I'm not sure whether to merge this as-is (after testing), or to wait until I implement some changes to the nnet3 "egs" creation and merging process.  The plan is to not have the extra left and right context for recurrent nnet training at the begin/end of files, and also to allow more variation in chunk length for chain training.  Before the changes made in this PR, this would have led to unacceptably high compilation times, but now it should be fast enough.

@GaofengCheng, is there any chance you could use this PR for one of your model-training runs, and see if it improves speed, and check that there is no substantial change in performance?  [It should be exactly the same, but of course it's not deterministic.]


danpovey:
Closing this PR and will open another one from the branch in the official Kaldi repo.

PR #1278, opened 2016-12-19
WIP: Learning scale of parameters

freewym:
Related to issue #1274. Created for discussion.

### DIFF
--- a/egs/swbd/s5c/local/chain/tuning/run_blstm_6j.sh
+++ b/egs/swbd/s5c/local/chain/tuning/run_blstm_6j.sh
@@ -185,6 +185,7 @@ if [ $stage -le 13 ]; then
     --trainer.optimization.initial-effective-lrate 0.001 \
     --trainer.optimization.final-effective-lrate 0.0001 \
     --trainer.optimization.momentum 0.0 \
+    --trainer.optimization.scale-learning-rate 0.1 \
     --trainer.deriv-truncate-margin 8 \
     --egs.stage $get_egs_stage \
     --egs.opts "--frames-overlap-per-eg 0" \
--- a/egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/common.py
@@ -84,12 +84,6 @@ def train_new_models(dir, iter, srand, num_jobs,
             cache_write_opt = "--write-cache={dir}/cache.{iter}".format(
                 dir=dir, iter=iter+1)
 
-        minibatch_opts = "--minibatch-size={0}".format(minibatch_size)
-
-        if chunk_level_training:
-            minibatch_opts = "{0} --measure-output-frames=false".format(
-                minibatch_size)
-
         process_handle = common_lib.run_job(
             """{command} {train_queue_opt} {dir}/log/train.{iter}.{job}.log \
                     nnet3-train {parallel_train_opts} {cache_read_opt} \
@@ -101,7 +95,8 @@ def train_new_models(dir, iter, srand, num_jobs,
             """ark:{egs_dir}/egs.{archive_index}.ark ark:- |"""
             """nnet3-shuffle-egs --buffer-size={shuffle_buffer_size} """
             """--srand={srand} ark:- ark:- | """
-            """nnet3-merge-egs {minibatch_opts} """
+            """nnet3-merge-egs --minibatch-size={minibatch_size} """
+            """--measure-output-frames=false """
             """--discard-partial-minibatches=true ark:- ark:- |" \
                     {dir}/{next_iter}.{job}.raw""".format(
                         command=run_opts.command,
@@ -115,12 +110,12 @@ def train_new_models(dir, iter, srand, num_jobs,
                         frame_opts=(""
                                     if chunk_level_training
                                     else "--frame={0}".format(frame)),
-                        minibatch_opts=minibatch_opts,
                         momentum=momentum, max_param_change=max_param_change,
                         deriv_time_opts=" ".join(deriv_time_opts),
                         raw_model=raw_model_string, context_opts=context_opts,
                         egs_dir=egs_dir, archive_index=archive_index,
-                        shuffle_buffer_size=shuffle_buffer_size), wait=False)
+                        shuffle_buffer_size=shuffle_buffer_size,
+                        minibatch_size=minibatch_size), wait=False)
 
         processes.append(process_handle)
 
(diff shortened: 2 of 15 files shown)


egs/swbd/s5c/local/chain/tuning/run_blstm_6j.sh
```diff
@@ -185,6 +185,7 @@ if [ $stage -le 13 ]; then
     --trainer.optimization.initial-effective-lrate 0.001 \
     --trainer.optimization.final-effective-lrate 0.0001 \
     --trainer.optimization.momentum 0.0 \
+    --trainer.optimization.scale-learning-rate 0.1 \
```
danpovey: in general with these tuning scripts I prefer you to add a new script rather than edit the existing ones.. then you can document what was changed and what the results were.
But of course, if this idea doesn't work, it won't matter.

egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/common.py
```diff
@@ -111,6 +113,8 @@ def train_new_models(dir, iter, srand, num_jobs,
                                     if chunk_level_training
                                     else "--frame={0}".format(frame)),
                         momentum=momentum, max_param_change=max_param_change,
+                        scale_learning_rate=scale_learning_rate,
+                        scale_update_frequency=scale_update_frequency,
```
danpovey: I don't think it's necessary to have this frequency passed through the python scripts, we can just set a reasonable value at the code level.  but this is OK for now; it might not even work.

src/nnet3/nnet-chain-training.cc
```diff
@@ -147,6 +149,44 @@ void NnetChainTrainer::ProcessOutputs(const NnetChainExample &eg,
...
+    if (comp->Properties() & kUpdatableComponent) {
+      UpdatableComponent *uc_delta = dynamic_cast<UpdatableComponent*>(comp);
+      UpdatableComponent *uc_nnet =
+        dynamic_cast<UpdatableComponent*>(nnet_->GetComponent(c));
+      if (uc_delta == NULL || uc_nnet == NULL)
+        KALDI_ERR << "Updatable component does not inherit from class "
+                  << "UpdatableComponent; change this code.";
+      // computes the dot product of the parameter in nnet_ with the parameter
+      // in delta_nnet_, which is the derivatives w.r.t. the scaling factor 
+      BaseFloat dot_prod =uc_delta->DotProduct(*uc_nnet);
```
danpovey: space after =.

src/nnet3/nnet-chain-training.cc
```diff
@@ -147,6 +149,44 @@ void NnetChainTrainer::ProcessOutputs(const NnetChainExample &eg,
...
+      // adds the param changes due to the change of parameter's scaling factor,
+      // to nnet_delta_.
+      BaseFloat alpha = scale_learning_rate * dot_prod;
+      uc_delta->Add(alpha, *uc_nnet);
```
danpovey: You'll have to be a bit careful here about the interaction with momentum.  This formula would not work correctly if momentum != 0.0; it would diverge, because you're adding to delta_nnet and next time you work out the scaling factor from delta_nnet.  One way to fix this would be to remember the 'alpha' values and subtract them again after adding delta_nnet to nnet, after calling  AddNnetComponents().

LvHang:
@danpovey 


1. In line 154, the first bug is change && to ||. Otherwise, the scale_learning_rate is nnet_config.scale_update_frequency times than we expected.

2. In line 164, we remove the factor(1.0-nnet_config.momentum). [Without considering the learning-rate]
2.1  If the momentum=0.0, the derivation of W[t] = delta_nnet_
2.2 [As your annotation, you try to compute the sum of geometric progression, so I think you assume the deri(W[t]) is the same in each time]
If the momentum!=0.0,  delta_nnet_(t)= (1-momentum)*deri(W[t]) + momentum * delta_nnet_(t-1)=(1-momentum)* deri(W[t])*(1+momentum+momentum^2+....) = deri(W[t])
In conclusion, the delta_nnet_ is approximately equal to the derivation of W[t]. So we don't need the factor (1.0-nnet_config.momentum).

3. I think dan's suggestion makes sense. Each time, the delta_nnet_ is updated. It means delta_nnet_ = delta_nnet_ + alpha * nnet_. If the momentum is not 0.0, it will be accumulated to next iteration. With the iteration raise, the accumulation will be large. We need minus the item alpha * nnet_ from delta_nnet_ after update the Weight matrix.
However, in chain model, the momentum is 0.0, but the result is not ideal. 

4.[report_after_fix_bug.pdf](https://github.com/kaldi-asr/kaldi/files/707236/report_after_fix_bug.pdf)
This is the log-probability curve after we fix two bug.

Best wishes,
Hang

danpovey:
Can you show me the directories?
I want to see if the model parameters seem to be getting larger or smaller.

LvHang:
The directory of report is /export/a11/hlyu/yiming-test/egs/swbd/s5c/report
The directories of models are /export/a11/hlyu/yiming-test/egs/swbd/s5c/exp/chain

danpovey:
BTW, what command do you use to generate that report?

LvHang:
steps/nnet3/report/generate_plots.py --is-chain true --start-iter 25 --comparison-dir exp/chain/blstm_6j_1_sp --comparison-dir exp/chain/blstm_6j_dot1_sp --comparison-dir exp/chain/blstm_6j_dot01_sp exp/chain_old/blstm_6j_scalezero_sp ./report

LvHang:
This is the newest WER result about learning a scale for each component on
AMI corpus:

Model Setup

dev

eval

momentum = 0.0; shrinkage = 1; scale-learning-rate = 0

39.0

42.7

momentum = 0.0; shrinkage = 0.99; scale-learning-rate = 0

38.7

42.1

momentum = 0.0; shrinkage = 1; scale-learning-rate = 0.01

41.0

44.4

momentum = 0.0; shrinkage = 0.99; scale-learning-rate = 0.01

40.2

43.6

The following figure shows the convergent tendency:


Apparently, the shrinkage is more useful. I and Yiming guess the
‘scale-learning-rate’ maybe a little big. Now, we scale down ‘
scale-learning-rate’ to 0.001/0.0001/0.0001 and keep momentum = 0.0;
shrinkage = 1. We hope the new setups can match the result of ‘momentum =
0.0; shrinkage = 0.99; scale-learning-rate = 0’ which means only using
shrinkage.

2017-01-16 1:00 GMT-05:00 Daniel Povey <notifications@github.com>:

> BTW, what command do you use to generate that report?
>

danpovey:
OK, but it's not looking very promising.  I'm prepared to just abandon the
idea.

danpovey:
closing PR as this does not seem to work.

PR #1279, opened 2016-12-20
My rnnlm

hainan-xv:
Not a ready PR yet; just to make it easier to see the changes


src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,145 @@
...
+    vector<std::pair<int32, BaseFloat> > p;
+    p.push_back(std::make_pair(word_ids_out[i], 1.0));
+    posterior.push_back(p);
+  }
+
+  eg.io.push_back(nnet3::NnetIo("output", output_dim, 0, posterior));
+  return eg;
+}
+
+vector<int> Select(const vector<BaseFloat> &u, int n) {
```
danpovey: ... also, this kind of sampling code needs to be tested.  I think we discussed testing strategies by email.. or maybe it was just with Ke.
Also, the name is not good... should contain the word Sample at least, not Select.

src/rnnlmbin/rnnlm-copy.cc
```diff
@@ -0,0 +1,81 @@
...
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    using namespace kaldi::nnet3;
+    using namespace kaldi::rnnlm;
+    typedef kaldi::int32 int32;
+
+    const char *usage =
+        "Copy 'raw' nnet3 neural network to standard output\n"
```
danpovey: update usage message

src/rnnlmbin/rnnlm-copy.cc
```diff
@@ -0,0 +1,81 @@
...
+                "are set to this value.");
+
+    po.Read(argc, argv);
+    
+    if (po.NumArgs() != 2) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    std::string raw_nnet_rxfilename = po.GetArg(1),
```
danpovey: update the variable names to reflect what they are.

src/rnnlmbin/rnnlm-compute-prob.cc
```diff
@@ -0,0 +1,80 @@
...
+    using namespace kaldi::rnnlm;
+    typedef kaldi::int32 int32;
+    typedef kaldi::int64 int64;
+
+    const char *usage =
+        "Computes and prints to in logging messages the average log-prob per frame of\n"
+        "the given data with an nnet3 neural net.  The input of this is the output of\n"
+        "e.g. nnet3-get-egs | nnet3-merge-egs.\n"
+        "\n"
+        "Usage:  nnet3-compute-prob [options] <raw-model-in> <training-examples-in>\n"
```
danpovey: update usage message.

src/rnnlmbin/rnnlm-eval.cc
```diff
@@ -0,0 +1,139 @@
...
+    const SimpleObjectiveInfo *info = computer.GetObjective("output");
+    ofile << info->tot_objective + oos_cost * num_oovs << endl;
+    computer.Reset();
+  }
+}
+
+int main(int argc, char *argv[]) {
+  using namespace kaldi;
+  const char *usage = "Get egs for rnnlm\n"
+    "e.g. rnnlm-get-egs rnnlm-model wordlist.in wordlist.out text-file logprob-file\n";
```
danpovey: usage message is wrong.

danpovey: i mean program name is wrong.. generic things should be in angle brackets.. there should be "Usage:" and an "e.g.:" line.  See other examples.

src/rnnlmbin/rnnlm-eval.cc
```diff
@@ -0,0 +1,139 @@
...
+  unordered_map<string, int> in_wlist = ReadWordlist(in_wlist_file);
+  unordered_map<string, int> out_wlist = ReadWordlist(out_wlist_file);
+
+  cout << "sizes are " << in_wlist.size() << " " << out_wlist.size() << endl;
+
+  if (num_words != -1) {
+    KALDI_ASSERT(num_words >= out_wlist.size());
+  }
+
+  BaseFloat oos_cost = log(1.0 / (num_words - out_wlist.size() + 1));
```
danpovey: Using this as the oos_cost will give you probs that are not comparable with probs where the probabilities for OOS words were "correctly" computed using, say, a unigram.  So it would be hard to compare with other toolkits.
It may be necessary to compute the extra cost term outside this program somehow, and manually add in the correction term at the script level.

src/rnnlmbin/rnnlm-get-egs.cc
```diff
@@ -0,0 +1,159 @@
...
+#include <sstream>
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "hmm/posterior.h"
+#include "nnet3/nnet-example.h"
+#include "rnnlm/rnnlm-utils.h"
+#include "nnet3/nnet-diagnostics.h"
+
+using std::string;
```
danpovey: Per Google style guide,
"Do not use using-directives (e.g. using namespace foo)"
You may find it easier to put your code inside a namespace, e.g. kaldi::rnnlm or kaldi::nnet3.  Or make namespace rnnlm a sub-namespace of nnet3.

src/rnnlmbin/rnnlm-get-egs.cc
```diff
@@ -0,0 +1,159 @@
...
+  int cur_line = 0;
+  int input_dim = wlist_in.size();
+  int output_dim = wlist_out.size();
+  while (getline(ifile, line)) {
+//    if (cur_line % 1000 == 0) {
+//      cout << "processing line " << cur_line << endl;
+//    }
+    vector<string> words = SplitByWhiteSpace(line);
+
+    vector<int> word_ids_in;
```
danpovey: use int32 as default integer type.

src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,145 @@
...
+  string word;
+  int id;
+
+  while (ifile >> word >> id) {
+    ans[word] = id;
+  }
+  return ans;
+}
+
+NnetExample GetEgsFromSent(const vector<int>& word_ids_in, int input_dim,
```
danpovey: It's inefficient to return a big struct from a function; use a pointer.

src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  for (int j = 0; j < word_ids_in.size(); j++) {
+    vector<std::pair<MatrixIndexT, BaseFloat> > pairs;
+    pairs.push_back(std::make_pair(word_ids_in[j], 1.0));
+    SparseVector<BaseFloat> v(input_dim, pairs);
+    input_frames.SetRow(j, v);
+  }
+
+  NnetExample eg;
+  eg.io.push_back(nnet3::NnetIo("input", 0, input_frames));
```
danpovey: [Important]
I notice that the egs are of variable length, equal to the sentence length.
This is going to be very inefficient when compiling computations, if the sentences are of variable length.
However, I am in the process of making some changes to the egs-merging that should make this much less of a problem-- it will enable us to batch up sentences that are the same length into minibatches, which will mean that the egs have a more regular structure and the structure is more frequently repeated.  I am also making the nnet3 compilation process much more efficient (on egs with regular structure).
This should be checked in in 2 or 3 weeks if all goes to plan.
In your logs keep track of the "total GPU time" reported, versus the total time taken you'll see in the Accounting:" line.  If the total GPU time is much less than the total, then you are taking time not on GPU (e.g. for compilation, or something like that).  Point me to one of your logs.

src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,145 @@
...
+  for (int k = n; k < u.size(); k++) {
+    tot_weight += u[k];
+//    cout << "  tot-weight = " << tot_weight << endl;
+    BaseFloat pi_k1_k1 = u[k] / tot_weight * n;
+
+    if (pi_k1_k1 > 1) {
+      pi_k1_k1 = 1;
+    }
+
+    BaseFloat p = BaseFloat(rand()) / RAND_MAX;
```
danpovey: Please use RandUniform().

src/rnnlmbin/rnnlm-eval.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  po.Read(argc, argv);
+
+  if (po.NumArgs() != 5) {
+    po.PrintUsage();
+    exit(1);
+  }
+
+  int i = 1;
+  string rnn_file = po.GetArg(i++);
```
danpovey: Please use explicit numbers in calls to GetArg().  It makes it easier to see what the usage is.

src/rnnlm/rnnlm-nnet.cc
```diff
@@ -0,0 +1,98 @@
...
+
+  // TODO(hxu) will allow for more flexible types
+//  input_projection_ = new LmLinearComponent();
+//  output_projection_ = new AffineComponent();
+//  output_layer_ =  new LogSoftmaxComponent();
+
+  std::vector<string> type(2);
+  std::vector<string> lines(2);
+  std::vector<ConfigLine> config_lines(2);
+
```
danpovey: I think it might be better not to put these special components in the config at all.
It is a bit confusing, because they aren't part of the network proper (the network has its own input and output nodes).
If you do want to put them in the same config file, it would be possible to give them different directives (e.g. instead of 'component' you could call it input-preprocessing-component ... and output-preprocessing-component)... and this function could process such lines and then strip them out of the main config file (you could rewrite the remaining lines of the config file to an ostringstream).

src/rnnlm/rnnlm-nnet.cc
```diff
@@ -0,0 +1,98 @@
...
+  std::vector<ConfigLine> config_lines(2);
+  
+  for (int i = 0; i < 2; i++) {
+    config_is >> type[i];
+    getline(config_is, lines[i]);
+    config_lines[i].ParseLine(lines[i]);
+  }
+
+  int i = 0; 
+  input_projection_ = dynamic_cast<LmInputComponent*>(LmComponent::NewComponentOfType(type[i]));
```
danpovey: Personally I don't think it should be necessary to set up a class hierarchy for this purpose.  I don't think the Component interface [or a modified copy of it] is at all appropriate for this purpose, you could just create two simple classes with a few suitable functions, for the input and output pre/post-processing, respectively.  This is serious overkill.

You could even consider merging the two of them into a single class, because it may turn out that it's helpful to have the same vectors represent the words at the input and the output of the network.

src/rnnlm/rnnlm-training.cc
```diff
@@ -0,0 +1,587 @@
...
+LmNnetSamplingTrainer::LmNnetSamplingTrainer(
+                          const LmNnetTrainerOptions &config,
+                          LmNnet *nnet):
+                          config_(config),
+                          nnet_(nnet),
+                          compiler_(*nnet->GetNnet(), config_.optimize_config),
+                          num_minibatches_processed_(0) {
+
+// TODO(hxu)
+  int s = nnet_->O()->OutputDim();
```
danpovey: O() ???

src/rnnlm/rnnlm-training.cc
```diff
@@ -0,0 +1,587 @@
...
+
+  vector<int> samples = Select(unigram, k);
+  vector<vector<int> > indexes(k);
+  for (int i = 0; i < k; i++) {
+    indexes[i] = samples;
+
+    const SparseVector<BaseFloat> &sv = post.Row(i);                              
+    int non_zero_index = -1;                                                    
+    sv.Max(&non_zero_index); 
+    indexes[i].push_back(non_zero_index);
```
danpovey: I think you need to think a bit more carefully about what happens when you already sampled one of the output labels.  IMO, what would make more sense here is to first say, "we're going to select all the indexes corresponding to output labels [and use weight 1 for them, because we sampled them w.p. 1], and then sample k items from the remaining unigram distribution [and weight them with 1.0 / [the probability with which we sampled them.]).  And note: the probability with which we sampled them != 1/unigram-prob.  Depends on k, and depends on the things we did not sample because they were labels.


Also, you don't seem to be doing the thing that we discussed, which is to select the same set of output labels for the entire minibatch.  That's the only way to make the matrix multiplication reasonably efficient on GPU.

src/rnnlm/rnnlm-training.cc
```diff
@@ -0,0 +1,587 @@
...
+  *tot_weight = post.Sum();
+  *tot_objf = 0;
+  for (int i = 0; i < k; i++) {
+    KALDI_ASSERT(out[i].size() == k + 1);
+//    KALDI_LOG << "out-" << i << " is " << out[i][k];
+    *tot_objf += out[i][k]; // last one (k) is the correct lable
+    for (int j = 0; j < k; j++) {
+//      KALDI_ASSERT(out[i][j] == out[i][j]);
+//      KALDI_LOG << "out-" << i << " " << j << " is " << out[i][j];
+      *tot_objf -= exp(out[i][j]) / unigram[indexes[i][j]];
```
danpovey: This is not right.  You need to divide by "the probability with which we sampled this element", which is not equal to 1/unigram-prob.  [for a start, because k != 1; but it's not just a question of multiplying by k, because some probs could get > 1.  You'd need to truncate to 1 and renormalize.]
Also, it would be more efficient to not treat the labels as a special case here, i.e. you should treat the set of labels as having been sampled with probability 1, for purposes of the denominator computation.  That way you are taking advantage of the fact that you computed them.
You should probably take the time to prepare a writeup of some kind on the math, and think about it in terms of importance sampling, and expected derivatives, and things like that.

src/rnnlm/rnnlm-training.cc
hainan-xv: Thanks for reviewing this.  I was actually planning to do a clean-up before asking you to do to review this... that's why there are some very bad practices which I merely used for debugging. 

About the algorithm, I think now I understand what you were proposing. I thought you meant for all training-egs to share the same samples, and each has a unique "correct" label; now I know you meant they ALL should compute the same set of labels, which include all samples, and all correct labels in all egs. I will make the changes.

src/rnnlm/rnnlm-training.cc
hainan-xv: Hmm right, thanks for pointing out this mistake. This might be why I'm getting very bad training likelihood so far...

I will fix all the issues that you mentioned in the review. 

One of the logs for an actual training run is here 
/export/b02/hxu/kaldi/egs/ami/s5/debug-rnnlm-rnn-0.001-0.0001-1.03-128-200-8--sample/log/train.rnnlm.4.log

src/rnnlm/rnnlm-component.cc
```diff
@@ -0,0 +1,788 @@
...
+      KALDI_ASSERT((*out)[i][j] == (*out)[i][j]);
+    }
+  }
+}
+
+void AffineSampleLogSoftmaxComponent::Backprop(
+                               const vector<vector<int> > &indexes,
+                               const MatrixBase<BaseFloat> &in_value,
+                               const MatrixBase<BaseFloat> &, // out_value
+                               const vector<vector<BaseFloat> > &output_deriv,
```
danpovey: vector<vector<BaseFloat> > is the wrong data-structure to use here...
all of this stuff should be done on GPU, you can use functions with names *CopyRows*() to selectively copy rows from a matrix.

danpovey: ... but of course don't access a CuMatrix via operator (), that's very slow..  You may be able to use CopyRows somehow with the matrix Exp() function to compute the various exp() quantities you need.. in any case think carefully about it...

danpovey:
OK... obviously you are getting very bad GPU utilization (GPU time 10 sec
vs. overall 171 seconds).

Objf doesn't seem to be in the expected range:

LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 0-99 is -3122.96 over 13564
frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 100-199 is -4374.56 over
13545 frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 200-299 is -3962.82 over
13563 frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 300-399 is -3624.18 over
13621 frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 400-499 is -3568.57 over
13696 frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 500-599 is -4852.24 over
13345 frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 600-699 is -3760.91 over
13636 frames.
LOG (rnnlm-train:PrintStatsForThisPhase():rnnlm-training.cc:385) Average
objective function for 'output' for minibatches 700-799 is -7981.68 over
13656 frames.
LOG (nnet3-shuffle-egs:main():nnet3-shuffle-egs.cc:103) Shuffled order of
13117 neural-network training examples using a buffer (partial
randomization)

Certainly the error in the objf computation wouldn't help here.
But there may be other issues.

The objf with exp in it will be vulnerable to divergence as the exp
function gets large very fast... I think I mentioned to you by email a
trick that can be used to get around this.  But maybe try to get it to work
without the trick first.  It may be helpful, in making it converge, to
introduce a fixed scalar bias into the setup so that if a word has a zero
vector representation (giving zero output), it's given a very small
log-prob, instead of a log-prob of 1 which is obviously way too large.
That way you don't have to separately learn, for every single word, that
it's supposed to have a small probability.

src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,233 @@
...
+    vector<std::pair<int32, BaseFloat> > p;
+    p.push_back(std::make_pair(word_ids_out[i], 1.0));
+    posterior.push_back(p);
+  }
+
+  eg.io.push_back(nnet3::NnetIo("output", output_dim, 0, posterior));
+  return eg;
+}
+
+void SelectWoReplacement(const vector<BaseFloat> &u, int n, vector<int> *out) {
```
danpovey: avoid things like Wo in fn names per google style guide, use Without.

src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,233 @@
...
+
+  for (int k = n; k < u.size(); k++) {
+    tot_weight += u[k];
+//    cout << "  tot-weight = " << tot_weight << endl;
+    BaseFloat pi_k1_k1 = u[k] / tot_weight * n;
+
+    if (pi_k1_k1 > 1) {
+      pi_k1_k1 = 1;  // must add
+    } else {
+      BaseFloat p = BaseFloat(rand()) / RAND_MAX;
```
danpovey: As I said before, use RandUniform().  Your code would break badly on Windows, for instance.

danpovey: As I said before, use RandUniform(). Your code would break badly on Windows, for instance.

src/rnnlm/rnnlm-utils.cc
danpovey: @keli78, can you please work on a PR against Hainan's repo to add test code for this?
Need to:
- test that the distribution of the samples converges to the correct distribution [have a test that fails to terminate if this is not true, and use a fixed Euclidean distance on the vector of probabilities, like 0.01.]
- test that probabilities 1.0 are always sampled and probabilities 0.0 are never sampled.

danpovey: @keli78, also make sure that the test is a randomized one that is run multiple times, and that different randomly chosen k and n values are used; and also specially test things that might be corner cases like when the vector is all ones and zeros, or all ones and k==n, or k=1 or k=2, things like that.  [I'm assuming here that k is the dimension of the 'u' vector and n is the number of things to be sampled... not sure what the paper used.]

src/rnnlm/rnnlm-utils.cc
```diff
@@ -0,0 +1,233 @@
...
+    vector<std::pair<int32, BaseFloat> > p;
+    p.push_back(std::make_pair(word_ids_out[i], 1.0));
+    posterior.push_back(p);
+  }
+
+  eg.io.push_back(nnet3::NnetIo("output", output_dim, 0, posterior));
+  return eg;
+}
+
+void SelectWithoutReplacement(const vector<BaseFloat> &u, int n, vector<int> *out) {
```
danpovey: as I said before, this should have 'sample' in the name, not 'select' (sample implies randomness).

jtrmal:
@hainan-xv this have conflicts, can you please resolve it. thanks

Issue #1268, opened 2016-12-15
Reg. local/score_sclite.sh in egs/gale_mandarin


vijayaditya:
I was just taking a look at the gale_mandarin setup and was confused to see 

```bash
if [ $stage -le 2 ]; then  
  $cmd LMWT=$min_lmwt:$max_lmwt $dir/scoring/log/score.LMWT.log \
    cp $data/stm $dir/score_LMWT/ '&&' \
    $hubscr -p $hubdir -V -l english -h hub5 -g $data/glm -r $dir/score_LMWT/stm $dir/score_LMWT/${name}.ctm || exit 1;
fi
```
in the local/score_sclite.sh where `-l english` was specified. I was wondering if this script is ever called from local/score.sh ie., is there an stm file in the data directories being decoded ? 

I guess it might have been unintentionally copied, so better delete it to avoid confusion.

vijayaditya:
Same issue even in egs/hkust.

naxingyu:
So, there is not stm in gale_mandarin or hkust. score_sclite.sh is deprecated.

vijayaditya:
@naxingyu In that case I would recommend cleaning up the score.sh script and deleting score_sclite.sh scripts.

naxingyu:
ok.

naxingyu:
This issue is addressed.

vijayaditya:
Thanks.

Vijay

jtrmal:
Closing.  Y.

PR #1281, opened 2016-12-21
cleanup-gale-mandarin-scoring

naxingyu:
addressing #1268


danpovey:
@jtrmal, can you please check this?

jtrmal:
Those are mostly removals which are OK -- I would, however, suggest  to replace the score.sh by the link to steps/scoring/score_kaldi.sh or creating a new simple scoring script (for scoring both WER and CER) as described  inside the file  steps/scoring/score_kaldi.sh

naxingyu:
You are right. It’s more clear. I’ve link it to score_kaldi_cer.sh

danpovey:
@jtrmal, please merge this if you think it's OK.

naxingyu:
Is this PR being reviewed?

jtrmal:
I'll review it today.   I didn't notice you've updated it.
Y.

naxingyu:
@jtrmal I'm testing this change on hkust. It should be ready to merge if the results don't get affected. I'll keep you posted.

naxingyu:
@jtrmal score_kaldi_cer.sh does not produce same result as the local score.sh. So I keep using the local score in hkust. What do you think about this one?

jtrmal:
Did you check what is the difference?  You can supply a custom tokenizer.
By default,  the score_cer script does not split (by character) nor filter
the non-speech events (which I believe is the correct behavior)
Y.

naxingyu:
Yes. That is the difference. In the local scoring script, I split the words into characters. I think that makes sense because it is the character error that is being computed, not word error.

Xingyu

> 在 2016年12月28日，16:19，jtrmal <notifications@github.com> 写道：
> 
> Did you check what is the difference? You can supply a custom tokenizer.
> By default, the score_cer script does not split (by character) nor filter
> the non-speech events (which I believe is the correct behavior)
> Y.
>

jtrmal:
Can you show us the difference in tokenization?
I'm not sure if we are talking about the same thing.  What I was saying is
that for example non-speech unit "<NOISE>" won't be split into characters.
Y.

naxingyu:
I see. In the local script, non-speech units are filtered. That should be the difference.

Xingyu

> 在 2016年12月28日，16:41，jtrmal <notifications@github.com <mailto:notifications@github.com>> 写道：
> 
> Can you show us the difference in tokenization?
> I'm not sure if we are talking about the same thing. What I was saying is
> that for example non-speech unit "<NOISE>" won't be split into characters.
> Y.
> 
>

jtrmal:
you can remove them by supplying the wer_filter scripts -- I think even the
score_kaldi_cer.sh supports using them
y.

naxingyu:
Filter scripts are added. Results are the same as before now. Ready.

X.

> 在 2016年12月28日，17:30，jtrmal <notifications@github.com> 写道：
> 
> you can remove them by supplying the wer_filter scripts -- I think even the
> score_kaldi_cer.sh supports using them
> y.
> 
> On Wed, Dec 28, 2016 at 10:24 AM, Xingyu Na <notifications@github.com>
> wrote:
> 
> > I see. In the local script, non-speech units are filtered. That should be
> > the difference.
> >
> > Xingyu
> >
> > > 在 2016年12月28日，16:41，jtrmal <notifications@github.com <mailto:
> > notifications@github.com>> 写道：
> > >
> > > Can you show us the difference in tokenization?
> > > I'm not sure if we are talking about the same thing. What I was saying is
> > > that for example non-speech unit "<NOISE>" won't be split into
> > characters.
> > > Y.
> > >
> > >
> > > On Dec 28, 2016 9:37 AM, "Xingyu Na" <notifications@github.com <mailto:

jtrmal:
THanks, merging. y.

PR #1287, opened 2016-12-24
Add a new -3,3 layer to nnet3/TDNN script

hhadian:


egs/swbd/s5c/local/nnet3/tuning/run_tdnn_c.sh
```diff
@@ -0,0 +1,106 @@
...
+if [ $stage -le 9 ]; then
+  echo "$0: creating neural net configs";
+
+  # create the config files for nnet initialization
+  steps/nnet3/tdnn/make_configs.py  \
+    --feat-dir data/${train_set}_hires \
+    --ivector-dir exp/nnet3/ivectors_${train_set} \
+    --ali-dir $ali_dir \
+    --relu-dim 1024 \
+    --splice-indexes "-2,-1,0,1,2 -1,2 -3,3 -3,3 -7,2 0"  \
```
danpovey: Hm.
I'm not too happy about checking in non-xconfig-based recipes at this point.
Is there any chance you could find the time to convert it to use xconfigs?  
It looks like we haven't checked in non-chain scripts using xconfigs yet, so this will be a good first step.
The non-chain system's xconfig is basically a simplification of the config for the chain-models... you could start with a chain TDNN's xconfig and modify it, while of course using the correct splice indexes for the TDNN.
You could do that as a separate experiment number in this sequence of experiments, so that if there are any changes arising from moving to xconfigs, we can see them clearly.

Basically I think you have to take a chain tdnn system using xconfigs,
remove the whole xent branch of the output, and change the output layer to be something like:

  output-layer name=output input=tdnn7 dim=$num_targets max-change=1.5                                       

(i.e. removing a bunch of options).
Check whether the max-change value is what we were using before by comparing with the configs of the regular nnet3 baseline.
You may not need the final "0" for the splice-indexes, because I think that is for the output layer.
Compare the config's structure (foo/bar/xconfigs/init.config, I think) with the structure from one of the later progress logs in your baseline nnet3 experiment: it should be about the same.

danpovey: Oops, I realized there is one more thing that needs to be done to get this working on non-chain nnet3 the same way as the baseline setup was working.  It relates to the presoftmax-prior-scale.  I *think* you can write:
```
output-layer name=output input=tdnn7 dim=$num_targets max-change=1.5 presoftmax-prior-scale-file=$dir/configs/presoftmax_prior_scale.vec
```
and it should work correctly with the current script.  At least that's the way it was designed to work.
I don't know if this has ever been tested.
@vijayaditya might know.

hhadian: I will do it. I will compare the resulting configs and then run the experiment.

hhadian: I got an error about presoftmax-prior-scale:

> Configuration
> value presoftmax-prior-scale-file=exp/nnet3/tdnn_a_sp/configs/presoftmax_prior_scale.vec was not expected in layer of type output-layer

I omitted this setting temporarily and generated the xconfigs and converted them to regular configs. The main difference from the old configs (generated without xconfig) is that there is no layer1.config, layer2.config, ..., layernN.config in the new configs dir, while `nnet3/train_dnn.py` seems to need these layer-by-layer configs.

danpovey:
Oh-- it's called presoftmax-scale-file.
train_dnn.py should still work with this output-- a soft link is made from
layer1.config so it looks to that script like the network has just one
layer.
When comparing with the old configs, best to look at the config from a
later progress log, since there it will be shown as one big config, not
many small ones.

hhadian:
OK. I will test it now. Everything else seems OK.

danpovey:
Please update check that the config-generation still works for you with the
latest master code [it should].  I improved the error-handling to print
more helpful error messages when there are errors parsing xconfig lines.

hhadian:
OK, I will check.

hhadian:
The WER after using xconfig (without chaning anything else) is: 
```
# %WER 20.0 | 4459 42989 | 82.3 12.4 5.3 2.3 20.0 57.5 | exp/nnet3/tdnn_3c_sp/decode_eval2000_hires_sw1_tg/score_10_0.0/eval2000_hires.ctm.filt.sys
# %WER 18.2 | 4459 42989 | 83.8 11.1 5.0 2.0 18.2 54.9 | exp/nnet3/tdnn_3c_sp/decode_eval2000_hires_sw1_fsh_fg/score_11_0.0/eval2000_hires.ctm.filt.sys
```
The baseline (i.e. using regular configs) WER is 19.7 (fg) and 17.9 (tg).
I compared the progress files. The networks are the same in terms of number of parameters and layers, and layer size and type (i.e. the network topologies are identical AFAIK). However, this is not true for the first few progress files, in which the old-config system has fewer layers.
I can share the progress files if necessary.

danpovey:
Show the output of chain_dir_info.pl on both directories.

hhadian:
The output, with the full path to the directories:

**new config**: ` /home/hhadian/kaldi-fork/egs/swbd/s5c/exp/nnet3/tdnn_3c_sp/: num-iters=165 nj=3..16 num-params=16.7M dim=40+100->8787 combine=-1.51->-1.49`
corresponding script: `/home/hhadian/kaldi-fork/egs/swbd/s5c/local/nnet3/tuning/run_tdnn_3c.sh`

**old config**: `/home/hhadian/kaldi-fork/egs/swbd/s5c/exp/nnet3/tdnn_a_sp/: num-iters=165 nj=3..16 num-params=16.7M dim=40+100->8787 combine=-1.52->-1.50`
corresponding script: `/home/hhadian/kaldi-fork/egs/swbd/s5c/local/nnet3/tuning/run_tdnn_a.sh`

hhadian:
OK, will prepare the script.

hhadian:
I pushed the script. There was no compute_prob_train.final log in nnet3 logs so I used compute_prob_train.combined. 
I also pushed the two scripts (baseline with xconfig, one extra -3,3 layer with xconfig)

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_3c.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+# 3c is as a, but uses xconfig.
+
+# At this script level we don't support not running on GPU, as it would be painfully slow.
+# If you want to run without GPU you'd have to call train_tdnn.sh with --gpu false,
+# --num-threads 16 and --minibatch-size 128.
+
+# System                   tdnn_a   tdnn_3c
+# WER on train_dev(tg)      17.41     17.37
```
danpovey: oh... you didn't show the train_dev numbers before... this seems to confirm that what we saw on eval2000 was just noise.

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_3c.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+# 3c is as a, but uses xconfig.
```
danpovey: This naming scheme is not what we normally use.  It should be just linear: a,b,c, ..., [or 1a,1b,1c,...] then 2a,2b,2c... and so on.

The link points to run_tdnn_c.sh but I don't see that file..

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_3c.sh
hhadian: That's interesting. What is the cause of this noise mainly?

danpovey:
Just statistical noise due to limited test-set size... differences of 0.2%
or so can't always be trusted.

egs/swbd/s5c/local/nnet3/tuning/run_tdnn_3c.sh
hhadian: I will fix it.

hhadian:
Is the naming correct now?

danpovey:
Looks good, thanks.  Merging.

PR #1288, opened 2016-12-24
Fix raw nnet training

lifeiteng:
minor fixes

### DIFF
--- a/egs/wsj/s5/steps/nnet3/get_egs.sh
+++ b/egs/wsj/s5/steps/nnet3/get_egs.sh
@@ -184,7 +184,7 @@ if [ -f $dir/trans.scp ]; then
 fi
 
 if [ ! -z "$online_ivector_dir" ]; then
-  ivector_dim=$(feat-to-dim scp:$online_ivector_dir/ivector_online.scp - 2>/dev/null) || exit 1;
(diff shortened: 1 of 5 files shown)


danpovey:
I don't think this is the right fix for the fact that feat-to-dim leads to some unnecessary output.
The issue is that if an error did happen at that place, there would be no output to indicate what had happened.
In train_mono.sh this is done:
```
  if ! feat_dim=`feat-to-dim "$example_feats" - 2>/dev/null` || [ -z $feat_dim ]; then
    feat-to-dim "$example_feats" -
    echo "error getting feature dimension"
    exit 1;
  fi
```
(because the issue is more likely to happen at the beginning of training), and I didn't do that everywhere because it's a little verbose.

Also, this issue exists in a lot of places and, if addressed, should be addressed everywhere.  And getting the iVector dim would not generate scary warnings, so no need to do it there.


So please remove that from the PR and I'll commit the other things... what to do about that excessive output is a different discussion that we can keep separate.

lifeiteng:
@danpovey update
but somewhere like https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/nnet3/get_egs_targets.sh#L272 still have this problem,
also in chain/get_egs.sh

danpovey:
@vimalmanohar, can you please look over this and make sure it's OK to merge?

PR #1293, opened 2016-12-29
adressing  python package suggestion issue reported by Alexander Gorodetski

jtrmal:
@danpovey how about this? I think we should be more conservative for the other architectures, i.e. I'm only changing the debian/ubuntu suggestion


danpovey:
Looks good.  I'll merge.
BTW, for a little while, when you've reviewed something, let me commit myself... I'm trying to develop a convention for commit messages.

PR #1294, opened 2016-12-30
[scripts][nnet3] Remove the ability to separately specify validation …

danpovey:
…l/r context; use regular chunk-size for valid/train-subset egs.

I'm not sure whether to aim to test this PR separately and merge it soon, or to wait till my other extensive changes to nnet3 stuff, including the egs creation process and the compilation process, need to be tested.

### DIFF
--- a/egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh
+++ b/egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh
@@ -10,7 +10,7 @@ set -e
 # note: this relies on having a cluster that has plenty of CPUs as well as GPUs,
 # since the lattice generation runs in about real-time, so takes of the order of
 # 1000 hours of CPU time.
-# 
+#
 
 
 stage=0
@@ -59,8 +59,8 @@ decode_start_epoch=1 # can be used to avoid decoding all epochs, e.g. if we deci
 
 if $use_gpu; then
   if ! cuda-compiled; then
-    cat <<EOF && exit 1 
-This script is intended to be used with GPUs but you have not compiled Kaldi with CUDA 
+    cat <<EOF && exit 1
+This script is intended to be used with GPUs but you have not compiled Kaldi with CUDA
 If you want to use GPUs (and have them), go to src/, and configure and make on a machine
 where "nvcc" is installed.  Otherwise, call this script with --use-gpu false
 EOF
@@ -95,7 +95,7 @@ if [ $frame_subsampling_factor -ne 1 ]; then
     rm ${train_ivector_dir}_fs/ivector_online.scp 2>/dev/null || true
 
     data_dirs=
-    for x in `seq -$[frame_subsampling_factor/2] $[frame_subsampling_factor/2]`; do 
+    for x in `seq -$[frame_subsampling_factor/2] $[frame_subsampling_factor/2]`; do
       steps/shift_feats.sh --cmd "$train_cmd --max-jobs-run 40" --nj 350 \
         $x $train_data_dir exp/shift_hires mfcc_hires
       utils/fix_data_dir.sh ${train_data_dir}_fs$x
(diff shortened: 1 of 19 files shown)


danpovey:
@vimalmanohar and @vijayaditya, you may want to glance at this, but no need to thoroughly check it, it's straightforward..;

@hhadian, would you mind testing this for me?  No hurry-- the process can take a couple of weeks.  It will be better to get this tested independently, so that later, when we check the more extensive changes to the nnet3 setup, we know that this one is already checked in and is not causing any problems.

The important things to run are the nnet3 LSTM[or, in general, recurrent-RNN] scripts and the discriminative training scripts on top of that.  Take note of how long it takes to compute the train/valid probs on each iteration, and the various diagnostic probs; also the time taken in the 'combine' step.
Best to do this on top of a modern, xconfig-based setup, maybe a TDNN+LSTM setup followed by discriminative training, you can create a discriminative training script for that setup and make a link to the tuning directory for that.
Actually more scripts than these are affected by the changes, but these are the most likely to be affected.
Also check the time taken in the compute-priors jobs in the discriminative training.

vijayaditya:
OK will do.

--Vijay

hhadian:
OK, I will do it.

hhadian:
So I am going to do the following:
1. Prepare a TDNN+LSTM setup for nnet3 using xconfig (I was going to do this anyway; for low frame rate experiments)
2. Prepare a discriminative setup on top of that
3. Try these scripts before and after applying the changes in this PR, and compare the results and runtime of different computations.
Are there any scripts/tools to report/compare the times taken for computing probs, combining, etc. ?
Hossein

danpovey:
Sounds great-- thanks.

There is no special tool but the relevant log files are the
compute_prob_*.log and the 'combine.log'... there should be a line with
'Accounting:' in it that gives you the time taken.

hhadian:
OK. I will get the times from the log files.
Regarding TDNN+LSTM for nnet3, I am preparing it based on chain scripts (local/chain/run_tdnn_lstm.sh) and current nnet3/LSTM script (local/nnet3/run_lstm.sh). These options are different:

`deriv-truncate-margin` --> no such thing in nnet3/train_rnn.py
`momentum` is 0.0 for chain, but it is 0.5 for nnet3/LSTM setup
initial and final learning rate for chain are `0.001, 0.0001` whereas they are `0.0003, 0.00003` for nnet3/LSTM (significant difference)

What should I do about these?

danpovey:
Hm.  The option to train_rnn.py is called
--trainer.deriv-truncate-margin
.. which is the same as in the chain script... make sure your repo is up to
date if not.
As for momentum-- you could try 0.5.  [I think we found that momentum is
not helpful for chain; anyway it was never a huge difference.]
As for the initial/final learning rates: maybe take them from the
nnet3+LSTM setup.  I'm not 100% sure if we have a rule of thumb about how
the learning rates should differ in nnet3 versus chain, @vijayaditya may
know; but anyway the objf has a lower dynamic range for chain so it's not
surprising if we'd want lower learning rates for chain.

hhadian:
Oh! I will make the code up-to-date now.
I will try 0.5 for momentum. 
Since the learning rates for nnet3/TDNN are much larger (0.0017, 0.00017) than nnet3/LSTM (0.0003, 0.00003), and since this setup is a combination of TDNN and LSTM, a value in between might make sense for the start (i.e. 0.001, 0.0001). Is that OK to start with?

danpovey:
No, I'd use the LSTM one.  The TDNN layers are much less sensitive to the
learning rate schedule because they are invariant to the scale of the
parameters [thanks to the renorm layer], so if the learning rate is too
large the parameters just get larger to compensate.

hhadian:
OK I'll go with LSTM's.

danpovey:
I'm closing this PR; the code is in the shortcut branch and will be merged when that branch is merged.

PR #1304, opened 2017-01-03
fix: chain do not need sp align

naxingyu:
This reverts part of the change in #1303 
I made a mistake to use alignments of sp data in chain training, which is not true. Now #1299 should be fixed.

### DIFF
--- a/egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
+++ b/egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
@@ -37,8 +37,9 @@ EOF
 fi
 
 dir=$dir${affix:+_$affix}
+build_tree_train_set=train_nodup
 train_set=train_nodup_sp
-ali_dir=exp/tri5a_ali_nodup_sp
+build_tree_ali_dir=exp/tri5a_ali
 treedir=exp/chain/tri6_tree
 lang=data/lang_chain
 
@@ -52,7 +53,7 @@ local/nnet3/run_ivector_common.sh --stage $stage \
 if [ $stage -le 9 ]; then
   # Get the alignments as lattices (gives the CTC training more freedom).
   # use the same num-jobs as the alignments
-  nj=$(cat $ali_dir/num_jobs) || exit 1;
+  nj=$(cat $build_tree_ali_dir/num_jobs) || exit 1;
   steps/align_fmllr_lats.sh --nj $nj --cmd "$train_cmd" data/$train_set \
     data/lang exp/tri5a exp/tri5a_lats_nodup_sp || exit 1;
   rm exp/tri5a_lats_nodup_sp/fsts.*.gz # save space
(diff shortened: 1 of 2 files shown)


danpovey:
This is still not right because when you are building the tree you use the perturbed data set:

 steps/nnet3/chain/build_tree.sh --frame-subsampling-factor 3 \
       --leftmost-questions-truncate -1 \
       --cmd "$train_cmd" 11000 data/$train_set $lang $ali_dir $treedir

[note: $train_set is train_nodup_sp].. and now the alignments are mismatched.
You may have to define a separate variable for the un-perturbed dataset, or at any rate do it differently somehow.

naxingyu:
Oh it's a mess now.... I'll figure out in a couple of minutes...

naxingyu:
OK now it seems to be correct.

X.

egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
```diff
@@ -37,8 +37,9 @@ EOF
 fi
 
 dir=$dir${affix:+_$affix}
+ali_set=train_nodup
```
danpovey: ok but can we call this build_tree_train_set instead?  ali_set is not very clear.

naxingyu:
Done.

X.

egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
```diff
@@ -37,8 +37,9 @@ EOF
 fi
 
 dir=$dir${affix:+_$affix}
+build_tree_train_set=train_nodup
 train_set=train_nodup_sp
-ali_dir=exp/tri5a_ali_nodup_sp
+ali_dir=exp/tri5a_ali
```
danpovey: and also-- should have said this before-- please rename ali_dir to build_tree_ali_dir

naxingyu:
Right… done and cleaned. What about the other chain example scripts?

X.

danpovey:
OK, merging, thanks.
I think this issue was specific to this example.

13265170340:
When I run steps/nnet3/chain/build_tree.sh 
WARNING (acc-tree-stats[5.5.88~3-8e30f]:main():acc-tree-stats.cc:89) Alignments has wrong size 67 vs. 200
WARNING (acc-tree-stats[5.5.88~3-8e30f]:main():acc-tree-stats.cc:89) Alignments has wrong size 43 vs. 129
WARNING (acc-tree-stats[5.5.88~3-8e30f]:main():acc-tree-stats.cc:89) Alignments has wrong size 60 vs. 180
.....
ali_dir was nnet model
I don't know where the problem is.

danpovey:
You may have to mess with the --alignment-subsampling-factor option.  E.g. try 3.

13265170340:
@danpovey Still wrong

danpovey:
sorry-- set it to 1.

PR #1308, opened 2017-01-03
[WIP] Add boosted-MMI to chain

hhadian:
Does it look OK? 
If yes, I am going to add the two options boost and hard_boost to chain's train.py and acoustic_model.py.

### DIFF
--- a/src/chain/chain-training.cc
+++ b/src/chain/chain-training.cc
@@ -42,6 +42,7 @@ void ComputeChainObjfAndDeriv(const ChainTrainingOptions &opts,
                                                 nnet_output.NumCols(),
                                                 kUndefined);
     is_output_deriv_temporary = true;
+    KALDI_LOG << "***** Temporary nnet-out-deriv allocation.";
   }
 
   BaseFloat num_logprob_weighted;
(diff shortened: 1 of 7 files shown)


src/chain/chain-training.h
```diff
@@ -61,8 +61,18 @@ struct ChainTrainingOptions {
   // should have a softmax as its final nonlinearity.
   BaseFloat xent_regularize;
 
+  // The boosting coefficient (i.e. b in bMMI)
+  BaseFloat boost;
+  
+  // This controls the boosting approach. If true, hard boosting
+  // will be used: the boosting mask consists of 1's and 0's and is 
+  // independent of numerator compuatation (i.e. is a function of numerator
```
galv: compuatation -> computation

galv: I'm not familiar with boosted MMI, but do you mean "independent of denominator computation"?

src/chain/chain-training.h
```diff
@@ -61,8 +61,18 @@ struct ChainTrainingOptions {
...
+  // graph only).
+  bool hard_boost;
+
   ChainTrainingOptions(): l2_regularize(0.0), leaky_hmm_coefficient(1.0e-05),
-                          xent_regularize(0.0) { }
+                          xent_regularize(0.0), boost(0.1),
```
galv: Won't this make boosting on by default, breaking backwards compatibility?

hhadian: Oh yes! will fix it.

src/chain/chain-training.cc
```diff
@@ -54,9 +64,24 @@ void ComputeChainObjfAndDeriv(const ChainTrainingOptions &opts,
       numerator.Backward(xent_output_deriv);
     }
   }
+  
+  num_logprob_weighted += -opts.boost * supervision.num_sequences *
+                          supervision.frames_per_sequence;
+  CuMatrixBase<BaseFloat> *boosting_mask = nnet_output_deriv;
+  if (opts.boost != 0.0 && opts.hard_boost) {
+    boosting_mask = new CuMatrix<BaseFloat>(*nnet_output_deriv);
+    boosting_mask->ApplyHeaviside();
```
danpovey: looks fine, but it occurred to me that if the hard mask turns out to be better, we should probably (a) hardcode it to use the hard mask, and (b) we can give the soft version of the mask to the denominator computation and have it do the heaviside itself.  It creates a copy of the matrix anyway, to do the transpose, so we can have it copy the mask before the actual probs, do the heaviside, and then add in the actual probs.  but no need to do this before testing the idea out, the speed difference would be extremely tiny.

hhadian: OK. 
BTW even for the purpose of testing the idea, I think this copying (`new CuMatrix<BaseFloat>(*nnet_output_deriv)`) might fail due to limited memory. I remember I had memory allocation problems with batch-sizes > 128 when I was testing on-GPU numerator computation with **two** nnet_output_deriv matrices.
Anyway I will test it to see what happens.

danpovey:
If you test with a fairly small network this should not be an issue.
But I think the chain training scripts often use a minibatch size of 64.

danpovey:
Since this did not seem to be helpful, I'm closing the pull request.. don't delete the branch, though, in case we want to revisit this in future.

Issue #861, opened 2016-06-23
[enhancement] Extend queue.pl to trap signals


danpovey:
it would be nice to modify queue.pl to trap the signals KILL and TERM, and kill the job that has been submitted before exiting.

vijayaditya:
Another enhancement would be to provide an option for early exit i.e., if one of the  jobs in the array fails the script should kill all the jobs and exit rather than wait till the end and report the percentage of failure.   If this feature is not activated the script should be printing details of the jobs being aborted/exiting with non-zero status as soon as the failure occurs.

This is essential in large chain recipes like ASpIRE and fisher_swbd where the `align.sh`/`make_denlats.sh` jobs run for more than a week and failure of individual jobs is not reported till the end.

kkm000:
Done in #1309

PR #1309, opened 2017-01-03
[enhancement] Extend queue.pl to trap signals #861

jyang126jhu:
Trap signal "KILL" and "TERM", and kill the job that has been submitted before exiting.

### DIFF
--- a/egs/wsj/s5/utils/queue.pl
+++ b/egs/wsj/s5/utils/queue.pl
@@ -65,8 +65,13 @@ my %cli_options = ();
 my $jobname;
 my $jobstart;
 my $jobend;
-
+my $logfile;
+my @logfiles = ();
+my %hash_log = ();
 my $array_job = 0;
+my $submit_flag = 0;
+my $sge_job_id;
+my $early_exit = "false";
 
 sub print_usage() {
   print STDERR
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/queue.pl
```diff
@@ -82,6 +87,7 @@ ()
    "and change its behavior.  Otherwise it uses qstat to work out when the job finished\n" .
    "Options:\n" .
    "  --config <config-file> (default: $config)\n" .
+   "  --early-exit <true | false> (default: false). If true, if one of the jobs in the array fails, all the jobs submitted will be deleted and the script will exit\n" .
```
danpovey: please break the line here to avoid very long line [see how it's done for --mem].
And let's have this default to true; it's normally what we want.

egs/wsj/s5/utils/queue.pl
```diff
@@ -553,17 +609,6 @@ ()
 # But we don't know about its exit status.  We'll look at $logfile for this.
 # First work out an array @logfiles of file-locations we need to
```
danpovey: You moved this code re setting `@logfiles` into the if-statement.  This might cause the script to break if the -sync y option is used.

egs/wsj/s5/utils/queue.pl
```diff
@@ -90,6 +98,37 @@ ()
...
 }
 
+sub caught_signal {
+	if ($submit_flag ==1) { #captured signal after submitting jobs
+		system ("qdel $sge_job_id");
+		die "Caught a signal: $!, deleting SGE task: $sge_job_id and exiting\n";
+	}
+}
+
+sub check_status {
```
danpovey: OK, this is in response to Vijay's comment about checking the status before the job is finished....
firstly, I'm not sure whether we should do it in the same PR because the queue.pl script is easy to break.

Secondly, the way you've done it is problematic for a few reasons:
   - it would cause too much NFS traffic as it's run too frequently and on too many files (the loop it's called from only accesses one file per iteration; not all of them).  The queue.pl script has been finely tuned to be responsive enough without hogging NFS traffic. 
   - it's not ideal to spawn too many processes (tail -1), they fill up the system logs and cause other problems...  better to fseek not too far from the end of the file, read a few lines, and take the last of the lines, without invoking new processes.
   - I don't see that you have made allowances for what happens when there are left-over logs from a previous iteration.  [need to check file times.]

Overall, it may be better to make this a separate PR.

jyang126jhu: Ok, I'll check it now.

egs/wsj/s5/utils/queue.pl
```diff
@@ -488,11 +538,11 @@ ()
...
         # is designed to check every 10 waits at first, and eventually every 50
         # waits.
-        if ( -f $f ) { next; }  #syncfile appeared: OK.
+        if ( -f $f ) { check_status; next; }  #syncfile appeared: OK.
         my $output = `qstat -j $sge_job_id 2>&1`;
         my $ret = $?;
-        if ($ret >> 8 == 1 && $output !~ m/qmaster/ &&
-            $output !~ m/gdi request/) {
+        if (( $ret >> 8 == 1 && $output !~ m/qmaster/ &&
+            $output !~ m/gdi request/) || ($ret == 0) ) {
```
danpovey: not sure what the `|| ($ret == 0)` is for; this if-statement seems to be for the case when the job does not exist.  If you're going to change things this much, it needs to be explained clearly at the very least..

danpovey:
Thanks-- can you please test this a bit?  You can just create a file test.sh that sleeps for a while and try submitting that and see if it behaves as expected when the script is terminated in various ways.

jyang126jhu:
I've tested it with terminating by Interrupt from keyboards and Kill process by pid,  should I consider more terminating ways?

egs/wsj/s5/utils/queue.pl
```diff
@@ -65,8 +65,9 @@
 my $jobname;
 my $jobstart;
 my $jobend;
-
+my $submit_flag = 0;
```
danpovey: I think you can simplify this by removing the $submit_flag and just doing if (defined $sge_job_id) {...

PR #1320, opened 2017-01-06
Remove touch in fisher_swbd/s5/local/chain/run_blstm_6h.sh

stanleyguan:
I think the directory does not exist before running `train.py`

### DIFF
--- a/egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
+++ b/egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
@@ -107,10 +107,6 @@ if [ $stage -le 13 ]; then
      /export/b0{5,6,7,8}/$USER/kaldi-data/egs/fisher_swbd-$(date +'%m_%d_%H_%M')/s5c/$dir/egs/storage $dir/egs/storage
   fi
 
-  # keep egs around when that run dies.
-  mkdir $dir/egs
-  touch $dir/egs/.nodelete 
-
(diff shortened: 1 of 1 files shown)


naxingyu:
@danpovey LGTM. But do we still need this? It was for tuning and should be deprecated now?

egs/fisher_swbd/s5/local/chain/run_blstm_6h.sh
```diff
@@ -107,7 +107,9 @@ if [ $stage -le 13 ]; then
      /export/b0{5,6,7,8}/$USER/kaldi-data/egs/fisher_swbd-$(date +'%m_%d_%H_%M')/s5c/$dir/egs/storage $dir/egs/storage
   fi
 
-  touch $dir/egs/.nodelete # keep egs around when that run dies.
+  # keep egs around when that run dies.
+  mkdir $dir/egs
```
danpovey: it would be better to just remove the "touch" command altogether-- in any case the 'mkdir' command would cause the script to die if run after partial run since the dir would already exist [would need the -p].

stanleyguan: So if training dies in one of the iterations, without the touch I can still restart from that iteration without having to create egs again?

danpovey:
Yes, it won't delete the egs if it fails.  the comment was wrong.

stanleyguan:
Made a new commit to remove `touch`.

PR #1321, opened 2017-01-06
Minor bug fixes with error statements

### LINKED ISSUE #1316: Minor bug in steps/libs/nnet3/train/common.py
I think the attribute to access there is strerror not strerr.

At [line 202]( https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/libs/nnet3/train/common.py#L202)

                 "malformed files: {1}".format(egs_dir, e.strerr))

vimalmanohar:
Fixes #1316

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/train/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/common.py
@@ -142,8 +142,9 @@ def copy_egs_properties_to_exp_dir(egs_dir, dir):
             if os.path.isfile(file_name):
                 shutil.copy2(file_name, dir)
     except IOError:
-        raise Exception("Error while trying to copy egs "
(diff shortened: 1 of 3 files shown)


vijayaditya:
Looks good to me.

vijayaditya:
@danpovey To answer your question. We were previously raising a new exception in the calling methods after adding some context to the error message. The strerror attribute, which the backtrace string of the original error IIRC, was being concatenated to this error message.

In the current case Vimal decided to add the necessary context in the logger's error stream and just reraise the original exception. He is using `raise` rather than `raise Exception` to preserve the back trace information of the original error, without the necessity to concatenate it to the new error message.

See http://stackoverflow.com/questions/4825234/exception-traceback-is-hidden-if-not-re-raised-immediately

PR #1324, opened 2017-01-09
change DropoutComponent into by-frame mode

GaofengCheng:
i.e. dropout matrix will be setted like
[[1,1,1,1],[0,0,0,0],[0,0,0,0],[1,1,1,1],[0,0,0,0]]


src/nnet3/nnet-simple-component.cc
```diff
@@ -119,15 +119,21 @@ void DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
...
+  // i.e. [[1,1,1,1],[0,0,0,0],[0,0,0,0],[1,1,1,1],[0,0,0,0]]
   // This const_cast is only safe assuming you don't attempt
   // to use multi-threaded code with the GPU.
   const_cast<CuRand<BaseFloat>&>(random_generator_).RandUniform(out);
-
-  out->Add(-dropout); // now, a proportion "dropout" will be <0.0
-  out->ApplyHeaviside(); // apply the function (x>0?1:0).  Now, a proportion "dropout" will
-                         // be zero and (1 - dropout) will be 1.0.
-
+  // now, a proportion "dropout" will be <0.0
```
danpovey: This wouldn't be compatible with regular uses of dropout.  You should add an option to the component.
And it's wasteful to do get a whole random matrix and only use one column; you should use a submatrix consisting of one column and randomly set just that.
Try again and I'll comment; this may take a few tries.

GaofengCheng: got it

GaofengCheng: @danpovey I add the option to choose by frame or the existing dropout in c++ level in this PR, after Vimal's code is merged, then adding the scripts level drop mode choice, will this be OK?

danpovey: There doesn't necessarily have to be an option in the scripts, it can be hardcoded; and anyway we'll consider that later; but the component should support both options.  e.g. adding "dropout-per-frame=true" in the config file to the component, will set the per-frame mode.

src/nnet3/nnet-simple-component.cc
```diff
@@ -119,15 +119,21 @@ void DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
...
+  out->Add(-dropout);
+  // apply the function (x>0?1:0).  Now, a proportion "dropout" will
+  // be zero and (1 - dropout) will be 1.0.
+  out->ApplyHeaviside();
+  CuVector<BaseFloat> *random_drop_vector = new CuVector<BaseFloat>(in.NumRows(), kSetZero);
```
danpovey: this is not how you create a subvector, and it's a memory leak.  Please read the following carefully:
http://kaldi-asr.org/doc/matrix.html

GaofengCheng: OK, thx for information!

src/nnet3/nnet-simple-component.cc
```diff
@@ -119,16 +136,35 @@ void DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
...
+    out->ApplyHeaviside(); // apply the function (x>0?1:0).  Now, a proportion "dropout" will
+                          // be zero and (1 - dropout) will be 1.0.
+    out->MulElements(in);
+  } else {
+    // randomize the dropout matrix by row,
+    // i.e. [[1,1,1,1],[0,0,0,0],[0,0,0,0],[1,1,1,1],[0,0,0,0]]
+    // This const_cast is only safe assuming you don't attempt
+    // to use multi-threaded code with the GPU.
+    CuSubMatrix<BaseFloat> out_col_submatrix = out->ColRange(0, 1);
+    const_cast<CuRand<BaseFloat>&>(random_generator_).RandUniform(&out_col_submatrix);
```
danpovey: Since you are having to copy the matrix to a vector, and having to allocate a vector anyway, it would be better to set the vector to random directly.  But it will be more convenient to allocate it as a matrix, like:
CuMatrix<BaseFloat> tmp(1, out->NumCols(), kUndefined);
const_cast<...>(random_generator_).RandUniform(&tmp);
..
out->AddVecToCols(1.0, tmp.Row(0), 0.0);

src/nnet3/nnet-simple-component.cc
```diff
@@ -150,11 +186,26 @@ void DropoutComponent::Backprop(const std::string &debug_info,
...
-  ExpectToken(is, binary, "</DropoutComponent>");
+  //back-compatibility code.
+  std::string token;
+  ReadToken(is, binary, &token);
+  if(token == "<DropoutComponent>"){
+    ReadToken(is, binary, &token);
+  }
+  KALDI_ASSERT(token == "<Dim>");
+  ReadBasicType(is, binary, &dim_); // read dimension.
+  ReadToken(is, binary, &token);
```
danpovey: You can just do
ExpectToken(is, binary, "<DropoutProportion>")
instead of reading the token adn then checking it.

src/nnet3/nnet-simple-component.cc
```diff
@@ -150,11 +186,26 @@ void DropoutComponent::Backprop(const std::string &debug_info,
...
+  KALDI_ASSERT(token == "<DropoutProportion>");
+  ReadBasicType(is, binary, &dropout_proportion_); // read dropout rate
+  ReadToken(is, binary, &token);
+  if(token == "<DropoutPerFrame>"){
+    ReadBasicType(is, binary, &dropout_per_frame_); // read dropout mode
+    ReadToken(is, binary, &token);
+    KALDI_ASSERT(token == "</DropoutComponent>");
```
danpovey: you can move the duplicated assert past the if-statement.
and remember the spaces:
`if (...) {`

src/nnet3/nnet-simple-component.h
```diff
@@ -87,11 +87,15 @@ class PnormComponent: public Component {
...
  public:
-  void Init(int32 dim, BaseFloat dropout_proportion = 0.0);
+  void Init(int32 dim, BaseFloat dropout_proportion = 0.0,
+            bool dropout_per_frame = true);
 
-  DropoutComponent(int32 dim, BaseFloat dropout = 0.0) { Init(dim, dropout); }
+  DropoutComponent(int32 dim, BaseFloat dropout = 0.0,
+                   bool dropout_per_frame = true) {
+                     Init(dim, dropout, dropout_per_frame);
+                   }
```
danpovey: wrong indentation levels.

src/nnet3/nnet-simple-component.h
```diff
@@ -87,11 +87,15 @@ class PnormComponent: public Component {
...
 
-  DropoutComponent(): dim_(0), dropout_proportion_(0.0) { }
+  DropoutComponent(): dim_(0), dropout_proportion_(0.0), dropout_per_frame_(true) { }
```
danpovey: run cpplint.py and check you don't go past 80 chars.

src/nnet3bin/nnet3-copy.cc
```diff
@@ -91,8 +94,11 @@ int main(int argc, char *argv[]) {
     if (scale != 1.0)
       ScaleNnet(scale, &nnet);
     
-    if (dropout > 0)
-      SetDropoutProportion(dropout, &nnet);
+    if (dropout >= 0)
+      KALDI_ERR << "--dropout option is deprecated. "
```
danpovey: [this for Vimal, but]... "deprecated" means still supported but not recommended.  Since this will crash, deprecated is not the right term.  I would suggest to just remove the set-dropout-proportion option and related code from the program.  If no scripts are using it, this is OK.

src/nnet3/nnet-simple-component.cc
GaofengCheng: @danpovey  `CuMatrix tmp(1, out->NumCols(), kUndefined); ` ? or We should use `CuMatrix tmp(out->NumRow(), 1, kUndefined);`  We want to copy the same col to the whole matrix(so matrix will be the same in row dim), the tmp should be row_num*1, not 1*col_num. Is this right?

src/nnet3/nnet-simple-component.cc
GaofengCheng: @danpovey maybe we should decide whether to be back compatible to previous model. Previous DropoutComponent do not have per-frame opts. This checking aims to avoid crash. `ExpectToken(is, binary, "<DropoutPerFrame>")` on elder model will go wrong.

danpovey:
No,  tmp(1, out->NumCols(), kUndefined); was correct.
Then you can do tmp.Row(0) to get the 1st row, and call CopyColsFromVec()
to copy that row to each column of the matrix.

danpovey:
It definitely needs to be back compatible to the old format.

Except that code seems to have a bug in it

if(token == "<DropoutComponent>"){
should be
if(token == "<DropoutPerFrame>"){

src/nnet3/nnet-simple-component.cc
```diff
@@ -136,34 +136,27 @@ void DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
...
-    out_col_submatrix.ApplyHeaviside();
-    CuVector<BaseFloat> *random_drop_vector = new CuVector<BaseFloat>(in.NumRows(), kSetZero);
-    random_drop_vector->CopyColFromMat(out_col_submatrix, 0);
+    const_cast<CuRand<BaseFloat>&>(random_generator_).RandUniform(&tmp);
+    tmp.Add(-dropout);
+    tmp.ApplyHeaviside();
     out->SetZero();
-    out->AddVecToCols(1.0 , *random_drop_vector, 1.0);
+    out->AddVecToCols(1.0, tmp.Row(0), 0.0);
     out->MulElements(in);
```
GaofengCheng: Change from my previous method (random a whole matrix and select one col) into this one, and a test on AMI SDM is done, they are equivalent on WER:
previous way:
image
new way:
image

src/nnet3/nnet-simple-component.h
danpovey: dropout_per_frame should default to false (default to "normal" dropout).

src/nnet3/nnet-simple-component.cc
```diff
@@ -87,27 +87,32 @@ void PnormComponent::Write(std::ostream &os, bool binary) const {
...
+                            bool dropout_per_frame) {
   dropout_proportion_ = dropout_proportion;
+  dropout_per_frame_ = dropout_per_frame;
   dim_ = dim;
 }
 
 void DropoutComponent::InitFromConfig(ConfigLine *cfl) {
   int32 dim = 0;
   BaseFloat dropout_proportion = 0.0;
+  bool dropout_per_frame = true;
```
danpovey: this should default to false.

src/nnet3/nnet-simple-component.cc
```diff
@@ -87,27 +87,32 @@ void PnormComponent::Write(std::ostream &os, bool binary) const {
...
   bool ok = cfl->GetValue("dim", &dim) &&
     cfl->GetValue("dropout-proportion", &dropout_proportion);
+    // for this stage, dropout is hard coded in by-frame mode
   if (!ok || cfl->HasUnusedValues() || dim <= 0 ||
```
danpovey: where is the "dropout-per-frame" configuration value being read?
You should be doing
```
 cfl->GetValue("dropout-per-frame", &dropout_per_frame);
```
[don't check the return value of GetValue().]

src/nnet3/nnet-simple-component.cc
```diff
@@ -119,16 +124,28 @@ void DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
...
+    const_cast<CuRand<BaseFloat>&>(random_generator_).RandUniform(out);
+
+    out->Add(-dropout);  // now, a proportion "dropout" will be <0.0
+    out->ApplyHeaviside();  // apply the function (x>0?1:0).  Now, a proportion
+                           // "dropout" will be zero and (1 - dropout) will be 1.0.
+    out->MulElements(in);
+  } else {
+    // randomize the dropout matrix by row,
+    // i.e. [[1,1,1,1],[0,0,0,0],[0,0,0,0],[1,1,1,1],[0,0,0,0]]
+    CuMatrix<BaseFloat> tmp(1, out->NumRows(), kUndefined);
```
danpovey: You should now be able to change this code to:
(a) use a vector instead of a matrix for 'tmp', and
(b) use CopyVecToCols() instead of AddVecToCols(), which avoids the need to call SetZero().

In about 20 minutes I'll push @kangshiyin's changes regarding implementing CopyVecToCols() to the master branch; and @vesis84 has already implemented RandUniform for vectors

src/nnet3/nnet-simple-component.h
```diff
@@ -87,11 +87,15 @@ class PnormComponent: public Component {
 // "Dropout: A Simple Way to Prevent Neural Networks from Overfitting".
 class DropoutComponent : public RandomComponent {
  public:
-  void Init(int32 dim, BaseFloat dropout_proportion = 0.0);
+  void Init(int32 dim, BaseFloat dropout_proportion = 0.0,
+            bool dropout_per_frame = true);
 
-  DropoutComponent(int32 dim, BaseFloat dropout = 0.0) { Init(dim, dropout); }
+  DropoutComponent(int32 dim, BaseFloat dropout = 0.0,
+                   bool dropout_per_frame = true) {
```
danpovey: the Init and the close-brace are at the wrong indentation level-- much too high.  And make sure you are writing spaces and not tabs.

src/nnet3/nnet-simple-component.cc
danpovey: you didn't address this comment about 80 chars and cpplint.py

src/nnet3/nnet-utils.cc
```diff
@@ -615,6 +615,29 @@ void ReadEditConfig(std::istream &edit_config_is, Nnet *nnet) {
...
+                  << config_line.WholeLine();
+      }
+      DropoutComponent *component = NULL;
+      int32 num_dropout_proportions_set = 0;
+      for (int32 c = 0; c < nnet->NumComponents(); c++) {
+        if (NameMatchesPattern(nnet->GetComponentName(c).c_str(),
+                               name_pattern.c_str()) &&
+            (component =
+             dynamic_cast<DropoutComponent*>(nnet->GetComponent(c)))) {
+          component->SetDropoutProportion(proportion);
```
danpovey: You should do:
```
if (component != NULL) {
  component->SetDropoutProportion(proportion);
  num_dropout_proportions_set++;
}
```
That way you can call this without specifying the name (your code would have segfaulted).
And please rename 'component' to 'dropout_component'

src/nnet3/nnet-utils.cc
```diff
@@ -615,6 +615,29 @@ void ReadEditConfig(std::istream &edit_config_is, Nnet *nnet) {
...
+          num_dropout_proportions_set++;
+        }
+      }
+      KALDI_LOG << "Set dropout proportions for "
+                << num_dropout_proportions_set << " nodes.";
```
danpovey: please change 'nodes' to 'components'.

src/nnet3bin/nnet3-copy.cc
```diff
@@ -87,11 +87,11 @@ int main(int argc, char *argv[]) {
...
     if (learning_rate >= 0)
       SetLearningRate(learning_rate, &nnet);
-    
+
     if (scale != 1.0)
       ScaleNnet(scale, &nnet);
-    
-    if (dropout > 0)
+
+    if (dropout >= 0)
```
danpovey: I believe we're removing these dropout options from the scripts.
[however, for this code to work as intended, the default value of 'dropout' should be -1, not zero].
Anyway, we should remove all related options and code and rely on the --edits and --edits-config options.

src/nnet3/nnet-simple-component.cc
GaofengCheng: @danpovey CuVector does not support Heaviside(), cumatrix remianed, but AddVecToCol is replaced with CopyColsFromVec, out->SetZero() is skipped

GaofengCheng: @danpovey Rerun with this code, results OK

src/nnet3/nnet-simple-component.cc
```diff
@@ -87,27 +87,34 @@ void PnormComponent::Write(std::ostream &os, bool binary) const {
...
+                 << Type() << ": \"" << cfl->WholeLine() << "\"";
+  Init(dim, dropout_proportion, dropout_per_frame);
 }
 
 std::string DropoutComponent::Info() const {
   std::ostringstream stream;
   stream << Type() << ", dim=" << dim_
-         << ", dropout-proportion=" << dropout_proportion_;
+         << ", dropout-proportion=" << dropout_proportion_
+         << ", dropout-per-frame=" << dropout_per_frame_;
```
danpovey: you can't do << to print out a bool, it will print 0 or 1.

src/nnet3/nnet-simple-component.cc
```diff
@@ -150,11 +170,26 @@ void DropoutComponent::Backprop(const std::string &debug_info,
 
 
 void DropoutComponent::Read(std::istream &is, bool binary) {
-  ExpectOneOrTwoTokens(is, binary, "<DropoutComponent>", "<Dim>");
-  ReadBasicType(is, binary, &dim_);
-  ExpectToken(is, binary, "<DropoutProportion>");
-  ReadBasicType(is, binary, &dropout_proportion_);
-  ExpectToken(is, binary, "</DropoutComponent>");
+  // back-compatibility code.
```
danpovey: remove this comment, it's not accurate (1st token may be missing due to how components are read in, not back compatibility).

src/nnet3/nnet-simple-component.cc
GaofengCheng: @danpovey fixed

Issue #1315, opened 2017-01-04
Remove sinusoid detection code; remove old SGMM code


danpovey:
I don't think the sinusoid-detection code is being used anywhere so I think it makes sense to remove it.
Also I don't believe the "old" sgmm code (as opposed to sgmm2, which is slightly better) has any reason to still be in the repository.  I'd like to remove that too.
I'm putting this up as an issue in case there are any objections.

danpovey:
@LvHang, can you please work on this?  Start with removing the sgmm and sgmmbin directories, detect-sinusoids.cc and sinusoid-detection.{h,cc}.  You'll have to update the Makefiles.
Check that things still compile, and that they still compile even after doing 'make clean'
Search in the scripts for things that call any of the sgmm binaries, and remove them also.  They'll have sgmm in the name but not sgmm2.  E.g. you can find candidates with something like
```
find egs -name '*sgmm*' -not -name '*sgmm2*' -print
```
but be careful, sometimes we have a script like run_sgmm.sh but it actually does the sgmm2 training.  We want to keep the sgmm2 code and scripts.

LvHang:
Ok，get it. I'll deal with it.

mpenagar:
Timit recipe (and I think that many other) use init-ubm command in sgmmbin:

$ cat exp/ubm4/log/cluster.log
# init-ubm --intermediate-num-gauss=2000 --ubm-num-gauss=400 --verbose=2 --fullcov-ubm=true exp/tri3_ali/final.mdl exp/tri3_ali/final.occs exp/ubm4/0.ubm
# Started at Wed Jan 11 18:16:33 CET 2017
#
bash: line 1: init-ubm: command not found

How should it be replaced? Thanks!

ngoel17:
Dan,
   I meant to visit the sinusoid detection code some point of time and
determine its usability for DTMF and ring tones. Can you please point me to
this.

Nagendra

LvHang:
Hi Dan,
The code--"init-ubm" is originally in sgmmbin/ directory. I check the dependencies of it. I think it is not relate to sgmm/. At the same time, it will be called by some scripts, such as "train_ubm.sh". Maybe we can move it to another directory, such as gmmbin/. Which one is better?

Hang

LvHang:
For sinusoid detection code, I think the code isn't be used anywhere in our standard scripts. I think this should be some personal needs. 

Hang

danpovey:
[...]

PR #1329, opened 2017-01-10
Remove sinusoid detection code and old SGMM code/scripts

LvHang:
@danpovey 
As your guidance, I modify something about #1315  
1) I remove the sinusoid-detection{-test}.{h,cc}  in src/feat and detect-sinusoids.cc in src/featbin and the sgmm/sgmmbin directories. At the same time, I fixed the makefile in the src/, src/feat/ and src/featbin. I try to make and "make clean; make". All goes smoothly.
2) I remove the scripts which call sgmm binaries. After generate the candidates with command "find egs -name '*sgmm*' -not -name '*sgmm2*' -print", I check each file and remove the files which call old sgmm.
3) I check whether the deleted files are called by other top scripts one by one and modify them.
Please Check it. Thanks

Best Wishes,
Hang


danpovey:
Build error, check the dependencies in the Makefile:
g++-4.9: error: ../sgmm//libkaldi-sgmm.so: No such file or directory

LvHang:
@danpovey 
Hi Dan,

I fix the bug--delete the ../sgmm//libkaldi-sgmm.so in the Makefile of decoder/.
After the update and git push. I download the newest codes from my remote repository and compile them. All passed.
Please check it. Thanks.

Best wishes,
Hang

danpovey:
Have you grepped in all the run.sh files for 'sgmm' to see if there are commented-out invocations to the SGMM scripts that you need to delete?

LvHang:
@danpovey 
Hi Dan,

I run the command--" find ./ -name '*run.sh*' -print | xargs grep -n 'sgmm' > info" and check the output file line by line. Now I think all in the run.sh isn't about sgmm.
In addition, I run the command--" grep -r 'sgmm' ./* " in both src/ and egs/ directories, and check all lines. I try my best to modify all of them. I hope I have fixed all about old sgmm.
Please check it. Thanks a lot.

Best wishes,
Hang

PR #1333, opened 2017-01-11
nnet1: replacing 'bc' by 'awk' in steps/nnet/train_scheduler.sh

vesis84:


danpovey:
Interesting that awk doesn't attempt to read stdin if there are no per-line or end commands, only a BEGIN one.

vesis84:
Yes, if the print was not in the BEGIN{...} section the awk would hang 
by waiting for 'stdin'.
There was the 'awk' originally, then year ago I changed it to 'bc' as it 
looked to me more
elegant, but seemingly less standard, now the circle has been completed 
by returning to 'awk' ;)

K.

PR #1337, opened 2017-01-13
Change default return value of install_sacc.sh

### LINKED ISSUE #1336: install_sacc.sh is returning an error code of 1 on success
The install script for SAcC is returning an error code of 1 when installation is successful.

I am not sure if this on purpose, but it cause headaches when trying to automate deployment (see for example https://github.com/tomjorquera/docker-kaldi/blob/master/extra/install_kaldi_extras.sh#L16-L18 for my personal encouter with the issue).

It seems it would be more sane for the script to return a value of 0 on success.

tomjorquera:
The install script for SAcC was returning an error code of 1 on a
successfull install.

Change it to the more "standard" return value of 0.

Fixes #1336.

### DIFF
--- a/tools/extras/install_sacc.sh
+++ b/tools/extras/install_sacc.sh
@@ -71,4 +71,4 @@ cd SAcC_GLNXA64
   && echo "**Error testing SAcC-- something went wrong." && exit 1;
 
 echo "Test succeeded."
-exit 1;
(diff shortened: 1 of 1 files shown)


danpovey:
Merging the change.  This script is not used in any up-to-date recipes though.

PR #1338, opened 2017-01-13
Update utils.py

pegahgh:


danpovey:
just curious, what was this needed for?

pegahgh:
It is needed  for e.g. input options with offsets
Append(input@-1,input,input@1)
Availability of this option is mentioned in comments.

PR #1342, opened 2017-01-15
Kaldi 5.1

danpovey:
This commit combines a bunch of changes to nnet3 (see also kaldi-asr.org/doc/versions.html for more info).

  - c++11-only code and use newer version of OpenFst.
  - faster compilation
  - ability separately specify extra-left-context for begin/end of utt
  - allow a variety of egs sizes in training
  - looped computation [for faster online computation, and allow infinite left-context]
  - introduce scaling factor into recurrence of LTSTMs where we add the old 'c', to make LSTM tolerant of seeing more context than it saw in training time.
  - refactor discriminative training scripts, removing a bunch of options and combining the lattice-generation and egs-dumping processes into one to avoid filling up the disk.


danpovey:
egs/tedlium/s5_r2/local/chain/tuning/run_tdnn_lstm_1k.sh

danpovey:
I am planning to merge this to master today.  I have created a 5.0 branch with the current state of master.

dogancan:
Do you want to merge the travis clang changes as well? I think the clang build is about 20% faster overall. It could be more but hard to say with the current variance in travis build times. I can rebase that PR and get rid of the test commits if you want to merge it today.

danpovey:
Could most of the variance in build times be due to the double-precision
randomization?
I just merged 5.1 so we can make this a separate issue.

dogancan:
That's what I thought at first as well but I don't think that is the case. Let me give it another look.

Issue #1340, opened 2017-01-15
Removing SetZero()


danpovey:
@freewym will do this.

I'd like to remove the SetZero() function from the UpdatableComponent and replace it with the SetAsGradient() function that will set the is_gradient_ flag and will set learning_rate_ to 1.0.  This function can be implemented *only* at the UpdatableComponent level, child classes will not have to override it.
Then all invocations of
`  SetZero(true)`
would be replaced by
```   Scale(0.0);
   SetAsGradient();```
and invocations of 
` SetZero(false);`
would be replaced by
`   Scale(0.0);`
The 'scale' functions will have to treat 0 as a special case, and call SetZero() on the matrices and vectors and set any floating-point counts to 0 directly, to avoid problems with inf/nan.
So code would be modified in the Scale() functions, but the individual overrides of SetZero() could be removed.

The function:
`void SetZero(bool is_gradient,
             Nnet *nnet);'
will be removed and invocations of it will be replaced by a call to:
    Scale(0.0, nnet);

It might actually be best if you make this a PR to the 'shortcut' branch.  This is still being actively developed but will be merged in the next 2-3 weeks.
I have just pushed the branch to the official Kaldi repo.

kkm000:
Done in #1343

PR #1343, opened 2017-01-16
Removing SetZero()

freewym:
resloved issue #1340.

Will run a tdnn recipe for test

### DIFF
--- a/src/nnet3/nnet-chain-training.cc
+++ b/src/nnet3/nnet-chain-training.cc
@@ -38,9 +38,7 @@ NnetChainTrainer::NnetChainTrainer(const NnetChainTrainingOptions &opts,
   KALDI_ASSERT(opts.nnet_config.momentum >= 0.0 &&
                opts.nnet_config.max_param_change >= 0.0);
   delta_nnet_ = nnet_->Copy();
-  bool is_gradient = false;  // setting this to true would disable the
-                             // natural-gradient updates.
-  SetZero(is_gradient, delta_nnet_);
+  ScaleNnet(0.0, delta_nnet_);
   const int32 num_updatable = NumUpdatableComponents(*delta_nnet_);
   num_max_change_per_component_applied_.resize(num_updatable, 0);
   num_max_change_global_applied_ = 0;
@@ -201,7 +199,7 @@ void NnetChainTrainer::UpdateParamsWithMaxChange() {
     if (param_delta > nnet_config.max_param_change) {
       if (param_delta - param_delta != 0.0) {
         KALDI_WARN << "Infinite parameter change, will not apply.";
-        SetZero(false, delta_nnet_);
+        ScaleNnet(0.0, delta_nnet_);
       } else {
         scale *= nnet_config.max_param_change / param_delta;
         num_max_change_global_applied_++;
--- a/src/nnet3/nnet-component-itf.h
+++ b/src/nnet3/nnet-component-itf.h
@@ -398,14 +398,14 @@ class UpdatableComponent: public Component {
   /// Sets the learning rate directly, bypassing learning_rate_factor_.
   virtual void SetActualLearningRate(BaseFloat lrate) { learning_rate_ = lrate; }
 
+  /// \brief Sets is_gradient_ to true and sets learning_rate_ to 1, ignoring
+  /// learning_rate_factor_.
+  virtual void SetAsGradient() { learning_rate_ = 1.0; is_gradient_ = true; }
+
   /// Gets the learning rate of gradient descent.  Note: if you call
   /// SetLearningRate(x), and learning_rate_factor_ != 1.0,
   /// a different value than x will returned.
   BaseFloat LearningRate() const { return learning_rate_; }
-
-  /// \brief Sets is_gradient_ to true and sets learning_rate_ to 1, ignoring
-  /// learning_rate_factor_.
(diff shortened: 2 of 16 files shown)


src/nnet3/nnet-chain-training.cc
```diff
@@ -38,9 +38,11 @@ NnetChainTrainer::NnetChainTrainer(const NnetChainTrainingOptions &opts,
   KALDI_ASSERT(opts.nnet_config.momentum >= 0.0 &&
                opts.nnet_config.max_param_change >= 0.0);
   delta_nnet_ = nnet_->Copy();
+  ScaleNnet(0.0, delta_nnet_);
   bool is_gradient = false;  // setting this to true would disable the
                              // natural-gradient updates.
-  SetZero(is_gradient, delta_nnet_);
+  if (is_gradient)
```
danpovey: we just set is_gradient = false, so there is no need to call SetAsGradeintNnet().
Sometimes I declare a variable is_gradient = false just before SetZero() to document what the 'false' means.

src/nnet3/nnet-component-itf.h
```diff
@@ -408,6 +403,10 @@ class UpdatableComponent: public Component {
   /// a different value than x will returned.
   BaseFloat LearningRate() const { return learning_rate_; }
 
+  /// \brief Sets is_gradient_ to true and sets learning_rate_ to 1, ignoring
+  /// learning_rate_factor_.
+  void SetAsGradient() { SetActualLearningRate(1.0); is_gradient_ = true; }
```
danpovey: i think doing `learning_rate_ = 1.0;` might be clearer than SetActualLearningRate().

src/nnet3/nnet-optimize-test.cc
```diff
@@ -97,7 +97,9 @@ static bool UnitTestNnetOptimizeWithOptions(int32 srand_seed,
   bool is_gradient = true;  // with natural gradient, the consolidation would
   // affect the final model params -> test just the
   // gradient.
-  SetZero(is_gradient, &nnet_to_update);
+  ScaleNnet(0.0, &nnet_to_update);
+  if (is_gradient)
```
danpovey: again, you can simplify this code.

src/nnet3/nnet-optimize-test.cc
```diff
@@ -107,7 +109,9 @@ static bool UnitTestNnetOptimizeWithOptions(int32 srand_seed,
   Nnet nnet_opt(nnet);  // copy of the nnet for the optimized computation.
   // necessary in case backprop changes parameters.
   Nnet nnet_opt_to_update(nnet_opt);
-  SetZero(is_gradient, &nnet_opt_to_update);
+  ScaleNnet(0.0, &nnet_opt_to_update);
+  if (is_gradient)
+    SetAsGradientNnet(&nnet_opt_to_update);
```
danpovey: .. and here too, is_gradient has not changed.

src/nnet3/nnet-simple-component.cc
```diff
@@ -1136,8 +1136,13 @@ void RectifiedLinearComponent::StoreStats(
 }
 
 void AffineComponent::Scale(BaseFloat scale) {
-  linear_params_.Scale(scale);
-  bias_params_.Scale(scale);
+  if (scale == 0.0) {
```
danpovey: you could comment:
// If scale == 0.0 we call SetZero() which will get rid of NaN's and inf's.

src/nnet3/nnet-simple-component.cc
```diff
@@ -4780,8 +4751,19 @@ std::string CompositeComponent::Info() const {
 
 // virtual
 void CompositeComponent::Scale(BaseFloat scale) {
-  for (size_t i = 0; i < components_.size(); i++)
-    components_[i]->Scale(scale);
+  if (scale == 0.0) {
```
danpovey: I don't think you have to modify the code of this function, just leave it as it is.

src/nnet3/nnet-utils.cc
```diff
@@ -280,10 +261,31 @@ void SetLearningRate(BaseFloat learning_rate,
   }
 }
 
+void SetAsGradientNnet(Nnet *nnet) {
```
danpovey: could you please rename this function to `SetNnetAsGradient()`?

src/nnet3/nnet-utils.cc
```diff
@@ -280,10 +261,31 @@ void SetLearningRate(BaseFloat learning_rate,
...
+      u_comp->SetAsGradient();
+    }
+  }
+}
+
 void ScaleNnet(BaseFloat scale, Nnet *nnet) {
   if (scale == 1.0) return;
   else if (scale == 0.0) {
-    SetZero(false, nnet);
+    for (int32 c = 0; c < nnet->NumComponents(); c++) {
```
danpovey: I don't think we need to have a special case here any more, for scale == 0.0.
Just make sure that all the components that store stats [which do all support the
Scale() function], have a Scale() function that really sets everything to zero if
called with 0.0.

src/nnet3/nnet-component-itf.h
freewym: SetActualLearningRate() is a virtual function that was overwritten in CompositeComponent, so we have to call it here.

danpovey:
I think CompositeComponent would have to override SetAsGradient() anyway,
to call the same function on all the owned components.

Issue #1339, opened 2017-01-14
cu-rand issues


danpovey:
@vesis84, do you have time to help with some cu-rand issues?

Firstly, I think it would be convenient to have the RandUniform() function work for CuVector also.

Also there are some optimizations that I think would be nice and a concern about safety that I have...

If the RandUniform()/RandGaussian() functions had different implementations for CuMatrix and CuSubMatrix, it would make the code more optimizable, as we know for a CuMatrix that the space between the strides is owned by us [well, modulo issues of how cudaMallocPitch() works.]

Also-- the even/odd thing only matters for Gaussian, not uniform, generation: search for 'even' here:
http://docs.nvidia.com/cuda/curand/group__HOST.html#group__HOST_1gb9280e447ef04e1dec4611720bd0eb69

I think it would be safer when allocating the 'tmp' matrix and dealing with the 'rounding-up' issue, to round up the numCols() to make it even when allocating the 'tmp' matrix, and then when calling curandGeneratexxx, to call it with the exact number of bytes from the first to the last [row,col].  I don't know if it's documented that when you call cudaMallocPitch it lets you own the memory past the last column up to the stride.  And I'm also a bit concerned what happens with matrices with one row.  The cu-allocator code just calls regular cudaMalloc if the matrix has one row, and there is no kind of padding, so this code would write one past the end.

vesis84:
Hi, okay, I'll take a look at it...

PR #1352, opened 2017-01-18
Fixing 'curand' issues #1339

vesis84:
added VectorBase::SetRandUniform(), CuVectorBase::SetRandUniform(), 

cudamatrix,curand: added wrapper template functions for 'curand' calls
to support both float and double by 1 function

cudamatrix,curand: refactored the code
- in 'SetRandn' the 'even' number of elements is done by appending a column to a matrix,
- optimized the code for cases solvable without 'tmp' matrix,
- updated the unit test, testing both with CuMatrix and CuMatrixBase,
  testing with various column number.

The unit tests in src/{cudamatrix,chain} and src/nnet{ ,2,3} passed OK...

Also the RBMs are training normally, but it would be wise to test it on some other task, before the merge is done (with dropout or similar).

### DIFF
--- a/src/cudamatrix/cu-rand-speed-test.cc
+++ b/src/cudamatrix/cu-rand-speed-test.cc
@@ -56,63 +56,166 @@ std::string MeanVariance(const CuMatrixBase<Real>& m) {
   return std::string("mean ") + ToString(mean) + ", std-dev " + ToString(std::sqrt(var));
 }
 
+template<typename Real>
+std::string MeanVariance(const CuVectorBase<Real>& v) {
+  std::ostringstream os;
+  Real mean = v.Sum() / v.Dim();
+  CuVector<Real> tmp(v);
+  tmp.Add(-mean);
+  tmp.ApplyPow(2.0);
+  Real var = tmp.Sum() / tmp.Dim();
+  return std::string("mean ") + ToString(mean) + ", std-dev " + ToString(std::sqrt(var));
+}
+
+
 template <typename Real>
-void CuRandUniformMatrixSpeedTest() {
+void CuRandUniformMatrixSpeedTest(const int32 iter) {
   Timer t;
   CuRand<Real> rand;
-  CuMatrix<Real> m(249,2011);
-  for (int32 i = 0; i < 200; i++) {
+  CuMatrix<Real> m(249,1001, kUndefined);
+  for (int32 i = 0; i < iter; i++) {
     rand.RandUniform(&m);
   }
-  KALDI_LOG << __func__ << NameOf<Real>() << " t = " << t.Elapsed() << "s, " << MeanVariance(m);
+  CuMatrix<Real> m2(256,1024, kUndefined);
+  for (int32 i = 0; i < iter; i++) {
+    rand.RandUniform(&m2);
+  }
+  // flops = number of generated random numbers per second,
+  Real flops = iter * (m.NumRows() * m.NumCols() + m2.NumRows() * m2.NumCols()) / t.Elapsed();
+  KALDI_LOG << __func__ << NameOf<Real>()
+            << " Speed was " << flops << " rand_elems/s. "
+            << "(debug " << MeanVariance(m) << ")";
 }
 
 template <typename Real>
-void CuRandGaussianMatrixSpeedTest() {
+void CuRandUniformMatrixBaseSpeedTest(const int32 iter) {
   Timer t;
   CuRand<Real> rand;
-  CuMatrix<Real> m(249,2011);
-  for (int32 i = 0; i < 200; i++) {
+  CuMatrix<Real> m(249,1001, kUndefined);
+  for (int32 i = 0; i < iter; i++) {
+    rand.RandUniform(dynamic_cast<CuMatrixBase<Real>*>(&m));
+  }
+  CuMatrix<Real> m2(256,1024, kUndefined);
+  for (int32 i = 0; i < iter; i++) {
+    rand.RandUniform(dynamic_cast<CuMatrixBase<Real>*>(&m2));
+  }
+  // flops = number of generated random numbers per second,
+  Real flops = iter * (m.NumRows() * m.NumCols() + m2.NumRows() * m2.NumCols()) / t.Elapsed();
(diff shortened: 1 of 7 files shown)


danpovey:
Cool.
Would you mind modifying the cu-rand speed tests to be faster?
Right now it's a bit slow:
Running cu-rand-speed-test ... 59s... SUCCESS
[from the travis log],
and we've been having challenges keeping the travis build within a
reasonable time.  For tests where the amount of code being tested is
relatively limited (as here), I prefer them to finish within about 1 to 3
seconds.
Some of the other speed tests will put a limit on the elapsed time, and
break from the loop when it hits a certain limit [this limits how many
times the test is repeated on CPU].  You can then compute a speed by
dividing by the number of times you ran the test [or some kind of fake
flops performance by dividing by the size of the problem].

Dan

vesis84:
Hi Dan, the unit test is faster now:
- The CPU/GPU 'for loop' is disabled if HAVE_CUDA != 1 (this should be done for other speed tests too, otherwise it may run twice the same test on CPU leading to slower Travis),
- the CPU iterations are limited to 10,
- the measuered speed unit is rand_elem/s (# of generated random numbers per second).
- the unit test took 3 seconds on my workstation.

danpovey:
Thanks!
I'll merge this soon, by hand, to the 'shortcut' branch to avoid the need to test too carefully.
@keli78, can you please have a look at what Karel did to ensure the test only runs once if there is no GPU, and do the same thing for the other tests in cudamatrix/?  Do this starting from the 'shortcut' branch, not the master.

danpovey:
Actually I'll merge this to the master, because IIRC the 'dropout' PR was against the master, and this is needed for that.

danpovey:
@keli78, for the other tests you don't have to add the 'iter' variable, just do an #ifdef that will ensure the loop only runs once if no GPU.

src/cudamatrix/cu-rand-speed-test.cc
```diff
@@ -56,63 +56,166 @@ std::string MeanVariance(const CuMatrixBase<Real>& m) {
...
+            << " Speed was " << flops << " rand_elems/s. "
+            << "(debug " << MeanVariance(v) << ")";
 }
 
 } // namespace kaldi
 
 
 int main() {
-  for (int32 loop = 0; loop < 2; loop++) {
+  int32 iter = 10; // Be quick on CPU,
```
keli78: As for "have a look at what Karel did to ensure the test only runs once if there is no GPU, and do the same thing for the other tests in cudamatrix/?", I think the variable iter should be changed from 10 to 1 in this line, right?

keli78: As for "the other tests in cudamatrix/", do you mean I should modify all the other cu-*-test.cc files? But I think that these test files only run the test once if no GPU. So I'm not sure what do you mean by "just do an #ifdef that will ensure the loop only runs once if no GPU".

danpovey:
It's not about the 'iter' stuff; the other tests won't have any 'iter'
variable, it's about how the loop works; make sure it only runs once, not
twice, when USE_CUDA is not defined.

vesis84:
Yes, what I did is following: The #ifdef is placed in such way that if 'HAVE_CUDA != 1', the for loop vanishes. And all the tests are done just once... The example is here: https://github.com/kaldi-asr/kaldi/blob/master/src/cudamatrix/cu-rand-speed-test.cc

keli78:
Got it. Thanks for the explanation. Will do it soon.

keli78:
Hi Dan, after I checkout the "shortcut" branch and compiled Kaldi, I did "make test" in cudamatrix/. While almost all tests failed: 
Running cu-vector-test .../bin/bash: line 1: 32485 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-vector-test
Running cu-matrix-test .../bin/bash: line 1: 32582 Terminated              ./$x > $x.testlog 2>&1
 6s... FAIL cu-matrix-test
Running cu-math-test .../bin/bash: line 1: 32640 Terminated              ./$x > $x.testlog 2>&1
 24s... FAIL cu-math-test
Running cu-test .../bin/bash: line 1: 32867 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-test
Running cu-sp-matrix-test .../bin/bash: line 1: 32887 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-sp-matrix-test
Running cu-packed-matrix-test .../bin/bash: line 1: 32926 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-packed-matrix-test
Running cu-tp-matrix-test .../bin/bash: line 1: 32946 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-tp-matrix-test
Running cu-block-matrix-test .../bin/bash: line 1: 32971 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-block-matrix-test
Running cu-matrix-speed-test .../bin/bash: line 1: 33086 Terminated              ./$x > $x.testlog 2>&1
 39s... FAIL cu-matrix-speed-test
Running cu-vector-speed-test .../bin/bash: line 1: 33332 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-vector-speed-test
Running cu-sp-matrix-speed-test ... 1s... SUCCESS
Running cu-array-test ... 0s... SUCCESS
Running cu-sparse-matrix-test ... 1s... SUCCESS
Running cu-device-test .../bin/bash: line 1: 33406 Terminated              ./$x > $x.testlog 2>&1
 4s... FAIL cu-device-test
Running cu-rand-speed-test .../bin/bash: line 1: 33480 Terminated              ./$x > $x.testlog 2>&1
 2s... FAIL cu-rand-speed-test
../makefiles/default_rules.mk:81: recipe for target 'test' failed
make: *** [test] Error 1

Does this make sense? Do I need to figure out the reasons that why these tests fail before I modify the  code to guarantee tests only run once if no GPU?

danpovey:
They failed because you did not reserve a GPU via qlogin and they got
killed.
Check your inbox.

PR #1354, opened 2017-01-18
totally removed truncate_deriv_weights option from chain train scripts

freewym:


danpovey:
I think I have already done this.  However for  some reason your changes still show up in the diff, even though locally I can't see that option any more and I pushed the changes.  Possible github glitch. I'll wait and see if it resolves, should be a null change.

PR #1282, opened 2016-12-21
Babel s5d to pr 

jtrmal:
It looks like a huge PR (and it is), however there is a couple of things that might help reduce the complexity:
most of the changes are contrained to egs/babel, i.e. benign for other setups

* there are some added executables to src/
* small changes to egs/wsj/steps (namely the keyword search scripts)
* lots of whitespace changes to egs/babel/s5c (I think those came from Vijay), no feature added
* egs/babel/s5d/ (the final babel year recipe) -- the actual content with reworked keyword search pipeline

what works for me is too find "egs/wsj" in the "Files changed" tab -- it should show around 14 files changed, most of it is just whitespaces... Similarly for "src/" -- should show around 10 files.

PR #1356, opened 2017-01-19
Babel s5d recipe

jtrmal:
replaces #1282 
couple of fixes as reported by Fred.
rebased
I did a new PR instead of rebasing the old one and doing push -f, as other people are probably using the branch and I don't want to mess up their clones in case they do pull.


src/kwsbin/print-proxy-keywords.cc
```diff
@@ -0,0 +1,129 @@
+// kwsbin/print-proxy-keywords.cc
+//
+// Copyright 2014  Johns Hopkins University (Author: Guoguo Chen)
```
danpovey: I assume this header is wrong?

src/kwsbin/print-proxy-keywords.cc
```diff
@@ -0,0 +1,129 @@
...
+    typedef kaldi::uint64 uint64;
+    typedef StdArc::StateId StateId;
+    typedef StdArc::Weight Weight;
+
+    const char *usage =
+        "Reads in the proxy keywords FSTs and print them to a file where each\n"
+        "line is \"kwid weight proxies\"\n"
+        "\n"
+        "Usage: print-proxy-keywords [options] <proxy-rspecifier> \\\n"
+        "                                      <kwlist-wspecifier>\n"
```
danpovey: should indicate the optional weights as:
`                                                [<weights-wspecifier>]`
here and remove the [] in the "e.g." line.

src/kwsbin/print-proxy-keywords.cc
```diff
@@ -0,0 +1,129 @@
...
+        " e.g.: print-proxy-keywords ark:proxy.fsts ark,t:kwlist.txt [ark,t:weights.txt]\n";
+
+    ParseOptions po(usage);
+
+    po.Read(argc, argv);
+
+    if ((po.NumArgs() < 2) && (po.NumArgs() > 3)) {
```
danpovey: && should be ||, and no need for parentheses.

src/kwsbin/print-proxy-keywords.cc
```diff
@@ -0,0 +1,129 @@
...
+      KALDI_ASSERT(paths.size() == weights.size());
+      for (int32 i = 0; i < paths.size(); i++) {
+        vector<int32> kwlist;
+        vector<double> weight;
+        weight.push_back(weights[i].Value());
+        for (int32 j = 0; j < paths[i].size(); j++) {
+          kwlist.push_back(paths[i][j]);
+        }
+        kwlist_writer.Write(key, kwlist);
+        weight_writer.Write(key, weight);
```
danpovey: if weight_wspecifier is really optional, then you should do:
` if (weight_wspecifier != "") `
here.

src/kwsbin/transcripts-to-fsts.cc
```diff
@@ -46,21 +60,34 @@ int main(int argc, char *argv[]) {
```
danpovey: In the usage message at line 51, need angle brackets around command line args, and need to indicate the optional weights-rspecifier using
        `[<weights-rspecifier>]`
there should probably be some explanation there of what the weights are.  
Actually I am not really happy about it being called "weights-rspecifier" if they actually represent
the minus-log of the weight, because "weights" on Kaldi command lines tend to refer to things
that should be >= 0, i.e. actual weights.  I'd prefer to see it called "costs-rspecifier" and the
variable be called a "cost".

src/kwsbin/transcripts-to-fsts.cc
```diff
@@ -81,6 +108,15 @@ int main(int argc, char *argv[]) {
 
       VectorFst<StdArc> fst;
       MakeLinearAcceptor(transcript, &fst);
+      if (weights_rspecifier != "" ) {
+        std::string weights_key = weights_reader.Key();
+        double weight = weights_reader.Value();
+        weights_reader.Next();
+
+        KALDI_ASSERT(weights_key == key);
+
```
danpovey: This code is a bit ugly, and I think unnecessary... the weights_reader should be made a RandomAccessDoubleReader, and you can use HasKey(key) and Value(key).  ,s,cs can be added on the command line to make it efficient, if they are actually sorted.  If they are not sorted (and this might relate to the special things that are done in the keyword-search pipeline), then we'll just take the memory hit.  I prefer tools like this to be general purpose.

jtrmal:
Thanks,  I'll address those.
Y.

jtrmal:
@danpovey this is open for review again

src/fstbin/fsts-project.cc
```diff
@@ -0,0 +1,80 @@
...
+    typedef kaldi::int32 int32;
+    typedef kaldi::uint64 uint64;
+
+    const char *usage =
+        "Reads a table of FSTs; for each element, performs the fst project\n"
+        "operation either on input (default) or on the output (if the option\n"
+        "--project-output is true).\n"
+        "\n"
+        "Usage: fsts-project [options] <fsts-rspecifier> <fsts-wspecifier>\n"
+        " e.g.: fsts-project ark:train.fsts ark,t:train.fsts\n";
```
danpovey: can we have a "see also:" line?  Like:
"see also: fstproject (from the OpenFst toolkit)"
and the same for similar tools?

jtrmal:
Ok.  Makes sense.
Y.

jtrmal:
ok, done.

tools/extras/check_dependencies.sh
```diff
@@ -41,10 +41,6 @@ if ! which libtoolize >&/dev/null && ! which glibtoolize >&/dev/null; then
   add_packages libtool libtool libtool
 fi
 
-if ! which svn >&/dev/null; then
```
danpovey: Did you intend to remove this?  Some recipes use it to download CMUDict.

jtrmal:
added back. I remember we were talking about it a couple of months back and
you were in favor of keeping it there. Sorry about this, just sneaked under
my radar.
y.

PR #1357, opened 2017-01-19
rnnlm-fix-style

keli78:
revert from int64 to long long and uint64 to unsigned long long

### DIFF
--- a/src/lm/arpa-file-parser-test.cc
+++ b/src/lm/arpa-file-parser-test.cc
@@ -220,18 +220,18 @@ ngram 2 = 2\n\
 ngram 3 = 2\n\
 \n\
 \\1-grams: \n\
--5.2 a -3.3\n\
--3.4 \xCE\xB2\n\
-0.0 <s> -2.5\n\
--4.3 </s>\n\
+-5.2	a -3.3\n\
+-3.4	\xCE\xB2\n\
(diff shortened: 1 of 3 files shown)


src/lm/arpa-file-parser-test.cc
```diff
@@ -220,18 +220,18 @@ ngram 2=2\n\
 ngram 3=2\n\
 \n\
 \\1-grams: \n\
--5.2    a -3.3\n\
--3.4    \xCE\xB2\n\
-0.0     <s> -2.5\n\
--4.3    </s>\n\
+-5.2	a -3.3\n\
```
danpovey: did you replace tabs with spaces or vice versa?  I can't tell from github.  These should be tabs; I believe the ARPA format specifies tabs at this point.

danpovey: actually, tabs should never appear in source-code files; they should all be \t instead of an explicit tab or a set of spaces.    The format is:
`prob [tab] n-gram`
or:
`prob [tab] n-gram [tab] backoff-prob`

keli78: The original version uses tab instead of \t or space. Do you want me to change it to \t?

danpovey:
yes change it, use \t

PR #1359, opened 2017-01-20
Enable Travis CI on the 'shortcut' branch

kkm000:


danpovey:
I merged, but I wonder whether it needs to be merged in the 'master' branch in order to be effective-- not sure at what stage this file is inspected.

kkm000:
Yes, that's confusing. But if I understand how Travis works, all they immediately see is the config file in the set of files they are testing. But it might be more complex than that, indeed.

May be it would be better to remove the branch restriction at all? It is also possible to exclude specific branches.

PR #1365, opened 2017-01-22
make sure loop run once if no GPU in tests in cudamatrix/

keli78:
make sure loop run once if no GPU in tests in cudamatrix/

have tested the code both with and without GPUs and the test results are expected.


danpovey:
This needs to be submitted to the 'shortcut' branch, not master (since you took the code from 'shortcut')...

danpovey:
.. and I don't see any commits from you in this branch; looks like just the 'shortcut' branch with no extra commits.

PR #1366, opened 2017-01-22
modify test code to guarantee loop only run once if no GPU

keli78:
have tested both with and without GPUs and the results are what we expect.

### DIFF
--- a/src/cudamatrix/cu-array-test.cc
+++ b/src/cudamatrix/cu-array-test.cc
@@ -116,8 +116,9 @@ static void UnitTestCuArray() {
 
 
 int main() {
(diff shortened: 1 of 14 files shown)


keli78:
@danpovey The travis CI build failed but there was no problem on my local branch. Not sure what's the reason for this.

danpovey:
Looks like a problem with travis itself.
Make a small change (e.g. add a newline) and push it, it will cause it to
re-build.

keli78:
Hi @danpovey the checks have passed.

danpovey:
ok thanks, merging.

PR #1368, opened 2017-01-23
change args of generate_plots.py to take a list of exp dirs

keli78:


egs/wsj/s5/steps/nnet3/report/generate_plots.py
```diff
@@ -52,9 +52,6 @@ def get_args():
         --comparison-dir exp/nnet3/tdnn1 --comparison-dir exp/nnet3/tdnn2 \\
         exp/nnet3/tdnn exp/nnet3/tdnn/report""")
 
-    parser.add_argument("--comparison-dir", type=str, action='append',
```
danpovey: It needs to be back-comptiable to the old usage method, but you can mention that this option is deprecated.

egs/wsj/s5/steps/nnet3/report/generate_plots.py
```diff
@@ -654,9 +653,13 @@ def main():
     else:
         output_nodes.append(('output', 'linear'))
 
-    generate_plots(args.exp_dir, args.output_dir, output_nodes,
-                   comparison_dir=args.comparison_dir,
-                   start_iter=args.start_iter)
+    if len(args.exp_dir) == 1: 
+      generate_plots(args.exp_dir[0], args.output_dir, output_nodes,
```
vijayaditya: Remember to specify in the help message that the first directory would be treated as the main experiment directory. It would be used to determine the components will have non-linearity and parameter stats plotted.

danpovey:
Vijay, you merge this if it looks OK.

vijayaditya:
I am assuming this is tested. LGTM.

danpovey:
Ke, did you actually test this?  Because it seems to be only generating a plot from one of the directories, when the directories are listed on the command line.

keli78:
hmm, I tried both the old (comparsion-dir arg) and the new args (no comparsion-dir) with three directories and checked the two generated plots. They work fine for me. Maybe the results in the directories you tested are too close?

danpovey:
Oh, now I try it again it works.. I must have mistyped the args somehow.

PR #1371, opened 2017-01-24
nnet3/report : Added support for parsing latest compute_{train,valid}…

vijayaditya:
…*.log

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
+++ b/egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
@@ -14,6 +14,18 @@ import libs.common as common_lib
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
+class KaldiLogParseException(Exception):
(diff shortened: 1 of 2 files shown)


danpovey:
Vijay, I  had a thought about plotting the avg-deriv stats, and the avg-value stats assuming they are currently being printed...
I think rather than plotting the mean and the mean +- stddev, it would more meaningful to plot the median (50th percentile) and the 5th and 95th percentile.  [note: these medians/percentiles are taken over the means of the individual dimensions/nodes of the layer].  This will reveal certain pathologies more clearly, i.e. it will clearly show if there are excessively over- or under-saturated units.  
What do you think about having @keli78 work on this?  It would require a certain amount of messing about with regexp's.

vijayaditya:
It should be pretty easy in the current setup. @keli78 might be able to do
this without any issues.

@Keli78 I can point you the methods which need to be changed for this.

--Vijay

keli78:
@vijayaditya Sure, I will have a try. The method I should change is generate_nonlin_stats_plots right?

PR #1373, opened 2017-01-24
fixing openfstwin patch for compat.h (Windows build)

pstanisl:
Source `compat.h` already contains checking of `_MSC_VER < 1900` and consequently it is not possible to apply the patch.

### DIFF
--- a/tools/extras/openfstwin-1.3.4.patch
+++ b/tools/extras/openfstwin-1.3.4.patch
@@ -2,16 +2,6 @@ diff --git a/src/include/fst/compat.h b/src/include/fst/compat.h
 index 00e2dba..ff8bacc 100644
 --- a/src/include/fst/compat.h
 +++ b/src/include/fst/compat.h
-@@ -23,7 +23,9 @@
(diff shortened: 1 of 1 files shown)


jtrmal:
I guess you're right. Does that "modified" patch apply cleanly?

pstanisl:
Yes, I tested it on forked repo and it works just fine.

danpovey:
Merging, but we may have to revisit this stuff when we upgrade to Kaldi 5.1 (currently the 'shortcut' branch) which uses a different OpenFst version.

danpovey:
wait.. but we previously built on Windows.  @pstanisl, are you sure you hadn't already applied the patch.    I think `patch` takes an -N option to say don't apply if already applied.

jtrmal:
I actually checked the sources and it's there. So I'm not sure how come it
worked for me before, but this should work also.
y.

pstanisl:
I tried it with git patch tool and it was not able to patch `compat.h` because source code around patched lines were different. But I am not expert in patching, maybe I miss some option for patch patch tool. :)

PR #1377, opened 2017-01-25
Use correct tree directory in chain model training

entn-at:
The changed chain model training scripts used a hard-coded path including '_sp' suffix in stage 12 irrespective of whether speed_perturb is set. They should use $treedir as path instead, as the tree built in stage 11 is written to $treedir.


danpovey:
Thanks for the patch!  Merging.

PR #1380, opened 2017-01-26
[src]: Add function to freeze natural gradient [to be used in adversa…

danpovey:
…rial training]; minor bugfix.

@freewym, this is for you to merge with your adversarial training setup.
You can declare a function
`virtual void UpdatableComponent::FreezeNaturalGradient(bool freeze) { }`
and then override it to call `Freeze(freeze)` on all the objects of type OnlineNaturalGradient,
where applicable.  Don't forget CompositeComponent.  And you'd declare in nnet-utils.h:

`void FreezeNaturalGradient(bool freeze, Nnet *nnet);`

This will enable us to have control (in a non-hacky way) over when the NG is updated in the adversarial training.  E.g. you'd do:
```
FreezeNaturalGradient(true, delta_nnet_);
[do the adversarial step here.]
FreezeNaturalGradient(false, delta_nnet_);
```

### DIFF
--- a/src/nnet2/nnet-precondition-online.cc
+++ b/src/nnet2/nnet-precondition-online.cc
@@ -607,6 +607,7 @@ OnlinePreconditioner& OnlinePreconditioner::operator = (
   update_period_ = other.update_period_;
   num_samples_history_ = other.num_samples_history_;
   alpha_ = other.alpha_;
+  delta_ = other.delta_;
   epsilon_ = other.epsilon_;
   t_ = other.t_;
   self_debug_ = other.self_debug_;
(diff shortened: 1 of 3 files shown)


danpovey:
@freewym, I'm deleting this PR to avoid clutter, but the branch still exists.

danpovey:
@freewym, I realized there was a problem in the way I did this.  It would mean that the natural-gradient object was not really doing natural gradient for the 1st minibatch, which could potentially lead to problems.  The problem is fixed now [one-line fix.]

danpovey:
closing since @freewym has probably already merged the changes to his branch.

PR #1381, opened 2017-01-27
handle wav.scp entries that are filenames, not just pipes, in run_ivector_common.sh

jason-levy:
Adjust Python script in run_ivector_common.sh to work with entries in wav.scp that are filenames, not just pipes. Only tedlium/s5 version is changed here, but the problem seems common to all versions.


danpovey:
The scripts in local/ are supposed to be specialized for the different datasets, so it's not really a bug if they don't work on other data.
However, there is a script utils/data/perturb_data_dir_volume.sh that is supposed to be used instead of that inline python script, e.g.
```
 utils/data/perturb_data_dir_volume.sh $datadir
```
That script handles pipes correctly.
If you were to change all those scripts to use that command instead, I would merge that PR.
But closing this one as it's not quite the right fix.

PR #1385, opened 2017-01-30
learn_lexicon.sh: few more minor fixes

gorinars:
Sorry, a few more bugs

The help message states retrain_src_mdl=false by default, but it is not. It could be the other way around (modify help message), but seems like "false" was planned as the default behavior according to the parameter passed explicitly in kaldi/egs/tedlium/s5_r2/local/run_learn_lex.sh

Second fix is just to make sure the script does not crash due to a different phone map


danpovey:
Thanks!  Merging.

Issue #1370, opened 2017-01-23
Change how the --frame argument is chosen in TDNN training.


danpovey:
This is for @LvHang (please ask @vimalmanohar and @freewym for help as needed),

I'll illustrate the issue here with reference to `train_tdnn.sh`, but the code I want
changed is the python-script version of the code, as called by train_dnn.py.

The --frame option controls which of the (e.g. 8) supervised frames in an eg is chosen to train on.
currently it's chosen as follows:
```
        archive=$[($k%$num_archives)+1]; # work out the 1-based archive index.                                                                                        
        frame=$[(($k/$num_archives)%$frames_per_eg)]; # work out the 0-based frame                                                                                    
 ```
If we define one sub-epoch at the time it takes to process all the archives, then that means that for the 1st sub-epoch we use frame 0, for the 2nd sub-epoch we use frame 1, and so on.
The problem here is that for big data, a sub-epoch might be quite long, and there may be certain pdfs that only occur at the very start/end of a file.  These will learn rather biased likelihoods.

What would be better is to change it to the equivalent of:
```
        archive=$[(k%num_archives)+1]; # work out the 1-based archive index.                                                                                        
        frame=$[((k/num_archives+archive)%frames_per_eg)]; # work out the 0-based frame                                                                                    
```
(note: I removed the dollar signs as they are not needed inside `$[..]` expressions.. anyway you'll be changing the python version of the script).

Please test this in the 'shortcut' branch on, say, the tedlium s5_r2 nnet3 TDNNs (not chain- it has to be with train_dnn.py).
This may make zero difference but we should test it anyway and I will probably merge it even if it makes no difference.

LvHang:
Thanks. Get it.

kkm000:
Done in #1389

PR #1389, opened 2017-02-01
change --frame argument

LvHang:
@danpovey 
Related to #1370 . Attention, this is in "shortcut" branch, not "master" branch.

Hi Dan,
1. Following you interpretation, I change the code about how to compute the --frame option.
2. I rewrite the tedlium/s5_r2/local/nnet3/run_tdnn.sh so that it will call steps/nnet3/train_dnn.py to train the model. The annotation (line 78-88) is the original shell version. The code (line 90-130) is what I converted.
3. The WER result is followed:
                                 |   after-change  |    ori
decode-dev              |          11.7         |  12.6
decode_dev_rescore |         10.9          |  11.5
decode_test              |         11.8          |  11.9
decode_test_rescore |         11.2          |  10.9
(The result in "ori" column is extracted from "RESULT" file in tedlium/s5_r2)

Hang

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/common.py
@@ -71,7 +71,7 @@ def train_new_models(dir, iter, srand, num_jobs,
         archive_index = (k % num_archives) + 1
 
         if not chunk_level_training:
-            frame = (k / num_archives) % frames_per_eg
+            frame = (k / num_archives + archive_index) % frames_per_eg
 
         cache_write_opt = ""
         if job == 1:


danpovey:
Interesting.  I'm not sure what run_tdnn_py.sh was modified from; anyway it is not following the correct pattern of these filenames; get Yiming to explain the system to you.
I think it's better if you rerun the baseline; numbers in the RESULTS file may be out of date.
Also the system that you are using is not up-to-date at all (it uses the deprecated config-generating scripts) so I would not want to check this in.  But it doesn't matter for purposes of testing the change.  Just run it again without the change.  If the egs were not deleted, then re-use them via the common-egs-dir option.

LvHang:
1. The "run_tdnn_py.sh" was modified from "tedlium/s5_r2/local/nnet3/run_tdnn.sh". In the 12th stage of "run_tdnn.sh", it call the shell script--"steps/nnet3/tdnn/train.sh". But in #1370 , I think you let me modify the "train_dnn.py". So I rewrite it. I asked Yiming and move them to tuning directory.
2. I will rerun the baseline right now.
3. You said the config-generating scripts is out-of-date. However, I check the "master" and "shortcut" branch in your repository, all of them use the "steps/nnet3/tdnn/make_configs.py". I guess you mean we need to write xconfig and use "steps/nnet3/xconfig_to_configs.py", right?
Hang

danpovey:
yes it should use the xconfigs, that tdnn sript in tedlium was out of date

LvHang:
Hi Dan,
I have already run the baseline.The WER result is below:
　　　　　　　　　　| after-change |  baseline
decode-dev 　　　　　|　 11.7 　　| 11.7
decode_dev_rescore　 | 10.9 　　　| 11.0
decode_test 　　　　　| 11.8 　　　| 11.6
decode_test_rescore 　| 11.2 　　　| 11.0
Do you think this is what you expect?
If you think it's ok, I'll use the xconfigs.

danpovey:
I think those differences are just the normal differences we see when we rerun.
I'll merge the change but I don't need any example script, just leave the one-line change
of common.py.

LvHang:
Ok. I need to modify it on "master" branch or "shortcut" branch? Thanks

danpovey:
shortcut

LvHang:
Update.
Thank you for giving me the chance to read the process of nnet3. Can I help something others? Thanks!
Hang

PR #1391, opened 2017-02-01
WIP: Backstitch training

freewym:


freewym:
@danpovey I encounter a problem when I was running the training: it says 

"...(nnet-compute.cc:506) Could not accept input for network node input (it is not expected at this point in the computation)"

I did't have this problem in my old adversarial training code against an olde master branch. So I suspect it might be due to some code changes in nnet-compute.cc. The log file is

/export/a12/ywang/kaldi/egs/ami/s5b/exp/sdm1/chain_cleaned/tdnn_lstm1i_adversarial0.2_fast_all_shortcut_sp_bi_ihmali_ld5/log/train.0.1.log

Yiming

danpovey:
This relates to some refactoring in the shortcut branch.  You have to
create a new NnetComputation object for the 2nd pass.

Dan

src/nnet3/nnet-chain-training.cc
```diff
@@ -70,21 +70,37 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
 
   NnetComputer computer(nnet_config.compute_config, *computation,
                         *nnet_, delta_nnet_);
+  // no adversarial training on the first minibatch to avoid some stats to be
+  // negative due to the scaling with the negative learning rate
+  if (nnet_config.alpha > 0.0 && num_minibatches_processed_ > 0) {
+    // adversarial training is incompatible with momentum > 0
+    KALDI_ASSERT(nnet_config.momentum == 0.0);
+    // give the inputs to the computer object.
```
freewym: @danpovey Do you mean I need to create additional NnetComputation and NnetComputer objects here?

danpovey:
sorry not NnetComputation, just NnetComputer.

egs/swbd/s5c/local/chain/tuning/run_blstm_6j.sh
```diff
@@ -185,6 +186,7 @@ if [ $stage -le 13 ]; then
     --trainer.optimization.initial-effective-lrate 0.001 \
     --trainer.optimization.final-effective-lrate 0.0001 \
     --trainer.optimization.momentum 0.0 \
+    --trainer.optimization.alpha $alpha \
```
danpovey: When we actually commit this I'd want to have a new script in 'tuning', with results properly declared at the top.
What alpha value was used and do you have the output of compare_wer_general.sh for some suitable comparison?

egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -757,6 +757,10 @@ def __init__(self,
                                  lstm*=0,0.2,0'.  More general should precede
                                  less general patterns, as they are applied
                                  sequentially.""")
+        self.parser.add_argument("--trainer.optimization.alpha", type=float,
```
danpovey: At this level, and probably also on the command line and in the code, best to call this adversarial-training-scale.  'alpha' does not really tell the user what it's all about.

src/nnet3/nnet-chain-training.cc
```diff
@@ -70,21 +70,40 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
 
   NnetComputer computer(nnet_config.compute_config, *computation,
```
danpovey: best to declare this object later on, just before it's needed.  otherwise it's confusing.

src/nnet3/nnet-chain-training.cc
```diff
@@ -70,21 +70,40 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
...
+    // creates a new NnetComputer object
+    NnetComputer computer_adv(nnet_config.compute_config, *computation,
+                              *nnet_, delta_nnet_);
+    // give the inputs to the computer object.
+    computer_adv.AcceptInputs(*nnet_, chain_eg.inputs);
+    computer_adv.Run();
+
+    this->ProcessOutputs(true, chain_eg, &computer_adv);
+    computer_adv.Run();
+    UpdateParamsWithMaxChange(true);
```
danpovey: please document the meaning of the true/false boolean options by declaring a variable
`bool is_adversarial_step = true;`
and calling it with that variable.  And likewise anywhere else you call this.

src/nnet3/nnet-training.cc
```diff
@@ -89,14 +108,17 @@ void NnetTrainer::ProcessOutputs(const NnetExample &eg,
       ComputeObjectiveFunction(io.features, obj_type, io.name,
                                supply_deriv, computer,
                                &tot_weight, &tot_objf);
-      objf_info_[io.name].UpdateStats(io.name, config_.print_interval,
-                                      num_minibatches_processed_++,
+      objf_info_[io.name + suffix].UpdateStats(io.name + suffix,
+                                      config_.print_interval,
+                                      (is_adversarial_step ?
+                                      num_minibatches_processed_ :
```
danpovey: no need to do the "?" expression.  Just call with num_minibatches_processed_, and then afterwards, do:
`if (!is_adversarial_step)
    num_minibatches_processed_++;`

danpovey:
COmpile failure (no hurry to fix, I'll sleep before using it):
```
nnet-training.cc: In member function ‘void kaldi::nnet3::NnetTrainer::Train(const kaldi::nnet3::NnetExample&)’:
nnet-training.cc:67:15: error: ‘const struct kaldi::nnet3::NnetTrainerOptions’ has no member named ‘alpha’
   if (config_.alpha > 0.0 && num_minibatches_processed_ > 0) {
               ^
nnet-training.cc: In member function ‘void kaldi::nnet3::NnetTrainer::UpdateParamsWithMaxChange(bool)’:
nnet-training.cc:201:15: error: ‘const struct kaldi::nnet3::NnetTrainerOptions’ has no member named ‘alpha’
   if (config_.alpha > 0.0) {
               ^
nnet-training.cc:204:41: error: ‘const struct kaldi::nnet3::NnetTrainerOptions’ has no member named ‘alpha’
         (is_adversarial_step ? -config_.alpha : (1 + config_.alpha));
                                         ^
nnet-training.cc:204:62: error: ‘const struct kaldi::nnet3::NnetTrainerOptions’ has no member named ‘alpha’
         (is_adversarial_step ? -config_.alpha : (1 + config_.alpha));
                                                              ^
```

freewym:
Probably I also need to add the freezing natural gradient code?

danpovey:
Yes, definitely

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +68,49 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
                              &request);
   const NnetComputation *computation = compiler_.Compile(request);
 
+  // no adversarial training on the first minibatch to avoid some stats to be
```
danpovey: I really want this code to be clear, and for the "simple path" (no adversarial training) not to be cluttered by the adversarial-training code.  I think it would be clearer if you were to create a function like this [and similar in the non-chain code]:
```
void NnetChainTrainer::TrainInternal(const NnetChainExample &eg,
                                                            const NnetComputation &computation,
                                                            bool is_adversarial_step);
```
You could call FreezeNaturalGradient() outside of that function, so that it would
have the same code in both cases (it would just have to pass 'is_adversarial_step' in).

BTW, regarding the asserts that failed before you disabled the 1st-minibatch adversarial step: we should probably remove those asserts and make sure the code does something reasonable when the stats are 
negative.  Because even disabling the 1st-minibatch adversarial step they could still get negative, due to being scaled by the max-change code.  The self-repair code could just return if the counts are negative.  Before calling the fast-l LSTM kernels you just need to make sure the counts are nonzero to prevent division by zero; the actual value hardly matters at all as this condition is so rare and the self-repair is so gentle.

src/nnet3/nnet-chain-training.h
```diff
@@ -70,13 +70,17 @@ class NnetChainTrainer {
...
   // Applies per-component max-change and global max-change to all updatable
   // components in *delta_nnet_, and use *delta_nnet_ to update parameters
-  // in *nnet_.
-  void UpdateParamsWithMaxChange();
+  // in *nnet_. If is_adversairal_step && opts_.alpha > 0, the update is
+  // an adversairal_step where the params are scaled by -opts_.alpha;
+  // if !is_adversairal_step && opts_.alpha > 0 the update is a normal step
+  // where the params are scaled by 1+opts_.alpha; otherwise
+  // !is_adversairal_step && opts_.alpha == 0, and it is a normal step
+  void UpdateParamsWithMaxChange(bool is_adversairal_step = false);
```
danpovey: adversarial is misspelled here and in the comment 3 times.

src/cudamatrix/cu-kernels.cu
```diff
@@ -2909,8 +2909,8 @@ static void _diff_lstm_nonlinearity(const int cell_dim, const int num_rows,
     const Real* sr_config = self_repair_config;
 #   pragma unroll
     for (int i = 0; i < 5; i++) {
-      update_sr[i] = deriv_sum_in[i * deriv_sum_in_stride + j] / count
-          < sr_config[i];
+      update_sr[i] = (count > 0 ? deriv_sum_in[i * deriv_sum_in_stride + j] /
```
freewym: This way we don't need to worry about when count <= 0. Is it good?

freewym: what if count < 0?  do we need to set update_sr[i] = 0 in this case?

freewym: nevermind, it would be false

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +68,43 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
                              &request);
   const NnetComputation *computation = compiler_.Compile(request);
 
-  NnetComputer computer(nnet_config.compute_config, *computation,
+  if (nnet_config.adversarial_training_scale > 0.0) {
```
danpovey: @freewym, you don't seem to have in this PR the stuff where we do the adversarial training with some specified probability for each minibatch.  I think that's important, as it makes it much faster.  Can you add that to the PR before I review it in detail?

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +68,43 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
...
+    // adversarial training is incompatible with momentum > 0
+    KALDI_ASSERT(nnet_config.momentum == 0.0);
+    bool freeze = true;
```
danpovey: actually, I don't think you need the 'freeze' variables in this case after all-- just call with true and false and for the false one add a comment: // un-freeze natural gradient.

src/nnet3/nnet-chain-training.cc
```diff
@@ -135,9 +156,12 @@ void NnetChainTrainer::ProcessOutputs(const NnetChainExample &eg,
...
     computer->AcceptInput(sup.name, &nnet_output_deriv);
 
-    objf_info_[sup.name].UpdateStats(sup.name, opts_.nnet_config.print_interval,
-                                     num_minibatches_processed_++,
+    objf_info_[sup.name + suffix].UpdateStats(sup.name + suffix,
+                                     opts_.nnet_config.print_interval,
+                                     num_minibatches_processed_,
                                      tot_weight, tot_objf, tot_l2_term);
+    if (!is_adversarial_step)
+      num_minibatches_processed_++;
```
danpovey: I think you can remove the if-statement and put this statement at the end of Train().

src/nnet3/nnet-training.cc
```diff
@@ -260,7 +258,7 @@ void ObjectiveFunctionInfo::UpdateStats(
     BaseFloat this_minibatch_tot_aux_objf) {
   int32 phase = minibatch_counter / minibatches_per_phase;
   if (phase != current_phase) {
-    KALDI_ASSERT(phase == current_phase + 1); // or doesn't really make sense.
+    //KALDI_ASSERT(phase == current_phase + 1); // or doesn't really make sense.
```
freewym: have to comment it out, since otherwise the assert will fail because of the adversarial training with probability

danpovey:
You can change it to
KALDI_ASSERT(phase > current_phase);
which anyway Pegah changed it to in a separate PR.

jtrmal:
@freewym  can you rebase on the new master? We did some changes to the travis configuration, so the builds should not fail now

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +68,43 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
                              &request);
   const NnetComputation *computation = compiler_.Compile(request);
 
-  NnetComputer computer(nnet_config.compute_config, *computation,
+  if (nnet_config.adversarial_training_scale > 0.0 && num_minibatches_processed_
+      % nnet_config.adversarial_training_interval == 0) {
+    // adversarial training is incompatible with momentum > 0
+    KALDI_ASSERT(nnet_config.momentum == 0.0);
+    FreezeNaturalGradient(true, delta_nnet_);
+    bool is_adversarial_step = true;
```
freewym: @danpovey For reversing the two steps, the only changes I made is setting is_adversarial_step=false here and is_adversarial_step=true at line 81.

PR #1393, opened 2017-02-02
fix clipped proportion ploting issue

keli78:
fixed the plotting issue of clipped proportion


egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
```diff
@@ -113,12 +113,11 @@ def parse_progress_logs_for_clipped_proportion(exp_dir):
     progress_log_files = "%s/log/progress.*.log" % (exp_dir)
     component_names = set([])
     progress_log_lines = common_lib.run_kaldi_command(
-        'grep -e "{0}" {1}'.format(
+        'grep -H -e "{0}" {1}'.format(
```
danpovey: this really fixed it?  Because -H is supposed to be the default if there are multiple files to search, and I would imagine there would be multiple progress log files.

PR #1394, opened 2017-02-02
add empty data checks in generate_plots.py

keli78:
I found three extra functions in generate_plots.py: generate_nonlin_stats_plots(), generate_clipped_proportion_plots(), and generate_parameter_diff_plots() and added empty data checks for the former two. As for the last one, it can fail itself if the number of iteration is zero.


danpovey:
Will wait for Vijay to review.  Vijay-- the issue I found with empty plots was due to zero values, that don't show up on the graph.  Also, please make the merge message start with [scripts] if you merge yourself.

PR #1395, opened 2017-02-02
ivector compatibility checks

jtrmal:
this replaces #1208, the change is that this PR was created against different target branch ("shortcut" branch).

@vimalmanohar @vijayaditya, it would be good you could review the changes especially with the respect to the python scripts -- I'm not sure if I've covered all the training scripts (dnn/rnn/raw_* and so on) and if it covers all use-cases.


jtrmal:
@danpovey , do you wanna merge, so that it would get more thoroughly tested before merging to master? I could start using the shortcut branch as default for non-babel stuff

PR #1400, opened 2017-02-07
Resolve conflicts due to cross compilation changes in master

dogancan:
@danpovey In addition to resolving the conflicts due to cross compilation changes in master, I also made some additional organizational changes to the configure script that made sense after merging. 
1. We now set `AR`, `AS` and `RANLIB` in the configure script (like `CXX`) and let the user overwrite these via environment variables. 
2. If `HOST` is set by the configure script, then it is prepended to `CXX`, `AR`, `AS` and `RANLIB` before they are written to `kaldi.mk`.
3. I removed the hard-coded `clang++` dependency in android builds. We now tell the user to set `CXX=clang++` during the configure call if the android flag is provided but `CXX != clang++`. 
4. I also removed the `-DHAVE_CLAPACK` flag from the android build flags for the moment since I believe it was not intentional but I will add it back if necessary after Eric responds.


src/configure
```diff
@@ -42,35 +57,43 @@ The default configuration is to build and link against static Kaldi libraries.
...
+  --speex-libdir=DIR    SPEEX library directory
+  --speex-incdir=DIR    SPEEX include directory
+  --host=HOST           Host 'triple' for cross compilation
+  --android-incdir=DIR  Andraid include directory
+
+Following environment variables can be used to override the default toolchain
+  CXX         C++ compiler [default=g++]
+  AR          Archive maintenance utility [default=ar]
+  AS          Assembler [default=as]
+  RANLIB      Archive indexing utility [default=ranlib]
```
danpovey: I'll merge this as-is, but a note here to the effect that $HOST will be prepended might be nice.

dogancan: I can do that right away. Give me 10 mins

danpovey:
Please check why travis build failed.
Seems configure-version related:
```
 make all -j4
The kaldi.mk file was generated using a different version of configure
script. Please rerun the configure again
test -d /home/travis/build/kaldi-asr/kaldi/src/lib || mkdir
/home/travis/build/kaldi-asr/kaldi/src/lib
make depend
```

src/configure
dogancan: Would you like me to put this note in the description of `--host` option or after these environment variables?

danpovey:
both would be ideal.

dogancan:
@danpovey Travis build failure is fixed. Sorry about that.

Do you mind reviewing the handling of clang++ requirement for android builds? Currently if android flag is given but CXX is not a variant of clang++, we are printing an error message asking the user to make sure clang++ is installed and to override CXX during the configure call. There are other options like overriding CXX automatically whenever android flag is given but I don't like that all that much.

danpovey:
I think printing a warning is a good solution about clang++.  I'll merge this.

PR #1401, opened 2017-02-07
little fix of the script.

caizexin:
when training ubm, we should just use the 40 dimention mfcc
so change the train directory for avoiding dimention mismatching between stage 2 and stage 3
this script won't get error when run after nnet2's scripts.


naxingyu:
Thanks for the fix!
@danpovey the current results are not effected because the this script is taking the iVector extractor trained in the nnet2 online setup when it is called, e.g. https://github.com/kaldi-asr/kaldi/blob/master/egs/hkust/s5/local/nnet3/run_tdnn.sh#L50 <https://github.com/kaldi-asr/kaldi/blob/master/egs/hkust/s5/local/nnet3/run_tdnn.sh#L50>

X.

PR #1402, opened 2017-02-07
Sprak nnet3

dresen:


danpovey:
Thanks!  It may be a day or two before I get to this, I am a bit busy
today.'

galv:
I noticed that you have a lot of early commits that merge the kaldi master branch. I recommend you get rid of those. In the future you can rebase your work on top of the kaldi master branch. But for now, you can cherry-pick your commits on top of a fresh master branch.

i.e., if you make a branch my-branc which is a copy of master:

```
git cherry-pick dbca511
git cherry-pick d3d4e41
```

and so on. git will tell you if there are conflicts but I doubt this will be the case for a new recipe.

danpovey:
I've got in the habit of doing 'squash and merge' rather than retaining the
history of the committer's branch.  So this may not matter.

egs/sprakbanken/s5/RESULTS
```diff
@@ -1,25 +1,21 @@
...
-%WER 23.19 [ 2507 / 10811, 435 ins, 520 del, 1552 sub ] exp/tri3b/decode_3g_test1k.si/wer_12
-%WER 19.06 [ 2061 / 10811, 427 ins, 384 del, 1250 sub ] exp/tri3b/decode_4g_test1k/wer_11
-%WER 23.20 [ 2508 / 10811, 447 ins, 520 del, 1541 sub ] exp/tri3b/decode_4g_test1k.si/wer_11
-%WER 17.42 [ 1883 / 10811, 416 ins, 359 del, 1108 sub ] exp/tri4a/decode_3g_test1k/wer_13
-%WER 20.86 [ 2255 / 10811, 403 ins, 473 del, 1379 sub ] exp/tri4a/decode_3g_test1k.si/wer_13
-%WER 17.52 [ 1894 / 10811, 396 ins, 372 del, 1126 sub ] exp/tri4b/decode_3g_test1k/wer_13
-%WER 20.82 [ 2251 / 10811, 399 ins, 471 del, 1381 sub ] exp/tri4b/decode_3g_test1k.si/wer_13
-%WER 17.53 [ 1895 / 10811, 403 ins, 375 del, 1117 sub ] exp/tri4b/decode_4g_test1k/wer_13
-%WER 20.99 [ 2269 / 10811, 438 ins, 436 del, 1395 sub ] exp/tri4b/decode_4g_test1k.si/wer_11
+%WER 22.87 [ 24286 / 106172, 3577 ins, 5321 del, 15388 sub ] exp/tri1/decode_fg_dev/wer_12_0.5
```
galv: Looks like you removed the results of existing recipes. Why?

egs/sprakbanken/s5/conf/online_cmvn.conf
```diff
@@ -0,0 +1 @@
+# configuration file for apply-cmvn-online, used in the script ../local/run_online_decoding.sh
```
galv: `local/run_online_decoding.sh` does not exist for this recipe. Maybe you accidentally copied this file from tedlium.

galv: If it is unused, delete it.

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+echo $0 $*
```
galv: Be sure to document what arguments are appropriate for $*. Looks like it's simply the names of the exp/ directories you want to compare. You should print this help message out if $* is empty.

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
```diff
@@ -0,0 +1,64 @@
...
+  printf "% 10s" $prob
+done
+echo
+
+echo -n "Final valid prob (xent)"
+for x in $*; do
+  prob=$(grep Overall $x/log/compute_prob_valid.final.log | grep -w xent | awk '{printf("%.4f", $8)}')
+  printf "% 10s" $prob
+done
+echo
```
galv: There is nothing about this file specific to the sprakbanken dataset. It should therefore belong in utils/ instead. Probably the best place to put it is utils/nnet3 but that doesn't exist, so maybe steps/nnet3 is a better place (Especially since there are other "utility"-like scripts there, like the reporting scripts and nnet3_to_dot.sh

Another maintainer should comment on whether this script is redundant with some other script and whether it should be included.

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+echo $0 $*
+
+echo -n "System               "
+for x in $*; do   printf "% 10s" " $(basename $x)";   done
```
galv: Change $* to "$@"

It's important in case someone has an experimental directory that contains a space. That's probably never going to happen, but you never know.

egs/sprakbanken/s5/local/chain/run_tdnn.sh
```diff
@@ -0,0 +1 @@
+tuning/run_tdnn_1b.sh
```
galv: It doesn't really matter, but best to use #!/bin/bash at the top here. Same for `run_tdnn_lstm.sh`

galv: Also, you have `run_tdnn_lstm.sh` and `run_tdnn.sh`, but you're missing `local/chain/run_lstm.sh` Any reason why you don't have the last file?

egs/sprakbanken/s5/local/chain/tuning/run_tdnn_1a.sh
```diff
@@ -0,0 +1,202 @@
...
+# by default, with cleanup:
+# local/chain/run_tdnn.sh
+
+# without cleanup:
+# local/chain/run_tdnn.sh  --train-set train --gmm tri3 --nnet3-affix "" &
+
+# note, if you have already run the corresponding non-chain nnet3 system
+# (local/nnet3/run_tdnn.sh), you may want to run with --stage 14.
+
+set -e -o pipefail
```
galv: @danpovey @vijayaditya Do we consider these settings okay? I feel like I remember being asked to remove these in one of my earlier pull requests, but I'm not sure. At the every least, I think I was asked to remove -e because you all prefer being explicit with `|| exit 1 ` clauses.

egs/sprakbanken/s5/run.sh
```diff
@@ -125,19 +124,43 @@ if [ $stage -le 9 ]; then
...
-##TODO: Add nnet3 and chain setups
 
 ## Works
 #local/sprak_run_nnet_cpu.sh tg dev 
 
 ## Works
 #local/sprak_run_sgmm2.sh dev
 
+# Run neural network setups based in the TEDLIUM recipe
+if [ $stage -le 11 ]; then
```
galv: @danpovey Is it typical to include nnet3 recipes inside run.sh? I feel that leaving in the lstm recipes is not good since those have the potential to run for a very long time without someone realizing.

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
danpovey: I believe this script may not have been properly updated to reflect the characteristics of the Sprakbanken database.   It is supposed to print out a little table for comparing WER and objective values between different training runs.  But this seems to be a copy of some other script.  It does contain some corpus-specific things-- the set of names of decoded directories, and what type of scoring they have (Kaldi vs. sclite).

egs/sprakbanken/s5/local/chain/tuning/run_tdnn_1a.sh
danpovey: The current situation is many scripts do have the 'set -e', and many do not.  Officially, either is OK right now.  It's not an ideal situation but that's the way it is.

egs/sprakbanken/s5/run.sh
danpovey: Hm.  I'm not sure.  I'd probably prefer to include the command (or commands) but commented out, with a comment saying that this is the command to run [such-and-such.]

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
galv: It's a fully new script according to git, so it's either newly created or copy-pasted from elsewhere, actually. It clearly works for @dresen if it's here now. Regardless, if it's not robust and clearly-documented, my guess is that we shouldn't keep it.

danpovey:
OK, let's see what he says about whether it's been tested.
It was probably copied from a similar script in egs/swbd/s5c or egs/ami/s5b
or egs/tedlium/s5_r2, but not sure if it has been used locally. Those
scripts need to be documented better.

egs/sprakbanken/s5/RESULTS
dresen: Those results are based on a subset of the whole test set that was used to do "fast testing". This ended up being more of a development set which is not acceptable for a subset of the test set, so I redid all GMM experiments with tuning on the actual dev set and rather than test1k and proceeded to the same for nnet3 and chain recipes.

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
dresen: I had forgotten to remove this script. I copied it over from tedlium beecause I wanted to get it to work, but as I understand score_sclite.sh, I need a 'glm' or 'slm' file, which is not part of the språkbanken data

egs/sprakbanken/s5/local/chain/run_tdnn.sh
dresen: It wasn't in the tedlium folder I copied, but I'll add it.

egs/sprakbanken/s5/run.sh
dresen: I'll comment it out

egs/sprakbanken/s5/conf/online_cmvn.conf
dresen: This is the next thing I want to work on.

egs/sprakbanken/s5/local/chain/compare_wer_general.sh
danpovey: You could easily get it to work, compare with the Switchboard setup where for the dev set it uses Kaldi scoring and not sclite scoring.  But either get it to work or remove it.

egs/sprakbanken/s5/local/chain/run_tdnn.sh
danpovey: Actually no need to have the 'just-lstm' number, generally TDNN+LSTM will work better.

danpovey:
Looks good.  I'll merge and if any issues are found in future we'll address them then.

PR #1404, opened 2017-02-07
WIP ivector-pca: adding support to replace LDA+MLLT with PCA for ivectors

### LINKED ISSUE #1123: Getting rid of need for alignments in iVector training
@david-ryan-snyder, it continues to be a pain that we need alignments to get the LDA+MLLT transform for purposes of training speaker-id systems.
Can you see if it's possible to replace that with a PCA transform without hurting results?  Did we ever look at this?

david-ryan-snyder:
* Adding support to replace LDA+MLLT with PCA for ivectors used in online ASR systems. Makding this the default in the Tedlium chain recipe, and updating the RESULTS using this.

* This is WIP until I add enough results that we're comfortable switching over to it. 

* Once merged, this will close #1123.


jtrmal:
David,  are the tedlium numbers really comparable?  In the original file I
see "chain_cleaned" which would indicate data cleanup was used - and your
directory is called differently.  Did you use the same cleaned data? I'm
not familiar with that recipe so I'm not sure if cleanup is now used by
default or not.
Thanks,  y.

david-ryan-snyder:
I think you're actually looking at an old commit in the PR. At one point I copied and pasted in the wrong info, but then fixed it in a later commit. If you check again, you'll see more comparable numbers.

But there's another problem I realized after you mentiond this: the current version of the recipe is using some new tdnn**1b**_sp_bi whereas all the results in the RESULTS file are reported on an older tdnn_sp_bi. So I better rerun some of the recipes in the RESULS file using the new TDNN script anyway, to make sure the comparison is fair (comparison of using PCA vs LDA features to the ivector system).

jtrmal:
yes, you are right -- not sure what github has displayed on my phone, but
now on my laptop the diff looks exactly as you described.
y.

david-ryan-snyder:
I just realized the Tedlium TDNN tuning recipes have been updated since I started working on this. Going to rerun this change (using PCA instead of LDA+MLLT ivectors) based on the latest recipe. After that, we we should have up-to-date results for Tedlium, WSJ, and AMI using the new ivector features.

danpovey:
@david-ryan-snyder, there are conflicts now..

david-ryan-snyder:
@danpovey, thanks. I'm taking care of it now.

david-ryan-snyder:
I made a new PR for this one. Too many parts were out of date. See https://github.com/kaldi-asr/kaldi/pull/1514

PR #1407, opened 2017-02-09
Resolve merge conflicts and add "make ext" to travis build

dogancan:
I noticed merge conflicts against master were not completely resolved after last merge. This PR addresses that and adds "make ext" to travis build.


danpovey:
thanks, will merge after travis build finishes (maybe to-morrow).

dogancan:
1. Earlier travis failure was due to a dependency bug in `src/Makefile` that manifested in parallel "make ext" executions. Fixed now.
2. Building with double precision mostly worked out of the box. There were a few places in the codebase that used a regular `float` that I changed to `BaseFloat` . Probably a good idea to review those changes.
3. I also added a random selection between double and single precision BaseFloats to the travis build.

PR #1408, opened 2017-02-09
Give prepare_lang the option to generate more phone disambig symbols

psmit:
Hi,

Could you consider merging this pull request. For some subword modeling work we do we use prepare_lang.sh for the general lang-directory setup after which we generate our own L_disambig.fst and we need a couple more disambig symbols then are generated by default.

Thanks.


danpovey:
Sure, merging.

PR #1409, opened 2017-02-09
making the Makefile checks serial

jtrmal:
@dogancan @danpovey , what do you think about this?
The issue I'm trying address is that if  make is run with -j, the output is usually quite convoluted and is not directly obvious what is the error.
y.


danpovey:
Looks reasonable (and it was probably a bug, not just a cosmetic issue);
but can you rename all0 -> subdirs ?

jtrmal:
ok, done.

dogancan:
LGTM.

PR #1415, opened 2017-02-10
adding missing --stage 0 to run_nnet2_ms.sh

weegreenblobbie:


danpovey:
Thanks for the fix, but what we were going to do is to change the default in the script itself to 0.  Also there was one other fix to that setup, which I forget what it was.  I had asked @keli78 to do it, but if you can do it yourself, then great.

weegreenblobbie:
conf/mfcc_hires.conf was missing, but it got added about a month ago, I was out of date :)

danpovey:
OK, can you modify your PR (or make a new one) to change the default in the script?

weegreenblobbie:
no problem

PR #1419, opened 2017-02-13
Add Exponential Linear Unit (ELU) component

hhadian:
Do I need to add an 'elu-layer' (similar to 'relu-layer') to parser.py?


danpovey:
yes.

danpovey:
or maybe basic_layers.py, you might ahve to change a couple of python files.

hhadian:
OK. I guess all previous commits have appeared in my pull request, should I make another pull request?

danpovey:
I guess you merged master; it's OK, when I merge master to shortcut those
commits will disappear, will do it to-morrow.

hhadian:
OK. (yes I did merge master!)

hhadian:
I added `elu` to parser.py (and to basic_layers.py) as a basic layer. I noticed there are` tdnn-relu-*` layers too (which are defined in tdnn.py). Should I add `elu` equivalents?
Also I have been using `relu-renorm-layer` in xconfig-based TDNN scripts. Shouldn't I use `tdnn-relu-renorm` instead?

src/nnet3/nnet-simple-component.cc
```diff
@@ -1171,6 +1171,118 @@ void RectifiedLinearComponent::StoreStats(
   StoreStatsInternal(out_value, &temp_deriv);
 }
 
+void EluComponent::Propagate(
+    const ComponentPrecomputedIndexes *indexes,
+    const CuMatrixBase<BaseFloat> &in,
+    CuMatrixBase<BaseFloat> *out) const {
+  // Apply exponential linear function (x >= 0 ? x : exp(x) - 1)
+  out->CopyFromMat(in);
+  CuMatrix<BaseFloat> exp_negated_relu(in);
```
danpovey: You can use the per-element Min() function [which was just now added], to make this more efficient. for propagate: copy in->output, exponentiate output, do output->Min(input).

danpovey:
For some reason the extra commits didn't disappear even though I merged master.  If they don't disappear after you make the next change, try rebasing against upstream/shortcut.

src/nnet3/nnet-simple-component.cc
hhadian: OK

hhadian:
OK

src/nnet3/nnet-simple-component.cc
hhadian: kPropagateInPlace is on so I will need to make a separate copy, right? Or do you want me to turn it off?

danpovey:
You can test whether out and in are the same, and make a copy if they are.
kPropagateInPlace just gives the compiler the ability to do in-place
propagation, it does not make it mandatory.

hhadian:
I guess I didn't get it exactly how you want me to use Min(). I think this is what you suggested:
```
out = exp(in) - 1
out->Min(in)
```
but after this, `out` will be equal to `in` because `in` is always smaller (in value, not magnitude). We will need to apply `abs` on `in` before calling `Min()`.

danpovey:
Hm, yes.  It might be possible to do:
out = exp(min(in, 0)) - 1
out->Min(in)
but you'd have to see whether what you are already doing is likely to be
faster.
Bear in mind that allocating matrices and zeroing them is not cost-free
(it's not just explicit operations that cost you something).

hhadian:
This isn't possible either (for `in` >= 0, `out` becomes 0). I will need to do that `abs()` thing which will make it almost the same as what I have already implemented.

danpovey:
ok, max not min.
abs() is problematic because you'd need an extra copy of 'in'.

hhadian:
Yes exactly, that's why it will be almost the same (in the current implementation there is a copy too).
Actually I considered `max` too, but then `exp` is not applied to negative values of `in`.

danpovey:
ok it's probably fine as-is.  later if this really works we can ask Shiyin
to write a kernel.

hhadian:
OK. BTW, I can write the kernel myself.

danpovey:
I don't have a problem with that.

hhadian:
Also mentioning @pegahgh who was interested. 
The results for CE/TDNN:
```
# System                tdnn_elu1a_sp tdnn_d_sp
# WER on train_dev(tg)      16.78     16.72
# WER on train_dev(fg)      15.52     15.31
# WER on eval2000(tg)        19.5      19.2
# WER on eval2000(fg)        18.0      17.8
# Final train prob         -1.216    -1.229
# Final valid prob         -1.321    -1.354
```

danpovey:
Odd that the WER is worse but there is a good valid-prob improvement.
Could you try it on another setup?
Dan

pegahgh:
Would you please generate plots for average relu values for elu vs. relu
for different layers?
It is interesting that it improves final train and valid prob!

hhadian:
I will try it on AMI since I have it ready.
@pegahgh, I will email them.

danpovey:
@hhadian, can you resolve the conflicts?
I'm considering merging this even though it's no better (so far) than ReLU, because it may be useful in future.

hhadian:
Will do it now

hhadian:
And this is the result for chain:
```
# System                tdnn_elu1a_sp tdnn_7h_sp
# WER on train_dev(tg)      14.37     13.96
# WER on train_dev(fg)      13.15     12.86
# WER on eval2000(tg)        17.1      16.5
# WER on eval2000(fg)        15.3      14.8
# Final train prob         -0.089    -0.093
# Final valid prob         -0.110    -0.117
# Final train prob (xent)        -1.296    -1.317
# Final valid prob (xent)       -1.3507   -1.4047
```

hhadian:
Obviously I am doing ELU experiments against shortcut (Kaldi 5.1), but the baseline (e.g. 7h) is against master, just asking to make sure that the TDNN baselines are not supposed to change in 5.1, right?

danpovey:
No, nothing was changed that would affect the TDNN results.
However, due to mean reversion, the expectation is that when we rerun
results that are checked in they will be slightly worse (because we
typically only check in things that are better, and that was originally
partly random).

PR #1423, opened 2017-02-15
Test clang on travis.

dogancan:
This is not for merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ notifications:
   email: false
 
 language: cpp
-sudo: required
+sudo: false
 dist: trusty
 
 os:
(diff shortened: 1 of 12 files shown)


dogancan:
@danpovey It does not seem like there is a correlation between single/double precision and travis build times. There isn't all that many clang builds to make a definitive conclusion but the gcc builds suggest the same thing and there is quite a bit of those so I am pretty sure this is not the reason.

Ignoring the builds that have timed out, average build time on the shortcut branch

compiler  |  # builds | # parallel jobs | average time
--------- | ---------|----------------|-------------
g++          |  19          |  4                      | 39mins
clang++   |  5            |  6                      | 29mins 

Note that 4 is the maximum number of jobs we can run with g++ due to the memory limit. We can go higher with clang++ but my completely unscientific tests suggest that build times do not improve beyond 6 jobs.

danpovey:
OK, I'm happy to merge the clang change if you think it helps, just let me
know when it's ready.

dogancan:
Closing this. I'm opening a new one against master.

Issue #1418, opened 2017-02-12
New CUDA functions


danpovey:
@kangshiyin, would you mind creating functions in class MatrixBase called ApplyFloor and ApplyCeiling that take a matrix argument, and can apply a different floor or ceiling per element?  So x.ApplyFloor(y) means: x = max(x, y) per element, and min for ApplyCeiling?

kangshiyin:
Sure, will do this week.

PR #1424, opened 2017-02-15
Element-wise matrix max(floor) and min(ceiling)

kangshiyin:
This is to address #1418 .
All GPU tests have passed.


danpovey:
Thanks!
I'm merging even though the tests failed as they timed out but were going OK.

PR #1429, opened 2017-02-17
Change Travis compiler to clang++-3.9.

dogancan:
Increase the number of parallel make jobs to 6.

Remove -rdynamic flag from CXXFLAGS. It is already in LDFLAGS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ notifications:
   email: false
 
 language: cpp
+sudo: false
+dist: trusty
 
 os:
   - linux
@@ -9,13 +11,12 @@ os:
 addons:
   apt:
     sources:
-      - ubuntu-toolchain-r-test
+      - llvm-toolchain-trusty-3.9
     packages:
       - gdb
-      - gcc-4.9
-      - g++-4.9
-      - gfortran-4.9
+      - gfortran
       - liblapack-dev
+      - clang-3.9
 
 branches:
   only:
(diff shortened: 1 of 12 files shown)


dogancan:
@danpovey @jtrmal This is ready for merge. Note that in addition to changing the compiler, I also removed the `-rdynamic` flag from `CXXFLAGS` on all platforms since it is a linker flag and we are already adding it to `LDFLAGS`. clang++ was printing a warning whenever we used this flag while compiling the object files.

dogancan:
@jtrmal Also see #1423 for build time averages with clang (vs gcc).

jtrmal:
I'm ok with merging this. My experience about the times is about the same.
y.

jtrmal:
BTW, another option how to speed things up  (after making kaldi 5.1
default) would be either download the stock openfst (hopefully it will be
available) or have a package already precompiled somewhere and just
download it (as a kinda desperate measure).
But I think Dogan or someone else has been already mentioning this.
y.

danpovey:
It seems to me that we have had more build failures since the switchover to clang; for some reason the build time seems more variable.
I am wondering whether we could revert this.
Any comments?

jtrmal:
I don't care too much about if it will be reverted or not. I checked the
two last failures and both were because of the timeout. So I guess we will
never know if it's really clang or something else unless we revert at least
for some time...
y.

danpovey:
Looks like that commit had other changes which were fixes, and should not
be reverted.  I'll wait till Dogan has time to make a PR which will change
it back to g++.

PR #1436, opened 2017-02-20
cleanup the HKUST scripts and add the scoring filters

jtrmal:
I added a couple if fixes to the hkust that I had not checked in. Mostly scoring stuff, which I don't think was done completely right. For example, every English word was originally split into characters while I believe it should be a single token, i.e. only CJK plane should be tokenized per-character.
It will change the numbers slightly, though.

Another change in scoring is the script produce both WER and CER.
y.


jtrmal:
It's ok to reject this  -- I was just trying to clean up my dir and sync with upstream.

danpovey:
@naxingyu, we believe this will affect the results slightly but Yenda doesn't have time to update them.  When you get a chance, could you please update them?

naxingyu:
Sure~

PR #1443, opened 2017-02-22
Change Travis distribution to ubuntu 12.04, clang to v3.8

dogancan:
This PR is for investigating why Travis builds are failing. ~Not for merge yet.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ notifications:
   email: false
 
 language: cpp
-sudo: false
-dist: trusty
 
 os:
   - linux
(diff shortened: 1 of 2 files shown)


dogancan:
@danpovey @jtrmal This is ready for merge.

I looked into the builds timing out. It seems the problem is not clang but the Ubuntu 14.04 virtual environments provided by Travis that we started using along with the clang-3.9 change. From what I can tell 14.04 VEs are in general slower than Ubuntu 12.04 VEs (Travis default) and there is also a large speed variance between 14.04 instances. We had a similar problem when Yenda first tried clang-3.9 with sudo enabled. In that case, Travis was allocating yet another virtual environment on google compute cloud (not a docker container but a full VM) and those were even slower.

Now the good news. After much trial and error, I managed to get clang-3.8 working in a 12.04 VE. I am now getting solid 20-25min builds with clang-3.8. 🎉 

Note that we are also installing gcc-4.9 and gfortran-4.9 along with clang-3.8 because otherwise the linker cannot find the correct gfortran library for some reason. Apparently this is a well known bug [[1](http://stackoverflow.com/questions/6302209/building-r-package-getting-error-ld-cannot-find-lgfortran),[2](http://askubuntu.com/questions/276892/cannot-find-lgfortran)]. I also modified the check for testable files so that changes in travis scripts trigger new builds.

danpovey:
Great work-- thanks!  Merging.

jtrmal:
Good job!
y.

PR #1444, opened 2017-02-22
[egs] xconfig

songmeixu:
1. xconfig: mv ‘add output-node’ from FixedAffineLayer to common xconfig_to_config

Thereafter output-node will appear in init.config, and it has no matter with FixedAffineLayer.

2. And add several basic nonlinear-layers in xconfig

3. fix swbd egs local/nnet3/run_lstm.sh


egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -15,25 +15,29 @@
...
-        'affine-layer' : xlayers.XconfigAffineLayer,
-        'lstm-layer' : xlayers.XconfigLstmLayer,
-        'lstmp-layer' : xlayers.XconfigLstmpLayer,
-        'lstmpc-layer' : xlayers.XconfigLstmpcLayer,
-        'fast-lstm-layer' : xlayers.XconfigFastLstmLayer,
-        'fast-lstmp-layer' : xlayers.XconfigFastLstmpLayer
-        }
+    'input' : xlayers.XconfigInputLayer,
+    'output' : xlayers.XconfigTrivialOutputLayer,
+    'output-layer' : xlayers.XconfigOutputLayer,
```
danpovey: please don't re-indent this, it messes up the git history.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -159,6 +159,9 @@ def write_config_files(config_dir, all_layers):
     for layer in all_layers:
         try:
             pairs = layer.get_full_config()
+            if layer.descriptors.has_key('input') and 'input' in layer.descriptors['input']['final-string']:
```
danpovey: can you explain what problem you were trying to solve with this change and the change in basic_layers.py?  I suspect that this is not the right fix.

songmeixu: Before, there must be FixedAffineLayer in xconfig, otherwise the 'output-node' would not appear in init.config, as 'output-node' is generated by get_full_config() of 'class FixedAffineLayer'.

Now, the generation of 'output-node' was moved into the final stage of xconfig, no matter whether there is FixedAffineLayer or not.

songmeixu: I know the prepended FixedAffineLayer is for LDA mat. However, i would like there is no limitation to implement other topos of neural networks without it. For saving computations for example.

danpovey:
What I think would make more sense is for the program to count the number
of lines in 'init.config' that are going to start with 'output-node', and:
 - if it's more than one, crash with an error message
 - if it's zero, then don't write init.config at all.
It can do this before actually writing init.config, by looking at the list.
 [if it's not going to write init.config, then it should remove such a file
if it exists; this will avoid confusing behavior if people re-use
directories with different configurations.]
Then, the training library should be modified so that if init.config
doesn't exist in the xconfig dir, it just doesn't do the
LDA-matrix-estimating stages of the training.
If you want you could submit your other unrelated fixes as a separate PR,
and then implement the change that I am describing.

songmeixu:
OK. I see what you mean, that's the correct way!

PR #1446, opened 2017-02-22
Bump Travis clang to v3.9

dogancan:
@danpovey I noticed I didn't try clang-3.9 after figuring out how to build with clang-3.8. I initially had trouble with clang-3.9 on Travis Ubuntu 12.04 VE but that was before I made the changes to get clang-3.8 working. Let's see if we can build with clang-3.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,14 @@ addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
-      - llvm-toolchain-precise-3.8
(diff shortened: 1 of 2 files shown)


dogancan:
@danpovey @jtrmal I tested building with clang-3.9 quite a few times. It seems like clang-3.9 is slightly slower (maybe around 10%) than clang-3.8 as far as build times are concerned. I think we are better off with clang 3.8 for the moment. Closing this PR.

PR #1447, opened 2017-02-22
[egs] xconfig: fix bug when there is not preconditioning LDA layer in xconfig

songmeixu:
Remove the generation of ‘output-node’ in get_full_config() of FixedAffineLayer. Now, FixedAffineLayer makes sense when it is not for precondiitoning LDA.


egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -444,12 +444,20 @@ def smooth_presoftmax_prior_scale_vector(pdf_counts,
 
 
 def prepare_initial_network(dir, run_opts, srand=-3):
-    common_lib.run_job(
-        """{command} {dir}/log/add_first_layer.log \
-                nnet3-init --srand={srand} {dir}/init.raw \
-                {dir}/configs/layer1.config {dir}/0.raw""".format(
-                    command=run_opts.command, srand=srand,
-                    dir=dir))
+    if os.path.exists(dir+"/configs/init.config"):
```
songmeixu: only if init.config exists, prepare with init.raw

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
```diff
@@ -343,7 +343,7 @@ def get_full_config(self):
         # unlike other layers the input layers need to be printed in
         # 'init.config' (which initializes the neural network prior to the LDA)
         ans = []
-        for config_name in [ 'init', 'ref', 'final' ]:
+        for config_name in [ 'ref', 'final' ]:
```
songmeixu: here, we don't know if LDA is used or not. We don't add input info into 'init.config' immediately, but depend on judgement by outter info.

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
```diff
@@ -817,13 +817,6 @@ def get_full_config(self):
         output_dim = self.output_dim()
         transform_file = self.config['affine-transform-file']
 
-
```
songmeixu: output-node for LDA should not always be generated inner FixedAffineLayer. we also generated when needed.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -170,6 +185,12 @@ def write_config_files(config_dir, all_layers):
             # preserves the backtrace
             raise
 
+    # remove previous init.config
```
songmeixu: the init.config is the switcher of LDA computing, previous version should remove anyway.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -156,11 +156,26 @@ def write_config_files(config_dir, all_layers):
 
     config_basename_to_header = get_config_headers()
 
+    input_pairs = []
```
songmeixu: keep input-nodes of networks, by assuming them only exists in front of config(first layer). If not, this is buggy, but previous version is also problematic. Should fix latter.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -156,11 +156,26 @@ def write_config_files(config_dir, all_layers):
...
+
     for layer in all_layers:
         try:
             pairs = layer.get_full_config()
+            # we judge whether the layer is preconditioning lda by retained name 'lda'.
```
songmeixu: Here, we start judge use LDA or not by retained name 'lda'. This is the only making sense method that i come up with.

songmeixu:
I give explanatory comments at each change

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
danpovey: it will be much simpler if you stick exactly to what I sketched out.
Please revert the changes to this class.
The main thing you have to change is the function 'write_config_files' in xconfig_to_configs.py.
This will have to add code in the loop where it does:
`   for basename, lines in config_basename_to_lines.items():`
to check how many lines starting with 'output-node' each config file has.
If any config file doesn't have at least one line starting with 'output-node' then
it won't be written; and if it's not basename 'init' then this will be an error.

most of the changes you've made can be reverted.

songmeixu: Hi, @danpovey , thank you for your detailed advice! But i think this will have problems when one situation that there is one or more FixedAffineLayer, which is not for LDA, just for feat transform for example. In this situation, we shouldn't do LDA also. I think the key is the FixedAffineLayer is not for LDA usage only, so how to decide the FixedAffineLayer is for LDA or not, i come up with the method using the 'name=lda' indication.

Maybe something that i misunderstanded, this is the first time i commit code to kaldi, and i hope it will be more in future. I'm very glad to communicate with you @danpovey ! I'm in china, and it's late in Baltimore. Thank you always replay as soon as you can!

danpovey:
Good point.
You could add a new boolean parameter to class

XconfigFixedAffineLayer with name 'write-init-config', defaulting to true.
Modify its set_default_configs function and its get_full_config function.

That will give a more explicit way to disable this; I don't like relying on
the name because people should be able to assume that the names are
arbitrary strings.

songmeixu:
@danpovey I add the bool 'write-init-config' and add counting of the lines num in write_config_files(). Please check it, thank you🙂

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -170,7 +170,31 @@ def write_config_files(config_dir, all_layers):
...
             raise
 
+    # remove previous init.config
+    try:
+        os.remove(config_dir + '/init.config')
+    except OSError:
+        pass
+
     for basename, lines in config_basename_to_lines.items():
+        # check the lines num before 'output-node':
```
danpovey: You don't have to check the num-lines before 'output-node'. 
Just check the number of times the line starts with 'output-node'.
E.g.:
```
 num_output_node_lines = sum( [ 1 if line.startswith('output-node' ) else 0 
                                                     for line in lines ] )
 ...
```

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -170,7 +170,24 @@ def write_config_files(config_dir, all_layers):
...
+    try:
+        os.remove(config_dir + '/init.config')
+    except OSError:
+        pass
+
     for basename, lines in config_basename_to_lines.items():
+        # check the lines num start with 'output-node':
+        num_output_node_lines = sum( [ 1 if line.startswith('output-node' ) else 0
+                                       for line in lines ] )
+        if num_output_node_lines == 0:
```
songmeixu: hi, @danpovey Check it again pls. 

Due to the output-node could > 1 in chain, I just check the 'num_output_node_lines == 0' condition.

danpovey:
Looks good.  Could you please verify  by running the first steps of training, that it works for the normal case?  (i.e. for existing checked-in setups?)

danpovey:
looks fine.  I won't merge this tonight though, I want to test it myself for a bit (it's not an urgent fix since the problem doesn't affect any recipes).

PR #1449, opened 2017-02-22
SWC Kaldi recipe - merge the updated version ("s5b" only)

wings619:
@danpovey  Hi Dan, I have removed the scripts of "s5" but kept the results and some text so the users know where to download the scripts in case of need. Does that look okay now please?


jtrmal:
I think you could use the utils/best_wer.pl to generate to RESULTS.txt --
that would make the format compatible with the other recipes.


something like
```
grep Avg SA1/tri4a/decode_dev_30k-4gram-int-swc.swc.v2.30k-combilex.4g-int.arpa.pr1-7/ascore_*/dev.ctm.filt.sys
| utils/best_wer.pl
```
should work

danpovey:
Why do you care so much about having the exact results in the paper
available as s5?  Is s5 giving better results than s5b?
I doubt that, in practice, people will care about the results being exactly
the same as the paper, and I think it would be best to keep it simple and
have just one official recipe, named s5.  But let me know if the results
are worse.

Dan

wings619:
@danpovey Hi Dan, just removed the s5 recipe. 

@jtrmal Thank you very much for the advice! I was not aware of the tool and it is absolutely very useful!

danpovey:
OK.
I think it's best if you rename s5b to s5-- otherwise it's a little
confusing.

Dan

danpovey:
There are quite a few scripts in the tuning/ directory.  I recommend that you just select the one  you actually used, rename it to 1a, and delete the others.
Also, in local/nnet3/, instead of, say, run_tdnn.sh, you should create a link to a script tuning/run_tdnn_1a.sh.  The AMI recipe may not have been fully up to date with this convention.
Grep for 'ami' in the scripts to replace with swc, there is at least one place where I saw it.
Make sure there are no old or irrelevant files that you don't need any more.  I don't think you need local/run_cleanup_segmentation.sh, for instance.

wings619:
@danpovey Hi Dan, is it possible to keep the name "s5b" please? This is because there was a previous release of the "s5" recipe in our own websites last year, and there might be confusion if I now rename the "s5b" into "s5" had any users downloaded and installed the old "s5" recipes. 

As for the run_tdnn_1a.sh files, I haven't really run the experiments and I am not clear whether that is stable or not as it is from the AMI folder. Do you think it is better to delete them when I am not certain please? Similarly, you are right that I don't really need the local/run_cleanup_segmentation.sh for the current recipe.

Thank you very much!

danpovey:
It looks like you have just moved the outdated DNN training recipes that
you were using from the old s5 setup to the s5b setup.  That's not really
what I was asking for.  Why can't you just run the more up-to-date, and
better, recipes, that are available in s5b, and delete the old recipes?

wings619:
@danpovey  Hi Dan, I am afraid that I am not sure I understand what you meant by "the more update-to-date, and better, recipes, that are available in s5b". 

I checked the recipe run.sh files line by line in the s5b recipe in AMI and the s5b recipe I updated. The only differences are in the TDNN ("local/chain/run_tdnn.sh") and the "local/run_cleanup_segmentation.sh" script. The rest are basically the same. 

If you are talking about the "run_tdnn.sh" script, I am afraid that I am not planning to include the TDNN results in the recipe in coming 3 months. If you insist I can pass the message to Thomas, he might ask the person who takes over the SWC data from me to run the TDNN experiments and to report the results.

danpovey:
I'm talking about the fact that you are using the older nnet1-based recipes
in your PR, and there are now much better (and better-supported) recipes
available.   The removal of the older recipes based on nnet1 (and possibly
nnet2) was one of the principal differences when I cleaned up the scripts
from s5 to s5b.

I really don't want to add new example scripts to Kaldi that are based on
these nnet1 recipes, as no-one really supports them, plus we have better
results based on nnet3 now.

Now that I look at the PR, you actually seem to have copied over the
nnet3-based scripts (from local/nnet3/ and local/chain/) from the ami/s5b
setup into this PR, but you're not invoking the in the run.sh; instead you
have copied over your older nnet1-based scripts from your original PR and
you are invoking those instead.

The PR also doesn't seem to have been created in a way that would allow it
to work on general grids: there's quite a bit of stuff specific to your
system in there.  That needs to be handled in a better way.

It seems to be important to you to for the checked-in recipe to be able to
create the baseline system that you published, but that's not really the
purpose of the checked-in recipes in Kaldi-- they are not intended as a
reference for published results; instead, we always to try to keep them
up-to-date with the latest methods, as time permits.

It should actually be extremely easy for someone with a minimum of
experience with Kaldi to run the recipes from ami/s5b, because your recipe
is basically a clone of the AMI recipe.  So I don't feel like I'm asking
for a lot.

vesis84:
Hi Dan, should I take a look at the `nnet1' scripts? AFAIK, they are not all that bad, after adding the support of i-vectors, the results are less than 1% WER behind your best system...

danpovey:
Sure, you could have a look if you want.  I haven't 100% decided what to do
about this PR.
It doesn't change that I still want to emphasize the nnet3-based recipes.
Even if the WER difference is small, the best nnet3 systems will be many
times faster to decode than the corresponding nnet1 systems, plus it's a
single pass of decoding and can be done online.  And it's being more
actively updated.

vesis84:
Hi, I'd consider removing the 'local/nnet/run_dnn_lda_mllt.sh', unless it is necessary to have it there. (in AMI it was required for the MDM condition, where we could not estimate the per-speaker fMLLR transforms as there is no diarization and single wav contains speech from several speakers). Is there a same or similar reason to have it?)

Then, it seems that I can't do the in-line comments, did the github online-app changed recently? The on-line editor seems to be a new feature, it it a replacement for the in-line comments?

danpovey:
If by inline comments you mean where you click on the 'conversation' tab,
click on a line and comment, I don't think that functionality was taken
away..
consider the 'start a review' button.

vesis84:
Aha, now I see it, thanks. In the end I did few cosmetic changes to the script 'local/nnet/run_dnn.sh', otherwise it looks good...

egs/swc/s5b/RESULTS.txt
```diff
@@ -0,0 +1,81 @@
...
+- DNN
+%WER 75.6 | 9338 58361 | 28.7 50.2 21.1 4.3 75.6 79.5 | -23.266 | exp/mdm8/SA1/dnn4_pretrain-dbn_dnn/decode_dev_30k-4gram-int-swc.swc.v2.30k-combilex.4g-int.arpa.pr1-7/ascore_10/dev_o4.ctm.filt.sys
+%WER 77.4 | 9069 56260 | 26.4 49.8 23.8 3.9 77.4 81.8 | -23.533 | exp/mdm8/SA1/dnn4_pretrain-dbn_dnn/decode_eval_30k-4gram-int-swc.swc.v2.30k-combilex.4g-int.arpa.pr1-7/ascore_12/eval_o4.ctm.filt.sys
+
+- DNN+SMBR
+%WER 73.4 | 9373 58361 | 30.7 47.7 21.6 4.1 73.4 78.5 | -22.653 | exp/mdm8/SA1/dnn4_pretrain-dbn_dnn_smbr/decode_dev_30k-4gram-int-swc.swc.v2.30k-combilex.4g-int.arpa.pr1-7_it4/ascore_10/dev_o4.ctm.filt.sys
+%WER 75.4 | 10621 56091 | 28.8 47.7 23.5 4.2 75.4 69.1 | -22.909 | exp/mdm8/SA1/dnn4_pretrain-dbn_dnn_smbr/decode_eval_30k-4gram-int-swc.swc.v2.30k-combilex.4g-int.arpa.pr1-7_it4/ascore_11/eval_o4.ctm.filt.sys
+
+
+======================
```
vesis84: Aha, I see, so all the resutls are with the 'nnet1' recipe. Are there also some experiments with Dan's best `nnet3' system? Also, the speaker adaptation does not bring the usual 2-3% improvements, not even for for the IHM recipe, can it be explained somehow?

egs/swc/s5b/local/chain/multi_condition/run_tdnn_lstm.sh
```diff
@@ -0,0 +1 @@
+tuning/run_tdnn_lstm_1a.sh
```
vesis84: There are too many: 'local/chain/tuning/run_tdnn_lstm_*.sh' scripts. I would include there single script, which provides the best results (Please ask Dan or Yenda, which one it is).

egs/swc/s5b/local/nnet3/run_ivector_common.sh
```diff
@@ -0,0 +1,198 @@
+#!/bin/bash
```
vesis84: What is the difference between: 'local/nnet3/multi_condition/run_ivector_common.sh' and 'local/nnet3/run_ivector_common.sh'. To me it looks like a duplication of code, but if it is in other setups as well, it still can be okay.

egs/swc/s5b/local/score.sh
```diff
@@ -0,0 +1,85 @@
...
+
+data=$1
+
+mic=$(echo $data | awk -F '/' '{print $2}')
+case $mic in
+##  ihm*)
+##    echo "Using sclite for IHM (close talk),"
+##    eval local/score_asclite.sh --asclite false $orig_args
+##  ;;
+  sdm*)
```
vesis84: This could be shortened by enumeration: 'sdm*|mdm*|DSBF*|SDBF*|...)'

egs/swc/s5b/local/score_asclite.sh
```diff
@@ -0,0 +1,120 @@
...
+for f in $data/stm $data/glm $lang/words.txt $lang/phones/word_boundary.int \
+     $model $data/segments $data/reco2file_and_channel $dir/lat.1.gz; do
+  [ ! -f $f ] && echo "$0: expecting file $f to exist" && exit 1;
+done
+
+name=`basename $data`; # e.g. eval2000
+
+mkdir -p $dir/ascoring/log
+
+if [ $stage -le 0 ]; then
```
vesis84: Perhaps, you can drop the 'reverse' thing, unless you really support the `ping-pong' decoding from Mirko Hannemann.

egs/swc/s5b/local/swc_ihm_data_prep.sh
```diff
@@ -0,0 +1,250 @@
...
+
+# Copyright 2016, The University of Sheffield (Author: Yulan Liu)
+# SWC IHM data preparation (train, dev and eval), as well as scoring pareparation.
+# Apache 2.0
+
+# To be executed from one directory above this script.
+
+. path.sh
+
+export LC_ALL=C
```
vesis84: The 'export LC_ALL=C' should be already done in 'path.sh'.

egs/swc/s5b/local/swc_ihm_data_prep.sh
```diff
@@ -0,0 +1,250 @@
...
+mkdir -p $dir
+
+
+# ---------------------------------------------------------------------------
+# Explanation of the "MODE":
+# 
+#       SA: stand-alone
+#       AD: adaptation
+#
+# SWC1, SWC2 and SWC3 all have three strips: A, B, C.
```
vesis84: What is the A,B,C? Are those the subsets of SWC* ? Also, it seems that there are many 'data/*' folders. Would it be possible to merge them into the standard 'train' 'dev' 'eval' dirs? The subsets can be defined in the 'STM' by using the 'subset identifiers enclosed in angle brackets' ('sclite' can print the WER for all the subsets in the *.lur file). ? Or is there a practical reason to have them separately (like different conditions => different processing)?

egs/swc/s5b/path.sh
```diff
@@ -0,0 +1,28 @@
...
+export PATH=$PWD/utils/:$KALDI_ROOT/src/bin:$KALDI_ROOT/tools/openfst/bin:$KALDI_ROOT/src/fstbin/:$KALDI_ROOT/src/gmmbin/:$KALDI_ROOT/src/featbin/:$KALDI_ROOT/src/lm/:$KALDI_ROOT/src/sgmmbin/:$KALDI_ROOT/src/sgmm2bin/:$KALDI_ROOT/src/fgmmbin/:$KALDI_ROOT/src/latbin/:$KALDI_ROOT/src/nnetbin:$KALDI_ROOT/src/nnet2bin/:$KALDI_ROOT/src/nnet3bin/:$KALDI_ROOT/src/kwsbin:$KALDI_ROOT/src/online2bin/:$KALDI_ROOT/src/ivectorbin/:$KALDI_ROOT/src/lmbin/:$PWD:$PATH
+export LC_ALL=C
+
+LMBIN=$KALDI_ROOT/tools/irstlm/bin
+SRILM=$KALDI_ROOT/tools/srilm/bin/i686-m64
+BEAMFORMIT=$KALDI_ROOT/tools/BeamformIt-3.51
+
+export PATH=$PATH:$LMBIN:$BEAMFORMIT:$SRILM
+
+export PERL5LIB=/share/spandh.ami1/sw/lib/std/libperl4-corelibs-perl/Perl4-CoreLibs-0.003/lib:${PERL5LIB}
```
vesis84: The folders '/share' are specific to your cluster. You have to avoid this! The scripts must prepared to run also for  other people... Otherwise, they would become unhappy either with you or with the kaldi maintainers...

egs/swc/s5b/run.sh
```diff
@@ -0,0 +1,286 @@
...
+mic=ihm
+MODE='SA1'      # Other options: 'SA2', 'AD1', 'AD2'
+fastrun=1	# 1: skip unnecessary decoding 
+spkadapt=1	# 1: include speaker adaptation
+nj=20 		# defalut number of parallel jobs,
+stage=0
+
+. utils/parse_options.sh
+
+base_mic=$(echo $mic | sed 's/[0-9]//g') # sdm, ihm or mdm
```
vesis84: This could be replaced by bash string-splits: ${mic:0:3} -> mdm , ${mic:3} -> 8

egs/swc/s5b/run.sh
```diff
@@ -0,0 +1,286 @@
...
+
+# Path where SWC gets downloaded (or where locally available):
+SWCDIR=$PWD/wav_db # Default, 
+case $(hostname -d) in
+  minigrid.dcs.shef.ac.uk) SWCDIR=/share/spandh.ami1/usr/yulan/sw/kaldi/dev/kaldi-trunk/egs/swc/swc ;; # UoS
+esac
+
+
+# Check microphone type
+if [ "$base_mic" == "ihm" ] ; then		# Individual headset microphone
```
vesis84: This logic seems to be a bit too complicated... Also, can it be merged with the 'stage 2' ? What is the purpose of 'mic_ch' and 'PROCESSED_DIR'? Is it used only in the data preparation? You could also explain the role of the variables by comments...

egs/swc/s5b/run_prepare_shared.sh
```diff
@@ -0,0 +1,59 @@
...
+rm tmp
+gawk '{$1="";print $0}' data/local/dict/lexicon.txt | sed 's/ /\n/g' | sort -u | awk '{if (NF==1){print $0}}' |  grep -v sil | grep -v oov  >  data/local/dict/nonsilence_phones.txt
+echo 'sil' > data/local/dict/optional_silence.txt
+echo -e "sil\noov" > data/local/dict/silence_phones.txt 
+utils/prepare_lang.sh data/local/dict "<unk>"  data/local/lang  $LMDIR
+
+
+# You need to download the LM as $SRCLM manually from the SWC
+# website "http://mini-vm20.dcs.shef.ac.uk/swc/LM.html",
+SRCLM=/share/spandh.ami1/asr/dev/mtg/swc/lib/nets/swc.v2.30k/30k-4gram-int-swc.swc.v2.30k-combilex.4g-int.arpa.gz
```
vesis84: $SRCLM is specific to your cluster.

egs/swc/s5b/local/chain/compare_wer_general.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
danpovey: @vesis84, this script is used for preparing tables for comparing the chain systems.. it was copied from ami/s5b.

danpovey:
I am requesting the corpus from Sheffield so that someone here can get this pull request into shape.  I think @wings619 doesn't have enough time (or Kaldi experience) to do it.

egs/swc/s5b/local/swc_ihm_data_prep.sh
wings619: @vesis84 The A, B, C are the so-called three "strips" of each file, they are the building units of dataset. With different research focus, different dataset could be made by selecting different strips combination in train, dev and eval. I think I had explained it in the comments somewhere. Therefore the STM file has the definition of strip, and the data preparation script hard coded a few recommended dataset definition. To avoid unexpected overwriting, I extended all the folder structure by one level vertically to distinguish different datasets.

egs/swc/s5b/RESULTS.txt
wings619: @vesis84 unfortunately there are no results with nnet3. When I wanted to re-run the experiments with nnet3, I realized that I need to first solve a lot of infrastructure issues with the libraries and paths in our grid. I did not have enough time to do everything so I decided to leave nnet3 for someone else in the future. Actually I am leaving Sheffield and I will start my new job in a few days. The whole last week I was engaged in the visa application and relocation things. So I am very sorry for the very slow reaction to the comments. Thank you and Dan a lot for reviewing the codes. I will do my best to catch up in coming days.

By the way, I am not very sure why the speaker adaptation doesn't work.

egs/swc/s5b/path.sh
wings619: @vesis84 Indeed! The cmd.sh and path.sh are both the direct copy from our setup. I should have cleaned them before uploading. Will do that later.

danpovey:
We can update the recipe here at Hopkins when we hear back from Thomas Hain
with how to get the dataset; I sent him a signed agreement about a week ago.

wings619:
@danpovey Dan thank you so much for helping out kindly! It is great that someone outside of Sheffield could help to run the scripts so that we could make sure things do run. 

Please talk with Thomas directly regarding the data, and wish you good luck! What I can do with that is very limited.

Once you get the data, if you find any files missing, please let me know. I am uploading the LM and dictionary to the SWC website now.

PR #1453, opened 2017-02-24
upgrade the sox calls to use more compatible options

jtrmal:
@vijayaditya @pswietojanski 
I did a run on the s5b recipe until mono-training and it seemed to work fine as far as I can tell.
do you wanna do more careful checks?


vijayaditya:
Could you just confirm that the dynamic range of the sox output is the same as before ?

Only concern I have is that the nnet3 models which do not have mean variance normalization might be affected if there is any change in the dynamic range of the waveforms.

vijayaditya:
otherwise LGTM.

danpovey:
i think it should be safe, I've found online refs to an error message
saying:
Error: /usr/bin/sox WARN sox: Option `-s' is deprecated, use `-e signed-
integer' instead

vijayaditya:
OK merging.

PR #1457, opened 2017-02-26
fix MacOSX incompatibilities in calls of paste

jtrmal:
I didn't test but should work.


johnjosephmorgan:
I found no more errors in my test.
Thanks,
John

jtrmal:
Great, thanks

PR #1461, opened 2017-02-27
SGE  style options replaced by the new-style options

jtrmal:
I did most of it  using sed  and other friends. I also fixed whitespacein the files being modified and one or two shell scripting small bugs I noticed when doing thesse changes.
probably easier to review on per-commit bsis (as the changes will be pretty systematic that way) than on the whole PR basis.


jtrmal:
* I didn't test the changes
* I didn't fix the babel recipes (I have it on my todo, but it's much more work than this)
* shall we add some code to the queue.pl/run.pl/others to inform the users the old-style options are not preferred anymore?

danpovey:
cool.  I'm OK with adding code to queue.pl etc. that will print warnings
about thing being deprecated.  but might be easier to do in a separate
commit?
I'll merge this.

PR #1483, opened 2017-03-08
Fixed absolute path used in voxforge eg script.

baali:


danpovey:
hm, that directory was supposed to be an example, better to set it to something that just doesn't exist and let the user decide.  and you could clarify in the comment at the top of run.sh that DATA_ROOT is set in path.sh.

baali:
There is already a comment in run.sh related to setting DATA_ROOT in path.sh:
`#       1. Setting the DATA_ROOT variable to point to a directory with enough free`
So you would rather comment out export DATA_ROOT line itself?
`export DATA_ROOT="$KALDI_ROOT/egs/voxforge/s5/voxforge"    # e.g. something like /media/secondary/voxforge`
And later there is a if check in path.sh on this variable which prompts user to set it correctly.

danpovey:
but the comment in run.sh doesn't say where it is.
yes, it might be best to just comment out the line.

baali:
Just fixed that.

danpovey:
and change run.sh to mention that DATA_ROOT is in path.sh, e.g. change
` Setting the DATA_ROOT variable to point to a directory with enough free`
to
` Setting the DATA_ROOT variable in path.sh to point to a directory with enough free`

baali:
Sorry for that mixup in comment grammar.

danpovey:
thanks.

PR #1490, opened 2017-03-11
patch the openfst 1.6.1 to compile in gcc 6.3

jtrmal:


rvenkidapathy:
Observation:

I tried it with ubuntu 16.10 lastnight with g++ 5.1 which led to the issue

I tried with ubuntu 16.04 now it compiles.

Thanks.

Ravi

tools/Makefile
```diff
@@ -77,6 +77,7 @@ endif
 
 openfst-$(OPENFST_VERSION): openfst-$(OPENFST_VERSION).tar.gz
 	tar xozf openfst-$(OPENFST_VERSION).tar.gz
+	[ -f ./extras/openfst-$(OPENFST_VERSION).patch ] && cd openfst-$(OPENFST_VERSION) && patch -p1 < ../extras/openfst-$(OPENFST_VERSION).patch
```
danpovey: shouldn't this line begin with a "-" (after the tab), so that the bad return status will be ignored if this version doesn't have the patch file?

jtrmal:
I think this version addresses both your comment and my concern of ignoring the code when the patching fails -- originaly, I didn't add the - there to have make stop in case for some reason patch fails.

danpovey:
will wait for checks.
Apparently this will be fixed in the next OpenFst release which should be
in a week or so.

jtrmal:
yes, I reported this to Kyle a few weeks back
y.

jtrmal:
Closing this, as the patch is a part of the newly released 1.6.2 (see PR #1492 )
y.

jtrmal:
Dan, I have just re-opened this PR.
I can have a look at the 1.6.2 issues later this week (maybe today but do not want to promise). I assume the issues in 1.6.2 relate to namespace lookup or something.

danpovey:
That guy clarified that his issues were actually for all the 1.6.*.  From
looking at the diffs from 1.5.*, I think it was to do with changing
'typedef' to 'using' which the standard says is a no-op.  So just a
straightforward compiler bug I think.  But IDK why it showed up just there.

jtrmal:
ok, then this is useless -- closing again.

PR #1498, opened 2017-03-17
update fisher_swbd recipe

naxingyu:
update fisher_swbd to adopt new path and srilm installer.

### DIFF
--- a/egs/fisher_swbd/s5/local/fisher_train_lms.sh
+++ b/egs/fisher_swbd/s5/local/fisher_train_lms.sh
@@ -6,7 +6,7 @@
 . path.sh
 
 text=data/train_all/text
-lexicon=data/local/dict_nosp/lexicon.txt 
+lexicon=data/local/dict_nosp/lexicon.txt
 
 for f in "$text" "$lexicon"; do
   [ ! -f $x ] && echo "$0: No such file $f" && exit 1;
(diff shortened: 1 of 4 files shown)


tools/extras/install_srilm.sh
```diff
@@ -14,7 +14,7 @@ if [ ! -d liblbfgs-1.10 ]; then
 fi
 
 # http://www.speech.sri.com/projects/srilm/download.html
-if [ ! -f srilm.tgz ]; then
+if [ ! -f srilm-1.7.2.tar ]; then
```
jtrmal: @naxingyu , I won't comment on the recipe changes per se (not familiar with that)
But I'm not fond of the changes here.
a) what srilm pages do offer for download if you go there is srilm-1.7.2.tar.gz, i.e. I'm not sure where the tar file comes from, but that script won't work for other people now.
b) I'm still in favor of asking users to do a symlink or copy the file as srilm.tgz.
I think it makes sense as we cannot download the file automatically -- so people can use whatever version they have available and do not have to undergo the downloading process every time.  Which I realize it's not true now as I'm thinking about it, as we need 1.7.1 at least, because of the maxent LMs. So I don't really object to having a fixed version and bump it now and then, but I think it's less convenient for us.
c) "$SRILM" is preferred, IMO.
 It's because of cases where the users have space (white character) in the path somewhere. Or perhaps not, but is there any rationale behind changing it?

naxingyu:
> 在 2017年3月17日，23:58，jtrmal <notifications@github.com> 写道：
> 
> @jtrmal commented on this pull request.
> 
> In tools/extras/install_srilm.sh <https://github.com/kaldi-asr/kaldi/pull/1498#discussion_r106684126>:
> 
> > @@ -14,7 +14,7 @@ if [ ! -d liblbfgs-1.10 ]; then
>  fi
>  
>  # http://www.speech.sri.com/projects/srilm/download.html
> -if [ ! -f srilm.tgz ]; then
> +if [ ! -f srilm-1.7.2.tar ]; then
> @naxingyu <https://github.com/naxingyu> , I won't comment on the recipe changes per se (not familiar with that)
> But I'm not fond of the changes here.
> a) what srilm pages do offer for download if you go there is srilm-1.7.2.tar.gz, i.e. I'm not sure where the tar file comes from, but that script won't work for other people now.
> 
The tar file is what I got from the download page, last month. I don’t if they have changed the offered file from tgz to tar, but that’s what I got.
> b) I'm still in favor of asking users to do a symlink or copy the file as srilm.tgz.
> I think it makes sense as we cannot download the file automatically -- so people can use whatever version they have available and do not have to undergo the downloading process every time. Which I realize it's not true now as I'm thinking about it, as we need 1.7.1 at least, because of the maxent LMs. So I don't really object to having a fixed version and bump it now and then, but I think it's less convenient for us.
> c) "$SRILM" is preferred, IMO.
> 
I agree. Will change it back.

danpovey:
Check with a couple ways of getting it, maybe some browsers or OS's
automatically unzip things.

naxingyu:
You are right… I’ll change the extension. And do I keep the version numbers or not?

> 在 2017年3月18日，09:12，Daniel Povey <notifications@github.com> 写道：
> 
> Check with a couple ways of getting it, maybe some browsers or OS's
> automatically unzip things.
> 
>

danpovey:
Let's not keep version numbers, it will be more compatible with existing
setups.

naxingyu:
OK. I’ve reverted the changes on installation script.

> 在 2017年3月18日，09:18，Daniel Povey <notifications@github.com> 写道：
> 
> Let's not keep version numbers, it will be more compatible with existing
> setups.
>

egs/fisher_swbd/s5/local/fisher_train_lms.sh
```diff
@@ -20,32 +20,16 @@ done
...
-   make || exit 1;
-   echo Done making the kaldi_lm tools
- fi
-) || exit 1;
-
-mkdir -p $dir
 
+kaldi_lm=`which train_lm.sh`
+if [ ! -x $kaldi_lm ]; then
+  echo "train_lm.sh is not found. Checkout tools/extra/install_kaldi_lm.sh"
```
danpovey: Say "look at" not "checkout".

naxingyu:
Changed and squashed.

danpovey:
Merging. Thanks!

PR #1499, opened 2017-03-17
Multistream

Phani-Nidadavolu:


danpovey:
Aargh!  There is too much stuff going on here.  I would only, even in principle, be willing to merge a small percentage of this.  But first explain at the top level what this is about, and what the improvements are on what setup.  I.e. what is the benefit to be had from reorganizing and then merging this.

vijayaditya:
@Phani-Nidadavolu Also this PR seems to be composed of individual pieces which are not necessarily tightly coupled. So it would be better to either isolate these into individual PRs, and submit each one with the details Dan had requested. 251 files being changed in a single PR would be very hard to review.

Also if it not very important for you to preserve the entire history of commits I would recommend isolating your changes to a single commit, once you have separated it into individual PRs. You can accomplish this using "git rebase -i" command.

vijayaditya:
I found lot of Keras code and things related to CIFAR data prep in the code.. if you are using an external toolkit I would strongly suggest that you just add the path to the external repository in your Kaldi specific code and import things necessary. Check for the things necessary before you launch the recipe. 

Please do not add Keras code into Kaldi !!!

danpovey:
I think it's unlikely that we'd accept Keras-based recipes into the
repository; but never say never.

vijayaditya:
Oops I just meant import from an external path, and don't add the Keras core code into the Kaldi repo. I didn't mean to dissuade the use of Keras modules.

danpovey:
Actually by making my comment above, I meant go further than you and say
that even if Keras is included as an external toolkit and not as part of
the repo, we probably wouldn't want to merge that recipe.  But of course if
the results are great, it might be a different story.

Phani-Nidadavolu:
Hello,

Sorry for the trouble. This PR was by mistake. Please ignore.

PR #1501, opened 2017-03-18
Sort cuda kernel function declarations to make searching easier.

kangshiyin:
All GPU tests have passed. Nothing should change.


danpovey:
thanks, merging.

PR #1506, opened 2017-03-21
based on 1387

GaofengCheng:
my change is (part of all): 
tedlium/s5_r2_povey_1387/local/chain/tuning/run_tdnn_lstm_1u_1024.sh
tedlium/s5_r2_povey_1387/tdnn_lstm_1u_newschedule_5epoch_1024.sh


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": false
```
danpovey: did you mean to check this in?

src/nnet3/nnet-general-component.cc
```diff
@@ -1374,5 +1374,105 @@ void ConstantComponent::UnVectorize(const VectorBase<BaseFloat> &params) {
...
+  stream << Type()
+         << ", output-dim=" << output_dim_
+         << ", dropout-proportion=" << dropout_proportion_
+         << ", exclusive=" << (exclusive_ ? "true" : "false");
+  return stream.str();
+}
+
+DropoutMaskComponent::DropoutMaskComponent():
+    output_dim_(-1), dropout_proportion_(0.5),
+    exclusive_(false) { }
```
danpovey: I think the 'exclusive' option can be removed (treat it as if it is always false).

egs/tedlium/s5_r2/local/chain/tuning/run_tdnn_lstm_1q.sh
```diff
@@ -0,0 +1,348 @@
+#!/bin/bash
+
+# 1q is as 1p, but add the "dropout-exclusive" option which means that
```
danpovey: in the next week or two, please work on getting a 'clean' pull request, something that we can check in.
The aim is to demonstrate just our optimal recipe, and to get rid of any tuning scripts that are not things
we plan to use in future-- for example, this script, which uses the 'dropout-exclusive' option.

Check again what is the currently recommended script in each case, and make sure that the script you are checking in is modified from that, and uses our recommended dropout setting.  Let's aim to check in
good-performing systems on AMI, Tedlium and Swbd, based on fast TDNN+LSTMs (and probably also
BLSTMs).  In each case there should just be one new script for each type of setup, e.g. just one new script
for Switchboard TDNN+LSTMs.

PR #1507, opened 2017-03-22
Some preliminary work for nnet3 support of convnets

danpovey:
This may be of interest to @galv.
I may ask you for help at some point, either to finish the code or to help compare with CuDNN.
I was inspired to do this after looking at your PR.  My hope is that for the normal fairly small-sized e.g. 3x3 convolution kernels, it will be unnecessary to use CuDNN in order to get optimal speed.  Read convolution.h and you'll see why.

### DIFF
--- a/src/nnet3/nnet-common.cc
+++ b/src/nnet3/nnet-common.cc
@@ -47,10 +47,13 @@ static void WriteIndexVectorElementBinary(
     const std::vector<Index> &vec,
     int32 i) {
   bool binary = true;
-  Index index = vec[i];
+  const Index &index = vec[i];
   if (i == 0) {
+    // we don't use std::abs(index.t) < 125 here because it doesn't have the
+    // right (or even well-defined) behavior for
+    // index.t == std::numeric_limits<int32>::min().
     if (index.n == 0 && index.x == 0 &&
-        std::abs(index.t) < 125) {
+        index.t > -125 && index.t < 125) {
       // handle this common case in one character.
       os.put(static_cast<signed char>(index.t));
     } else {  // handle the general case less efficiently.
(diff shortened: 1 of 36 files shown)


galv:
@danpovey I've seen this in my inbox. I haven't looked at this in detail yet, though.

src/nnet3/nnet-common.cc
```diff
@@ -535,7 +535,8 @@ void PrintIntegerVector(std::ostream &os,
   os << "]";
 }
 
-
+// this will be the most negative number representable as int32.
+const int kNoTime = std::numeric_limits<BaseFloat>::min();
```
galv: Shouldn't BaseFloat be int32 here?

danpovey:
yes, thanks..

src/nnet3/nnet-convolutional-component.h
```diff
@@ -0,0 +1,120 @@
...
+  DistributeComponent(int32 input_dim, int32 output_dim) {
+    Init(input_dim, output_dim);
+  }
+  DistributeComponent(): input_dim_(0), output_dim_(0) { }
+  virtual int32 InputDim() const { return input_dim_; }
+  virtual int32 OutputDim() const { return output_dim_; }
+
+  // use the default Info() function.
+  virtual void InitFromConfig(ConfigLine *cfl);
+  virtual std::string Type() const { return "DistributeComponent"; }
```
galv: Should this really return "DistributeComponent"?

danpovey:
that was just a draft, I'll work on that later.

danpovey:
Closing this PR, combined it with batch-norm.

PR #1509, opened 2017-03-22
fixes to babel pipeline

jtrmal:
the patch is huge but I tested it on all op3 langs with different configs.
some of the issues emerged when Fred was testing out the pipeline in their setup.
Mostly small changes a fixes -- only one major change is the fix for the race condition in creating the split datadirs


jtrmal:
BTW, everything is localized to s5d/local

PR #1510, opened 2017-03-22
Change return code of extract-rows

happyalu:
num_done > 0 is success. Updated the return statement to return 0 when successful.

### DIFF
--- a/src/featbin/extract-rows.cc
+++ b/src/featbin/extract-rows.cc
@@ -149,7 +149,7 @@ int main(int argc, char *argv[]) {
     KALDI_LOG << "Processed " << num_done << " segments successfully; "
               << "errors on " << num_err;
 
-    return (num_done > 0);
+    return (num_done == 0);
   } catch(const std::exception &e) {
     std::cerr << e.what();
     return -1;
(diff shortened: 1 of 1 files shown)


happyalu:
POSIX utitilies use a zero return value to indicate success and a non-zero value to indicate failure.  This is a pattern that most of the Kaldi utilities also follow, but extract-rows returns non-zero on success which breaks compatibility with bash options like set -e.  Changed the return value to use zero to indicate success. Similar to extract-feature-segments.cc, assumed success when at least one segment is processed.

src/featbin/extract-rows.cc
```diff
@@ -149,7 +149,7 @@ int main(int argc, char *argv[]) {
     KALDI_LOG << "Processed " << num_done << " segments successfully; "
               << "errors on " << num_err;
 
-    return (num_done > 0);
+    return (num_done == 0);
```
danpovey: Yes, that was a bug; that code has not been used much so far.
thanks, but can you please change it to
(num_done > 0 ? 0 : 1)
which is how we normally write it?
returning the truth value causes me to do a double-take when looking at the code.

happyalu: Thanks.

Changed the expression.

PR #1511, opened 2017-03-22
WIP: Epsilon Backstitch

freewym:
For discussion. Modified on top of #1391


src/nnet3/nnet-chain-training.cc
```diff
@@ -167,6 +188,14 @@ void NnetChainTrainer::UpdateParamsWithMaxChange() {
                   << "UpdatableComponent; change this code.";
       BaseFloat max_param_change_per_comp = uc->MaxChange();
       KALDI_ASSERT(max_param_change_per_comp >= 0.0);
+      if (nnet_config.backstitch_training_scale > 0.0 &&
```
danpovey: You need to do a much better job, via comments, of explaining what is going on here.   Even I don't understand this.  It's possible that the code structure here is the problem, and it will have to be changed so that it's easier to understand.  

Here is my proposal.

We take UpdateParamsWithMaxChange out of the class (probably declare it in nnet-utils.h), remove the is_backstitch_step  argument, take out the ScaleNnet() from it [that can be done separately by the 
calling code], and give it this interface:
```
/**
       This function does the operation '*nnet += scale * delta_nnet', while respecting
       any max-parameter-change (max-param-change) specified in the updatable
       components, plus the global max-param-change specified as 'max_param_change'.

       With max-changes taken into account, the operation of this function is equivalent
       to the following, although it's done more efficiently:
        \code
             Nnet temp_nnet(delta_nnet);
             ScaleNnet(1.0 / max_change_scale, &temp_nnet);
             [ Scale down parameters for each component of temp_nnet as needed so
               their Euclidean norms do not exceed their per-component max-changes ]
             [ Scale down temp_nnet as needed so its Euclidean norm does not exceed
               the global max-change ]
             ScaleNnet(max_change_scale, &temp_nnet);  // undo the previous scaling.
             AddNnet(temp_nnet, scale, nnet);
        \endcode

       @param [in] delta_nnet  The copy of '*nnet' neural network that contains 
                                  the proposed change in parameters.  Normally this will previously
                                  have been set to: (delta_nnet = 
                                  parameter-derivative-on-current-minibatch * learning-rate per parameter),
                                  with any natural gradient applied as specified in the components; but
                                  this may be different if momentum or backstitch are used.
      @param [in] max_param_change  The global max-param-change specified on the 
                                  command line (e.g. 2.0), which specifies the largest change allowed
                                  to '*nnet' in Euclidean norm.  If <= 0, no global max-param-change will
                                  be enforced, but any max-change values specified in the components
                                  will still be enforced;  see UpdatableComponent::MaxChange(), and 
                                  search for 'max-change' in the configs or nnet3-info output).
      @param [in] max_change_scale  This value, which will normally be 1.0, is used to scale
                                  all per-component max-change values and the global 'max_param_change',
                                 before applying them (so we use 'max_param_change * uc->MaxChange()'
                                 as the per-component max-change, and 'max_param_change * max_change_scale'
                                 as the global max-change).
      @param [in] scale  This value, which will normally be 1.0, is a scaling factor used when adding 
                                 to 'nnet', applied after any max-changes.
       @param [in,out] nnet  The nnet which we add to.
*/
void UpdateNnetWithMaxChange(const Nnet &delta_nnet, 
                                              BaseFloat max_param_change,
                                              BaseFloat max_change_scale,
                                              BaseFloat scale, Nnet *nnet);
```

You may have to have the backstitch code in a separate code branch from the regular/momentum
training code, to keep the simple case simple.

src/nnet3/nnet-chain-training.cc
```diff
@@ -217,17 +253,33 @@ void NnetChainTrainer::UpdateParamsWithMaxChange() {
...
+        nnet_config.backstitch_training_epsilon;
+    BaseFloat scale_backstitch =
+        (is_backstitch_step ? -nnet_config.backstitch_training_epsilon :
+        backstitch_ratio);
+    scale_factors.Scale(scale * scale_backstitch);
+    AddNnetComponents(*delta_nnet_, scale_factors, scale * scale_backstitch,
+                      nnet_);
+    scale_backstitch = (is_backstitch_step ? (-backstitch_ratio + 1 +
+        nnet_config.backstitch_training_epsilon) / backstitch_ratio :
+        nnet_config.momentum);
```
danpovey: This code wouldn't work with momentum (you'd probably need a separate copy of the nnet for that), so better to use 0.0 instead of nnet_config.momentum, and assert that momentum is zero.   After restructuring there may be 2 paths for momentum vs backstitch, anyway.

I can't figure out whether this code is correct.  My rule when I can't easily figure out correctness is to ask for the code to be restructured, because if present me can't figure it out, then future me definitely won't be able to, not to mention present or future anyone-else.

freewym:
The reorganized code (alpha=alpha_old+alpha_old^2=0.24, epsilon=0.2) gave the same result as the old backstitch code (alpha_old=0.2).

danpovey:
ok, good; try with smaller epsilon, eg. 0.1 or less, if you haven't already.
[but don't waste CPUs: if it seems to be progressing the same as the old
code for the same parameters, using old runs might be OK.]

freewym:
@danpovey the latest backstitch code has been tested on the chain model.

jtrmal:
btw, I'm testing it as well, actually with combination with dropout. Just
FYI
y.

danpovey:
Yiming, have you modified the code with the random-seed stuff so that it's
fully compatible with dropout?  [Yenda, unless we take care to make the
random seed the same in the 1st and second passes of it, it may not give
 as much improvement as it could.]  Yiming, please get on that right away
if you haven't already done it.

Dan

freewym:
Yes, I have done that in the commit "resets random generator's seed for RandomComponent..."

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +69,80 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
...
+    TrainInternal(chain_eg, *computation, is_backstitch_step);
+  }
+
+  num_minibatches_processed_++;
+}
+
+void NnetChainTrainer::TrainInternal(const NnetChainExample &eg,
+                                     const NnetComputation &computation,
+                                     bool is_backstitch_step) {
+  srand(srand_seed_ + num_minibatches_processed_);
```
danpovey: You shouldn't do this if you're not doing backstitch, because it's not free of cost.
So only execute these two statements if you're doing backstitch; calling it outside of (and just before) TrainInternal will make more sense.

src/nnet3/nnet-training.h
```diff
@@ -158,13 +169,12 @@ class NnetTrainer {
 
   ~NnetTrainer();
  private:
-  void ProcessOutputs(const NnetExample &eg,
-                      NnetComputer *computer);
+  void TrainInternal(const NnetExample &eg,
```
danpovey: we'll need documentation for this function.

danpovey:
I just noticed that this PR is to master, and it isn't even up to date with master.  Can you please first try to merge with master, and then try to merge with the kaldi_52 branch?  We'll be wanting to merge this with kaldi_52.  You will likely have conflicts.
Obviously save this branch, as bad things can  happen when merges fail.

PR #1513, opened 2017-03-23
Unexpected behaviour of ConvertStringToReal in Android

jcsilva:
ConvertStringToReal was implemented using strtod and strtof. These 2 functions
depend on locale. When trying to run Kaldi in Android, you can't
set locales from native code (https://code.google.com/p/android/issues/detail?id=57313), and these functions have an unexpected behaviour.

This PR uses stringstream for dealing with it. You may see that strtod and strtof is still in the source code. It was necessary because there are two tests that verify if NAN and INF were converted to float and stringstream doesn't handle them.

### DIFF
--- a/src/util/text-utils-test.cc
+++ b/src/util/text-utils-test.cc
@@ -186,10 +186,12 @@ void TestConvertStringToReal() {
   KALDI_ASSERT(!ConvertStringToReal("GARBAGE inf", &d));
   KALDI_ASSERT(!ConvertStringToReal("GARBAGEinf", &d));
   KALDI_ASSERT(!ConvertStringToReal("infGARBAGE", &d));
+  KALDI_ASSERT(!ConvertStringToReal("inf_GARBAGE", &d));
   KALDI_ASSERT(!ConvertStringToReal("inf GARBAGE", &d));
   KALDI_ASSERT(!ConvertStringToReal("GARBAGE infinity", &d));
   KALDI_ASSERT(!ConvertStringToReal("GARBAGEinfinity", &d));
   KALDI_ASSERT(!ConvertStringToReal("infinityGARBAGE", &d));
+  KALDI_ASSERT(!ConvertStringToReal("infinity_GARBAGE", &d));
   KALDI_ASSERT(!ConvertStringToReal("infinity GARBAGE", &d));
   KALDI_ASSERT(ConvertStringToReal("1.#INF", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("-1.#INF", &d) && d < 0 && d - d != 0);
@@ -218,6 +220,7 @@ void TestConvertStringToReal() {
   KALDI_ASSERT(!ConvertStringToReal("GARBAGE nan", &d));
   KALDI_ASSERT(!ConvertStringToReal("GARBAGEnan", &d));
   KALDI_ASSERT(!ConvertStringToReal("nanGARBAGE", &d));
+  KALDI_ASSERT(!ConvertStringToReal("nan_GARBAGE", &d));
   KALDI_ASSERT(!ConvertStringToReal("nan GARBAGE", &d));
   KALDI_ASSERT(!ConvertStringToReal("GARBAGE 1.#QNAN", &d));
   KALDI_ASSERT(!ConvertStringToReal("GARBAGE1.#QNAN", &d));
--- a/src/util/text-utils-test.cc
+++ b/src/util/text-utils-test.cc
@@ -243,11 +243,16 @@ void TestNan() {
 template<class Real>
 void TestInf() {
   Real d;
-  std::ostringstream strs;
-  strs << exp(10000);
-  std::string s = strs.str();
+  std::ostringstream pos_strs;
+  pos_strs << exp(10000);
+  std::string pos = pos_strs.str();
 
-  KALDI_ASSERT(ConvertStringToReal(s, &d) && d > 0 && d - d != 0);
+  std::ostringstream neg_strs;
+  neg_strs << -exp(10000);
+  std::string neg = neg_strs.str();
+
+  KALDI_ASSERT(ConvertStringToReal(pos, &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal(neg, &d) && d < 0 && d - d != 0);
 }
 
 
--- a/src/util/text-utils-test.cc
+++ b/src/util/text-utils-test.cc
@@ -178,6 +178,10 @@ void TestConvertStringToReal() {
   KALDI_ASSERT(ConvertStringToReal("Inf", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("INF", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("InF", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("infinity", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("-infinity", &d) && d < 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("1.#INF", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("-1.#INF", &d) && d < 0 && d - d != 0);
 
   KALDI_ASSERT(ConvertStringToReal("nan", &d) && d != d);
   KALDI_ASSERT(ConvertStringToReal("+nan", &d) && d != d);
@@ -185,6 +189,8 @@ void TestConvertStringToReal() {
   KALDI_ASSERT(ConvertStringToReal("Nan", &d) && d != d);
   KALDI_ASSERT(ConvertStringToReal("NAN", &d) && d != d);
   KALDI_ASSERT(ConvertStringToReal("NaN", &d) && d != d);
+  KALDI_ASSERT(ConvertStringToReal("1.#QNAN", &d) && d != d);
+  KALDI_ASSERT(ConvertStringToReal("-1.#QNAN", &d) && d != d);
 }
 
 
--- a/src/util/text-utils-test.cc
+++ b/src/util/text-utils-test.cc
@@ -173,47 +173,58 @@ void TestConvertStringToReal() {
 
   // it also works for inf or nan.
   KALDI_ASSERT(ConvertStringToReal("inf", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal(" inf", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("inf ", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal(" inf ", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("+inf", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("-inf", &d) && d < 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("Inf", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("INF", &d) && d > 0 && d - d != 0);
   KALDI_ASSERT(ConvertStringToReal("InF", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("infinity", &d) && d > 0 && d - d != 0);
+  KALDI_ASSERT(ConvertStringToReal("-infinity", &d) && d < 0 && d - d != 0);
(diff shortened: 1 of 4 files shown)


src/util/text-utils.cc
```diff
@@ -221,20 +219,43 @@ bool ConvertStringToReal(const std::string &str,
...
+      while (isspace(*end)) end++;
+    if (end == this_str || *end != '\0' || errno != 0)
+      return false;
+
+    *out = d;
+    return true;
+  }
+
+  // if istringstream was successfully converted to a number, we
+  // need to garantee that there is not any other token in str
```
danpovey: spelling issue: guarantee.

src/util/text-utils.cc
```diff
@@ -243,12 +264,38 @@ bool ConvertStringToReal(const std::string &str,
...
+
+  if (!(i >> *out)) {
+    // Number conversion failed.
+    // I know it's not a numeric value.
+    // These lines are here just do check if str is nan or inf
+    const char *this_str = str.c_str();
+    char *end = NULL;
+    errno = 0;
+
+    float d = KALDI_STRTOF(this_str, &end);
```
danpovey: It seems to me that if we're going to change the code in this way, it would be better to get rid of the strtof and strtod functions entirely, and remove the ifdefs related to them, since they are a compatibility headache, plus it's not clear to me whether their behavior would depend on local in these cases.

Could you try writing code to handle those cases without using strtof/strtod?
Based on
http://stackoverflow.com/questions/11420263/is-it-possible-to-read-infinity-or-nan-values-using-input-streams
I think you should make sure to test for inf, -inf, +inf, nan, -nan, +nan, and also case variants with NaN and Inf.  The testing may need to be extended to cover more cases than before.

src/util/text-utils.cc
```diff
@@ -188,67 +190,77 @@ inline bool is_nan_text(const std::string &in, const std::string &prefix) {
...
+
+    std::string c;
+    in_.clear();
+    if (!(in_ >> c)) return *this; //If the stream is broken even before trying to read from it, it's pointless to try.
+
+    std::transform(c.begin(), c.end(), c.begin(), ::tolower); // transform c to lowercase.
+
+    if(inf_nam_map.find(c) != inf_nam_map.end()) {
+      x = inf_nam_map[c];
+      if(neg) x = -x;
```
danpovey: need space after if and for `} else {`... try running cpplint.py on this.

src/util/text-utils.cc
jcsilva: I wrote the code and the tests following your suggestion. 
I also completely removed strtof and strtod from the source code.

src/util/text-utils.cc
```diff
@@ -188,67 +190,77 @@ inline bool is_nan_text(const std::string &in, const std::string &prefix) {
...
+{
+ public:
+
+  number_istream (std::istream &i) : in_(i) {}
+
+  number_istream & operator >> (T &x) {
+    bool neg = false;
+    char c;
+    if (!in_.good()) return *this;
+    while (isspace(c = in_.peek())) in_.get();
```
danpovey: not sure if it's faster to do 
`in_ >> std::ws;`
or some similar magic.

src/util/text-utils.cc
```diff
@@ -188,67 +190,77 @@ inline bool is_nan_text(const std::string &in, const std::string &prefix) {
...
-    return true;
-#endif  // defined(_MSC_VER)
+ private:
+  std::istream &in_;
+
+  number_istream & parse_on_fail (T &x, bool neg)
+   {
+    std::map<std::string, T> inf_nam_map;
+    // we'll keep just lowercase values.
+    inf_nam_map["inf"] = std::numeric_limits<T>::infinity();
```
danpovey: do you mean nan_map instead of nam_map?  If 'nam' means name, please don't abbreviate like that, google syle guide doesn't like that.

danpovey:
thanks- will probably merge to-morrow.

src/util/text-utils.cc
```diff
@@ -188,67 +190,76 @@ inline bool is_nan_text(const std::string &in, const std::string &prefix) {
   return true;
 }
 
-template<typename T>
-bool convert_special_number(const std::string &str, T *out) {
-  if (stricmp(str, "infinity") || stricmp(str, "inf") ||
-      starts_with(str, "1.#INF")) {
```
danpovey: I'm looking at this deleted code and it seems to be using alternative printed representations of nan.
Maybe this is how Window prints them.  I think it would be best to modify the code you've written to call
this function in cases where it's not a regular number, and to extend the test code to support the
cases that this function seems to be trying to handle.
This will make it less likely that we'll experience regressions on Windows.

jcsilva:
PR was updated to support alternative representations of nan and inf.

Please, note that:

1. it uses starts_with. So, "1.#INF" and "1.#INFINITY" or "1#.INFinity" or "1#.INF something else" should give the same results. I think it can be "dangerous" sometimes... but it was the original behaviour. Should it be changed?
2. It doesn't check for initial spaces in "1.#INF" or "1.#QNAN". Again, it was the original behaviour. Should it be changed?
3. There is no +1.#INF or +1.#INF. The same in the original commit. Should it be changed?

danpovey:
All those things should be OK.

src/util/text-utils.cc
```diff
@@ -191,14 +191,16 @@ inline bool is_nan_text(const std::string &in, const std::string &prefix) {
 template<typename T>
 bool convert_special_number(const std::string &str, T *out) {
```
danpovey: Please name this according to the Google style guide, i.e. ConvertSpecialNumber, and put 'static' before bool so that it's clear it's a local function.

src/util/text-utils.cc
```diff
@@ -208,47 +210,38 @@ bool convert_special_number(const std::string &str, T *out) {
...
-                         float *out) {
-  const char *this_str = str.c_str();
-  char *end = NULL;
-  errno = 0;
-
-#ifdef _MSC_VER
-  // TODO: check if the new MSVC already supports it
-  // depending on claims of the C++11 support, it should have
-  if (convert_special_number(str, out))
+  // deals with nan and inf
```
danpovey: If I understand this correctly, it does the nan stuff first, which is quite inefficient as it's not the normal case and it involves constructing maps, etc.   This function is called quite often.
Please make sure that the 'normal case' of this function, where it contains an actual number, is as efficient as possible.  So try to have it so that you just construct an istringstream, input it via >> to a float or double, and check its statuses and then return, as the normal code path.  All that stuff about Trim() should be unnecessary in the principal code path of the function.  Reading via >> to a float should probably skip over spaces by itself; if not, use >> std::ws or something like that.

src/util/text-utils.cc
```diff
@@ -165,90 +167,89 @@ inline bool starts_with(const std::string &in, const std::string &prefix) {
...
+  if (iss.fail()) {
+    // Number conversion failed.
     return false;
-  *out = f;
+  }
+
+  // if something remains in the istringstream,
+  // we'll check if it is #INF or #QNAN (to deal with
+  // MSVC stuffs), or if it is some garbage text.
+  if (iss.tellg() != -1) {
```
danpovey: I don't understand why you are putting the different types of infinity (the MSVC ones and the regular ones) in different places.  It makes the code quite ugly.  And also I think you should be skipping over other kinds of space character, not just ' '.

Also, it seems to me that the way this code is written, it would falsely accept strings that consisted of some kind of garbage followed by `#INF`: for instance, `hello #INF`.

danpovey:
BTW, can you please clarify the problem that this PR was intended to fix?
I followed the link https://code.google.com/p/android/issues/detail?id=57313
and it's saying that it always using the US locale and you can't set any other.  Isn't this a good thing?  Or is it the case that if you have a French phone it would always use the French locale?
Did you encounter a particular problem?

jcsilva:
This problem annoyed me a few times.

First, we tried to use the ParseOptions class in one application we were writing. It worked fine in some computers. But, when we tried to run it in a machine where LOCALE was pt_BR, it got crazy with real numbers (because decimal point is ',' (comma)). And it was not easy to find out the problem, because you the file is parsed and the number is read, but it is value is wrong. The workaround we used was _std::setlocale(LC_ALL, "en_US.UTF-8");_  in the begging of our code.

Another problem happened a few days ago. I tried to run online2-wav-nnet2-latgen-faster.cc in my Android phone (which was bought in Brazil) to validate the cross-compilation. I got similar problems when reading the variable --posterior-scale (which is in the file ivector_extractor.conf). I tried to use setlocale in my source code, but it didn't give me any useful results. So, I found that link (https://code.google.com/p/android/issues/detail?id=57313) and saw that it was not possible to change the locale from native code... and I thought it would be useful to use something independent of locales to convert from string to float/double. After using istringstream to deal with this convertion, I didn't have problems anymore.

I also thought it would be a useful implementation for other people who might face this kind of problem. This was the reason I wrote this PR. But, please, if you think it is not the best solution, or if it is not useful, abandon this PR. Or, if you have another opinion or suggestion, please tell me.

danpovey:
OK, it seems to me that it's a real problem and this solution seems like it
should work.

danpovey:
@jcsilva, there shouldn't be too much work to do on this..
just reorganize the code a bit so that different types of infinity (including MSVC ones) are handled in the same place.  And can you please add tests where you put various kinds of garbage before and after the different types of infinity, to check that errors are caught as they should be?  Doesn't need all combinations.. .just a reasonable number of them.

jcsilva:
Sorry, I have been busy this week and I didn't have time to work on this.

So, I updated the PR. Just some observations:

1. I have one question about "nan". In the current Kaldi source coude, there is a function called is_nan_text that returns true if it is called like is_nan_text("nanGARBAGE", "nan"), but it is false if it is called as is_nan_text("nan_GARBAGE", "nan"). This function is called when MSVC is used. This was tested before? This is the desired behaviour? If yes, we should keep some "ifdef", because nanGARBAGE should return false in Linux, and true in MSVC, right?

2. When dealing with 1.#QNAN or 1.#INF in MSVC, the current Kaldi implementation will return true if there is 1.#INF_GARBABE or 1.#INFGARBAGE. This is different from the previous situation (there is no difference when "_" is present or not). For me, it seems a bit strange, but I kept this behaviour in this PR.

3. As it was requested, more tests were added. Please, check if it necessary to add more cases.

danpovey:
Sorry, I have been busy this week and I didn't have time to work on this.
>
> So, I updated the PR. Just some observations:
>
>    1.
>
>    I have one question about "nan". In the current Kaldi source coude,
>    there is a function called is_nan_text that returns true if it is called
>    like is_nan_text("nanGARBAGE", "nan"), but it is false if it is called as
>    is_nan_text("nan_GARBAGE", "nan"). This function is called when MSVC is
>    used. This was tested before? This is the desired behaviour? If yes, we
>    should keep some "ifdef", because nanGARBAGE should return false in Linux,
>    and true in MSVC, right?
>
> Let's not allow nanGARBAGE.  Don't put too much stress on how that
MSVC-specific code looks and behaves, it wasn't carefully reviewed.  Let's
just try to support the key points of how NaN's are written on Windows.



>
>    1.
>
>    When dealing with 1.#QNAN or 1.#INF in MSVC, the current Kaldi
>    implementation will return true if there is 1.#INF_GARBABE or
>    1.#INFGARBAGE. This is different from the previous situation (there is no
>    difference when "_" is present or not). For me, it seems a bit strange, but
>    I kept this behaviour in this PR.
>
>
I think best not to accept the garbage at the end, unless it's space.


>    1.
>
>    As it was requested, more tests were added. Please, check if it
>    necessary to add more cases.
>
>
Thanks... will check after these changes.

danpovey:
It's looking good to me.
Could you please add one more test that generates a nan and an inf (e.g. division by floating-point zero for nan; exp of a larger number for inf), and makes sure that when read back in they still test as nan (x != x) and as inf (x == x && x - x != 0)?
That will mean that if there is a problem on Windows, the automated tests will catch it.

src/util/text-utils-test.cc
```diff
@@ -173,7 +173,81 @@ void TestConvertStringToReal() {
...
+  std::string s = strs.str();
+
+  KALDI_ASSERT(ConvertStringToReal(s, &d) && d != d);
+}
+
+template<class Real>
+void TestInf() {
+  Real d;
+  std::ostringstream strs;
+  strs << 1.0/0.0;
```
danpovey: I don't think this is a safe way to get infinity.
According to wikipedia:
`Depending on the programming environment and the type of number (e.g. floating point, integer) being divided by zero, it may generate positive or negative infinity by the IEEE 754 floating point standard`.
Try exp(10000).

danpovey: actually it would be good to test both positive and negative infinity, try -exp(10000) too.

danpovey:
Thanks
@jtrmal, do you think we should just merge this, assuming the tests are OK, or wait till you can test on Windows?  I'm OK either way.

jtrmal:
I'm ok with merging it without  testing on windows.
@pstanisl, can you please check after merge if the tests pass? Thanks!
Y.

danpovey:
Just realized we have not merged this yet, I'll mege it.

Issue #1059, opened 2016-09-25
cleanup when training iVector extractor


danpovey:
this issue probably exists in both the speaker-id and speech-reco scripts..
I think more cleanup should be done when training the iVector extractor, to remove intermediate models (1.ie, etc.), and post.*.gz.

david-ryan-snyder:
@danpovey, I think this can be closed now. The ASR ivector part was done in #1514 and the rest was in #1543.

PR #1514, opened 2017-03-24
Getting rid of need for alignments in iVector training in AMI

david-ryan-snyder:
This only partially satisfies Issue #1123. I think it will be easier to commit support for this feature one recipe at a time. This starts with AMI. 

* Adds a script in steps/online/nnet2/get_pca_transform.sh which estimates a PCA transform to be used as an alternative to LDA+MLLT for ivectors

**In egs/ami/s5b**

* Adding local/chain/tuning/run_tdnn_1d.sh which is like local/chain/tuning/run_tdnn_1b.sh but uses PCA features for the ivector instead of LDA

* Updating RESULTS_ihm to show results with the above tuning script (run_tdnn_1d)

* Updating local/nnet3/run_ivector_common.sh to support either PCA or LDA features (LDA is by default)

### DIFF
--- a/egs/ami/s5b/RESULTS_ihm
+++ b/egs/ami/s5b/RESULTS_ihm
@@ -40,7 +40,6 @@
 %WER 24.0 | 13098 94470 | 79.4 12.1 8.5 3.4 24.0 57.1 | -0.153 | exp/ihm/nnet3_cleaned/tdnn_sp/decode_dev/ascore_12/dev_hires.ctm.filt.sys
 %WER 25.5 | 12643 89984 | 77.7 14.2 8.2 3.2 25.5 56.4 | -0.139 | exp/ihm/nnet3_cleaned/tdnn_sp/decode_eval/ascore_11/eval_hires.ctm.filt.sys
 
-
 # local/nnet3/run_tdnn.sh --mic ihm --train-set train --gmm tri3 --nnet3-affix ""
 # nnet3 xent TDNN without data cleaning [cleaning makes very small and
 #  inconsistent difference on this dat]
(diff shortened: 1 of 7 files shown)


egs/ami/s5b/RESULTS_ihm
```diff
@@ -55,17 +61,21 @@
...
-
-# local/chain/run_tdnn.sh --mic ihm --stage 12 &
 # cleanup + chain TDNN model
-# for d in exp/ihm/chain_cleaned/tdnn_sp_bi/decode_*; do grep Sum $d/*sc*/*ys | utils/best_wer.sh; done
+# local/chain/run_tdnn.sh --mic ihm --stage 12 &
+# for d in exp/ihm/chain_cleaned/tdnn_sp_bi/decode_*; do grep sum $d/*sc*/*ys | utils/best_wer.sh; done
 %WER 22.5 | 13098 94490 | 80.6 10.8 8.6 3.1 22.5 55.0 | 0.072 | exp/ihm/chain_cleaned/tdnn_sp_bi/decode_dev/ascore_10/dev_hires.ctm.filt.sys
 %WER 22.5 | 12643 89978 | 80.3 12.5 7.2 2.7 22.5 53.1 | 0.149 | exp/ihm/chain_cleaned/tdnn_sp_bi/decode_eval/ascore_10/eval_hires.ctm.filt.sys
 
+# cleanup + chain TDNN model. Uses PCA instead of LDA for ivector features.
```
danpovey: Could you rerun the baseline?  I suspect something else may have changed since it was last run.
And can you put the output of one of those compare_wer.sh scripts inside the 1d script, showing
how it differs from 1b?
If 1d is the best script so far, you could update the link.

david-ryan-snyder: Yes. Starting it now.

david-ryan-snyder:
@danpovey,

I'm not sure how we should be handling the interaction between the older ivector extractors used in the other nnet3 recipes in AMI, and the one I'm updating (local/chain/run_tdnn.sh). To make my concern concrete:

If the user runs some nnet3 script in local (something that isn't local/chain/run_tdnn.sh) he will get an ivector extractor in exp/ihm/nnet3_cleaned/ that is trained using the default LDA features. However, if the user then runs local/chain/run_tdnn.sh (the one updated in this PR) from --stage 4 or earlier, the old ivector extractor will be obliterated. 

One solution is to change nnet3_affix=_cleaned to nnet3_affix=**1d**_cleaned in local/chain/run_tdnn_1d.sh. This will solve the problem of different versions of the ivector overriding the other (e.g., LDA vs PCA version). However, there are problems with this approach as well. For example, consider what happens when the user wants to train a new nnet3 recipe using the _same_ PCA ivectors. Now he has to set nnet3_affix=1d_cleaned, which I think is confusing and will make the contents of exp/ uglier, or alternatively, he has to train a new ivector extractor for each new nnet3 system, but that's unnecessarily wasteful.

danpovey:
That's a pre-existing issue and it's independent of the change to the
scripts, because even running the same script twice will give you a
different iVector extractor that's incompatible.

The scripts currently use those 'ids' to detect when an incompatible
iVector extractor has been used.  But it doesn't give you any way to make
the system usable, because the old iVector extractor would have been lost
by that point.

How about modifying the script train_ivector_extractor.sh so that instead
of just nuking the old iVector extractor, it creates a subdirectory
backup.XXX (via mktemp) where it puts final.ie and final.ie.id?  Remember
to check both the mac and Linux versions of mktemp to make sure your usage
is compatible.




On Mon, Mar 27, 2017 at 5:08 PM, david-ryan-snyder <notifications@github.com

david-ryan-snyder:
@danpovey,

OK, we can do that if that looks to be the best option. 

But, elsewhere you simply check to see if some model or directory already exists, and if it does, you print out something like "already exists, refusing to overwrite it." Why not do a similar thing for the UBM and i-vector directories or models?

danpovey:
I would be fine with that solution also-- but I think it should be done in
the calling script (run_ivector_common.sh), preferably near the top so that
the user sees it immediately and not after a delay.

But, IMO, better to do the backup anyway.

On Mon, Mar 27, 2017 at 5:26 PM, david-ryan-snyder <notifications@github.com

danpovey:
also, because final.ie is a soft link, doing the backup stuff at the end of
the script might be tricky-- better to do it at the start.

david-ryan-snyder:
OK. Backup it is. 

This issue also applies to and needs to be taken care of in the UBM as well, right?

danpovey:
Yes, it looks like it.


On Mon, Mar 27, 2017 at 5:51 PM, david-ryan-snyder <notifications@github.com

david-ryan-snyder:
@danpovey, you can review this now when you get the chance.

The main feature is supporting PCA instead of LDA for the ivector features. For now, results are only reported in the AMI recipe for the default TDNN system. Once we get this through code review, I can start expanding this to other recipes.

In addition to the PCA vs LDA stuff, there's also some changes to the ivector scripts. These consist of the model backup we talked about and some minor changes so that the --cleanup option works correctly (See https://github.com/kaldi-asr/kaldi/issues/1059).

egs/ami/s5b/local/nnet3/run_ivector_common.sh
```diff
@@ -17,8 +17,8 @@ train_set=train   # you might set this to e.g. train_cleaned.
 gmm=tri3          # This specifies a GMM-dir from the features of the type you're training the system on;
                   # it should contain alignments for 'train_set'.
 
-
 num_threads_ubm=32
+feat_type=lda
```
danpovey: let's call this ivector_transform_type so as not to confuse it with the feat_type used in other scripts.

egs/wsj/s5/steps/online/nnet2/train_diag_ubm.sh
```diff
@@ -89,6 +89,15 @@ for f in $data/feats.scp "$online_cmvn_config" $srcdir/splice_opts $srcdir/final
    [ ! -f "$f" ] && echo "$0: expecting file $f to exist" && exit 1
 done
 
+if [ -d "$dir" ]; then
+  bak_dir=$(mktemp -d ${dir}/backup.XXX);
+  echo "$0: Directory $dir already exists. Backing up models in ${bak_dir}";
```
danpovey: models -> i-vector extractor

egs/ami/s5b/local/nnet3/run_ivector_common.sh
david-ryan-snyder: OK.

egs/wsj/s5/steps/online/nnet2/train_diag_ubm.sh
david-ryan-snyder: This is actually the script for training the diagonal UBM.  So I can just refer to it as such here, and say "i-vector extractor" in the next script.

danpovey:
ok.

david-ryan-snyder:
@danpovey, I addressed your last comments.

danpovey:
Thanks!  Merging.

PR #1516, opened 2017-03-28
modified {left,right}-context determination code to take max of {init,ref}.config

vijayaditya:
…ax of {init,ref}.config

### DIFF
--- a/egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
+++ b/egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
@@ -207,21 +207,30 @@ def write_config_files(config_dir, all_layers):
 def add_back_compatibility_info(config_dir):
     """This will be removed when python script refactoring is done."""
 
-    common_lib.run_kaldi_command("nnet3-init {0}/ref.config "
-                                 "{0}/ref.raw".format(config_dir))
-    out, err = common_lib.run_kaldi_command("nnet3-info {0}/ref.raw | "
-                                            "head -4".format(config_dir))
-    # out looks like this
-    # left-context: 7
-    # right-context: 0
-    # num-parameters: 90543902
-    # modulus: 1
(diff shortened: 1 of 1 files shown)


danpovey:
Looks like it's too late and you've already done it-- but since we spoke
last, I realized that there is a problem with doing it this way.  It causes
a mismatch with the testing condition, where the extra-left-context would
be determined from the model itself.  This could impact results (in these
unusual situations where there is this type of context mismatch).

I thought of various code-level fixes and they were all quite complicated
to implement.

But I've thought of a different script-level fix that will re-use some of
the code that you've already written - if you have time, I know you have a
job now.

Firstly, extend the lda-layer or whatever it's called by adding a "delay"
parameter, defaulting to zero.  If it's not zero, then when creating the
init.config, instead of creating the output node directly, it will use that
Descriptor that normally would have gone to the output node, and create an
internal node (just call it delay-node).  Then the output-node would be
something like
output-node name=output input=Offset(delay-node, t, $delay)

If the script detects that the context of the init.config is not contained
within the context of the ref.config, then it will die with an error
message telling the user to add a "delay" option to the LDA layer.  The
script can actually work out the delay value that's closest to zero while
also being sufficient to prevent the error from happening again, and can
tell it to the user.

danpovey:
should have mentioned- the internal node would just be a NoOpComponent.

vijayaditya:
Seems simple enough. I will do this tomorrow.

--Vijay

vijayaditya:
Yep realized that.

danpovey:
ping...

vijayaditya:
@danpovey got busy with paper writing will do it by EoD tomorrow.

danpovey:
ping..

vijayaditya:
working on it .. should be done soon.

vijayaditya:
See #1540 .

PR #1517, opened 2017-03-28
Update run.sh

songmeixu:
use $nj, but `nj` never be assigned.

### DIFF
--- a/egs/wsj/s5/run.sh
+++ b/egs/wsj/s5/run.sh
@@ -123,7 +123,7 @@ if [ $stage -le 2 ]; then
 
     for data in dev93 eval92; do
       nspk=$(wc -l <data/test_${data}/spk2utt)
-      steps/decode.sh --nj $nj --cmd "$decode_cmd" exp/tri1/graph_nosp_tgpr \
+      steps/decode.sh --nj 10 --cmd "$decode_cmd" exp/tri1/graph_nosp_tgpr \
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/run.sh
```diff
@@ -123,7 +123,7 @@ if [ $stage -le 2 ]; then
 
     for data in dev93 eval92; do
       nspk=$(wc -l <data/test_${data}/spk2utt)
-      steps/decode.sh --nj $nj --cmd "$decode_cmd" exp/tri1/graph_nosp_tgpr \
+      steps/decode.sh --nj 10 --cmd "$decode_cmd" exp/tri1/graph_nosp_tgpr \
```
danpovey: oh-- that should be $nspk

egs/wsj/s5/run.sh
```diff
@@ -123,7 +123,7 @@ if [ $stage -le 2 ]; then
 
     for data in dev93 eval92; do
       nspk=$(wc -l <data/test_${data}/spk2utt)
-      steps/decode.sh --nj $nj --cmd "$decode_cmd" exp/tri1/graph_nosp_tgpr \
+      steps/decode.sh --nj $nspk --cmd "$decode_cmd" exp/tri1/graph_nosp_tgpr \
```
songmeixu: yeah, it should be $nspk

PR #1521, opened 2017-03-30
Implement batch normalization at the code level

danpovey:
Includes some deeper modifications to the nnet3 framework-- the concept of 'memos' that can be passed between the Propagate and the corresponding Backprop. 

@tomkocse will work on testing this out with some script level changes, and seeing how it compares in terms of performance with using NormalizeLayer.


danpovey:
Closing PR, as I'll be merging together with convolution.

PR #1526, opened 2017-04-02
Modify get_version.sh to deal better with whitespace (avoid space in …

danpovey:
…version); minor fixes

@dogancan, this is because I was getting whitespace in the version string on mac:

```
WARNING ([5.1.      44~       1-cfd87]:Check():convolution.cc:197) the input at the 3'th height is never used.
```


dogancan:
It seems the BSD version of the `wc` utility prints extra space before printing the counts. I did not notice this on my end since I am using GNU coreutils on Mac. My bad.

PR #1531, opened 2017-04-04
tools: remove openfst check

vesis84:
* we will use the local open-fst, while we still can have the system-wide open-fst (not used by kaldi)
* (note: this setup was previously forbidden)

### DIFF
--- a/tools/extras/check_dependencies.sh
+++ b/tools/extras/check_dependencies.sh
@@ -181,14 +181,6 @@ if which grep >&/dev/null && pwd | grep -E 'JOB|LMWT' >/dev/null; then
   status=1;
 fi
 
-if [ -f /usr/lib64/libfst.so.1 ] || [ -f /usr/local/include/fst.h ] || \
-   [ -f /usr/include/fst/fst.h ] || [ -f /usr/local/bin/fstinfo ]; then
-  echo "*** $0: Kaldi cannot be installed (for now) if you have OpenFst"
-  echo "***   installed in system space (version mismatches, etc.)"
-  echo "***   Please try to uninstall it."
-  status=1
-fi
-
 if ! $printed && [ $status -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


vesis84:
I better double-checked that this causes no problems...
In `kaldi.mk' there are the absolute paths to the local OpenFST:

OPENFSTINC = /mnt/matylda5/iveselyk/DEVEL/kaldi-official/tools/openfst/include
OPENFSTLIBS = /mnt/matylda5/iveselyk/DEVEL/kaldi-official/tools/openfst/lib/libfst.so
OPENFSTLDFLAGS = -Wl,-rpath=/mnt/matylda5/iveselyk/DEVEL/kaldi-official/tools/openfst/lib

The headers are okay, as the '-I${OPENFSTINC}' option has priority over the system dirs.
The library is also okay, as the LD_LIBRARY_PATH is stored to the binary by the '-rpath' options.
The related documentation is here:

Headers [4th paragraph]: https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html#Search-Path
Library [section 3.4]: http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html

danpovey:
But I believe the OpenFst headers would be installed to a system location,
and system-level includes always take precedence over local ones.

vesis84:
Hi apparently this is not true:

> There are a number of command-line options you can use to add additional directories to the search path. The most commonly-used option is -Idir, which causes dir to be searched after the current directory (for the quote form of the directive) and ahead of the standard system directories. You can specify multiple -I options on the command line, in which case the directories are searched in left-to-right order.

The order is following:
1. local directory
2. -I dirs [our local openfst]
3. system includes [the system installation of openfst]

danpovey:
hm, OK.  Merging.

vesis84:
Thanks Dan!

Issue #1097, opened 2016-10-04
In SID recipes, provide a script-level option for simple-length-normalization


david-ryan-snyder:
Make this --simple-length-normalization flag into variable (true|false) in the SRE10 and SRE08 scripts. It should be true in SRE08 and false in SRE10. This emphasizes that it's a configurable variable rather than an essential part of the pipeline.

Assignee: @david-ryan-snyder

david-ryan-snyder:
Resolved as of #1543.

Issue #1096, opened 2016-10-04
SID and LID recipes should use --num-threads N


david-ryan-snyder:
Several SID and LID scripts use  the"-pe smp N" format to specify the number of threads. This is not recommended now. Replace that with  "--num-threads N" instead.

Assigned to @david-ryan-snyder

david-ryan-snyder:
Resolved as of #1543.

Issue #165, opened 2015-09-21
SRE10 v2 Improvements


david-ryan-snyder:
This issue (for david-ryan-snyder) tracks some improvements to the DNN-based SRE10 example (egs/sre10/v2)
- Remove num-components option from init_full_ubm_from_dnn.sh. It should be worked out from nnet-am-info.
- Change full-gmm-init-from-stats.cc so that it uses MleFullGmmUpdate() do the estimation. This has things like variance flooring, and limits on the condition number, to make sure that there are never problems inverting.
- Provide GPU support for the DNN posterior computation in init_full_ubm_from_dnn.sh, train_ivector_extractor_dnn.sh and extract_ivectors_dnn.sh. The binary nnet-am-compute already supports GPU usage, but the bash scripts have not been updated.

david-ryan-snyder:
First two bullet points are addressed with #244. Still need to handle the GPU support.

david-ryan-snyder:
Resolved in #1543.

PR #1538, opened 2017-04-10
WIP: Speaker ID script improvements 

david-ryan-snyder:
This WIP PR is for various improvements to the SID recipes. This is primarily motivated by issues tracked on Git or in the Kaldi forums.

* The i-vector scripts that use nnet2 in sid/*_dnn.sh now support using GPU explicitly. This is in response to complaints of slowness in these scripts. This will close Issue #165. 

* A fix to src/fgmmbin/fgmm-global-init-from-accs.cc so that when a Gaussian has very low occupancy, we don't just crash (See https://groups.google.com/d/msg/kaldi-help/U_L_6IWBN1c/L8oPTcE5AgAJ)

* LID and SID scripts now do more cleanup with the --cleanup=true option. This closes Issue #1059.

* SID i-vector training scripts now use '--num-threads N' instead of '-pe smp N.' This closes Issue #1096. (The LID scripts were already doing the right thing). 

* In sre10/v1/local/plda_scoring.sh added an option for --simple-length-norm (which defaults to 'false' since it gives better performance in SRE10). This closes issue #1097. 

* In egs/sre10/{v1,v2}/run.sh, the PLDA scores are now written to exp instead of local. This is better, since v1 and v2 share the same local directory and would override each other otherwise. Also changing the old-style memory options to the new ones (E.g., --mem 5G))

* The scripts to train the DNN for SRE10 have been moved from sre10/v1/local to sre08/v1/sid/nnet2 (this mirrors what we did with lre07/v1/lid/nnet2). This is consist with other setups, and makes them easier to access by new (or user created) SID recipes. 

* In sre10/v1/local/dnn/run_nnet2_multisplice.sh we now use 8 GPUs to the train the DNN, instead of 18 (which is excessive, and might've been a typo).

* Various cosmetic fixes: fixed indentation in several sid and lid scripts. Fixed a typo in src/gmm/full-gmm.cc . Changed wording in egs/sre10/v1/local/dnn/train_dnn.sh so that it, which is an nnet2 pnorm recipe, is no longer referred to as the "current best recipe" but rather as an "older nnet2 recipe," which is now the correct thing to say.


david-ryan-snyder:
Something went wrong during the rebasing. Going to try this again in another PR.

PR #1542, opened 2017-04-12
This pull request combines batch-norm and convolutional component.

danpovey:


danpovey:
For anyone on the CLSP grid, this is running at
`/home/dpovey/kaldi-trunk/egs/wsj/s5/exp/chain/cnn_tdnn1a_sp`
Now that I start running it, I am reminded of why we originally avoided CNNs-- it's quite slow, about 5 times slower than the TDNN-only baseline.  This is explainable by having more layers running at the input frame rate (of 100 frames per second).  But the objective values are better than the TDNN ones at the same iterations.  This is a rather shallow CNN with only 5 convolutional layers.

danpovey:
@GaofengCheng, you could use this as the starting point for some experimentation (it's not fully debugged yet, of course; I haven't seen the WERs).
@galv, if you have some time at some point I would be interested to see some kind of speed-test versus your CuDNN based component, and against the older component in nnet3.  I may do the latter myself eventually.  I *believe* what I have done is pretty much as optimized as it can be for setups with small patch sizes, but you never know.

GaofengCheng:
@danpovey will look into this in parallel with gated-linear

danpovey:
I have added an example script with results.
The WERs and final objf values are worse than the TDNN baseline, but this system is pretty small and shallow.

danpovey:
closing, will re-open from upstream/kaldi_52

PR #1545, opened 2017-04-14
AMI TDNN Results Update

david-ryan-snyder:
This is a continuation of Issue #1123 (which can't be closed yet). The main point of this is to update more of the results as we continue to switch to using PCA (instead of LDA) for the ivector transform. 

* Update to the TDNN results in RESULTS_{ihm,sdm,mdm}. [All of the results improve slightly.]

* Change to egs/ami/s5b/local/chain/tuning/run_tdnn_1{a,b,c,d}.sh so that the file $dir/egs/.nodelete is not created by default. User should now run with --cleanup false to prevent the egs from being deleted after training. 

* Bug fix in  egs/ami/s5b/run.sh. Previously, if BeamformIt was not installed, the user was instructed to go to ../../../tools/ and run make beamformit. However, this rule doesn't exist in the Makefile. I replaced the instructions with ../../../tools/; extras/install_beamformit.sh; cd -. [Alternatively, we could add just add a rule to the Makefile to do that, but I didn't see many rules involving  extras/ in the Makefile.]


egs/ami/s5b/local/chain/tuning/run_tdnn_1b.sh
```diff
@@ -212,9 +212,9 @@ if [ $stage -le 16 ]; then
      /export/b0{5,6,7,8}/$USER/kaldi-data/egs/ami-$(date +'%m_%d_%H_%M')/s5b/$dir/egs/storage $dir/egs/storage
   fi
 
- touch $dir/egs/.nodelete # keep egs around when that run dies.
+  mkdir -p $dir/egs && touch $dir/egs/.nodelete # keep egs around when that run dies.
```
danpovey: Actually, would you mind just deleting these lines?  The 'proper' way now to stop the egs being removed is to run with --cleanup false.  Anyway it isn't necessarily a good default to keep the egs around.

danpovey:
@david-ryan-snyder (and this isn't specifically about this commit but about your changes to how the iVector extraction is done in the example scripts),
What kind of WER differences, if any, are you seeing by changing LDA->PCA for the i-vector space?
What I am wondering is whether they are substantial enough that we should ask people like @GaofengCheng and @tomkocse who may be adding new results in the near future, to run the script from the start so that iVectors get re-computed.

david-ryan-snyder:
@danpovey,

I don't think they need to. The difference is usually very small (or no difference) and the recipes they are working on are probably still using LDA by default anyway. 

At the moment, the _only_ recipe that uses PCA instead of LDA features by default is egs/ami/s5b/local/chain/run_tdnn.sh . All of recipes in AMI, and all other examples continue to use the LDA transform as before (at least until we update them, which we are in the process of). 

An apples-to-apples comparison in AMI ihm gives a WER of 22.0/22.2 for PCA and 21.9/22.3 for LDA (dev/eval). So they're basically the same. Looking at some of my older results for this (which did not get committed yet), it was slightly worse for Tedilium (a difference of about 0.08% absolute), and slightly better on WSJ.

david-ryan-snyder:
@danpovey, also I removed the "$dir/egs/.nodelete" line.

david-ryan-snyder:
Thanks!

danpovey:
OK, merged.
My feeling is it's probably best to be consistent and do this everywhere-- the results changes are probably just noise.

david-ryan-snyder:
@danpovey, 

I can make that this the default wherever possible. But, I think it will still require targeted updates for each recipe, since different recipes use different ivector extractor training scripts. So, there's an aim for consistency, but I think in the short term there will have to settle for some discrepancies...

The problem with making all of the changes simultaneously in one PR is that by the time I test all the recipes, there will be updates to at least a few of them that make the new changes out of date...

danpovey:
Not urgent at all, I just mean eventually.


On Fri, Apr 14, 2017 at 1:06 PM, david-ryan-snyder <notifications@github.com

david-ryan-snyder:
Oh, got it. Yeah, that's the plan.

PR #1548, opened 2017-04-15
generic read fst, the version supports constfst and vectorfst

LvHang:
Hi Dan,
I open the PR to track the solution about generic read FST.
Now, I modify the fstext/kaldi-fst-io.h/cc so that it can read constfst or vectorfst and return a Fst<StdArc> pointer. I change the utils/mkgraph.sh and gmmbin/gmm-latgen-faster.cc and test it. I confirm it can get the same result as original code.
Then, I will modify other *-lattgen-faster and test them.
Hang

### DIFF
--- a/egs/wsj/s5/utils/mkgraph.sh
+++ b/egs/wsj/s5/utils/mkgraph.sh
@@ -135,7 +135,7 @@ fi
 trap "rm -f $dir/HCLG.fst.$$" EXIT HUP INT PIPE TERM
 if [[ ! -s $dir/HCLG.fst || $dir/HCLG.fst -ot $dir/HCLGa.fst ]]; then
   add-self-loops --self-loop-scale=$loopscale --reorder=true \
-    $model < $dir/HCLGa.fst > $dir/HCLG.fst.$$ || exit 1;
+    $model < $dir/HCLGa.fst | fstconvert --fst_type=const > $dir/HCLG.fst.$$ || exit 1;
   mv $dir/HCLG.fst.$$ $dir/HCLG.fst
   if [ $tscale == 1.0 -a $loopscale == 1.0 ]; then
     # No point doing this test if transition-scale not 1, as it is bound to fail.
@@ -168,6 +168,4 @@ cp $lang/phones/disambig.{txt,int} $dir/phones/ 2> /dev/null
 cp $lang/phones/silence.csl $dir/phones/ || exit 1;
 cp $lang/phones.txt $dir/ 2> /dev/null # ignore the error if it's not there.
 
-# to make const fst:
-# fstconvert --fst_type=const $dir/HCLG.fst $dir/HCLG_c.fst
 am-info --print-args=false $model | grep pdfs | awk '{print $NF}' > $dir/num_pdfs
--- a/src/fstext/kaldi-fst-io.cc
+++ b/src/fstext/kaldi-fst-io.cc
@@ -42,6 +42,71 @@ VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename) {
   return fst;
 }
 
+Fst<StdArc> *ReadFstKaldiGeneric(std::string rxfilename, bool throw_on_err) {
+  if (rxfilename == "") rxfilename = "-"; // interpret "" as stdin,
+  // for compatibility with OpenFst conventions.
+  kaldi::Input ki(rxfilename);
+  fst::FstHeader hdr;
+  // Read FstHeader which contains the type of FST
+  if (!hdr.Read(ki.Stream(), rxfilename)) {
+    if(throw_on_err) {
+      KALDI_ERR << "Reading FST: error reading FST header from "
+                << kaldi::PrintableRxfilename(rxfilename);
+    } else {
+      KALDI_WARN << "We fail to read FST header from "
+                 << kaldi::PrintableRxfilename(rxfilename) 
+                 << ". A NULL pointer is returned.";
+      return NULL;
+    }
+  }
+  // Check the type of Arc
+  if (hdr.ArcType() != fst::StdArc::Type()) {
+    if(throw_on_err) {
+      KALDI_ERR << "FST with arc type " << hdr.ArcType() << " is not supported.";
+    } else {
+      KALDI_WARN << "Fst with arc type" << hdr.ArcType()
+                 << " is not supported. A NULL pointer is returned.";
+      return NULL;
+    }
+  }
+  // Read the FST
+  FstReadOptions ropts("<unspecified>", &hdr);
+  Fst<StdArc> *fst = NULL;
+  if (hdr.FstType() == "const") {
+    fst = ConstFst<StdArc>::Read(ki.Stream(), ropts);
+  } else if (hdr.FstType() == "vector") {
+    fst = VectorFst<StdArc>::Read(ki.Stream(), ropts);
+  }
+  if (!fst) {
+    if(throw_on_err) {
+     KALDI_ERR << "Could not read fst from "
+               << kaldi::PrintableRxfilename(rxfilename);
+    } else {
+      KALDI_WARN << "Could not read fst from "
+                 << kaldi::PrintableRxfilename(rxfilename)
+                 << ". A NULL pointer is returned.";
+      return NULL;
+    }
+  }
+  return fst;
+}
+
+VectorFst<StdArc> *CastToVectorFst(Fst<StdArc> *fst) {
+  // This version currently supports ConstFst<StdArc> or VectorFst<StdArc>         
+  std::string real_type = fst->Type();
+  KALDI_ASSERT(real_type == "vector" || real_type == "const");
+  if (real_type == "vector") {
+    return dynamic_cast<VectorFst<StdArc> *>(fst);
+  } else {
+    // As the 'fst' can't cast to VectorFst, I'm creating a new 
+    // VectorFst<StdArc> initialized by 'fst', and deletes 'fst'.
+    VectorFst<StdArc> *new_fst = new VectorFst<StdArc>(*fst);
+    KALDI_WARN << "The 'fst' is deleted.";
+    delete fst;
+    return new_fst;
+  }
+}
+
 void ReadFstKaldi(std::string rxfilename, fst::StdVectorFst *ofst) {
   fst::StdVectorFst *fst = ReadFstKaldi(rxfilename);
   *ofst = *fst;
--- a/src/fstext/kaldi-fst-io.h
+++ b/src/fstext/kaldi-fst-io.h
@@ -41,6 +41,22 @@ namespace fst {
 // doesn't support the text-mode option that we generally like to support.
 VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename);
 
+// Read a binary FST using Kaldi I/O mechanisms (pipes, etc.)
+// If it can't read the FST, if throw_on_err == true it throws using KALDI_ERR;
+// otherwise it prints a warning and returns. Note:this
+// doesn't support the text-mode option that we generally like to support.
+// This version currently supports ConstFst<StdArc> or VectorFst<StdArc>
+// (const-fst can give better performance for decoding).
+Fst<StdArc> *ReadFstKaldiGeneric(std::string rxfilename,
+                                 bool throw_on_err = true);
+
+// This function attempts to dynamic_cast the pointer 'fst' (which will likely
+// have been returned by ReadFstGeneric()), to the more derived 
+// type VectorFst<StdArc>. If this succeeds, it returns the same pointer;
+// if it fails, it converts the FST type (by creating a new VectorFst<stdArc>
+// initialized by 'fst'), prints a warning, and deletes 'fst'.
+VectorFst<StdArc> *CastToVectorFst(Fst<StdArc> *fst);
+
 // Version of ReadFstKaldi() that writes to a pointer.  Assumes
 // the FST is binary with no binary marker.  Crashes on error.
 void ReadFstKaldi(std::string rxfilename, VectorFst<StdArc> *ofst);
@@ -54,17 +70,16 @@ void WriteFstKaldi(const VectorFst<StdArc> &fst,
 // This is a more general Kaldi-type-IO mechanism of writing FSTs to
 // streams, supporting binary or text-mode writing.  (note: we just
 // write the integers, symbol tables are not supported).
-// On error, throws using KALDI_ERRR.
+// On error, throws using KALDI_ERR.
 template <class Arc>
 void WriteFstKaldi(std::ostream &os, bool binary,
                    const VectorFst<Arc> &fst);
 
 // A generic Kaldi-type-IO mechanism of reading FSTs from streams,
-// supporting binary or text-mode reading/writing
+// supporting binary or text-mode reading/writing.
 template <class Arc>
 void ReadFstKaldi(std::istream &is, bool binary,
                   VectorFst<Arc> *fst);
-
 
 // This is a Holder class with T = VectorFst<Arc>, that meets the requirements
 // of a Holder class as described in ../util/kaldi-holder.h. This enables us to
--- a/src/bin/latgen-faster-mapped-parallel.cc
+++ b/src/bin/latgen-faster-mapped-parallel.cc
@@ -36,7 +36,7 @@ int main(int argc, char *argv[]) {
     using namespace kaldi;
     typedef kaldi::int32 int32;
     using fst::SymbolTable;
-    using fst::VectorFst;
+    using fst::Fst;
     using fst::StdArc;
 
     const char *usage =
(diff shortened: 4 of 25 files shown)


egs/wsj/s5/utils/mkgraph.sh
```diff
@@ -169,5 +169,7 @@ cp $lang/phones/silence.csl $dir/phones/ || exit 1;
 cp $lang/phones.txt $dir/ 2> /dev/null # ignore the error if it's not there.
 
 # to make const fst:
-# fstconvert --fst_type=const $dir/HCLG.fst $dir/HCLG_c.fst
```
danpovey: I wasn't thinking that you would make a separate 'const' version of HCLG.fst, since they take up quite a lot of space.  Instead, insert fstconvert into the pipe after add-self-loops, and just make HCLG.fst const.
Of course this will cause a crash unless *all* the programs that read decoding graphs are converted to be able to read regular or const FST.
Bear in mind that there may be other scripts that write HCLG.fsts.scp for purposes like segmentation of training data; those should be converted too, to write const FSTs.

LvHang:
Hi Dan,
I update the codes in if branches, and CastToVectorFst function. After testing, it looks good.
I will update the "FstHolder" in today or tomorrow, and test them.
Hang

LvHang:
Hi Dan,
I add the "FstHolder" code, Fst read/write function and update the else branches code. It can be compiled and I'm testing it now.

For ReadFstKaldi(std::istream &is, bool binary, Fst<Arc> *fst) and WriteFstKaldi(std::ostream &os, bool binary, const Fst<Arc> &fst) code, currently, I just complete the "binary" mode. 

For text-format model, maybe we can do like this:
1)ReadFstKaldi(std::istream &is, bool binary,Fst<Arc> *fst):
fst::FstHeader hdr;                                                      
hdr.Read(is, "");                                                        
FstReadOptions ropts("<unspecified>", &hdr);                             
Fst<Arc> *ans = NULL                                                     
if( hdr.FstType() == "vector") {                                         
  // use the "else" branch of ReadFstKaldi(std::istream &is, bool binary,VectorFst<Arc> *fst)                       
} else if( hdr.FstType() == "const") {                                   
  // I need to check the ConstFst content and write.
}
2)WriteFstKaldi(std::ostream &os, bool binary,const Fst<Arc> &t):
if (t.Type() == "vector") {
  // use the "else" branch of  WriteFstKaldi(std::ostream &os, bool binary,const VectorFst<Arc> &t)
} else if (t.Type() == "const") {
  // I need to check the ConstFst content and write.
)
Do you have any comments? Thanks a lot.
Hang
For ReadFstKaldi(std::istream &is, bool binary, Fst<Arc> *fst) and WriteFstKaldi(std::ostream &os, bool binary, const Fst<Arc> &fst) code, currently, I just complete the "binary" mode. 

For text-format model, maybe we can do like this:
1)ReadFstKaldi(std::istream &is, bool binary,Fst<Arc> *fst):
fst::FstHeader hdr;                                                      
hdr.Read(is, "");                                                        
FstReadOptions ropts("<unspecified>", &hdr);                             
Fst<Arc> *ans = NULL                                                     
if( hdr.FstType() == "vector") {                                         
  // use the "else" branch of ReadFstKaldi(std::istream &is, bool binary,VectorFst<Arc> *fst)                       
} else if( hdr.FstType() == "const") {                                   
  // I need to check the ConstFst content and write.
}
2)WriteFstKaldi(std::ostream &os, bool binary,const Fst<Arc> &t):
if (t.Type() == "vector") {
  // use the "else" branch of  WriteFstKaldi(std::ostream &os, bool binary,const VectorFst<Arc> &t)
} else if (t.Type() == "const") {
  // I need to check the ConstFst content and write.
)
Do you have any comments? Thanks a lot.
Hang

danpovey:
In text format there will be no distinction between vector and const-fst, and you should read as vector-fst.  However, I don't have time to help with the details; @hainan-xv, can you please help him?

hainan-xv:
Sure.

LvHang:
@hainan-xv 
Hi Hainan,
Maybe you can review the code. I have been testing. I tested the "if branch" with the "gmm-latgen-faster" function in steps/decode_si.sh and "else branch" with the ""gmm-latgen-faster" function in steps/align_fmllr_lats.sh. It goes well. 
Most of other binaries are similar.  I will test few binaries which look different, such as "gmm-latgen-biglm-faster.cc".
And I think we can do it paralleled. Thanks!
Hang

hainan-xv:
OK. I will give you my feedback by tomorrow.

LvHang:
Hi all,
I think I have finished the test.
I used the steps/decode.sh to test the ReadFstKaldiGeneric() function.
I used the steps/align_fmllr_lats.sh to test the Read() function of FstHolder in SequentialTableReader case.
I modified the bin/compile-train-graph.cc to test the Write() function of FstHolder.
I used the steps/decode_with_map.sh to test the FstHolder in RandomAccessTableReader case.
I used the steps/decode_biglm.sh to test the CastToVectorFst() function.
All of them passed. Now utils/mkgraph.sh will generate constfst and *-latgen-faster can deal with correctly.
Through the tests, I think the FstHolder works well.

@hainan-xv
Thank you for your help about coding yesterday. Maybe you can give me some other feedback about the code, then we can  perfect it. Thanks a lot.

Bests,
Hang

src/fstext/kaldi-fst-io.cc
```diff
@@ -42,6 +42,67 @@ VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename) {
...
+  return fst;
+}
+
+VectorFst<StdArc> *CastToVectorFst(Fst<StdArc> *fst) {
+  std::string real_type = fst->Type();
+  if (real_type == "vector") {
+    return dynamic_cast<VectorFst<StdArc> *>(fst);
+  } else {
+    VectorFst<StdArc> *new_fst = new VectorFst<StdArc>(*fst);
+    std::cout << "Warning: The 'fst' is deleted." << std::endl;
```
hainan-xv: use KALDI_WARN instead (then there is no need for std::endl)

src/fstext/kaldi-fst-io.cc
```diff
@@ -42,6 +42,67 @@ VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename) {
...
+    delete fst;
```
hainan-xv: I am not sure why you would delete fst?

src/fstext/kaldi-fst-io.cc
```diff
@@ -42,6 +42,67 @@ VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename) {
...
+                << ". A NULL pointer is returned." << std::endl;
+      return NULL;
+    }
+  }
+  // Check the type of Arc
+  if (hdr.ArcType() != fst::StdArc::Type()) {
+    if(throw_on_err) {
+      KALDI_ERR << "FST with arc type " << hdr.ArcType() << " is not supported.";
+    } else {
+      std::cout << "Warning: Fst with arc type" << hdr.ArcType()
```
hainan-xv: KALDI_WARN

src/fstext/kaldi-fst-io.cc
```diff
@@ -42,6 +42,67 @@ VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename) {
...
+    }
+  }
+  return fst;
+}
+
+VectorFst<StdArc> *CastToVectorFst(Fst<StdArc> *fst) {
+  std::string real_type = fst->Type();
+  if (real_type == "vector") {
+    return dynamic_cast<VectorFst<StdArc> *>(fst);
+  } else {
```
hainan-xv: Probably a good idea to have a KALDI_ASSERT here like
 KALDI_ASSERT(real_type == "const" && real_type == "some other type that it supports");

so that if it's a type that isn't supported the assertion would fail

src/fstext/kaldi-fst-io.cc
LvHang: In fact, I declare its declaration in .h file.

LvHang:
@hainan-xv 
Hi Hainan,
1. As your comments, I update the codes.
2. I used compile-train-graph and gmm-latgen-faster to test the text-format Read()/Write() in FstHolder.
3. I used the valgrind tool to test the gmm-latgen-faster which contains FstHolder and gmm-latgen-biglm-faster which contains  CastToVectorFst(). It looks good.
Do you have any other comments? Thanks a lot.
Hang

hainan-xv:
Looks good to me now.

src/fstext/kaldi-fst-io.h
```diff
@@ -131,6 +161,86 @@ class VectorFstTplHolder {
...
+
+  // Depending on the fst type, we cope t into this holder.
+  void Copy(const T &t) {
+    Clear();
+    if(t.Type() == "const") {
+      t_ = new ConstFst<Arc>(t);
+    } else if( t.Type() == "vector") {
+      t_ = new VectorFst<Arc>(t);
+    } else {
+      KALDI_WARN << "The version currently only support VectorFst and ConstFst.";
```
danpovey: this should be an error not a warning (anyway the code t_ = &t would later cause a double free error).

src/fstext/kaldi-fst-io.h
```diff
@@ -65,6 +81,20 @@ template <class Arc>
 void ReadFstKaldi(std::istream &is, bool binary,
                   VectorFst<Arc> *fst);
 
+// A generic Kaldi-type-IO mechanism of reading FSTs from streams,
+// The FSTs contains different types, such as ConstFst, VectorFst, etc.
+// Now, only supporting binary reading
+template <class Arc>
+void ReadFstKaldi(std::istream &is, bool binary,
```
danpovey: I'd prefer it if you rename thsi to ReadFstKaldiGeneric.

src/fstext/kaldi-fst-io.h
```diff
@@ -131,6 +161,86 @@ class VectorFstTplHolder {
...
+  // it's really binary, so we need to read in binary mode to be on the safe
+  // side.  Extra linefeeds won't matter, the text-mode reading code ignores
+  // them.
+  static bool IsReadInBinary() { return true; }
+
+  const T &Value() {
+    // code error if !t_.
+    if (!t_) KALDI_ERR << "FstTplHolder::Value() called wrongly.";
+    if(t_->Type() == "const") {
+      return *(dynamic_cast<ConstFst<Arc>*>(t_));
```
danpovey: There is no need to cast here- you are pointlessly casting and then implicitly casting back again.
And there is no need to check the type here.

LvHang:
Hi Dan,
Following your comments, I fixed the bugs which includes--fixing Copy() and Value() of FstHolder, renaming Read/WriteFstKaldi(std::istream &is, bool binary,Fst **fst) to Read/WriteFstKaldi(std::istream &is, bool binary,Fst **fst) and some comments.
Thank you for your suggestions.
Hang

danpovey:
Hang, have you run the mini_librispeech example scripts in this branch?  If
not, you should, to test it out.

LvHang:
Oh, I haven' done it. Let me test it out.
Daniel Povey <notifications@github.com>于2017年5月3日 周三19:14写道：

> Hang, have you run the mini_librispeech example scripts in this branch? If
> not, you should, to test it out.
>
>

LvHang:
Hi Dan,
I have tested it out with mini_librispeech. Compared with the results in "Result" file, the results looks good. 
I think the results in "Result" file maybe are out-of-date, because my chain model result is 12.66% better than 13.35 in "Result" file.
However, there still has a problem. When we run "fstisstochastic", it will report error.
`"ERROR: FstImpl : ReadHeader:Fst not of type vector:<unspecified>"`
I checked the binary of fstisstochastic, it use
`VectorFst<StdArc> *fst = ReadFstKaldi(fst_in_filename)`
But the real fst is a const fst. At the same time, I check other binaries in fstbin directory, they have the same cases.
So I think we need to modify all the files in fstbin with
`fst::CastToVectorFst(fst::ReadFstKaldiGeneric(fst_in_filename));`, right?
Hang

LvHang:
Maybe we just need modify " fstisstochastic" rather than all the files in fstbin directory.
Because only " fstisstochastic" is used after `fstconvert --fst_type=const` in utils/mkgraph.sh
What do you think about it?
Hang

danpovey:
With fstisstochastic, I looked at the code it calls and there is nothing
there that requires it to be VectorFst, the internal code uses Fst, so you
could change the reading code to ReadFstGeneric.

I'm OK to just change the binaries which actually need to read ConstFst.

LvHang:
Hi Dan,
I looked at the code about fstisstochastic. Although IsStochasticFst() function only uses Fst, IsStochasticFstInLog() function needs a real Fst type, such as VectorFst or ConstFst.
The signature of IsStochasticFstInLog() is:
`bool IsStochasticFstInLog(const VectorFst<StdArc> &fst,
                                           float delta,                                          
                                           StdArc::Weight *min_sum,                              
                                           StdArc::Weight *max_sum){
  VectorFst<LogArc> logfst;                                                     
  Cast(fst, &logfst); }`
The key point is that Cast() function needs the impl_ object which declare in class ImplToFst. Class ImplToFst inert from Class Fst. So Class Fst doesn't have impl_ object.

So the easiest way to slove the problem is using ReadFstGeneric() and CastToVectorFst() together.
Or overload the IsStochasticFstInLog() function and do casting inside.
Hang

danpovey:
I think it's best to define that function to work on Fst, but have it
interrogate the type internally, so it will work only if it's a VectorFst
or a ConstFst (otherwise die with an error).

HCLG.fst is very large, so if we cast it inside that program it will tend
to blow up in memory.

Dan

LvHang:
Hi Dan,
I fix the bug about fstisstochastic, please check.
And test the newest code with mini_libraryspeech. All goes well.
Hang

LvHang:
Hi Dan,
There are some information about the speed up and memory consumption for constfst and vectorfst.
For more details, you can look into /export/a11/hlyu/fst_speed_test/ directory.

I use /usr/bin/time -v  and top to monitor memory consumption; and /usr/bin/time -v and kaldi's timer to monitor the time. I will show the value which comes from /usr/bin/time in bracket.
All the tests conduct with eval92 dataset.
(1)About wsj/tri1 (use gmm-latgen-faster) on b16
constfst: Accounting time=591(user time:578.17; system time:3.98); memory=0.2%mem(603116kb)
vectorfst: Accounting time=623(user time:617.93; system time:2.95); memory=0.4%mem(1217000kb)
(2)About wsj/nnet2_online(use nnet-latgen-faster) on b10
constfst: Accounting time=1458(user time:1449.03; system time:4.43); memory=0.3%mem(835632kb)
vectorfst: Accounting time=1526(user time:1513.47; system time:4.96); memory=0.6%mem(1595844kb)
(2)About wsj/nnet3/tdnn1a_sp(use nnet3-latgen-faster-parallel --thread=4) on b10
constfst: Accounting time=277(user time:1107.6; system time:13.14); memory=0.5%mem(995228kb)
vectorfst: Accounting time=283(user time:1124.62; system time:12.09); memory=1.0%mem(1766412kb)

So the memory consumption will save about 50%, and speed up is about 5%.

Most of the commands of run.sh are tested.
local/run_nnet2.sh is testing.
The rest are local/run_rnnlms.sh and local/nnet3/run_lstm.sh. I think they are out-of-date. Maybe I need to fix them slightly.

Bests,
Hang

danpovey:
@LvHang, this seems to have a conflict now- can you please resolve it?

LvHang:
Ok

2017-05-27 21:18 GMT-04:00 Daniel Povey <notifications@github.com>:

danpovey:
I just looked through the code and I saw a problem.
It looks like as far as archives are concerned you are only ever reading, never writing, const fsts.
This means that we are not getting the efficiency and memory efficiency improvements of using const FSTs.
You should definitely change the program 'compile-train-graphs' so that it writes as const-fst.  That is responsible for a lot of disk I/O so changing it to use const FST will be a saving.  You can do this by using the FstHolder and by passing it a ConstFst (you can just initialize the ConstFst from a VectorFst in order to convert the fst type). Obviously you will also have to change all programs with 'align' in their name that read Fsts, to use the generic holder object.

Also, there will continue to be some programs that will need to read  vector FSTs (because they do some operation which is not supported by const FSTs, like arc sorting).  So you will have to modify the VectorFstHolder so that it is capable of reading const FSTs (if it isn't already capable of that...  I don't know if OpenFst does any automatic type conversion of FSTs in the reading code).  This is only a potentially problem in binary mode.
Once you change compile-train-graphs so that it writes ConstFst, you can test whether the existing reading code can read ConstFst by passing one of the graphs.*.gz as input to fstcopy.  Use that to verify that the VectorFstHolder can read a ConstFst.
After that, change the program fstcopy to use the generic holder type FstHolder, there is no reason to force it to use vector FSTs.

PR #1552, opened 2017-04-17
[src] Make it possible to do 'egs' extraction without un-compressing …

danpovey:
…matrices.

Also add new forms of compression, e.g. to be able to reproduce 8-bit image data.


danpovey:
I was hoping that the travis checks would run here but they didn't even after I changed the .yml file... merging into the kaldi_52 branch and will look at the tests there.

PR #1554, opened 2017-04-17
Setting up basic structure for CIFAR directory.

danpovey:
@hhadian, you may want to start your data-preparation script from this branch.

Not sure if I mentioned, I want the filter-dim (==color) to vary the fastest in the columns of the image matrices.


hhadian:
OK. So for a CIFAR image which is `32x32` with 3 image channels red, green, and blue you want the feature matrix to be of size `96x32` where for e.g. the pixel at row `i` and col `j` of the red channel of the image is stored at index `(i*3, j)` of the feat matrix, and this pixel in the blue channel is stored at index `(i*3+1, j)` of the feat matrix, etc. Right?

danpovey:
No I want 32 x 96, where the row-index (0...31) corresponds to the x
position, and the '96' corresponds to the combination of y position and
color (with the color varying fastest).

hhadian:
OK, I am going to write the data-prep script in python since I'll have to read binary image data and write Kaldi format matrices (where I'll use the common python lib in steps/libs). Is that fine?

danpovey:
Yes, that's fine.

PR #1557, opened 2017-04-19
Add batch result

tomkocse:


egs/ami/s5b/local/chain/tuning/run_tdnn_1e.sh
```diff
@@ -0,0 +1,266 @@
+#!/bin/bash
+
+# same as 1b but uses batchnorm components instead of renorm
+
+# Results on 03/27/2017:
+# local/chain/compare_wer_general.sh ihm tdnn1b_sp_bi tdnn1e_sp_bi
+# System                   tdnn1b_sp_bi tdnn1e_sp_bi
+# WER on dev               21.9         21.4
+# WER on eval              22.2         21.5
```
danpovey: Thanks for doing this!  This is a very nice result.

egs/ami/s5b/RESULTS_mdm
```diff
@@ -64,11 +62,18 @@
 
 
 # cleanup + chain TDNN model, alignments from IHM data (IHM alignments help).
-# local/chain/run_tdnn.sh --mic mdm8 --use-ihm-ali true --stage 12 &
```
danpovey: Actually, I think I'll wait to merge, I'd like you to clean up this RESULTS file.
We are not putting the detailed tuning results in this RESULTS file.  Remove the old results.  We don't  want any detailed comparative results in this file, just the results with and without cleanup, and possibly with and without IHM alignments (but you don't have to update all the results necessarily ).  Don't say "Uses batchnorm components instead of renorm" at this level.

danpovey:
Merging this; I'm looking into the reasons for the travis failure, which are unrelated.

Issue #1569, opened 2017-04-22
CNN experiments


danpovey:
@GaofengCheng, I was hoping  you could help with this (this is in the kaldi_52 branch).
In wsj/s5/local/chain/tuning/run_cnn_tdnn_1a.sh I have an example of using CNNs, but it's not working great.
Something that I think is worth trying is as follows:

  - Start from our best setup, say a TDNN+LSTM setup with dropout and batch-norm.
  - Try changing the feature type (first do this without using CNNs, just with the same config as before).  I recommend using 64 mel bins and ceps instead of 40 (so change --num-mel-bins and --num-ceps both to 64).   And make sure it has --use-energy=false, otherwise it's not suitable for CNNs because it's not just a linear transformation of the filterbanks.  It looks like our hires configs already have --use-energy=false
  - Next, try adding just 2 convolutional layers at the start of those networks, with small dimensions so that it's not too slow.  And remove the learning-rate-factor=2.0 on the ivector branch, I realized it's inappropriate, and that branch is training too fast.

```
  # pre-process the ivector with fully connected layers; later we will combine                                                                              
  # it with the output of the CNN layers.                                                                                                                   
  # The learning-rate-factor >1.0 is because these layers will be instantiated                                                                              
  # once per sequence (for t==0), which will tend to make it train too slowly.                                                                              
  relu-renorm-layer name=ivector-1 input=ivector dim=200                                                                    
  relu-renorm-layer name=ivector-2 dim=200
                                                                                                                                                            
  conv-batchnorm-layer name=cnn1 input=idct height-in=64 height-out=64 time-offsets=-1,0,1 height-offsets=-1,0,1 num-filters-out=32                         
  conv-batchnorm-layer name=cnn2 height-in=64 height-out=64 time-offsets=-1,0,1 height-offsets=-1,0,1 num-filters-out=32      
 
  # the first TDNN layer:
   relu-renorm-layer input=Append(-1,0,1,ReplaceIndex(ivector-2, t, 0)) ....

```

It will be necessary to try a version of this experiment without the iVectors (you can re-use the egs by including the input node but just not using it, i.e. replacing `relu-renorm-layer input=Append(-1,0,1,ReplaceIndex(ivector-2, t, 0))...` with `relu-renorm-layer input=Append(-1,0,1)`.  The reason is that with this new way of processing ivectors, we will want to verify that they are still helping.

[...]

PR #1571, opened 2017-04-24
CNN experiments

GaofengCheng:
This PR is for discussion about code and results related to issue #1569
For this stage, experiments are based on AMI IHM.
1. (done) running a baseline TDNN+LSTM based on 40dim feature with-or-without iVector
3. (done) if 2. works, adding the batchnorm and dropout ( this is another work that I'm doing) to network from 2 (CNN+TDNN+LSTM) .


egs/ami/s5b/local/nnet3/run_ivector_common.sh
```diff
@@ -64,18 +65,18 @@ if [ $stage -le 2 ]; then
...
+    utils/copy_data_dir.sh data/$mic/$datadir data/$mic/${datadir}_hires${feature_type}
   done
 
   # do volume-perturbation on the training data prior to extracting hires
   # features; this helps make trained nnets more invariant to test data volume.
-  utils/data/perturb_data_dir_volume.sh data/$mic/${train_set}_sp_hires
+  utils/data/perturb_data_dir_volume.sh data/$mic/${train_set}_sp_hires${feature_type}
 
   for datadir in ${train_set}_sp dev eval; do
     steps/make_mfcc.sh --nj $nj --mfcc-config conf/mfcc_hires.conf \
```
pegahgh: @GaofengCheng  Don't you plan to use fbank instead of MFCC in your CNN experiments?

danpovey:
we project back to filterbanks inside the network with a special transform
(idct-layer).  MFCCs are more compressible.

egs/ami/s5b/local/nnet3/run_ivector_common.sh
GaofengCheng: @pegahgh As Dan said, we use idct. Also I found, at least on my experiment setup for this stage, 64-dim feature does not give improvement compared with 40-dim feature.

pegahgh: @GaofengCheng Did you just try it on AMI? AMI is 16kHz setup with 80 hrs training, it may doesn't have enough examples to distinguish classes using higher order features. Did you try 64-dimMFCC  on 300 hrs SWBD data?

GaofengCheng: @pegahgh I'm doing experiments on SWBD now. Do you think it's necessary to re-train an ivector-extractor and re-extract the online-ivector for different dim feature (or it's OK for different dim feature using the same ivector)

egs/ami/s5b/local/chain/tuning/run_tdnn_lstm_1i.sh
```diff
@@ -184,9 +184,14 @@ if [ $stage -le 15 ]; then
...
+  idct-layer name=idct input=input dim=40 cepstral-lifter=22 affine-transform-file=$dir/configs/idct.mat
+
+  conv-relu-batchnorm-layer name=cnn1 input=idct height-in=40 height-out=40 time-offsets=-1,0,1 height-offsets=-1,0,1 num-filters-out=32 learning-rate-factor=0.333 max-change=0.25
+  conv-relu-batchnorm-layer name=cnn2 input=cnn1 height-in=40 height-out=40 time-offsets=-1,0,1 height-offsets=-1,0,1 num-filters-out=32
+  relu-batchnorm-layer name=tdnn0 input=cnn2 dim=512
 
   # the first splicing is moved before the lda layer, so no splicing here
-  relu-renorm-layer name=tdnn1 dim=1024
+  relu-renorm-layer name=tdnn1 input=Append(tdnn0,lda) dim=1024
   relu-renorm-layer name=tdnn2 input=Append(-1,0,1) dim=1024
```
GaofengCheng: @danpovey  I try feature fusion, but init nnet failed, config line from init.config
```
input-node name=ivector dim=100
input-node name=input dim=40
output-node name=output input=Append(Offset(input, -1), input, Offset(input, 1), ReplaceIndex(ivector, t, 0))
output-node name=output input=input
```

GaofengCheng: setting basic_layer.py line995 to False solve this problem

danpovey:
But changing defaults is dangerous; I'm concerned about what would happen
if you ran the scritpts that used IDCT but not LDA.  Anyway I see no reason
why it should be writing to init.config at all, but it could be that we
need to change xconfig_to_configs.py so that it doesn't create init.config
if it would have no outputs.

danpovey:
@GaofengCheng, should this PR still be up here?  Is there any chance we'll merge this?  I don't know what your currently favored CNN recipe looks like and what the timeline is for merging that.

GaofengCheng:
@danpovey I'll update my CNN scripts under this PR soon (after my regularization exps finish), let it be here for this stage...

danpovey:
Gaofeng, I discovered that the default value of num-minibatches-history in the convolutional components (at 4.0) is too low.  If I set it to 40.0 in the config line for convolutional layers and res-blocks, I see a consistent improvement in the image recognition setups (it's only 0.2% or so, absolute, but fairly consistent). 
I am wondering whether to start adding this to the config lines, or just change the default in the code.  Can you see for ASR setups, whether 40.0 is better than 4.0, or at least not worse?

GaofengCheng:
@danpovey I have checked under AMI IHM and SDM (SWBD is running), num-minibatches-history 40 indeed helped a little, details see my email to you

egs/ami/s5b/local/chain/tuning/run_cnn_tdnn_lstm_1i.sh
```diff
@@ -0,0 +1,333 @@
...
+    --chain.xent-regularize $xent_regularize \
+    --chain.leaky-hmm-coefficient 0.1 \
+    --chain.l2-regularize 0.00005 \
+    --chain.apply-deriv-weights false \
+    --chain.lm-opts="--num-extra-lm-states=2000" \
+    --egs.dir "$common_egs_dir" \
+    --egs.opts "--frames-overlap-per-eg 0" \
+    --egs.chunk-width $chunk_width \
+    --egs.chunk-left-context $chunk_left_context \
+    --egs.chunk-right-context $chunk_right_context \
```
danpovey: I believe you should be setting the chunk-left-context-initial and chunk-right-context-final to zero, and the chunk-size should be a comma-separated list.  I.e. this script is very old.   I don't remember all the changes we made recently, but try to refer to recently added scripts.

GaofengCheng: @danpovey Ok Dan, I think you mean I should transfer this experiments to recently added scripts and then test the online decode

danpovey:
The problem with the online decoding may be a separate issue, I'm just
pointing out that these scripts are outdated.

GaofengCheng:
@danpovey Ok get it

GaofengCheng:
@danpovey I will transfer it to new scripts soon

egs/ami/s5b/local/chain/tuning/run_cnn_tdnn_lstm_1i.sh
```diff
@@ -0,0 +1,332 @@
...
+  for decode_set in dev eval; do
+    (
+      nspk=$(wc -l <data/$mic/${decode_set}_hires/spk2utt)
+      # note: we just give it "data/${data}" as it only uses the wav.scp, the
+      # feature type does not matter.
+      steps/online/nnet3/decode.sh \
+        --nj $nspk --cmd "$decode_cmd" \
+        --acwt 1.0 --post-decode-acwt 10.0 \
+        --extra-left-context-initial 0 \
+        --frames-per-chunk "$frames_per_chunk" \
```
danpovey: Please don't supply frames-per-chunk here.
And because you don't specify chunk-left-context-initial when training, the value of
--extra-left-context-initial should be the same as the chunk-left-context.

PR #1573, opened 2017-04-24
nnet3: removing add_lda and include_log_softmax options

vimalmanohar:


danpovey:
Thanks a lot... I'm afraid I am working on a bunch of other changes that may conflict with this so, so I think it's better if we wait and have you commit this in a day or two.  I probably won't need the lda-related code as I think already resolved that issue everywhere, but the other part is relevant.

danpovey:
OK I have committed those other changes I was talking about now.  you'll probably have conflicts.

danpovey:
@vimalmanohar, can you please try to resolve the conflicts in this PR?

danpovey:
@vimalmanohar, sorry, I dropped the ball on this, and new conflicts have appeared.  Would you mind resolving them?

egs/wsj/s5/steps/nnet3/train_raw_dnn.py
```diff
@@ -286,7 +288,9 @@ def train(args, run_opts):
     # use during decoding
     common_train_lib.copy_egs_properties_to_exp_dir(egs_dir, args.dir)
 
-    if (args.stage <= -3) and os.path.exists(args.dir+"/configs/init.config"):
```
danpovey: Are there other reasons we use init.config other than to make the LDA-like transform?

egs/wsj/s5/steps/nnet3/train_raw_dnn.py
```diff
@@ -286,7 +288,9 @@ def train(args, run_opts):
...
+    add_lda = common_train_lib.is_lda_added(config_dir)
+
```
danpovey: Are there other reasons we use init.config other than to make the LDA-like transform?

vimalmanohar: Probably not. It seems like to be removed in the transfer learning PR.

danpovey:
In that case can we just keep the current code instead of adding a new
function to check it?

egs/wsj/s5/steps/nnet3/train_raw_dnn.py
vimalmanohar: init.config is still used to create the initial model. The check is needed to know if the LDA needs to be trained.

danpovey:
I think you are mistaken, if we are talking about the current kaldi_52
code; init.config is only used if we are doing the LDA thing.

egs/wsj/s5/steps/nnet3/train_raw_dnn.py
danpovey: did you see my comment:
"I think you are mistaken, if we are talking about the current kaldi_52
code; init.config is only used if we are doing the LDA thing."
can you please update the PR?

vimalmanohar: I do not understand what needs to be done. The current solution is needed to know if LDA needs to be trained. The function is_lda_added can be changed to read init.raw if needed.

danpovey:
Look at the current xconfig_to_configs.py in kaldi_52: it skips creating
the init.config if there are no output nodes in init.config, and the only
layer that adds an output node there (I believe) is the LDA layer.  So I
believe the script only needs to check for the existence of the
init.config.  the add_lda variable is not needed at all.

danpovey:
Closing this as I'm working on a version of this myself, along with certain other code cleanups.

PR #1577, opened 2017-04-25
Updated run_ihm.sh

felixs8696:
Removed $ before file path that caused errors on resuming download.


danpovey:
Thanks, merged.  Bear in mind that we recommend the s5b recipe over the s5 recipe.

PR #1578, opened 2017-04-25
Add 'test mode' to dropout component

freewym:
For issue #1570. Failure on derivative test need to be resolved.

### DIFF
--- a/src/nnet3/nnet-simple-component.cc
+++ b/src/nnet3/nnet-simple-component.cc
@@ -104,6 +104,7 @@ void DropoutComponent::InitFromConfig(ConfigLine *cfl) {
   bool ok = cfl->GetValue("dim", &dim) &&
     cfl->GetValue("dropout-proportion", &dropout_proportion);
   cfl->GetValue("dropout-per-frame", &dropout_per_frame);
+  // It only makes sense to set test-mode in the config for testing purposes.
   cfl->GetValue("test-mode", &test_mode_);
     // for this stage, dropout is hard coded in
     // normal mode if not declared in config
@@ -131,8 +132,8 @@ void* DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
   BaseFloat dropout = dropout_proportion_;
   KALDI_ASSERT(dropout >= 0.0 && dropout <= 1.0);
   if (test_mode_) {
-    out->Set(1.0 - dropout);
-    out->MulElements(in);
+    out->CopyFromMat(in);
+    out->Scale(1.0 - dropout);
     return NULL;
   }
   if (!dropout_per_frame_) {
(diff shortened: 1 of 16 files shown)


src/nnet3/nnet-simple-component.cc
```diff
@@ -100,9 +100,11 @@ void DropoutComponent::InitFromConfig(ConfigLine *cfl) {
   int32 dim = 0;
   BaseFloat dropout_proportion = 0.0;
   bool dropout_per_frame = false;
+  test_mode_ = false;
   bool ok = cfl->GetValue("dim", &dim) &&
     cfl->GetValue("dropout-proportion", &dropout_proportion);
   cfl->GetValue("dropout-per-frame", &dropout_per_frame);
+  cfl->GetValue("test-mode", &test_mode_);
```
danpovey: Can you comment here that it only makes sense to set test-mode in the config for testing purposes?

src/nnet3/nnet-simple-component.cc
```diff
@@ -128,6 +130,11 @@ void* DropoutComponent::Propagate(const ComponentPrecomputedIndexes *indexes,
 
   BaseFloat dropout = dropout_proportion_;
   KALDI_ASSERT(dropout >= 0.0 && dropout <= 1.0);
+  if (test_mode_) {
+    out->Set(1.0 - dropout);
```
danpovey: This would not work correctly for in-place propagation, if we ever change the code to make that possible, so I think it would be nicer to replace this code with:
```
  out->CopyFromMat(in);
  out->Scale(1.0  - dropout);
  return NULL;
```

freewym:
Done

danpovey:
@hhadian, I haven't really done any testing of this, but I'm hoping you can help with scripting and testing it in the CIFAR setup.
In your Keras-based recipes, where is the dropout applied?  I.e. on which layers?

hhadian:
Sure. Do you want me to prepare another script in local/nnet3 to test it on CIFAR?
The dropout is applied after ReLU activations (or after max-pooling which itself is applied on CNN). I.e. In the recipe I'm using it is either `CNN -> ReLU -> MaxPool -> Dropout` or `FullyConnected -> ReLU -> Dropout`

danpovey:
Yes, make another script like that.
But you'll have to add the capability for dropout to the xconfig layers for
CNN and fully-connected layers, i.e. XconfigConvLayer and
XconfigBasicLayer.  I suggest to have it as an extra nonlinearity in
addition to relu, sigmoid, etc., in XconfigBasicLayer, and an extra
component in addition to 'conv', 'batchnorm' and 'relu', in
XconfigConvLayer.  You'll see in the code how to change it.  In addition
you may need to add extra options for the layer name in parser.py (don't go
overboard, just add what you need).  And you'll need a way to set the
dropout proportion.  Make it default to 0.5 (you'll modify
set_default_configs in both classes to add 'dropout-proportion' : 0.5).

hhadian:
OK. Will do.

PR #1581, opened 2017-04-26
Kaldi 52

YiwenShaoStephen:


YiwenShaoStephen:
@hhadian It can only support color==3 (rgb) matrix by now. But I think it might be enough for cifar sets.

hhadian:
This is working. Before merging, I need to test it with matrices from nnet egs too.

YiwenShaoStephen:
Good luck with it! As we have a 12 hours gap in time, I couldn't response until 9 or 10 hours later. If there is any other problem, leave it to me. I will try to fix it tomorrow.

hhadian:
It is working fine. I added a script to make it easier to see images in egs. 
Augmentation works fine when only flipping but corrupts the image when shift is enabled.

danpovey:
Hossein, make sure you give the --num-channels=3 option to the augmentation
binary.

hhadian:
I am using that option.
I did more checking now, vertical shift is working fine too. Horizontal shift corrupts the image though.

danpovey:
OK, have a look at the code and fix it; I never reviewed it thoroughly.

danpovey:
I see now that Hossein said this was OK to merge.  Merging it.

PR #1585, opened 2017-04-26
Add augmentation option to train_raw_dnn.py

hhadian:
It seems github is acting odd. It includes my previous commits everytime but merges them w/o a problem, so we are fine I guess.

I tried augmentation (only flipping and vertical shift, I'll fix horizontal shift after I'm done with trying dropout)  with the same config as 1a, and I got 79.5% final valid accuracy.


danpovey:
I think it's because I previously committed using 'squash and merge', so those exact commits are not showing up.  If you were to rebase against kaldi_52, the issue would probably go away.  But I can  merge anyway.

egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/common.py
```diff
@@ -27,6 +27,7 @@ def train_new_models(dir, iter, srand, num_jobs,
                      raw_model_string, egs_dir,
                      momentum, max_param_change,
                      shuffle_buffer_size, minibatch_size_str,
+                     image_augmentation_opts,
```
danpovey: You need to have this default to None; otherwise this  function would crash when called from elsewhere.

hhadian: OK. I guess the same goes with train_new_models, right?

danpovey:
I think I commented on the wrong function.  It's train_one_iteration where
it's important-- that's called from elsewhere.  But I think
train_new_models is only called from there; grep around, though, to check.

hhadian:
`train_new_models` is only called `train_one_iteration` but I guess it might be safer to default it to None anyway.

PR #1587, opened 2017-04-27
[src] Change how combination works with batch-norm.  Only for nnet3, …

danpovey:
…needs to be duplicated for chain.

@GaofengCheng, you'll have to do some work to duplicate all these changes for the 'chain' tools.  Check each and every diff line because I think all or nearly all of them need to also be made on similar 'chain'-based code.  And that code might be in nnet3/ or chain/ or nnet3bin/ or chainbin/, so grep around for similarly named things.


danpovey:
Also, I have not tested this at all, I am leaving that to you.

GaofengCheng:
@danpovey  get it

danpovey:
I've tested this enough that I'm pretty confident in it, so merging.
Also contains some CIFAR recipe changes I tested with it.

PR #1589, opened 2017-04-28
Add dropout to Xconfig

hhadian:


hhadian:
It gives 78.5% final accuracy on valid set (w/o data augmentation) which is 4% more than similar Keras setup.
I have not used any normalization either.

danpovey:
Can you rebase this against kaldi_52?  Github is showing some changes that I think I already merged.

hhadian:
Rebased it

danpovey:
I merged this, but I was surprised that you didn't combine dropout with batchnorm, e.g. conv-relu-batchnorm-dropout-layer.

hhadian:
I will do that. First I wanted to try the similar setup as Keras (which does not use batchnorm).

danpovey:
I'm trying myself adding batchnorm to all 4 of your 1c layers.

hhadian:
OK.

hhadian:
BTW, I also tried 1c setup w/o dropout, and the final accuracy was 68% and there was significantly more overfitting.

danpovey:
Adding batch-norm to 1c seems to help, I am nearly finished training (as
1d), and test accuracy is 0.82 (vs 0.78 in 1c)

hhadian:
That's interesting. It is also interesting that 1c was very fast to train too (even though it has similar setup to Keras with a rather big fully connected layer 8192x512). I wonder why Keras takes so long to train.

PR #1596, opened 2017-04-29
Add image augmentation

hhadian:
The transform matrix preparation might not be very efficient. For the start, I aimed for good code structure.

### DIFF
--- a/src/nnet3bin/nnet3-egs-augment-image.cc
+++ b/src/nnet3bin/nnet3-egs-augment-image.cc
@@ -1,7 +1,7 @@
 // nnet3bin/nnet3-egs-augment-image.cc
 
 // Copyright      2017  Johns Hopkins University (author:  Daniel Povey)
-//                2017  Yiwen Shao
+//                2017  Hossein Hadian
 
 // See ../../COPYING for clarification regarding multiple authors
 //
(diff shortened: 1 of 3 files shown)


src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -1,7 +1,7 @@
 // nnet3bin/nnet3-egs-augment-image.cc
 
 // Copyright      2017  Johns Hopkins University (author:  Daniel Povey)
-//                2017  Yiwen Shao
```
danpovey: keep Yiwen's name, I think a little of that code is still there, anyway he will likely add more.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -63,77 +63,62 @@ struct ImageAugmentationConfig {
...
+      //                              c1  <=  c_old  <=  c2
+
+      int32 r1 = static_cast<int32>(floor(r_old));
+      int32 c1 = static_cast<int32>(floor(c_old));
+      int32 r2 = r1 + 1;
+      int32 c2 = c1 + 1;
+
+      // TODO(hhadian): it might be better to use the following 'if' only when
+      // none of the 4 points are available. For now, we only check one of them.
+      if (r1 < 0 || c1 < 0 || r1 > num_rows - 1 ||
```
danpovey: I don't think you need a separate branch at this level.
Instead, why not do the following (edited for style):
```
if (r1 < 0) { r1 = 0; if (r2 < 0) r2 = ;  }
if (r2 <= num_rows) { r2 = num_rows - 1; if (r1 <= num_rows) r1 = num_rows - 1; }
```
and similar code for c1,c2.  Then you can treat the edge case the same as the regular case.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -147,37 +132,91 @@ void PerturbImage(const ImageAugmentationConfig &config,
...
+  BaseFloat horizontal_shift = (2.0 * RandUniform() - 1.0) *
+                               config.horizontal_shift * image_width;
+  BaseFloat vertical_shift = (2.0 * RandUniform() - 1.0) *
+                             config.vertical_shift * image_height;
+  shift_mat(0, 2) = floor(horizontal_shift);  // best to floor shifts to avoid blurring
+  shift_mat(1, 2) = floor(vertical_shift);
+  // since we will center the image before applying the transform,
+  // horizontal flipping is simply achieved by setting [0, 0] to -1:
+  if (WithProb(config.horizontal_flip_prob))
+    shift_mat(0, 0) = -1;
```
danpovey: We might need to test this.  In the Keras code they don't do it this way, and I wonder whether there might be some subtle reason for thi.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -147,37 +132,91 @@ void PerturbImage(const ImageAugmentationConfig &config,
...
+  Matrix<BaseFloat> reset_origin_mat(3, 3, kUndefined);
+  reset_origin_mat.SetUnit();
+  reset_origin_mat(0, 2) = -image_width / 2.0 + 0.5;
+  reset_origin_mat(1, 2) = -image_height / 2.0 + 0.5;
+
+  // transform_mat = set_origin_mat * transform_mat * reset_origin_mat
+  transform_mat.AddMatMatMat(1.0, set_origin_mat, kNoTrans,
+                                  transform_mat, kNoTrans,
+                                  reset_origin_mat, kNoTrans, 0.0);
+  ApplyAffineTransform(transform_mat, config.num_channels, image);
```
danpovey: BTW, let's not worry about efficiency here; these matrices are very tiny.

danpovey:
It would be great if you could test this by manually inspecting the images in the same way that you did for the old code.

src/nnet3bin/nnet3-egs-augment-image.cc
hhadian: There might be some reason, or maybe not because I have tested this (alone and with other operations). I have also tested rotation. All works fine.

hhadian:
> It would be great if you could test this by manually inspecting the images in the same way that you did for the old code

This code is tested.

danpovey:
Great.  How about you add a script that uses this and make it part of the
same PR?

src/nnet3bin/nnet3-egs-augment-image.cc
hhadian: But this way we will be using the edges (of the old image) for all unavailable pixels. In the current code, I am using the nearest (in row axis) point (of the new image). I guess it makes a difference in rotation for e.g.

danpovey: Your current code seems to me rather strange, I can't see what the intention is.  It will cause a setrange 'double-vision' effect, with parts of the image repeated.
I think it's best if you change it to the way I said.

danpovey: BTW, I think if we want to preserve textures and so on, the best option is to reflect, like the 'reflect' mode of keras.  My intention here was to emulate the 'nearest' mode of Keras; later if it's needed we  can add an option and give alternatives like 'reflect' and 'wrap'.

hhadian: OK I'll change it.

hhadian:
I'm currently running the augmentation script to see if there are any problems.

danpovey:
Looks good, but could you please put a little more information in the new scripts?
Say what the accuracy without augmentation was in the comment, and also add the output of steps/info/nnet_dir_info.pl, like I did for the existing scripts.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -63,77 +64,73 @@ struct ImageAugmentationConfig {
...
+
+      // we are going to do bilinear interpolation between 4 closest points
+      // to the point [r_old, c_old]. We have: r1  <=  r_old  <=  r2
+      //                                       c1  <=  c_old  <=  c2
+
+      int32 r1 = static_cast<int32>(floor(r_old));
+      int32 c1 = static_cast<int32>(floor(c_old));
+      int32 r2 = r1 + 1;
+      int32 c2 = c1 + 1;
+
```
danpovey: I think it would be cleaner if you obtained the interpolation factors at this stage, before modifying r1, r2, c1 and c2.
E.g.
`BaseFloat weight_11 = ..., weight_12 = ...`
Then you don't have to modify r_old and  c_old, which is confusing.
Changing it like this would also be more efficient.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -63,77 +64,73 @@ struct ImageAugmentationConfig {
...
+      if (r1 < 0) {
+        r1 = 0;
+        r_old = 0.0;  // hold the above conditions
+        if (r2 < 0)
```
danpovey: This code is very long vertically.  I don't think the google style guide has anything against:
`if (x) y;`

hhadian:
I did. BTW, it is interesting how much improvement we get due to augmentation with vs. w/o dropout. With dropout, the improvement is very little (1.5%). I tried more epochs with a smaller lr too but no difference.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -63,77 +64,73 @@ struct ImageAugmentationConfig {
...
-        origin_row(*image, origin_r);
-    current_row.CopyFromVec(origin_row);
-  }
-}
-
-/* Shifts the image vertically by 'vertical_shift' (+ve == to the top).*/
-void VerticalShift(int32 vertical_shift,
-                   int32 num_channels,
-                   MatrixBase<BaseFloat> *image) {
+void ApplyAffineTransform(MatrixBase<BaseFloat> &transform,
```
danpovey: I think it would be nice to have some kind of pointer or link to the Keras code that we were referring to.

hhadian: For applying the transform? Or using this method of geometric transformation altogether?
I have not used any code as a reference for writing this code (applying the transform).

src/nnet3bin/nnet3-egs-augment-image.cc
hhadian: Will do it

src/nnet3bin/nnet3-egs-augment-image.cc
danpovey: OK, not here, then; but in the place where the transform is computed.
And see if you can reduce the amount of vertical space a little bit throughout this code.

hhadian: OK. Jus write something like `reference: https://github.com/fchollet/keras/blob/master/keras/preprocessing/image.py#L524`?
BTW, this method of applying an affine transform for geometric transformation is very well known in image processing (and is explained exactly like this code in text books such as Gonzalez book which I used as a reference).

danpovey:
OK, maybe refer to gonzalez book and also to keras-- but see if there is a
more 'upstream' version of the keras code, I assume fchollet is not the
official maintainer?

hhadian:
OK. I think it is the official one. https://keras.io points to it too.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -64,6 +64,16 @@ struct ImageAugmentationConfig {
   }
 };
 
+/**
+  This function applies a geometric transformation 'transform' to the image.
+  References: [1] Digital Image Processing book by Gonzalez and Woods.
```
danpovey: was there supposed to be a [2] here?

danpovey: oh I see there wasn't.  the [1] confused me.

hhadian: I just numbered the references. Delete the numbers?

danpovey:
I'd delete them, it's confusing in this context.

hhadian:
Deleted.

PR #1597, opened 2017-04-30
add orthogonal matrix initialization for nnet3

keli78:
add orthogonal matrix initialization for nnet3


pegahgh:
@keli78 I wonder how much does it help in faster convergence for convolution layer?
Did you check the relative parameter difference for convolution layer?

keli78:
For the convolution layers, the relative parameter differences are almost the same as before. 
It makes the training more stable and improves the accuracy on cifar10 a little bit.

PR #1603, opened 2017-05-03
Add another augmentation experiment script + final-combination flag

hhadian:


egs/wsj/s5/steps/nnet3/train_raw_dnn.py
```diff
@@ -360,7 +365,7 @@ def learning_rate(iter, current_num_jobs, num_archives_processed):
 
         num_archives_processed = num_archives_processed + current_num_jobs
 
-    if args.stage <= num_iters:
+    if args.stage <= num_iters and args.final_combination:
```
danpovey: we'l want to have it force-symlink (function force_symlink in libs/common.py) the last model to final.mdl, if final_combination was false, so this would work if the next stage was ASR decoding.  [this would require an if-statement inside the existing if-statement].

hhadian: I fixed it

hhadian:
SVHN is ready to push. If I push it now, it will pop up in this PR. Do you want me to push it? Or wait to merge this?

egs/wsj/s5/steps/nnet3/train_raw_dnn.py
```diff
@@ -361,13 +366,17 @@ def learning_rate(iter, current_num_jobs, num_archives_processed):
...
+        if args.final_combination:
+            logger.info("Doing final combination to produce final.raw")
+            train_lib.common.combine_models(
+                dir=args.dir, num_iters=num_iters,
+                models_to_combine=models_to_combine, egs_dir=egs_dir,
+                minibatch_size_str=args.minibatch_size, run_opts=run_opts,
+                get_raw_nnet_from_am=False,
+                sum_to_one_penalty=args.combine_sum_to_one_penalty)
+        else:
+            common_lib.force_symlink("{}.raw".format(num_iters),
```
danpovey: I wasn't aware that {} could be used in place of {0}, but I think it's best  to use the same convention as the other code, i.e. {0}.

hhadian: It can, actually {} is very common AFAIK. I will change it.

PR #1604, opened 2017-05-03
Add SVHN dataset

hhadian:
This is really odd that github lists previous commits.


danpovey:
It's because I did squash and merge and you didn't rebase.  But it's OK, I'll squash this time too so list is irrelevant.

danpovey:
Actually, why don't you rebase.  I notice the old changes are there, that shouldn't be, and it makes me nervous.

hhadian:
I rebased just before my previous PR today (less than 2 hours ago)!
I will rebase everytime before making a PR from now on. I will rebase this too.

hhadian:
Rebased it.

danpovey:
It looks great, but I'm not sure that I want to be the first one to write example scripts for it.  Do you want to add a first guess at some training setup?

Incidentally, when dumping egs, you might want to aim for fewer archives, by using a larger number of images-per-archive (e.g. 50k).  I'm not sure how many archives it produces currently.  The training time per job is a little low right now, with the cifar setup.

hhadian:
Actually (to test things) I adopted 1e (from cifar) and started training. It is currently at epoch 30 with accuracy ~95%. 
The training time per job for SVHN is currently ~22 seconds. I will try to put more images in archives.

danpovey:
FYI: For ASR tasks we generally aim for about 3 to 5 minutes per job.
However we don't want to split the data into fewer archives than num-jobs-final.
And it could be that these models train fast enough that we want to do the averaging more frequently than that.
Also we don't really know what type of models we are going to end up using.
E.g. the wide-resnet configurations I am experimenting with are slower.

But we can always dump the egs in different sizes later if needed.

hhadian:
I am not sure I understood the part about averaging. 
But currently 3 archives are created., which leads to 2N iterations (N is number of epochs).
If we make 1 archive instead of 3 and set num-jobs-initial and final to 1, then we will have N iterations.

If I understood things right, the only way to make it considerably faster (i.e do only N/5 iterations for e.g.) is to change the code to enable more than 1 epoch per iteration (through a flag like epochs_per_iter) by feeding an archive more than once to nnet3-train.
Then we can simply choose a suitable epochs_per_iter for each experiment (i.e. 1a, 1b, res1a, ...)

danpovey:
@hhadian, I can merge this soon, but could you add at least one example script with numbers?

danpovey:
actually, you focus on the CIFAR stuff; I'm running this setup and I can add an example script.

hhadian:
I had 2 scripts from last week and I pushed them (before seeing your last message).

danpovey:
it says I closed this; I didn't mean to.  I'd reopen, but the button is grayed out.
 I'm not seeing your example scripts, either from the github interface or when I check out your repo from the command line.

danpovey:
can you please create a new PR?  It won't let me reopen this one.

hhadian:
Will do now

PR #1605, opened 2017-05-04
Backstitch kaldi 52

freewym:
Continued work from #1511


danpovey:
@freewym, can you please fix conflicts?
I'm considering merging this version soon, since it looks like doing it at the egs level would be more complicated than I thought, due to the interaction with natural gradient.

freewym:
OK, will do it tomorrow.

freewym:
@danpovey Do you prefer to add all the backstitch experiments scripts to this PR, or create a separate PR after merging this one?

danpovey:
Probably a separate PR after merging this one.

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +69,81 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
...
+    ScaleNnet(0.0, delta_nnet_);
 }
 
-
-void NnetChainTrainer::ProcessOutputs(const NnetChainExample &eg,
+void NnetChainTrainer::ProcessOutputs(bool is_backstitch_step,
+                                      const NnetChainExample &eg,
                                       NnetComputer *computer) {
   // normally the eg will have just one output named 'output', but
   // we don't assume this.
```
danpovey: you should probably clarify in this comment that in the diagnostics, 
the output-name with the "_backstitch" suffix is the one
computed after the first, backward step of backstitch.

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +69,81 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
                              &request);
   const NnetComputation *computation = compiler_.Compile(request);
 
-  NnetComputer computer(nnet_config.compute_config, *computation,
+  if (nnet_config.backstitch_training_scale > 0.0 && num_minibatches_processed_
+      % nnet_config.backstitch_training_interval == 0) {
+    // backstitch training is incompatible with momentum > 0
+    KALDI_ASSERT(nnet_config.momentum == 0.0);
+    FreezeNaturalGradient(true, delta_nnet_);
+    bool is_backstitch_step = true;
```
danpovey: The name 'is_backstitch_step' is not very clear as backstitch has two steps.
You should probably say 'is_backstitch_step1'.

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +69,81 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
...
+  } else {
+    // conventional training
+    bool is_backstitch_step = false;
+    TrainInternal(chain_eg, *computation, is_backstitch_step);
+  }
+
+  num_minibatches_processed_++;
+}
+
+void NnetChainTrainer::TrainInternal(const NnetChainExample &eg,
```
danpovey: I think it would make the code easier to understand if you split TrainInternal into two functions:
TrainInternal(), with an interface like the current function but with no is_backstitch_step argument, and
TrainInternalBackstitch(),  taking the is_backstitch_step1 argument.  Currently the code relies on the condition
```
if (nnet_config.backstitch_training_scale > 0.0 && num_minibatches_processed_
      % nnet_config.backstitch_training_interval == 0) {
```
being identical in 2 places, which is a bit ugly.  Obviously simplify in reasonable ways, e.g. momentum
is not allowed in backstitch, so no need to refer to it at all in the backstitch version of the function.

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +69,81 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
...
+  // Scales delta_nnet
+  if (success)
+    ScaleNnet(scale_delta_nnet, delta_nnet_);
+  else
+    ScaleNnet(0.0, delta_nnet_);
 }
 
-
-void NnetChainTrainer::ProcessOutputs(const NnetChainExample &eg,
+void NnetChainTrainer::ProcessOutputs(bool is_backstitch_step,
```
danpovey: I think it would  be better if you give this an argument is_backstitch_step2 and only added the
suffix _backstitch if that were true.  Because the regular "output" is the one we should really be plotting,
in most circumstances.  I'm a little unclear how you are obtaining your plots, because I assume you
need to pick up the one with the _backstitch suffix, given the way the code currently works.

Note: once you split TrainInternal() into two functions, doing this will be easier, because in the calling
code you'll know whether you are doing backstitch or not.

src/nnet3/nnet-chain-training.cc
```diff
@@ -258,14 +235,18 @@ void NnetChainTrainer::PrintMaxChangeStats() const {
         KALDI_LOG << "For " << delta_nnet_->GetComponentName(c)
                   << ", per-component max-change was enforced "
                   << (100.0 * num_max_change_per_component_applied_[i]) /
-                     num_minibatches_processed_ << " \% of the time.";
+                     num_minibatches_processed_ /
+                     (opts_.nnet_config.backstitch_training_scale > 0.0 ? 2.0 : 1.0)
```
danpovey: instead of `a / b / c`, prefer `a / (b * c)` which is easier to read by humans.

src/nnet3/nnet-chain-training.cc
```diff
@@ -258,14 +235,18 @@ void NnetChainTrainer::PrintMaxChangeStats() const {
...
+                  << " \% of the time.";
       i++;
     }
   }
   if (num_max_change_global_applied_ > 0)
     KALDI_LOG << "The global max-change was enforced "
               << (100.0 * num_max_change_global_applied_) /
-                 num_minibatches_processed_ << " \% of the time.";
+                 num_minibatches_processed_ /
```
danpovey: and here also.

src/nnet3/nnet-component-itf.h
```diff
@@ -471,6 +471,9 @@ class UpdatableComponent: public Component {
   /// learning_rate_factor_.
   virtual void SetAsGradient() { learning_rate_ = 1.0; is_gradient_ = true; }
 
+  /// freezes/unfreezes NaturalGradient updates
```
danpovey: add: `, if applicable (to be overriden by components that use Natural Gradient).`.
Also, please make sure that all components that use natural gradient actually do
override this, including ones that were added since you started this work.  
Does TimeHeightConvolutionCompoent override this?  And LstmNonlinearityComponent,
if it uses natural gradient? (I don't recall).
If not, it might impact @GaofengCheng's work.

src/nnet3/nnet-training.cc
```diff
@@ -62,20 +63,78 @@ void NnetTrainer::Train(const NnetExample &eg) {
                         &request);
   const NnetComputation *computation = compiler_.Compile(request);
 
-  NnetComputer computer(config_.compute_config, *computation,
+  if (config_.backstitch_training_scale > 0.0 &&
```
danpovey: My comments about the chain training code all apply to the regular nnet3 training code also.

src/nnet3/nnet-training.h
```diff
@@ -70,6 +74,13 @@ struct NnetTrainerOptions {
                    "so that the 'effective' learning rate is the same as "
                    "before (because momentum would normally increase the "
                    "effective learning rate by 1/(1-momentum))");
+    opts->Register("backstitch-training-scale", &backstitch_training_scale,
+                   "backstitch traning factor. "
```
danpovey: typo traning; and mention that is alpha in our publications.

src/nnet3/nnet-training.h
```diff
@@ -70,6 +74,13 @@ struct NnetTrainerOptions {
...
+                   "if 0 then in the normal training mode.");
+    opts->Register("backstitch-training-interval",
+                   &backstitch_training_interval,
+                   "do backstitch training with the specified interval of "
+                   "minibatches.");
```
danpovey: mention that this is the interval n in publications.

src/nnet3/nnet-utils.h
```diff
@@ -254,6 +266,61 @@ void FindOrphanNodes(const Nnet &nnet, std::vector<int32> *nodes);
 */
 void ReadEditConfig(std::istream &config_file, Nnet *nnet);
 
+/**
+   This function does the operation '*nnet += scale * delta_nnet', while
+   respecting any max-parameter-change (max-param-change) specified in the
+   updatable components, plus the global max-param-change specified as
```
danpovey: plus -> and also

src/nnet3/nnet-utils.h
```diff
@@ -254,6 +266,61 @@ void FindOrphanNodes(const Nnet &nnet, std::vector<int32> *nodes);
...
+   @param [in] max_change_scale  This value, which will normally be 1.0, is used
+               to scale all per-component max-change values and the global
+               'max_param_change', before applying them (so we use
+               'max_change_scale * uc->MaxChange()' as the per-component
+               max-change, and 'max_change_scale * max_param_change' as the
+               global max-change).
+   @param [in] scale  This value, which will normally be 1.0, is a scaling
+               factor used when adding to 'nnet', applied after any max-changes.
+   @param [in,out] nnet  The nnet which we add to.
+   @param [in,out] num_max_change_per_component_applied  Stats for per-component
```
danpovey: Clarify that we add to the elements of this (if that is what this function does). 
Also I would write this as just [out].

src/nnet3/nnet-utils.h
```diff
@@ -254,6 +266,61 @@ void FindOrphanNodes(const Nnet &nnet, std::vector<int32> *nodes);
...
+                   max-change.
+   @param [in,out] num_max_change_global_applied  Stats for global max-change.
```
danpovey: Clarify that we add to this.    And I'd write this as just [out].  I might be wrong in doing that, but other documentation uses this convention.

danpovey:
@freewym, there is something else, and @GaofengCheng, pay attention to this, as it may affect your experiments with dropout and backstitch.
I am concerned that it may be insufficient to rely on the egs-shuffling to ensure that on different epochs, different data gets backstitch applied to it.  The issue is that if there are 'rare-sized' egs, they won't get shuffled enough-- we use a buffer size of 2000, but rare sizes of egs may get spit out in almost the same order as they came in, if we consider egs of only one type.
So instead of a condition like
` num_minibatches_processed_ % interval == 0`
you should use a condition like
` num_minibatches_processed_ % interval == srand_seed_ % interval`.
But that means that you will have to start setting the srand-seed via the --srand option of nnet3-train and nnet3-chain-train (add the option to the scripts if needed; set it to the same value as is provided to nnet3-shuffle-egs).
This may have a small impact on experiments with dropout (I doubt it, but it could happen).

freewym:
srand_seed_ is initialized with a random integer in the constructor. which are almost different across all calling of the training binary. So it might be suffice to solve the issue for 'rare-size' egs by changing the condition to what you proposed?

danpovey:
No that's not sufficient because the output of RandInt() is determined by
rand(), which is deterministic unless you set srand() to a different value
at the start of the program.

freewym:
Oh yes.

src/nnet3/nnet-chain-training.cc
freewym: The plot is based on parsing compute_prob_{train,valid}.*.log, not train.*.*.log. So it would not affect the plot.

GaofengCheng:
@danpovey I can try backstitch+dropout after Yiming fixing as you recommended and see

jtrmal:
guys, does this have a chance to be merged soon (today) or does it still
need some more time?
y.

danpovey:
Probably more like 2-3 days, it depends how busy Yiming is-- I asked for
some small code refactorings and it will require testing just to make sure
the rewritten code doesn't crash.

src/nnet3/nnet-chain-training.cc
```diff
@@ -68,23 +69,94 @@ void NnetChainTrainer::Train(const NnetChainExample &chain_eg) {
                              &request);
   const NnetComputation *computation = compiler_.Compile(request);
 
-  NnetComputer computer(nnet_config.compute_config, *computation,
+  if (nnet_config.backstitch_training_scale > 0.0 && num_minibatches_processed_
+      % nnet_config.backstitch_training_interval ==
+      srand_seed_ % nnet_config.backstitch_training_interval) {
```
freewym: I tested the new refactored code usng chain model (tdnn-lstm system on AMI, with backstitch enabled or disabled respectively, if enabled, test interval=1 or 5 respectively), and xent models (using the CIFAR recipe, with backstitch enabled or disabled, if enabled, only test interval=1), and compared their results with my old runs. Among those results, the only one that cannot match the old result is the chain+tdnn-lstm AMI system with backstitch interval=5 setting, where WER is ~0.8 worse than the old run, while the valid log-prob is also a bit lower. So now I am changing this expression to 
`nnet_config.backstitch_training_interval == 0`
to see if I can reproduce the old result.

danpovey:
It could be this is about which minibatches are chosen for the natural
gradient update, but I'd only expect to see this effect with
backstitch-interval=4, since it's the same as the update-period for the
natural gradient.

PR #1607, opened 2017-05-04
nnet1: minor update the of i-vector / mpe script

vesis84:
- mpe: backward compatibility is provided
- ivec: the ivectors get stored in binary format (saves space)

### DIFF
--- a/egs/wsj/s5/steps/nnet/ivector/extract_ivectors.sh
+++ b/egs/wsj/s5/steps/nnet/ivector/extract_ivectors.sh
@@ -164,13 +164,13 @@ if [ $stage -le 2 ]; then
       weight-post ark:- "ark,s,cs:gunzip -c $dir/weights.gz|" ark:- \| \
       ivector-extract --acoustic-weight=$posterior_scale --compute-objf-change=true \
         --max-count=$max_count --spk2utt=ark:$this_sdata/JOB/spk2utt \
-      $srcdir/final.ie "$feats" ark,s,cs:- ark,t:$dir/ivectors_spk.JOB.ark
+      $srcdir/final.ie "$feats" ark,s,cs:- ark:$dir/ivectors_spk.JOB.ark
   else
     $cmd JOB=1:$nj $dir/log/extract_ivectors.JOB.log \
       gmm-global-get-post --n=$num_gselect --min-post=$min_post $srcdir/final.dubm "$gmm_feats" ark:- \| \
       ivector-extract --acoustic-weight=$posterior_scale --compute-objf-change=true \
         --max-count=$max_count --spk2utt=ark:$this_sdata/JOB/spk2utt \
-      $srcdir/final.ie "$feats" ark,s,cs:- ark,t:$dir/ivectors_spk.JOB.ark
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/steps/nnet/ivector/extract_ivectors.sh
```diff
@@ -181,39 +181,35 @@ if [ $stage -le 3 ]; then
...
-  copy-vector "ark:cat $dir/ivectors_spk-as-utt.*.ark |" ark,scp:$absdir/ivectors_spk-as-utt.ark,$dir/ivectors_spk-as-utt.scp
-  rm $dir/ivectors_spk-as-utt.*.ark
   copy-vector "ark:cat $dir/ivectors_utt.*.ark |" ark,scp:$absdir/ivectors_utt.ark,$dir/ivectors_utt.scp
   rm $dir/ivectors_utt.*.ark
 fi
 
+# duplicate the `speaker' i-vectors to all its `utterances',
+if [ $stage -le 5 ]; then
+{ copy-vector scp:$dir/ivectors_spk.scp ark,t:- | \
+    awk -v spk2utt=$data/spk2utt 'BEGIN{ while (getline < spk2utt) { spk=$1; $1=""; spk_utts[spk]=$0; }}
```
danpovey: Here it would be more compact to do:
```
utils/apply_map.pl -f 2 $dir/ivectors_spk.scp <$data/utt2spk >$dir/ivectors_spk-as-utt.scp
```
instead of this whole block of code.

vesis84:
I see, that is a good remark. I have simplified it accordingly. 
(the spk->utt cloning of i-vectors is done at the scp file... no need to have the real per-utterance 'copies' of i-vector data...)

egs/wsj/s5/steps/nnet/ivector/extract_ivectors.sh
```diff
@@ -181,39 +181,28 @@ if [ $stage -le 3 ]; then
...
   rm $dir/ivectors_spk.*.ark
-  copy-vector "ark:cat $dir/ivectors_spk-as-utt.*.ark |" ark,scp:$absdir/ivectors_spk-as-utt.ark,$dir/ivectors_spk-as-utt.scp
-  rm $dir/ivectors_spk-as-utt.*.ark
   copy-vector "ark:cat $dir/ivectors_utt.*.ark |" ark,scp:$absdir/ivectors_utt.ark,$dir/ivectors_utt.scp
   rm $dir/ivectors_utt.*.ark
 fi
 
+# duplicate the `speaker' i-vector to all `utterances' of that speaker,
+if [ $stage -le 5 ]; then
+  utils/apply_map.pl -f 2 --permissive $dir/ivectors_spk.scp <$data/utt2spk >$dir/ivectors_spk-as-utt.scp
```
danpovey: Why did you need to add the --permissive option?  That option seems to just leave something unaffected if the mapping is not there, so in this case it would probably leave the speaker-id in case of the ivector.   That would just cause a crash later.

vesis84: Well, I thought there might be a mapping for which there is no i-vector. Okay, I see, this will work well even without the '--permissive' option. I'll remove it... (it was a friday afternoon commit :) )

Issue #886, opened 2016-07-07
nnet2 to nnet3 model converter


vijayaditya:
There have been multiple requests for a tool to convert nnet2 models into nnet3 models. 
There is no such tool currently available. However it would be simple to write and does not even need accessing the C++ parts of the nnet2 or nnet3 code.

The procedure is 
1. Convert nnet2 model to text format using nnet-am-copy
2. Isolate the parameter matrices corresponding to components (e.g. `NaturalGradientAffineComponent`) into separate files
3. Use these files to initialize the components in the nnet3 model using the `file-name` argument in the config string. For this write a nnet3 config which replicates the nnet2 model architecture and `nnet3-am-init` to initialize this model and write it out.

This would be a simple python script involving some string parsing.

jfainberg:
I'll start working on this (unless anyone else already is).

vijayaditya:
Please go ahead, no one has taken this yet.
Btw it would be helpful if you discuss your algorithm before implementation
as some changes can be a bit involved e.g. SpliceComponent to Descriptor
conversion.

Vijay

jfainberg:
Ok, this is getting there and I'll make WIP PR soon once it's a bit cleaned up.

I have one question: when initialising affine components with nnet3-init, it expects a matrix with column dim = input_dim + 1 (bias). The isolated parameter matrices from the nnet2 model are split into <LinearParameters> and <Bias>. I could either use something like `paste` to manually append the bias to the linear parameters, or modify ::InitFromConfig to take either matrix or linear_params and bias separately. The first is a bit of a hack, but perhaps ok for the intended use of this?

danpovey:
Either way is OK with me.

yujunlhz:
Dear Kaldi developers:

I'd like to know whether there is any update for this thread?

thanks

danpovey:
I am not aware of any update but it would be nice of Joachim comes through
with this.
Actually this is not a priority for me at all because as time goes on,
fewer people will be using nnet2 models.

jfainberg:
Sorry, other things took priority and I forgot to come back to this. I have done some work on it; I believe it was nearly ready. I'll have a look at it the coming weekend and make a PR.

hyuezhi:
[...]

PR #1611, opened 2017-05-07
Python script to convert nnet2 to nnet3 models

jfainberg:
Re #886 .

Initial implementation - I'll test it a bit more over the next few days, but open for comments.

Couple of points:
- Currently considers everything but the final model temporary (including the generated config it uses to initialise the model).
- Uses `numpy` to combine matrices and vectors into a single matrix for initialisation. Possible to do using just file manipulation, just a bit more messy.
- Ignores ValueSum and DerivSum stat fields of e.g. normalize and softmax components.

### DIFF
--- a/egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
+++ b/egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
@@ -1,0 +1,441 @@
+#!/usr/bin/env python
+# Copyright 2016
+
+# This script converts nnet2 models into nnet3 models.
+# It requires knowledge of valid components which
+# can be modified in the configuration section below.
+
+import argparse, subprocess, os, tempfile, logging, sys, shutil, fileinput, re
+from collections import defaultdict, namedtuple
+import numpy as np
+
+# Begin configuration section
+# Components and their corresponding node names
+
+NODE_NAMES = {
+    "<AffineComponent>":"affine",
+    "<AffineComponentPreconditioned>":"affine",
+    "<AffineComponentPreconditionedOnline>":"affine",
+    "<BlockAffineComponent>":"affine",
+    "<BlockAffineComponentPreconditioned>":"affine",
+    "<SigmoidComponent>":"nonlin",
+    "<TanhComponent>":"nonlin",
+    "<PowerComponent>":"nonlin",
+    "<RectifiedLinearComponent>":"nonlin",
+    "<SoftHingeComponent>":"nonlin",
+    "<PnormComponent>":"nonlin",
+    "<NormalizeComponent>":"renorm",
+    "<MaxoutComponent>":"maxout",
+    "<MaxpoolingComponent>":"maxpool",
+    "<ScaleComponent>":"rescale",
+    "<DropoutComponent>":"dropout",
+    "<SoftmaxComponent>":"softmax",
+    "<LogSoftmaxComponent>":"log-softmax",
+    "<FixedScaleComponent>":"fixed-scale",
+    "<FixedAffineComponent>":"fixed-affine",
+    "<FixedLinearComponent>":"fixed-linear",
+    "<FixedBiasComponent>":"fixed-bias",
+    "<PermuteComponent>":"permute",
+    "<AdditiveNoiseComponent>":"noise",
+    "<Convolutional1dComponent>":"conv",
+    "<SumGroupComponent>":"sum-group",
+    "<DctComponent>":"dct",
+    "<SpliceComponent>":"splice",
+    "<SpliceMaxComponent>":"splice"
+}
+
+SPLICE_COMPONENTS = [c for c in NODE_NAMES if "Splice" in c]
+AFFINE_COMPONENTS = [c for c in NODE_NAMES if "Affine" in c]
+
+KNOWN_COMPONENTS = NODE_NAMES.keys()
+# End configuration section
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter('%(asctime)s [%(filename)s:%(lineno)s - %(funcName)s - %(levelname)s ] %(message)s')
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+def GetArgs():
+    parser = argparse.ArgumentParser(description="Converts nnet2 into nnet3 models.",
+                                     epilog="e.g. steps/convert_nnet2_to_nnet3.py exp/tri4_nnet2 exp/tri4_nnet3");
+    parser.add_argument("--tmpdir", type=str, default="./",
+                        help="Set a custom location for the temporary directory.")
+    parser.add_argument("--skip-cleanup", action='store_true',
+                        help="Will not remove the temporary directory.")
+    parser.add_argument("--model", type=str, default='final.mdl',
+                        help="Choose a specific model to convert (default: final.mdl).")
+    parser.add_argument("--binary", type=str, default="false",
+                        help="Whether to write the model in binary or not.")
+    parser.add_argument("nnet2_dir", metavar="src-nnet2-dir", type=str,
+                        help="")
+    parser.add_argument("nnet3_dir", metavar="src-nnet3-dir", type=str,
+                        help="")
+
+    print(' '.join(sys.argv))
+
+    args = parser.parse_args()
+
+    if not os.path.exists(args.nnet3_dir):
+        os.makedirs(args.nnet3_dir)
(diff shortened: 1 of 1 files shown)


vijayaditya:
*"Uses numpy to combine matrices and vectors into a single matrix for initialisation. Possible to do using just file manipulation, just a bit more messy."*

I think this dependence on `numpy` for cleaner code is fine, as this binary is going to be used by people who might have sufficient experience with Kaldi.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+    "<SpliceMaxComponent>":"splice"
+}
+
+SPLICE_COMPONENTS = [c for c in NODE_NAMES if "Splice" in c]
+AFFINE_COMPONENTS = [c for c in NODE_NAMES if "Affine" in c]
+
+KNOWN_COMPONENTS = NODE_NAMES.keys()
+# End configuration section
+
+logger = logging.getLogger(__name__)
```
vijayaditya: I am not familiar with the latest logging guidelines. @vimalmanohar could you take a look.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+
+        # add priors
+        result += subprocess.call('nnet3-am-adjust-priors --binary={0} {1} {2} {3}'.format(binary, os.path.join(tmpdir, 'nnet3_no_prior.mdl'), 
+                                                                                           self.priors, 
+                                                                                           model), shell=True)
+
+        if (result != 0):
+            raise OSError('Encountered an error writing the model.')
+
+def ParseNnet2(line_buffer):
```
vijayaditya: Better to rename this as ParseNnet2ToNnet3

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+    '''
+    <InputDim> -> input-dim
+    '''
+    string = token[1:-1]
+    string = re.sub(r'((?<=[a-z])[A-Z]|(?<!\A)[A-Z](?=[a-z]))', r'-\1', string).lower()
+    return string
+
+def ConsumeToken(token, line):
+    '''Returns line without token'''
+    if token != line.split()[0]:
```
vijayaditya: IIRC these lines can be very large. So it might be better to just check for this using regexes or at least using split() method using something like a maxsplit=1 option.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+                    previous_component=MakeSpliceString(previous_component, component.pairs['<Context>'])
+                    continue
+                f.write('component-node name={name} component={name} input={inp}\n'.format(name=component.ident, inp=previous_component))
+                previous_component = component.ident
+            logger.warning('Assuming linear objective.')
+            f.write('output-node name=output input={inp} objective={obj}\n'.format(inp=previous_component, obj='linear'))
+
+    def WriteModel(self, model, binary="true"):
+        result = 0
+
```
vijayaditya: Check that self.config is a proper nnet3 config file.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+        result += subprocess.call('nnet3-am-init --binary=true {0} {1} {2}'.format(self.transition_model,
+                                                                                    os.path.join(tmpdir, 'nnet3.raw'),
+                                                                                    os.path.join(tmpdir, 'nnet3_no_prior.mdl')), shell=True)
+
+        # add priors
+        result += subprocess.call('nnet3-am-adjust-priors --binary={0} {1} {2} {3}'.format(binary, os.path.join(tmpdir, 'nnet3_no_prior.mdl'), 
+                                                                                           self.priors, 
+                                                                                           model), shell=True)
+
+        if (result != 0):
```
vijayaditya: Do you think it might be worth adding a check in a top level shell script which performs forward prop through the nnet2 and nnet3 models and asserts that the values are within acceptable threshold ?

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+                f.write('component-node name={name} component={name} input={inp}\n'.format(name=component.ident, inp=previous_component))
+                previous_component = component.ident
+            logger.warning('Assuming linear objective.')
+            f.write('output-node name=output input={inp} objective={obj}\n'.format(inp=previous_component, obj='linear'))
+
+    def WriteModel(self, model, binary="true"):
+        result = 0
+
+        # write raw model
+        result += subprocess.call('nnet3-init --binary=true {0} {1}'.format(self.config, os.path.join(tmpdir, 'nnet3.raw')), shell=True)
```
vijayaditya: Do you want to consider using the KaldiCommand function available in the nnet3 python libraries ? This might help you handle some common errors.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
vijayaditya: I don't know if there is a easy way to check the correctness of the transition model without doing a decode.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
...
+
+                f.write('component name={name} type={comp_type} {config_string}\n'.format(name=component.ident, comp_type=component.component, config_string=config_string))
+
+            f.write('\n# Component nodes\n')
+            f.write('input-node name=input dim={0}\n'.format(self.input_dim))
+            previous_component='input'
+            for component in self.components:
+                if component.ident == "splice":
+                    # Create splice string for the next node
+                    previous_component=MakeSpliceString(previous_component, component.pairs['<Context>'])
```
vijayaditya: It is not clear how you are handling the `SpliceMaxComponent` here ?

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
```diff
@@ -0,0 +1,441 @@
+#!/usr/bin/env python
+# Copyright 2016
+
```
vijayaditya: Please add your name to the author list.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
vimalmanohar: The logging is fine. Although, the guidelines for variable names and spacing in the Google style guide are not followed. https://google.github.io/styleguide/pyguide.html

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
jfainberg: Sorry, I had forgot about SpliceMaxComponent - it's not explicitly handled at the minute. What is the equivalent in nnet3? I couldn't find a max-descriptor (e.g. 'Max(Offset, ...)')

vijayaditya:
This is not something that can be done using a descriptor. You would need
to add a component. I think Max pooling component can be used but I don't
remember if the data needs to be rearranged. However I know of no
architectures which were using splicemaxcomponent in nnet2, so I think you
can just raise an error rather than handling this case.

egs/wsj/s5/steps/nnet3/convert_nnet2_to_nnet3.py
jfainberg: Good idea. How about another argument to the argparser for some features, and if present, will run a forward pass and compare the results using `difflib`; so keeping it all in the single Python script?

The transition model isn't read by Numpy, so shouldn't have any numerical errors I think.

danpovey:
So @jfainberg, this is ready to commit?  Still says WIP.

danpovey:
@vijayaditya  if you say it's still good I'll merge.

jfainberg:
Yes @danpovey, please go ahead. I've tested it (forward pass) with typical nnet2 pnorm and tanh networks.

vijayaditya:
@danpovey Went through it briefly. LGTM.

PR #1612, opened 2017-05-07
[WIP/not for merge]  Spatial averaging in CNNs.

danpovey:
This is about adding an extra pixel to the input image patch that consists of the average over the input image.  Doesn't seem to helpful for CNNs, at least on CIFAR-10.
I won't merge this; however I may leave the PR up for a bit for colloraborators to look at.
Also there are some changes/extension to AddMatBlocks() which will likely be useful for some other purpose in future, so we may at some point want to extract those changes.

### DIFF
--- a/egs/cifar/v1/local/nnet3/run_cnn_aug_1b2.sh
+++ b/egs/cifar/v1/local/nnet3/run_cnn_aug_1b2.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+
+# aug_1b2 is as aug_1b but using spatial averages.
+#  Doesn't help: valid acc 0.844 -> 0.8401, train acc 0.8973 -> 0.8932.
+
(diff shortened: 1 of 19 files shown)


danpovey:
@hhadian, would you mind re-testing this with the latest res-net recipe, or instructing Ashish or Jonathan how to do so?
I suspect with the merge conflict you can just accept the current kaldi_52 version, because the extra   cudamatrix function I added in this PR was added to kaldi_52 later to meet other needs, and may have been modified since.  Teaching those guys how to resolve merge conflicts would probably be a good thing to do anyway, if they don't know already.

danpovey:
BTW, the config-level change here is to add use-spatial-averages=true to the options of convolutional layers (or maybe just to some convolutional layers).  This could include convolutional layers inside res-blocks.

hhadian:
Sure, I will help them resolve the conflict and test this.
BTW, the cu-matrix changes in this PR are unrelated to spatial averaging, right?

danpovey:
They aren't unrelated as AddMatBlocks is used in this PR, but I believe the
version of that function currently in kaldi_52 should work fine, so you
could discard those changes and it should still work.

PR #1613, opened 2017-05-07
[WIP/not-for-merge] Initial resnet implementation

danpovey:
This is an implementation of ResNets.  They do seem helpful for CIFAR, e.g. we're getting 91% accuracy, although based on the papers we're referring to we should be getting better accuracy, around 95%.

I made it a bit more general than it needed to be, with some features that I now realize are not helpful, so I'm going to modify and simplify the implementation before this is ready for check in.  I am putting up this PR here just in case collorabors like @hhadian want to have a look.

This PR also contains some stuff that didn't really work regarding image augmentation in RGB space.

@freewym, you probably shouldn't use this, I'm working today on a more definitive implementation.


danpovey:
@freewym, actually maybe you can start from this point.
Before I merge this I want to clean this up and get rid of most of the resnet experiments from the repo, and some of the image augmentation stuff that didn't help; but if you look at local/nnet3/run_resnet_1b24.sh, that should be the best starting point.  Later I'll rename it to something more normal.

I'm running it in /home/dpovey/kaldi-chain/egs/cifar/v1; it's not finished running yet, but it's the cleanest version of ResNet (i.e. the closest to published setups).

PR #1614, opened 2017-05-08
Ivectors on images

david-ryan-snyder:
This PR adds the following:

* Binary src/featbin/extract-patches.cc which is used to extract all possible nxm patches in an image
* Scripts in egs/cifar/v1/images/ivector for computing PCA transform, training UBM/extractor and extracting i-vectors
* Script egs/cifar/v1/local/run_ivector.sh which demos extracting ivectors from images. It's invoked in the run.sh.

### DIFF
--- a/egs/cifar/v1/image/ivector/extract_ivectors.sh
+++ b/egs/cifar/v1/image/ivector/extract_ivectors.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+# Copyright     2013  Daniel Povey
+#               2014  David Snyder
+# Apache 2.0.
+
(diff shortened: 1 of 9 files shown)


danpovey:
looks like you merged with master.. probably a mistake.  I need to merge master to kaldi_52 and manually fix the conflicts.

david-ryan-snyder:
Yes, I just realized that... It's OK now though. No need to worry about this PR until after I get results adding this into the CNN. I'm looking into that now.

david-ryan-snyder:
@hhadian FYI

PR #1615, opened 2017-05-10
cudamatrix: accelerating the BLSTM training

vesis84:
- by removing some of the 'cudaDeviceSynchronize()' calls, the training
  time of a BLSTM epoch drops to 55.5% of original time (nnet1)
- consequently, the 'time-stats' in the 'profile' are a little inaccurate for some functions,

### DIFF
--- a/src/cudamatrix/cu-array.cc
+++ b/src/cudamatrix/cu-array.cc
@@ -33,11 +33,11 @@
 
 namespace kaldi {
 
-template<> 
+template<>
 void CuArray<int32>::Set(const int32 &value) {
   if (dim_ == 0) return;
 #if HAVE_CUDA == 1
-  if (CuDevice::Instantiate().Enabled()) { 
+  if (CuDevice::Instantiate().Enabled()) {
     Timer tim;
 
     dim3 dimBlock(CU2DBLOCK);
@@ -45,7 +45,7 @@ void CuArray<int32>::Set(const int32 &value) {
     ::MatrixDim d = { 1, Dim(), Dim() };
 
     cuda_int32_set_const(dimGrid, dimBlock, data_, value, d);
-    CU_SAFE_CALL(cudaGetLastError());
+    CUDA_GET_LAST_ERROR;
 
     CuDevice::Instantiate().AccuProfile(__func__, tim.Elapsed());
   } else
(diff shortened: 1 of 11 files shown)


danpovey:
Another side-effect of this change would be to make it hard to figure out where errors come from-- e.g. if a kernel produces an error, it may now appear to come from the next kernel invocation.

I wonder whether it could be made dependent on the verbose level.  That way if there is a problem we could run with higher verbose level.

vesis84:
Aha, I see, the unit-tests could then have high default 'verbose' level, while the training binaries would have smaller level to be fast by default... We could even support the fastest 'mode' which would produce the 'improper' time-stats (verbosity: 0 ... improper time-stats (fastest), 1 ... partially correct stats (faster), 2 ... exact stats (current code)).

I'll probably let it be for a moment, so your students can do the speed tests. We will test it on building the MGB eval. systems meanwhile (there's up to 1200h of data). Will return to it, once we have the eval systems (the eval data are realeased in a week).

danpovey:
@LvHang, could you please check this PR out and run with it on the mini-librispeech setup?  I want to look at the times and compare it with a previous run.

LvHang:
ok，i will get it right now
Daniel Povey <notifications@github.com>于2017年5月10日 周三00:00写道：

danpovey:
I think the timing stats will become pretty meaningless if the synchronize calls are not made, because it only launches the kernel in that case, it's not waiting for it to complete.  If they are correct, IMO it's only by chance of what order things were invoked in, so maybe best not to print such stats.

What I propose is that at verbose level 0 it doesn't accumulate the timing statistics at all, and doesn't print them, and doesn't call cudaDeviceSynchronize at all.  At verbose level 1 or higher it would do all of those things.

We could add code to neural net training scripts so that it runs with verbose level 1 every 20 iterations, so that if we really want to see those stats we can see them.

vesis84:
Dan, yes, the stats have 'less' of a meaning with the asynchronous invocation of kernels. Yes, they could be all disabled... Or eventually, we could accumulate the stats for the calls which are the 'blocking calls' as the GPU-CPU I/O (these do a 'sync' implicitly). Plus we could measure some selected functions like cuBLAS calls (with a sync before, and a sync after)... It would be good to play with it a bit and see carefully compare the runtimes :)

Depending on the measurements, the possible scenarios are:
A)
verbose 0 : no stats (or only the no-overhead stats)
verbose 1 : some stats (I/O, cuBLAS, ???)
verbose 2 : all the stats (2x slower BLSTM)
B)
verbose 0 : some stats (I/O, cuBLAS, ??? ; almost no overhead)
verbose 1 : all the stats (2x slower BLSTM)

danpovey:
I don't think there is much value in stats that are partial or wrong; if
people want stats they can set verbose=1.
In addition, a lot of code is already printing out extremely verbose stuff
at verbose level 2, so the verbose levels would all have to be bumped up if
we did what you propose, which would be quite painful.

danpovey:
Closing this PR since merged alternative version.

PR #1616, opened 2017-05-10
[src] Fix bug found by Hainan and Yiming regarding looped computation.

danpovey:


danpovey:
@freewym, this is a bug fix for the problem that you and Hainan encountered with the RNNLM stuff.
Unfortunately it would not be trivial to fix it in master, so the fix is only available in kaldi_52.  This means you will have to upgrade all your RNNLM work to start from the kaldi_52 branch.
However, it needs testing to make sure that it doesn't break regular online decoding.

Could you please run online decoding in some setup with this version of the code?  Often the scripts don't run it by default even if they support it, you have to add  some kind of flag.

freewym:
OK

freewym:
Do I need to do like what is in `egs/mini_librispeech/s5/local/chain/run_tdnn.sh`, i.e.,
call `steps/online/nnet3/prepare_online_decoding.sh` and `steps/online/nnet3/decode.sh`?
If yes, can I just use mini_libirspeech to test?

danpovey:
You can use mini-librispeech to test.
But make sure you test both TDNN and TDNN-LSTM.
Also please, once you can merge, test that the fix resolves the problem
that you and Hainan found; I no longer intend to test this myself.

freewym:
I tested the online decoding on mini-librispeech using chain+TDNN and xent+TDNN-LSTM respectively, and got exactly the same results as before.

danpovey:
OK, thanks; I'll merge, and if any other problems with this show up we'll address them at that time.

freewym:
@hainan-xv just merged it with kaldi_52 and tested it on our rnnlm rescoring, and the problem looks resolved.

PR #1622, opened 2017-05-15
Add SVHN dataset

hhadian:


danpovey:
I ran the resnet setup from cifar (with just 30 epochs), and I got 1.60% error.
See /home/dpovey/kaldi-chain/egs/svhn/v1/exp/resnet1b.
This is better than most of the results in the WRN paper https://arxiv.org/pdf/1605.07146.pdf, which I believe has the best results, and very close to their best result of 1.54% which was with a much larger model with way more epochs.  A tiny bit of tuning and we'd have the best published results on SVHN.

hhadian:
This is really amazing! I guess these results should be published (at least in arxiv papers) for everyone to see.

danpovey:
Closing this because I merged it indirectly.

PR #1630, opened 2017-05-18
[src,egs,scripts] Add SVHN example; fix asymmetry in image-augmentation; minor script changes.

danpovey:


danpovey:
This is a tuned version of @hhadian's SVHN recipe.
@freewym, can you please start right away a backstitch version of the resnet1b SVHN recipe?  The baseline results are just about as good as the best published ones, so it will be impressive if we manage to beat the best published numbers on SVHN.

PR #1632, opened 2017-05-18
Minor fix regarding first/last chunk's right-context in chain models

hhadian:


danpovey:
Note: the bug that this is a fix for, can cause a train/test mismatch in the right context for BLSTMs, so if there are any "newer" BLSTM scripts, in which --extra-right-context-final is set to 0, this might improve the results for those scripts.  However, most of the BLSTM scripts currently in the repo seem to be older scripts that would not be affected by this bug.
@hhadian is testing this.

hhadian:
while testing right-context-final=0 with blstm, it fails at iteration 0:
`ERROR (nnet3-chain-copy-egs[5.1.93~1-7af21]:ModifyChainExampleContext():nnet3-chain-copy-egs.cc:229) You requested --right-context=42, but example only has right-context of 6 (will warn only once; this may be harmless if using any --*right-context-final options.`

I guess the `KALDI_ERR` at [line 229](https://github.com/kaldi-asr/kaldi/blob/master/src/chainbin/nnet3-chain-copy-egs.cc#L229) is supposed to be `KALDI_WARN` maybe?

danpovey:
yes it should be kaldi_warn.

hhadian:
Using a final right-context of 0 gave a mere 0.2% improvement (all of it is on the callhome subset, which is improved from 19.0 to 18.6):
```
# local/chain/compare_wer_general.sh blstm_6k_sp blstm_6k_final0_sp
# System                blstm_6k_sp blstm_6k_final0_sp
# WER on train_dev(tg)      13.28     13.26
# WER on train_dev(fg)      12.31     12.27
# WER on eval2000(tg)        15.6      15.4
# WER on eval2000(fg)        14.2      14.0
# Final train prob         -0.055    -0.054
# Final valid prob         -0.076    -0.078
# Final train prob (xent)        -0.762    -0.754
# Final valid prob (xent)       -0.8616   -0.8748
```

danpovey:
OK, well anyway it's a bug fix, even if it didn't help we should be
committing it.
Thanks for testing it..
I may just go ahead and merge it-- but I believe you need to check in one
more fix WARN->ERR

hhadian:
I just checked in the other commit.

PR #1633, opened 2017-05-18
Transfer learning wsj->rm

pegahgh:
This is the Transfer learning setup which transfers the model trained on wsj to rm.
The preliminary result without any tuning shows improvement(2.71 -> 2.21)


danpovey:
Thanks-- I'll try to look at this tonight.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -49,11 +49,63 @@ def xconfig_line_to_object(config_line, prev_layers = None):
               "*** {0}".format(config_line), file=sys.stderr)
         raise e
 
+# This function reads existing model file with nnet3 format and returns it as
+# list of layers with name and dimension to be used as auxilary information
+# to generate xconfig.
+def read_model(model_filename):
```
danpovey: there is no documentation of what the return format of this function is, and the name is too non-specific (e.g. get_model_component_info).  Don't use the word 'layer' if what you really mean is component, they are different things.  And grep for '^component-node', not -node.  You could otherwise get problems in future if something that looks like that appears in the file.

egs/babel_multilang/s5/local/nnet3/run_tdnn_multilingual.sh
```diff
@@ -204,9 +204,12 @@ EOF
     echo " output-layer name=output-${lang_index} dim=$num_targets max-change=1.5"
   done >> $dir/configs/network.xconfig
 
+  cat <<EOF > $dir/configs/edits.config
+  rename-node old-name=output-0 new-name=output
+EOF
   steps/nnet3/xconfig_to_configs.py --xconfig-file $dir/configs/network.xconfig \
     --config-dir $dir/configs/ \
-    --nnet-edits="rename-node old-name=output-0 new-name=output"
+    --nnet-edits=$dir/configs/edits.config
```
danpovey: nnet-edits expects a string, not a filename.  also you are already supplying --config-dir as an arg to the script and I see that elsewhere you are treating edits.config as a standard part of that config directory, so there seems to be some disagreement about whether scripts are supposed to pick that up automatically, or whether it needs to be supplied explicitly.
If you are intending to change how that option works you should rename it and chang the documentation.

egs/rm/s5/local/chain/run_tdnn_wsj_rm.sh
```diff
@@ -0,0 +1,184 @@
+#!/bin/bash
+set -e
```
danpovey: please add explanation of what this script does.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -28,6 +28,12 @@ def get_args():
         epilog='Search egs/*/*/local/{nnet3,chain}/*sh for examples')
     parser.add_argument('--xconfig-file', required=True,
                         help='Filename of input xconfig file')
+    parser.add_argument('--existing-model',
```
danpovey: The doc-string doesn't explain what this option expects, which is presumably a filename of a .mdl
file containing a previously trained neural net.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -219,17 +225,19 @@ def write_config_files(config_dir, all_layers):
             raise
 
 
-def add_back_compatibility_info(config_dir, nnet_edits=None):
+def add_back_compatibility_info(config_dir, existing_model=None,
```
danpovey: This function is differently named in kaldi_52.  Please make this PR against kaldi_52 as it will save me hassle in merging.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -310,7 +310,8 @@ def train(args, run_opts, background_process_handler):
         logger.info("Creating denominator FST")
         chain_lib.create_denominator_fst(args.dir, args.tree_dir, run_opts)
 
-    if (args.stage <= -4):
+    if (args.stage <= -4) and os.path.exists("{dir}/config/init.config".format(
```
vimalmanohar: The directory should be `configs` instead of `config`

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
pegahgh: I grep  '-node', since we need both component-node and input-node!

egs/babel_multilang/s5/local/nnet3/run_tdnn_multilingual.sh
pegahgh: I modified the script to accept nnet-edits as filename and use it using edits-config. If you agree with that, I will update the description and option name as edits-config.
Also edits.config passed explicitly to xconfig_to_configs.py but train.py automatically picks up edits.config at prepare_acoustic_model function.

danpovey:
OK.

pegahgh:
@vimalmanohar you mentioned you already tested this PR. Would you let me update me with your point about this PR.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,66 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+            if len(key_value) == 2:
+                key = key_value[0]
+                value = key_value[1]
+                if key == "name":
+                    layer_name = value
+                    #layer_name, auxiliary_output = xutils.split_layer_name(value)
+                elif key == "input-dim":
+                    input_dim = int(value)
+                elif key == "output-dim":
+                    output_dim = int(value)
```
vimalmanohar: The key can also be "dim", in which case it would be both input-dim and output-dim.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,66 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+                elif key == "input":
+                    input_str = value
+
+        if layer_name is not None and layer_name not in layer_names:
+            key_to_value = dict()
+            layer_names.append(layer_name)
+            key_to_value['name'] = layer_name
+            if  input_dim != -1:
+                if output_dim == -1:
+                    # The layer is input layer type.
```
vimalmanohar: I am not sure what this is for. There may never be only input_dim defined with output_dim not defined.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,66 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+                    key_to_value['dim'] = input_dim
+                elif input_str is not None:
+                    key_to_value['dim'] = output_dim
```
vimalmanohar: input_str is None for input-node, but its dim is still defined by 'dim'

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,66 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
 
+# This reads raw existing model (*.raw) and returns array of
+# XconfigInputLayer one per input-node or component-node with same 'name' used
+# in raw model and 'dim' equal to 'output-dim' for component-node and input-dim for
+# input-node.
+def get_model_component_info(model_filename):
+    all_layers = []
+    try:
+        f = open(model_filename, 'r')
+    except Exeption as e:
```
vimalmanohar: Exception

pegahgh:
@vimalmanohar I fixed some issues about this PR and merged it with master. Since you used this PR, do you have any comments on that?

egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -424,6 +424,19 @@ def prepare_initial_acoustic_model(dir, run_opts, srand=-1):
     common_train_lib.prepare_initial_network(dir, run_opts,
                                              srand=srand)
 
+    # edits 0.raw using edits.config before adding transition model.
```
vimalmanohar: Add an example for edits.config here and explain that it is used for transfer learning.

egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -424,6 +424,19 @@ def prepare_initial_acoustic_model(dir, run_opts, srand=-1):
...
+    edits_config_file = "{0}/configs/edits.config".format(dir)
+    if os.path.exists(edits_config_file):
+        logger.info("edits 0.raw model using {0}/configs/edits.config."
+                    "".format(dir))
+        common_lib.execute_command(
+            """{command} {dir}/log/edit.log \
+                nnet3-copy --edits-config={edits_config} {dir}/0.raw \
```
vimalmanohar: I suggest you move this inside prepare_initial_network as this is not just for chain objf, but generic for anything. 
Also you are modifying 0.raw itself. Its better to write an intermediate model like pre_edit.raw.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,59 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+    layer_names = []
+    for line in out.split("\n"):
+        parts = line.split(" ")
+        input_dim = -1
+        for  field in parts:
+            key_value = field.split("=")
+            if len(key_value) == 2:
+                key = key_value[0]
+                value = key_value[1]
+                if key == "name":
```
vimalmanohar: Give examples for different types of lines that will be parsed. input-node, component-node etc.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,59 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+                elif key == "dim":
+                    input_dim = int(value)
+                elif key == "output-dim":
+                    input_dim = int(value)
+
+        if layer_name is not None and layer_name not in layer_names:
+            key_to_value = dict()
+            layer_names.append(layer_name)
+            key_to_value['name'] = layer_name
+            if  input_dim != -1:
```
vimalmanohar: Comment which type of lines go to which case.

egs/wsj/s5/steps/libs/nnet3/xconfig/utils.py
```diff
@@ -60,6 +60,10 @@ def get_dim_from_layer_name(all_layers, current_layer, full_layer_name):
     for layer in all_layers:
         if layer is current_layer:
             break
+        # if "." used in layer name
```
vimalmanohar: Give example for when this is applicable.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -219,17 +227,20 @@ def write_config_files(config_dir, all_layers):
             raise
 
 
-def add_nnet_context_info(config_dir, nnet_edits=None):
+def add_nnet_context_info(config_dir, existing_model=None,
+                                edits_config=None):
```
vimalmanohar: Check indentation.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
```diff
@@ -219,17 +227,20 @@ def write_config_files(config_dir, all_layers):
...
-    if nnet_edits is not None:
-        model = "nnet3-copy --edits='{0}' {1} - |".format(nnet_edits,
-                                                          model)
-    out = common_lib.get_command_stdout('nnet3-info "{0}" | head -n 4 '
-                                        .format(model))
+    if edits_config is not None:
+        model = """ - | nnet3-copy --edits-config={0} - {1}""".format(edits_config,
+                                                              model)
+    common_lib.execute_command("""nnet3-init {0} {1}/ref.config """
+                                 """ {2} """.format(existing_model if
```
vimalmanohar: Check indentation. 
I am not sure if @danpovey wants this to be in Google standards, since the rest of this file is not.

egs/babel_multilang/s5/local/nnet3/run_tdnn_multilingual.sh
```diff
@@ -202,11 +202,13 @@ EOF
     echo " relu-renorm-layer name=prefinal-affine-lang-${lang_index} input=tdnn_bn dim=1024"
```
vimalmanohar: I this supposed to be in this PR? Also I do not see any WSJ -> RM examples.

egs/wsj/s5/steps/nnet3/xconfig_to_configs.py
danpovey: Thanks for checking this, Vimal.
I'd like it to conform to Google standards as much as possible, but don't be too strict with the column-length of 80, it's hard to stick to that  in Python, what with the indentation.

pegahgh:
@vimalmanohar  would you 1) review the PR and comment on that and also 2) test it on MGB to verify if you can get the best results you had?
The new weight transfer degrades performance on wsj->rm, but it gives good improvement on libri->wsj.
Since the PR name is transfer-learning-wsj-rm, I didn't add scripts for libri->wsj, but I can add them on next commit and rename PR.

egs/wsj/s5/steps/nnet3/chain/train_more.py
```diff
@@ -0,0 +1,590 @@
...
+
+    arg_string = pprint.pformat(vars(args))
+    logger.info("Arguments for the experiment\n{0}".format(arg_string))
+
+    # Check files
+    chain_lib.check_for_required_files(args.feat_dir, args.tree_dir,
+                                       args.lat_dir)
+
+    # Set some variables.
+    num_jobs = 4 #common_lib.get_number_of_jobs(args.lat_dir)
```
vimalmanohar: Why is this fixed at 4?

egs/wsj/s5/steps/nnet3/chain/train_more.py
```diff
@@ -0,0 +1,590 @@
...
+    if (args.num_jobs_final > num_archives_expanded):
+        raise Exception('num_jobs_final cannot exceed the '
+                        'expanded number of archives')
+
+    # copy the properties of the egs to dir for
+    # use during decoding
+    logger.info("Copying the properties from {0} to {1}".format(egs_dir, args.dir))
+    common_train_lib.copy_egs_properties_to_exp_dir(egs_dir, args.dir)
+
+    if (args.stage <= -2) and os.path.exists(args.dir+"/configs/init.config"):
```
vimalmanohar: This stage is not required in this script. But you should check that init.raw exists.

egs/babel_multilang/s5/local/nnet3/run_tdnn_multilingual.sh
```diff
@@ -204,9 +204,16 @@ EOF
...
 
+  cat <<EOF > $dir/configs/edits.config
+  rename-node old-name=output-0 new-name=output
+EOF
   steps/nnet3/xconfig_to_configs.py --xconfig-file $dir/configs/network.xconfig \
     --config-dir $dir/configs/ \
-    --nnet-edits="rename-node old-name=output-0 new-name=output"
+    --edits-config=$dir/configs/edits.config
+
+  cat <<EOF >> $dir/configs/vars
```
pegahgh: I updated xconfig_to_config to accept edits-config instead of nnet-edit, so this script should be updated.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm.sh
```diff
@@ -0,0 +1,192 @@
+#!/bin/bash
```
pegahgh: This should be soft link to _1a

egs/rm/s5/local/prepare_wsj_rm_lang.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+# Copyright 2017 Pegah Ghahremani
+
+# This script prepares a dictionary for sourc-target using source phone set, lexicon and dict and
```
pegahgh: I planned to make it generic, but it is tough to make it generic. 
The name is wsj_rm, so it should not mislead the user!

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm.sh
vimalmanohar: This is not a softlink. Git will show it as softlink if it is. There is already local/chain/run_tdnn_wsj_rm.sh which is a softlink to _1a.

egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -502,13 +502,34 @@ def smooth_presoftmax_prior_scale_vector(pdf_counts,
 
 
 def prepare_initial_network(dir, run_opts, srand=-3):
-    if os.path.exists(dir+"/configs/init.config"):
+    if not os.path.exists("{0}/configs/init.config".format(dir)):
```
pegahgh: We accept edit.config just in cases where init.raw exists does not generated using init.config like copying init.raw model from another dataset

danpovey: Pegah, I don't understand your reply *or* this code.

egs/wsj/s5/steps/nnet3/chain/train_more.py
```diff
@@ -0,0 +1,590 @@
+#!/usr/bin/env python
+
+# Copyright 2016    Vijayaditya Peddinti.
+#           2016    Vimal Manohar
+# Apache 2.0.
+
+""" This script is based on steps/nnet3/chain/train.sh
```
pegahgh: ok, I will do that, but I believe amount of changes to train.py is very low and we can still apply them to train.py!

vimalmanohar: Add in the description how this script is different from train.py and what it is for.

egs/wsj/s5/steps/nnet3/chain/train_more.py
pegahgh: I am not still sure, if it is good idea to have separate script train_more.py!!

egs/wsj/s5/steps/libs/nnet3/train/common.py
pegahgh: I mean, we should change the code to check os.path.exists(dir+"/init.raw") not init.config
Also We should have another assertion check to check at most one of the two files init.config and edit.config should exist.

danpovey:
I don't believe there is a current use-case where init.config will exist--
unless you have something like that in the multilingual setup.
We removed support for layer-by-layer pretraining from the xconfig creation
scripts, and I think the training scripts no longer support it either (i.e.
there is no support for adding new layers from layer1.config, layer2.config
and so on).

egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -21,31 +22,54 @@
...
 def create_phone_lm(dir, tree_dir, run_opts, lm_opts=None):
-    """Create a phone LM for chain training
+    """Create a phone LM for chain training.
 
-    This method trains a phone LM for chain training using the alignments
+    This method trains a weighted phone LM for chain training generated by
+    weighted combination of counts in alignments
     in "tree_dir"
+    "tree_dir" is a list of comma-separated alignments directories with
+    same tree and phone sets (each containing tree and ali.*.gz)
```
pegahgh: you are right! removed it from "_1b"!

egs/wsj/s5/steps/libs/nnet3/train/common.py
pegahgh: We still check existence of init.config and use it for stage -4 (init.raw generation) and stage -2 (lda generation)

danpovey:
Oh yes, I forgot, init.config is expected to exist after all, in the normal
case.
But I don't like the way you are changing all these things that are mixed
up in the normal code path-- it's not clear in your changed code what
relates to the transfer learning (or training an already-changed system)
and what does not.  That needs to be clarified and it needs to be easy to
verify, from reading the changes on github, that for non-transfer learning,
it would behave the same as previously.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
```diff
@@ -0,0 +1,192 @@
...
+  mkdir -p $dir/configs
+  cat <<EOF > $dir/configs/network.xconfig
+  relu-renorm-layer name=tdnn7-target input=Append(tdnn6.renorm@-3,tdnn6.renorm@0) dim=$dim
+  ## adding the layers for chain branch
+  relu-renorm-layer name=prefinal-chain-target input=tdnn7-target dim=$dim target-rms=0.5
+  output-layer name=output-target include-log-softmax=false dim=$num_targets max-change=1.5
+  relu-renorm-layer name=prefinal-xent-target input=tdnn7-target dim=$dim target-rms=0.5
+  output-layer name=output-xent-target dim=$num_targets learning-rate-factor=$learning_rate_factor max-change=1.5
+EOF
+  # edits.config contains edits required to train transferred model.
```
pegahgh: If we have xconfig-file containing new output-layers with same name 'output' and 'output-xent', it fails since we used more than one layer with same name!

danpovey:
Show me the error.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
pegahgh: RuntimeError: Name 'output' is used for more than one layer

pegahgh: File "steps/libs/nnet3/xconfig/basic_layers.py", line 47, in __init__
    "layer.".format(self.name))
RuntimeError: Name 'output' is used for more than one layer.

danpovey:
You were the one who modified the xconfig setup to take the 'aux_layers'.
So in effect that's an error produced by your own code, since the name
conflict is with something from there.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
pegahgh: I fixed the issue by adding extra class as XconfigAuxiliaryLayer and all layers in existing model are parsed as an object of this class and there is no error if the new layers have name conflict with auxiliary layer names.

src/chainbin/chain-est-phone-lm.cc
```diff
@@ -32,37 +32,51 @@ int main(int argc, char *argv[]) {
...
         "gunzip -c input_dir/ali.*.gz | ali-to-phones input_dir/final.mdl ark:- ark:- | \\\n"
         " chain-est-phone-lm --leftmost-context-questions=dir/leftmost_questions.txt ark:- dir/phone_G.fst\n";
 
     bool binary_write = true;
     LanguageModelOptions lm_opts;
+    std::string scales_str;
 
     ParseOptions po(usage);
     po.Register("binary", &binary_write, "Write output in binary mode");
+    po.Register("scales", &scales_str, "comma-separated list of integer valued scale weights used to scale different phone sequences.");
```
pegahgh: I think it is a simple change and it is better to keep scales, since in some cases scale as or 100x  and we need to repeat 10 or 100 times, which takes 100x times to process alignments.

danpovey:
This will never dominate the training time because the only purpose of the
scales is to repeat 'small' datasets so they are not too much smaller than
'large' datasets.  So I'd prefer that you do it the way I said-- I don't
like too many code changes when they are not necessary

vimalmanohar:
The command fails if its too long, which is possible for reasonably sized names. num-jobs and operating systems. So I do not understand why such a simple change of code is problematic when it makes things more flexible and faster.
```
var=(); for n in `seq 300`; do var+=(/home/vmanoha1/workspace_chain_semisup/egs/fisher_english/s5/exp/semisup_15k/chain_semi15k_250k/tdnn7b_ex250k_prun2.0_lmwt0.0_tol2_comb1a/decode_dev/lat.*.gz); done; gunzip -c ${var[@]}| lattice-copy ark:- ark,t:-
lattice-copy ark:- ark,t:-
bash: /bin/gunzip: Argument list too long
```

egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -20,32 +21,64 @@
 logger.addHandler(logging.NullHandler())
 
 
-def create_phone_lm(dir, tree_dir, run_opts, lm_opts=None):
-    """Create a phone LM for chain training
+def create_phone_lm(dir, tree_dir, run_opts, alignment_dirs=None, lm_opts=None):
```
danpovey: Vimal pointed out that this might fail due to argument list too long.  But you can work around that by doing:
``` for x in [arg_list]; do gunzip -c $x; done | ... ```
I don't like the C++ code changes because having a scale that is required to be an integer is kind of weird, and because IMO it's just more elegant to repeat the input that many times, since that is what the code does.  And also because of the principle that the "unusual case" shouldn't cause code to be modified in things used  in the "common case".  I want to keep the common case simple, which makes the codebase easier to maintain.

File: egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh [hunk 30]
```diff
@@ -66,14 +66,23 @@ done
...
-  alignments+=("ark:gunzip -c $adir/ali.*.gz | ali-to-phones $adir/final.mdl ark:- ark:- |")
+  w=`echo $weights | cut -d, -f$[$n+1]`
+  if ! [[ $w =~ ^[+]?[0-9]+$ ]]; then
+    echo "no positive integer weight specified for alignment $adir" && exit 1;
+  fi
+  repeated_ali_to_process=""
+  for x in `seq $w`;do
+    repeated_ali_to_process="ark:gunzip -c $adir/ali.*.gz $repeated_ali_to_process"
+  done
+  alignments+=("$repeated_ali_to_process | ali-to-phones $adir/final.mdl ark:- ark:- |")
```
vimalmanohar: This will not work, also it will exceed the line limit. I don't think the solution @danpovey mentioned is applicable here. You might have to write the combined and alignments to disk.

pegahgh:
I am not sure if we can process for loops inside queue.pl!

danpovey:
I think you can; you just need to escape the semicolon so it isn't
interpreted by the shell too soon.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
```diff
@@ -0,0 +1,199 @@
...
+where "nvcc" is installed.
+EOF
+fi
+
+# The iVector-extraction and feature-dumping parts are the same as the standard
+# nnet3 setup, and you can skip them by setting "--stage 8" if you have already
+# run those things.
+
+
+lang_dir=data/lang_wsj_rm
```
vimalmanohar: Add a comment that this will be created using phones from WSJ and words (and G.fst?) from RM.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
```diff
@@ -0,0 +1,199 @@
...
+    echo "$0: no such file $f"
+  fi
+done
+
+if [ $stage -le -1 ]; then
+  echo "$0: prepare lexicon.txt for RM using WSJ lexicon."
+  if ! cmp -s <(grep -v "^#" $src_lang/phones.txt) <(grep -v "^#" data/lang/phones.txt); then
+    local/prepare_wsj_rm_lang.sh  $srcdir/data/local/dict_nosp $srcdir/data/lang $lang_dir
+  else
+    rm -rf $lang_dir
```
vimalmanohar: `rm -r $lang_dir 2>/dev/null || true`

egs/rm/s5/local/prepare_wsj_rm_lang.sh
```diff
@@ -0,0 +1,69 @@
...
+rm -r $output_dir 2>/dev/null
+mkdir -p $output_dir
+mkdir -p $output_dir/local
+# copy *phones.txt from source to target.
+cp -r $src_dict $output_dir/local/dict
+rm $output_dir/local/dict/lexicon*.txt
+
+oov_word=`cat $src_lang/oov.txt`
+# common word list in rm lexicon with lexicon in wsj
+comm -12 <(awk '{print $1}' data/local/dict/lexicon.txt | sed "s/\+/\'/g" | sort) \
```
vimalmanohar: I don't think the previous comments about this file were addressed. You can do this more easily using perl instead of multiple sort statements.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# begin configuration section.
+cmd=run.pl
+stage=-10
+weights= # comma-separated list of integer valued scale weights used
+         # to scale different phone sequences for different alignments.
+lm_opts='num_extra_lm_state=2000'
+#end configuration section.
+
+help_message="Usage: "$(basename $0)" [options] <ali-dir1> [<ali-dir2> ...] <out-dir>
```
pegahgh: I didn't change it. It was old thing and I think you wrote it!!!

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
```diff
@@ -0,0 +1,199 @@
+#!/bin/bash
+# _1b is as _1a, but different as follows
```
pegahgh: I think it is good to keep it. Just in cases that chain model doesn't give good alignments as gmm model and it may need tuning left and right tolerance.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
```diff
@@ -0,0 +1,199 @@
...
+num_jobs_initial=2
+num_jobs_final=4
+minibatch_size=32
+frames_per_eg=150
+remove_egs=false
+xent_regularize=0.1
+
+# configs for transfer learning
+common_egs_dir=
+srcdir=../../wsj/s5/
```
pegahgh: I think it is better to clarify them, but it is better to keep them, since the chain src model can be in different directory . I think it is better to remove srcdir.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
vimalmanohar: You can move the variable help_message to below parse_options.sh. Then it might be fine.

pegahgh: I prefer to have the same style as other usage messages in other kaldi files.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
```diff
@@ -80,12 +86,13 @@ fi
...
 
-lang_dir=data/lang_wsj_rm
+lang_src_tgt=data/lang_wsj_rm # This dir is prepared using phones.txt and lexicon from
+                          # WSJ and wordlist and G.fst from RM.
 ali_dir=exp/tri4b${src_tree_dir:+_wsj}_ali
 lat_dir=exp/tri3b_lats${src_tree_dir:+_wsj}
 dir=exp/chain/tdnn_wsj_rm${tdnn_affix}
 
-required_files="$src_mdl $src_lang/phones.txt $srcdir/data/local/dict_nosp/lexicon.txt $src_gmm_dir/final.mdl $src_tree_dir/tree"
+required_files="$src_mdl $srcdir/exp/nnet3/extractor/final.mdl $src_lang/phones.txt $srcdir/data/local/dict_nosp/lexicon.txt $src_gmm_dir/final.mdl $src_tree_dir/tree"
```
vimalmanohar: There is not final.mdl in extractor. It should be final.ie

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1c.sh
```diff
@@ -162,8 +166,11 @@ if [ $stage -le 8 ]; then
   if [ $train_stage -lt -4 ]; then
     train_stage=-4
   fi
-
-  steps/nnet3/chain/train_more.py --stage $train_stage ${chain_opts[@]} \
+  # we used chain model from source to generate lats for target and the
+  # tolerance used in chain egs generation using this lats should be 1 or 2 which is
+  # (source_egs_tolerance/frame_subsampling_factor)
```
vimalmanohar: Mention source_egs_tolerance is 5.

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
```diff
@@ -1104,27 +1105,23 @@ def get_input_descriptor_names(self):
 
         return []  # there is no 'input' field in self.config.
 
-    def output_name(self, auxiliary_outputs = None):
+    def output_name(self, auxiliary_outputs=None):
```
vimalmanohar: Remove the blank lines after function name.

egs/wsj/s5/steps/nnet3/align_lats.sh
```diff
@@ -0,0 +1,190 @@
...
+fi
+echo "$0: feature type is $feat_type"
+
+cmvn_opts=`cat $srcdir/cmvn_opts 2>/dev/null`
+cp $srcdir/cmvn_opts $dir 2>/dev/null
+
+case $feat_type in
+  raw) feats="ark,s,cs:apply-cmvn $cmvn_opts --utt2spk=ark:$sdata/JOB/utt2spk scp:$sdata/JOB/cmvn.scp scp:$sdata/JOB/feats.scp ark:- |"
+   ;;
+  lda)
```
vimalmanohar: Remove the lda feat_type. It is no longer supported in nnet3.

egs/wsj/s5/steps/nnet3/align_lats.sh
```diff
@@ -0,0 +1,190 @@
...
+    echo "$0: expected $transform_dir/num_jobs to contain the number of jobs." && exit 1;
+  nj_orig=$(cat $transform_dir/num_jobs)
+
+  if [ $feat_type == "raw" ]; then trans=raw_trans;
+  else trans=trans; fi
+  if [ $feat_type == "lda" ] && ! cmp $transform_dir/final.mat $srcdir/final.mat; then
+    echo "$0: LDA transforms differ between $srcdir and $transform_dir"
+    exit 1;
+  fi
+  if [ ! -f $transform_dir/$trans.1 ]; then
```
vimalmanohar: Fmllr is also not supported. Compare with steps/nnet3/align.sh.

egs/wsj/s5/steps/nnet3/align_lats.sh
```diff
@@ -0,0 +1,190 @@
...
+lang=$2
+srcdir=$3
+dir=$4
+
+oov=`cat $lang/oov.int` || exit 1;
+mkdir -p $dir/log
+echo $nj > $dir/num_jobs
+sdata=$data/split${nj}utt
+[[ -d $sdata && $data/feats.scp -ot $sdata ]] || \
+   split_data.sh --per-utt $data $nj || exit 1;
```
vimalmanohar: It might be better to make this without per-utt. Check how it is in steps/nnet3/align.sh

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -323,7 +323,8 @@ def train(args, run_opts):
         shutil.copy('{0}/tree'.format(args.tree_dir), args.dir)
         chain_lib.create_denominator_fst(args.dir, args.tree_dir, run_opts)
 
-    if (args.stage <= -4) and os.path.exists(args.dir+"/configs/init.config") and args.input_model is None:
+    if (args.stage <= -4) and os.path.exists(args.dir+"/configs/init.config")
```
vimalmanohar: You have to add paranthesis when you break lines, otherwise it will fail.

egs/wsj/s5/steps/nnet3/align_lats.sh
pegahgh: we split per-utt in align.sh

egs/wsj/s5/steps/nnet3/align_lats.sh
pegahgh: Fmllr is still supported in steps/nnet3/align.sh.

danpovey:
Then it's out of date.  We've been removing support for LDA/fMLLR features
from those scripts, to simplify them.

egs/wsj/s5/steps/nnet3/align_lats.sh
pegahgh: we still have raw fmllr, but fmllr after lda removed.

danpovey:
Oh.  Maybe keeping raw fMLLR is OK, if still supported by most scripts;
however, in practice we never use it.  We could try it again though; maybe
there are scenarios where it really helps.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
```diff
@@ -0,0 +1,189 @@
...
+num_jobs_initial=2
+num_jobs_final=4
+minibatch_size=128
+frames_per_eg=150
+remove_egs=false
+xent_regularize=0.1
+
+# configs for transfer learning
+srcdir=../../wsj/s5   # base dir for source dataset.
+src_tdnn_affix=1d
```
vimalmanohar: It's not a good idea to have `src_tdnn_affix`. The affix is used differently in different places. It will be easier to adapt to other recipes if add an option like `src_mdl_dir`. Also add an option called "extractor" or "ivector_extractor". Don't name the variable `src_extractor`, because there is going to be only one extractor i.e. no target extractor.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
```diff
@@ -0,0 +1,189 @@
...
+
+ali_dir=exp/tri3b_ali
+treedir=exp/chain/tri4_5n_tree
+lang=data/lang_chain_5n
+
+local/online/run_nnet2_common.sh  --stage $stage \
+                                  --ivector-dim 100 \
+                                  --nnet-affix "$nnet_affix" \
+                                  --mfcc-config $srcdir/conf/mfcc_hires.conf \
+                                  --extractor $srcdir/exp/nnet3/extractor || exit 1;
```
vimalmanohar: Pass the extractor directory input from the user.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
```diff
@@ -0,0 +1,208 @@
...
+train_stage=-4
+get_egs_stage=-10
+
+# training options
+num_epochs=2
+initial_effective_lrate=0.005
+final_effective_lrate=0.0005
+leftmost_questions_truncate=-1
+max_param_change=2.0
+final_layer_normalize_target=0.5
```
vimalmanohar: This option is not used in this script. Check other options like leftmost_questions_truncate is not used in _b and _c. You can hard code some of the options directly instead of allowing the used to control is. e.g. frames_per_eg is always going to be 150.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
```diff
@@ -0,0 +1,189 @@
...
+final_layer_normalize_target=0.5
+num_jobs_initial=2
+num_jobs_final=4
+minibatch_size=128
+frames_per_eg=150
+remove_egs=false
+xent_regularize=0.1
+
+# configs for transfer learning
+srcdir=../../wsj/s5   # base dir for source dataset.
```
vimalmanohar: Add comment that this `srcdir` will be used for the mfcc config. Since `src_mdl_dir` and `extractor` are taken separately from the user.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
```diff
@@ -0,0 +1,189 @@
...
+src_tdnn_affix=1d
+common_egs_dir=
+primary_lr_factor=0.25 # The learning-rate factor for transferred layers from source
+                       # model.
+dim=450
+nnet_affix=_online
```
vimalmanohar: Is this default of online right? Should it not be online_wsj?

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1c.sh
```diff
@@ -0,0 +1,221 @@
...
+
+# The iVector-extraction and feature-dumping parts are the same as the standard
+# nnet3 setup, and you can skip them by setting "--stage 8" if you have already
+# run those things.
+
+lang=data/lang_chain_5n_wsj
+lang_src_tgt=data/lang_wsj_rm # This dir is prepared using phones.txt and lexicon from
+                              # source(WSJ) and and wordlist and G.fst from target(RM)
+ali_dir=exp/chain/chain_ali_wsj
+treedir=exp/chain/tri4_5n_tree_wsj
```
vimalmanohar: Is this variable used?

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1c.sh
```diff
@@ -0,0 +1,221 @@
...
+                                  --nnet-affix "$nnet_affix" \
+                                  --mfcc-config $srcdir/conf/mfcc_hires.conf \
+                                  --extractor $srcdir/exp/nnet3/extractor || exit 1;
+src_mdl_dir=`dirname $src_mdl`
+if [ $stage -le 4 ]; then
+  echo "$0: Generate alignment using source chain model."
+  steps/nnet3/align.sh --nj 100 --cmd "$train_cmd" \
+  --online-ivector-dir exp/nnet2${nnet_affix}/ivectors \
+  --extra-left-context-initial 0 --extra-right-context-final 0 \
+  --scale-opts "--transition-scale=1.0 --acoustic-scale=1.0 --self-loop-scale=1." \
```
vimalmanohar: self-loop-scale=1.0
Also in other places.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
pegahgh: the user no longer provide src_mdl_dir and extractor and srcdir is only taken from user.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
pegahgh: I think src_extractor is better, since it is the source extractor and the user should know not to use target extractor.

pegahgh: Most tdnn scripts have tdnn_affix and it is compatible with them.
@danpovey What is your preference? 1) user provides src_mdl src_extractor (old version) 2) user just provides srcdir and src_tdnn_affix (new version)

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
pegahgh: I will remove unused options, but about other options, I kept the ones which are not hardcoded in default chain script for rm!

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1a.sh
danpovey: I prefer all things to be given explicitly and directly, e.g. src_mfcc_config, src_mdl, extractor, without making too many assumptions about the directory structure where things are coming from.  Document the meaning of all of these in the script you are calling (e.g. in the usage message and via comments when you declare the variables).  Check the ivector extractor id if feasible.

egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1b.sh
danpovey: The default chain script for rm may have too many options.  In general I always try to discourage using options in these example scripts if they are only used once.  The chain script for RM must be old.  Remove the leftmost_questions_truncate option entirely, even when calling the build_tree.sh script, it is deprecated.  Things like final_effective_lrate are used once and should just be hardcoded in the call to train.py, unless they are used twice.

pegahgh: I will remove all options used once in the scripts, but I think it is better to keep some training options which are used (max-param-change, *-lr, ...) in all other scripts e.g. swbd/s5c/local/chain/run_tdnn.sh! It gives user idea which options are important for training and needs to be tuned!

danpovey:
Hm.  A valid point but at this point I prefer to remove the options.
Generally people ask on the list and we tell them which things to tune, and
if someone follows the list (c.f. kaldi-asr.org/forums.html, commenting for
others' sake) then they know which should be tuned.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +67,73 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+        output_dim = -1
+        for  field in parts:
+            key_value = field.split("=")
+            if len(key_value) == 2:
+                key = key_value[0]
+                value = key_value[1]
+                if key == "name":           # name=**
+                    layer_name = value
+                elif key == "dim":          # for input-node
+                    input_dim = int(value)
```
danpovey: you should be setting output_dim here, because logically an input node has no input, only an output.
that would enable some of the code below to be simplified.  In fact, I don't see why you need separate input_dim and output_dim variables; you could just have a single 'dim' which is the output dim of the layer, and simplify the code quite a bit.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -208,11 +215,13 @@ def process_args(args):
                 args.deriv_truncate_margin))
 
     if (not os.path.exists(args.dir)
-            or not os.path.exists(args.dir+"/configs")):
-        raise Exception("This scripts expects {0} to exist and have a configs "
+            or (not os.path.exists(args.dir+"/configs") and
+                not os.path.exists(args.input_model))):
+        raise Exception("This scripts expects {0} to exist. Also either of --trainer.input-model "
```
danpovey: scripts -> script

danpovey: also this sentence does not make sense. Maybe "Either the --trainer.input-model option should be supplied or ....>"
and check line length.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -208,11 +215,13 @@ def process_args(args):
...
+                        " as '0.raw' model should exist or {0} should have a configs "
                         "directory which is the output of "
-                        "make_configs.py script".format(
-                        args.dir))
+                        "make_configs.py script.".format(
+                        args.dir, args.input_model))
```
danpovey: I don't see where you are using args.input_model  in the exception text.

egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -426,13 +426,16 @@ def compute_preconditioning_matrix(dir, egs_dir, num_lda_jobs, run_opts,
...
 
 
-def prepare_initial_acoustic_model(dir, run_opts, srand=-1):
-    """ Adds the first layer; this will also add in the lda.mat and
-        presoftmax_prior_scale.vec. It will also prepare the acoustic model
-        with the transition model."""
-
-    common_train_lib.prepare_initial_network(dir, run_opts,
-                                             srand=srand)
+def prepare_initial_acoustic_model(dir, run_opts, srand=-1, input_mdl=None):
```
danpovey: variable name should be input_model, not input_mdl (check other code, you'll see).

egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -426,13 +426,16 @@ def compute_preconditioning_matrix(dir, egs_dir, num_lda_jobs, run_opts,
...
+    """ Adds the first layer; It will also prepare the acoustic model
+        with the transition model.
+        If input_mdl is specified, no initial network preparation(adding
+        first layer) is done on that and this model is prepared instead of
+        '0.raw' acoustice model with the transition model.
```
danpovey: spelling.  maybe you should run a spell-checker because there are too many of these problems.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -272,14 +281,18 @@ def train(args, run_opts):
...
-    config_dir = '{0}/configs'.format(args.dir)
-    var_file = '{0}/vars'.format(config_dir)
+    if args.input_model is None:
+        config_dir = '{0}/configs'.format(args.dir)
+        var_file = '{0}/vars'.format(config_dir)
 
-    variables = common_train_lib.parse_generic_config_vars_file(var_file)
+        variables = common_train_lib.parse_generic_config_vars_file(var_file)
+    else:
+        # if args.input_model specified, the model left and right context
```
danpovey: specified -> is specified.  computed -> is computed.

egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -357,6 +357,37 @@ def parse_generic_config_vars_file(var_file):
     raise Exception('Error while parsing the file {0}'.format(var_file))
 
 
+def parse_input_model(input_model):
```
danpovey: rename to get_input_model_info, because it doesn't parse the model (it's not  a text file).
The documentation is very bad: it doesn't even explain what the function returns.  It says
it "outputs" the left and right context.  I don't see it opening any files to output it.  Think carefully
about the meaning of the words you use.

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, first_token, key_to_value, all_layers):
             raise RuntimeError("Invalid value: name={0}".format(
                 key_to_value['name']))
         for prev_layer in all_layers:
-            if self.name == prev_layer.name:
+            if self.name == prev_layer.name and prev_layer.layer_type is not 'auxiliary':
```
danpovey: line length

egs/wsj/s5/steps/libs/nnet3/xconfig/utils.py
```diff
@@ -14,14 +14,16 @@
 # Given a list of objects of type XconfigLayerBase ('all_layers'),
 # including at least the layers preceding 'current_layer' (and maybe
 # more layers), return the names of layers preceding 'current_layer'
+# other than layers of type 'auxiliary'.
```
danpovey: change to:
`other than layers of type 'auxiliary', which correspond to component-node names
 from an existing model that we are adding layers to, and which can't be used in
 expressions like [-1].`
obviously make sure line length is OK.

egs/wsj/s5/steps/libs/nnet3/xconfig/utils.py
```diff
@@ -85,6 +91,10 @@ def get_string_from_layer_name(all_layers, current_layer, full_layer_name):
     for layer in all_layers:
```
danpovey: Note that this comment supersedes all previous and following comments and will require changes
to the comment text that I have provided.
Your use of the term 'Auxiliary' as in AuxiliaryLayer and so on, conflicts with the use of the term
'auxiliary' as in 'auxiliary_output', and this could lead to a lot of confusion as the same word is
used in two very different senses.   I think it would be better to change all instances of
'auxiliary' (as used in this new sense) to 'existing'.  Of course names in code would have to
change as well as in comments.
If some sentences no longer make sense, you may have to reword them: don't just blindly
search and replace.

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
```diff
@@ -1075,6 +1075,51 @@ def get_full_config(self):
         return ans
 
 
+class XconfigAuxiliaryLayer(XconfigLayerBase):
+    """This class is for lines like
+    'auxiliary name=aux dim=40'
```
pegahgh: no, it is not presented in config line. I mentioned that in get_full_config function in this class.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,97 @@
...
+lm_opts='num_extra_lm_state=2000'
+#end configuration section.
+
+
+[ -f ./path.sh ] && . ./path.sh
+. parse_options.sh || exit 1;
+
+if [ $# -lt 2 ]; then
+  echo "Usage: $0 [options] <ali-dir1> [<ali-dir2> ...] <out-dir>";
+  echo "e.g.: $0 exp/tri1_ali exp/tri2_ali exp/chain/tdnn_1a_sp";
```
pegahgh: it is explained in local/chain/tuning/run_tdnn_wsj_rm_1c.sh.

pegahgh:
@danpovey thanks for comment. I will fix them now.

danpovey:
It is not sufficient for the interface of anything to be documented in
calling code.  It needs to be documented in the actual place where it is
declared/defined.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,97 @@
...
+cp ${ali_dirs[0]}/tree $dir/ || exit 1
+
+if [ $stage -le 1 ]; then
+  for n in `seq 0 $[num_alignments-1]`; do
+    w=$(echo $weights | cut -d, -f$[$n+1])
+    if ! [[ $w =~ ^[+]?[0-9]+$ ]] ; then
+      echo "no positive int weight specified for alignment ${ali_dirs[$n]}" && exit 1;
+    fi
+  done
+  $cmd $dir/log/make_phone_lm_fst.log \
```
vimalmanohar: This will exceed the line length since there are num-jobs x num-repeats alignments, which is why all this escaping was done in the first place. The original solution of modifying the code is probably the best.

danpovey:
you could do
cat filename  | while read x; do ....; done

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
pegahgh: it is documented in line 30.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,97 @@
...
+  done
+  utils/lang/check_phones_compatible.sh ${ali_dirs[0]}/phones.txt \
+    ${ali_dirs[$n]}/phones.txt || exit 1;
+done
+
+cp ${ali_dirs[0]}/tree $dir/ || exit 1
+
+if [ $stage -le 1 ]; then
+  for n in `seq 0 $[num_alignments-1]`; do
+    w=$(echo $weights | cut -d, -f$[$n+1])
```
pegahgh: I think it is better to keep it optional. I will change the code.

egs/wsj/s5/steps/libs/nnet3/train/common.py
pegahgh: I will change it, also it is good if definition for 'parse_generic_config_vars_file' function is added.

egs/rm/s5/local/prepare_wsj_rm_lang.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# Copyright 2017 Pegah Ghahremani
+
+# This script prepares a dictionary for wsj-rm experiment using wsj phone set, lexicon and dict and
+# rm's words.txt are copied from wsj lexicon for common words in wsj
+# and rm. words in rm that are not available in the wsj lexicon are added
+# as oov in lexicon.txt.
+# The oov word "<SPOKEN_NOISE>" in wsj is also added to words.txt and G.fst is recompiled using
```
pegahgh: This is for training.

egs/wsj/s5/steps/libs/nnet3/xconfig/parser.py
```diff
@@ -67,11 +68,77 @@ def xconfig_line_to_object(config_line, prev_layers = None):
...
+def get_model_component_info(model_filename):
+    """ This function reads existing model (*.raw or *.mdl) and returns array
+        of XconfigExistingLayer one per {input,output}-node or component-node
+        with same 'name' used in the raw model and 'dim' equal to 'output-dim'
+        for component-node and 'dim' for {input,output}-node.
+
+        i.e. layer in *.mdl -> corresponding 'XconfigExistingLayer' layer
+             'input-node name=ivector dim=100' ->
+             'existing name=ivector dim=100'
+             'component-node name=tdnn1.affine ** input-node=1000 '
```
vimalmanohar: input-node -> input-dim and output-node -> output-dim

egs/wsj/s5/steps/libs/nnet3/xconfig/basic_layers.py
```diff
@@ -1075,6 +1087,59 @@ def get_full_config(self):
         return ans
 
 
+class XconfigExistingLayer(XconfigLayerBase):
+    """
+    This class is for lines like 
+    'existing name=tdnn1.affine dim=40'.
```
danpovey: This documentation is not meaningful.  Where I wrote:
```
 This class is for lines like 
     xxxxxx
```
the line xxxxx would  appear in an xconfig file.  But these wouldn't appear in an xconfig file.
I don't want to have to give you the documentation word for word.  Figure it out.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,124 @@
...
+
+[ -f ./path.sh ] && . ./path.sh
+. parse_options.sh || exit 1;
+
+if [ $# -lt 2 ]; then
+  echo "Usage: $0 [options] <ali-dir1> [<ali-dir2> ...] <out-dir>";
+  echo "e.g.: $0 exp/tri1_ali exp/tri2_ali exp/chain/tdnn_1a_sp";
+  echo "Options: "
+  echo " --cmd (run.pl|queue.pl...)      # specify how to run the sub-processes.";
+  echo "--lm-opts                        # options for phone LM generation";
```
danpovey: please document the 'weights' here.  It's important so it should go in the usage message as well as in the comment.

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,124 @@
...
+  utils/lang/check_phones_compatible.sh ${ali_dirs[0]}/phones.txt \
+    ${ali_dirs[$n]}/phones.txt || exit 1;
+done
+
+cp ${ali_dirs[0]}/tree $dir/ || exit 1
+
+if [ -z $weights ]; then
+  # If 'weights' is not specified, comma-separated array '1' with dim
+  #'num_alignments' is defined as 'weights'.
+  for n in `seq 1 $num_alignments`;do weights="$weights,1"; done
```
danpovey: this doesn't look right, there would be a leading comma.
And it's not ignored by cut:
```
echo ,a,b,c | cut -d, -f1

mac:nnet3: echo ,a,b,c | cut -d, -f2
a
```
That means this code path was not tested!

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
```diff
@@ -0,0 +1,124 @@
...
+    w=$(echo $weights | cut -d, -f$[$n+1])
+    adir=${ali_dirs[$n]}
+    num_jobs=$(cat $adir/num_jobs)
+    if ! [[ $w =~ ^[+]?[0-9]+$ ]] ; then
+      echo "no positive int weight specified for alignment ${ali_dirs[$n]}" && exit 1;
+    fi
+    rm $adir/alignment_files.txt 2>/dev/null || true
+    for x in `seq $w`;do
+      for j in `seq $num_jobs`;do
+        echo $adir/ali.$j.gz >> $adir/alignment_files.txt
```
danpovey: You are making this more complicated than it needs to be by writing to separate files.  Write them all to the same file, $dir/alignment_files.txt.  Then you don't need a loop in the $cmd script.

src/nnet3/nnet-utils.cc
```diff
@@ -654,6 +654,32 @@ void ReadEditConfig(std::istream &edit_config_is, Nnet *nnet) {
...
+        if (NameMatchesPattern(nnet->GetComponentName(c).c_str(),
+            name_pattern.c_str()) &&
+            (component =
+            dynamic_cast<UpdatableComponent*>(nnet->GetComponent(c)))) {
+          component->SetLearningRateFactor(learning_rate_factor);
+          num_learning_rate_factors_set++;
+        }
+      }
+      KALDI_LOG << "Set learning rate factors for " << num_learning_rate_factors_set
+                << " nodes.";
```
danpovey: change 'nodes' to 'components' since the learning rate factor applies to components, not nodes.

src/nnet3/nnet-utils.h
```diff
@@ -283,6 +283,9 @@ void CollapseModel(const CollapseModelConfig &config,
        Note: this sets the 'underlying' learning rate, i.e. it will get
        multiplied by any 'learning-rate-factor' set in the nodes.
```
danpovey: nodes -> components

src/nnet3/nnet-utils.h
```diff
@@ -283,6 +283,9 @@ void CollapseModel(const CollapseModelConfig &config,
...
 
+    set-learning-rate-factor [name=<name-pattern>] learning-rate-factor=<learning-rate-factor>
+       Sets the learning rate factor for any updatable nodes matching the name pattern.
```
danpovey: nodes -> components

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
pegahgh: I need loop, since we need different final.mdl for "ali-to-phones \$adir/final.mdl"

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
vimalmanohar: @pegahgh Since you are already creating w_arr below, you can use w_arr = (); and then for num_alignment times do w_arr += (1.0). Then in line 90, you can use w=${w[$n]}

egs/wsj/s5/steps/nnet3/chain/make_weighted_den_fst.sh
vimalmanohar: I think the complexity of this command is underestimated, and this is being made more difficult that it previously was.

@pegahgh it is better to write the alignments in $dir/alignment.$n.txt instead of inside $adir, since you may not have write access to that. Also there would be race-conditions if multiple experiments are run using the same alignment directories.

pegahgh: I can also write model in the file in a same line as alignment is written. Then we can have single file.

vimalmanohar: There will still be one loop inside the command. There needs to be more escaping to split each line into model and alignment, which might again make it more complicated.

danpovey:
Oh, OK as-is then.

src/nnet3/nnet-utils.h
pegahgh: I agree! Thanks, also probably I should change this issue for set-learning-rate. Do you agree?

danpovey:
yes.

PR #1637, opened 2017-05-22
Fixed steps/cleanup/make_biased_lm_graphs.sh to actually add the top-n-words

Gastron:
Added --top-words to the lm_opts.
The top_words.int file also had the probabilities and word integers in wrong columns for internal/make_one_biased_lm.py


danpovey:
Thanks for this PR.   Before merging it I'd like to understand how it affects the behavior of the script; it looks like a serious bug so I'm wondering how the script ran before?  And whether this affects the result of the cleanup recipes, e.g. clean_and_segment_data.sh or segment_long_utterances.sh?
Was this a bug that was introduced when segment_long_utterances.sh was added?  @vimalmanohar , perhaps you can comment?

vimalmanohar:
This seems to have been around before. The top words was never passed to the LM creation script.

danpovey:
OK, merging.  Thanks.

Issue #1640, opened 2017-05-23
Get a topsorted latice?


psmit:
I'm creating some rescoring scripts of my own and it would be very convenient if I can make lattices topologically sorted. It does not seem this is possible with the normal fst tools as I would need to define a custom weight to read the special lattice weights.

Is there any way to easily TopSort a lattice? If not, would you accept a patch that either introduces `lattice-topsort` or a flag to `lattice-copy` that TopSorts the lattices?

jtrmal:
IMO lattice-topsort is a better option than adding flag.
y.

vdp:
@psmit: BTW, why not use OpenFst's weight registration mechanism? If I'm not mistaken OpenFst's command line tools are build on top of their "script" infrastructure, so you should be able to use them with (compact) lattice weights. See [this comment](https://sourceforge.net/p/kaldi/mailman/message/32968915/) by Paul Dixon. He has a blog post about this too.

psmit:
Thanks for the tip @vdp ! I'll certainly look to that as well to see if it would suit my purposes

PR #1641, opened 2017-05-24
[src] Add lattice-topsort binary to Topologically sort lattices

psmit:
See #1640

Based the suggestion by @jtramal I made a separate binary that does topsorting of lattices.


danpovey:
Thanks, but I think I'd rather leave the pull request up and unmerged for
now.
I don't generally like adding programs that are not used anywhere in the
scripts, as it bulks up the repo and adds to compile time.

PR #1645, opened 2017-05-26
fixes for nnet3

osadj:
fixed several bugs and compatibility issues for libs/nnet3/report/log_parse.py, libs/nnet3/train/common.py, and nnet3/train_raw_dnn.py


egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
```diff
@@ -310,7 +310,7 @@ def parse_progress_logs_for_param_diff(exp_dir, pattern):
 def parse_train_logs(exp_dir):
     train_log_files = "%s/log/train.*.log" % (exp_dir)
     train_log_lines = common_lib.run_kaldi_command(
-        'grep -e Accounting {0}'.format(train_log_files))[0]
+        'grep -e Accounting {0}'.format(train_log_files))[0].decode()
```
danpovey: what problem did this solve?

osadj:
In Python3, run_kaldi_command returns a bytes object (as opposed to str).
This will convert that into a str that can be properly consumed by the next
step (which is regex I believe).

danpovey:
Those scripts are not supposed to be fully compatible with python3.  In
python2, calling decode() on a regular string returns a unicode string, so
i'm concerned this may produce its own unintended consequences.

If that is to be changed, it should be done in a better way, by someone
with a full understanding of this stuff.   For now just revert that change,
please.

osadj:
I just successfully tested this with Python 2.7 as well. It works just fine.

If you are really concerned about backward compatibility, I can check the
output of kaldi command, and if it is of type str, we don't decode it. How
does that sound to you?

danpovey:
That would be OK, but it should be done inside run_kaldi_command, not in
the calling code.
However I don't want you to do it on the master branch because it will
cause a merge conflict, the command is 'get_command_stdout' in kaldi_52.
So please make that a separate issue, you can make a PR to kaldi_52.

osadj:
Sure. I just reverted back those changes to master. As you suggested, I
will commit these to kaldi_52.

osadj:
I can't seem to find 'get_command_stdout' in kaldi_52. Can you please point
me to this?

danpovey:
*find ../../egs/wsj/s5/steps/libs -name '*.py' -exec grep 'def
get_command_stdout' {} \; -print*

def get_command_stdout(command, require_zero_status = True):

../../egs/wsj/s5/steps/libs/common.py

osadj:
oops! I was still looking for it in kaldi.

I just made a PR to kaldi_52.

PR #1646, opened 2017-05-26
decode the output of get_command_stdout if not str

osadj:


danpovey:
This did not work, it crashed the scripts.
```
LOG (nnet3-init[5.1]:main():nnet3-init.cc:80) Initialized raw neural net and wrote it to exp/cnn_aug_1b_cifar100/configs//ref.raw
nnet3-info exp/cnn_aug_1b_cifar100/configs//ref.raw
  File "steps/nnet3/xconfig_to_configs.py", line 306, in <module>
    main()
  File "steps/nnet3/xconfig_to_configs.py", line 301, in main
    check_model_contexts(args.config_dir, args.nnet_edits)
  File "steps/nnet3/xconfig_to_configs.py", line 265, in check_model_contexts
    .format(model))
  File "steps/libs/common.py", line 117, in get_command_stdout
    return stdout if type(stdout) is str else stdout.decode()
UnboundLocalError: local variable 'str' referenced before assignment
```

I changed it to:
```
return stdout if type(stdout) is type('str') else stdout.decode()
```
which I now realize is not very clear because any string would have been OK, not just 'str'.

osadj:
Thanks for letting me know. That is strange because:

Python 3.5.2 (default, Nov 17 2016, 17:05:23)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> type('hello') is str
True

Python 2.7.12 (default, Nov 19 2016, 06:48:10)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> type('hello') is str
True

How would you like to make modify this to make it more clear?

Regards,
Omid

danpovey:
There is a local variable in that function called 'str'.
Maybe that could be renamed to something like 'output' and you could change
it back to "is str".

jtrmal:
there is python's package types that exports
types.StringType
types.UnicodeType
which I think might be more clear to use
y.

osadj:
I like this. Thanks, Yenda.

osadj:
I take that back. This is no longer available (and recommended) in Python3.

jtrmal:
In that case, I think it's fine to ignore python3 as the developers are
idiots
y.

danpovey:
It's OK, the solution he had originally was fine ("is str"), it works in
both; the only issue was there was a local variable named str which should
be renamed.

osadj:
Sounds good. Should I make a PR then?

danpovey:
yes please

PR #1652, opened 2017-05-27
WIP: Fixes for Threading Bugs

dogancan:
@danpovey I noticed that one of the sgmm2 tests (estimate-am-sgmm2-test) started failing after merging #1350. I found an initialization bug in the new sgmm2 threading code but that did not fix the failing test. I am not sure whether the test failure is caused by a bug in the new threading code or if the new code revealed an already existing problem. I am not familiar with the sgmm2 code, so I can not really tell what the problem is. Could you take a look at it?

### DIFF
--- a/src/sgmm2/estimate-am-sgmm2.h
+++ b/src/sgmm2/estimate-am-sgmm2.h
@@ -442,8 +442,9 @@ class UpdateWClass: public MultiThreadable {
       MultiThreadable(other),
       accs_(other.accs_), model_(other.model_), w_(other.w_),
       log_a_(other.log_a_), F_i_ptr_(other.F_i_ptr_), g_i_ptr_(other.g_i_ptr_),
-      F_i_(other.F_i_), g_i_(other.g_i_), tot_like_ptr_(other.tot_like_ptr_),
-      tot_like_(0.0) { }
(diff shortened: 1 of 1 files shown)


danpovey:
I just pushed a fix.
It seems that the arguments of the constructor of thread are copied by
value, which is not what this setup assumes (it may cause the destructors
to be called simultaneously, and since they do unsynchronized stuff, this
can cause problems).
Per
http://en.cppreference.com/w/cpp/thread/thread/thread
I used std::ref for that.
It seems to fix the problem.
The change in your PR fixes an efficiency issue, not really a bug, but I
can merge it anyway once the checks finish.

dogancan:
Hmm, makes sense. Thanks for the fix :) Indeed this other change does not affect anything as far as this test is concerned. I guess whether we should merge it depends on what behavior we want when copying these objects. As it is the matrices F_i and g_i are copied from the source object. With the change in this PR, they are initialized as zero matrices. From what I can tell F_i and g_i matrices are not set to zero elsewhere in the code when copying these objects.

danpovey:
Actually, I think I'll close this PR to keep the history cleaner.  Either way is fine, and the efficiency issue is tiny.  The matrix is always zero at the time we copy it.

PR #1627, opened 2017-05-17
[egs,scripts] Add, and use the --proportional-shrink option (approximates l2 regularization).

danpovey:

PR #1654, opened 2017-05-29
adding proportional-shrink scripts to AMI-SDM

GaofengCheng:


GaofengCheng:
Related to #1627

danpovey:
Thanks!  These are very nice improvements.  If you ever get a chance to do something similar in the Switchboard s5c recipe or the tedlium s5_r2 recipe, that would be great too.  Not urgent.

egs/ami/s5b/local/chain/tuning/run_tdnn_1f.sh
```diff
@@ -0,0 +1,266 @@
+#!/bin/bash
+
+# same as 1e but uses batchnorm components instead of renorm also adding
+# proportional-shrink 10, trained with 4 epochs
+
+# System               tdnn1e_sp_bi_ihmali tdnn1f_sp_bi_ihmali
```
danpovey: actually I just realized that you don't specify what microphone these results are with.  I assume sdm, but you should probably state that, and say with what command line options this script was run.

GaofengCheng: @danpovey Oh..... after the SWBD scripts are OK, I'll adding the command lines in these scripts together with the SWBD scripts

PR #1663, opened 2017-06-01
fix Minimize() call

jtrmal:
I'm not sure about the is_partial semantics but this way it makes more sense to me (we won't lose potential hits in case the determinization fails). But I can remove it.


jtrmal:
let me close that for now, I have to find out if the option allow_partial is actually needed at all

jtrmal:
For now, let's just fix the minimize call 
the allow_partial option is true by default so perhaps we should think about if we want to keep it as it is or do something else

PR #1664, opened 2017-06-03
Remove deriv weights

hhadian:
Should I also delete it from all tuning scripts (i.e. `local/chain/tuning`)? [I have deleted it from the scripts in `local/chain`]
Note: I am running an experiment to make sure everything is alright. I will leave a comment here after it is successfully finished.


egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -81,11 +81,6 @@ def get_args():
     parser.add_argument("--chain.leaky-hmm-coefficient", type=float,
                         dest='leaky_hmm_coefficient', default=0.00001,
                         help="")
-    parser.add_argument("--chain.apply-deriv-weights", type=str,
```
danpovey: You'll have to leave this option present and just say that it's deprecated in the usage message 
and do nothing with it.    Otherwise, if people have locally modified scripts with this option, they
will crash.

src/nnet3/nnet-chain-example.cc
```diff
@@ -31,30 +31,20 @@ void NnetChainSupervision::Write(std::ostream &os, bool binary) const {
...
 
 void NnetChainSupervision::Read(std::istream &is, bool binary) {
   ExpectToken(is, binary, "<NnetChainSup>");
   ReadToken(is, binary, &name);
   ReadIndexVector(is, binary, &indexes);
   supervision.Read(is, binary);
-  std::string token;
-  ReadToken(is, binary, &token);
-  // in the future this back-compatibility code can be reworked.
-  if (token != "</NnetChainSup>") {
```
danpovey: You'll need to keep the code that reads the deriv-weights, but just do nothing with them.  Otherwise the code will crash when running on previously dumped egs.

hhadian:
> Not necessary to remove the option from older scripts, e.g. the old tuning ones in swbd, but try to remove it from any fairly new and currently-recommended scripts.

Will do.

danpovey:
we decided not to merge this, the feature was needed for something else.

PR #1665, opened 2017-06-03
WIP: adding hub4-ne broadcast spanish recipe

jtrmal:
There are some details I intend to address but should be complete feature-wise


jtrmal:
I'm still re-running it from scratch but seems to be ready for review

danpovey:
So the pitch helped for this setup?

egs/hub4_spanish/s5/local/prepare_unicode_lexicon.py
```diff
@@ -0,0 +1,198 @@
...
+
+# ======= Prepare data/local directory for babel data with unicode tags ======
+# This script creates all files in the data/local directory for babel formats,
+# except for the filtered_lexicon.txt file which is created by the
+# make_lexicon_subset.sh script.
+#
+# This script basically takes the place of the prepare_lexicon.pl script. It
+# creates the following files.
+#
+# 1. lexicon.txt (via local/lexicon/make_unicode_lexicon.py which happens prior
```
danpovey: I don't really understand from this comment what script creates lexicon.txt.

egs/hub4_spanish/s5/local/prepare_unicode_lexicon.py
```diff
@@ -0,0 +1,198 @@
...
+def main():
+    #  --------------- Extract unicode_graphemes from the table --------------
+    if(len(sys.argv[1:]) == 0):
+        print("Usage: local/prepare_unicode_lexicon.txt <lex_table>"
+              " <lex_dir>" )
+        sys.exit(1)
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("table", help="Table containing all information about"
+                        " how to map unicode graphemes to unicode descriptors")
```
danpovey: It would be nice if it were obvious what the format of the table was.

jtrmal:
I didn't evaluate the difference. It did help for a significant number of
babel langs, though.
I didn't see any harm in using it here as well.
y.

egs/hub4_spanish/s5/local/prepare_unicode_lexicon.py
```diff
@@ -0,0 +1,198 @@
...
+    parser.add_argument("lex_dir", help="Directory to which all files"
+                        " should be written")
```
danpovey: This gives the impression that lex_dir is omly an input, but aren't you getting some information from there too?

egs/hub4_spanish/s5/local/prepare_unicode_lexicon.py
```diff
@@ -0,0 +1,198 @@
...
+    parser.add_argument("--nonspeech", help="File with map of nonspeech words",
```
danpovey: please clarify what the format of the file is.  e.g. map from what to what?

egs/hub4_spanish/s5/local/reestimate_langp.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# Copyright (c) 2015, Johns Hopkins University ( Yenda Trmal <jtrmal@gmail.com> )
+# License: Apache 2.0
+
+# Begin configuration section.
+cmd=run.pl
```
danpovey: this seems not to be used.  looks like it relies on train_cmd being set.

egs/hub4_spanish/s5/local/write_kaldi_files.pl
```diff
@@ -0,0 +1,113 @@
...
+use List::Util qw(max);
+
+my $audio_width=1;
+my $speaker_width=1;
+my $time_width=1;
+
+binmode(STDOUT, ":utf8");
+binmode(STDERR, ":utf8");
+
+die "$0: Error: Unsupported number of arguments: " . scalar @ARGV . ", 3 supported "
```
danpovey: would be nicer to have a usage message showing what the argumnts are, with an example.

egs/hub4_spanish/s5/run.sh
```diff
@@ -0,0 +1,316 @@
...
+    --boost-silence $boost_sil --nj $train_nj --cmd "$train_cmd" \
+    data/train data/langp/tri5 exp/tri5 exp/tri5_ali
+
+  local/reestimate_langp.sh --cmd "$train_cmd" --unk "$unk" \
+    data/train data/lang data/local \
+    exp/tri5_ali data/local/dictp/tri5_ali data/local/langp/tri5_ali data/langp/tri5_ali
+fi
+
+
+if [ $stage -le 12 ]; then
```
danpovey: I don't see why it's necessary to include SGMM systems and MMI.  Surely the performance is much worse
than the chain model?  IMO it would be better to remove all these stages and just have the script call the best chain-model script (maybe commented out so people can choose to run it manually after checking the GPU setup).
If you leave it like this, everyone will waste their time running that stuff.

danpovey:
Regarding the pitch-- I would prefer if you test whether it helps and remove it if not (or if the
improvement is tiny).
The issue is that people may want to use this setup in some real application, requiring online
decoding, and pitch increases the latency of that.

jtrmal:
ok, I can remove it, np.
y.

egs/hub4_spanish/s5/run.sh
jtrmal: Yes, I agree -- I just wasn't sure if it will be the case given the fact we have ~30 hrs of data.
But I can remove it or move to a separate script.
y.

egs/hub4_spanish/s5/local/chain/run_ivector_common_pitch.sh
```diff
@@ -0,0 +1,247 @@
...
+set -e -o pipefail
+
+# This script is called from local/nnet3/run_tdnn.sh and local/chain/run_tdnn.sh (and may eventually
+# be called by more scripts).  It contains the common feature preparation and iVector-related parts
+# of the script.  See those scripts for examples of usage.
+
+
+stage=0
+nj=30
+min_seg_len=1.55  # min length in seconds... we do this because chain training
```
danpovey: This script is kind of out of date.  min_seg_len is something we don't use any more, since we use a range of chunk sizes in the chain model training, so losing small segments is less of an issue.
I suspect there are other issues too.  Better to copy and modify the run_ivector_common.sh of a setup that *is* fully up-to-date.
As far as I can see, the only ones that are fully up-to-date in the sense that they use PCA instead of LDA
for the i-vector UBM (@david-ryan-snyder , we need to update more of them...) are mini_librispeech and ami/s5b, of which mini_librispeech is the better starting point because it doesn't have the microphone issues.
Of course you might want to check whether the pitch matters first, because the mini_librispeech setup
doesn't have that.  [and see if we used a smaller num-gauss for the mini_librispeech because it has
so little data.]

Naturally for the chain training script, you should get one that's up-to-date too.  I think the
mini-librispeech one should be up-to-date, but it doesn't have a TDNN+LSTM system which
 is usually the best.  You can look at WSJ for that system.

I don't like adding new setups that are not using all of the current best practices.  Old setups
are grandfathered in, but for new ones, we should use up-to-date scripts.

egs/hub4_spanish/s5/local/arpa2G.sh
```diff
@@ -0,0 +1,115 @@
...
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+# WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+# MERCHANTABLITY OR NON-INFRINGEMENT.
+# See the Apache 2 License for the specific language governing permissions and
+# limitations under the License.
+
+#Simple utility script to convert the gzipped ARPA lm into a G.fst file
```
danpovey: Is it necessary to use the "local" arpa2G script?  Does it do anything that the one in utils/ doesn't do, that we need?

jtrmal:
as far as I can tell, the training script itself should be reasonably
recent as it was taken from swbd couple of weeks back,
I'll fix the ivector script.
y.

egs/hub4_spanish/s5/local/chain/run_tdnn_lstm_pitch.sh
```diff
@@ -0,0 +1,224 @@
...
+    --chain.apply-deriv-weights false \
+    --chain.lm-opts="--num-extra-lm-states=2000" \
+    --egs.dir "$common_egs_dir" \
+    --egs.opts "--frames-overlap-per-eg 0" \
+    --egs.chunk-width 150 \
+    --trainer.num-chunk-per-minibatch 128 \
+    --trainer.frames-per-iter 1500000 \
+    --trainer.num-epochs 4 \
+    --trainer.optimization.num-jobs-initial 2 \
+    --trainer.optimization.num-jobs-final 12 \
```
danpovey: This is an LSTM model so you should be adding the chunk-left-context option, together with left-context-initial 0; and the same in decoding.  (the option names are different in training versus decoding, check other scripts).  Best to have a variable at the top of the script for the chunk_left_context, since it should be the same in training and decoding.  Also check that the learning rates are ones that we use for LSTMs with about this amount of data.  (I think we use a different lrate for LSTMs versus TDNNs; and for less data the lrate is sometimes a bit higher).

egs/hub4_spanish/s5/local/chain/run_tdnn_lstm_pitch.sh
```diff
@@ -0,0 +1,224 @@
...
+                                  --gmm $gmm \
+                                  --num-threads-ubm $num_threads_ubm \
+                                  --nnet3-affix "$nnet3_affix"
+
+
+gmm_dir=exp/$gmm
+ali_dir=exp/${gmm}_ali_${train_set}_sp_comb
+tree_dir=exp/chain${nnet3_affix}/tree_bi${tree_affix}
+lat_dir=exp/chain${nnet3_affix}/${gmm}_${train_set}_sp_comb_lats
+dir=exp/chain${nnet3_affix}/tdnn_flstm${tdnn_affix}_sp_bi
```
danpovey: let's call it just lstm, not flstm, it's more standard.
And you can remove the _bi suffix, we always build the chain systems that way now so it's redundant.

jtrmal:
ok. I checked the AMI s5b and it does have the min_seg_len so I guess I'll
start with the minilibrispeech.
I am using pca in this script BTW. I'll address the rest of the issues in
 the next couple of days or weeks
y.

danpovey:
Do test whether the pitch makes a difference before removing it, I'm
curious.

danpovey:
Also please tune the proportional-shrink option.  I'd start with 100 and change it in factors of 2 till you find a local optimum, then tune it slightly more precisely (e.g. to within a factor of 1.5 or so).

jtrmal:
without pitch
%WER 14.67 [ 1442 / 9827, 174 ins, 254 del, 1014 sub ]
exp/chain/tdnn_flstm_sp_bi_online/decode_test/wer_10_1.0

vs with pitch
%WER 14.82 [ 1456 / 9827, 196 ins, 244 del, 1016 sub ]
exp/chain_pitch/tdnn_flstm_sp_bi_online/decode_test/wer_10_1.0

egs/hub4_spanish/s5/local/reestimate_langp.sh
jtrmal: addressed

jtrmal:
@danpovey I addressed all the comments (I believe).

PR #1671, opened 2017-06-07
nnet3 model-collapsing code

danpovey:
This PR adds code that modifies an nnet3 model in a way that will improve efficiency in test time,
combining successive layers into one where possible.
@GaofengCheng, would you mind merging this into the branch you are using for your experiments?
I want to make sure there are no important bugs before I merge it-- I don't plan to merge immediately.

### DIFF
--- a/egs/mini_librispeech/s5/local/chain/run_tdnn.sh
+++ b/egs/mini_librispeech/s5/local/chain/run_tdnn.sh
@@ -1,1 +1,1 @@
-tuning/run_tdnn_1b.sh
+tuning/run_tdnn_1c.sh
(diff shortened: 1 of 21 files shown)


vince62s:
does this reduce the graph size a lot ? I mean in way where we could imagine increasing the LM significantly ?

danpovey:
It doesn't reduce the graph size at all, only the nnet3 model size
slightly; but within the past month we committed a change (look for
const-fst/ConstFst) that reduces the memory usage for the graph by a factor
of 2.  So you should already have a factor of 2 to play with.

PR #1673, opened 2017-06-09
[src] Fix bug in looped computation

danpovey:


danpovey:
This fixed the problem reported here
https://groups.google.com/forum/#!topic/kaldi-help/y-PV8ssSwMQ
by Alim Misbullah.  Thanks to @LvHang who reproduced the problem locally.

There are actually two separate changes here either of which is sufficient to fix the problem, but only one of which is the "right" fix.  The change to nnet-optimize-utils.cc make it possible to compile, using looped computation, computations that have descriptors things like ReplaceIndex(ivector, t, 0).  However, we are never supposed to have expressions like that showing up by the time we get to that compilation stage, because the decoder binary is supposed to modify the network to remove such things.  So the "correct" fix is the change to nnet-compile-looped.cc which makes it replace such expressions in a way that is now insensitive to the name of the node-index (doesn't have to be exactly "ivector" now).  I made that fix also.

The changes to nnet-optimize-utils.cc are not, then, strictly necessary, but they make the code slightly more elegant; and who knows, the ability to compile networks containing those kinds of expressions using looped computation may at some point become needed for a legitimate reason.

Issue #1672, opened 2017-06-08
Installation of sequitur and mmseg fails (Prerequisites for GALE Mandarin)


akshayc11:
I am not sure if this is system dependent, but here is my setup:
I have a custom pythonpath where I have installed virtualenv, from which I have created a separate virtual environment.

When I try to run 
```
cd $KALDI_ROOT/tools
extras/install_sequitur.sh
```

I noticed that the installation script expects the setup.py to install the libraries at <sequitur_dir>/lib
However, the system installs it in <sequitur_dir>/lib64
A similar problem arose in the installation of mmseg 
```
extras/install_mmseg.sh
```

Changing the installation lines that refer to lib/ to lib64/ fixed that issue for me. 

I have attached a patch that I have created for that here. If you think this should be done, please feel free to apply the patch and commit it.

[fix_extra_tool_install.patch.txt](https://github.com/kaldi-asr/kaldi/files/1062037/fix_extra_tool_install.patch.txt)

Akshay

jtrmal:
this is definitely system dependent. The patch itself is not
valid for other systems.
what system do you use?
y.

danpovey:
The sequitur installation commands in install_sequitur.py explicitly select
the current directory as the installation target:

cd sequitur-g2p
make CXX=g++ CC=gcc
python setup.py install --prefix `pwd`

cd ../

So I don't understand why it would install in lib64.
Dan

akshayc11:
When you specify the prefix, I believe the setup.py creates a lib (in my case lib64) folder, into which it installs the python-specific site-packages. I believe the issue may be arising due to something akin to this: https://github.com/Linuxbrew/legacy-linuxbrew/issues/443 

This is being run on a centOS (which is an RHEL fork)

One of the comments in the above issue was:
```
Looking at the Python bug tracker, this appears to be a change that certain Linux vendors make to their system Pythons to support mixed 32/64-bit systems. OpenSuSE does this; Ubuntu trusty (afaict, in my hands) does not. Apparently RHEL does. Linuxbrew or pyenv pythons should not.
```

I guess I am one of the unlucky few facing this issue.

danpovey:
Hm.  If some OSs still do this, I think it might be worthwhile fixing it in
some way.  @jtrmal, any chance you could make an 'issue'?  Unless you
disagree.

jtrmal:
I will. I think it should be fairly easy to fix -- will look into it.
y.

[...]

PR #1677, opened 2017-06-10
when installing sequitur and mmseg, look for lib64 as well

jtrmal:
should be addressing #1672

### DIFF
--- a/tools/extras/install_mmseg.sh
+++ b/tools/extras/install_mmseg.sh
@@ -62,21 +62,21 @@ python setup.py install --prefix `pwd`
 ## so that should be pretty reliable) and then we work out the location of
 ## the site-packages directory (typically it would be one level up from
 ## the location of the mmseg.py file but using find seems more reliable
-mmseg_file_lib=$(find ./mmseg-1.3.0/lib/ -type f -name mmseg.py | head -n 1)
-mmseg_file_lib64=$(find ./mmseg-1.3.0/lib64/ -type f -name mmseg.py | head -n 1)
-if [ -f $mmseg_file_lib ]; then
+mmseg_file_lib=$(find ./mmseg-1.3.0/lib/ -type f -name mmseg.py | head -n1)
+mmseg_file_lib64=$(find ./mmseg-1.3.0/lib64/ -type f -name mmseg.py | head -n1)
+if [ ! -z ${mmseg_file_lib+x} ]; then
   lib_dir=./lib/
-elif [ -f $mmseg_file_lib64 ]; then
+elif [ ! -z ${mmseg_file_lib64+x} ]; then
   lib_dir=./lib64/
 else
   echo >&2 "$0: ERROR: Didn't find ./mmseg-1.3.0/lib/ or ./mmseg-1.3.0/lib64/"
-  echo >&2 "  Perhaps your python or system installs python into a different dir?"
-  echo >&2 "  or some other unknown issues arised - review the script output"
-  echo >&2 "  and try to figure out what went wrong."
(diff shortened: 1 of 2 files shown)


tools/extras/install_mmseg.sh
```diff
@@ -39,29 +39,47 @@ if [ -d ./mmseg-1.3.0 ] ; then
...
+)
 
-  [ -f ./env.sh ] && . ./env.sh
+## we first find the mmseg.py file (the module name which will be imported,
+## so that should be pretty reliable) and then we work out the location of
+## the site-packages directory (typically it would be one level up from
+## the location of the mmseg.py file but using find seems more reliable
+mmseg_file_lib=$(find ./mmseg-1.3.0/lib/ -type f -name mmseg.py | head -n 1)
+mmseg_file_lib64=$(find ./mmseg-1.3.0/lib64/ -type f -name mmseg.py | head -n 1)
+if [ -f $mmseg_file_lib ]; then
```
akshayc11: For some reason, this check is failing on my machine: The $mmseg_file_lib variable is empty, but despite that, this condition is being triggered.

How about this:

```
if [ "x${mmseg_file_lib}" != "x" ]; then
  lib_dir=./lib/
elif [ "x${mmseg_file_lib64}" != "x" ]; then
  lib_dir=./lib64
fi
```

The above worked for me, although, as far as I am aware, there seems to be no reason why the one you have proposed should not work

tools/extras/install_mmseg.sh
```diff
@@ -39,29 +39,47 @@ if [ -d ./mmseg-1.3.0 ] ; then
...
+else
+  echo >&2 "$0: ERROR: Didn't find ./mmseg-1.3.0/lib/ or ./mmseg-1.3.0/lib64/"
+  echo >&2 "  Perhaps your python or system installs python into a different dir?"
+  echo >&2 "  or some other unknown issues arised - review the script output"
+  echo >&2 "  and try to figure out what went wrong."
+  exit 1
+fi
 
-  echo "export PYTHONPATH=\$PYTHONPATH:$wd/mmseg-1.3.0/lib/python${pyver}/site-packages"
+site_packages_dir=$(cd ./mmseg-1.3.0; find $libdir -name "site-packages" -type d | head -n1)
```
akshayc11: Should be 
```
+site_packages_dir=$(cd ./mmseg-1.3.0; find $lib_dir -name "site-packages" -type d | head -n1)
```

tools/extras/install_mmseg.sh
akshayc11: For some reason this was failing in my code. I tried the following:
```
if [ ! -z ${mmseg_file_lib+x} ]; then
  lib_dir=./lib/
elif [ ! -z ${mmseg_file_lib64+x} ]; then
  lib_dir=./lib64/
else
...
```
Based on https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

akshayc11:
There was 1 typo, and 1 correction (using -z instead of -f) in the install_mmseg.sh. Please refer to the inline comments for more details.. Otherwise, worked in my machine. Looks good to me.

jtrmal:
Thanks for testing it. Now those issues should be addressed.
Just to be sure: did you test the sequitur install as well?
y.

akshayc11:
I did... Sequitur passed.

danpovey:
ok to merge, by me.

jtrmal:
Great. Thanks a lot.
@danpovey I think ready to merge
Y.

PR #1685, opened 2017-06-14
cnn_tdnn_lstm with online decode

GaofengCheng:
1. cnn_tdnn_lstm_1{a,b} is based on tdnn_lstm_1j ( I based on tdnn_lstm_1i in my experiments)
2. online decode results are added
@danpovey I think this should be OK
@LvHang It will be great if you could have a look at this PR


danpovey:
Very nice improvements!
I'll merge this even though we are still optimizing it, because it will be of interest to others too.

PR #1691, opened 2017-06-16
Fix multilingual setup issues

pegahgh:


egs/babel_multilang/s5/local/nnet3/run_shared_ivector_extractor.sh
```diff
@@ -12,8 +12,8 @@ stage=4
 suffix=_sp
 feat_suffix=_hires # feat_suffix used in train_set for lda_mllt training.
 nnet3_affix=
-numLeavesMLLT=5500
-numGaussMLLT=90000
+numLeavesMLLT=2500
```
danpovey: did you intend to check in this change?

pegahgh:
I retested the setup with use_pitch_for_ivec=false. it is still running (DNN training stage).

egs/babel_multilang/s5/local/nnet3/run_shared_ivector_extractor.sh
```diff
@@ -67,7 +67,7 @@ fi
 
 if [ $stage -le 5 ]; then
   # To train a diagonal UBM we don't need very much data, so use the smallest subset.
-  steps/online/nnet2/train_diag_ubm.sh --cmd "$train_cmd" --nj 200 --num-frames 500000 \
+  steps/online/nnet2/train_diag_ubm.sh --cmd "$train_cmd" --nj 100 --num-frames 200000 \
```
danpovey: and this?

egs/babel_multilang/s5/local/nnet3/run_shared_ivector_extractor.sh
pegahgh: These parameters are defined in lang.conf (for babel languages) and I set the initial values to be equal to values in lang.conf.

egs/babel_multilang/s5/local/nnet3/run_shared_ivector_extractor.sh
pegahgh: I didn't test this, but 500000 frames is very large and it takes long time to train ubm model!(200000 is the number of frames we use in fisher_swbd setup)
I will test it in small setup to see if it makes any difference.

egs/babel_multilang/s5/local/nnet3/run_shared_ivector_extractor.sh
jtrmal: IMO it would be better just not add more options and remove the MLLT training completely (i.e. replace it by PCA, as it's only the transforms we need). I think Dan Snyder has done enough experiments to show the difference is negligible if any at all.

pegahgh: I am not sure, if @david-ryan-snyder  did any test on Babel languages and also we didn't remove MLLT from lots of our scripts yet!

david-ryan-snyder: @pegahgh, @jtrmal 

I never tested the PCA (vs MLLT) transform in any Babel recipes.

Since we want to switchover to PCA completely, for new recipes I think we might as well make it the default (and potentially only) option.

danpovey:
I agree, better to make it non-optional and use only PCA.  We never seem to
see a difference.

danpovey:
But I'd rather get this committed in a working  state before making more
changes.
Pegah, is this working now?

PR #1692, opened 2017-06-17
Code base for Tdnn lstm paper

vijayaditya:


vijayaditya:
Created against the wrong master by mistake.

PR #1693, opened 2017-06-17
Kaldi 5.3 (includes nnet3-based RNNLM setup)

danpovey:
This RNNLM branch will probably exist for a month or two before it's merged, I'm putting it in the official Kaldi repo and others will be making PRs to it as we finalize the setup.  A lot of code already exists but I'm going to start from a clean slate rather than from existing branches so I can review code as it gets added.


nshmyrev:
RNN is going out of fashion it seems https://arxiv.org/pdf/1706.03762.pdf

danpovey:
Ha, well that's for translation.  Let's see who else picks it up.

vince62s:
works also for LM... if interested, implementation here: https://github.com/tensorflow/tensor2tensor/tree/master/tensor2tensor/models

danpovey:
Are there any papers with results for language modeling with that
architecture, and does it seem to do better than conventional topologies?

Anyway the stuff that is going on in the 'rnnlm' branch is all independent
of the specific architecture used, I just call it RNNLM as a generic term
for neural net based language models.

Also Hainan Xu, with Yishay Carmiel at Spoken, is working on ways to
rescore Kaldi lattices with TensorFlow-based language models.  Hopefully
both nnet3-based and TensorFlow-based alternatives will be available in the
end, and we can compare.

src/rnnlm/sampler.cc
```diff
@@ -0,0 +1,494 @@
...
+  }
+}
+
+Sampler::Sampler(const std::vector<BaseFloat> &unigram_probs) {
+  KALDI_ASSERT(!unigram_probs.empty());
+  double total = std::accumulate(unigram_probs.begin(),
+                                 unigram_probs.end(),
+                                 0.0);
+  KALDI_ASSERT(std::abs(total - 1.0) < 1.0e-02);
+  double inv_total = total / 1.0;
```
wantee: Should this be '1.0 / total' ?

danpovey:
thanks, will fix.

danpovey:
@kangshiyin and @hhadian, if you look in the cudamatrix directory in header files, there are a bunch of TODOs; if @kangshiyin could take on the SparseMatrix-related ones and @hhadian the other ones, it would be great.

@kangshiyin, for the sparse-matrix stuff it's not just a question of implementing the functions-- it would require redesigning the internals of the class and re-implementing the existing functions too.  If you think you won't have time for this in the next, say, 3 weeks, let me know and I'll see if someone else is willing to do it.

Obviously I'd like test code for all of these things.

hhadian:
Will do

kangshiyin:
Will do.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -85,11 +85,37 @@ class CuSparseMatrix {
...
+  ///  CUDA kernels for this, to just do it by sparse-matrix multiplication.
+  ///  In that case it could be done by implementing a function as follows:
+  ///   // Set *this to alpha * op(A) * op(B), e.g. A * B if transA = kNoTrans
+  ///   // and transB = kNoTrans.
+  ///   SetSmatSmat(BaseFloat alpha,
+  ///            const CuSparseMatrix<Real> &A, MatrixTransposeType transA,
+  ///            const CuSparseMatrix<Real> &B, MatrixTransposeType transB);
+  ///    ... this would call cusparseScsrgemm/cusparseDcsrgemm, if using the GPU.
+  void SelectRows(const CuArray<int32> &row_indexes,
+                  const CuSparseMatrix<Real> &smat_other);
```
kangshiyin: What is the typical matrix size and number of nonzeros per row which I should optimize for? I'm choosing from two sets of block/grid settings, one for 1~30 nnz per row, and the other for more than 100 nnz per row.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -114,11 +140,53 @@ class CuSparseMatrix {
...
+  /// 'indexes' is expected to contain elements in the
+  /// range [-1, dim - 1].  Each row 'i' of *this after
+  /// calling the constructor will contain zero elements
+  /// if indexes[i] == -1, otherwise it will contain a single
+  /// element at column-index indexes[i] with value 1.0.
+  ///
+  /// If trans == kTrans, the result will be the transpose
+  /// of the sparse matrix described above.
+  ///  TODO: implement this.
+  CuSparseMatrix(const CuArray<int32> &indexes,
```
kangshiyin: Is it ok to do this on CPU and then copy back to GPU ?

danpovey:
I'd say less than 30 nnz per row.

And for the constructor that takes the indexes-- no, please don't do this
on the CPU.  Since it has a very regular structure, it should be
straightforward to write a kernel that will do what you want.

If it makes your life easier, you could modify the comment to disallow
-1's, since for what I need it for, there won't be -1's present.

danpovey:
@hhadian, can you please implement the function RnnlmEmbeddingTrainer::PrintStats()?
For the max-change stuff you can look for inspiration in nnet3::NnetTrainer.  For the parameter-change parts, you'll be comparing `*embedding_mat_` with `initial_embedding_mat_`, and you can look for inspiration in what nnet3-progress does.  No need to convert to vectors, the analog of `VecVec()` for matrices is `TraceMatMat(A, B, kTrans)`.  The exact way the output looks is not critical as long as it's clear.  This function is called just once per job, in the destructor of the object.

hhadian:
Will do.

danpovey:
@hhadian, would you mind implementing the Write and Read functions of RnnlmExample, and also all the functions declared in rnnlm-test-utils.h except for EstimateAndWriteLanguageModel()?

@wantee, do you have time to write the function EstimateAndWriteLanguageModel() declared in rnnlm-test-utils.h?  It's OK to create a class in the .cc file as needed to implement this.  It's for testing purposes; I'd prefer if it were some interpolated Kneser-Ney like thing with a fixed D value (e.g. just use 0.6 or something).  But as long as it writes a correctly normalized language model with the property that the probability you'd get from backoff from any n-gram is always less than probability explicitly computed from that n-gram, it's OK.  "Interpolated" Kneser-Ney, as defined in "A Bit of Progress in Language Modeling", and as implemented by SRILM with the -interpolate option, satisifes this.

wantee:
I may do it in this weekend.

hhadian:
Will do

scripts/rnnlm/initialize_matrix.py
```diff
@@ -0,0 +1,58 @@
...
+import math
+
+parser = argparse.ArgumentParser(description="This script is used to randomly initialize "
+                                 "Kaldi-format matrices, with supplied dimensions and "
+                                 "standard deviation.",
+                                 epilog="E.g. " + sys.argv[0] + " --num-rows=10 --num-cols=20 "
+                                 "--param-stddev=0.05 > exp/rnnlm1/embedding.mat",
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+parser.add_argument("--num-rows", type=int, required=True,
```
danpovey: @hhadian, would you mind translating this program from python into perl?
It turns out that python's random number generation is unacceptably slow and this takes too long.
numpy would be faster but I try to avoid a dependency on numpy.

danpovey: BTW I did a quick test and it looks like perl is only a factor of 3 faster than python for this type of thing, but I think that's enough for this purpose.

hhadian: I wrote the script in perl. If we generate uniform random numbers (in both perl and python), perl is 2.1 times faster. I tried with num-rows=200, num-cols=50000.

danpovey:
@wantee, would you mind adding some kind of script that will make the word features human readable?  i.e. so that after running make_word_features.py, the output, along with the features file could be given to some other script that would print them in a human readable format, e.g. turn
```
  3   0 1.0 3 1
```
into e.g.
```
  3  "constant"=1.0 "special </s>"=1.0
```
and likewise for things like "unigram", "length", "word of", "prefix uh", "match in"
and so on.

danpovey:
@hhadian, can you please look at the changes I made to the CPU version of AddSmatMat in commit 27851a2, and get someone to make similar changes to AddMatSmat, and where applicable, to AddSmat?  Mostly micro-optimizations.

danpovey:
@wantee, since you did such a great job with the other things, I wonder if I can ask you to help with something else?   I think you are the only person other than me who is familiar enough with pocolm to do this.

In order for the sampling code to work, we'll need an ARPA-format language model, and it would be nice if this were well pruned and had been interpolated using the weights obtained by multiplying by weights proportional to the (multiplicity * weighting) from the data_weights.txt file.  I wonder if you could work on some kind of wrapper of pocolm that would do this?

My concern right now is that while it's good for interpolation and pruning, vanilla pocolm is going to be too slow for this purpose, once we go to large data sizes.  So what I'm thinking is that we could decide the n-gram order in advance (say, trigram), compute "reasonable" discounting parameters on some setup for trigram, and just leave them fixed, by supplying them using the --bypass-metaparameters option.  IIRC the discounting parameters are defined in such a way that they are scale-invariant w.r.t. the magnitude of the individual count (i.e. w.r.t. the value of the dataset weights), so I think one set of discounting parameters for everything will do.  Of course the data-source weights will ultimately be derived from data_weights.txt, but you'll have to normalize them so that the largest one is (say) 0.99, as IIRC they have to be strictly between 0 and 1.

There are probably a lot of details involved in this, that I haven't thought of, but I'm feeling a bit exhausted today and don't have the energy to work out the details.  I'm sure you'll do something reasonable.

wantee:
I am not very clear what do we need. I think the pocolm already can do pruning and interpolation.

About the scalability matter, actually, from the first time I knew pocolm, I was thinking that we can write a Hadoop version program to do the whole work. The only thing that would be complicated is the parameter optimization  process. But, as you said, if we can precompute and fixed the discounting parameters, this become a feasible thing to do. The most advantage of Hadoop is obviously the scalability of the distributed framework. And I feel natural to implement pocolm on it, since the various `sort` commands in the pocolm pipeline could be done by MapReduce framework automatically.

danpovey:
I'm not talking about any changes to pocolm itself.  Hadoop is out of the
question as this is just a small preprocessing stage before the main RNNLM
training and we can't take on any big dependencies.

All I'm talking about is writing  a small wrapper script for pocolm, within
the rnnlm tools, that will just call it with --bypass-metaparameters, using
weights derived from the data_weights.txt or whatever we call it, and some
"reasonable" discounting parameters that you can just estimate from some
random corpus.  Discounting parameters estimated on one dataset will
perform quite well enough on other datasets.  We are going to prune
heavily, probably (we might need to do this in order to ensure that the
sampling doesn't use up 100% of the CPU in the pipe), so the exact
perplexity of the pre-pruning LM will not make much difference to the
perplexity of the pruned LM, and the perplexity of the pruned LM will make
very little difference to the training speed of the RNNLM with sampling.

danpovey:
@wantee, actually forget this.  The sampling is a bit too slow, and I'm starting to realize that the pruning procedure used in pocolm is not very optimal for this application; it would make more sense to write a custom program to estimate the LM for this case.  I have to work out some details.

wantee:
OK.

And, yes I knew that the Hadoop is out of question, we are not going to
make big changes for now. It just came out to my mind when you said the
discounting parameters could be fixed. Since In my opinion, the scalability
is always a key feature of a LM tool.

scripts/rnnlm/validate_word_features.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
jtrmal: I did notice this wile looking through that random initialization speedup PR -- do we really want to use python3 in general?

danpovey:
For these new scripts I decided to just go with python3, to avoid future
compatibility headaches.  These days python3 is always installed on any
normal system.  (However, I notice that we're not checking it in
check_dependencies.sh, which we should probably do at some point).

vince62s:
Dan, do you have first results for the ptb training ? (I mean PPL on the test / valid set)

Also I was wondering if you guys tested the tfrnnlm with a weight higher than 0.5 (if so, did 0.5 give you the best result ?)

cheers.

danpovey:
I have it to the stage where I've debugged that the current binaries all
work in the 4 different modes (sampling/not-sampling and
sparse-features/not-sparse-features) and the training is stable (and I've
made sure the sampling is sufficiently fast to not dominate), but I have
not finished any complete training runs or evaluated perplexities with test
data; I have to write that binary, and also I haven't started work on the
official training scripts.  I hope to do that in the next few days.
The perplexities won't be that good initially anyway as this stuff requires
tuning and I'll likely wait till the students concerned are back and have
them do that.

danpovey:
This is mostly a note to nobody in particular.
I had a strange problem with the rnnlm-train binary where I noticed one iteration was very slow, and I logged into the node, and on 'nvidia-smi', the same binary (and the same process-id) was listed as occupying two different GPUs.  The output is below.  That node for some reason had 3 of the 4 the GPUs accidentally set to default mode instead of process-exclusive mode.  The two GPUs were getting 0% use.  Unfortunately I restarted the job without looking at the output, and since then I have not been able to reproduce the issue, even after setting some of all of the GPUs to default mode and running the job manually.

The nnet3-train binary is 'special' because it has two threads that access the GPU.  It runs fine but I cannot be 100% sure that the synchronization is being done correctly; also I suspect I am not using the streams as efficiently as I could be.  @kangshiyin, is there any chance you could  look into the synchronization issues and whether there are compilation options that need to be set for greater efficiency?  The background thread is not doing any serious computation, it's just doing a little memory transfer, and I don't want the main thread to have to wait for those memory transfers to finish before continuing to do whatever it's doing (except, of course, when we start training on the minibatch that the background thread has just prepared; explicit synchronization might be needed at those times, but we know when that is).  The run.sh in the ptb/s5 setup in this branch can be used to run the RNNLM training setup.
```
root@c04:~# nvidia-smi
Sun Aug 27 19:51:58 2017
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 381.22                 Driver Version: 381.22                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GTX 108...  Off  | 0000:02:00.0     Off |                  N/A |
| 23%   32C    P8     9W / 250W |    193MiB / 11172MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  GeForce GTX 108...  Off  | 0000:04:00.0     Off |                  N/A |
| 30%   53C    P2    62W / 250W |    309MiB / 11172MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   2  GeForce GTX 108...  Off  | 0000:83:00.0     Off |                  N/A |
| 40%   71C    P2   180W / 250W |   7035MiB / 11172MiB |     97%      Default |
+-------------------------------+----------------------+----------------------+
|   3  GeForce GTX 108...  Off  | 0000:84:00.0     Off |                  N/A |
| 30%   57C    P2   176W / 250W |   5815MiB / 11172MiB |     53%   E. Process |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID  Type  Process name                               Usage      |
|=============================================================================|
|    0     39572    C   rnnlm-train                                    183MiB |
|    1     39572    C   rnnlm-train                                    299MiB |
|    2     40169    C   nnet3-chain-train                             7025MiB |
|    3     40087    C   nnet3-chain-train                             5805MiB |
+-----------------------------------------------------------------------------+
root@c04:~# kill -KILL 39572
root@c04:~# nvidia-smi -i 0 -c 3
Set compute mode to EXCLUSIVE_PROCESS for GPU 0000:02:00.0.
All done.
root@c04:~# nvidia-smi -i 1 -c 3
Set compute mode to EXCLUSIVE_PROCESS for GPU 0000:04:00.0.
All done.
root@c04:~# nvidia-smi -i 2 -c 3
Set compute mode to EXCLUSIVE_PROCESS for GPU 0000:83:00.0.
All done.   
root@c04:~# nvidia-smi -i 2 -c 3
Compute mode is already set to EXCLUSIVE_PROCESS for GPU 0000:83:00.0.
All done.   
root@c04:~# nvidia-smi -i 3 -c 3
Compute mode is already set to EXCLUSIVE_PROCESS for GPU 0000:84:00.0.
```

kangshiyin:
I'm not sure if I understand your design intention correctly. Are you
trying to overlap the CPU-GPU data copy of the next batch with the GPU
computing of the current batch? Occupying 2 GPUs are not desired?

danpovey:
Yes, that's what I'm trying to do.
I definitely didn't intend to use 2 GPUs.
As I said, I was not even able to replicate the situation where it occupied
2 GPUs.

kangshiyin:
That's probably because each host thread created a separate CUDA context
and selected one GPU to run the code.

Overlapping data transfer with kernel execution typically do not need 2
host threads. It can be done with a single host thread, by assigning all
the operations of a batch, including data transfer and kernel execution, to
a separate CUDA stream. The synchronized data transfer API
cudaMemcpyAsync() should
be used. Currently Kaldi uses cudaMemcpy(), and all CUDA kernels are run in
default stream, which is designed not to run concurrently with any other
CUDA streams. So make the overlapping happen may not be easy. Here is a
link for your reference.

https://devblogs.nvidia.com/parallelforall/how-overlap-data-transfers-cuda-cc/

danpovey:
> That's probably because each host thread created a separate CUDA context
> and selected one GPU to run the code.
>

OK, but this isn't what seems to happen normally.  It seems like normally
just one GPU is used.

Overlapping data transfer with kernel execution typically do not need 2
> host threads. It can be done with a single host thread, by assigning all
> the operations of a batch, including data transfer and kernel execution, to
> a separate CUDA stream. The synchronized data transfer API
> cudaMemcpyAsync() should
> be used. Currently Kaldi uses cudaMemcpy(), and all CUDA kernels are run in
> default stream, which is designed not to run concurrently with any other
> CUDA streams. So make the overlapping happen may not be easy. Here is a
> link for your reference.
>
> https://devblogs.nvidia.com/parallelforall/how-overlap-
> data-transfers-cuda-cc/


OK.  The reason for the separate stream was for other reasons than just the
transfer to GPU; it was doing some stuff on CPU as well.  My intention was
to use the nvcc option --default-stream per-thread, so that the two threads
would have different default streams.  However, there may be some
subtleties involved here, e.g. I think I saw somewhere that even if you
intend to use the default streams, you still have to pass them explicitly
into kernels when you invoke your own kernels.

Can you figure out whether there is a practical way to have the two threads
somehow automatically (via the default threads) do the right thing, and
only synchronize per-thread instead of globally (except when we call an
explicit synchronization function)?



Dan



Read more at:
http://docs.nvidia.com/cuda/cuda-runtime-api/index.html#ixzz4rOp7FPiG
Follow us: @GPUComputing on Twitter | NVIDIA on Facebook

PR #1694, opened 2017-06-17
Replacement for readlink -f

johnjosephmorgan:
Replace the gnu tool redline and option -f with a script.


danpovey:
Thanks for the PR, but I'm a bit confused.
Is that "yaounde" example supposed to be included?
And it seems to me that the readlink changes are a separate issue.
Can you say what the motivation for replacing all the "readlink -f" invocations in Kaldi is?

johnjosephmorgan:
I did not mean to put all the Yaounde stuff in there. I only meant to replace the redline command with a script under utils. The readlink and the -f option are gnu specific. It does not work on MacOSX Sierra.

danpovey:
This does seem to be a genuine issue but I think we should talk about the
right solution first before accepting your fix.  maybe it's better if
@jtrmal does it.

Yenda, I notice that you use readlink -f in a bunch of places, and in other
places other people use readlink -e.  Neither of these will work on mac,
and its version of readlink doesn't seem to have an option to recursively
follow links.

I wonder whether a different name, like utils/make_absolute.sh, would be
more suitable than utils/readlink_f.sh.  If you fingers aren't well enough
healed to do it yourself, let us know along with your opinions on the
topic, and maybe someone else will volunteer to do it.

johnjosephmorgan:
The Yaounde example is not supposed to be included. Only the readlink replacement script should be considered.
The readlink program with the -f option is gnu specific. It does not work on MacOSX.
I was running the online ivector extraction script when I found the problem.
j

jtrmal:
@johnjosephmorgan thanks for taking care of this, I wasn't online those days. 
y.

danpovey:
Yenda, please decide about committing this, and commit it yourself if it's
OK by you.
I believe there are a bunch of 'readlink -e' invocations which we should
also replace; I think it's being used for the same purpose, but don't take
my word for it, please check.

egs/wsj/s5/utils/make_absolute.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
jtrmal: @johnjosephmorgan please note your authorship and license (Apache 2.0, if you agree).

jtrmal:
@johnjosephmorgan , can you please change the `readlink -e` calls as well? Replacing the call by your script seems good enough, as far as I  look through the scripts. Or I can do it myself later, if you busy with something else now.

johnjosephmorgan:
OK, I'll work on it now.

egs/wsj/s5/utils/make_absolute.sh
danpovey: Also, please add a few lines of comment at the top of this file saying what it does: something like:
```
# This script is a replacement for the command readlink -f (which is not portable).
# It turns a pathname into an absolute pathname, including following soft links.
```
Also, please use lowercase variable names in the bash script, for compatibility with the
other bash scripts in Kaldi.

jtrmal:
@johnjosephmorgan thanks, merging

PR #1696, opened 2017-06-19
Enhancing confidences,

vesis84:
Hi, during the MGB evals I have updated the way how confidecnes are computed inside 'lattice-to-ctm-conf'. The idea is to be able to score the 'reference transcripts', to calculate its confidence by using a lattice 'unioned' with the aligned transcripts. To make it work, we need to properly initialize the 'sausage bins' both with the word-sequence as well as begin/end times of bins...

Could you please look at the code? I also added a class `MinimumBayesRiskOptions` and added the support to print the CTM even with the epsiolon bins, some of them have zero duration... But other <eps> bins will have non-zero duration and a low confidence here will indicate that there's a word missing from 'reference transcript' ...

Best, Karel.

### DIFF
--- a/src/bin/copy-vector.cc
+++ b/src/bin/copy-vector.cc
@@ -35,16 +35,19 @@ int main(int argc, char *argv[]) {
         " e.g.: copy-vector --binary=false 1.mat -\n"
         "   copy-vector ark:2.trans ark,t:-\n"
         "see also: dot-weights, append-vector-to-feats\n";
-    
+
     bool binary = true;
     int32 change_dim = -1;
+    float scale = 1.0;
     ParseOptions po(usage);
 
     po.Register("binary", &binary, "Write in binary mode (only "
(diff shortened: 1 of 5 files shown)


vesis84:
Hmm, it says that we run out of time for the Travis test :(
`The job exceeded the maximum time limit for jobs, and has been terminated.`

Is it possible to configure it somehow?

jtrmal:
no, but we can restart that test again.
y.

jtrmal:
(which I did just now).
y.

vesis84:
Perfect! Now it passed ;)

vesis84:
@danpovey Can you please take a look ?

danpovey:
Sorry I didn't get to it yesterday, will try to get to it today.

danpovey:
This looks straightforward-- merging.

Issue #1695, opened 2017-06-18
CUSPARSE library linking etc.


danpovey:
@kangshiyin, there is  something I need to be done in the 'rnnlm' branch.
We'll need the cuSPARSE library.  It will include some changes to the headers and the Makefiles.
Also, there seems to be something called a "library handle" that needs to be initialized, once, at the start of the program, and then supplied to any calls to the cuSPARSE library.  This will require some changes to the CuDevice class to initialize the handle and to make it available on request (e.g. in addition to GetHandle(), add a function GetCusparseHandle()).

For your info, in case you want to get a little bit ahead of this specific request, here is where we are going with the sparse-matrix stuff.
We will adding more functionality to the CuSparseMatrix class.  Currently it can just add to a dense matrix.  We will add functionality including  dense += dense * sparse, and (sparse = transpose-of-sparse), and possibly eventually (sparse = sparse * sparse).  The way CuSparseMatrix stores its data will be changed, and the "backoff case" (when we're not using CUDA) will be changed, by adding a regular SparseMatrix member to be used for that purpose.  We won't be using CuArray any more, or MatrixElement.   CuSparseMatrix will be stored in the CSR format, and will contain 3 pointers, although I'm considering only having the first of them actually be allocated, as a single memory region.  For "dense += sparse", we might just create a temporary vector of integers and convert to COO format.

Incidentally, unlike the regular (dense) matrices, we will not be having a "transposed view" of the sparse-matrices.  That is, the things that CUDA calls the rows, we will call the rows too, as far as sparse matrices are concerned.  [Note: above, when I say we'll support dense += dense * sparse, I might have meant dense += dense * sparse^T.  There is one case that's efficient and that's the case we need to support.]  [Also, note: for sparse = transpose-of-sparse, there is a library function which changes to CSC format, which can be used for that purpose.]

What I have written is just the bottom-line, it doesn't contain the reasoning and the all details.  I hope to get to this in more detail soon, but this might be enough for you to get started.

danpovey:
@kangshiyin, correction to the above...

I just remembered that one of the important operations we support is: trace(dense * sparse).
We'll need to implement in CUDA an operation to gather the relevant elements of a dense matrix (as given by a sparse-matrix structure) into a vector, so we can do a dot-product.  I believe that's how the trace operation is currently done.  cuSPARSE has a "gthr" (gather) operation for vectors, but , it seems, not for matrices.

[...]

PR #1699, opened 2017-06-20
Link cusparse lib and handle it with CuDeivce

kangshiyin:
This is to tackle the first part of #1695 .


danpovey:
Thanks!  Merging.

PR #1701, opened 2017-06-20
Arpa reading code

keli78:
version 1

### DIFF
--- a/src/lm/Makefile
+++ b/src/lm/Makefile
@@ -4,9 +4,9 @@ all:
 
 include ../kaldi.mk
 
-TESTFILES = arpa-file-parser-test arpa-lm-compiler-test
+TESTFILES = arpa-file-parser-test arpa-lm-compiler-test arpa-sampling-test
 
-OBJFILES = arpa-file-parser.o arpa-lm-compiler.o const-arpa-lm.o \
+OBJFILES = arpa-file-parser.o arpa-lm-compiler.o arpa-sampling.o const-arpa-lm.o \
 	   kaldi-rnnlm.o mikolov-rnnlm-lib.o
 
 LIBNAME = kaldi-lm
--- a/src/lm/Makefile
+++ b/src/lm/Makefile
@@ -4,7 +4,7 @@ all:
 
 include ../kaldi.mk
 
-TESTFILES = arpa-file-parser-test arpa-lm-compiler-test 
+TESTFILES = arpa-file-parser-test arpa-lm-compiler-test
 
 OBJFILES = arpa-file-parser.o arpa-lm-compiler.o const-arpa-lm.o \
 	   kaldi-rnnlm.o mikolov-rnnlm-lib.o
--- a/src/lm/Makefile
+++ b/src/lm/Makefile
@@ -4,9 +4,9 @@ all:
 
 include ../kaldi.mk
 
-TESTFILES = arpa-file-parser-test arpa-lm-compiler-test arpa-sampling-test
+TESTFILES = arpa-file-parser-test arpa-lm-compiler-test 
 
-OBJFILES = arpa-file-parser.o arpa-lm-compiler.o arpa-sampling.o const-arpa-lm.o \
+OBJFILES = arpa-file-parser.o arpa-lm-compiler.o const-arpa-lm.o \
 	   kaldi-rnnlm.o mikolov-rnnlm-lib.o
 
 LIBNAME = kaldi-lm
--- a/src/lm/arpa-sampling.cc
+++ b/src/lm/arpa-sampling.cc
@@ -36,7 +36,7 @@ void ArpaSampling::HeaderAvailable() {
   probs_.resize(ngram_order_);
 }
 
-BaseFloat ArpaSampling::GetProb(int32 order, int32 word, const HistType& history) {
+BaseFloat ArpaSampling::GetProb(int32 order, int32 word, const HistType &history) {
   BaseFloat prob = 0.0;
   NgramType::const_iterator it = probs_[order - 1].find(history);
   if (it != probs_[order - 1].end() &&
@@ -58,7 +58,8 @@ BaseFloat ArpaSampling::GetProb(int32 order, int32 word, const HistType& history
   return prob;
 }
 
-BaseFloat ArpaSampling::GetBackoffWeight(int32 order, int32 word, const HistType& history) {
+BaseFloat ArpaSampling::GetBackoffWeight(int32 order, int32 word, 
+    const HistType &history) {
   BaseFloat bow = 0.0;
   KALDI_ASSERT(order >= 1);
   NgramType::const_iterator it = probs_[order - 1].find(history);
@@ -72,11 +73,11 @@ BaseFloat ArpaSampling::GetBackoffWeight(int32 order, int32 word, const HistType
 }
 
 void ArpaSampling::GetUnigramDistribution(std::vector<BaseFloat> *unigram_probs) {
-  for (int32 i = 0; i < num_words_; i++) {
+  for (int32 i = 0; i < num_words_; ++i) {
     HistType h; // empty history
     WordToProbsMap::const_iterator it = probs_[0][h].find(i);
     if (it != probs_[0][h].end()) {
-      (*unigram_probs)[i] = exp(it->second.first);
+      (*unigram_probs)[i] = Exp(it->second.first);
     }    
   }
 }
@@ -88,7 +89,7 @@ void ArpaSampling::ComputeHistoriesWeights(
       it = histories.begin(); it != histories.end(); ++it) {
     HistType history((*it).first);
     KALDI_ASSERT(history.size() <= ngram_order_);
-    for (int32 i = 0; i < history.size() + 1; i++) {
+    for (int32 i = 0; i < history.size() + 1; ++i) {
       HistType h_tmp = history;
       // (*it).second is the input weight of a history 
       BaseFloat prob = 1.0 / histories.size() * ((*it).second);
@@ -96,7 +97,7 @@ void ArpaSampling::ComputeHistoriesWeights(
         HistType::iterator last = h_tmp.end() - 1;
         HistType h(h_tmp.begin(), last);
         int32 word = h_tmp.back();
-        prob *= exp(GetBackoffWeight(h_tmp.size(), word, h));
+        prob *= Exp(GetBackoffWeight(h_tmp.size(), word, h));
         HistType h_up(h_tmp.begin() + 1, h_tmp.end());
         h_tmp = h_up;
       }
@@ -107,12 +108,13 @@ void ArpaSampling::ComputeHistoriesWeights(
   }
 }
 
-BaseFloat ArpaSampling::GetOutputWordsAndAlpha(const std::vector<std::pair<HistType, BaseFloat> > &histories,
-    unordered_map<int32, BaseFloat> *pdf_w) {
+BaseFloat ArpaSampling::GetOutputWordsAndAlpha(const std::vector<std::pair<HistType, 
+    BaseFloat> > &histories, std::unordered_map<int32, BaseFloat> *pdf_w) {
   HistWeightsType hists_weights;
   ComputeHistoriesWeights(histories, &hists_weights); 
   BaseFloat prob = 0;
-  for (HistWeightsType::const_iterator it = hists_weights.begin(); it != hists_weights.end(); ++it) {
+  for (HistWeightsType::const_iterator it = hists_weights.begin(); 
+      it != hists_weights.end(); ++it) {
     HistType h(it->first);
     int32 order = h.size();
     NgramType::const_iterator it_hist = probs_[order].find(h);
@@ -125,14 +127,10 @@ BaseFloat ArpaSampling::GetOutputWordsAndAlpha(const std::vector<std::pair<HistT
           HistType::iterator first = h.begin() + 1;
           HistType h1(h.begin(), last);
           HistType h2(first, h.end());
-          prob = it->second * (exp(probs_[order][h][word].first) - 
-                  exp(GetBackoffWeight(order, h.back(), h1) + GetProb(order, word, h2)));
-          unordered_map<int32, BaseFloat>::iterator map_it = (*pdf_w).find(word);
-          if (map_it != (*pdf_w).end()) {
-            (*pdf_w)[word] += prob;
-          } else {
-            (*pdf_w).insert({word, prob});
-          }
+          prob = it->second * (Exp(probs_[order][h][word].first) - 
+                  Exp(GetBackoffWeight(order, h.back(), h1) + GetProb(order, word, h2)));
+          
+          (*pdf_w)[word] += prob;
         }
       }
     }
@@ -141,29 +139,25 @@ BaseFloat ArpaSampling::GetOutputWordsAndAlpha(const std::vector<std::pair<HistT
   BaseFloat alpha = 0.0;
   BaseFloat total_weights = 0;
   std::vector<std::pair<HistType, BaseFloat> >::const_iterator it = histories.begin();
-  for(; it != histories.end(); it++) {
+  for(; it != histories.end(); ++it) {
     total_weights += (*it).second;
     HistType h = (*it).first;
-    int32 order = h.size();
     BaseFloat backoff_weights = 0.0;
-    while (order > 0) {
+    for (int32 order = h.size(); order > 0; --order) {
       HistType::iterator last = h.end() - 1;
       HistType h1(h.begin(), last);
       BaseFloat backoff_logprob = GetBackoffWeight(order, h.back(), h1);
       backoff_weights += backoff_logprob;
-      order -= 1;
       h = h1;
     }
-    alpha += (*it).second * exp(backoff_weights);
+    alpha += (*it).second * Exp(backoff_weights);
   }
+  // if total input weights equals 0, each input weight is zero. In this case
+  // the only output is the unigram distribution and alpha should be exactly 1
   if (total_weights == 0) {
-    KALDI_ASSERT(alpha == 1.0);
+    alpha = 1.0;
   }
   return alpha;
-}
-
-int32 ArpaSampling::GetNumWords() {
-  return num_words_;
 }
 
 } // end of kaldi
--- a/src/lm/arpa-sampling.cc
+++ b/src/lm/arpa-sampling.cc
@@ -1,0 +1,169 @@
+// arpa-sampling.cc
+
+#include "arpa-sampling.h"
+#include <iostream>
+#include <string>
+#include <iterator>
+#include <algorithm>
+#include <math.h>
+
+namespace kaldi {
+
+// This function reads in each ngram line from an ARPA file
+void ArpaSampling::ConsumeNGram(const NGram& ngram) {
+  int32 cur_order = ngram.words.size();
+  int32 word = ngram.words.back(); // word is the last word in a ngram term 
+  HistType history(ngram.words.begin(), ngram.words.begin() + cur_order - 1);
+  KALDI_ASSERT(history.size() == cur_order - 1);
+
+  // log probability of a ngram term (natural log)
+  BaseFloat log_prob = ngram.logprob;
+  // backoff log probability of a ngram term (natural log)
+  BaseFloat backoff_weight = ngram.backoff;
+  std::pair <BaseFloat, BaseFloat> probs_pair;
+  probs_pair = std::make_pair(log_prob, backoff_weight);
+  probs_[cur_order - 1][history].insert({word, probs_pair});
+ 
+  // get the total number of words from an ARPA file
+  if (cur_order == 1) {
+    num_words_++;
+  }
+}
+
+void ArpaSampling::HeaderAvailable() {
+  ngram_counts_ = NgramCounts();
+  ngram_order_ = NgramCounts().size(); 
+  probs_.resize(ngram_order_);
+}
+
+BaseFloat ArpaSampling::GetProb(int32 order, int32 word, const HistType& history) {
+  BaseFloat prob = 0.0;
+  NgramType::const_iterator it = probs_[order - 1].find(history);
+  if (it != probs_[order - 1].end() &&
+      probs_[order-1][history].find(word) != probs_[order-1][history].end()) {
+    prob += probs_[order-1][history][word].first;
+  } else { // backoff to the previous order
+    order--;
+    if (order >= 1) {
+      HistType::const_iterator first = history.begin() + 1;
+      HistType::const_iterator last = history.end();
+      HistType h(first, last);
+      prob += GetProb(order, word, h);
+      int32 word_new = history.back();
+      HistType::const_iterator last_new = history.end() - 1;
+      HistType h_new(history.begin(), last_new);
+      prob += GetBackoffWeight(order, word_new, h_new);
+    }
+  }
+  return prob;
+}
+
+BaseFloat ArpaSampling::GetBackoffWeight(int32 order, int32 word, const HistType& history) {
+  BaseFloat bow = 0.0;
+  KALDI_ASSERT(order >= 1);
+  NgramType::const_iterator it = probs_[order - 1].find(history);
+  if (it != probs_[order - 1].end()) {
+    WordToProbsMap::const_iterator it2 = probs_[order - 1][history].find(word);
+    if (it2 != probs_[order - 1][history].end()) {
+      bow = it2->second.second;
+    }
+  }
+  return bow;
+}
+
+void ArpaSampling::GetUnigramDistribution(std::vector<BaseFloat> *unigram_probs) {
+  for (int32 i = 0; i < num_words_; i++) {
+    HistType h; // empty history
+    WordToProbsMap::const_iterator it = probs_[0][h].find(i);
+    if (it != probs_[0][h].end()) {
+      (*unigram_probs)[i] = exp(it->second.first);
+    }    
+  }
+}
+ 
+void ArpaSampling::ComputeHistoriesWeights(
+    const std::vector<std::pair<HistType, BaseFloat> > &histories,
+    HistWeightsType *hists_weights) {
+  for (std::vector<std::pair<HistType, BaseFloat> >::const_iterator
+      it = histories.begin(); it != histories.end(); ++it) {
+    HistType history((*it).first);
+    KALDI_ASSERT(history.size() <= ngram_order_);
+    for (int32 i = 0; i < history.size() + 1; i++) {
+      HistType h_tmp = history;
+      // (*it).second is the input weight of a history 
+      BaseFloat prob = 1.0 / histories.size() * ((*it).second);
+      while (h_tmp.size() > (history.size() - i)) {
+        HistType::iterator last = h_tmp.end() - 1;
+        HistType h(h_tmp.begin(), last);
+        int32 word = h_tmp.back();
+        prob *= exp(GetBackoffWeight(h_tmp.size(), word, h));
+        HistType h_up(h_tmp.begin() + 1, h_tmp.end());
+        h_tmp = h_up;
+      }
+      HistType::iterator begin = history.begin() + i;
+      HistType h(begin, history.end());
+      (*hists_weights)[h] += prob;
+    }
+  }
+}
+
+BaseFloat ArpaSampling::GetOutputWordsAndAlpha(const std::vector<std::pair<HistType, BaseFloat> > &histories,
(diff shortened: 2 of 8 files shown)


src/lm/arpa-sampling.cc
```diff
@@ -0,0 +1,169 @@
...
+            (*pdf_w).insert({word, prob});
+          }
+        }
+      }
+    }
+  }
+  // compute alpha
+  BaseFloat alpha = 0.0;
+  BaseFloat total_weights = 0;
+  std::vector<std::pair<HistType, BaseFloat> >::const_iterator it = histories.begin();
```
keli78: so you want me to use auto in the arpa reading code?

src/lm/arpa-sampling.cc
```diff
@@ -0,0 +1,169 @@
...
+    NgramType::const_iterator it_hist = probs_[order].find(h);
+    if (it_hist != probs_[order].end()) {
+      for(WordToProbsMap::const_iterator it_word = probs_[order][h].begin(); 
+          it_word != probs_[order][h].end(); ++it_word) {
+        int32 word = it_word->first;
+        if (order > 0) {
+          HistType::iterator last = h.end() - 1;
+          HistType::iterator first = h.begin() + 1;
+          HistType h1(h.begin(), last);
+          HistType h2(first, h.end());
```
kkm000: Same point about reallocations as below applies here too (yes, I am reading code bottom up! :) )

danpovey:
no.

src/lm/arpa-sampling.cc
```diff
@@ -0,0 +1,169 @@
...
+          prob = it->second * (exp(probs_[order][h][word].first) - 
+                  exp(GetBackoffWeight(order, h.back(), h1) + GetProb(order, word, h2)));
+          unordered_map<int32, BaseFloat>::iterator map_it = (*pdf_w).find(word);
+          if (map_it != (*pdf_w).end()) {
+            (*pdf_w)[word] += prob;
+          } else {
+            (*pdf_w).insert({word, prob});
+          }
```
kkm000: `std::unordered_map::operator[]` [specification says](http://en.cppreference.com/w/cpp/container/unordered_map/operator_at): 
> Returns a reference to the value that is mapped to a key equivalent to *key*, performing an insertion if such key does not already exist [...] When the default allocator is used, this results in the key being copy constructed from *key* and the mapped value being value-initialized.  

So if `word` in not in the map, applying the `[]` inserts a new entry for that word with the value initialized to 0, the  (since the default allocator is used). Therefore, there is no need to call `find` to check whether an element in the map or not; just writing
```
(*pdf_w)[word] += prob;  // 'word' will be added if not found in map.
```
is sufficient. However, a comment on using the explicit but not extremely obvious behavior is helpful, or rather even necessary!

src/lm/arpa-sampling.cc
kkm000: Thanks.

danpovey:
@kkm, thanks for the comments.  Ke, please try to take action on them.
But regarding the test code, I think rather than the "test harness"
approach, it's better that you just figure out a way to test it using its
public interfaces, from external code that doesn't rely on seeing the class
internals.   And remove all the testing code from the class itself.

On Tue, Jun 20, 2017 at 5:32 PM, Kirill Katsnelson <notifications@github.com

keli78:
Sure I'll modify the code based on your and @kkm000 's comments. Thanks!

keli78:
@danpovey @kkm000 I modified the code based on your comments and separated the test code from the ArpaSampling class. One remaining issue is how to properly test the correctness of alpha. While I haven't figured out a proper way to do it.

src/lm/arpa-sampling.cc
```diff
@@ -0,0 +1,163 @@
...
+  }
+  // if total input weights equals 0, each input weight is zero. In this case
+  // the only output is the unigram distribution and alpha should be exactly 1
+  if (total_weights == 0) {
+    alpha = 1.0;
+  }
+  return alpha;
+}
+
+} // end of kaldi
```
kkm000:
```
}  // namespace kaldi
```

src/lm/arpa-sampling.h
```diff
@@ -0,0 +1,129 @@
...
+       bos_symbol_ = "<s>";
+       eos_symbol_ = "</s>";
+       unk_symbol_ = "<unk>";
+  }
+  
+  // This function computes the unigram distribution of all vocab words which are
+  // represented by integers from 1 to num_words (0 is reserved for epsilon)
+  // Probabilities of integers are not represented by any valid word is 0, e.g.
+  // 0 has probability 0.0
+  void GetUnigramDistribution(std::vector<BaseFloat> *unigram_probs);
```
kkm000: I'll look more into it later today or maybe tomorrow, but my immediate concern is the assumptions that which this interface makes about symbols. OpenFST assumes that the *states* are dense, in the sense that if a state s_i exists, then s_j exists for any j<i. AFAIK, no such assumption is ever made about symbol indices. Generally, numbering of symbols allows arbitrary gaps. The interface (both this and the next functions) assumes that symbols are numbered 1, ..., N_1, where N_1 is the number of unigrams in the model, i. e. that the set of unigrams read from the model is just a permutation of the range [1..N_1]. This would not hold if the symbol map (normally our `words.txt`) contains numbering gaps (which is not normally the case but was not prohibited until now). This would also not hold if not all words are present in the LM (which might often be the case!) For example of the latter, if the file `words.txt` defines words A=1, B=2, C=3 and D=4, but only C and D are in the LM, this function is set to return the vector of the length 3, with the probs for symbols 0, 1, and 2, which are all 0.

I think it would be more consistent to make no assumptions about the values of the symbol IDs, and rather return a partial mapping of symbols to the probs (unordered_map<int32, BaseFloat>), or provide a method to return a probability given symbol ID. I do not know about the intended use for this class to make a sensible suggestion here, though.

danpovey:
We already discussed this offline.  Any unigram probs for words that aren't
mentioned in the ARPA file will just be zero.  It's important for the
efficiency of the sampling algorithm that the unigrams be represented as a
dense vector.


On Fri, Jun 23, 2017 at 1:38 AM, Kirill Katsnelson <notifications@github.com

kkm000:
> We already discussed this offline. Any unigram probs for words that aren't mentioned in the ARPA file will just be zero. It's important for the efficiency of the sampling algorithm that the unigrams be represented as a dense vector.

So what are the indices into the returned vector? If they are symbols, then the vector size should be the maximum observed symbol ID, not the count of unigrams in the grammar.

danpovey:
Oh, yes, it should be the the maximum observed symbol ID (plus one).  The
documentation may need changing if it says otherwise.
I think the documentation for num_words_ should say it's the largest
encountered symbol-ID plus one.  (this is normally the same as the
num-words).
Also the function GetUnigramDistribution() should clear the vector and then
resize it to num_words_ + 1, filing with zeros.  Relying on the caller to
initialize things in a certain way is not good practice.

On Fri, Jun 23, 2017 at 1:43 AM, Kirill Katsnelson <notifications@github.com

kkm000:
`num_words_` as implemented is the count of unigrams. This is probably not correct.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // HistWeightsType represents the map between histories (with weights) and
+  // their computed weights
+  typedef std::unordered_map<HistType, BaseFloat, VectorHasher<int32> > HistWeightsType;
+
+  // constructor
+  ArpaSampling(ArpaParseOptions options, fst::SymbolTable* symbols)
+     : ArpaFileParser(options, symbols) { 
+       ngram_order_ = 0;
+       num_words_ = 0;
+       bos_symbol_ = "<s>";
```
danpovey: please remove these 3 special symbols, they aren't needed by this code.

src/rnnlm/arpa-sampling-test.cc
```diff
@@ -0,0 +1,216 @@
...
+    std::istringstream is(line);
+    std::istream_iterator<std::string> begin(is), end;
+    std::vector<std::string> tokens(begin, end);
+    HistType history;
+    int32 word;
+    BaseFloat hist_weight = 0;
+    for (int32 i = 0; i < tokens.size() - 1; ++i) {
+      word = sym->Find(tokens[i]);
+      if (word == fst::SymbolTable::kNoSymbol) {
+        word = sym->Find(arpa_->unk_symbol_);
```
danpovey: You can remove this.  We won't be doing anything special about words not in the vocabulary; they would just be an error.

src/rnnlm/arpa-sampling-test.cc
```diff
@@ -0,0 +1,216 @@
...
+      }
+      history.push_back(word);
+    }
+    HistType h1;
+    if (history.size() >= ngram_order) {
+      HistType h(history.end() - ngram_order + 1, history.end());
+      h1 = h;
+    }
+    #define PARSE_ERR (KALDI_ERR << arpa_->LineReference() << ": ")
```
danpovey: there's no point defining something and using it once.  Just write the code directly.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
+// arpa_sampling.h
+
+// Copyright     2016  Ke Li
```
danpovey: it's 2017.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+
+class ArpaSampling : public ArpaFileParser {
+ public:
+  friend class ArpaSamplingTest;
+
+  // HistType represents a history
+  typedef std::vector<int32> HistType;
+  // WordToProbsMap represents the words and their probabilities given
+  // an arbitrary history
+  typedef std::unordered_map<int32, std::pair<BaseFloat, BaseFloat> > WordToProbsMap;
```
danpovey: Any typdefs which the public interface doesn't need should be protected or private.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // an arbitrary history
+  typedef std::unordered_map<int32, std::pair<BaseFloat, BaseFloat> > WordToProbsMap; 
+  // NgramType represents the map between a history and the map of existing words
+  // and their probabilities given this history
+  typedef std::unordered_map<HistType, WordToProbsMap, VectorHasher<int32> > NgramType;
+  // HistWeightsType represents the map between histories (with weights) and
+  // their computed weights
+  typedef std::unordered_map<HistType, BaseFloat, VectorHasher<int32> > HistWeightsType;
+
+  // constructor
```
danpovey: Please add a comment explaining how you are supposed to read in the arpa file-- I assume it requires using functions from the base-class.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+       ngram_order_ = 0;
+       num_words_ = 0;
+       bos_symbol_ = "<s>";
+       eos_symbol_ = "</s>";
+       unk_symbol_ = "<unk>";
+  }
+  
+  // This function computes the unigram distribution of all vocab words which are
+  // represented by integers from 1 to num_words_ (0 is reserved for epsilon)
+  // Probabilities of integers not represented by any valid word is 0, e.g.
```
danpovey: "is 0"-> "are 0.0".

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // 0 has probability 0.0
+  void GetUnigramDistribution(std::vector<BaseFloat> *unigram_probs);
+
+  // This function reads in a list of histories with input weights and returns
```
danpovey: reads in -> accepts.  ("reads in" implies file I/O.)
input weights -> associated weights
and it doesn't return the non-unigram probs, it outputs to that variable.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  BaseFloat GetBackoffWeight(int32 order, int32 word, const HistType& history);
+
+  // This function returns the computed weights of histories with input weights
+  void ComputeHistoriesWeights(const std::vector<std::pair<HistType, BaseFloat> > 
+      &histories, HistWeightsType *hists_weights);
+
+  // Highest N-gram order of the read-in ARPA LM
+  int32 ngram_order_;
+  
+  // Total number of words of the read-in ARPA LM
```
danpovey: it should be the highest symbol-id present in the read-in ARPA LM, plus one.  and make sure that's what is implemented.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // Begining of sentence symbol
+  std::string bos_symbol_;
+
+  // End of sentence symbol
+  std::string eos_symbol_;
+
+  // Unkown word symbol
+  std::string unk_symbol_;
+
+  // Counts of each LM state
```
danpovey: no, these are the number of n-grams for each n-gram order, indexed by n-gram order minus one.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  friend class ArpaSamplingTest;
+
+  // HistType represents a history
+  typedef std::vector<int32> HistType;
+  // WordToProbsMap represents the words and their probabilities given
+  // an arbitrary history
+  typedef std::unordered_map<int32, std::pair<BaseFloat, BaseFloat> > WordToProbsMap; 
+  // NgramType represents the map between a history and the map of existing words
+  // and their probabilities given this history
+  typedef std::unordered_map<HistType, WordToProbsMap, VectorHasher<int32> > NgramType;
```
danpovey: Rename this to MapType.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // represented by integers from 1 to num_words_ (0 is reserved for epsilon)
+  // Probabilities of integers not represented by any valid word is 0, e.g.
+  // 0 has probability 0.0
+  void GetUnigramDistribution(std::vector<BaseFloat> *unigram_probs);
+
+  // This function reads in a list of histories with input weights and returns
+  // 1) the non_unigram_probs (maps the higher-than-unigram words to their
+  // corresponding probabilities given the list of histories) and 
+  // 2) a scalar unigram_weight = sum of history_weight * backoff_weight of that history
+  BaseFloat GetDistribution(const std::vector<std::pair<HistType, BaseFloat> >
```
danpovey: change the whitespace, so it looks like
```
  BaseFloat GetDistriubution(
      const std::....
```

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+}
+
+BaseFloat ArpaSampling::GetProb(int32 order, int32 word, const HistType &history) {
+  BaseFloat prob = 0.0;
+  NgramType::const_iterator it = probs_[order - 1].find(history);
+  if (it != probs_[order - 1].end() &&
+      probs_[order-1][history].find(word) != probs_[order-1][history].end()) {
+    prob += probs_[order-1][history][word].first;
+  } else { // backoff to the previous order
+    order--;
```
danpovey: Best not to change 'order', instead refer hereafter to 'order - 1'.  Modifying variables
leads to code that's more bug-prone in general.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+      &histories, std::unordered_map<int32, BaseFloat> *non_unigram_probs);
+ 
+ protected:
+  // ArpaFileParser overrides.
+  virtual void HeaderAvailable(); 
+  virtual void ConsumeNGram(const NGram& ngram);
+  virtual void ReadComplete() {}
+
+ private:
+  // This function returns the log probability of a LM state, [history word],
```
danpovey: There is no such thing as logprob of LM state.
It returns the log probability of a word 'word' given a history 'history'.
Please remove 'order' from the interface, as it's redundant with 'history'
(it has to be the same as history.size() + 1).   Make the function const.
Rename it to GetLogprob, as it returns the log-prob.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // from the read-in ARPA file if it exists. If the LM state does not exist,
+  // it backs off to a lower order until the updated LM state found.
+  // Note: a LM state is a ngram term in a ARPA language model
+  BaseFloat GetProb(int32 order, int32 word, const HistType& history);
+
+  // This function returns the back-off log probability of a LM state
+  // from the read-in ARPA file 
+  BaseFloat GetBackoffWeight(int32 order, int32 word, const HistType& history);
```
danpovey: Rename to GetBackoffLogprob.

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+
+  // This function returns the computed weights of histories with input weights
+  void ComputeHistoriesWeights(const std::vector<std::pair<HistType, BaseFloat> >
```
danpovey: it would be more accurate to say that this function changes format from [x] to [y]....
but better to remove it.  If it's easier for your code to consume the map type, we can just have the calling
code provide the map code directly, i.e.  change the interface of GetDistribution().

src/lm/Makefile
```diff
@@ -4,7 +4,7 @@ all:
 
 include ../kaldi.mk
 
-TESTFILES = arpa-file-parser-test arpa-lm-compiler-test
+TESTFILES = arpa-file-parser-test arpa-lm-compiler-test
```
kkm000: Remove EOL whitespace. The change is accidental.

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+
+#include "arpa-sampling.h"
+
+namespace kaldi {
+
+// This function reads in each ngram line from an ARPA file
+void ArpaSampling::ConsumeNGram(const NGram& ngram) {
+  int32 cur_order = ngram.words.size();
+  int32 word = ngram.words.back(); // word is the last word in a ngram term 
+  HistType history(ngram.words.begin(), ngram.words.begin() + cur_order - 1);
```
kkm000: Replace `ngram.words.begin() + cur_order - 1` with `ngram.words.end() - 1`

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+  BaseFloat log_prob = ngram.logprob;
+  // backoff log probability of a ngram term (natural log)
+  BaseFloat backoff_weight = ngram.backoff;
+  std::pair <BaseFloat, BaseFloat> probs_pair;
+  probs_pair = std::make_pair(log_prob, backoff_weight);
+  probs_[cur_order - 1][history].insert({word, probs_pair});
+ 
+  // get the total number of words from an ARPA file
+  if (cur_order == 1) {
+    num_words_++;
```
kkm000: I think you intend to find the maximum symbol index, not the number of unigrams. This should be  
`if (num_words_ < word) num_words_ = word;`

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+  } else { // backoff to the previous order
+    order--;
+    if (order >= 1) {
+      HistType::const_iterator first = history.begin() + 1;
+      HistType::const_iterator last = history.end();
+      HistType h(first, last);
+      prob += GetProb(order, word, h);
+      int32 word_new = history.back();
+      HistType::const_iterator last_new = history.end() - 1;
+      HistType h_new(history.begin(), last_new);
```
kkm000: Every time you construct a vector, the *object* itself is constructed on the stack, so is free in memory sense. The vector object contains  essentially a pointer to data in memory and vector's size. But this *data* is always allocated with the operator new on the heap. Even those small allocations that freed soon lead to memory fragmentation, and are potentially expensive at runtime. `h_new` is the `h` but without the last element, and `h` is not needed any more. Instead, remove the last element of `h` and get rid of `h_new`.  `h.pop_back()` will do the trick.

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+    WordToProbsMap::const_iterator it2 = probs_[order - 1][history].find(word);
+    if (it2 != probs_[order - 1][history].end()) {
+      bow = it2->second.second;
+    }
+  }
+  return bow;
+}
+
+void ArpaSampling::GetUnigramDistribution(std::vector<BaseFloat> *unigram_probs) {
+  (*unigram_probs).clear();
```
kkm000: The idiomatic way to call a function by class pointed is `unigram_probs->clear();` (the next line too).

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+  (*unigram_probs).resize(num_words_ + 1);
+  for (int32 i = 0; i < num_words_ + 1; ++i) {
```
kkm000: The loop may as well start at 1. There is no word 0 in the LM.

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+    HistType h; // empty history
+    WordToProbsMap::const_iterator it = probs_[0][h].find(i);
+    if (it != probs_[0][h].end()) {
+      (*unigram_probs)[i] = Exp(it->second.first);
+    }
```
kkm000: Remove EOL space. Watch for it in general.

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+    if (it != probs_[0][h].end()) {
+      (*unigram_probs)[i] = Exp(it->second.first);
+    }    
+  }
+}
+
+void ArpaSampling::ComputeHistoriesWeights(
+    const std::vector<std::pair<HistType, BaseFloat> > &histories,
+    HistWeightsType *hists_weights) {
+  (*hists_weights).clear();
```
kkm000: ->

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+  for (std::vector<std::pair<HistType, BaseFloat> >::const_iterator
+      it = histories.begin(); it != histories.end(); ++it) {
+    HistType history((*it).first);
```
kkm000: This copies the whole array to a newly allocated memory. And you are not modifying this array! There is no need to keep a copy. This is what const references are for:  
`const HistType& history = it->first;`

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+  pdf_w->clear();
+  HistWeightsType hists_weights;
+  ComputeHistoriesWeights(histories, &hists_weights); 
+  BaseFloat unigram_weight = 0.0;
+  BaseFloat total_weights = 0;
+  BaseFloat probsum = 0;
+  BaseFloat prob = 0;
+  for (HistWeightsType::const_iterator it = hists_weights.begin(); 
+      it != hists_weights.end(); ++it) {
+    HistType h(it->first);
```
kkm000: A const reference here will work for you again, no need to copy the vector.

src/rnnlm/arpa-sampling.cc
```diff
@@ -0,0 +1,145 @@
...
+  }
+  // If total input weights is zero, then each input weight is zero. In this case
+  // the output is the unigram distribution and unigram weight should be 1 
+  if (total_weights == 0) {
+    unigram_weight = 1.0;
+  }
+  return unigram_weight;
+}
+
+}  // end of kaldi
```
kkm000: Our coding style (inherited from the Google's coding style) asks for the *exact* comment string
```
}  // namespace kaldi
```
with exactly 2 spaces before the `//` (as all EOL comments are separated from code with 2 spaces).

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ARPA_SAMPLING_H_
+#define ARPA_SAMPLING_H_
```
kkm000: This file name is `kaldi/rnnlm/arpa-sampling.h`, so the include guard should be named after the directory name, with the `-` and the `/` represented with the `_`, and the `_` at the end:  
`KALDI_RNNLM_ARPA_SAMPLING_H_`

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+#include <vector>
+
+#include <math.h>
+#include <sys/time.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include "fst/fstlib.h"
+#include "util/common-utils.h"
+#include "lm/arpa-file-parser.h"
+
```
kkm000: The coding style asks for this: first you list all system includes sorted alphabetically (directory prefix is considered part of the name), then empty string, then our includes, also sorted alphabetically:
```
#include <algorithm>
#include <iostream>
#include <fstream>
#include <map>
#include <math.h>
#include <sstream>
#include <stdlib.h>
#include <string>
#include <sys/time.h>
#include <unistd.h>
#include <vector>

#include "fst/fstlib.h"
#include "lm/arpa-file-parser.h"
#include "util/common-utils.h"
```
@danpovey: C++11 calls for <cmath> and <cstdlib> in favor of <math.h> and <stdlib.h>. But I am not sure if we won't open a can of worms with it. Probably safer to leave as they are?

src/rnnlm/arpa-sampling.h
```diff
@@ -0,0 +1,131 @@
...
+  // LM state probabilities
+  std::vector<NgramType> probs_;
+
+  // Histories' weights
+  HistWeightsType hists_weights_;
+  
+};
+
+}  // end of namespace kaldi
```
kkm000: Fix comment,

src/rnnlm/arpa-sampling-test.cc
kkm000: But print a KALDI_WARNING please, `"Unknown word in ARPA file: " << "tokens[i];`

keli78:
Sure, I'll address these issues by tomorrow.

src/rnnlm/arpa-sampling-test.cc
danpovey: actually this should be KALDI_ERR (fatal).

src/rnnlm/arpa-sampling.h
keli78: Both types are ok for my code if there are no repeated history in the input history list. I did this conversion because I assumed the history list may have repeated history (then loop over those repeated histories are not efficient). If there are no repeated histories, I can remove this function and we don't need to change the interface of GetDistribution().

danpovey:
Remove the function-- the calling code can take care of repeated histories
if I feel there is an efficiency issue.

src/rnnlm/arpa-sampling.cc
keli78: Good idea. Thanks!

src/rnnlm/arpa-sampling.h
keli78: hm, the google C++ style suggests the order as c system, c++ system, other

danpovey: Actually, sys/time.h and math.h and unistd.h and stdlib.h shouldn't be included at all.  Those kinds of things should be included indirectly via util/common-utils.h so we can do it in a portable way.  And I suspect that most of the other includes aren't really necessary either but I'd be OK to leave them.  
We're not really following google's directive to directly include all things that you use.

keli78:
While all the tests I wrote are passed, I'm still not one hundred percent sure about the correctness of computing the unigram weight.

danpovey:
I'll hopefully look at the code at some point.  Also I am hoping that Jian
Wang (@wantee) will look through it, he knows a lot about language models
so he would be able to spot any problems.

danpovey:
I'm merging this since I am doing some stuff that requires it, but there may still be issues to be addressed.

PR #1704, opened 2017-06-20
[src] Fix to previously added back-compatibility code from PR #1646

kkm000:
(cherry picked from commit c0cd9c543b5fa708213978e59991034019b446eb)

This is in kaldi_52 but not in master. Dan, please accept or close--I do not know if it should be, this is a *graph-theoretical MR* : )


danpovey:
It's OK, it turns out that was not the right fix, we fixed it another way by removing a variable called 'str'.

PR #1706, opened 2017-06-22
Added an boost-unk option to utils/prepare_lang.sh for unknown word modeling

xiaohui-zhang:


egs/wsj/s5/utils/prepare_lang.sh
```diff
@@ -67,6 +67,10 @@ extra_word_disambig_syms=        # if set, add disambiguation symbols from this
 num_extra_phone_disambig_syms=1 # Standard one phone disambiguation symbol is used for optional silence.
                                 # Increasing this number does not harm, but is only useful if you later
                                 # want to introduce this labels to L_disambig.fst
+boost_unk=1.0 # for unknown-word modeling, we provide this option to boost the unknown-word's probability
```
danpovey: please call this unk_factor and rework the explanation.
and instead of 'improper', say, 'too low or too high'.

egs/wsj/s5/utils/prepare_lang.sh
```diff
@@ -273,9 +281,9 @@ fi
 # FST due to the optional-silence [the last phone of any word gets 2 arcs].
 if [ ! -z "$unk_fst" ]; then  # if the --unk-fst option was provided...
   if "$silprob"; then
-    utils/lang/internal/modify_unk_pron.py $tmpdir/lexiconp_silprob.txt "$oov_word" || exit 1
+    utils/lang/internal/modify_unk_pron.py --boost-unk=$boost_unk $tmpdir/lexiconp_silprob.txt "$oov_word" || exit 1
```
danpovey: This will only modify the probability of the unk word in the lexicon if the --unk-fst option is provided.
I don't like that,  because  you might want to modify the unk prob even if unk is just a  normal word.
I'd prefer if you implemented this in a more generic way, not dependent on the unk-fst option being provided.

PR #1712, opened 2017-06-26
neural and i-vector LID recipes for AP17-OLR Challenge

ABHISHEKDEY173:


danpovey:
Thanks for the pull request.  Just from glancing at it, it looks like there are probably scripts in there that were copied from elsewhere but are not actually used.
@david-ryan-snyder, can you please look into what this is, and maybe give a preliminary opinion about whether we should look into merging it?  I haven't heard of this challenge.

david-ryan-snyder:
@ABHISHEKDEY173,

Before we look into this further, could you add some description as to what this is about (e.g., what is being added and where)?

Also, as Dan said, there appear to be lots of files that are added from other recipes but may not actually used in the recipe. This makes it hard to figure out what functionality this pull request is adding. Could you remove the files that are not needed? Also, most recipes have a run.sh which is the entry point in the example. I couldn't find that by searching file names in the pull request. 

Best,
David

ABHISHEKDEY173:
Sorry Sir my mistake I clicked on the pull request button. Nothing is
edited from my side. I apologize for the inconvenience caused.

david-ryan-snyder:
@danpovey sounds like this is an accidental PR! So I think we can close it.

Issue #1107, opened 2016-10-11
maybe set CPPFLAGS="-stdlib=libstdc++" in tools/extras/install_sequitur.sh


jasonrayles-nbcuni:
When installing sequitur on OS X, I think that
`make CPPFLAGS="-stdlib=libstdc++"` is required.

This solution was proposed at https://groups.google.com/forum/#!msg/kaldi-help/x3AJzKHe5lY/EpBJORtFAgAJ. I am not sure what the implications of explicitly setting this option is for all users, but according to http://stackoverflow.com/a/16788372, libstdc++ is the standard GNU/Linux library, anyway. 

If nothing else, perhaps a comment in the install script will could save new users some time.

jtrmal:
Thanks for reporting it, I'll have a look.
MacOS is annoying as the default libc is clang's one (libc++) as far as I understand.  I will investigate, but do not expect a quick fix -- if you have enough technical experience, can you please add that flag just to see what happens?

Also, my feeling is it should go into CXXFLAGS, but perhaps that does not matter

jasonrayles-nbcuni:
` make CPPFLAGS="-stdlib=libstdc++" ` definitely works on OS X. I could not install sequitur without it.

jtrmal:
would you be willing to remove your sequitur and then change the extras/install_sequitur.sh so that the line 67 would say:
```
make CXX=g++ CC=gcc
```
My feeling is that is more correct solution anyway.

jasonrayles-nbcuni:
Ok, I just had a chance to try this. 

` make CXX=g++ CC=gcc ` did not work for me. 

I get
```
[...]

PR #1716, opened 2017-06-27
sequitur: try to work out the correct  compilation flags

jtrmal:
should be resolving #1107


jtrmal:
@danpovey  I didn't head from that guy, but I think this is a safe change w.r.t. linux, sounless you object, please merge. Or we can drop it, as we didn't have any more reports on issues about this.

danpovey:
Can you see why the check failed?  It looks like it's not finding src/base/version.h.  Could it be a problem with the recent changes to the travis setup?

jtrmal:
@kkm000  added some code that prints additional info after a run (and the version is one of those info) but the file won't get created when there was only a change outside src/ (when we were starting with travis, we decided not to compile everything if there was only a change in egs and/or tools)

danpovey:
Hm, OK.  I guess that needs to be fixed somehow?  Because it caused the
build to fail.

jtrmal:
Hopefully will be addressed by #1827 (I'll restart build of this if/when the fix gets merged).

Issue #1707, opened 2017-06-22
RNNLM data-preparation scripts


danpovey:
@wantee, I hope this catches you at a good time.
I have been figuring out some of the details of how the RNNLM setup will work, and I have in this 'issue' some notes about how I want to the frontend (data preparation) to work.  It would be great if you could help with this-- I know from our work on pocolm that you are able to do it.

```
 # The training data for RNNLM training will be formatted a little like for pocolm:
 # there will a directory like
 # e.g. data/rnnlm/train/
 #  with files:
 #              dev.txt foo.txt bar.txt


 # get the word-list and the mapping to integers.  Note: by default this
 # includes the dev word-list.


# there will be a script to validate this data-dir: it should contain things like
# foo.txt, bar.txt, and dev.txt (dev.txt is a special filename that's
# obligatory).
# These files are normal text files (lines like "hello there"),
# but these lines may contain internal </s>'s to demarcate parts of
# a sequence of sentences or utterances.  (this won't always be the
# case, it's something you can do optionally).
# Still, the lines shouldn't be excessively long (warn if there
# are lines longer than 1000 words).
# There is no need to read everything by default for validating the directory,
# you can do spot-checks by default.
# For now let's not support gzipped versions of them-- there is a reason
# for this that will become clear in future.

rnnlm/validate_data_dir.py  data/rnnlm/data


# The script get_unigram_counts.xx (xx could be sh or py),
# creates text files containing the unigram counts of words.
# things like dev.counts, foo.counts, bar.counts,
# with lines like:
# hello  12341
# probably a simple awk script, maybe with sort and uniq and then another
# awk script, is the easiest way to get this.
# note: </s> is included in the counts; it will contain any </s> symbols that
# appear inside the sentences (to separate parts of a sequence, as mentioned
# above), plus one for each end-of-line (these are implicit).
rnnlm/get_unigram_counts.xx data/rnnlm/data

# get_vocab.py gets a word-list/vocabulary, in an OpenFst compatible format,
# with a mapping to integers.  By default this will include all words present
# in all data-sources including dev, plus <eps> as symbol zero, and
# <s>, </s> and <brk>.
# Note: because of our feature-based representation
# of the words and the sampling algorithm, adding an extra word
# does not really impose an extra cost, either in time taken or
# in parameters.
# There is no reason to *not* use this script while training, because
# the models are inherently transferable to new vocabularies.
[...]

PR #1717, opened 2017-06-27
RNNLM data-preparation (#1707)

wantee:
This PR is for Issue #1707.

Testings are done with tedlium s5_r2 setup. Main script is ./local/rnnlm/prepare_rnnlm_data.sh.

Testing data is generated by following commands:
```
dir=data/rnnlm/data/
mkdir -p $dir
for s in dev train test; do
  cut -d ' ' -f 2- data/$s/text | tr 'a-z' 'A-Z' > $dir/$s.txt
done
cat $dir/train.txt $dir/test.txt > $dir/ted.txt
rm $dir/train.txt $dir/test.txt
```
And add some '\</s\>' and Chinese words are added to the first lines of dev.txt manually to test encoding.

There are two questions I still need to confirm:
1) Does the BOW and EOW need to be considered as a word when we computing the ngram-order? More specifically, when user set --max-ngram-order=3, do we produce 'final ing' or only 'final ng', current implementation is the latter one.

2) I use a naive way to loop over all the features with every words in make_word_features.py, this turns out to be slow. Do we need to do some advance trick to speedup it?

### DIFF
--- a/scripts/rnnlm/get_unigram_probs.py
+++ b/scripts/rnnlm/get_unigram_probs.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(description="This script gets the unigram probabilities of words.",
+                                 epilog="E.g. " + sys.argv[0] + " --vocab-file=data/rnnlm/vocab/words.txt "
+                                         "--data-weights-file=exp/rnnlm/data_weights.txt data/rnnlm/data "
+                                         "> exp/rnnlm/unigram_probs.txt",
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+parser.add_argument("--vocab-file", type=str, default='', required=True,
+                    help="Specify the vocab file.")
+parser.add_argument("--data-weights-file", type=str, default='', required=True,
+                    help="Specify the data-weights file.")
+parser.add_argument("--smooth-unigram-counts", type=float, default=1.0,
+                    help="Specify the constant for smoothing. We will add "
+                         "(smooth_unigram_counts * num_words_with_non_zero_counts / vocab_size) "
+                         "to every unigram counts.")
+parser.add_argument("data_dir",
+                    help="Directory in which to look for data")
+
+args = parser.parse_args()
+
+SPECIAL_SYMBOLS = ["<eps>", "<s>", "<brk>"]
+
+# get the name with txt and counts file path for all data sorces
+def GetAllDataSources(data_dir):
+    data_sources = {}
+    for f in os.listdir(data_dir):
+        full_path = data_dir + "/" + f
+        if f == 'dev.txt' or f == 'dev.counts' or os.path.isdir(full_path):
+            continue
+        if f.endswith(".txt"):
+            name = f[0:-4]
+            if name in data_sources:
+                data_sources[name] = (full_path, data_sources[name][1])
+            else:
+                data_sources[name] = (full_path, None)
+        elif f.endswith(".counts"):
+            name = f[0:-7]
+            if name in data_sources:
+                data_sources[name] = (data_sources[name][0], full_path)
+            else:
+                data_sources[name] = (None, full_path)
+        else:
+            sys.exit(sys.argv[0] + ": Text directory should not contain files with suffixes "
+                     "other than .txt or .counts: " + f)
+
+    for name, (txt_file, counts_file) in data_sources.items():
+        if txt_file is None or counts_file is None:
+            sys.exit(sys.argv[0] + ": Missing .txt or .counts file for data source: " + name)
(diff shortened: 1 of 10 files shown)


wantee:
Please ignore the second question I asked. I realize that we can use the similar procedure in choose_features.py for make_word_features.py to match the ngram features, so there would be no speed problems. The code is updated.

danpovey:
@wantee, thanks a lot!!  It looks great, just from scrolling through it.  I asked @keli78 to look into it and try to apply it to the datasets that she is working with for language modeling.   I am making progress on some of the core RNNLM stuff, meanwhile (while also doing other things).  There are other threads that have to complete before we can really test this properly-- mostly things I am doing, but I believe @kangshiyin is going to work on some sparse matrix stuff too.  

Incidentally, for Kaldi stuff we are not using the exact same coding style we were using in pocolm-- @vimalmanohar pointed out that the use of function names LikeThis is not pep8 compliant, they should be like_this.  Also we try to stay within an 80 char limit, but this is not super important and I personally sometimes break that rule.

kangshiyin:
I'm quite busy recently.. My progress is slow until now. I'm now trying to implement the classes SparseMatrix2 and CuSparseMatrix2 which have similar API as the old ones. Currently I only work on the arithmetic operations you mentioned in #1695, but leave the memory ops like Resize() and CopyFrom/ToMat/SMat() untouched. Hopefully when the 2nd version is strong enough, we could then replace the original classes.

wantee:
I rename the function names, and fix the errors output by pep8 except the 'lines too long' error.

danpovey:
@wantee-- thanks a lot.  @kangshiyin, it's OK-- I think I am going to be the limiting factor in terms of the timeline of this project, not you.  BTW, I'm starting some interesting work with attention models, inspired by Google's "Attention is all you need" paper, and once we have the basic version working I may ask you to help optimize it.  But that will probably be a couple of weeks' time.

danpovey:
Commenting on this now-merged PR...
@keli78, since your work found that in the Keras setup, even zero-mean, unit-stddev unigram was harmful, I think we should try to support scaling down the unigram features.  Can you add a --unigram-scale option to make_word_features.py, and have it print something like:

`unigram <entropy> <scale>`
e.g.
`unigram 2.2317 0.1`
and then, unigram features will be scaled also by the second number, so instead of
`(logprob + entropy) / entropy`
it will be
`(logprob + entropy) * scale / entropy`

Obviously you need to also change the code that dumps the features per word.   Please try to run the testing setup that @wantee was using, to validate that your changes work.

scripts/rnnlm/validate_data_dir.py
```diff
@@ -0,0 +1,110 @@
...
+            if line is None:
+                break
+            lineno += 1
+            if args.spot_check == 'true' and lineno > 10:
+                break
+            words = line.split()
+            if len(words) != 0:
+                found_nonempty_line = True
+                for word in words:
+                    if word.lower() in disallowed_symbols:
```
danpovey: Let's not do 'lower()' here; I'd rather allow the upper-case versions of those symbols as real words since in general the whole system will be case-sensitive.

keli78:
I'm running an experiment of zero-mean, 0.1 std dev unigram and it will finish within several hours. Do you think we should wait for that first?

danpovey:
Sure you can wait.

scripts/rnnlm/get_unigram_probs.py
```diff
@@ -0,0 +1,159 @@
...
+                                        "> exp/rnnlm/unigram_probs.txt",
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+parser.add_argument("--vocab-file", type=str, default='', required=True,
+                    help="Specify the vocab file.")
+parser.add_argument("--data-weights-file", type=str, default='', required=True,
+                    help="Specify the data-weights file.")
+parser.add_argument("--smooth-unigram-counts", type=float, default=1.0,
+                    help="Specify the constant for smoothing. We will add "
+                         "(smooth_unigram_counts * num_words_with_non_zero_counts / vocab_size) "
```
keli78: I think we should do add-lambda smoothing (lambda here is the smooth-unigram-counts) on unigram counts instead of scaling them right?

danpovey:
I'm not sure what you mean by add-lambda smoothing.
I think we're talking about adding something to the *counts*, not the
probabilities.  This is a different way of specifying the lambda, probably;
but specified this way, a single value (like 1.0) is suitable in even
pathological cases.

keli78:
oh ok, I think it's equivalent to what I said. Add-lambda smoothing is just a smoothing method by adding a small number (lambda) to each count. 
BTW, the zero-mean 0.1 std dev unigram doesn't improve the performance (ppl 56.4 vs. 56). I'll implement the unigram scaling option soon.

PR #1725, opened 2017-06-28
Fix for threaded nnet2 decoding

alumae:
With STL threads, a thread can only be called `join()` once. However, `SingleUtteranceNnet2DecoderThreaded::WaitForAllThreads()` is called many times: once via `Wait()` from online2-wav-nnet2-latgen-threaded.cc, and the second time in the destructor, which causes a system error when running `online2-wav-nnet2-latgen-threaded` (right before the exit).

With this patch we check whether a thread is joinable before calling `join()`. This fixes the problem mentioned above.


jtrmal:
I might not understand this enough, but I though join() is blocking (http://en.cppreference.com/w/cpp/thread/thread/join), i.e. it will wait until the thread is joinable and then join it and only then proceeds on the next thread. Your solution would be actively polling state of the thread many times per second.

PR #1729, opened 2017-06-29
[src] Make sure softmax-related functions can work in-place.

danpovey:


danpovey:
This is just a little efficiency thing; it's something that came up when preparing some code for attention models.

Issue #1660, opened 2017-05-31
Default dropout has changed?


psmit:
I noticed that a xconfig-based tdnn_blstm recipe (like the one in swdb but for other data) suddenly had a very bad performance in kaldi52 compared to the master branch a month ago. I'm using the same egs and scripts and finally I found differences in the generated `configs/xconfigs.expanded.1` and `configs/xconfigs.expanded.2` . All relu-renorm-layers have gotten default values `dropout-proportion=0.5 learning-rate-factor=1.0 ng-affine-options=`.

I am assuming that `dropout-proportion=0.5` is different from earlier defaults, I guess that would be 
 `dropout-proportion=0.0`. Am I correct in my assumptions? If so, is this change intentional?

Related question, does this relate to the `--trainer.dropout-schedule` option in the `chain/train.py`? If so, how do they interact?

danpovey:
That wouldn't be the reason.  dropout-proportion only has an effect if the
layer is something like relu-dropout-layer.  And dropout-schedule only has
an effect if there are dropout layers present.
Look for other reasons, e.g. see when the training likelihoods start to
differ.

Dan

danpovey:
Also you can share the output of steps/info/chain_dir_info.pl for both
dirs, I might see something.

psmit:
exp/chain/model/gr_tdnn_blstm_9_a: num-iters=2150 nj=4..4 num-params=48.3M dim=40+100->3426 xent:train/valid[1431,2149]=(-0.655,-0.595/-0.716,-0.718) logprob:train/valid[1431,2149]=(-0.046,-0.037/-0.064,-0.066)
    exp/chain/model/gr_tdnn_blstm_9_g: num-iters=1200 nj=4..4 num-params=48.3M dim=40+100->3426 xent:train/valid[798,1199]=(-0.745,-0.680/-0.785,-0.749) logprob:train/valid[798,1199]=(-0.060,-0.051/-0.072,-0.068)


_a is the old, well working model, _g is the one that has worse performance (on the same iterations)

Asr results on iteration 200 and 1200:

    %WER 20.47 [ 11780 / 57534, 1055 ins, 2275 del, 8450 sub ] exp/chain/model/gr_tdnn_blstm_9_a/decode200/wer_8_0.0
    %WER 18.27 [ 10509 / 57534, 1136 ins, 1809 del, 7564 sub ] exp/chain/model/gr_tdnn_blstm_9_a/decode1200/wer_7_0.5
[...]

PR #1730, opened 2017-06-30
New tdnn-blstm scripts under swbd

GaofengCheng:
related to #1660


danpovey:
You can move the soft link to point to 1d instead of 1a.

danpovey:
remember to move the soft-link to point to the latest script

GaofengCheng:
@danpovey redirect the soft link to the newest script

PR #1734, opened 2017-07-02
First provisional version of extras/install_phonetisaurus.sh script

AdolfVonKleist:
This includes an ```install_phonetisaurus.sh``` script, which borrows as much as possible from the existing ```install_sequitur.sh``` script and update to .gitignore to add the ```tools/phonetisaurus-g2p``` directory created during installation.

It is tied to the ```kaldi``` tag in the Phonetisaurus repository (https://github.com/AdolfVonKleist/Phonetisaurus/tree/kaldi).


danpovey:
Thanks!  @jtrmal is in charge of this aspect of things so I'll wait for him
to look at it.

jtrmal:
Thanks -- script looks good.
Before merging this I'd propose to add the train and apply script as well
(so that everything will get merged together).
y.

AdolfVonKleist:
Sounds good.  Just to clarify, you mean adding the python bindings, correct?  The c++ utilities include: ```phonetisaurus-align```, ```phonetisaurus-arpa2wfst```, ```phonetisaurus-g2pfst```, and ```phonetisaurus-g2prnn```.   

If yes, should I make sure that that python ```setup.py``` installs everything locally/to user space?  This doubt was the main reason I did not include the bindings in the first iteration.

jtrmal:
I think the python bindings are not that important.
I was thinking of providing two scripts,
The first one would take the standard-format lexicon and did the m2m
alignment and LM training and conversion of the LM to fst and second one
would be taking the wordlist and the directory with the trained g2p and
generate pronunciation lexicon
Y.

AdolfVonKleist:
Ah OK, got it.  I will add these back in, but still omit the bindings.

tools/extras/install_phonetisaurus.sh
```diff
@@ -0,0 +1,90 @@
...
+fi
+
+
+if [ ! -d ./phonetisaurus-g2p ] ; then
+  git clone https://github.com/AdolfVonKleist/Phonetisaurus.git phonetisaurus-g2p ||
+  {
+    echo  >&2 "$0: Warning: git clone operation ended unsuccessfully"
+    echo  >&2 "  I will assume this is because you don't have https support"
+    echo  >&2 "  compiled into your git "
+    git clone https://github.com/AdolfVonKleist/Phonetisaurus.git phonetisaurus-g2p
```
psmit: Given your comment about https, I guess you want to clone the http (or git://) version here?

AdolfVonKleist: Yes, thanks good catch.

AdolfVonKleist:
I've added two wrapper scripts, ```src/bin/phonetisaurus_train```, and ```src/bin/phonetisaurus_apply```, and updated the ```kaldi``` tag.  The scripts can be used like so:
```
# Vanilla training with some basic, sane defaults
$ phonetisaurus_train --lexicon cmudict.formatted.dict --seq2_del

# Apply the model to a word list
$ phonetisaurus_apply --model train/model.fst --word_list wordlist
```
The README.md file and help messages from the scripts have additional info.  Change, bugfix or update requests to these are also welcome.

jtrmal:
Oh, great. Thanks a lot. 
@danpovey I think this is ok to merge -- I'll try to adapt some of the recipes using sequitur to use this instead.

jtrmal:
@AdolfVonKleist I just realized one possible feature request -- sequitur allows to specify that instead of nbest alternatives it generates that many alternatives so that at least p% mass of the total probability is covered. In many cases this makes more sense than nbest. I was wondering if you could add that as well?

AdolfVonKleist:
Yes that should not be too big an add.  I probably would still restrict the search space a bit, or at least leave it tweakable as an option, since it will slow things down.  At present I use a tweaked version of the OpenFst [NShortestPath](https://github.com/AdolfVonKleist/Phonetisaurus/blob/master/src/include/PhonetisaurusRex.h#L209) that accepts a custom filter.  By default this just filters for unique output phoneme sequences.  

I guess the easiest thing to do here would be to create an alternative path filter and aggregate/keep paths that share the same unique output phoneme sequence for some tunable fraction of the lattice.  Then we push in log towards the final state,  remove total weight and from there we can output the list of prons up to %p mass of the (remaining) total.  Would that be a reasonable compromise?

jtrmal:
Yes, I think that might be OK. And yes, I meant that only as an option (and
disabled by default)
Y.

On Jul 13, 2017 8:45 PM, "Josef Novak" <notifications@github.com> wrote:

Yes that should not be too big an add. I probably would still restrict the
search space a bit, or at least leave it tweakable as an option, since it
will slow things down. At present I use a tweaked version of the OpenFst
NShortestPath
<https://github.com/AdolfVonKleist/Phonetisaurus/blob/master/src/include/PhonetisaurusRex.h#L209>
that accepts a custom filter. By default this just filters for unique
output phoneme sequences.

I guess the easiest thing to do here would be to create an alternative path
filter and aggregate/keep paths that share the same unique output phoneme
sequence for some tunable fraction of the lattice. Then we push in log
towards the final state, remove total weight and from there we can output
the list of prons up to %p mass of the (remaining) total. Would that be a
reasonable compromise?

—
You are receiving this because you were mentioned.

<https://github.com/kaldi-asr/kaldi/pull/1734#issuecomment-315167508>, or mute
the thread

danpovey:
Guys,  is there a way to get the probabilities for the different prons that Phonetisaurus generates?  Is this already supported in this PR?

AdolfVonKleist:
I am adding this as an option as part of the response to @jtrmal request.  At present it just returns the raw scores for each hypothesis, along with the raw multigram values [you can also dump the complete lattice related to a given word].

danpovey:
@jtrmal and josef-- do you think it would be better, or desirable, to
obtain probabilities instead of raw scores?  They could just be obtained
based on the assumption that the top-100 (or some specifiable top-n)
contains all the probability mass.  It seems to me that that would be more
convenient, especially for people who are used to a Sequitur-based workflow.

AdolfVonKleist:
@danpovey yes, this makes sense to me, and is a pretty small add-on, and of a piece with the request made my @jtrmal .  I started work on this last weekend, but did not quite finish.  This weekend I am away, but I see no obstacle to wrapping it up next weekend.  

One additional question that might be worth considering is: what do you prefer to do with the reference lexicon?  In the binaries there is no concept of a reference lexicon, only the learned model, and there is no guarantee that the model will always generate correct pronunciations, even for seen words.  You can probably partially mitigate this using corpus statistics during training, but I think the point still stands.

In the scripts, the reference lexicon can be loaded, and the G2P called either only as a back-off - i.e. when there is no reference; or in 'greedy' mode, where it will be called in either case, but entries already present in the reference will be ignored.

If you combine the above with the idea of, "_generating all pronunciation hypotheses associated with at least %p of the mass related to a given word_", then you might also want to consider how much of that mass you want to allocate a priori to any available references.

danpovey:
I would have thought that dealing specially with reference words would be
done by a calling script in a local directory, not by the
phonetisaurus-related scripts in steps/,  but Yenda might have something to
say about this.

jtrmal:
I;m not sure if I understand fully, but from my POV, the fact that g2p can
generate pronunciations not matching the lexicon pronunciations is a
non-issue and higher level scripts would have to resolve it somehow.
y.

AdolfVonKleist:
OK, that sounds fine and keeps things simple.

AdolfVonKleist:
@jtrmal @danpovey I have updated the distribution so it will produce normalized probs rather than the default raw scores.  The ```phonetisaurus_apply``` script can now be called (optionally) like so:

```
$ phonetisaurus_apply --model train/model.fst --word_list words.list \
    --verbose --accumulate --pmass 0.88 --probs 2> g2p.log
test    0.68    T EH1 S T
test    0.21    T EH2 S T
right   0.81    R AY1 T
right   0.13    R AY0 T
junkify 0.64    JH AH1 NG K AH0 F AY2
junkify 0.23    JH AH1 NG K IH0 F AY2
junkify 0.05    JH AH1 NG K AH0 F AY0
```

The ```--accumulate``` flag indicates that scores should be accumulated over identical output pronunciations, ignoring any epsilons.  

The ```--pmass``` flag indicates the probability mass cutoff to be used.  In the above example this means, "output hypotheses until these sum to at least 88% of the total mass for this word."  

The ```--probs``` flag indicates that the scores should be output as human-readable probs (by default they will still be written out as raw scores, even when ```--pmass``` is in play).

The optional ```--beam``` parameter will tweak the size of the lattice.  By default it is set to 10000; probably a bit high.  If you set it to something very low, like ```--beam 3```, it should be clear how this affects the number of nbest, and the relative mass of the remaining hypotheses.

-joe

danpovey:
Thanks!  
I guess we can merge this PR because it's just the installation script and won't break anything.
@jtrmal @xiaohui-zhang and @ngoel17, please take notice of what he said here-- it might be useful in future.  I think it would be great if we had examples of using this within Kaldi. but I guess that's a separate issue and not part of this PR.
@jtrmal, can you merge if you think it's OK?

jtrmal:
Oh, great -- this was exactly how I was imagining it.
@danpovey -- I'll add it to babel (replace the sequitur), but that's probably not really interesting to you?

xiaohui-zhang:
@AdolfVonKleist was MITLM installed by default? Otherwise we can use other existing LM toolkits in kaldi in phonetisaurus_train?

xiaohui-zhang:
Also there seems to be a problem when I run phonetisarus-align:
src/bin/phonetisaurus-align: error while loading shared libraries: libfst.so.7: cannot open shared object file: No such file or directory

AdolfVonKleist:
@xiaohui-zhang No.  I wrote a hook in the script for this, but only added command details for mitlm [L134](https://github.com/AdolfVonKleist/Phonetisaurus/blob/master/src/bin/phonetisaurus_train#L134) it will throw a ```NotImplemented``` error if you try to set it to something else.  If mitlm is not present it will throw an ```EnvironmentError```.

The ```phonetisaurus-arpa2wfst``` binary only cares that the input is a valid ARPA format LM, so there is no inherent restriction to or depency on mitlm.  I am not familiar with the kaldi LM tool but I guess it should be short work to extend the script with a command for it.

Regarding the second comment, OpenFst needs to be in your ```$LD_LIBRARY_PATH```.   Maybe that needs to be double checked in the install script, if it is not already present in ```env.sh```.

xiaohui-zhang:
Thanks @AdolfVonKleist

danpovey:
Is it possible to add the OpenFst library path to the phonetisaurus
binaries via -rpath?   That's what we do with Kaldi tools.  It seems a
shame to have one more thing to keep track of; we don't normally ensure
that OpenFst's library directory is on our LD_LIBRARY_PATH.

AdolfVonKleist:
Sure, should be no problem.  I will make this change.

danpovey:
Thanks!

PR #1749, opened 2017-07-08
Add configuration check for CUDA incompatible g++ compilers

dogancan:
These checks are fairly loose since we do not want to be too restrictive with g++ version unless we know for a fact that a particular g++ version is incompatible with the particular CUDA toolkit installed on the system.

I added the check inside configure_cuda function since this check is relevant only when compiling with CUDA.

Relates to #1744.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -380,18 +380,39 @@ function configure_cuda {
       failure "Cannot cross compile with CUDA support"
     fi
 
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks a lot!  Will merge when the check finishes.

dogancan:
I did not add any checks for compilers other than g++ since I don't know what versions are incompatible with what versions of CUDA. Let me know if you would like checks for other compilers as well.

PR #1752, opened 2017-07-10
fix a no formal arpa file parse problem

LvHang:
This is a small bug fix about parsing no formal arpa file.
The case is if a blink contains some whitespaces, "arpa-file-parse.cc" will report "line [  ]: Invalid n-gram data line".


danpovey:
did you run arpa2fst with this change?

LvHang:
Yes. I did.
I use the "arpa2fst" to deal with the problematic arpa file. It can pass.
At the same time, I try the normal swbd arpa file (data/local/lm/sw1_fsh.o3g.kn). It goes smoothly.

PR #1757, opened 2017-07-11
make the combine stage optional

LvHang:
Hi Dan,
I take steps/nnet3/train_dnn.py as an example to show how I modify the codes.

There is one question in line 367 of steps/nnet3/train_dnn.py. For convenience, I copy the last-numbered model to combine model so that the code conduct adjusting the priors uniformly. It's a little bit opportunistic. In another way, I can adjust the priors directly but it will make the codes looks a bit redundancy. How do you think about it?

Hang

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/train/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/common.py
@@ -818,6 +818,14 @@ class CommonParser(object):
                                  the final model combination stage.  These
                                  models will themselves be averages of
                                  iteration-number ranges""")
+        self.parser.add_argument("--trainer.optimization.do-final-combination",
+                                 dest='do_final_combination', type=str,
+                                 action=common_lib.StrToBoolAction,
+                                 choices=["true", "false"], default=True,
+                                 help="""Use this to specify conducting final
+                                 model combination or not. If the option is set
+                                 to false, for now just copy the last-numbered
+                                 model to final model.""")
         self.parser.add_argument("--trainer.optimization.combine-sum-to-one-penalty",
                                  type=float, dest='combine_sum_to_one_penalty', default=0.0,
                                  help="""If > 0, activates 'soft' enforcement of the
--- a/egs/wsj/s5/steps/libs/nnet3/train/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/common.py
@@ -822,10 +822,9 @@ class CommonParser(object):
                                  dest='do_final_combination', type=str,
                                  action=common_lib.StrToBoolAction,
                                  choices=["true", "false"], default=True,
-                                 help="""Use this to specify conducting final
-                                 model combination or not. If the option is set
-                                 to false, for now just copy the last-numbered
-                                 model to final model.""")
+                                 help="""Set this to false to disable the final
+                                 'combine' stage (in this case we just use the
+                                 last-numbered model as the final.mdl).""")
         self.parser.add_argument("--trainer.optimization.combine-sum-to-one-penalty",
                                  type=float, dest='combine_sum_to_one_penalty', default=0.0,
                                  help="""If > 0, activates 'soft' enforcement of the
--- a/egs/wsj/s5/steps/nnet3/chain/train.py
+++ b/egs/wsj/s5/steps/nnet3/chain/train.py
@@ -401,10 +401,15 @@ def train(args, run_opts):
     num_iters = ((num_archives_to_process * 2)
                  / (args.num_jobs_initial + args.num_jobs_final))
 
-    models_to_combine = common_train_lib.get_model_combine_iters(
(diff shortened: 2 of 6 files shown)


egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -818,6 +818,14 @@ def __init__(self,
                                  the final model combination stage.  These
                                  models will themselves be averages of
                                  iteration-number ranges""")
+        self.parser.add_argument("--trainer.optimization.do-final-combination",
+                                 dest='do_final_combination', type=str,
+                                 action=common_lib.StrToBoolAction,
+                                 choices=["true", "false"], default=True,
+                                 help="""Use this to specify conducting final
```
danpovey: change the help string to:
`Set this to false to disable the final 'combine' stage (in this case we just use the last-numbered model as the final.mdl.`

egs/wsj/s5/steps/nnet3/train_dnn.py
```diff
@@ -281,10 +281,14 @@ def train(args, run_opts):
...
-        num_archives_expanded, args.max_models_combine,
-        args.num_jobs_final)
+    # If do_final_combination is True, compute the set of models_to_combine.
+    # Otherwise, models_to_combine will be none.
+    models_to_combine = None
+    if args.do_final_combination:
+        models_to_combine = common_train_lib.get_model_combine_iters(
+            num_iters, args.num_epochs,
+            num_archives_expanded, args.max_models_combine,
+            args.num_jobs_final)
```
danpovey: I'd put the 'models_to_combine = None' in the 'else' branch.

egs/wsj/s5/steps/nnet3/train_dnn.py
```diff
@@ -352,13 +356,22 @@ def train(args, run_opts):
...
+            logger.info("Doing final combination to produce final.mdl")
+            train_lib.common.combine_models(
+                dir=args.dir, num_iters=num_iters,
+                models_to_combine=models_to_combine,
+                egs_dir=egs_dir,
+                minibatch_size_str=args.minibatch_size, run_opts=run_opts,
+                sum_to_one_penalty=args.combine_sum_to_one_penalty)
+        else:
+            logger.info("""Producing final.mdl without doing final combination.
+                        Copying the last-numbered model to combined model so that
```
danpovey: I think it's potentially confusing to create a combined.mdl which is not really the combined.mdl.
Best to use different source models later on.

egs/wsj/s5/steps/nnet3/train_dnn.py
```diff
@@ -353,27 +358,43 @@ def train(args, run_opts):
...
-            num_archives=num_archives,
-            prior_subset_size=args.prior_subset_size, run_opts=run_opts)
-
-        logger.info("Re-adjusting priors based on computed posteriors")
-        combined_model = "{dir}/combined.mdl".format(dir=args.dir)
-        final_model = "{dir}/final.mdl".format(dir=args.dir)
-        train_lib.common.adjust_am_priors(args.dir, combined_model,
-                                          avg_post_vec_file, final_model,
-                                          run_opts)
+        if args.do_final_combination:
```
danpovey: there is too much code here.  Can't you do things like
   `iter=('combined' if args.do_final_combination else num_iters),`
and the like?

LvHang:
Hi Dan,
I updated all the scripts and am testing it with minilibrispeech.

LvHang:
I have been tested nnet3/train_rnn.py and chain/train_rnn.py with `--trainer.optimization.do-final-combination=false` on mini-librispeech. All goes well.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -496,17 +501,28 @@ def train(args, run_opts):
...
+                leaky_hmm_coefficient=args.leaky_hmm_coefficient,
+                l2_regularize=args.l2_regularize,
+                xent_regularize=args.xent_regularize,
+                run_opts=run_opts,
+                sum_to_one_penalty=args.combine_sum_to_one_penalty)
+        else:
+            logger.info("Copying the last-numbered model to final.mdl")
+            common_lib.force_symlink("{0}.mdl".format(num_iters),
+                                     "{0}/final.mdl".format(args.dir))
+            chain_lib.compute_train_cv_probabilities(
```
danpovey: Instead of calling compute_train_cv_probabilities, you could just force-symlink the final-numbered logs to the 'final' logs.

LvHang:
I have been tested the "steps/nnet3/chain/train.py" script after modifying the symbolic link of log file and run the experiments for "steps/nnet3/chain/train.py" and "steps/nnet3/train_rnn.py" without setting `do-final-combination=true`. Those are not broken.

PR #1760, opened 2017-07-12
changed --nj value to  instead of hard coded to 100

johnjosephmorgan:
Also, shouldn't the ali_dir be used instead of the gmm_dir?The alignment with lats is hard coded to use 100 jobs, but it is supposed to use the same number of jobs that were used to build i-vectors.


danpovey:
Hm.  I think that may have been intentional-- the run_ivector_common.sh will mostly be limited by disk speed which is why we have a lower number of jobs.  I know it's not very elegant but you are supposed to just change that 100 to whatever number you really want to use.  It only affects speed, it won't change the results.  There is no requirement that it have the same value as you dumped ivectors with.

PR #1762, opened 2017-07-14
[src] Add GruNonlinearityComponent; moving around some sources in nne…

danpovey:
…t3 to avoid very large files.


danpovey:
@GaofengCheng, if you read the source it should be obvious how to use this to make fast-gru and fast-pgru layers.

GaofengCheng:
@danpovey Thx Dan, will look into this

GaofengCheng:
@danpovey I will upload a gru.py as soon as I can , I'm working on the GRU, seeking for more properties of it

danpovey:
@GaofengCheng and @LvHang, I notice this has conflicts now.  Since we are still using GRUs, I think it might make sense not to lose track of this.  Not urgent though, since the TDNN-F are the recommended system under most circumstances.

LvHang:
So we should take the code from your branch "danpovey/gru" and fix these conflicts, right? Let me look into it.

danpovey:
It looks like you already merged that code into *something* in your
branch... I'm not sure where that went.

LvHang:
@danpovey 
Hi Dan, I haven't found the code in my branch. So I just merged your "gru" code into the newest kaldi master.
As your branch is far more behind the newest kaldi master, so I just keep your "gru" code and other code(ConvolutionComponent, LstmNonlinearityComponent and MaxPoolingComponent) in "nnet3-special-component.{h,cc}" is replaced by the newest kaldi code.
I sloved the conflicts and it can be compiled.
You can update your gru branch with https://github.com/LvHang/kaldi/tree/gru

danpovey:
Thanks.
Just leave the branch there for now.  I'm hoping that at some point @GaofengCheng may have a chance to look at it, and maybe make GRU recipes based on it.
Dan

PR #1763, opened 2017-07-14
minor bugfix to nnet1 alignment script when creating lattices

wombat78:
nnet1 alignment script breaks when trying to optionally create lattices


wombat78:
need more fix.
will resubmit.

PR #1767, opened 2017-07-17
Sparse matrix 2 CPU version

kangshiyin:
Hi @danpovey Here's some code snippets for #1695

### DIFF
--- a/src/matrix/sparse-matrix.cc
+++ b/src/matrix/sparse-matrix.cc
@@ -20,6 +20,7 @@
 
 #include <algorithm>
 #include <limits>
+#include <map>
 #include <string>
 
 #include "matrix/sparse-matrix.h"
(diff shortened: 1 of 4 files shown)


kangshiyin:
Redesign SparseMatrix to use the internal storage of CSR format.

Keep all APIs unchanged except for .Data() and .Row() which depend on the old storage scheme.
Pass all unit tests in the follow folders that do not involve GPU,
base matrix util feat tree gmm transform sgmm2 fstext hmm lm decoder lat ivector kws

danpovey:
Shiyin, I don't want you to change the SparseMatrix code at all.  All changes should be to class CuSparseMatrix.

kangshiyin:
Oops... I thought it would be better if we use CSR format in both classes. The data copy between CPU and GPU will be much simpler, and CPU sparse BLAS libraries may be used in the future.

danpovey:
I'd rather change the CPU code in the future if it's ever needed; changing
the CPU code will cause hassles with testing the I/O back compatibility and
I just don't like to change code unless there is a good reason.  Right now
the CPU and GPU formats are already different.

danpovey:
BTW, instead of replacing the Data() function in CuSparseMatrix with a
bunch of functions, I think it would be better to remove it and use friend
declarations-- as long as the number of friend declarations is reasonable.

kangshiyin:
OK. I will then use the original SparseMatrix.

Regarding `CsrRowPtr()`, `CsrColIdx()` and `CsrVal()`, how about we make them private, so we don't need to write `csr_row_ptr_col_idx_ + num_rows_ + 1` for `csr_col_idx` every time?

danpovey:
Fine if private (or even protected).

PR #1768, opened 2017-07-17
add default args for arpa sampling test

keli78:
The two args are 1) a small ARPA file and 2) a history file which is in directory rnnlm/test/

### DIFF
--- a/src/rnnlm/arpa-sampling-test.cc
+++ b/src/rnnlm/arpa-sampling-test.cc
@@ -141,7 +141,8 @@ int main(int argc, char **argv) {
   const char *usage = "";
   ParseOptions po(usage);
   po.Read(argc, argv);
-  std::string arpa_file = po.GetArg(1), history_file = po.GetArg(2);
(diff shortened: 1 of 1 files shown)


danpovey:
Please don't do it this way, it will be confusing for people who are
looking at that makefile.  Instead, if no args are supplied, act as if the
args have those default values.

keli78:
@danpovey  I fixed it.

jtrmal:
I don't think this is right...   Your change will cause each test binary to be called with the given arguments.
I'd suggest modifying the default rule so that it looks for a variable called
binary_name_ARGS (for example rnnlmtest_ARGS) and if it does exist, it runs the binary with the given arguments.
y.

danpovey:
Ke, I think you misunderstood what I was asking for.
I meant to change the C++ code of the binary.
It's almost never a good idea to change something global just to fix
something local.

keli78:
Oh, got it.

PR #1770, opened 2017-07-18
aishell recipe fixes

naxingyu:
fix aishell url bug.


naxingyu:
@danpovey ready to merge

PR #1773, opened 2017-07-21
added new ComputeFeatures function and modified binaries to allow do…

pegahgh:
…wnsampling waveform(waveform downsampling function is not added yet).

### DIFF
--- a/src/feat/feature-common-inl.h
+++ b/src/feat/feature-common-inl.h
@@ -25,24 +25,62 @@
 namespace kaldi {
 
 template <class F>
+void OfflineFeatureTpl<F>::ComputeFeatures(
+    const VectorBase<BaseFloat> &wave,
+    BaseFloat sample_freq,
+    BaseFloat vtln_warp,
+    Matrix<BaseFloat> *output) {
+  KALDI_ASSERT(output != NULL);
+  BaseFloat new_sample_freq = computer_.GetFrameOptions().samp_freq;
+  if (sample_freq == new_sample_freq)
+    Compute(wave, vtln_warp, output);
+  else {
+    if (new_sample_freq < sample_freq) {
+      if (! computer_.GetFrameOptions().allow_downsample)
+        KALDI_ERR << "Waveform and config sample Frequency mismatch: "
+                  << sample_freq << " .vs " << new_sample_freq
+                  << "--allow_downsample should be true to allow "
+                  << " downsampling the waveform.";
+
+      // Downsample the waveform.
+      Vector<BaseFloat> downsampled_wave(wave);
+      Compute(downsampled_wave, vtln_warp, output);
+    } else
+      KALDI_ERR << "The waveform is allowed to get downsampled."
+                << "New sample Frequency " << new_sample_freq
+                << " is larger than waveform original sampling frequency "
+                << sample_freq;
+
+  }
+}
+
+template <class F>
+void OfflineFeatureTpl<F>::ComputeFeatures(
+    const VectorBase<BaseFloat> &wave,
+    BaseFloat sample_freq,
+    BaseFloat vtln_warp,
+    Matrix<BaseFloat> *output) const {
+  OfflineFeatureTpl<F> temp(*this);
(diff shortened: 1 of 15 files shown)


pegahgh:
@danpovey 
I made all changes to allow downsampling the waveform during feature extraction.
I didn't add function to do waveform downsampling yet.
I wonder if you have specific plan to implement this function.

src/feat/feature-common.h
```diff
@@ -118,25 +118,48 @@ class OfflineFeatureTpl {
...
+     This is the newer interface where you specify the sample frequency
+     of the input waveform.
+       @param [in] wave   The input waveform
+       @param [in] sample_freq  The sampling frequency with which
+                                'wave' was sampled.
+     @param [in] vtln_warp  The VTLN warping factor (will normally
+                            be 1.0)
+     @param [out]  output  The matrix of features, where the row-index
+                           is the frame index.
+
```
danpovey: remove blank line

src/feat/feature-common.h
```diff
@@ -118,25 +118,48 @@ class OfflineFeatureTpl {
...
-               Vector<BaseFloat> *deprecated_wave_remainder = NULL) const;
+               Matrix<BaseFloat> *output) const;
+
+  /**
+     Computes the features for one file (one sequence of features).
+     This is the newer interface where you specify the sample frequency
+     of the input waveform.
+       @param [in] wave   The input waveform
+       @param [in] sample_freq  The sampling frequency with which
+                                'wave' was sampled.
```
danpovey: clarify that if sample_freq is higher than the frequency specified in the config,
we'll downsample, but if lower, it's an error.

src/feat/feature-window.h
```diff
@@ -57,7 +58,8 @@ struct FrameExtractionOptions {
       window_type("povey"),
       round_to_power_of_two(true),
       blackman_coeff(0.42),
-      snip_edges(true){ }
+      snip_edges(true),
+      allow_downsample(false){ }
```
danpovey: space after ) [this was wrong in the original code]

src/feat/feature-common-inl.h
```diff
@@ -25,24 +25,62 @@
...
+    Compute(wave, vtln_warp, output);
+  else {
+    if (new_sample_freq < sample_freq) {
+      if (! computer_.GetFrameOptions().allow_downsample)
+        KALDI_ERR << "Waveform and config sample Frequency mismatch: "
+                  << sample_freq << " .vs " << new_sample_freq
+                  << "--allow_downsample should be true to allow "
+                  << " downsampling the waveform.";
+
+      // Downsample the waveform.
```
danpovey: OK, so I guess this is where you'd call the 'real' downsample function.  
Declare it in resample.h, and maybe make it something like:
```
/// Downsample a waveform.  This is a convenience wrapper for the 
/// class 'LinearResample'.
void DownsampleWaveform(BaseFloat orig_freq, const VectorBase<BaseFloat> &wave,
                BaseFloat new_freq, Vector<BaseFloat> *new_wave);
```
Regarding the filter-cutoff, please find out what sox uses and use that rule to find it;
and document it in the declaration of the function.  For 'num_zeros', maybe use 10.

src/featbin/compute-spectrogram-feats.cc
```diff
@@ -109,16 +109,16 @@ int main(int argc, char *argv[]) {
         }
       }
 
-      if (spec_opts.frame_opts.samp_freq != wave_data.SampFreq())
+      if (spec_opts.frame_opts.samp_freq > wave_data.SampFreq())
         KALDI_ERR << "Sample frequency mismatch: you specified "
-                  << spec_opts.frame_opts.samp_freq << " but data has "
-                  << wave_data.SampFreq() << " (use --sample-frequency "
-                  << "option).  Utterance is " << utt;
+                  << spec_opts.frame_opts.samp_freq << " larger than data sample freq "
```
danpovey: You can remove this check, it's checked internally.

PR #1776, opened 2017-07-26
Some fixes re CuArray + Add CuMatrixBase::AddToElements + test

hhadian:
Note: Re CuMatrixBase::AddToElements, I did not add a corresponding member function for the MatrixBase class. I did the on-CPU computation in place (i.e. in the CuMatrixBase class) just like CuMatrixBase::Lookup.

### DIFF
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -40,7 +40,7 @@ namespace kaldi {
 template<typename T>
 void CuArray<T>::Resize(MatrixIndexT dim, MatrixResizeType resize_type) {
   KALDI_ASSERT((resize_type == kSetZero || resize_type == kUndefined) && dim >= 0);
-  if (dim_ == dim) {
+  if (this->dim_ == dim) {
     if (resize_type == kSetZero)
       this->SetZero();
     return;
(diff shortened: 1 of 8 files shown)


hhadian:
Seems the previous commit has appeared here. Will rebase now...

src/matrix/kaldi-matrix.h
```diff
@@ -324,6 +324,14 @@ class MatrixBase {
   /// the same).
   void AddToRows(Real alpha, Real *const *dst) const;
 
+  /// For each row i of *this, adds this->Row(i) to
+  /// dst->Row(indexes(i)) if indexes(i) >= 0, else do nothing.
+  /// Requires that all the indexes[i] that are >= 0
+  /// be distinct, otherwise the behavior is undefined.
+  void AddToRows(Real alpha,
```
danpovey: In general I don't like adding too many of this type of function in class Matrix, because many of these functions only make sense in the GPU context (on CPU you could just make a loop).  But because the other AddToRows() function already existed in class Matrix, it probably makes sense to put this one there too, for symmetry.

hhadian: Yes, I did this for symmetry. 
But I didn't do this for AddToElements. Is that OK?

danpovey:
yes that's OK.

hhadian:
Rebased and ready.

PR #1778, opened 2017-07-27
Add rnnlm::ReadSparseWordFeatures

hhadian:


src/rnnlm/rnnlm-utils.cc
```diff
@@ -20,20 +21,45 @@
...
+  std::vector<std::vector<std::pair<MatrixIndexT, BaseFloat> > > sparse_rows;
+  std::string line;
+  int32 line_number = 0;
+  while (std::getline(is, line)) {
+    std::vector<std::pair<MatrixIndexT, BaseFloat> > row;
+    std::istringstream line_is(line);
+    int32 word_id;
+    line_is >> word_id;
+    line_is >> std::ws;
+    KALDI_ASSERT(word_id == line_number++);
```
danpovey: For things that could be a problem in user input, better to fail with KALDI_ERR and an informative error message, instead of KALDI_ASSERT.

src/rnnlm/rnnlm-utils.cc
```diff
@@ -1,6 +1,7 @@
 // rnnlm/rnnlm-utils.cc
 
 // Copyright 2017  Daniel Povey
+//                 Hossein Hadian
```
danpovey: this line should give the year (2017).

src/rnnlm/rnnlm-utils.cc
```diff
@@ -20,20 +21,45 @@
...
+    int32 word_id;
+    line_is >> word_id;
+    line_is >> std::ws;
+    KALDI_ASSERT(word_id == line_number++);
 
+    int32 feature_index;
+    BaseFloat feature_value;
+    while (line_is >> feature_index)
+    {
+      KALDI_ASSERT(feature_index >= 0 && feature_index < feature_dim);
```
hhadian: use KALDI_ERR for here too?

danpovey:
yes.

PR #1779, opened 2017-07-28
CuSparseMatrix using CSR storage format

kangshiyin:
Redesign CuSparseMatrix using CSR storage format; Implement a few new APIs and add unit tests.

All tests in matrix/ and cudamatrix/ have passed.

### DIFF
--- a/src/cudamatrix/cu-sparse-matrix.h
+++ b/src/cudamatrix/cu-sparse-matrix.h
@@ -2,6 +2,7 @@
 
 // Copyright      2015  Johns Hopkins University (author: Daniel Povey)
 //                2015  Guoguo Chen
+//                2017  Shiyin Kang
 
 // See ../../COPYING for clarification regarding multiple authors
 //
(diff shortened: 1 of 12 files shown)


src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -43,82 +44,63 @@ Real TraceMatSmat(const CuMatrixBase<Real> &A,
...
-
   /// Copy from CPU-based matrix.
-  CuSparseMatrix<Real> &operator = (const SparseMatrix<Real> &smat);
+  CuSparseMatrix<Real> &operator =(const SparseMatrix<Real> &smat);
 
   /// Copy from possibly-GPU-based matrix.
-  CuSparseMatrix<Real> &operator = (const CuSparseMatrix<Real> &smat);
+  CuSparseMatrix<Real> &operator =(const CuSparseMatrix<Real> &smat);
+
+  void Resize(const MatrixIndexT num_rows, const MatrixIndexT num_cols,
```
danpovey: I think this should probably be private or protected, or at least should have a warning saying that users won't normally have to use this.  And I don't understand why having nnz defaulting to 0 is the right thing.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -140,79 +122,123 @@ class CuSparseMatrix {
...
   inline const SparseMatrix<Real> &Mat() const {
-    return *(reinterpret_cast<const SparseMatrix<Real>* >(this));
+    return *(reinterpret_cast<const SparseMatrix<Real>*>(this));
   }
   inline SparseMatrix<Real> &Mat() {
-    return *(reinterpret_cast<SparseMatrix<Real>* >(this));
+    return *(reinterpret_cast<SparseMatrix<Real>*>(this));
+  }
+
+  /// Returns pointer to the data array of lenght nnz_ that holds all nonzero
```
danpovey: typo

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -140,79 +122,123 @@ class CuSparseMatrix {
...
+  /// values in zero-based CSR format
+  const Real* CsrVal() const {
```
danpovey: It's not clear to me that it's the right decision, to make these things public.  Is it feasible to make these things private and make use of friend declarations?  I want to leave the door open to refactoring this class later on if there is a need.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -140,79 +122,123 @@ class CuSparseMatrix {
...
+
+private:
   // This member is only used if we did not compile for the GPU, or if the GPU
   // is not enabled.  It needs to be first because we reinterpret_cast this
   std::vector<SparseVector<Real> > cpu_rows_;
 
-  MatrixIndexT num_rows_;
-  MatrixIndexT num_cols_;
-
   // This is where the data lives if we are using a GPU.  Notice that the format
```
danpovey: this comment is no longer correct.  I think you should give a pointer to some NVidia or other documentation about how the csr format works, and mention that we store a couple of those arrays in one allocated block of memory.

src/cudamatrix/cu-sparse-matrix.h
```diff
@@ -140,79 +122,123 @@ class CuSparseMatrix {
...
+  // matrix size num_rows_ x num_cols_
+  MatrixIndexT num_rows_;
+  MatrixIndexT num_cols_;
+
+  // number of non-zeros
+  MatrixIndexT nnz_;
+
+  // csr row ptrs and col indices in a single int array
+  // of the length (num_rows_ + 1 + nnz_)
+  int* csr_row_ptr_col_idx_;
```
danpovey: normally we'd put the * to the right of the space in this context, but let me know if the Google style guide states that this is wrong.

src/cudamatrix/cu-sparse-matrix.h
kangshiyin: Defaulting to 0 was trying to be compatible with other classes, by using two arguments only, but probably not reasonable. I will add a comment for it.

src/cudamatrix/cu-sparse-matrix.h
kangshiyin: They are now protected.

src/cudamatrix/cu-sparse-matrix.h
kangshiyin: Will fix.

src/cudamatrix/cu-sparse-matrix.h
kangshiyin: Both are fine according to this link.
https://google.github.io/styleguide/cppguide.html#Pointer_and_Reference_Expressions

src/cudamatrix/cu-sparse-matrix.h
kangshiyin: Will fix.

kangshiyin:
I've debugged the code with original and additional unit tests. I think it should be fine.

src/cudamatrix/cu-sparse-matrix.h
danpovey: Don't make the Resize method public and don't make nnz have a default.  It should not be called from outside the class anyway.

kangshiyin: Done.

danpovey:
Thanks!  I'm merging now; I'll review the whole rnnlm PR in more detail before I merge to master but I doubt there are any significant problems in this part.
Please try to get it to the stage where there are no more linking errors in the cudamatrix directory (even if not everything is implemented); this will help me to make progress in the rnnlm directory as I can then check for linking errors there.

PR #1780, opened 2017-07-29
Add CuVectorBase::CopyElements() and VecMatVec() + tests

hhadian:
Two notes:
1. After adding a new unit test to cu-vector-test.cc, [ApplyExp](https://github.com/hhadian/kaldi/blob/625518d2b06481d0fa663242dad2cf9ce7da4df7/src/cudamatrix/cu-vector-test.cc#L522) unit test fails [which is completely unrelated anyway]. I guess it is simply because the random state has changed and a random number is generated on which ApplyExp has an error > 0.000001
2. There are some lines in cu-vector-test.cc where a random int is generated using `int32 M = 10 % Rand() % 100;`. I guess the first `%` should be `+`. Should I fix them?

### DIFF
--- a/src/cudamatrix/cu-kernels-ansi.h
+++ b/src/cudamatrix/cu-kernels-ansi.h
@@ -502,6 +502,14 @@ void cudaD_matrix_lookup(dim3 Gr, dim3 Bl, const double *data, MatrixDim dim,
 void cudaF_matrix_lookup(dim3 Gr, dim3 Bl, const float *data, MatrixDim dim,
                          const Int32Pair *indices, int indices_size,
                          float *output);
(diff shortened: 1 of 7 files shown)


danpovey:
Thanks.  I'll merge, but please separately fix those bugs in cu-vector-test.cc and also see if you can fix the ApplyExp() problem by changing the threshold e.g. by a factor of 10.

hhadian:
Will do. A factor of 10 fixes that problem.

hhadian:
Also something else, I have not written 2 separate kernels for transposed and non-transposed versions of [CopyElements](https://github.com/kaldi-asr/kaldi/pull/1780/files#diff-ac2b19fd476b7714f65049829e244de0R1191) because the difference in implementation was so small. If that's a problem, I'll fix it.

danpovey:
that's fine.

PR #1782, opened 2017-07-31
Add CuMatrixBase::AddSmat()

kangshiyin:
Implement CuMatrixBase::AddSmat() .

All tests in `cudamatrix/` have passed.

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -251,6 +251,48 @@ static void _select_rows(const int* out_row_ptr, int* out_col_idx,
   }
 }
 
+// mat += alpha * smat
+//
+// We use warpSize threads per row to access only the nonzero elements.
+// Every CU1DBLOCK/warpSize rows share one thread block.
+// 1D grid to cover all rows of smat.
+template<typename Real>
+__global__
+static void _add_smat(Real* mat, MatrixDim mat_dim, Real alpha,
+                      const int* smat_row_ptr, const int* smat_col_idx,
+                      const Real* smat_val) {
+  const int i = blockIdx.x * blockDim.y + threadIdx.y; // row idx
+  if (i < mat_dim.rows) {
+    const int row_start = smat_row_ptr[i];
+    const int row_end = smat_row_ptr[i + 1];
+    for (int n = row_start + threadIdx.x; n < row_end; n += warpSize) {
+      const int j = smat_col_idx[n]; // col idx of smat
+      mat[i * mat_dim.stride + j] += alpha * smat_val[n];
+    }
+  }
+}
+
+// mat += alpha * smat^T
+//
+// We use warpSize threads per row to access only the nonzero elements.
+// Every CU1DBLOCK/warpSize rows share one thread block.
+// 1D grid to cover all rows of smat.
+template<typename Real>
+__global__
+static void _add_smat_trans(Real* mat, MatrixDim mat_dim, Real alpha,
+                            const int* smat_row_ptr, const int* smat_col_idx,
+                            const Real* smat_val) {
+  const int i = blockIdx.x * blockDim.y + threadIdx.y; // row idx
+  if (i < mat_dim.cols) {
+    const int row_start = smat_row_ptr[i];
+    const int row_end = smat_row_ptr[i + 1];
+    for (int n = row_start + threadIdx.x; n < row_end; n += warpSize) {
+      const int j = smat_col_idx[n]; // col idx of smat
+      mat[j * mat_dim.stride + i] += alpha * smat_val[n];
+    }
+  }
+}
+
 /// Fill the array 'data' with the sequence [base ... base + length)
 /// Use 1D block and 1D grid
 template<typename T>
@@ -5080,5 +5122,29 @@ void cudaF_select_rows(dim3 Gr, dim3 Bl, const int* out_row_ptr,
                        const int* in_col_idx, const float* in_val) {
   _select_rows<<<Gr, Bl>>>(out_row_ptr, out_col_idx, out_val, row_indexes,
                            num_selected_rows, in_row_ptr, in_col_idx, in_val);
+}
+void cudaD_add_smat(dim3 Gr, dim3 Bl, double* mat, MatrixDim mat_dim,
+                    double alpha, const int* smat_row_ptr,
+                    const int* smat_col_idx, const double* smat_val) {
+  _add_smat<<<Gr, Bl>>>(mat, mat_dim, alpha, smat_row_ptr, smat_col_idx,
+                        smat_val);
+}
+void cudaF_add_smat(dim3 Gr, dim3 Bl, float* mat, MatrixDim mat_dim,
+                    float alpha, const int* smat_row_ptr,
+                    const int* smat_col_idx, const float* smat_val) {
+  _add_smat<<<Gr, Bl>>>(mat, mat_dim, alpha, smat_row_ptr, smat_col_idx,
+                        smat_val);
+}
+void cudaD_add_smat_trans(dim3 Gr, dim3 Bl, double* mat, MatrixDim mat_dim,
+                          double alpha, const int* smat_row_ptr,
+                          const int* smat_col_idx, const double* smat_val) {
+  _add_smat_trans<<<Gr, Bl>>>(mat, mat_dim, alpha, smat_row_ptr, smat_col_idx,
+                              smat_val);
+}
+void cudaF_add_smat_trans(dim3 Gr, dim3 Bl, float* mat, MatrixDim mat_dim,
+                          float alpha, const int* smat_row_ptr,
+                          const int* smat_col_idx, const float* smat_val) {
+  _add_smat_trans<<<Gr, Bl>>>(mat, mat_dim, alpha, smat_row_ptr, smat_col_idx,
+                              smat_val);
 }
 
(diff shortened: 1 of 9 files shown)


src/cudamatrix/cu-matrix.cc
```diff
@@ -968,6 +969,46 @@ void CuMatrixBase<Real>::AddMat(Real alpha, const CuMatrixBase<Real>& A,
...
+      KALDI_ASSERT(NumRows() == A.NumCols());
+      KALDI_ASSERT(NumCols() == A.NumRows());
+    }
+
+    CuTimer tim;
+
+    // We use warpSize threads per row to access only the nonzero elements.
+    // Every CU1DBLOCK/warpSize rows share one thread block.
+    // 1D grid to cover all rows of A.
+    const int warpSize = 32;
```
danpovey: At this point in the code you seem to be using warpSize fixed to 32, but in the CUDA code you use just 'warpSize' which is not bound to this variable so as far as I can see might be dependent on the hardware.  Is this correct?

src/cudamatrix/cu-kernels.cu
```diff
@@ -251,6 +251,48 @@ static void _select_rows(const int* out_row_ptr, int* out_col_idx,
...
+// We use warpSize threads per row to access only the nonzero elements.
+// Every CU1DBLOCK/warpSize rows share one thread block.
+// 1D grid to cover all rows of smat.
+template<typename Real>
+__global__
+static void _add_smat(Real* mat, MatrixDim mat_dim, Real alpha,
+                      const int* smat_row_ptr, const int* smat_col_idx,
+                      const Real* smat_val) {
+  const int i = blockIdx.x * blockDim.y + threadIdx.y; // row idx
+  if (i < mat_dim.rows) {
```
danpovey: It's not clear to me that this code would be very efficient in the case where there's exactly one nonzero element per row, which is quite common in practice.  It seems like it would launch an excessive number of thread blocks.
 I'm concerned that for the regular neural net training, where we use this function, we could experience a slowdown.  
One option here is to call the function csr2coo() with a temporarily allocated piece of memory, and then make a kernel that uses the coo format, which will be much more efficient.  Then we can rely on the NVidia people having implemented csr2coo() in a smart way.
Let me know what you think.

danpovey:
It looks like there is no existing speed test for AddSmat().  It would be great if you could make one and use it to compare speed of this with the master version.  A little slowdown is OK but I don't want it to be huge.  The reason this is an issue, is that it's used in existing code (TraceMatSmat() is also used in existing code, and similar issues occur for that).

src/cudamatrix/cu-matrix.cc
kangshiyin: The built-in variable can only be accessed in .cu file, so I use a constant here. According to [this table](http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications), it is 32 for all hardware until now.

src/cudamatrix/cu-kernels.cu
kangshiyin: I think the speed is bounded by un-coalesced memory access to the dense matrix, so it might not change a lot with CSR format.  Additional conversion with csr2coo() also takes time, which can be as long as _add_smat() itself, so may not be a good idea.  The nnz in a sparse matrix is usually much less compared to the number of elements in a typical dense matrix, so the sparse-mat-related kernel functions seems not critical to the over-all speed.

I will add a speed test for TraceMatSmat(), to compare with the old implementation. It has a similar access pattern with AddSmat().

danpovey:
OK, thanks.  I'm OK to merge it the way you did it, if empiricall the speed
is not so different.

kangshiyin:
CSR format is slightly slower but I think it is OK.

```
Benchmark(Gflops)                                 dim  old    new   speedup
CuSparseMatrix::TraceMatSmat<double>               16  0.006  0.006 0.931x
CuSparseMatrix::TraceMatSmat<double>               32  0.025  0.023 0.926x
CuSparseMatrix::TraceMatSmat<double>               64  0.093  0.086 0.921x
CuSparseMatrix::TraceMatSmat<double>              128  0.292  0.258 0.885x
CuSparseMatrix::TraceMatSmat<double>              256  1.104  1.061 0.961x
CuSparseMatrix::TraceMatSmat<double>              512  3.643  3.496 0.960x
CuSparseMatrix::TraceMatSmat<double>             1024  8.108  8.226 1.015x
CuSparseMatrix::TraceMatSmat<double>[transposed]   16  0.006  0.006 0.932x
CuSparseMatrix::TraceMatSmat<double>[transposed]   32  0.025  0.024 0.939x
CuSparseMatrix::TraceMatSmat<double>[transposed]   64  0.093  0.086 0.922x
CuSparseMatrix::TraceMatSmat<double>[transposed]  128  0.273  0.259 0.950x
CuSparseMatrix::TraceMatSmat<double>[transposed]  256  1.086  1.069 0.985x
CuSparseMatrix::TraceMatSmat<double>[transposed]  512  3.643  3.523 0.967x
CuSparseMatrix::TraceMatSmat<double>[transposed] 1024  8.973  9.094 1.013x
CuSparseMatrix::TraceMatSmat<float>                16  0.007  0.006 0.962x
CuSparseMatrix::TraceMatSmat<float>                32  0.025  0.025 0.967x
CuSparseMatrix::TraceMatSmat<float>                64  0.096  0.092 0.967x
CuSparseMatrix::TraceMatSmat<float>               128  0.307  0.292 0.952x
CuSparseMatrix::TraceMatSmat<float>               256  1.155  1.122 0.971x
CuSparseMatrix::TraceMatSmat<float>               512  3.975  3.977 1.000x
CuSparseMatrix::TraceMatSmat<float>              1024 10.596 10.459 0.987x
CuSparseMatrix::TraceMatSmat<float>[transposed]    16  0.006  0.006 0.962x
CuSparseMatrix::TraceMatSmat<float>[transposed]    32  0.025  0.025 0.967x
CuSparseMatrix::TraceMatSmat<float>[transposed]    64  0.095  0.091 0.962x
CuSparseMatrix::TraceMatSmat<float>[transposed]   128  0.308  0.294 0.957x
CuSparseMatrix::TraceMatSmat<float>[transposed]   256  1.168  1.126 0.964x
CuSparseMatrix::TraceMatSmat<float>[transposed]   512  4.130  4.092 0.991x
CuSparseMatrix::TraceMatSmat<float>[transposed]  1024 11.364 11.063 0.973x
```

danpovey:
those differences are tiny-- fine with me.  But how dense were those sparse
matrices?  Test them also with matrices that have one element per row, as
that's a common case encountered in practice.

kangshiyin:
Changed from `zero_prob=0.8` to one element per row. The result is very similar.

```
Benchmark(Gflops)                                 dim  old    new   speedup
CuSparseMatrix::TraceMatSmat<double>               16  0.006  0.006 0.968
CuSparseMatrix::TraceMatSmat<double>               32  0.025  0.025 0.984
CuSparseMatrix::TraceMatSmat<double>               64  0.101  0.099 0.981
CuSparseMatrix::TraceMatSmat<double>              128  0.392  0.395 1.008
CuSparseMatrix::TraceMatSmat<double>              256  1.565  1.552 0.992
CuSparseMatrix::TraceMatSmat<double>              512  6.052  5.982 0.988
CuSparseMatrix::TraceMatSmat<double>             1024 22.860 22.395 0.980
CuSparseMatrix::TraceMatSmat<double>[transposed]   16  0.006  0.006 0.993
CuSparseMatrix::TraceMatSmat<double>[transposed]   32  0.026  0.025 0.987
CuSparseMatrix::TraceMatSmat<double>[transposed]   64  0.101  0.100 0.989
CuSparseMatrix::TraceMatSmat<double>[transposed]  128  0.398  0.390 0.981
CuSparseMatrix::TraceMatSmat<double>[transposed]  256  1.561  1.545 0.990
CuSparseMatrix::TraceMatSmat<double>[transposed]  512  6.070  5.995 0.988
CuSparseMatrix::TraceMatSmat<double>[transposed] 1024 22.830 22.336 0.978
CuSparseMatrix::TraceMatSmat<float>                16  0.006  0.006 0.981
CuSparseMatrix::TraceMatSmat<float>                32  0.025  0.025 0.982
CuSparseMatrix::TraceMatSmat<float>                64  0.102  0.100 0.977
CuSparseMatrix::TraceMatSmat<float>               128  0.407  0.397 0.977
CuSparseMatrix::TraceMatSmat<float>               256  1.596  1.565 0.980
CuSparseMatrix::TraceMatSmat<float>               512  6.190  6.057 0.979
CuSparseMatrix::TraceMatSmat<float>              1024 23.724 23.108 0.974
CuSparseMatrix::TraceMatSmat<float>[transposed]    16  0.006  0.006 0.967
CuSparseMatrix::TraceMatSmat<float>[transposed]    32  0.026  0.025 0.984
CuSparseMatrix::TraceMatSmat<float>[transposed]    64  0.102  0.100 0.980
CuSparseMatrix::TraceMatSmat<float>[transposed]   128  0.409  0.400 0.977
CuSparseMatrix::TraceMatSmat<float>[transposed]   256  1.598  1.553 0.971
CuSparseMatrix::TraceMatSmat<float>[transposed]   512  6.228  6.044 0.970
CuSparseMatrix::TraceMatSmat<float>[transposed]  1024 23.765 23.044 0.970
```

danpovey:
Thanks!   Merging.

PR #1783, opened 2017-07-31
Circumwent the windows correct name resolution issues

jtrmal:
VS 2017 have a problem figuring that KwsLexicographicArc  typedef'd as Arc is actually correct. This looks like the simplest and nicest fix

### DIFF
--- a/src/kwsbin/kws-search.cc
+++ b/src/kwsbin/kws-search.cc
@@ -264,7 +264,7 @@ int main(int argc, char *argv[]) {
       StateId state_id = siter.Value();
       for (MutableArcIterator<KwsLexicographicFst>
            aiter(&index, state_id); !aiter.Done(); aiter.Next()) {
-        Arc arc = aiter.Value();
+        KwsLexicographicArc arc = aiter.Value();
         // Skip the non-final arcs
         if (index.Final(arc.nextstate) == Weight::Zero())
           continue;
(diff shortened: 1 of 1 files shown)


danpovey:
I'll merge.  Who was looking after the travis?  We might have to tune it again.
Some of the tests appear to take many minutes, but when run locally take a fraction of a second...those tests create temporary files, and I wonder whether some file system weirdness or latency on Travis is the cause.  Perhaps reducing the num-jobs (-p) in the Makefile would help?  I don't know when we last tuned this.

jtrmal:
IIRC it was mostly @dogancan and partially me. I can try later this week if
Dogan does not have time.
Y.

dogancan:
Unfortunately I don't have time at the moment. We last tuned Travis when switching to c++11. At the time g++ was significantly slower, so we switched to clang-3.8. I think we also increased the number of parallel jobs since clang was using less memory and we could get faster builds that way.

kkm000:
I'll take a look on Monday when I'm back in town.

On 31 Jul 2017 3:59 p.m., "Daniel Povey" <notifications@github.com> wrote:

Merged #1783 <https://github.com/kaldi-asr/kaldi/pull/1783>.

—
You are receiving this because you are subscribed to this thread.
<https://github.com/kaldi-asr/kaldi/pull/1783#event-1186686906>, or mute
the thread

kkm000:
I'll see how far we can get with ccache--travis seem to have added the [support for keeping caches between builds](https://docs.travis-ci.com/user/caching/#ccache-cache).

jtrmal:
Cool, that might actually substantially
Y.

PR #1785, opened 2017-08-01
Fixed bugs in Multi-database English LVCSR recipe

xiaohui-zhang:


danpovey:
@jtrmal, could you please review and merge?

xiaohui-zhang:
The purpose is to map acronyms like u._c._l._a. to separate characters so that it could be correctly scored. It's used in local/score_sclite.sh. They should have been copied from the fisher_swbd recipe but Allen forgot to copy them. They are not identical, because there are different mapping rules of eval2000 and rt03. There are authors in header of the scripts.

jtrmal:
Yeah, ok, I know, but while you are at it, could you please add the standard header to the file, i.e. stating copyright and authors the way we ask all authors to do it?

xiaohui-zhang:
addressed! @jtrmal

jtrmal:
Thanks, merging

PR #1790, opened 2017-08-02
Update tedlium nnet3-CE-lstm recipe to use xconfig

hhadian:
Should I add $label_delay to the LDA layer too?
If this seems OK, I'll prepare a similar one for ASPIRE.


danpovey:
No, don't add label delay unless it was already present in the original setup (it should be obvious if it was).

egs/tedlium/s5_r2/local/nnet3/tuning/run_lstm_1a.sh
```diff
@@ -101,22 +97,32 @@ for f in $train_data_dir/feats.scp $train_ivector_dir/ivector_online.scp \
...
+  lstm_opts+=" non-recurrent-projection-dim=$non_recurrent_projection_dim"
+
+  mkdir -p $dir/configs
+  cat <<EOF > $dir/configs/network.xconfig
+  input dim=100 name=ivector
+  input dim=40 name=input
+
+  # please note that it is important to have input layer with the name=input
+  # as the layer immediately preceding the fixed-affine-layer to enable
+  # the use of short notation for the descriptor
```
danpovey: This looks plausible.
It's been a while and I can't be 100% sure it's the same as the old setup, without someone looking at the generated configs.  But I guess something working is better than something that's broken.

danpovey: For this configuration I think it won't be necessary to add label-delay to the fixed-affine-layer.

hhadian:
There is a `--label-delay $label_delay ` option in the original setup, so I added `output-delay=$label_delay` to the output layer (based on other LSTM xconfigs that I have seen before).

egs/tedlium/s5_r2/local/nnet3/tuning/run_lstm_1a.sh
danpovey: Actually, sorry, I changed my mind.  I can't be 100% sure without thinking about it carefully, and adding label-delay for the fixed-affine-layer is safer.

But since we're changing the script, I think it's a good idea to make it more up-to-date.

This means adding the options
`--egs.chunk-left-context-initial 0 --egs.chunk-right-context-final 0`
in both training and decoding.
The chunk-width is on the small side but let's not mess with that right now.
Change the num_chunk_per_minibatch from 100 to 128,64.

hhadian: I asked about label delay because I guess I had seen some warning/error before in another setup regarding label delay. Do you want me to run this command only to at least make sure it runs w/o problems?

hhadian: I ran it. It has a problem in check_model_context:
`Exception: Model specified in {0}/init.config requires greater context than the model specified in {0}/ref.config. This might be due to use of label-delay at the output in ref.config. Please use delay=$label_delay in the initial fixed-affine-layer of the network, to avoid this issue.`

hhadian:
Now it generates the configs w/o problems.

hhadian:
Should I update ASPIRE too?

egs/aspire/s5/local/chain/tuning/run_blstm_7b.sh
```diff
@@ -183,6 +209,8 @@ if [ $stage -le 12 ]; then
     --egs.chunk-width $chunk_width \
     --egs.chunk-left-context $chunk_left_context \
     --egs.chunk-right-context $chunk_right_context \
+    --egs.chunk-left-context-initial=0 \
```
danpovey: please add these options to prep_test_aspire.sh below and modify the script to pass them through. the defaults in the script can be -1.  If you don't do this, the decoding will be mismatched.

egs/aspire/s5/local/nnet3/run_lstm.sh
```diff
@@ -116,6 +119,8 @@ if [ $stage -le 8 ]; then
     --egs.chunk-width=$chunk_width \
     --egs.chunk-left-context=$chunk_left_context \
     --egs.chunk-right-context=$chunk_right_context \
+    --egs.chunk-left-context-initial=0 \
```
danpovey: again, pass these through to the testing script.

egs/aspire/s5/local/nnet3/run_tdnn.sh
```diff
@@ -32,17 +32,29 @@ dir=exp/nnet3/tdnn
...
+  # the use of short notation for the descriptor
+  fixed-affine-layer name=lda input=Append(-2,-1,0,1,2,ReplaceIndex(ivector, t, 0)) affine-transform-file=$dir/configs/lda.mat
+
+  # the first splicing is moved before the lda layer, so no splicing here
+  relu-renorm-layer name=tdnn1 dim=1248
+  relu-renorm-layer name=tdnn2 dim=1248 input=Append(-1,2)
+  relu-renorm-layer name=tdnn3 dim=1248 input=Append(-3,3)
+  relu-renorm-layer name=tdnn4 dim=1248 input=Append(-3,3)
+  relu-renorm-layer name=tdnn5 dim=1248 input=Append(-7,2)
+  relu-renorm-layer name=tdnn6 dim=1248
```
danpovey: I think you should remove this layer, for equivalence with the old setup.  However I'm not 100% sure of this.  If we have the old models to compare with, you could check the num-params.  @vijayaditya, do you have a run of this system?

vijayaditya:
I had to delete it during one of the recent cleanups. I no longer have it.

danpovey:
We have them at http://kaldi-asr.org/models.html, you could download them
from there.

egs/aspire/s5/local/nnet3/run_lstm.sh
hhadian: There is no decoding step in this recipe.

vijayaditya:
We never ran an LSTM recipe on ASpIRE. This script is called from

https://github.com/kaldi-asr/kaldi/blob/master/egs/aspire/s5/local/nnet3/run_blstm.sh

which has the decode step.

--Vijay

hhadian:
> We have them at http://kaldi-asr.org/models.html, you could download them
from there.

I did, but it was chain not CE. 
Anyway I generated the configs using both methods. It seems that we should have that last layer `  relu-renorm-layer name=tdnn6 dim=1248 ` because the old method generates configs where there is an affine component (with input_dim = output_dim) before the final affine component.

hhadian:
> We never ran an LSTM recipe on ASpIRE.

In that case, I guess I should merge run_blstm.sh and run_lstm.sh into a single script and rewrite the xconfig for BLSTM (because it is for LSTM now)?

vijayaditya:
Better just rename the script to run_lstm_train.sh. then create the
run_lstm.sh which calls it with appropriate arguments and has the decode
command.

hhadian:
The point is, when I change run_lstm.sh to use xconfig, it can be either BLSTM or LSTM. The current setup seems to support both BLSTM and LSTM, although if one wants to run LSTM, they should call the decode script manually [so I guess effectively it is only BLSTM].

vijayaditya:
Oh yeah.. I forgot the main point of this PR. Yes just keep the BLSTM
recipe.

PR #1791, opened 2017-08-03
Add CuMatrix::AddSmatMat() and unit test

kangshiyin:
All test in cudamatrix/ have passed.

### DIFF
--- a/src/cudamatrix/cu-matrix-test.cc
+++ b/src/cudamatrix/cu-matrix-test.cc
@@ -769,6 +769,44 @@ static void UnitTextCuMatrixAddMatSmat() {
 }
 
 template<typename Real>
+static void UnitTextCuMatrixAddSmatMat() {
+  for (int i = 0; i < 2; ++i) {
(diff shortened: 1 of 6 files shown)


danpovey:
I didn't realize when I sketched that out that it would be necessary to tranpose C twice, but I can't see an easy way around it now.    Let me know if there are any 'natural-looking' operations of this type that look like they would be more efficient to implement.
The decision to reverse the row-major vs. col-major format of matrices vs. CuBLAS was taken a long time ago by Karel, and by the time I first saw the code, too much had been built on top of it to make it a point that was even worthwhile to query. 
Anyway, I'll merge this.

kangshiyin:
I can not find a better way either.  My rough test shows that those 2 dense mat transpose ops takes 1/3 of the total time of the whole API, with the matrix size setting,

 ```
    int m = 10 + Rand() % 400;
    int k = 10 + Rand() % 600;
    int n = 10 + Rand() % 500;
```
repeating 200 times.

PR #1794, opened 2017-08-04
Change KALDI_LOG to KALDI_VLOG(1) in ComputeDerivedVars()

davidavdav:
In our production system, this logging is the only verbose part in normal operation, and it appears 
to be a bit spurious, it would be nicer to have this level of verbosity reduced.

### DIFF
--- a/src/ivector/ivector-extractor.cc
+++ b/src/ivector/ivector-extractor.cc
@@ -180,7 +180,7 @@ class IvectorExtractorComputeDerivedVarsClass {
 };
 
 void IvectorExtractor::ComputeDerivedVars() {
-  KALDI_LOG << "Computing derived variables for iVector extractor";
+  KALDI_VLOG(1) << "Computing derived variables for iVector extractor";
   gconsts_.Resize(NumGauss());
   for (int32 i = 0; i < NumGauss(); i++) {
     double var_logdet = -Sigma_inv_[i].LogPosDefDet();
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks for the PR, but I don't think I'll merge-- that message is useful in explaining a delay to the user when run directly from the command line (else they'd think the decoding was inherently slow).

davidavdav:
Mmm okay, sure, in our perception this takes hardly any time compared to
regular posterior computations, but I can't vouch for others.

Cheers,

---david ( 大卫)

Op 4 aug. 2017 20:54 schreef "Daniel Povey" <notifications@github.com>:

Closed #1794 <https://github.com/kaldi-asr/kaldi/pull/1794>.

—
You are receiving this because you authored the thread.
<https://github.com/kaldi-asr/kaldi/pull/1794#event-1193643682>, or mute
the thread

danpovey:
For a normal-size ivector extractor it takes many seconds, and it also uses
multiple threads by default I think.  If you have set up your production
system in such a way that this message is printed other than at startup,
then you are wasting a lot of CPU cycles.

davidavdav:
Yes, sure, for a proper speaker recognition setup this probably takes quite some time, but in our ASR setups, with i-vector dimensions of 100, it is fairly fast. Anyway, it is only two lines at the start up, so we can probably deal with it.  Or we will just rebase this patch for ourselves every time we sync with upstream Kaldi.

PR #1796, opened 2017-08-05
Add a simple implementation for EstimateAndWriteLanguageModel

hhadian:
It compiles but rnnlm-example-test fails. It says that the unigram distribution does not sum to 1.0.
It might be the ARPA I write but it seems fine AFAIK. Maybe it's a problem in ARPRA reading code.
I will look into it later.

### DIFF
--- a/src/rnnlm/rnnlm-test-utils.cc
+++ b/src/rnnlm/rnnlm-test-utils.cc
@@ -89,12 +89,10 @@ void EstimateAndWriteLanguageModel(
     int32 bos_symbol, int32 eos_symbol,
     std::ostream &os) {
 
-  std::vector<int32> counts(symbol_table.NumSymbols(), 0);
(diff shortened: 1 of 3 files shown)


src/rnnlm/rnnlm-test-utils.cc
```diff
@@ -82,6 +82,36 @@ void ConvertToInteger(
...
+    }
+  }
+
+  os << "\\data\\\n";
+  os << "ngram 1=" << counts.size() - 1 << "\n";  // not including <eps>
+  os << "\n";
+  os << "\1-grams:\n";
+
+  for (int32 i = 1; i < counts.size(); i++)
+    os << Log(1.0f * counts[i] / tot_count) << " "
```
danpovey: needs to be log base 10.

PR #1797, opened 2017-08-06
Add xent_regularize parameter to Aspire chain recipes

adrianbg:
Fixes a bug in `learning_rate_factor=$(echo "print 0.5/$xent_regularize" | python)`

### DIFF
--- a/egs/aspire/s5/local/chain/tuning/run_blstm_7b.sh
+++ b/egs/aspire/s5/local/chain/tuning/run_blstm_7b.sh
@@ -20,6 +20,7 @@ num_data_reps=3
 
 
 min_seg_len=
+xent_regularize=0.1
 chunk_width=150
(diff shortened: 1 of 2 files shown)


galv:
@adrianbg Just want to mention that it is helpful to include text like `Fixes a bug in learning_rate_factor=$(echo "print 0.5/$xent_regularize" | python)` in the git commit message itself, not just on github (github will automatically populate the textbox with all but the first line of the commit message). When I first saw an email about this commit, I was confused because the git commit message did not justify the change.

danpovey:
Hm.
I distinctly remember editing the commit message to clarify that it was a bug-fix, before merging, but now I see his original commit message is the one that is recorded.  Next time I squash-merge a commit, I'll check that my changes to the commit message take effect.

PR #1809, opened 2017-08-11
reverting the recent change of perturb_data_dir_speed_3way.sh

tomkocse:


danpovey:
The change was to fix a problem related to utterance/speaker sorting in switchboard, but caused problems in several other example scripts.  We're going to fix the switchoard problem by fixings its data-prep script to make speakers prefixes of the utterances.

PR #1814, opened 2017-08-13
remove the setting of PYTHONIOENCODING in prepare_rnnlm_data.sh

wantee:


scripts/rnnlm/choose_features.py
```diff
@@ -7,6 +7,7 @@
 import argparse
 import sys
 import math
+sys.stdout = open(1, 'w', encoding='utf-8', closefd=False)
```
danpovey: I'll merge this as I assume it's right, but can you explain why this is necessary and what the default behavior is?
I would have thought that UTF-8 is automatically compatible with/convertible-to raw bytes, and that sys.stdout would by default be outputting raw bytes.  But perhaps the issue is that it's locale dependent and it might try to convert to other encodings dependent on the locale?

wantee: I think sys.stdout would not output raw bytes. I ran some simple tests and found that, for python 3, it would output with encoded string, whose encoding is controlled by PYTHONIOENCODING environment variable or locale of the system, for python 2.7, you have to explicitly pass a encoded the string to it.
Actually, since I set the locale for my system to UTF-8, there is no problem when I run the script directly from shell, but the problem is in the path.sh, where there is a line with `export LC_ALL=C`, so it will always throw an error if I do not set the encoding for sys.stdout explicitly.

Issue #1813, opened 2017-08-12
RNNLM feature scales


danpovey:
@wantee, I'm having problems with instability in the sparse-word-feature version of RNNLM training.  I was able to work around the problem by scaling the constant feature down and by removing the length feature and the unigram feature, but I think we need a more general solution.

My proposal is that all the features be given a scaling factor (e.g. an extra field in features.txt), that the choose_features.py script be given an option --max-feature-rms (example default: 0.01), representing the maximum allowed root-mean-square value (i.e. uncentered standard deviation) of any feature; and that the scaling factors will all be 1.0 by default but then will be made smaller than 1.0 as needed to ensure that the RMS value of that feature not exceed the max-feature-rms.  

This will obviously require changes to various scripts.  Can you please work on the changes required to make this work, assuming you have time?
The only thing which might not be clear is what to do with the unigram-prob feature.  I recommend to subtract the negative-entopy from it, and then compute the scaling factor according to the --max-feature-rms limitation.

wantee:
OK. Will do.

danpovey:
addressed.

PR #1816, opened 2017-08-13
add scales for rnnlm features (#1813)

wantee:


wantee:
Dan, please have a look. I am not very sure whether this is what you wanted.

scripts/rnnlm/choose_features.py
```diff
@@ -131,16 +141,30 @@ def read_unigram_probs(unigram_probs_file):
...
-    print("{0}\tunigram\t{1}\t{2}".format(num_features, entropy, args.unigram_scale))
+    rms = 0.0
+    for idx, p in enumerate(unigram_probs):
+        if wordlist[idx] in SPECIAL_SYMBOLS:
+            continue
+        if p > 0.0:
+            rms += (p - entropy) ** 2
+    rms = math.sqrt(rms / vocab_size)
+
+    print("{0}\tunigram\t{1}\t{2}".format(num_features, entropy, get_feature_scale(rms)))
```
danpovey: I think this could do with some more explanation at this point.  But maybe later- I may merge this as-is.

scripts/rnnlm/validate_word_features.py
```diff
@@ -62,27 +67,21 @@
...
-            else:
-                if feat_id > max_feat_id:
-                    sys.exit(sys.argv[0] + ": Wrong feat_id: {0}.".format(line))
-                if int(feat_value) < 1:
-                    sys.exit(sys.argv[0] + ": Value of ngram feature must be >= 1: {0}.".format(line))
+            else: # all feat_value would be float
+                try:
+                    float(feat_value)
+                except ValueError:
+                    sys.exit(sys.argv[0] + ": Value of unigram feature should be a float number: {0}.".format(line))
```
danpovey: float number -> floating point number (or just "float")

scripts/rnnlm/choose_features.py
wantee: Actually I don't know how to properly explain it. Would you please add some comments on this later?

PR #1819, opened 2017-08-14
Enable Travis cache

### LINKED ISSUE #1805: travis compile or test times out
Opening a separate issue to track this, xref #1783.

I made some modifications to the script under travis to not compile fst binaries. That shaved a few minutes off it, but this is not an entire solution.

I tried to use ccache, which reduced the `make all` time from 6:25 (with ccache enabled and empty cache) to 0:47 (cache warm) on my machine, absent any source changes. I hoped with the cache warm the build will take a shorter time, but the catch 22 is travis does not save the cache directory when a run times out, so the next compile starts with the blank slate. I'll see if it is possible to force travis to push the cache explicitly after compile but before running tests.

Another thing I am going to try is skip linking the binaries (compile only). I can think of bugs that would slip through the holes this way, but there are potential much worse ones that would not be caught even by linking anyway.

Other ideas welcome.

/cc @dogancan @jtrmal

kkm000:
Using ccache to cache object files. Also, the work is split into two jobs: one compiles everything (but skips linking binaries), and another compiles and runs tests only in libraries.

I am getting as low as 5 minutes in the first job without changes in source files, around 25 in the second. But Travis is quite inconsistent in performance: One run took almost 40 minutes, although the cache hit ratio was nominal.

Most of the changes are related to quoting the variable `CXX` properly.

**Notable omissions:**
 * Compiling in single precision only. Old way was randomize on each run, but that would rub ccache's fur the wrong way. Do we need both? I'll add separate jobs for the other mode if need both.
* Not running the target `ext_test`. It adds a lot to compile time, partly because portaudio, but runs only 1 test in src/online2.

Closes #1805

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -7,7 +7,10 @@ CC = gcc         # used for sph2pipe
 
 # Note: OpenFst requires a relatively recent C++ compiler with C++11 support,
 # e.g. g++ >= 4.7, Apple clang >= 5.0 or LLVM clang >= 3.3.
-OPENFST_VERSION = 1.6.2
+OPENFST_VERSION ?= 1.6.2
+
+# Default features configured for OpenFST; can be overridden in the make command line.
+OPENFST_CONFIGURE ?= --enable-static --enable-shared --enable-far --enable-ngram-fsts
 
 OPENFST_VER_NUM := $(shell echo $(OPENFST_VERSION) | sed 's/\./ /g' | xargs printf "%d%02d%02d")
 ifeq ("$(shell expr $(OPENFST_VER_NUM) \< 10600)","1")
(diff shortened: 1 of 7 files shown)


tools/Makefile
```diff
@@ -87,7 +94,7 @@ sclite: sclite_compiled
 .PHONY: sclite_compiled
 sclite_compiled: sctk sctk_configured
 	cd sctk; \
-	$(MAKE) CC=$(CC) CXX=$(CXX) all && $(MAKE) install && $(MAKE) doc
+	$(MAKE) CC="$(CC)" CXX="$(CXX)" all && $(MAKE) install && $(MAKE) doc
```
danpovey: why do CC and CXX need to be quoted here?

kkm000: Generally, because they can be (and CXX is) a multi-token value. ccache wraps a compiler:
```
- CXX="ccache clang++-3.8 -Qunused-arguments -fcolor-diagnostics -Wno-tautological-compare"
```

danpovey:
I don't object to anything specific in it, but @jtrmal may want to have a look; and it's still timing out.

kkm000:
Yes, I just restarted it for the 3rd time. I did see it succeed though, and when caches warms up, timeouts will be less likely. But this is not a complete solution; tests should be able to finish with a cold cache.

I think the next step is to split testing between two or more jobs. I need to tally a successful run to understand what libs have the least interdependencies, and which of them have the longest running tests.

kkm000:
Huh. Travis is mind-boggling. The testing job timed out thrice, then on the fourth run completed in 15 minutes (and that's with a completely cold cache!). Test run times would be more those of a high-end CPU. So much variance that its hard to even plan splitting the test job into multiple.

jtrmal:
I think that was both mine and Dogan's experience as well.
Y.

kkm000:
If that continues to be a problem, and alternative is e. g. Jenkins on Amazon EC. But that requires some budget and, more importanly, some work, and I can think of the way to spend time better... Let's see if Travis is still an option with all its quirks and vagaries.

kkm000:
@jtrmal, PTAL if you have time. I am not pushing you, just pointing this is a complete change ready for a code review, and I think this is already better that what we already have, and will in any case improve the chances of a CI job running to completion. Travis pulls the master branch cache for pull requests when they do not have one on their own, then reuses this PR-specific cache if the PR branch is changed, so having a warm cache for the master branch will help.

jtrmal:
@danpovey LGTM, please merge.

kkm000:
Thanks!

danpovey:
Thanks!  Merging.

PR #1834, opened 2017-08-20
AddMatSmat, AddSmat optimization

sas91:
Function calls removed from the for loop continuation condition. "sparse-matrix-test" works fine with the change. For AddSmatMat, micro-optimizations has been performed in a previous commit "27851a29444803531fd027d318795f90110339c8" with blas functions by writing one row at a time in the result matrix. For similar optimization in AddMatSmat, I think as the sparse matrix is in the right side of the multiplication, one column can be written at a time but the matrix seems to be stored in a row-major order. By taking transpose blas functions may be used but that might not be efficient. Please let me know how to optimize AddMatSmat and AddSmat with blas.

### DIFF
--- a/src/matrix/kaldi-matrix.cc
+++ b/src/matrix/kaldi-matrix.cc
@@ -405,7 +405,7 @@ void MatrixBase<Real>::AddSmat(Real alpha, const SparseMatrix<Real> &A,
     KALDI_ASSERT(NumCols() == A.NumCols());
     MatrixIndexT a_num_rows = A.NumRows();
     for (MatrixIndexT i = 0; i < a_num_rows; ++i) {
-      const auto & row = A.Row(i);
+      const SparseVector<Real> &row = A.Row(i);
       MatrixIndexT num_elems = row.NumElements();
       for (MatrixIndexT id = 0; id < num_elems; ++id) {
         (*this)(i, row.GetElement(id).first) += alpha
@@ -417,7 +417,7 @@ void MatrixBase<Real>::AddSmat(Real alpha, const SparseMatrix<Real> &A,
     KALDI_ASSERT(NumCols() == A.NumRows());
     MatrixIndexT a_num_rows = A.NumRows();
     for (MatrixIndexT i = 0; i < a_num_rows; ++i) {
-      const auto & row = A.Row(i);
+      const SparseVector<Real> &row = A.Row(i);
       MatrixIndexT num_elems = row.NumElements();
       for (MatrixIndexT id = 0; id < num_elems; ++id) {
         (*this)(row.GetElement(id).first, i) += alpha
@@ -493,6 +493,7 @@ void MatrixBase<Real>::AddMatSmat(Real alpha, const MatrixBase<Real> &A,
     this->Scale(beta);
     MatrixIndexT b_num_rows = B.NumRows(),
         this_num_rows = this->NumRows();
+    // Iterate over the rows of sparse matrix B and columns of A.
     for (MatrixIndexT k = 0; k < b_num_rows; ++k) {
       const SparseVector<Real> &B_row_k = B.Row(k);
       MatrixIndexT num_elems = B_row_k.NumElements();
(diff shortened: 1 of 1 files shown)


src/matrix/kaldi-matrix.cc
```diff
@@ -487,11 +491,16 @@ void MatrixBase<Real>::AddMatSmat(Real alpha, const MatrixBase<Real> &A,
...
 
     Matrix<Real> buf(NumRows(), NumCols(), kSetZero);
-    for (int i = 0; i < NumRows(); ++i) {
-      for (int k = 0; k < B.NumRows(); ++k) {
-        for (int e = 0; e < B.Row(k).NumElements(); ++e) {
-          int j = B.Row(k).GetElement(e).first;
-          buf(i, j) += A(i, k) * B.Row(k).GetElement(e).second;
+    MatrixIndexT a_num_rows = A.NumRows();
+    MatrixIndexT b_num_rows = B.NumRows();
+    for (int i = 0; i < a_num_rows; ++i) {
```
danpovey: please replace all these ints with MatrixIndexT.

src/matrix/kaldi-matrix.cc
```diff
@@ -487,11 +491,16 @@ void MatrixBase<Real>::AddMatSmat(Real alpha, const MatrixBase<Real> &A,
...
+      for (int k = 0; k < b_num_rows; ++k) {
+        const SparseVector<Real> &B_row_k = B.Row(k);
+        MatrixIndexT num_elems = B_row_k.NumElements();
+        for (int e = 0; e < num_elems; ++e) {
+          const std::pair<MatrixIndexT, Real> &p = B_row_k.GetElement(e);
+          int j = p.first;
+          buf(i, j) += A(i, k) * p.second;
```
danpovey: Also, the most important optimization here is the use of cblas_xAxpy.  You should do that too.
Here, you can do it by first moving the loop over 'i' from the outside to the inside, and then, once you verify that it still works, turn the loop over 'i' into a call to cblas_xAxpy.  You might find it interesting to read the first chapter of the book 'Matrix computations' by golub et al, if you haven't already.  There is a copy on my desk (or just above the desk); you might find Yenda in our shared office and you can borrow it.

sas91:
Thank you Dr. Povey. I will borrow the book tomorrow and try the
optimization with "cblas_xAxpy".

*Thanks & Regards*,
Aswin

sas91:
Blas optimization added for "AddMatSmat". Test cases in "sparse-matrix-test" passed with the changes.

src/matrix/kaldi-matrix.cc
```diff
@@ -403,19 +403,23 @@ void MatrixBase<Real>::AddSmat(Real alpha, const SparseMatrix<Real> &A,
...
             * row.GetElement(id).second;
       }
     }
   } else {
     KALDI_ASSERT(NumRows() == A.NumCols());
     KALDI_ASSERT(NumCols() == A.NumRows());
-    for (int i = 0; i < A.NumRows(); ++i) {
+    MatrixIndexT a_num_rows = A.NumRows();
+    for (MatrixIndexT i = 0; i < a_num_rows; ++i) {
       const auto & row = A.Row(i);
```
danpovey: this is a pre-existing issue in the old code, but we prefer not to use 'auto' in cases like this; use the actual type.  And no space between & and row.

src/matrix/kaldi-matrix.cc
```diff
@@ -486,33 +490,47 @@ void MatrixBase<Real>::AddMatSmat(Real alpha, const MatrixBase<Real> &A,
...
+        this_num_rows = this->NumRows();
+    for (MatrixIndexT j = 0; j < b_num_rows; ++j) {
+      const SparseVector<Real> &B_row_j = B.Row(j);
+      MatrixIndexT num_elems = B_row_j.NumElements();
+      Real *this_col_j = this->Data() + j;
+      for (MatrixIndexT e = 0; e < num_elems; ++e) {
+        const std::pair<MatrixIndexT, Real> &p = B_row_j.GetElement(e);
+        MatrixIndexT k = p.first;
+        Real alpha_B_jk = alpha * p.second;
+	const Real *a_col_k = A.Data() + k;
```
danpovey: indent level

sas91:
the tests pass in cudamatrix/. Pushed with changes.

danpovey:
Thanks!  Merging.

PR #1839, opened 2017-08-23
Added dumping/non-dumping option and a pointer to copy decoded output on it to kaldi decoder.

MohamadIsmail:
added dumping option for decoded features and a pointer passed by user to copy information on it when you use class "DecodeUtteranceLatticeFaster".
before that, the user was forced to pass a valid path to a dump file to receive output on it. now you have an option to pass to its constructor either:
- "compact lattice writer"/"lattice writer" as NULL, dumping option as false and an instantiated pointer to receive output on it.
- or "compact lattice writer"/"lattice writer" as an absolute path to a dump file you want to create, dumping option as true and NULL pointer.

### DIFF
--- a/src/decoder/decoder-wrappers.cc
+++ b/src/decoder/decoder-wrappers.cc
@@ -44,7 +44,9 @@ DecodeUtteranceLatticeFasterClass::DecodeUtteranceLatticeFasterClass(
     int64 *frame_sum, // on success, adds #frames to this.
     int32 *num_done, // on success (including partial decode), increments this.
     int32 *num_err,  // on failure, increments this.
-    int32 *num_partial):  // If partial decode (final-state not reached), increments this.
+    int32 *num_partial ,  // If partial decode (final-state not reached), increments this.
+    CompactLattice* lat_out, // for storing decoded features. 
+    bool write):    // dumping option.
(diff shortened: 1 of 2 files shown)


danpovey:
thanks for the PR, but we don't merge things like this unless they are needed for something we are checking in; this extra code would not be called anywhere so it's unnecessary for the current project.

lordadamson:
Hi Dan,
Thanks for the reply. We use the decoder wrapper in our project and we do not wish to use file IO.
We're currently running a modified version of kaldi that has @MohamadIsmail patch.
I'm sure there are other projects that would need to avoid disk IO.
Is there an alternative that we're not aware of?
Thanks in advance.

danpovey:
That wrapper function is just a convenience function created in order to
avoid duplication of the same code in multiple different command-line
decoder binaries.  It's not really a core utility.  You can adapt the code
as you see fit for your purposes.

lordadamson:
Ah! I understand now. Thank a lot for the explanation.

On Aug 25, 2017 2:47 AM, Daniel Povey <notifications@github.com> wrote:
That wrapper function is just a convenience function created in order to
avoid duplication of the same code in multiple different command-line
decoder binaries. It's not really a core utility. You can adapt the code
as you see fit for your purposes.

PR #1848, opened 2017-08-29
Csj  scripts and results updated with tdnn+chain

rickychanhoyin:
with kaldi version in mid of august


egs/csj/s5/RESULTS
```diff
@@ -1,134 +1,212 @@
...
 %WER 11.15 [ 2901 / 26028, 297 ins, 697 del, 1907 sub ] exp/nnet3/nnet_tdnn/decode_eval1_csj/wer_15_0.0
 %WER 8.47 [ 2259 / 26661, 289 ins, 381 del, 1589 sub ] exp/nnet3/nnet_tdnn/decode_eval2_csj/wer_12_0.0
-%WER 9.21 [ 1583 / 17189, 240 ins, 258 del, 1085 sub ] exp/nnet3/nnet_tdnn/decode_eval3_csj/wer_12_0.0
\ No newline at end of file
+%WER 9.21 [ 1583 / 17189, 240 ins, 258 del, 1085 sub ] exp/nnet3/nnet_tdnn/decode_eval3_csj/wer_12_0.0
+
+
+
+##### These experiments use the below line in csj_data_prep.sh
+## cat $CSJ/*/*/*-wav.list 2>/dev/null | sort > $dir/wav.flist # Using All data
```
danpovey: I think it would be better if instead of relying on someone editing csj_data_prep.sh, you added an option to csj_data_prep.sh and put one version of the invocation of that script (hopefully the better version) in the  run.sh, and the other version commented out in the run.sh.

rickychanhoyin: updated

egs/csj/s5/local/csj_data_prep.sh
```diff
@@ -20,12 +20,19 @@
 set -e # exit on error
 
 #check existing directories
-if [ $# != 1 ]; then
-  echo "Usage: csj_data_prep.sh <csj-data dir>"
+if [ $# -ne 1 ] && [ $# -ne 2 ]; then
+  echo "Usage: csj_data_prep.sh <csj-data dir> [mode_number]"
+  echo " mode_number can be 0, 1, 2, 3, (0=default using academic lecture and other data, 1=using academic lecture data, 2=using all data except for dialog data, 3=using all data)"
```
danpovey: better to split this message over multiple lines; this is hard to read.

egs/csj/s5/local/csj_data_prep.sh
```diff
@@ -20,12 +20,19 @@
 set -e # exit on error
 
 #check existing directories
-if [ $# != 1 ]; then
-  echo "Usage: csj_data_prep.sh <csj-data dir>"
+if [ $# -ne 1 ] && [ $# -ne 2 ]; then
+  echo "Usage: csj_data_prep.sh <csj-data dir> [mode_number]"
```
danpovey: mode_number should also be in angle brackets, i.e. `[<mode_number>]`

egs/csj/s5/RESULTS
```diff
@@ -1,134 +1,212 @@
-## These are results using the third edition of CSJ.
-for eval_num in `seq 3`; do echo "=== evaluation set $eval_num ===" ;\
- for x in exp/{tri,dnn}*/decode_eval${eval_num}*; do [ -d $x ] && grep WER $x/wer_* | utils/best_wer.sh; done ; done
-
-## Results of using training set that contains "academic" and "other" speech data (default).
-## If you want to use "simulated public speaking" and "dialog" data, you should check the following script [local/csj_data_prep.sh line 44].
+##### These experiments use the below line in csj_data_prep.sh
+## cat $CSJ/*/{A,M}*/*-wav.list 2>/dev/null | sort > $dir/wav.flist # Using "Academic lecture" and "other" data
```
danpovey: Please change these comments to be consistent with how csj_data_prep.sh works now.

rickychanhoyin: ok

danpovey:
Actually I just had a thought while I was away from the computer.
Would you mind giving the "modes" names instead of numbers?  I think that
would be much clearer.

rickychanhoyin:
mu.... no idea which "modes" names would be appropriate

danpovey:
OK, leave it as is, I'll check to-morrow.

egs/csj/s5/local/chain/tuning/run_tdnn_7k.sh
```diff
@@ -0,0 +1,177 @@
...
+
+
+set -e
+
+# configs for 'chain'
+affix=
+stage=8
+train_stage=-10
+get_egs_stage=-10
+speed_perturb=false
```
danpovey: I think speed_perturb should default to true.  We always do it like this.  If you're concerned about speed you can reduce the number of epochs slightly.

egs/csj/s5/local/chain/tuning/run_tdnn_7k.sh
```diff
@@ -0,0 +1,177 @@
+#!/bin/bash
```
danpovey: please rename to run_tdnn_1a.sh and have the experimental directory be something like tdnn_1a_sp (assuming speed_perturb is true).

egs/csj/s5/run.sh
```diff
@@ -277,6 +283,9 @@ wait
 # Karel's DNN recipe on top of fMLLR features
 local/nnet/run_dnn.sh
```
danpovey: can you please move all the MMI training to something like local/run_mmi.sh and have it commented out in the top-level script, and also comment out local/nnet/run_dnn.sh?  (Once you add the speed perturbing, run_tdnn_chain.sh will be comfortably better than that result).

egs/csj/s5/local/chain/tuning/run_tdnn_7k.sh
```diff
@@ -0,0 +1,177 @@
...
+initial_effective_lrate=0.001
+final_effective_lrate=0.0001
+leftmost_questions_truncate=-1
+max_param_change=2.0
+final_layer_normalize_target=0.5
+num_jobs_initial=1
+num_jobs_final=1
+minibatch_size=128
+frames_per_eg=150
+remove_egs=false
```
danpovey: better to have remove_egs to default to true.  And for the frames_per_eg I recommend 150,140,100
for minibatch_size I recommend 128,64
(they are both comma-separated strings).
I think it would be best if you copied the current Swbd configuration, which is as follows:

```
 relu-batchnorm-layer name=tdnn1 dim=625                                                                                                       
  relu-batchnorm-layer name=tdnn2 input=Append(-1,0,1) dim=625                                                                                  
  relu-batchnorm-layer name=tdnn3 dim=625                                                                                                       
  relu-batchnorm-layer name=tdnn4 input=Append(-1,0,1) dim=625                                                                                  
  relu-batchnorm-layer name=tdnn5 dim=625                                                                                                       
  relu-batchnorm-layer name=tdnn6 input=Append(-3,0,3) dim=625                                                                                  
  relu-batchnorm-layer name=tdnn7 input=Append(-3,0,3) dim=625                                                                                  
  relu-batchnorm-layer name=tdnn8 input=Append(-3,0,3) dim=625                                                                                  
  relu-batchnorm-layer name=tdnn9 input=Append(-3,0,3) dim=625       
```

egs/csj/s5/local/chain/tuning/run_tdnn_7k.sh
```diff
@@ -0,0 +1,177 @@
...
+
+# training options
+num_epochs=4
+initial_effective_lrate=0.001
+final_effective_lrate=0.0001
+leftmost_questions_truncate=-1
+max_param_change=2.0
+final_layer_normalize_target=0.5
+num_jobs_initial=1
+num_jobs_final=1
```
danpovey: ... I guess you ran this somewhere where you had only one GPU?  If possible I'd increase num-jobs-{initial,final} to 3 and 10.  But you'll have to install GridEngine to run this. 
This will make it easier for others who have grids to reproduce your results quickly and reproducibly.

jtrmal:
btw, if there is anyone working on this recipe, it would be great if you
could fix the formatting (spaces instead of tabs) for the files. And I
don't mean the DNN scripts specifically nor exclusively -- I just remember
there was a lot of files with tabs in it in this recipe.
y.

danpovey:
Ricky, if you're too busy for this we can have someone local make the
changes I asked for.

LvHang:
@danpovey 
How are you thinking about merging the local/run_tdnn_chain.sh (which extracts ivector, call run_tdnn.sh and decoding) into local/chain/run_tdnn.sh. I think we always do it in this way. Leave local/run_tdnn_chain.sh and local/chian/run_tdnn.sh separately looks a little weird.
Or just leave it here?

danpovey:
You're right, it is odd-- do it the normal way, for consistency.
Unless you can find some good reason why it was done that way after looking
at the script carefully.

LvHang:
@danpovey 
Do we have the CSJ dataset on our grid?
I think maybe I need to test the code. His local/nnet3/run_ivector_common.sh code is too old. It doesn't support speed_perturb. That means, in his recipe, he didn't deal with the speed_perturb case at all.
So the "speed_perturb" option only can be "false". I'm rewriting it. So I hope to test.

danpovey:
OK, make sure to use a fairly modern run_ivector_common.sh.  the
mini_librispeech one is quite up to date.

jtrmal:
we have the corpora in /export/corpora5/CSJ/
@sw005320  can probably help if the format has changed -- I think he was saying the /export/corpora5/CSJ/Disk/ is what is of interest to us
y.

LvHang:
Yeah, I modify it according to swbd.
Do you know where can I find the CSJ dataset in our grid?

LvHang:
Ok, I see. Jtrmal show it.

jtrmal:
@LvHang , swbd has an obsolete version of the run_ivector_common.sh (I thnk), do not use that one.

LvHang:
@jtrmal , so tedlium is better? 
It use "utils/data/perturb_data_dir_speed_3way.sh data/${train_set} data/${train_set}_sp"

jtrmal:
I think mini_librispeech is the recent one -- not sure about tedlium
y.

danpovey:
It's not ideal, as the tedlium one uses LDA+MLLT for the ivector extractor,
we prefer PCA.

LvHang:
OK, get it. Thanks

rickychanhoyin:
Yes, I am busy in these days. Many thanks for LvHang, Dan and Jan following
on it.

Ricky

2017-11-20 10:07 GMT+08:00 Daniel Povey <notifications@github.com>:

> Ricky, if you're too busy for this we can have someone local make the
> changes I asked for.
>
>
>

LvHang:
@danpovey 
In new version run_ivector_common.sh, we always use 3way speed perturbation method. Do you think we need to leave a "speed_perturb" option for user to choose?

danpovey:
I don't think it's necessary to give an option, just have it done by
default.

danpovey:
... and remove the _sp suffix on the directory name in that case.

danpovey:
Addressed in #2035

PR #1850, opened 2017-08-29
fixed some issues for bottleneck feature extraction in run-4-anydecod…

pegahgh:
…e-lang.sh.


egs/babel_multilang/s5/run-4-anydecode-langs.sh
```diff
@@ -29,10 +28,14 @@ ivector_suffix=
 iter=final
 
 # params for extracting bn features
-multidir=exp/nnet3/multi_bnf_sp
+use_bnf=false # If true, bottleneck feature is extracted and appended as input
+              # for nnet3 model.
```
danpovey: you should probably clarify that if you set it to true, you have to set 'bnf_node_name'.
if you have a script that trains a bottleneck model, why not set bnf_node_name to the bottleneck node?

egs/babel_multilang/s5/run-4-anydecode-langs.sh
```diff
@@ -29,10 +28,14 @@ ivector_suffix=
...
+bnf_nnet_dir=exp/nnet3/multi_bnf_sp
+use_ivector_bnf=false # If true, ivector used in extracting bottleneck features.
+bnf_node_name= # Component node name correspond to bottleneck layer in bottleneck
+               # network. Output of this node is extracted
+               # as bottleneck features. i.e. tdnn_bn.renorm
```
danpovey: 'correspond' should be 'corresponding'.
'i.e.' should be 'e.g.'.

egs/babel_multilang/s5/run-4-anydecode-langs.sh
pegahgh: Do you mean to hardcode it in the script? 
People can use this script to extract feature for bottlenek layer after affine, nonlinearity or renorm layer. So bnf_node_name can be tdnn_bn.affine, tdnn_bn.renorm or tdnn_bn.relu.
Also people can use this script to extract non-bottleneck layer and probably do dimensionality reduction on that using PCA.

danpovey:
These 'run' scripts are supposed to be examples of what you can do; in
general they are not supposed to be general purpose scripts.  If it's
becoming a general-purpose script that you are supposed to invoke with
arguments, then there needs to be a higher-level script or README that
demonstrates the workflow, and it needs to be pointed to from that script.
I'm not familar enough with the BABEL script structure to know where that
might be.  @jtrmal might be able to comment.

jtrmal:
IMO  the better location for that script would be somewhere in the original
babel recipes directory -- otherwise, it will need to pull in too many of
other mess (scripts) just to make it function. Even now there are too many
scripts in local/ for the multilang example, that are not necessary from
the multilang POV. And, worse, it will complicate the setup for users --
and the multilang is very non-trivial to set up correctly even now.

y.

jtrmal:
To explain: The original script is a modified version of the "normal"
decoding script from the babel/s5d recipe. That script is doing a lot of
things, including setting-up keyword search for the given dataset
(typically dev set), segmenting the data, running parametrization and so
on. Essentially, it is a push-button script that does everything exactly as
the program requested. Because of this, it contains a lot of calls of
scripts from the babel local/ directory. So if Pegah decided to transplant
the script into multilang_babel, she will have to copy/symlink handful of
local/ scripts as well. I think it's not a trivial thing and will only
pollute the multilang/local directory and becloud which scripts are
important and which are not

So IMO it would be better to provide a script in babel/s5d based on
assumption that there is already trained system in ../../multilang/s5/
y.

danpovey:
@jtrmal, is this PR OK to merge?

jtrmal:
seems ok. I didn't test the new script, though. I assume Pegah did.
y.

PR #1863, opened 2017-08-31
add special_symbol_opts

wantee:


scripts/rnnlm/get_special_symbol_opts.py
```diff
@@ -0,0 +1,52 @@
...
+
+# Copyright  2017  Jian Wang
+# License: Apache 2.0.
+
+import os
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(description="This script checks whether the special symbols "
+                                 "appear in words.txt with expected values.",
```
danpovey: I just noticed this PR.
Can you please make the usage message of this script explain that it prints out
these options to its stdout, and explain what they look like?

wantee: Yes, done.

PR #1868, opened 2017-09-02
Ap17 olr

lindada-summer:


danpovey:
Thanks for the PR,
We do accept PRs for new datasets, but this PR seems to have too many extra files in.  It looks to me like you just copied over the WSJ scripts and added the entire directory.   You even have an NFS lock file checked in!  It needs to be much cleaner than this.  If you clean it up we will review it.  I'll leave it open for now though.

danpovey:
This PR seems to still have a bunch of stuff copied over from some other setup that are not necessary for it to run, and that probably haven't been run in this context.  I'm not going to mention specific files because it's still the bulk of the files.  Please look at each file individually, and if in doubt, delete it.  And after that, look and see if the PR still seems to make sense.

PR #1871, opened 2017-09-05
prevented a potential leak in DecodeUtteranceLatticeFasterClass::operator()()

lordadamson:


danpovey:
Actually I think that code was correct.  Look at the destructor of that class.

PR #1875, opened 2017-09-06
A new speaker verification system trainning method

cdw1994:
Speaker verification system with tandem feature using DNN frame level phoneme posterior probabilities.

This recipe replaces the standard MFCC trained GMM of the v1 recipe with a  tandem feature trained GMM. A time-delay deep neural network (TDNN) acoustic model is used to extract the frame level phoneme posterior probabilities. After log, PCA, the resulted low dimensional features are fused with MFCC at the feature level to get hybrid tandem feature. The recipe is based on http://www.isca-speech.org/archive/archive_papers/interspeech_2014/i14_1120.pdf. See run.sh for updated results.


danpovey:
@david-ryan-snyder, could you look at this and give your reaction?  I don't have time to scan it right now.

david-ryan-snyder:
@cdw1994, I think a more popular method involves training a DNN with a bottleneck layer, extracting BNFs from the DNN, and appending them to the MFCCs. Have you compared that method with what you're doing in the pull request (which appears to be applying PCA to the log posteriors)?

danpovey:
Also I believe we have a recipe checked in that uses regular MFCC features
with posteriors (for i-vector extraction) set to the state-level
posteriors.  Is your proposed method better than that?

On Wed, Sep 6, 2017 at 12:38 PM, david-ryan-snyder <notifications@github.com

david-ryan-snyder:
@danpovey, yes, this PR uses the same DNN trained in in that recipe, but uses the posteriors (after dimensionality reduction) as features instead of replacing the GMM posteriors. They did compare with that other approach, and reported slightly better results. In my (and I think, others) experience, BNFs+MFCCs perform about the same as using the DNN posteriors (I've also done similar things with the same DNN recipe). So, I'm inclined to believe that differences in performance between the two techniques isn't significant.

I think this PR will be useful if @cdw1994 is willing to refactor the scripts. In brief, my suggestion is to generalize the scripts so that it supports difference kinds of features extracted from DNN (e.g., BNFs or something after PCA). There are other issues that need to get addressed in the PR, e.g., it looks like log-posteriors (very large) are written to disk as features and some scripts are unnecessarily duplicated. My suggestions won't require much code.  I will come up with a more detailed proposal for @cdw1994.

cdw1994:
@david-ryan-snyder, it's my pleasure to refactor the scripts. Looking forward to your detailed proposal.

david-ryan-snyder:
Thanks! BTW, @cdw1994 did you get a chance to compare this with extracting BNFs or applying PCA to earlier layers of the DNN?

cdw1994:
@david-ryan-snyder We've train a 512-GMM on these two tandem features(BNF based and DNN/PCA based) with our code. We conducted our experiment on NIST SRE extended condition 5 (tel-to-tel) female part task. The results are as follows:

tandem feature | EER(%) | norm minDCF08 | norm minDCF10
----------------- | -------- | ------------------ |------------------
MFCC+Bottleneck | 1.53  | 0.087 | 0.185 |
MFCC+DNN/PCA | 1.40  | 0.076 | 0.179 |

It seems that DNN/PCA based tandem features are slightly better than BNF based tandem features. I also think that the differences in performance between the two techniques isn't significant.

I will try to compare these two tandem features in the script.

david-ryan-snyder:
Cool! Glad you did that experiment. I don't think you need to compare with the other tandem features in the script... If a user wants to do BNFs instead of DNN/PCA it should be clear what to change. My suggestion is to add some script to sid/nnet2 that extracts features from the DNN (e.g., compute_bnf.sh). There can be an option for applying transform-feats to support PCA before writing to disk. I'll elaborate more tomorrow.

david-ryan-snyder:
@cdw1994, I overestimated when I'd get a chance to respond to this. I will try to get back to you with a plan this week. Sorry for the delay.

david-ryan-snyder:
@cdw1994,

Here are some more detailed suggestions.  My goal is to make the scripts more general, enabling support for things like bottleneck features (in addition to your PPP features).  It will also improve disk usage; your current version writes big things like posteriors to disk. It's going to require some refactoring, but if you're willing to do it, I think it will come in handy for other things too.

**Computing PCA transform**

I recommend creating a script called `sid/nnet2/get_pca_transform.sh 
`
Instead dumping the posteriors directly to disk, the script will get the posteriors (or output of any layer for that matter) in memory, compute the PCA transform, and then write the matrix to disk. 

The following is an example of what I'm talking about. Of course, in your case, instead of splicing features you're going to do the nnet2 computation.
https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/online/nnet2/get_pca_transform.sh

Notice that the features in that script are also fairly high-dimensional after the splicing, but we never need to write it directly to disk. Also, note that we do things to make the computation more efficient, e.g., see how the variable `$max_utts` is used, and how the output of the splicing is subsampled before computing PCA. You can do something similar for your script. 

**Computing BNFs or "Pseudo" BNFs**

My suggestion is to create a script called sid/nnet2/compute_bnf.sh. By default, the script does the nnet2 forward computation and writes the output to the disk. However, you could provide an option e.g., --transform “<path>/transform.mat”.  If the option is present, it will apply the transform before writing out the DNN features. You can use the --transform option to support you PPP features, while simultaneously supporting bottleneck features.

**Changes to SID Scripts**

You should replace all of the scripts you created ending in _v3.sh (e.g., sid/extract_ivectors_v3.sh) with versions that end in _tandem.sh (e.g., sid/extract_ivectors_tandem.sh).

The tandem version will take in two feature directories, instead of just one. E.g., 
`data=$2` and `data_bnf=$3.` In your case, data_bnf will point to a directly of PPP features that have had PCA applied to them, but you could easily imagine them coming from a bottleneck layer of a DNN.  You’ll also have to create two feature streams, e.g., feats and feats_bnf. The feats stream can support the usual delta and CMVN options that the original scripts (e.g., sid/extract_ivectors.sh) supports. The `$feats_bnf` will do whatever is appropriate for the phonetic features (might be that it needs nothing beyond applying VAD). Both should include select-voiced-frames in their respective pipes. After creating the two feature streams, you can append them together using the binary paste-feats.

There are several advantages to doing it this way. Firstly, you never have to write the pasted feats to disk. You only need to write the PPP and the MFCCs. Then, this script can compute the deltas and append the feats together in memory. Another advantage is that this script can also support things like BNFs.

cdw1994:
@david-ryan-snyder    Very thanks for your suggestions, I will try to refactor it this week.

danpovey:
Should this be closed?  No progress for a long time.

david-ryan-snyder:
I think it should be closed, yes. 

This PR is (essentially) about using bottleneck features (from an ASR DNN) in an i-vector system. It would be nice to have support for this, but I'd prefer it was done differently. For one, it should be using nnet3 now, not nnet2. Also, cumbersome systems like this are losing favor now that we have good DNN embedding-based systems. So, I don't think it's urgent that we continue developing this, but it would be nice to have.

Issue #1592, opened 2017-04-28
Failures when python is python3


psmit:
At this moment kaldi assumes that python is actually python2. Most distro's have this indeed configured this way, but I have at least 2 cases that this doesn't hold

1) Arch linux 
2) When using a module system and both python2 and python3 versions are loaded, sometimes this causes python to be linked to python2.

I think the best solution would be to write explicitly `#!/usr/bin/env python2` instead of `#!/usr/bin/env python3`. Afaik this would work in all situations correctly. Would you be open to a PR that implements this change?

jtrmal:
I think that could work. 
https://www.python.org/dev/peps/pep-0394/

I'm not sure, however, if all distributions and other OS (Mac?) do support python2 link.  Also, according to the PEP, it's actually Arch that is causing the incompatibility, so I don't think we should spend too much of our time trying to deal with that.

dresen:
I'm in the middle of installing Kaldi on a 2015 macbook pro and the solution with bash alias'ing python to python2.7 doesn't work - extras/check_dependencies.sh still reports version 3.*

jtrmal:
@dresen -- I think that's because bash by default does not expand aliases when in non-interactive mode. I think 
shopt -s expand_aliases
might help

jtrmal:
btw, when I'm thinking about it -- expanding the aliases could have a lot of negative results. Perhaps it would be better to define a shell function
```
function python() {
  python2 "$@"
}
```
or something like that -- I don't have time to test it out now.

danpovey:
python2 is not defined on my mac.
I believe the Python community has accepted that python3 is essentially a
separate language for compatibility purposes, and that python2 will always
be invoked as "python", and python3 as "python3".

In my opinion Arch linux has painted itself into a corner by this choice,
because it's not compatible with the rest of the world.


Dan

psmit:
I disagree Dan. In the linked PEP is written that this choice will be revisited every couple years. Also, this PEP came only after Arch changed it's default python. There was no python community opinion before they did that; the community saw the effects and created this PEP as a reaction.

Anyway, the advise of this PEP is to use python2 in the shebang. Could you maybe look why your mac does not have python2 defined; the PEP indicates that the mac installers have been updated to provide that.

danpovey:
Hm.

The PEP does seem to say that python2 is the preferred way to invoke
python.  However, since my mac doesn't have python2 installed, and I think
[...]

PR #1876, opened 2017-09-08
More thorough python dependency checking

psmit:
After extras/check_dependencies.sh is now run, the following is ensured:

- python2.7 is installed and the binary python2 exists (if not, a link to python2.7 is added)
- python3 is installed
- python links to python2.7. If it is not the case a symlink is added that is placed early on the path

This is basically the first step in resolving #1592 and does the basic checking for python3 as it is already used in some places (I saw it e.g. mentioned in #1707)

This has been tested on our linux machines and clusters. I do not have access to a Mac, so it would need to be verified to be working there.

### DIFF
--- a/tools/extras/check_dependencies.sh
+++ b/tools/extras/check_dependencies.sh
@@ -105,16 +105,20 @@ fi
 [ -f ./env.sh ] && . ./env.sh
 if $pythonok && ! which python2 >&/dev/null; then
   mkdir -p $PWD/python
-  echo "$0: python2.7 is installed, but the python2 binary does not exist. Adding this to tools/env.sh"
+  echo "$0: python2.7 is installed, but the python2 binary does not exist. Creating a symlink and adding this to tools/env.sh"
   ln -s $(which python2.7) $PWD/python/python2
   echo "export PATH=$PWD/python:\${PATH}" >> env.sh
 fi
 
-if $pythonok && which python >&/dev/null && [[ ! -f $PWD/python/stubborn ]]; then
+if [[ -f $PWD/python/.use_default_python && -f $PWD/python/python ]]; then
+  rm $PWD/python/python 
+fi
+
+if $pythonok && which python >&/dev/null && [[ ! -f $PWD/python/.use_default_python ]]; then
   version=`python 2>&1 --version | awk '{print $2}' `
   if [[ $version != "2.7"* ]] ; then
     echo "$0: WARNING python 2.7 is not the default python. We fixed this by adding a correct symlink more prominently on the path."
-    echo "$0: If you really want to use python3 as default, remove $PWD/python/python and add an empty file $PWD/python/stubborn"  
+    echo "$0: If you really want to use python $version as default, add an empty file $PWD/python/.use_default_python and run this script again."  
     mkdir -p $PWD/python
     ln -s $(which python2.7) $PWD/python/python
     echo "export PATH=$PWD/python:\${PATH}" >> env.sh
--- a/tools/extras/check_dependencies.sh
+++ b/tools/extras/check_dependencies.sh
@@ -87,28 +87,40 @@ if ! which awk >&/dev/null; then
   add_packages gawk gawk gawk
 fi
 
-if which python >&/dev/null ; then
+pythonok=true
+if ! which python2.7 >&/dev/null; then
+  echo "$0: python2.7 is not installed"
+  add_packages python2.7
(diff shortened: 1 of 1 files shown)


danpovey:
Seems reasonable but @jtrmal should check as it interacts with env.sh.

tools/extras/check_dependencies.sh
```diff
@@ -87,28 +87,40 @@ if ! which awk >&/dev/null; then
...
+fi
+
+( 
+#Use a subshell so that sourcing env.sh does not have an influence on the rest of the script
+[ -f ./env.sh ] && . ./env.sh
+if $pythonok && ! which python2 >&/dev/null; then
+  mkdir -p $PWD/python
+  echo "$0: python2.7 is installed, but the python2 binary does not exist. Adding this to tools/env.sh"
+  ln -s $(which python2.7) $PWD/python/python2
+  echo "export PATH=$PWD/python:\${PATH}" >> env.sh
```
jtrmal: you are creating python2 link, but we usually use shebang python
it would be OK I guess in case when you would be working on PR that changes the shebangs to python2 -- without that I'm not sure it would fix any issue

tools/extras/check_dependencies.sh
```diff
@@ -87,28 +87,40 @@ if ! which awk >&/dev/null; then
...
+if ! which python2.7 >&/dev/null; then
+  echo "$0: python2.7 is not installed"
+  add_packages python2.7
+  pythonok=false
+fi
+
+if ! which python3 >&/dev/null; then
+  echo "$0: python3 is not installed"
+  add_packages python3
+  pythonok=false
```
jtrmal: I see you are adding requirement for python3 -- is this necessary? I remeber @danpovey has a couple of scripts for LM stuff, but those were not merged yet.

tools/extras/check_dependencies.sh
```diff
@@ -87,28 +87,40 @@ if ! which awk >&/dev/null; then
...
-else
-  if which python2.7 >&/dev/null  || which python2 >&/dev/null ; then
-    echo "$0: python 2.7 is not the default python. You should either make it"
-    echo "$0: default or create an bash alias for kaldi scripts to run correctly"
-    status=1
-  else
-    echo "$0: python is not installed (we need python 2.7)"
-    add_packages python2.7 python python2.7
+    echo "$0: WARNING python 2.7 is not the default python. We fixed this by adding a correct symlink more prominently on the path."
+    echo "$0: If you really want to use python3 as default, remove $PWD/python/python and add an empty file $PWD/python/stubborn"
```
jtrmal: this is a confusing message imho.
creating that stubborn file does not really imply using python3 -- it implies using whatever is linked to "python"

danpovey:
Regarding the requirement for python3-- it's true that we don't require it
yet, but we will soon, and I'd rather not have to rewrite the dependency
script at that time, so it's OK with me to add this dependency now.


Dan

tools/extras/check_dependencies.sh
psmit: Yes, my thought has been that we switch to use python2 shebangs, either at once or gradually. I just did not want to include it in this same PR, but that would be coming.

tools/extras/check_dependencies.sh
psmit: You are absolutely right. I created the "stubborn" hook for some testing locally. I'm thinking I could remove that, as those that would want to be stubborn can just edit this script themselves. Does that sound better?

jtrmal:
I don't personally object to the idea - but naming it ".use_default_python"
sound better to me.
Also please verify what happens when you run the depends script once,
 create the file afterwards and then run it again - in that case it
shouldn't fail and perhaps the links should be removed - maybe it does that
already, I didn't check.
Or the message should inform the user about the necessity of removing the
links manually

Y.

jtrmal:
Sorry -- the previous reaction should go to the "stubborn" thread -- Seems like github does not handle nicely when replying via e-mail.

psmit:
There was only the removal message on the first time running the script. I now updated it to actually removing the symlink when python/.use_default_python is present.

I tested all scenarios I could think of for the python==python3.x case. I do not give a message when .use_default_python exists, but I'm not certain that is necessary?

The only thing "untested" in real situations is the "/usr/bin/python2 doesn't exist" case. Can someone else test that?

jtrmal:
LGTM

PR #1879, opened 2017-09-11
utilizing smart pointers in the decoder-wrappers

lordadamson:
We've been discussing the ability to utilize smart pointers in the forums. Here's my attempt to a high level convenience class without breaking compatibility.

### DIFF
--- a/src/decoder/decoder-wrappers.cc
+++ b/src/decoder/decoder-wrappers.cc
@@ -21,6 +21,8 @@
 #include "decoder/faster-decoder.h"
 #include "lat/lattice-functions.h"
 
+using namespace std;
+
 namespace kaldi {
 
 
@@ -55,8 +57,7 @@ DecodeUtteranceLatticeFasterClass::DecodeUtteranceLatticeFasterClass(
     like_sum_(like_sum), frame_sum_(frame_sum),
     num_done_(num_done), num_err_(num_err),
     num_partial_(num_partial),
-    computed_(false), success_(false), partial_(false),
-    clat_(NULL), lat_(NULL) { }
+	computed_(false), success_(false), partial_(false) { }
 
 
 void DecodeUtteranceLatticeFasterClass::operator () () {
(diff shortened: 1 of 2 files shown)


jtrmal:
Please note you've broken the formatting of the file and I believe "using
namespace something" is generally discouraged - it should be either
explicit import (using sts::unique_ptr) or just add explicitly the std::
prefix whenever you need the given class.
No need to fix that now, as I assume it's on KY a proof of concept.
Y.

lordadamson:
Hi @jtrmal 

Thanks for the feedback.

How else did I break the format of the file?

And regarding the namespace thing: The way I go about the namespace business is that I never do a `using namespace XYZ;` in a .h file because it could propagate to other cpp files including it, which can often be undesirable. However in cpp files I find no problem using it because it can't propagate beyond that particular source file.
So if you could shed some light on what I'm missing here. And of course if that is a convention that kaldi uses throughout the code I'll definitely respect and abide by regardless of my opinions.

lordadamson:
Hi @jtrmal 

Thanks for the feedback.

How did I break the format of the file?

jtrmal:
I see different indentation in the diff from my cell phone -- maybe it
looks fine when looking at normal (I..e not mobile) website?
Y.

lordadamson:
I guess I fixed it. Can you look at it now?

jtrmal:
see here:
https://google.github.io/styleguide/cppguide.html#Namespace_Names

I think so far in kaldi we are fairly benevolent about this in executables
codes but I think we should be stricter about that in libs.
I'd be happier to see std::unique_ptr and std::shared_ptr  explicitly in
the code but that might be just my personal taste.
Anyway, I will let Dan decide (and it's not important anyway for now).
y.

danpovey:
Hm.
I can see that it's safer, but I don't think it does much for the clarity of the code, especially for users not familiar with C++11 features.

lordadamson:
Alright then I'll close the issue now

PR #1882, opened 2017-09-12
WIP: Heroico Recipe

johnjosephmorgan:
Heroico Recipe


jtrmal:
John, I've just noticed that the recipe is in egs/heroico/ -- can you please move it  to egs/heroico/s5/ --actually, maybe even egs/heroico_spanish/s5/ would be better but that I will leave to you -- heroico is probably just fine.

johnjosephmorgan:
DONE. I'd rather stick with just heroic, but I could be convinced to change.
J

jtrmal:
It's OK, let's stick with heroico. I'll try to review  tomorrow. I have
noticed that some of the files you wrote do not have (c) and license, can
you please add those? Another thing I have noticed is that you actually
process files and store them on disk, which is not that effective.
Typically, we create wav.scp that contains the command that has to be used
to process the audio on-the-fly, the format of wav.scp is very simple:
Utt-name command-line-dumping-audio-to-stdout |

For example :
Utt sox wav.in - t wav - r 8000 - |

Which essentially just downsample the file. The pipe character at the end
is very important.

But actually, I might not be right, I didn't check the code too carefully
yet.
Thanks.
Y.

johnjosephmorgan:
Hi Yenda,
I've tried to write a script that uses a command line in the wav.scp file.
Here is the problem:
The wav files in the corpus have the form:
PATH/SPEAKER/UTTERANCE.wav
The SPEAKER is not a prefix of the UTTERANCE.wav file name.
The utt2spk and spk2utt do not get sorted correctly.

How do I get around this without  copying the wav files to files with new speaker-prefixed names.
John

jtrmal:
you could call the entries in wav.scp using the pattern

SPEAKER_UTTERANCE  sox PATH/SPEAKER/UTTERANCE.wav -r 16000 -t wav - |

segments file then can contain
SPEAKER_UTTERANCE_001 SPEAKER_UTTERANCE start-time end-time

and utt2spk then
SPEAKER_UTTERANCE_001 SPEAKER

does that make sense?
y.

johnjosephmorgan:
I think I got this working. It should be pushed to the repo.
I changed the names of the scripts from *copy_wav_files.pl to *make_lists.pl.
Thanks Yenda. That's pretty cool how that works.
John

egs/heroico/s5/local/heroico_answers_copy_wav_files.pl
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/perl -w
```
jtrmal: is this file used? If not, can you please remove it (and the other files not used as well)?

egs/heroico/s5/local/heroico_answers_make_lists.pl
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/perl -w
```
jtrmal: We prefer to use /usr/bin/env perl  shebang, i.e.
```
#!/usr/bin/env perl
```

env typically allows only one parameter, so you cannot supply -w switch, but you are using `use warnings` anyway, so it should not be necessry.

egs/heroico/s5/local/score_mbr.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
jtrmal: this file is probably not needed -- the score_kaldi.sh can do mbr scoring as well (use the --use-mbr true switch, if needed)

egs/heroico/s5/path.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
jtrmal: for path.sh, I'd suggest using the standard path.sh such as the one that can be found in the ws recipe + add modifications you need.

egs/heroico/s5/local/score_combine.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
jtrmal: this could be probably deleted too, you don't seem to use the lattice-combination anywhere

egs/heroico/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/bash
```
jtrmal: I have just noticed this file -- what recipe did you use as the starting code?
We do these things now more effectively -- unless it's a lot of modifications, could you use 
mini_librispeech/s5/local/nnet3/run_ivector_common.sh instead of this?
We didn't upgrade the older recipes, yet, but we are trying to make the new ones to use that.

egs/heroico/s5/local/prepare_dict.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash -u
+
+# Copyright 2017 John Morgan
+# Apache 2.0.
+
+set -o errexit
+
+[ -f path.sh ] && . path.sh  # Sets the PATH to contain necessary executables
```
jtrmal: this should be probably written as 
```
[ -f ./path.sh ] && . ./path.sh
```
otherwise the first path.sh on the PATH would be sourced.

egs/heroico/s5/local/prepare_lm.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Copyright 2017 John Morgan
+# Apache 2.0.
+
+. cmd.sh
```
jtrmal: this should be probably also 
```
. ./cmd.sh
```

egs/heroico/s5/README.md
```diff
@@ -0,0 +1,28 @@
+# heroico
+Scripts  for Kaldi recipe for Spanish using the Heroico corpus.
+
+The Heroico corpus includes a set of recordings from nonnative speakers.
+
+WER Results:
```
jtrmal: Hi John, could you please split this file into two -- README and RESULTS, please?

egs/heroico/s5/cmd.sh
```diff
@@ -0,0 +1,23 @@
...
+# queue names and different ways of specifying things like memory;
+# to account for these differences you can create and edit the file
+# conf/queue.conf to match your queue's configuration.  Search for
+# conf/queue.conf in http://kaldi-asr.org/doc/queue.html for more information,
+# or search for the string 'default_config' in utils/queue.pl or utils/slurm.pl.
+
+export train_cmd=queue.pl
+export decode_cmd="queue.pl --mem 2G"
+# the use of cuda_cmd is deprecated, used only in 'nnet1',
+export cuda_cmd="queue.pl --gpu 1"
```
jtrmal: I'd actually recomend to remove the cuda_cmd altogether if you are not using it.

egs/heroico/s5/local/prepare_dict.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash -u
+
+# Copyright 2017 John Morgan
+# Apache 2.0.
+
+set -o errexit
+
+[ -f ./path.sh ] && . ./path.sh
+
+wget http://www.openslr.org/resources/34/santiago.tar.gz
```
jtrmal: I think it should be better 
```
if [ ! -f ./santiago.tar.gz ];  then
  wget ...
fi
```
You could also look at the wget's `-O ` switch -- probably it could be used to download the file directly into the destination directory.
Also, I'm mildly leaning towards downloading the file in run.sh, not here.

egs/heroico/s5/run.sh
```diff
@@ -0,0 +1,451 @@
...
+fi
+
+if [ $stage -le 3 ]; then
+    # prepare lm on training transcripts
+    local/prepare_lm.sh
+
+    # make lm fst
+    mkdir -p data/local/lm
+
+    gunzip language_models/lm_threegram.arpa.gz
```
jtrmal: there is a script utils/format_lm.sh that handles the whole conversion process -- unless you need something very specific, I think it would be better to rely on that script.

egs/heroico/s5/run.sh
```diff
@@ -0,0 +1,451 @@
...
+	    exp/make_mfcc\
+	    mfcc || exit 1;
+
+	utils/fix_data_dir.sh \
+	    data/$fld || exit 1;
+	done
+fi
+
+if [ $stage -le 5 ]; then
+        echo "monophone training"
```
jtrmal: indentation

egs/heroico/s5/run.sh
jtrmal: I mean conversion ARPA->G.fst

egs/heroico/s5/run.sh
```diff
@@ -0,0 +1,451 @@
...
+	exp/tri3b \
+	exp/tri3b_ali
+fi
+
+if [ $stage -le 8 ]; then
+    # i-vector  extractor training
+    local/nnet3/run_ivector_common.sh
+
+    # train and test chain models
+    local/chain/run_tdnn.sh
```
jtrmal: the sequence of the two commands will y probably cause the run_ivector_common.sh to run twice. you can just call  
```
 local/chain/run_tdnn.sh
```

or other option that might be more "educative" is to run it like this
```
local/nnet3/run_ivector_common.sh
# we need to call the run_tdnn.sh with the stage param so that 
# the ivector script won't get run twice.
local/chain/run_tdnn.sh --stage 10
```

johnjosephmorgan:
Yenda,
I did not understand this message about arpa -> G.fst.
Thanks,
John

jtrmal:
Oh, I was just trying to emphasize that the whole conversion process from
arpa format LM to the G graph can be done just using the format_lm script.
Which you did figure out:)
y.

johnjosephmorgan:
ERROR (nnet3-chain-train[5.2.62~1-a2342]:MallocPitchInternal():cu-allocator.cc:114) Cannot allocate the requested memory (5696 x 12032 = 68534272
bytes)
Yenda, I got the above error message when training the chain models. I made the change to the num-chunk-per-minibatch option. I reduced the numbers to 64,32,16 and the training is now running . 
j

jtrmal:
What gpu do you use?
y.

johnjosephmorgan:
Tesla k20.
I had similar problems a couple of weeks ago. My sysadmin  found
problems with memory sticks. We removed those memory sticks. I'm not
sure this has anything to do with the current memory problem.
J

jtrmal:
I'm not sure either, but if you are running multiple jobs at the same GPU
(i.e. the GPU is not in exclusive mode), that might be explaining the
memory issues.
y.

johnjosephmorgan:
Is exclusive mode something I have to set?

danpovey:
nvidia-smi -c 3

The code will print a warning if it is not set that way.

jtrmal:
yes (using nvidia-smi -c ... ), but it will also make things complicated if
you are running the recipe via run.pl and not queue.pl

johnjosephmorgan:
My sysadmin set the GPU to exclusive process compute mode. I
apparently killed my job :(
Here is his message:
Previously, the compute mode was set to Default.  The options are Default,
Exclusive Process, Exclusive Thread and Prohibited.  Prohibited means no
jobs can run on that device.  Default means any jobs can run on that device.
Exclusive Thread means only one thread can run at a time.  Exclusive Process
means only one job can run, but it can be multi-threaded.

I'm still not sure this is the problem though.
I guess I should try again with the minibatch chunk set to 256,128,64?

danpovey:
That likely was the problem.
However when exclusive mode is set, you have to be careful to set
--num-jobs-final to no more than the number of GPUs on your machine.

johnjosephmorgan:
Thanks Dan, Yes, it looks like that was the problem.
I set initial and final number of jobs to 1 and the mini batch sizes
back to 256,128,64 and the training is up and running again.
I noticed that the number of iterations jumped up to 140 from 40.
John

johnjosephmorgan:
Yenda,
The recipe  runs end to end on my single node computer with run.pl

I am trying to run the recipe on the clsp cluster with queue.pl.
This is failing  at the estimate pca transform step.

I am using a qsub command line like:
qsub -cwd -V -o out_chain.log -e err_chain.log -l
"hostname=b06,mem_free=12g,ram_free=12g" -q g.q ./run.sh --stage 9

johnjosephmorgan:
I am also using the gpu=1 flag.

danpovey:
You are not supposed to submit the top-level run.sh to the queue.  You
should run it from the shell-- but first edit cmd.sh so that the desired
command is used to run jobs.

johnjosephmorgan:
Dan and Yenda,
My run.sh script runs end2end on my GPU  using run.pl. I have not succeeded in running end2end on the clsp cluster using queue.pl. 
It crashed at the pca estimation step.
I got passed this step by inserting -l hostname=b03 in  the decode_cmd and train_cmd in cmd.sh.

jtrmal:
Can you send me e-mail directly with location of the system? I'll have a
look what was the issue. It _should_ work
y.

jtrmal:
@johnjosephmorgan I have run the recipe successfully on the CLSP cluster and did some modifications to it. You can include the modifications by doing
```
git add remote jtrmal https://github.com/jtrmal/kaldi.git
git cherry-pick 30d339780b163e36b08ac78371233f4f198a6e7e
git push
```
I just tested it and it worked, so hopefully it will work for you as well

After that, I think it can be merged

jtrmal:
BTW, I just updated the command sequence (make sure you call the fetch --all)
y.

johnjosephmorgan:
Did you mean git remote add in the first command?
J

jtrmal:
yes, "remote add", sorry
y.

danpovey:
We often mirror things on openslr.org to reduce problems from flaky
servers.

johnjosephmorgan:
I am guessing you have to get permission to mirror the subs corpus on openslr.org.
Meanwhile, Is it ok to use an lm trained on the acoustic model training tex?
I could also use the reference GlobalPhone LMs.
The Spanish GlobalPhone LM is available at:
http://www.csl.uni-bremen.de/GlobalPhone/lm/SP.3gram.lm.gz
J

danpovey:
Permission might not be needed depending on the license of the existing
thing.  Anyway I'll wait for Yenda to deal with this.  We'll need to wait
till the website is up until we can do anything.  Probably using the
acoustic training text would be fine, but if you've already figured out how
to use the other database, it would be nice to make use of your existing
work.

johnjosephmorgan:
The open subtitles site is working now.

http://opus.lingfil.uu.se/OpenSubtitles.php
says
IMPORTANT: If you use the OpenSubtitle corpus, please, add a link to
http://www.opensubtitles.org/ to your website and to your reports and
publications produced with the data! I got the data under this
condition!

So it looks like mirroring these corpora on openslr.org should not be a problem.
John

jtrmal:
John,  do you have comparison how the subtitles help compared to the LM on
the training data text?
If you want to make this completely correct, you should interpolate the lm
on the training data text and the es subtitles text to minimize perplexity
-- I don't think that movies subtitles would be a correct LM for your
domain,
y.

johnjosephmorgan:
Yenda,

I've been using the subs corpus because I think it consists mostly of
dialogues.
However, I don't think it fits with the purpose of this recipe.
If it's OK with you, I will go back to training the lm on the AM training text.
j

johnjosephmorgan:
The scores in the RESULTS file were obtained with the AM training text.
It's disappointing that the chain model scores do not outperform the
tri3b scores.
Could this be because the corpus is so small and narrow?
j

egs/heroico/s5/run.sh
```diff
@@ -0,0 +1,295 @@
...
+  #local/subs_prepare_data.pl
+
+  #rm $tmpdir/subs/OpenSubtitles2016.en-es.es
+#fi
+
+if [ $stage -le 4 ]; then
+  # build lm
+  local/prepare_lm.sh $lm_training_data
+
+  utils/format_lm.sh \
```
danpovey: Please don't put each argument on a separate line like this -- it makes the script hard to read without excessive scrolling.

egs/heroico/s5/local/chain/run_tdnn.sh
```diff
@@ -0,0 +1,286 @@
...
+  # those.  The num-leaves is always somewhat less than the num-leaves from
+  # the GMM baseline.
+   if [ -f $tree_dir/final.mdl ]; then
+     echo "$0: $tree_dir/final.mdl already exists, refusing to overwrite it."
+     exit 1;
+  fi
+  steps/nnet3/chain/build_tree.sh \
+    --frame-subsampling-factor 3 \
+    --context-opts "--context-width=2 --central-position=1" \
+    --cmd "$train_cmd" 3500 ${lores_train_data_dir} \
```
danpovey: You'll normally want to use a number of leaves that's less than your baseline GMM system.

egs/heroico/s5/local/chain/run_tdnn.sh
```diff
@@ -0,0 +1,286 @@
...
+  input dim=40 name=input
+
+  # please note that it is important to have input layer with the name=input
+  # as the layer immediately preceding the fixed-affine-layer to enable
+  # the use of short notation for the descriptor
+  fixed-affine-layer name=lda input=Append(-2,-1,0,1,2,ReplaceIndex(ivector, t, 0)) affine-transform-file=$dir/configs/lda.mat
+
+  # the first splicing is moved before the lda layer, so no splicing here
+  relu-batchnorm-layer name=tdnn1 dim=512
+  relu-batchnorm-layer name=tdnn2 dim=512 input=Append(-1,0,1)
```
danpovey: It's hard to judge without seeing the diagnostics from your training (e.g. the output of steps/info/chain_dir_info.pl), but I suspect it is overfitting and that it would be helpful to reduce these dimensions from 512 to, say, 400 or reduce the num-epochs.  Also you might want to modify this script to be more like the current WSJ run_tdnn.sh, which points to the 1e version (includes adding l2 related options, removing proportional-shrink,  and changing num-jobs-final).

Normally when we check in these scripts we put some kind of diagnostics and WERs at the top of the script.  the output of chain_dir_info.pl would be a good start.

egs/heroico/s5/README
```diff
@@ -0,0 +1,4 @@
+ heroico
+Scripts  for Kaldi recipe for Spanish using the Heroico corpus.
```
danpovey: Some information about the Heroico corpus here would be nice.  E.g. how much train and test data is there, and what kind of speech does it contain?

dresen:
A word of caution: When I tried to use the Danish language data from opensubtitles, I got many surprises because the people who created the subtitles use odd tricks e.g. double capital 'i' in words like caIIing. In the end I abandoned the corpus because it needed too much cleaning.

jtrmal:
Yes, I think the recipe shouldn't be using the subtitles.
y.

johnjosephmorgan:
@trmal should I continue tuning?

jtrmal:
@danpovey , I think this is OK to merge, unless you have more suggestions what else could @johnjosephmorgan tune in the neural nets -- but it's a fairly small train data (10 hrs), so he's using only tdnn.

egs/heroico/s5/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
```
danpovey: I don't like including the Apache license here--  It makes it look almost like it's a separate project, which will lead to confusion.
There is already a COPYING file at the top of the Kaldi project which includes the full text of the Apache 2.0 license (after some preamble).

egs/heroico/s5/local/chain/tuning/run_tdnn_1i.sh
```diff
@@ -0,0 +1,337 @@
...
+# 1i
+# set number of epochs to 7 (avoid overfitting like 1e)
+# set network hidden layer dimensions to 256
+
+# info
+# exp/chain/tdnn1i_sp:
+# num-iters=105
+# nj=1..1
+# num-params=2.0M
+# dim=40+100->1392
```
danpovey: Since this is a new recipe, I'd prefer to have just a single recipe here, whichever is the best, and call it 1a.

danpovey:
Thanks!  Merging.

PR #1884, opened 2017-09-13
Add scales and constant values to Descriptors

danpovey:
This allows you to do expressions like (1-x) to be done as part of Descriptors, which could be useful for things like RNNLMs (although you still can't do dot-products as part of Descriptors).

This seems like something that might be useful for the future, even if it's not immediately helpful.

@GaofengCheng, it would be great if you could test whether RNNLMs or LSTMs can be implemented this way.  I have done some basic testing of this code but it needs more testing.

Please note, this PR is against the 'rnnlm' branch, and the 'rnnlm' branch will not be merged to master for at least a few weeks.  So don't make any important work dependent on this-- keep it on a side branch.


GaofengCheng:
@danpovey Just saw it, will do it

GaofengCheng:
@danpovey I have tested new descriptor.
```
# ./local/chain/compare_wer_general.sh tdnn_opgru_traditional tdnn_opgru_scale_const 
# System                tdnn_opgru_traditional  tdnn_opgru_scale_const
# WER on train_dev(tg)      12.65     12.82
# WER on train_dev(fg)      11.56     11.75
# WER on eval2000(tg)        14.9      14.9
# WER on eval2000(fg)        13.4      13.4
# Final train prob         -0.067    -0.066
# Final valid prob         -0.086    -0.087
# Final train prob (xent)        -0.879    -0.878
# Final valid prob (xent)       -0.9628   -0.9697
```
config for scale and const:
`configs.append("component-node name={0}.y1_t component={0}.y1 input=Append({0}.h_t, Sum(Scale(-1.0,{0}.z_t), Const(1.0, {1})))".format(name, cell_dim))`

parsed:
`component-node name=gru2.y1_t component=gru2.y1 input=Append(gru2.h_t, Sum(Scale(-1.0,gru2.z_t), Const(1.0, 1024)))`

danpovey:
OK, thanks.  If there was a bug it would have been a bug difference.  That
looks the same.  So I guess I'll merge it.

Issue #1885, opened 2017-09-13
code snippet seem never be executed


yzmyyff:
In swbd/s5c/local/swbd1_train_lms.sh, lines from 49 to 51 seem would never be executed.

I do not know if there is any problem here.

psmit:
Why do you think they are never executed? To me it seems standard code that checks if those two files exist.

yzmyyff:
```bash
for f in "$text" "$lexicon"; do 
  [ ! -f $x ] && echo "$0: No such file $f" && exit 1;  # $x is a missing variable, or it should be $f?
done
```

psmit:
Yes, there is a bug. Maybe you can make a PR to fix it?

yzmyyff:
Thanks, I will.

PR #1886, opened 2017-09-14
fix bug #1885

yzmyyff:


galv:
I want to mention that we could have caught this bug with shellcheck, actually. I could probably set it up pretty easily for CI. It has a few false positives (mostly about quoting variables), but those rules could be disabled.

danpovey:
Hm.
might be a good idea to fix any existing bugs before adding it to travis.
probably a good idea though.

galv:
I tried running this on only `egs/wsj/s5/{steps,utils}` Quite a few warnings show up! Pretty amazing. I'll have to get back to this, hopefully this weekend.

PR #1887, opened 2017-09-14
[scripts] Formatting ivector extractor

lparam:
Ident using spaces 2.


danpovey:
thanks for the PR, but I don't like to commit edits that are just about whitespace.
the issue is that later when someone does "git blame" to find who wrote code (or where a bug came from), they don't find the person who really wrote that code.

PR #1889, opened 2017-09-15
[scripts] Add --train-stage to rerun ivector extrator trainning

lparam:


danpovey:
This script already supports the --stage option, which does what you want.

lparam:
--train-stage is for trainning iteration like --stage in steps/nnet3/chain/train.py. 
if ivector-extractor-sum-accs crash or some other reason,  we could use --train-stage to rerun trainning.

PR #1891, opened 2017-09-17
Updates related to refactoring of the phonetisaurus build system.

AdolfVonKleist:
These changes reflect some minor refactoring of the g2p build system as contributed by @giuliopaci .  There are two changes which entailed updating the ```install_phonetisaurus.sh``` build script: 

1. The configure scripting and location have been updated
2. The compiled binaries and ```phonetisaurus-apply``` scripts have been moved to more canonical locations.  

Note also that the wrapper scripts, ```phonetisaurus_apply```, and ```phonetisaurus_train``` have been renamed to ```phonetisaurus-apply```, and ```phonetisaurus-train```.  While this last does not affect the ```install_phonetisaurus.sh``` script directly, it may affect any egs/experiments that used the old names.

### DIFF
--- a/tools/extras/install_phonetisaurus.sh
+++ b/tools/extras/install_phonetisaurus.sh
@@ -60,7 +60,6 @@ fi
     cd phonetisaurus-g2p
     #checkout the current kaldi tag
     git checkout -b kaldi kaldi
-    cd src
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks.  I'll merge it now; if errors are found I'll fix it later.

PR #1896, opened 2017-09-20
Xvectors: DNN Embeddings for Speaker Recognition 

david-ryan-snyder:
**Overview**
This pull request adds xvectors for speaker recognition. The system consists of a feedforward DNN with a statistics pooling layer. Training is multiclass cross entropy over the list of training speakers (we may add other methods in the future). After training, variable-length utterances are mapped to fixed-dimensional embeddings or “xvectors” and used in a PLDA backend. This is based on http://www.danielpovey.com/files/2017_interspeech_embeddings.pdf, but includes recent enhancements not in that paper, such as data augmentation.

This PR also adds a new data augmentation script, which is important to achieve good performance in the xvector system. It is also helpful for ivectors (but only in PLDA training).

This PR adds a basic SRE16 recipe to demonstrate the system. An ivector system is in v1, and an xvector system is in v2. 

**Example Generation**
An example consists of a chunk of speech features and the corresponding speaker label. Within an archive, all examples have the same chunk-size, but the chunk-size varies across archives. The relevant additions:

+ sid/nnet3/xvector/get_egs.sh — Top-level script for example creation
+ sid/nnet3/xvector/allocate_egs.py — This script is responsible for deciding what is contained in the examples and what archives they belong to.
+ src/nnet3bin/nnet3-xvector-get-egs — The binary for creating the examples. It constructs examples based on the ranges.* file.

**Training**
This version of xvectors is trained with multiclass cross entropy (softmax over the training speakers). Fortunately, steps/nnet3/train_raw_dnn.py is compatible with the egs created here, so no new code is needed for training. Relevant code:

+ sre16/v1/local/nnet3/xvector/tuning/run_xvector_1a.sh — Does example creation, creates the xconfig, and trains the nnet

**Extracting XVectors**
After training, the xvectors are extracted from a specified layer of the DNN after the temporal pooling layer. Relevant additions

+ sid/nnet3/xvector/extract_xvectors.sh — Extracts embeddings from the xvector DNN. This is analogous to extract_ivectors.sh.
+ src/nnet3bin/nnet3-xvector-compute — Does the forward computation for the xvector DNN (variable-length input, with a single output).  

**Augmentation**
We’ve found that embeddings almost always benefit from augmented training data. This appears to be true even when evaluated on clean telephone speech. Relevant additions:

+ steps/data/augment_data_dir.py — Similar to reverberate_data_dir.py but only handles additive noise. 
+ egs/sre16/v1/run.sh — PLDA training list is augmented with reverb and MUSAN audio
+ egs/sre16/v2/run.sh — DNN training and PLDA list are augmented with reverb and MUSAN.

**SRE16 Recipe**
The PR includes a bare bones SRE16 recipe. The goal is primarily to demonstrate how to train and evaluate an xvector system.  The version in egs/sre16/v1/ is a straightforward i-vector system.  The recipe in egs/sre16/v2 contains the DNN embedding recipe. Relevant additions

+ egs/sre16/v1/local/ — A bunch of dataprep scripts
+ egs/sre16/v2/local/nnet3/xvector/prepare_feats_for_egs.sh -- A script that applies cmvn and removes silence frames and writes the results to disk. This is what the nnet examples are generated from. 
+ egs/sre16/v1/run.sh — ivector top-level script
+ egs/sre16/v2/run.sh — xvector top-level script

Results for this example:
```
  xvector (from v2) EER: Pooled 8.76%, Tagalog 12.73%, Cantonese 4.86%
  ivector (from v1) EER: Pooled 12.98%, Tagalog 17.8%, Cantonese 8.35%
```
Note that the recipe is somewhat "bare bones." We could improve the results for the xvector system further by adding even more training data (e.g., Voxceleb: http://www.robots.ox.ac.uk/~vgg/data/voxceleb/). Both systems would improve from updates to the backend such as adaptive score normalization or more effective PLDA domain adaptation techniques. However, I believe that is orthogonal to this PR.


osadj:
@David, which i-vector system is this? On SRE'16 evaluation set, we are getting 9.63% EER (minDCF~0.63) with a basic single i-vector system. It is promising to see you are reporting improvement with "xvectors" over your own i-vector system, but you also need to compare the results with that of a state-of-the-art system.

danpovey:
Your 9.63% EER was pooled on Cantonese and Tagalog?
And was there any nonstandard training data in there?

osadj:
Of course. And no data augmentation was used. We only used previous SRE data (SRE04-SRE10) with GMM based system. Please see NIST's presentation at SRE'16 workshop for more details.

danpovey:
I think the point of the PR was to compare the embeddings, there was not a
lot of attention given to the backend.  As pointed out in the PR, there is
a lot of score normalization and domain adaptation of the PLDA which is
possible but which was not done in this PR.  The numbers on sre10 are
basically state of the art, to my knowledge.
A paper on this is here
http://www.danielpovey.com/files/2017_interspeech_embeddings.pdf.

osadj:
I think it would be minimal work to add a simple PLDA adaptation (using unlabeled data from Major languages) to this PR. Also, I don't believe the numbers you are referring to are state-of-the-art results in SRE10. IBM has published the best numbers so far on SRE10 C5 and 10s-10s (obtained using a single system):

https://arxiv.org/pdf/1605.01635.pdf

danpovey:
It's great that you have such good numbers.  What is your timeline to open
source your system?

We may add PLDA adaptation at some point but I don't want to delay this PR
over that.  There are many scenarios where that is not an issue and the
main point of xvectors is that they help a lot when enrolment or test cuts
are very short.

osadj:
Thanks. We have already released a system for LRE'17 (at https://lre.nist.gov). Time permitting, we may also release our SRE'16 system (the base code will be the same, need to publish the recipe).

I agree that xvectors really shine on short test segments. It's great work.

david-ryan-snyder:
@osadj, Thanks for your feedback.

As @danpovey pointed out, the point of this PR wasn't really to produce a state of art SRE16 recipe. My main goal is to make a complete DNN embedding (aka xvector) recipe available. My expectation in providing this recipe is that the user will likely have his own backend that he can plug into, and will likely use different training lists anyway.

Both the ivector and the xvector recipes can be improved by tuning the training lists and adding things like score normalization and PLDA adaptation. At some point, I would like to add more of this backend stuff to Kaldi. But, since these things are often not too hard to implement in matlab or python, the priority is lower for now.

david-ryan-snyder:
> On SRE'16 evaluation set, we are getting 9.63% EER (minDCF~0.63)

Is this balanced or unbalanced? What is the p(target) for your minDCF?

osadj:
That is equalized minCprimary (average over the two Pt's). 

I also noticed you have not mentioned what scoring tool you used to compute the results. I think you might want to use the NIST official scoring tool (or mention it if you used it).

david-ryan-snyder:
The run.sh is reporting unequalized scores. After applying score normalization and using the NIST scoring script, I get the following for the xvectors:

EER = 10.61
min_Cprimary = 0.6038

david-ryan-snyder:
After haphazardly applying some agglomerative cluster based adaptation, I get the following for the xvectors. Results are using your scoring script and are equalized.

**Pooled**
EER = 9.23
min_Cprimary = 0.5697

**Cantonese**
EER = 4.59
min_Cprimary = 0.3937

**Tagalog**
EER = 12.92
min_Cprimary = 0.7411

Probably these numbers would be better with a little more thought put into the adaptation. But as it stands, it seems competitive with what you reported.

osadj:
@david-ryan-snyder, thanks for reporting back. Indeed, these numbers look great! I would report these instead, not only because they look much better, also they are computed using the official SRE'16 metric.

david-ryan-snyder:
@osadj,

Thanks!

I agree that the better numbers are preferable. However, I used my own score normalization and the clustering-based PLDA adaptation which is not in Kaldi yet. I agree we should add those things, especially now that we have a "bare bones" SRE16 recipe (which, as you pointed out, clearly benefits from domain adaptation). However, I think this is better left as a separate pull request so that I, or whoever decides to take this on, can focus on adding that one feature and doing it well.

If anyone has access to the SRE16 data, they must've participated in the evaluation, which means they still have access to your script... which means they should be able to compute the official equalized scores. To address your concern (which I think is about misrepresenting the evaluation), I can make a comment in the results section of the run.sh script with the equalized EER and minDCF in the current version. I can keep those numbers, and Kaldi's unequalized EER up-to-date as the recipe is improved with things domain adaptation.

osadj:
@david-ryan-snyder, we are on the same page on this. In my latest comment, I was suggesting that it would be more helpful if you reported (in RESULTS should also do) SRE'16 official metrics so people in the community could make comparisons/interpretations. I didn't mean you should add/report results with score normalization/PLDA adaptation, although that would make the recipe more complete.

david-ryan-snyder:
@danpovey,

Travis keeps failing on the "all ext" build. After the most recent commit, it appears to have failed after exceeding a time limit for the jobs.

However, most of the time this fails for other reasons. For example, in https://travis-ci.org/kaldi-asr/kaldi/jobs/279276106 I got the following error:

`ln: failed to create symbolic link ‘/home/travis/build/kaldi-asr/kaldi/src/lib/libkaldi-matrix.so’: File exists
make[1]: *** [libkaldi-matrix.so] Error 1
make[1]: Leaving directory `/home/travis/build/kaldi-asr/kaldi/src/matrix'
make: *** [matrix] Error 2`

I'm assuming this is unrelated to my PR (I only added 2 C++ binaries in nnet3bin/, the rest of the additions are scripts), but I don't see other active PRs failing as often.

danpovey:
In default_rules.mk, there is an "rm" statement followed by a "ln -s"
statement.  If you could remove the "rm" and replace the "ln -s" with a "ln
-sf" it might fix the issue.

It looks to me like ccache may have parsed the Makefile and assumed that
that sequence of commands is some part of the C compiler, and is trying to
test it, which somehow causes an error.  This may be a workaround.

david-ryan-snyder:
Thanks. I'll try that in the next commit and see if it fixes things.

david-ryan-snyder:
I tried applying the default PLDA adaptation binary (ivector-adapt-plda). The xvector results improve significantly, but the ivector results are only slightly improved. 

Here are the current xvector results. This is taken from the official SRE16 scoring script (equalized). I believe this is now pretty competitive. Of course, additional adaptation or score normalization will improve results further. 

```
  # -- Pooled --
  # EER:          8.97
  # min_Cprimary: 0.61
  # act_Cprimary: 0.63
  #
  # -- Cantonese --
  # EER:          4.71
  # min_Cprimary: 0.42
  # act_Cprimary: 0.43
  #
  # -- Tagalog --
  # EER:         13.23
  # min_Cprimary: 0.77
  # act_Cprimary: 0.82
```

osadj:
Great. Thanks for sharing, @david-ryan-snyder .

david-ryan-snyder:
@danpovey, I made the changes to default_rules.mk and the checks pass now.

I think this is PR is ready to be reviewed.

osadj:
BTW, what data did you use for PLDA adaptation?

david-ryan-snyder:
Just the SRE16 major dataset. I didn't include the minor languages.

osadj:
Thanks. Adding the minor language data should not have a significant impact.

entn-at:
Out of curiosity: Did FBank features give better performance, or is there another reason for using them over MFCCs in the v2 setup?

danpovey:
I can't see that fbanks are actually being used-- other than that the
config file exists (it's probably not needed).

entn-at:
See egs/sre16/v2/run.sh (https://github.com/david-ryan-snyder/kaldi/blob/9cbf06978ffe559cf0cf244b27821e03b6b11826/egs/sre16/v2/run.sh):

steps/make_fbank.sh --fbank-config conf/fbank.conf --nj 40 --cmd "$train_cmd" \
      data/${name} exp/make_fbank $fbankdir

danpovey:
Oh, I see now, they are being used.
I'd like to know why, David...
generally speaking we prefer to dump full-dimensional MFCCs (with
num-fbanks== mfcc-dim) over fbanks, because they are more compressible.
With fbanks, the standard compression algorithm does not work as well
because they are too correlated.

david-ryan-snyder:
Actually I am using fbanks in the embedding recipe. I _did_ find that it works a little better than the 20 MFCCs typically used in the i-vector recipes. This hasn't been extensively explored yet, though.

osadj:
My 2 cents: I have not seen any benefits from using 40-D MFCC of FBank
features over 39-D MFCCs (13+d+dd).

david-ryan-snyder:
I actually haven't tested this with num-fbanks== mfcc-dim. My guess is we can switch to using MFCCs in the embeddings recipe without loss of performance.

danpovey:
I think you should try the full-dimensional MFCCs-- I won't hold up the PR
over it, but it is the way we nornally do such things.

Omid- I'm surprised that extra dimensions of MFCC did not help, esp. if
this was an nnet system.



On Fri, Sep 29, 2017 at 2:49 PM, david-ryan-snyder <notifications@github.com

osadj:
I do see benefits with 20-D MFCCs over 13-D MFCCs (for speaker reco) using
GMMs, but not nnet. This observation has been consistent irrespective of
the ASR toolkit used to train the nnet (i.e., Attila or Kaldi).

entn-at:
@osadj For features used in calculating (GMM/DNN senone) posteriors or as features for calculating BW stats? Or are you talking about DNN embedding type systems such as this PR?

osadj:
I am not talking about this PR (I have yet to try it, will do after
installing the new P100s we got). I meant DNN based speaker reco systems in
general (either for computing the posteriors or BNFs). For BW stats, if I
use DNN posterior alignments, 20-D MFCC may or may not be better than 13-D
MFCC, it depends on the task. On SRE10 I just used 13-D MFCCs to computed
BW stats with DNN alignments (see IBM 2016 speaker reco system).

david-ryan-snyder:
@danpovey, FYI, I'm rerunning the v2 recipe with MFCCs (same dim as the filter banks). I think by the time we go through the PR, I'll be able to update the results (which I doubt will change much).

egs/sre08/v1/sid/nnet3/xvector/allocate_egs.py
```diff
@@ -0,0 +1,325 @@
+#!/usr/bin/env python
```
danpovey: If this script would be trivial to make a python3 script, and you can test it, it would be great to have this be python3.
I'm encouraging python3 for new scripts now; we have accepted the dependency.

egs/sre08/v1/sid/nnet3/xvector/extract_xvectors.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+
+# Copyright     2017  David Snyder
+#               2017  Johns Hopkins University (Author: Daniel Povey)
+#               2017  Johns Hopkins University (Author: Daniel Garcia Romero)
+# Apache 2.0.
+
+# This script extracts embeddings (called "xvectors" here) from a set of
+# utterances, given features and a trained DNN.
```
danpovey: it would be great if you could summarize how it differs from the regular extract_ivectors.sh script

egs/sre08/v1/sid/nnet3/xvector/get_egs.sh
```diff
@@ -0,0 +1,241 @@
+#!/bin/bash
+
+# Copyright      2017 Johns Hopkins University (Author: Daniel Povey)
+#                2017 Johns Hopkins University (Author: Daniel Garcia-Romero)
+#                2017 David Snyder
+# Apache 2.0
+#
+# This script dumps training examples (egs) for multiclass xvector training.
+# These egs consist of a data chunk and a label. Each archive of egs has,
```
danpovey: explain that the label corresponds to the speaker, and whether it's one-based or zero-based.

egs/sre16/v1/local/make_musan.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
```
danpovey: if this script could trivially be upgraded to python3, that would  be great;
I now prefer python3 for new scripts.

src/nnet3bin/nnet3-xvector-get-egs.cc
```diff
@@ -0,0 +1,238 @@
...
+                  << "output files.";
+      (*example_writers)[chunk->output_archive_id]->Write(
+                         chunk->name, eg);
+      (*num_egs_written) += 1;
+    }
+  }
+}
+
+// Delete the dynamically allocated memory.
+static void Cleanup(unordered_map<std::string,
```
danpovey: Instead of having its own function I'd prefer to just put this in its own code block ({...}) with
a comment like
  ` {  // Free memory`
It feels like it should be part of the main function.
For the vector you can use DeletePointers() rather than writing a loop.

danpovey: Actually you could modify the DeletePointers() template so that it works with any container, not just vectors.  You could template it on the STL container, e.g.
`template <typename C> `...
and internally do:
```
  // T is the pointer type that the container contains, e.g. T == int*.
  typedef typename C::value_type T;
``

**Comment by david-ryan-snyder** [comment]
Thanks Dan! I'll look into addressing these.

**Comment by dgromero** [comment]
@danpovey, the reason why we are using Fbanks is for the current research we are doing in bandwidth extension/multi-bandwidth systems, as well as some frequency axis invariance (using convolutions). In the experiments, we saw that the results were very similar with MFCCs and Fbanks, but Fbanks are more in line with our future plans.

**Comment by dgromero** [comment]
@david-ryan-snyder, @danpovey I also wanted to mention that you are being too humble. xvectors not only shining for short durations! The next round of ICASSP papers will show that. The main advantage of the xvectors (or in general, DNN embedding architectures) is that they are able to leverage large amounts of data much better than the ivector architectures. Specifically, data augmentation helps in the embedding training part (whereas in standard ivector systems it typically hurts, and it is the back-end that needs to handle the multi-condition/ data-augmented training). The reason is that augmenting data to train the UBM and T matrices is not such a good idea since it is an unsupervised task. However, the xvectors are trained in a supervised way with multi class cross-entropy to discriminate among speakers. That makes the xvectors more invariant to those distorsions. Also, you can still use all the same back-end bag-of-tricks that we have accumulated over the years for ivector systems.

**Comment by david-ryan-snyder** [comment]
Except for the fbanks vs MFCC issue, I believe I've addressed all the comments. I'm still waiting for the experiment using full dim MFCCs to finish, and then this should be finished.

**Comment by entn-at** [comment]
I adapted the setup to SRE10, using the same corpora that are used in the sre10/v2 recipe (SRE04-08, SWB2 p2/3, SWB Cell p1/2, Fisher) and got an EER of 2% (no further tuning, just using the same data augmentation procedure, fbank features, LDA dimension, no PLDA model adaptation (trained on SRE04-08) etc.).
Would be interesting to run this on the PRISM set or similar to see how it compares under noise/reverberation (but I assume you've already done those kinds of experiments - I'm just posting to confirm that it runs fine ;-)).

**Comment by david-ryan-snyder** [comment]
Here are the updated results using MFCCs instead of Fbanks. The results are no worse. Maybe it's even slightly better. I'll go ahead and change it over in the recipe.
 
```
  # -- Pooled --
  #                             Fbanks     MFCC
  # EER:                        8.97       8.66
  # min_Cprimary:               0.61       0.61
  # act_Cprimary:               0.63       0.62
  #
  # -- Cantonese --
  # EER:                        4.71       4.69
  # min_Cprimary:               0.42       0.42
  # act_Cprimary:               0.43       0.43
  #
  # -- Tagalog --
  # EER:                       13.23      12.63
  # min_Cprimary:               0.77       0.76
  # act_Cprimary:               0.82       0.81

```

osadj:
Thanks, Ewald, for sharing this. It is good to know that the setup can be
(easily?) adapted for other tasks as well. Can you also please provide
further details on the SRE10 task, e.g., test condition (C1..C5 or pooled),
male vs female (or pooled)? I understand the details are in the recipe ;-)
but it would be great if you could share those.

david-ryan-snyder:
@entn-at,

Very cool, thanks for running this!

I've done a similar experiment and got 1.7% EER on SRE10. I used more datasets (+ Voxceleb), though.

david-ryan-snyder:
`SRE04-08, SWB2 p2/3, SWB Cell p1/2, Fisher`

So you used Fisher in the xvector recipe? I've been wondering if this will work, since there aren't a lot of recordings per speaker. Maybe we could try an experiment where Fisher is augmented more aggressively than the other datasets. Might help to offset the limited data.

osadj:
David, I do not find any details on the SRE10 task in run.sh or the README.
Can you please add those information to either the README or run.sh where
you report the results? Thanks.



On Tue, Oct 3, 2017 at 12:47 PM, david-ryan-snyder <notifications@github.com

entn-at:
Omid: I ran it on C5 extended (same as the sre10/{v1,v2} recipes), pooled male/female (though it would be easy to get those numbers).

osadj:
Thanks, Ewald. Is it then fair to compare the EER you reported to "# ind
pooled: 1.01" (that is a gender-independent system evaluated on pooled
male/female trials on C5)? hmm, EER~2% does not look right to me for a DNN
based system.

entn-at:
David: Yes, that is indeed the issue and the reason why most Fisher speakers get discarded during example preparation (I end up with 7170 speakers in the combined training set, 5767 of which are from SWB/SRE). More augmentation or treating Fisher differently when pruning the lists may make a big difference. Adding VoxCeleb is a good idea, I've used it in a different context for training in-domain PLDA/calibration. Another reason for the difference in EER I got could be that I reduced the number of training epochs from 3 to 2, as I only used 3 training jobs (num-jobs-initial/final=3). 

Omid: Note that this was merely a proof-of-concept ("making it run"), I'm sure you can get better performance by tuning hyperparameters and adding more datasets. Without any tuning whatsoever, I get EER=2, minDCF10=0.435

david-ryan-snyder:
@osadj: I don't have the 1.7% EER on SRE10 using xvectors documented anywhere yet. In the future we'll probably work on a recipe for this (unless someone else finds a good one first, e.g., @entn-at :-)). 

I think we can still do a lot better on SRE10 by using more augmentation on Fisher, additional datasets, etc. Maybe including ASR BNFs will help as well (although it's aesthetically undesirable, it's a fairer comparison with sre10/v2).

david-ryan-snyder:
@danpovey, I've addressed those comments. I think the PR should be ready to ship now.

danpovey:
thanks!  Merged.

497354078:
Hi, I have a question about TDNN, 
[sre16/v2/local/nnet3/xvector/run_xvector.sh] line 105
  # The stats pooling layer. Layers after this are segment-level.
  # In the config below, the first and last argument (0, and ${max_chunk_size})
  # means that we pool over an input segment starting at frame 0
  # and ending at frame ${max_chunk_size} or earlier.  The other arguments (1:1)
  # mean that no subsampling is performed.
  stats-layer name=stats config=mean+stddev(0:1:1:${max_chunk_size}) 
--this layer's output dims is 3001, but mean(1500) concat stddev(1500) is 3000dims. Can you tell the detail? thanks.

david-ryan-snyder:
@497354078, in nnet3, this pooling layer is implemented as two components: StatisticsExtractionComponent and StatisticsPoolingComponent.

The first component computes the sum over X, X^2, and a _count_. That's why this component has 3001 rather than 3000 components (1500 + 1500 + 1). This is then received as input by the second layer, which uses these sums and the count to compute the avg(X) and stddev(X). The output of this second layer is actually 3000, as you would expect.

It might help to look at the forward computation for each component: https://github.com/kaldi-asr/kaldi/blob/master/src/nnet3/nnet-general-component.cc#L447 and https://github.com/kaldi-asr/kaldi/blob/master/src/nnet3/nnet-general-component.cc#L771

PR #1899, opened 2017-09-21
Fast lattice rescoring based on pruned composition

danpovey:


danpovey:
Hainan will extend this to the RNNLM case.

PR #1900, opened 2017-09-22
FIX: CU_SAFE_CALL used with cublasXXX func

selaselah:
all CuBLAS func return a cublasStatus_t, which is not same as cudaError_t and the value mean is not compatible

### DIFF
--- a/src/cudamatrix/cu-common.cc
+++ b/src/cudamatrix/cu-common.cc
@@ -68,10 +68,8 @@ void GetBlockSizesForSimpleMatrixOperation(int32 num_rows,
   dimGrid->z = 1;
 }
 
-const char* cublasGetStatusString(cublasStatus_t status)
-{
-  switch(status)
-  {
+const char* cublasGetStatusString(cublasStatus_t status) {
+  switch(status) {
     case CUBLAS_STATUS_SUCCESS:           return "CUBLAS_STATUS_SUCCESS";
     case CUBLAS_STATUS_NOT_INITIALIZED:   return "CUBLAS_STATUS_NOT_INITIALIZED";
     case CUBLAS_STATUS_ALLOC_FAILED:      return "CUBLAS_STATUS_ALLOC_FAILED";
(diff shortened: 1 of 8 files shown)


src/cudamatrix/cu-common.cc
```diff
@@ -67,6 +67,25 @@ void GetBlockSizesForSimpleMatrixOperation(int32 num_rows,
                "Matrix has too many rows to process");
   dimGrid->z = 1;
 }
+
+const char* cublasGetStatusString(cublasStatus_t status)
```
danpovey: thanks for the PR.
So there wasn't an official version of this function, that we could have used?

danpovey: I see that there wasn't an official function like this, but please use Google style w.r.t. open-braces: put them on the same line as the previous one, after a space, e.g.
`const char* cublasGetStatusString(cublasStatus_t status) {`

And please put a comment before the declaration of this function, saying:
`// This is analogous to the CUDA function cudaGetErrorString().`

selaselah:
the CI failed randomly. I have reopened this request several times but all are failed. Here is some error log:

    ln: failed to create symbolic link ‘/home/travis/build/kaldi-asr/kaldi/src/lib/libkaldi-matrix.so’: File exists

src/cudamatrix/cu-common.cc
selaselah: Reformat done.
Howerver, the CI raise some reasonless error :(

danpovey:
Thanks.  Maybe @jtrmal or @kkm can look into the CI errors-- probably an
issue with the latest scripts there.

jtrmal:
The compilation went fine, only the scripts weren't  able to finish storing
the ccache on time -- not sure what to do with that.
y.

danpovey:
We have some CuSparse library calls in the 'rnnlm' branch and I think the same issue exists there.  Do you have time to address that via another PR?

selaselah:
I check the CuSparse's doc,. The status value also conflict with cudaError_t. I'll take a PR tomorrow.

PR #1902, opened 2017-09-22
[scripts] zero-fill indexes in segment_ctm_edits scripts

psmit:
Fixes main problem of #1892

I have not added any fix_data_dir.sh calls, but can do that if so desired. I would at least suggest to add a validate_data_dir.sh call on the end of clean_and_segment_data.sh to capture any other bugs that might come up.

### DIFF
--- a/egs/wsj/s5/steps/cleanup/clean_and_segment_data.sh
+++ b/egs/wsj/s5/steps/cleanup/clean_and_segment_data.sh
@@ -192,6 +192,9 @@ if [ $stage -le 8 ]; then
   echo "$0: based on the segments and text file in $dir/segments and $dir/text, creating new data-dir in $data_out"
   padding=$(cat $dir/segment_end_padding)  # e.g. 0.02
   utils/data/subsegment_data_dir.sh --segment-end-padding $padding ${data} $dir/segments $dir/text $data_out
+  # utils/data/subsegment_data_dir.sh can output directories that have e.g. to many entries left in wav.scp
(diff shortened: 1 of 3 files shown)


danpovey:
@vimalmanohar, can you check this?
Maybe the validate_data_dir.sh call would be a good idea.  Peter, maybe you
could add that to the PR?

psmit:
Note I revised the first commit. I had forgot to do `git add` after fixing the last off-by-one error.

Regarding adding validate_data_dir.sh, I just checked and that actually fails if there are unused entries in wav.scp (which is very possible to happen after clean_and_segment). Adding "fix_data_dir.sh" would do the trick, but it might possibly hide bugs... I can also try to have a look to utils/data/subsegment_data_dir.sh so that it would filter out unused id's in wav.scp, but that would take me a bit longer to fix and test.

danpovey:
@vimalmanohar, do you have an opinion on this?
I'd lean towards just adding fix_data_dir.sh, so the script will run
without error.

vimalmanohar:
This looks fine. Adding fix_data_dir.sh after clean_and_segment_data.sh is fine. But it might be better to add it outside the clean_and_segment_data.sh in order to ensure we do not miss any errors there.

danpovey:
@psmit, can you just add the fix_data_dir.sh call at the end of clean_and_segment_data.sh?  I think that's the simplest solution.  THen I'll merge the PR.

psmit:
Ok, done. I added it almost on the end, before the CMVN stats are calculated.

PR #1903, opened 2017-09-22
 Do not assume the decoded text is in utf-8

jtrmal:
This should address some of the issues when the decoded text is not utf-8 compatible and we do not know the encoding (as in Kaldi we do not care). In that case, we handle the stream as a bytestream.
Will have to do more testing, but I have already passed one case when the scoring was failing before (Windows-1250).
@danpovey -- Opened to comments


egs/wsj/s5/utils/scoring/wer_ops_details.pl
```diff
@@ -31,16 +31,44 @@
...
-binmode STDOUT, ":utf8";
-
 my $help;
 my $special_symbol= "<eps>";
 my $separator=";";
 my $extra_size=4;
 my $max_size=16;
 
 
+sub get_utf8_or_ascii {
```
danpovey: would be nice to have documentation for this function.

jtrmal:
@danpovey , checked on three different corpora: AmEn, Cantonese(UTF-8) and Czech(Win-1250). Works well on all of them now.

danpovey:
cool.
merge yourself if you think it's ready.

PR #1907, opened 2017-09-24
Add Switchboard RNNLM script

keli78:


keli78:
I'll update the results later

PR #1909, opened 2017-09-25
[not-for-merge]: support multi GPU and Thread

selaselah:
kaldi's cudamatrix is not support multi GPU:
It make gpu-config(CuDevice) singleton, so all thread in the process will get same config.
And according cuda's doc, all thread must call cudaSetDevice(). If not, new thread will use the default GPU.
The cublas's cublasCreate init context according specified GPU. So all thread should call cublasCreate afte cudaSetDevice.

I fix it without change origin API and keep CuDevice singleton. So all other kaldi's code need NO change. All I have done:
- make CuDevice thread_local, so each thread will use its own config.
- move the device reset code out of IsComputeExclusive(). Origin version is bugly and the reset behavior should be done by caller.
- add a new function SelectGpuId(use_gpu, forced_gpu_id). This version will keep try use given gpu_id.

I make a simple test on my own server and everything is OK.

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -56,7 +56,6 @@ namespace kaldi {
    initialization).  But regardless, changing to this new mechanism should be
    harmless even if the problem was specific to the CLSP grid.
 */
-
 static bool GetCudaContext(int32 num_gpus, std::string *debug_str) {
   std::ostringstream debug_stream;
   debug_stream << "num-gpus=" << num_gpus << ". ";
@@ -71,6 +70,32 @@ static bool GetCudaContext(int32 num_gpus, std::string *debug_str) {
     cudaGetLastError();  // Make sure the error state doesn't get returned in
                          // the next cudaGetLastError().
   }
+  *debug_str = debug_stream.str();
+  return false;
+}
+
+/**
+ * Similar as GetCudaContext(num_gpus, debug_str). But only run on given GPU.
+ */
+static bool GetCudaContextById(int32 gpu_id, std::string *debug_str) {
+  std::ostringstream debug_stream;
+  cudaSetDevice(gpu_id);
+  cudaError_t e = cudaDeviceSynchronize(); // << CUDA context gets created here.
+  if (e == cudaSuccess) {
+    int32 act_gpu_id;
+    e = cudaGetDevice(&act_gpu_id);
+    if (e == cudaSuccess) {
+      if (act_gpu_id == gpu_id) {
+        *debug_str = debug_stream.str();
+        return true;
+      }
+      debug_stream << "Trty to set GPU#" << gpu_id << ", but get GPU#" << act_gpu_id;
+    }
+    debug_stream << "Device " << gpu_id << ": " << cudaGetErrorString(e) << ". ";
+  }
+  debug_stream << "Device " << gpu_id << ": " << cudaGetErrorString(e) << ".  ";
+  cudaGetLastError();  // Make sure the error state doesn't get returned in
+  // the next cudaGetLastError().
   *debug_str = debug_stream.str();
   return false;
 }
@@ -180,6 +205,11 @@ void CuDevice::SelectGpuId(std::string use_gpu) {
   } else {
     // Or suggest to use compute exclusive mode
     KALDI_WARN << "Suggestion: use 'nvidia-smi -c 3' to set compute exclusive mode";
+    // in this case we release the GPU context...
+    e = cudaThreadExit(); // deprecated, but for legacy reason not cudaDeviceReset
+    if (e != cudaSuccess) {
+      KALDI_CUDA_ERR(e, "Failed to release CUDA context on a GPU");
+    }
 
     // And select the GPU according to proportion of free memory
     if (SelectGpuIdAuto()) {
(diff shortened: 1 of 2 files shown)


danpovey:
This is definitely interesting, and I'll leave it up for discussion purposes.  However, it's not going to get merged anytime soon.  There are a couple of issues:

(1) It would need to be part of an overall strategy for how to use multiple local GPUs for parallel model training.  Currently the strategy we use is to treat multiple GPUs as if they were on different machines, and periodically average the models, so each job uses one GPU.  If we decided to use multiple local GPUs, decisions would need to be made on how to handle the model or gradient averaging.  In fact, it's not at all clear how communication between different GPUs could be handled, e.g. for model averaging, with the way the current code is structured.  Initializing the neural nets in the different threads would be complicated, unless each thread read from disk.   A neural net from a different thread would contain pointers to the wrong GPU, so they'd be incompatible.

(2) Your changes conflict with what I am doing in the 'rnnlm' branch.  In the 'rnnlm' branch, we use multiple threads, but the different threads use the same GPU.  The main thread does training, and the background thread prepares the examples for training; it does mostly CPU work but it does occasionally execute transfers to the GPU.  These need to be to the same GPU, and your changes would break that code.

(3) Too much stuff to test.  If we did introduce a way to make use of multiple GPUs on the same machine, in the recipes we'd have to introduce a --num-gpus-per-job, in addition to --num-jobs-initial and --num-jobs-final.  This would affect the results, and it might be necessary to tune the recipes differently depending on the hardware configuration.

It is possible-- even likely-- that taking advantage of multiple GPUs attached to the same node would lead to a speed improvement in training.. probably of the order of 20% or so, depending how many jobs there were.  But it would be a big drain on our time and would require a lot of testing.

In fact, if we did go in this direction I think it would make more sense to completely rewrite the training scripts and instead of submitting multiple small jobs to the queue, submit a single job which would run the training from start to finish.  That would mean that some of the logic which currently exists in the python training script, would have to be moved into the C++ binary (either that, or call the C++ from python which starts to get very complicated).  However, this isn't strictly necessary-- it would be possible to keep fairly short-duration jobs that would use multiple GPUs, and leave the scripts broadly structured as they currently are, even if the original reason for the short-duration jobs no longer exists.

Anyway, this would be a very large-scope project and not something we can take on right now.

selaselah:
Thanks for reviewing. I think some issue can be ignore:
1) I think multi-process-single-gpu are appropriate for training. This request will not change this pattern. I intro single-process-multi-gpu for decoding service. In this case, multi-GPU is just scale out, and gpu communication and traning averaging don't take palce. My machine has 4 GPU cards and multi-gpu can boost performance 4x.
2) In rnnlms branches, there has a background_thread, the thread body is RunBackgroundThread.
If apply my request, the func will crash because the cublas and cusparse's handle are uninited. To overcome it, backgournd_thread should call SelectGpuId also. Althoung the code will run without my request, but it will cause bug maybe. There is a example,
A machine has 2 GPU, and GPU#0 run another process. So the SelectGpuId will use GPU#1, and init cublas_handle and cusparse_handle with this GPU. Then, the backgroup_thread forked will running on the default GPU, which is GPU#0, but use the cublas_handle and cusparse_handle belong to GPU#1. 
With my request, each gpu-thread must call SelectGpuId once, rnnlms' code need a little change. But there are some benifit: background_thread can running on the specified GPU and avoid bug.
3) prev binary don't need any change, because they are single-thread process, the thread_local CuDevice worked as same as Singleton. So no '--num-jobs-per-gpu' needed. Change will take place on the binaries which use multi-gpu-thread.

danpovey:
>
>    1.
>    I think multi-process-single-gpu are appropriate for training. This
>    request will not change this pattern. I intro single-process-multi-gpu for
>    decoding service. In this case, multi-GPU is just scale out, and gpu
>    communication and traning averaging don't take palce. My machine has 4 GPU
>    cards and multi-gpu can boost performance 4x.
>
> We generally don't recommend using GPUs at all for decoding, I think it
would be a waste of a GPU (we can generally decode faster than real-time
even on CPU).


>    1. In rnnlms branches, there has a background_thread, the thread body
>    is RunBackgroundThread.
>    If apply my request, the func will crash because the cublas and
>    cusparse's handle are uninited. To overcome it, backgournd_thread should
>    call SelectGpuId also. Althoung the code will run without my request, but
>    it will cause bug maybe. There is a example,
>    A machine has 2 GPU, and GPU#0 run another process. So the SelectGpuId
>    will use GPU#1, and init cublas_handle and cusparse_handle with this GPU.
>    Then, the backgroup_thread forked will running on the default GPU, which is
>    GPU#0, but use the cublas_handle and cusparse_handle belong to GPU#1.
>    With my request, each gpu-thread must call SelectGpuId once, rnnlms'
>    code need a little change. But there are some benifit: background_thread
>    can running on the specified GPU and avoid bug.
>
>
This seems quite complicated and I suspect there is something wrong in your
logic (note: a lot of invocations of the GPU do not use any handle at all,
as they are to custom CUDA functions or to things like cudaMemcpy).


>
>    1.  prev binary don't need any change, because they are single-thread
>    process, the thread_local CuDevice worked as same as Singleton. So no
>    '--num-jobs-per-gpu' needed. Change will take place on the binaries which
>    use multi-gpu-thread.
>
>
I don't see a good reason to use multiple GPUs for decoding.  Anyway I
think this whole approach would be quite wasteful of GPUs.  If neural net
evaluation is on the GPU, it's possible in principle to use a single GPU to
decode many models  at the same time (while having multiple CPU threads do
the graph search).  There is no need to use multiple GPUs at alo--
multilple threads on the same GPU would work, and something else that would
work would be to combine the different decoding threads in a minibatch,
just the same was as is done for training.  This would of course be
complicated, but if you were serious about fast decoding with GPUs it would
make sense to do it.

selaselah:
Yes, CPU is enough for real-time. However in my case, a machine need process about 100secs audio per secend in read-time. I'll do more investigation, maybe one-machine-one-gpu is ideal.

PR #1910, opened 2017-09-25
validate_lang checks for incompatible UTF-8 whitespaces

jtrmal:
This is the first batch of changes in the validation scripts
@danpovey one of the maybe controversial design decision is that the change will cause the files being read twice (read until end and then seek()d back) -- but I think in these specific cases it should be OK

### DIFF
--- a/egs/wsj/s5/utils/validate_data_dir.sh
+++ b/egs/wsj/s5/utils/validate_data_dir.sh
@@ -108,6 +108,7 @@ fi
 
 num_utts=`cat $tmpdir/utts | wc -l`
 if [ -f $data/text ]; then
+  utils/validate_text.pl $data/text || exit 1;
   check_sorted_and_uniq $data/text
   text_len=`cat $data/text | wc -l`
   illegal_sym_list="<s> </s> #0"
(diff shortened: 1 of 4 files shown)


jtrmal:
BTW, I could also check the other files (silence.txt, nonsilence.txt, disambig.txt and so on), but not sure if it makes a lot of sense as they will be (typically) generated automatically or will be a subset of phones.txt
y.

jtrmal:
@danpovey checked for cp1250 and cantonese, seems working as expected.

danpovey:
OK with me if you merge when you think it's ready.

jtrmal:
I'll merge this. I was thinking about deduplicating the code, as all three files contain the same function, but apparently in perl it;'s not really easy to achieve hassle-less relative import.
WIll have to look at it later

danpovey:
I think its easiest to have duplicated code and keep everything separate.

jtrmal:
I actually forgot to add [scripts] in from of the merge message... do you
want me to revert?
y.

danpovey:
no need.

AdolfVonKleist:
Quick question, ignore if not relevant: does this also handle the full-width space encountered frequently in Japanese ('　' [U+3000](https://www.cs.tut.fi/~jkorpela/chars/spaces.html))?

PR #1921, opened 2017-10-03
fix run_ivector_common.sh to use data/lang

LvHang:
Check all the subdirectories in "egs" directory and use "data/lang"(with silence probs) to replace "data/lang_nosp" if it has already be gotten.
Hang


danpovey:
I'm just merging this without having all the setups rerun; I don't believe this will make any difference.

PR #1926, opened 2017-10-09
fix " Argument list too long " in grep

GaofengCheng:
Under Fisher+SWBD, grep -e will cause list too long problem.
Use find + xargs to solve this problem.

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
+++ b/egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
@@ -310,9 +310,10 @@ def parse_progress_logs_for_param_diff(exp_dir, pattern):
 
 
 def parse_train_logs(exp_dir):
-    train_log_files = "%s/log/train.*.log" % (exp_dir)
+    train_log_files = "%s/log/" % (exp_dir)
+    train_log_names = "train.*.log"
     train_log_lines = common_lib.get_command_stdout(
-        'grep -e Accounting {0}'.format(train_log_files))
+        'find {0} -name {1} | xargs grep -e Accounting'.format(train_log_files,train_log_names))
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/steps/libs/nnet3/report/log_parse.py
```diff
@@ -310,9 +310,10 @@ def parse_progress_logs_for_param_diff(exp_dir, pattern):
 
 
 def parse_train_logs(exp_dir):
-    train_log_files = "%s/log/train.*.log" % (exp_dir)
+    train_log_files = "%s/log/" % (exp_dir)
+    train_log_names = "train.*.log"
     train_log_lines = common_lib.get_command_stdout(
-        'grep -e Accounting {0}'.format(train_log_files))
+        'find {0} -name {1} | xargs grep -e Accounting'.format(train_log_files,train_log_names))
```
danpovey: You need to put quotes (double-quotes will do) around `{1}` because otherwise if there is a file
called train.X.log in the current directory it will be substituted by bash.
Also you need to replace `grep -e Accounting` with `grep -H -e Accounting` because otherwise the
filename of the matched file will not be printed and the regex will fail to match.
Also please change the function name parse_train_logs to get_train_times, which is more informative.

GaofengCheng: @danpovey fixed
Also, the total training time is calculated by selecting the max duration per iteration and then adding, will it be better if we use the total computation time (adding all the times in train logs instead of selecting the max).

danpovey:
I think that code is attempting to print an approximation of the wall-clock
time (assuming jobs had immediately launched on the queue on the better
machines.)  Leave that as it is.

PR #1928, opened 2017-10-10
[scripts] Added a dollar sign before sed commmands when it uses \t, to force us…

xiaohui-zhang:
…ing bash's string syntax, since OS X sed doesn't recognize \t as a tab.

### DIFF
--- a/egs/chime1/s5/local/chime1_prepare_data.sh
+++ b/egs/chime1/s5/local/chime1_prepare_data.sh
@@ -40,8 +40,9 @@ scp="$data/train/wav.scp"
 rm -f "$scp"
 for sid in `seq 34`; do
   sid2=`printf "s%02d" $sid`
+  tab=$'\t'
   ls -1 $wav_train/id$sid/*.wav \
-    | sed $"s/\(.*\)\/\(.*\).wav/${sid2}_\2\t\1\/\2.wav/" \
+    | sed "s/\(.*\)\/\(.*\).wav/${sid2}_\2$tab\1\/\2.wav/" \
     | sort >> $scp
 done
 for x in "devel" "test"; do
--- a/egs/chime1/s5/local/chime1_prepare_data.sh
+++ b/egs/chime1/s5/local/chime1_prepare_data.sh
@@ -71,10 +71,10 @@ for x in $set_list; do
     # Create utt2spk files 
     # No speaker ID
     tab=$'\t'
-    sed "s/\(.*\)\t.*/\1${tab}\1/" < "$scp" > "$data/$x/utt2spk"
+    sed "s/\(.*\)${tab}.*/\1${tab}\1/" < "$scp" > "$data/$x/utt2spk"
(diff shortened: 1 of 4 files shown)


egs/chime1/s5/local/chime1_prepare_data.sh
```diff
@@ -41,7 +41,7 @@ rm -f "$scp"
 for sid in `seq 34`; do
   sid2=`printf "s%02d" $sid`
   ls -1 $wav_train/id$sid/*.wav \
-    | sed "s/\(.*\)\/\(.*\).wav/${sid2}_\2\t\1\/\2.wav/" \
+    | sed $"s/\(.*\)\/\(.*\).wav/${sid2}_\2\t\1\/\2.wav/" \
```
danpovey: this is not right: the dollar sign syntax doesn't work with double-quotes, and with single-quotes the backslash-then-digit things won't work right.  I suggest to do
` tab=$'\t'`
and then use `$tab` instead of `\t`.

xiaohui-zhang:
fixed. thanks

egs/chime1/s5/local/chime1_prepare_data.sh
```diff
@@ -68,9 +70,10 @@ for x in $set_list; do
 
     # Create utt2spk files 
     # No speaker ID
-    sed 's/\(.*\)\t.*/\1\t\1/' < "$scp" > "$data/$x/utt2spk"
+    sed $'s/\(.*\)\t.*/\1\t\1/' < "$scp" > "$data/$x/utt2spk"
```
danpovey: this won't work, the backslash-1 disappears.

egs/chime1/s5/local/chime1_prepare_data.sh
```diff
@@ -68,9 +70,11 @@ for x in $set_list; do
 
     # Create utt2spk files 
     # No speaker ID
-    sed 's/\(.*\)\t.*/\1\t\1/' < "$scp" > "$data/$x/utt2spk"
+    tab=$'\t'
+    sed "s/\(.*\)\t.*/\1${tab}\1/" < "$scp" > "$data/$x/utt2spk"
```
danpovey: you missed a \t.

egs/chime1/s5/local/chime1_prepare_data.sh
```diff
@@ -68,9 +70,11 @@ for x in $set_list; do
...
     # Use speaker ID
-    #sed "s/\(s..\)\(.*\)\t.*/\1\2\t\1/" < "$scp" > "$data/$x/utt2spk"
+    # tab=$'\t'
+    # sed "s/\(s..\)\(.*\)\t.*/\1\2${tab}\1/" < "$scp" > "$data/$x/utt2spk"
```
danpovey: better fix this one too although it's commented.

danpovey:
Thanks!  merging.

PR #1932, opened 2017-10-11
Use KALDI_ROOT env if it's initialized

yamachu:
Previous script, used relative PATH that used `pwd`.
When cloning only CSJ-recipe for tuning it,  it is hard to keep the directory structure.

IMO, we should use KALDI_ROOT env that initialized in `path.sh` instead of `pwd`'s relative path.

### DIFF
--- a/egs/csj/s5/local/csj_train_lms.sh
+++ b/egs/csj/s5/local/csj_train_lms.sh
@@ -42,7 +42,7 @@ fi
 text=$1    
 lexicon=$2 
 dir=$3     
-. path.sh
+. ./path.sh
 
 for f in "$text" "$lexicon"; do
   [ ! -f $x ] && echo "$0: No such file $f" && exit 1;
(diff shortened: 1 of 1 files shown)


egs/csj/s5/local/csj_train_lms.sh
```diff
@@ -50,9 +50,9 @@ done
 loc=`which ngram-count`;
 if [ -z $loc ]; then
   if uname -a | grep 64 >/dev/null; then # some kind of 64 bit...
-    sdir=`pwd`/../../../tools/srilm/bin/i686-m64 
+    sdir=${KALDI_ROOT:-`pwd`/../../..}/tools/srilm/bin/i686-m64
```
danpovey: i think it would make more sense to have the script source path.sh
(`. ./path.sh`)
and then jsut use ${KALDI_ROOT} with no backoff to ../../../.

yamachu: Thank you for your advice.
Looks good to me.

yamachu:
Should I squash commits?

egs/csj/s5/local/csj_train_lms.sh
```diff
@@ -42,6 +42,7 @@ fi
 text=$1    
 lexicon=$2 
 dir=$3     
+. path.sh
```
danpovey: do `. ./path.sh`
(in case there is a path.sh in a directory on the current PATH, or . is not on the path).
no need to squash
I'll merge to-morrow

PR #1935, opened 2017-10-11
Replaced all sed commands using \t and \n with perl commands for OS X…

xiaohui-zhang:
… compatiblity issues.

### DIFF
--- a/egs/babel/s5b/local/train_g2p.sh
+++ b/egs/babel/s5b/local/train_g2p.sh
@@ -55,7 +55,7 @@ fi
 if [ ! -z $icu_transform ] ; then
   paste \
     <(cat $lexicon | awk '{print $1}' | uconv -f $encoding -t $encoding -x "$icu_transform") \
-    <(cat $lexicon | sed $'s/^[^ \t][^ \t]*[ \t]//g') \
+    <(cat $lexicon | perl -ape 's/^[^ \t][^ \t]*[ \t]//g;') \
   > $wdir/lexicon_transformed.txt
   lexicon=$wdir/lexicon_transformed.txt
 fi
--- a/egs/chime1/s5/local/chime1_prepare_data.sh
+++ b/egs/chime1/s5/local/chime1_prepare_data.sh
@@ -40,9 +40,8 @@ scp="$data/train/wav.scp"
 rm -f "$scp"
 for sid in `seq 34`; do
   sid2=`printf "s%02d" $sid`
-  tab=$'\t'
   ls -1 $wav_train/id$sid/*.wav \
-    | sed "s/\(.*\)\/\(.*\).wav/${sid2}_\2${tab}\1\/\2.wav/" \
+    | perl -ape "s/(.*)\/(.*).wav/${sid2}_\2\t\1\/\2.wav/;" \
     | sort >> $scp
 done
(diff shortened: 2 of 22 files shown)


egs/babel/s5b/local/train_g2p.sh
```diff
@@ -55,7 +55,7 @@ fi
 if [ ! -z $icu_transform ] ; then
   paste \
     <(cat $lexicon | awk '{print $1}' | uconv -f $encoding -t $encoding -x "$icu_transform") \
-    <(cat $lexicon | sed $'s/^[^ \t][^ \t]*[ \t]//g') \
+    <(cat $lexicon | perl -ape 's/^[^ \t][^ \t]*[ \t]//g;') \
```
danpovey: because this regex is more complicated you should check with some example input that it does what you think it does (different progrmas have different conventions).

egs/chime1/s5/local/chime1_prepare_data.sh
```diff
@@ -40,9 +40,8 @@ scp="$data/train/wav.scp"
 rm -f "$scp"
 for sid in `seq 34`; do
   sid2=`printf "s%02d" $sid`
-  tab=$'\t'
   ls -1 $wav_train/id$sid/*.wav \
-    | sed "s/\(.*\)\/\(.*\).wav/${sid2}_\2${tab}\1\/\2.wav/" \
+    | perl -ape "s/(.*)\/(.*).wav/${sid2}_\2\t\1\/\2.wav/;" \
```
danpovey: You can't assume that these advanced features are compatible; you'll have to test it.  I suggest to just figure out a single line that you think should probably match, and try both versions.  However, in this particular case  you could probably leave the sed there.  It's the newline which we really have a problem with, the tab can be dealt with using the existing solution.

egs/chime1/s5/local/chime1_prepare_data.sh
```diff
@@ -53,9 +52,8 @@ for x in "devel" "test"; do
     wav_dir="${!wav_var}"
     for sid in `seq 34`; do
       sid2=`printf "s%02d" $sid`
-      tab=$'\t'
       ls -1 $wav_dir/*/s${sid}_*.wav \
-        | sed "s/\(.*\)\/\(.*\)\/s.*_\(.*\).wav/${sid2}_\3_\2${tab}\1\/\2\/s${sid}_\3.wav/" \
+        | perl -ape "s/(.*)\/(.*)\/s.*_(.*).wav/${sid2}_\3_\2$\t\1\/\2\/s${sid}_\3.wav/;" \
```
danpovey: again, for complex things like this you should either test somehow, or leave the sed command.

xiaohui-zhang: yep. I actually tested every complex perl command I wrote...  The major feature not compatible I've encountered that is that "(" is already a special symbol in perl so we don't need the backslash before it.

egs/chime1/s5/local/chime1_prepare_data.sh
xiaohui-zhang: yep that's what I did regarding tests. Do you still want to switch back to our old solution using sed?

PR #1947, opened 2017-10-19
Adding new non-splicing layers in the AMI TDNN setup

freewym:


danpovey:
nice that it gives such a large improvement.

Issue #332, opened 2015-11-16
Documentation needed: decoding just one file


danpovey:
We repeatedly get questions in the forum about how to decode just one file.
It would be good if we had a documentation page or sub-page about this, that would describe the various steps required to do this, in the context of one of the example scripts.

achernetsov:
Daniel, hello!
I want to propose solution, here it is: https://github.com/achernetsov/kaldi-docker-example

The point is: someone need to install docker and run simple script to watch kaldi in action. Docker images heavily commented, to help understand how this all works.

I do not pulled images yet. If you will find this solution acceptable, I can finish it (I need little guidance, where to place this work: im egs, in some other folder, or in separate repository under kaldi-asr), and then I will pull images in dockerhub. I will also create page in documentation.

Also I found it quite ugly that I mapped volume with audio, but I don't get yet how to pass audio as stream, maybe in future I could handle this also.

danpovey:
Hm.
It's a cool project, but I don't think it's something that would be natural
to include in the Kaldi project itself.  Maybe a separate repo that we
could point people to?
Dan

On Fri, Oct 13, 2017 at 2:49 PM, Artyom Chernetsov <notifications@github.com

jtrmal:
Btw, there is the Florians virtual speech kitchen, which has  (AFAIK) a
docker  image with kaldi
Y.

achernetsov:
@jtrmal I created image with intention to clarify what's going on. It is not optimal and with lots of comments. This is quick start example for newcomers. 
But thanks for info about speech kithen!

victorlock:
I’m in

PR #1952, opened 2017-10-20
example of decoding 1 file added to documentation

achernetsov:
Created demonstration of decoding 1 file, using prebuild images (with lots of description) and just 1 shell script to run; Work is located in repo: https://github.com/achernetsov/kaldi-docker-example;

Write 1 page and updated pages index

#332

### DIFF
--- a/src/doc/example_decode_one_file.dox
+++ b/src/doc/example_decode_one_file.dox
@@ -1,0 +1,36 @@
+// doc/example_decode_one_file.dox
+
+// Copyright 2009-2011 Microsoft Corporation
+//           2013-2014 Johns Hopkins University (author: Daniel Povey)
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


danpovey:
Thanks-- I want to think about this for a bit and see if people have comments.
I wonder whether this is the right place to put it-- I have to think about that.
And it should make clear that it's an external contribution so people don't expect the core Kaldi team to debug issues with it.

achernetsov:
@danpovey it's probably better to place into kaldi-asr/newbie-docker-example repository, to be part of project

danpovey:
I don't have much experience with docker, and I don't even have a machine
where I can run it-- I'd want to test it myself, or have a core Kaldi
person test it, before making it  part of the repo.  Anyway this is not
urgent; I'll leave the issue up and think about it for a while.



On Fri, Oct 20, 2017 at 3:01 PM, Artyom Chernetsov <notifications@github.com

saurabhvyas:
@achernetsov thanks a lot for your example, it is very helpful, since current documentation doesn't address that explicitly, also I think using docker has become increasingly important, with open source projecs, since it avoid dependency and install conflicts, with docker I installed kaldi with just one line command, and manual install on Ubuntu took a lot of time, and had to tweak and adjust a lot of stuff.

Issue #1951, opened 2017-10-20
cmd.sh and path.sh pathnames


danpovey:
can someone please go through the repo with some kind of script and change all instances of
`. cmd.sh` to `. ./cmd.sh`
and the same for path.sh?
Please someone volunteer!

PR #1953, opened 2017-10-20
[egs,scripts] Fix cmd.sh and path.sh pathnames

entn-at:
Addresses #1951. Changes were made using the following two commands:

```
find -name '*.sh' -type f -exec sed -i 's;^. cmd.sh$;. ./cmd.sh;g' "{}" \;
find -name '*.sh' -type f -exec sed -i 's;^. path.sh$;. ./path.sh;g' "{}" \;
```


danpovey:
Thanks!  There were more than I expected.

PR #1955, opened 2017-10-20
rotation for image augmentation WIP

YiwenShaoStephen:


danpovey:
Some issues I notice here are: it should default to cifar10 in the script not cifar100 (next time don't change the script, run with "--dataset cifar100").  And the rotation-with-some-probability thing probably needs a flag for it, e.g. --rotation-prob.

YiwenShaoStephen:
Which scripts do I need to add the new image augmentation options (i.e. rotation-degree, rotation-prob)? Only the resnet_1* or all scripts?

danpovey:
When we add options like this we add a new script, so e.g. copy and modify
resnet_1c to resnet_1d.  No need to do it for the non-resnets right now.

YiwenShaoStephen:
Do I need to also include the changes on the number of filters and epochs in the new recipe that achieves better result? For example, I can make a resnet_1d with rotation only, resnet_1e with changes on epochs and filters, and finally resnet_1f with them together.

danpovey:
Do 1d with rotation only and 1e with more epochs and filters added to 1d.
The 1d may be useful for experimental-turnaround-time reasons.

YiwenShaoStephen:
OK. I notice that in each recipe, there are some lines of comments on the top showing the results of that recipe. For this reason, it might take me some time to get completed results of these new recipes. (My previous experiments are mainly on cifar100)

YiwenShaoStephen:
I've pushed these two recipes. But the results in top few lines are not completed. Only the result in resnet_1e for cifar100 is up-to-date. I will update other information upon getting the results.

YiwenShaoStephen:
All results updated. Ready for merging.

danpovey:
@hhadian, please review this for me.

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -231,6 +241,14 @@ void PerturbImage(const ImageAugmentationConfig &config,
   // [ cos(theta)  -sin(theta)  0
   //   sin(theta)  cos(theta)   0
   //   0           0            1 ]
+  if (RandUniform() <= config.rotation_prob) {
+    BaseFloat theta = (2 * config.rotation_degree * RandUniform() -
+		       config.rotation_degree) / 180.0 * 3.1415926;
```
hhadian: I guess you could use Kaldi constant `M_PI`

src/nnet3bin/nnet3-egs-augment-image.cc
```diff
@@ -231,6 +241,14 @@ void PerturbImage(const ImageAugmentationConfig &config,
   // [ cos(theta)  -sin(theta)  0
   //   sin(theta)  cos(theta)   0
   //   0           0            1 ]
+  if (RandUniform() <= config.rotation_prob) {
```
hhadian: I guess you could use Kaldi's `WithProb()` which makes it nicer.
Not important though.

danpovey:
Actually don't use WithProb().  I think that was an unnecessary feature and
if it could be removed and all instances replaced with RandUniform() < x,
it would be clearer to someone with general computer knowledge.

YiwenShaoStephen:
Thanks. I've replaced 3.1415 with M_PI.

PR #1958, opened 2017-10-21
two small updates 

TianyanZhou:


egs/sre10/v1/local/cosine_scoring.sh
```diff
@@ -22,7 +22,7 @@ scores_dir=$6
 
 mkdir -p $scores_dir/log
 run.pl $scores_dir/log/cosine_scoring.log \
-  cat $trials \| awk '{print $1, $2}' \| \
+  cat $trials \| awk '{print $1" "$2}' \| \
```
danpovey: I'm curious about what version of awk you are using.  I'm not aware of any version for which this would be a syntax error.

danpovey: @TianyanZhou, please notice my comment above RE your version of awk.

TianyanZhou: I'm using GNU Awk 4.0.1. I forget to mention that there is another problem in this script. Line 28: 'ark:ivector-normalize-length scp:${test_ivec_dir}/ivector.scp ark:- |'. Shell cannot refer to variable ${test_ivec_dir} and report an error "Failed to open script file /ivector.scp". When I surround line 28 with a double quotation mark, it can refer to ${test_ivec_dir} successfully but reports "awk syntax error '{print $1, $2}'". And after I change '{print $1, $2}' to '{print $1" "$2}', this script works finally. I agree that "awk '{print $1, $2}" should not be a syntax error. It's kind of weird shell reports this. I also noticed the original command (cat data/sre10_test/trials | awk '{print $1, $2}' \| ) became (cat data/sre10_test/trials | awk "{print $1, $2}" | ) in file "exp/scores_cosine/log/cosine_scoring.log", why it replaces the awk 'print' command with a double quotation mark. I'm not sure whether this is because my awk version or not.

egs/sre10/v1/local/cosine_scoring.sh
```diff
@@ -22,7 +22,7 @@ scores_dir=$6
...
  ivector-compute-dot-products - \
   scp:${enroll_ivec_dir}/spk_ivector.scp \
   'ark:ivector-normalize-length scp:${test_ivec_dir}/ivector.scp ark:- |' \
```
danpovey: Here, the single quotes should be double quotes-- otherwise the shell will not expand the variable.

egs/sre10/v1/local/cosine_scoring.sh
danpovey: This seems to be an issue with  run.pl.  You can fix it in this PR if you want, but revert this change (and test the fix).
You have to replace the lines in run.pl:
```
foreach $x (@ARGV) {
    if ($x =~ m/^\S+$/) { $cmd .=  $x . " "; }
    elsif ($x =~ m:\":) { $cmd .= "'$x' "; }
    else { $cmd .= "\"$x\" "; }
}
```
with the lines beginning `foreach my $x (@ARGV) {` in queue.pl.

TianyanZhou: I tried this, it still failed with the same "awk syntax error".

danpovey: OK, I see why now-- I'll merge your changes as-is.

PR #1959, opened 2017-10-22
[src,scripts,egs] Make it possible to specify component-level l2 regularization

danpovey:


danpovey:
Hm, my previous comment on this PR seems to have been lost somehow.
This is quite an important PR.  It allows us to replace the rather ad-hoc use of '--proportional-shrink'
for l2 regularization, with a per-component specification of the amount of l2 regularization
we want to use.  There are certain components which we do not want to apply l2 regularization
to at all (e.g. the trainable scaling factor that is sometimes associated with batch-norm; regularizing
this can be quite harmful as it makes it too small.  This is not available in the master right now).  
Also, it seems to be quite helpful to use less l2 regularization for the output layer vs. the relu layers.  There are other issues that need attention here:
  - We need to tune the configuration for larger setups than mini_librispeech
Various issues come up in configurations that have LSTMs:
  - Do we need to do l2 on all parts of the LSTM?
  - Should we have the same amount of l2 on LSTM versus TDNN layers?
These same considerations apply to GRUs.  @hhadian it would be good if you can tune this for the image recognition setups.  @GaofengCheng, I hope you can help figure out how to do this for LSTMs and GRUs.  I did not implement the l2-regularize option on LSTM or convolutional layers but it is quite straightforward.


To find the right ballpark value of the l2 regularization constant to begin tuning, a good rule of thumb is to compare it with a "good" working setup, train a few iterations, and grep for "Norms" in one of the progress.X.log files.  You want these to have about the same values as the good working baseline.   If they are too large you need to increase the amount of l2 regularization; if too small, you may need to decrease it.  Once you get close to a good value you may want to tune on WER, but start by changing it in factors of two.

We also need to tune this for regular nnet3 (non-chain) setups.  There is an interaction with the chunk size; setups with fewer labels per chunk (like TDNNs for nnet3) will need much smaller amounts of l2 regularization.  

I should mention that @keli78 has been working with something not that different from this in the RNNLM setup, and @pegahgh suggested to implement it at the component level.

pegahgh:
@danpovey , thnx, I had same concern specially in direct-from-signal setup, where I just need to apply l2-regularization on convolution layer rather than all other affine components. 
I will test this PR on my best setup.

PR #1962, opened 2017-10-24
nnet1: various minor changes,

vesis84:
- ivectors: filter out speakers with no i-vectors,
- steps/nnet/train.sh: allow by-pass of phone-set check,
- nnet-randomizer.cc: make '# of extra frames' relative to 'capacity',
- nnet-train-frmshuff: introduce --max-frames (limits utterance length),


egs/wsj/s5/utils/subset_data_dir_tr_cv.sh
```diff
@@ -20,6 +20,8 @@
 # especially for the cases where dataset contains multiple different corpora,
 # where type of speakers or recording channels may be quite different 
 
+echo "$0 This script is deprecated, please use: 'utils/nnet/subset_data_tr_cv.sh'"
+
```
danpovey: If the interfaces are compatible, can you make this a soft link?

src/nnet/nnet-randomizer.cc
```diff
@@ -64,9 +64,11 @@ void MatrixRandomizer::AddData(const CuMatrixBase<BaseFloat>& m) {
   }
   // extend the buffer if necessary,
   if (data_.NumRows() < data_end_ + m.NumRows()) {
-    CuMatrix<BaseFloat> data_aux(data_);
-    // Add extra 1000 rows, so we don't reallocate soon:
-    data_.Resize(data_end_ + m.NumRows() + 1000, data_.NumCols());
+    // CuMatrix -> Matrix -> CuMatrix (needs less GPU memory),
+    Matrix<BaseFloat> data_aux(data_);
+    // Add extra 3% rows, so we don't reallocate soon:
```
danpovey: typically in situations like this (for a growing buffer) one would increase by a factor closer to 1.5 or 2, like stl vector does.  Or simply allocate the rows separately as vectors, if feasible- but there may be some reason why this is impractical.  I'm not sure if something is different here...

egs/wsj/s5/utils/nnet/subset_data_tr_cv.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash +x
```
danpovey: I don't think you mean to check it in with "+x".

src/nnet/nnet-randomizer.cc
vesis84: The 'data_' buffer is supposed to have a 'fixed size', plus extra frames coming from 'last utterance', plus the 3% as a reserve for even longer utterances. The buffer is quite large (32k rows or more) and the dimensions can be large as well. So the 3% are there to give some extra space without consuming too much memory...

egs/wsj/s5/utils/nnet/subset_data_tr_cv.sh
vesis84: '+x' does nothing, it can be manually changed to '-x' which enables printing of the expanded bash commands, line by line as they are executed,

egs/wsj/s5/utils/subset_data_dir_tr_cv.sh
vesis84: okay, I'll merge them and create symlink...

vesis84:
The updates are finished...

PR #1966, opened 2017-10-24
Correct the data dir for Chime4 

Szu-JuiChen:
The merged method can not be used on CLSP grid.  I checked that this method can be execute on CLSP grid. But I am not sure if this is portable.


danpovey:
Actually it seems that if you use the '-f' option, and the match for *.clsp.jhu.edu, it will work on the CLSP grid but also on mac.

Szu-JuiChen:
Yes, you are right. Files are modified as your suggestion. Thanks!

PR #1968, opened 2017-10-24
[egs,scripts] Use command passed as --egs.cmd for get_egs.sh instead of command passed as --cmd

entn-at:
This PR fixes a minor issue with the acoustic model training libraries. The command passed as --egs.cmd was not passed to get_egs.sh. In raw_model.py, the command passed as --egs.cmd is being passed correctly to get_egs_targets.sh.

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/train/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/common.py
@@ -922,8 +922,7 @@ class CommonParser(object):
                                  """, default="queue.pl")
         self.parser.add_argument("--egs.cmd", type=str, dest="egs_command",
                                  action=common_lib.NullstrToNoneAction,
-                                 help="Script to launch egs jobs",
-                                 default="queue.pl")
+                                 help="Script to launch egs jobs")
         self.parser.add_argument("--use-gpu", type=str,
                                  action=common_lib.StrToBoolAction,
                                  choices=["true", "false"],
--- a/egs/wsj/s5/steps/libs/nnet3/train/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/common.py
@@ -922,7 +922,8 @@ class CommonParser(object):
                                  """, default="queue.pl")
         self.parser.add_argument("--egs.cmd", type=str, dest="egs_command",
                                  action=common_lib.NullstrToNoneAction,
-                                 help="Script to launch egs jobs")
+                                 help="Script to launch egs jobs",
+                                 default="queue.pl")
         self.parser.add_argument("--use-gpu", type=str,
                                  action=common_lib.StrToBoolAction,
                                  choices=["true", "false"],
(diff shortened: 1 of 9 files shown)


egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -922,7 +922,8 @@ def __init__(self,
                                  """, default="queue.pl")
         self.parser.add_argument("--egs.cmd", type=str, dest="egs_command",
                                  action=common_lib.NullstrToNoneAction,
-                                 help="Script to launch egs jobs")
+                                 help="Script to launch egs jobs",
+                                 default="queue.pl")
```
danpovey: This change is not right, it will break the example scripts.  
Look for statements like this:
```
    run_opts.egs_command = (args.egs_command
                            if args.egs_command is not None else
                            args.command)
```

danpovey: BTW I don't like the way the code is currently (as pointedout in my comment above),
 as in-place modification of variables is confusing.  I'd prefer it if where `args.egs_command` is used, we instead use
`(args.egs_command if args.egs_command is not None else args.command)`

entn-at:
Thanks! I reverted the change adding a default value.
I added a `cmd` function parameter to `generate_egs`/`generate_chain_egs`/`generate_egs_using_targets`. Where these methods are called, I passed `(args.egs_command if args.egs_command is not None else args.command)` to the new `cmd` parameter.

egs/wsj/s5/steps/libs/nnet3/train/frame_level_objf/raw_model.py
```diff
@@ -18,7 +18,7 @@
 
 
 def generate_egs_using_targets(data, targets_scp, egs_dir,
-                               left_context, right_context,
+                               left_context, right_context, cmd,
```
danpovey: I don't understand why cmd needs to be provied here if it's available from run_opts.
best to figure out the command at the innermost level rather than changing interfaces.

entn-at: Perhaps I misunderstood your previous comment about the assignment to `run_opts.egs_command`. `args`, which contains `egs_command`, is not being passed to `generate_egs`, only `run_opts` is. That's presumably why `run_opts.egs_command` is assigned to either `args.egs_command` or `args.command` if the former is `None`. 

If you prefer, I can revert the changes introducing the `cmd` parameter and use `command=(run_opts.egs_command if run_opts.egs_command is not None else run_opts.command)` inside `generate_egs` where the command line for `get_egs.sh` is being built.

danpovey:
Sorry, I misunderstood the original code when I scanned it- I thought it was modifying a variable, but it was assigning a new variable.  So the original code was mostly OK.  Just make your minimal original changes, except for the changed default to "queue.pl"

entn-at:
OK, I'm closing this PR and I'm going to create a new one with only the minimal set of changes.

PR #1974, opened 2017-10-26
Remove download of ATLAS header files from tools/.

galv:
Commit the header files cblas.h and clapack.h from ATLAS directly
instead for simplicity.

Use a new directory for the atlas header files. This way, when people
pull this commit from github, they won't get an error from git saying
that the pull will overwrite untracked files (namely, clapack.h and
cblas.h).

Also remove custom installation of ATLAS from tools/ since it's rarely
used and apparently broken at the moment.


galv:
I don't fully justify myself in the commit message, but I anticipate this to be 100% backwards compatible. Let me know if you have any concerns.

danpovey:
I am testing right now, I hope to commit soon.

PR #1980, opened 2017-10-29
register max_warnings option, referred to in parse error message

jimregan:


src/lmbin/arpa2fst.cc
```diff
@@ -67,6 +67,8 @@ int main(int argc, char *argv[]) {
                 "would be lost entirely)");
     po.Register("ilabel-sort", &ilabel_sort,
                 "Ilabel-sort the output FST");
+    po.Register("max_warnings", &options.max_warnings,
```
danpovey: Registering options that are part of options classes is done inside the options class.  Look at its 'Register' function.

jimregan:
Ah, ok. I fixed the warning instead

PR #1982, opened 2017-10-30
Add a tdnn+lstm+attention+backstitch recipe for tedlium

hhadian:


danpovey:
Thanks- I'll wait to see if @freewym has any comments before I merge.

egs/tedlium/s5_r2/local/chain/tuning/run_tdnn_lstm_attention_bs_1a.sh
```diff
@@ -0,0 +1,347 @@
...
+  input dim=40 name=input
+
+  # please note that it is important to have input layer with the name=input
+  # as the layer immediately preceding the fixed-affine-layer to enable
+  # the use of short notation for the descriptor
+  fixed-affine-layer name=lda input=Append(-2,-1,0,1,2,ReplaceIndex(ivector, t, 0)) affine-transform-file=$dir/configs/lda.mat
+
+  # the first splicing is moved before the lda layer, so no splicing here
+  relu-renorm-layer name=tdnn1 dim=1024
+  relu-renorm-layer name=tdnn2 dim=1024 input=Append(-1,0,1)
```
danpovey: hm.  I think it would be better if you replaced these renorm layers with batchnorm layers.  You'd probably get a small improvement.
And while you're at it, why not try removing the proportional-shrink and replacing it with component-level l2 regularization?  You can use my recently checked-in WSJ setup (the TDNN one) for reference.  In general if there is more data we'll want a little bit less l2 but I think tedlium and WSJ are quite similar in size so the same values might work (or try them divided by 2).

If we're updating scripts I think we should update them all the way.

freewym:
The backstitch part looks good to me.

egs/tedlium/s5_r2/local/chain/tuning/run_tdnn_lstm_attention_bs_1a.sh
hhadian: Will do. What about attention's renorm layer? Do you want me to replace that as well?

danpovey:
Yes, replace that too.
batchnorm is always better than renorm (unless recurrence is involved) so
you might as well just hardcode the change in the script.
Dan

vince62s:
is there any rason why this is still pending ?
also, does it make sense to compare a model with 13M parameters vs 33M ?

danpovey:
It's still pending because we're trying to get version with l2-regularize instead of relying on 'shrink-value'.
RE comparing different model sizes-- the idea is that for each model type, we tune it for optimal WER.  This can lead to quite different numbers of parameters.  The models with backstitch tend to optimally have more parameters because it's kind of a regularizer.  These models are also slower to train because they tend to reach optimal performance after more epochs, so it's something we'll tend to do only when we really care about reaching the best WER.

vince62s:
the layer dimension difference is very significant. is there an obvious reason why the other baselines (tdnn, tdnn_lstm_attend) would overfit with a layer size of 1024, and this setup would not ? is BS the only impact? regularization ? thanks.

danpovey:
with backstitch we normally try larger models as it prevents overfitting.

egs/tedlium/s5_r2/local/chain/tuning/run_tdnn_lstm_attention_bs_1a.sh
ngoel17: Did you ever try batchnorm? Did it really help? How about component level regularization?

danpovey:
Hossein is working on tuning a setup with batchnorm and component-level
regularization.  So far he has not got better numbers so we haven't checked
it in yet.

vince62s:
Dan, when looking at the PR you said "replace prop shrink by l2 regul" but actually there was no proportional shrink in the PR, am I right ?
Also, do you know what l2 values were tested ? we are conducting similar experiences on double the amount of hours, so I would like to know what suits the best.

danpovey:
Vincent:
I don't recall w.r.t. the proportional shrink, whatever you say...
Regarding the l2 values, here is some off-github correspondence between
Hossein and myself:

====
I tried it (indicated below by *e*):

# System                      a          c        d         e
# WER on dev(orig)            8.0       8.4       8.2       8.2
#         [looped:]           8.0       8.4       8.2       8.2
# WER on dev(rescored)        7.4       7.9       7.5       7.6
#         [looped:]           7.4       7.9       7.5       7.6
# WER on test(orig)           8.2       8.3       8.1       8.0
#         [looped:]           8.2       8.3       8.2       8.2
# WER on test(rescored)       7.7       7.8       7.7       7.6
#         [looped:]           7.7       7.8       7.8       7.7
# Final train prob        -0.0605   -0.0900   -0.0751   -0.0698
# Final valid prob        -0.0842   -0.0946   -0.0853   -0.0856
# Final train prob (xent)   -0.7602   -1.0608   -0.9333   -0.8515
# Final valid prob (xent)   -0.8740   -1.0950   -0.9900   -0.9321


a: the unchanged script (i.e. with renorm and fixed shrink value 0.99)
c: using batchnorm + l2-regularize (*0.005* for tdnn and attention layers
and *0.001* for LSTM and output layers)
d: same as 1c but with l2-regularize *0.0015* for tdnn and attention layers
and *0.0005* for LSTM and output layers
e: d: same as 1c but with l2-regularize *0.0005* for tdnn and attention
layers and *0.0002* for LSTM and output layers

Norms:
*a:* Norms of parameter matrices are [ tdnn1.affine:44.9218
tdnn2.affine:54.7379 lstm1.W_all:97.368 lstm1.lstm_nonlin:47.4662
lstm1.W_rp:39.5179 tdnn3.affine:53.9454 tdnn4.affine:54.9937
lstm2.W_all:94.147 lstm2.lstm_nonlin:48.2632 lstm2.W_rp:44.7119
tdnn5.affine:58.0611 tdnn6.affine:59.836 attention1.affine:64.4435
output.affine:55.752 output-xent.affine:196.565 ]
*d:* Norms of parameter matrices are [ tdnn1.affine:19.9774
tdnn2.affine:23.385 lstm1.W_all:48.1358 lstm1.lstm_nonlin:19.5765
lstm1.W_rp:25.7464 tdnn3.affine:20.646 tdnn4.affine:22.1167
lstm2.W_all:51.131 lstm2.lstm_nonlin:22.5379 lstm2.W_rp:31.0513
tdnn5.affine:21.7788 tdnn6.affine:25.1947 attention1.affine:28.33
output.affine:33.3455 output-xent.affine:79.8698 ]
*e:* Norms of parameter matrices are [ tdnn1.affine:30.3378
tdnn2.affine:39.933 lstm1.W_all:76.2599 lstm1.lstm_nonlin:35.9059
lstm1.W_rp:36.0933 tdnn3.affine:36.1792 tdnn4.affine:36.536
lstm2.W_all:74.9741 lstm2.lstm_nonlin:37.244 lstm2.W_rp:40.2443
tdnn5.affine:36.5974 tdnn6.affine:39.5641 attention1.affine:45.6915
output.affine:39.7231 output-xent.affine:109.688 ]

Should I try even smaller l2-regularize factors?
Daniel Povey <dpovey@gmail.com>
Nov 25 (9 days ago)
to Hossein
Yes, try approximately halving them again.  It still seems a little
over-regularized.

vince62s:
sorry but I must be mistaken by something, I thought prop-shrinnk was set by:
--trainer.optimization.proportional-shrink 20
(as I saw in some other setups)
but in the PR, there is:
--trainer.optimization.shrink-value 0.99
is this just a different way to set up proportional shrink ?

on l2, I didn't know there should be 2 different values for tdnn vs lstm....
I see now a good reason why our run is too much regularized.
thanks

danpovey:
proportional-srhink and shrink-value are not really equivalent but they
both result in shrinkage.  The --shrink value applies it conditionally, if
it detects that on average, sigmoid and tanh neurons are too saturated.

PR #1984, opened 2017-10-30
Add OCR/Handwriting Recognition examples

hhadian:


egs/iam/s5/local/unk_arc_post_to_transcription.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import sys
+import numpy as np
+from scipy import misc
+parser = argparse.ArgumentParser(description="""uses phones to convert unk to word""")
```
pegahgh: It is good to add extended description for this function and its arguments. This script can be used in other applications.

egs/iam/s5/local/chain/run_cnn_1a.sh
```diff
@@ -0,0 +1,231 @@
...
+nnet3_affix=    # affix for exp dirs, e.g. it was _cleaned in tedlium.
+affix=_1a  #affix for TDNN+LSTM directory e.g. "1a" or "1b", in case we change the configuration.
+ali=tri3_ali
+common_egs_dir=
+reporting_email=
+
+# chain options
+train_stage=-10
+xent_regularize=0.1
+frame_subsampling_factor=4
```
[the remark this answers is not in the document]
hhadian: It has not been tested yet. The reason for choosing
a larger factor was that the average number of frames per
word in OCR (when the line images are scaled to have a height of 40) is almost 2 times that of ASR.

egs/iam/s5/local/chain/run_cnn_1a.sh
```diff
@@ -0,0 +1,231 @@
...
+alignment_subsampling_factor=1
+# training chunk-options
+chunk_width=340,300,200,100
```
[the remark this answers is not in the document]
hhadian: We can change this to something less unusual like `300,200,100`
Again, the reason for choosing this is because the average number of frames per phone/letter is almost 2 times larger for OCR.

danpovey:
Actually I prefer it when the chunk widths are not too regularly spaced...
then we get more combinations so we can more closely approximate the
lengths of longer utterances.  That's why I sometimes use slightly
random-seeming numbers.

egs/iam/s5/local/chain/run_cnn_1a.sh
```diff
@@ -0,0 +1,231 @@
...
+where "nvcc" is installed.
+EOF
+fi
+
+gmm_dir=exp/${gmm}
+ali_dir=exp/${ali}
+lat_dir=exp/chain${nnet3_affix}/${gmm}_${train_set}_lats
+dir=exp/chain${nnet3_affix}/cnn${affix}
+train_data_dir=data/${train_set}
+lores_train_data_dir=$train_data_dir  # for the start, use the same data for gmm and chain
```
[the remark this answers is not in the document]
hhadian: This was modified from an ASR recipe and we didn't remove this
variable so we could optionally experiment with different resolutions for
the gmm and chain systems. Anyway, the gmm system does not give good results
so I guess we can remove this and focus on the chain setup only.

hhadian:
> Actually I prefer it when the chunk widths are not too regularly spaced...
then we get more combinations so we can more closely approximate the
lengths of longer utterances.  That's why I sometimes use slightly
random-seeming numbers.

Oh OK. I won't change it then.

egs/iam/s5/local/score.sh
```diff
@@ -0,0 +1,154 @@
+#!/bin/bash
```
jtrmal: this should be a symlink to steps/score_kaldi.sh, not a copy

jtrmal:
BTW, lot of the files does not have authors and copyrights -- please add those.

egs/iam/s5/cmd.sh
```diff
@@ -0,0 +1,15 @@
...
+# commands one by one: most recipes will exhaust the memory on your
+# machine).  queue.pl works with GridEngine (qsub).  slurm.pl works
+# with slurm.  Different queues are configured differently, with different
+# queue names and different ways of specifying things like memory;
+# to account for these differences you can create and edit the file
+# conf/queue.conf to match your queue's configuration.  Search for
+# conf/queue.conf in http://kaldi-asr.org/doc/queue.html for more information,
+# or search for the string 'default_config' in utils/queue.pl or utils/slurm.pl.
+
+export cmd="queue.pl"
```
jtrmal: cmd is definitely weird and non-standard. If you have any use-case for it, please name it more self-descriptively.

egs/iam/s5/local/prepare_lm.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
jtrmal: I think this file should be replaced by utils/format_lm.sh (or at least most of the things from this file should be replaced by that one)

egs/uw3/v1/local/prepare_lm.sh
```diff
@@ -0,0 +1,62 @@
...
+
+  compile-lm $tmpdir/tmp_lm.ilm.gz -t=yes /dev/stdout | \
+    grep -v unk > $tmpdir/lm_phone_bg.arpa
+
+  arpa=$tmpdir/lm_phone_bg.arpa
+fi
+
+cat $arpa | utils/find_arpa_oovs.pl $lang/words.txt  > $tmpdir/oovs.txt
+
+cat $arpa | \
```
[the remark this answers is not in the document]
jtrmal: I think more preferable way would be just call utils\format_lm.sh (I noticed in the previous recipe)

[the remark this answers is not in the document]
jtrmal: Also, it seems you are osciallating between IRSTLM and pocolm? is that necessary? Cannot you use just one toolkit?

egs/uw3/v1/local/score.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+[ -f ./path.sh ] && . ./path.sh
+
+# begin configuration section.
+cmd=run.pl
+stage=0
+decode_mbr=false
+stats=true
```
jtrmal: perhaps you should add `stats` to scoring_opts as well.
Also, if don't wanna do it the way suggested in kaldi_score_cer.sh (I assume because you want to modify the default lmwts, even though I'm not sure if it's worth the efort), please make sure the stage parameter is processed correctly

egs/iam/s5/cmd.sh
danpovey: Actually I think I am ok with just using "$cmd". The distinction between train_cmd and decode_cmd became less necessary now that we have a common interface for those tools-- we mostly keep them around just out of inertia.

jtrmal:
I agree but now we would have three instead of using just one?
y.

danpovey:
But in this particular example it looks like there would be just one.
The only way to start using just one is to do it example by example-- too
much hassle to upgrade existing scripts.

jtrmal:
Ok, I thought I see three different variables definition. But if there is
only one, then I don't have any complaints
Y

egs/iam/s5/local/unk_arc_post_to_transcription.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+
+import argparse
```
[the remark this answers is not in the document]
xiaohui-zhang: will do

egs/uw3/v1/local/prepare_lm.sh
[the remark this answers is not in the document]
hhadian: Yes, this is very old. I'll fix it.
Re LM toolkits, I'm not sure which toolkit to use.
Here (i.w. UW3) we just need a simple LM trained on training text.
But in IAM, we have 3 LM sources and pocolm might be more suited.

egs/uw3/v1/local/score.sh
hhadian: @ChunChiehChang, could you please change this script
to follow the instructions mentioned in the header of `steps/scoring/score_kaldi_cer.sh`?

egs/iam/s5/local/score.sh
hhadian: It's because it calls `local/unk_arc_post_to_transcription.py` somewhere
in the scoring pipeline. However, looking at the script, 
it seems to me that it can be done
through `hyp_filtering_cmd`. @aarora8, could you please make it
a symlink and do the "unk scoring" using `hyp_filtering_cmd`?

egs/iam/s5/local/chain/align_nnet3_lats.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
hhadian: Should we move this script to `steps/nnet3`?

danpovey: See how it differs from steps/nnet3/align_lats.sh.

egs/uw3/v1/local/process_data.py
```diff
@@ -0,0 +1,57 @@
...
+    if line.endswith('.txt'):
+      text_path = os.path.join(args.database_path, page, line)
+      image_name = line.split('.')[0]
+      image_path = os.path.join(args.database_path, page, image_name + '.png')
+      utt_id = page + '_' + image_name
+      gt_fh = open(text_path, 'r')
+      text = gt_fh.readlines()[0].strip()
+      
+      coin = random.randint(0, 20)
+      if coin >= 1:
```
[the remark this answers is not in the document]
ChunChiehChang: Okay, it was done to separate the training and testing data as I couldn't find any mention of official training and testing sets for UW3

egs/uw3/v1/local/prepare_lm.sh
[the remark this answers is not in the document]
danpovey: I think I'm OK with using 2 different LM toolkits since it's 2 different recipes.  The only benefit in having a standard is if it's globally enforced, and that isn't practical for various reasons (there just isn't a clear leader).

egs/iam/s5/local/chain/align_nnet3_lats.sh
hhadian: Actually, they are basically the same, however `steps/nnet3/align_lats.sh` was not there at the time.
@aarora8, could you please update the chainali
recipes to use `steps/nnet3/align_lats.sh` instead and remove this script?

jtrmal:
I'm not gonna argue any more, but I don't see reason why they _should_ be
different, given the fact that the recipes are prepared by the same team
(and in the same PR) -- I think the more things will be idiomatic and fixed
across related recipes, the better (of course only where it makes sense).
y.

egs/iam/s5/local/chain/align_nnet3_lats.sh
aarora8: done, made the change for the pull request.

egs/iam/s5/local/chain/run_cnn_1a.sh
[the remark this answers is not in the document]
aarora8: Best word error rate with subsampling factor(FSF) 4 is slightly better than 3 and 5. WER(%) with FSF=3,5 is close to 14.80% and for FSF=4 it is close to 14.50%.

FSF | Best WER(%)
-- | --
3 | 14.81
4 | 14.51
5 | 14.80

egs/iam/s5/local/score.sh
aarora8: ok, it is also converting upper case to lower case. But in the mail response Paul from RWTH-aachen mention that   they are not converting upper case to lower case. should I not include the conversion in the change as well.

egs/iam/s5/local/unk_arc_post_to_transcription.py
```diff
@@ -0,0 +1,86 @@
...
+  if word == unk_val: # get character sequence for unk
+    phone_key_vect = phones.split(" ")
+    phone_val_vect = list()
+    for pkey in phone_key_vect:
+      phone_val_vect.append(phone_dict[pkey])
+    phone_2_word = list()
+    for phone_val in phone_val_vect:
+      phone_2_word.append(phone_val.split('_')[0])
+    phone_2_word = ''.join(phone_2_word)
+    utt_word_dict[uttID][count] = phone_2_word
```
xiaohui-zhang: This part is the core of this script and definitely needs more explanation,.e.g. "Since in OCR, the lexicon is purely graphemic, we can just concatenate the phones from the most probable phone sequence given by the unk-modle to produce the predicted word.", because usually we need a P2G model in order to map a phone sequence to a predicted word.

xiaohui-zhang: Also please add some explanation at the header,.e.g. "then it will replace 
 the \<unk\> with the word predicted by \<unk\> model" => "then it will replace 
 the \<unk\> with the word predicted by \<unk\> model by concatenating phones decoded from the unk-model."

aarora8: Done, thanks.

egs/iam/s5/local/unk_arc_post_to_transcription.py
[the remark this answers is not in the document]
xiaohui-zhang: @danpovey BTW after I figured out the purpose and context of this script, I agree that its function definitely makes sense (I've asked Ashish to add more explanation though). There's no other existing solution for now. I have some scripts (not committed yet) which pick up unks from arc-post files, and then insert recognized unks from P2G into a ctm. But they are not very related since we don't need P2G here.

egs/iam/s5/cmd.sh
danpovey: you should probably remove either $cmd, or $train_cmd and $decode_cmd.

danpovey:
I haven't merged this because it's still marked WIP.  Let me know when you think it's ready.

PR #1987, opened 2017-10-31
Basic setup for IARPA Material corpora

jtrmal:
This does not train anything, just sets up the kaldi data directories and prepares basic test LM


danpovey:
Wow, that was fast!

@freewym, can you please do something for us-- can you take the branch in this PR and use it as a starting point for a more complete system build?

You can use the mini-librispeech setup as a guide for the structural elements, but for tuning purposes you may want to use other examples, based on data size (I don't recall right now how much data this has).

This is kind of time-sensitive.

jtrmal:
its around 40 hrs
I will still mess around the recipe, but at least it will get us running.
some of the settings for nnet3 can be taken from the babel/s5d recipe, I
guess, I have updated recipes in
`/export/a09/jtrmal/kaldi-babel/egs/babel/s5d/`
I managed to track the progress upto backstitch and proportional-shrink
introduction, after that, I got swamped by other things.
y.

freewym:
OK. Will do soon.

jtrmal:
BTW, let's not overtune -- it is my belief we will be getting some other
data from broadcast news, after which point we will have to decide how  to
proceed.
So let's not spend too much time on fine tuning of the parameters, as it
might be very transitory setup
y.

jtrmal:
@freewym @danpovey I did some cleanup of the data preparation (regarding your concerns about creating unnecessary files and the `~` character) and actually found utf-8 encoding issue in lexicon preparation, which all should be resolved now.

jtrmal:
@freewym any progress on this, btw?

freewym:
Thanks, will re-run based on your latest changes. 

After 4 stages of mono phone/triphone deltas/LDA+MLLT/SAT training, WER on dev is 64.47. Does that looks reasonable?

jtrmal:
For swahili, I have `/export/b19/jtrmal/babel/202-swahili-flp-g-ext` that
gives around 59 % WER
For tagalog, I have `/export/b17/jtrmal/babel/106-tagalog-flp-g-ext` that
gives around 63.18 % WER
those are babel system and your setup might be different, but I'd say you
are within the ballpark
y.

jtrmal:
@freewym I have added a filter for scoring of the decoded text -- can you please merge (should be painless, as it's a new file) and rescore some of the systems you care about?
What the filter does is that it deletes all non-speech events before scoring -- typically, you don't care about them.

freewym:
OK

freewym:
@jtrmal After filtering, the WER was reduced from 44.43 to 39.14, using the chain tdnn-lstm recipe.

jtrmal:
ok, I have 41.x on a tdnn-lstm without dropout and backstitch, so I guess
we're good
```
/export/b19/jtrmal/babel/202-swahili-flp-p-basic/exp/chain_cleaned/tdnn_lstm_sp/decode_dev10h.pem
```
BTW, don't forget to try the other language (tagalog, iirc) as well.
y.

jtrmal:
I think there is no point in keeping this open

PR #1995, opened 2017-11-02
The outline of fvector

LvHang:
This RP is reserved for fvector project.
Now, it just is the outline of the work. A part is the code, and the other is the comment.
In fvector_run.sh, I just write the core function. It will be instead by a bash script respectively.
The binearies in fvector/fvectorbin are used to cut the wavdata into chunk, add noise for each chunk and generate the NnetExample.
The  specific code which implement the comments is writing and will be finished in this week.

### DIFF
--- a/src/fvector/fvector-perturb.cc
+++ b/src/fvector/fvector-perturb.cc
@@ -1,0 +1,289 @@
+#include "fvector/fvector-perturb.h"
+
+namespace kaldi {
+
+void FvectorPerturb::ApplyPerturbation(const MatrixBase<BaseFloat>& input_chunk,
+                                       Matrix<BaseFloat>* perturbed_chunk) {
+  // The original_dim_matrix is a matrix whose dimension is same with input_chunk.
+  // Assume the sample_frequency=8kHz, the original length is 120ms.
+  // It will be a (4, 960) matrix.
+  Matrix<BaseFloat> original_dim_matrix(input_chunk);
+  // Firstly, we add additive noise with probability.
+  AddNoise(opts_.add_noise, &original_dim_matrix);
+  // we do Resize() here, because Resize() belongs to Matrix<> rather than MatrixBase<>
+  original_dim_matrix.Resize(2, original_dim_matrix.NumCols(), kCopyData);
+  KALDI_ASSERT(original_dim_matrix.NumRows() == 2);
+  // After AddNoise(), the shape of original_dim_matrix is (2, original_dim).
+  if (opts_.volume_perturbation) {
+    VolumePerturbation(&original_dim_matrix);
+  }
+  // The expected_dim_matrix is a matrix (input_chunk.NumRows(), expected-chunk-length
+  // * sample_frequency / 1000). E.g. it is a (4, 800) matrix.
+  Matrix<BaseFloat> expected_dim_matrix(original_dim_matrix.NumRows(),
+      opts_.expected_chunk_length * opts_.sample_frequency / 1000);
+  if (opts_.speed_perturbation) {
+    //1. generate speed perturb factor randomly(Noice: the expected_length is
+    //always smaller than original_length) for each line.
+    //(1) a=min{original_length/expected_length -1, max-speed-perturb-rate}
+    //(2) the range of factor is (1-a, 1+a)
+    BaseFloat boundary = std::min(static_cast<BaseFloat>((original_dim_matrix.NumCols() * 1.0 / opts_.sample_frequency)
+          * 1000 / opts_.expected_chunk_length - 1), opts_.max_speed_perturb_rate);
+    for (MatrixIndexT i = 0; i < original_dim_matrix.NumRows(); ++i) {
+      //caculate the speed factor
+      BaseFloat factor =static_cast<BaseFloat> (RandInt(
+          (int)((1-boundary)*100),(int)((1+boundary)*100)) * 1.0 / 100.0);
+      
+      Vector<BaseFloat> speed_input_vector(original_dim_matrix.Row(i));
+      
+      MatrixIndexT speed_output_dim = static_cast<MatrixIndexT>(ceil(original_dim_matrix.NumCols() / factor));
+      KALDI_ASSERT(speed_output_dim >= opts_.expected_chunk_length * opts_.sample_frequency / 1000);
+      Vector<BaseFloat> speed_output_vector(speed_output_dim);
+
+      SpeedPerturbation(speed_input_vector, opts_.sample_frequency, factor, &speed_output_vector);
+      
+      Vector<BaseFloat> time_shifted_vector(expected_dim_matrix.NumCols());
+      if (opts_.time_shift) {
+        TimeShift(speed_output_vector, &time_shifted_vector);
+      } else {
+        time_shifted_vector.CopyFromVec(speed_output_vector.Range(0, expected_dim_matrix.NumCols()));
+      }
+      expected_dim_matrix.CopyRowFromVec(time_shifted_vector, i); 
+    }
+  } else { //no speed_perturbation
+    if (opts_.time_shift) {
+      for (MatrixIndexT i = 0; i < original_dim_matrix.NumRows(); ++i) {
+        Vector<BaseFloat> input_vector(original_dim_matrix.Row(i));
+        Vector<BaseFloat> time_shifted_vector(expected_dim_matrix.NumCols());
+        TimeShift(input_vector, &time_shifted_vector);
+        expected_dim_matrix.CopyRowFromVec(time_shifted_vector, i);
+      }  
+    } else {
+      expected_dim_matrix.CopyFromMat(original_dim_matrix.Range(0, expected_dim_matrix.NumRows(),
+                                                                0, expected_dim_matrix.NumCols()));
+    }
+  }
+  // Now we operate the "expected_dim_matrix"  
+  perturbed_chunk->Resize(2, expected_dim_matrix.NumCols());
+  MatrixIndexT indices[2] = {0, 1};
+  perturbed_chunk->CopyRows(expected_dim_matrix, indices);
+}
+
+void FvectorPerturb::VolumePerturbation(MatrixBase<BaseFloat>* chunk) {
+  //1. Randomly generate 2 number from (1-max-volume-variance, 1+max-volume-variance)
+  std::vector<BaseFloat> volume_factors;
+  for (MatrixIndexT i = 0; i < chunk->NumRows(); ++i) {
+    BaseFloat factor = static_cast<BaseFloat>(
+        RandInt((int)((1-opts_.max_volume_variance)*100),
+                (int)((1+opts_.max_volume_variance)*100)) / 100.0);
+    volume_factors.push_back(factor);
+  }
+  //2. scale each line respectively.
+  for (MatrixIndexT i = 0; i < chunk->NumRows(); ++i) {
+    chunk->Row(i).Scale(volume_factors[i]);
+  }
+}
+
+// we stretch the signal from the beginning to end.
+// y(t) = x(s*t) for t = 0,...,n. If s>0, the output will be shorter than
+// input. It represents speeding up. Vice versa.
+// Use ArbitraryResample deal with each line.
+//
+// In ArbitraryResample, according to num_zeros and filter_cutoff, it generates
+// the "filter_with". And then each output_sample(t) corresponds to few input_samples
+// from (t-filter_with) to (t+filter_with), which is stored in "first_index_".
+// And "weights_" will be adjust by a Hanning window in function FilterFunc.
+// In brief, you can think each output sample is the weighted sum of few input_samples.
+void FvectorPerturb::SpeedPerturbation(VectorBase<BaseFloat>& input_vector,
+                                       BaseFloat samp_freq,
+                                       BaseFloat speed_factor,
+                                       VectorBase<BaseFloat>* output_vector) {
+  if (speed_factor == 1.0) {
+    output_vector->CopyFromVec(input_vector);
+  } else {
+    Vector<BaseFloat> in_vec(input_vector),
+                      out_vec(output_vector->Dim());
+    int32 input_dim = in_vec.Dim(),
+          output_dim = out_vec.Dim();
+    Vector<BaseFloat> samp_points_secs(output_dim);
+    int32 num_zeros = 4; // Number of zeros of the sinc function that the window extends out to.
+    // lowpass frequency that's lower than 95% of the Nyquist.
+    BaseFloat filter_cutoff_hz = samp_freq * 0.475; 
+    for (int32 i = 0; i < output_dim; i++) {
+      samp_points_secs(i) = static_cast<BaseFloat>(speed_factor * i / samp_freq);
+    }
+    ArbitraryResample time_resample(input_dim, samp_freq,
+                                    filter_cutoff_hz, 
+                                    samp_points_secs,
+                                    num_zeros);
+    time_resample.Resample(in_vec, &out_vec);
+    output_vector->CopyFromVec(out_vec);
+  }
+}
+
+void FvectorPerturb::TimeShift(VectorBase<BaseFloat>& input_vector,
+                               VectorBase<BaseFloat>* output_vector) {
+  //1. generate start point randomly whose range is
+  // [0, row.NumCols()- expected_chunk_length * sample_frequency)
+  int32 start_point = static_cast<int32>(RandInt(0, input_vector.Dim() - output_vector->Dim()));
+  //2. get the successive expected_chunk_length * sample_frequency data.
+  output_vector->CopyFromVec(input_vector.Range(start_point, output_vector->Dim()));
+}
+
+void FvectorPerturb::AddNoise(BaseFloat probability_threshold, 
+                              MatrixBase<BaseFloat>* chunk) {
+  //1. generate 2 SNR from (min-snr, max-snr)
+  //2. add N1(line3) to S1(line1) with snr1 with probability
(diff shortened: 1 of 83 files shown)


src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,95 @@
...
+                   "affect the speed_perturb_rate, the speed_perturb_rate factor will "
+                   "in the range of min{expected-chunk-length/original-length, "
+                   "max-speed-perturb-rate}. default=100 ms.");
+    opts->Register("max-speed-perturb-rate", &max_speed_perturb_rate,
+                   "Max speed perturbation applied on matrix. It will work together "
+                   "with expected_chunk_length, default=0.1.");
+    opts->Register("max-volume-variance", &max_volume_variance, "The variation in "
+                   "volume will vary form -max-volume-variance to max-volume-variance randomly."
+                   "default=0.03.")
+    opts->Register("max-snr",&max_snr,"Specify a upperbound Signal to Noise Ration. We will scale the noise according "
```
pegahgh: ration -> ratio

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,95 @@
...
+                   "to the original signal and SNR. Normally, it's a non-zero number between -30 and 30"
+                   "default=-5.");
+    opts->Register("min-snr",&min_snr,"Specify a lowerbound Signal to Noise Ration. We will scale the noise according "
+                   "to the original signal and SNR. Normally, it's a non-zero number between -30 and 30"
+                   "default=10");
+    opts->Register("do-volume-variation", &do_volume_variation, "If ture, we will "
+                   "conduct variations in volume.");
+    opts->Register("do-speed-perturbation", &do_speed_perturbation, "If ture, we will "
+                   "conduct variations in speed.");
+    opts->Register("do-time-shift", &do_time_shift, "If ture, we will "
```
pegahgh: remove do from all names. e.g. do-time-shift -> time-shift

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,95 @@
...
+                   "conduct reverberation operation.");
+  }
+};
+
+class FvectorPerturb {
+ public:
+  PerturbXvector(FvectorPerturbOptions opts) { opts_ = opts; }
+  void ApplyPerturbation(const MatrixBase<BaseFloat> &input_chunk,
+                         Matrix<BaseFloat> *perturbed_chunk);
+  void DoVolumeVariation(MatrixBase<BaseFloat> &chunk);
```
pegahgh: Also Remove Do from function names.

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,95 @@
...
+                   "get the successive 'expected_chunk_length' data. Otherwise, we get "
+                   "the data from the head.");
+    opts->Register("do-add-noise", &do_add_noise, "If ture, we will "
+                   "conduct add additive noise to source chunk.");
+    opts->Register("do-reverberation", &do_reverberation, "If ture, we will "
+                   "conduct reverberation operation.");
+  }
+};
+
+class FvectorPerturb {
```
pegahgh: Add Description for this class and its functions.

src/fvector/fvector-perturb.cc
```diff
@@ -0,0 +1,54 @@
...
+
+void FvectorPerturb::DoSpeedPerturbation(MatrixBase<BaseFloat>& chunk) {
+  //1. generate 4 speed perturb factor randomly
+  //(1) a=min{expected_chunk_length/original_length, max-speed-perturb-rate}
+  //(2) the range of factor is (1-a, 1+a)
+  //2. Use ArbitraryResample deal with each line
+  // Reference Pegah's Time-stretch
+}
+
+void FvectorPerturb::DoTimeShift(MatrixBase<BaseFloat>& chunk) {
```
pegahgh: it is better to pass pointer instead of reference to functions.

src/fvectorbin/fvector-add-noise.cc
```diff
@@ -0,0 +1,53 @@
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "fvector/fvector-perturb.h"
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    const char *usage =
+        "Perturb the chunk data.\n"
+        "Usage:  fvector-add-noise [options...] <chunk-rspecifier> <perturbed-wspecifier>\n";
```
pegahgh: describe function in more details e.g. what type of perturbation it covers

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,95 @@
...
+    opts->Register("do-speed-perturbation", &do_speed_perturbation, "If ture, we will "
+                   "conduct variations in speed.");
+    opts->Register("do-time-shift", &do_time_shift, "If ture, we will "
+                   "conduct time shift. That means randomly select the start point and "
+                   "get the successive 'expected_chunk_length' data. Otherwise, we get "
+                   "the data from the head.");
+    opts->Register("do-add-noise", &do_add_noise, "If ture, we will "
+                   "conduct add additive noise to source chunk.");
+    opts->Register("do-reverberation", &do_reverberation, "If ture, we will "
+                   "conduct reverberation operation.");
```
pegahgh: remove reverberation for now!

src/fvectorbin/fvector-add-noise.cc
```diff
@@ -0,0 +1,53 @@
...
+
+    int64 num_read = 0, num_written = 0;
+    for (; !chunk_reader.Done(); chunk_reader.Next(), num_read++) {
+      std::string key = chunk_reader.Key();
+      // input_chunk has 3 lines.
+      const Matrix<BaseFloat> &input_chunk = chunk_reader.Value();
+      // whole_chunk has 4 lines, it copies the first line and will be operate.
+      Matrix<BaseFloat> whole_chunk(4, input_chunk.NumCols());
+      MatrixIndexT indices[4] = {0, 0, 1, 2};
+      whole_chunk.CopyRows(input_chunk, indices);
```
pegahgh: I am not sure, what you are doing here. Would you describe it in more details?

src/fvectorbin/fvector-get-egs.cc
```diff
@@ -0,0 +1,90 @@
...
+    SequentialBaseFloatMatrixReader feature_reader(feature_rspecifier);
+
+    int32 num_read = 0,
+          num_egs_written = 0;
+    for (; feature_reader.Done(); feature_reader.Next(), num_read++) {
+      std::string key = feature_reader.Key();
+      const GeneralMatrix &feats = feature_reader.Value();
+      SubMatrix<BaseFloat> chunk1(feats, 0, 1, 0, feats.NumCols()),
+                           chunk2(feats, 1, 1, 0, feats.NumCols());
+      NnetIo nnet_io1 = NnetIo("input", 0, chunk1),
```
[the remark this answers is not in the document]
LvHang: In this binary, each GeneralMatrix(feats) has two rows. They are expected length(e.g 100ms), and come from the same source 120ms. But they were added different kinds of perturbation(e.g different volume variation, different speed perturbation, etc.) At the same time, the first row was added "noise1" chunk and the second row was added "noise2" chunk.

src/fvectorbin/fvector-add-noise.cc
LvHang: In "input_chunk" matrix, it contains 3 lines(S1, N1, N2). Here, we copy the S1. So, in "whole_chunk" matrix, it contains 4 lines(S1, S2, N1, N2). When we do "add-noise", we will add additive noise(N1->S1 and N2->S2). So after "add-noise" operation, we has a matrix which contains two rows. And the two rows(100ms) come from the same source(120ms).

pegahgh: What is the number of rows in  your input chunk? is this 3? If it is 3, I don't think it is necessary to have this implementation and you can have noise_mat and orig_mat and the output is perturbed version of each row in orig_mat with rows in noise_mat. 

Also you mean 'adding another speech chunk' by noise? 
Please describe it with detail in code!

src/fvectorbin/fvector-get-egs.cc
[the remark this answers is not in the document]
pegahgh: at least, in this code, chunk1 and chunk2 are the same. You need to pass different feats to generate chunk2

pegahgh:
@LvHang You mentioned you prepared scripts, but I didn't see any updates. Would you commit your new changes e.g fixing whatever I asked you in the comment?

LvHang:
The binaries are added. Now the FvectorPerturbation class is implemented.
You can read in the order: fvector-chunk.cc, fvector-add-noise.cc and fvector-get-egs.cc.
I'm testing and try to make sure there is no bugs. The bash scripts will update soon.

LvHang:
Now, All the codes can be compiled well.
At the same time, I add the fvector-perturb-test.cc file to compare our speed perturbation function, which use ArbitraryResample Class, with sox.
In fvector/test_data directory, there are two wavs. The "test.wav" is a test wav. The "test_speed12.wav" is the speed pertrubed wav by sox (with speed factor = 1.2)
You can download and compile the code directly. Then you can see the two results. 
When we use ApproxEqual to compare the distance, it will be fail. In numerical value, there is a gap.(Sorry, I didn't have enough experience to say it's big or not.) At the same time, I compute the cosin distance is 0.999444. (If you can help to check the code, I will be appreciate.)
Besides, I think the two results is hard to be excatly the same. Because ArbitraryResample has few super-parameters, such as num_zeros, filter_cutoff and so on. In addition, in ArbitraryResample, we use hanning window, but I don't know what sox uses.
Best,
Hang

LvHang:
@pegahgh , when you are free, could you help to check the C++ code ,bash scripts and the above question. Thanks a lot.
I think the PR can generate the fvector examples if the four kinds of perturb operations are right. I think there is no bugs in C++ code, but I'm afraid there is any interpretive mistake in speed perturbation part. I hope not.
If the above codes are all appropriate. Maybe we can move on to construct the topology of network and train it.
Best,
Hang

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,116 @@
...
+                   "conduct time shift. That means randomly select the start point and "
+                   "get the successive 'expected_chunk_length' data. Otherwise, we get "
+                   "the data from the head.");
+    opts->Register("add-noise", &add_noise, "If ture, we will "
+                   "conduct add additive noise to source chunk.");
+  }
+};
+
+/* This class is used to do 4 kinds of perturbation operation to fvector.
+ * The input always is a Matrix which contains four lines(S1, S2, N1, N2)[S1=S2]
```
pegahgh: I am not still sure why you selected fixed number of perturbation as 4?

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,116 @@
...
+ public:
+  FvectorPerturb(FvectorPerturbOptions opts) { opts_ = opts; }
+  void ApplyPerturbation(const MatrixBase<BaseFloat>& input_chunk,
+                         Matrix<BaseFloat>* perturbed_chunk);
+
+  // Randomly Generate 4 scale number and scale each line respectively
+  void VolumePerturbation(MatrixBase<BaseFloat>* chunk);
+
+  // Use ArbitraryResample. For each line, randomly generate a speed factor. 
+  // Then do time axis strench. As speed factor is different, so we deal with
```
pegahgh: strench -> stretch

src/fvector/fvector-perturb.h
```diff
@@ -0,0 +1,116 @@
...
+  // each vector sepeartely. The dim of output_vector is bigger than
```
pegahgh: sepeartely -> separately

src/fvectorbin/fvector-add-noise.cc
```diff
@@ -0,0 +1,58 @@
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "fvector/fvector-perturb.h"
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    const char *usage =
+        "Perturb the chunk data. Each input chunk is a four-lines matrix(S1, S2, N1, N2).\n"
+        "After use 4 kinds of perturbation opertation, each output chunk is a 2-lines matrix.\n"
```
pegahgh: you need to rewrite usage!

pegahgh: 2-line matrix -> 2 consecutive rows of output matrix

src/fvector/fvector-perturb.cc
```diff
@@ -0,0 +1,148 @@
...
+    for (int32 i = 0; i < output_dim; i++) {
+      samp_points_secs(i) = static_cast<BaseFloat>(speed_factor * i / samp_freq);
+    }
+    ArbitraryResample time_resample(input_dim, samp_freq,
+                                    filter_cutoff_hz, 
+                                    samp_points_secs,
+                                    num_zeros);
+    time_resample.Resample(in_vec, &out_vec);
+    output_vector->CopyFromVec(out_vec);
+  }
```
pegahgh: Did you check it with sox speed perturbation? would you send me some figure comparing kaldi versus sox?

src/fvector/fvector-perturb.cc
```diff
@@ -0,0 +1,148 @@
...
+void FvectorPerturb::TimeShift(VectorBase<BaseFloat>& input_vector,
+                               VectorBase<BaseFloat>* output_vector) {
+  //1. generate start point randomly whose range is
+  // [0, row.NumCols()- expected_chunk_length * sample_frequency)
+  int32 start_point = static_cast<int32>(RandInt(0, input_vector.Dim() - output_vector->Dim()));
+  //2. get the successive expected_chunk_length * sample_frequency data.
+  output_vector->CopyFromVec(input_vector.Range(start_point, output_vector->Dim()));
+}
+
+void FvectorPerturb::AddNoise(MatrixBase<BaseFloat>* chunk) {
```
pegahgh: These type of functions are very specific for your design. 
It is better to have generic function with noise and chunk as input to function!

LvHang:
@pegahgh 
According to our discussion, I do some modification. And I upload some result files for your convenience.
(1) According to your comments, I fixed some problems.
(2) I rearrange the order of 4 kinds of perturbation operations. Now we add additive noise with probability firstly. Then we operate the 2 consecutive rows matrix. Do volume perturbation, speed perturbation and time-shift.
(3) I modify the fvector-get-egs. Now, for each chunk(100ms in one row), we cut it into few piece, which depend on "frame_length_ms" and "frame_shift_ms", so that it is similiar with general features(e.g. mfcc).
(4) I modify the  "fvector-perturb-test.cc" file. Now it can generate wav file. In fvector/test_data directory, the "test.wav" is original wav, the "test_speed12.wav" is wav which is processed by sox , and the "test_speedperturbed.wav" is generated by my SpeedPerturbation() function.
Then I extract the 39 dims mfcc features for the 2 perturbed wav files. I have no idea about how to plot it elegantly, so I just store the mfcc features separately. Please see "test_speed12.mfcc.txt" and "test_speedperturbated.mfcc.txt".
At the same time, I use "compare_feats" to compare the two files. The value is 0.975012. So I think my SpeedPerturbation() is right.
When you are free, you can check it.

Now I'm writing the topology of network. Could you show me the setup you used in paper "Acoustic modeling from the signal domain using CNN". Thanks a lot.

Best,
Hang

LvHang:
Source Code: pegah time/frequency domain cnn + david plda xvector + fvector add noise + 18-02-12 kaldi/master
Scripts: mini_librispeech scripts

LvHang:
The whole pipeline of kaldi

LvHang:
The following information is mainly to describe the current situation of fvector and the framework of the code.
1. The current experiments are conducting on mini_librispeech dataset.

The baseline(mfcc chain model) result is WER=9.66%(tglarge_dev_clean_2)
The fvector chain model result is WER=15.10%(tglarge_dev_clean_2) 

[This result is much better than dump Nnet3Egs which have already add noise into disk(WER=21.82%, log-likelihood is about -1.6). It dumped the raw data vector into disk. And then we add the noise on the fly.]

[But the fvector model is not well trained since the training data is not enough. The current final log-likelihood is -0.63. The volume of dataset is 8epoches*13hrs. The volume of training data is about 2000hrs when David trained a same-different model.]

Otherwise, I will test the four kinds of add noise ways separately.

2. About compile code.
You can download the newest kaldi-master. And "git cherry-pick" the lastest two commits of this PR. One is about doing frequency domain training. The other is about fvector code.

3. The framework summary.
(1)In xvector/xvectorbin directories, it provides the same-different xvector training code. Maybe it looks old, I mean its style is a little different current kaldi-master. I will update it later. Anyway, it works now.
(2)The frequency domain raw waveform code is mainly in xconfig and nnet3.
(3)The fvector/fvectorbin directories contain other codes which is needed in this project.
~.The code with "_block" suffix means it deal with a block data together, maybe 64 source chunks. Up to you. But it restricts flexibility. For example, when we do volume perturbation in this code, the 64 chunks will be scaled with the same factor.
~.The code with "_separate" suffix is the newest. In this version, We dump the KaldiVector into disk rather than Nnet3Eg. And on each training iteration, we add the noise on the fly. I make the added noise is different each time rather than a fixed noise. So it will prevent the network to learn a specific noise so that it sink into a specific case.
In this setup, we will store the source chunks and noise chunks separately into archives（which I called eg.ark and eg.noise.ark. The latter one is twice that of the former.）
~. The other codes is public or "Nnet3Eg" version.
~. The Nnet3Eg version recipe is "s5/run_fvector.sh" and the "separate" version recipe is "s5/run_fvector_separate". You can go along with it.

PR #1999, opened 2017-11-05
add RNNLM recipes for 5 datasets

keli78:
add RNNLM recipes for swbd, ami, tedlium, wsj and hub4_spanish; 
set test modes of dropout and batchnorm True;
some other fix about getting word features

### DIFF
--- a/egs/ami/s5b/local/rnnlm/tuning/run_lstm_tdnn_a.sh
+++ b/egs/ami/s5b/local/rnnlm/tuning/run_lstm_tdnn_a.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# Copyright 2012  Johns Hopkins University (author: Daniel Povey)  Tony Robinson
+#           2017  Hainan Xu
+#           2017  Ke Li
+
+# 2-layer LSTM hidden 200:
+# rnnlm/train_rnnlm.sh: best iteration (out of 18) was 10, linking it to final iteration.
+# Train objf: -620.20 -4.49 -4.33 -4.21 -4.13 -4.06 -4.00 -3.96 -3.92 -3.88 -3.84 -3.81 -3.77 -3.74 -3.71 -3.68 -3.65 -3.59
+# Dev objf:   -10.76 -4.74 -4.54 -4.44 -4.38 -4.33 -4.31 -4.29 -4.28 -4.28 -4.28 -4.29 -4.29 -4.30 -4.32 -4.33 -4.34 -4.36
+
+# 2-layer LSTM + 1 TDNN layer hidden 200:
+# rnnlm/train_rnnlm.sh: best iteration (out of 18) was 7, linking it to final iteration.  
+# Train objf: -664.80 -4.39 -4.21 -4.09 -4.01 -3.95 -3.89 -3.85 -3.81 -3.77 -3.74 -3.71 -3.68 -3.65 -3.63 -3.61 -3.59 -3.57
+# Dev objf:   -10.76 -4.65 -4.45 -4.34 -4.30 -4.29 -4.28 -4.27 -4.27 -4.28 -4.28 -4.29 -4.29 -4.30 -4.31 -4.32 -4.33 -4.34
+
+# Begin configuration section.
+cmd=run.pl
+dir=exp/rnnlm_lstm_tdnn_a
+embedding_dim=200
+epochs=60
+mic=sdm1
+stage=-10
+train_stage=0
+
+. utils/parse_options.sh
+train=data/$mic/train/text
+dev=data/$mic/dev/text
+wordlist=data/lang/words.txt
+text_dir=data/rnnlm/text
+mkdir -p $dir/config
+set -e
+
+for f in $train $dev $wordlist; do
+  [ ! -f $f ] && \
+    echo "$0: expected file $f to exist; search for run.sh and utils/prepare_lang.sh in run.sh" && exit 1
+done
+
+if [ $stage -le 0 ]; then
+  mkdir -p $text_dir
+  cat $train | cut -d ' ' -f2- > $text_dir/ami.txt
+  cat $dev | cut -d ' ' -f2- > $text_dir/dev.txt
+fi
+
+if [ $stage -le 1 ]; then
+  cp $wordlist $dir/config/
+  n=`cat $dir/config/words.txt | wc -l`
+  echo "<brk> $n" >> $dir/config/words.txt
+
+  # words that are not present in words.txt but are in the training or dev data, will be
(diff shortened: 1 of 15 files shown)


scripts/rnnlm/train_rnnlm.sh
```diff
@@ -175,14 +177,19 @@ while [ $x -lt $num_iters ]; do
...
 
         # Run the training job or jobs.
         $cmd $queue_gpu_opt $dir/log/train.$x.$n.log \
            rnnlm-train \
              --rnnlm.max-param-change=$rnnlm_max_change \
+             --rnnlm.l2_regularize_factor=$rnnlm_l2_factor \
              --embedding.max-param-change=$embedding_max_change \
              --embedding.learning-rate=$embedding_lrate \
+             --embedding.l2_regularize=$embedding_l2 \
+             --embedding.l2_regularize_factor=$embedding_l2_factor \
```
danpovey: I don't think it's necessary to have two separate options here-- just have a single l2-regularize option and compute it at the script level.  Specifying the factor was only necessary for the model-level one because the component-level l2 values can be different.

keli78: Ok, will remove the embedding l2 factor.

egs/ami/s5b/local/rnnlm/tuning/run_lstm_tdnn_a.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
+# Copyright 2012  Johns Hopkins University (author: Daniel Povey)  Tony Robinson
+#           2017  Hainan Xu
+#           2017  Ke Li
+
+# 2-layer LSTM hidden 200:
+# rnnlm/train_rnnlm.sh: best iteration (out of 18) was 10, linking it to final iteration.
```
danpovey: Please modify train_rnnlm.sh so that after this line, it prints out something like:
```
  train/dev perplexity was -83.7/-128.4.
```
These should be extracted from the compute_prob_{train,valid} logs of the 'best' iteration, the one that we linked to the final iteration,  and exponentiated.  If you have to write a perl or python script to do this, then do it (although a simple inline awk script would probably suffice).

danpovey:
Merging this as-is, though.

francoishernandez:
Hi @keli78 ,
I'm looking at the recipe for tedlium.
Just for the sake of clarity, it seems like you are building your RNNLM dev and training sets from `data/train/text`. Shouldn't we take some more consequent data, such as the data provided in the TEDLIUM_r2/LM directory and used in 'normal' arpa rescoring?
Anyways thanks for the scripts!
F

keli78:
Hi @francoishernandez, sorry for this late reply. You can use the data in the LM directory definitely.

francoishernandez:
Hmm. I tried running my version with the whole LM data and your version with only the 'text from audio' data and in both cases it fails.
It seems to begin correctly, doing the first iterations, but 'diagnostics' fail. Looks like there are some issues in the model. @danpovey any idea? We really would like to run experiments on this.

Here are the 'general' logs ran on @keli78 's script:
```
rnnlm dir done
rnnlm/train_rnnlm.sh: will train for 10 iterations
Training neural net (pass 0)
Training neural net (pass 1)
Training neural net (pass 2)
Training neural net (pass 3)
Training neural net (pass 4)
Training neural net (pass 5)
run.pl: job failed, log is in exp/rnnlm_lstm_tdnn_a/log/compute_prob.5.log
rnnlm/train_rnnlm.sh: error with diagnostics on iteration 5 of training.
```
And the contents of `compute_prop.5.log`
```
# rnnlm-get-egs --bos-symbol=152215 --eos-symbol=152216 --brk-symbol=152217 --vocab-size=152218 exp/rnnlm_lstm_tdnn_a/text/dev.txt ark:- | rnnlm-compute-prob exp/rnnlm_lstm_tdnn_a/5.raw "rnnlm-get-word-embedding exp/rnnlm_lstm_tdnn_a/word_feats.txt exp/rnnlm_lstm_tdnn_a/feat_embedding.5.mat -|" ark:- 
# Started at Thu Nov 23 18:30:02 CET 2017
#
rnnlm-compute-prob exp/rnnlm_lstm_tdnn_a/5.raw 'rnnlm-get-word-embedding exp/rnnlm_lstm_tdnn_a/word_feats.txt exp/rnnlm_lstm_tdnn_a/feat_embedding.5.mat -|' ark:- 
LOG (rnnlm-compute-prob[5.2.287~1-3348b]:SelectGpuId():cu-device.cc:110) Manually selected to compute on CPU.
rnnlm-get-egs --bos-symbol=152215 --eos-symbol=152216 --brk-symbol=152217 --vocab-size=152218 exp/rnnlm_lstm_tdnn_a/text/dev.txt ark:- 
rnnlm-get-word-embedding exp/rnnlm_lstm_tdnn_a/word_feats.txt exp/rnnlm_lstm_tdnn_a/feat_embedding.5.mat - 
LOG (rnnlm-get-egs[5.2.287~1-3348b]:Process():rnnlm-example.cc:703) Processed 10000 lines of input.
LOG (rnnlm-compute-prob[5.2.287~1-3348b]:PrintStatsOverall():rnnlm-core-training.cc:119) Overall objf is (-nan + -nan) = -nan over 0 words (weighted) in 0 minibatches; exact = (-nan + -nan) = -nan
LOG (rnnlm-compute-prob[5.2.287~1-3348b]:~CachingOptimizingCompiler():nnet-optimize.cc:714) 0.283 seconds taken in nnet3 compilation total (breakdown: 0.114 compilation, 0.136 optimization, 0.00346 shortcut expansion, 0.0266 checking, 6.2e-06 computing indexes, 0.00294 misc.)
std::bad_alloc
# Accounting: time=115 threads=1
# Ended (code 255) at Thu Nov 23 18:31:57 CET 2017, elapsed time 115 seconds
```
By the way, when this fails, some rnnlm-compute-prob instances keep running on CPU.

danpovey:
It looks to me like what's happening is that you're running this on a
single machine without much memory, and the rnnlm-compute-prob jobs take a
while since they don't use GPU, so they accumulate in the background and
eventually exhaust your memory.

francoishernandez:
Well this was launched on a machine with 32Gb of memory and 2 GPUs, but using only one of them.
Would running the rnnlm-compute-prob jobs on the second GPU be a solution? Or shall we make the script wait for those to finish before going on?

danpovey:
Yes you could do that.  It would probably naturally finish fast enough that
contention for GPUs wouldn't normally be a problem.  Or you could change
the python script to wait (I think there is a different function call for
background vs. foreground jobs).

PR #2005, opened 2017-11-09
Update RNNLM recipes with L2 regularization

keli78:


egs/ami/s5b/local/rnnlm/tuning/run_lstm_tdnn_b.sh
```diff
@@ -0,0 +1,99 @@
...
+
+  # choose features
+  rnnlm/choose_features.py --unigram-probs=$dir/config/unigram_probs.txt \
+                           --use-constant-feature=true \
+                           --top-word-features 10000 \
+                           --min-frequency 1.0e-03 \
+                           --special-words='<s>,</s>,<brk>,<unk>,[noise],[laughter]' \
+                           $dir/config/words.txt > $dir/config/features.txt
+
+lstm_opts="l2-regularize=$comp_l2"
```
danpovey: did you ever try using the smaller amount of L2 ($output_l2) for the lstm layers?  I seem to remember that was better in an ASR setup where I tried it.

keli78: For AMI no, but I tried smaller output regularization (0.001) and it gives a little worse perplexity (70.1 vs. 68.7) than using 0.005. So I just use 0.005. 
For other datasets, I tuned the L2 value.

egs/swbd/s5c/local/rnnlm/tuning/run_lstm_tdnn_b.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# Begin configuration section.
+cmd=run.pl
+dir=exp/rnnlm_lstm_tdnn_b
+embedding_dim=1024
+lstm_rpd=256
+lstm_nrpd=256
+embedding_l2=0.001 # embedding layer l2 regularize
+comp_l2=0.001 # component-level l2 regularize
+output_l2=0.001 # output-layer l2 regularize
```
pegahgh: I know that in ASR, Dan used smaller l2_reg for output layers to allow larger weights for output layer, but in all of these scripts, output_l2 and comp_l2 are the same values.
Is there any specific reason to have two variables?
Did you get degradation by reducing output_l2?

keli78: Oh because at the beginning I have thought they may be different so I kept using two variables. I tried smaller output_l2 (than comp_l2) on AMI and the performance degrades. For other datasets, I just tried to match the norms of parameters with those of the best results that I used proportional shrink. So I didn't separately tune the output_l2 much.  I think I can have a try on one or two datasets.

PR #2007, opened 2017-11-09
fix some tiny details

jtrmal:
@vesis84 please review removal of the src/nnet/nnet-linear-transform.h.orig
Seems like something that got in by accident, but please confirm


danpovey:
I think it's just a mistake, I'll merge.

PR #2008, opened 2017-11-10
replace tabs by spaces and reformat if necessary

jtrmal:
I'm almost ashamed of this PR -- doesn't do anything, only replaces tabs with whitespaces and reformats (so it's not plain sed work, I actually tried to reformat things correctly).
Files I didn't do:
```
egs/wsj/s5/utils/prepare_lang.sh
egs/fame/s5/local/fame_dict_prep.sh
egs/fame/v1/local/fame_dict_prep.sh
egs/wsj/s5/steps/cleanup/lattice_oracle_align.sh
egs/wsj/s5/steps/cleanup/find_bad_utts_nnet.sh
egs/wsj/s5/steps/cleanup/find_bad_utts.sh
```
which use tabs for other reasons, not for formatting

and I skipped these recipes
```
egs/gp/
egs/csj
egs/chime3
```
which have HUGE number of tabs everywhere (I might fix it later or never)

I checked this changeset using 
```
git diff -b  ..upstream/master
```
which returned empty diff, i..e. the changes are really only whitespaces


jtrmal:
BTW, I didn't do perl and python scrpts

danpovey:
Merging-- thanks.

PR #2012, opened 2017-11-13
Implementation of TDNN for chime4

Szu-JuiChen:
Adding TDNN option for chime4


egs/chime4/s5_1ch/local/chain/tuning/run_tdnn_1d.sh
```diff
@@ -0,0 +1,346 @@
+#!/bin/bash
+
+# 1d is as 1c but introducing two non-splicing layers towards the beginning of
+#   the network.
+
+# local/chain/compare_wer.sh exp/chain/tdnn1c_sp exp/chain/tdnn1e3_sp
```
danpovey: You seem to have copied these comments from WSJ.  They report the results and diagnostics from WSJ.  You should replace these with suitable results and diagnostics from chime4.
In fact, you seem to have just copied the entire tuning directory from WSJ without really looking into its contents.  The idea is that when you commit the inital recipe for a setup you'll start with 1a.  Don't copy scripts that you haven't tested (e.g. the LSTM scripts).

Szu-JuiChen:
Got it. I'll update this part. Thanks!

egs/chime4/s5_1ch/local/chain/tuning/run_tdnn_1a.sh
```diff
@@ -0,0 +1,349 @@
...
+  input dim=100 name=ivector
+  input dim=40 name=input
+
+  # please note that it is important to have input layer with the name=input
+  # as the layer immediately preceding the fixed-affine-layer to enable
+  # the use of short notation for the descriptor
+  fixed-affine-layer name=lda input=Append(-2,-1,0,1,2,ReplaceIndex(ivector, t, 0)) affine-transform-file=$dir/configs/lda.mat
+
+  # the first splicing is moved before the lda layer, so no splicing here
+  relu-batchnorm-layer name=tdnn1 dim=600
```
danpovey: You might want to try the same thing with dim=700 or 750: I think this could be overfitting.
Unless you want to get this checked in quick because of some kind of deadline.

Szu-JuiChen: Sure. But I tried dim=700 before and got an error message about running out of memory. I will try it with larger memory requisition. Thanks.

danpovey:
No, if that happens you have to reduce the num-chunk-per-minibatch.  E.g.
change it from 256,128,64 to 128,64

sw005320:
@Szu-JuiChen I have run you codes, and found some issues. Can you change as follows in local/run_beamform_2ch_track.sh and local/run_beamform_6ch_track.sh?
`cd ../../../tools/; make beamformit`
->
`cd ../../../tools/; extras/install_beamformit.sh`

Szu-JuiChen:
@sw005320 Sure. I'll modify it.
@danpovey  I have tried some experiments these days and the system is still not converge.  
image

The initial setting in tdnn1a is dimension=750, epoch=4, initial-lrate=0.003, and final-lrate=0.0003
In tdnn1b, I increase the epoch to 6
In tdnn1c, I change the final-lrate to 0.0001
In tdnn1d, I increase the epoch to 8

I would like to ask if you have any suggestions for parameters?

By the way, tdnn1b can get the best WER 4.12% after rnnlm rescore. On the other hand, my previous best WER is 4.13% with dimension=600, epoch=4, initial-lrate=0.0025, and final-lrate=0.00025

danpovey:
It's strange that your train and valid probs end up so similar.
Is it the case that you augment your training data in such a way that the
original files get duplicated to different names?  In that case the default
train/valid split would not be correct, and it would be necessary to add to
the input data-dir an "utt2uniq" file that maps the utterance id to the
original utterance id before duplication.

However, this would mainly affect the diagnostics, not the actual results.

I think the main issue with your script as it was, was that you were not
doing the speed-perturbing.

sw005320:
@danpovey About the speed-perturbation, I tested it in this setup, and did not have some improvements. So, I think we don't have to include it. Note that we already augment the training data by using all 6 channel audios as training data (6 times more than the original, which is widely used in the CHiME4 challenge top systems).

danpovey:
OK.  IMO it would have been a good idea when augmenting with the channel
audios, to do some speed perturbation, that would probably improve the
results.  But I guess that's a separate isue.
I suspect what was happening is that your systems were overfitting (and
maybe you had too many epochs) but you were not seeing that in the
diagnostics because of not having the 'utt2orig' file (so the validation
data was just different channels of utterances that were already seen in
training).

danpovey:
Also, I don't know what you do with the speaker ids when you multiply the channels, but when you do the data perturbing you should make sure that the channel-id is part of the speaker-id.   If you don't do this, the i-vector adaptation won't work correctly as it will mix up the channels.

sw005320:
We use the same speaker-id across channels. I agree that it's worth to consider to make the i-vector adaptation work channel-wise. @Szu-JuiChen can you also perform such experiments when you make time?

Szu-JuiChen:
@sw005320  No problem. I will complete the i-vector adaptation portion.
@danpovey  Thanks a lot for your opinions.

Szu-JuiChen:
@danpovey  Hi Prof. Dan, I checked the files train_subset_uttlist and valid_uttlist. You are right that some validation data was just different channels of utterances that were already seen in training. As a result, I am finding methods to generate utt2orig in Kaldi repository. However, only three files mentioned it and I still don't know how to produce it. Or I can just copy get_egs.sh and modify the "Get list of validation utterances" part in it? Thanks.

danpovey:
utt2orig is just a text file where the first field is the utterance-id and
the second field is the utterance-id that you had before the data
perturbation (or addition of noise, whatever).
e.g.

my_utt101_modified_somehow my_utt101
my_utt101_modified_another_way my_utt101
my_utt102_modified_somehow my_utt102

danpovey:
You need to call utils/fix_data_dir.sh to fix the sorting issue.

danpovey:
@Szu-JuiChen, what is happening with this?  It seems to hvae gone quiet and I think there were outstanding issues.

Szu-JuiChen:
@danpovey  I already add channel id part of speaker id. The utt2spk now looks like this:
011_CH1_011C0201_PED.CH1_SIMU 011_CH1
011_CH1_011C0202_PED.CH1_SIMU 011_CH1
011_CH1_011C0203_BUS.CH1_SIMU 011_CH1
......

However, the result of run_gmm.sh is not good:
gmm_7data_channel_wise

Then I found that in the training monophone model step, lots of utterance do not have alignment. I just found this issue last Thursday and I am still working on it.

By the way, I also modified run_tdnn_lstm.sh for Chime4 but have not tunned yet. Because Shinji hopes that I can complete adding Hainan's lmrescore into Chime4 first. I can upload those scripts if needed.

danpovey:
Regarding the run_gmm.sh results, I don't know how bad that is because I don't know what the baseline results were (or how many channels that was run for).
It can be normal to have quite a few alignments failed at the monophone stage, it's not necessarily a problem.
In future when you paste results, please paste as text.  You can enclose in triple backticks to make sure it's formatted in fixed width font.

Szu-JuiChen:
@danpovey Hi Prof. Dan,
The original WER is 13.66% as the shown below:
```
-------------------
best overall dt05 WER 13.55% (language model weight = 11)
-------------------
dt05_simu WER: 14.01% (Average), 12.24% (BUS), 17.05% (CAFE), 11.61% (PEDESTRIAN), 15.15% (STREET)
-------------------
dt05_real WER: 13.10% (Average), 15.68% (BUS), 13.35% (CAFE), 10.55% (PEDESTRIAN), 12.82% (STREET)
-------------------
et05_simu WER: 20.43% (Average), 15.13% (BUS), 21.68% (CAFE), 22.26% (PEDESTRIAN), 22.66% (STREET)
-------------------
et05_real WER: 21.34% (Average), 28.32% (BUS), 21.46% (CAFE), 18.95% (PEDESTRIAN), 16.64% (STREET)
-------------------
```
and the overall WER with channel id is 21.49% as shown in the photo above.

The error message looks like this:
```
exp/mono0a_tr05_multi_noisy/log/align.20.9.log:WARNING (gmm-align-compiled[5.2.240~1-994ca]:AlignUtteranceWrapper():decoder-wrappers.cc:477) Did not successfully decode file 020_CH1_020C020I_PED.CH1_SIMU, len = 408
```
and there are around 15-20% of the utterances got this error messages. 
It happened in train_mono.sh (line 117) -> gmm-align-compiled (line 128) -> AlignUtteranceWrapper() in decoder-wrappers.cc (line 479)-> ReachedFinal() in faster-decoder.cc. 
After viewing the codes, I still not sure why this would happened.

sw005320:
@danpovey it seems that including the channel id in the speaker id significantly degraded the performance. Szu-Jui and I will investigate the reason, but since the original his LMF-TDNN already has significant improvement. So, can you merge his PR? Actually, we're also preparing to improve the beamforming and RNNLM rescoring parts in this CHiME-4 task (with these, this recipe will reach top-level performance compared with the challenge systems), and if this PR is delayed all the other activities are also delayed. But again, we'll definitely revisit this in near future.

sw005320:
Thanks Dan!!!

egs/chime4/s5_1ch/local/chain/run_tdnn_lstm_recog.sh
```diff
@@ -0,0 +1,223 @@
...
+This script is intended to be used with GPUs but you have not compiled Kaldi with CUDA
+If you want to use GPUs (and have them), go to src/, and configure and make on a machine
+where "nvcc" is installed.
+EOF
+fi
+
+if [ $# -ne 2 ]; then
+  printf "\nUSAGE: %s <enhancement method> <model dir>\n\n" `basename $0`
+  echo "First argument specifies a unique name for different enhancement method"
+  echo "Second argument specifies acoustic and language model directory"
```
danpovey: (for next time you update: always give an example command line in usage messages-- it tends to be the most useful part of the usage message).

danpovey:
OK, I merged it, but please look into this and if you can, share with me more detailed diagnostics, and if on our system, pathnames, because it shouldn't be getting worse. 
(Incidentally, those tables you have been showing, it would be nice if they could indicate what stage of the system you are displaying somehow.. up to you.)  At what stage of system building do the WERs start getting worse-- is it already worse in the GMM system?  Do the objf improvements from tree-building change? (Look for build_tree.log, IIRC... generally higher is better).  What percentage of monophone alignments failed in the baseline (you told me 20% with channel-id, but not the percentage for the baseline).   You can show me the output of steps/info/gmm_dir_info.pl for original and modified runs.

Szu-JuiChen:
@danpovey  Hi Dan, the alignment issue is fixed. Here is the report of the experiments:
The TDNN result is a little bit better than the system without channel-wise training as shown below:
```
The TDNN result without channel-wise training
-------------------
best overall dt05 WER 4.41% (language model weight = 10)
-------------------
dt05_simu WER: 4.48% (Average), 4.01% (BUS), 5.10% (CAFE), 4.25% (PEDESTRIAN), 4.54% (STREET)
-------------------
dt05_real WER: 4.35% (Average), 5.03% (BUS), 4.14% (CAFE), 3.89% (PEDESTRIAN), 4.32% (STREET)
-------------------
et05_simu WER: 5.61% (Average), 4.93% (BUS), 5.79% (CAFE), 5.70% (PEDESTRIAN), 6.03% (STREET)
-------------------
et05_real WER: 5.98% (Average), 6.51% (BUS), 5.83% (CAFE), 5.90% (PEDESTRIAN), 5.70% (STREET)
-------------------
```
```
The TDNN result with channel-wise training
-------------------
best overall dt05 WER 4.34% (language model weight = 10)
-------------------
dt05_simu WER: 4.32% (Average), 3.92% (BUS), 5.16% (CAFE), 3.82% (PEDESTRIAN), 4.37% (STREET)
-------------------
dt05_real WER: 4.37% (Average), 4.99% (BUS), 4.32% (CAFE), 4.00% (PEDESTRIAN), 4.16% (STREET)
-------------------
et05_simu WER: 5.57% (Average), 4.84% (BUS), 6.11% (CAFE), 5.49% (PEDESTRIAN), 5.85% (STREET)
-------------------
et05_real WER: 5.90% (Average), 6.62% (BUS), 5.57% (CAFE), 5.72% (PEDESTRIAN), 5.70% (STREET)
-------------------
```

However, the result after RNN LM rescoring is almost the same:
```
The RNNLM rescore result without channel-wise training
-------------------
best overall dt05 WER 2.88% (language model weight = 11)
-------------------
dt05_simu WER: 2.94% (Average), 2.83% (BUS), 3.67% (CAFE), 2.57% (PEDESTRIAN), 2.68% (STREET)
-------------------
dt05_real WER: 2.83% (Average), 3.42% (BUS), 2.55% (CAFE), 2.71% (PEDESTRIAN), 2.62% (STREET)
-------------------
et05_simu WER: 3.80% (Average), 3.25% (BUS), 3.98% (CAFE), 3.81% (PEDESTRIAN), 4.17% (STREET)
-------------------
et05_real WER: 4.01% (Average), 4.28% (BUS), 3.87% (CAFE), 3.98% (PEDESTRIAN), 3.92% (STREET)
-------------------
```
```
The RNNLM rescore result with channel-wise training
-------------------
best overall dt05 WER 2.89% (language model weight = 11)
-------------------
dt05_simu WER: 2.99% (Average), 2.74% (BUS), 3.92% (CAFE), 2.54% (PEDESTRIAN), 2.76% (STREET)
-------------------
dt05_real WER: 2.80% (Average), 3.20% (BUS), 2.63% (CAFE), 2.76% (PEDESTRIAN), 2.60% (STREET)
-------------------
et05_simu WER: 3.81% (Average), 3.06% (BUS), 4.30% (CAFE), 3.92% (PEDESTRIAN), 3.98% (STREET)
-------------------
et05_real WER: 3.98% (Average), 4.41% (BUS), 3.83% (CAFE), 3.74% (PEDESTRIAN), 3.94% (STREET)
-------------------
```

danpovey:
OK.  Well, it wasn't expected to make a huge difference, but I think you should still fix it, because the way you did it was effectively a bug; and people might take your scripts and use them in other scenarios where it might make more of a difference.

PR #2015, opened 2017-11-14
[WIP][not-for-merge] Universal am draft

m-wiesner:
Just a draft of the universal acoustic models based on the BABEL data.


m-wiesner:
The main file to look at is 

egs/babel_universal_am/run.sh

jtrmal:
thanks Matthew
I will look into it tomorrow. Your PR seems to change things all over the repo -- does that have any reason? For now, do not worry too much to make the scripts pretty, we have to figure out the recipe layout first.

m-wiesner:
It's probably a combination of a few things: The branch isn't up to date, it's based on some babel scripts which got modified, the adaptation was based off of Pegha's branch .... This is the stuff I was cleaning up.
________________________________

jtrmal:
Matthew, I think we should start building this differently, probably from scratch. This is almost non-reviewable. The adaptation of the model is fairly orthogonal issue to what we are doing, so let's ignore it for now.

jtrmal:
I know what's missing -- can you add example (to the run.sh) how to run
decoding for a predefined set of languages? You can assume the devset will
be always called dev10h.pem
y.

m-wiesner:
That's one thing I've been working on. I'll add the example where you manually replace phonemes, but hopefully I'll have this tree thing done and tested at least some so that replacing missing phonemes is no longer a problem.

________________________________

jtrmal:
For now, just decoding the training language's dev10h.pem would be good
enough.
y.

m-wiesner:
Ok. I have the script for the unseen phonemes though, so I'll use that. If there are no missing phonemes, then everything works with no manual work.


If there are missing phonemes and I don't specify a seen phoneme with which to replace the missing phonemes, then all missing phonemes and resulting words with empty pronunciations get removed from the lexicon. I retrain the LM replacing removed words with <unk> as in the standard training recipe.

________________________________

PR #2025, opened 2017-11-21
[bug fix] Fix resize's silent failure when resize_type == kCopyData

galv:
and stride_type does not equal old stride_type.

### DIFF
--- a/src/matrix/kaldi-matrix.cc
+++ b/src/matrix/kaldi-matrix.cc
@@ -676,13 +676,14 @@ void Matrix<Real>::Resize(const MatrixIndexT rows,
   // resize_type == kCopyData.
   if (resize_type == kCopyData) {
     if (this->data_ == NULL || rows == 0) resize_type = kSetZero;  // nothing to copy.
-    else if (rows == this->num_rows_ && cols == this->num_cols_) { return; } // nothing to do.
+    else if (rows == this->num_rows_ && cols == this->num_cols_ &&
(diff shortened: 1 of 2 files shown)


galv:
Motivation is that I need to convert a matrix with 16-byte alignment on its strides to one with no padding in its strides.

danpovey:
OK, but I think the logic needs to be more like:
```
else if (rows == this->num_rows_ && cols == this->num_cols_ &&
(stride_type == kDefaultStride || this->stride_ == this->num_cols_))
return;
```

: interpret kStrideEqualNumCols as it *needs* to be equal to the num-cols,
and

galv:
That is a little bit more clear, thank you. I pushed a new commit (it's a force push, so you won't see the previous one). All tests passed on my local machine.

moshimo2019:
The `Resize` function doesn't change `stride_` if resize_type == kSetZero and `rows == MatrixBase<Real>::num_rows_ && cols == MatrixBase<Real>::num_cols_`.
The following code will crash if the column of the matrix to be loaded isn't multiple of 4
```
kaldi::Matrix<kaldi::BaseFloat> sli(0, 0, kaldi::kSetZero, kaldi::kStrideEqualNumCols);
std::ifstream ifs("matrix.txt");
sli.Read(ifs, false);
sli.Resize(sli.NumRows(), sli.NumCols(), kaldi::kSetZero, kaldi::kStrideEqualNumCols);
assert(sli.NumCols() == sli.Stride());
```

PR #2030, opened 2017-11-22
added an option in the configure script to build with -fPIC

lordadamson:
I personally need that option because I statically link kaldi into a shared library so kaldi has to be built with `-fPIC`.
To achieve that I'd have to manually change the `kaldi.mk` every time I built kaldi which because redundant and frustrating especially when I forget to do it.

Hopefully other people would find this useful as well.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -61,6 +61,7 @@ Configuration options:
   --version             Display the version of 'configure' and exit
   --static              Build and link against static libraries [default=no]
   --shared              Build and link against shared libraries [default=no]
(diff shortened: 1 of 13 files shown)


galv:
Actually, you don't need to do this. I've done the same thing as well. If you do `./configure --shared`, it will still build the .a files (except on mac, I think, which as far as I know doesn't like static libraries anyway) in their respective directories with position-independent-code, even though it produces .so files in the lib/ directory.

lordadamson:
Oh! I guess that solves it. Thank you!

PR #2031, opened 2017-11-22
Various nnet3 changes

danpovey:
This PR needs to be cleaned up a bit and have experiments added.  It contains various nnet3 changes:
  - Add ScaleAndOffsetComponent, usable for batchnorm with the scale and offset term.
  - Add LinearComponent for when we don't need the bias.
  - Scale down the batch-norm stats slightly on every iteration, to keep them fresh (improves the objective-function values in the `compute_prob_*.*.log` files).
  - Different versions of the res-blocks for image recogntition (I need to add examples of this).
  - Add bottleneck LSTM layer, in which the projection (as in LSTMP) is done to the input and cell values jointly, instead of just the cell values.
  - Some misc code cleanups
  - Add the nnet3 edit command 'apply-svd', which can be used to decompose affine components into two smaller pieces.


danpovey:
I have just merged this PR.  Please be careful when comparing objective-function values after updating to this code, because by default it now scales down the batchnorm stats on each minibatch, keeping the batchnorm stats fresher.  In setups that include batchnorm, this will tend to improve the objective values printed in compute_prob_train.X.log and compute_prob_valid.X.log (where X != "final").

PR #2038, opened 2017-11-26
Fix storage path for mfcc features for tedlium

hhadian:
While cleaning my egs, my mfcc features (for tedlium) were removed too.
It turns out it was because the mfcc features are stored in the same path as egs.
I created this fix to prevent accidental deletion of features on CLSP grid in the future.
I'm not sure if it's worth merging but it seems other databases have this problem too. Should I fix them?


danpovey:
Sure, fix others if you find any.  I'll wait till you update the PR and merge them all at once.

hhadian:
I fixed the rest.

PR #2042, opened 2017-11-27
[build,src] Increase configure version

danpovey:
.. prevent potential build problems due to some changes made when upgrading to 5.3.

### DIFF
--- a/src/cudamatrix/cu-vector.h
+++ b/src/cudamatrix/cu-vector.h
@@ -206,7 +206,12 @@ class CuVectorBase {
   // Set each element to y = (x == orig ? changed : x).
   void ReplaceValue(Real orig, Real changed);
 
+  // Multiplies (*this) by v elementwise: (*this)[i] *= v
   void MulElements(const CuVectorBase<Real> &v);
+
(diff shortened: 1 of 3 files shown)


src/cudamatrix/cu-vector.h
```diff
@@ -206,7 +206,12 @@ class CuVectorBase {
   // Set each element to y = (x == orig ? changed : x).
   void ReplaceValue(Real orig, Real changed);
 
+  // Multiplies (*this) by v elementwise: (*this)[i] *= v
```
galv: I think you should make this change in the comment: `v` ->`v[i]`

src/cudamatrix/cu-vector.h
```diff
@@ -206,7 +206,12 @@ class CuVectorBase {
...
   void MulElements(const CuVectorBase<Real> &v);
+
+  // Divides (*this) by v elementwise: (*this)[i] /= v
```
galv: `v` ->`v[i]`

danpovey:
That's actually unrelated to these changes.  The issue is that some
RNNLM-related code requires cuSparse.

PR #2046, opened 2017-11-29
Sanitize console strings

sdrobert:
In #2044, StringToReadable was employed to ensure invalid buffer
strings were actually printable on console. However, there are many
more places where error messages could contain unprintable chars
than just kaldi-matrix and kaldi-vector (e.g. keys, rxfilenames
from corrupted scp files, etc.). As part of the debug process in
I/O, these partial results are printed all over the place, which
lead to segfaults.

This commit pushes the work to the destructor of
kaldi::MessageLogger. It uses the function StringToReadable to
sanitize the strings before sending them off to callbacks or
stderr.

A couple of reasons why this might _not_ be desirable to merge:
- Every message will have an additional overhead incurred as
  StringToReadable reconstructs a printable string, even when
  it's already printable.
- Whitespace and control characters are sanitized to
  ``[character <int>]``.

### DIFF
--- a/src/base/kaldi-error.cc
+++ b/src/base/kaldi-error.cc
@@ -148,10 +148,12 @@ MessageLogger::MessageLogger(LogMessageEnvelope::Severity severity,
 
 
 MessageLogger::~MessageLogger() KALDI_NOEXCEPT(false) {
-  // sanitize string and remove trailing '\n',
-  std::string str = StringToReadable(ss_.str());
+  // remove trailing '\n',
+  std::string str = ss_.str();
   while (!str.empty() && str[str.length() - 1] == '\n')
     str.resize(str.length() - 1);
+  // sanitization comes after - newlines aren't printable
+  str = StringToReadable(str);
 
   // print the mesage (or send to logging handler),
   MessageLogger::HandleMessage(envelope_, str.c_str());
--- a/src/base/kaldi-error.cc
+++ b/src/base/kaldi-error.cc
@@ -148,8 +148,8 @@ MessageLogger::MessageLogger(LogMessageEnvelope::Severity severity,
 
 
 MessageLogger::~MessageLogger() KALDI_NOEXCEPT(false) {
-  // remove trailing '\n',
-  std::string str = ss_.str();
+  // sanitize string and remove trailing '\n',
+  std::string str = StringToReadable(ss_.str());
   while (!str.empty() && str[str.length() - 1] == '\n')
     str.resize(str.length() - 1);
 
--- a/src/matrix/kaldi-matrix.cc
+++ b/src/matrix/kaldi-matrix.cc
@@ -21,7 +21,6 @@
 // See the Apache 2 License for the specific language governing permissions and
 // limitations under the License.
 
-#include "base/kaldi-utils.h"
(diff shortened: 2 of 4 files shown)


jtrmal:
Sorry, I have never seen the phenomena you are mentioning, i.e. SEGFAULT
because of printing binary data string. Actually, I have seen something
similar only in python.

How would utf-8 be handled? I didn't check the StringToReadable(), do you
sanitize also characters above 128, or are you keeping those intact?
y.

sdrobert:
Ah, the faults might come from when I wrap Kaldi for python, then. Sorry about that.

StringToReadable uses isprint on narrow characters, so no, it can't handle utf-8. Is it possible to infer the encoding cheaply given that LC_ALL=C? Anyways, I'll close this.

jtrmal:
I didn't mean to imply there is something wrong with it, I was just
confused.
y.

jtrmal:
Ad your question -- the encoding is quite hard to infer. Also, sometimes
(for example when the file is damaged or have incorrect structure), the
string being printed does not have any real encoding at all.
y.

sdrobert:
Ah, no, I wasn't offended or anything like that. I was on the fence about this merge in the first place.

Something similar might be worthwhile, however. I wonder if we can modify ``StringToReadable`` to handle characters like ``IsToken``, but include spaces? It might be worthwhile for the sake of better logs.

danpovey:
Actually I'm not convinced, now that I look into it, that
StringToReadable() as it stands is consistent with the overall design of
Kaldi.

Kaldi is designed to be able to read and write strings that may be in
arbitrary encodings, without ever encoding or decoding them but just
treating them as sequences of bytes.  Since we never do anything more
complicated with strings than split on space, this is fine.  We ignore the
locale to the extent possible, and Kaldi C++ programs are intended to be
run with LC_ALL=C.  I don't really like the fact that std:isprint uses the
locale.  StringToReadable(), if applied to UTF-8 encoded tokens under the
condition that LC_ALL=C (which is expected), would turn them into an
unreadable mess instead of actually printing them.


I'm actually considering reverting that previous PR that I merged.  I'm not
convinced that there was a problem to be solved at all.  The worst that
might have happened is that some binary gunk might have been printed to the
terminal.  I think your actual problem might have been at the python
wrapping level.

sdrobert:
Makes sense, thought the bit about the TokenHolder in the previous PR is still true (check the logic in SequentialReader::Open). If you plan on reverting, I suggest you patch that yourself afterwards.

danpovey:
You're right, that KALDI_ERR should be KALDI_WARN.
And I'm fine with the changes that introduce the use of CharToString
instead of bare characters.
I think StringToReadable should be removed; and where it prints that line,
it should,
if the line is longer than 20 characters, truncate it and add "...". before
printing the message.
Also the extra includes are OK.

Is there any chance you can take care of this?

sdrobert:
The ellipses in matrix.cc and vector.cc? Sounds good. Will do tomorrow.

danpovey:
thanks.   What I mean by "..." is if it's printing a line in an error
message, if the line is long, truncate it and add "..." before printing.
No need to change code that you haven't previously touched though- maybe
best not to, it will minimize the chance of problems.

PR #2056, opened 2017-12-01
[src] Bug-fix to conceptual bug in Minimum Bayes Risk/sausage code. Thx:@jtrmal

danpovey:
Fixing this bug makes the sausage stats from the MBR code significantly less sparse.

### DIFF
--- a/src/lat/sausages.cc
+++ b/src/lat/sausages.cc
@@ -114,11 +114,11 @@ double MinimumBayesRisk::EditDistance(int32 N, int32 Q,
       for (int32 q = 0; q <= Q; q++) {
         if (q == 0) {
           alpha_dash_arc(q) = // line 15.
-              alpha_dash(s_a, q) + l(w_a, 0) + delta();
+              alpha_dash(s_a, q) + l(w_a, 0, true);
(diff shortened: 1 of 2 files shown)


jtrmal:
I finally got to test this briefly (on babel pashto)

before:
```
%WER 38.5 | 21825 101803 | 65.4 24.6 10.0 3.8 38.5 29.7 | -0.418 | 
```

after:
```
%WER 38.4 | 21825 101803 | 65.4 24.6 10.0 3.9 38.4 29.7 | -0.419 |
```
For that given setup, I haven't seen more than 0.1 or 0.2 (rarely) % improvement.

On my second test-case (where I have noticed the issue), everything seems ok, but I don't have any experiment set up yet.

jtrmal:
BTW, the key for interpreting the columns is the same as in the ctm files:
```
Err | # Snt # Wrd | Corr    Sub    Del    Ins    Err  S.Err | NCE
```

jtrmal:
To explain, the babel recipe uses MBR decoding even during scoring of a single system.
I don't have any intuition if the gain would be bigger or smaller for let's say lattice combination (via lattice-union+mbr).

PR #2057, opened 2017-12-02
[egs] Add recipe for Mozilla Common Voice corpus v1

entn-at:
This a basic recipe for the recently released Mozilla Common Voice corpus (v1, CC-0 licensed). See https://voice.mozilla.org/data

Some of the data preparation scripts were taken from the voxforge recipe (dict, LM). The systems and chain model setup were adapted from mini_librispeech (including speed perturbation, PCA transform for i-vector extraction, etc.).

I did not tune the setup, the chain system already has WERs of about 5% (see RESULTS).


entn-at:
Note that this recipe currently only uses the "valid" portion of the corpus, that is, utterances that have had at least 2 people listen to them, and the majority of those listeners said the audio matches the text.

egs/commonvoice/s5/local/chain/tuning/run_tdnn_1a.sh
```diff
@@ -0,0 +1,288 @@
...
+  input dim=100 name=ivector
+  input dim=40 name=input
+
+  # please note that it is important to have input layer with the name=input
+  # as the layer immediately preceding the fixed-affine-layer to enable
+  # the use of short notation for the descriptor
+  fixed-affine-layer name=lda input=Append(-2,-1,0,1,2,ReplaceIndex(ivector, t, 0)) affine-transform-file=$dir/configs/lda.mat
+
+  # the first splicing is moved before the lda layer, so no splicing here
+  relu-batchnorm-layer name=tdnn1 dim=512
```
danpovey: This system is rather small for a 500-hour dataset.  You may want to try dim=768 instead of 512.

I also notice that in the RESULTS file you called this 1e (IIRC).

egs/commonvoice/s5/local/score.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
jtrmal: can you replace this script by a symlink steps/score_kaldi.sh, please?

egs/commonvoice/s5/run.sh
```diff
@@ -0,0 +1,159 @@
...
+. ./path.sh
+
+stage=0
+
+set -euo pipefail
+
+if [ $stage -le 0 ]; then
+  mkdir -p $data
+
+  local/download_and_untar.sh $(/usr/bin/dirname $data) $data_url
```
jtrmal: Does the absolute pathname `/usr/bin/dirname` have any particular reason?

egs/commonvoice/s5/local/chain/tuning/run_tdnn_1a.sh
```diff
@@ -0,0 +1,288 @@
...
+    --chain.leaky-hmm-coefficient=0.1 \
+    --chain.l2-regularize=0.00005 \
+    --chain.apply-deriv-weights=false \
+    --chain.lm-opts="--num-extra-lm-states=2000" \
+    --trainer.srand=$srand \
+    --trainer.max-param-change=2.0 \
+    --trainer.num-epochs=4 \
+    --trainer.frames-per-iter=1500000 \
+    --trainer.optimization.num-jobs-initial=3 \
+    --trainer.optimization.num-jobs-final=3 \
```
danpovey: If your setup allows it, it would be a good idea, for speed, to increase num-jobs-final to something like 12.

egs/commonvoice/s5/local/format_data.sh
```diff
@@ -0,0 +1,53 @@
...
+# and the corresponding lang_test_* directory.
+
+echo "--- Preparing the grammar transducer (G.fst) for testing ..."
+
+test=data/lang_test
+mkdir -p $test
+for f in phones.txt words.txt phones.txt L.fst L_disambig.fst phones; do
+    cp -r data/lang/$f $test
+done
+cat $lmdir/lm.arpa | \
```
jtrmal: I'd prefer the rest of the script to be replaced by utils/format_lm.sh

danpovey:
Sometimes we use the script utils/make_absolute.sh which helps with
portability.

egs/commonvoice/s5/local/chain/tuning/run_tdnn_1a.sh
entn-at: Unfortunately, I only have 3 GPUs, but I will change it in the script to 12.

danpovey:
It would be better if you installed GridEngine and set up the 'gpu'
resource, if you haven't already-- that way you  could run with more jobs
even though you only have 3 GPUs.   I don't like scripts checked in that
haven't been tested with those settings.

entn-at:
No problem, I have gridengine set up. I'm going test it with jobs-final=12 (It's just going to take a while longer).

danpovey:
It actually won't take any longer because the total number of jobs is the
same as before, it's just that (conceptually) more of them are in parallel.
I recommend to increase the dim from 512 to 768 at the same time.

entn-at:
I made the following changes:

- change `score.sh` to a symlink to `steps/score_kaldi.sh`
- remove absolute path to `dirname`
- replace `local/format_data.sh` with a call to `utils/format_lm.sh`
- use `<unk>` instead of `SIL` in lexicon

I'm currently running the whole recipe from start to finish. Once that's done I'll add another commit with the changes to `run_tdnn_1a.sh` and `RESULTS`.

danpovey:
Thanks a lot!  @jtrmal, please merge when and if you're OK with it.  No need to check more, necessarily.

entn-at:
Thanks for the quick review and the helpful comments!

jtrmal:
all right, I'll merge. Thanks a lot!

Cyrix126:
It would be nice to adapt it to the updated version of the mozilla dataset and to add option for choosing the language.

danpovey:
You're right it would be nice.  And I am hoping someone will contribute
it.  But I have more pressing matters to attend to.

PR #2059, opened 2017-12-03
Pruned compose

hainan-xv:
Pruned lattice rescoring for both Kaldi-RNNLM and TensorFlow RNNLM; supports both G.fst and G.carpa as the old LM.


src/latbin/lattice-lmrescore-const-arpa-kaldi-rnnlm-pruned.cc
```diff
@@ -0,0 +1,181 @@
+// latbin/lattice-lmrescore-kaldi-rnnlm-pruned.cc
+
+// Copyright 2017 Johns Hopkins University (author: Daniel Povey)
```
danpovey: I don't believe it's necessary to have an extra program for the const-arpa case.  You could just have an option to lattice-lmrescore-kaldi-rnnlm-pruned.  They OnDemandDeterministicFst objects share the same interface, so the if-statement wouldn't have to encompass the bulk of the code, just the code that initializes that object.

hainan-xv: Oh sure. I thought about having both in one binary but then I saw there are separate "lattice-lmrescore" and "lattice-lmrescore-const-arpa" under latbin/ so I separated them.

I will make the changes.

egs/swbd/s5c/local/rnnlm/tuning/run_lstm_1e.sh
```diff
@@ -95,12 +96,17 @@ fi
 
 if [ $stage -le 4 ] && $run_rescore; then
   echo "$0: Perform lattice-rescoring on $ac_model_dir"
-  LM=sw1_fsh_fg
+  LM=sw1_fsh_fg # using the 4-gram const arpa file as old lm
+#  LM=sw1_tg # if using the original 3-gram G.fst as old lm
+  pruned=
+  if $pruned_rescoring; then
```
danpovey: the option above is named 'pruned_rescore', not 'rescoring'.

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat.sh
```diff
@@ -0,0 +1,103 @@
...
+    --max-ngram-order=$max_ngram_order \
+    $rnnlm_dir/unk.probs $rnnlm_dir/wordlist.rnn.final $oldlang/words.txt ark:- "$rnnlm_dir/rnnlm" \
+    "ark,t:|gzip -c>$outdir/lat.JOB.gz" || exit 1;
+fi
+if ! $skip_scoring ; then
+  err_msg="Not scoring because local/score.sh does not exist or not executable."
+  [ ! -x local/score.sh ] && echo $err_msg && exit 1;
+  local/score.sh --cmd "$cmd" $data $oldlang $outdir
+else
+  echo "Not scoring because requested so..."
```
danpovey: All messages printed (including 'err_msg' above) should start with "$0: ". 
Reword to "$0: Skipping scoring because --skip-scoring was specified."

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+# Copyright 2015  Guoguo Chen
+#           2017  Hainan Xu
+# Apache 2.0
+
+# This script rescores lattices with RNNLM trained with TensorFlow.
+# It uses a pruend algorithm to speed up the runtime and improve the accuracy.
```
danpovey: typo

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+# which is an improved version over steps/tfrnnlm/lmrescore_rnnlm_lat.sh,
+# which uses the exact same interface
+# The details of the pruning algorithm is described in
+# http://www.cs.jhu.edu/~hxu/tf.pdf
```
danpovey: this may not be a very permanent location.  My homepage is more permanent, it better to put it there, you can send me the pdf if it's not there already(remind me where it's being published).

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# Begin configuration section.
+cmd=run.pl
+skip_scoring=false
+max_ngram_order=4 # Approximate the lattice-rescoring by limiting the max-ngram-order
+                  # if it's set, it merges histories in the lattice if they share
+                  # the same ngram history and this prevents the lattice from 
+                  # exploding exponentially. Details of the n-gram approximation
+                  # method are described in section 2.3 of the paper
+                  # http://www.cs.jhu.edu/~hxu/tf.pdf
```
danpovey: same here- more permanent location.

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+inv_acwt=10
```
danpovey: I'm not sure why you are having inv_acwt being specified, instead of acwt.  The reason we sometimes use inv_acwt is because it interacts better with queue.pl and the like, but in this case that issue doesn't exist, so I think it would be better to specify acwt directly.

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+
+. ./utils/parse_options.sh
+
+if [ $# != 5 ]; then
+   echo "Does language model rescoring of lattices (remove old LM, add new LM)"
+   echo "with RNNLM."
+   echo ""
+   echo "Usage: $0 [options] <old-lang-dir> <rnnlm-dir> \\"
+   echo "                   <data-dir> <input-decode-dir> <output-decode-dir>"
+   echo " e.g.: $0 ./rnnlm data/lang_tg data/test \\"
```
danpovey: This example doesn't seem to match up with the required args; and I don't like './rnnlm': why use the "./" on just that one arg?

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+oldlang=$1
+rnnlm_dir=$2
+data=$3
+indir=$4
+outdir=$5
+
+oldlm=$oldlang/G.fst
+carpa_option=
+if [ ! -f $oldlm ]; then
+  echo "$0: file $oldlm not found; looking for $oldlang/G.carpa"
```
danpovey: I don't like "looking for".  If you want to print which old LM you're using, just say which one you're using.
If you fail to find one then print an error message.

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+  oldlm=$oldlang/G.carpa
+  carpa_option="--use-const-arpa=true"
+fi
+  
+[ ! -f $oldlm ] && echo "$0: Missing file $oldlm" && exit 1;
+[ ! -f $rnnlm_dir/rnnlm ] && [ ! -d $rnnlm_dir/rnnlm ] && echo "$0: Missing file $rnnlm_dir/rnnlm" && exit 1;
```
danpovey: I don't understand why it would be a directory.  Since you are testing for several files, better to do "for f in ...".  You can test for lat.1.gz, which should always be present.

egs/wsj/s5/steps/tfrnnlm/lmrescore_rnnlm_lat_pruned.sh
```diff
@@ -0,0 +1,92 @@
...
+  $carpa_option $oldlm $oldlang/words.txt \
+  $rnnlm_dir/unk.probs $rnnlm_dir/wordlist.rnn.final "$rnnlm_dir/rnnlm" \
+  "ark:gunzip -c $indir/lat.JOB.gz|" "ark,t:|gzip -c>$outdir/lat.JOB.gz" || exit 1;
+
+if ! $skip_scoring ; then
+  err_msg="Not scoring because local/score.sh does not exist or not executable."
+  [ ! -x local/score.sh ] && echo $err_msg && exit 1;
+  local/score.sh --cmd "$cmd" $data $oldlang $outdir
+else
+  echo "Not scoring because requested so..."
```
danpovey: again, fix this message and the err_msg above.

scripts/rnnlm/lmrescore_pruned.sh
```diff
@@ -0,0 +1,108 @@
...
+if $normalize; then
+  normalize_opt="--normalize-probs=true"
+fi
+special_symbol_opts=$(cat $rnnlm_dir/special_symbol_opts.txt)
+
+word_embedding=
+if [ -f $rnnlm_dir/word_embedding.final.mat ]; then
+  word_embedding=$rnnlm_dir/word_embedding.final.mat
+else
+  word_embedding="\"rnnlm-get-word-embedding $rnnlm_dir/word_feats.txt $rnnlm_dir/feat_embedding.final.mat -|\""
```
danpovey: normally, the interior quotes would be single quotes rather than escaped double quotes.

src/fstext/kaldi-fst-io.h
```diff
@@ -81,6 +81,8 @@ template <class Arc>
 void ReadFstKaldi(std::istream &is, bool binary,
                   VectorFst<Arc> *fst);
 
+fst::VectorFst<fst::StdArc> *ReadAndPrepareLmFst(std::string rxfilename);
```
danpovey: Needs documentation

src/latbin/lattice-lmrescore-kaldi-rnnlm-pruned.cc
```diff
@@ -0,0 +1,206 @@
...
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    typedef kaldi::int32 int32;
+    typedef kaldi::int64 int64;
+    using fst::SymbolTable;
+    using fst::VectorFst;
+    using fst::StdArc;
+    using fst::ReadFstKaldi;
+    using std::unique_ptr;
```
danpovey: remove this using std::unique_ptr, from here and the other files where it appears, as you don't seem to need it.

src/latbin/lattice-lmrescore-kaldi-rnnlm-pruned.cc
```diff
@@ -0,0 +1,206 @@
...
+    const char *usage =
+        "Rescores lattice with kaldi-rnnlm. This script is called from \n"
+        "scripts/rnnlm/lmrescore_pruned.sh. An example for rescoring \n"
+        "lattices is at egs/swbd/s5/local/rnnlm/run_rescoring.sh \n"
+        "\n"
+        "Usage: lattice-lmrescore-kaldi-rnnlm-pruned [options] \\\n"
+        "             <old-lm-rxfilename> <embedding-file> \\\n"
+        "             <raw-rnnlm-rxfilename> \\\n"
+        "             <lattice-rspecifier> <lattice-wspecifier>\n"
+        " e.g.: lattice-lmrescore-kaldi-rnnlm-pruned --lm-scale=-1.0 fst_words.txt \\\n"
```
danpovey: you should probably add an example with const-arpa.

src/tfrnnlmbin/Makefile
```diff
@@ -33,7 +33,7 @@ ADDLIBS = ../lat/kaldi-lat.a ../lm/kaldi-lm.a ../fstext/kaldi-fstext.a \
           ../base/kaldi-base.a ../tfrnnlm/kaldi-tensorflow-rnnlm.a
 
 LDLIBS +=  -lz -ldl -fPIC -lrt
-LDLIBS += -L$(TENSORFLOW)/bazel-bin/tensorflow -ltensorflow_cc -ltensorflow_framework
+LDLIBS += -L$(TENSORFLOW)/bazel-bin/tensorflow -ltensorflow_cc# -ltensorflow_framework
```
danpovey: this hash symbol is strange... not sure why it's here.  explain what's going on.

src/tfrnnlmbin/lattice-lmrescore-tf-rnnlm.cc
```diff
@@ -35,17 +35,18 @@ int main(int argc, char *argv[]) {
...
         "             <word-symbol-table-rxfilename> <lattice-rspecifier> \\\n"
         "             <rnnlm-rxfilename> <lattice-wspecifier>\n"
-        " e.g.: lattice-lmrescore-tf-rnnlm --lm-scale=-1.0 unkcounts.txt rnnwords.txt \\\n"
-        "              words.txt ark:in.lats rnnlm ark:out.lats\n";
+        " e.g.: lattice-lmrescore-tf-rnnlm --lm-scale=0.5 "
+        "    data/tensorflow_lstm/unkcounts.txt data/tensorflow_lstm/rnnwords.txt \\\n"
+        "    data/lang/words.txt ark:in.lats data/tensorflow_lstm/rnnlm ark:out.lats\n";
 
     ParseOptions po(usage);
     int32 max_ngram_order = 3;
```
ngoel17: what's ngram_order needed for?

src/tfrnnlmbin/lattice-lmrescore-tf-rnnlm-pruned.cc
```diff
@@ -0,0 +1,201 @@
...
+
+    if (po.NumArgs() != 7 && po.NumArgs() != 6) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    std::string lm_to_subtract_rxfilename, lats_rspecifier, rnn_word_list,
+      word_symbols_rxfilename, rnnlm_rxfilename, lats_wspecifier, unk_prob_file;
+    if (po.NumArgs() == 6) {
+      lm_to_subtract_rxfilename = po.GetArg(1),
```
ngoel17: what does this do? lm_to_subtract_rxfilename? I think this the G.fst from the arpa model. Maybe call it decode_arpa_fst_rxfilename ? and change other names accordingly.

src/tfrnnlmbin/lattice-lmrescore-tf-rnnlm.cc
danpovey: it's registered on the command line, you can see the help string there...

src/tfrnnlmbin/lattice-lmrescore-tf-rnnlm-pruned.cc
danpovey: i see your point but in general it might be the result of a previous rescoring pass, it might not be the one we decoded with.

Issue #2071, opened 2017-12-08
There's something wrong in steps/data/augment_data_dir.py


by2101:
I use this script to augment my data, and there's something wrong.
Specifically, the generated wav.scp is like this:
`863-000002.wav-noise wav-reverberate --shift-output=true --additive-signals='/data4/music_noise_16K_norm/norm_110.wav wav-reverberate --duration=6.16 - - |,/data4/music_noise_16K_norm/norm_136.wav' --start-times='0,0' --snrs='10,5' /data4/f001/wav/000002.wav - |`
It can not be used for wav-to-duration. And I directly run command
 ` wav-reverberate --shift-output=true --additive-signals='/data4/music_noise_16K_norm/norm_110.wav wav-reverberate --duration=6.16 - - |,/data4/music_noise_16K_norm/norm_136.wav' --start-times='0,0' --snrs='10,5' /data4/f001/wav/000002.wav -` failed.

So there's something wrong in steps/data/augment_data_dir.py line 82 to line 84. 
`    if len(bg_noise_utts) > 0:
        num = random.choice(num_opts)
        for i in range(0, num):
            noise_utt = random.choice(bg_noise_utts)
            noise = noise_wavs[noise_utt] + " wav-reverberate --duration=" \
            + dur_str + " - - |"
            snr = random.choice(bg_snr_opts)
            snrs.append(snr)
            start_times.append(0)
            noises.append(noise)`

I wrote a version fixed this bug. 
`    # Now handle the background noises
    if len(bg_noise_utts) > 0:
        num = random.choice(num_opts)
        for i in range(0, num):
            noise_utt = random.choice(bg_noise_utts)
            noise = "wav-reverberate --duration=" \
            + dur_str + " " + noise_wavs[noise_utt] + " - |"
            snr = random.choice(bg_snr_opts)
            snrs.append(snr)
            start_times.append(0)
            noises.append(noise)` 

I would like to provide this version. Otherwise, I wrote a script to generate some sample from wav.scp to test. I think it is useful to test some wave samples. I also would like to provide it.

danpovey:
Yes, please do provide your changed version as a pull request.  It's best
if you find someone local who can explain how github works, if you don't
already know how, but if not, please contact Samuel Zhang <
samuelzhang1104@gmail.com>.  Please bear in mind that in general the
wav-file "rspecifiers" can be of two forms:
 (1) plain filenames, that don't end in the pipe symbol
and
[...]

PR #2073, opened 2017-12-09
Fix bug in augment_data_dir.py

by2101:
I fixed bug in augment_data_dir.py. Now it creates wav.scp file like this:
`
noise_f001_000006.wav wav-reverberate --shift-output=true --additive-signals='wav-reverberate --duration=5.863812 /data4/pointsource_noises_norm/norm_noise-free-sound-0731.wav - |,/data4/Nonspeech_16k_norm/norm_n3.wav' --start-times='0,0' --snrs='15,5' /data4/wavdata/863-ASCCD/f001/wav/000006.wav - |
`
gen_wav_sample.py can generate some wave file samples based on wav.scp for user to test the effect of adding noise.


egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -21,7 +22,7 @@ def GetArgs():
         "babble, for example.  Foreground noises are added sequentially, according to a specified "
         "interval.  See also steps/data/reverberate_data_dir.py "
         "Usage: augment_data_dir.py [options...] <in-data-dir> <out-data-dir> "
-        "E.g., steps/data/augment_data_dir.py --utt-suffix aug --fg-snrs 20:10:5:0 --bg-snrs 20:15:10 "
+        "E.g., steps/data/augment_data_dir.py --utt_prefix aug --fg-snrs 20:10:5:0 --bg-snrs 20:15:10 "
```
danpovey: This seems to be unrelated to the fix you said you were going to make, and anyway it's not compatible-- there are many scripts that use that option, and you are changing its name.
Also, it was a suffix (not a prefix) for a good reason-- making it a prefix would destroy a certain data-being-sorted property that is needed.

egs/wsj/s5/steps/data/gen_wav_sample.py
```diff
@@ -0,0 +1,108 @@
+#encoding:utf8
+# Copyright     2017  Ye Bai
+# Apache 2.0
+#
```
danpovey: it's good if you tested it, but please don't check in the testing script.

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -83,8 +84,8 @@ def AugmentWav(utt, wav, dur, fg_snr_opts, bg_snr_opts, fg_noise_utts, \
         num = random.choice(num_opts)
         for i in range(0, num):
             noise_utt = random.choice(bg_noise_utts)
-            noise = noise_wavs[noise_utt] + " wav-reverberate --duration=" \
-            + dur_str + " - - |"
+            noise = "wav-reverberate --duration=" \
```
danpovey: This fix doesn't seem right- it looks like you didn't undertsand what I was saying about the pipe at the end.  There needs to be some conditional code.  Read my responses to you on github and/or kaldi-help more carefully.

egs/wsj/s5/steps/data/augment_data_dir.py
by2101: In my data, utt id is like this: 00001.wav, so I prefer add prefix rather than suffix. I don't want to break the extend name.

egs/wsj/s5/steps/data/gen_wav_sample.py
by2101: This script was tested ok.

egs/wsj/s5/steps/data/augment_data_dir.py
by2101: Oh, I see. The noise file may be a pipline too.

by2101: I think I can add "" to the pipeline, the result will be like this:
wav-reverberate --duration=4.588438 "wav-copy /data4/wavdata20171208/music_noise_16K_norm/norm_238.wav - |" - 
I think it works.

danpovey: My point was that the original file in the wav.scp may already have been a pipeline.

danpovey:
@david-ryan-snyder, can you please take responsibility for reviewing this?

egs/wsj/s5/steps/data/gen_wav_sample.py
danpovey: Please don't check in this gen_wav_sample.py script.

egs/wsj/s5/steps/data/augment_data_dir.py
by2101: I think it is tackled between line 111 to 118.

david-ryan-snyder:
@danpovey,@by2101, Is there a description of the bug that motivated this somewhere? What functionality is this PR adding?

by2101:
Here #2071 
Specifically, the original version generated wav.scp like this:
863-000002.wav-noise wav-reverberate --shift-output=true --additive-signals='/data4/music_noise_16K_norm/norm_110.wav wav-reverberate --duration=6.16 - - |,/data4/music_noise_16K_norm/norm_136.wav' --start-times='0,0' --snrs='10,5' /data4/f001/wav/000002.wav - |

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -123,7 +124,7 @@ def CopyFileIfExists(utt_suffix, filename, input_dir, output_dir):
         if len(utt_suffix) > 0:
             new_dict = {}
             for key in dict.keys():
-                new_dict[key + "-" + utt_suffix] = dict[key]
+                new_dict[key + "_" + utt_suffix] = dict[key]
```
david-ryan-snyder: I don't see why this needs to change.

by2101: uh, I want to keep it consistent with reverberate_data_dir.py generated wav.scp.

egs/wsj/s5/steps/data/augment_data_dir.py
david-ryan-snyder: Lines 111-118 handle that for the speech file you're adding noise to. You need to do the same kind of thing for the noise file too. I think what you have will fail if the noise wav.scp uses pipes.

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -175,13 +176,15 @@ def main():
         new_wav = AugmentWav(utt, wav, dur, fg_snrs, bg_snrs, fg_noise_utts,
             bg_noise_utts, noise_wavs, noise_reco2dur, args.fg_interval,
             num_bg_noises)
-        new_utt = utt + "-" + args.utt_suffix
+        new_utt = utt + "_" + args.utt_suffix
```
david-ryan-snyder: I would err on the side of not changing these things unless there's a good reason.

egs/wsj/s5/steps/data/augment_data_dir.py
by2101: I add quotes to the noise wav. Now it generates like this:
wav-reverberate --duration=4.588438 "wav-copy /data4/wavdata20171208/music_noise_16K_norm/norm_238.wav - |" -
It can deal with pipeline case and non-pipeline case

by2101:
Because in reverberate_data_dir.py, the generated utt id in wav.scp use "_". I want to keep add noise version consistent with add reverberation version. So I changed "-" to "_".

egs/wsj/s5/steps/data/augment_data_dir.py
david-ryan-snyder: There are other scripts that use "-" for this purpose instead of "_", such as perturb_data_dir_speed.sh.

Anyway, we already have recipes that are using this script, so to prevent any potential back compatibility issues, it's better if you leave it how it was.

by2101: OK, I will modify it to "-".

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -175,13 +176,15 @@ def main():
         new_wav = AugmentWav(utt, wav, dur, fg_snrs, bg_snrs, fg_noise_utts,
             bg_noise_utts, noise_wavs, noise_reco2dur, args.fg_interval,
             num_bg_noises)
-        new_utt = utt + "-" + args.utt_suffix
+        new_utt = utt + "-" + args.utt_suffix
```
david-ryan-snyder: Sorry, can you also remove the trailing whitespace? You can set up your vimrc to do this for you automatically: https://stackoverflow.com/questions/356126/how-can-you-automatically-remove-trailing-whitespace-in-vim.

egs/wsj/s5/steps/data/gen_wav_sample.py
```diff
@@ -0,0 +1,103 @@
+#encoding:utf8
```
david-ryan-snyder: Dan is encouraging using python3 for all new python scripts... so you might want to put #!/usr/bin/env python3 at the top of the script.

egs/wsj/s5/steps/data/gen_wav_sample.py
```diff
@@ -0,0 +1,103 @@
...
+                                            " The format is like this:"
+                                            "rvb_noise_f001_000008.wav \n"
+                                            "rvb_noise_f001_000009.wav \n"
+                                            "...")
+                                        
+    parser.add_argument("scp_file",
+                        help="Input data")
+    parser.add_argument("output_dir",
+                        help="Output data directory")
+
```
david-ryan-snyder: Indent appears inconsistent. Might help to look at some recent Kaldi python scripts for guidance.

egs/wsj/s5/steps/data/gen_wav_sample.py
```diff
@@ -0,0 +1,103 @@
...
+                reco = " ".join(items[1:]).strip()
+                if reco[-1] == "|":
+                    execute_command("{wav} > {outpath}/{out_name}".format(wav=reco[0:-1],
+                                outpath=output_dir,out_name=prefix+id+".wav"))
+                else:
+                    execute_command("wav-copy '{wav}' {outpath}/{out_name}".format(wav=reco,
+                                outpath=output_dir,out_name=prefix+id+".wav"))
+                sample_num -=1          
+    print("Done.")
+
```
david-ryan-snyder: lines 92-103 can probably be removed, along with any other trailing whitespace.

egs/wsj/s5/steps/data/gen_wav_sample.py
david-ryan-snyder: Also, can you add some comment at the top of the file explaining what it does?

egs/wsj/s5/steps/data/gen_wav_sample.py
```diff
@@ -0,0 +1,104 @@
...
+                                                 "E.g. python gen_wav_sample.py wav.scp sample",
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    parser.add_argument("--num", type=int, dest = "sample_num", default = 10,
+                                        help="Number of samples to generate.")
+
+    parser.add_argument("--prefix", type=str, dest = "prefix", default = "",
+                                        help="Prefix of generated samples.")
+
+    parser.add_argument("--record_file", type=str, dest = "record_file", default = "",
```
david-ryan-snyder: This should be --record-file

egs/wsj/s5/steps/data/gen_wav_sample.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+from __future__ import print_function
```
danpovey: I wanted him to delete gen_wav_sample.py from the PR anyway-- it's just something he used for testing, I think.

by2101: Oh，I think some people like me may want to listen some samples after doing adding reverbration or adding noises. Or do you think it should be moved to utils directory?

danpovey: The purpose of the script wasn't clear, and the script doesn't clearly explain what its arguments mean (i.e. what the interface is).
People may want to do that, but I don't want you to mix that up with the bug-fix commit.  You are making this too complicated. Talk to David by email  (David Snyder <david.ryan.snyder@gmail.com>) about this possibility, but in any case, do it as a separate commit.

by2101: OK, I will remove it from this PR.

egs/wsj/s5/steps/data/augment_data_dir.py
david-ryan-snyder: @by2101, could you also apply this fix to the foreground noises in the next loop? It looks like the fix has only been done for background noises so far. After that, I think this will be basically ready to merge.

david-ryan-snyder:
Also, before we merge this (please add the fix to the "fg" noises first), we'll want you to squash the previous commits into a single one to clean up the commit history. This will be useful to read if you haven't done it before: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git .

Squashing all of the commits isn't always necessary. But in this case, it's going to be preferred as files have been added and removed within the same pull request. Doing this will clean up the commit history.

egs/wsj/s5/steps/data/augment_data_dir.py
by2101: Well, in foreground noise loop, it works well.
The bug is caused by using wav-reberberate to change duration: 
noise = noise_wavs[noise_utt] + " wav-reverberate --duration=" + dur_str + " - - |"
when noise_wavs[noise_utt] is not a pipeline, it fail.
But in foreground, noise = noise_wavs[noise_utt] works well whenever it is pipeline or a file.

by2101:
OK, I will squash it. Thank you for your review!

by2101:
Sorry, I run this command to squash these commit:
git reset --soft HEAD~15 && 
git commit --edit -m"$(git log --format=%B --reverse HEAD..HEAD@{1})"
It does not work. Maybe I did some wrong command. Could you please help me to squash it? Or maybe I can create a new pull request to commit this file? @david-ryan-snyder

david-ryan-snyder:
Unfortunately I don't have time to figure that out right now (why you couldn't squash the commits). You could create a new PR for this if you want. But you might find it helpful to figure out how to do this yourself, incase it comes up later. Up to you though.

PR #2074, opened 2017-12-09
Minor cleanup and clarification when running online2-wav-nnet3-latgen-faster with aishell recipe model.

qin:


xiaosdawn:
Hi，
Recently, I have trained my chain model, then I tried to use 'online2-wav-nnet3-latgen-faster' for online decoding by calling 'steps/online/nnet3/decode.sh'. Before decoding, 'steps/online/nnet3/prepare_online_decoding.sh' was called to generate the online config. The details of my script as follows:
`
steps/online/nnet3/prepare_online_decoding.sh \
        --mfcc-config conf/mfcc_hires.conf \
        data/lang ${dir} ${dir}/decode_online

    rm $dir/.error 2>/dev/null || true
    for decode_set in test; do
    (
        # note: we just give it "$decode_set" as it only uses the wav.scp, the
        # feature type does not matter.
        local/online/nnet3/decode.sh --nj 10 --cmd "$decode_cmd" \
            --acwt 1.0 --post-decode-acwt 10.0 \
            $graph_dir data/${decode_set}_hires \
            ${dir}/decode_online/decode_${test_data}_${decode_set} || exit 1;
     ) || touch $dir/.error &
    done
`
I just used MFCC for network input, when training my chain model. So I didn't give the 'ivector_extractor' parameter when calling prepare_online_decoding.sh.

danpovey:
These kinds of questions should go to kaldi-help, but you'd need to provide more details, e.g. some of the decoding log; what data you were running with; what recipe you based it on; what changes you made.

Issue #2068, opened 2017-12-07
[RNNLM] Fail with large vocab on billion-word LM


francoishernandez:
Hi,
As part of our experiments with Kaldi-RNNLM, we wanted to train a baseline on the One Billion Word LM benchmark dataset.
Trying with a vocab of ~800k it keeps failing at the first 'compute-prob' iteration. On an extract of ~20M words with the same vocab, it fails too. But on the same extract, using Tedlium's 152k vocab, it works.
Is such a big vocab too large for such models?

Thanks in advance!
F

Here are the logs from `compute_prob.0.log`:
```
# rnnlm-get-egs --bos-symbol=793471 --brk-symbol=793473 --eos-symbol=793472 --vocab-size=793474 exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/text/dev.txt ark:- | rnnlm-compute-prob --use-gpu=yes exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/0.raw "rnnlm-get-word-embedding exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/word_feats.txt exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/feat_embedding.0.mat -|" ark:- 
# Started at Thu Dec  7 00:32:09 CET 2017
#
rnnlm-compute-prob --use-gpu=yes exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/0.raw 'rnnlm-get-word-embedding exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/word_feats.txt exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/feat_embedding.0.mat -|' ark:- 
rnnlm-get-egs --bos-symbol=793471 --brk-symbol=793473 --eos-symbol=793472 --vocab-size=793474 exp/rnnlm_2lstmp_3tdnn_2x256_e1024_1B/text/dev.txt ark:- 
LOG (rnnlm-get-egs[5.3.4~1-4ac4]:Process():rnnlm-example.cc:703) Processed 6075 lines of input.
WARNING (rnnlm-compute-prob[5.3.4~1-4ac4]:SelectGpuId():cu-device.cc:183) Not in compute-exclusive mode.  Suggestion: use 'nvidia-smi -c 3' to set compute exclusive mode
[...]

PR #2077, opened 2017-12-11
Minor change to cudamatrix - cu-sparse-matrix.cc

francoishernandez:
As discussed in issue #2068 .
Large vocabularies caused int32 overflow issues in matrix creation.
Solution proposed by @danpovey: cast num_cols to int64 to prevent it.
The use of a large vocabulary may still cause memory issues.

### DIFF
--- a/src/cudamatrix/cu-sparse-matrix.cc
+++ b/src/cudamatrix/cu-sparse-matrix.cc
@@ -388,7 +388,7 @@ void CuSparseMatrix<Real>::CopyFromSmat(const CuSparseMatrix<Real>& smat,
       Resize(smat.NumCols(), smat.NumRows(), smat.NumElements(), kUndefined);
       CuTimer tim;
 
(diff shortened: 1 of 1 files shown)


src/cudamatrix/cu-sparse-matrix.cc
```diff
@@ -388,7 +388,7 @@ void CuSparseMatrix<Real>::CopyFromSmat(const CuSparseMatrix<Real>& smat,
       Resize(smat.NumCols(), smat.NumRows(), smat.NumElements(), kUndefined);
       CuTimer tim;
 
-      CUSPARSE_SAFE_CALL(
+      CU_SAFE_CALL(
```
danpovey: Looks to me like there was a merge problem here.  It should be CUSPARSE_SAFE_CALL

francoishernandez: Fixed it, thanks.

PR #2081, opened 2017-12-13
fix augment_data_dir.py

by2101:
Fixed a bug in augment_data_dir.py.
I create a new PR because there are too many commits in old one.


danpovey:
@david-ryan-snyder, if you say this is OK  I'll merge.

david-ryan-snyder:
Yes, it's fine. @by2101, thanks!

PR #2094, opened 2017-12-21
fix a bug in RNNLM where it might exhaust GPU memory when computing o…

hainan-xv:
…bjf without sampling

### DIFF
--- a/src/rnnlm/rnnlm-example-utils.h
+++ b/src/rnnlm/rnnlm-example-utils.h
@@ -33,8 +33,6 @@
 namespace kaldi {
 namespace rnnlm {
 
-const int64 MAX_MAT_SIZE = 1000000000;
-
 /**
    @file rnnlm-example-utils.h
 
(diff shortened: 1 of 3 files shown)


src/rnnlm/rnnlm-example-utils.h
```diff
@@ -33,6 +33,8 @@
 namespace kaldi {
 namespace rnnlm {
 
+const int64 MAX_MAT_SIZE = 1000000000;
```
danpovey: It's rather ugly to make this a const int like this (and it would, I think, be misnamed, should probably be kMaxMatSize).
Put it in the config class.  And clarify the units, whether it's bytes or elements.  e.g.  max-logprob-elements or max-logprob-bytes.

src/rnnlm/rnnlm-example-utils.h
```diff
@@ -165,6 +167,11 @@ struct RnnlmObjectiveOptions {
                  "average denominator part of the objective, for this minibatch, "
                  "is more negative than this value.  Set this to 0.0 to use "
                  "unmodified objective function.");
+    po->Register("max-logprob-elements", &max_logprob_elements,
+                 "Maximum number of elements when we allocate a matrix of size "
+                 "[minibatch-size, num-words] for computing logprobs of words. "
+                 "If the size is exceed, we will break the matrix along the "
```
danpovey: exceeded

PR #2097, opened 2017-12-21
Keeping tree leaves to a multiple of 8

vimalmanohar:


src/tree/build-tree.cc
```diff
@@ -212,8 +214,26 @@ EventMap *BuildTree(Questions &qopts,
...
-    int32 num_leaves = 0;
-    EventMap *tree_renumbered = RenumberEventMap(*tree_clustered, &num_leaves);
+    int32 num_leaves_out = 0;
+    EventMap *tree_renumbered;
+    if (round_num_leaves) {
+      int32 num_leaves_required = ((num_leaves - num_removed) / 8) * 8;
+      std::vector<EventMap*> leaf_mapping;
+
+      int32 num_actually_removed = ClusterEventMapGetMapping(
+          *tree_clustered, stats, -kLogZeroBaseFloat, &leaf_mapping,
```
galv: Perhaps a more idiomatic way to get positive infinity is `std::numeric_limits<BaseFloat>::infinity()` instead of `-kLogZeroBaseFloat`

danpovey: I agree with Daniel on how to get infinity.  Is this tested?  And did you test that the regular tree-building (without this option set) runs the same way it did before?

vimalmanohar: I made the changes. I tested it that it works in the normal tree building case. I also tested that it works the same without the option set.

danpovey:
Sorry, I merged this too soon and had to revert it.  After merging I saw a problem.
The original code is using ClusterEventMapRestrictedByMap().  The purpose of this is to ensure that only tree-leaves that were originally on the same line of roots.txt can be merged together.
I believe there is an important reason for this, and there is a situation where if, for instance, silence and non-silence phones are merged together, something bad happens.  This might be in some rarely-used scenario such as (model-space) MLLR.  But I'm pretty sure I didn't write ClusterEventMapRestrictedByMap() for no reason.
Sorry if this makes your life harder... I'm sure there must be a way to modify ClusterEventMapRestrictedByMap() so it does what is needed.

PR #2104, opened 2017-12-26
Parallel lattice rescoring with TF RNNLMs + GRU example for TF RNNLM

hainan-xv:


src/fstext/deterministic-fst.h
```diff
@@ -84,10 +84,29 @@ class DeterministicOnDemandFst {
...
+template<class Arc>
+class DeterministicOnDemandFstParallel {
+ public:
+  typedef typename Arc::StateId StateId;
+  typedef typename Arc::Weight Weight;
+  typedef typename Arc::Label Label;
+
+  virtual StateId Start() = 0;
+
+  // dgao
```
danpovey: this comment shouldn't be here, and various aspects of this aren't right, both from a Google style guid point of view and a documentation point of view. I'm not going to get into details as this doesn't seem review-ready at this point.f

hainan-xv: Oh right. We will spend some time carefully going over all the code. I will send you an explicit "ready-to-review" message when I feel it's ready.

hainan-xv:
closing before I'm planning to add significant amount of new code before re-opening the PR.

PR #2106, opened 2017-12-27
Round number of leaves to multiple of 8

vimalmanohar:
Fixed issue in #2097

### DIFF
--- a/egs/rm/s5/run.sh
+++ b/egs/rm/s5/run.sh
@@ -4,6 +4,7 @@
 set -e # exit on error
 
 
+false && {
 # call the next line with the directory where the RM data is
 # (the argument below is just an example).  This should contain
 # subdirectories named as follows:
@@ -71,6 +72,7 @@ local/test_decoders.sh # This is a test program that we run only in the
                        # RM setup, it does some comparison tests on decoders
                        # to help validate the code.
 #draw-tree data/lang/phones.txt exp/tri1/tree | dot -Tps -Gsize=8,10.5 | ps2pdf - tree.pdf
+}
 
 # align tri1
 steps/align_si.sh --nj 8 --cmd "$train_cmd" \
@@ -133,6 +135,8 @@ steps/train_sat.sh 1800 9000 data/train data/lang exp/tri2b_ali exp/tri3b
 utils/mkgraph.sh data/lang exp/tri3b exp/tri3b/graph
 steps/decode_fmllr.sh --config conf/decode.config --nj 20 --cmd "$decode_cmd" \
   exp/tri3b/graph data/test exp/tri3b/decode
+
+exit 0
 
 (
  utils/mkgraph.sh data/lang_ug exp/tri3b exp/tri3b/graph_ug
--- a/egs/rm/s5/run.sh
+++ b/egs/rm/s5/run.sh
@@ -4,7 +4,6 @@
 set -e # exit on error
 
 
-false && {
 # call the next line with the directory where the RM data is
 # (the argument below is just an example).  This should contain
 # subdirectories named as follows:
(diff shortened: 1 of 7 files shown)


danpovey:
Thanks a lot... have you run any recipes with this and checked that the diagnostics like tree-impr are similar to before?

vimalmanohar:
I am checking this on RM. The WER results seem similar to before in the RESULTS file. I will check the logs to see if they are similar.

vimalmanohar:
The objective change in build-tree looks similar in new and old codes.

danpovey:
Fail on test:
```
  -553.239 -1260.74 1106.56 472.166 86.5589 226.666 6.26406 -561.831 -272.928 -434.097 -34.3362 133.59 -134.117 37.8137 -195.921 50.3036 159.485 343.228 -739.929 389.961 -186.367 290.25 -260.87 208.242 139.037 223.33WARNING ([5.3]:InitRand():build-tree-questions.cc:94) Questions::InitRand(), using union of all keys.  This may work depending on how you are building the tree but may crash (depends if you have already ensured that stats currently on the same leaf all share the same set of keys.)
LOG ([5.3]:BuildTree():build-tree.cc:161) BuildTree: before building trees, map has 2 leaves.
LOG ([5.3]:SplitDecisionTree():build-tree-utils.cc:577) DoDecisionTreeSplit: split 2 times, #leaves now 4
LOG ([5.3]:BuildTree():build-tree.cc:161) BuildTree: before building trees, map has 2 leaves.
LOG ([5.3]:SplitDecisionTree():build-tree-utils.cc:577) DoDecisionTreeSplit: split 2 times, #leaves now 4
WARNING ([5.3]:ClusterEventMapToNClustersRestrictedByMap():build-tree-utils.cc:873) num-clusters is less than size of map. Not doing anything.
LOG ([5.3]:BuildTree():build-tree.cc:273) BuildTree: Rounded num leaves to multiple of 8 by removing 26400 leaves.
ASSERTION_FAILED ([5.3]:BuildTree():build-tree.cc:276) : 'num_actually_removed < 8' 
[ Stack-Trace: ]
kaldi::MessageLogger::HandleMessage(kaldi::LogMessageEnvelope const&, char const*)
kaldi::MessageLogger::~MessageLogger()
kaldi::KaldiAssertFailure_(char const*, char const*, int, char const*)
kaldi::BuildTree(kaldi::Questions&, std::vector<std::vector<int, std::allocator<int> >, std::allocator<std::vector<int, std::allocator<int> > > > const&, std::vector<int, std::allocator<int> > const&, std::vector<bool, std::allocator<bool> > const&, std::vector<bool, std::allocator<bool> > const&, std::vector<std::pair<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > >, kaldi::Clusterable*>, std::allocator<std::pair<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > >, kaldi::Clusterable*> > > const&, float, int, float, int, bool)
kaldi::TestBuildTree()
main
__libc_start_main
./build-tree-test() [0x40d81e]
```

vimalmanohar:
I fixed the bug.
The test succeeds now.

egs/rm/s5/run.sh
```diff
@@ -4,6 +4,7 @@
 set -e # exit on error
 
 
+false && {
```
danpovey: temp?

vimalmanohar:
This is ready for review.

src/tree/build-tree-test.cc
```diff
@@ -88,7 +88,7 @@ void TestGenRandStats() {
 
 
 void TestBuildTree() {
-  for (int32 p = 0; p < 3; p++) {
+  for (int32 p = 0; p < 30; p++) {
```
danpovey: please make sure this doesn't take too long (e.g. more than 0.1 second).. I may have set it to a small number because it was slow.

PR #2108, opened 2017-12-28
update the best rnnlm setup (trained on swbd+fisher) with hesitation …

xiaohui-zhang:
…mapping,.etc


egs/swbd/s5c/local/rnnlm/tuning/run_tdnn_lstm_1e.sh
```diff
@@ -45,7 +51,14 @@ if [ $stage -le 0 ]; then
...
   # hold out one in every 50 lines as dev data.
   cat $text | cut -d ' ' -f2- | awk -v text_dir=$text_dir '{if(NR%50 == 0) { print >text_dir"/dev.txt"; } else {print;}}' >$text_dir/swbd.txt
-  zcat $fisher_text > $text_dir/fisher.txt
+  cat > $dir/config/hesitation_mapping.txt <<EOF
+hmm hum
+mmm um
+mm um
+mhm um-hum 
+EOF
+  zcat $fisher_text | awk 'NR==FNR{a[$1]=$2;next}{for (n=1;n<=NF;n++) if ($n in a) $n=a[$n];print $0}' \
```
danpovey: I prefer gunzip -c to zcat, it's used more often.

vince62s:
out of curiosity, did you actually get better WER ?
For Tedlium, you guys committed the same network (3 Tdnn + 2 lstm) with 800 + 200 + 200.
We tried to increase the same as this PR 1024 + 256 + 256, leading to a better PPL, BUT when rescoring it led to worse WER.

danpovey:
Just FYI, when tuning these things if the differences are small I tend to trust the perplexities more than the WERs; they are less noisy.

PR #2118, opened 2018-01-01
Add swbd-only RNNLM setup

keli78:


egs/swbd/s5c/local/rnnlm/run_tdnn_lstm_swbd.sh
```diff
@@ -4,29 +4,28 @@
 #           2017  Hainan Xu
```
danpovey: Hm.  I thought we were going to have a tuning/ directory in these setups.
Also, I think there should be a note at the top of the _swbd.sh version of the script explaining that this script trains on only the Switchboard transcripts, not Fisher data.

keli78: Ok. Since we only have one version I have thought it should be put in local/rnnlm. I will move it to the tuning/ directory if necessary.

PR #2122, opened 2018-01-03
Make a more detailed description for tdnn_lstm_1a in Fisher_SWBD

GaofengCheng:


danpovey:
@GaofengCheng, for future experiments I think it would make sense to make chunk_size a comma-separated list like it is for other setups.

GaofengCheng:
@danpovey Sorry, for not giving quick answer. OK, I will rerun the baseline with comma-style chunk size.

PR #2124, opened 2018-01-04
update configure and android_openblas.mk for the latest ndk in android.

lightslife:
Support for 64bit ARMv8(AArch64) architecture in Android, and update the android_openblas.mk  as  jcsilva put it.


danpovey:
Is there a reason you closed this?

lightslife:
The PR  can compile in my environment, and it's working in Android. But PR#2125 build failed in Travis CI, and I don't know how to fix the link error.

danpovey:
That was just a problem with Travis.  @galv, do you think his changes here were reasonable?
I'm concerned about the -lm_hard option, I don't know the implications of that.

lightslife:
The link problem is solved.  "-lm_hard" option means it linked with libm_hard.a. This is the same math library as libm.a, except that it follows hard-float ABI conventions. in the latest ndk, we use the soft-float calling conventions.

danpovey:
I'm not comfortable to merge this because you are changing some options and I think it might break the older setups.  I'd like to hear from e.g. @jcsilva or @galv about whether I should merge this.

ognjentodic:
FWIW, I can confirm the comment about -lm_hard being obsoleted (a while back actually, not just the latest NDK). I'll try to test the change, but probably won't have time until later next week.

jcsilva:
Those changes in android_openblas.mk seem fine. The other, about supporting AArch64, seems to be ok too, but I've never touched this architecture, so I'm not certain to guarantee it will work. 

Hard float ABI is no more supported by Android NDK (see [this](https://android.googlesource.com/platform/ndk/+/master/docs/HardFloatAbi.md)). And for compiling Kaldi with the most recent Android NDK version, you **must** use a recent version of OpenBLAS (release [0.2.20] (https://github.com/xianyi/OpenBLAS/releases/tag/v0.2.20), when softfp ABI was officially supported or newer).

galv:
I'm a little late to the game here, but I agree that Android needs -lm instead of -lm_hard. And if Ogi says it's fine, I'm pretty sure it's fine :D

Issue #2123, opened 2018-01-04
Kaldi C++ interface is leaky.


galv:
I've noticed some issues when compiling a C++ library against Kaldi.

1. f2c.h, which is included via matrix/kaldi-blas.h, leaks various symbols into the global namespace. This was a big pain today, as f2c.h defines the symbol `real` to be `float`, which caused compilation errors when I included kaldi headers before Eigen headers (Eigen uses the symbol real as well).

2. To compile my separate library I must add `-DHAVE_CLAPACK` to the compiler command line and `${kaldi_PREFIX}/kaldi/tools/ATLAS_headers/include/` and `${kaldi_PREFIX}/kaldi/tools/CLAPACK/` to my include paths, since kaldi's header files depend recursively on those include paths being present, which is leaky, as it demands a library user know to include kaldi-specific dependencies. Note that this works only for ATLAS. It won't work with OpenBLAS, which is another reason why this workaround is bad.

From what I can tell, both of these can be fixed by removing the `#include matrix/kaldi-blas.h` from `matrix/matrix-common.h`, since this is the only header file that includes `kaldi-blas.h`. This means that kaldi-blas.h would have to be included in various .cc files afterwards, but I think that this is fine because it makes the library much cleaner to interface with.

danpovey:
I think what you propose would be a good idea, but it would require a
little bit of work, and testing with different matrix libraries.
The following lines in matrix-common.h would be a problem:
typedef enum {
  kTrans    = CblasTrans,
  kNoTrans = CblasNoTrans
} MatrixTransposeType;
but we could just figure out what the numeric values are, put them in the
header, and use KALDI_STATIC_ASSERT in a c file that they are the same.

galv:
Hmmm... I know you can declare enums without defining them, but it's of course not clear how you would access the enum values without defining the enum right in the .h file. I'll look into it.

This is something I'd like to look at maybe next week, after some deadlines have passed for me.

PR #2127, opened 2018-01-06
issue 2123: eliminating global symbols leak (like 'float' etc)

bob177:
https://github.com/kaldi-asr/kaldi/issues/2123
Compile time assert could be placed in some test file like matrix/matrix-lib-test.cc, but it seems not compiled by default

### DIFF
--- a/src/matrix/cblas-wrappers.h
+++ b/src/matrix/cblas-wrappers.h
@@ -26,6 +26,7 @@
 #include "matrix/kaldi-vector.h"
 #include "matrix/kaldi-matrix.h"
 #include "matrix/matrix-functions.h"
+#include "matrix/kaldi-blas.h"
 
 // Do not include this file directly.  It is to be included
 // by .cc files in this directory.
@@ -335,6 +336,9 @@ inline void cblas_Xsbmv1(
               1, x, 1, beta, y, 1);
 }
 
+inline void cblas_compile_time_assert (){
+	  KALDI_COMPILE_TIME_ASSERT (kNoTrans == CblasNoTrans && kTrans == CblasTrans);
+}
 
 /// This is not really a wrapper for CBLAS as CBLAS does not have this; in future we could
 /// extend this somehow.
--- a/src/matrix/cblas-wrappers.h
+++ b/src/matrix/cblas-wrappers.h
@@ -336,10 +336,6 @@ inline void cblas_Xsbmv1(
               1, x, 1, beta, y, 1);
 }
 
-inline void cblas_compile_time_assert (){
-	  KALDI_COMPILE_TIME_ASSERT (kNoTrans == CblasNoTrans && kTrans == CblasTrans);
-}
-
 /// This is not really a wrapper for CBLAS as CBLAS does not have this; in future we could
 /// extend this somehow.
 inline void mul_elements(
--- a/src/matrix/matrix-common.h
+++ b/src/matrix/matrix-common.h
@@ -26,11 +26,12 @@
 #include "base/kaldi-common.h"
 
 namespace kaldi {
-// this enums equal to CblasTrans and SblasNoTrans constants from CBLAS library
-// we are writing them as numbers because we didn't include that library headers here  
+// this enums equal to CblasTrans and CblasNoTrans constants from CBLAS library
+// we are writing them as literals because we don't want to include here matrix/kaldi-blas.h,
+// which puts many symbols into global scope (like "real") via the header f2c.h 
 typedef enum {
-  kTrans    = 112,
-  kNoTrans  = 111
+  kTrans    = 112, // = CblasTrans
+  kNoTrans  = 111  // = CblasNoTrans
 } MatrixTransposeType;
 
 typedef enum {
--- a/src/matrix/matrix-common.h
+++ b/src/matrix/matrix-common.h
@@ -24,12 +24,11 @@
 // files in this directory.
 
 #include "base/kaldi-common.h"
-#include "matrix/kaldi-blas.h"
 
 namespace kaldi {
 typedef enum {
-  kTrans    = CblasTrans,
-  kNoTrans = CblasNoTrans
+  kTrans    = 112,
+  kNoTrans  = 111
 } MatrixTransposeType;
 
 typedef enum {
--- a/src/matrix/matrix-common.h
+++ b/src/matrix/matrix-common.h
@@ -26,6 +26,8 @@
 #include "base/kaldi-common.h"
 
 namespace kaldi {
+// this enums equal to CblasTrans and SblasNoTrans constants from CBLAS library
+// we are writing them as numbers because we didn't include that library headers here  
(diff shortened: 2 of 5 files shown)


src/matrix/matrix-common.h
```diff
@@ -24,12 +24,11 @@
 // files in this directory.
 
 #include "base/kaldi-common.h"
-#include "matrix/kaldi-blas.h"
 
 namespace kaldi {
 typedef enum {
```
danpovey: there needs to be some comment to the effect that these are the same as CblasTrans and CblasNoTranss from the CBLAS library, but we are writing them as numbers because we don't include that library in these headers.

src/matrix/cblas-wrappers.h
```diff
@@ -335,6 +336,9 @@ inline void cblas_Xsbmv1(
               1, x, 1, beta, y, 1);
 }
 
+inline void cblas_compile_time_assert (){
+	  KALDI_COMPILE_TIME_ASSERT (kNoTrans == CblasNoTrans && kTrans == CblasTrans);
```
danpovey: I'd rather you put this in a test program (e.g. matrix-lib-test.cc) to avoid polluting the header.

galv: Since we have C++11 now, you can simple do `static_assert(kNoTrans == CblasNoTrans && kTrans == CblasTrans, "informative error message")` in global scope in one of the .cc files. Probably easier to understand for the average C++ developer because it's part of the language.

danpovey: makes sense.  Our macro should be removed and replaced everywhere with that.

src/matrix/matrix-common.h
bob177: done

src/matrix/cblas-wrappers.h
bob177: put static_assert to the matrix-lib-test.cc. But we need make sure that numbers don't change from version to version and from matrix lib to another matrix lib.
from matrix-lib.h I simply removed include of cblas-wrappers.h - all compiles correctly, no need to add extra include to *.cc, header jama-svd.h also includes cblas-wrappers.h, but if it is not used outside it can not leaks symbols

danpovey:
Thanks.  @galv, do you have any time to check this?
@bob177, there are a few options for which BLAS library to use that are available when compiling Kaldi.  Can you please try to verify that it still compiles with some of the other options than ATLAS?  e.g. OpenBLAS at least-- MKL if possible as well would be great.

src/matrix/matrix-common.h
bob177: 1) CLAPACK - well compiling, but when linking there is an error: undefined reference to 'cblas_*'. I think error in src/configure, because no linking blas library is specified, for instance when linking align-egual. And I suppose CBLAS should be istalled independtly.
2) OpenBLAS - well compiling and linking. (But there is difficulties, which I fixed) There are many many warnings 'linker input file unused because linking not done'. Matrix-lib-text is passed. But we need include cblas-wrappers.h into matrix-lib-test.cc.
3) mkl - tomorrow

danpovey:
I think it's been a while since we linked with CLAPACK.  Probably the
errors you're seeing are independent of the changes you made, you may not
have installed it.

src/matrix/matrix-common.h
```diff
@@ -24,12 +24,13 @@
 // files in this directory.
 
 #include "base/kaldi-common.h"
-#include "matrix/kaldi-blas.h"
 
 namespace kaldi {
+// this enums equal to CblasTrans and SblasNoTrans constants from CBLAS library
+// we are writing them as numbers because we didn't include that library headers here
```
galv: I'd rather say this: 'we are writing them as literals because we don't want to include "matrix/kaldi-blas.h", which puts many symbols things into global scope like "real" via the header "f2c.h". '

State why you the code does what it does, not what it does.

src/matrix/matrix-common.h
```diff
@@ -24,12 +24,13 @@
...
 #include "base/kaldi-common.h"
-#include "matrix/kaldi-blas.h"
 
 namespace kaldi {
+// this enums equal to CblasTrans and SblasNoTrans constants from CBLAS library
+// we are writing them as numbers because we didn't include that library headers here  
 typedef enum {
-  kTrans    = CblasTrans,
-  kNoTrans = CblasNoTrans
+  kTrans    = 112,
```
galv: I would add comments like `// = CblasTrans` and `// = CblasNoTrans` to clarify.

src/matrix/matrix-lib-test.cc
```diff
@@ -4740,10 +4740,11 @@ template<typename Real> static void MatrixUnitTest(bool full_test) {
 
 int main() {
   using namespace kaldi;
+  static_assert(int(kNoTrans) != int(CblasNoTrans) && int(kTrans) == int(CblasTrans),
```
galv: Don't put this here. I recommend putting it into one of the non-test .cc files. By putting it into the test files, you are preventing the static_assert from being run when someone just types "make" instead of "make test".  There's no good "general" .cc file I can think of, so I recommend putting it at the top of kaldi-matrix.cc.

Also, I'm not positive, but can you do `kNoTrans == CblasNoTrans` here? Casting to int should be fine since the literal values will fit into an int, but it's actually not defined in C/C++ what kind of int backs an enum, so you should prefer to avoid the cast if possible.

Also, you can put the static_assert into global scope. It does not need to sit inside a function.

src/matrix/cblas-wrappers.h
```diff
@@ -26,6 +26,7 @@
 #include "matrix/kaldi-vector.h"
 #include "matrix/kaldi-matrix.h"
 #include "matrix/matrix-functions.h"
+#include "matrix/kaldi-blas.h"
```
galv: I just want to note that you need to be careful to make sure that none of the files #include'd by matrix-lib.h #include cblas-wrappers.h recursively, since it exposes the bad symbols via kaldi-blas.h. I don't think that any of them do from having spent some time just now grepping through the source code.

It would in general be preferable to #include kaldi-blas.h only in the .cc files, but we would have to factor out cblas-wrappers.h into a separate cblas-wrappers.cc implementation file, so I think this is good enough.

galv:
It's important to check that OpenBLAS, ATLAS, Intel MKL, and Mac Accelerate don't break with this change.

I will check the mac build and report back on it.

danpovey:
Daniel is right about it being better to put the assert in kaldi-matrix.cc.

galv:
Mac build works. Specifically, `make -j && make -j test` works

src/matrix/cblas-wrappers.h
bob177: Compiled well with Intel MKL, matrix test passed. If you think that warning is ok, I would remove casting to int in static_assert.

PR #2137, opened 2018-01-09
multi_en: Fixed acronym normalization, swbd lexicon preparation, OOV …

xiaohui-zhang:
…pronunciation generation, acoustic data sub-sampling,.etc; Added hub4_97 data. @vimalmanohar can you please review it? Thanks!


egs/multi_en/s5/run.sh
```diff
@@ -137,19 +140,22 @@ if [ $stage -le 7 ]; then
...
-      $data exp/make_mfcc/$c/train || exit 1;
-    steps/compute_cmvn_stats.sh \
-      $data exp/make_mfcc/$c/train || exit 1;
+    (
+     data=data/$c/train
+     steps/make_mfcc.sh --mfcc-config conf/mfcc.conf \
+       --cmd "$train_cmd" --nj 40 \
+       $data exp/make_mfcc/$c/train || exit 1;
+     steps/compute_cmvn_stats.sh \
+       $data exp/make_mfcc/$c/train || exit 1;
```
vimalmanohar: Exiting here is not very useful. Better to do `|| touch $dir/.error` and check for that later.

egs/multi_en/s5/run.sh
```diff
@@ -98,17 +103,15 @@ if [ $stage -le 3 ]; then
...
 
-# Prepare the dictionary and train G2P model using the combined (CMUDict+Tedlium+swbd) lexicon
-# in data/local/dict_combined, and then synthesize pronounciations for all OOV words 
-# across all training transcripts.
+# Synthesize pronounciations for OOV words across all training transcripts and produce the final lexicon.
 if [ $stage -le 4 ]; then
-  # We prepare the dictionary in data/local/dict_combined.
-  local/prepare_dict.sh $swbd $tedlium2
-  local/g2p/train_g2p.sh --stage 0 --silence-phones "data/local/dict_combined/silence_phones.txt" data/local/dict_combined exp/g2p
+  wait
```
vimalmanohar: Comment that you are waiting for train_g2p.sh to finish.

vimalmanohar: Also you should check for errors if it has failed.

xiaohui-zhang:
fixed

egs/fisher_swbd/s5/path.sh
```diff
@@ -1,4 +1,4 @@
-export KALDI_ROOT=`pwd`/../../..
+export KALDI_ROOT=../../kaldi
```
vimalmanohar: Don't change the standard, which is the same in all recipes. Also fix this in multi-en and elsewhere you changed.

egs/fisher_swbd/s5/run.sh
```diff
@@ -14,7 +14,7 @@ if [ $? -ne 0 ]; then
 fi
 
 # prepare fisher data and put it under data/train_fisher
-local/fisher_data_prep.sh /export/corpora3/LDC/LDC2004T19 /export/corpora3/LDC/LDC2005T19 \
+ local/fisher_data_prep.sh /export/corpora3/LDC/LDC2004T19 /export/corpora3/LDC/LDC2005T19 \
```
vimalmanohar: Fix indentation.

egs/mini_librispeech/s5/run.sh
```diff
@@ -80,7 +80,7 @@ if [ $stage -le 3 ]; then
   steps/align_si.sh --boost-silence 1.25 --nj 5 --cmd "$train_cmd" \
     data/train_clean_5 data/lang_nosp exp/mono exp/mono_ali_train_clean_5
 fi
-
```
vimalmanohar: This is not meant to be changed.

egs/multi_en/s5/path.sh
```diff
@@ -1,7 +1,8 @@
-export KALDI_ROOT=`pwd`/../../..
+export KALDI_ROOT=../../kaldi
```
vimalmanohar: Change this.

egs/swbd/s5c/path.sh
```diff
@@ -1,4 +1,4 @@
-export KALDI_ROOT=`pwd`/../../..
+export KALDI_ROOT=/export/b04/xzhang/kaldi
```
vimalmanohar: Fix this

egs/fisher_swbd/s5/local/chain/run_tdnn_lstm_1a.sh
```diff
@@ -107,7 +108,7 @@ lang=data/lang_chain
 local/nnet3/run_ivector_common.sh --stage $stage \
   --speed-perturb $speed_perturb \
   --generate-alignments $speed_perturb || exit 1;
-
+exit
```
vimalmanohar: Remove this.

egs/swbd/s5c/local/chain/tuning/run_tdnn_lstm_1l.sh
```diff
@@ -20,8 +20,8 @@
 set -e
 
 # configs for 'chain'
-stage=12
-train_stage=-10
+stage=15
```
vimalmanohar: Undo this change.

egs/wsj/s5/utils/mkgraph.sh
```diff
@@ -17,7 +17,7 @@ set -o pipefail
 
 tscale=1.0
 loopscale=0.1
-
+unkscale=1.0
```
vimalmanohar: This does not seem to be used.

xiaohui-zhang: finished cleaning up

egs/multi_en/s5/run.sh
```diff
@@ -137,19 +140,22 @@ if [ $stage -le 7 ]; then
...
+    ) &
   done
+  wait
 fi
 
 # fix and validate training data directories
 if [ $stage -le 8 ]; then
   # get rid of spk2gender files because not all corpora have them
-  rm -f data/*/train/spk2gender
+  rm data/*/train/spk2gender 2>/dev/null
```
vimalmanohar: There should be -f. Otherwise the script might exit because of set -e.

egs/multi_en/s5/run.sh
```diff
@@ -98,17 +103,15 @@ if [ $stage -le 3 ]; then
...
-  local/prepare_dict.sh $swbd $tedlium2
-  local/g2p/train_g2p.sh --stage 0 --silence-phones "data/local/dict_combined/silence_phones.txt" data/local/dict_combined exp/g2p
+  wait # Waiting for train_g2p.sh to finish
   dict_dir=data/local/dict_nosp
   mkdir -p $dict_dir
+  rm $dict_dir/lexiconp.txt 2>/dev/null
   cp data/local/dict_combined/{extra_questions,nonsilence_phones,silence_phones,optional_silence}.txt $dict_dir
-  local/g2p/apply_g2p.sh --var-counts 1 exp/g2p/model.fst data/local/g2p_phonetisarus data/local/dict_combined/lexicon.txt $dict_dir/lexicon.txt
+  local/g2p/apply_g2p.sh --var-counts 1 exp/g2p/model.fst data/local/g2p_phonetisarus \
+    data/local/dict_combined/lexicon.txt $dict_dir/lexicon.txt || touch $dict_dir/.error
```
vimalmanohar: This does seem to be running in the background. So better to use `|| exit`

egs/multi_en/s5/run.sh
```diff
@@ -137,19 +140,22 @@ if [ $stage -le 7 ]; then
...
-    steps/make_mfcc.sh --mfcc-config conf/mfcc.conf \
-      --cmd "$train_cmd" --nj 40 \
-      $data exp/make_mfcc/$c/train || exit 1;
-    steps/compute_cmvn_stats.sh \
-      $data exp/make_mfcc/$c/train || exit 1;
+    (
+     data=data/$c/train
+     steps/make_mfcc.sh --mfcc-config conf/mfcc.conf \
+       --cmd "$train_cmd" --nj 40 \
+       $data exp/make_mfcc/$c/train || touch $data/.error
```
vimalmanohar: Are you checking if there is an error in a later stage? Also you should first remove the `rm -f` the `.error` files.

egs/multi_en/s5/run.sh
```diff
@@ -71,20 +83,13 @@ if [ $stage -le 1 ]; then
   local/wsj_data_prep.sh $wsj0/??-{?,??}.? $wsj1/??-{?,??}.?
   local/wsj_format_data.sh
   utils/copy_data_dir.sh --spk_prefix wsj_ --utt_prefix wsj_ data/wsj/train_si284 data/wsj/train
-  rm -rf data/wsj/train_si284
+  rm -r data/wsj/train_si284 2>/dev/null
```
vimalmanohar: Without `-f` the script will exit if error since you gave `set -e`

egs/multi_en/s5/run.sh
danpovey: I don't like using -f to make 'rm' be quiet, because it overrides when people change modes to save something.  Instead, do `|| true`

egs/multi_en/s5/run.sh
xiaohui-zhang: hmm I actually made it back to || exit 1; since I think we should quit if the feature extraction fails

vimalmanohar: `|| exit 1` will have no effect inside a subshell running in background.

xiaohui-zhang: fixed

egs/multi_en/s5/RESULTS
```diff
@@ -36,21 +36,24 @@ exit 0
...
-multi_a  tri5a  tg_eval2000        ||  %WER 24.3 | 4459 42989 | 78.8 15.6 5.6 3.2 24.3 63.3 | exp/multi_a/tri5a/decode_tg_eval2000/score_13_0.0/eval2000.ctm.filt.sys
-multi_a  tri5a  tg_eval2000.si     ||  %WER 30.6 | 4459 42989 | 73.7 20.2 6.1 4.3 30.6 67.9 | exp/multi_a/tri5a/decode_tg_eval2000.si/score_10_1.0/eval2000.ctm.filt.sys
-multi_a  tri5a  tg_sp_eval2000     ||  %WER 24.2 | 4459 42989 | 79.1 15.6 5.3 3.3 24.2 63.2 | exp/multi_a/tri5a/decode_tg_sp_eval2000/score_12_0.0/eval2000.ctm.filt.sys
-multi_a  tri5a  tg_sp_eval2000.si  ||  %WER 30.5 | 4459 42989 | 73.7 20.3 6.0 4.2 30.5 67.8 | exp/multi_a/tri5a/decode_tg_sp_eval2000.si/score_10_1.0/eval2000.ctm.filt.sys
-multi_a  tri5b  tg_eval2000        ||  %WER 24.3 | 4459 42989 | 79.3 15.7 5.0 3.6 24.3 63.5 | exp/multi_a/tri5b/decode_tg_eval2000/score_11_0.0/eval2000.ctm.filt.sys
-multi_a  tri5b  tg_eval2000.si     ||  %WER 30.7 | 4459 42989 | 73.6 20.4 6.0 4.3 30.7 68.1 | exp/multi_a/tri5b/decode_tg_eval2000.si/score_10_1.0/eval2000.ctm.filt.sys
+# On eval2000 the final GMM results is 24.5, which is better than the above result (24.9).
```
danpovey: I'm pretty sure this comment doesn't belong here.

danpovey: Also the results seems lightly worse than before.  Is there an explanation for this?  Is it the case that for other test sets it was better?
What exactly is better about this new setup than how it was before?

xiaohui-zhang: I fixed some bugs of acronym normalization in swbd training data (previously these acronyms were wrongly normalized as OOVs), and I fixed the way we apply G2P to OOVs in training data, i.e. we won't apply G2P to words containing special characters. But these changes are too tiny to affect WERs. The major change which may affect WER is we now have 72h more training data (hub4 97). Since there are mismatch between hub4's bn style and eval2000's style, there could be some degradation at GMM level (actually looking tri5a.si and tri5b.si it's slightly better than before). I think this could be compensated at chain model level.  btw I intentionally made that comment there...

xiaohui-zhang:
ready to merge I think

PR #2138, opened 2018-01-09
Swbd stats vad recipe

GoVivace:
Stats VAD recipe for Switchboard, with some bugfixes.

### DIFF
--- a/egs/swbd/s5c/local/run_asr_segmentation.sh
+++ b/egs/swbd/s5c/local/run_asr_segmentation.sh
@@ -36,7 +36,7 @@ prepare_targets_stage=-10
 nstage=-10
 train_stage=-10
 test_stage=-10
-num_data_reps=1
+num_data_reps=2
 base_rirs=simulated
 affix=_1a
 stage=-1
@@ -164,7 +164,7 @@ fi
 
 if [ $stage -le 5 ]; then
   utils/copy_data_dir.sh ${rvb_data_dir} ${rvb_data_dir}_hires
-  steps/make_mfcc.sh --mfcc-config conf/mfcc_hires.conf --nj 40 \
+  steps/make_mfcc.sh --mfcc-config conf/mfcc_hires.conf --nj 10 \
     ${rvb_data_dir}_hires
   steps/compute_cmvn_stats.sh ${rvb_data_dir}_hires
 fi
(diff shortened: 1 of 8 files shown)


danpovey:
@david-ryan-snyder, is there someone working on things related to this who can try this for us and make sure it runs cleanly?

david-ryan-snyder:
Possibly @mmaciej2. After we merge the speaker segmentation stuff, he planned on adding support to use Vimal's SAD instead of oracle segmentation in the recipe. I believe he's already worked with Vimal's SAD to some extent.

danpovey:
OK, cool.  @mmaciej2, could you please do this-- fairly soon, like before the weekend?  It should take very little time, just shepherding the jobs.  And if you notice anything confusing when doing so, let me know.

mmaciej2:
Sure, @danpovey, I will try to do this as soon as possible.

egs/swbd/s5c/local/segmentation/copy_targets_dir.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
+# Copyright    2017  Nagendra Kumar Goel
```
vimalmanohar: Can you move this file to steps/segmentation? I think it will be useful in other cases.

egs/swbd/s5c/local/run_asr_segmentation.sh
```diff
@@ -40,8 +42,8 @@ affix=_1a
 stage=-1
 nj=80
 
-. ./path.sh
-. ./cmd.sh
+. path.sh
```
vimalmanohar: The standard now is to use `. ./cmd.sh` and `if [ -f ./path.sh ]; then . ./path.sh; fi`
Better to change it to this in all files.

egs/swbd/s5c/local/segmentation/copy_targets_dir.sh
vimalmanohar: Since there is no change to the contents of the targets file, it might be better to copy the scp file and add the suffix to the utterance-ids.

egs/swbd/s5c/local/run_asr_segmentation.sh
```diff
@@ -1,15 +1,17 @@
...
-# Copyright 2017  Vimal Manohar
+# Copyright  2017  Nagendra Kumar Goel
+#            2017  Vimal Manohar
 # Apache 2.0
 
-# Features configs (Must match the features used to train the models
-# $sat_model_dir and $model_dir)
+# We assume the run-1-main.sh (because we are using model directories like
+# exp/tri4) and later we assumme run-4-anydecode.sh was run to prepare
+# data/dev10h.pem
```
mmaciej2: This comment looks like it is for a babel recipe, not switchboard, unless I am mistaken.

egs/swbd/s5c/local/segmentation/copy_targets_dir.sh
GoVivace: @vimalmanohar  I had some other functionality in mind - like creating multiple reps of data for noise, which I apparently didn't finish although the script should work in its current form. How about I move it after I finish that?

egs/swbd/s5c/local/run_asr_segmentation.sh
GoVivace: fixed and pushed

GoVivace:
Please review this latest version.

egs/swbd/s5c/local/segmentation/tuning/train_lstm_asr_sad_1a.sh
```diff
@@ -1,12 +1,15 @@
...
+# Apache 2.0
+
 # This is a script to train a TDNN-LSTM for speech activity detection (SAD) 
 # using LSTM for long-context information.
 
 set -o pipefail
 set -u
 
-. ./cmd.sh
+. cmd.sh
```
danpovey: this should be ./cmd.sh

egs/swbd/s5c/local/segmentation/copy_targets_dir.sh
vimalmanohar: Even then it would be better to copy the scp file and change the utterance-id. Also I think most scripts expect targets.scp, which is not being created by this script.

GoVivace:
I was going to ask you... Your final dnn training script expects a
target.scp in the root folder of the experiment. Is it ok to change that.
Both changes need to happen together.

GoVivace:
@vimalmanohar Currently copy-data-dir does not create the scp file, but then later combine-data-dir creates it. The two are expected to be used together. I am running the scripts once again to make sure no other inadvertent bugs were introduced. Fixed "cmd.sh" pointed out by @danpovey

danpovey:
Closing this PR since we're now planning to merge Vimal's modified version in #2193

PR #2153, opened 2018-01-15
[build] set 'rpath' for ATLAS shared libraries,

vesis84:
- in BUT we have installed both ATLAS (includes LAPACK 3.0) and
  standalone LAPACK 3.4.2
- the standalone LAPACK 3.4.2 had priorty over LAPACK from ATLAS,
  which was causing runtime errors
- we cannot remove standalone LAPACK, as other programs are using it
- Hence we propose to select the LAPACK from ATLAS by using 'rpath',
  which stores the library location in the binaries.
  (if on other systems $(ATLASLDFLAGS) remains undefined, it causes no harm)

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -800,6 +800,7 @@ function linux_configure_dynamic {
 
   echo ATLASINC = $ATLASROOT/include >> kaldi.mk
   echo ATLASLIBS = $ATLASLIBS >> kaldi.mk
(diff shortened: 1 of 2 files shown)


galv:
I see no harm in this. In fact, something like cmake would automatically set the rpath whenever you link a shared object anyway, so this is the more "correct" thing to do.

jtrmal:
I'm ok with this as well.
Alternatively, should be solvable by setting LD_LIBRARY_PATH appropriately.
y.

PR #2154, opened 2018-01-15
[egs] ami, changing path for storing downloaded wavs,

vesis84:


egs/ami/s5b/run.sh
```diff
@@ -28,7 +28,7 @@ set -euo pipefail
 # Path where AMI gets downloaded (or where locally available):
 AMI_DIR=$PWD/wav_db # Default,
 case $(hostname -d) in
-  fit.vutbr.cz) AMI_DIR=/mnt/scratch05/iveselyk/KALDI_AMI_WAV ;; # BUT,
+  fit.vutbr.cz) AMI_DIR=/mnt/scratch05/tmp/$USER/KALDI_AMI_WAV ;; # BUT,
```
jtrmal: Hi Karel, this does not look particularly effective, w.r.t. sites that host the data, each user will download them (needlessly) again. Also, it will be fairly slow. Are you sure that on your side (FIT), you cannot store the corpus somewhere globally, available to all people (in similar way we do at CLSP)?

vesis84: Well, of course we could, cache it here globally... But then we have to download both the IHM and MDM wavs. Okay, I get your point, it will save the traffic of the server, as a courtesy...

PR #2155, opened 2018-01-15
Copy of lang directory fails on MACOSX

johnjosephmorgan:
In the utils/format_lm.sh file the data/lang/phones directory is copied to a new lang_test directory. This operation fails on MACOSX.


danpovey:
Thanks, I'll fix that soon.  I think you didn't mean to make this PR-- closing it.

johnjosephmorgan:
I was pretty sure it was a problem, but it was a mystery and I did not know how to fix it.
After doing a search, it looks like the source directory should drop the trailing slash.
cp -R lang/phones

not
cp -R phones/
J

danpovey:
Fixed in commit 1fb7fd92e

PR #2168, opened 2018-01-21
fst-type optimization for decoder & make fst reader to support future fst-types

chenzhehuai:
1. To reduce VTable cost in openfst based graph search, make ProcessEmitting() and ProcessNonemitting()  be templated on the Fst type, and create wrapper functions like the following: BaseFloat ProcessEmittingWrapper(DecodableInterface *decodable). In the .cc file, the ProcessEmittingWrapper() function will check the actual Fst type; if it's ConstFst or VectorFst, it will call ProcessEmitting() with one of those two types; otherwise it will call it with the regular Fst interface (to cover the other cases). 
Just do this for LatticeFasterDecoder and for LatticeFasterOnlineDecoder; as the others are rarely used.

test results are in /export/a12/zchen/export/fsttype-opt/README with 3-8% speed improvement in VectorFst & ConstFst

2. change the implementation of ReadFstKaldiGeneric to directly refer to openfst implementation to support future fst-types. e.g. olabel_lookahead fst


src/configure
```diff
@@ -1090,22 +1090,26 @@ fi
 echo "OPENFSTINC = $FSTROOT/include" >> kaldi.mk
 if $static_fst ; then
   OPENFSTLIBS="$FSTROOT/lib/libfst.a"
+  OPENFSTSCRIPTLIBS="$FSTROOT/lib/libfstscript.a"
 else
   if [ "`uname`" == "Darwin"  ]; then
     OPENFSTLIBS="$FSTROOT/lib/libfst.dylib"
     OPENFSTLDFLAGS="-Wl,-rpath -Wl,${FSTROOT}/lib"
+    OPENFSTSCRIPTLIBS="$FSTROOT/lib/libfstscript.dylib"
```
danpovey: oh.  I didn't realize that we have to link an extra library in order to use the FstScript stuff.
Doing that would require us to bump up the configure version, forcing everyone to re-run configure.
Since this change does not produce any measurable benefit for the user, let's revert that part of the PR.

src/decoder/lattice-faster-decoder.cc
```diff
@@ -683,6 +684,7 @@ BaseFloat LatticeFasterDecoder::GetCutoff(Elem *list_head, size_t *tok_count,
   }
 }
 
+template <typename Fst>
```
danpovey: Please call this FstType instead of Fst: otherwise the code is hard to read (confusion with fst::Fst).

src/decoder/lattice-faster-decoder.cc
```diff
@@ -715,7 +717,7 @@ BaseFloat LatticeFasterDecoder::ProcessEmitting(DecodableInterface *decodable) {
     StateId state = best_elem->key;
     Token *tok = best_elem->val;
     cost_offset = - tok->tot_cost;
-    for (fst::ArcIterator<fst::Fst<Arc> > aiter(fst_, state);
+    for (fst::ArcIterator<Fst > aiter(dynamic_cast<Fst &>(const_cast<fst::Fst<Arc> &>(fst_)), state);
```
danpovey: figure out a way to reduce this line length, I suggest declaring a reference variable as follows, at the top of this function:
`FstType &fst = dynamic_cast<FstType&>(const_cast<fst::Fst<Arc> &>(fst_));`
This will also save a lot of vtable lookups, since dynamic_cast is a nontrivial operation.

src/decoder/lattice-faster-online-decoder.h
```diff
@@ -337,12 +337,16 @@ class LatticeFasterOnlineDecoder {
 
   /// Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
   /// Returns the cost cutoff for subsequent ProcessNonemitting() to use.
-  BaseFloat ProcessEmitting(DecodableInterface *decodable);
+  template <typename Fst> BaseFloat ProcessEmitting(DecodableInterface *decodable);
+
+  BaseFloat ProcessEmittingWrapper(DecodableInterface *decodable);
```
danpovey: need some documentation here.

src/decoder/lattice-faster-online-decoder.h
```diff
@@ -337,12 +337,16 @@ class LatticeFasterOnlineDecoder {
 
   /// Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
   /// Returns the cost cutoff for subsequent ProcessNonemitting() to use.
-  BaseFloat ProcessEmitting(DecodableInterface *decodable);
+  template <typename Fst> BaseFloat ProcessEmitting(DecodableInterface *decodable);
```
danpovey: need some explanation about why it's templated.   And change Fst to FstType.

chenzhehuai:
@danpovey  done.

src/decoder/lattice-faster-decoder.cc
```diff
@@ -3,6 +3,7 @@
 // Copyright 2009-2012  Microsoft Corporation  Mirko Hannemann
 //           2013-2014  Johns Hopkins University (Author: Daniel Povey)
 //                2014  Guoguo Chen
+//                2018  SJTU Zhehuai Chen
```
danpovey: You probably should have just your name-- affiliation is only normally required for employees.

One other point: please be careful with the 80-character limit.  We try to respect it except in exceptional circumstances.

galv:
I am pretty curious about this change. How did you discover that vtable look ups were a bottleneck? 

When I looked into decoding myself, I ran valgrind's profiling mode on online2-nnet3-latgen-faster (results below), and found that overwhelmingly most instructions were executed doing matrix multiplication, even with MKL. I realize that instruction frequencies don't take into account things like memory latency, but I was always of the impression that a better avenue would be to, e.g., use 16-bit floating point.

Otherwise, very interesting, thanks for doing this.

```
--------------------------------------------------------------------------------
             Ir
--------------------------------------------------------------------------------
141,403,258,439  PROGRAM TOTALS

--------------------------------------------------------------------------------
            Ir  file:function
--------------------------------------------------------------------------------
83,559,869,118  ???:mkl_blas_avx2_sgemm_kernel_nocopy_TN_b1 [/opt/intel/compilers_and_libraries_2017.4.196/linux/mkl/lib/intel64_lin/libmkl_avx2.so]
 3,230,401,861  <path-to-kaldi>/src/decoder/lattice-faster-online-decoder.cc:kaldi::LatticeFasterOnlineDecoder::ProcessEmitting(kaldi::DecodableInterface*) [<path-to-kaldi>/src/online2bin/online2-wav-nne\
t3-latgen-faster]
 3,112,799,568  <path-to-kaldi>/tools/openfst/include/fst/fst.h:fst::ArcIterator<fst::Fst<fst::ArcTpl<fst::TropicalWeightTpl<float> > > >::Done() const [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-l\
atgen-faster]
 3,016,805,312  <path-to-kaldi>/tools/openfst/include/fst/fst.h:fst::ArcIterator<fst::Fst<fst::ArcTpl<fst::TropicalWeightTpl<float> > > >::Value() const [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-\
latgen-faster]
 3,016,805,312  <path-to-kaldi>/tools/openfst/include/fst/fst.h:fst::ArcIterator<fst::Fst<fst::ArcTpl<fst::TropicalWeightTpl<float> > > >::Next() [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-latgen-\
faster]
 2,716,377,118  <path-to-kaldi>/src/decoder/lattice-faster-online-decoder.cc:kaldi::LatticeFasterOnlineDecoder::ProcessNonemitting(float) [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-latgen-faster]
 2,059,711,448  /build/glibc-bfm8X4/glibc-2.23/malloc/malloc.c:_int_malloc [/lib/x86_64-linux-gnu/libc-2.23.so]
 1,887,359,520  <path-to-kaldi>/src/nnet3/decodable-online-looped.cc:kaldi::nnet3::DecodableAmNnetLoopedOnline::LogLikelihood(int, int) [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-latgen-faster]
 1,589,841,265  <path-to-kaldi>/src/matrix/kaldi-matrix.cc:kaldi::MatrixBase<float>::ApplyFloor(float) [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-latgen-faster]
 1,533,493,740  <path-to-kaldi>/src/nnet3/../nnet3/decodable-online-looped.h:kaldi::nnet3::DecodableNnetLoopedOnlineBase::EnsureFrameIsComputed(int) [<path-to-kaldi>/src/online2bin/online2-wav-nnet3-latg\
en-faster]
 1,532,433,177  <path-to-kaldi>/src/nnet3/../matrix/kaldi-matrix.h:kaldi::MatrixBase<float>::operator()(int, int)
```

chenzhehuai:
@danpovey Done.

danpovey:
Can you please re-test any speed change and report it here?  This will also answer @galv's query.

chenzhehuai:
@galv Yes, actually I didn't take NN propagation into account here but only concentrate on searching. 
 **details of test**
I firstly do nnet3 propagate and save the posteriors in a ark file. And then I compare the decoding RTF logs from different methods. The result shows 3-8% difference. Notably, it's already with g++ -O3 optimization.

src/decoder/lattice-faster-online-decoder.h
```diff
@@ -337,12 +338,23 @@ class LatticeFasterOnlineDecoder {
 
   /// Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
   /// Returns the cost cutoff for subsequent ProcessNonemitting() to use.
-  BaseFloat ProcessEmitting(DecodableInterface *decodable);
+  /// To reduce vtable lookup cost in openfst based graph search, 
+  //make ProcessEmitting() and ProcessNonemitting() be templated on the FstType,
```
danpovey: should be space between // and comment.

danpovey: ... also the comment is too long.  And keep the triple ///, it relates to Doxygen.
 I'd just add a single line:
/// Templated on Fst type for speed; called via ProcessEmittingWrapper().

src/decoder/lattice-faster-decoder.cc
```diff
@@ -837,6 +856,19 @@ void LatticeFasterDecoder::ProcessNonemitting(BaseFloat cutoff) {
   } // while queue not empty
 }
 
+template void LatticeFasterDecoder::ProcessNonemitting<fst::ConstFst<fst::StdArc>>(BaseFloat cutoff);
+template void LatticeFasterDecoder::ProcessNonemitting<fst::VectorFst<fst::StdArc>>(BaseFloat cutoff);
+template void LatticeFasterDecoder::ProcessNonemitting<fst::Fst<fst::StdArc>>(BaseFloat cutoff);
+
+void LatticeFasterDecoder::ProcessNonemittingWrapper(BaseFloat cost_cutoff) {
+  if (typeid(fst_) == typeid(fst::ConstFst<Arc>)) {
```
danpovey: We haven't used typeid elsewhere and I'm concerned there may be subtleties involved or portability problems.  I think it would be better to use fst_.FstType(), which might be "vector" or "const".
After you do that, re-run your test.  I'm surprised you didn't see a speedup for VectorFst.  You may have to run it a couple times to correct for effects of nfs caching and the like.

chenzhehuai:
@danpovey  Done.  Some new results in the same experiment setup

| decoder  |   graph   | RTF  |
|----------|-----------|------|
| baseline | vectorfst | 0.62 |
| opt.     | vectorfst | 0.58 |
| baseline | constfst  | 0.58 |
| opt.     | constfst  | 0.54 |

src/decoder/lattice-faster-online-decoder.h
```diff
@@ -335,14 +336,25 @@ class LatticeFasterOnlineDecoder {
...
 
-  /// Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
-  /// Returns the cost cutoff for subsequent ProcessNonemitting() to use.
-  BaseFloat ProcessEmitting(DecodableInterface *decodable);
+  // Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
+  // Returns the cost cutoff for subsequent ProcessNonemitting() to use.
+  // To reduce vtable lookup cost in openfst based graph search, 
+  // make ProcessEmitting() and ProcessNonemitting() be templated on the FstType,
+  // and create wrapper functions, ProcessEmittingWrapper(), which will check 
+  // the actual FstType; if it's ConstFst or VectorFst, it will call
```
danpovey: fix this comment.

danpovey: @chenzhehuai, did you see this?
I asked you to fix this comment many times, by github comment and by email.  For your convenience I'll repeat here an email I sent you separately, which you also seem to have missed.  It's a very small point, but after asking 4 or 5 times, it starts to get annoying.
```
thanks.
Regarding the comment: please change:
// Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
   // Returns the cost cutoff for subsequent ProcessNonemitting() to use.
  // To reduce vtable lookup cost in openfst based graph search, 
   // make ProcessEmitting() and ProcessNonemitting() be templated on the FstType,
    // and create wrapper functions, ProcessEmittingWrapper(), which will check 
    // the actual FstType; if it's ConstFst or VectorFst, it will call 
    // ProcessEmitting() with one of those two types explicitly; otherwise it 
  // will call it with the regular Fst interface (to cover the other cases).
   // will call it with the regular Fst interface .

to:

   /// Processes emitting arcs for one frame.  Propagates from prev_toks_ to cur_toks_.
   /// Returns the cost cutoff for subsequent ProcessNonemitting() to use.
   /// Templated on FST type for speed; called via ProcessEmittingWrapper().


that's what I was trying to say before.
Be careful: this is in 2 places.
```

src/decoder/lattice-faster-online-decoder.cc
```diff
@@ -883,6 +885,7 @@ BaseFloat LatticeFasterOnlineDecoder::ProcessEmitting(
 
   BaseFloat cost_offset = 0.0; // Used to keep probabilities in a good
   // dynamic range.
+  FstType &fst = dynamic_cast<FstType&>(const_cast<fst::Fst<Arc> &>(fst_));
```
danpovey: Actually I think this can be `const FstType &fst`-- no need for const_cast.

src/decoder/lattice-faster-online-decoder.cc
```diff
@@ -891,7 +894,7 @@ BaseFloat LatticeFasterOnlineDecoder::ProcessEmitting(
     StateId state = best_elem->key;
     Token *tok = best_elem->val;
     cost_offset = - tok->tot_cost;
-    for (fst::ArcIterator<fst::Fst<Arc> > aiter(fst_, state);
+    for (fst::ArcIterator<FstType> aiter(fst, state);
          !aiter.Done();
          aiter.Next()) {
       Arc arc = aiter.Value();
```
danpovey: I just realized that there are some obvious optimizations here that can be done.
It seems likely that makeing Arc a const reference (i.e. `const Arc &arc...`) might be more efficient.

And (this is related), below, instead of `arc.weight == ...`, it would be more efficient to just work in cost space, i.e. do:
  `BaseFloat new_weight = arc.weight.Value() + cost_offset - decodable->LogLikelihood(frame, arc.ilabel);`
(keep within the 80-char limit).

Wherever you have these non-const references to Fst, make them const references.  And if there is anywhere else where there are unnecessary copies of Arc, then change that too.  This particular case isn't that important because it's only done for the best token, not all of them.

src/decoder/lattice-faster-decoder.cc
```diff
@@ -715,15 +718,13 @@ BaseFloat LatticeFasterDecoder::ProcessEmitting(DecodableInterface *decodable) {
     StateId state = best_elem->key;
     Token *tok = best_elem->val;
     cost_offset = - tok->tot_cost;
-    for (fst::ArcIterator<fst::Fst<Arc> > aiter(fst_, state);
+    for (fst::ArcIterator<FstType> aiter(fst, state);
          !aiter.Done();
          aiter.Next()) {
-      Arc arc = aiter.Value();
+      const Arc arc = aiter.Value();
```
danpovey: The point here is not the const-- that doesn't change anything.  The point is to make it a reference, e.g.
`const Arc &arc`.  That avoids a copy.  There is no need to make it const if it's not a reference.

src/decoder/lattice-faster-decoder.cc
```diff
@@ -170,7 +171,7 @@ bool LatticeFasterDecoder::GetRawLattice(Lattice *ofst,
           KALDI_ASSERT(f >= 0 && f < cost_offsets_.size());
           cost_offset = cost_offsets_[f];
         }
-        Arc arc(l->ilabel, l->olabel,
+        const Arc arc(l->ilabel, l->olabel,
```
danpovey: revert this change.

PR #2136, opened 2018-01-09
updating fisher_swbd nnet3/chain scripts: removing non-chain training…

xiaohui-zhang:
… scripts, updating ivector training scripts,.etc. @GaofengCheng please re-run and check in your latest chain model training scripts on top of this. Thanks.

PR #2173, opened 2018-01-23
Some updates for Fisher_Swbd( related to #2136)

GaofengCheng:
Related to #2136
What this PR about:
1. rerun baseline after the update from Xiaohui, including blstm_6j, tdnn_opgru_1a
2. adding new training setups for tdnn_opgru and tdnn_lstm, see tdnn_lstm_1b and tdnn_opgru_1b.

Difference between tdnn_lstm_1b and tdnn_lstm_1a : adding batchnorm in TDNN layers, adding dropout for LSTMP, string-style chunk width, 64,32 minibatch,  frames-per-iter 1500000 instead of 1200000, change xent-reg from 0.0025 to 0.01

Difference between tdnn_opgru_1b and tdnn_opgru_1a : using string-style chunk width, 64,32 minibatch,  frames-per-iter 1500000 instead of 1200000, change xent-reg from 0.0025 to 0.01


egs/fisher_swbd/s5/local/chain/run_tdnn_lstm_1b.sh
```diff
@@ -0,0 +1,296 @@
+#!/bin/bash
+# Copyright 2017 University of Chinese Academy of Sciences (UCAS) Gaofeng Cheng
+# Apache 2.0
+
+# Using string-style chunk width, batchnorm in TDNN layers and perframe dropout for LSTMP.
```
danpovey: please make sure the chain_dir_info.pl output is put here, I want to be able to see the number of parameters.

egs/fisher_swbd/s5/local/chain/run_tdnn_opgru_1b.sh
```diff
@@ -0,0 +1,296 @@
+#!/bin/bash
+# Copyright 2017 University of Chinese Academy of Sciences (UCAS) Gaofeng Cheng
```
danpovey: why do you never have any comparisons between different experiments?  That's what the compare_wer_general.sh script is for.  It would allow us to see what the point of the experiments is.
You have 1a and 1b scripts for OPGRU.  There is no comment here saying how 1b differs from 1a.

PR #2174, opened 2018-01-23
Fix upper bound bug in get_num_frames.sh

GaofengCheng:
Problem description:
    When calling utils/data/get_num_frames.sh in large training set (like fisher+swbd+speed perturb), the total frame number may exceed printf() Int value range (<2147483647 ).

example:
bash utils/data/get_num_frames.sh ../s5/data/train_nodup_sp_hires/
output: 2147483647 (it should be 2267406655)

after fix, the output is OK:
bash utils/data/get_num_frames.sh ../s5/data/train_nodup_sp_hires/
output:2267406655


egs/wsj/s5/utils/data/get_num_frames.sh
```diff
@@ -22,4 +22,4 @@ fi
 
 frame_shift=$(utils/data/get_frame_shift.sh $data) || exit 1
 
-awk -v s=$frame_shift '{n += $2} END{printf("%d\n", int(n / s))}' <$data/utt2dur
+python utils/data/get_num_frames.py --frame-shift $frame_shift --utt2dur-dir $data/utt2dur
```
danpovey: Hm.  So one line of awk transforms into 24 lines of python.
Did you try the fix of just changing ` n += $2` to `n += 1.0 * $2` so it uses floating-point, not integer arithmetic?
The result doesn't have to be exact.

PR #2190, opened 2018-01-25
A couple of unrelated minor fixes

akreal:


src/Makefile
```diff
@@ -171,7 +171,7 @@ cudamatrix: base util matrix
 nnet: base util hmm tree matrix cudamatrix
 nnet2: base util matrix lat gmm hmm tree transform cudamatrix
 nnet3: base util matrix lat gmm hmm tree transform cudamatrix chain fstext
-rnnlm: base util matrix cudamatrix nnet3 lm hmm
+rnnlm: base util matrix cudamatrix nnet3 chain lm hmm
```
danpovey: what linking error was this intended to solve?

akreal: sorry, it was caused by a problem in my environment. removed this.

egs/librispeech/s5/local/chain/tuning/run_tdnn_1b.sh
```diff
@@ -173,6 +173,7 @@ if [ $stage -le 15 ]; then
      /export/b0{5,6,7,8}/$USER/kaldi-data/egs/librispeech-$(date +'%m_%d_%H_%M')/s5c/$dir/egs/storage $dir/egs/storage
   fi
 
+  mkdir -p $dir/egs
```
danpovey: hm.
Actually I think the better fix would be to just remove the following line `   touch $dir/egs/.nodelete `. 
that's what would fail.

danpovey: @akreal, could you please change this so it removes the line `touch $dir/egs/.nodelete` instead of adding this line?

akreal: Sure, done.

PR #2194, opened 2018-01-28
[src] Work around problem related to ungetc failures on ifstream

danpovey:


danpovey:
Below is a comment from Mael Primet on the kaldi-help list; this fix should resolve that problem.
I had this in another branch as I'd noticed it there, and I'm cherry-picking that changes to master.
The issue seems to be that while in C, you were always allowed to `unget` one character, it seems that the C++ standard doesn't guarantee this (even though in practice it is possible almost always).

===
While doing a chain training I get this error after the first iteration of NN training

```
nnet3-copy ./exp/tdnn_0/1.1.raw -
nnet3-am-copy --set-raw-nnet=- ./exp/tdnn_0/0.mdl ./exp/tdnn_0/1.mdl
WARNING (nnet3-am-copy[5.3.100-3d33e]:PeekToken():io-funcs.cc:182) Error ungetting '<' in PeekToken
ERROR (nnet3-am-copy[5.3.100-3d33e]:ReadToken():io-funcs.cc:159) ReadToken, failed to read token at file position -1
```

it works if I replace the "-" with a file

what could cause this? I recently updated the version of Kaldi and I did not have the error before

PR #2200, opened 2018-02-01
Fixed rounding issue in utils/data/get_uniform_subsegments.py

mmaciej2:
Using the "int" function instead of proper rounding while creating the segment ID could lead to cases where two segments would erroneously receive the same ID.

### DIFF
--- a/egs/wsj/s5/utils/data/get_uniform_subsegments.py
+++ b/egs/wsj/s5/utils/data/get_uniform_subsegments.py
@@ -87,8 +87,8 @@ def run(args):
         else:
             end = end_time
         new_utt = "{utt_id}-{s:08d}-{e:08d}".format(
-            utt_id=utt_id, s=int(100 * (start - start_time)),
-            e=int(100 * (end - start_time)))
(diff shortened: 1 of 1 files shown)


danpovey:
Can you give an example of the problem?    I have a hard time seeing how this could happen.

mmaciej2:
@danpovey The problem happens when the start_time (I assume as some kind of machine precision issue) that is read out of the segments file is slightly larger than its value rounded to two decimal places. So the sub-segment is listed as being from something like time 0.99999 to 2.9999 and will produce a segments file with something like this:
utt_000099_000299 1.00 3.00
If there is a segment going from 0.99 to 2.99, it will receive the same ID, i.e.
utt_000099_000299 0.99 2.99

I think there's an argument that you should not be trying to produce subsegments that are only one frame apart, so the ID collision isn't an issue, but nevertheless the code in the master branch produces IDs that are inconsistent with the time marks by 0.01 seconds.

If you would like me to point you to a particular segments file and set of parameters that have this problem, I can set that up as well.

PR #2205, opened 2018-02-04
Fix speed-perturbatin for swbd + execuation permissions for some fisher_swbd scripts

hhadian:


egs/fisher_swbd/s5/local/chain/run_tdnn_lstm_1b.sh
```diff
@@ -108,7 +108,7 @@ dir=${dir}$suffix
 build_tree_train_set=train_nodup
 train_set=train_nodup_sp
 build_tree_ali_dir=exp/tri5a_ali
-treedir=exp/chain/tri6_tree
+treedir=exp/chain/tri5_tree
```
danpovey: I think you should change this  line back because it could problems when running from a later --stage.

egs/wsj/s5/utils/data/perturb_data_dir_speed_3way.sh
```diff
@@ -38,9 +38,17 @@ utils/data/get_utt2dur.sh ${srcdir}
 
 utils/data/perturb_data_dir_speed.sh 0.9 ${srcdir} ${destdir}_speed0.9 || exit 1
 utils/data/perturb_data_dir_speed.sh 1.1 ${srcdir} ${destdir}_speed1.1 || exit 1
-utils/data/combine_data.sh $destdir ${srcdir} ${destdir}_speed0.9 ${destdir}_speed1.1 || exit 1
 
-rm -r ${destdir}_speed0.9 ${destdir}_speed1.1
+utils/copy_data_dir.sh --spk-prefix sp1.0- --utt-prefix sp1.0- ${srcdir} data/temp0
```
danpovey: since you are already creating the dir with _speed0.9 and  _speed1.0 as suffixes, I think it would be clearer and more consistent to use _speed1.0 as the suffix for this.

PR #2211, opened 2018-02-10
[WIP] generate egs for age, emotion or TS task

LvHang:
@pegahgh 
I cleaned up my repository. Now the codes and scripts focus on your age, emotion or teacher-student task. Bear in mind, these codes make the egs in the same archive have distinct lengths and use "ranges in script-file lines" method. Please check the codes and see whether you have any other requirements.

In the comment of #2169 , you mentioned "The idea is that in future, specially the label can be even vector like posteriors (e.g. TS learning application)" for function "ProcessUttToLabelFile()".
According to my knowledge, the loss function of Teacher-Student Training is 
  loss = - sum_all_state {teacher_output(state, t) * log student_output(state, t)}
That means, for each frame, we need a posterior vector.
I think to supply this requirement, we simply modify the "ProcessUttToLabelFile()" is not feasible. Function "ProcessUttToLabelFile()" only uses utt_id to find the label so that we can use the label generates the one-hot vector later. However, to slove the problem, we need provide another file which contains the output of teacher system in matrix format or posterior format. If possible, please show me a file which you can provide and I will figure out a way to slove it.

Best,
Hang

### DIFF
--- a/egs/sre08/v1/sid/nnet3/xvector/allocate_egs_seg_v1.py
+++ b/egs/sre08/v1/sid/nnet3/xvector/allocate_egs_seg_v1.py
@@ -1,0 +1,302 @@
+#!/usr/bin/env python3
+
+# Copyright 2017 Johns Hopkins University (author: Hang Lyu)
+
+"""This script is suitable for the training set whose size is not very big, such
+as TS learning, age or emotion id.
+
+For generating egs for xvector in a more effecient way, we use "ranges in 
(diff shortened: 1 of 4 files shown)


danpovey:
@LvHang I'm not sure if you want to finish this?  It seems to lack an example script.

LvHang:
@danpovey Sure, I will revisit the code and pegah's paper about age/TS. Then, I will provide an example script.

PR #2213, opened 2018-02-11
[scripts,egs] Fix failures in some run_ivector_common.sh scripts (since #2205)

danpovey:


danpovey:
The issue that this PR resolves will have caused a failure like this in the AMI setup
steps/online/nnet2/get_pca_transform.sh: Error summing cmvn stats
(or possibly other error messages in setups where run_ivector_common.sh uses PCA) since
commit #2205.  The issue is, I forgot that the older run_ivector_common.sh scripts rely
on the fact that the un-perturbed data has utterance-ids that are subsets of the utterance-ids
in the perturbed data, so we can take a subset and use the old alignments.

PR #2215, opened 2018-02-13
Updated aspire data prep to use LDC2017S21 (and removing window/overlap options that are not used any more from prep_test_aspire.sh). Thanks Jon Nichols

xiaohui-zhang:


xiaohui-zhang:
maybe @vijayaditya can review?

vijayaditya:
Sure will do it tomorrow.

egs/aspire/s5/local/multi_condition/aspire_data_prep.sh
```diff
@@ -22,12 +21,11 @@ tmpdir=`pwd`/data/local/data
...
 if [ $stage -le 0 ]; then
 
-  find $dev_transcript/ -name '*.stm'  > $tmpdir/transcripts.flist
+  find $dev_transcript/ -name 'dev.stm'  > $tmpdir/transcripts.flist
   find $dev_audio/ -name '*.wav'  > $tmpdir/wav.flist
   find $test_audio/ -name '*.wav'  > $tmpdir/wav_test.flist
-  find $eval_audio/ -name '*.wav'  > $tmpdir/wav_eval.flist
 
-  n=`cat $tmpdir/transcripts.flist | wc -l`
+  n=`cut -d' ' -f1 $(cat $tmpdir/transcripts.flist) | uniq | wc -l`
```
vijayaditya: better use awk. Relying cut might cause issues if the field separator was changed to other whitespaces e.g. tabs.

egs/aspire/s5/local/multi_condition/aspire_data_prep.sh
```diff
@@ -22,12 +21,11 @@ tmpdir=`pwd`/data/local/data
...
   if [ $n -ne 30 ]; then
     echo "Expected to find 30 transcript files in the aspire_single_dev_transcript directory, found $n"
```
vijayaditya: Change the error message. We are no longer looking for $n files. Actually just check for dev.stm and throw an error if it is not found at early stages of the script. 

Also it might be better to move all these commands to `aspire_data_sanity_check.sh` script which can also be run individually, and call it from here.

egs/aspire/s5/local/multi_condition/aspire_data_prep.sh
```diff
@@ -122,12 +115,12 @@ for line in sys.stdin.readlines():
   print '{0} sox --vol {1} {2} -r 8000 -t wav - |'.format(file_id, out_rms, line)
 "| sort -k1,1 -u  > $dev/wav.scp || exit 1;
   cat $dev/wav.scp |awk '{printf("%s %s A\n", $1, $1)}' > $dev/reco2file_and_channel
-  cp $aspire_data/glm $dev
+  cp $aspire_data/my_english.glm $dev
 fi
 
 # prepare the eval and test data
```
vijayaditya: update comment.

egs/aspire/s5/local/multi_condition/run_nnet2_common.sh
```diff
@@ -5,6 +5,7 @@
 
 
 stage=1
+aspire_data=
```
vijayaditya: Describe this variable.

egs/aspire/s5/local/nnet3/prep_test_aspire.sh
```diff
@@ -186,8 +186,6 @@ if [ $stage -le 8 ]; then
     --ctm-beam 6 \
     --iter $iter \
     --decode-mbr true \
-    --window $window \
```
vijayaditya: Add a note in the PR detailing why these variables have been removed.

egs/aspire/s5/run.sh
```diff
@@ -160,7 +161,7 @@ steps/train_sat.sh  --cmd "$train_cmd" \
 local/build_silprob.sh
 
 # train the neural network model
-local/multi_condition/run_nnet2_ms.sh
+local/multi_condition/run_nnet2_ms.sh --aspire-data $aspire_data
```
vijayaditya: Could you also please call local/chain/run_tdnn.sh rather than the nnet2 script.

egs/aspire/s5/run.sh
```diff
@@ -9,6 +9,7 @@
 . ./path.sh
 
 mfccdir=`pwd`/mfcc
+aspire_data=/export/corpora/LDC/LDC2017S21/IARPA-ASpIRE-Dev-Sets-v2.0/data
```
vijayaditya: Better to check for the cluster and set the variable e.g. see AMI script ([line](https://github.com/kaldi-asr/kaldi/blob/master/egs/ami/s5b/run.sh#L30))

egs/aspire/s5/local/multi_condition/aspire_data_prep.sh
xiaohui-zhang: Thanks for the comments.
Hmm since we are doing "$(cat $tmpdir/transcripts.flist)" here, we could still count the number of files in the stm. Of course I should check whether the stm exists at an earlier stage.

vijayaditya: Yep however you decide to write the check ensure that the error message is updated according to the new directory structure, so that the users can work out what went wrong.

xiaohui-zhang: sure man XD

xiaohui-zhang:
I've addressed problems mentioned by @vijayaditya. @vimalmanohar has offered to add results and his TDNN training scripts on top of this. Thanks

vimalmanohar:
All these commits are moved to #2230. This PR can be closed.

PR #2219, opened 2018-02-14
[scripts] Add switch 'share-state-roots' for share/non-share roots through states.

swabe:
if set True, single root per phoneme (all states) is created. False, create root for each phoneme state.


danpovey:
Is there a particular scenario where you found this useful, or where it's needed for some special reason?  I don't like to pass through options at the script level unless there is some concrete use-case.

swabe:
We are using triphone based alignement to train phoneme recognizer which has 3states per phoneme. Therefore, it is necessary to switch-off across state clustering.

PR #2221, opened 2018-02-14
[scripts] Add delta_opts for train_mono.sh

swabe:


danpovey:
this is not unreasonable, but in general I don't like to add extra options unless there's an immediate need for it (e.g. one of the example scripts requires it).

danpovey:
merged at karel's request.

vesis84:
Thank you :)

PR #2223, opened 2018-02-15
[not-for-merge] Replacing the statistic pooling layer with self-attention layer in sr…

tomkocse:
…e16 sv task

### DIFF
--- a/egs/sre16/v1/local/nnet3/xvector/run_xvector.sh
+++ b/egs/sre16/v1/local/nnet3/xvector/run_xvector.sh
@@ -1,1 +1,154 @@
-tuning/run_xvector_1a.sh
+#!/bin/bash
+# Copyright      2017   David Snyder
+#                2017   Johns Hopkins University (Author: Daniel Garcia-Romero)
+#                2017   Johns Hopkins University (Author: Daniel Povey)
+# Apache 2.0.
+
+# This script trains a DNN similar to the recipe described in
+# http://www.danielpovey.com/files/2017_interspeech_embeddings.pdf .
+
+. ./cmd.sh
+set -e
+
+stage=1
+train_stage=0
+use_gpu=true
+remove_egs=false
+
+data=data/swbd_sre_combined_no_sil
+nnet_dir=exp/xvector_nnet_1a/
+egs_dir=exp/xvector_nnet_1a/egs
+
+. ./path.sh
+. ./cmd.sh
+. ./utils/parse_options.sh
+
+num_pdfs=$(awk '{print $2}' $data/utt2spk | sort | uniq -c | wc -l)
+
+# Now we create the nnet examples using sid/nnet3/xvector/get_egs.sh.
+# The argument --num-repeats is related to the number of times a speaker
+# repeats per archive.  If it seems like you're getting too many archives
+# (e.g., more than 200) try increasing the --frames-per-iter option.  The
+# arguments --min-frames-per-chunk and --max-frames-per-chunk specify the
+# minimum and maximum length (in terms of number of frames) of the features
+# in the examples.
+#
+# To make sense of the egs script, it may be necessary to put an "exit 1"
+# command immediately after stage 3.  Then, inspect
+# exp/<your-dir>/egs/temp/ranges.* . The ranges files specify the examples that
+# will be created, and which archives they will be stored in.  Each line of
+# ranges.* has the following form:
+#    <utt-id> <local-ark-indx> <global-ark-indx> <start-frame> <end-frame> <spk-id>
+# For example:
+#    100304-f-sre2006-kacg-A 1 2 4079 881 23
+
+# If you're satisfied with the number of archives (e.g., 50-150 archives is
+# reasonable) and with the number of examples per speaker (e.g., 1000-5000
+# is reasonable) then you can let the script continue to the later stages.
+# Otherwise, try increasing or decreasing the --num-repeats option.  You might
+# need to fiddle with --frames-per-iter.  Increasing this value decreases the
+# the number of archives and increases the number of examples per archive.
+# Decreasing this value increases the number of archives, while decreasing the
+# number of examples per archive.
+if [ $stage -le 4 ]; then
+  echo "$0: Getting neural network training egs";
+  # dump egs.
+  if [[ $(hostname -f) == *.clsp.jhu.edu ]] && [ ! -d $dir/egs/storage ]; then
+    utils/create_split_dir.pl \
+     /export/b{03,04,05,06}/$USER/kaldi-data/egs/sre16/v2/xvector-$(date +'%m_%d_%H_%M')/$egs_dir/storage $egs_dir/storage
+  fi
+  sid/nnet3/xvector/get_egs.sh --cmd "$train_cmd" \
+    --nj 8 \
+    --stage 0 \
+    --frames-per-iter 1000000000 \
+    --frames-per-iter-diagnostic 100000 \
+    --min-frames-per-chunk 200 \
+    --max-frames-per-chunk 400 \
+    --num-diagnostic-archives 3 \
+    --num-repeats 35 \
+    "$data" $egs_dir
+fi
+
+if [ $stage -le 5 ]; then
+  echo "$0: creating neural net configs using the xconfig parser";
+  num_targets=$(wc -w $egs_dir/pdf2num | awk '{print $1}')
+  feat_dim=$(cat $egs_dir/info/feat_dim)
+
+  # This chunk-size corresponds to the maximum number of frames the
+  # stats layer is able to pool over.  In this script, it corresponds
+  # to 100 seconds.  If the input recording is greater than 100 seconds,
+  # we will compute multiple xvectors from the same recording and average
+  # to produce the final xvector.
+  max_chunk_size=10000
+
+  # The smallest number of frames we're comfortable computing an xvector from.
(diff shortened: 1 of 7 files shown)


egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
```diff
@@ -0,0 +1,218 @@
...
+"""
+
+from __future__ import print_function
+import re
+from libs.nnet3.xconfig.basic_layers import XconfigLayerBase
+
+
+class XconfigSelfLayer(XconfigLayerBase):
+    """This class is for parsing lines like
+    self-layer name=attention config=mean+stddev(-99:3:9:99) input=tdnn1
```
vimalmanohar: Better to use a different name like weighted-stats-layer instead of a generic one like self-layer.

egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(self, first_token, key_to_value, prev_names=None):
+        assert first_token in ['self-layer']
+        XconfigLayerBase.__init__(self, first_token, key_to_value, prev_names)
+
+    def set_default_configs(self):
+        self.config = {'input': '[-1]',
+                       'dim': -1,
+                       'config': '',
+                       'affine-dim': 300,
+                       'num-heads': 0}
```
vimalmanohar: Can num-heads be 0? Is it checked anywhere that it should not be zero?

src/nnet3/nnet-general-component.h
```diff
@@ -775,7 +775,149 @@ class DropoutMaskComponent: public RandomComponent {
...
+ # the input matrix is the count from the extraction component, it will not be used in 
+ # this component.
+ # The tricky here is that a softmax computation is involved in this component because
+ # the weights are arrange as column vector of the matrix, so we can't use the original SoftMax
+ # component and have to do the softmax copmutation here.
+ # We need to transpose the weights to row vector, then call ApplySoftMaxPerRow(),
+ # then transpose the weights back to column vector.
+
+ configs and their defaults:  input-dim=-1, input-period=1, left-context=-1, right-context=-1,
+    num-heads=1, num-log-count-features=0, output-stddevs=true
```
vimalmanohar: num-heads=0 seems to be the default.

egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
```diff
@@ -0,0 +1,218 @@
...
+
+        configs.append(
+             'component-node name={0}.batchnorm'
+             ' component={0}.batchnorm input={0}.relu'
+             ''.format(self.name))
+
+        # The second affine node.
+        affine_options = 'param-stddev=0.04472135955 bias-stddev=1.0 bias-mean=0.0 max-change=0.75 l2-regularize=0.0'
+        configs.append(
+            'component name={0}.second_affine type=NaturalGradientAffineComponent'
```
vimalmanohar: Is this needed if num-heads=0?

egs/sre16/v1/local/nnet3/xvector/run_xvector.sh
```diff
@@ -1 +0,0 @@
-tuning/run_xvector_1a.sh
```
vimalmanohar: Is dropout used?

egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
```diff
@@ -0,0 +1,218 @@
...
+             ' target-rms={2}'
+             ''.format(self.name, da_dim, target_rms))
+
+        configs.append(
+             'component-node name={0}.batchnorm'
+             ' component={0}.batchnorm input={0}.relu'
+             ''.format(self.name))
+
+        # The second affine node.
+        affine_options = 'param-stddev=0.04472135955 bias-stddev=1.0 bias-mean=0.0 max-change=0.75 l2-regularize=0.0'
```
vimalmanohar: Explain what the stddev value is..

egs/sre16/v1/local/nnet3/xvector/run_xvector.sh
```diff
@@ -1 +0,0 @@
...
\ No newline at end of file
```
vimalmanohar: sd -> stddev

egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
```diff
@@ -0,0 +1,218 @@
...
+        # The statistic extraction node.
+        configs.append(
+            'component name={name}-extraction-{lc}-{rc} '
+            'type=StatisticsExtractionComponent input-dim={dim} '
+            'input-period={input_period} output-period={output_period} '
+            'include-variance={var} '.format(
+                name=self.name, lc=self._left_context, rc=self._right_context,
+                dim=input_dim, input_period=self._input_period,
+                output_period=self._stats_period,
+                var='true' if self._output_stddev else 'false'))
```
danpovey: @david-ryan-snyder, you should probably review this too.
This uses StatisticsExtractionComponent but without the StaticsPoolingComponent, which is odd;
it's, in effect, pooling short groups of frames before the self-attention.  
Also, this layer seems to be a bit against the spirit of a layer; it has a couple of Affine+Relu+Batchnorm before the poling and attention, so should probably be split into two layers.
@vimalmanohar, if you have specific ideas to improve this, please comment too.

david-ryan-snyder: @tomkocse can correct me if I'm wrong, but I don't think this PR is meant to be merged in it's current state. It's related to some experiments we're doing. 

I think there's more work to be done on this, both to establish it's value as a layer and to improve the code (again, @tomkocse, correct me if I'm wrong).

david-ryan-snyder:
Also, I think this is closely related to the weighted pooling that we experimented with in the past. It appears to perform similarly. We should revisit that before committing to this implementation, as the weighted pooling is simpler to implement. Anyway, @tomkocse is looking into this stuff more, let's see where this goes.

src/nnet3/nnet-component-itf.cc
```diff
@@ -59,6 +59,8 @@ ComponentPrecomputedIndexes* ComponentPrecomputedIndexes::NewComponentPrecompute
     ans = new StatisticsExtractionComponentPrecomputedIndexes();
   } else if (cpi_type == "StatisticsPoolingComponentPrecomputedIndexes") {
     ans = new StatisticsPoolingComponentPrecomputedIndexes();
+  } else if (cpi_type == "SelfAttentionComponentPrecomputedIndexes") {
```
danpovey: @tomkocse, this seems to be a bug fix, can you please make a separate PR for it?

tomkocse: @danpovey I don't see the bug here in the original code, can you point that out ?

egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
tomkocse: @david-ryan-snyder Yes, I agree what you said, let's wait for more results from further experiments.

src/nnet3/nnet-component-itf.cc
danpovey: oh sorry, I didn't realize that SelfAttentionComponent was new.

src/nnet3/nnet-general-component.h
```diff
@@ -775,7 +775,149 @@ class DropoutMaskComponent: public RandomComponent {
...
+/*
+  Class SelfAttentionComponent is used together with
+  StatisticsExtractionComponent to extract weighted mean and
+  standard-deviation statistics.
+
+  SelfAttentionComponent is a part of the self-attention mechanism, which does 
+  a linear combination of a series of input frames
+
+ # In SelfAttentionComponent, the first n columns of the input matrix are interpreted
+ # as the weight vectors in multi-head attentions. If only a single-head attention is used,
```
pegahgh: mention that n is number of heads in the attention layer.

src/nnet3/nnet-general-component.h
```diff
@@ -775,7 +775,149 @@ class DropoutMaskComponent: public RandomComponent {
...
+ # then the first column of the input matrix is the weight vector. The n + 1 th column of
+ # the input matrix is the count from the extraction component, it will not be used in
```
pegahgh: Does it mean that you always expect the previous layer to be StatisticExtraction component?

egs/wsj/s5/steps/libs/nnet3/xconfig/self_layer.py
```diff
@@ -0,0 +1,218 @@
...
+        da_dim = self._affine_dim
+        self_repair_scale = 1e-05
+        target_rms = 1.0
+        affine_options = 'max-change=0.75'
+
+        configs = []
+        # The first affine node.
+        configs.append(
+            'component name={0}.first_affine type=NaturalGradientAffineComponent'
+            ' input-dim={1} output-dim={2} {3}'
```
pegahgh: I am not sure, why do you need to add separate part for having affine+nonlin+batchnorm within this component, when you can add this part using relu-batchnorm-layer at script level.

PR #2224, opened 2018-02-15
Fixed issue with dowsampling the waveform during online decoding.

pegahgh:
I checked this PR by doing online decoding on 8kHz trained model by adding options --allow-downsample=true and --sample-frequency=4000 in conf/mfcc_hires.conf.

### DIFF
--- a/egs/tedlium/s5_r2/local/ted_download_lm.sh
+++ b/egs/tedlium/s5_r2/local/ted_download_lm.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+#
+# Copyright  2018  David Snyder
+# Apache 2.0
+#
+# This script downloads pre-built language models trained on the Cantab-Tedlium
+# text data and Tedlium acoustic training data.  If you want to build these
+# models yourself, run the script local/ted_train_lm.sh.
+
+set -e
+
+echo "$0: downloading Tedlium 4 gram language models (it won't re-download if it was already downloaded.)"
+wget --continue http://kaldi-asr.org/models/5/4gram_small.arpa.gz -P data/local/local_lm/data/arpa || exit 1
+wget --continue http://kaldi-asr.org/models/5/4gram_big.arpa.gz -P data/local/local_lm/data/arpa || exit 1
+
+exit 0
+
--- a/egs/tedlium/s5_r2/run.sh
+++ b/egs/tedlium/s5_r2/run.sh
@@ -57,10 +57,12 @@ if [ $stage -le 3 ]; then
 fi
 
 if [ $stage -le 4 ]; then
-  # later on we'll change this script so you have the option to
-  # download the pre-built LMs from openslr.org instead of building them
+  # Download the pre-built LMs from kaldi-asr.org instead of building them
   # locally.
-  local/ted_train_lm.sh
+  local/ted_download_lm.sh
+  # Uncomment this script to build the language models instead of
+  # downloading them from kaldi-asr.org.
+  # local/ted_train_lm.sh
 fi
 
 if [ $stage -le 5 ]; then


src/feat/online-feature.cc
```diff
@@ -41,20 +41,37 @@ template<class C>
 void OnlineGenericBaseFeature<C>::AcceptWaveform(BaseFloat sampling_rate,
                                                  const VectorBase<BaseFloat> &waveform) {
   BaseFloat expected_sampling_rate = computer_.GetFrameOptions().samp_freq;
-  if (sampling_rate != expected_sampling_rate)
-    KALDI_ERR << "Sampling frequency mismatch, expected "
-              << expected_sampling_rate << ", got " << sampling_rate;
-  if (waveform.Dim() == 0)
+  bool allow_downsample = computer_.GetFrameOptions().allow_downsample;
```
danpovey: This won't give the right results if you call this multiple times with separate pieces of waveform.
You'll have to add a class member of type LinearResample*.  That object retains all the state necessary to do this right when different pieces of waveform are being appended together.  When the user says the waveform is done, you can, I think, call its Resample() function with an empty piece of waveform and flush == true.

You should probably add a function InputSampleRate() to the LinearResample object so that you can check the user doesn't call AcceptWaveform() with different sampling_rate values on different calls.  It's this class's responsibility to make sure that the sampling_rate values provided on different calls are all the same, so make sure if that is done, an error is raised and it doesn't just silently do the wrong thing in that case.

pegahgh: @danpovey, I wonder if this commit contains what you asked me!
I did simple test in online setup and it works properly, but probably someone should test it.

src/feat/online-feature.cc
```diff
@@ -35,26 +35,63 @@ template<class C>
...
+  Vector<BaseFloat> downsampled_wave;
+  int32 waveform_dim = waveform.Dim();
+  if (sampling_rate < expected_sampling_rate)
+    KALDI_ERR << "The waveform is allowed to get downsampled."
```
danpovey: This error message is confusing.  Just do
```   KALDI_ERR << "Sample frequency " << sampling_rate
                     << " is less than the expected value " << expected_sampling_rate;```

File: src/feat/online-feature.cc [hunk 3]
```diff
@@ -35,26 +35,63 @@ template<class C>
...
+                                                 expected_sampling_rate,
+                                                 lowpass_cutoff,
+                                                 lowpass_filter_width);
+    }
+    BaseFloat old_sampling_rate = waveform_downsampler_->InputSampleRate();
+    // Check wheter the sampling rate values provided on different calls are
+    // the same.
+    if (sampling_rate != old_sampling_rate)
+      KALDI_ERR << "Waveform sampling rate value provided in this call "
+                << "mismatches the value in older calls:"
```
danpovey: mismatches -> differs from; older calls -> previous calls.

src/feat/online-feature.cc
```diff
@@ -35,26 +35,63 @@ template<class C>
...
+                << sampling_rate << " .vs " << old_sampling_rate;
+
+    // flush out the last few samples of input waveform, if input_finished_
+    // is true
+    const bool flush = input_finished_;
```
danpovey: AcceptWaveform() should never be called if input_finished_ is false.  It will always be false.

src/feat/online-feature.h
```diff
@@ -117,13 +118,18 @@ class OnlineGenericBaseFeature: public OnlineBaseFeature {
...
+  // already discarded, i.e. that were prior to 'waveform_remainder_'.
   int64 waveform_offset_;
 
   // waveform_remainder_ is a short piece of waveform that we may need to keep
   // after extracting all the whole frames we can (whatever length of feature
   // will be required for the next phase of computation).
   Vector<BaseFloat> waveform_remainder_;
+
+  // This object is used to down-sample the waveform, when
+  // sampling rate in frame extraction is less than
```
danpovey: sampling rate->the sampling rate; waveform sampling rate -> the waveform sampling rate.

danpovey:
@galv, since you first reported this issue, would you mind testing if this patch works, and including whether it works with features that don't need downsampling?  This is mostly just to check for fatal bugs.

danpovey:
@pegahgh, is there any chance you could run online decoding for a setup that you have previously run online decoding for (with nothing changed) and make sure the output, or at least the WER, is identical?

pegahgh:
Sure, I will test it in AMI-SDM!

PR #2226, opened 2018-02-17
Update gst-plugin Makefile

rjcobain:
The plugin is not recognised as valid when using the master gstreamer branch because the symname is extracted from the filename, https://github.com/GStreamer/gstreamer/blob/master/gst/gstplugin.c#L685

### DIFF
--- a/egs/voxforge/gst_demo/run-simulated.sh
+++ b/egs/voxforge/gst_demo/run-simulated.sh
@@ -26,8 +26,8 @@ ac_model=${data_file}/models/$ac_model_type
 trans_matrix=""
 audio=${data_file}/audio
 
-if [ ! -s $KALDI_ROOT/src/gst-plugin/libgstkaldi.so ]; then
-    echo "Kaldi Gstreamer plugin libarary $KALDI_ROOT/src/gst-plugin/libgstkaldi.so not present, make it first"
+if [ ! -s $KALDI_ROOT/src/gst-plugin/libgstonlinegmmdecodefaster.so ]; then
+    echo "Kaldi Gstreamer plugin libarary $KALDI_ROOT/src/gst-plugin/libgstonlinegmmdecodefaster.so not present, make it first"
     exit 1
 fi
 
--- a/src/doc/online_programs.dox
+++ b/src/doc/online_programs.dox
@@ -163,7 +163,7 @@ repository (http://backports.debian.org). Install the 'good' GStreamer plugins (
 and GStreamer 1.0 tools (package `gstreamer1.0-tools`). A demo program also requires the PulseAudio Gstreamer plugins 
 (package gstreamer1.0-pulseaudio).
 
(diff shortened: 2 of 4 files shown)


danpovey:
I don't follow the gstreamer enough to really know whether this is right, but you sound like you know what you are talking about -> merging.

danpovey:
Actually before I merge, can you confirm that this doesn't break the example in Kaldi (I think it was vystadial_cz or something).

rjcobain:
Hi Dan, will check them now, there are a couple of other files that reference libgstkaldi.so

rjcobain:
Having some trouble getting the vystadial demos to run, will try again tomorrow.  The demo gst_demo/run-simulated.sh was updated and it runs fine.

rjcobain:
The vystadial demos are horribly broken and out of date, but I managed to get some of them running from a docker image ufaldsg/pykaldi.  They do not appear to have any dependency on the gst-plugin

danpovey:
OK, I'll merge.  If you think it makes sense to just delete the vystadial demo, let me know.  They may be doing more harm than good.

rjcobain:
I wouldn't like to make that call, maybe if someone else could have a look at them, maybe I made a mistake.  The problem, as far as I can see, is that they depend on an old python wrapper for Kaldi, https://github.com/UFAL-DSG/pykaldi/  I tried using the current wrapper, https://github.com/pykaldi/pykaldi but it didn't work due to api changes.  PyKaldi also relies on a fork of Kaldi.  The project says that "We are hoping to upstream these changes over time.".  Once their changes get merged then it might be possible to update/fix these demos.

danpovey:
I don't think the old pykaldi is being maintained, but the new one is
(although I think it's not officially released yet; they have an upcoming
ICASSP paper).  @dogancan, maybe you could look into this, to see whether
we should still be supporting the vystadial example scripts?

Dan

dogancan:
@danpovey I don't know much about the vystadial example. I also don't know if the old pykaldi package the vystadial scripts depend on is compatible with kaldi any more. Maybe we should ask @oplatek?

@rjcobain The new pykaldi package has no connection with other python wrappers written for kaldi. We have no intention of adding support for the Python APIs provided by other packages. If by "it didn't work due to api changes", you meant it does not support the API of the old pykaldi package, unfortunately that won't change in the future. Updating vystadial demo to work with new pykaldi is doable but not trivial since it uses a custom decoder defined in the old pykaldi package. On the other hand, we have every intention of supporting kaldi C++ API. In fact, I just finished updating pykaldi for kaldi v5.4. I also updated our kaldi fork so it is caught up with the upstream. I am hoping we can upstream the changes we made to Kaldi in the near future. I just haven't had the opportunity yet.

rjcobain:
Thanks for the explanation Dogan, I am not a Python developer but even I should have noticed that they are completely different!

oplatek:
@dogancan I do not think that vystadial scripts have ever been dependent on pykaldi (my very old Python wrapper of kaldi)

PR #2228, opened 2018-02-20
WIP: enable using utterance weights for chain training

jtrmal:
@danpovey, this is so far what I have done.
For some reasons I started to get CUDA crash because of exhausted GPU memory, but I also merged current master and I'm not sure if some of your changes actually increase memory consumption/fragmentation. But it's also possible it's a bug in the egs merging code (I'll look at that).
It's also possible that the egs need to be generated again, even though I doubt that.

### DIFF
--- a/src/nnet3/nnet-chain-example.cc
+++ b/src/nnet3/nnet-chain-example.cc
@@ -428,11 +428,17 @@ bool NnetChainExampleStructureCompare::operator () (
   for (size_t i = 0; i < size; i++)
     if (!io_compare(a.inputs[i], b.inputs[i]))
       return false;
+
   size = a.outputs.size();
   for (size_t i = 0; i < size; i++)
     if (a.outputs[i].name != b.outputs[i].name ||
         a.outputs[i].indexes != b.outputs[i].indexes)
       return false;
+  // verify the weights are compatible (i.e. equal) for merging
+  for (size_t i = 0; i < size; i++)
+    if (a.outputs[i].supervision.weight != b.outputs[i].supervision.weight)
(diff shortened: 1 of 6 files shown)


src/nnet3/nnet-chain-example.cc
```diff
@@ -428,11 +428,17 @@ bool NnetChainExampleStructureCompare::operator () (
   for (size_t i = 0; i < size; i++)
     if (!io_compare(a.inputs[i], b.inputs[i]))
       return false;
+
```
danpovey: It would be more efficient if you somehow included the weight in the return of NnetChainExampleStructureHasher.  Otherwise it will have to do a lot of comparisons if people use many different weights.

jtrmal: True, I'll try to figure something out.

danpovey:
@jtrmal I think it might not be much work to get this checked in. 
It might be a good idea to note at the script level that you shouldn't have too many distinct weights, as only examples with a particular weight can be merged into a minibatch.  (IIRC).

jtrmal:
@danpovey I just noticed I forgot about this one. Is this still something that makes sense to merge?

danpovey:
Hm.
It depends whether we have any use-cases for it.  I am cautious about adding complexity to the original chain training setup, partly because I plan to begin a rewrite of it starting this month.
Also there are conflicts.  Let me know whether there are immediate use-cases.

jtrmal:
At this time I cannot think of any -- I will close the PR and we can always necromance it in case it will be needed.

Issue #2227, opened 2018-02-18
curand return codes


danpovey:
I notice we don't seem to print the curand return status values right.  Can someone please fix this?

danpovey:
@galv, do you have time for this? I think you fixed the CUBLAS ones.

PR #2232, opened 2018-02-21
[src] Print curand return codes

entn-at:
Adresses #2227. I tried to make it as similar as possible to the specialized status code functions for CUBLAS and CUSPARSE. I don't have much experience with this part of the code. Please let me know if you had something different in mind.

Running `make test` in `cudamatrix/` results in SUCCESS for all tests.


danpovey:
Thanks!
Did you run this test on a machine where CUDA was installed and Kaldi had been configured to use CUDA?  Just checking.

entn-at:
Yes, with CUDA toolkit 9.0 and tested using a GTX 1080.

danpovey:
Thanks!

PR #2238, opened 2018-02-23
[scripts] Fix typo in comment

mpuels:
According to http://kaldi-asr.org/doc/transform.html#transform_mllt it should be "STC" instead of "CTC".


danpovey:
Merging but if in future you have more typo fixes, try to batch them up in a single PR.

mpuels:
Ok, will do it.

PR #2247, opened 2018-03-01
Added utils/prepare_extended_lang.sh (extending vocab of a lang-dir) and rnnlm/change_vocab.sh (changing vocab of an existing rnnlm-dir)

xiaohui-zhang:
…s into the lexicon/language model and prepare lang

### DIFF
--- a/egs/wsj/s5/utils/lang/add_unigrams_arpa.pl
+++ b/egs/wsj/s5/utils/lang/add_unigrams_arpa.pl
@@ -1,0 +1,93 @@
+#!/usr/bin/env perl
+
+# Copyright 2018  Xiaohui Zhang
+# Apache 2.0.
+#
+use strict;
+use warnings;
+use Getopt::Long;
+
+my $Usage = <<EOU;
+# This is a simple script to add unigrams to an ARPA lm file.
+Usage: utils/lang/add_unigrams_arpa.pl [options] <oov-prob-file> <scale> <input-arpa >output-arpa
+<oov-prob-file> contains a list of words and their probabilities, e.g. "jack 0.2". All probs will be
+scaled by a positive scalar <scale> and then be used as the unigram prob. of the added word.
+The scale should approximiately relect the OOV rate of the language in concern.
+EOU
+
+my @F;
+my @OOVS;
+
+if (@ARGV != 2) {
+  die $Usage;
+}
+
+# Gets parameters.
+my $oov_prob_file = shift @ARGV;
+my $scale = shift @ARGV;
+my $arpa_in = shift @ARGV;
+my $arpa_out = shift @ARGV;
+
+# Opens files.
+open(F, "<$oov_prob_file") || die "$0: Fail to open $oov_prob_file\n";
+while (<F>) { push @OOVS, $_; }
+my $num_oovs = @OOVS;
+
+$scale > 0.0 || die "Bad scale";
+print STDERR "$0: Creating LM file with additional unigrams, using $oov_prob_file\n";
+
+my %vocab;
+my $unigram = 0;
+my $num_unigrams = 0;
+my @lines;
+
+# Parse and record the head and unigrams in the ARPA LM.
+while(<STDIN>) {
+  if (m/^ngram 1=(\d+)/) { $num_unigrams = $1; }
+  
+  if (m/^\\2-grams:$/) { last; }
+  if (m/^\\1-grams:$/) { $unigram = 1; push(@lines, $_); next; }
+  if (m/^\\2-grams:$/) { $unigram = 0; }
+
+  my @col = split(" ", $_);
+  if ( $unigram == 1 ) {
+    # Record in-vocab words into a map.
+    if ( @col > 0 ) {
+      my $word = $col[1];
+      $vocab{$word} = 1;
+      push(@lines, $_);
+    } else {
+      # Insert out-of-vocab words and their probs into the unigram list.
+      foreach my $l (@OOVS) {
+        my @A = split(" ", $l);
+        @A == 2 || die "bad line in oov2prob: $_;";
+        my $word = $A[0];
+        my $prob = $A[1];
+        if (exists($vocab{$word})) { next; }
+        $num_unigrams ++;
+        my $log10prob = (log($prob * $scale) / log(10.0));
+        $vocab{$word} = 1;
+        my $line = sprintf("%.6f\t$word\n", $log10prob);
+        push(@lines, $line);
+      }
+    }
+  } else { push(@lines, $_); }
+}
+
+# Print the head and unigrams, with the updated # unigrams in the head.
+foreach my $l (@lines) {
+  if ($l =~ m/ngram 1=/) {
+    print "ngram 1=$num_unigrams\n";
+  } else {
+    print $l;
+  }
+}
+
+# Print the left fields.
+print "\n\\2-grams:\n";
+while(<STDIN>) {
+  print;
+}
+
+close(F);
+exit 0
--- a/egs/wsj/s5/utils/prepare_extended_lang.sh
+++ b/egs/wsj/s5/utils/prepare_extended_lang.sh
@@ -18,12 +18,10 @@
 # derived files, that go in data/lang/.
 
 # Begin configuration section.
-oov_unigram_prob=0.0001 # Unigram LM probability we'll assign to each word outside the original vocab.
-srilm_opts="-subset -prune-lowprobs -unk"
 prep_lang_opts=
-lm=
-to_carpa=false # convert the language model to const arpa format rather than FST 
 stage=0
+word_list= # if a word list (mapping words from the srcdict to IDs) is provided,
+# we'll make sure the IDs of these words are kept as before.
 # end configuration sections
 
 echo "$0 $@"  # Print the command line for logging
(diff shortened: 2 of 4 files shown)


jtrmal:
Hi Samuel, how this is different in functionality from utils/format_lm_srilm.sh ?

xiaohui-zhang:
Hi Yenda, this basically allow the user insert new word & pronunciations (e.g. words we want to add during decoding, and their g2p-generated prons) into an existing lexicon and LM, and then rebuild lang, without re-training LM.

jtrmal:
Yes, well I didn't read the format_lm_srilm.sh carefully, but I have
thought that's exactly what that script is doing (I might be mistaken)
y.

xiaohui-zhang:
Hmm format_lm_srilm.sh is capable of shrinking the vocabulary of an LM, while mine aims to extend the vocab :-)

danpovey:
I'll look at this later.  I think there are problems with format_lm_srilm,
but this new script may not be perfect either.

xiaohui-zhang:
@danpovey ready for review.

egs/wsj/s5/utils/prepare_extended_lang.sh
```diff
@@ -0,0 +1,182 @@
...
+stage=0
+# end configuration sections
+
+echo "$0 $@"  # Print the command line for logging
+
+. utils/parse_options.sh
+
+if [ $# -ne 7 ]; then
+  echo "usage: utils/prepare_extended_lang.sh <dict-src-dir> <oov-dict-entry> <extra-lexicon> <phone-symbol-table> <extended-dict-dir> <tmp-dir> <extended-lang-dir>"
+  echo "e.g.: utils/prepare_extended_lang.sh data/local/dict <SPOKEN_NOISE> lexicon_extra.txt data/lang/phones.txt data/local/dict_ext data/local/lang_ext data/lang_ext"
```
danpovey: you should probably put single quotes around `<SPOKEN_NOISE>` here.

egs/wsj/s5/utils/prepare_extended_lang.sh
```diff
@@ -0,0 +1,182 @@
...
+  echo "The goal is to extend the lexicon from <dict-src-dir> with extra lexical entries from <extra-lexicon>, putting the extended lexicon into <extended-dict-dir>,"
```
danpovey: we don't strictly enforce 80 chars per line, but try not to go over around 100.

egs/wsj/s5/utils/prepare_extended_lang.sh
```diff
@@ -0,0 +1,182 @@
...
+  echo "and then build a valid lang dir <extended-lang-dir>."
+  echo "<dict-src-dir> must be a valid dictionary dir and <oov-dict-entry> is the oov word (see utils/prepare_lang.sh for details)."
+  echo "If a language model is provided, we'll also extend the language model and re-build G.fst (or G.carpa), by inserting extra words as extra unigrams,"
```
danpovey: change "If a language model is provided," to "If a language model is provided via the --lm option,"

egs/wsj/s5/utils/prepare_extended_lang.sh
```diff
@@ -0,0 +1,182 @@
...
+  echo "with probalities as the same as the unigram prob. (optionally boosted) of <oov-dict-entry> from the original LM."
```
danpovey: It's not clear what 'optionally boosted' here means-- and I don't think it would be appropriate to give the OOV-word prob to each individual OOV word, as there could be many of them.

egs/wsj/s5/utils/prepare_extended_lang.sh
```diff
@@ -0,0 +1,182 @@
...
+    $oov_word $tmpdir $extlang || exit 1;
+fi
+
+if [ $stage -le 2 ]; then
+  [ -z "$lm" ] && echo "$(basename $0): Not extending the language model since it's not provided." && exit 0;
+  awk '{print $1}' $extlang/words.txt > $tmpdir/voc || exit 1;
+  
+  echo "$(basename $0): Inserting extra unigrams into '$lm'."
+  gunzip -c $lm | \
+    awk -v s=$extdict/lexicon_extra.txt -v oov=$oov_word -v boost=$oov_unigram_prob -v prob=$oov_prob \
```
danpovey: a few lines of comment here explaining what is going on would be helpful.

egs/wsj/s5/utils/prepare_extended_lang.sh
```diff
@@ -0,0 +1,182 @@
...
+        prob = log(oov_unigram_prob); 
+        for (x in dict) { $1 = prob; $2 = x; print $0; }
+      } 
+      if ($1 == "\\2-grams:") unigram = 0;
+    }' | change-lm-vocab -vocab $tmpdir/voc -lm - -write-lm - $srilm_opts | gzip -c > $tmpdir/lm_tmp.gz
+
+  if $to_carpa; then
+    echo "$(basename $0): Convert the LM to FST."
+    unk=`cat $extlang/oov.int`
+    bos=`grep -w "<s>" $extlang/words.txt | awk '{print $2}'`
```
danpovey: Better to use something like `awk '{if($1 == "<s>") print $2;}'`, in case there are words like `<s>foo`.

scripts/rnnlm/change_vocab.sh
```diff
@@ -0,0 +1,76 @@
...
+# Copyright 2018  Xiaohui Zhang
+
+# This script prepares a new rnnlm-dir commpatible with a new vocab from a provided word-list,
+# given an exisiting rnnlm-dir containing a trained rnnlm. Basically, we copy the feature 
+# embedding, a trained rnnlm and some config files from the old rnnlm-dir. And then we re-
+# generate the unigram_probs.txt (a fixed unigram prob is assigned to words out of the orignal vocab),
+# word_feats.txt and word embeddings.
+
+cmd=run.pl
+oov_unigram_prob=0.01
```
danpovey: This seems kind of high... more like 10^-6 is what I'd choose.   I don't know how critical this is though.

egs/wsj/s5/utils/prepare_extended_lang.sh
xiaohui-zhang: oh that's was the old comment created when I used a scaled version os the <unk>'s lm unigram prob before. I'll change it. BTW, when you said we shouldn't give the OOV-word prob to each individual OOV word, did you mean we should allow different probs for each OOV word? I think it's exactly what this script does: https://github.com/kaldi-asr/kaldi/blob/master/egs/babel/s5d/local/arpa2G.sh Actually, I just realized I, should probably remove the part dealing with expanding G.fst in my script and just use the above script for this purpose. Also I think we should move arpa2G.sh to utils level, e.g. utils/format_lm_extended.sh or even just modify utils/format_lm.sh to allow adding OOVs. What do you think? @jtrmal your opinion is also welcomed (basically I'm creating scripts for people to add words into L and G in test time).

danpovey: What I meant was that, regardless of whether the individual probs of the previously-unknown words are all the same or not, they should likely be smaller than the 'unk' class.  However, we have to be a bit careful with dividing the prob of the 'unk' class in the LM, between the previously-OOV words.  Different estimation formulas give very different values for the unk prob, that may or may not be appropriate here (and this depends on how the vocab and the data were prepared too),  so it might well be necessary to allow the user to manually override the unk-prob from the LM somehow, in case the value in the ARPA was unsuitable for some reason.

I do think that separating the 'G' parts of it from the rest of the 'lang' preparation is a good idea.  However, arpa2G.sh is not a good starting point.  That is a rather outdated script; these days, we do it more efficiently from the C++, as you can see from egs/ami/s5b/local/ami_format_data.sh:
```
gunzip -c "$arpa_lm" | \
  arpa2fst --disambig-symbol=#0 \
           --read-symbol-table=data/lang_test/words.txt - data/lang_test/G.fst
```
So no script in utils/ is needed just to format the LM.
I think what might be cleaner is to have some kind of script that would modify the ARPA file as part of a pipe, so you could do
```
gunzip -c "$arpa_lm" | \
  some_command_that_messes_with_unk_probs | \
  arpa2fst --disambig-symbol=#0 \
           --read-symbol-table=data/lang_test/words.txt - data/lang_test/G.fst
```


I think it probably would be a good idea to

xiaohui-zhang: Thanks! I agree. Regarding overriding the <unk> prob. I think it's an independent issue. As we discussed before we might not want to consider the prob of <unk> class in LM when adding new words, since we don't know how it's estimated. But we should allow the user to change it somewhere else. For now I'm doing this by scaling all arcs with output symbol <unk> in HCLG. Do you think we want to commit such a script under utils? 

Sure I think it's a good idea to create a script (Did you mean some local commands or a script to put under utils? I suggest a script since it might be long) to add OOV & probs to an ARPA file. I think the mechanism from arpa2G.sh might still be good: ask the user the provide an oov2prob file, and an --unk-fraction scalar scaling the OOV probs. These probs will be then ceilinged by the smallest unigram prob existing in the ARPA. What do you think? 

At the end, you mentioned "I think it probably would be a good idea to". to what?

danpovey: In `some_command_that_messes_with_unk_probs`, my intention was that it was some kind of script under utils/lang/.  I don't think applying a ceiling based on the smallest unigram prob in the ARPA is a good idea, as there are situations where, due to domain mismatch, the new OOV words might be quite probable.  For the rest of the design-- I suggest you come up with something, and we discuss afterward.
I don't know what I was going to say with 'I think it probably would be a good idea to'-- forget that.

xiaohui-zhang: done. I also added a perl script for inserting extra unigrams into ARPA lm.

egs/wsj/s5/utils/lang/add_unigrams_arpa.pl
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env perl
+
+# Copyright 2018  Xiaohui Zhang
+# Apache 2.0.
```
danpovey: I notice there are no scripts in this commit that use this.  Is there any example that it wouldn't be too disruptive or time-consuming to check in?

xiaohui-zhang: I've finished tuning the OOV recovery recipe (which used all scripts involved in this PR) for swbd-eval2000 & rt03. Basically, this recipe (unk-decoding + p2g + expanding lexicon & (RNN)LMs + 4-gram rescoring + RNNLM rescoring) results in much better CER on OOVs and matching overall WERs on eval2000/rt03 vs. a clean baseline without unk modeling (sorry, there's no improvement on overall WER, which makes sense since there are only hundreds of OOVs. Previously I mis-interpreted the improvements by pruned-RNNLM re-scoring as from my recipe.) The recipe is a bit ugly for now but you can skim it if you have time /export/b15/xzhang/swbd/s5c/oov_recovery.sh. I'll discuss more details with you on tmr's meeting. Besides this script, there are a few other changes needed to be committed: 1. modify lattice-arc-post so that the starting and ending state of each arc could be printed.  2. a script doing P2G based on sequitur is also needed.

xiaohui-zhang: I'm planning to do more experiments on other datasets with more OOVs though.

PR #2114, opened 2017-12-30
[WIP]  This branch contains some acoustic modeling improvements... still cleaning up.

danpovey:

PR #2248, opened 2018-03-01
add rnnlm script on tedlium+lm1b; add rnnlm rescoring results

keli78:


egs/tedlium/s5_r2/local/rnnlm/tuning/run_lstm_tdnn_with_lm1b.sh
```diff
@@ -0,0 +1,156 @@
...
+                           --special-words='<s>,</s>,<brk>,<unk>' \
+                           $dir/config/words.txt > $dir/config/features.txt
+fi
+
+cat >$dir/config/xconfig <<EOF
+input dim=$embedding_dim name=input
+relu-renorm-layer name=tdnn1 dim=$embedding_dim input=Append(0, IfDefined(-1))
+fast-lstmp-layer name=lstm1 cell-dim=$embedding_dim recurrent-projection-dim=$lstm_rpd non-recurrent-projection-dim=$lstm_nrpd
+relu-renorm-layer name=tdnn2 dim=$embedding_dim input=Append(0, IfDefined(-2))
+fast-lstmp-layer name=lstm2 cell-dim=$embedding_dim recurrent-projection-dim=$lstm_rpd non-recurrent-projection-dim=$lstm_nrpd
```
danpovey: I'll merge this but I think you should play with the l2-regularize option for both the fast-lstmp-layer and the relu-renorm-layers.  See my email on the list to that guy who had problems with long utterances, for tuning suggestions.   Right now there is no shrinkage or anything that would keep the LSTM parameters in the right range.

egs/tedlium/s5_r2/local/rnnlm/tuning/run_lstm_tdnn_with_lm1b.sh
```diff
@@ -0,0 +1,156 @@
...
+ngram_order=4 # approximate the lattice-rescoring by limiting the max-ngram-order
+              # if it's set, it merges histories in the lattice if they share
+              # the same ngram history and this prevents the lattice from 
+              # exploding exponentially
+pruned_rescore=true
+
+. ./cmd.sh
+. ./utils/parse_options.sh
+
+text=/export/a07/keli1/1-billion-word-language-modeling-benchmark/training-monolingual.tokenized.shuffled/news.en
```
danpovey: scripts that use external data like this need to be able to work directly from the downloaded data and need to have instructions how to download it, at the very least.  If you did any data preparation for this, then you need to figure out how to check that in-- possibly a script in local/, or possibly in this script.  Probably a script in local/ would make more sense.

vince62s:
@keli78 did you run and get these results with PR #2114 checked in ?

keli78:
@vince62s I think this PR #2114 is mainly about the new layer (factorized parameter matrices) which is not used in RNNLM in my case. So the answer is not yet. I can try it later.

vince62s:
My point was not about the use of that PR in the rnnlm itself but more in the Acoustic model when reporting the WER after lattice rescoring.

keli78:
I see. The acoustic model I used for reporting the WER doesn't have the new architecture. It's a chain model with backstitch. Considering tedlium is not a small dataset, I'm not sure whether this new architecture (which works pretty good on small dataset) can further improve WER a lot. @danpovey , how do you think? Should I have a try?

danpovey:
If you're talking about the TDNN-F architecture, it's not specifically for
small datasets, in fact it works great on Switchboard, which is about the
same size as Tedlium.
But our GPUs are pretty busy right now, I don't put a high priority on
running this.

Let me know whether this PR should be merged-- I thought there was some
outstanding issue with it.

keli78:
Got it. I may try the factorized model after the interspeech deadline (next Friday) when the GPUs are not very busy. I'll let you know when I push my newest changes in for the previous issues (probably by next week).

egs/tedlium/s5_r2/local/rnnlm/tuning/run_lstm_tdnn.sh
```diff
@@ -2,29 +2,52 @@
...
 epochs=20
 
-. ./cmd.sh
-. utils/parse_options.sh
-[ -z "$cmd" ] && cmd=$train_cmd
+# variables for lattice rescoring
+run_lat_rescore=true
+run_nbest_rescore=true
+decode_dir_suffix=rnnlm_tedlium
+ac_model_dir=/export/a12/ywang/kaldi/egs/tedlium/s5_r2/exp/chain_cleaned/tdnn_lstm1i_adversarial1.0_interval4_epoches7_lin_to_5_sp_bi
```
danpovey: There shouldn't be absolute pathnames in these scripts, it's supposed to work when someone runs the checked-in scripts.     Also, this tuning/ directory should contain numbered scripts with suffixes like _1a.sh, _1b.sh, etc, and there should be a link from one level above to here.

danpovey:
@keli78 can you please look into this again?  I'd like to get this merged.

egs/tedlium/s5_r2/local/rnnlm/tuning/run_lstm_tdnn.sh
keli78: sure, I'll look into this today.

keli78:
Hi @danpovey, this PR is ready to be merged. The rescoring results of rnnlm with the google one billion data will be updated soon.

PR #2257, opened 2018-03-06
IFN/ENIT Dataset Handwritten Recognition Example

babakrkb:


danpovey:
@hhadian, let me know if you've already reviewed this.  This seems to include a duplicate version of the iam setup-- it's not clear to me why that's needed.

babakrkb:
It was my mistake. My project was not up-to-date. I think the Problem was solved.

hhadian:
No, I have not reviewed it yet. I'll do it.

egs/ifnenit/v1/local/data_prep.sh
```diff
@@ -0,0 +1,142 @@
...
+export LC_ALL=C
+utils/prepare_lang.sh --num-sil-states 3 --num-nonsil-states 4 --position-dependent-phones false data/local/dict "<unk>" data/local/lang data/lang_pregdl_nolm
+export LC_ALL=$oldLC
+
+
+cut -d ' ' -f 2- data/train/text > data/local/lm/train.lines
+cat data/test/text | awk '{ for(i=2;i<=NF;i++) print $i;}' | sort -u >test_words.txt
+cat data/train/text | awk '{ for(i=2;i<=NF;i++) print $i;}' | sort -u >train_words.txt
+utils/filter_scp.pl --exclude train_words.txt test_words.txt > diff.txt
+cat diff.txt >> data/local/lm/train.lines
```
hhadian: please remove this line (and any other code that might leak any part of the test data into the LM or lexicon). I guess these lines were used for the initial tests and debugging.

hhadian: And move all the code related to LM training to local/train_lm.sh or maybe run.sh

egs/ifnenit/v1/local/format_lm_sri.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
```
hhadian: how is this different from `utils/format_lm_sri.sh`?

egs/ifnenit/v1/local/get_atb_pronun.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/python
```
hhadian: This script needs documentation and probably a better name

egs/ifnenit/v1/local/ienit-initialize.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+source cmd.sh
+source path.sh
+
+data_dir="/export/b01/babak/IFN-ENIT/ifnenit_v2.0p1e"
+
+export LC_ALL=en_US.utf8
+
+# Fetch transcriptions, invert images
```
hhadian: This needs more documentation. It doesn't seem to be inverting images.

egs/ifnenit/v1/local/map_to_rareA.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/python
```
hhadian: This needs documentation.

egs/ifnenit/v1/local/reduce-to-vocabulary.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/python
```
hhadian: This and other scripts in local/ need documentation. Please use underscores in the file name.

egs/ifnenit/v1/local/score.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
```
hhadian: how is this different from steps/scoring/score_kaldi_wer.sh?

egs/ifnenit/v1/local/score_combine.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
hhadian: Do you do system combination?

egs/ifnenit/v1/path.sh
```diff
@@ -0,0 +1,25 @@
...
+
+export LD_LIBRARY_PATH=/home/dpovey/libs:$KALDI_ROOT/src/chainbin:/usr/local/lib:$LD_LIBRARY_PATH
+
+export PATH=$KALDI_ROOT/tools/srilm/bin/:$KALDI_ROOT/tools/srilm/bin/i686-m64/:$KALDI_ROOT/src/bin:$KALDI_ROOT/src/chainbin:$KALDI_ROOT/src/chain:$KALDI_ROOT/src/nnet3bin:$KALDI_ROOT/src/lmbin:$KALDI_ROOT/tools/openfst/bin:$KALDI_ROOT/src/fstbin/:$KALDI_ROOT/src/gmmbin/:$KALDI_ROOT/src/featbin/:utils:$KALDI_ROOT/src/fgmmbin:$KALDI_ROOT/src/sgmm2bin:$KALDI_ROOT/src/nnetbin:$KALDI_ROOT/src/nnet2bin:$KALDI_ROOT/src/sgmmbin:$KALDI_ROOT/src/lm:$KALDI_ROOT/src/latbin:$PATH 
+# path to the directory in which the subset of RM corpus is stored
+
+export LC_ALL=en_US.utf8
+export LC_LOCALE_ALL=C
+
+export MONO_GAUSS="40000"
```
hhadian: These variables should not be here. They should be in run.sh. Actually, it's better not to use any variables and instead use the values in the corresponding commands.

egs/wsj/s5/utils/format_lm_sri.sh
```diff
@@ -91,4 +91,4 @@ fstisstochastic $out_dir/G.fst
 # we do expect the first of these 2 numbers to be close to zero (the second is
```
hhadian: I guess you can revert the changes to this file.

src/nnet3/nnet-example-utils.cc
```diff
@@ -816,6 +823,21 @@ void UtteranceSplitter::GetGapSizes(int32 utterance_length,
...
+    info->left_context = (config_.left_context_initial >= 0 ?
+                         config_.left_context_initial : config_.left_context);
+    info->right_context = (config_.right_context_final >= 0 ?
+                          config_.right_context_final : config_.right_context);
+
+    (*chunk_info).push_back(*info);
+
+  }
+  else {
   std::vector<int32> chunk_sizes;
```
hhadian: These lines need indentation. Also it should be ` } else {`.

egs/ifnenit/v1/local/add_sclite.sh
```diff
@@ -0,0 +1,34 @@
...
+#. parse_options.sh || exit 1;
+
+if [ $# -ne 3 ] && [ $# -ne 4 ]; then
+  echo "Usage: local/add_sclite.sh <score-dir> <words.txt> <ref-trn> [<ref-clevel-trn>]"
+  echo "  sclite evaluation on word and phoneme level for all *.tra files in score-dir"
+  exit 1;
+fi
+
+oldLC=$LC_ALL
+export LC_ALL="en_US.utf8"
```
danpovey: please clarify why you need to do this (if not, what breaks?)
I prefer if possible to have scripts not be dependent on any locale other than LC_ALL, but if it's an external script that's OK.

And please set LC_ALL to C in path.sh.  That is expected by the rest of Kaldi.

babakrkb: This script was redundant and I've removed it, but because I am working with unicode format of transcripts I should set LC_ALL to some string that contain "utf8". If I don't do this I get some error like this: `UnicodeDecodeError: 'ascii' codec can't decode byte 0xd8 in position 89: ordinal not in range(128)`

danpovey: If that is an error coming from some kind of python script, there are ways from the python script itself to resolve this.  That's better than relying on external environment variables.
Remember to use the python3 mechanisms for this (for new scripts, we prefer to explicitly use python3, via #!/usr/bin/env python3).

egs/ifnenit/v1/run.sh
```diff
@@ -0,0 +1,155 @@
...
+  echo "dict folder preparation"
+  local/prepare_dict.sh
+  utils/prepare_lang.sh --num-sil-states 3 --num-nonsil-states 4 --position-dependent-phones false data/local/dict "<unk>" data/local/lang data/lang_nolm
+fi
+
+if [ $stage -le 2 ]; then
+  # LM preparation
+  echo "LM preparation"
+  cp -R $data_dir/lang_nolm -T $data_dir/lang
+  local/prepare_lm.sh --ngram 1 $data_dir/train/text $data_dir/lang || exit 1;
```
hhadian: Could you please (instead of using `local/prepare_lm.sh`) make the grammar (G.fst) using `utils/make_unigram_grammar.pl` (and `fstcompile --isymbols=$lang/words.txt --osymbols=$lang/words.txt`) and remove `local/prepare_lm.sh`?

egs/ifnenit/v1/run.sh
```diff
@@ -0,0 +1,155 @@
...
+  echo "data preparation"
+  local/ienit_initialize.sh
+  local/prepare_data.sh
+fi
+
+if [ $stage -le 1 ]; then
+  # dict folder preparation
+  echo "dict folder preparation"
+  local/prepare_dict.sh
+  utils/prepare_lang.sh --num-sil-states 3 --num-nonsil-states 4 --position-dependent-phones false data/local/dict "<unk>" data/local/lang data/lang_nolm
```
hhadian: I guess it's better to call the lang you use for training `data/lang` and the one used for testing `data/lang_test` (similar to other Kaldi recipes)

egs/ifnenit/v1/run.sh
```diff
@@ -0,0 +1,155 @@
...
+           ## This relates to the queue.
+. utils/parse_options.sh  # e.g. this parses the --stage option if supplied.
+
+
+numSilStates=4
+numStates=8
+num_gauss=10000
+
+numLeavesTri=500
+numGaussTri=20000
```
hhadian: please remove all these variables except `nj` and `stage`.

egs/ifnenit/v1/path.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# path to Kaldi's root directory
```
hhadian: Please fix this file (make it look like the one in other examples in egs/)

egs/ifnenit/v1/local/replace_arabic_punctuation.py
```diff
@@ -0,0 +1,28 @@
...
+		     .replace("؛", ":")
+		     .replace("؟", "?")
+		     .replace("–", "-")
+		     .replace("‘", "'")
+			 .replace("[", "(")
+		     .replace("{", "(")
+		     .replace("﴾", "(")
+		     .replace("]", ")")
+		     .replace("}", ")")
+		     .replace("}", "﴿")
```
hhadian: Shouldn't this line be `.replace("﴿", ")")` ?

egs/ifnenit/v1/local/make_feature_vect.py
```diff
@@ -0,0 +1,83 @@
...
+        if im_x >= (28 + (20*i)) and im_x <= (28 + (20*(i+1))):
+           padding_x = (30 + (20*(i+1))) - im_x
+           padding_y = im_y
+        else:
+           continue
+    # padding_x = max(5,int((args.padding/100)*im.shape[1]))
+    # padding_y = im.shape[0]
+    im_pad = np.concatenate((255 * np.ones((padding_y, math.ceil(1.0 * padding_x / 2)) , dtype=int), im), axis=1)
+    im_pad1 = np.concatenate((im_pad,255 * np.ones((padding_y, int(1.0 * padding_x / 2)), dtype=int)), axis=1)
+    # im_pad = np.concatenate((255 * np.ones((padding_y,5) , dtype=int), im), axis=1)
```
hhadian: remove such comments here and everywhere else

egs/ifnenit/v1/local/ienit_initialize.sh
```diff
@@ -0,0 +1,35 @@
...
+    cp $data_dir/data/$folder/tru/*.tru data/tru
+    ls $data_dir/data/$folder/png/*.png >> tmp.flist
+  done
+  cat tmp.flist | xargs -n 1 -IBLA basename BLA '.png' > tmp.uttids
+  cat tmp.flist | sed 's/\/png\/\([a-z0-9_]\+\).png$/\/tru\/\1.tru/' | xargs egrep '^LBL:' | iconv -f 'cp1256' -t 'UTF-8' | python3 local/remove_diacritics.py | cut -d':' -f5- | cut -d';' -f1 | paste -d' ' tmp.uttids - > data/text.$set
+  folders="set_d"
+done
+rm tmp.flist tmp.uttids
+
+export LC_ALL=C
```
hhadian: I guess it's good to have a newline at the end of the files.

egs/ifnenit/v1/RESULTS.txt
```diff
@@ -0,0 +1,3 @@
+WER%                    abc-d       abcd-e      abcde-f     abcde-s 
+cnn1a                   1.46%       4.18%       5.79%       9.15%
+cnn_chainali_1b         1.96%       3.28%       4.80%       7.63%
```
hhadian: Could you please update the results (after your changes)?

egs/ifnenit/v1/run.sh
danpovey: There is no need to have separate lang vs. lang_test directories, if you only use one language model for testing.  The distinction is historical and has no real use any more.

egs/ifnenit/v1/RESULTS.txt
babakrkb: Actually the `diff.txt` file was empty. So every test words are in the train set. I've decoded `abc-e` version and the result was the same. I think result would be the same for other versions too.

egs/ifnenit/v1/local/add_sclite.sh
babakrkb: I've tried every solution that I've found but I still get same error, even when I set `PYTHONIOENCODING = utf8`. I think this error might be of terminal encoding then I should set `LC_ALL="en_US.utf8` for those specific lines of script that have a pipeline for processing some UTF-8 text.

jtrmal: Can you please provide small example (expected input, expected output and script) that needs LC_ALL to be set? I'll have a look at it -- I don't have time to debug this from the recipe ATM.
The problem with system locale (LC_*) is that it needs locale database for that given combination to be installed and compiled and it's not guaranteed -- so unless there are very specific reasons (and so far it does not look like they are), it's better to fix the python script (which will be more portable).
y.

babakrkb: Thank you @jtrmal . I've removed every LC_ALL statements.

egs/ifnenit/v1/README.txt
```diff
@@ -0,0 +1 @@
+You can fill the registration form in http://www.ifnenit.com/ to download dataset. Then you sould set the path of the dataset in the beginnig of the "local/ienit-initialize.sh" and "local/process_data.py"
```
hhadian: Please change this to something like:
```
This directory contains example scripts for handwriting recognition on
the Arabic IFN/ENIT dataset: http://www.ifnenit.com
You'll need to register at their website to be able to download the dataset.
```

egs/ifnenit/v1/cmd.sh
```diff
@@ -0,0 +1,27 @@
+# "queue.pl" uses qsub.  The options to it are
```
hhadian: Please make this file like iam/v1/cmd.sh and update `$train_cmd` and `$decode_cmd` to `$cmd` in your scripts.

egs/ifnenit/v1/config/ligatures
```diff
@@ -0,0 +1,69 @@
...
+llmLM lM lM mM
+lmLE lB mE
+lmLE lM mE
+lmLM lB mM
+lmLM lM mM
+mjLE mB jE
+mjLE mM jE
+mjLM mB jM
+mjLM mM jM
+mohammad maB haM maM daE
```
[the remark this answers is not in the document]
hhadian: Is this line an error (it doesn't look like the other lines)?

egs/ifnenit/v1/local/ienit_initialize.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+. ./cmd.sh
+. ./path.sh
+
+# This script is originally from qatip project (http://qatsdemo.cloudapp.net/qatip/demo/)
```
hhadian: @danpovey, does this need some kind of copyright statement?

egs/ifnenit/v1/local/ienit_initialize.sh
```diff
@@ -0,0 +1,31 @@
...
+# of Qatar Computing Research Institute (http://qcri.qa/)
+
+data_dir="/export/b01/babak/IFN-ENIT/ifnenit_v2.0p1e"
```
hhadian: This should be set via a command line option. 
Rename the variable to `database_dir`, move the line to run.sh, and pass it to related scripts from there.

egs/ifnenit/v1/local/make_feature_vect.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
```
hhadian: Please rename this file to `make_features.py`

egs/ifnenit/v1/local/prepare_data.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+. ./cmd.sh
+. ./path.sh
+
+# To be run from one directory above this script.
+# Creat text, utt2spk, spk2utt, images.scp, and feats.scp for test and train.
+
+database='/export/b01/babak/IFN-ENIT/ifnenit_v2.0p1e/data'
```
hhadian: this should be set via a command line option

egs/ifnenit/v1/local/prepare_data.sh
```diff
@@ -0,0 +1,45 @@
...
+
+  local/process_data.py $database data/$set --dataset $set --model_type word || exit 1
+  sort data/$set/images.scp -o data/$set/images.scp
+  sort data/$set/utt2spk -o data/$set/utt2spk
+
+  utils/utt2spk_to_spk2utt.pl data/$set/utt2spk > data/$set/spk2utt
+
+  mkdir -p data/{train,test}/data
+
+  local/make_feature_vect.py data/$set --scale-size 40 | \
```
hhadian: I guess it's good to change the `--scale-size` option to `--feat-dim` to be consistent with other OCR recipes.

egs/ifnenit/v1/run.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
+
+stage=0
+nj=8
+data_dir=data
+exp_dir=exp
```
hhadian: remove these two variables and hardcode them in the code.

src/nnet3/nnet-example-utils.cc
```diff
@@ -816,30 +823,43 @@ void UtteranceSplitter::GetGapSizes(int32 utterance_length,
...
+      info = new ChunkTimeInfo;
+      info->first_frame = 0;
+      info->num_frames = utterance_length;
+      info->left_context = (config_.left_context_initial >= 0 ?
+                          config_.left_context_initial : config_.left_context);
+      info->right_context = (config_.right_context_final >= 0 ?
+                            config_.right_context_final : config_.right_context);
+
+      (*chunk_info).push_back(*info);
+  } else {
```
hhadian: The indentation of the `else` block does not look right to me.

src/nnet3/nnet-example-utils.cc
```diff
@@ -816,30 +823,43 @@ void UtteranceSplitter::GetGapSizes(int32 utterance_length,
...
-                          config_.right_context_final : config_.right_context);
-    t += chunk_sizes[i];
+    int32 t = 0;
+    if (config_.num_frames_str == "-1" ) {
+      ChunkTimeInfo *info;
+      info = new ChunkTimeInfo;
+      info->first_frame = 0;
+      info->num_frames = utterance_length;
+      info->left_context = (config_.left_context_initial >= 0 ?
+                          config_.left_context_initial : config_.left_context);
```
hhadian: fix indentation (also for L835)

src/nnet3/nnet-example-utils.cc
```diff
@@ -816,30 +823,43 @@ void UtteranceSplitter::GetGapSizes(int32 utterance_length,
...
+      info->right_context = (config_.right_context_final >= 0 ?
+                            config_.right_context_final : config_.right_context);
+
```
hhadian: I guess you could remove this empty line

src/nnet3/nnet-example-utils.cc
```diff
@@ -816,30 +823,43 @@ void UtteranceSplitter::GetGapSizes(int32 utterance_length,
...
-    t += gaps[i];
-    ChunkTimeInfo &info = (*chunk_info)[i];
-    info.first_frame = t;
-    info.num_frames = chunk_sizes[i];
-    info.left_context = (i == 0 && config_.left_context_initial >= 0 ?
-                         config_.left_context_initial : config_.left_context);
-    info.right_context = (i == num_chunks - 1 && config_.right_context_final >= 0 ?
-                          config_.right_context_final : config_.right_context);
-    t += chunk_sizes[i];
+    int32 t = 0;
```
hhadian: I guess the main issue with how this looks is that you are using a 4-space indentation instead of 2-space. Please re-configure your editor and make sure the indentations are right.

egs/ifnenit/v1/run.sh
```diff
@@ -0,0 +1,145 @@
...
+  local/chain/run_cnn_1a.sh --stage 0 \
+   --gmm tri3 \
+   --ali tri3_ali \
+   --nnet3_affix $nnet3_affix \
+   --affix $affix \
+   --lang_test lang
+fi
+
+if [ $stage -le 13 ]; then
+  local/chain/run_cnn_chainali_1b.sh --stage 0 \
```
hhadian: Make these options default and remove them from here (also for stage 12). See run.sh in iam or uw3 and try to follow their style. Remove all those affix variables as well.

egs/ifnenit/v1/config/ligatures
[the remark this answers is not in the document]
babakrkb: No it is not. The only difference is that we can assign a meaningful name for this ligature and we can not for the others.

egs/ifnenit/v1/run.sh
babakrkb: I have followed their style in my scripts, but I think there is some problem in iam/v1/local/run_cnn_chainali_1b.sh options. (It won't work with default options).
If you check the options you can realize that the {affix} option which is used in line 30 should not be the same {affix} which is used in line 71. The first one is for the alignment model but the second is for output model.
I have changed my code a little to handle this issue.

babakrkb:
Yes I've got your point, I did that in my last commit. I only wanted to
mention that problem.

egs/ifnenit/v1/run.sh
babakrkb: Yes I've got your point, I did that in my last commit. I only wanted to
mention that problem.

egs/ifnenit/v1/local/chain/run_cnn_chainali_1b.sh
```diff
@@ -0,0 +1,231 @@
...
+    steps/nnet3/chain/gen_topo.py $nonsilphonelist $silphonelist >$lang/topo
+  fi
+fi
+
+if [ $stage -le 2 ]; then
+  # Get the alignments as lattices (gives the chain training more freedom).
+  # use the same num-jobs as the alignments
+  steps/nnet3/align_lats.sh --nj $nj --cmd "$cmd" \
+                            --scale-opts '--transition-scale=1.0 --self-loop-scale=1.0' \
+                            ${train_data_dir} data/$lang_test $chain_model_dir $lat_dir
```
hhadian: Please change `data/$lang_test` here to `data/lang`. Also in other places except for graph making (stage 6).
Also pass `--acoustic-scale 1.0` (see the same script in IAM). 
You can remove chainali_1a recipe from the PR if you are not using it (rename this script to chainali_1a in that case).

@danpovey, are you planning to merge this PR?

src/nnet3/nnet-example-utils.cc
```diff
@@ -341,11 +344,13 @@ UtteranceSplitter::UtteranceSplitter(const ExampleGenerationConfig &config):
     total_num_utterances_(0), total_input_frames_(0),
     total_frames_overlap_(0), total_num_chunks_(0),
     total_frames_in_chunks_(0) {
-  if (config.num_frames.empty()) {
-    KALDI_ERR << "You need to call ComputeDerived() on the "
+  if (config.num_frames_str != "-1") {
```
danpovey: these changes need to be documented in the header.  To this:
```
   po->Register("num-frames", &num_frames_str, "Number of frames with labels "
                "that each example contains (i.e. the left and right context "
                "are to be added to this).  May just be an integer (e.g. "
                "--num-frames=8), or a principal value followed by "
                "alternative values to be used at most once for each utterance "
                "to deal with odd-sized input, e.g. --num-frames=40,25,50 means "
                "that most of the time the number of frames will be 40, but to "
                "deal with odd-sized inputs we may also generate egs with these "
                "other sizes.  All these values will be rounded up to the "
                "closest multiple of --frame-subsampling-factor.");
```
add the text: "As a special case, --num-frames=-1 means 'don't do any splitting'."

egs/ifnenit/v1/run.sh
```diff
@@ -0,0 +1,138 @@
...
+
+if [ $stage -le 2 ]; then
+  # LM preparation
+  echo "LM preparation"
+  cat data/train/text | cut -d' ' -f2- | utils/make_unigram_grammar.pl | \
+    fstcompile --isymbols=data/lang/words.txt --osymbols=data/lang/words.txt > data/lang/G.fst
+fi
+
+if [ $stage -le 3 ]; then
+  steps/train_mono.sh --nj $nj \
```
danpovey: please don't use so much vertical space.

babakrkb:
I've done adding documentation and some changes to `run.sh` and `run_cnn_chainali_1a.sh`.

PR #2265, opened 2018-03-08
Fix subsampling factor in nnet3 egs

vimalmanohar:


vimalmanohar:
I guess this should be committed even though it affects performance a little.

PR #2276, opened 2018-03-14
fix beamformit installation

sw005320:
- fix beamformit installation pointed out in the CHiME-5 google group.


jtrmal:
Thanks, I'm merging this, as it is my overlooking.

PR #2279, opened 2018-03-14
[src] Fix to nnet3 looped compilation bug (thanks: Alim Misbulah)

danpovey:


danpovey:
Please try the fix in #2286.  Note: it's very possible that you will get nonsense results when decoding your already-trained model; let me know.  Unfortunately this was a prior bug that, when fixed, may affect some people... not too many of them, though, it's architecture-specific.

PR #2286, opened 2018-03-16
[src] Fix to nnet3 compilation issue affecting BLSTMP.  Thanks: Alim Misbullah

danpovey:


danpovey:
WARNING: This bug-fix may affect inference using already-trained models, most likely with architectures that involve backward-in-time recurrence and dim-range nodes (e.g. BLSTMP).  The bug, which has existed from the start of nnet3, likely caused these architectures to use the wrong time-offset of some inputs.  We never used these types of models much (perhaps because of this bug), so hopefully this won't be a big inconvenience.

misbullah:
It can work well now.

Thanks a lot for help.

Best,
Alim

PR #2288, opened 2018-03-17
[src] Make CachingOptimizingCompiler thread safe.  Thx: Arseniy Gorin

danpovey:


danpovey:
This code isn't really intended for use with GPUs.
In any case I'm not sure what you mean-- perhaps there is some bug in a
particular version of the CUDA library that you are talking about.  CUDA is
definitely supposed to support multi-threaded programs.

btiplitz:
Oh I misread the comment. Ignore what I said

src/nnet3/nnet-am-decodable-simple.cc
```diff
@@ -248,7 +248,7 @@ void DecodableNnetSimple::DoNnetComputation(
   request.outputs.resize(1);
   request.outputs[0].Swap(&output_spec);
 
-  const NnetComputation *computation = compiler_.Compile(request);
+  std::shared_ptr<const NnetComputation> computation = compiler_.Compile(request);
```
gorinars: please apply the same fix in nnet3-xvector-compute.cc:47

PR #2289, opened 2018-03-17
Add support for --use-gpu=wait to nnet3 scripts

hhadian:
@danpovey, could you please see if this is how you meant for this to be done?
If yes, I'll make changes to the rest of the nnet3 scripts.

### DIFF
--- a/egs/wsj/s5/steps/libs/nnet3/train/common.py
+++ b/egs/wsj/s5/steps/libs/nnet3/train/common.py
@@ -942,7 +942,9 @@ class CommonParser(object):
                                  help="Script to launch egs jobs")
         self.parser.add_argument("--use-gpu", type=str,
                                  choices=["true", "false", "yes", "no", "wait"],
-                                 help="Use GPU for training", default="true")
+                                 help="Use GPU for training. "
+                                 "Note 'true' and 'false' are deprecated.",
+                                 default="yes")
         self.parser.add_argument("--cleanup", type=str,
                                  action=common_lib.StrToBoolAction,
                                  choices=["true", "false"], default=True,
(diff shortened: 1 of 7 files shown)


egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -228,7 +228,9 @@ def process_args(args):
         args.transform_dir = args.lat_dir
     # set the options corresponding to args.use_gpu
     run_opts = common_train_lib.RunOpts()
-    if args.use_gpu:
+    if args.use_gpu in ["true", "false"]:
+        args.use_gpu = {'true': 'yes', 'false': 'no'}[args.use_gpu]
```
danpovey: I'd prefer if you wrote this as
 `args.gpu = ("yes" if args.gpu == "true" else "no")`
as it's easier for a human to parse.
Otherwise this is OK.

egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -941,9 +941,8 @@ def __init__(self,
                                  action=common_lib.NullstrToNoneAction,
                                  help="Script to launch egs jobs")
         self.parser.add_argument("--use-gpu", type=str,
-                                 action=common_lib.StrToBoolAction,
-                                 choices=["true", "false"],
-                                 help="Use GPU for training", default=True)
+                                 choices=["true", "false", "yes", "no", "wait"],
```
danpovey: You should probably mention that "true" and "false" are deprecated.

danpovey:
Thanks-- merging.

PR #2290, opened 2018-03-18
xvector extractor cache config

gorinars:
Following the discussion in
https://groups.google.com/forum/#!topic/kaldi-help/ROtSHHe3Z_I

Using larger cache speeds up extraction if equal length segments have already been processed

### DIFF
--- a/egs/sre08/v1/sid/nnet3/xvector/extract_xvectors.sh
+++ b/egs/sre08/v1/sid/nnet3/xvector/extract_xvectors.sh
@@ -14,9 +14,11 @@
 # Begin configuration section.
 nj=30
 cmd="run.pl"
-chunk_size=-1 # The chunk size over which the embedding is extracted.
-              # If left unspecified, it uses the max_chunk_size in the nnet
-              # directory.
+
+cache_capacity=64 # Cache capacity for x-vector extractor
+chunk_size=-1     # The chunk size over which the embedding is extracted.
+                  # If left unspecified, it uses the max_chunk_size in the nnet
+                  # directory.
 use_gpu=false
 stage=0
 
@@ -34,6 +36,7 @@ if [ $# != 3 ]; then
   echo "  --use-gpu <bool|false>                           # If true, use GPU."
   echo "  --nj <n|30>                                      # Number of jobs"
   echo "  --stage <stage|0>                                # To control partial reruns"
+  echo "  --cache-capacity <n|64>                          # To speed-up xvector extraction"
   echo "  --chunk-size <n|-1>                              # If provided, extracts embeddings with specified"
   echo "                                                   # chunk size, and averages to produce final embedding"
 fi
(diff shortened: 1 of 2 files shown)


gorinars:
Travis build failed with a strange error

`./matrix/libkaldi-matrix.so: file not recognized: File truncated`

I feel like, if possible, just re-running it would solve the issue. If I should re-create this PR to trigger travis, please let me know

jtrmal:
I have restarted the build.
y.

danpovey:
the build isn't necessary when code is not changed.  it's a travis issue.  I'll merge.

danpovey:
I see that the default in the code is 64.  Perhaps it would make sense to use a larger value as the default in the script, assuming this makes the code run faster?

gorinars:
64 is the setting we use now by default. 
I usually use 475 for chunk-size=500 to make sure computer is compiled only once for a given segment size. For SRE the right value would be 10000, but I'm not sure about the memory consumption then.

danpovey:
Hm, that would be quite a bit of I/O for each job, but I guess it's OK if
it gives a speedup in practice.  If you just set the limit to a large value
it would work for any setup, but IMO in the longer term the right fix is to
change the setup so that a relatively small number of segment sizes.
@david-ryan-snyder, I don't know if there is progress on that front?

david-ryan-snyder:
@danpovey, those issues are both important but actually orthogonal. @LvHang PR speeds up training. @gorinars PR only affects the x-vector system after it's been trained. To answer your question... I still haven't made progress on it, but I haven't forgotten about it either.

BTW, @gorinars, feel free to add your name to the author list on those files, so that your contribution is more visibly credited.

david-ryan-snyder:
It occurred to me I might've misunderstood what you meant, @danpovey.

If you're talking about reducing the number of segment-lengths during x-vector extracting, I think the issue is basically solved already (especially given @gorinars PR).

In diarization, we already extract embeddings from fixed-length chunks (e.g., 150 length chunks). For speaker recognition, we usually set a (near) infinite right context, because we found that it's empirically better to extract embeddings from the entire recording. However, if speed is an issue, the user can set chunk-length=1000 or something like that, and the binary will extract embeddings from 1000 frame chunks, and average them to produce the final embedding.

gorinars:
@david-ryan-snyder that exactly what led to the PR issue.
I was setting chunk-length=1000, but the actual chunks are sometimes smaller, like 900, 101, etc. Re-compilation for each of them was slowing down the system significantly.
So we finally ended up with stuff implemented in nnet3-xvector-compute-parallel.cc in #2303 by passing the pre-compiled cache for all segment lengths and increasing the cache size.

gorinars:
@david-ryan-snyder thanks for remark about adding the name. Not sure for this particular change - it is really a very small modification.

danpovey:
I'll merge, it won't break anything.

PR #2292, opened 2018-03-18
[WIP] Separate background noises max count from foreground noises max count

nshmyrev:
When wav files are longer than 20 minutes the current implementation adds about 10 background music noises at once for the full length of audio effectively adding the white noise. The proposal is to separate parameters for background and for foreground and control maximum background noises with a separate option to ensure you do not have more than 1-2 of the total.

Another approach probably would be to avoid repeat of the background noise for the full audio length.

Please let me know what do you think.

### DIFF
--- a/egs/wsj/s5/steps/data/reverberate_data_dir.py
+++ b/egs/wsj/s5/steps/data/reverberate_data_dir.py
@@ -61,6 +61,8 @@ def GetArgs():
     parser.add_argument("--noise-smoothing-weight", type=float, default = 0.3,
                         help="Smoothing weight for the noise probabilties, e.g. 0 <= p <= 1. If p = 0, no smoothing will be done. "
                         "The noise distribution will be mixed with a uniform distribution according to the smoothing weight")
+    parser.add_argument("--max-bg-noises", type=int, default = 1,
+                        help="This controls the maximum number of point-source noises that could be added to a recording according to its duration")
     parser.add_argument("--max-noises-per-minute", type=int, default = 2,
                         help="This controls the maximum number of point-source noises that could be added to a recording according to its duration")
     parser.add_argument('--random-seed', type=int, default=0, help='seed to be used in the randomization of impulses and noises')
@@ -113,6 +115,9 @@ def CheckArgs(args):
 
     if args.max_noises_per_minute < 0:
         raise Exception("--max-noises-per-minute cannot be negative")
+
+    if args.max_bg_noises < 0:
+        raise Exception("--max-bg-noises cannot be negative")
 
     if args.source_sampling_rate is not None and args.source_sampling_rate <= 0:
         raise Exception("--source-sampling-rate cannot be non-positive")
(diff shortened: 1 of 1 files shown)


danpovey:
@david-ryan-snyder  and @vimalmanohar, please comment.

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -63,6 +63,8 @@ def GetArgs():
                         "The noise distribution will be mixed with a uniform distribution according to the smoothing weight")
     parser.add_argument("--max-noises-per-minute", type=int, default = 2,
```
vimalmanohar: This looks good. But perhaps the name and the comments can be changed to reflect that these are foreground noises. e.g. --max-fg-noises-per-minute and so on.

nshmyrev: This will make all recipe scripts incompatible but ok, I'll work on that.

vimalmanohar: Ok. Perhaps you can add the new option for future use and temporarily keep --max-noises-per-minute with a message that it is deprecated. But use the old option if that is supplied. But @danpovey might have other opinions on that.

danpovey:
I think in this case it's OK to just ignore the old --max-noises-per-minute option when deciding the number of background noises, and instead specify the number of background noises as a number.  The way it was done before didn't really make sense.  As long as @david-ryan-snyder agrees.
You can rename --max-noises-per-minute to --num-fg-noises-per-minute but retain the old name as a deprecated alias; I think that's possible in the Python options parsing somehow.  (The "max" wasn't really appropriate either, as it wasn't a maximum; it specified the actual number, subject to the probability of adding noises in the first place.)

As for the --max-bg-noises: again, it isn't really a maximum, it's a number, so I think better to call it --num-bg-noises instead.

 The options mention isotropic noises and pointsource.  Those seem to be the same same as background and foreground noises?  I'd prefer to have consistent notation if so-- perhaps we can register 2 versions of the same option for those, too; say the isotropic and pointsource versions are deprecated; and replace "isotropic"/"iso" and "pointsource" with "foreground" and "background", and normalize other variable names too, for instance, change max_noises_recording to num_noises_foreground.  I know this is a pain but it could probably be done fairly mechanically.

We might want to re-test the Aspire setup before committing the change though.

david-ryan-snyder:
> The way it was done before didn't really make sense. As long as @david-ryan-snyder agrees

This was actually @tomkocse code, not mine--I think you confused this with a similar augmentation script that I wrote. Let's get Tom's input.

danpovey:
@david-ryan-snyder, I knew it's not your script, I just thought you might have a comment as you've worked with these kinds of things.

@nshmyrev, if you have time, please go ahead and make as many of the changes I asked for as you think you have time to do.  I'll ask @tomkocse to make the rest, if needed.
It looks like to add aliases, you have to just change e.g. add_argument("--foo", ...) to add_argument("--foo", "--alias-for-foo", ... ), where the target name will be derived from the first string ("foo" in this case).

nshmyrev:
Thanks, I'll work on it next week.

david-ryan-snyder:
@nshmyrev were you still planning on merging this one?

As Dan mentioned, you could just add in the essential things (e.g., an alias for num-bg-noises, num-fg-noises so existing recipes don't break) and someone else could improve the remaining option names later on.

nshmyrev:
@david-ryan-snyder sorry, still planning to do this, just a bit distracted by other tasks.

danpovey:
@david-ryan-snyder can you please review this?

david-ryan-snyder:
Thanks @nshmyrev!

I will take a look at this in a couple of days.

PR #2293, opened 2018-03-18
fix reading wavs produced on iPhones with extra RIFF tags.

vesis84:
- Apple creates wavs with extra RIFF tags for optimizing memory layout,
  - instead of standard 'RIFF,WAVE,fmt ,data'
    we observed a file with 'RIFF,WAVE,JUNK,fmt ,FLLR,data'
  - according to RIFF specs, we should skip the unknown tags, which is
    what the fix does, as it skips the 'JUNK', the skip of 'FLLR' was
    already in the code.
  - similar problem discussed in: https://stackoverflow.com/questions/6284651/avaudiorecorder-doesnt-write-out-proper-wav-file-header

- Please let me know if you'd like to do further refactoring of the code,
  i can imagine there could be a method 'ProceedTo4ByteTag("ABCD")',
  which would skip tags until it reaches 'ABCD' and returns number of bytes read...
- On the other hand, the code in this PR already solves the issue, and the changes are the smallest possible.

### DIFF
--- a/src/feat/wave-reader.cc
+++ b/src/feat/wave-reader.cc
@@ -132,7 +132,23 @@ void WaveInfo::Read(std::istream &is) {
   uint32 riff_chunk_read = 0;
   riff_chunk_read += 4;  // WAVE included in riff_chunk_size.
 
-  reader.Expect4ByteTag("fmt ");
+  // Possibly skip any RIFF tags between 'WAVE' and 'fmt '.
+  // Apple devices produce a filler tag 'JUNK' for memory alignment.
+  reader.Read4ByteTag();
+  riff_chunk_read += 4;
+  while (strcmp(reader.tag,"fmt ") != 0) {
+    uint32 filler_size = reader.ReadUint32();
+    riff_chunk_read += 4;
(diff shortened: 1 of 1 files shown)


src/feat/wave-reader.cc
```diff
@@ -132,7 +133,25 @@ void WaveInfo::Read(std::istream &is) {
...
 
-  reader.Expect4ByteTag("fmt ");
+  // Skip any RIFF tags between 'WAVE' and 'fmt '.
+  // Apple devices produce filler tag 'JUNK' for memory alignment.
+  reader.Read4ByteTag();
+  riff_chunk_read += 4;
+  while (strcmp(reader.tag,"fmt ") != 0) {
+    uint32 filler_size = reader.ReadUint32();
+    riff_chunk_read += 4;
+    KALDI_WARN << "Skipping RIFF tag '" << reader.tag
```
danpovey: I don't imagine you want that warning there permanently...

vesis84:
Okay, the warning-print is now removed. The code is tested on both the Apple and non-Apple wavefile...

PR #2295, opened 2018-03-20
[src] Fix issue with CUDA device initialization if 'wait' specified.  Thx: @olix20

danpovey:


src/cudamatrix/cu-device.cc
```diff
@@ -65,6 +65,13 @@ static bool GetCudaContext(int32 num_gpus, std::string *debug_str) {
   if (cudaFree(0) == 0)
     return true;
 
+  if (debug_str == NULL) {
+    // this function called when "wait" is specified... don't get debug_str,
```
galv: You might want to reword this comment... it's rather funky.

PR #2298, opened 2018-03-21
fixing LM issue in CHiME5

jtrmal:
due working with several channels having the same transcription (several microphone recording the same conversation), the training data look very artificial from the POV of LM methods, so we have to remove the duplicate transcriptions. This script that does this wasn't modified after changing the format of the utterance ids.


kamo-naoyuki:
The development text may need same changes.

https://github.com/jtrmal/kaldi/blob/957826603b4c031d8010589ee12eaf1f6c855d1e/egs/chime5/s5/local/train_lms_srilm.sh#L114-L116

PR #2299, opened 2018-03-22
[tools, extras] morfessor installation script

jty016:
added install_morfessor.sh and its symbolic link

### DIFF
--- a/tools/extras/install_morfessor.sh
+++ b/tools/extras/install_morfessor.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+# Copyright  2017  Atlas Guide (Author : Lucas Jo)
+#
+# Apache 2.0
+#
+
+echo "#### installing morfessor"
+dirname=morfessor
+if [ ! -d ./$dirname ]; then
+	mkdir -p ./$dirname
+	git clone https://github.com/aalto-speech/morfessor.git morfessor ||
+		{
+			if [ $? -ne 0 ]; then
+				echo  >&2 "$0: Error git clone operation "
+				echo  >&2 "  Failed in cloning the github repository (https://github.com/aalto-speech/morfessor.git)"
+				echo  >&2 "  Download morfessor-2.0.1.tar.gz instead"
+				wget http://morpho.aalto.fi/projects/morpho/morfessor-2.0.1.tar.gz
+				tar -zxvf morfessor-2.0.1.tar.gz
+				mv Morfessor-2.0.1/ $dirname
+			fi
+		}
+fi
+
+# local installation
+site_packages_dir=$(python -m site --user-site | grep -oE "lib.*")
+wd=`pwd`
+wd=`readlink -f $wd || pwd`
+export MORFESSOR="$wd/morfessor"
+export PYTHONPATH="${PYTHONPATH:-}:$MORFESSOR/${site_packages_dir}"
+(
+cd $dirname
+mkdir -p $site_packages_dir
+python setup.py install --prefix `pwd` --record files.txt
+)
+wait
+
+# env.sh setup
+site_packages_dir=$(cd morfessor; find ./lib{,64} -type d -name site-packages | head -n 1)
+(
+  set +u
+  [ ! -z "${MORFESSOR}" ] && \
+    echo >&2 "morfessor variable is aleady defined. undefining..." && \
+    unset MORFESSOR
+
+  [ -f ./env.sh ] && . ./env.sh
+
+  [ ! -z "${MORFESSOR}" ] && \
+    echo >&2 "MORFESSOR config is already in env.sh" && exit
+
+  wd=`pwd`
+  wd=`readlink -f $wd || pwd`
+
+  echo "export MORFESSOR=\"$wd/morfessor\""
+  echo "export PATH=\"\$PATH:\${MORFESSOR}/bin\""
+  echo "export PYTHONPATH=\"\${PYTHONPATH:-}:\$MORFESSOR/${site_packages_dir}\""
+) >> env.sh
+
+echo >&2 "installation of MORFESSOR finished successfully"
+echo >&2 "please source tools/env.sh in your path.sh to enable it"
+echo >&2 'when uninstall it, try: sudo rm $(cat ./morfessor/files.txt)'
--- a/tools/extras/install_morfessor.sh
+++ b/tools/extras/install_morfessor.sh
@@ -7,18 +7,18 @@
 echo "#### installing morfessor"
 dirname=morfessor
 if [ ! -d ./$dirname ]; then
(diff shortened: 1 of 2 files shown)


tools/extras/install_morfessor.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+# Copyright  2017  Atlas Guide (Author : Lucas Jo)
+#
+# Apache 2.0
+#
+
+echo "#### installing morfessor"
+dirname=morfessor
+if [ ! -d ./$dirname ]; then
+	mkdir -p ./$dirname
```
jtrmal: could you please change the tabs to spaces (each tab = two spaces)?

tools/install_morfessor.sh
```diff
@@ -0,0 +1 @@
+extras/install_morfessor.sh
```
jtrmal: I think you don't have to add this symlink (that is done only for historical reasons for some packages)

tools/extras/install_morfessor.sh
```diff
@@ -0,0 +1,60 @@
...
+				echo  >&2 "  Download morfessor-2.0.1.tar.gz instead"
+				wget http://morpho.aalto.fi/projects/morpho/morfessor-2.0.1.tar.gz
+				tar -zxvf morfessor-2.0.1.tar.gz
+				mv Morfessor-2.0.1/ $dirname
+			fi
+		}
+fi
+
+# local installation
+site_packages_dir=$(python -m site --user-site | grep -oE "lib.*")
```
jtrmal: typically we try to not to install the packages system-wise (user-wise) -- we keep the libs in kaldi/tools/<package> and just set up the shell variables in env.sh (PYTHONPATH). Otherwise, you could just use pip install morfessor.

Also, in that case, the uninstallation is not needed -- you just delete the directory.

tools/extras/install_morfessor.sh
jty016: applied retab with size 2

tools/install_morfessor.sh
jty016: okay got it. I deleted the symlink

tools/extras/install_morfessor.sh
jty016: @jtrmal 
That line is not intended to install morfessor user-wise.
It is just to grep the name of site-package name and use the name for PYTHONPATH variable,
otherwise, sometime installation process could get an error.

Anyway after installation process, the morfessor libs will be located inside kaldi/tools/

danpovey:
@jtrmal, please merge when you think it's OK.

jtrmal:
Yep, I will get to this tmrw, I'm still not fully comfortable with my
understanding how the script works.
Y.

tools/extras/install_morfessor.sh
```diff
@@ -0,0 +1,60 @@
...
+# Apache 2.0
+#
+
+echo "#### installing morfessor"
+dirname=morfessor
+if [ ! -d ./$dirname ]; then
+  mkdir -p ./$dirname
+  git clone https://github.com/aalto-speech/morfessor.git morfessor ||
+    {
+      if [ $? -ne 0 ]; then
```
jtrmal: this is confusing -- why you are testing  the return code if you have the guarantee done by  `||`

tools/extras/install_morfessor.sh
```diff
@@ -0,0 +1,60 @@
...
+      fi
+    }
+fi
+
+# local installation
+site_packages_dir=$(python -m site --user-site | grep -oE "lib.*")
+wd=`pwd`
+wd=`readlink -f $wd || pwd`
+export MORFESSOR="$wd/morfessor"
+export PYTHONPATH="${PYTHONPATH:-}:$MORFESSOR/${site_packages_dir}"
```
jtrmal: I still think this is too complex for no clear reason (reason clear to me) -- you could just install into $MORFESSOR/local/ (or whatever) and this would greatly simplify the script.

tools/extras/install_morfessor.sh
```diff
@@ -0,0 +1,60 @@
...
+  wd=`readlink -f $wd || pwd`
+
+  echo "export MORFESSOR=\"$wd/morfessor\""
+  echo "export PATH=\"\$PATH:\${MORFESSOR}/bin\""
+  echo "export PYTHONPATH=\"\${PYTHONPATH:-}:\$MORFESSOR/${site_packages_dir}\""
+) >> env.sh
+
+echo >&2 "installation of MORFESSOR finished successfully"
+echo >&2 "please source tools/env.sh in your path.sh to enable it"
+echo >&2 'when uninstall it, try: sudo rm $(cat ./morfessor/files.txt)'
```
jtrmal: and when uninstalling -- you should delete the whole $MORFESSOR fir (and sudo shouldn't be needed to be used?)

psmit:
Morfessor author here. We normally recommend installing through pip or conda, but in this case there is a much simpler way:

 - Clone the github
 - Set PYTHONPATH to include the repo
 - Set PATH to include the scripts directory

Running setup.py has no benefit here in my opinion.

Uninstalling is as simple as removing the repo (and cleaning up the variables if so desired)

jtrmal:
Thanks, Peter!
y.

psmit:
One issue, I noticed that the scripts in scripts/ are actually not having the executable bit set when cloning, I can fix that upstream (in a couple hours)

tools/extras/install_morfessor.sh
jty016: @jtrmal yep I agree on it. 
Anyway it seems like @psmit shows us more clear way. 
After scripts in scripts/ has been changed to executables, I'll update this code in more simpler way.

tools/extras/install_morfessor.sh
jty016: it should be removed

tools/extras/install_morfessor.sh
jty016: It was intended to inform where exist the real files.  Will be removed either

psmit:
Script have executable bit set now: https://github.com/aalto-speech/morfessor/commit/6c5ad6d8d6cc7c1404b0af91c2853c3f39075a1f

Changing the gist of the script to:

    git clone https://github.com/aalto-speech/morfessor.git morfessor
    (
        echo 'export MORFESSOR="$(pwd)/morfessor"'
        echo 'export PATH="$(pwd)/morfessor/scripts:$PATH"'
        echo 'export PYTHONPATH="$(pwd)/morfessor"'
    ) >> env.sh

should be enough.

psmit:
I just tested this script and it works for me (Ubuntu 16.04)

jtrmal:
Thanks to both of you! Merging.

PR #2303, opened 2018-03-22
[not-for-merge] Thread compiler xvector example and few optimizations

gorinars:
Example of running xvector extractor in multiple threads (includes e4da298 that makes it safe)

Some features include:
* --num-threads
* extra argument cache_rxfilename  that allows to store cache and re-use it in extractor
* --chunk-sampling-rate - dummy implementation of frame rounding to make segments less diferse and make code spend less time on compilation (not tested properly - better use default)

### DIFF
--- a/egs/aspire/s5/local/run_asr_segmentation.sh
+++ b/egs/aspire/s5/local/run_asr_segmentation.sh
@@ -48,7 +48,6 @@ reco_nj=40
 
 # test options
 test_nj=30
-test_stage=1
 
 . ./cmd.sh
 if [ -f ./path.sh ]; then . ./path.sh; fi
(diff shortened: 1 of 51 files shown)


danpovey:
This is related to #2288.

@gorinars, would you mind fixing the conflicts?

As previously discussed, I will close this PR but I wanted it on github for discoverability, since this might come in useful in future.

gorinars:
I was not sure if it was better to keep your commit e4da298 in this PR. Basically without it the thing will not work properly, so I guess it is better to keep it and later when you merge your part the diff will be not so huge

danpovey:
closing but please keep that branch up for reference.

PR #2304, opened 2018-03-23
[src] changes to make kaldi can be installed on cygwin

qbetterk:
mostly the job is to add "defined(__CYGWIN__)" when needed and choose correct function as compiled on cygwin.

One thing worth noticing is that there is a line of code should be changed in "/usr/include/ctype.h". (still the problem of add "defined(__CYGWIN__)"). But this seems a problem of cygwin itself. Maybe user need to add it manually.

### DIFF
--- a/src/base/get_version.sh
+++ b/src/base/get_version.sh
@@ -89,7 +89,6 @@ fi
 # Overwrite ./version.h with the temporary file if they are different.
 if ! cmp -s $temp version.h; then
   cp $temp version.h
-  sed -i 's/\r//g' version.h
   chmod 644 version.h
 fi
 exit 0
--- a/src/base/get_version.sh
+++ b/src/base/get_version.sh
@@ -89,6 +89,7 @@ fi
 # Overwrite ./version.h with the temporary file if they are different.
 if ! cmp -s $temp version.h; then
   cp $temp version.h
+  sed -i 's/\r//g' version.h
   chmod 644 version.h
 fi
 exit 0
--- a/src/base/kaldi-utils.cc
+++ b/src/base/kaldi-utils.cc
@@ -46,7 +46,7 @@ void Sleep(float seconds) {
 #if defined(_MSC_VER) || defined(MINGW)
   ::Sleep(static_cast<int>(seconds * 1000.0));
 #elif defined(__CYGWIN__)
-  kaldi::Sleep(static_cast<int>(seconds * 1000.0));
+  sleep(static_cast<int>(seconds * 1.0));
 #else
   usleep(static_cast<int>(seconds * 1000000.0));
 #endif
--- a/src/base/kaldi-utils.cc
+++ b/src/base/kaldi-utils.cc
@@ -45,6 +45,8 @@ std::string CharToString(const char &c) {
 void Sleep(float seconds) {
 #if defined(_MSC_VER) || defined(MINGW)
   ::Sleep(static_cast<int>(seconds * 1000.0));
+#elif defined(__CYGWIN__)
+  kaldi::Sleep(static_cast<int>(seconds * 1000.0));
 #else
   usleep(static_cast<int>(seconds * 1000000.0));
 #endif
--- a/src/base/kaldi-utils.cc
+++ b/src/base/kaldi-utils.cc
@@ -46,7 +46,7 @@ void Sleep(float seconds) {
 #if defined(_MSC_VER) || defined(MINGW)
   ::Sleep(static_cast<int>(seconds * 1000.0));
 #elif defined(__CYGWIN__)
-  sleep(static_cast<int>(seconds * 1.0));
+  sleep(static_cast<int>(seconds));
 #else
   usleep(static_cast<int>(seconds * 1000000.0));
 #endif
--- a/src/base/kaldi-utils.h
+++ b/src/base/kaldi-utils.h
@@ -52,7 +52,7 @@
 #  define KALDI_MEMALIGN_FREE(x) _aligned_free(x)
 #elif defined(__CYGWIN__)
 #  define KALDI_MEMALIGN(align, size, pp_orig) \
-  (*(pp_orig) = aligned_alloc(size, align))
+  (*(pp_orig) = aligned_alloc(align, size))
 #  define KALDI_MEMALIGN_FREE(x) free(x)
 #else
 #  define KALDI_MEMALIGN(align, size, pp_orig) \
--- a/src/base/kaldi-utils.h
+++ b/src/base/kaldi-utils.h
@@ -46,20 +46,6 @@
 #endif
 #endif
 
-#if defined(_MSC_VER)
-#  define KALDI_MEMALIGN(align, size, pp_orig) \
-  (*(pp_orig) = _aligned_malloc(size, align))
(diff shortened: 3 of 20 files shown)


src/base/kaldi-utils.cc
```diff
@@ -45,6 +45,8 @@ std::string CharToString(const char &c) {
 void Sleep(float seconds) {
 #if defined(_MSC_VER) || defined(MINGW)
   ::Sleep(static_cast<int>(seconds * 1000.0));
+#elif defined(__CYGWIN__)
+  kaldi::Sleep(static_cast<int>(seconds * 1000.0));
```
danpovey: this makes me think you haven't run the tests ("make test").  It would just recurse and crash when it exceeds the max stack depth.
Try doing "man usleep" or "man -k sleep" in cygwin, assuming man pages are installed, to see what sleep functions are available and what you have to include to find them.
If there isn't something like

src/util/kaldi-io.cc
```diff
@@ -34,16 +34,27 @@
...
+  return popen(command, mode);
 #else  // KALDI_CYGWIN_COMPAT
   return _popen(command, mode);
 #endif  // KALDI_CYGWIN_COMPAT
 }
 #endif  // _MSC_VER
 
+#ifdef __CYGWIN__
+static pclose(FILE* file) {
+  //FILE *file = (FILE*)_file;
```
jtrmal: I don't see the purpose of this function -- seems almost like a recursive call or somethign
y.

qbetterk: Sorry, this change(about 'popen' and 'pclose') is a mistake. The original code was right. I find that what I need to change is in the /usr/include/stdio.h rather than in kaldi.

qbetterk:
@hainan-xv Can you review this?

src/base/get_version.sh
```diff
@@ -89,6 +89,7 @@ fi
 # Overwrite ./version.h with the temporary file if they are different.
 if ! cmp -s $temp version.h; then
   cp $temp version.h
+  sed -i 's/\r//g' version.h
```
jtrmal: try to avoid sed -i (or sed altogether)

jtrmal:
LGTM (besides  the sed issue)
BTW, there was (used to be) some compatibility  feature to make the
commands actiing like file (ie. those "command|command2|") specifiers work
correctly within CYGWIN, make sure you didn't break those by
adding/changing the ifdefs -- I think that relates to kaldi-io file.
Also, did you run all tests successfully (`make test`)?
y

src/base/get_version.sh
danpovey: actually, try to figure out where the linefeed character comes from originally (does 'echo' add it?) and I'd like to know what problems it causes.

danpovey:
Yes, testing should include running a setup like "yesno" on cygwin.

src/base/get_version.sh
qbetterk: but there would be '\r' in the 'version.h' file since it is on windows. should I leave it to users to delete '\r' themselves?

jtrmal:
Actually, I think the logic of your change is wrong, your code will cause
the if ! cmd code always succeed, so you will always replace the file --
you should modify the $temp file instead
y.

src/base/get_version.sh
qbetterk: kaldi itself does not have '\r'. this character is automatically generated when I download kaldi to a windows machine. '\r' in windows means 'end of line' like '\n' in Unix and the Unix environment(cygwin) cannot recognize it.

jtrmal:
does gcc/g++ mind the \r character there?
y.

src/base/get_version.sh
jtrmal: Actually, I think the logic of your change is wrong, your code will cause
the `if ! cmd ` test always succeed, so you will always replace the file --
you should modify the $temp file instead

qbetterk:
No,  gcc/g++ seems ok with \r

Sent from Mail for Windows 10

jtrmal:
OK, sorry, in that case I don't understand if the edit is needed at all? I
might be missing somehting, tho.
y.

qbetterk:
Yes, I have run the ‘make test’ and there is no error.

Kun

src/fstext/lattice-weight-test.cc
```diff
@@ -34,14 +34,15 @@ typedef CompactLatticeWeightCommonDivisorTpl<LatticeWeight, int32>
...
-    return LatticeWeight( 1, 2);  // sometimes return special values..
-  } else if (kaldi::Rand() % 3 ==  0) {
-    return LatticeWeight( 2, 1);  // this tests more thoroughly certain properties...
+  int tmp = kaldi::Rand() % 4
+  if (tmp == 0) {
+  	return LatticeWeight::Zero();
+  } else if (tmp == 1) {
+  	return LatticeWeight( 1, 2);  // sometimes return special values..
+  } else if (tmp == 2) {
+  	return LatticeWeight( 2, 1);  // this tests more thoroughly certain properties...
```
hainan-xv: you're not supposed to use \tab in the code; use 2 spaces instead.

jtrmal:
Yes, please read the Google Style Guide for C++

qbetterk:
Thanks! This guide seems useful!

Best,
Kun

src/base/kaldi-utils.cc
```diff
@@ -45,6 +45,8 @@ std::string CharToString(const char &c) {
 void Sleep(float seconds) {
 #if defined(_MSC_VER) || defined(MINGW)
   ::Sleep(static_cast<int>(seconds * 1000.0));
+#elif defined(__CYGWIN__)
+  sleep(static_cast<int>(seconds * 1.0));
```
danpovey: the `* 1.0` here is unnecessary.

src/base/kaldi-utils.h
```diff
@@ -46,10 +46,14 @@
...
 #endif
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER)
 #  define KALDI_MEMALIGN(align, size, pp_orig) \
   (*(pp_orig) = _aligned_malloc(size, align))
 #  define KALDI_MEMALIGN_FREE(x) _aligned_free(x)
+#elif defined(__CYGWIN__)
+#  define KALDI_MEMALIGN(align, size, pp_orig) \
+  (*(pp_orig) = aligned_alloc(align, size))
```
danpovey: Actually I now see that we don't need this macro at all any more, as aligned_alloc is part of the C++11 standard.  So please remove the macro and replace all instances of
  `foo = KALDI_MEMALIGN(x, y, &z)`
with
  `foo = aligned_alloc(x, y);`
and remove the temporary variables that previously bound to z, and replace KALDI_MEMALIGN_FREE with free (there are instances of this in the cudamatrix/ directory as well, that should also be replaced).

qbetterk:
OK I would change it.

By the way, the test of ‘yesno’  on both Cygwin and Linux are all finished with no error.

Best,
Kun

jtrmal:
and do they have comparable scores?
y.

qbetterk:
I remember it is 0% WER.

PR #2311, opened 2018-03-26
[WIP] end2end OCR on IAM

hhadian:


egs/cifar/v1/image/get_allowed_lengths.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python
```
danpovey: can this be python3?  (python3 preferred for new scripts).

PR #2313, opened 2018-03-26
use openfst 1.6.7

jtrmal:
@danpovey for your consideration
I wasn't able to come up with another solution than explicit full specialization of the templates. 
I left the original template there so that the error will be triggered again if someone uses other types than int and double/float

### DIFF
--- a/src/fstext/lattice-weight.h
+++ b/src/fstext/lattice-weight.h
@@ -621,9 +621,9 @@ class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<FloatType>, IntType> 
   }
 };
 template<>
-class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<float>, int> > {
+class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<float>, int32> > {
  public:
-  typedef CompactLatticeWeightTpl<LatticeWeightTpl<float>, int> Weight;
+  typedef CompactLatticeWeightTpl<LatticeWeightTpl<float>, int32> Weight;
 
   NaturalLess() {}
 
@@ -635,9 +635,9 @@ class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<float>, int> > {
   }
 };
 template<>
-class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<double>, int> > {
+class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<double>, int32> > {
  public:
-  typedef CompactLatticeWeightTpl<LatticeWeightTpl<double>, int> Weight;
+  typedef CompactLatticeWeightTpl<LatticeWeightTpl<double>, int32> Weight;
 
   NaturalLess() {}
 
--- a/src/fstext/lattice-weight.h
+++ b/src/fstext/lattice-weight.h
@@ -330,6 +330,34 @@ class NaturalLess<LatticeWeightTpl<FloatType> > {
     return (Compare(w1, w2) == 1);
   }
 };
+template<>
+class NaturalLess<LatticeWeightTpl<float> > {
+ public:
+  typedef LatticeWeightTpl<float> Weight;
+
+  NaturalLess() {}
+
+  bool operator()(const Weight &w1, const Weight &w2) const {
+    // NaturalLess is a negative order (opposite to normal ordering).
+    // This operator () corresponds to "<" in the negative order, which
+    // corresponds to the ">" in the normal order.
+    return (Compare(w1, w2) == 1);
+  }
+};
+template<>
+class NaturalLess<LatticeWeightTpl<double> > {
+ public:
+  typedef LatticeWeightTpl<double> Weight;
+
+  NaturalLess() {}
+
+  bool operator()(const Weight &w1, const Weight &w2) const {
+    // NaturalLess is a negative order (opposite to normal ordering).
+    // This operator () corresponds to "<" in the negative order, which
+    // corresponds to the ">" in the normal order.
+    return (Compare(w1, w2) == 1);
+  }
+};
 
 template<class FloatType>
 inline LatticeWeightTpl<FloatType> Times(const LatticeWeightTpl<FloatType> &w1,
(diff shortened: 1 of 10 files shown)


danpovey:
I just realized that if I merge this, it will break the builds because fst::kNoSymbol doesn't exist in the current OpenFst versions.  We'd have to bump up the build version so 'make' would give the right error message, and force the user to install the latest OpenFst version.
I don't think I'll do this tonight.
One possibility-- that would make it unnecessary to bump up the build version-- is to change e.g.
`if (value == fst::kNoSymbol) {`
to
`if (value == -1) { // fst::kNoSymbol`

jtrmal:
changed as you suggested

danpovey:
@jtrmal, please merge once you feel it's tested enough.

src/fstext/lattice-weight.h
```diff
@@ -330,6 +330,34 @@ class NaturalLess<LatticeWeightTpl<FloatType> > {
     return (Compare(w1, w2) == 1);
   }
 };
+template<>
+class NaturalLess<LatticeWeightTpl<float> > {
+ public:
+  typedef LatticeWeightTpl<float> Weight;
```
kkm000: Just wondering, did `LatticeWeightTpl` break for you too? I tried only cl and icc on Windows so far, and they only complained about the 2-argument template `CompactLatticeWeightTpl`. But the both compilers are minimum C++14, I think. A 1-argument specialization is "more specific" in C++, so both specialization should be fine, if you bake in the `int32` as the Gallic label type into `CompactLatticeWeightTpl`.

src/fstext/lattice-weight.h
```diff
@@ -330,6 +330,34 @@ class NaturalLess<LatticeWeightTpl<FloatType> > {
...
+
+  NaturalLess() {}
```
kkm000: This ever-multiplying constructor is not needed.

src/fstext/lattice-weight.h
```diff
@@ -592,6 +620,34 @@ class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<FloatType>, IntType>
     return (Compare(w1, w2) == 1);
   }
 };
+template<>
+class NaturalLess<CompactLatticeWeightTpl<LatticeWeightTpl<float>, int> > {
```
kkm000: Should the `int` be `int32_t` for consistency? I think `int` is not generally used in the codebase.

src/fstext/lattice-weight.h
danpovey: Yes it broke for us, that's why we changed this code.  Kaldi only requires c++11 to compile, so we don't want to depend on c++14 features.

src/fstext/lattice-weight.h
danpovey: yes, this should be int32 (which is what we call it Kaldi code).

jtrmal:
I'm traveling today, but I'll try and provide fixes today (worst case tmrw).
Y.

Excalib88:
I try to build kaldi but i get error "Error	C1083	Cannot open include file: 'sys/socket.h': No such file or directory	online-audio-server-decode-faster	\kaldi\src\onlinebin\online-audio-server-decode-faster.cc	37	" and "Error	C1083	Cannot open include file: 'netdb.h': No such file or directory	online-net-client	\kaldi\src\onlinebin\online-net-client.cc" 
I such files cannot find

jtrmal:
Those errors do not relate to openfst.
Also, if you are taking about building on windows, it is expected you won't
find those, IIRC. The online net binaries weren't ported (too os specific).
Y.

jtrmal:
This relates mainly to
https://github.com/espnet/JSALT18main/issues/15
and
https://github.com/espnet/JSALT18main/issues/18

src/fstext/lattice-weight.h
jtrmal: I'm adressing the changes from @kkm000 
Removing the constructor, howevcer, is not completely trivial -- after removal, I get the following errors:
```
g++ -std=c++11 -I.. -I/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include  -Wall -Wno-sign-compare -Wno-unused-local-typedefs -Wno-deprecated-declarations -Winit-self -DKALDI_DOUBLEPRECISION=0 -DHAVE_EXECINFO_H=1 -DHAVE_CXXABI_H -DHAVE_ATLAS -I/export/a09/jtrmal/kaldi-clean-git/tools/ATLAS_headers/include -msse -msse2 -pthread -g  -fPIC -DHAVE_CUDA -I/usr/local/cuda/include   -c -o feature-mfcc.o feature-mfcc.cc
In file included from /export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/fst.h:27:0,
                 from /export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/expanded-fst.h:16,
                 from /export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/fstlib.h:28,
                 from ../fstext/fstext-lib.h:22,
                 from ../lat/kaldi-lattice.h:24,
                 from ../hmm/hmm-utils.h:25,
                 from hmm-utils.cc:22:
/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/arc.h: In instantiation of ‘fst::ArcTpl<Weight>::ArcTpl() [with W = fst::LatticeWeightTpl<float>]’:
/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/test-properties.h:113:11:   required from ‘uint64 fst::ComputeProperties(const fst::Fst<Arc>&, uint64, uint64*, bool) [with Arc = fst::ArcTpl<fst::LatticeWeightTpl<float> >; uint64 = long unsigned int]’
/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/test-properties.h:216:50:   required from ‘uint64 fst::TestProperties(const fst::Fst<Arc>&, uint64, uint64*) [with Arc = fst::ArcTpl<fst::LatticeWeightTpl<float> >; uint64 = long unsigned int]’
/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/fst.h:871:52:   required from ‘uint64 fst::ImplToFst<Impl, FST>::Properties(uint64, bool) const [with Impl = fst::internal::VectorFstImpl<fst::VectorState<fst::ArcTpl<fst::LatticeWeightTpl<float> >, std::allocator<fst::ArcTpl<fst::LatticeWeightTpl<float> > > > >; FST = fst::MutableFst<fst::ArcTpl<fst::LatticeWeightTpl<float> > >; uint64 = long unsigned int]’
/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/vector-fst.h:430:7:   required from here
/export/a09/jtrmal/kaldi-clean-git/tools/openfst/include/fst/arc.h:38:12: error: no matching function for call to ‘fst::LatticeWeightTpl<float>::LatticeWeightTpl()’
   ArcTpl() {}
```

I'll look into it in the evening -- just letting you know I didn't abandon this.

kkm000: Wow! C++ loves us!

jtrmal:
I've given up on removing the default constructor... It goes back to openfst code where the constructor is defined and even setting it as =default leads to issues as the constructor gets deleted in some cases as move constructor is not provided. I decided I don't really want to spend a day studying how move semantics affects default constructors and how this interacts with =default and =delete.

PR #2323, opened 2018-03-29
make the install script fetch changes from github repo and fix instal…

jtrmal:
…lation issue for new setuptools


langep:
Oh sorry, I thought this was the original issue. I will take a look but it should work with newest sequitur version

jtrmal:
Thanks -- I introduced these things because even the setup.py was refusing
to work without being it set up.
@danpovey, if you are ok with this, please merge (this is orthogonal
to the error report we had yesterday).
y.

PR #2324, opened 2018-03-30
[src,scripts] Add option --boundary-tolerance for chain training

danpovey:


danpovey:
One of the tunable parameters in the 'chain' systems, that affects the dumping of egs, is the left-tolerance/right-tolerance.  It affects how much we allow the time aligments in the numerator paths to differ from those in the lattice.  We usually set both tolerances to 5.  (It's measured in frames, at whatever frame rate the original lattice was dumped at).

This PR (which I want to ask people to test in a couple different scenarios) makes it possible to have a tighter tolerance at the boundaries of chunks (the 1.5-second chunks that we train on), than in the interior of the chunk.  I had been concerned, particularly for recurrent setups, that the model might figure out it was at the edge of a chunk and might find it easier to just delete all the phones near that edge.  If you set a smaller boundary-tolerance at the edge, it makes it impossible for the model to do this, it is forced to model those phones.


The fact that --boundary-tolerance can now be set to a smaller value, now also allows us to set the --left-tolerance and --right-tolerance higher than before.  This becomes a bit more like full forward-backward over the numerator graph.  The default of --left-tolerance and --right-tolerance is 5.  One thing that you can try is as follows: set the boundary-tolerance to 1 but the left/right tolerance to 8:
```
 steps/nnet3/chain/train.py  \
   ...
    --chain.left-tolerance=8 \
    --chain.right-tolerance=8 \
    --chain.boundary-tolerance=1 \
```
@GaofengCheng, would you mind trying this for some kind of recurrent system, e.g. on a Switchboard setup?  Remember that you need to re-dump egs.

@hhadian, please make a note to test this (or get someone to test it) on some kind of handwriting/OCR setup.  I suspect it might help there due to poor alignments from the GMM system.

(Note: if the lattices were generated by a non-chain system, i.e. they were at the non-subsampled sampling rate, you probably shouldn't set --boundary-tolerance to a value smaller than about int(frame-subsampling-factor/2)).

@vimalmanohar, please have a look at this to make sure you understand it.  It might be useful for some aspect of your work.

GaofengCheng:
Will try on TDNN-OPGRU and TDNN-LSTM

PR #2331, opened 2018-04-03
add --num-threads option

alex-ht:
Adding --num-threads option to make $cmd called properly.


danpovey:
Thanks, but that num-threads is not for the ivector extraction in general, it's just for a startup task that is very quick.  So it would be quite wasteful to reserve that many threads on the queue.
Closing this for now-- if it turns out there is an important reason for needing to set that, I might consider re-opening it.

PR #2333, opened 2018-04-03
Bug fix: oscillating cluster threshold in agglomerative-cluster

mmaciej2:
There was a bug in agglomerative-cluster.cc that would cause the threshold used for clustering to oscillate between positive and negative per recording.


danpovey:
Merging.  Please comment whether this would have affected any checked-in recipes.

mmaciej2:
This should not affect any checked-in recipes, as they tune the threshold as part of the recipe. It's possible that the error rate will decrease slightly, but using PLDA scoring should result in an optimal threshold of 0 in theory; and, a threshold of 0 would be unaffected by this bug, so the effect should be minimal. I will verify that there is no appreciable performance difference on the CallHome recipe, but this bug should really only affect scoring methods that are not "centered" around 0, which we do not currently use.

PR #2334, opened 2018-04-03
[scripts] Fix to per-utt issue after changes to chain/get_egs.sh in #2140

danpovey:


danpovey:
Note: the problem probably only affected the IAM recipe and ./rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1c.sh.

PR #2335, opened 2018-04-03
Single threaded sequential taskrunner

sikoried:
Introduce a `synchronous_` flag internal to `TaskSequencer` that turns `Run()` effectively into a synchronous call (ie. no threading), as advised by @dogancan.

Fixed `online2bin/online2-wav-nnet[23]-latgen-{faster,threaded}.cc` which used `--num-threads-startup` instead of `--num-threads` to overwrite `g_num_threads` (the global external variable defining the number of threads). The option `--num-threads-startup` seemed never used in the examples, anyways.

Background: For some experiments to run kaldi in a secure enclave (Interspeech submission pending), we needed to limit the number of threads; unfortunately, the `ivector/ivector-extractor.cc` manually sets the threads to `g_num_threads` (lines 198ff).


src/util/kaldi-thread.h
```diff
@@ -181,11 +181,21 @@ class TaskSequencer {
     KALDI_ASSERT((config.num_threads_total <= 0 ||
                   config.num_threads_total >= config.num_threads) &&
                  "num-threads-total, if specified, must be >= num-threads");
+    if (config.num_threads == 0) {
```
danpovey: I think it would be nicer if instead of synchronous_, you added a member num_threads_, which was just a copy of config.num_threads.  There's a chance we might need that later for something else anyway.

src/online2bin/online2-wav-nnet2-latgen-faster.cc
```diff
@@ -124,7 +124,7 @@ int main(int argc, char *argv[]) {
                 "--use-most-recent-ivector=true and --greedy-ivector-extractor=true "
                 "in the file given to --ivector-extraction-config, and "
                 "--chunk-length=-1.");
-    po.Register("num-threads-startup", &g_num_threads,
+    po.Register("num-threads", &g_num_threads,
```
danpovey: I don't understand why you want to change this option name.
The reason it was called --num-threads-startup was to avoid giving users the impression that this binary is actually able to do multi-threaded decoding.   (The num-threads is only used at the very beginning of the program).

sikoried: Sure. How about `num-threads-ivectors`? The point is that it uses multithreading even if the user is not aware of it (and it has nothing to do with startup)

sikoried: The ivector extractor [always starts multithreaded](https://github.com/kaldi-asr/kaldi/blob/831dd38cf7f94dff0fe8325d11e27e1eb8946389/src/ivector/ivector-extractor.cc#L198) using `g_num_threads`, ie. if someone uses the ivector extractor, they should be setting `g_num_threads` (which is currently done by the obscure `num-threads-startup` option).

danpovey:
That code should only be called once when the program is started, when
the  OnlineNnet2FeaturePipelineInfo  object is initialized.  If you have
evidence that that is not the case, let me know.

sikoried:
Ah, I guess I wasn't clear why I'm trying to get rid of the threads regardless if they only happen in the init phase: SGX and in turn the [Graphene](https://github.com/oscarlab/graphene) library that we use to lift the computation into the secure enclave has a limit of how many threads you can create over the process lifetime, that's why I'd like total control over thread creation. Does that make sense?

danpovey:
Makes sense.  But please make the changes requested.

src/util/kaldi-thread.h
```diff
@@ -260,6 +268,8 @@ class TaskSequencer {
     args->me->tot_threads_avail_.Signal();
   }
 
+  int num_threads_; // copy of config.num_threads (since Semaphore doesn't store original count)
```
danpovey: thanks.
please make this int32, just to forestall compiler warnings on some systems and compilers.f

danpovey:
thanks, will merge when checks done.

Issue #2003, opened 2017-11-08
Change printed output for non-chain nnet training


danpovey:
This is super non-urgent, but making an issue to keep track of this.
I'd like the diagnostics for the non-chain nnet3 training to be changed in the same way that @hhadian previously did for chain training in this PR
https://github.com/kaldi-asr/kaldi/pull/1880

PR #2338, opened 2018-04-05
Cosmetic changes to nnet3 printed output

Ore-an:
As for #2003, the logging code was made identical to the chain logging code.


danpovey:
Thanks-- I assume you have tested this?

Ore-an:
I've tested for 3-4 epochs for each script besides train_raw_rnn.py (as I'm not really familiar with the segmentation scripts) and for full training for train_rnn.py. Considering how the code works, the only problem it could create is the absence of logging information if there are other python training scripts calling the frame-level common.py besides the four I've changed, but as far as I know there shouldn't be other.
rnn

danpovey:
OK, sounds good enough.  Merging.  Thanks!

PR #2339, opened 2018-04-06
Bug fix in egs generation script

vimalmanohar:


egs/wsj/s5/steps/nnet3/chain/get_egs.sh
```diff
@@ -488,7 +488,7 @@ if [ $stage -le 6 ]; then
     # 'storage' directory.
     rm cegs_orig.*.ark 2>/dev/null
   )
-  if [ $archives_multiple -gt 1 ]; then
+  if ! $generate_egs_scp && [ $archives_multiple -gt 1 ]; then
```
danpovey: Vimal, can you explain why this is necessary?  I'm having trouble understanding the script now.

danpovey: Forget it, I figured it out.

PR #2343, opened 2018-04-08
IAM: Add a regular chain training recipe which uses the e2e model for alignments and tree

hhadian:


egs/iam/v1/local/chain/run_cnn_e2eali_1a.sh
```diff
@@ -0,0 +1,245 @@
+#!/bin/bash
+
+# e2eali_1a is the same as chainali_1c but uses the e2e chain model to get the
+# lattice alignments and to build a tree
+
+# local/chain/compare_wer.sh exp/chain/e2e_cnn_1a exp/chain/cnn_chainali_1c exp/chain/cnn_e2eali_1a
+# System                      e2e_cnn_1a cnn_chainali_1c cnn_e2eali_1a
+# WER                             14.05     13.14     12.79
+# CER                              6.59      6.40      5.73
```
danpovey: This is very interesting.   Merging.  It would be interesting to explore whether this approach is better for ASR.

PR #2350, opened 2018-04-11
[src] Fix nnet3-copy-egs bug for multiling. non-chain training (thx:Joshua Meyer)

danpovey:


JRMeyer:
It works! Thanks, Dan:)

PR #2354, opened 2018-04-12
[WIP] Iam_ocr Adding Wellington corpus

ChunChiehChang:
I haven't updated the new results on the top of the scripts yet.


egs/iam/v1/run.sh
```diff
@@ -78,12 +78,12 @@ if [ $stage -le 4 ]; then
     data/lang exp/mono
 fi
 
-if [ $stage -le 5 ]; then
-  utils/mkgraph.sh --mono data/lang_test exp/mono exp/mono/graph
-
-  steps/decode.sh --nj $nj --cmd $cmd exp/mono/graph data/test \
-    exp/mono/decode_test
-fi
+#if [ $stage -le 5 ]; then
```
danpovey: are these commented-out parts supposed to be commented out?

ChunChiehChang: Yes, I intentionally commented out the decoding parts because the results from the first couple of stages (monophone/triphone) are not very good and I didn't think I needed to compute them. I can just pass their alignments on to the later CNN stages of the script.

egs/iam/v1/local/remove_utterances_from_corpus.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
```
aarora8: A cleaned version of this file './local/remove_test_utterances_from_lob.py' is already part of kaldi/master.

egs/iam/v1/run.sh
danpovey: Rather than comment them out, you can set at the top
`decode_gmm=false`
and do things like
`if [ $stage -le 5 ] && $decode_gmm; then`

danpovey:
I am thinking of merging this within 24 hours-- I want to be a bit more aggressive about merging these recipes as I think it will help our work move faster.
@hhadian do you have time to review in a bit more detail?  If you are busy it's OK, I can merge and we can address things as we find them.

danpovey:
oh wait, I see this PR is very small.   Merging.

PR #2356, opened 2018-04-14
Madcat Arabic handwritten text line recognition

aarora8:
It's using 750k utterances from madcat Arabic data. It is getting the line image from the MAR (minimum area rectangle). It contains the recent TDNN training recipe for End2end and regular chain recipe. @hhadian 

add README(README.txt)
add scripts for data preparation (text, wav.scp and utt2spk file) (local/prepare_data.sh, local/process_data.py, local/create_line_image_from_page_image.py)
add scripts for feature extraction (local/make_features.py)
add scripts for lexicon, language modeling, grammar (local/train_lm.sh, local/prepare_lexicon.py, local/prepare_dict.sh)
add script for GMM-HMM training and using chain model (local/chain/run_cnn_1a.sh, run.sh, run_end2end.sh, local/chain/run_cnn_chainali_1b.sh, local/chain/run_flatstart_cnn1a.sh, local/chain/compare_wer.sh)
Other (cmd.sh, link to image/steps/utils, v1/local/score.sh, path.sh, local/check_tools.sh)

-It is getting the line image from the MAR (minimum area rectangle). 
-It is currently building the language model from training utterances only. 
-Its lexicon size is 95k words, OOV rate is around 1.5%. 
-For quick debugging and experiments, it can be run with a subset of the dataset based on writing conditions (writing style, speed, carefulness) of the image. 
-It contains the recent TDNN training recipe for End2end and regular chain.
-WER 12.97% with line image formed by stitching the word images.
-WER 15.03% with line image formed using MAR.

To do:
replace PIL in create_line_image_from_page_image
replace convex_hull library routine
update configs


egs/madcat_ar/v1/local/create_line_image_from_page_image.py
```diff
@@ -0,0 +1,300 @@
...
+    corner_points = []
+    for i1 in (.5, -.5):
+        for i2 in (i1, -1 * i1):
+            corner_points.append((rectangle['rectangle_center'][0] + i1 * rectangle['length_parallel'],
+                            rectangle['rectangle_center'][1] + i2 * rectangle['length_orthogonal']))
+
+    return rotate_points(rectangle['rectangle_center'], rectangle['unit_vector_angle'], corner_points)
+
+
+def minimum_bounding_box(points):
```
danpovey: All these functions need more documentation.  You need to document the format of the inputs and outputs and what the function does, using a doc-string-- don't rely on the user having to read the code.

egs/madcat_ar/v1/local/create_line_image_from_page_image.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3
+# Minimum Bounding Box part in this script is originally (https://github.com/BebeSparkelSparkel/MinimumBoundingBox)
+
+
+import argparse
+import os
+import xml.dom.minidom as minidom
+from PIL import Image
```
danpovey: We have to decide if this (PIL) is the image library we want to use and maybe we should have a script that checks if it and other dependencies are installed (called from the run.sh) and an installation script?  It might be just a case of doing "pip install Pillow".  (?).  I notice PIL is not available by default on my mac-- it must not be a core Python library.  @hhadian, comments?  For now this doesn't have to be changed though.

Also, scipy is a big dependency and probably one whose presence should be checked by the script.  It might turn out to be easier just to write our own convex-hull code.  But for now it's not a priority to address this.

egs/madcat_ar/v1/local/create_line_image_from_page_image.py
```diff
@@ -0,0 +1,300 @@
...
+from scipy.misc import toimage
+
+from scipy.spatial import ConvexHull
+from math import sqrt
+from math import atan2, cos, sin, pi, degrees
+from collections import namedtuple
+
+parser = argparse.ArgumentParser(description="""Creates line images from page image.""")
+parser.add_argument('database_path1', type=str,
+                    help='Path to the downloaded (and extracted) mdacat data file 1')
```
danpovey: i think you mean madcat, not mdacat.

Please add an example of usage of this program.  Below is an example from an rnnlm script:
```
parser = argparse.ArgumentParser(description="This script get a vocab from unigram counts "
                                 "of words produced by get_unigram_counts.sh",
                                 epilog="E.g. " + sys.argv[0] + " data/rnnlm/data > data/rnnlm/vocab/words.txt",
                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
And I think instead of 'data file 1' you mean data directory 1.

egs/madcat_ar/v1/local/create_line_image_from_page_image.py
```diff
@@ -0,0 +1,300 @@
...
+parser.add_argument('database_path2', type=str,
+                    help='Path to the downloaded (and extracted) mdacat data file 2')
+parser.add_argument('database_path3', type=str,
+                    help='Path to the downloaded (and extracted) mdacat data file 3')
+parser.add_argument('data_splits', type=str,
+                    help='Path to file that contains the train/test/dev split information')
+args = parser.parse_args()
+
+bounding_box = namedtuple('bounding_box', ('area',
```
danpovey: This needs documentation.

egs/madcat_ar/v1/local/create_line_image_from_page_image.py
```diff
@@ -0,0 +1,300 @@
...
+data_path2 = args.database_path2
+data_path3 = args.database_path3
+height_buffer = int(args.height_buffer)
+width_buffer = int(args.width_buffer)
+char_width_buffer = int(args.char_width_buffer)
+char_height_buffer = int(args.char_height_buffer)
+line_images_path_list = args.database_path1.split('/')
+line_images_path = ('/').join(line_images_path_list[:3])
+
+writing_condiiton_folder_list = args.database_path1.split('/')
```
danpovey: watch spelling.  it's 'condition'.

egs/madcat_ar/v1/local/create_line_image_from_page_image.py
hhadian: I guess Pillow (which is the new PIL) is not very actively maintained. **scikit-image** is more up-to-date but also bigger I guess. I'll need to do some checking to see what other options we might have.
[P.S. I just tried installing Pillow on my fedora and it failed but I could easily install scikit-image.]

For checking, Ashish could add a similar script to `egs/iam/v1/local/check_tools.sh` to check for all the requirements.

jtrmal:
I have sent an email about this about two months ago. There was a drop-in
solution - I think Scikit-image has the image reading functions obsoleted.
Or something like that.
Y.

egs/madcat_ar/v1/README.txt
```diff
@@ -0,0 +1,15 @@
+MADCAT (Multilingual Automatic Document Classification Analysis and Translation) 
+Arabic Corpus is a LDC dataset (LDC2012T15, LDC2013T09, LDC2013T15) for handwriting recognition. 
+The dataset contains abstracts from News related passages and blogs. The xml file for each page 
+provides line segmentation and word segmentation information and also provide the writing
```
hhadian: provide --> provides

egs/madcat_ar/v1/local/chain/run_cnn_1a.sh
```diff
@@ -0,0 +1,233 @@
...
+affix=_1a  #affix for TDNN+LSTM directory e.g. "1a" or "1b", in case we change the configuration.
+ali=tri3_ali
+common_egs_dir=
+reporting_email=
+
+# chain options
+train_stage=-10
+xent_regularize=0.1
+frame_subsampling_factor=4
+alignment_subsampling_factor=1
```
hhadian: Remove this variable. It's not needed in this script. 
Also remove it in other scripts and hardcode it where needed.

egs/madcat_ar/v1/local/chain/run_cnn_chainali_1b.sh
```diff
@@ -0,0 +1,245 @@
+#!/bin/bash
+
+# chainali_1b is as chainali_1a except it has 3 more cnn layers and 1 less tdnn layer.
+# ./local/chain/compare_wer.sh exp/chain/cnn_chainali_1a/ exp/chain/cnn_chainali_1b/
+# System                      cnn_chainali_1a cnn_chainali_1b
+# WER                              6.69     6.25
```
hhadian: Are these results for MADCAT? Please update them.

egs/madcat_ar/v1/local/create_line_image_from_page_image_word_stitching.py
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env python3
```
hhadian: These scripts are not very small. I guess it's better to follow the standard style as in `image/get_allowed_lengths.py` and maybe even do logging.

egs/madcat_ar/v1/local/prepare_lang.sh
```diff
@@ -0,0 +1,521 @@
+#!/bin/bash
```
hhadian: How is this script different from utils/prepare_lang.sh?

egs/madcat_ar/v1/local/prepare_lm.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
hhadian: Please remove this script.

egs/madcat_ar/v1/local/process_data.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python3
+
+# Copyright  2018  Ashish Arora
+
+""" This script reads the extracted IAM database files and creates
```
hhadian: Please update the docs.

egs/madcat_ar/v1/path.sh
```diff
@@ -0,0 +1,7 @@
+export KALDI_ROOT=`pwd`/../../..
+[ -f $KALDI_ROOT/tools/env.sh ] && . $KALDI_ROOT/tools/env.sh
+export PATH=$PWD/utils/:$KALDI_ROOT/tools/openfst/bin:$PWD:$PATH
+[ ! -f $KALDI_ROOT/tools/config/common_path.sh ] && echo >&2 "The standard file $KALDI_ROOT/tools/config/common_path.sh is not present -> Exit!" && exit 1
+. $KALDI_ROOT/tools/config/common_path.sh
+export LC_ALL=C
+export CUDA_CACHE_DISABLE=1
```
hhadian: I'm not sure what this line is for.

egs/madcat_ar/v1/run_end2end.sh
```diff
@@ -0,0 +1,76 @@
...
+if [ $stage -le 5 ]; then
+  echo "$0: Estimating a language model for decoding..."
+  local/train_lm.sh
+  utils/format_lm.sh data/lang data/local/local_lm/data/arpa/3gram_unpruned.arpa.gz \
+                     data/local/dict/lexicon.txt data/lang_test
+fi
+
+
+if [ $stage -le 6 ]; then
+  echo "$0: estimating phone language model for the denominator graph"
```
hhadian: In the latest IAM scripts I have moved this stage to the actual chain script (it makes more sense). Please do the same.

egs/madcat_ar/v1/local/create_line_image_from_page_image_word_stitching.py
aarora8: This was an old script, I removed it. I will try to follow the style of image/get_allowed_length.py for create_line_image_from_page_image.py

egs/madcat_ar/v1/local/prepare_lang.sh
aarora8: It was used for smaller punctuation topology for Arabic data. But since chainali code used to take around 7-9 days. I was not able to test it. I removed it.

egs/madcat_ar/v1/local/prepare_data.sh
```diff
@@ -0,0 +1,49 @@
...
+    cp data/$dataset/utt2spk data/$dataset/utt2spk_tmp
+    cp data/$dataset/text data/$dataset/text_tmp
+    cp data/$dataset/images.scp data/$dataset/images_tmp.scp
+    sort data/$dataset/utt2spk_tmp > data/$dataset/utt2spk
+    sort data/$dataset/text_tmp > data/$dataset/text
+    sort data/$dataset/images_tmp.scp > data/$dataset/images.scp
+    rm data/$dataset/utt2spk_tmp data/$dataset/text_tmp data/$dataset/images_tmp.scp
+  done
+
+  utils/utt2spk_to_spk2utt.pl data/train/utt2spk > data/train/spk2utt
```
hhadian: This stuff will be done by `fix_data_dir.sh` too.

egs/madcat_ar/v1/local/train_lm.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+
+# Copyright 2016  Vincent Nguyen
+#           2016  Johns Hopkins University (author: Daniel Povey)
+#           2017  Ashish Arora
+#           2017  Hossein Hadian
+# Apache 2.0
+#
+# This script trains an LM on the LOB+Brown text data and IAM training transcriptions.
```
hhadian: fix these comments please.

danpovey:
I agree having a scripts  image/fix_data_dir.sh and
image/validate_date_dir.sh and the like is  a good idea.

egs/madcat_ar/v1/local/chain/run_cnn_e2eali_1b.sh
```diff
@@ -0,0 +1,233 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
aarora8: I ran the default run_end2end.sh script. The script was calling run_cnn_e2eali_1a.sh in it. But am planning to run it.

egs/madcat_ar/v1/local/chain/run_cnn_e2eali_1a.sh
```diff
@@ -0,0 +1,232 @@
...
+    --trainer.optimization.shrink-value=1.0 \
+    --trainer.num-chunk-per-minibatch=64,32 \
+    --trainer.optimization.momentum=0.0 \
+    --egs.chunk-width=$chunk_width \
+    --egs.chunk-left-context=$chunk_left_context \
+    --egs.chunk-right-context=$chunk_right_context \
+    --egs.chunk-left-context-initial=0 \
+    --egs.chunk-right-context-final=0 \
+    --egs.dir="$common_egs_dir" \
+    --egs.opts="--frames-overlap-per-eg 0" \
```
danpovey: are we going to move towards unconstrained egs?  (not necessary to do it in this PR though).

aarora8: I updated this branch with kaldi/master and added unconstrained egs script: run_cnn_e2eali_1b.sh from IAM branch. I will run it to update the results.

hhadian: Actually Ashish has added `run_cnn_e2eali_1b.sh` to the PR which uses unconstrained egs.

egs/madcat_ar/v1/local/prepare_data.sh
```diff
@@ -0,0 +1,49 @@
...
+if [ $stage -le 1 ]; then
+  local/process_data.py $download_dir1 $download_dir2 $download_dir3 $dev_split_file data/dev data/local/lines/images.scp || exit 1
+  local/process_data.py $download_dir1 $download_dir2 $download_dir3 $test_split_file data/test data/local/lines/images.scp || exit 1
+  local/process_data.py $download_dir1 $download_dir2 $download_dir3 $train_split_file data/train data/local/lines/images.scp || exit 1
+
+  for dataset in train test dev; do
+    cp data/$dataset/utt2spk data/$dataset/utt2spk_tmp
+    cp data/$dataset/text data/$dataset/text_tmp
+    cp data/$dataset/images.scp data/$dataset/images_tmp.scp
+    sort data/$dataset/utt2spk_tmp > data/$dataset/utt2spk
```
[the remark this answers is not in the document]
aarora8: Should I add the similar script in images folder: 'image/fix_data_dir .sh', 'image/validate_date_dir.sh'

danpovey:
yes.

egs/madcat_ar/v1/local/train_lm.sh
```diff
@@ -0,0 +1,113 @@
...
+
+  echo "$0: Getting the Data sources"
+
+  rm ${dir}/data/text/* 2>/dev/null || true
+
+  # use the validation data as the dev set.
+  # Note: the name 'dev' is treated specially by pocolm, it automatically
+  # becomes the dev set.
+
+  cat data/dev/text | cut -d " " -f 2-  > ${dir}/data/text/dev.txt
```
hhadian: If we are using the dev text here, we should never report WERs for the dev set. I guess it's safer to use, say 5000 lines from the train text here and the rest for training.

egs/cifar/v1/image/get_image2num_frames.py
```diff
@@ -14,7 +14,7 @@
 import os
 import sys
 import numpy as np
-from scipy import misc
+from PIL import Image
```
danpovey: I'm a little surprised to see you replace scipy with PIL.  I though we were going to move, in the long term at least, from PIL to scikit-image, at least I see a conversation in my email about this.  @hhadian might want to comment.

Also, I won't let this block merging this PR, but we should add a script to check python library dependencies.
One options is to have a script image/check_dependencies.sh and have it check dependencies for all the things used in the image/ scripts-- we can add options to enable or disable certain checks as needed depending on where it's called form. @hhadian, what do you think?

aarora8: sorry, In one of the online forms, it was mentioned that we can find the size of the image without loading it into memory (https://stackoverflow.com/questions/15800704/get-image-size-without-loading-image-into-memory). Since in the get_image2num_frames.py script, it was obtaining size of images, I updated it after discussing with @hhadian. It was saving around 100 mins for madcat arabic data. 

get_image2num_frames with PIL is taking around 70 mins
get_image2num_frames with scipy is taking around 168 mins

jtrmal:
PIL is dead.
y.

danpovey:
I thought scikit-image was supposed to contain a copy of PIL.
Also, I hope you are running that script with some kind of parallelization?

jtrmal:
did you test the imageio which now seems to be the suggested way how to do
IO with bitmaps? It seems to include PIL or some subset of this, but in
contrast to PIL it's maintained.
y.

egs/cifar/v1/image/get_image2num_frames.py
aarora8: I was planning to add parallelization, I was not able to get a chance. 
I will try imageio and add parallelizartion

aarora8: sorry, I tried skimage for create_line_image_from_page_image.py, but I was getting an error which i was not able to fix. so i shifted back to PIL there also. I will fix it in both places.

jtrmal:
BTW, Ashish, what format are those pictures?
Thanks, y

egs/cifar/v1/image/get_image2num_frames.py
aarora8: It is in .tif format. I was able to do all the processing required in create_line_image_from_page_image. I was able to view the processed line images in my local machine. But while saving, the image it was saving a completely black image after rounding off the values which were close to zero.

aarora8: I tried to find an option in save function in skimage which will fix it but i was not able to find a particular option.

jtrmal:
OK, I don't know what the problem is, but as a general suggestion, do not
use tiff to save anything, those files can have a very complex structure
with a lot of different extensions and as a result, it's hard to generate a
tiff file that will be interpreted correctly by all programs.
Png would be probably more suitable for that.
y.

hhadian:
> I'm a little surprised to see you replace scipy with PIL. I though we were going to move, in the long term at least, from PIL to scikit-image, at least I see a conversation in my email about this. @hhadian might want to comment.
> 
> One options is to have a script image/check_dependencies.sh and have it check dependencies for all the things used in the image/ scripts-- we can add options to enable or disable certain checks as needed depending on where it's called form. @hhadian, what do you think?

As Ashish explained, it was because scipy inefficiently would load the whole image just to get its size and since, AFAIK, PIL is required (when we use the recommended `imageio`) regardless, I imagined it would make sense to use it here. 
As an alternative approach (to `get_image2num_frames.py`) I think we can skip this step and get the features (w/o enforcing image lengths to allowed lengths) and then compute `image2num_frames` from the features (which will be fast), and finally enforce the utterance length when getting the e2e egs in `nnet3-chain-e2e-get-egs.cc`. This can work for OCR because the beginning/end frames are always white pixel padding and we can simply repeat them.

I agree with ` image/check_dependencies.sh`. We can remove `local/check_tools.sh` and instead call that in run.sh. Also, we can move `local/make_features.py` to `image/` because I guess all the OCR recipes use the same copy (is that right Ashish?).

Hossein

aarora8:
local/make_features.py  is almost same for OCR recipes. We can add option like,  adding Gaussian noise (UW3), for IFN (it might need option to flip image).

Thanks,

Ashish

________________________________

jtrmal:
I wonder if we should use imagemagick or similar package for either
converting the image  on-the-fly to some easier format and/or querying the
image properties (size for example).
I know it adds another dependency, but it's a fairly common package and it
might simplify the whole issue for us.
y.

danpovey:
Guys, let's leave PIL for now and sort this out later.
It's not a deep dependency.

egs/madcat_ar/v1/local/chain/run_cnn_chainali_1b.sh
```diff
@@ -0,0 +1,226 @@
+#!/bin/bash
+
```
danpovey: @aarora8, there should be results at the top of these files, obtained from the compare_wer.sh script.

danpovey:
you can just put the current results for now and change it later.
But I think you should create a tuning/ directory and make those run*
scripts links to it.  I.e. all 1a and 1b suffixes (etc.) would be within
the tuning/ directory and there would be soft links from one directory
above.
Experiments that differ only in suffix like 1a or 1b are supposed to be
diffferent versions of the same experiment (differently tuned).  And you
generally won't change the results after doing the initial experiment.  At
least that's how we normally do it.
But before we merge this, it's OK to just take the best current experiment
and make it 1a.

aarora8:
ok, thank you. I will add the results and make it 1a.

________________________________

egs/madcat_ar/v1/local/chain/run_cnn_chainali_1b.sh
aarora8: sorry, I realized I haven't run run_cnn_1a.sh, run_cnn_chainali_1b and run_cnn_end2endali_1a, with the recent code. I have currently ran run_flatstart_cnn1a.sh and run_cnn_end2endali_1b (UC) scripts. I updated the results on those two scripts, I will run and update the results for other scripts aswell.

egs/madcat_ar/v1/local/train_lm.sh
aarora8: I tried using first 5000 lines from the train text here and the rest for training, but the first 5k lines are occurring again in remaining training. train_lm.sh, gives an error due to it. currently, to remove the error, I reverted the change. I was working on a task of adding Arabic gigaworld corpus text data in the language model. currently, it is not complete. should i update it later with that some portion of Arabic gigaworld corpus text data.

danpovey:
Yes, update it later.

PR #2359, opened 2018-04-15
Use member function in FrameOptions for window size.

wanguanglu:


src/feat/mel-computations.cc
```diff
@@ -37,13 +37,10 @@ MelBanks::MelBanks(const MelBanksOptions &opts,
   int32 num_bins = opts.num_bins;
   if (num_bins < 3) KALDI_ERR << "Must have at least 3 mel bins";
   BaseFloat sample_freq = frame_opts.samp_freq;
-  int32 window_length = static_cast<int32>(frame_opts.samp_freq*0.001*frame_opts.frame_length_ms);
-  int32 window_length_padded =
-      (frame_opts.round_to_power_of_two ?
-       RoundUpToNearestPowerOfTwo(window_length) :
-       window_length);
+  int32 window_length = frame_opts.WindowSize();
```
danpovey: is 'window_length' needed any more?  remove the variable if not.
please check the code carefully, that this would not change its behavior-- i.e. see if the member function is equivalent, and if not I should look into it more closely.

wanguanglu: yes, `window_length` is not used, I will removed it later.
I have checked the code and tested it, they are equivalent.

Another question, why don't we add `-Wunused-value` for unused variable check?

PR #2362, opened 2018-04-16
Minor issues in the Regex

adelra:
I was facing an error on this line so I fixed it this way, due to some issues regarding Wellington Corpus I still cannot verify the output but the error is gone :)


egs/iam/v1/local/prepare_data.sh
```diff
@@ -147,7 +147,7 @@ dict=OrderedDict([("^",""), ("|",""), ("_",""), ("*0",""), ("*1",""), ("*2",""),
   ("*+$",""), ("$",""), ("*+","£"), ("*-","-"), ("*/","*"), ("*|",""), ("*{","{"), ("*}","}"),
   ("**#",""), ("*#",""), ("*?",""), ("**\"","\""), ("*\"","\""), ("**'","'"), ("*'","'"),
   ("*<",""), ("*>",""), ("**[",""), ("**]",""), ("**;",""), ("*;",""), ("**:",""), ("*:",""),
-  ("\\\0",""), ("\\\15",""), ("\\\1",""), ("\\\2",""), ("\\\3",""), ("\\\6",""), ("\\\"",""),
+  ("\\0",""), ("\\15",""), ("\\1",""), ("\\2",""), ("\\3",""), ("\\6",""), (r"\"",""),
   ("{0",""), ("{15",""), ("{1",""), ("{2",""), ("{3",""), ("{6","")]);
```
hhadian: @ChunChiehChang, I was wondering if there is an easier way for removing the annotations?

ChunChiehChang: I'm not too sure. I couldn't find a file without the annotations. Most of the annotations convey information such as italics, bold, subscripts, and things like that. The GUIDE.txt file from the corpus has a list of all their added annotations. I haven't tried any experiments without removing the tags but I thought it was better to remove them.

danpovey: I'm a little unclear how this ever worked.  Could this be due to python version differences in how they treat escaping in strings?  
And can you test on our grid to see if it works for us?
I wish Guido would stop messing with core language features.

danpovey: oh wait, I think the bug he likely had was the one that existed already that I already merged a fix for, which he just fixed a different way (the last thing on the line), but he also probably fixed an issue in the other things on that line that meant they previously would not have matched the text properly.
Let me know if I should merge this though-- please test.

adelra: @danpovey the second one fixes another issue in the regex plus fixing the issue in another way (because the first patch had an unexpected output) but I am not sure if this was the desired output since I don't have access to the corpus.

danpovey: Thanks a lot for this fix, but it looks like @ChunChiehChang  has taken your fix in his separate PR now, that also resolves other issues, so I'll close this PR.

PR #2368, opened 2018-04-18
Updating code on TDNN runner scripts of HKUST nnet3 repo

underdogliu:
Motivation: I was playing with HKUST repo and found the TDNN scripts are out-dated (while most other egs are not, e.g. aishell) with low robustness on execution so I made the following simple modifications without changing the logistic essentially. Hope this will benefit future users.

Things done: update run_tdnn.sh in HKUST nnet3 scripts, for both xent and chain-xent trials. The code has been executed with similar results on RESULTS obtained (will not update RESULTS though since still there are variations at numerical level of 0.01). This is the first time I've made PR on Kaldi as a user for while. Please make sharp comments!

Future TODO: If needed, will update LSTM scripts as well, which is slightly harder task.


danpovey:
Thanks a lot for doing this, but can you please do it slightly differently...
- make a tuning directory and make run_tdnn.sh a soft link.  make the current one _1a.sh and the new one _1b.sh and put results from both.
- add a compare_wer.sh script.  these are somewhat dataset dependent, look at the mini_librispeech one for an example but you'll have to change it.
- for the chain scripts, use the latest wsj recipe as an example.  (note: will update this soon but use the current one).
-  i notice some things are not quite right in the script, e.g. frames_per_eg should be a comma-separated list.  other training options may not be right, I suggest to  take them from e.g. the wsj example.
- in the run_ivector_common.sh script it is using steps/train_lda_mllt.sh, it should be using a PCA transform, look to other example scripts for details.  

if you aren't able to do this I may try to find someone else to do it.

underdogliu:
@danpovey I'll take care of it and make a WER comparison of TDNN in the script. Thanks!

underdogliu:
@danpovey I'm working on filling the requirements. But maybe I'm a little bit lack of knowledge but I can't tell why PCA is needed here - theoretically yes PCA gives more flexibility in terms of adaptation and unsupervised learning but I doubt it will be a big issue here. Practically also, I found at least switchboard and ami are using LDA as default for now (ami is providing options though - maybe I wanna do in same way).

jtrmal:
Actually, we are using PCA in the newer version of the recipes. The older
recipes are using LDA.
Please use PCA and do not bother with LDA -- it's much faster and we didn't
observe any negative impact on WER>
y.

underdogliu:
@jtrmal OK I'm still brousing and found WSJ is using PCA. But if you can explain more about the motivation of doing so please? It will be very helpful. Thank you!

jtrmal:
Mostly, it's because it greatly simplifies the ivector training script --
do not try to read too much into the fact that some of the recipes are
using LDA -- those recipes haven't been upgraded yet, owing to our limited
bandwidth. Typically, the mini_librispeech scripts will be the most current
ones so try to use/modify those.
y.

underdogliu:
@danpovey @jtrmal I will run the experiment and finish the number filling some point next week. If you have time please have a look in advance for style checks. Thank you!

danpovey:
Don't worry about those older scripts.
Small differences like 0.2% are expected when we rerun things-- it's not
deterministic.  Don't worry.

underdogliu:
so I've made code modifications with results updated in the script (not in RESULTS though). I guess the results are not good (0.2% and 1% degradation for nnet3 and chain-xent, respectively) potentially because of the fact that it was ran in my place and maybe there are some degradations in terms of resource saturation or something else (no bugs reported though). So if anyone interested please have a shot and update results if you got stronger ones. Also @danpovey  please make sharp comments if you see anything weird. Thanks!

egs/hkust/s5/local/chain/tuning/run_tdnn_2a.sh
```diff
@@ -0,0 +1,227 @@
+#!/bin/bash
+# Copyright 2018  Emotech LTD (Author: Xuechen Liu)
+
+# This script is based on run_tdnn_7h.sh in swbd chain recipe.
+# exp 2a: change the step of making configs, using xconfig
+#         some minor changes on training parameters, referencing wsj
+
+# Results
+# local/nnet3/compare_wer_general.sh --online exp/chain/tdnn_7h_chain_2asp
```
danpovey: this compare_wer_general.sh script doesn't seem to exist and the results don't seem very consistent.

I'm surprised you took such an old Swbd script as the source.  The latest one probably would have given better results.

underdogliu: Thanks for the reply!

I will re-run online decoding for a further check. 

The swbd comment is inherited from old script and I forogt to delete it. The one I modified from is actually run_tdnn_1a.sh in chain/nnet3 directory, which was the one for hkust results.

Speaking of output of chain info:
```
exp/chain/tdnn_7h_sp: num-iters=435 nj=4..4 num-params=11.8M dim=43+100->4000 combine=-0.100->-0.100 (over 9) xent:train/valid[289,434,final]=(-1.79,-1.76,-1.75/-1.72,-1.71,-1.71) logprob:train/valid[289,434,final]=(-0.119,-0.107,-0.107/-0.153,-0.155,-0.154)
```
(the one above is actually 'chain_2a')
```
exp/chain/tdnn_7h_chain_2b_sp: num-iters=435 nj=3..4 num-params=11.8M dim=43+100->4000 combine=-0.114->-0.114 (over 3) xent:train/valid[289,434,final]=(-1.74,-1.71,-1.71/-1.95,-1.96,-1.95) logprob:train/valid[289,434,final]=(-0.124,-0.114,-0.114/-0.172,-0.177,-0.176)
```

danpovey:
Your system is overtraining quite a bit, probably because there is no model-level l2 regularization (this is an older recipe).
I suspect what would work quite a bit better would be to take the current swbd recipe which is 7p (make sure your code is up to date), and reduce the dims from 1280->1024 and 1536->1280.

It would be good if you would add a compare_wer.sh script (look at the mini_librispeech one for inspiration but they are somewhat dataset dependent due to scoring differences).

Once you get a good recipe there is no need to include intermediate experiments, and in fact you can just delete the old recipe and make your best new recipe 1a, because the old recipe wasn't working.

underdogliu:
@danpovey got it.

egs/hkust/s5/local/nnet3/compare_wer_general.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+# Copyright 2018  Emotech LTD (Author: Xuechen Liu)
```
danpovey: Please add another script local/chain/compare_wer.sh (or just move this and modify it if there are no non-chain nnet3 example scripts).
Make the name 'compare_wer.sh' not compare_wer_general.sh.
The chain compare_wer.sh script will compare the chain and xent objective functions, see other examples.

egs/hkust/s5/local/nnet3/tuning/run_tdnn_1a.sh
```diff
@@ -0,0 +1,135 @@
...
+if [ $stage -le 0 ]; then
+  local/nnet3/run_ivector_common.sh --stage $stage \
+    --ivector-extractor exp/nnet3/extractor || exit 1;
+fi
+
+if [ $stage -le 8 ]; then
+  echo "$0: creating neural net configs";
+
+  # create the config files for nnet initialization
+  python steps/nnet3/tdnn/make_configs.py  \
```
danpovey: is this the script you meant to put as 1a?  this looks like the old, non-working one.
I was suggesting to just discard that and put the current best one.  but I guess you have not finished tuning.

underdogliu: Oh sorry...It's in the plan to remove it. Haven't push the changes to the online repo though

danpovey:
Thanks!

PR #2370, opened 2018-04-18
[WIP] Making Wellington Corpus optional in iam_ocr

ChunChiehChang:
Finished making the wellington corpus optional and fixed some regex issues. I still need to update the results and write a cleaner regex command.


danpovey:
please resolve conflicts.

adelra:
@ChunChiehChang  I took a quick look at the corpus, correct me if I'm wrong, wouldn't just this work for the cleaning part: `<(.*?)>`

ChunChiehChang:
There are some cases where the "{0" and "}" pairs of annotation tags spans multiple lines and just to be safe I did all of the regex parts separately. There are also some issues with the original corpora where the annotations have typos. For example, some open brackets do not have a corresponding close bracket.

PR #2377, opened 2018-04-20
Use generic fst::Fst<Arc> superclass when fst.Type() is not const or vector

mike-clark-8192:
There was a typo made when the lattice decoders were being updated to support const FSTs.  When the fst.Type() is not "vector" or "const" the code is defaulting to fst::ConstFst.  This is incorrect.  It should default to fst::Fst.  I made minimal changes to fix this mistake.


danpovey:
Thanks!  Will merge when checks finish.

mike-clark-8192:
Travis CI_TARGETS="all text" timed out because it ran on too long.  I don't think this was the fault of my changes, but I am new to this.

jtrmal:
I have restarted the build, hopefully, it will finish now.
y.

PR #2378, opened 2018-04-21
Fix some comment mismatch.

wanguanglu:


src/hmm/transition-model.h
```diff
@@ -288,8 +288,8 @@ class TransitionModel {
 
   HmmTopology topo_;
 
-  /// Triples indexed by transition state minus one;
-  /// the triples are in sorted order which allows us to do the reverse mapping from
+  /// Tuples indexed by transition state minus one;
+  /// the tuples are in sorted order which allows us to do the reverse mapping from
   /// triple to transition state
```
danpovey: there is one  more 'triple' on this line.

wanguanglu: yeal, I fixed it.

wanguanglu:
I don't know why the continuous integration sometimes fail?

danpovey:
the travis build sometimes fails due to issues in how it was set up.  i am hoping that e.g. @galv might have a chance to figure out what's going wrong-- I haven't had the time.

PR #2381, opened 2018-04-23
[WIP] DFSMN related code and example scripts

GaofengCheng:
Some information : 
    I tried DFSMN (https://arxiv.org/pdf/1803.05030.pdf) on 1000hrs Chinese CTS (Nnet3 CE), which has context [-57,57]. The semi-orth-DFSMN reached WER 25.17%. (3layer-1024-128-128 BLSTM : WER 25.2%; 3layer-1024-256-256 BLSTM : WER 24.6%).  Semi-orth was found beneficial, so we used semi-orth-DFSMN, see TDNN-F (http://www.danielpovey.com/files/2018_interspeech_tdnnf.pdf).

The DFSMN exps on Switchboard are still running. The DFSMN related issues will be discussed under this PR.


Eric-Chen-007:
There is a "$opts2" option in configs of DFSMN. But i can't find any initialization or assignment for it.

GaofengCheng:
ignore $opts2, we do not use any regularization in the first layer.

Eric-Chen-007:
@GaofengCheng , get it. Thanks.

GaofengCheng:
@danpovey 
en, the l2 is in a mess.
I will try one with uniform l2 value (double for output).
Gaofeng

fanlu:
@GaofengCheng 
Will you implement dfsmn + chian model?
Why do you use steps/nnet3/train_rnn.py to implement dfsmn? this version doesn't have frame_subsampling_factor and alignment_subsampling_factor config?
How can we use Lower Frame Rate?
thanks

GaofengCheng:
@fanlu 
Will you implement dfsmn + chian model?
    I have already tried it under chain, and compared it with the tdnn_7q (switchboard). And I will not make the Chain-DFSMN PR. 
    The bad news is DFSMN is little worse than skinny TDNN-F (the difference is not large, 0.2~0.3 abs), at least on the Switchboard task.  (This should be tuning problem)
    The good news is DFSMN can be smaller if you want a very broad context, like [-100,100]. 
    DFSMN is like a diagonal-TDNN-F with skip connections, so I think there should not be big difference between DFSMN and TDNN-F. Choosing which one depends on your favor.

Why do you use steps/nnet3/train_rnn.py to implement dfsmn? this version doesn't have frame_subsampling_factor and alignment_subsampling_factor config?
How can we use Lower Frame Rate and LFMMI?

    Using train_rnn is for the purpose of reducing redundant context computation. 
    I do not remember exactly whether there is the sub-sampling factor option in this python scripts. But I do implement the LFR system under similar setup. Maybe you should try more.

danpovey:
> @fanlu <https://github.com/fanlu>
> Will you implement dfsmn + chian model?
> I have already tried it under chain, and compared it with the tdnn_7q
> (switchboard). And I will not make the Chain-DFSMN PR.
> The bad news is DFSMN is little worse than skinny TDNN-F (the difference
> is not large, 0.2~0.3 abs), at least on the Switchboard task. (This should
> be tuning problem)
>

We are still trying DFSMN with other configurations.  I asked Gaofeng to
try it on a smaller setup.  I still have hope that with the right settings,
and the right initialization, it might be better than TDNN-F.

The good news is DFSMN can be smaller if you want a very broad context,
> like [-100,100].
> DFSMN is like a diagonal-TDNN-F with skip connections, so I think there
> should not be big difference between DFSMN and TDNN-F. Choosing which one
> depends on your favor.
>
> Why do you use steps/nnet3/train_rnn.py to implement dfsmn? this version
> doesn't have frame_subsampling_factor and alignment_subsampling_factor
> config?
> How can we use Lower Frame Rate and LFMMI?
>

The chain system is what has lower frame rate and LF-MMI.  You should use
steps/nnet3/chain/train.py in that case.  Gaofeng has some scripts for
this, maybe he can add them to his PR.

fanlu:
@GaofengCheng Thanks for sharing your experiment
when I use aishell chain model, the num-params is 12.4M
But dfsmn_1b_sp's num-params is 22.1M，almost twice more than chain but get similar cer
So I want to know whether the chain will help dfsmn.
It will very appreciate if you make the chain-dfsmn pr.
@danpovey Tanks for your reply，Your work if very helpful to us, Thank you very very much.

naxingyu:
What's the conclusion of this PR?

danpovey:
@GaofengCheng  did a bunch more experiments with variants of D-FSMN and we were not able to find a scenario where it was better than TDNN-F, so I decided not to merge.

GaofengCheng:
I didn't find FSMN(and its variants) better than TDNN-F, that is to say they perform similarly, also we don't want to maintain too many achitectures at the same time. So I stopped this work.

underdogliu:
@GaofengCheng Speaking of that, I also tried this one due to curiosity. Yes they reached similar performance but Semi-orth DFSMN actually has smaller size, say roughly 30% fewer parameters. Guess it's thus still useful.

danpovey:
I think that's one of the things Gaofeng looked at-- whether, after tuning,
DFSMN might be more compact than a well-tuned TDNN-F.  But I don't recall
that we had positive results on that.  I might be wrong though.

GaofengCheng:
@underdogliu DFSMN is more compact than TDNN-F because we use diagonal-matrix for DFSMN. I also tried using diagonal-TDNN-F, it's very compact and also good performance.
All in all, the difference is very small. Both good, both recommended.
And the example is here, it should be very easy if people want to try DFSMN under nnet3.

ucasyouzhao1987:
@GaofengCheng Did you try very deep fsmn？ Like 30 layers dfsmn.

whaozl:
@GaofengCheng this dfsmn have keras layer?

PR #2383, opened 2018-04-23
[WIP] Add support for unconstrained-egs for chain training + example recipes

hhadian:
This adds a few scripts (with results) which use unconstrained-egs (in PR #2341).

### DIFF
--- a/egs/iam/v1/local/chain/run_cnn_chainali_1d.sh
+++ b/egs/iam/v1/local/chain/run_cnn_chainali_1d.sh
@@ -1,0 +1,249 @@
+#!/bin/bash
+
+# chainali_1d is as chainali_1c except it uses unconstrained egs
+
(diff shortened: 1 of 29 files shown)


danpovey:
Cool.
I'll merge this soon.

danpovey:
Looks like the test code doesn't compile-- my branch has the same issue.  Can you please address that?

francisr:
I haven't followed closely the whole thing, but it seemed to me that the time constraint was found to be useful when LF-MMI was first made, why is it better without it now?

hhadian:
The time constraints are still used but indirectly, i.e. we can still change left/right tolerance but they are less likely to make a difference, I guess. The supervision FSTs are converted to unconstrained FSTs (i.e. e2e-style) only after time constraints are applied and after they are split up into chunks.

danpovey:
If you can quickly add any tests, e.g. that the I/O works, then do it.  Just don't make it a blocker.

PR #2387, opened 2018-04-26
add wordlists mismatch test for rnnlm rescoring

hainan-xv:


scripts/rnnlm/lmrescore.sh
```diff
@@ -72,6 +72,12 @@ awk -v n=$0 -v w=$weight 'BEGIN {if (w < 0 || w > 1) {
   print n": Interpolation weight should be in the range of [0, 1]"; exit 1;}}' \
   || exit 1;
 
+if [ ! -z "$(head -n -1 $rnnlm_dir/config/words.txt | diff $oldlang/words.txt -)" ]; then
```
danpovey: It would be cleaner here to use cmp, e.g.
`if ! head -n -1 $rnnlm_dir/config/words.txt | cmp $oldlang/words.txt  -; then`

PR #2392, opened 2018-04-28
make numerator computation in the e2e code more effective

jtrmal:


jtrmal:
@danpovey and @hhadian 
I'm creating it now just in case the cluster will go down but should be mostly ready anyway.  Please comment.

src/base/kaldi-math.h
```diff
@@ -183,6 +183,11 @@ inline Float RandPrune(Float post, BaseFloat prune_thresh,
 
 inline double LogAdd(double x, double y) {
   double diff;
+  if (!std::isfinite(y))
```
danpovey: I'm concerned about this.. there are other parts of the code where we call LogAdd, and these changes will make it slower.  Was this really necessary?

src/chain/chain-generic-numerator.h
```diff
@@ -98,71 +96,79 @@ class GenericNumeratorComputation {
...
-
   // Defining this constant as an enum is easier.  it controls a memory/speed
   // tradeoff, determining how many frames' worth of the transposed derivative
   // we store at a time.  It's not very critical; the only disadvantage from
   // setting it small is that we have to invoke an AddMat kernel more times.
   enum { kMaxDerivTimeSteps = 8 };
 
-  // sets up the alpha for frame t = 0.
-  void AlphaFirstFrame();
+  // for the remapped fsts, copy the appropriate activations to CPU memory
```
danpovey: I think this needs quite a bit more documentation.  And say what the 'remapped fsts' are, e.g. provide a pointer to where the concept is explained.

src/chain/chain-generic-numerator.h
```diff
@@ -98,71 +96,79 @@ class GenericNumeratorComputation {
...
 
-  // sets up the beta for frame t = num_time_steps_.
-  void BetaLastFrame();
+  // the beta computation for 0 <= t < supervision_.frames_per_sequence
+  // for some 0 <= seq < supervision_.num_sequences.
+  void BetaLastFrame(int seq);
 
-  // the beta computation for 0 <= beta < num_time_steps_.
-  void BetaGeneralFrame(int32 t);
+  // the beta computation for t = supervision_.frames_per_sequence
```
danpovey: Are these comments swapped somehow?  Surely BetaGeneralFrame should not be for the last frame.
And I'm not sure it makes sense to have two functions?

jtrmal:
I will remove it, should be working just fine without it as well -- I was
testing something, but now I cannot recall what.
y.

src/chain/chain-generic-numerator.h
```diff
@@ -98,71 +96,79 @@ class GenericNumeratorComputation {
...
-  void BetaGeneralFrameDebug(int32 t);
+  void BetaGeneralFrameDebug(int32 seq);
 
 
   const Supervision &supervision_;
 
-  // the transposed neural net output.
-  Matrix<BaseFloat> exp_nnet_output_transposed_;
+  // a reference to the nnet output.
+  const CuMatrixBase<BaseFloat> &nnet_output_x_;
```
danpovey: what does the x_ mean?  variable names should generally be based on words, anyway.

src/chain/chain-generic-numerator.h
```diff
@@ -98,71 +96,79 @@ class GenericNumeratorComputation {
...
+  std::vector<Matrix<BaseFloat>> nnet_output_deriv_;
 
   // forward and backward probs matrices. These have the
   // same dimension and functionality as alpha_ and beta_
   // in DenominatorComputation except here we don't use beta
   // sums (becasue we don't use leakyHMM). However, we use
   // alpha sums to help avoid numerical issues.
-  Matrix<double> alpha_;
-  Matrix<double> beta_;
+  std::vector<Matrix<BaseFloat>> alpha_;
```
danpovey: Is it necessary to have  these things as class members instead of as local variables inside functions (possibly passed between functions as necessary)?  It seems wasteful in terms of memory, although I know in the big picture it is trivial.

src/nnet3/nnet-optimize-utils.cc
```diff
@@ -696,18 +696,15 @@ void RenumberComputation(NnetComputation *computation) {
 }
 
 
+bool IsNoop(const NnetComputation::Command &command) {
```
danpovey: I'd change to 'static bool' to clarify it's not declared in the header.

src/chain/chain-generic-numerator.h
jtrmal: This is the only part that can take more time -- for now, I didn't change the way this class is used (in chain/chain-training.cc), i.e. the interface is the same. 

I can change it so that we can have only one Matrix as a member variables (instead of vector<Matrix>) or even make a new member function (say with signature ForwardBackward(int sequence_id, CuMatrix *out)) which will have only local variables alpha, beta and probs (they will need to be allocated for each sequence again).

danpovey:
The ForwardBackward approach sounds good.

src/chain/chain-generic-numerator.cc
```diff
@@ -98,246 +92,269 @@ GenericNumeratorComputation::GenericNumeratorComputation(
...
-
-  for (int32 seq = 0; seq < num_sequences; seq++) {
-    double inv_arbitrary_scale = prev_alpha(num_states, seq);
+    num_sequences = supervision_.num_sequences,
+    num_frames = supervision_.frames_per_sequence;
+
+  KALDI_ASSERT(seq >= 0 && seq < num_sequences);
+  CopySpecificPdfsIndirect(seq, index_to_pdf_[seq], &(probs_[seq]));
+
+  SubMatrix<BaseFloat> alpha(alpha_[seq],
```
hhadian: I guess something that can make the code even nicer, is to have a separate `alpha_sum` variable (maybe `std::vector<Vector<BaseFloat> > alpha_sum`) which stores the alpha-sums. I guess it will simplify some parts of the code. 
If you want to merge, I can do this in a separate PR?

hhadian: Actually, I'm not even sure if alpha sums help (with numeric stability) in log-space?

jtrmal:
Yeah, me neither -- I will prepare a separate PR that will remove the sums
computation.  That way, we can easily return back if we find out it is
needed even for the log-space
y.

jtrmal:
when the checks finish, it should be ready for review/merge

src/chain/chain-generic-numerator.cc
```diff
@@ -33,59 +37,42 @@ GenericNumeratorComputation::GenericNumeratorComputation(
...
-    exp_nnet_output_transposed_gpu.ApplyExp();
-    exp_nnet_output_transposed_.Resize(nnet_output.NumCols(),
-                                       nnet_output.NumRows(), kUndefined);
-    exp_nnet_output_transposed_.CopyFromMat(exp_nnet_output_transposed_gpu);
-  }
 
   using std::vector;
-  int32 B = supervision_.num_sequences,
-      num_frames = supervision_.frames_per_sequence;
+  int B = supervision_.num_sequences;
```
danpovey: I know this is existing code, but let's rename B to num_seq.

src/chain/chain-generic-numerator.cc
```diff
@@ -98,246 +85,294 @@ GenericNumeratorComputation::GenericNumeratorComputation(
...
-               supervision_.e2e_fsts[seq], s);
+             supervision_.e2e_fsts[seq], s);
            !aiter.Done();
            aiter.Next()) {
         const fst::StdArc &arc = aiter.Value();
         DenominatorGraphTransition transition;
-        transition.transition_prob = exp(-(arc.weight.Value() - offset));
-        transition.pdf_id = arc.ilabel - 1;
+        transition.transition_prob = -(arc.weight.Value() - offset);
+        int pdf_id = arc.ilabel - 1;
```
danpovey: make this int32, and please add a comment like
// note: the FST labels were pdf-id plus one.

src/chain/chain-generic-numerator.h
```diff
@@ -81,90 +100,94 @@ namespace chain {
...
+
+  // returns total prob for the given matrix alpha (assumes the alpha
+  // matrix was computed using AlphaFirstFrame() and AlphaGeneralFrame()
+  // (it's exactly like 'tot_probe_' in DenominatorComputation)
+  BaseFloat GetTotalProb(const Matrix<BaseFloat> &alpha);
 
   // some checking that we can do if debug mode is activated, or on frame zero.
   // Sets ok_ to false if a bad problem is detected.
-  void BetaGeneralFrameDebug(int32 t);
+  bool BetaGeneralFrameDebug(int32 seq,
```
danpovey: This can be renamed to something like CheckValues(); and make it const if it doesn't change any class members.

src/chain/chain-generic-numerator.h
```diff
@@ -81,90 +100,94 @@ namespace chain {
...
-  // sets up the beta for frame t = num_time_steps_.
-  void BetaLastFrame();
-
-  // the beta computation for 0 <= beta < num_time_steps_.
-  void BetaGeneralFrame(int32 t);
+  void AlphaFirstFrame(int seq, Matrix<BaseFloat> *alpha);
+
+  // the alpha computation for 0 < t <= supervision_.frames_per_sequence
+  // for some 0 <= seq < supervision_.num_sequences.
+  BaseFloat AlphaGeneralFrame(int seq,
```
danpovey: please rename GeneralFrame -> RemainingFrames wherever it occurs here.

src/chain/chain-training.cc
```diff
@@ -75,24 +75,17 @@ void ComputeChainObjfAndDerivE2e(const ChainTrainingOptions &opts,
...
-    } else if (nnet_output_deriv && numerator_ok) {
-      numerator_ok = numerator.Backward(nnet_output_deriv);
-      if (!numerator_ok)
-        KALDI_LOG << "Numerator backward failed.";
+    } else {
+      numerator_ok = numerator.ForwardBackward(&num_logprob_weighted,
+                                               nnet_output_deriv);
     }
+    if (!numerator_ok)
+        KALDI_LOG << "Numerator forward-backward failed.";
```
danpovey: I'd make this KALDI_WARN.

jtrmal:
I implemented the proposed idea with collapsing the transfers into one + did some further some tweaks (removed one transposition and used the Swap() trick)

PR #2395, opened 2018-04-30
iam_ocr: Updating results on top of scripts

ChunChiehChang:
Also moved regex command to a separate python script. The previous PR that I made for adding the wellington corpus didn't include the updated results on the top of the scripts.


jtrmal:
I have to admit I don't track this very carefully, but wasn't discussed the wellington corpus should be an optional part? From the changes, it seems it's not optional.

PR #2398, opened 2018-05-01
[WIP][egs][aishell2] AIShell2 database and recipe

dophist:
hi Dan & Kaldi community,

After aishell1, we are planning to release a much larger open-source Mandarin corpus (AIShell-2). Corpus & recipe descriptions are attached below, currently we are still cleaning up the scripts and codes.

Any comments and suggestions are welcomed.

---

# AISHELL-2

AISHELL-2 is by far the largest open-source speech corpus available for Mandarin ASR research.
### training data
* 1000 hours of speech data (around 1 million utterances)
* 1991 speakers (845 male and 1146 female)
* clean recording environment(studio or quiet living room)
* read speech
* reading prompts from various domain: entertainment, finance, technology, sports, control command, place of interest etc.
* near field recording via 3 parallel channels(iOS, Android, Microphone).
* data from iOS channel is open-sourced for research community

### evaluation data:
Currently we open-sourced AISHELL2-2018A-EVAL, containing:
* dev: 2500 utterances from 5 speaker
* test: 5000 utterances from 10 speakers

you can download above evaluation set from:
http://www.aishelltech.com/aishell_eval

we may update and release other evaluation sets on the website later, targeting on different applications and senarios. 

## 2. RECIPE
Based on Kaldi standard system, AISHELL-2 provides a self-contained Mandarin ASR recipe, with:
* a word segmentation module, which is a must-have component for Chinese ASR systems
* an open-sourced Mandarin lexicon(DaCiDian)
* a simplified GMM training recipe
* 80-dim FBank without pitch as NN input feature
* slightly different data augmentation setup(tempo perturbation, max-volume perturbation)
* acoustic channel adaptation recipe(AM fine-tuning)
* a real-time streaming ASR demo via laptop microphone (tested on MacOS)

# CONTACT
AISHELL foundation is a non-profit online organization, with members from speech industry and research institutes.

We hope the corpus and recipes that we open-sourced could be beneficial to the entire speech community.

Due to the scale of AISHELL-2 (more than 100G data for the open-sourced iOS channel), we have decided to distribute AISHELL-2 via hard-disk delivery.

To apply for the AISHELL-2 disk, or you come across any problem with the recipe, contact us via:

aishell.foundation@gmail.com


danpovey:
Thanks a lot!
Just from looking at the file list I notice a couple of things that I think should be changed, e.g. you chose the name m5 instead of the more standard s5; and I see cross-entropy nnet3 systems but it's more normal to provide chain systems as the results would be better, and it's faster.
I would rather review it later on after changes.  I wonder if we can somehow get @underdogliu involved because he recently updated the HKUST example and some of the issues that he resolved there might also apply to this recipe.

dophist:
Hi Dan,

1. renaming to s5 done.
2. chain model is definitely our plan. For the moment, chain model is commented out in run.sh, because we are still tuning AM training setup and overall recipe. This may take some time due to the limited computation power we got in hand. We'll let you know when we get solid RESULTS file. 

For temporary recording, most recent results are listed below:

model | training time | dev(cer) | test(cer)
-- | -- | -- | --
mono | 0.5h | 62.93 | 62.20
tri1 | 0.5h | 27.70 | 27.90
tri2 | 1.4h | 25.72 | 26.12
tri3 | 2h | 22.29 | 22.48
tdnn | 25.5h | 11.75 | 11.18
lstm | 17.5h | 10.81 | 10.34
chain tdnn | 24.5h | 10.50 | 10.18

@underdogliu It would be great if you can share your experience with us on HKUST Mandarin ASR recipe, feel free to contact us via aishell.foundation@gmail.com, thanks a lot : )

underdogliu:
@dophist Hi ya! So I guess I'd like to get more information on the file structure you're currently using and the network recipe. I'll send a email to the address above. Please check whether inbox or span area.

naxingyu:
@underdogliu I'll refine the local/nnet3/train_tdnn_ft.sh script. Please leave it when as is. Thanks.

underdogliu:
@naxingyu No worries! At the meantime I'd like to work on chain model. I've already sent request of aishell2 data to the aishell mailbox to recover the baseline.

naxingyu:
@danpovey I've re-structured the nnet3/chain folders. Also, as per discussion in the forum, I've updated the nnet3 python lib to accept input model instead of initializing from config. This change is being tested. After this PR is merged, I'll make changes to other training python scripts accordingly.
@underdogliu I've updated this PR. Please check out. Thanks!

egs/aishell2/s5/local/chain/tuning/run_lstm_2a.sh
```diff
@@ -0,0 +1,217 @@
+#!/bin/bash
+
+# run_lstm_1a.sh is a first attempt at an LSTM system, based on xconfigs-- it's
```
danpovey: i think this file is supposed to be named 1a.

egs/aishell2/s5/local/data_augmentation/perturb_data_dir_speed.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
```
danpovey: I don't understand why these files in local/data_augmentation/ were included.

egs/aishell2/s5/local/nnet3/tuning/run_tdnn_aishell2_ft_3a.sh
```diff
@@ -0,0 +1,66 @@
+# !/bin/bash
+
+# This script uses weight transfer as a transfer learning method to transfer
```
danpovey: I don't think the name 'aishell2' should be part of this filename.
What is the motivation for  including this script?

egs/aishell2/s5/local/nnet3/tuning/run_lstm_2a.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
+
+#    This is the standard "lstm" system, built in nnet3; this script
```
danpovey: Was it necessary to include the non-chain nnet3 recipes?   They are generally quite a bit worse than the chain recipes. 
Some of the comments seem out of date (referring to cleanup).
Also: normally these directories would have results at the top, showing what the results are and how they differ from other results in the same experimental directory.
And it's a good idea to make a 'compare_wer.sh' script to compare the WER results and diagnostics in a neat way.  E.g. look at the mini_librispeech one.  You should always put the results in comments at the top of these types of files.

dophist:
Hi, Dan,

Thanks for your time and suggestions, a brief update here:

### Data augmentation:
I've merged code changes in local/data_augmentation into kaldi utils/data, which essentially:
* add tempo perturbation support into current sp tool: "speed" perturbation distorts vocal characteristics and pitch, where tempo does not. Tempo mode is more useful transforming read-speech wav(already with good ages/gender coverage) into spontaneous scenarios.

* add absolute volume perturbation support into current vp tool: with improper scale setup, current "scale" vp often results in saturated or silence-like wavs, this could be detrimental. With absolute volume control, vp can be more robust.

the above data augmentation enhancements are base on our experience from industry, and they should be useful to others.  The code change doesn't break compatibility, existed scripts and recipes should behave the same way as before, please have a review, anything need to be change further, just let me know, I'll fix them as quickly as I can.

### finetune recipe
the reason why we add a fine-tune recipe(feed new data to an existing model) is as follows:
* Firstly, since the release of aishell1, we have been asked over and over again, about how to adapt/tune aishell1 model with their own speech data, to improve the performance of their own system/service. these requests are mainly from startups or individuals who don't know Kaldi that well. So we hope the release of this fine-tune recipe serves as a drop-and-run tool for them.

* Secondly, from an industry point of view, we found run_{tdnn,lstm}.sh etc usually involve model training on tens of thousand hours of speech, which is costly in terms of time and computation power.  In company we normally train a base model from scratch every 2 or 3 month, when there are major algorithm/data updates.  However, the fine-tune recipe is pretty much being used everyday, to deal with acoustic mismatch from various speech applications or products.


### chain results
Besides that, during last few weeks we tweak chain model setup and the current results are:

model | test(cer)
-- | --
mono | 50.13
tri1 | 27.91
tri2 | 26.12
tri3 | 22.54
tdnn | 11.13
chain tdnn | 9.90

%WER 50.13 [ 24828 / 49532, 373 ins, 2618 del, 21837 sub ] exp/mono/decode_test/cer_8_0.0
%WER 27.91 [ 13824 / 49532, 463 ins, 1289 del, 12072 sub ] exp/tri1/decode_test/cer_9_0.0
%WER 26.16 [ 12959 / 49532, 515 ins, 1089 del, 11355 sub ] exp/tri2/decode_test/cer_9_0.0
%WER 22.54 [ 11166 / 49532, 335 ins, 1158 del, 9673 sub ] exp/tri3/decode_test/cer_11_0.0
%WER 11.13 [ 5513 / 49532, 113 ins, 428 del, 4972 sub ] exp/nnet3/tdnn_sp/decode_test/cer_10_0.0
%WER 9.90 [ 4904 / 49532, 70 ins, 929 del, 3905 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

egs/wsj/s5/utils/data/perturb_data_dir_volume.sh
```diff
@@ -68,6 +82,19 @@ if $volume_perturb_done; then
...
+  reco2vol=$data/reco2vol
+  wav-to-volume scp:$data/wav.scp ark,t:- | awk -v awk_seed=${local_seed}  \
+    -v awk_low=${abs_low} -v awk_high=${abs_high}                          \
+    -v awk_out=${reco2vol}                                                 \
+    'BEGIN{ srand(awk_seed); }                                             \
+    {                                                                      \
+      dst_vol=int(rand() * (awk_high - awk_low) + awk_low);                \
+      scale=(dst_vol * 1.0) / ($2 * 1.0);                                  \
+      print $1,scale > awk_out;                                            \
+    }'
```
danpovey: Wouldn't this be quite slow, since it would have to read all the wav files without parallelization?
Also, I think you can redirect the awk command to that file instead of doing the '> awk_out' inside the awk script, that's simpler.
But I'll ask separately for more details about why you are adding this.

danpovey:
I think the finetuning recipe is a good idea.  I'll have a look at the details later.

Regarding the tempo vs. speed perturbation: actually, changing the pitch and formant locations of the speech is the main point.  We did extensive experiments about this (see http://www.danielpovey.com/files/2017_icassp_reverberation.pdf) and found that speed perturbation was the best (comparing with the various other options available in sox).  So unless you have some compelling experimental evidence, I think you shouldn't be changing that.

Regarding the volume perturbation:  do you have experimental evidence that doing it your way is better?  The way the script is set up by default, it will (at a maximum) increase the volume only by a factor of 2; and if the waveform wasn't already clipped, it will likely not be very badly clipped after that.  And anyway, real recording are sometimes clipped (or very quiet), so I don't think it's really a problem.  In fact, it's kind of the point.  Do you have some kind of experimental evidence that doing it your way is better?

danpovey:
Oh, and I noticed another thing.
You are dumping filterbanks instead of MFCCs.  This is not a good idea.  MFCCs (if you use the full dimension) contain exactly  the same information as filterbanks, but they are more compressible; and when we dump features, we use compression.  The compression is lossy and you lose more when you do it on filterbanks.  You can lose, IIRC, about 0.5% WER absolute by using filterbanks instead of MFCCs, due to the compression.  Please change it to MFCCs.  And don't use 80-dimensional ones unless you can demonstrate an improvement over 40-dimensional ones which is what we normally use.  (40-dim MFCCs will be kinder on disk access, and we haven't, in the past, seen a difference from increasing the dimension above that.. but who knows, it may be different when you have a lot of Chinese data).  You also changed the window type to hamming, and set the dither option (although to the default).  Please take out those options, just for consistency with other recipes.    It won't make a difference.

underdogliu:
@danpovey Regarding of comments above, I suggest including it into some tuning recipes since as you said, things can be different w.r.t different corpus. @dophist I'll work on some of such suggestions above (and with pitch features I guess as well as other feature fusion methods) as soon as I get the data (we're close XD)

danpovey:
Guys, it's pretty clear to me that you have changed certain things just because you thought it would be better a different way, without actually doing any experiments.  For instance, those window functions really wouldn't make a difference; and if you had done experiments comparing filterbanks vs. MFCCs of the same dimension, I'm confident that either you wouldn't have found a difference or MFCCs would have been better (for compression reasons). 
And I'm not OK with changing these defaults for no good reason.  If you want to change anything vs. how the other recipes do it, I want you do demonstrate via experiments that it's better your way.  We always keep the settings consistent across recipes unless there is a compelling reason.  At the very least you need to tell me what specific experiments you did that made you think your ways of doing things were better.

danpovey:
... and I agree the results are nice, and it's a nice recipe.  But it would be very simple to just take out all of those changes (regarding the speed and volume perturbation and the feature configuration) and see how well it does.  I'm confident that using the normal defaults, it would be the same or better.  That's just one experiment.  And from that baseline, they could experiment with their changes one by one.  But I'm not going to check in all of these changes without careful experiments for each one.

egs/aishell2/s5/local/nnet3/tuning/run_lstm_1a.sh
```diff
@@ -0,0 +1,170 @@
...
+  [ -z $frames_per_chunk ] && frames_per_chunk=$chunk_width;
+  rm $dir/.error 2>/dev/null || true
+  for dset in dev test; do
+   (
+    steps/nnet3/decode.sh --nj $decode_nj --cmd "$decode_cmd"  --num-threads 4 \
+        --extra-left-context $extra_left_context \
+        --extra-right-context $extra_right_context \
+        --extra-left-context-initial 0 \
+        --extra-right-context-final 0 \
+        --online-ivector-dir exp/nnet3${nnet3_affix}/ivectors_${dset}_hires \
```
danpovey: i'm guessing this stage (or this script?) was not run, because it has the online-ivector-dir argument and you are not using ivectors.

src/online2bin/online2-mic-asr-demo.cc
```diff
@@ -0,0 +1,303 @@
+// online2bin/online2-mic-asr-demo.cc
+
+// Copyright 2014  Johns Hopkins University (author: Daniel Povey)
```
danpovey: This online demo is potentially an interesting project but I'l like to separate it from the rest of the aishell2 recipe.
I am a little cautious, in general, about these types of demos because they tend to be the cause of a lot of installation difficulties and also tend to be hard to maintain; and they potentially are a distraction from the core parts of Kaldi, attracting what I consider to be the wrong kind of user- i.e. clueless users doing small-scale projects.
But it's nice to have something that works on mac because a lot of people have macs and they are sufficiently homogeneous that installation difficulties are less likely than they would be on linux or windows.  So I'll definitely consider this-- as a separate pull request after the main recipe is merged.

danpovey:
What is the process to get the training data for this recipe?  You provide a gmail address and say that you ship a disk.  How much do you charge for that?  And are there limitations on data use and redistribution?  I just want to understand the situation.

underdogliu:
@danpovey I'm not from academic institute so I can't get the data access for free. Luckily my company is glad to support us to do some speech research and collaborations on open-sourced tools like kaldi using this new corpus but it takes time to go through the process and get the data.

egs/aishell2/README.md
```diff
@@ -0,0 +1,43 @@
+# AISHELL-2
+
+AISHELL-2 is by far the largest open-source speech corpus available for Mandarin ASR research.
```
danpovey: Please make this a .txt file to match other recipes.
And please don't use the term "open-source" since you are selling the dataset.  Perhaps "free*" would work.
*for research use.
Right now it's not absolutely clear from the sentence that the dataset is only free for non-commercial use... if it is even free for non-commercial use; that isn't absolutely clear to me.

dophist:
Hi Dan,

This week we finished some recipe modifications and experiments according to your suggestions, a brief update here:

#### NN feature
we change the NN feature from 80-dim fbank to high-resolution mfcc, and re-start the NN training from scratch (this change doesn't affect GMM, so we re-use GMM model and resources), results are as follows:

model | 80-fbank | 40-mfcc_hires |
-- | -- | --
tdnn | 11.13 | 11.2
chain tdnn | 9.90 | 10.10

details:
For 80-fbank
%WER 11.13 [ 5513 / 49532, 113 ins, 428 del, 4972 sub ] exp/nnet3/tdnn_sp/decode_test/cer_10_0.0
%WER 9.90 [ 4904 / 49532, 70 ins, 929 del, 3905 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

For 40-mfcc_hires
%WER 11.20 [ 5547 / 49532, 99 ins, 475 del, 4973 sub ] exp/nnet3/tdnn_sp/decode_test/cer_10_0.0
%WER 10.10 [ 5001 / 49532, 63 ins, 967 del, 3971 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

the only difference in the above experiments are NN feature and input layer dimension(80 vs 40), all other factors(training hyper params, etc) are exactly the same.  I don't think this result is enough to reach conclusion that one feature is better than another, the difference is marginal, and may vary when we make more tuning towards each feature setup. 

My experience is that when we get really large dataset, 80 fbank is slight better(0.3%-0.5%) than old-style mfcc/plp, but since I haven't actually compared 80-fbank with mfcc_hires on industrial dataset,  we would like to make aishell-2 consistent with Kaldi's standard setup, as this commit update.  And later we can try out ideas step by step based on this standard recipe, as Dan suggested. @underdogliu 

#### tempo and abs-vol perturbation
We remove tempo perturbation and abs-volume perturbation codes from the PR.

Here are some pitfalls that we'd like to share about these perturbations:

* In terms of the speed perturb over tempo: I came across situations occasionally that some training data are extremely mismatched from target application in terms of "speech rate" (such as slow crowd-sourced recording for spontaneous application).  During data simulation, and perturb factor may be more intense(say 1.1-1.3 instead of 0.9-1.1).  The "speed" perturbation produces non-human-like wavs when the perturb factor significantly deviates 1.0(say 1.3). "tempo" is a remedy in such cases. "speed" is superior to "tempo" in cases that training data is somehow limited to cover the variability of population vocal & pitch characteristics, and this is often not the case for really large industrial dataset.

* In terms of abs-volume perturbation, a pitfall I often come across is that careless setup of lower volume scale may produce lots of silence-like wavs for training. The resulting ASR system will present particular vulnerability towards fricative sound & silence modeling, especially in noisy environment.  Fricative phone mis-aligned with silence(noisy), may further degrade ASR modules that rely on duration (e.g. ASR confidence calculation or Wakeup-Word system)

Actually these perturbations are not particularly suitable for AISHELL-2, because AISHELL-2 is well-recording of clean read speech with typical volume range. That's the reason why we didn't actually use these perturbation in the recipe.  We only add those perturbation codes as a data-aug tool-chain enhancement.  We remove these modifications at the moment, to be more cautious about community code changes and recipe consistency. 

#### laptop streaming mic asr demo
I've stripped mic asr demo out from this PR, and later I will create a separate PR for the demo. 

#### misc
README renamed, and descriptions of the AISHELL-2 data release are refined in both readme and run.sh comments.

best,
Jiayu

danpovey:
Thanks for doing the experiment about 80-dim fbank fs. 40-dim MFCC.
Perhaps you could try 60-dim MFCC as well-- and 80 if you have time
(although that's up to you)?
I'm open to using higher-dim MFCCs.  But fbanks will interact badly with
our compression scheme.

I believe we have also found that when using a lot of data (like 1000
hours), the speed perturbation does not really help.

Regarding the volume perturbation: I'm surprised you'd get any significant
amount of silence-like wavs because the volume is chosen uniformly between
the low and high scales, and even if you set the low scale very close to
zero, you just won't get that many perturbed to a very small volume.

The volume perturbation is particularly important if the dataset you are
training on was automatically pre-normalized for volume-- it allows the
nnet to learn invariance to gain differences.  You'd never see an
improvement on a matched test set (that was normalized in the same way),
but it would allow for robustness to more natural data.

Can I ask why you took out the i-vectors?

It may be OK with me to leave out the i-vectors even though we do normally
use them.

Pitch is normally helpful for Chinese ASR.  I know it may create extra
latency in the online demo though.  I think @naxingyu said he would
experiment with that.

Dan

underdogliu:
@dophist Speaking of features, a 43-dim mfcc hires features with pitch would give quite good benefit to the result I believe - I did multiple experiments in hkust and aishell and the improvement was significant, at least to chain model. So perhaps we can use steps/make_mfcc_pitch_online.sh to generate the features (I do remember pitch features was used in aishell btw...).

danpovey:
I agree that pitch should help-- I wonder whether there was a specific
reason why they left it out, e.g. some interaction with their online demo.

I am open to the possibility that greater dim than 40 would help, although
I expect we'll see most or all of the improvement by going to 60.

dophist:
Hi Dan,

Sorry for the late update. 

#### 1. We've got results on the dimensions of hires-mfcc:

model | xent tdnn | chain tdnn 
-- | -- | -- 
80-fbank | 11.13 | 9.90 
40-mfcc_hires | 11.2 | 10.10
60-mfcc_hires | 11.2 | 10.03
80-mfcc_hires | 11.14 | 9.90

details:
For 80-fbank
%WER 11.13 [ 5513 / 49532, 113 ins, 428 del, 4972 sub ] exp/nnet3/tdnn_sp/decode_test/cer_10_0.0
%WER 9.90 [ 4904 / 49532, 70 ins, 929 del, 3905 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

For 40-mfcc_hires
%WER 11.20 [ 5547 / 49532, 99 ins, 475 del, 4973 sub ] exp/nnet3/tdnn_sp/decode_test/cer_10_0.0
%WER 10.10 [ 5001 / 49532, 63 ins, 967 del, 3971 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

For 60-mfcc_hires
%WER 11.20 [ 5550 / 49532, 89 ins, 506 del, 4955 sub ] exp/nnet3/tdnn_sp/decode_test/cer_10_0.5
%WER 10.03 [ 4969 / 49532, 70 ins, 939 del, 3960 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

For 80-mfcc_hires
%WER 11.14 [ 5520 / 49532, 98 ins, 562 del, 4860 sub ] exp/nnet3/tdnn_sp/decode_test/cer_11_0.0          
%WER 9.90 [ 4903 / 49532, 68 ins, 903 del, 3932 sub ] exp/chain/tdnn_1g_sp/decode_test/cer_7_0.0

#### 2. Release of Android and Mic dev/test set.
We've restructured the dev/test resources, and made corresponding recipe modifications, so that now AISHELL-2 data users should be able to download and play with dev/test sets from Android & Mic channels(the released training data is from iOS).  We believe these resources can help the community with acoustic transfer learning research.

#### 3. about pitch and iVector
We also believe pitch and iVector setup can help to achieve lower CER on AISHELL-2 corpus. @underdogliu is working on such recipe and experiments, and AFAIK he's getting clear improvements over our baseline.

The reason we didn't add these in our baseline recipe:
iVector: most academic corpus contains info about speakers(as AISHELL-2 does), but for industry data, it is not always feasible to collect speaker info for each utterances, due to privacy issue.  And assigning a dummy speaker id to each utterance, may help with CER, but it complicates the baseline system.  At least in this "baseline" recipe, we want to make minimal assumptions about the training data(it only requires wav.scp & trans.txt as general purpose ASR dataset)

For pitch, the computation introduces latency, we may add pitch into the baseline later.  But for now, we would rather leave pitch setup to @underdogliu 's improved recipe(as iVector), if Dan is OK with this.  @underdogliu can share more results on how much pitch and iVector are helping with Mandarin ASR, that's also a fruitful result for me : )

Dan, let me know when you want a more clean commit history and code style consistency check towards this PR, I'll do it

best,
Jiayu

danpovey:
OK, thanks for the results.  Those WER differences are actually small
enough that I normally wouldn't really trust them-- the differences could
easily be noise.  Can you rerun the chain TDNN training with a different
random seed (not that the random seed makes much difference) on the 3 MFCC
experiments to get another number for us to see?  You can re-use the egs if
you kept them, and if you want, you can run with slightly fewer epochs for
speed.

I am a little confused by your comment that "assigning a dummy speaker id
to each utterance, may help with CER, but it complicates the baseline
system.".  Kaldi requires you to do that-- data validation would fail if
there were no utt2spk file, and you'd normally have a dummy one-to-one
mapping. Perhaps what you mean is that the ivectors complicate the baseline
system.

Dan



model xent tdnn chain tdnn

dophist:
Sure, we'll rerun those 3 mfcc dimension experiments with a different seed. It may take around more than a week, I'll update as long as we get the results.

"Perhaps what you mean is that the ivectors complicate the baseline
system." Yes, that's what I mean, sorry to be confusing. 

best,
Jiayu

naxingyu:
@danpovey Does srand=123 works for you?

danpovey:
sure.

naxingyu:
@danpovey by setting srand=123 we got results below:

model | chain | chain srand=123
-- | -- | --
40-mfcc_hires | 10.10 | 9.89
60-mfcc_hires | 10.03 | 10.00
80-mfcc_hires | 9.90 | 9.99

danpovey:
Oh, OK.  So as I suspected, the differences were likely just noise.  Then
let's just use the default 40-dimensional MFCC features, they are kinder to
the disk.

naxingyu:
So we update related results and start another round of review?

egs/aishell2/s5/run.sh
```diff
@@ -0,0 +1,57 @@
...
+gmm_stage=1
+
+. ./cmd.sh
+. ./path.sh
+. ./utils/parse_options.sh
+
+# we should probably move jieba(for word segmentation) into 
+# Kaldi's "tools" dir with Dan's approval
+# before you run the entire recipe, run the following command
+# local/install_jieba.sh local/jieba
```
danpovey: can you please move the install_jieba.sh script into the tools directory and make it so that this script dies with an informative error message (telling you exactly what to run) if it's not installed?

egs/aishell2/s5/run.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+# GMM
+if [ $stage -le 2 ]; then
+  local/run_gmm.sh --nj $nj --stage $gmm_stage
+fi
+
+# xent
+if [ $stage -le 3 ]; then
+  local/nnet3/run_tdnn.sh --nj $nj
```
danpovey: let's leave this commented out by default; people can run it if they want.  Obviously make sure the 'stage' in the chain training is set to a low enough value that it will run without running the nnet3 script.

egs/aishell2/README.txt
```diff
@@ -0,0 +1,50 @@
...
+
+Depends on your location and internet speed, we distribute the corpus in two ways:
+* hard-disk delivery
+* cloud-disk downloading
+
+To apply for AISHELL-2 corpus for free, you need to fill in a very simple application form, confirming that:
+* university department / education institute info
+* only for non-commercial research / education use
+
+After comfirmation, AISHELL-foundation will covers all fees (including the corpus, hard-disk cost etc)
```
danpovey: confirmation

danpovey:
I quickly scanned through it and only saw small issues.  If you fix them and update the results, it should be OK to merge.

egs/aishell2/s5/run.sh
naxingyu: OK. I'll make it into env.sh as other tools do.

egs/aishell2/s5/run.sh
naxingyu: Sure.

egs/aishell2/README.txt
naxingyu: Confirmed. I'll update.

naxingyu:
@danpovey issues are fixed and results are updated.

danpovey:
OK, merging.  Thanks.

bambamochu:
Great work, I downlowaded the dev and eval and trained began the training, but one thing I noticed was that the the words in data/local/dictlexicon.txt were not segemented with Jieba its as if the pronunications were generated at the sentence level and not word level, however the text in data/dev/text data/test/text and data/train/text all used the local/word_segmentation.py correctly and were seperated

dophist:
bambamo, every entry in lexicon *IS* a word, a word doesn't need to be segmented. DaCiDian contains half a million Chinese words, some words are just a bit longer than you expected.

bambamochu:
Thank you dophist, i didn't know chinese had 10+ character words, just double check it please, thanks

PR #2401, opened 2018-05-02
remove other un-needed files

huangruizhe:
Remove un-needed files, which are not mentioned in any Makefile or not included in any other src files.
  
**First,** in kaldi/src/ directory and its subdirectories, the following shell script finds out all candidates which are *.cc or *.h files whose basic file names (without the extension) are not mentioned in any of the following:
- Makefile*
- *.mk
- kaldi/src/configure

[find-non-used.txt](https://github.com/kaldi-asr/kaldi/files/1968630/find-non-used.txt) (change extention to .sh and chmod+x before use)

Here are 6 files we found:
- src/chain/context-dep-topology.h
- src/cudamatrix/cu-kernels-ansi.h      ----x
- src/decoder/decodable-mapped.h
- src/decoder/decodable-sum.h
- src/util/basic-filebuf.h                        ----x
- src/util/kaldi-cygwin-io-inl.h               ----x
  
**Second,** we use the following script to double-check the candidates and see if they are referenced by others but not explicitly mentioned in Makefiles:
`grep -r base_name kaldi/src/`
The "false-positives" are those marked by "----x".
  
**At last,** we remove the remaining 3 files.


danpovey:
Thanks for finding these.
Please update  your PR to un-delete the two decodable headers.  They may be useful for something in future, and may possibly be used by people already for things like system combination.

huangruizhe:
The two decodable headers are un-deleted, while preserving their version history.

PR #2403, opened 2018-05-04
[egs] Updating VoxCeleb Recipe

david-ryan-snyder:
Minor updates to the VoxCeleb recipe:

1. **Using more data** 
* Including dev part of VoxCeleb1 in training (moved make_voxceleb1_test.pl to make_voxceleb1.pl, it prepares both dev and test)
* Reducing the minimum chunk-length from 500 to 400. There are a lot of short segments in VoxCeleb, so this allows us to retain more of them for training
* In the x-vector system, increasing --num-repeats from 35 to 50 to create more examples per speaker
2. **Updating the MFCC config**
* I've found that a different MFCC config works well for wideband speaker recognition in general, see config changes below (@entn-at, I believe the previous config was adapted from an ASR recipe)
3. **VoxCeleb1 trials are now downloaded from our mirror at http://www.openslr.org/49/**

The x-vector system improved by about 20%, from 4.173% to 3.224% EER. The i-vector system goes from 5.748% to 5.53% EER, so it's about the same after the update. 

FYI: @entn-at


egs/voxceleb/v2/run.sh
```diff
@@ -18,24 +18,27 @@ vaddir=`pwd`/mfcc
 
 
 voxceleb1_trials=data/voxceleb1_test/trials
```
danpovey: is data/voxceleb1_test/trials an input to the script?  do we need to make it downlaodable?
you might want to add a comment to e.g. 
```
# change this to where you have downloaded (or want the script to
# download) the VoxCeleb corpus).
```
(obviously edited for accuracy, I don't know if it auto-downloads).

entn-at: `local/make_voxceleb1.pl` creates `data/voxceleb1_test/trials` by converting the trials file provided by the authors of VoxCeleb to a format usable in Kaldi. It tries to download it from the official URL if it does not already exist in the VoxCeleb1 data distribution. Perhaps it would be possible to mirror that file on openslr.org?

danpovey: If @david-ryan-snyder has time to deal with it, mirroring that sounds like a good idea.

david-ryan-snyder: Sure, we can mirror it, and download the file from the mirror instead.

egs/voxceleb/README.txt
```diff
@@ -1,6 +1,13 @@
...
-http://www.robots.ox.ac.uk/~vgg/data/voxceleb2/ for additional details and
-information on how to obtain them.
+ 
+ This is a Kaldi recipe for speaker verification using the VoxCeleb1 and
+ VoxCeleb2 corpora.  See http://www.robots.ox.ac.uk/~vgg/data/voxceleb/ and 
+ http://www.robots.ox.ac.uk/~vgg/data/voxceleb2/ for additional details and
+ information on how to obtain them.
 
-Note: This recipe requires ffmpeg to be installed and its location included in $PATH
+ Note: This recipe requires ffmpeg to be installed and its location included
```
danpovey: It would be nice if run.sh or a script in local/ would check for this, if not done already.

david-ryan-snyder: Good idea--I'll add that.

david-ryan-snyder: Done.

david-ryan-snyder:
@danpovey, I added a check to see if ffmpeg exists in egs/voxceleb/v1/local/make_voxceleb2.pl.

david-ryan-snyder:
> I'd like to add a simple BNF i-vector recipe as v3.

That would be great... If you do this, could you make it in nnet3?

entn-at:
The setup I'm currently working on adding is based on an nnet3 TDNN model trained on TEDLIUM with a 60-dim BN layer. The TDNN does not use i-vectors as input (trained with `--norm-means=true`). Custom versions of `train_ivector_extractor.sh` and `extract_ivectors.sh` use BNF for Gaussian selection/posterior calculation and (optionally) add BNFs on top of MFCCs (+D+DD) to compute stats. 

It's still running, I'll create a PR once I have results to confirm that it works as intended.

PR #2412, opened 2018-05-10
lattice-copy: check sorted and segfault fix

### LINKED ISSUE #687: lattice-copy bugs
I saw the message
LOG (lattice-copy:CopySubsetLattices():lattice-copy.cc:104) The utterance 20000907_0930_1030_rfi_fm_dga-0093333-0093857 is larger than the last key in the include list. Not reading further.
from the command:
gunzip -c decode_nosp_dev_tgpr/lat.2.gz| lattice-copy --include=foo ark:- ark:- 
This is not right in general, as unless the ",s" option is included in the rspecifier, we cannot assume the lattices are sorted [they are typically sorted in practice, per archive, but the framework is supposed to be general].

Also I see the code includes the line
  if (include && lattice_reader->Key() > *(subset_list.rbegin())) {
with no check for whether subset_list is empty.. this could segfault.
Making an issue rather than fixing it myself, so someone else can help.

jfainberg:
Fixes #687 . Check for s / cs options in rspecifier and avoid segfault if include list is empty.

### DIFF
--- a/src/latbin/lattice-copy.cc
+++ b/src/latbin/lattice-copy.cc
@@ -178,7 +178,7 @@ int main(int argc, char *argv[]) {
 
     RspecifierOptions opts;
     ClassifyRspecifier(lats_rspecifier, NULL, &opts);
-    bool sorted = opts.sorted || opts.called_sorted;
+    bool sorted = opts.sorted;
 
     int32 n_done = 0;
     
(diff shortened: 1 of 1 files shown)


src/latbin/lattice-copy.cc
```diff
@@ -174,6 +176,10 @@ int main(int argc, char *argv[]) {
     std::string lats_rspecifier = po.GetArg(1),
         lats_wspecifier = po.GetArg(2);
 
+    RspecifierOptions opts;
+    ClassifyRspecifier(lats_rspecifier, NULL, &opts);
+    bool sorted = opts.sorted || opts.called_sorted;
```
danpovey: just the 'sorted' option is relevant here. (called_sorted is an assertion that the program will access them in sorted order, which is only relevant for random-access readers)>

jfainberg: Makes sense, thanks! Fixed.

Issue #2388, opened 2018-04-26
Create utils/data/get_reco2dur.sh


david-ryan-snyder:
> That seems to be a deficiency in the scripts.  We should have a script-level mechanism to create the reco2dur file, e.g. a script
> utils/data/get_reco2dur.sh
> that is modeled after
> utils/data/get_utt2dur.sh
> (although it should obtain the recording lengths by copying the utterance lengths if it notices that the utterance-ids and the recording-ids are the same with a one to one relationship and the utt2dur file is present),
> and the script
> utils/data/perturb_data_dir_speed_3way.sh 
> and related scripts should copy and modify the reco2dur files the same way they do with the utt2dur file, if present.
> In addition, validate_data_dir.sh and copy_data_dir.sh should be extended to properly deal with the reco2dur file.

See Kaldi-Help: https://groups.google.com/d/msg/kaldi-help/px-rPmGIK7A/eGbjC18OAgAJ

jtrmal:
@danpovey I think this can be closed

johnjosephmorgan:
The script utils/data/get_reco2dur.sh seems to be causing a problem. The file data/train/reco2dur is not getting written. I see a potential problem on lines 53 and 54 of utils/data/get_reco2dur.sh.

david-ryan-snyder:
@jtrmal, @danpovey, I think this issue got resolved in https://github.com/kaldi-asr/kaldi/pull/2544

PR #2414, opened 2018-05-11
Add util to compute recordings durations before data perturbation

Ore-an:
As per #2388. I'd like some help to test more extensively, especially in the case where recording-ids are the same as utt-ids, as a bug could have a big impact

### DIFF
--- a/egs/wsj/s5/utils/data/get_reco2dur.sh
+++ b/egs/wsj/s5/utils/data/get_reco2dur.sh
@@ -1,0 +1,143 @@
+#!/bin/bash
+
+# Copyright 2016  Johns Hopkins University (author: Daniel Povey)
+#           2018  Andrea Carmantini
+# Apache 2.0
+
+# This script operates on a data directory, such as in data/train/, and adds the
+# reco2dur file if it does not already exist.  The file 'reco2dur' maps from
+# recording to the duration of the recording in seconds.  This script works it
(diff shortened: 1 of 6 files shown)


egs/wsj/s5/utils/data/get_reco2dur.sh
```diff
@@ -0,0 +1,143 @@
...
+
+
+if [ -s $data/reco2dur ] && \
+  [ $(cat $data/wav.scp | wc -l) -eq $(cat $data/reco2dur | wc -l) ]; then
+  echo "$0: $data/reco2dur already exists with the expected length.  We won't recompute it."
+  exit 0;
+fi
+
+if [ -s $data/utt2dur ] && \
+   [ $(cat $data/utt2spk | wc -l) -eq $(cat $data/utt2dur | wc -l) ] && \
```
danpovey: you could do
`wc -l <$data/utt2spk` to avoid the 'cat'.

danpovey:
Thanks!  I commented on the original thread- I hope the guy who reported the problem can help test it.

danpovey:
Thanks.  Have you at least tested that this runs?  I may merge without waiting for that guy to test.

Ore-an:
Yes, I've tested it on a dataset with a segments file and everything works as it should; durations are as reported by soxi, validate_data_dir outputs if there's missing lines in the file and fix_data_dir filters out what's not in wav.scp.

PR #2420, opened 2018-05-14
[egs] BNF i-vector based speaker verification recipe for VoxCeleb

entn-at:
This is a recipe demonstrating a bottleneck feature (BNF) based i-vector PLDA system. 
A nnet3 TDNN model is trained on TEDLIUM with a 60-dim BN layer. Custom versions of `train_ivector_extractor.sh` and `extract_ivectors.sh` use BNF for Gaussian selection/posterior calculation and (optionally) add BNFs on top of MFCCs (+D+DD) to compute stats.

This is currently work in progress. I haven't yet confirmed (for this particular setup) that the system achieves reasonable performance. Due to problems with my SGE setup, i-vector extractor training progress has been very slow and it will take some more time until I have results. I'm opening this PR to invite general comments and suggestions for changes.

FYI: @david-ryan-snyder


egs/voxceleb/v3/run.sh
```diff
@@ -0,0 +1,232 @@
...
+if [ $stage -le 14 ]; then
+  eer=`compute-eer <(local/prepare_for_eer.py $voxceleb1_trials exp/scores_onlymfcc_voxceleb1_test) 2> /dev/null`
+  mindcf1=`sid/compute_min_dcf.py --p-target 0.01 exp/scores_onlymfcc_voxceleb1_test $voxceleb1_trials 2> /dev/null`
+  mindcf2=`sid/compute_min_dcf.py --p-target 0.001 exp/scores_onlymfcc_voxceleb1_test $voxceleb1_trials 2> /dev/null`
+  echo "only MFCC EER: ${eer}%"
+  echo "minDCF(p-target=0.01): $mindcf1"
+  echo "minDCF(p-target=0.001): $mindcf2"
+  # only MFCC EER: 4.608%
+  # minDCF(p-target=0.01): 0.4370
+  # minDCF(p-target=0.001): 0.6094
```
david-ryan-snyder: @entn-at, what is this second system? The only called "onlymfcc?"

david-ryan-snyder:
FYI @leibny

egs/voxceleb/v3/run.sh
david-ryan-snyder: OK, I get it. So you use BNFs + MFCCs to get the posteriors from the GMM, and just the MFCCs (in conjunction with the posteriors) to get the i-vector stats.

entn-at: I only used BNFs to compute posteriors, but used MFCC+BNF (`paste-feats`) to get i-vector stats. Perhaps I could explore other variants as well.

entn-at:
The recipe demonstrates two BNF variants, one using BNF for calculating posterior probabilities and using MFCC+BNF as speaker ID features (EER: 4.205%, minDCF(p-target=0.01): 0.4509, minDCF(p-target=0.001): 0.6528), and another using BNF for calculating posterior probabilities but only MFCCs as speaker ID features (EER: 4.608%, minDCF(p-target=0.01): 0.4370, minDCF(p-target=0.001): 0.6094).

leibny:
@entn-at  if I am understanding it correctly, you are using BNF to obtain UBM posteriors, and then you get stats using MFCC+BNF (BNF-UBM posteriors NOT longer used for stats).

The pipeline that usually works is:
* Train 2 UBMs one for BNF and another one for BNF+MFCC, though you can just use only MFCC. * For both you use the posteriors given the BNF-UBM to compute stats for your systems.
* Concatenate your new features at i-vector level (MFCC_i-vectors+BNF_i-vector). Usually, fusion at any level improves results.

Have you tried this approach? 

To prevent your system to go slow, you can reduce the delta (i.e., --delta=1)

entn-at:
I used the following pipeline for this recipe:

1. Train a UBM on BNF. 
2. Initialize another UBM with BNF+MFCC (or just MFCC) using posteriors obtained from the previously trained BNF-UBM (using `fgmm-global-acc-stats-post` and `fgmm-global-init-from-accs`).
3. Initialize i-vector extractor with BNF+MFCC (or just MFCC) UBM.
4. Use BNF-UBM posteriors with BNF+MFCC (or just MFCC) frame-level features to train i-vector extractor and later to extract i-vectors

Whenever posteriors are used, they are computed from the BNF-only UBM.

Perhaps I could add i-vector level fusion of the i-vectors described above with BNF-only i-vectors.

entn-at:
I'm currently running a BNF-only i-vector system (BNF-UBM for posteriors, BNF+1delta for stats) in order to try i-vector level fusion. I added a version of `paste-feats` that works with `Vector<BaseFloat>`/`BaseFloatVectorReader`/`BaseFloatVectorWriter` (I put it in `ivectorbin/paste-ivectors.cc`).

entn-at:
I'd prefer to add a SITW BNF-ivector recipe later, after the initial SITW recipe (https://github.com/kaldi-asr/kaldi/pull/2422) has been merged. This BNF-ivector recipe generally "works", i.e., it achieves results that are better than those of standard MFCC i-vector systems and that, depending on the data used for testing, are similar or a bit worse than TDNN posterior (sre10/v2) type systems and x-vector systems. 

I should add that the recipe hasn't been tuned much on this data; perhaps a modified Tedlium TDNN config with wider BN layer gives better results. Unfortunately, I don't have the resources to do exhaustive tuning of the recipe, but it provides a good starting point for others to improve it.

danpovey:
I'm happy to merge something simple that works.

leibny:
It is a little bit counterintuitive to have an MFCC only system with better results for low FA than the one that contains MFCC+BNF. 
Do you have the results for the BNF only system? I am wondering if the results for this system are not good enough in the low FA region.

entn-at:
I agree that it is counterintuitive; however, I'd like to point out that the VoxCeleb1 test set only has 37720 trials (18860 target, 18860 nontarget), which may result in too few errors at the MinDCF10 operating point to give a good estimate. NIST SRE 2010 extended core condition 5 has 416119 trials (7169 target, 408950 nontarget), and a similar BNF recipe achieves substantially lower MinDCF10 values than the standard MFCC i-vector recipe (though somewhat higher values than the TDNN posterior recipe). MinDCF10 may not be meaningful for the VoxCeleb1 evaluation set.

entn-at:
Below are the results for the different BNF system variants that I tested. All variants use BNF any time posteriors are computed (UBM training, i-vector extractor training, i-vector extraction). "MFCC" / "BNF" / "MFCCBNF" denotes the features (or combination) used for computing i-vector stats (while using BNF UBM based posteriors). i-vector level fusion is done by concatenating two 600-dim i-vectors and using LDA to reduce the dimension to 400.

System                 | EER | minDCF(p-target=0.01) | minDCF(p-target=0.001) | Notes
---------------|------|---------------------------|-----------------------|-------------
i-vector MFCCBNF | 4.205% | 0.4509 | 0.6528 | 24MFCC_D_A + 60BNF = 132 feat-dim for stats
i-vector MFCC | 4.608% | 0.4370 | 0.6094 | 24MFCC_D_A = 72 feat-dim for stats
i-vector BNF | 12.32% | 0.8766 | 0.9323 | 60BNF_D = 120 feat-dim for stats
i-vector level fusion MFCCBNF + MFCC | 3.918% | 0.3945 | 0.6288 |
i-vector level fusion MFCCBNF + BNF | 3.929% | 0.4257 | 0.5960 |
i-vector level fusion MFCC + BNF | 3.887% | 0.4307 | 0.6205 |
../v1 | 5.419% | 0.4701 | 0.5981 |
../v2 | 3.224% | 0.3492 | 0.5452 |

Let me know if you have any further suggestions. If you prefer not to merge this, I can close the PR and keep the branch up for anyone who's interested.

leibny:
The results are quite interesting and make more sense now. It seems that BNF system is not very good (as we already discussed), but gives some improvement in EER when combined with MFCCs systems at several stages.  Is there a way to have a better BNF system? (Using other database for training, tuning the BNF a bit?)

entn-at:
The approach taken here in this recipe is somewhat similar to the hybrid alignment framework in [1]. All the tested systems above use *only* BNF for UBM posterior computation ("alignments") while using different features for stats computation. That seems to provide improvements over the standard MFCC GMM-UBM i-vector approach. At the same time, using these BNFs as the only speakerID-features (to compute i-vector stats) doesn't work too well, which shouldn't come as a surprise, as they are computed from the 6th TDNN layer and presumably retain less speaker-dependent information.

I realize this "hybrid" approach is somewhat non-standard. If you prefer, I can close this PR and later submit a more standard BNF-MFCC recipe that does not make a distinction between alignment features and stats features.

[1] McLaren, M., Castan, D., Ferrer, L., & Lawson, A. (2016). On the Issue of Calibration in DNN-Based Speaker Recognition Systems. In INTERSPEECH (pp. 1825-1829). [link](https://pdfs.semanticscholar.org/7f30/55cf2b7b7ba20b82e171f3e751828f364601.pdf)

david-ryan-snyder:
@ewald,

Thank you for working on this. We definitely want to add something along these lines.

With the current pull request, my concern is that the performance doesn’t compensate for the added complexity. To justify building an ASR system to obtain BNFs, I think the i-vector system that incorporates the BNFs should be clearly better than the traditional i-vector system at reasonable operating points. Also, the x-vector system is still much better, despite being simpler in some respects (it doesn’t need transcribed training data, and consists of fewer models), so it’s not clear what the advantage the v3 system provides. 

If you have time to experiment with this more, I have a few suggestions:

+ Can you perform future experiments on either SITW or SRE10? Performance on these datasets is better understood, so it's easier to get a sense of how good the performance should be. For example, if you made an SRE10 v3 recipe, you should be able to match the performance of the v2 recipe. As you pointed out before, the VoxCeleb 1 test set is small, and there may be some unusual characteristics in the test set that we’re not aware of yet.

+ I think it would be better if the features for the initial recipe are just MFCCs (and deltas, optionally) appended to BNFs, for both UBM and i-vector extractor. I think it's more standard, and will probably perform better anyways. When we were doing this with nnet2 BNFs, this is what my scripts looked like:

https://github.com/david-ryan-snyder/kaldi/blob/for-ewald/egs/sre08/v1/sid/train_diag_ubm_bnf.sh
https://github.com/david-ryan-snyder/kaldi/blob/for-ewald/egs/sre08/v1/sid/train_full_ubm_bnf.sh
https://github.com/david-ryan-snyder/kaldi/blob/for-ewald/egs/sre08/v1/sid/extract_ivectors_bnf.sh
https://github.com/david-ryan-snyder/kaldi/blob/for-ewald/egs/sre08/v1/sid/train_ivector_extractor_bnf.sh

Normally I apply the CMVN to the BNFs as well as the the MFCCs + deltas. You don't apply any feature normalization to your BNFs. Have you checked to see if it makes a difference?

Best,
David

entn-at:
Thanks, David! 
I tested different variants in the past (including more standard setups). Results varied a lot depending on the datasets used. I didn't apply CMVN to BNFs for UBM posterior computation in this recipe, as it actually resulted in better performance (that may not necessarily be true when BNFs are used together with MFCCs (+deltas) as speakerID feature for i-vector stats computation). 

I'm going to close this PR and plan to open another one later with a more standard setup on SRE10.

Best,
Ewald

PR #2422, opened 2018-05-15
[egs] Add recipes for Speakers in the Wild (SITW)

david-ryan-snyder:
The PR adds an i-vector and an x-vector recipe for Speakers in the Wild (SITW) (http://www.speech.sri.com/projects/sitw/). The results using x-vectors are currently state-of-the-art, as far as I know.

The recipe is trained on VoxCeleb1 and VoxCeleb2 (http://www.robots.ox.ac.uk/~vgg/data/voxceleb/). Sixty speakers in VoxCeleb1 overlap with SITW. We remove those from VoxCeleb1 prior to training. 

FYI @entn-at, @danpovey, @leibny


david-ryan-snyder:
@danpovey, when you get a chance, I think this is OK to merge. The v1 (i-vector) and v2 (x-vector) recipes are similar to the existing speaker recognition recipes.

PR #2431, opened 2018-05-19
[scripts] Adding script to compute minDCF for speaker recognition 

david-ryan-snyder:
This pull request adds a script called sid/compute_min_dcf.py, which computes the minimum of the detection cost function.

The PR also updates the voxceleb recipe, to include minDCF results along with equal error-rate. 

This performance metric is a very common in speaker recognition literature. It's usually used to see performance at operating points to the "left" of the EER. See Section 3 in https://www.nist.gov/sites/default/files/documents/2016/10/07/sre16_eval_plan_v1.3.pdf for details. 

I compared the output of this script with the NIST scoring tools. The results are basically identical. 

FYI @leibny, @entn-at


david-ryan-snyder:
@danpovey, I think this PR can be merged. It should be a relatively uncontroversial addition.

PR #2432, opened 2018-05-19
Add a warning message to nnet3/align.sh about using GPUs.

hhadian:
Do you want me to highlight it more? Or maybe add a warning message in the `if $use_gpu; then` block?


danpovey:
that's good. merging.

PR #2433, opened 2018-05-20
add word_ins_penalty option to src/latbin/lattice-1best.cc; add word_…

ronggong:
add word_ins_penalty option to src/latbin/lattice-1best.cc; 
add word_ins_penalty as a grid search parameter to egs/wsj/s5/steps/get_ctm.sh

### DIFF
--- a/egs/wsj/s5/steps/get_ctm.sh
+++ b/egs/wsj/s5/steps/get_ctm.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 # Copyright Johns Hopkins University (Author: Daniel Povey) 2012.  Apache 2.0.
+#           Music Technology Group, Universitat Pompeu Fabra (Rong Gong) 2018. Apache 2.0.
+
 
 # This script produces CTM files from a decoding directory that has lattices
-# present.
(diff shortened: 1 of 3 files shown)


danpovey:
Thanks, but I now realize that I overlooked something when I asked you to do this.

I thought that get_ctm.sh only got one ctm; I see now that it gets multiple ctm's.  Adding the WIP grid search will make it slower and will change the directory names which may break other scripts.
Can you please edit the comment at the top of get_ctm.sh to the following:
```
# This script produces CTM files from a decoding directory that has lattices                                                                         
# present.  It does this for a range of language model weights; see also 
# get_ctm_fast.sh which does it for just one LM weight and also supports
# the word insertion penalty.
```
and edit its `see also` line to:
```
  echo "See also: steps/get_train_ctm.sh, steps/get_ctm_fast.sh"
```
...and revert your other changes to get_ctm.sh.  Instead make the change to get_ctm_fast.sh.

psmit:
Just out of interest, why the choice of adding the word-insertion-penalty to lattice-1best? From what I can see isn't the "normal" kaldi way to add `lattice-add-penalty` in the scripts? Also lattice-best-path does not have that flag.

PR #2434, opened 2018-05-20
BPE version of MADCAT

babakrkb:
This is the byte-pair encoding version of MADCAT ARABIC OCR. 
I'm not really sure that `local/reverse.py` is the best way to reverse the latin and digits sequences. But I test it on some example and it works fine.


babakrkb:
Oh, I'll handle these conflicts and push again.

babakrkb:
About some tuning issues:
1-We can change the number in the `local/learn_bpe.py -s 700`. I am not sure that this number is the optimal one (because I am not very familiar with the BPE algorithm). If we decrease this number, words would be broken into more small pieces.
2-Maybe we can build a better lexicon that consider a right position for each character. In this setup position dependent phones was disable.

egs/madcat_ar/v1/local/apply_bpe.py
```diff
@@ -0,0 +1,339 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Author: Rico Sennrich
```
danpovey: this code, from their repo, seems to be released under the MIT license.  You can add
`# Released under the MIT License.`

egs/madcat_ar/v1/local/reverse.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import re, os, sys, io
+
```
danpovey: you should document carefully what this does via a comment.

egs/madcat_ar/v1/path.sh
```diff
@@ -1,4 +1,5 @@
 export KALDI_ROOT=`pwd`/../../..
+export LD_LIBRARY_PATH=/home/dpovey/libs:$LD_LIBRARY_PATH
```
danpovey: this should be in your .bashrc, not here.

egs/madcat_ar/v1/local/prepare_lexicon.py
```diff
@@ -19,7 +19,7 @@
         line_vect = line.strip().split(' ')
         for i in range(1, len(line_vect)):
             characters = list(line_vect[i])
-            characters = " ".join(characters)
+            characters = " ".join([ 'SIL' if char == '|' else char for char in characters])
```
danpovey: please add a comment here to explain what is going on. 
BTW, does the | occur at the begining or end of these things?

egs/madcat_ar/v1/local/apply_bpe.py
```diff
@@ -0,0 +1,339 @@
+#!/usr/bin/env python
```
danpovey: Let's consider putting these BPE related scripts in steps/bpe/.  @hhadian, do you have any thoughts on the location?

egs/madcat_ar/v1/run_end2end.sh
```diff
@@ -74,8 +74,15 @@ fi
...
+  cut -d' ' -f2- data/train/text | python3 local/reverse.py | python3 local/prepend_words.py | python3 local/learn_bpe.py -s 700 > data/train/bpe.out
+  for set in test train dev ; do
+    cut -d' ' -f1 data/$set/text > data/$set/ids
+    cut -d' ' -f2- data/$set/text | python3 local/reverse.py | python3 local/prepend_words.py | python3 local/apply_bpe.py -c data/train/bpe.out | sed 's/@@//g' > data/$set/bpe_text
+    mv data/$set/text data/$set/text.old
+    paste -d' ' data/$set/ids data/$set/bpe_text > data/$set/text
+  done
   local/prepare_dict.sh
-  utils/prepare_lang.sh --num-sil-states 4 --num-nonsil-states 8 --sil-prob 0.9999 \
+  utils/prepare_lang.sh --num-sil-states 4 --num-nonsil-states 8 --sil-prob 0.0 --position-dependent-phones false \
```
danpovey: you might want to comment that because this recipe uses BPE encoding, the silences are part of the words' pronunciations.

This recipe is not 100% correct in terms of how the word-final silence of the sentence is treated.  Basically, our model won't allow it as an option.  It can probably be handled by the context dependency but it's not ideal (e.g. it will mess up the alignments, if we cared about that) and I think it makes sense to fix it.

Please try the following fix.   Add a script in a shared location (e.g.  steps/bpe/ if that's what we want to do)

steps/bpe/add_final_optional_silence.sh


which will have usage:
```
if [ $# -ne 1 ]; then
  echo "Usage: $0  <lang> "
  echo " Add final optional silence to lexicon FSTs (L.fst an L_disambig.fst) in  "
  echo " lang/ directory <lang>."
fi
```
It can use "fstconcat" for this, concatenating with an FST like this:
```
0  1   $opt_sil_phone 0  -0.69
0 -0.69
1  0.0
```
The -0.69 is log(0.5) so it has equal probability for silence and not-silence.  This is all based on the assumption that your BPE puts the | at the beginning of words and the silence is at the beginning of words.   However, you should probably make the silence prob configurable via the --sil-prob option like prepare_lang.sh.  Be sure to treat the special case sil_prob=1.0 specially (disallow sil_prob=0.0, in that we just shouldn't call the script).

We can consider making the final silence non-optional, i.e. prob=1.0, since the initial silence is effectively non-optional, given how the BPE works.  You can test that, and also the topologies e.g. the num-states in silence.

egs/wsj/s5/steps/scoring/score_kaldi_wer.sh
```diff
@@ -18,7 +18,7 @@ max_lmwt=17
 iter=final
 #end configuration section.
 
-echo "$0 $@"  # Print the command line for logging
+echo "EDITED $0 $@"  # Print the command line for logging
```
danpovey: this may be not for commi

egs/madcat_ar/v1/local/prepare_lexicon.py
babakrkb: It occurs at the beginning of the words.

egs/wsj/s5/steps/scoring/score_kaldi_wer.sh
babakrkb: I think I should put all the ` sed 's/ //g' | sed 's/|/ /g' ` in the `local/wer_output_filter`. So I'll revert every changes in the score files.

egs/madcat_ar/v1/run_end2end.sh
babakrkb: I will test all of these strategies, but I think I should get some help from @hhadian to do them right. Because I don't have much experience in working with FSTs. 

Should I test BPE on IAM with this kind of recipe or wait to fix the final-word space issue?

egs/madcat_ar/v1/local/apply_bpe.py
hhadian: Maybe `utils/lang/bpe` because it's related to dict/lang preparation?

egs/madcat_ar/v1/run_end2end.sh
hhadian: @danpovey, what would you think about adding a `SIL` at the end of the transcriptions (i.e. in `text`) given the fact that we always have padding for OCR recipes (so we always have a final silence)?

danpovey:
I'm OK with putting those things in utils/lang/bpe/.
How would adding a SIL at the end interact with language modeling?
I'm concerned that for it to work in test time, you'd have to have the
language model allow SIL at the end, which would complicate LM preparation.

egs/wsj/s5/steps/scoring/score_kaldi_cer.sh
```diff
@@ -58,7 +58,7 @@ ref_filtering_cmd="cat"
 [ -x local/wer_output_filter ] && ref_filtering_cmd="local/wer_output_filter"
 [ -x local/wer_ref_filter ] && ref_filtering_cmd="local/wer_ref_filter"
 hyp_filtering_cmd="cat"
-[ -x local/wer_output_filter ] && hyp_filtering_cmd="local/wer_output_filter"
+[ -x local/wer_output_filter ] && hyp_filtering_cmd="local/wer_output_filter | sed 's/ //g' | sed 's/|/ /g' "
```
danpovey: @babakrkb, you shouldn't make changes to this file, it's in the shared part of the scripts. 
I think you already said you were going to fix this, though.

egs/wsj/s5/utils/lang/bpe/add_final_optional_silence.sh
```diff
@@ -0,0 +1,53 @@
...
+
+if [ $sil_ge_zero -eq 1 ] && [ $sil_le_one -eq 1 ]
+then
+	if [ $sil_zero -eq 1 ]
+	then
+		echo "final-sil-prob = 0 => Final silence was not added."
+	elif [ $sil_one -eq 1 ]
+	then
+		log=$(echo "l($final_sil_prob)" | bc -l)
+		mv $langdir/L.fst $langdir/preL.fst
```
hhadian: I guess it's better to use `.orig` instead, e.g. L.fst.orig

egs/wsj/s5/utils/lang/bpe/add_final_optional_silence.sh
```diff
@@ -0,0 +1,53 @@
...
+		mv $langdir/L_disambig.fst $langdir/preL_disambig.fst
+		echo -e "0\t1\t1\t0\n1" > $langdir/final_sil.fst.txt
```
hhadian: Better to write directly to final_sil.fst, e.g. echo -e "0\t1\t1\t0\n1" | fstcompile > $langdir/final_sil.fst

egs/wsj/s5/utils/lang/bpe/add_final_optional_silence.sh
```diff
@@ -0,0 +1,53 @@
...
+sil_le_one=$(echo "$final_sil_prob <= 1.0" | bc -l)
+
+if [ $sil_ge_zero -eq 1 ] && [ $sil_le_one -eq 1 ]
+then
+	if [ $sil_zero -eq 1 ]
+	then
+		echo "final-sil-prob = 0 => Final silence was not added."
+	elif [ $sil_one -eq 1 ]
+	then
+		log=$(echo "l($final_sil_prob)" | bc -l)
```
hhadian: I guess `bc` is not recommended. You could use perl -e.

egs/wsj/s5/utils/lang/bpe/add_final_optional_silence.sh
```diff
@@ -0,0 +1,53 @@
...
+langdir=$1
+
+sil_zero=$(echo "$final_sil_prob == 0.0" | bc -l)
+sil_one=$(echo "$final_sil_prob == 1.0" | bc -l)
+sil_ge_zero=$(echo "$final_sil_prob >= 0.0" | bc -l)
+sil_le_one=$(echo "$final_sil_prob <= 1.0" | bc -l)
+
+if [ $sil_ge_zero -eq 1 ] && [ $sil_le_one -eq 1 ]
+then
+	if [ $sil_zero -eq 1 ]
```
hhadian: 1. please use spaces instead of tabs.
2. use 2 spaces not 4.

egs/wsj/s5/utils/lang/bpe/add_final_optional_silence.sh
```diff
@@ -0,0 +1,53 @@
...
+		log=$(echo "l($final_sil_prob)" | bc -l)
+		mv $langdir/L.fst $langdir/preL.fst
+		mv $langdir/L_disambig.fst $langdir/preL_disambig.fst
+		echo -e "0\t1\t1\t0\t$log\n0\t$log\n1\t0.0" > $langdir/final_sil.fst.txt
+		fstcompile $langdir/final_sil.fst.txt > $langdir/final_sil.fst
+		fstconcat $langdir/preL.fst $langdir/final_sil.fst | fstarcsort --sort_type=olabel > $langdir/L.fst
+		fstconcat $langdir/preL_disambig.fst $langdir/final_sil.fst | fstarcsort --sort_type=olabel > $langdir/L_disambig.fst
+	fi
+else
+	echo "final-sil-prob should be between 0.0 and 1.0. Final silence was not added."
```
hhadian: better to always use "$0" in echo.

egs/wsj/s5/utils/lang/bpe/add_final_optional_silence.sh
```diff
@@ -0,0 +1,50 @@
...
+final_sil_prob=0.5
+
+echo "$0 $@"  # Print the command line for logging
+
+. utils/parse_options.sh
+
+if [ $# -ne 1 ]; then
+  echo "Usage: $0  <lang>"
+  echo " Add final optional silence to lexicon FSTs (L.fst and L_disambig.fst) in  "
+  echo " lang/ directory <lang>."
```
danpovey: Please explain why this might bre useful, e.g.:
```This can be useful in systems with byte-pair encoded (BPE) lexicons, in which
 the word-initial silence is part of the lexicon, so we turn off the standard optional
 silence in the lexicon```

**Comment by danpovey** [comment]
Before I merge this, there's just something I want to clarify.
Is there concept of an OOV when you do BPE encoding?
I mean-- is it possible that when trying to apply a BPE encoding to some previously unseen data, there would be words that couldn't be covered by the encoding; and what happens to them?

**Comment by hhadian** [comment]
The lexicon generated by BPE on MADCAT includes all the singleton characters (Babak, correct me if I'm wrong), so I guess there won't be any OOVs. However, this might be different for other databases.

**Comment by babakrkb** [comment]
I think Hossein is right. I was checking some unseen words in test set, and I have seen that if the algorithm can not cover a word, or part of a word by the `bpe.out` file, it breaks the word to its characters. But if you want a more precise answer please give me a little time to test more examples.

**Comment by danpovey** [comment]
See if you can find what happens when there are previously unseen
characters.
I just want to make sure it doesn't just die if it sees unexpected things.

**Comment by babakrkb** [comment]
Ok, I've got it. This algorithm will never die in any circumstances. It works like this:
First, when we run `learn_bpe.py` on the train text, it generates a file named `bpe.out` which contains a dictionary that shows the mapping of the characters like this:
```
b a -> ba
ba k -> bak
gr een -> green
...
```
Then, when we want to apply BPE  on some text file according to the `bpe.out` , it first break all the text into characters, and starts to replacing pair of characters according to `bpe.out` until it can't find any joinable pair of characters.  So wherever it sees the `b` followed by `a`, it replace these two character with `ba` and so on. So we won't have any problem if the text has unseen words or even unseen characters.

PR #2436, opened 2018-05-21
Backward rnnlm

hainan-xv:
lattice rescoring with backward RNNLMs


egs/swbd/s5c/local/rnnlm/tuning/run_tdnn_lstm_back_1e.sh
```diff
@@ -0,0 +1,143 @@
...
+    exit 1
+  fi
+
+  for decode_set in eval2000; do
+    decode_dir=${ac_model_dir}/decode_${decode_set}_${LM}_looped
+
+    # Lattice rescoring
+    rnnlm/lmrescore_back.sh \
+      --cmd "$decode_cmd --mem 4G" \
+      --weight 0.5 --max-ngram-order $ngram_order \
```
danpovey: If you are using a weight of 0.5 for both the forward and backward passes, I'd be surprised if this weight was optimal.  Because effectively the weight on the n-gram LM would be zero.

hainan-xv: I tuned this a bit and it seems 0.5 is the best weight.

I found another small in issue in the code which I should fix before merging this PR. Do you think we should just leave 0.5 as it is now?

danpovey: Did you try using 0.4 for both forward and backward?  It's very surprising if it doesn't want any weight at all on the n-gram LM-- that always helps, as they are so complementary.

hainan-xv: Yes. 0.4/0.4/0.2 is also worse than 0.5/0.5/0

Issue #2438, opened 2018-05-21
Breaks on wjs/utils/fix_data_dir.sh


Fhrozen:
Hello There:

I am being trying to execute the the feature extraction recipe in ESPnet, but I found that every time that the code executes fix_data_dir.sh it just stopped the main routine. I found that this stops at the line 100, int he filter_recordings function, just before finishes. 
I modified with the following:

```
    filter_file $tmpdir/recordings $data/wav.scp
    [ -f $data/reco2file_and_channel ] && filter_file $tmpdir/recordings $data/reco2file_and_channel
    [ -f $data/reco2dur ] && filter_file $tmpdir/recordings $data/reco2dur
    true
  fi
} 
```
adding a true before the fi. (it also works with echo "")

My question is that if this problem is related to the bash version or any specific configuration in the main sh file? The set flags used in the main routine are `-e ,-u, and -o pipeline `, I tried removing the second and the third but did not worked.

PR #2439, opened 2018-05-21
[scripts] Fix fix_data_dir.sh bug introduced in #2414; thanks, @Fhrozen

danpovey:


danpovey:
Resolves #2438.  @Ore-an, please note: that 'true' was needed to ensure the function had zero return status.   Due to `set -e`, it causes the script to exit prematurely if it returns nonzero.

Ore-an:
Sorry about that, I probably didn't notice it was inside the function.

PR #2443, opened 2018-05-22
fix fail: All outputs were removed

jtrmal:
@danpovey, for your consideration -- I had the script failing  after
```
nnet3-copy --edits='remove-output-nodes name=output
```
with the error message "All output layers were removed" or something like that.
Plus, improved some other things, most notably support for *.mdl files, not just *.raw nnets.


jtrmal:
Ok, I will update accordingly.
Y.

jtrmal:
updated

PR #2450, opened 2018-05-24
[scripts] Add retry.pl, to be used when queue is flaky.

danpovey:


danpovey:
Note: if you cherry-pick this commit, you should also change the final
`exit($return_status);`
to 
`exit(1)`
as in #2453.

Issue #2448, opened 2018-05-24
lattice-lmrescore-kaldi-rnnlm-pruned, Error: "Possibly not all states of input lattice are co-accessible?"


vesis84:
Hi Hainan, Dan,
the problem with 'lattice-lmrescore-kaldi-rnnlm-pruned' is still there, even after re-building the GMM model/lattices, and when using nnet3-rnnlm:

`ASSERTION_FAILED (lattice-lmrescore-kaldi-rnnlm-pruned[5.4.107-f0333]:ComputeLatticeStateInfo():compose-lattice-pruned.cc:585) : '!arc_costs.empty() && "Possibly not all states of input lattice are co-accessible?"'`

I looked into the lattice generation command, and the `'steps/decode_fmllr.sh'`
generates lattices using options: `'--determinize-lattice=false --allow-partial=true'`.

So, there are 3 ways to fix it, and I am not sure, which one is the best:
A) enable determinization/minimization in the 'steps/decode_fmllr.sh',
B) add determinization/minimization into the RNNLM re-scoring scripts
(for example into the 'rnnlm/lmrescore_pruned.sh')
C) add determinization/minimization into the re-scoring binary: 'lattice-lmrescore-kaldi-rnnlm-pruned'

To me "C" seems best, as it imposes no assumption on the input lattices, and all changes would be done in one place. What would you suggest?

I can change the code, or you (Hainan/Dan) can do it as well,
please let me know what you prefer...

Thanks,
Karel

vesis84:
(after discussing it by email w/o github...)

Hi all,
aha, I see. I tried adding the 'fst::Connect(&clat)' to lattice rescoring tool, and it fixed the problem.

The lattices were generated with 'steps/decode_fmllr.sh', I did not modify anything.
According to the script, the last tool called was: lattice-determinize-pruned-parallel
(and I used the current HEAD from the kaldi-asr/master on github)

Should I open a PR, in which the 'fst::Connect(&clat)' is added to all the RNNLM lattice re-scoring tools?

Thanks,
Karel

vesis84:
And the lattice, which is causing the trouble is far from empty:

```
-rw-r--r--  1 iveselyk fit  620622 May 25 16:14 ...0014251_0015030.lat
-rw-r--r--  1 iveselyk fit  620442 May 25 16:27 ...0014251_0015030.lat.connect
```

And the diff is:

```
[...]

PR #2455, opened 2018-05-26
rnnlm dataprep: the only valid 2-col splitter is: ' ' (space)

vesis84:
- helps if the words in 'words.txt' contain special UTF whitespaces,
  which otherwise lead to having >2 columns,
- it will make the code a little more robust to 'dirty' dataprep,

### DIFF
--- a/scripts/rnnlm/choose_features.py
+++ b/scripts/rnnlm/choose_features.py
@@ -10,6 +10,8 @@ import math
 from collections import defaultdict
 sys.stdout = open(1, 'w', encoding='latin-1', closefd=False)
 
+import re
+tab_or_space = re.compile('[ \t]')
 
 parser = argparse.ArgumentParser(description="This script chooses the sparse feature representation of words. "
                                              "To be more specific, it chooses the set of features-- you compute "
@@ -86,7 +88,7 @@ def read_vocab(vocab_file):
     vocab = {}
     with open(vocab_file, 'r', encoding="latin-1") as f:
         for line in f:
-            fields = line.split()
+            fields = re.split(tab_or_space, line)
             assert len(fields) == 2
             if fields[0] in vocab:
                 sys.exit(sys.argv[0] + ": duplicated word({0}) in vocab: {1}"
@@ -115,7 +117,7 @@ def read_unigram_probs(unigram_probs_file):
     unigram_probs = []
     with open(unigram_probs_file, 'r', encoding="latin-1") as f:
         for line in f:
-            fields = line.split()
+            fields = re.split(tab_or_space, line)
             assert len(fields) == 2
             idx = int(fields[0])
             if idx >= len(unigram_probs):
--- a/scripts/rnnlm/choose_features.py
+++ b/scripts/rnnlm/choose_features.py
@@ -8,7 +8,7 @@ import argparse
 import sys
 import math
 from collections import defaultdict
-sys.stdout = open(1, 'w', encoding='utf-8', closefd=False)
+sys.stdout = open(1, 'w', encoding='latin-1', closefd=False)
 
 
 parser = argparse.ArgumentParser(description="This script chooses the sparse feature representation of words. "
@@ -84,7 +84,7 @@ SPECIAL_SYMBOLS = ["<eps>", "<s>", "<brk>"]
 #  and 'wordlist' is a list indexed by integer id, that returns the string-valued word.
 def read_vocab(vocab_file):
     vocab = {}
-    with open(vocab_file, 'r', encoding="utf-8") as f:
+    with open(vocab_file, 'r', encoding="latin-1") as f:
         for line in f:
             fields = line.split()
             assert len(fields) == 2
@@ -113,7 +113,7 @@ def read_vocab(vocab_file):
 # id of the word, which evaluates to the unigram prob of the word.
 def read_unigram_probs(unigram_probs_file):
     unigram_probs = []
-    with open(unigram_probs_file, 'r', encoding="utf-8") as f:
+    with open(unigram_probs_file, 'r', encoding="latin-1") as f:
         for line in f:
             fields = line.split()
             assert len(fields) == 2
--- a/scripts/rnnlm/get_unigram_probs.py
+++ b/scripts/rnnlm/get_unigram_probs.py
@@ -74,7 +74,7 @@ def get_all_data_sources_except_dev(text_dir):
 #                    value is a tuple (repeated_times_per_epoch, weight)
 def read_data_weights(weights_file, data_sources):
     data_weights = {}
-    with open(weights_file, 'r', encoding="utf-8") as f:
+    with open(weights_file, 'r', encoding="latin-1") as f:
         for line in f:
             try:
                 fields = line.split()
@@ -99,7 +99,7 @@ def read_data_weights(weights_file, data_sources):
 # return the vocab, which is a dict mapping the word to a integer id.
 def read_vocab(vocab_file):
     vocab = {}
-    with open(vocab_file, 'r', encoding="utf-8") as f:
+    with open(vocab_file, 'r', encoding="latin-1") as f:
         for line in f:
             fields = line.split()
             assert len(fields) == 2
@@ -128,10 +128,11 @@ def get_counts(data_sources, data_weights, vocab):
         if weight == 0.0:
             continue
 
-        with open(counts_file, 'r', encoding="utf-8") as f:
+        with open(counts_file, 'r', encoding="latin-1") as f:
             for line in f:
                 fields = line.split()
-                assert len(fields) == 2
+                if len(fields) != 2: print("Warning, should be 2 cols:", fields, file=sys.stderr);
+                assert(len(fields) == 2)
                 word = fields[0]
                 count = fields[1]
                 if word not in vocab:
--- a/scripts/rnnlm/get_unigram_probs.py
+++ b/scripts/rnnlm/get_unigram_probs.py
@@ -7,6 +7,9 @@ import os
 import argparse
 import sys
 
+import re
+tab_or_space = re.compile('[ \t]')
+
 parser = argparse.ArgumentParser(description="This script gets the unigram probabilities of words.",
                                  epilog="E.g. " + sys.argv[0] + " --vocab-file=data/rnnlm/vocab/words.txt "
                                         "--data-weights-file=exp/rnnlm/data_weights.txt data/rnnlm/data "
@@ -77,7 +80,7 @@ def read_data_weights(weights_file, data_sources):
     with open(weights_file, 'r', encoding="latin-1") as f:
         for line in f:
             try:
-                fields = line.split()
+                fields = re.split(tab_or_space, line)
                 assert len(fields) == 3
                 if fields[0] in data_weights:
                     raise Exception("duplicated data source({0}) specified in "
@@ -101,7 +104,7 @@ def read_vocab(vocab_file):
     vocab = {}
     with open(vocab_file, 'r', encoding="latin-1") as f:
         for line in f:
-            fields = line.split()
+            fields = re.split(tab_or_space, line)
             assert len(fields) == 2
             if fields[0] in vocab:
                 sys.exit(sys.argv[0] + ": duplicated word({0}) in vocab: {1}"
@@ -130,8 +133,8 @@ def get_counts(data_sources, data_weights, vocab):
 
         with open(counts_file, 'r', encoding="latin-1") as f:
             for line in f:
-                fields = line.split()
-                if len(fields) != 2: print("Warning, should be 2 cols:", fields, file=sys.stderr);
+                fields = re.split(tab_or_space, line)
+                if len(fields) != 2: print("Warning, should be 2 cols:", fields, line, file=sys.stderr);
                 assert(len(fields) == 2)
                 word = fields[0]
                 count = fields[1]
--- a/scripts/rnnlm/train_rnnlm.sh
+++ b/scripts/rnnlm/train_rnnlm.sh
@@ -188,12 +188,16 @@ while [ $x -lt $num_iters ]; do
         else dest_number=$[x+1]; fi
         # in the normal case $repeated data will be just one copy.
         repeated_data=$(for n in $(seq $num_repeats); do echo -n $dir/text/$split.txt ''; done)
-        
+
         rnnlm_l2_factor=$(perl -e "print (1.0/$this_num_jobs);")
         embedding_l2_regularize=$(perl -e "print ($embedding_l2/$this_num_jobs);")
 
+        # allocate queue-slots for threads doing sampling,
+        num_threads_=$[$num_egs_threads*2/3]
+        [ -f $dir/sampling.lm ] && queue_thread_opt="--num-threads $num_threads_" || queue_thread_opt=
+
         # Run the training job or jobs.
-        $cmd $queue_gpu_opt $dir/log/train.$x.$n.log \
+        $cmd $queue_gpu_opt $queue_thread_opt $dir/log/train.$x.$n.log \
            rnnlm-train \
              --rnnlm.max-param-change=$rnnlm_max_change \
              --rnnlm.l2_regularize_factor=$rnnlm_l2_factor \
--- a/scripts/rnnlm/get_best_model.py
+++ b/scripts/rnnlm/get_best_model.py
@@ -22,7 +22,7 @@ args = parser.parse_args()
 
 num_iters=None
 try:
-    with open(args.rnnlm_dir + "/info.txt", encoding="utf-8") as f:
+    with open(args.rnnlm_dir + "/info.txt", encoding="latin-1") as f:
         for line in f:
             a = line.split("=")
             if a[0] == "num_iters":
@@ -41,7 +41,7 @@ best_iter=-1
 for i in range(1, num_iters):
     this_logfile = "{0}/log/compute_prob.{1}.log".format(args.rnnlm_dir, i)
     try:
-        f = open(this_logfile, 'r', encoding='utf-8')
+        f = open(this_logfile, 'r', encoding='latin-1')
     except:
         sys.exit(sys.argv[0] + ": could not open log-file {0}".format(this_logfile))
     this_objf=-1000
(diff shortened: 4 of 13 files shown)


scripts/rnnlm/get_unigram_probs.py
```diff
@@ -130,8 +130,9 @@ def get_counts(data_sources, data_weights, vocab):
 
         with open(counts_file, 'r', encoding="utf-8") as f:
             for line in f:
-                fields = line.split()
-                assert len(fields) == 2
+                fields = line.split(' ')
```
danpovey: This was definitely a bug.  The rule for things like words.txt in Kaldi is that they should never contain anything which, *when interpreted as an ASCII character*, is a space.
And in general, we don't even assume that things like text files are actually encoded in UTF-8 or a compatible encoding; we only require that the spaces between words be ASCII space (' ').
So I believe the correct fix here would be to change encoding="utf-8" to encoding="latin-1".
Would you mind testing whether that change works for your setup?
Make sure there are no similar issues in the rnnlm/ subdirectory.

vesis84: Hi, okay, so, if the only requirement is to contain ASCII-space ' ' as a separator, this is also fine for texts in UTF encoding. The code for whitespace is always '0x20'. 

If we changed encoding in python script `encoding="utf-8"` to `encoding="latin-1"`, the further development of the python scripts would become more difficult, as the prints would not be able to print the correct symbols. So I'd keep the encoding filter as it is.

The problem was that the `line.split()` splits string with on any white-space character, while with `line.split(' ')` we are narrowing to split only with the ASCII-space `0x32`, and it does not matter if the `line` is a UTF string.

So, I'll look for other `split()` commants in the directory and change them if necessary...

vesis84:
Done! The data preparation seems to work well. There is no error. It would be good to test it with one of your other recipes too... (maybe one of the JHU students could do it, before merging-in?)

danpovey:
Karel, there is another reason to use latin-1, and it explains why we are
intending to standardize its use everywhere.  It's that the scripts are
supposed to support other non-UTF8 (but ASCII-compatible) encodings, such
as GBK.

So please change as I asked, by reading as latin-1.

Dan

danpovey:
... and regarding printing: for programs like this, we should probably set the encoding for sys.stdout to latin-1 as well.  The intention is that the output should look the same as the input, when interpreted as a bytestring, because programs like this have no need to break up words internally or even know what the encoding is, other than being able to split on whitespace.  But we can leave that till later; I'm not sure of the correct invocation to do it.

vesis84:
Aha, well, still I have mixed feelings about using 'latin-1'. It seems to be against the current trend, in which the UTF-8 is more and more wide-spread than ASCII : https://en.wikipedia.org/wiki/UTF-8#/media/File:Utf8webgrowth.svg

On the other hand I checked that the ASCII-space '0x20' cannot be the 2nd/3rd/4th UTF-8 byte, nor the 2nd GBK byte. So I will do the change and test it, in the way you prefer it...

jtrmal:
Hi Karel, the Latin encoding in this context in effect says 'I don't care,
as long as space is 0x20', which should be true for  encodings where the
smallest unit is 8 bits, I. E including utf8. Will not work with ucs
encodings (smallest units 16 or 32 bits). But we don't really support those.
Y.

vesis84:
Well, I tested it partially. The data-prep is running fine. While I have problem with training the model (getting the CUSPARSE error I described before in #2448): 
`ERROR (rnnlm-train[5.4.146~5-6b94e]:CopyFromSmat():cu-sparse-matrix.cc:395) cusparseStatus_t 6 : "CUSPARSE_STATUS_EXECUTION_FAILED" returned from 'cusparse_csr2csc(GetCusparseHandle(), smat.NumRows(), smat.NumCols(), smat.NumElements(), smat.CsrVal(), smat.CsrRowPtr(), smat.CsrColIdx(), CsrVal(), CsrColIdx(), CsrRowPtr(), CUSPARSE_ACTION_NUMERIC, CUSPARSE_INDEX_BASE_ZERO)'
`
I keep getting this much more often then I did last week. But AFAIK this is independent to the change in the data preparation for RNNLM training.

But it is true that I did not finish the 'testing by lattice rescoring'.

vesis84:
Hi, I found another issue, in rnnlm training there is multi-threaded sampling of targets, but the script was getting just one slot for all the threads... See the fix : c6193c5 (Bascially, it adds `--num-threds N` to the call of the `queue.pl`.) This is likely to improve stability of the training jobs...
K.

vesis84:
Hmm, still getting that: "CUSPARSE_STATUS_EXECUTION_FAILED" I don't think I can finish the test in our cluster, it keeps on crashing every 3-5th iteration, this is really annoying :(

vesis84:
But the data preparation seems to be fine...

danpovey:
Karel, did you remember the "export", i.e. "export CUDA_CACHE_DISABLE=1"?
It's in your path.sh?

danpovey:
OK, looking back at my emails, it looks like others have encountered this problem before.
However, we were never able to fix it because we could not replicate it here at Hopkins.

I would really appreciate it if you could run the program in cuda-memcheck-- preferably repeatedly, on some kind of automated loop, on a machine where you previously had the error, and wait till you get the error.  I really don't understand what the issue is.  It could be some subtle concurrency problem.

In parallel, you could just go ahead and train your model with retry.pl.  But obviously back up some files that are sufficient to reproduce the problem.

Just for the record, can you also let us know the CUDA version and driver version and which GPU hardware types you get the failure on?

danpovey:
Karel: 
We started getting failures like this at JHU.  This is good news as it means we can investigate the problem and find the cause.  Previously when I reported the problem to NVidia as a likely driver bug they told me that it was probably an error in our code.  But now I think I have something to show them: I am seeing things in the system logs at the same time as these errors occur.  This is from the output of `dmesg -T`:

```
[Fri Jun  1 15:35:11 2018] NVRM: GPU at PCI:0000:02:00: GPU-56dda1e1-ec05-17a6-ec14-d33d60198868
[Fri Jun  1 15:35:11 2018] NVRM: GPU Board Serial Number: 
[Fri Jun  1 15:35:11 2018] NVRM: Xid (PCI:0000:02:00): 31, Ch 00000010, engmask 00000101, intr 10000000
```

Can you see whether you can find the same on your system?
You may need to ask your sysadmins: not all Linux flavors allow users to see `dmesg` output.

scripts/rnnlm/choose_features.py
```diff
@@ -84,9 +84,9 @@
 #  and 'wordlist' is a list indexed by integer id, that returns the string-valued word.
 def read_vocab(vocab_file):
     vocab = {}
-    with open(vocab_file, 'r', encoding="utf-8") as f:
+    with open(vocab_file, 'r', encoding="latin-1") as f:
         for line in f:
-            fields = line.split()
+            fields = line.split(' ')
```
danpovey: @vesis84, you can remove the ' ' here and elsewhere.  If the file contained tabs, we do want to split on those as well, even though that's now how we expect people will write the file.

vesis84: Hi, we could eventually support only the spaces and tabs and nothing else by: 
`re.split("[ \t]", str)` Does that sound good? (it's more explicit and readable...)
K.

vesis84:
Hi, I just found an answer to the stability problem. 
- **The cause:** recent upgrade of GPU driver to `384.130`, this **disabled the compute exclusive mode** as a by-product.
- **The consequence:** 2 processes sharing 1 GPU (part of GPU-RAM taken by other process), and the CUDA allocator in Kaldi already has a lot of GPU-RAM, so there is not enough GPU-RAM for the `cusparse_csr2csc(.)` call
- **The outcome:** error message `"CUSPARSE_STATUS_EXECUTION_FAILED"`, which means _"we just ran out of GPU memory..."_

After enabling the 'process exclusive mode' and using GPUs with `>=8GB RAM`, the training is stable, and the problem is solved...

vesis84:
And I found similar error message in `dmesg -t` (seems to be a bit old, it's from March 21st):
```
NVRM: loading NVIDIA UNIX x86_64 Kernel Module  384.130  Wed Mar 21 03:37:26 PDT 2018 (using threaded interrupts)
NVRM: GPU at PCI:0000:02:00: GPU-7fef535b-ed05-2a7d-3398-57aa66ebd4b9
NVRM: GPU Board Serial Number: 
NVRM: Xid (PCI:0000:02:00): 31, Ch 00000020, engmask 00000101, intr 10000000
NVRM: GPU at PCI:0000:02:00: GPU-7fef535b-ed05-2a7d-3398-57aa66ebd4b9
NVRM: GPU Board Serial Number: 
NVRM: Xid (PCI:0000:02:00): 31, Ch 00000020, engmask 00000101, intr 10000000
```
But I am not sure how to interpret the codes there... But I did not investigate it yet...
K.

scripts/rnnlm/choose_features.py
danpovey: I'd prefer just split(), I think, it's easier to remember and duplicate.

danpovey:
We are actually getting the CUSPARSE_STATUS_EXECUTION_FAILED on our grid and we do have exclusive mode.  From our initial investigations it seems to be some subtle concurrency problem  @hainan-xv is working on it.  (Right, hainan?)

scripts/rnnlm/train_rnnlm.sh
```diff
@@ -25,7 +25,10 @@ embedding_lrate_factor=0.1  # the embedding learning rate is the
                             # nnet learning rate times this factor.
 backstitch_training_scale=0.0    # backstitch training scale
 backstitch_training_interval=1   # backstitch training interval
+
 cmd=run.pl  # you might want to set this to queue.pl
+queue_gpu_opt="--gpu 1" # you may change the GPU opt externally,
```
danpovey: @vesis84, please don't do this-- this is not really consistent with our normal way of working.
These kinds of things are configurable by changing queue.conf.

scripts/rnnlm/train_rnnlm.sh
```diff
@@ -182,18 +185,21 @@ while [ $x -lt $num_iters ]; do
...
         else dest_number=$[x+1]; fi
         # in the normal case $repeated data will be just one copy.
         repeated_data=$(for n in $(seq $num_repeats); do echo -n $dir/text/$split.txt ''; done)
-        
+
         rnnlm_l2_factor=$(perl -e "print (1.0/$this_num_jobs);")
         embedding_l2_regularize=$(perl -e "print ($embedding_l2/$this_num_jobs);")
 
+        # allocate queue-slots for threads doing sampling,
+        [ -f $dir/sampling.lm ] && queue_thread_opt="--num-threads $num_egs_threads" || queue_thread_opt=
```
danpovey: Normally it won't actually use that much CPU on average because it's limited by GPU time.  I don't know how many it uses though.
The num_egs_threads is more like an upper bound to make sure it is not limited by the sampling.

scripts/rnnlm/train_rnnlm.sh
vesis84: Yes, originally I did this with 'queue.conf'. But is it possible somehow to have 2 queues, one with all the GPUs and other with GPUs that have more than, say 7G RAM? (the big GPUs are not always necessary...)

scripts/rnnlm/train_rnnlm.sh
vesis84: Aha, at our servers, it was running with 600% CPU, while the num-threads was set to 10. So it was 'eating' the CPU time of other slots (GPU process is supposed to consum 100% CPU only).

danpovey:
For the queue thing, you can have one command be e.g.
"queue.pl --config conf/queue_big_gpu_mem.conf"
For the slots thing-- yes, I guess it makes sense to add the requirement,
but maybe as a middle ground, set it to half the requested num-thread?  Too
many reserved slots aren't good either, normally.

Dan

scripts/rnnlm/choose_features.py
vesis84: Well, `split()` splits on any whitespace character (this can be a problem for example with `\r` characters). Are you sure you want to allow to split any "whitespace" character? (and sorry for explaining if you already are aware of it...)

danpovey:
We don't anticipate \r or \t in text lines, but if it's there, we'd want it
to be treated as space.  What we support is ASCII-compatible encodings,
i.e. encodings where spaces have the same meaning as they do in ASCII.

vesis84:
Good, I have incorporated the comments: 
( `split(' ') -> split()`, num threads is 2/3 of `$num_egs_threads`).

Now, at testing, I am getting again the original error I had before...
It is in `rnnlm/get_unigram_probs.py`, line 133.

A 'counts' file is read in 'latin-1' encoding, the line is: 
`w\xc3\xa0nazokumbana 1` 
and it is wrongly split as 
`['w\xc3', 'nazokumbana', '1']`

This means that for some reason `split()` interprets the char `\xa0` as whitespace.
 An obvious solution would be to use: `split(' ')` or `re.split('[ \t]', line)`

What should be the next step?

danpovey:
OK, so that is unicode character U+00A0, encoded as the bytes "C3" "A0"
(written as hex).  The second byte, A0 (==160) does map to non-breaking
space in the latin-1 character set.

We have to decide whether to allow &nbsp (non-breaking space) in words--
that will determine the course of action.  Currently, validate_lang.pl
actually does not allow &nbsp: you can see in validate_utf8_whitespaces.
But there may actually be situations in, say, Arabic, where &nbsp and
similar non-standard whitespaces might validly appear inside a word--
intended, for example, to alter the way letters join together.   @hhadian
and @jtrmal, do you have any comment?  Yenda, you were the one who added
the code to check for those types of whitespace in UTF text, so I assume
you remember the original reason why we banned them.

danpovey:
OK, I had another look at how the Unicode format works
https://www.fileformat.info/info/unicode/utf8.htm
and it looks like the byte "A0" (i.e. byte value 160) could potentially appear in a lot of Unicode characters; for instance c3a0 is a valid utf-8 byte sequence, meaning a grave-accented "a".  That means that whenever we split data that is encoded as latin-1, we can't do just "split()", because it could spuriously split a lot of UTF characters.

I think Karel's original plan of splitting on " " explicitly would actually make the most sense.
I now see that sym2int.pl (which is the other program that mainly deals with splitting of text data) explicitly splits on " ", and keeping those two things consistent would probably make sense.

The other way we sometimes split is implicitly using awk (always with LC_ALL=C exported).  That probably splits on tabs and \r as well, but probably the easiest way to resolve this difference is to just say that tabs and linefeed characters are banned in text data such as in 'data/text'.  We could change validate_lang.pl to enforce this.

That still leaves open the question of whether &nbsp is banned in utf8-encoded text, and we do need to resolve that, but it makes it a separate issue.

jtrmal:
I'm trying to remember, but I don't know. Those were the times we didn't
have to lie about the encoding just to make python3 happy.
It was certainly because people weren't cleaning up the data properly and
were getting whitespaces into the lexicon and similar files (where,
consequently, got split). But I don't remember the concrete case that was
the one that made us do it -- certainly, there was one.
y.

jtrmal:
I'd split on tabs and spaces, because we are pretty careless sometimes (for
example in the lexicon stuff).
Plus, speaking of the lexicon, those tabs were sometimes used to split on
syllabs, but I don't recall now.
y.

hhadian:
In Farsi (and probably Arabic), zero-width space and zero-width non-joiner (&zwnj) are commonly used in the middle of compound words, not sure about $nbsp but I don't remember seeing it in Farsi texts (I guess it's mostly popular in HTML).

jtrmal:
I think we used to recommend to remove them or map them to something
visible anyway as it can cause troubles hard to debug.
ZWJ does not have the unicode isWhitespace set (
https://www.fileformat.info/info/unicode/char/200d/index.htm)
y.

danpovey:
OK, so I think we're good then, assuming Hossein thinks it's OK for
Arabic.  Our rule will be: we disallow any space characters that UTF-8
considers as space, that are different from space, tab and newline.  That
permits zero-width non-joiner but not zero-width space.  Hossein, do you
think we can get by with that?  E.g. can we map any zero-width spaces to
something else, like regular space, without badly distorting something or
losing meaning?

That requires us to
 -- modify validate_text.pl and validate_lang.pl to explicitly disallow
linefeed (they should probably print something that's specific to linefeed,
so people don't think it's a UTF-8 issue).

I realized that sym2int.pl is actualy splitting on all whitespaces
including tabs, spaces and linefeeds, since in perl, split(" ", ...) has a
special meaning.

Karel, so you should be splitting on spaces or tabs; and if you feel like
it, make those modifications to validate_text.pl and validate_lang.pl to
reject linefeed (if not, someone else can do it).

hhadian:
Yes, I think. Any zero-width space can be mapped to zero-width non-joiner (actually, I guess zero-width non-joiner [is the standard way](https://en.wikipedia.org/wiki/Zero-width_non-joiner)) or even space without losing meaning.

johnjosephmorgan:
I remember running into these zero width characters in Dari and Pashto. people would get mad at me when I would delete them, so I think they might be important.

danpovey:
Hm, OK.  But Hossein is saying that it should be possible to replace
zero-width space with zero-width joiner.  What I am *hoping* is that for
each character that has Unicode's "space" property, there is another usable
equivalent that isn't considered a space.  But I don't know this for sure.
It will definitely make our lives easier if we can ensure that words don't
contain spaces internally.

johnjosephmorgan:
Yes. I think that works for zero width space.
I was worried about ligatures.
If you want a ligature, you replace zero width space with zero width joiner.
If you do not want a ligature, you replace zero width space with zero width non-joiner.

vesis84:
Aha, okay, so, you mean... I should modify the `validate_text.pl` and `validate_lang.pl` to disallow the CR (not LF) both in 'words.txt' and in 'data/text'. Is that correct?

vesis84:
I am not sure how to make the change in `validate_*.pl` correctly and I need to leave for today. Maybe someone else could do it? Or, if it can wait, I'll look into it tomorrow...

danpovey:
Thanks.  Once you confirm that you've tested it (at least that it doesn't crash in the early stages), I'll merge.  @jtrmal, do you have time to change the validation scripts to ban CR (\r)?

jtrmal:
I'll do it.
y.

vesis84:
Hi, for me the dataprep and RNNLM training runs fine. A.t.m. waiting for the training to finish... K.

PR #2459, opened 2018-05-29
use ‘bc’ command to process float math

songmeixu:
When use 'float or < 1' percent args, get error like this:
image
Thus, use 'bc' instand.

### DIFF
--- a/egs/wsj/s5/utils/nnet/subset_data_tr_cv.sh
+++ b/egs/wsj/s5/utils/nnet/subset_data_tr_cv.sh
@@ -39,7 +39,7 @@ if [ -z "$cv_spk_list" ]; then
   # Select 'cv_spk_percent' speakers randomly,
   cat $src_data/spk2utt | awk '{ print $1; }' | utils/shuffle_list.pl --srand $seed >$tmp/speakers
   n_spk=$(wc -l <$tmp/speakers)
-  n_spk_cv=$(echo "($cv_spk_percent * $n_spk) / 100" | bc)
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/nnet/subset_data_tr_cv.sh
```diff
@@ -39,7 +39,7 @@ if [ -z "$cv_spk_list" ]; then
   # Select 'cv_spk_percent' speakers randomly,
   cat $src_data/spk2utt | awk '{ print $1; }' | utils/shuffle_list.pl --srand $seed >$tmp/speakers
   n_spk=$(wc -l <$tmp/speakers)
-  n_spk_cv=$((cv_spk_percent * n_spk / 100))
+  n_spk_cv=$(echo "($cv_spk_percent * $n_spk) / 100" | bc)
```
danpovey: We prefer to use "perl -e" for this, as bc is not one of the required things we check for when installing Kali.
But I'm kind of surprised you'd need to use a floating point value.  Setting such a small validation set (<1%) would normally only make sense if your dataset was huge, and the nnet1 tools don't support training on multiple GPUs so I don't know how much sense it makes to process huge datasets with those tools.

songmeixu: Hi, dan. Yeah, my dataset is huge(2.5M utts). I use this script not for nnet1, but for all stages of training, a small dev set is needed for validation.

I like "perl -e" method also, and it is committed.

PR #2461, opened 2018-05-29
[src] Fix bug in lattice-oracle relating to wildcards in lattice output.

dogancan:
@danpovey @francisr After this fix wildcards are preserved in the lattice output. Note that the text output does not include wildcards as before. The warning about oracle path not being found in the original lattice should never be printed after this change.

### DIFF
--- a/src/latbin/lattice-oracle.cc
+++ b/src/latbin/lattice-oracle.cc
@@ -26,18 +26,16 @@
 
 namespace kaldi {
 
-using std::vector;
-using std::set;
-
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks!  @francisr, can you please test?

francisr:
It's working for me.   
I've added a `TopSortCompactLatticeIfNeeded` before writing the lattices because I like them to be sorted for quick inspection, but that's just a nice to have.

PR #2463, opened 2018-05-30
TopSort lattice-oracle lattice output

dogancan:
@danpovey @francisr This commit topologically sorts the lattice output of lattice-oracle before writing it. It also sorts the lattice arcs after relabeling wildcards (before composing with the mask) which might improve the runtime for large input lattices. 

By the way, the oracle computation implementation seems to be somewhat inefficient. I tried to improve the runtime by doing the composition lazily but that did not give a significant improvement over the current approach on the example lattices. Factoring the edit transducer should help but I did not look into that. Is this something worth looking into? It seems like this binary is not used all that much.

### DIFF
--- a/src/latbin/lattice-oracle.cc
+++ b/src/latbin/lattice-oracle.cc
@@ -26,18 +26,16 @@
 
 namespace kaldi {
 
-using std::vector;
-using std::set;
-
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks.  I'll merge when checks are done.
I don't think this is used that much, you're right.
Probably mainly for diagnostics.

PR #2469, opened 2018-06-01
 Added tdnn_opgru recipe for multi-en

xiaohui-zhang:


danpovey:
can you please start the numbering for the opgru things from 1a, and also switch that setup over to using a tuning/ directory?

xiaohui-zhang:
Sure Dan. will do. Previously I started the numbering from 5 in order to differentiate nnet systems trained on top of tri5a (without librispeech) or tri6a (with librispeech). Now that we've already known librispeech won't help, I will start numbering from 1.

xiaohui-zhang:
done

PR #2473, opened 2018-06-03
Add backstitch option to steps/nnet3/chain/train.py

teinhonglo:
Adding args "trainer.optimization.backstitch-training-scale" and "backstitch-training-interval".
All file change is in "option to steps/nnet3/chain/train.py".


danpovey:
This script already has that option.  Run it without args and you'll see.
see steps/libs/nnet3/train/common.py

PR #2475, opened 2018-06-05
 [egs,scripts] Fix bug in make_voxceleb1.pl

weichcai:
Fix train/test data split bug in make_voxceleb1.pl


danpovey:
@david-ryan-snyder  or @leibny, please check.

david-ryan-snyder:
@weichcai, what you said make sense, thanks for catching it.

The results will need to be updated. Did you happen to rerun either of the recipes (e.g., the v1 recipe or the v2 recipe)?

david-ryan-snyder:
Also adding @entn-at for his info, since he's been working with this recipe.

weichcai:
Not yet. I have only run a small experiments just use Voxceleb1 (rather than Voxceleb1 + Voxceleb2 in Kaldi)  as my training set. In my case, after running my own fixed script, it didn't change too much, only a little bit worse.

david-ryan-snyder:
OK, thanks.

@danpovey, could you merge this?

I'll make a new PR to update the results.

PR #2477, opened 2018-06-06
Using nj variable in run_ivector_common.sh

sih4sing5hong5:


egs/wsj/s5/local/nnet3/run_ivector_common.sh
```diff
@@ -110,7 +110,7 @@ if [ $stage -le 4 ]; then
   # can be sensitive to the amount of data.  The script defaults to an iVector dimension of
   # 100.
   echo "$0: training the iVector extractor"
-  steps/online/nnet2/train_ivector_extractor.sh --cmd "$train_cmd" --nj 10 \
+  steps/online/nnet2/train_ivector_extractor.sh --cmd "$train_cmd" --nj $nj \
```
danpovey: Each job actually uses 16 threads by default, so this may be too high.
It might make more sense to make this separately configurable.
Also it looks like $num_threads_ubm is not passed into this script.
If someone has fewer slots on their queue it may make more sense for them to just edit variables at the top of run_ivector_common.sh rather than pass them in though.

sih4sing5hong5:
Thank you for your detailed explanation.

PR #2479, opened 2018-06-07
framework of batch version nnet compute（WIP）

LvHang:
@vimalmanohar @pegahgh 
This is the framework of the batch version nnet compute. I write a new class which is named "DecodableNnetSimpleBatch" to deal with it. The new class is similar with class "DecodableNnetSimple". It tries to deal "minibatch_size" utterances at the same time.
The "nnet3-compute-batch.cc" is the fundamental way to use this class in my mind. And the comments in each function in "nnet-am-decodable-simple.h" show the flow.
I hope my understanding is rigth. I'm not sure.
When you are free, please to check it. Thank you for your feedback.

### DIFF
--- a/src/nnet3/Makefile
+++ b/src/nnet3/Makefile
@@ -31,7 +31,7 @@ OBJFILES = nnet-common.o nnet-compile.o nnet-component-itf.o \
   nnet-compile-looped.o decodable-simple-looped.o \
   decodable-online-looped.o convolution.o \
   nnet-convolutional-component.o attention.o \
-  nnet-attention-component.o nnet-tdnn-component.o
+  nnet-attention-component.o nnet-tdnn-component.o nnet-batch-compute.o
 
 
 LIBNAME = kaldi-nnet3
--- a/src/nnet3/nnet-batch-compute.cc
+++ b/src/nnet3/nnet-batch-compute.cc
@@ -1,0 +1,740 @@
+// nnet3/nnet-batch-compute.cc
+
+// Copyright 2012-2018  Johns Hopkins University (author: Daniel Povey)
+//           2018       Hang Lyu
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include <algorithm>
+#include "nnet3/nnet-batch-compute.h"
+#include "nnet3/nnet-utils.h"
+
+namespace kaldi {
+namespace nnet3 {
+
+BatchNnetComputer::BatchNnetComputer(
+    const NnetSimpleComputationOptions &opts,
+    const Nnet &nnet,
+    const VectorBase<BaseFloat> &priors,
+    int32 online_ivector_period,
+    bool  ensure_exact_final_context,
+    int32 minibatch_size):
+    opts_(opts),
+    nnet_(nnet),
+    output_dim_(nnet_.OutputDim("output")),
+    log_priors_(priors),
+    online_ivector_period_(online_ivector_period),
+    compiler_(nnet_, opts_.optimize_config),
+    ensure_exact_final_context_(ensure_exact_final_context),
+    minibatch_size_(minibatch_size) {
+  // Check the Nnet
+  KALDI_ASSERT(IsSimpleNnet(nnet));
+  ComputeSimpleNnetContext(nnet, &nnet_left_context_, &nnet_right_context_);
+  log_priors_.ApplyLog();
+  CheckAndFixConfigs();
+  // Prepare ComputationRequest and store a NnetComputation object
+  PrepareComputationRequest();
+}
+
+
+void BatchNnetComputer::CheckAndFixConfigs() {
+  static bool warned_frames_per_chunk = false;
+  int32 nnet_modulus = nnet_.Modulus();
+  if (opts_.frame_subsampling_factor < 1 ||
+      opts_.frames_per_chunk < 1) {
+    KALDI_ERR << "--frame-subsampling-factor and "
+              << "--frames-per-chunk must be > 0";
+  }
+  KALDI_ASSERT(nnet_modulus > 0);
+  int32 n = Lcm(opts_.frame_subsampling_factor, nnet_modulus);
+
+  if (opts_.frames_per_chunk % n != 0) {
+    // round up to the nearest multiple of n.
+    int32 frames_per_chunk = n * ((opts_.frames_per_chunk + n - 1) / n);
+    if (!warned_frames_per_chunk) {
+      warned_frames_per_chunk = true;
+      if (nnet_modulus == 1) {
+        // simpler error message.
+        KALDI_LOG << "Increasing --frames-per-chunk from "
+                  << opts_.frames_per_chunk << " to "
+                  << frames_per_chunk << " to make it a multiple of "
+                  << "--frame-subsampling-factor="
+                  << opts_.frame_subsampling_factor;
+      } else {
+        KALDI_LOG << "Increasing --frames-per-chunk from "
+                  << opts_.frames_per_chunk << " to "
+                  << frames_per_chunk << " due to "
+                  << "--frame-subsampling-factor="
+                  << opts_.frame_subsampling_factor << " and "
+                  << "nnet shift-invariance modulus = " << nnet_modulus;
+      }
+    }
+    opts_.frames_per_chunk = frames_per_chunk;
+  }
+}
+
+
+void BatchNnetComputer::PrepareComputationRequest() {
+  int32 input_dim = nnet_.InputDim("input"),
+        ivector_dim = nnet_.InputDim("ivector");
+  KALDI_ASSERT(input_dim > 0);
+
+  std::pair<int32, int32> context;
+  context = std::make_pair(opts_.extra_left_context + nnet_left_context_,
+                           opts_.extra_right_context + nnet_right_context_);
+  batch_info_[context] = new BatchInfoQueue();
+
+  if (opts_.extra_left_context_initial != opts_.extra_left_context &&
+      opts_.extra_left_context_initial >= 0) {
+    context = std::make_pair(opts_.extra_left_context_initial +
+                             nnet_left_context_,
+                             opts_.extra_right_context + nnet_right_context_);
+    batch_info_[context] = new BatchInfoQueue();
+  }
+
+  if (opts_.extra_right_context_final != opts_.extra_right_context &&
+      opts_.extra_right_context_final >= 0) {
+    context = std::make_pair(opts_.extra_left_context + nnet_left_context_,
+                             opts_.extra_right_context_final +
+                             nnet_right_context_);
+    batch_info_[context] = new BatchInfoQueue();
+  }
+
+  for (BatchInfoMap::iterator iter =
+       batch_info_.begin(); iter != batch_info_.end(); iter++) {
+    context = iter->first;
+    int32 left_context = context.first,
+          right_context = context.second;
+    // Actually, when opts_.frame_subsampling_factor > 1,  the chunk size will
+    // be less than opts_.frames_per_chunk.
+    int32 num_input_rows = left_context + opts_.frames_per_chunk +
+      right_context - opts_.frame_subsampling_factor + 1,
+          num_output_rows = opts_.frames_per_chunk /
+                            opts_.frame_subsampling_factor;
+
+    ComputationRequest* request = new ComputationRequest();
+    request->need_model_derivative = false;
+    request->store_component_stats = false;
+    request->inputs.reserve(2);
+    request->outputs.reserve(1);
+    std::vector<Index> input_indexes, ivector_indexes, output_indexes;
+
+    for (int32 n = 0; n < minibatch_size_; n++) {
+      for (int32 t = 0; t < num_input_rows; t++) {
+        input_indexes.push_back(Index(n, t - left_context, 0));
+      }
+    }
+    if (ivector_dim > 0) {
+      for (int32 n = 0; n < minibatch_size_; n++) {
+        ivector_indexes.push_back(Index(n, 0, 0));
+      }
+    }
+    for (int32 n = 0; n < minibatch_size_; n++) {
+      for (int32 i = 0; i < num_output_rows; i++) {
+        output_indexes.push_back(Index(n, i * opts_.frame_subsampling_factor,
+                                       0));
+      }
+    }
+
+    request->inputs.push_back(IoSpecification("input", input_indexes));
+    if (ivector_dim > 0) {
+      request->inputs.push_back(IoSpecification("ivector", ivector_indexes));
+    }
+    request->outputs.push_back(IoSpecification("output", output_indexes));
+    context_to_request_[context] = request;
+  }
+
+  if (ensure_exact_final_context_) {
+    context = std::make_pair(-1, -1);
+    batch_info_[context] = new BatchInfoQueue();
+  }
+}
+
+
+void BatchNnetComputer::AcceptInput(
+    const std::string &utt_id,
+    const Matrix<BaseFloat> *feats,
+    const Vector<BaseFloat> *ivector,
+    const Matrix<BaseFloat> *online_ivectors) {
+  // Check the input fits with the nnet.
+  CheckInput(feats, ivector, online_ivectors);
+
+  utt_list_.push_back(utt_id);
+  feats_[utt_id] = feats;
+  if ( ivector != NULL ) {
+    ivectors_[utt_id] = ivector;
+  }
+  if ( online_ivectors != NULL ) {
+    online_ivector_feats_[utt_id] = online_ivectors;
+  }
+  is_computed_[utt_id] = false;
+
+  // Compute number of output frames
+  int32 cur_num_subsampled_frames =
+    (feats->NumRows() + opts_.frame_subsampling_factor - 1) /
+    opts_.frame_subsampling_factor;
+
+  num_subsampled_frames_[utt_id] = cur_num_subsampled_frames;
+
+  // Allocate storage space for output
+  Matrix<BaseFloat> *log_post = new Matrix<BaseFloat>(
+      cur_num_subsampled_frames, output_dim_);
+  log_post_[utt_id] = log_post;
+
+  PrepareBatchInfo();
+}
+
+
+void BatchNnetComputer::CheckInput(const Matrix<BaseFloat> *feats,
+                                   const Vector<BaseFloat> *ivector,
+                                   const Matrix<BaseFloat> *online_ivectors) {
+  KALDI_ASSERT(!(ivector != NULL && online_ivectors != NULL));
+  KALDI_ASSERT(!(online_ivectors != NULL && online_ivector_period_ <= 0 &&
+                 "You need to set the --online-ivector-period option!"));
+  int32 feature_dim = feats->NumCols();
+  int32 ivector_dim = 0;
+  if (ivector != NULL) {
+    ivector_dim = ivector->Dim();
+  }
+  if (online_ivectors != NULL) {
+    ivector_dim = online_ivectors->NumCols();
+  }
+  int32 nnet_input_dim = nnet_.InputDim("input"),
+        nnet_ivector_dim = std::max<int32>(0, nnet_.InputDim("ivector"));
+  if (feature_dim != nnet_input_dim)
+    KALDI_ERR << "Neural net expects 'input' features with dimension "
+              << nnet_input_dim << " but you provided "
+              << feature_dim;
+  if (ivector_dim != std::max<int32>(0, nnet_.InputDim("ivector")))
+    KALDI_ERR << "Neural net expects 'ivector' features with dimension "
+              << nnet_ivector_dim << " but you provided " << ivector_dim;
+}
+
+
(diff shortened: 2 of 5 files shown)


src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,239 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
 };
 
 
+/*
+  This class handles the neural net computation in parallel.
+
+  It can accept just input features, or input features plus iVectors.  */
+class DecodableNnetSimpleBatch {
```
danpovey: Let's not have 'Decodable' in the name since it doesn't fit that interface.

LvHang:
I updated the complete code. The nnet3bin/nnet-compute-batch.cc shows the main flow. The GetFinishedUtterance() is a non-blocking call. The Compute() is a blocking call. The core code is PrepareBatchInfo() and DoNnetComputation().
I'm testing it and trying to fix any bug.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
 };
 
 
+/*
+  This class handles the neural net computation in parallel.
```
danpovey: I think you should be clearer about this since the code itself is no more parallel than the normal code.
Maybe something like:
 ```
This class does neural net inference in a way that is optimized for GPU use: it combines
chunks of multiple utterances into minibatches for more efficient computation. 
```

File: src/nnet3/nnet-am-decodable-simple.h [hunk 3]
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+                        has no locking mechanism (and it would be tricky to design one,
+                        as we'd need to lock the individual computations also),
+                        so the calling code has to make sure that if there are
+                        multiple threads, they do not share the same compiler
+                        object.
+     @param [in] online_ivector_period If you are using iVectors estimated 'online'
+                        (i.e. if online_ivectors != NULL) gives the periodicity
+                        (in frames) with which the iVectors are estimated.
+     @param [in] minibatch_size The capacity of the batch. In general, it means
+                        the number of chunks will be process.
```
danpovey: you might want to mention where the chunk size comes from (I assume somewhere in opts).

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+                    CachingOptimizingCompiler *compiler,
+                    int32 online_ivector_period = 0,
+                    int32 minibatch_size = 1,
+                    int32 max_capacity = 10);
+
+
+  // returns the number of frames of likelihoods.  The same as feats_.NumRows()
+  // in the normal case (but may be less if opts_.frame_subsampling_factor !=
+  // 1).
+  inline int32 NumFrames(std::string& utt_id) const { return *num_subsampled_frames_.find(utt_id); }
```
danpovey: I don't think we need this interface function.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+
+  inline int32 OutputDim() const { return output_dim_; }
+
+  inline bool Ready() const { return batch_info_queue_.size() == minibatch_size_; }
```
danpovey: Please explain in a comment what this means.  Perhaps it means that the class contains enough data that you can call Compute() and it would do something even with flush == false?

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+                        so the calling code has to make sure that if there are
+                        multiple threads, they do not share the same compiler
+                        object.
+     @param [in] online_ivector_period If you are using iVectors estimated 'online'
+                        (i.e. if online_ivectors != NULL) gives the periodicity
+                        (in frames) with which the iVectors are estimated.
+     @param [in] minibatch_size The capacity of the batch. In general, it means
+                        the number of chunks will be process.
+     @param [in] max_capacity The number of feature matrixes can be stored in
+                        this class. In the case of "very small batch size" or
```
danpovey: I don't feel like this comment really explains what max_capacity does and how the class interprets it.
Actually I don't see why it would be necessary.  And these feature matrices-- do these correspond to utterances, or chunks of utterances?

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+  // single iVector input, estimated in batch-mode ('ivector'), or 'online'
+  // iVectors ('online_ivectors' and 'online_ivector_period', or none at all.
+  void AcceptInput(const std::string &utt_id,
+                   const MatrixBase<BaseFloat> &feats,
+                   const VectorBase<BaseFloat> *ivector = NULL,
+                   const MatrixBase<BaseFloat> *online_ivectors = NULL);
+  
+  
+  // Gets the output for a finished utterance. This is a non-blocking call.
+  bool GetFinishedUtterance(std::string *uttid, Matrix<BaseFloat> *output_matrix);
```
danpovey: Explain the return status.  Are the utterances going to be returned in the same order they were provided to the class?

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+
+  inline int32 OutputDim() const { return output_dim_; }
+
+  inline bool Ready() const { return batch_info_queue_.size() == minibatch_size_; }
+  
+  // The "very small batch size" or "super long utterance" case will cause the
+  // data is piled up. To prevent it, we lead in this function.
+  // When this function is triggered. We always call Compute() and 
+  // GetFinishedUtterance to reduce the load of the class.
+  inline bool Overload() const { return utt_list_.size() > max_capacity_; }
```
danpovey: I hope there is a way to get rid of the need for this function.  It seems rather ugly.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+
+  NnetSimpleComputationOptions opts_;
+  const Nnet &nnet_;
+  int32 nnet_left_context_;
+  int32 nnet_right_context_;
+  int32 output_dim_;
+  // the log priors (or the empty vector if the priors are not set in the model)
+  CuVector<BaseFloat> log_priors_;
+
+  std::unordered_map<std::string, Matrix<BaseFloat>> feats_;
```
danpovey: You should never store things like Matrix or Vector in an unordered_map directly, you should store pointers to them.
The issue is that any time the size of the map changes, they will have to be copied which is a fairly big operation.  Obviously you'll have to delete the pointers any time you delete a key from the map.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+  std::unordered_map<std::string, Matrix<BaseFloat>> online_ivector_feats_;
+
+  // online_ivector_period_ helps us interpret online_ivector_feats_; it's the
+  // number of frames the rows of ivector_feats are separated by.
+  int32 online_ivector_period_;
+
+  // a reference to a compiler passed in via the constructor, which may be
+  // declared at the top level of the program so that we don't have to recompile
+  // computations each time.
+  CachingOptimizingCompiler &compiler_;
```
danpovey: For google-style-guide reaons, please make this a pointer not a reference.  (It's non-const).

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,170 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+  std::unordered_map<std::string, int32> num_subsampled_frames_;
+
+  // store each utterance id in order. We don't use a queue for here as,
+  // in function "compute()" which is a blocking call, we will keep on 
+  // computing without taking the results out.
+  std::list<std::string> utt_list_;
+
+  // store the information of the current batch. The meaning of the tuple is
+  // (utt_id, first_output_subsampled_index, last_output_frame_index)
+  std::queue<std::tuple<std::string, int32, int32>> batch_info_queue_;
```
danpovey: separate the two '>' characters.  It matters in older compilers.

LvHang:
Now, the code can be compiled.
I use "compare-feature" to compare the result of "nnet3-compute" and "nnet3-compute-batch" with a chain-tdnn model. They are the same.
Nnet3 compilation takes a lot of time. I'm optimizing it.

LvHang:
Yes, I'm modifying the part of preparing ComputationRequest. In fact, we only need one kind of ComputationRequest for speed. So I think the inialization of CachingOptimizingCompiler and CompuationRequest should be completed in constructor.

[The CompuationRequestHasher will call StringHasher and IndexVectorHasher]

left_context = nnet_left_context_ + max(opts_.extra_left_context, opts_.extra_left_context_initial)
right_context = nnet_right_context + max(opts_.extra_right_context, opts_.extra_right_context_final)
The "input" indexes should be [ (n,0,0) , (n, opts_.frames_per_chunk + left_context + right_context, 0 ) )
The "output" indexes should be (n, 0, 0) ~ (n, opts_.frames_per_chunk/opts_.subsampling_factor, 0)
if (nnet_.InputDim("ivector") > 0)
  The "ivector" indexes should be (n, 0, 0)
The range of n is [0, minibatch_size).

src/nnet3/nnet-am-decodable-simple.cc
```diff
@@ -361,6 +361,444 @@ BaseFloat DecodableAmNnetSimpleParallel::LogLikelihood(int32 frame,
...
+  KALDI_ASSERT(input_dim > 0);
+  
+  int32 left_context = std::max(opts_.extra_left_context,
+                                opts_.extra_left_context_initial) +
+                       nnet_left_context_,
+        right_context = std::max(opts_.extra_right_context,
+                                 opts_.extra_right_context_final) +
+                       nnet_right_context_;
+  int32 num_input_rows = left_context + opts_.frames_per_chunk + right_context,
+        num_output_rows = opts_.frames_per_chunk / opts_.frame_subsampling_factor;
```
pegahgh: Is it correct num_output_rows? sometimes, we use use different frames_per_chunk to be multiple of frame_subsampling_factor.

src/nnet3/nnet-am-decodable-simple.cc
```diff
@@ -361,6 +361,444 @@ BaseFloat DecodableAmNnetSimpleParallel::LogLikelihood(int32 frame,
...
+  request.outputs.push_back(IoSpecification("output", output_indexes));
+  computation_ = compiler_.Compile(request);
+
+  num_batch_input_rows_ = num_input_rows;
+  num_batch_output_rows_ = num_output_rows;
+}
+
+
+void BatchNnetComputer::CheckAndFixConfigs() {
+  static bool warned_frames_per_chunk = false;
```
pegahgh: add small description for this param!

src/nnet3/nnet-am-decodable-simple.cc
```diff
@@ -361,6 +361,444 @@ BaseFloat DecodableAmNnetSimpleParallel::LogLikelihood(int32 frame,
...
+  }
+  request.outputs.push_back(IoSpecification("output", output_indexes));
+  computation_ = compiler_.Compile(request);
+
+  num_batch_input_rows_ = num_input_rows;
+  num_batch_output_rows_ = num_output_rows;
+}
+
+
+void BatchNnetComputer::CheckAndFixConfigs() {
```
pegahgh: This function is not specific to Batch decoding. Do we need to reimplement that?

src/nnet3/nnet-am-decodable-simple.cc
```diff
@@ -361,6 +361,444 @@ BaseFloat DecodableAmNnetSimpleParallel::LogLikelihood(int32 frame,
...
+  Matrix<BaseFloat> tot_input(num_batch_input_rows_ * minibatch_size_,
+                              nnet_.InputDim("input"), kSetZero);
+  Matrix<BaseFloat> tot_ivector;
+  int32 ivector_dim = nnet_.InputDim("ivector");
+  if (ivector_dim > 0) {
+    tot_ivector.Resize(minibatch_size_, ivector_dim, kSetZero);
+  }
+
+  // 1.Prepare data
+  for(int32 n=0; !batch_info_queue_.empty(); n++) {
```
pegahgh: wrong indentation n = 0

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,6 +392,161 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
 };
 
 
+/*
+  This class handles the neural net computation in parallel.
+
+  It can accept just input features, or input features plus iVectors.  */
+class BatchNnetComputer {
```
pegahgh: It can be friend of class DecodableNnetSimple to have no need to copy some functions!

src/nnet3/nnet-am-decodable-simple.cc
LvHang: opts_.frames_per_chunk will be rounded up to nearest multiple of opts_.frame_subsampling_factor by function CheckAndFixConfigs().

src/nnet3/nnet-am-decodable-simple.cc
LvHang: I explain it below. See it in the answer of "friend class" comment.

src/nnet3/nnet-am-decodable-simple.cc
LvHang: OK. Actually, I copied this part of code from DecodableNnetSimple, so I didn't modify it.

src/nnet3/nnet-am-decodable-simple.cc
LvHang: Thanks

src/nnet3/nnet-am-decodable-simple.h
LvHang: If I want class BatchNnetComputer use some functions of class DecodableNnetSimple. I should declare "friend class BatchNnetComputer" in class DecodableNnetSimple. But I think DecodableNnetSimple is a footstone of many other classes, such as DecodableAmNnetSimple. So I don't want to modify it actually.
Otherwise, only the function "CheckAndFixConfigs()" is copied. So I didn't do as your comment before.

src/nnet3/nnet-am-decodable-simple.cc
danpovey: and need space after for.   See if misc/maintenance/cpplint.py still works.

LvHang: Ok, I will use cpplint to check all the code.

LvHang:
Update the code.  

Now the code will have 3 different type batches depending on extra context. For details, they are  (extra_left_context, extra_right_context), (extra_left_context, extra_right_context_final)
 and (extra_left_context_initial, extra_right_context). If the utterance length is >= chunk_size, then leave the final chunk in the utterance the normal length, and just make it overlap with the previous chunk.

Otherwise, it will have an option "--ensure-exact-final-context" which controls whether we deal with those "shorter-than-chunk-size" utterances specially. If not, it will be arranged to (extra_left_context_initial, extra_right_context)

Test the code on "tdnn", "tdnn_lstm" and "blstm" models with "nnet3-compute". It always gets the same results. The "final chunk" result will be different in "lstm" and "blstm" models, as we make it overlap with previous chunk. 
If "ensure-exact-final-context == true", the "shorter than chunk size" utterances will have the same results with "nnet3-compute" in each kind of model.
(subsampling_factor =3, extra_left_context_initial/extra_right_context_final <0 or >=0 has been tested.)

Test the speed on eval2000 dataset which has 4458 utterances with GTX 1080 Ti GPU.
For tdnn model,  "nnet3-compute" is 1274.09s, "nnet3-compute-batch" is 414.55s
For tdnn_lstm model, "nnet3-compute" is 1539.87s, "nnet3-compute-batch" is 371.68s
For blstm model, "nnet3-compute" is 4251.25s, "nnet3-compute-batch" is 355.05s

Fix the coding style misc/maintenance/cpplint.py.

francisr:
Are the speed gains only for GPU?

LvHang:
@francisr
I think so. It's designed for GPU accelerate. I haven't do too much tests on CPU.
In my opinion, it doesn't have advantages in accelerating CPU decoding. And preparing batch information will take a little bit additional time.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,8 +395,247 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+  BatchInfoMap batch_info_;
+  BatchInfo last_batch_info_;
+
+  // The key is (tot_left_context, tot_right_context), which would equal the
+  // model left/right context plus the extra left/right context. The value is
+  // corresponding ComputationRequest pointer. It is updated in function
+  // PrepareCompuationRequest().
+  typedef std::unordered_map<std::pair<int32, int32>, ComputationRequest *,
+                             IntPairHasher, IntPairEqual> NormalRqType;
+  NormalRqType RqType_;
```
danpovey: This variable name RqType_ is not style-guide compliant.  Maybe you could call it something like `context_to_request_`.  And I don't like the typename NormalRqType.  Perhaps ComputatationRequestMap would be better.

LvHang:
I did some rough speed tests. That means I didn't add the "Timer" in class, I use it in "nnet3-compute" and "nnet3-compute-batch" binaries.

The first test is conducted on swbd eval2000 dataset which contains 4458 utterances.

|    | compute | batch | batch_exact_final
|-- | ----------| -- | --|
tdnn(gpu) | 946.234 | 245.257 | 189.425
tdnn_lstm(gpu) | 1551.37 | 199.65 | 197.43
blstm(gpu) | 3827.52 | 196.722 | 290.213
tdnn(cpu) | 4063.43 | 5911.52 | 6476.77
tdnn_lstm(cpu) | 21103.9 | 11662.1 | 11773.8
blstm(cpu) | 23470.8 | 6515.24 | 6677.56

For the second test, I select 200 utterances from eval2000, and then run 10 times. After that, get the average.

|  | compute | batch | batch_exact_final|
|-- | -- | -- | --|
tdnn(gpu) | 58.93 | 22.96 | 23.47
tdnn_lstm(gpu) | 81.69 | 17.59 | 16
blstm(gpu) | 227.56 | 24.23 | 26.46
tdnn(cpu) | 127.17 | 177.05 | 173.35
tdnn_lstm(cpu) | 576.09 | 275.36 | 265.2
blstm(cpu) | 759.58 | 230.88 | 237.38

From the tables, we can see that we can get a prominent rise in speed in all kinds of models if we use gpu. For the cpu situtation,  we will get some benifit for lstm and blstm. I'm not sure how it is caused. But for tdnn model, the batch version will be slow. Otherwise, the more the utterances are, the better the batch version gets.

Compare "nnet3-compute" and "nnet3-compute-batch". The aim of them are all get the posterior for each utterance.
"nnet3-compute" use "class DecodableAmNnetSimple", and get the posterior frame by frame with function GetOutputForFrame().
"nnet3-compute-batch" also can complete the assignment to get the posterior for each utterance. It only can get the result of whole utterance with function GetFinishedUtterance(), but it doesn't supply the function to get some frames in an un-finished utterance.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,8 +395,247 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+                    const VectorBase<BaseFloat> &priors,
+                    int32 online_ivector_period = 0,
+                    bool  ensure_exact_final_context = false,
+                    int32 minibatch_size = 128);
+  ~BatchNnetComputer();
+
+  // It takes features as input, and you can either supply a
+  // single iVector input, estimated in batch-mode ('ivector'), or 'online'
+  // iVectors ('online_ivectors' and 'online_ivector_period', or none at all.
+  void AcceptInput(const std::string &utt_id,
```
danpovey: I assume that this object takes ownership of the pointers?  You should clarify.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,8 +395,247 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+
+  NnetSimpleComputationOptions opts_;
+  const Nnet &nnet_;
+  int32 nnet_left_context_;
+  int32 nnet_right_context_;
+  int32 output_dim_;
+  // the log priors (or the empty vector if the priors are not set in the model)
+  CuVector<BaseFloat> log_priors_;
+
+  std::unordered_map<std::string, const Matrix<BaseFloat> * > feats_;
```
danpovey: I believe the default hasher for std::string is very bad, it just uses the length so it would hash them all to one bucket.  You should use StringHasher.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,8 +395,247 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
...
+  // when opts_.frame_subsampling_factor == 1.
+  std::unordered_map<std::string, int32> num_subsampled_frames_;
+
+  std::unordered_map<std::string, bool> is_computed_;
+  // store each utterance id in order. We don't use a queue for here as,
+  // in function "compute()" which is a blocking call, we will keep on
+  // computing without taking the results out.
+  std::list<std::string> utt_list_;
+
+  // The meaning of the tuple is
```
danpovey: You should use a struct for this, it will be more readable.  And you need to explain much more clearly what is the meaning of these members-- not jut what they are used for.

src/nnet3/nnet-am-decodable-simple.h
```diff
@@ -392,8 +395,247 @@ class DecodableAmNnetSimpleParallel: public DecodableInterface {
 };
 
 
+struct IntPairHasher {
```
danpovey: This stuff shouldn't be in nnet-am-decodable-simple.{h,cc}, please move to something like nnet-batch-compute.{h,cc}

LvHang:
@danpovey 
I have modified the codes according to your comments.

danpovey:
thanks!  will look at it to-morrow

danpovey:
This was merged indirectly -> closing.

PR #2483, opened 2018-06-08
Limit arpa unk history

olix20:
Added limit_arpa_unk_history.py to limit unk history of an existing ARPA LM, following the discussion [here](https://groups.google.com/forum/#!topic/kaldi-help/uGiExyOiyjc).

### DIFF
--- a/egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
+++ b/egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
@@ -3,129 +3,155 @@
 # Copyright 2018    Armin Oliya
 # Apache 2.0.
 
-# This script takes an existing ARPA lanugage model and limits the <unk> history 
-# to make it suitable for downstream <unk> modeling. This is for the case when
-# you don't have access to the original text corpus that is used for creating the LM.
-# If you do, you can use pocolm with the option --limit-unk-history=true. This
-# keeps the graph compact after adding the unk model.
+'''
+This script takes an existing ARPA lanugage model and limits the <unk> history
+to make it suitable for downstream <unk> modeling.
+This is for the case when you don't have access
+to the original text corpus that is used for creating the LM.
+If you do, you can use pocolm with the option --limit-unk-history=true.
+This keeps the graph compact after adding the unk model.
+'''
 
 import argparse
-import io 
-import os
-import re 
-import sys 
-import gzip 
+import io
+import re
+import sys
 from collections import defaultdict
 
 
 parser = argparse.ArgumentParser(
-	description = "This script takes an existing ARPA lanugage model and limits the <unk> history to make it suitable for downstream <unk> modeling.\nIt supports up to 5-grams.", 
-	usage = "utils/lang/limit_arpa_unk_history.py <oov_dict_entry> <input_arpa >output_arpa")
+    description='''This script takes an existing ARPA lanugage model
+    and limits the <unk> history to make it suitable
+    for downstream <unk> modeling.
+    It supports up to 5-grams.''',
+    usage='''utils/lang/limit_arpa_unk_history.py
+    <oov-dict-entry> <input-arpa >output-arpa''',
+    epilog='''E.g.: gunzip -c src.arpa.gz |
+    utils/lang/limit_arpa_unk_history.py "<unk>" | gzip -c >dest.arpa.gz''')
 
-parser.add_argument('oov_dict_entry', help='oov identifier, for example <unk>', type=str)
-args = parser.parse_args()  
+parser.add_argument(
+    'oov_dict_entry',
+    help='oov identifier, for example "<unk>"', type=str)
+args = parser.parse_args()
 
 
 def get_ngram_stats(old_lm_lines):
-	ngram_counts = defaultdict(int) 
+    ngram_counts = defaultdict(int)
 
-	for i in range(10):
-		g = re.search(r"ngram (\d)=(\d+)",old_lm_lines[i])
-		if g:
-			ngram_counts[int(g.group(1))] = int(g.group(2))
-	
-	if len(ngram_counts) == 0:
-		sys.exit("Couldn't get counts per ngram section. The input doesn't seem to be a valid ARPA language model.")
+    for i in range(10):
+        g = re.search(r"ngram (\d)=(\d+)", old_lm_lines[i])
+        if g:
+            ngram_counts[int(g.group(1))] = int(g.group(2))
 
-	max_ngrams = list(ngram_counts.keys())[-1]
-	skip_rows = ngram_counts[1] 
+    if len(ngram_counts) == 0:
+        sys.exit("""Couldn't get counts per ngram section.
+            The input doesn't seem to be a valid ARPA language model.""")
 
-	if max_ngrams > 5: 
-		sys.exit("This script supports up to 5-gram language models.")
+    max_ngrams = list(ngram_counts.keys())[-1]
+    skip_rows = ngram_counts[1]
 
-	return max_ngrams, skip_rows, ngram_counts
+    if max_ngrams > 5:
+        sys.exit("This script supports up to 5-gram language models.")
+
+    return max_ngrams, skip_rows, ngram_counts
 
 
-def find_and_replace_unks(old_lm_lines,max_ngrams,skip_rows):
-	ngram_diffs =  defaultdict(int)
-	unk_pattern = re.compile("[0-9.-]+(?:[\s\\t]\S+){1,3}[\s\\t]"+args.oov_dict_entry+"[\s\\t](?!-[0-9]+\.[0-9]+).*")
-	backoff_pattern = re.compile("[0-9.-]+(?:[\s\\t]\S+){1,3}[\s\\t]<unk>[\s\\t]-[0-9]+\.[0-9]+")
-	passed_2grams, last_ngram = False, False
-	unk_row_count, backoff_row_count = 0, 0
+def find_and_replace_unks(old_lm_lines, max_ngrams, skip_rows):
+    ngram_diffs = defaultdict(int)
+    unk_pattern = re.compile(
+        "[0-9.-]+(?:[\s\\t]\S+){1,3}[\s\\t]" + args.oov_dict_entry +
+        "[\s\\t](?!-[0-9]+\.[0-9]+).*")
+    backoff_pattern = re.compile(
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,150 @@
...
+
+    return max_ngrams, skip_rows, ngram_counts
+
+
+def find_n_replace_unks(old_lm_lines,max_ngrams,skip_rows):
+
+    new_lm_lines = old_lm_lines[:skip_rows]
+    ngram_diffs =  defaultdict(int)
+
+    unk_pattern = re.compile("[0-9.-]+(?:[\s\\t]\S+){1,3}[\s\\t]<unk>[\s\\t](?!-[0-9]+\.[0-9]+).*")
```
danpovey: In general we don't assume that the written form of unk is `<unk>`: it may be capitalized , for instance. So it should be an option to the scripts.
And please make it _and_ instead of _n_.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,150 @@
...
+
+            l = old_lm_lines[i].strip()
+
+        
+            if "\{}-grams:".format(3) in l:
+                passed_2grams = True
+            if "\{}-grams:".format(max_ngrams) in l:
+                last_ngram = True
+                
+
```
danpovey: there is too much vertical whitespace in this file.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,150 @@
...
+import argparse
+import codecs 
+import os
+import re 
+from collections import defaultdict
+
+
+
+parser = argparse.ArgumentParser(description='This script takes an existing ARPA lanugage model and limits the <unk> history to make it suitable for downstream <unk> modeling.')
+parser.add_argument('--old_lm',required=True, help='path to the old ARPA language model', type=str)
```
danpovey: I think it would be better to have this program read from stdin and write to stdout, because these files are usually stored gzipped.  You should then make any informational statements be printed to stderr.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,150 @@
...
+    with codecs.open(os.path.expanduser(args.new_lm),"w", encoding="utf-8") as f: 
+        f.writelines(new_lm_lines)
+
+
+
+
+def main(): 
+
+
+    with codecs.open(os.path.expanduser(args.old_lm),"r", encoding="utf-8") as f:
```
danpovey: please don't read as utf-8: instead, read as latin-1.  It's more compatible with other encodings e.g. GBK.

jtrmal:
I'm a bit disappointed it's a python2 script, especially for something that will be probably staying with us.

olix20:
thanks, i tried to include all your comments. 
Not sure abotu latin-1 though .. when i include it i get unicode errors, but it goes fine without specifying an encoding.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python
```
danpovey: please change this to python3 and make sure it still works.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,144 @@
...
+    sys.stderr.write("Removed {} lines including {} as second-to-last term\n".format(unk_row_count, args.oov_dict_entry))
+    sys.stderr.write("Removed backoff probabilties from {} lines\n".format(backoff_row_count))
+
+    return new_lm_lines, ngram_diffs
+
+
+def read_old_lm():
+    sys.stderr.write("Reading ARPA LM frome input stream .. \n")
+    
+    if args.compressed:
```
danpovey: if you are going to explicitly gzip/unzip things it actually defeats the purpose of using stdin/stdout.  You should do one or the other.  The point was that you might want to do
`gunzip -c src.gz | your_program.py | gzip -c > dest.gz`

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,144 @@
...
+        input_stream = gzip.open(sys.stdin.buffer, 'rt')
+    else:
+        input_stream = io.TextIOWrapper(sys.stdin.buffer) #encoding="latin-1"
```
danpovey: Your problem with the latin-1 may be that you are not doing the encoding to both input and output streams.
Please do it and if it doesn't work, let me know specifically what the error is.

olix20:
got it, tested with python3 shebang and latin-1 encoding for both stdin/out.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
...
+import sys 
+import gzip 
+from collections import defaultdict
+
+
+parser = argparse.ArgumentParser(
+	description = "This script takes an existing ARPA lanugage model and limits the <unk> history to make it suitable for downstream <unk> modeling.\nIt supports up to 5-grams.", 
+	usage = "utils/lang/limit_arpa_unk_history.py <oov_dict_entry> <input_arpa >output_arpa")
+
+parser.add_argument('oov_dict_entry', help='oov identifier, for example <unk>', type=str)
```
danpovey: I think for stylistic consistency I'd prefer to have this be 'oov-dict-entry'.  (The variable itself will still be oov_dict_entry).
And please have an example line, I think you can use the "epilog" for that.
For example:
`  epilog="E.g.: gunzip -c src.arpa.gz | utils/lang/limit_arpa_unk_history.py '<unk>' | gzip -c >dest.arpa.gz"`

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
...
+import io 
+import os
+import re 
+import sys 
+import gzip 
+from collections import defaultdict
+
+
+parser = argparse.ArgumentParser(
+	description = "This script takes an existing ARPA lanugage model and limits the <unk> history to make it suitable for downstream <unk> modeling.\nIt supports up to 5-grams.",
```
danpovey: please split up this string over a couple lines.  You can use triple quotes.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python3
+
+# Copyright 2018    Armin Oliya
+# Apache 2.0.
+
+# This script takes an existing ARPA lanugage model and limits the <unk> history
```
danpovey: I think it's more standard to put this in triple-quotes as a doc-string, instead of as a comment.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
...
+	sys.stderr.write("Removed {} lines including {} as second-to-last term.\n".format(unk_row_count, args.oov_dict_entry))
+	sys.stderr.write("Removed backoff probabilties from {} lines.\n".format(backoff_row_count))
+
+	return new_lm_lines, ngram_diffs
+
+
+def read_old_lm():
+	sys.stderr.write("Reading ARPA LM frome input stream .. \n")
+
+	with io.TextIOWrapper(sys.stdin.buffer,encoding="latin-1") as input_stream:
```
danpovey: spce after comma please.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
...
+				
+			new_lm_lines.append(l+"\n")
+
+	sys.stderr.write("Removed {} lines including {} as second-to-last term.\n".format(unk_row_count, args.oov_dict_entry))
+	sys.stderr.write("Removed backoff probabilties from {} lines.\n".format(backoff_row_count))
+
+	return new_lm_lines, ngram_diffs
+
+
+def read_old_lm():
```
danpovey: you seem to be using an indent level of 8.  Can you please use a level of 4?  That's what we use elsewhere.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
...
+				if g_backoff:
+					updated_row = g_backoff.group(0).split()[:-1]
+					updated_row = updated_row[0]+"\t"+(" ".join(updated_row[1:]))+"\n"
+					new_lm_lines.append(updated_row)
+					backoff_row_count += 1
+					continue
+				
+			new_lm_lines.append(l+"\n")
+
+	sys.stderr.write("Removed {} lines including {} as second-to-last term.\n".format(unk_row_count, args.oov_dict_entry))
```
danpovey: please try to keep the line length below about 100: you can split strings if needed.  Look at other code e.g. in steps/libs/nnet3/, for examples of style.

egs/wsj/s5/utils/lang/limit_arpa_unk_history.py
```diff
@@ -0,0 +1,131 @@
...
+	sys.stderr.write("Removed backoff probabilties from {} lines.\n".format(backoff_row_count))
+
+	return new_lm_lines, ngram_diffs
+
+
+def read_old_lm():
+	sys.stderr.write("Reading ARPA LM frome input stream .. \n")
```
danpovey: I think it's more standard to use
   `print("....", file=sys.stderr)`
and in that case you don't need the \n at the end.

jtrmal:
You can also run through flake8 (http://flake8.pycqa.org/en/latest/)
That would help you to get rid of most of the style issues
Also, please use 2-space to indent (definitely not tabs)
y.

danpovey:
Yenda, I'm looking at the rest of our python code, and it seems 4 spaces is
the standard.  (We use 2 for bash and C++ though).

jtrmal:
all right, my bad, sorry -- I was more concerned about the fact that it
seemed that the tabs are used for indentation and didn't check what are we
using.
y.

olix20:
Thanks, i think i covered all your style comments, all is clear with flake8.

PR #2487, opened 2018-06-10
[src] Make it so very wide-context nnet3 models don't cause a crash.

danpovey:


danpovey:
I notice the tests are failing in an interesting way now: in the "make test" here, we're getting "illegal instructions" in some tests.  I'm wondering whether the travis build runs on heterogeneous hardware.  Anyway, it has nothing to do with the changes in this PR so I'm merging it anyway.

Issue #2484, opened 2018-06-09
CUDA 9.2 supports GCC 7


leomao:
CUDA 9.2 is now listing gcc-7.3.1 in its supported compiler.
Should add following into `src/configure`.

```
        9_2)
          MIN_UNSUPPORTED_GCC_VER="8.0"
          MIN_UNSUPPORTED_GCC_VER_NUM=80000;
        ;;
```

danpovey:
Sure, that makes sense.

leomao:
0f954c27228436a088417cd1d7b1557357f012ec miss `;;` at the end.

jtrmal:
resolved, closing

PR #2488, opened 2018-06-10
FIX: add missing semicolons in configure

nanma:
The `configure` script in `src` directory fails due to missing semicolons. I try to fix this small issue. Hope it will help.


danpovey:
Both #2486 and this commit resolve #2484.

Incidentally, this highlights that our travis build needs to be fixed.  It seems I ignored the error because it errors so frequently.  @galv, do you have time to look at it?

PR #2494, opened 2018-06-11
[egs] updating results in voxceleb recipe

david-ryan-snyder:
Updating the voxceleb recipe results after applying @weichcai's bug fix in https://github.com/kaldi-asr/kaldi/pull/2475 .

I reran the v1 and v2 recipes from scratch. The i-vector results get a little worse, and the x-vector results get a little better. Probably just noise since the evaluation dataset is small.


david-ryan-snyder:
@danpovey, please merger when you get the chance.

PR #2504, opened 2018-06-15
small improvement over pruned lattice-rescoring algorithm

hainan-xv:
This was first tested on the material recipe, and also on swbd. It was giving the almost identical performance as before but faster.

### DIFF
--- a/src/lat/compose-lattice-pruned.cc
+++ b/src/lat/compose-lattice-pruned.cc
@@ -159,7 +159,8 @@ class PrunedCompactLatticeComposer {
     int32 lat_state;
     int32 lm_state;
 
-    // the distance between the state from the start state
+    // the number of arcs on the path from the start state to this state, in the
+    // composed lattice, by which this state was first reached.
     int32 depth;
 
     // If you have just called RecomputePruningInfo(), then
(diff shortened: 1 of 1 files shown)


src/lat/compose-lattice-pruned.cc
```diff
@@ -159,6 +159,9 @@ class PrunedCompactLatticeComposer {
     int32 lat_state;
     int32 lm_state;
 
+    // the distance between the state from the start state
+    int32 depth;
```
danpovey: it might be more accurate to describe this as:
```
the number of arcs on the path from the start state to this state, in the composed
lattice, by which this state was first reached.
```

src/lat/compose-lattice-pruned.cc
```diff
@@ -292,6 +295,18 @@ class PrunedCompactLatticeComposer {
   // it always follows an automatic schedule based on the num-arcs of the output
   // lattice.
   bool output_reached_final_;
+
+  // This variable, which we set initially to -1000, makes sure that in the
+  // beginning of the algorithm, we always prioritize exploring the lattice
+  // in a depth-first way. Once we find a path reaching a final state, this
+  // variable will be reset to 0.
+  // The reason we do this is because the beam-search depends on a good estimate
+  // of the composed-best-cost, which before we reach a fina state, we instead
```
danpovey: final

PR #2511, opened 2018-06-18
disable -march=native in travis build

jtrmal:
-march=native seems badly interacting with ccache (which kinda makes sense)
the other changes is just to trigger rebuild (removal of tabs in cc sources0


jtrmal:
@danpovey  this should resolve the travis compilation issues

PR #2514, opened 2018-06-19
Change smart ptr copy to move sematics

Lallapallooza:


danpovey:
Thanks, but I'm trying to keep the code simple and obvious to the un-initiated.
Since the extra work here is very tiny (maybe an increment and decrement of a counter and a copy of a pointer or two), I'm not inclined to merge this. 
Thanks though.

vdp:
Dan, as far as I understand the cost is not so much due to counter inc/dec, but because of the thread-safety requirement for the shared_ptrs (https://stackoverflow.com/a/41874953). IIRC the compilation is not called in a super tight loop, though, so as you say it probably doesn't matter here.

danpovey:
oh, OK.  Worth merging then, I guess.

PR #2521, opened 2018-06-26
added chime-5 evaluation parts

sw005320:
- [x] change directory path from test one (https://github.com/kaldi-asr/kaldi/compare/master...sw005320:chime5_eval?expand=1#diff-a94243e58c145345e08267e902b2101cR27) to general one 
- [x] update RESULTS

### DIFF
--- a/egs/chime5/s5/local/score_for_submit.sh
+++ b/egs/chime5/s5/local/score_for_submit.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+# Copyright 2012-2014  Johns Hopkins University (Author: Daniel Povey, Yenda Trmal)
+# Apache 2.0
+
+cmd=run.pl
+dev=exp/chain_train_worn_u100k_cleaned/tdnn1a_sp/decode_dev_beamformit_ref
+eval=exp/chain_train_worn_u100k_cleaned/tdnn1a_sp/decode_eval_beamformit_ref
+do_eval=false
+
+echo "$0 $@"  # Print the command line for logging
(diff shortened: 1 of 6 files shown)


egs/chime5/s5/local/score_for_submit.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+# Copyright 2012-2014  Johns Hopkins University (Author: Daniel Povey, Yenda Trmal)
+# Apache 2.0
+
```
jtrmal: you could add explanation what the script does (probably also to the help message)

egs/chime5/s5/run.sh
```diff
@@ -118,12 +112,33 @@ if [ $stage -le 5 ]; then
...
+  done
+
   # Split speakers up into 3-minute chunks.  This doesn't hurt adaptation, and
   # lets us use more jobs for decoding etc.
-  for dset in ${train_set} ${test_sets}; do
+  for dset in ${train_set}; do
     utils/copy_data_dir.sh data/${dset} data/${dset}_nosplit
     utils/data/modify_speaker_info.sh --seconds-per-spk-max 180 data/${dset}_nosplit data/${dset}
   done
+  for dset in ${test_sets}; do
```
jtrmal: this will probably be a problem when you decide to have per-speaker stats (not an issue per se).

jtrmal:
@danpovey I think this should be ready to merge (reviewed today).

PR #2522, opened 2018-06-26
Small fix on scripts of Aishell2

underdogliu:
This PR includes some slight modifications on Aishell2 scripts, including:

- Add a mode option, which decides whether to train simple model with 40-dim mfcc or 'normal' model with 43-dim pitch-added mfcc, i-vector and dropout.

- Some minor comment fix.

@danpovey This PR has been primarily checked by @dophist although I believe you may have more comments before merging. Thanks for checking!


egs/aishell2/s5/run.sh
```diff
@@ -20,26 +20,35 @@ tst_set=
...
+  local/run_gmm.sh --nj $nj --stage $gmm_stage --mode $mode
 fi
 
 # nnet3 & chain
+# so far _1d is the SoTA chain recipe for normal use. If better model get
+# trained & tested it will hopefully be manually updated in later PRs
 if [ $stage -le 3 ]; then
-  # local/nnet3/run_tdnn.sh --nj $nj
-  local/chain/run_tdnn.sh --nj $nj
+  [ $mode == "simple" ] && local/chain/run_tdnn.sh --nj $nj --stage 5 || \
```
danpovey: This should be an if-statement.
But also-- you're really not supposed to go into the tuning directory like that.  If there are going to be two sequences of training scripts, one for simple, one for "normal", there should be two soft links, pointing to two sequences of  numbered scripts.   But this gets very complicated.

I general, I'm not really excited about maintaining two different versions of this recipe, especially if mixed up like this.  I'd much rather have just one version-- it's easier to maintain.  Maybe you can just mention the older script and say that at that time we weren't using pitch and ivectors, and the difference is such-and-such?

Also, since you are changing the recipe in this PR, why aren't the results changing?  Were they previously run with these changes, but the changes just weren't checked in?

underdogliu: Thanks for your reply! w.r.t your questions:

1) Thanks for pointing out and I'll make fixes abut if and soft-linking with chatting with @dophist ;

2) We are doing this since it depends on actual needs from the user, in terms of whether they wanna use pitch+ivec+dropout or not. So we wanna keep both versions even though _1d does have better results. @dophist perhaps has more comments to tell;

3) We are not maintaining a 'RESULTS' file so there is no global results about GMM model decoding results. If one wanna run _1a, he/she should run in 'simple' mode and the result of chain is in the script local/chain/tuning/run_tdnn_1a.sh; correspondingly, if one wanna run _1d, he/she should use pitch feats from the beginning of mono training ('normal' mode). So the results in the chain scripts are not changing because they are using diff. type of feats from the beginning (i.e. they previously run with these changes and the changes are actually checking in w.r.t the mode option).

Hope it helps without any potentially offensive statement. If so please point it out. Thanks~

danpovey:
>
>    1.
>
>    Thanks for pointing out and I'll make fixes abut if and soft-linking
>    with chatting with @dophist <https://github.com/dophist> ;
>    2.
>
>    We are doing this since it depends on actual needs from the user, in
>    terms of whether they wanna use pitch+ivec+dropout or not. So we wanna keep
>    both versions even though _1d does have better results. @dophist
>    <https://github.com/dophist> perhaps has more comments to tell;
>    3.
>
>    We are not maintaining a 'RESULTS' file so there is no global results
>    about GMM model decoding results. If one wanna run _1a, he/she should run
>    in 'simple' mode and the result of chain is in the script
>    local/chain/tuning/run_tdnn_1a.sh; correspondingly, if one wanna run
>    _1d, he/she should use pitch feats from the beginning of mono training
>    ('normal' mode). So the results in the chain scripts are not changing
>    because they are using diff. type of feats from the beginning (i.e. they
>    previously run with these changes and the changes are actually checking in
>    w.r.t the mode option).
>
>
OK, then please add a RESULTS file.  It's important for debugging and
making sure that someone else's system build  is working.

The current setup in the tuning/ directory is a bit of a mess, and it's not
properly explained.  For instance, it jumpts from 1a to 1d-- it should be
consecutive.  And the script in 1d says how it compares to the non-existent
script 1c.  It's not explained anywhere that the 1d was run with pitch but
the 1c was not.

You should probably rename those scripts and add proper explanations.

underdogliu:
@danpovey In my commits in this PR I made some fixes (like, proper explanations for _1d) but got it and will do more. Thanks.

egs/aishell2/s5/local/run_gmm.sh
```diff
@@ -18,13 +19,22 @@ stage=1
 dev_nj=$(wc -l data/dev/utt2spk | awk '${print $1}' || exit 1;)
 test_nj=$(wc -l data/test/utt2spk | awk '${print $1}' || exit 1;)
 
-
-# Now make MFCC plus pitch features.
+# Now make MFCC features.
+# For normal use where academic=false, we produce mfcc feats;
```
danpovey: this comment may be out of date

egs/aishell2/s5/local/chain/tuning/run_tdnn_1d.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# _1d is as _1c, but with dropout schedule added, referenced from wsj
+# _1d is as _1a, but with i-vector and dropout schedule added, referenced from wsj
```
danpovey: OK, I am assuming this is not the entire difference because this one has pitch too.

danpovey:
can you please rename 1d to 1b?  Also, please make clear at the top of each chain script, which 'mode' it was run with, i.e. whether it was done in 'normal' or 'simple' mode.  Make sure that local/chain/compare_wer.sh prints out the number of parameters, and also please include the output chain_dir_info.pl in the comments at the top.  (That will also help clarify what the feature type was).

egs/aishell2/s5/local/run_gmm.sh
```diff
@@ -9,22 +9,32 @@ set -euxo pipefail
...
+mode=normal        # Please refer to ../run.sh for comments. should be either 'normal' or 'simple'
 
 . ./cmd.sh
 [ -f ./path.sh ] && . ./path.sh;
 . ./utils/parse_options.sh
 
 # nj for dev and test
-dev_nj=$(wc -l data/dev/utt2spk | awk '${print $1}' || exit 1;)
-test_nj=$(wc -l data/test/utt2spk | awk '${print $1}' || exit 1;)
+dev_nj=$(wc -l data/dev/spk2utt | awk '${print $1}' || exit 1;)
```
dophist: @underdogliu hi, Xuechen, still syntax error in awk command

egs/aishell2/s5/local/chain/tuning/run_tdnn_1b.sh
```diff
@@ -1,24 +1,30 @@
 #!/bin/bash
 
-# _1d is as _1c, but with dropout schedule added, referenced from wsj
+# _1b is as _1a, but with pitch feats, i-vector and dropout schedule added, referenced from wsj
+# this script is for 'normal' mode
```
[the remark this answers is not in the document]
underdogliu: it would only run correctly in 'normal' mode due to feature dimension. I think it has been indicated in run.sh, where setting different modes leads to different scripts. But of course I can make changes to make it potentially simpler.

danpovey:
That script works out the feature dimension from the features-- it looks to
me like it would work.

underdogliu:
For alignments and question tree generation there may be some dimension mismatch. For example when doing alignment conversion, if the number of columns of final.mat from ali_dir is something like, for example, 144=16x9 (this is with pitch) while the dimension of input features is 13, then it would fail. But there is a possibility that I got the wrong memory from experience. Also, even if they is not, it's still good practice from my point of view to keep the dimension consistent. But if it's not please point out sharply and I'd be happy to correct my mistakes in the script.

danpovey:
The only thing I see in the script that wouldn't be comptible with the
"simple" setting is that it does steps/make_mfcc_pitch.sh instead of
steps/make_mfcc.sh.  But that shouldn't cause a crash, it's just that it
isn't really compatible with your intentions in the "simple" setting.

Anyway, I suppose my main problem is as follows: if your training script is
going to have the "simple" option, that option should actually work.
Currently it wouldn't work, because local/chain/run_tdnn.sh points to a
script which does use pitch.  IMO there are two reasonable possibilities:

 (1) Take out the "simple" option
 (2) Split the scripts into local/chain_simple/ and local/chain/ (the
latter for the 'normal' setup), with separate tuning sequences (e.g. 1a in
both).

And obviously, in the "simple" case, the script would call the version in
local/chain_simple/.

Dan

danpovey:
Regarding --stage 5: those scripts, if run without the --stage option,
should always run from the beginning.  Since that was the first nnet3 or
chain script in the run.sh, it should run from the beginning when invoked
there, so it shouldn't require the --stage option.

underdogliu:
ah I got what you meant. Fixed.

danpovey:
Thanks!  Merging.

PR #2528, opened 2018-07-02
Add a check for utterance length when creating end2end egs

hhadian:


danpovey:
I assume this is tested?

hhadian:
Yes, I tested this on IAM (where there was no bad utterance) and some artificial bad utterances to trigger the warning.

PR #2531, opened 2018-07-03
Reco2dur fix

pswietojanski:
I believe this is a fix to this PR: https://github.com/kaldi-asr/kaldi/pull/2414/files

It originally added reco2dur bits to speed perturbation scripts (so perturb_data_dir_speed_3way.sh takes care of that, and then reverberation scripts do not need to recompute durations at the recording level). 

The issue is, that although perturb_data_dir_speed_3way.sh computes reco2dur for the source folder (if it's missing), perturb_data_dir_speed.sh is not modifying it for speed perturbed variants. As a result reco2dur does not exists in speed perturbed folders, and combine_data.sh later omits it when creating _sp data dir. Which is a problem for reverberate_data_dir.py later (as it tries to recompute it from scratch).


danpovey:
Looks plausible.  It's been tested somehow?

pswietojanski:
yes, I am doing some large scale experiments on multi_en, and this issue was very evident. I did test the fix today and seem to work as expected.

PR #2538, opened 2018-07-07
add compute_sentence_probs_arpa.py

DongjiGao:
Adding compute_sentence_probs_arpa.py that can compute sentence log probability of each sentence from the input file given an arpa language model and write the log prob into an output file.
This script has been checked with SRILM for 10k sentences.
Can also support stdin and stdout.

### DIFF
--- a/egs/wsj/s5/utils/lang/compute_sentence_probs_arpa.py
+++ b/egs/wsj/s5/utils/lang/compute_sentence_probs_arpa.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python
+
+# Dongji Gao
+
+# We're using python 3.x style but want it to work in python 2.x
+
+from __future__ import print_function
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(description="This script evaluates the lob probabilty (10 based) of each sentence "
+                                             "from data (in text form), given a language model in arpa form"
+                                             "and a specific ngram order.",
+                                 epilog="e.g. ./compute_sentence_probs_arpa.py ARPA_LM NGRAM_ORDER TEXT_IN PROB_FILE",
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument("arpa_lm", type=str,
+                    help="Input language model in arpa form.")
+parser.add_argument("ngram_order", type=int,
+                    help="Order of ngram")
+parser.add_argument("text_in", type=str,
+                    help="Filename of input text file (each line will be interpreted as a sentence).")
+parser.add_argument("prob_file", type=str,
+                    help="Filename of output probability file.")
+args = parser.parse_args()
+
+def check_args(args):
+    args.text_in_handle = sys.stdin if args.text_in == "-" else open(args.text_in, "r")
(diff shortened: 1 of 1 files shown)


xiaohui-zhang:
looks good and I've tested it.

danpovey:
has it been tested in python2 and python3, and compared with SRILM output?

DongjiGao:
Yes. The only difference between python2 and 3 is the precision:
e.g -36.599961699999994 in python3 and -36.6 in python2

xiaohui-zhang:
ready to merge?

egs/wsj/s5/utils/lang/compute_sentence_probs_arpa.py
```diff
@@ -0,0 +1,150 @@
...
+
+# Dongji Gao
+
+# We're using python 3.x style but want it to work in python 2.x
+
+from __future__ import print_function
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(description="This script evaluates the lob probabilty (10 based) of each sentence "
```
danpovey: This should be 'log probability'.
And wouldn't it be more convenient for it to default to outputting log base e (like Kaldi uses) and maye we can add an option to output log base 10 if it's needed in the future?

DongjiGao: The log base in arpa is 10, so I directly used it. I will fix it and add a log base option.

egs/wsj/s5/utils/lang/compute_sentence_probs_arpa.py
```diff
@@ -0,0 +1,150 @@
...
+    ngram_dict, tot_num = load_model(args.arpa_lm)
+
+    num_valid, max_ngram_order = check_number(args.arpa_lm, tot_num)
+    if not num_valid:
+        sys.exit("get_sentence_prob.py: Wrong loading model.")
+    if args.ngram_order <= 0 or args.ngram_order > max_ngram_order:
+        sys.exit("get_sentence_prob.py: " +   
+            "Invalid ngram_order (either negative or greater than maximum ngram number ({}) allowed").format(max_ngram_order)
+
+    output_result(args.text_in_handle, args.prob_file_handle, args.ngram_order)
```
pengyulong: args.text_in_handle is 'str' type，may be text_in filename,but in line 129 :`lines = text_in_handle.readlines()`, this code must be wrong!

DongjiGao:
ready to merge?

egs/wsj/s5/utils/lang/compute_sentence_probs_arpa.py
```diff
@@ -0,0 +1,156 @@
...
+    return logprob
+        
+
+def output_result(text_in_handle, output_file_handle, ngram_order):
+    index = 1
+    lines = text_in_handle.readlines()
+    for line in lines:
+        new_line = "<s> " + line[:-1] + " </s>"
+        logprob = compute_sentence_prob(new_line, ngram_order)
+        new_logprob = logprob * math.log(10, args.log_base)
```
danpovey: it would be better to evaluate `math.log(10, args.log_base)` outside the loop.

PR #2544, opened 2018-07-11
[scripts] Fix to get_reco2dur.sh (thanks: Zhangfeifan)

danpovey:


danpovey:
@jtrmal, maybe this fixes the problem you noticed recently?
I'll merge this right away as it was obviously a bug.

jtrmal:
I think @johnjosephmorgan  reported something in this direction -- I just re-opened the bug, but I cannot find it now. Strange.

johnjosephmorgan:
I pulled the latest kaldi version and the problem was fixed. I'm not sure when the fix happened.
Thanks,
J

PR #2546, opened 2018-07-12
remove extra comma produced by bsd seq in macos

LiChenda:
In macos, the bsd version "seq" produce extra comma in `$(seq -s',' $nj)`, which makes `$auto_vocab_splits` wrong.
remove extra comma.


danpovey:
Thanks; can you please search using some kind of 'find' command for other instances where we are using the "-s" option to seq?  
It's annoying that the unixes are this incompatible.

LiChenda:
@danpovey Thank you. It's done. Here is a new commit.

PR #2547, opened 2018-07-12
[scripts] Fix an issue in BPE-related scripts

hhadian:
The only real fix is to use the integer id of the sil phone in `add_final_optional_silence.sh`
The rest of the changes are cosmetic.


hhadian:
I think this is ready for merge.

PR #2548, opened 2018-07-13
fix phonetisaurus install issue when python links to python3

jtrmal:
I'm not able to test if phonetisaurus will work with python3, though

### DIFF
--- a/tools/extras/install_phonetisaurus.sh
+++ b/tools/extras/install_phonetisaurus.sh
@@ -21,7 +21,7 @@ if ! $(python -c "import distutils.sysconfig" &> /dev/null); then
     echo "Proceeding with installation." >&2
 else
   # get include path for this python version
-  INCLUDE_PY=$(python -c "from distutils import sysconfig as s; print s.get_config_vars()['INCLUDEPY']")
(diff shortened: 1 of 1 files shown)


AdolfVonKleist:
The bindings do not explicitly support python3.  The apply and train scripts have a couple of unicode-escape regexes (`ur"someregex"`), and the `raise` invocations use the old print syntax.   Those need to be fixed before those scripts will run under python3 without complaint.

danpovey:
Do you think the internal parts of phonetisaurus even support python3?
Might be easier just to die saying it won't work, than les the install
script support it?

It was always a stupid policy, to make 'python' point to python3.

AdolfVonKleist:
The python wrapper scripts are all written in 2.7.  The only step towards 3 compatibility is use of the `print_function` from `__future__`.  The wrappers [but not the bindings] are just calling the c++ binaries.  I don't know how important and upgrade to python3 is/would be.

danpovey:
One possibility is to update the scripts to invoke python as 'python2.7'.
We already require that python2.7 be installed and be accessible by typing
'python2.7'.

danpovey:
I think we ended up fixing this a different way.

PR #2549, opened 2018-07-14
[src] Change decodable code so nnet context does not have to be recomputed.

danpovey:


galv:
LGTM

danpovey:
Thanks!

I'm waiting for Ashish to test in a setup where it was a problem; when
that's done I'll merge.

PR #2554, opened 2018-07-17
remove redundant check

gaob13:
already checked at l.65.


jtrmal:
Can you make the L.65 message more informative, similarly as the message being removed is?

PR #2562, opened 2018-07-21
new tool fsts-concat, similar to fsts-union...

vesis84:
- concatenates all vector-fsts with the same 'key' from the sorted
  'rspecifier' in the order as they appear in I/O,
- handy for concatenating HCLG graphs: 'phn-loop . some_word . phn-loop',

### DIFF
--- a/src/fstbin/fsts-concat.cc
+++ b/src/fstbin/fsts-concat.cc
@@ -1,0 +1,112 @@
+// fstbin/fsts-concat.cc
+
+// Copyright 2016  Johns Hopkins University (Authors: Jan "Yenda" Trmal)
+//           2018  Soapbox Labs (Author: Karel Vesely)
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "fstext/fstext-utils.h"
+#include "fstext/kaldi-fst-io.h"
+
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    using namespace fst;
+    typedef kaldi::int32 int32;
+    typedef kaldi::uint64 uint64;
+
+    const char *usage =
+        "Reads kaldi archives with FSTs. Concatenates the fsts from all the rspecifiers.\n"
+        "The fsts to concatenate must have same key. The sequencing is given by the position of arguments.\n"
+        "\n"
+        "Usage: fsts-concat [options] <fsts-rspecifier1> <fsts-rspecifier2> ... <fsts-wspecifier>\n"
+        " e.g.: fsts-concat scp:fsts1.scp scp:fsts2.scp ... ark:fsts_out.ark\n"
+        "\n"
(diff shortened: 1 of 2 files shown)


danpovey:
I don't think I want to merge this with its current design.
The reason is that I'm not crazy about the way it concatenates all FSTs with the same key in a single input.
Typically if you want to concatenate things, you'd have several pipelines producing different types of thing that you'd then want to concatenate.  The most natural usage would be something like

  `fsts-concat 'ark:command1|' 'ark:command2|' 'ark:command3|' ... 'ark:-' | ...`

in my opinion, which would require different code.
But if you don't have the time to change it, it's OK; we can just wait till it's needed for something that needs to be checked in.

vesis84:
Well, I was thinking of what you just proposed, but then I decided to use the original API that was used for 'fsts-union' (I understand the difference between union and concatenation)...
And then I realized it might be less flexible than the "current" solution. Imagine a situation in which one would like to concatenate different number of FSTs for individual keys... This is easily doable with the scps, but not with the arks...

The "current" version is easy to use by preparing the 'scp' from any script language. Typically, some fsts will be "change" on per-key basis and some will be "fixed".

To check this in, is it absolutely necessary to have some calling script? (I am developing this for a company, and I am not sure how this intellectual property things work...)
K.

danpovey:
I think the version as you have it will create a lot of scripting headaches
in the most likely use-cases, if we commit it as-is, that's why I don't
plan to merge it with that command-line interface.

It's not 100% necessary to have a calling script.

vesis84:
Okay, in this case, I will change the interface, and I'll send an update when it's done... Thank you, Karel.

vesis84:
Hi, the API was changed as Dan suggested. (and rebased on top of the 'main master'...)

src/fstbin/fsts-concat.cc
```diff
@@ -0,0 +1,111 @@
...
+          n_skipped = 0;
+
+    for (; !fst_reader.Done(); fst_reader.Next()) {
+      std::string key = fst_reader.Key();
+
+      // check that the key exists in all 'fst_readers',
+      bool skip_key = false;
+      for (int32 i=0; i<fst_readers.size(); i++) {
+        if (!fst_readers[i]->HasKey(key)) {
+          KALDI_WARN << "Skippng '" << key << "'"
```
danpovey: typo skipping

src/fstbin/fsts-concat.cc
```diff
@@ -0,0 +1,111 @@
...
+            << " due to missing the fst in " << i+2 << ". <rspecifier> : "
```
danpovey: please add parens around i+2, just to avoid doubt about precedence; and maybe replace . with `'th`.

src/fstbin/fsts-concat.cc
```diff
@@ -0,0 +1,111 @@
...
+
+    int32 n_done = 0,
+          n_skipped = 0;
+
+    for (; !fst_reader.Done(); fst_reader.Next()) {
+      std::string key = fst_reader.Key();
+
+      // check that the key exists in all 'fst_readers',
+      bool skip_key = false;
+      for (int32 i=0; i<fst_readers.size(); i++) {
```
danpovey: some compilers warn about signed/unsigned comparisons... I think it would be helpful to have a variable
  `int32 fst_readers_size = (int32)fst_readers.size();`

src/fstbin/fsts-concat.cc
```diff
@@ -0,0 +1,111 @@
...
+      }
+      // finally, prepend the fst from the 'Sequential' reader,
+      fst::Concat(fst_reader.Value(), &fst_out);
+
+      // write the output,
+      fst_writer.Write(key, fst_out);
+      n_done++;
+    }
+
+    // cleanup,
```
danpovey: please start comments with capitals and end with period, not comma.  It's google style guide... sorry to be picky.

src/fstbin/fsts-concat.cc
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    std::string fsts_rspecifier = po.GetArg(1),
+        fsts_wspecifier = po.GetArg(po.NumArgs());
+
+    SequentialTableReader<VectorFstHolder> fst_reader(fsts_rspecifier);
+    std::vector<RandomAccessTableReader<VectorFstHolder>*> fst_readers;
+    TableWriter<VectorFstHolder> fst_writer(fsts_wspecifier);
+
+    for (int32 i=2; i<po.NumArgs(); i++)
```
danpovey: please add space around operators `=` and `<` (also later on in this file).

vesis84:
okay, everything is incorporated... (and rebased...)

PR #2573, opened 2018-07-25
[scripts] Fix to analyze_alignments/lats.sh (double-counting)

eginhard:
If a single phoneme is aligned to the whole utterance, it is counted both as
`begin` and `end`, but is added to the total only once. This caused
`assert count >= 0` in `analyze_phone_length_stats.py` to fail (`count = all - begin - end`).

I'm not sure what the ideal behavior is, but in this change only the `begin` is counted in such cases.


danpovey:
OK.  It seems to mess with the expected meanings of the counts, but I can't see a better way to resolve the issue that's easy, so I'll merge this.

PR #2575, opened 2018-07-26
[src,build] Refactor CuDevice code for compatibility with GPU decoding.

danpovey:


src/cudamatrix/cu-kernels-ansi.h
```diff
@@ -789,6 +789,9 @@ void cuda_uncompress_uint8(dim3 Gr, dim3 Bl, BaseFloat *dest,
                           MatrixDim dim, const uint8_t *src,
                           int src_stride, float scale);
 
+// Launches a kernel that does nothing, explicitly using the legacy default stream;
+// this will synchronize all threads without blocking.
+void cuda_legacy_noop();
```
galv: I'm sure you're aware already, but this is unused. 

I would consider renaming it to something like "cuda_synchronize_all_devices".

Consider adding a `cudaGetLastError()` to the end of the function definition. Up to you.

It also looks like you could do a for loop with `CuDevice::Instantiate::SelectGpuId(id)` and [cudaDeviceSynchronize](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g10e20b05a95f638a4071a655503df25d).

src/cudamatrix/cu-allocator.cc
```diff
@@ -123,14 +127,68 @@ void* CuMemoryAllocator::MallocPitchInternal(size_t row_bytes,
   return ans;
 }
 
+
+std::string GetFreeGpuMemory(int64* free, int64* total) {
+#ifdef _MSC_VER
+  size_t mem_free, mem_total;
+  cuMemGetInfo_v2(&mem_free, &mem_total);
+#else
+#if (CUDA_VERSION >= 3020)
```
galv: Perhaps I am missing something, but why are we supporting < cuda 3.2?

src/cudamatrix/cu-allocator.cc
```diff
@@ -123,14 +127,68 @@ void* CuMemoryAllocator::MallocPitchInternal(size_t row_bytes,
...
+      // define the function signature type
+      // and get the symbol
+#if (CUDA_VERSION >= 3020)
+      typedef CUresult (*cu_fun_ptr)(size_t*, size_t*);
+      cu_fun_ptr dl_cuMemGetInfo = (cu_fun_ptr)dlsym(libcuda,"cuMemGetInfo_v2");
+#else
+      typedef CUresult (*cu_fun_ptr)(int*, int*);
+      cu_fun_ptr dl_cuMemGetInfo = (cu_fun_ptr)dlsym(libcuda,"cuMemGetInfo");
+#endif
+      if (NULL == dl_cuMemGetInfo) {
```
galv: This is subtle and doesn't really matter, but the *right* way to do this is to call dlerror() and check if the return value is NULL. This is because it is possible that a shared object may have a symbol which is initialized to NULL. Of course, that never happens in this case.

chenzhehuai:
it seems undefine GetFreeMemory()  in  cudamatrix/cu-device.h

chenzhehuai:
@danpovey  now it seems fine. btw, I didn't test other kaldi tools except GPU decoders.
@galv you might check the latest GPU decoder code (Hugo Braun works on) in this PR. 
https://github.com/kaldi-asr/kaldi/pull/2322
Please feel free to give any comments on this revision, including the style, readability and code comments. I will refine it based on your comments.

danpovey:
Note: this PR is probably superseded by #2593, which is a more fundamental rewrite of the CuDevice code.  Closing it now.

PR #2584, opened 2018-07-31
How to adapt two AMs?

xiaosdawn:
Hi all. If I have a relatively large acoustic model trained from standard materials,say broadcast news recordings, and a small model trained from recordings of specific fields,i.e. telephone speech or even speech control for machines. Is there a way to combine and adapt them to generate a new model? If so, is there a recipe for me to start with? or someone could give me a start guide?
Any advises or ideas will be appreciate! Thanks.


ericyue:
how to do it right way when am is chain/nnet3 ? @danpovey

danpovey:
You can't really merge already-trained models; the easiest way is to train on combined training data.

xiaosdawn:
Thank you so much. @danpovey 
I wonder if I could use the relatively large acoustic model to align for the new data ( some data recorded from some specific fields), then re-train a new acoustic model ?  or use the multilingual scripts to merge the new small model and the well-trained large acoustic model ?  I guess if these two acoustic models could be treated as two "languages" or "tasks" ?

danpovey:
The system won't be very sensitive to the model used for the alignments.
I don't recommend to use the multilingual setup for this.  I think it will be better to just make a combined dataset and train on it.

xiaosdawn:
OK, I got it. Thank you so much for the quickly replay. @danpovey

PR #2586, opened 2018-07-31
fix embedding training bug for RNNLM without letter features

hainan-xv:
Addressing issue #2569

### DIFF
--- a/src/rnnlm/rnnlm-embedding-training.cc
+++ b/src/rnnlm/rnnlm-embedding-training.cc
@@ -175,7 +175,7 @@ void RnnlmEmbeddingTrainer::Train(
   if (config_.l2_regularize > 0.0) {
     BaseFloat l2_term = -2 * config_.l2_regularize;
     if (l2_term != 0.0) {
-      embedding_deriv->AddToRows(l2_term, active_words, embedding_mat_);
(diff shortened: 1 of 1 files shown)


src/rnnlm/rnnlm-embedding-training.cc
```diff
@@ -175,7 +175,10 @@ void RnnlmEmbeddingTrainer::Train(
   if (config_.l2_regularize > 0.0) {
     BaseFloat l2_term = -2 * config_.l2_regularize;
     if (l2_term != 0.0) {
-      embedding_deriv->AddToRows(l2_term, active_words, embedding_mat_);
+      CuMatrix<BaseFloat> l2_matrix(embedding_deriv->NumRows(),
+                                    embedding_deriv->NumCols(), kUndefined);
```
danpovey: Can you explain what the difference is, or why the old code was wrong?

hainan-xv: The old code does

a subset of rows of embedding_mat_ += derivs * l2_factor

But it should be

derivs += a subset of rows of embedding_mat * l2_factor


Later derivs will be added to embeeding_mat with learning rate (and possibly momentum) which makes it different.

danpovey:
OK, thanks.  Merging.

PR #2587, opened 2018-07-31
Call the right script when fixing/validating data directories in OCR …

hhadian:
…recipes

### DIFF
--- a/egs/wsj/s5/utils/validate_data_dir.sh
+++ b/egs/wsj/s5/utils/validate_data_dir.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+cmd="$@"
 
 no_feats=false
 no_wav=false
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/utils/validate_data_dir.sh
```diff
@@ -44,6 +45,12 @@ if [ ! -d $data ]; then
   exit 1;
 fi
 
+if [ -f $data/images.scp ]; then
+  cmd=${cmd/--no-wav/--no-image}  # change --no-wav to --no-image, if supplied
```
danpovey: I think you should just remove the --no-wav rather than adding  --no-image.  If someone added --no-image, that implies they knew they would be calling the image version, in which case they could just call it directly.

PR #2590, opened 2018-08-04
Fix typo in diarization plda scoring

SedatDe:
Seems like there is a typo there.

### DIFF
--- a/src/ivectorbin/ivector-plda-scoring-dense.cc
+++ b/src/ivectorbin/ivector-plda-scoring-dense.cc
@@ -194,8 +194,8 @@ int main(int argc, char *argv[]) {
         } else {
           KALDI_WARN << "Unable to compute conversation dependent PCA for"
             << " recording " << reco << ".";
-          ivector_mat_plda.Resize(ivector_mat.NumRows(), ivector_mat.NumCols());
-          ivector_mat_plda.CopyFromMat(ivector_mat);
+          TransformIvectors(ivector_mat, plda_config, this_plda,
+          &ivector_mat_plda);
         }
         for (int32 i = 0; i < ivector_mat_plda.NumRows(); i++) {
           for (int32 j = 0; j < ivector_mat_plda.NumRows(); j++) {
--- a/src/ivectorbin/ivector-plda-scoring-dense.cc
+++ b/src/ivectorbin/ivector-plda-scoring-dense.cc
@@ -194,8 +194,8 @@ int main(int argc, char *argv[]) {
         } else {
           KALDI_WARN << "Unable to compute conversation dependent PCA for"
             << " recording " << reco << ".";
-          ivector_mat_pca.Resize(ivector_mat.NumRows(), ivector_mat.NumCols());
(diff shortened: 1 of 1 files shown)


src/ivectorbin/ivector-plda-scoring-dense.cc
```diff
@@ -194,8 +194,8 @@ int main(int argc, char *argv[]) {
         } else {
           KALDI_WARN << "Unable to compute conversation dependent PCA for"
             << " recording " << reco << ".";
-          ivector_mat_pca.Resize(ivector_mat.NumRows(), ivector_mat.NumCols());
-          ivector_mat_pca.CopyFromMat(ivector_mat);
+          ivector_mat_plda.Resize(ivector_mat.NumRows(), ivector_mat.NumCols());
```
danpovey: @david-ryan-snyder or @mmaciej2,  please check.

david-ryan-snyder:
@mmaciej2, could you please look at this? I have family visiting until Friday.

After glancing at this, it looks like there was indeed a bug in that branch of the if-else. We don't usually enter in to the else branch in normal use (it only happens for very short utterances, which we tend to not want to diarize anyway), which is why we didn't notice it earlier. Of course we need to fix it.

However, it looks like there's still a bug in this PR. I'll will comment inline with what I think is wrong, but @mmaciej2 please double check.

mmaciej2:
I have taken a look at this and the fix that @david-ryan-snyder commented is correct.

This part of the conditional is entered if the conversation PCA transform cannot be estimated. The current, bugged code bypasses the PCA transform and just copies the input ivectors into the "pca-transformed ivectors", presumably with the assumption that the PLDA would then be applied after that, so the bug is that the PLDA transform is never applied. @SedatDe 's fix instead copies the input ivectors to the "plda-transformed ivectors" (which is where the final output for this part of the code should be). But we don't want to skip the PLDA transform, we just want to apply it without a PCA reduction. So David's fix is correct, to apply the PLDA transform directly to the input ivectors.

danpovey:
OK but can you tell me what bug this fixes so I can edit the commit message?  I want to know what would have failed before-- it was a crash?  In what circumstance?

david-ryan-snyder:
PLDA::TransformIvectors was not applied to i-vectors when the recording was too short to compute a conversation-dependent PCA.

danpovey:
and that caused a crash or a degradation in results?

david-ryan-snyder:
It would cause a degradation in the results, since we'd still be using Plda::LogLikelihoodRatio to score pairs of i-vectors, without having first applied Plda::TransformIvectors.

However, this bug was never encountered in the existing recipe that uses this binary.

mmaciej2:
Actually, David, I think you are incorrect. As it stands in the Kaldi master branch, it will either crash or produce garbage results. On line 203 it's using the ivector_mat_plda matrix, but if EstPca fails, ivector_mat_plda is never touched outside of its declaration on line 177. The EstPca 'else' branch only assigns anything to ivector_mat_pca.

david-ryan-snyder:
Right, it will produce garbage results.

PR #2594, opened 2018-08-06
fix bug 'vector iterator + offset out of range' in nnet-common

qindazhu:


src/nnet3/nnet-common.cc
```diff
@@ -412,6 +412,12 @@ size_t IndexVectorHasher::operator () (
     ans += iter->n * 1619;
     ans += iter->t * 15649;
     ans += iter->x * 89809;
+	// check 'n2 > index_vector.size()' first,
+	// otherwise, 'end - n2' will be less than 'index_vector.begin()' and
+	// cause error "vector iterator + offset out of range" in STL vector.
+	if (n2 > index_vector.size() || iter >= end - n2)
```
danpovey: make 'size' a variable (declare it next to 'ans') as it's inefficient to keep using size() in a loop.

qindazhu: updated! Define 'len' as 'index_vector.size()'. Pls review again. Thanks!

src/nnet3/nnet-common.cc
```diff
@@ -412,6 +413,12 @@ size_t IndexVectorHasher::operator () (
     ans += iter->n * 1619;
     ans += iter->t * 15649;
     ans += iter->x * 89809;
+	// check 'n2 > index_vector.size()' first,
+	// otherwise, 'end - n2' will be less than 'index_vector.begin()' and
+	// cause error "vector iterator + offset out of range" in STL vector.
```
danpovey: Please make sure this code is correctly indented (and make sure you use spaces not tabs).
Also, you can change the comment to the following:
  // The following if-statement was introduced in order to fix an
  // out-of-range iterator problem on Windows.

src/nnet3/nnet-common.cc
```diff
@@ -392,11 +392,12 @@ size_t IndexVectorHasher::operator () (
...
                             // n2 to 1 would make the hasher consider all
                             // elements.
+  size_t len = index_vector.size();
   // all long-ish numbers appearing below are randomly chosen primes.
-  size_t ans = 1433 + 34949  * index_vector.size();
+  size_t ans = 1433 + 34949 * len;
   std::vector<Index>::const_iterator iter = index_vector.begin(),
       end = index_vector.end(), med = end;
-  if (med > iter + n1)
+  if (n1 <= len)
```
danpovey: .. also I believe this should be `if n1 < len`, not `<=`.

qindazhu: updated again. Also checked indentation and spaces. With another question:

why when `med == end == iter + n1`, we would like to do `for (; iter != med; ++iter)`  instead of skip this? This will make the hash better?

PR #2600, opened 2018-08-07
 Add a BPE-based recipe for IAM

hhadian:
Add a BPE version of the IAM recipe.

Since there were quite a lot of changes in the scripts, I created a `v2` recipe (which only has e2e and e2e+chain training). I think there are useful scripts in the `v1` recipe so it might be worth to keep it (maybe for a while). 
Here is the comparison of the same model with/without BPE:

```
# local/chain/compare_wer.sh v1/exp/chain/cnn_e2eali_1b v2/exp/chain/cnn_e2eali_1b
# System                         non-BPE       BPE
# WER                             12.40     10.33
# WER (rescored)                     --     10.10
# CER                              5.59      5.00
# CER (rescored)                     --      4.88
# Final train prob              -0.0322   -0.0428
# Final valid prob              -0.0563   -0.0666
# Final train prob (xent)       -0.6891   -0.9210
# Final valid prob (xent)       -0.8309   -1.0264
# Parameters                      3.95M     3.98M

```

Also added (in v2) a new chain script (fewer but bigger layers + smaller l2 + dropout + more epochs) which improves 1b a bit:
```
# local/chain/compare_wer.sh exp/chain/cnn_e2eali_1b exp/chain/cnn_e2eali_1c
# System                      cnn_e2eali_1b cnn_e2eali_1c
# WER                             10.33     10.05
# WER (rescored)                  10.10      9.75
# CER                              5.00      4.76
# CER (rescored)                   4.88      4.68
# Final train prob              -0.0428   -0.0317
# Final valid prob              -0.0666   -0.0630
# Final train prob (xent)       -0.9210   -0.5413
# Final valid prob (xent)       -1.0264   -0.7096
# Parameters                      3.98M     5.12M
```


danpovey:
I notice there no separate "rescored" line in the new results.  Was the LM
on the BPE system small enough to create a graph directly?  Would there
have been a benefit in using a higher-order LM, in that case, to rescore
with?

hhadian:
Do you mean for non-BPE? because there are actually "rescored" lines in the results for the BPE models.

The graph for the BPE system was very big so I used a highly-pruned 6-gram LM (created using pocolm) to create the graph and decode, and then rescore with an unpruned 6-gram pocoLM. I also tried rescoring with an 8-gram LM (created using SRILM) but it was not helpful.

danpovey:
oh, I see.

PR #2607, opened 2018-08-09
Update Windows INSTALL

naxingyu:
By following the updated instruction, the compilation works with the current openfst repo from @kkm000 without pain.


danpovey:
Thanks!

PR #2613, opened 2018-08-11
updating results and cleaning up scripts for heroico recipe, and fixed a bug in utils/lang/limit_arpa_unk_history.py

xiaohui-zhang:


xiaohui-zhang:
@danpovey good to merge

PR #2615, opened 2018-08-12
Grammar decoding

danpovey:


src/doc/grammar.dox
```diff
@@ -0,0 +1,384 @@
...
+
+ - A sequence starting and ending at the loop-state, with olabel `\#nonterm_begin`
+   and ilabels consisting of, `\#nonterm_begin` followed by all possible left-context
+   phones (and `\#nonterm_bos`) in parallel.
+ - A self-loop arc at the loop-state, with olabel `\#nonterm_end`
+   and ilabel `\#nonterm_end`.
+ - For each user-defined nonterminal (e.g. `\#nonterm:foo`) and for
+   `\#nonterm_begin`, the following:
+    - An arc from the loop-state to a newly created state, with ilabel
+      and olabel equal to the symbol in question (say `\#nonterm:foo`).
```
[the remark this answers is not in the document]
danpovey: @ngoel17 
This is just a mechanism to pass those symbols through the lexicon, so if they appear on the ilabels of the grammar, they will also appear on the ilabels of LG.fst; and to add all possible phonetic contexts where appropriate.
I'm not sure what scenario you want to support in combination with this?  An example might help

egs/mini_librispeech/s5/local/grammar/simple_demo.sh
```diff
@@ -0,0 +1,169 @@
...
+
+fi
+
+if [ $stage -le 7 ]; then
+  echo "$0: will test equivalece using fstequivalent"
+  if fstequivalent --delta=0.01 --random=true --npath=100 $tree_dir/grammar1/HCLG.fst $tree_dir/grammar2/HCLG.fst; then
+    echo "$0: success: the two were equivalent"
+  else
+    echo "$0: failure: the two were equivalent"
+  fi
```
chenzhehuai: Cool. In my understanding, we might need to further verify it by a complexer example (e.g. unk word recognition). My experience (from the previous AFST idea) is when we have a complexer top-level graph (e.g. the following one), I fail in making the grammar1 and grammar2 equivalent and also obtain some WER drops if decoding.
0    1    GROUP   GROUP
1    2    #nonterm:contact_list  <eps>
2    3    #nonterm:phone_number  <eps>
3    4    ASSIST   ASSIST  0.69314718055994
4    5    #nonterm:contact_list  <eps>
5

danpovey:
Thanks... yes, I'll try to make more examples so we can test it more
thoroughly.  Right now I'm working on an example for augmenting the
vocabulary with new words.

src/doc/grammar.dox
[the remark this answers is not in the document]
ngoel17: Example is : input:  one hundred and seven
                   output:  107
   It can be achieved by having a G.fst where ilable and olabel are not the same.

danpovey:
oh, OK.
The framework does not have a problem with that.  It only deals with the
special symbols on the ilabels on G.fst, it does not inspect the olabels.

egs/wsj/s5/utils/lang/make_lexicon_fst.py
```diff
@@ -0,0 +1,407 @@
...
+                    # has a silence transition to loop_state.
+    new_state = 3  # the next un-allocated state, will be incremented as we go.
+
+
+    print('{src}\t{dest}\t{phone}\t{word}\t{cost}'.format(
+        src=start_state, dest=loop_state,
+        phone='<eps>', word='<eps>', cost=no_sil_cost))
+    print('{src}\t{dest}\t{phone}\t{word}\t{cost}'.format(
+        src=start_state, dest=sil_state,
+        phone='<eps>', word='<eps>', cost=sil_cost))
```
DongjiGao: I guess the cost from start_state to sil_state should be 0.0 here or it would add sil_cost twice if the sentence starts with SIL.

danpovey:
Thanks for checking it.  This is the penalty for seeing silence... you take
the penalty when you enter the sil state, not when you leave.  (this was a
translation of the existing perl script, and I tested it produced
equivalent output).

egs/wsj/s5/utils/lang/make_lexicon_fst.py
```diff
@@ -0,0 +1,407 @@
...
+    ans = []
+    found_empty_prons = False
+    found_large_pronprobs = False
+    # See the comment near the top of this file, RE why we use latin-1.
+    with open(filename, 'r', encoding='latin-1') as f:
+        for line in f:
+            a = line.split()
+            if len(a) < 2:
+                print("{0}: error: found bad line '{1}' in lexicon file {1} ".format(
+                    sys.argv[0], line.strip(), filename), file=sys.stderr)
```
DongjiGao: The "{1}" appears twice in string formatting. The second "{1}" should be "{2}". It also happens in later parts.

danpovey:
Thanks a lot-- will fix.

egs/mini_librispeech/s5/local/grammar/simple_demo.sh
danpovey: @DongjiGao, when you make your more thorough test example, make sure to have nonterminals following each other directly, and also invoking each other.  (Not recursively though).

PR #2618, opened 2018-08-13
Update configure to handle rpm based openblas

btiplitz:
as I have both ubuntu and centos systems, I've updated configure such that it can handle openblas as installed by rpm or packages


src/configure
```diff
@@ -1317,18 +1317,29 @@ elif [ "`uname`" == "Linux" ]; then
...
+    if [ -f $OPENBLASROOT/lib/libopenblas.so ]; then
+      OPENBLASLIBDIR=$OPENBLASROOT/lib
+    elif [ -f $OPENBLASROOT/lib64/libopenblas.so ]; then
+      OPENBLASLIBDIR=$OPENBLASROOT/lib64
+    else
       failure "Expected to find the file $OPENBLASROOT/lib/libopenblas.so"
     fi
+    if [ -f $OPENBLASROOT/include/openblas_config.h ] ; then
+      OPENBLASINCDIR=$OPENBLASROOT/include
+    elif [ -f $OPENBLASROOT/include/openblas/openblas_config.h ] ; then
```
galv: Could you comment that this is the debian-based install path, and that the former is (I assume) the rpm install path.

src/configure
```diff
@@ -1317,18 +1317,29 @@ elif [ "`uname`" == "Linux" ]; then
     if [ -z "$OPENBLASROOT" ]; then
       failure "Must specify the location of OPENBLAS with --openblas-root option (and it must exist)"
     fi
-    if [ ! -f $OPENBLASROOT/lib/libopenblas.so ]; then
+    if [ -f $OPENBLASROOT/lib/libopenblas.so ]; then
```
galv: Can you comment that this is the deb-based install path and the second case is the rpm-based install path for clarity to others?

src/configure
galv: Actually, never mind, on a Centos7 machine and a Ubuntu 16.04 machine, I get /usr/include/openblas after installing the appropriate packages. What is this case `if [ -f $OPENBLASROOT/include/openblas_config.h ] ; then` handling? Is it when openblas is installed via kaldi's tools/ directory?

tools/extras/check_dependencies.sh
```diff
@@ -96,8 +96,8 @@ fi
 
 if ! which python3 >&/dev/null; then
   echo "$0: python3 is not installed"
-  add_packages python3
-  pythonok=false
+#  add_packages python3
```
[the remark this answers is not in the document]
btiplitz: Did not mean to checkin.  my desktop is missing python3 and I can't install it.  (No permissions and did not feel like waiting to build it)

src/configure
btiplitz: will do

src/configure
btiplitz: The build fails.  I wonder if that is why it's failing to build at kaldi-asr. Hmm

danpovey: I don't think search for openblas_config.h is appropriate in the case where it's a system package.  I believe the dev package may provide cblas.h and lapacke.h.   Just replacing openblas_config.h with cblas.h in all these checks should be fine, and may allow it to build on travis.

btiplitz:
is there anything else I need to do before this could be merged ?

src/configure
```diff
@@ -1309,18 +1309,32 @@ elif [ "`uname`" == "Linux" ]; then
...
+    else
       failure "Expected to find the file $OPENBLASROOT/lib/libopenblas.so"
     fi
+    if [ -f $OPENBLASROOT/include/cblas.h ] ; then
+      OPENBLASINCDIR=$OPENBLASROOT/include
+    elif [ -f $OPENBLASROOT/include/openblas/cblas.h ] ; then
+      # in REDHAT/CentOS/Ubuntu package installs, the includes are located here
+      OPENBLASINCDIR=$OPENBLASROOT/include/openblas
+    else 
+      echo "Assuming openblas is aleady in a default include path"
```
danpovey: You should probably print a scarier warning here, e.g.
echo "$0: ***** Using OpenBlas from $OPENBLASROOT but cblas.h is not found.  "
echo "     ******  Assuming openblas is aleady in a default include path, but"
echo "      *****  if you get compilation messages about not finding files like cblas.h,"
echo "      ***** you should look into this (e.g. make sure to install the 'openblas-dev' package,"
echo "      ***** if it is a package-based install)."


Also, please confirm that you tested that this still works if you use the default OPENBLASROOT based on the install from tools/.

btiplitz: Incorporated.  And yes I am testing this with the tools openblas.

PR #2621, opened 2018-08-14
Tamil ocr with training data for yomdle and testing data from slam

aarora8:
It's using 30k utterances from yomdle English and Tamil dataset. It is getting the line image from the MAR (minimum area rectangle) @jtrmal @hhadian @ChunChiehChang 

- It is getting the line image from the MAR (minimum area rectangle).
- BPE based model with vocab size=700, grayscale images, normalized height 40
- It is currently building the language model from corpus text shared during Scale workshop (5M lines). It is performing 3-gram decoding and 6-gram rescoring.
- shared model with English and Tamil line images.
- directly downscaling and upscaling based data augmentation
- semi-supervised training (22k untranscribed line images)
- normalized scoring scripts (shared by Stephen Rawls)
- decoding beam = 12


danpovey:
Thanks!  @hhadian, I hope you can do the first passes of review...

hhadian:
Sure, will do.

egs/yomdle_tamil/v1/local/chain/run_cnn_e2eali_1b.sh
```diff
@@ -0,0 +1 @@
+tuning/run_cnn_e2eali_1b.sh
```
hhadian: The scripts which are not in the tuning directory do not need to have a suffix (like 1b). They usually point to the best recipe in the tuning directory. Please rename this to `run_cnn_e2eali.sh`

egs/cifar/v1/image/ocr/yomdle/create_download_dir.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
hhadian: I'm not sure if this `yomdle` directory should be a subdirectory of `image/ocr` because I guess the shared scripts should not be dataset-specific. Maybe this could be a subdirectory of `egs/yomdle_tamil/v1/local` and other yomdle recipes can point to it?

aarora8:
Ok, thank you. I will make the modifications from the above comments.

ChunChiehChang:
We might want to add an option to flip the image in ./image/ocr/make_features.py for Farsi and Arabic

aarora8:
Ok, thanks. I will add it and test it with the Farsi setup.

egs/yomdle_tamil/v1/local/process_augment_data.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+import os
+import argparse
+
+parser = argparse.ArgumentParser(
+    description="""Regenerate images.scp, text, utt2spk and spk2utt from feats.scp for augment data""")
```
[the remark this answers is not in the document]
aarora8: This script is reading the feats.scp file and regenerating images.scp, text, utt2spk and spk2utt. Apart from original images.scp, text, utt2spk new values were created due to augmentation (it is currently doubling the data). Hence to map newly created feats, to text, images.scp and utt2spk this script is used. But I am not sure, if this is the best way, should we do it some other way.

egs/yomdle_tamil/v1/local/chain/run_e2e_cnn.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
```
hhadian: Would it be possible to add some results to these scripts?

egs/yomdle_tamil/v1/local/train_lm.sh
```diff
@@ -0,0 +1,128 @@
...
+fi
+
+if [ $stage -le 1 ]; then
+  # decide on the vocabulary.
+  # Note: you'd use --wordlist if you had a previously determined word-list
+  # that you wanted to use.
+  # Note: if you have more than one order, use a certain amount of words as the
+  # vocab and want to restrict max memory for 'sort',
+  echo "$0: training the unpruned LM"
+  min_counts='train=2 madcat=1'
```
hhadian: this shouldn't be `madcat`

egs/yomdle_tamil/v1/local/yomdle/normalized_scoring/utils/normalize_common.py
```diff
@@ -0,0 +1,37 @@
+import sys
```
hhadian: If possible, please add a one-line description for all the scripts under `yomdle`.

egs/cifar/v1/image/ocr/process_augment_data.py
```diff
@@ -0,0 +1,78 @@
...
+text_file = os.path.join(args.dir + '/', 'text')
+text_fh = open(text_file, 'w')
+
+utt2spk_file = os.path.join(args.dir + '/', 'utt2spk')
+utt2spk_fh = open(utt2spk_file, 'w')
+
+feats_scp_file = os.path.join(args.dir, 'feats.scp')
+with open(feats_scp_file) as feats_scp_fh:
+    for uttid_image_path in feats_scp_fh:
+        uttid_image_path = uttID_image_path.strip()
```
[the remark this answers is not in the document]
aarora8: sorry, I missed some places, I will fix it at all places and run the setup again to check if it is working properly. It is currently not getting used, I will call with the new modifications.

egs/yomdle_tamil/v1/local/chain/run_e2e_cnn.sh
aarora8: Ok, thank you, I will add it.

egs/yomdle_tamil/v1/local/train_lm.sh
aarora8: sorry, I will fix it.

aarora8:
Sorry, currently, I did not add bidi-related scripts. I was not sure should we add it now or with the Farsi setup.

hhadian:
Yes, it makes sense to add it with the Farsi setup.

egs/yomdle_tamil/v1/local/yomdle/normalized_scoring/utils/normalize_common.py
aarora8: Thanks, I added the description by going through the scripts. Since the codes were small, I feel it is correct but I might have missed some details.

egs/yomdle_tamil/v1/local/extract_features.sh
```diff
@@ -0,0 +1,68 @@
...
+    copy-feats --compress=true --compression-method=7 \
+    ark:- ark,scp:$featdir/images.JOB.ark,$featdir/images.JOB.scp
+
+## aggregates the output scp's to get feats.scp
+for n in $(seq $nj); do
+  cat $featdir/images.$n.scp || exit 1;
+done > $data/feats.scp || exit 1
+
+# re-map utt2spk, images.scp and text if doing image augmentation
+image/ocr/process_augment_data.py $data
```
ChunChiehChang: I think there's a bug with this were it will give an error for the unaugmented test data directory. Or are you augmenting the test data as well?

aarora8: Thanks, currently in the extract features script, the augmentation is turned off for both training and test data. I will add the modification to run augmentation for train data but not for test data.

egs/cifar/v1/image/ocr/process_augment_data.py
```diff
@@ -11,68 +11,54 @@
...
-        uttid_spk_vect = uttid_spk.split(" ")
-        uttid = uttid_spk_vect[0]
-        image_id = uttid
+        uttid_spk_vect = uttid_spk.strip().split(" ")
+        utt_id = uttid_spk_vect[0]
         spk = uttid_spk_vect[1]
-        uttid_to_spk[image_id] = spk
+        uttid_to_spk[utt_id] = spk
 
 image_file = os.path.join(args.dir, 'backup', 'images.scp')
```
hhadian: Ah! I see now what you are doing. You make a copy of the data dir (as a backup), then call `make_features.py` to generate feats.scp with 2 times as many utterances as there are. Then you call this script to update text, images.scp and utt2spk to cover all the entries in feats.scp. Is that right?

I think it would be better to take this approach instead: remove anything related to augmentation in `extract_features.sh` add a make a new script `local/augment_data.sh <src-dir> <out-dir>` which does something like this:
```
utils/copy_data_dir.sh --spk-prefix aug1- --utt-prefix aug1- $srcdir ${outdir}_aug1
# call extract_features.sh on ${outdir}_aug1 with appropriate augmentation options (which will simply be passed to make_features.py)
# ... do as many augmentations as needed, e.g. aug2, aug3, ... and then combine them all:
utils/combine_data_dir.sh --extra-files images.scp $outdir  $srcdir ${outdir}_aug1 ${outdir}_aug2 ...
# remove the temporary augX dir's:
rm -r ${outdir}_aug1 ${outdir}_aug2 ...
```
You might need to modify `utils/copy_data_dir.sh` to be able to handle `images.scp`.

aarora8: Thank you. Yeah, the process_augment_data script is doing exactly same as you said (updating the text, images.scp and utt2spk to cover all the entries in feats.scp). I will surely make the above mentioned modifications for augmenting.

hhadian:
Thanks Ashish, is this ready? Have you tested the new changes?

aarora8:
Thanks, it is almost done,  during testing there was a small bug, I will surely fix it today.

aarora8:
Thanks, i tested it. It is working ok now.

danpovey:
@hhadian please let me know when I should merge.

egs/wsj/s5/utils/copy_data_dir.sh
```diff
@@ -86,10 +86,16 @@ fi
 
 if [ -f $srcdir/segments ]; then
   utils/apply_map.pl -f 1 $destdir/utt_map <$srcdir/segments >$destdir/segments
-  cp $srcdir/wav.scp $destdir
+  if [ -f $srcdir/wav.scp ]; then
+    cp $srcdir/wav.scp $destdir
+  elif [ -f $srcdir/images.scp ]; then
+    cp $srcdir/images.scp $destdir
+  fi
```
hhadian: Since we don't use a segments file in OCR recipes, we might not even need to add these changes. i.e. when there is a segments file, then it's not an OCR data dir. @danpovey, what do you think?

Also, I suggested editing `utils/copy_data_dir.sh` instead of creating a new `image/copy_data_dir.sh` because I expected the changes to be small. @danpovey what would you suggest?

egs/wsj/s5/utils/lang/bpe/prepend_words.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# This script, prepend '|' to every words in the transcript to mark
+# the beginning of the words for finding the initial-space of every word
+# after decoding.
+
+import sys, io
+
+infile = io.TextIOWrapper(sys.stdin.buffer, encoding='utf-8')
```
danpovey: For anything that doesn't need to split inside words (like this script) I prefer the latin-1 encoding; that will also work on data like GBK that's not utf-8 but which is compatible with ASCII.

danpovey:
@hhadian , after thinking about this for a bit (and it was a difficult
decision) I think it might be better to use the copy-and-modify approach.
The issue is that we don't know, in future, how extensive the changes might
be, and once we go down the road of using the same shared scripts, it may
lead us into problems.  The scripts for the OCR could be substantially
simpler as there are many things that don't need to be supported.  And
there may be people only interested in OCR who would appreciate having
relatively clean/simple scripts.
But bear in mind that validate_data_dir.sh (and maybe fix_data_dir.sh)
should be copied and modified along with copy_data_dir.sh.

danpovey:
Great.

hhadian:
This looks good to me.

PR #2622, opened 2018-08-14
fixing Dockerfiles not working

keighrim:
`Dockerfile`s in `misc/docker` are not building kaldi as there are missing dependencies.


chheang620:
- [ ] 
Chheang620 
- 

> ### <key>GADApplicationIdentifier</key>
<string>ca-app-pub-3940256099942544~1458002511</string>psi

PR #2625, opened 2018-08-15
Fix backwards-incompatibility of changes made in #2383

funous:
This fixes the backwards-incompatible changes made to AddSelfLoops in #2383. Instead of changing the original function signature we introduce a new function that provides the new functionality, calling this new function from the old one and from all the places where the new arguments were used with non-default values.

This makes AddSelfLoops backwards-compatible - external calls won't have to be updated (unless they were updated since #2383).

### DIFF
--- a/src/bin/add-self-loops.cc
+++ b/src/bin/add-self-loops.cc
@@ -97,13 +97,11 @@ int main(int argc, char *argv[]) {
     if (!fst)
       KALDI_ERR << "add-self-loops: error reading input FST.";
 
-    bool check_no_self_loops = true;
-
     // The work gets done here.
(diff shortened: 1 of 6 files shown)


danpovey:
Sorry... I don't think I want to merge this.  My attitude is that "that ship has sailed", meaning that likely people will have already solved any back-compatibility issue (we do very clearly explain how to fix it, and it's an easy fix).  Making them fix it again would, IMO, be a bit perverse.  Also, the new code is obviously a bit more verbose than the old code.   I think if anyone is using Kaldi as a library they have to understand that occasionally little changes like that will have to be made.

PR #2629, opened 2018-08-17
Fixes to chain and relevant nnet3 python scripts to support python3.x

akshayc11:
Details:
1. thread module has been renamed to _thread from python3

https://docs.python.org/2/library/thread.html

```
Python 3.4.8 (default, Mar 23 2018, 10:04:27)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import thread
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-e75c663b2a08> in <module>()
----> 1 import thread

ImportError: No module named 'thread'
```

2. any division operation results in float by default
```
Python 3.6.3 (default, Jan  4 2018, 16:40:53)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: 32/2
Out[1]: 16.0

In [2]: int(32)/2
Out[2]: 16.0

In [3]: int(int(32)/2)
Out[3]: 16

In [5]: 16/4 % 3
Out[5]: 1.0
```


egs/wsj/s5/steps/libs/nnet3/train/chain_objf/acoustic_model.py
```diff
@@ -167,8 +167,8 @@ def train_new_models(dir, iter, srand, num_jobs,
         # work out the 1-based archive index.
         archive_index = (k % num_archives) + 1
         # previous : frame_shift = (k/num_archives) % frame_subsampling_factor
-        frame_shift = ((archive_index + k/num_archives)
-                       % frame_subsampling_factor)
+        frame_shift = int((archive_index + k/num_archives)
```
danpovey: thanks.  I think a better fix here would be to write k // num_archives.  Same for the other place-- change / to //.

PR #2639, opened 2018-08-23
[scripts] Add basis-fMLLR version of align_lats.sh

danpovey:


danpovey:
I accidentally did a 'git push' that pushed this to master.  But it's OK.  If I find a problem in this script I'll fix it but it won't break anything.

PR #2641, opened 2018-08-24
Only include tokens with <eps> arc in queue

desh2608:


danpovey:
Thanks!  There seems to be a compilation error in lattice-simple-decoder.cc though, from the travis buld.

desh2608:
@danpovey oh alright, I'll try to fix it.

dophist:
We have already had all frontier token in the queue, why do we add tokens(with epsilon ilabel) again? Just from a quick skim, not for sure. Should this check be done in the first place of queue's push_back() ?

src/decoder/lattice-faster-decoder.cc
```diff
@@ -818,6 +818,12 @@ void LatticeFasterDecoder::ProcessNonemitting(BaseFloat cutoff) {
     }
   }
 
+  for (const Elem *e = toks_.GetList(); e != NULL;  e = e->tail) {
```
danpovey: Yes, @dophist  is right; you need to replace the loop at line 812 with this.

src/decoder/lattice-simple-decoder.cc
```diff
@@ -564,8 +565,11 @@ void LatticeSimpleDecoder::ProcessNonemitting() {
   for (unordered_map<StateId, Token*>::iterator iter = cur_toks_.begin();
        iter != cur_toks_.end();
        ++iter) {
-    queue.push_back(iter->first);
-    best_cost = std::min(best_cost, iter->second->tot_cost);
+    StateId key = iter->first;
+    if (fst_->NumInputEpsilons(key) != 0) {
```
desh2608: @vimalmanohar I'm getting an error here because of this method. Can you help with this?

vimalmanohar: It looks like fst_ is not a pointer. In Dan's PR, it was changed to a pointer. So here you need to use `fst_.`

src/decoder/lattice-faster-online-decoder.cc
```diff
@@ -989,8 +989,11 @@ void LatticeFasterOnlineDecoder::ProcessNonemitting(BaseFloat cutoff) {
   // problem did not improve overall speed.
 
   KALDI_ASSERT(queue_.empty());
-  for (const Elem *e = toks_.GetList(); e != NULL;  e = e->tail)
-    queue_.push_back(e->key);
+  for (const Elem *e = toks_.GetList(); e != NULL;  e = e->tail) {
+    StateId key = e->key;
+    if (fst_.NumInputEpsilons(key) != 0)
```
danpovey: there is another place where this check should be done, in the loop below this, before you push to the queue.  (applies to all the decoders).

src/decoder/lattice-simple-decoder.cc
```diff
@@ -564,8 +564,11 @@ void LatticeSimpleDecoder::ProcessNonemitting() {
   for (unordered_map<StateId, Token*>::iterator iter = cur_toks_.begin();
        iter != cur_toks_.end();
        ++iter) {
-    queue.push_back(iter->first);
-    best_cost = std::min(best_cost, iter->second->tot_cost);
+    StateId key = iter->first;
+    if (fst_.NumInputEpsilons(key) != 0) {
+      queue.push_back(key);
+      best_cost = std::min(best_cost, iter->second->tot_cost);
```
danpovey: the assignment of best_cost should not be inside the if-statement.  Check that in my original PR I did not make this mistake.

desh2608: Oh okay. Actually the PR did not have changes in the lattice-simple-decoder, so I figured the best_cost is only over those objects which are actually being pushed to the queue. Anyway, I'll make this change.

danpovey: Thanks.  The idea is to make the code equivalent to what it was before (but faster).  The best_cost is over all live tokens, which includes states that don't have epsilon transitions frm them.

src/decoder/lattice-faster-decoder.cc
```diff
@@ -809,8 +809,11 @@ void LatticeFasterDecoder::ProcessNonemitting(BaseFloat cutoff) {
   // problem did not improve overall speed.
 
   KALDI_ASSERT(queue_.empty());
-  for (const Elem *e = toks_.GetList(); e != NULL;  e = e->tail)
-    queue_.push_back(e->key);
+  for (const Elem *e = toks_.GetList(); e != NULL;  e = e->tail) {
+    StateId key = e->key;
```
danpovey: actually, I think it would be clearer if this variable were named state, and not key.  I know it was key in my branch; I'll fix the conflict later.

PR #2656, opened 2018-08-30
Remove pre-C++11 compatibility for noexcept.

galv:
I'm happy to remove the KALDI_NOEXCEPT macro altogether, since no one sane should have depended on this macro in their code that uses Kaldi as an external library. I'm just being paranoid about backwards compatibility here.


danpovey:
I'd just remove it altogether, I don't think anyone would be using it and anyway it would just be a compilation error to fix.

galv:
You're right. I updated this PR. This change is good to go.

PR #2658, opened 2018-09-01
Iam

aarora8:
It contains some modifications related to as follows: 
- adding smaller HMM/GMM (4 states instead of 8) punctuation topology
- augmentation (shearing,  contrast normalization)  
- run.sh BPE based script
- parameter tuning of chain scripts


danpovey:
Thanks!!  @hhadian please review

egs/iam/v2/run_end2end.sh
```diff
@@ -27,35 +28,70 @@ wellington_database=/export/corpora5/Wellington/WWC/
...
+            phone_dict[phone] = phone;
+for phone in phone_dict.keys():
+      output.write(phone+ '\n');
+END
+   ) > data/local/phones.txt
+
+  cut -d' ' -f2- data/train/text > data/local/train_data.txt
+  cat data/local/phones.txt data/local/train_data.txt | \
     local/prepend_words.py | \
     utils/lang/bpe/learn_bpe.py -s 700 > data/local/bpe.txt
```
hhadian: Could you please check to see if there is an option that we can pass to learn_bpe.py to make sure all singleton characters are included (so that you don't need to create local/phones.txt)?

egs/iam/v2/run.sh
```diff
@@ -0,0 +1,200 @@
+#!/bin/bash
```
hhadian: Is there an obvious gain from using this script? If I'm right, the results are almost the same. I guess it's better to keep v2 small and only add scripts which give a clear improvement.

egs/cifar/v1/image/copy_data_dir.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
```
hhadian: Won't this be added in the yomdle PR?

egs/iam/v2/run.sh
aarora8: Thanks, I am trying to tune run_end2end.sh to give better results. currently, run.sh is giving a slightly better result as compared to run_end2end.sh (9.0 vs 9.3). I will remove it.

egs/iam/v2/run_end2end.sh
aarora8: ok, thank you. I will check it.

egs/cifar/v1/image/copy_data_dir.sh
aarora8: Sorry, I will remove it.

egs/iam/v2/local/chain/tuning/run_cnn_e2eali_1a.sh
```diff
@@ -243,3 +244,21 @@ if [ $stage -le 7 ]; then
     --nj $nj --cmd "$cmd" \
     $dir/graph data/test $dir/decode_test || exit 1;
 fi
+
+if [ $stage -le 8 ] && $decode_val; then
```
hhadian: If it's not a hassle, please decode both `val` and `test` in the same stage using a for loop, e.g. `for testset in test $maybe_val; do  ... done` where maybe_val can be set to "val" or empty. See swbd/s5c/local/chain/run_tdnn.sh.

egs/iam/v2/local/chain/tuning/run_cnn_e2eali_1d.sh
```diff
@@ -0,0 +1,269 @@
+#!/bin/bash
+
+# This script does end2end chain training (i.e. from scratch)
```
hhadian: Please add a line explaining how this is different from `1c`

egs/iam/v2/run_end2end.sh
```diff
@@ -27,56 +29,97 @@ wellington_database=/export/corpora5/Wellington/WWC/
 ./local/check_tools.sh
 
 if [ $stage -le 0 ]; then
+  if [ -f data/train/text ] && ! $overwrite; then
+    echo "Not processing, probably script have run from wrong stage"
```
hhadian: Don't forget `$0`. Also I'm not sure how this will help if someone calls this script from stage 1, maybe you meant to put this check in stage 1 (before `echo "$0: Preparing data..."`)?

egs/iam/v2/run_end2end.sh
```diff
@@ -27,56 +29,97 @@ wellington_database=/export/corpora5/Wellington/WWC/
...
+            phone_dict[phone] = phone;
+for phone in phone_dict.keys():
+      output.write(phone+ '\n');
+END
+   ) > data/local/phones.txt
+
+  cut -d' ' -f2- data/train/text > data/local/train_data.txt
+  cat data/local/phones.txt data/local/train_data.txt | \
+    utils/lang/bpe/prepend_words.py | \
     utils/lang/bpe/learn_bpe.py -s 700 > data/local/bpe.txt
```
hhadian: Did you have a chance to check learn_bpe.py to see if it has an option to include all singletons?

egs/iam/v2/run.sh
hhadian: I see, but the results are the same on the other 3 test conditions so this 0.3 difference could be noise.

egs/iam/v2/run_end2end.sh
aarora8: sorry, I tried going through the code, but i was not able to complete it. I will surely do it now.

aarora8:
Ok thank you, I will make the modifications from the review.

egs/iam/v2/run_end2end.sh
aarora8: I tried checking the options and docs, there is no option to include singletons

egs/iam/v2/local/chain/tuning/run_cnn_e2eali_1d.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# e2eali_1c is the same as e2eali_1c but has more CNN layers, different filter size
```
hhadian: e2eali_1c is --> e2eali_1d is
less minibatch, frams-per-iter --> smaller minibatches and frames-per-iter

hhadian:
This looks good to me.

PR #2666, opened 2018-09-01
[src] Avoid accidentally using double precision

christoph-conrads:
* avoid an accidental double-precision computation (`1.0/x`)
* replace multiplications with the reciprocal of a value with divisions

### DIFF
--- a/src/feat/mel-computations.cc
+++ b/src/feat/mel-computations.cc
@@ -190,13 +190,12 @@ BaseFloat MelBanks::VtlnWarpFreq(BaseFloat vtln_low_cutoff,  // upper+lower freq
   BaseFloat one = 1.0;
   BaseFloat l = vtln_low_cutoff * std::max(one, vtln_warp_factor);
   BaseFloat h = vtln_high_cutoff * std::min(one, vtln_warp_factor);
-  BaseFloat scale = 1.0 / vtln_warp_factor;
-  BaseFloat Fl = scale * l;  // F(l);
-  BaseFloat Fh = scale * h;  // F(h);
+  BaseFloat Fl = l / vtln_warp_factor;  // F(l);
+  BaseFloat Fh = h / vtln_warp_factor;  // F(h);
   KALDI_ASSERT(l > low_freq && h < high_freq);
   // slope of left part of the 3-piece linear function
   BaseFloat scale_left = (Fl - low_freq) / (l - low_freq);
-  // [slope of center part is just "scale"]
+  // [slope of center part is just "1/vtln_warp_factor"]
 
   // slope of right part of the 3-piece linear function
   BaseFloat scale_right = (high_freq - Fh) / (high_freq - h);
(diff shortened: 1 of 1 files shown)


galv:
Actually, I'm certain this code does not generate any double precision instructions in the first place.

Take a look below:

```
galv@training-Silverback-WS:~/development/test/double$ cat double.cc
int double_promotion() {
  float divisor = 2.0;
  float divided = 1.0 / divisor;
  return divided;
}
galv@training-Silverback-WS:~/development/test/double$ g++ -g -c double.cc 
galv@training-Silverback-WS:~/development/test/double$ objdump -d -S double.o | less

double.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <_Z16double_promotionv>:
int double_promotion() {
   0:   55                      push   %rbp
   1:   48 89 e5                mov    %rsp,%rbp
  float divisor = 2.0;
   4:   f3 0f 10 05 00 00 00    movss  0x0(%rip),%xmm0        # c <_Z16double_promotionv+0xc>
   b:   00 
   c:   f3 0f 11 45 f8          movss  %xmm0,-0x8(%rbp)
  float divided = 1.0 / divisor;
  11:   f3 0f 10 05 00 00 00    movss  0x0(%rip),%xmm0        # 19 <_Z16double_promotionv+0x19>
  18:   00 
  19:   f3 0f 5e 45 f8          divss  -0x8(%rbp),%xmm0
  1e:   f3 0f 11 45 fc          movss  %xmm0,-0x4(%rbp)
  return divided;
  23:   f3 0f 10 45 fc          movss  -0x4(%rbp),%xmm0
  28:   f3 0f 2c c0             cvttss2si %xmm0,%eax
}
  2c:   5d                      pop    %rbp
  2d:   c3                      retq   
```

You can look up divss to see that it is for single precision floating point.

I'm hazy on the details, but I believe that C (and C++) take into account the type of the variable that an arithmetic operation is assigned to, and in that way avoids promoting the output of `1.0 / divisor` to a double, only to truncate it to a single-precision float.

I'm happy to change my mind on this if you can convince me otherwise.

christoph-conrads:
tl;Dr: The code converts to double precision according to the C++ standard but after computing `1.0/x` and `1.0f/x` for all single-precision floats `x`, I can say that the rounded double-precision result is always equal to the direct computation in single precision.

> Actually, I'm certain this code does not generate any double precision instructions in the first place.

I compiled this code with GCC 4.8.3 and GCC 7.3.0 and both compilers use a single-precision division, even for divisors different from 2 but read on.

> I'm hazy on the details, but I believe that C (and C++) take into account the type of the variable that an arithmetic operation is assigned to, and in that way avoids promoting the output of 1.0 / divisor to a double, only to truncate it to a single-precision float.

I never heard of this rule and I do not believe it exists. For example, the statement `unsigned long ul = 184467440737095516151;` causes a compiler warning with both GCC and Clang because the right-hand side is interpreted as `int` or `long int`. Also, according to the wiki page [floating point literal](https://en.cppreference.com/w/cpp/language/floating_literal), a float without suffix must be interpreted as double.

You can also use C++11 `auto` and have the compiler infer the type:
```
#include <cstddef>
std::size_t double_promotion() {
  float divisor = 2.0;
  auto divided = 1.0 / divisor;
  return sizeof(divided);
}
```
On my computer, I get eight implying a `double`. Moreover, if you compile the code with GCC 4.8.3, GCC 7.3.0, or Clang 3.4 and the option `-Wconversion`, you will see the following warning:
> double.cc: In function ‘float double_promotion()’:
> double.cc:3:22: warning: conversion to ‘float’ from ‘double’ may alter its value [-Wconversion]
>  float divided = 1.0 / divisor;

Finally, Clang does generate a double-precision division but only if optimizations are disabled.

PR #2668, opened 2018-09-01
[src] Diagnose invalid `VtlnWarpFreq` input

christoph-conrads:
* remove an assertion that was always true
* always check if the VTLN low, high cut-off form a non-empty interval
  after multiplication with the VTLN warp factor


danpovey:
Sorry, closing this because I don't have time to analyze it to figure out whether it is correct (and in any case it's likely not necessary).

christoph-conrads:
> (and in any case it's likely not necessary).

Once the Nyquist frequency is known, there are several checks in the code that ensure the following invariant:
```
freq_low < vtln_low < vtln_high < freq_high <= freq_nyquist.
```
There is no such check for the warped VTLN cutoffs. If `l > h` in `VtlnWarpFrequency`, then frequencies `f` higher than `vtln_low` will be mapped to larger values than frequencies `f > h`.

PR #2683, opened 2018-09-05
update logging to allow threshold controls

btiplitz:
KALDI uses KALDI_LOG when the output message is too verbose.  I've updated several places to use KALDI_VLOG... as to allows the reduction of logging messages


jtrmal:
sory, the PR is empty -- see the "Files changed" card.

btiplitz:
I just noticed that.  I'll check

btiplitz:
Sorry I'm new to this tool.  I missed a step so changes should now be present

danpovey:
Sorry, I don't want to merge this, those logging messages are useful enough that I'd rather have them visible by default.
If you have an application where they are a hassle, you can just keep those changes in  your repo and merge changes from master into it.

PR #2691, opened 2018-09-08
WIP: Add CUDNN as an optional dependency to cu-device.{cc,h}

galv:
Currently, this assumes that cudnn is installed in
/usr/local/cuda/lib64 and /usr/local/cuda/include, because that's what
my current machine has it installed. This can be cleaned up later.

@danpovey You may want to base your work on this starting point

### DIFF
--- a/src/cudamatrix/cu-device.h
+++ b/src/cudamatrix/cu-device.h
@@ -35,11 +35,7 @@
 #include "base/kaldi-common.h"
 #include "base/timer.h"
 #include "cudamatrix/cu-allocator.h"
-
-// Forward declare the cudnnHandle_t found in cudnn.h so that we don't
-// need to #include <cudnn.h>. This allows us to make cudnn an
-// optional dependency.
-typedef struct cudnnContext *cudnnHandle_t;
+#include "cudamatrix/cu-cudnn-helper.h"
 
 namespace kaldi {
 
--- a/src/cudamatrix/cu-device.h
+++ b/src/cudamatrix/cu-device.h
@@ -2,6 +2,7 @@
 
 // Copyright 2009-2012  Karel Vesely
 //           2012-2015  Johns Hopkins University (author: Daniel Povey)
+//           2018       Daniel Galvez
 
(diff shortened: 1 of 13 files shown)


src/cudamatrix/cu-device.h
```diff
@@ -184,6 +191,7 @@ class CuDevice {
   /// (i.e. from outside the class), call this only if Enabled() returns true.
   bool IsComputeExclusive();
 
+  // Shouldn't this be a private constructor?
```
galv: @danpovey I think we may be breaking the Singleton nature of the CuDevice class by exposing this constructor. I'm not completely sure about the implications of this yet, but it seems like we should remove this from the public interface if possible.

danpovey:
RE the constructor-- if you can figure out how to make it private and have
the code still compile, great

galv:
Okay, I'll look into the publicly exposed constructor issue elsewhere.

src/nnet3/convolution-cudnn.h
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Returns the size of the workspace required for each stage, in
+   * bytes (not 32-bit words).
+   */
+  size_t TempSpaceRequiredForward() const;
+  size_t TempSpaceRequiredBackwardData() const;
+  size_t TempSpaceRequiredBackwardFilter() const;
+
+  // Why aren't these const methods? That would make things a lot simpler
```
danpovey: RE the 'const' comments: it's fine to have them be const methods.

galv:
@daanpovey Check out the latest change I made. It documents the expected input formats of the member functions of the cudnn::ConvolutionComputation class. It also fixes some implementation details. If you have more questions about it, it's best just to ask me directly.

Regarding how to handle the cudnn dependency: What I've been trying to do is avoid any header files from having a dependency on cudnn.h (we can do this by forward-declaring the necessary types). This makes the cudnn dependency easy to disable. I don't have a strong opinion, though.

PR #2692, opened 2018-09-10
[src] Make cublas and cusparse use per-thread streams.

galv:
This will reduce synchronization overhead when we actually use multiple
cuda devices in one process go down drastically, since we no longer
synchronize on the legacy default stream.

More details here: https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -109,9 +109,10 @@ void CuDevice::Initialize() {
     cudaSetDevice(device_id_);
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
+    CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
     // Initialize the cuSPARSE library
     CUSPARSE_SAFE_CALL(cusparseCreate(&cusparse_handle_));
-
+    CUSPARSE_SAFE_CALL(cusparseSetStream(cusparse_handle_, cudaStreamPerThread));
   }
 }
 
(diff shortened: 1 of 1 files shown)


galv:
Unit tests pass. Running a tdnn model in librispeech for sanity, but I think it's fine since none of our current code depends on running with multiple GPUs in a single process yet.

danpovey:
Thanks!
I'll look at this to-morrow.
There is something else that I just realized needs to be done, which if you
are in the mood to code, you might want to start on.
A class like

template <class Real>
class PinnedMatrix: public MatrixBase<Real> {
 ...
};

in cudamatrix/pinned-matrix.h
which will be just like Matrix<Real> except for its allocator/deallocator.
To reduce allocation overhead you can write an extremely simple allocation
object (can be put in cu-allocator.h).  It would round any request for
pinned memory up to the next greater power of two, and allocate it directly
by calling CUDA functions; when such blocks are freed it would cache them
and supply them to future request for the same size block.  Thus, it would
be similar to the old CuAllocator code, but much simpler as there would be
no provision made for truly freeing the memory blocks.

Dan

galv:
Huh, okay. To clarify, "pinned" here means that that memory is never freed for the duration of the process, right? It is not the "pinned" memory which allows cudaMemcpyAsync and other various functions to access host memory by pinning memory to physical RAM, right?

Regardless, I don't see myself beginning on something like this until next weekend, if at all. Fair warning.

danpovey:
It's the pinned memory given by cudaMallocHost, which allows
cudaMemcpyAsync etc.
It also isn't freed for the duration of the process, but that's just to
simplify the allocator code.

danpovey:
Actually, I think the same mechanism with 'thread_id' as is already used in
cu-allocator.cc would need to be used to avoid data races when multiple
threads are used.

francisr:
Are you planning to have the training run on multiple GPUs in a single process?

danpovey:
No, there is no plan for that right now.  This is more about multi-threaded
decoders sharing the same GPU.
Maybe at some point in the future I'll look into better mechanisms for
model syncing for multi-GPU training, though.

src/cudamatrix/cu-device.cc
```diff
@@ -109,9 +109,10 @@ void CuDevice::Initialize() {
     cudaSetDevice(device_id_);
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
+    CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
```
danpovey: @galv, thanks.  I believe justin mentioned that we also need to use this cudaStreamPerThread for other API calls, like memcpy.

Issue #2663, opened 2018-09-01
Throw exception when I do LDA


YihengJiang:
Dear professor：
My experiments are speaker recognition. I do LDA after extracted vectors, I use the way like x-vector but change the neural network. It's unfortune that I get an error as follow(excerpts from lda's log file):

LOG (ivector-subtract-global-mean[5.4.1851-2fa7]:main():ivector-subtract-global-mean.cc:76) Read 200000 iVectors.
LOG (ivector-subtract-global-mean[5.4.1851-2fa7]:main():ivector-subtract-global-mean.cc:79) Norm of iVector mean was 2.13583
LOG (ivector-subtract-global-mean[5.4.1851-2fa7]:main():ivector-subtract-global-mean.cc:108) Wrote 200000 mean-subtracted iVectors
LOG (ivector-compute-lda[5.4.1851-2fa7]:main():ivector-compute-lda.cc:267) Read 200000 utterances, 0 with errors.
LOG (ivector-compute-lda[5.4.1851-2fa7]:main():ivector-compute-lda.cc:273) Computing within-class covariance.
LOG (ivector-compute-lda[5.4.1851-2fa7]:main():ivector-compute-lda.cc:278) 2-norm of iVector mean is 5.94342e-08
LOG (ivector-compute-lda[5.4.185~1-2fa7]:ComputeLdaTransform():ivector-compute-lda.cc:122) Stats have 6430 speakers, 200000 utterances.
### WARNING (ivector-compute-lda[5.4.185~1-2fa7]:Cholesky():tp-matrix.cc:110) Cholesky decomposition failed. Maybe matrix is not positive definite. Throwing error
Cholesky decomposition failed.

I try to understand Cholesky decomposition,and consider where is wrong,but it has no effective. So I want to ask for some help.
Now,I have an idea for this problem is that the training of my network did not sufficient. Cause I just train one epoch(84 egs file). In this case, I train network again(but result has not came out, now), at the mean time, I put this issue to you for some help. Dear professor, I want to know if it is the point of this problem, or it has another reason? Hope response,thank you!!!

danpovey:
Cholesky decomposition would fail if the covariance of your i-vectors was
not full rank.  We could fix the code to accept this kind of input, but
this shouldn't happen if you are doing everything right.  I suspect either
your network  has a linear bottleneck just before the end, of a dimension
less than the output dimension (which doesn't make sense as an i-vector
extractor) or something went badly wrong during training.

[...]

PR #2693, opened 2018-09-10
[src] improve error handling in ivector-compute-lda

david-ryan-snyder:
This PR resolves issue https://github.com/kaldi-asr/kaldi/issues/2663

The Cholesky decomposition is replaced with an eigen decomposition. We floor the eigen values to a small positive number (max_eigenvalue * std::numeric_limits<Real>::min()). 

FYI, @danpovey, @YihengJiang

### DIFF
--- a/src/ivectorbin/ivector-compute-lda.cc
+++ b/src/ivectorbin/ivector-compute-lda.cc
@@ -84,11 +84,23 @@ template<class Real>
 void ComputeNormalizingTransform(const SpMatrix<Real> &covar,
                                  MatrixBase<Real> *proj) {
   int32 dim = covar.NumRows();
-  TpMatrix<Real> C(dim);  // Cholesky of covar, covar = C C^T
-  C.Cholesky(covar);
-  C.Invert();  // The matrix that makes covar unit is C^{-1}, because
-               // C^{-1} covar C^{-T} = C^{-1} C C^T C^{-T} = I.
-  proj->CopyFromTp(C, kNoTrans);  // set "proj" to C^{-1}.
+  Matrix<Real> U(dim, dim);
+  Vector<Real> s(dim);
+  covar.Eig(&s, &U);
+  // Sort from greatest to smallest eigenvalue.
+  SortSvd(&s, &U);
+  // Floor eigenvalues to a small positive value.
+  Real floor = s(0) * std::numeric_limits<Real>::min();
+  int32 num_floored;
+  s.ApplyFloor(floor, &num_floored);
+  if (num_floored > 0) {
+    KALDI_WARN << "Floored " << num_floored << " eigenvalues of covariance "
+               << " to " << floor;
+  }
+  // Next two lines computes projection proj, such that
+  // proj * covar * proj^T = I.
+  s.ApplyPow(-0.5);
+  proj->AddDiagVecMat(1.0, s, U, kTrans, 0.0);
 }
 
 void ComputeLdaTransform(


danpovey:
thanks.  will check it to-morrow.

src/ivectorbin/ivector-compute-lda.cc
```diff
@@ -84,11 +84,23 @@ template<class Real>
...
-  C.Invert();  // The matrix that makes covar unit is C^{-1}, because
-               // C^{-1} covar C^{-T} = C^{-1} C C^T C^{-T} = I.
-  proj->CopyFromTp(C, kNoTrans);  // set "proj" to C^{-1}.
+  Matrix<Real> U(dim, dim);
+  Vector<Real> s(dim);
+  covar.Eig(&s, &U);
+  // Sort from greatest to smallest eigenvalue.
+  SortSvd(&s, &U);
+  // Floor eigenvalues to a small positive value.
+  Real floor = s(0) * std::numeric_limits<Real>::min();
```
danpovey: @david-ryan-snyder, this  floor is too small.. the problem is you take the square root and invert it before constructing the LDA matrix, and certain directions will be too large-- for roundoff reasons, if nothing else.  I suggest to make it configurable, with a default of 0.0001.  This will likely never make a difference for Kaldi recipes because our i-vectors and x-vectors have a reasonable distribution.   It may affect other scenarios, though.
(note: if this was set to 1.0, we'd have PCA, not LDA.)

src/ivectorbin/ivector-compute-lda.cc
```diff
@@ -209,14 +221,17 @@ int main(int argc, char *argv[]) {
...
+              covariance_floor = 1.0e-04;
     bool binary = true;
 
     po.Register("dim", &lda_dim, "Dimension we keep with the LDA transform");
     po.Register("total-covariance-factor", &total_covariance_factor,
                 "If this is 0.0 we normalize to make the within-class covariance "
                 "unit; if 1.0, the total covariance; if between, we normalize "
                 "an interpolated matrix.");
+    po.Register("covariance-floor", &covariance_floor, "Floor on the eigenvalues "
+                " of the interpolated covariance matrix");
```
danpovey: I'd add (relative to the highest eigenvalue)

david-ryan-snyder: Now it's directly configurable, rather than being multiplied with the max eigenvalue. 

I was finding that setting the floor to max_eigenvalue * 1.0e-04 was changing my results slightly. That's why I was using a std::numeric_limits::min() in the first place--to get a smaller floor.

danpovey:
And was it worse?  And how many did it say it floored?  And with which
recipe?  x-vector or i-vector?

david-ryan-snyder:
It's an x-vector system.

Using the larger floor made it slightly worse at 6/8 operating points I looked, and slightly better at 2/8 operating points. The relative differences are **very** small, like 1%. It's probably just noise, but I'd rather be cautious about changing results, since this option is almost never needed.

In my system, a floor of 1.0e-04 * max_eigenvalue floors 126 out of 512 eigenvalues to 0.00602884. The max eigenvalue is 38.4097.

danpovey:
OK.. see if 1.0e-06 is more similar to the baseline.

david-ryan-snyder:
@danpovey, thanks, that default seems fine. I updated the PR.

Issue #2694, opened 2018-09-11
Find a bug in src/nnet/nnet-lstm-projected.h


cbtpkzm:
`

    // apply the gradient clipping,
    if (clip_gradient_ > 0.0) {
      w_gifo_x_corr_.ApplyFloor(-clip_gradient_);
      w_gifo_x_corr_.ApplyCeiling(clip_gradient_);
      w_gifo_r_corr_.ApplyFloor(-clip_gradient_);
      w_gifo_r_corr_.ApplyCeiling(clip_gradient_);
      bias_corr_.ApplyFloor(-clip_gradient_);
      bias_corr_.ApplyCeiling(clip_gradient_);
      w_r_m_corr_.ApplyFloor(-clip_gradient_);
      w_r_m_corr_.ApplyCeiling(clip_gradient_);
      peephole_i_c_corr_.ApplyFloor(-clip_gradient_);
      peephole_i_c_corr_.ApplyCeiling(clip_gradient_);
      peephole_f_c_corr_.ApplyFloor(-clip_gradient_);
      peephole_f_c_corr_.ApplyCeiling(clip_gradient_);
      peephole_o_c_corr_.ApplyFloor(-clip_gradient_);
      peephole_o_c_corr_.ApplyCeiling(clip_gradient_);
    }
`
Here the clip_gradient_ should be grad_clip_?

danpovey:
fixed, thanks.

PR #2696, opened 2018-09-11
[src] Fix nnet1 proj-lstm bug where gradient clipping not used; thx:@cbtpkzm

danpovey:


danpovey:
Resolves #2694

vesis84:
Hi, is this 'issue' something that was in the code for a long time, or was it 'newly created' ?
Thanks, Karel.

danpovey:
Looks like it was there for a year at least.

PR #2702, opened 2018-09-13
Adding yomdle_fa egs

ChunChiehChang:


egs/yomdle_fa/v1/local/normalized_scoring/normalized_scoring.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This script provides normalized scoring for the various slam dataset languages.
```
hhadian: Could you please show me where this script is used in the recipe?

ChunChiehChang: Everything in the normalized_scoring dir was specifically to use the same normalization as Stephen. It's separate from the Kaldi stuff not actually used in the scripts if ./run_end2end.sh is run. I can remove the entire directory if need be.

aarora8: we can also, run the normalize scoring without ref_files. ref_files are the test utterances. in the normalize scoring script we can replace the path to ref_file with a path to data/test/text.

ChunChiehChang: Usually it's around 3-4% WER improvement. I removed the normalization that was done in the ./local/wer_output_filter. I can put it back in to remove the reliance on an external scoring script. This was mostly added to ensure that over the summer everyone was normalizing in the same way. Some of the stuff are things like "farsi vs arabic yeh" and "tatweel".

hhadian: I guess we can keep them and you can put them back in wer_output_filter.

ChunChiehChang: okay, I've added them back

ChunChiehChang:
I've also added the yomdle Chinese scripts to this PR. Should I have created a separate one?

danpovey:
I don't have an opinion; @hhadian might.

hhadian:
I guess that's OK (assuming we are planning on checking in all the yomdle recipes) to have them in the same PR.
But, it might be cleaner if all languages were subsets of `/egs/yomdle` (e.g. egs/yomdle/fa, egs/yomdle/zh, etc.)

PR #2704, opened 2018-09-13
[scripts] Fix to steps/cleanup/debug_lexicon.sh

danpovey:


danpovey:
This resolved a sorting-related issue.  It's not clear why it happened, but the fix is safe.

PR #2718, opened 2018-09-17
madcat arabic scripts cleaning and adding higher order language model

aarora8:


danpovey:
Thanks!  When @hhadian says it's good to go, I'll merge.

hhadian:
This looks good to me. 
Good improvement! BTW, how long does training the 6-gram LM take? I guess since it's BPE it shouldn't take too long, right?

aarora8:
Thank you, building 6-gram it was very fast as compared to the Tamil setup. I am not completely sure but i guess it is between 1-2 hrs.

PR #2721, opened 2018-09-19
Add encoding fix of #2680

quadr:
While applying this issue for the sliprob version, he missed it at `read_lexiconp` function.

### DIFF
--- a/egs/wsj/s5/utils/lang/make_lexicon_fst_silprob.py
+++ b/egs/wsj/s5/utils/lang/make_lexicon_fst_silprob.py
@@ -131,8 +131,9 @@ def read_lexiconp(filename):
     found_large_pronprobs = False
     # See the comment near the top of this file, RE why we use latin-1.
     with open(filename, 'r', encoding='latin-1') as f:
+        whitespace = re.compile("[ \t]+")
         for line in f:
-            a = line.split()
+            a = whitespace.split(line.strip())
             if len(a) < 2:
                 print("{0}: error: found bad line '{1}' in lexicon file {1} ".format(
                     sys.argv[0], line.strip(), filename), file=sys.stderr)


egs/wsj/s5/utils/lang/make_lexicon_fst_silprob.py
```diff
@@ -131,8 +131,9 @@ def read_lexiconp(filename):
     found_large_pronprobs = False
     # See the comment near the top of this file, RE why we use latin-1.
     with open(filename, 'r', encoding='latin-1') as f:
+        whitespace = re.compile("[ \t]+")
         for line in f:
-            a = line.split()
+            a = whitespace.split(line.strip())
```
danpovey: Thanks.  But we discovered a problem with doing line.strip(); it can remove character 255 I think which is part of the UTF-8 encoding range.  Can you change this and the other script to do line.rstrip('\n'), and test them?  (just to make sure they work in the normal case).

danpovey: Actually that's a more widespread problem, so I'll merge this and figure that problem out later.

quadr: How about this way?
```python3
with open(filename, 'r', encoding='utf-8') as f:
        for line in f:
            a = [ p.encode().decode('latin-1') for p in line.split() ]
```
I think this way doesn't have the problem that you mentioned.

danpovey:
The whole point of using the latin-1 encoding was that we want to be
agnostic about the encoding used as long as it's ASCII-compatible and the
spaces are also spaces in ASCII.  E.g. that way we can support GBK encoding.

danpovey:
Maybe you mistyped because in the code snipped you show you are reading in
utf-8.  If you assume that, you might as well do all the processing in
utf-8.

quadr:
```python3
with open(filename, 'r', encoding='latin-1') as f:
        for line in f:
            a = [ p.decode('latin-1') for p in line.encode('latin-1').split() ]
```
It only use latin-1 and ascii.

danpovey:
Oh, I see.  Yes, that is interesting.  I wonder if it is locale-independent, though?  I would hope so.. anyway we normally set the LC_ALL=C, but I'd like to know if the locale affects it.
It's a shame I just merged a fix.

quadr:
locale(LC_ALL) affects only printing non-ascii charactors.
If this way has a locale problem, the latest release also has a same problem.

danpovey:
Thanks for this, but since we resolved the issue a different way I'm not planning to merge this.  I'll bear the method in mind for the future though.

PR #2725, opened 2018-09-19
resolved conflicts for #2722

xiaohui-zhang:


egs/heroico/s5/run.sh
```diff
@@ -179,4 +179,8 @@ if [ $stage -le 12 ]; then
   local/chain/run_tdnn.sh
 fi
 
+if [ $stage -le 13 ]; then
```
danpovey: this should be removed too.

egs/tunisian_msa/s5/local/chain/tuning/run_tdnn_1a.sh
```diff
@@ -0,0 +1,282 @@
+#!/bin/bash
+
```
danpovey: There should be results  here (and a compare_wer.sh script to compare them.).
@xiaohui-zhang, would you mind either running it yourself locally or asking one of the newer guys to do the same?  It would be good experience for them.

xiaohui-zhang: sure I'll take care of it

xiaohui-zhang:
ready for merging @danpovey

PR #2728, opened 2018-09-21
spelling mistake

xu-song:


danpovey:
thanks but I don't want to bulk up the commit history with this-- it's too trivial and the script is anyway outdated.

PR #2739, opened 2018-09-26
Randomization modification to unite the random numbers generated from different machines

MohamadIsmail:
I'm currently working on ASR system and I have noticed that the RandUniform() function in src/math/kaldi-math.cc generates different random numbers from different machines(training machine and testing machine). this led us to generating different mfcc features and -in the end- different results so we came with a better idea -I think- to generate the same set of random real numbers on different environments.


MohamadIsmail:
They're pretty much the same performance with a slight edge to my version on so many iterations.

```c++
for (size_t i = 0; i < 300000000; i++)
{
	random_numbers[i] = distribution(generator);
}
```
```c++
for (size_t i = 0; i < 300000000; i++)
{
       random_numbers[i] = static_cast<float>((Rand(state) + 1.0) / (RAND_MAX+2.0));
}
```

`Rand()`: 2.566
`std::uniform_real_distribution<double>`: 2.445

MohamadIsmail:
I can't understand the CI log. Is it related to some kind of timeout?

danpovey:
it's probably OK-- it often fails for no reason.

danpovey:
But check that all the tests work for you.  I'm rerunning the tests.  It
failed in an odd way.

francisr:
You're not setting the generator with the seed.

danpovey:
@MohamadIsmail, I don't think I can merge this unless you figure out how to set the random seed and figure out parallel-programming issues.  I don't know much about this area.  The reason why we needed the RandomState object was mainly because each time you call rand(), it locks a mutex, and it can get very slow if multiple threads do that.

danpovey:
Closing for now as those comments were un-addressed.  I'm open to doing it this way in future, but I think it might be better to have the 'default_random_engine' be the only  thing in the RandomState, and having a  separate distribution object... I'm not sure how it's supposed to be done though.

PR #2740, opened 2018-09-26
[egs][aishell2] turn off new word discovery in word segmentation stage

dophist:
word segmentation was setup with new word discovery(Jieba's default), this commit turns it off so that there will be less OOVs in training text.


danpovey:
If this affects results, can you update the results?

danpovey:
How have you tested this?  E.g. how does the word list size change?  I want to make sure it doesn't break something.

dophist:
Yes I re-run the data-prep stage, the only difference is that there are a few hundred less <UNK> in word segmented training text now, nothing else changed.  The CER difference introduced by this should be less than a different random seed.

Issue #2743, opened 2018-09-29
Cuda 10 support


leomao:
Cuda 10 was released. The support gencode values are:
```
'compute_30','compute_32','compute_35',
'compute_37','compute_50','compute_52','compute_53','compute_60','compute_61',
'compute_62','compute_70','compute_72','compute_75','sm_30','sm_32','sm_35',
'sm_37','sm_50','sm_52','sm_53','sm_60','sm_61','sm_62','sm_70','sm_72','sm_75'.
```

danpovey:
Thanks.
If you want to try to get Kaldi to compile with CUDA 10, and maybe make the
changes to the Makefile and configure script to get it to work, I'd look at
your PR.  Until someone has got it to compile and run, we won't know if
additional things need to be done.
Dan

PR #2747, opened 2018-10-02
update configure to support cuda 10

leomao:
Related issue: #2743
It seems that it just compiled without any error. But I am not sure if there is something broken.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -407,7 +407,7 @@ function configure_cuda {
           MIN_UNSUPPORTED_GCC_VER="7.0"
           MIN_UNSUPPORTED_GCC_VER_NUM=70000;
         ;;
-        9_2 | 9_*)
+        9_2 | 9_* | 10_*)
           MIN_UNSUPPORTED_GCC_VER="8.0"
           MIN_UNSUPPORTED_GCC_VER_NUM=80000;
         ;;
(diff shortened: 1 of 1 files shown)


jtrmal:
What would be also good to figure out is the range of gcc compilers cuda 10 supports.

leomao:
cuda 10 suppports gcc from 4.4.7 to 7.3.1 according to this [table](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#system-requirements). 
 Also, in `cuda/include/crt/host_config.h`, it explicitly reject to use gcc with version greater than 7.

jtrmal:
ok, great, in that case, no other changes necessary, I guess. I'll let Dan
merge/decide.
Thanks for the PR!
y.

ououkhin63:
Hi 
Now, I am using cuda =10 and gcc version is  7.3.0.
I am trying to compile kaldi with gpu.But I can't. When I run ./configure, it shows like that
Configuring ...
Checking compiler g++ ...
Checking OpenFst library in /home/neuro/kaldi/tools/openfst ...
Atlas found in /usr/lib/x86_64-linux-gnu
Validating presence of ATLAS libs in /usr/lib/x86_64-linux-gnu
Using library /usr/lib/x86_64-linux-gnu/liblapack.so as ATLAS's CLAPACK library.
Unsupported CUDA_VERSION (CUDA_VERSION=10_0), please report it to Kaldi mailing list, together with 'nvcc -h' or 'ptxas -h' which lists allowed -gencode values...

PR #2748, opened 2018-10-02
Use get_python_inc and print function in install scripts

leomao:
Use the function form of `print` to support both python 2 and 3.
Use get_python_inc from sysconfig, this function is simpler.

### DIFF
--- a/tools/extras/install_mmseg.sh
+++ b/tools/extras/install_mmseg.sh
@@ -18,7 +18,7 @@ if ! $(python -c "import distutils.sysconfig" &> /dev/null); then
     echo "Proceeding with installation." >&2
 else
   # get include path for this python version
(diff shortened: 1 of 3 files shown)


danpovey:
Thanks.  However, I think for sequitur we have to be a bit more careful-- you actually cannot use sequitur if python resolves to python3, I believe.  It might be possible to somehow rejig it to invoke python2 directly, but that would be complicated and @jtrmal would have to help.  For that one, could you just make it so the installation script, if it detects that python is really python3, just dies with a suitable error message.

danpovey:
@jtrmal, any comment?

jtrmal:
Actually, looking at the code, I think there is the statement "from
__future__ import print_function" missing in the oneliners.
Y.

danpovey:
I think it works in this case as there's just one arg to the print
function, which python2.7 would treat as an expression with parentheses
around it.

PR #2761, opened 2018-10-05
Fixed rm recipe not working

keighrim:
The recipe stops running after first delta training as [`local/test_decoders.sh`](https://github.com/kaldi-asr/kaldi/blob/2d163f4df27b8b7b6e989e54b8b863b6545a7d87/egs/rm/s5/local/test_decoders.sh) fails with a message;  
`Decoders were not equivalent, check exp/tri1/decode/tmp/equivalent_2_3.log and contact maintainers`

Log says [`gmm-decode-simple`] expects "vector" type of fst, so this fix passes a converted `HCLG.fst`.


danpovey:
Actually, I think a better fix would be to change gmm-decode-simple so that it can read the right FST type.  It's probably just a question of changing which wrapper is called to read the FST.  Would you mind figuring that out?

danpovey:
@desh2608 would you mind doing this if @keighrim doesn't get to it?

desh2608:
Ok I'll look into it.

keighrim:
I'm not a c/cpp programmer, and not sure how to change `gmm-decode-simple` to swap the wrapper. Thanks for taking charge of this @desh2608 !

desh2608:
Fixed in this PR: https://github.com/kaldi-asr/kaldi/pull/2787

danpovey:
Thanks-- fixed another way, so closing.

PR #2769, opened 2018-10-08
Update ami_download.sh

puneetbawa:
URL for the corpus was pointing towards a file which doesn't exist. Changed that to a valid one.

### DIFF
--- a/egs/ami/s5/local/ami_download.sh
+++ b/egs/ami/s5/local/ami_download.sh
@@ -53,8 +53,8 @@ cat local/split_train.orig local/split_eval.orig local/split_dev.orig > $wdir/am
 wgetfile=$wdir/wget_$mic.sh
 
 # TODO fix this with Pawel, files don't exist anymore,
-manifest="wget --continue -O $adir/MANIFEST.TXT http://groups.inf.ed.ac.uk/ami/download/temp/amiBuild-04237-Sun-Jun-15-2014.manifest.txt"
-license="wget --continue -O $adir/LICENCE.TXT http://groups.inf.ed.ac.uk/ami/download/temp/Creative-Commons-Attribution-NonCommercial-ShareAlike-2.5.txt"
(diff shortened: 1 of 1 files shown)


egs/ami/s5/local/ami_download.sh
```diff
@@ -53,8 +53,8 @@ cat local/split_train.orig local/split_eval.orig local/split_dev.orig > $wdir/am
 wgetfile=$wdir/wget_$mic.sh
 
 # TODO fix this with Pawel, files don't exist anymore,
-manifest="wget --continue -O $adir/MANIFEST.TXT http://groups.inf.ed.ac.uk/ami/download/temp/amiBuild-04237-Sun-Jun-15-2014.manifest.txt"
-license="wget --continue -O $adir/LICENCE.TXT http://groups.inf.ed.ac.uk/ami/download/temp/Creative-Commons-Attribution-NonCommercial-ShareAlike-2.5.txt"
+manifest="wget --continue -O $adir/MANIFEST.TXT http://groups.inf.ed.ac.uk/ami/download/temp/amiBuild-0153-Tue-Oct-2-2018.manifest.txt"
+# license="wget --continue -O $adir/LICENCE.TXT http://groups.inf.ed.ac.uk/ami/download/temp/Creative-Commons-Attribution-NonCommercial-ShareAlike-2.5.txt"
```
danpovey: If this isn't obtained you should remove the echo statement further down that mentions the license.

ervaibhavkumar:
@puneetbawa Please do that echo part also

ervaibhavkumar:
@danpovey As LICENCE.TXT does not exist on the server, shouldn't it be removed / commented ?

danpovey:
If it doesn't exist then remove the lines referring to it.  There is an echo statement at the bottom of that script that you need to remove also.

puneetbawa:
@danpovey The changes have been committed. Please review and let me know if any issues.
Thanks

PR #2772, opened 2018-10-08
Total number of words for WER calculation was incorrect.

siddalmia:
It did not remove the special symbol completely (The quotes ('') were left behind).


danpovey:
@sw005320 please check

sw005320:
Thanks @siddalmia, could you give me an example of where the number becomes different? I usually checked the number of total word counts carefully for all cases, and I did not found problems, but your way seems to be more elegant.

siddalmia:
Hi @sw005320 , You should get a wrong answer whenever there is an insertion error because then the ref would contain  `'***'` but your `sed` is removing only the `***` and leaving `''` behind. So the total number of reference words would be more than it should be.

sw005320:
OK. It may be come from the difference of the format of `per_utt`. When I tested this script, it used `***` (not `'***'`), and there was no problem (CHiME-5 official score was also no problem). And yes, your modification must be correct anyway. Thanks for pointing out the bug.

@danpovey, could you merge this PR?

Issue #2707, opened 2018-09-14
SAD produces consecutive segments


stanleyguan:
I have been using `egs/wsj/s5/steps/segmentation/detect_speech_activity.sh` and a pre-trained model to do SAD on recordings. One thing I noticed is that the resulting segments may be consecutive. As an example, an utterance might be

```
291.30 296.91 k e c g u at email dot com
```

while the SAD may produce segments as

```
291.22 292.39
       292.39 292.99
              292.99 293.85
                     293.85 294.66
                            294.66 295.40
                                   295.40 297.05
```

where there are no gaps between the segments and it makes more sense in this case to have just a single combined segment.

My understanding is that this has to do with the padding of the segments (defaults to 0.2 seconds on both ends). Before padding, the segments probably had gaps in between. From an end-result point of view though I don't see much value in keeping these separate since there is no way to go back to the original segments (you can't just subtract the paddings since the padding is clipped to not overlap with adjacent segments). As it is [currently implemented](https://github.com/kaldi-asr/kaldi/blob/eef2a093a6de35aca9cdf773b0e443c6a6bebaa2/egs/wsj/s5/steps/segmentation/internal/sad_to_segments.py#L161-L196) the padding choice is arbitrary for this situation since the earlier segment is padded first and if that extends beyond the next segment, that next segment won't be padded at the start. For ASR purposes it can also be beneficial to have a single segment so that the model can pool information together. Therefore it might make sense to just merge the segments if there are no gaps, or at least have an option to do so.

I'm happy to make a PR if this is deemed desirable.

danpovey:
Hm.  I think it would make sense to merge consecutive segments as long as they are less than, say, 7 seconds long.  @vimalmanohar  what do you think about this?

vimalmanohar:
I think this is an application specific issue. We decided not to do any post-processing. Combining segments together can also hurt the performance e.g. if they are from different sentences. Although it might help in this particular utterance to combine. You can get longer segments by tuning the graph_opts parameters.

danpovey:
I think it's still interesting to have a look at the effect of this simple
post-processing step though, and maybe make it an option that's off by
default.  I think it may be useful for quite a few scenarios.

danpovey:
yes, that would be good.

chenfuouc:
May I ask which example egs used the scripe to do SAD?

vimalmanohar:
[...]

PR #2776, opened 2018-10-09
Enhancements & minor bugfix to segmentation postprocessing

stanleyguan:
Addresses #2707 

* Option to merge consecutive segments provided the combined segment is no longer than a given duration
* Option to filter out segments shorter than a given duration
* Bugfix to post-processing stats computation, with also added stats related to the items above
* Modified `detect_speech_activity.sh` to allow relevant arguments to be passed to `sad_to_segments.py`

The PR should leave the behavior of existing scripts unchanged.


egs/wsj/s5/steps/segmentation/detect_speech_activity.sh
```diff
@@ -57,6 +57,9 @@ acwt=0.3
 transform_probs_opts=""
 
 segment_padding=0.2   # Duration (in seconds) of padding added to segments 
+min_segment_dur=0   # Filter out segments that are shorter than this
```
danpovey: These options seem useful to me and I think probably make sense to include.  However, I think there is an option when you build the HMM, that controls the minimum duration of segments, and I think the documentation for this option should mention that, so that the users don't think this is the only (or primary) way to control the minimum segment duration.  Also the documentation for these options is also given below, in the usage message, so you should duplicate it there.   And please make the documentation, both here, in the usage message, and in the
python script, more explicit.  For example, mention that the unit is seconds, and that consecutive segments are merged only if they are touching.

stanleyguan:
@danpovey Hopefully the comment has been addressed.

egs/wsj/s5/steps/segmentation/detect_speech_activity.sh
```diff
@@ -56,7 +56,13 @@ acwt=0.3
 # e.g. --speech-in-sil-weight=0.0 --garbage-in-sil-weight=0.0 --sil-in-speech-weight=0.0 --garbage-in-speech-weight=0.3
 transform_probs_opts=""
 
+# Postprocessing options
 segment_padding=0.2   # Duration (in seconds) of padding added to segments 
+min_segment_dur=0   # Minimum duration (in seconds) required for a segment to be included
+                    # This is before any padding. This is an alternative to --min-speech-duration above.
```
danpovey: Please clarify that speech segments with length below this (before padding) will be discarded.  And for merge_consecutive_max_dur, you should probably clarify that this is after padding by --segment-padding seconds.

stanleyguan:
Additional clarification added.

danpovey:
Thanks!  @vimalmanohar  can you please look through it?  If you say it's OK I'll merge it.

egs/wsj/s5/steps/segmentation/internal/sad_to_segments.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
```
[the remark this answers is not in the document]
stanleyguan: Missed the permission change. Now fixed.

PR #2778, opened 2018-10-09
[not-for-merge] [src] Adding binary lattice-to-ngram-counts

david-ryan-snyder:
This PR adds a binary called lattice-to-ngram-counts that computes n-label softcounts over input lattices.

Note that this binary may need to be used in conjunction with lattice-expand-ngram (https://github.com/kaldi-asr/kaldi/blob/master/src/latbin/lattice-expand-ngram.cc) so that each arch as a unique n-label history. 

The PR is currently "not-for-merge" but may be added to the main branch at some point.

@m-wiesner

### DIFF
--- a/src/lat/lattice-functions.cc
+++ b/src/lat/lattice-functions.cc
@@ -6,6 +6,7 @@
 //                2013  Cisco Systems (author: Neha Agrawal) [code modified
 //                      from original code in ../gmmbin/gmm-rescore-lattice.cc]
 //                2014  Guoguo Chen
+//                2018  David Snyder
 
 // See ../../COPYING for clarification regarding multiple authors
 //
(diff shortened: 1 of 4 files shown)


david-ryan-snyder:
@m-wiesner was this code, or something similar to it, ever added to the main branch?

kkm000:
@david-ryan-snyder, @m-wiesner, where we are at this?

kkm000:
@david-ryan-snyder, @m-wiesner, friendly ping?

danpovey:
I think  it's OK to merge but it's not used in any recipes.  We can just leave the PR up, I think, in case people need it, and if it seems useful for anything we want to commit we can merge it.

kkm000:
Agree, it makes no sense to maintain unused code.

PR #2779, opened 2018-10-10
Add support for context independent phones in gmm-init-biphone (for e2e)

hhadian:
This PR adds a feature to gmm-init-biphone so it can treat some phones (e.g. silence phones) as context independent. 
@aarora8 and @ChunChiehChang, if you have time, it would be good to test this using some of the OCR recipes. Please let me know if you will have time for this.


aarora8:
Thank, i will test it with rimes dataset. Should compare it with a baseline or something similar.

aarora8:
ok, thank you, i will do it.

ChunChiehChang:
I'll probably run this on the iam egs as well.

hhadian:
Based on Jonathan's and Ashish's results, this does not give WER improvements but the objf is improved. I saw similar results on ASR tasks. 
But I think it might be useful to have this feature, anyway.

danpovey:
OK, merging.

PR #2796, opened 2018-10-19
[src] Adding rir-simulate program

funcwj:
I implement rir-simulate reference from [https://github.com/ehabets/RIR-Generator](https://github.com/ehabets/RIR-Generator) which could got same result as that. For example, 
```
rir-simulate --sound-velocity=340 \
  --samp-frequency=16000 --receiver-location=2,1.5,2 \
  --source-location=2,3.5,2 --room-topo=5,4,6 \
  --beta=0.4 --number-samples=4096 rir1.wav

rir-simulate --sound-velocity=340 \
  --samp-frequency=16000 --receiver-location=2,1.5,2 \
  --source-location=2,3.5,2 --room-topo=5,4,6 \
  --beta=0.4 --number-samples=2048 --order=2 \
  --microphone-type=omnidirectional \
  --angle=0 --hp-filter=true rir2.wav
```
produce same results with example_{1,2}.m

### DIFF
--- a/src/feat/rir-generator.cc
+++ b/src/feat/rir-generator.cc
@@ -1,0 +1,247 @@
+// feat/rir-generator.cc
+
+// Copyright 2018 Jian Wu
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+
+#include "feat/rir-generator.h"
+
+namespace kaldi {
+
+void RirGenerator::ComputeDerived() {
+    if (!str_to_pattern_.count(opts_.microphone_type))
+        KALDI_ERR << "Unknown option values: --microphone-type=" << opts_.microphone_type;
+    KALDI_ASSERT(frequency_ >= 1);
+    KALDI_ASSERT(velocity_ >= 1);
+    KALDI_ASSERT(order_ >= -1);
+
+    // Process room topo
+    KALDI_ASSERT(opts_.room_topo != "" && "Options --room-topo is not configured");
+    std::vector<BaseFloat> topo;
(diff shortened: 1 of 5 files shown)


danpovey:
Great!  @david-ryan-snyder, can you please take a look?
@funcwj, please bear in mind the Google C++ style guide; I notice for starters that you have lines >>80chars, and an indent level of 4 instead of 2, but there may be other issues.  I believe there is a cpplint.py somewhere in the repo that you could check it with.

funcwj:
Sorry, I'm used to use tab indent, and I will fix the format problem.

src/feat/rir-generator.cc
```diff
@@ -0,0 +1,247 @@
...
+    KALDI_ASSERT(frequency_ >= 1);
+    KALDI_ASSERT(velocity_ >= 1);
+    KALDI_ASSERT(order_ >= -1);
+
+    // Process room topo
+    KALDI_ASSERT(opts_.room_topo != "" && "Options --room-topo is not configured");
+    std::vector<BaseFloat> topo;
+    KALDI_ASSERT(SplitStringToFloats(opts_.room_topo, ",", false, &topo));
+    room_dim_ = topo.size();
+    // KALDI_ASSERT(room_dim_ == 2 || room_dim_ == 3);
```
david-ryan-snyder: I'm guessing you meant to remove this line.

src/feat/rir-generator.cc
```diff
@@ -0,0 +1,247 @@
...
+            angle_tmp.push_back(0.0);
+    } else {
+        KALDI_ASSERT(SplitStringToFloats(opts_.orientation, ",", false, &angle_tmp));
+    }
+    if (angle_tmp.size() == 1)
+        angle_tmp.push_back(0.0);
+    KALDI_ASSERT(angle_tmp.size() == 2);
+    angle_.swap(angle_tmp);
+
+    // Process beta
```
david-ryan-snyder: I think this section needs some comments. What is 6 for?

david-ryan-snyder:
In addition to what Dan said about style, I think it would be a good idea to add more substantial comments describing the library code in feat/. E.g., I think you should relate this to the matlab code this is based on (you only mentioned that in the binary), what the functions are doing, etc.

For some inspiration, here's an example of a reasonably well documented .h file: https://github.com/kaldi-asr/kaldi/blob/master/src/ivector/plda.h.

funcwj:
In fact, I suppose some tasks requires simulated rirs with customized configures, and using matlab code is not very convenient to do it, because some server may not have it, so I implement it based on kaldi. 
Now I‘m working on some enhancement tasks, so this requirement do not shown mach in kaldi's egs. My idea is using some python scripts to generate a rir database with customized configures instead of using matlab, which is similar to [https://github.com/tomkocse/sim-rir-preparation](https://github.com/tomkocse/sim-rir-preparation)

src/feat/rir-generator.cc
funcwj: beta means T60 or reflection coefficients for each space(3D room), so dim of it is 1 or 6.

funcwj:
Thanks a lot for giving suggustions, I will update comments in the future.

danpovey:
@phanisankar-nidadavolu can you please look carefully at this code?   You may find this is the easiest way to get RIRs.  @vimalmanohar just pointed out to me that this PR existed-- very convenient.

PR #2800, opened 2018-10-24
Simplifying AMI recipe

vimalmanohar:
Simplying AMI recipe (only the one with data augmentation)
1. LDA transformation is replaced by PCA for i-vector training
2. An extra alignment stage is removed and alignments is obtained from the supervision lattice best path.
3. Short segments are not combined to get 1.5s long segments. Instead, multiple smaller chunk-widths are used in training.

WER results are almost the same.

### DIFF
--- a/egs/ami/s5b/local/chain/multi_condition/tuning/run_tdnn_lstm_1a.sh
+++ b/egs/ami/s5b/local/chain/multi_condition/tuning/run_tdnn_lstm_1a.sh
@@ -19,7 +19,6 @@ set -e -o pipefail
 stage=0
 mic=ihm
 nj=30
-min_seg_len=1.55
 use_ihm_ali=false
 train_set=train_cleaned
 gmm=tri3_cleaned  # the gmm for the target data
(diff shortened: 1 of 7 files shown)


egs/ami/s5b/local/chain/multi_condition/tuning/run_tdnn_lstm_1a.sh
```diff
@@ -109,13 +104,13 @@ local/nnet3/multi_condition/run_ivector_common.sh --stage $stage \
 local/nnet3/prepare_lores_feats.sh --stage $stage \
                                    --mic $mic \
                                    --nj $nj \
-                                   --min-seg-len $min_seg_len \
+                                   --min-seg-len "" \
```
danpovey: wouldn't it be cleaner to remove the min-seg-len option from this script entirely?  The run_ivector_common.sh script already doesn't seem to support it.

But I'm concerned that this PR will break the other chain example scripts and the other nnet3 example scripts.

vimalmanohar:
The prepare_lores_feats.sh is common to when using data augmentation (in
multi_condition) and also when not. The run_ivector_common.sh has separate
versions for the two cases. I only updated the version using data
augmentation, which is only one tdnn+lstm script. The min-seg-len option is
retained for compatibility with the recipes that don't use data
augmentation.

PR #2806, opened 2018-10-30
Fix openblas build for aarch64

ryanleary:
Aarch64 builds with openblas tries to use the sse and sse2 compiler directives. Add a tailored aarch64 makefile.


danpovey:
Merging, since it doesn't look like it could break anything that might have previously worked.

Issue #2814, opened 2018-11-02
Should this be a bug for GeneralDropoutComponent?


GaofengCheng:
In my understanding, this should be a bug.
See: https://github.com/kaldi-asr/kaldi/blob/master/src/nnet3/nnet-general-component.cc#1699
`    ans->Scale(1.0 / dropout_proportion); ---->
    ans->Scale(1.0 / (1.0- dropout_proportion));
`

GaofengCheng:
@danpovey

PR #2815, opened 2018-11-02
[src] Fixing the scaling bug of the general-dropout component

GaofengCheng:
fix the scale factor bug for sequence-shared dropout

related to #2814


GaofengCheng:
@danpovey 
Hi Dan, you may be interested in this PR. This may solve a bug in GeneralDropoutComponnet, which will lead the breakdown of training (when we set a small dropout value like 0.1). This does not influence the continuous dropout.

danpovey:
Hm, interesting.
I suspect in setups with batch-norm (which is most setups),this would not have made much of a difference in practice.  But it might depend where exactly the batchnorm is.
I'll merge it.

PR #2816, opened 2018-11-03
[WIP] Gale Arabic BPE

aarora8:
It is updating the current word based Gale Arabic recipe to a byte pair encoding based recipe.

- BPE based model with vocab size=700
- It is building a 6-gram sub-word language model and a with RNNLM. Both language models are currently built with training text.
- updating I-vector extraction routine (replacing the LDA+MLLT with PCA, and removing the combine_short_segments.sh)
- updating nnet training setup, with TDNN-F
- adding run_chain_common.sh routine
- minor cleanup

our current results with the word-based and bpe based setup are as follows:
Word-based WER: 16.37
BPE based WER: 16.58
@hhadian @DongjiGao


francoishernandez:
Hi,
This looks very interesting. Do you have a paper or some report somewhere going into some details about such a system?
Thanks!
François

danpovey:
Not yet; it's very much in flux.
Basically, we can get about the same WER with the BPE-based system as with
the baseline system, using n-gram LMs.  I think we configure it to use a
few thousand word-pieces.
Initial experiments with RNNLMs seemed to show they didn't interact well
with BPE, but I believe that may have been due to a too-small ngram-order
approximation in the lattice-rescoring; we are experimenting with that.

aarora8:
Our WER for BPE-based setup is slightly worse (WER difference: 0.58) as compared to word-based setup. Currently, the best WER after RNNLM rescoring for both setups are 15.29(BPE-based) and 14.71(word-based). In RNNLM rescoring, increasing the n-gram order from 7 to 12 was degrading the performance for BPE-based setup. As suggested, I was going to print best-path likelihood and check whether it gets better or worse with an increase in n-gram order.  To check if it's a search problem or a model problem. Sorry due to the thesis deadline I was not able to investigate further. I will surely do it today.

PR #2817, opened 2018-11-05
New AMI recipes

vimalmanohar:


jtrmal:
@danpovey I think this is ok to merge -- it adds just some new scripts, shouldn't interact with anything else. We will be probably  using the scripts, so in case there is some problem, we will notice and report it.

PR #2819, opened 2018-11-06
[build] Add cub library for cuda projects

ryanleary:
This adds the [NVIDIA CUB library](https://nvlabs.github.io/cub/) to the tools/dependencies group and makes it available to modules using cuda.

Future PRs are going to require features from this library. In an effort to make that patchset easier to review, we're breaking it up into smaller pieces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,6 @@ GSYMS
 /tools/mmseg-1.3.0.tar.gz
 /tools/mmseg-1.3.0/
 /kaldiwin_vs*
+/tools/cub-1.8.0.zip
+/tools/cub-1.8.0/
+/tools/cub
--- a/src/configure
+++ b/src/configure
@@ -863,6 +863,7 @@ android=false
 MATHLIB='ATLAS'
 ATLASROOT=`rel2abs ../tools/ATLAS_headers/`
 FSTROOT=`rel2abs ../tools/openfst`
+CUBROOT=`rel2abs ../tools/cub`
 
 # Save the command line to include in kaldi.mk
 cmd_line="$0 $@"
@@ -954,6 +955,9 @@ do
     shift ;;
   --fst-root=*)
     FSTROOT=`read_dirname $1`;
+    shift ;;
+  --cub-root=*)
+    CUBROOT=`read_dirname $1`;
     shift ;;
   --clapack-root=*)
     CLAPACKROOT=`read_dirname $1`;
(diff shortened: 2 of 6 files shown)


danpovey:
Thanks!
@huangruizhe, please check whether this installs cleanly on the CLSP grid, and also on a mac (please find someone else to do it, or help if you don't have one, or haven't installed xcode).

ryanleary:
Want me to add the quiet flag?

danpovey:
yes please, I think it's a good idea.

ryanleary:
Done.

ryanleary:
@danpovey could you merge this one too? Trying to get a clean branch of the other work prepared for you.

Worldexe:
In patch `741b75b474a03a584e64dec36e96a47bcc2a1117` you added `-lcuda` library, that actually lives inside nvidia driver, not cuda library.

So, now we actually need nvidia drivers to build cuda-enabled kaldi?

The scenario is like this: I have nvidia cards in cluster, but kaldi is being built on machine without them. Also, there are machines in the cluster without nvidia cards (to run cpu-only kaldi tasks). So, now I have to install nvidia drivers on all the machines?

danpovey:
Oh.  I think you are right.  That is a problem.  Thanks for pointing it out. Can you test whether, if you just remove -lcuda from makefiles/cuda_64bit.mk and makefiles/cuda_32bit.mk, it works?  @ryanleary  why did you add -lcuda?

danpovey:
@btiplitz I think you noticed this issue to.

ryanleary:
Good spot. Two things going on here: 
1. The addition of the flag leaked from a larger patchset, that shouldn't have been included here. 
2. Changing the flag to `-cudart` is probably sufficient (which links against the runtime, only requiring the SDK be installed and not the drivers). I will confirm and update the PR.

danpovey:
The flag -lcudart was already supplied.
I hope the flag was not really needed to get the CUDA decoder to work.  If
it's needed, it should only be included in the binaries that really need
it-- it will break things if included for things that only optionall use
CUDA.

luitjens:
I suspect it was added when we were playing around with multi-context drive
API device selection.  We removed that code but never removed the link
flag.  As you suggested we should just try and build without it and see if
there is any linking errors.

ryanleary:
Validated that cuda decoder will build correctly without that flag. Sorry for the trouble.

btiplitz:
Thanks for fixing this.  This did cause problems as we build on systems without cuda cards then run on systems with them.  I did add the necessary packages

PR #2821, opened 2018-11-06
[src] Use cuda streams in matrix library

ryanleary:
When multiple threads are used within Kaldi, cudamatrix will now leverage an independent cuda stream for that thread's matrix operations. This will improve pipelining and GPU utilization in multi-threaded cases.

### DIFF
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -139,8 +139,8 @@ void CuArray<T>::CopyFromArray(const CuArrayBase<T> &src) {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) {
     CuTimer tim;
-    CU_SAFE_CALL(cudaMemcpy(this->data_, src.data_, this->dim_ * sizeof(T),
-                            cudaMemcpyDeviceToDevice));
+    CU_SAFE_CALL(cudaMemcpyAsync(this->data_, src.data_, this->dim_ * sizeof(T),
+                            cudaMemcpyDeviceToDevice, cudaStreamPerThread));
     CuDevice::Instantiate().AccuProfile(__func__, tim);
   } else
 #endif
@@ -158,8 +158,8 @@ void CuArrayBase<T>::CopyFromArray(const CuArrayBase<T> &src) {
   if (CuDevice::Instantiate().Enabled()) {
     CuTimer tim;
     CU_SAFE_CALL(
-        cudaMemcpy(this->data_, src.data_, dim_ * sizeof(T),
-                   cudaMemcpyDeviceToDevice));
+        cudaMemcpyAsync(this->data_, src.data_, dim_ * sizeof(T),
+                   cudaMemcpyDeviceToDevice, cudaStreamPerThread));
     CuDevice::Instantiate().AccuProfile(__func__, tim);
   } else
 #endif
--- a/src/cudamatrix/cu-matrix.cc
+++ b/src/cudamatrix/cu-matrix.cc
@@ -229,8 +229,9 @@ void CuMatrixBase<Real>::CopyFromMat(const CuMatrixBase<OtherReal> &M,
       MatrixIndexT dst_pitch = stride_ * sizeof(Real);
       MatrixIndexT src_pitch = M.Stride() * sizeof(Real);
       MatrixIndexT width = M.NumCols() * sizeof(Real);
-      CU_SAFE_CALL(cudaMemcpy2D(data_, dst_pitch, M.data_, src_pitch,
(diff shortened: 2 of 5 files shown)


danpovey:
Thanks!  Please observe the 80-character-per-line limit, even if the existing code was not always compliant.  And try to make sure the code is justified correctly (e.g. lines up with the open-parenthesis on the previous line).

ryanleary:
Pushed fix for line width.

danpovey:
Thanks.  I might not merge this right away because I have to do some careful thinking about whether this might break existing code, and if so the best way to fix it.  But it does make sense to make this change.

luitjens:
Hi Dan,

I had thought carefully about whether this might break existing code and I don't believe it will.  Here is why.

All the changes are to DeviceToDevice copies thus in order to verify that there is no race conditions we need to consider three cases:

1)  Work in the same stream
2)  Work in stream 0
3   Work in other streams.

1)  This is automatically handled in hardware by stream semantics.
2)  This still works as the global default stream is synchronous with these streams
3)  As far as I can tell Kaldi doesn't use other streams.  I greped the code for cudaStreamCreate and do not see it anywhere. 

Thus I firmly believe there are no race conditions here.  Maybe we should have DanG see if he agrees with my assessment.

luitjens:
> I saw nvidia did the same changes. If these operations are done async, then there needs to be some call to ensure these background tasks have completed right ? Otherwise the memory might not be complete when accessed ?

Please see my earlier response.  For all cases in Kaldi this should not be an issue.  Stream synchronization is already handled in the hardware and does not require host synchronization.  Where we would have a potential race condition is if we were copying from device to host.  In this case we would want to still do the async copy in the per-stream default thread but then immediately call cudaStreamSynchronize(cudaDefaultStreamPerThread);  This way we would avoid synchronizing other streams.

btiplitz:
Oh, I see you only changed the devicetodevice calls.  Would changing hostToDevice calls also work ?

galv:
Well, it certainly would break the timers surrounding these calls. This is because cudaMemcpyAsync is asynchronoous. It returns immediately. Regarding correctness of code, my hunch is that this change won't break anything, but would appreciate having a time to sit down and really think about it. cudaMemcpy calls are normally used as synchronization points, but as Justin said, that is only for host<->device transfers.

luitjens:
> Oh, I see you only changed the devicetodevice calls. Would changing hostToDevice calls also work ?

For hostToDevice we would have one more additional detail to worry about.   That is we would have to ensure the host memory was not deallocated prior to the copy finishing.  However,  this is only a concern if the host memory is in page-locked memory.  If the host memory is not page locked then the call will still be synchronous with respect to the host.  As far as I can tell Kaldi does not use page locked memory.  So again this would be safe.

luitjens:
> Well, it certainly would break the timers surrounding these calls. This is because cudaMemcpyAsync is asynchronoous. It returns immediately. Regarding correctness of code, my hunch is that this change won't break anything, but would appreciate having a time to sit down and really think about it. cudaMemcpy calls are normally used as synchronization points, but as Justin said, that is only for host<->device transfers.

Yes timers would be impacted.  Is that considered a big issue?  

If we really wanted to be conservative we could add cudaStreamSynchronize(cudaStreamPerThread); following each copy.  I'd prefer we did not do that as it adds unnecessary synchronization but that would get the transfers out of the default stream and maintain the exact same per-thread semantics.

danpovey:
Actually we don't have to worry about the timers.  They call
cudaStreamSynchronize() themselves, if enabled (verbose level >= 1).

The thing I was worrying about is this: in Kaldi we compile with
-DCUDA_API_PER_THREAD_DEFAULT_STREAM (default stream per thread), and we
may sometimes do things like reading objects of type CuMatrix in a
background thread (which happens automatically if you use
"ark,bg:foo.ark"). Those would copy from host to device, and possibly from
device to device if something gets called in an initializer (say, of a more
complex object that is being read).  If another stream might read that
memory, it would be necessary, in a loop like:
  for (; !reader.Done(); reader.Next()) {
to add:
  for (; !reader.Done(); reader.Next()) {
     SynchronizeGpu();

I may have omitted such calls to SynchronizeGpu() in the past after
reasoning that the only thing that was happening in the reader was
cudaMemcpy (host to device), which was synchronous.
But that may be in RNNLM-related code that was already deleted to work
around driver problems.  I would have to check if there is any more code of
that type.   I doubt there is though.



Dan

luitjens:
I think it would be good to run a battery of tests on this and see if any issues pop up.  In the end this will provide more performance but cases where you use a background thread will need to be explicit in their synchronization.   This would mean calling cudaStreamSynchronize prior to launching/activating the background thread.

ryanleary:
Thanks for getting this merged, Dan.

PR #2828, opened 2018-11-09
 move get_ctm_conf.sh from aspire to steps/conf/

xiaohui-zhang:


danpovey:
OK.  This is good, but I think we should also have a 'fast' version of this script, that's more like get_ctm_fast.sh (e.g. get_ctm_conf_fast.sh).  I.e. one that would do it for only one lmwt, but parallelize over jobs.  I think for many applications, this would be the most natural way to do it.  
Just to get all this done in go, would you mind working on that script?

xiaohui-zhang:
sure I agree. I realized we need that some time ago. will do.

xiaohui-zhang:
updated

danpovey:
Thanks a lot!  I'll merge this and we'll let users do testing.

PR #2840, opened 2018-11-14
wall clock time is a bad timer base

btiplitz:
When doing some performance testing, I noticed the times varied by run sometimes significantly.  using the thread clock time I believe is more accurate, so I thought I would suggest this change to see if it seemed line a good change or not.


danpovey:
Sorry, I don't think this would work for some of the use-cases.  E.g. we use this to compute the total wall-clock time of the program, which would be different if some of the work happens in background threads.  And it would likely fail to account for time spent on the GPU.

btiplitz:
For threads, there is also a CLOCK_PROCESS_CPUTIME_ID which would reflect the threads.  For the gpu, I believe that would not be a problem as I use this technique and I can compare these times to the cudaEvent with the same result.  For my issue, using the above clock would suffice

danpovey:
This change would be disruptive as it would require changes to all the code
that prints out real-time factors from programs.  I suggest you just change
it locally when you are doing performance testing.

PR #2851, opened 2018-11-19
[build] Remove -lcuda from cuda makefiles which breaks operation w/o driver (#2819)

danpovey:


danpovey:
Fixes bug introduced in #2819

PR #2856, opened 2018-11-20
fixed errors when built with --DOUBLE_PRECISION=1

huangruizhe:


galv:
This seems odd. While your solution works, it is probably depending on an implicit cast from `std::numeric_limits<float>::min()` to BaseFloat. It should work just as well if you do `std::numeric_limits<BaseFloat>::min()` (and it would also be more correct, as then you are comparing with correct minimum).

danpovey:
I think the reason we made that float and not BaseFloat is to get a consistent  minimum feature value regardless of how it's compiled.

PR #2859, opened 2018-11-20
WIP: Cluster nonsilence phones for aishell

naxingyu:
to share the roots of the different tones of each phone.
This will change the results.

### DIFF
--- a/egs/aishell/s5/RESULTS
+++ b/egs/aishell/s5/RESULTS
@@ -1,18 +1,18 @@
-%WER 33.82 [ 35432 / 104765, 743 ins, 3991 del, 30698 sub ] exp/mono/decode_test/cer_12_0.0
-%WER 19.39 [ 20310 / 104765, 903 ins, 1452 del, 17955 sub ] exp/tri1/decode_test/cer_13_0.5
-%WER 19.23 [ 20147 / 104765, 910 ins, 1287 del, 17950 sub ] exp/tri2/decode_test/cer_14_0.5
-%WER 17.14 [ 17961 / 104765, 812 ins, 1024 del, 16125 sub ] exp/tri3a/decode_test/cer_14_0.0
-%WER 13.64 [ 14294 / 104765, 669 ins, 736 del, 12889 sub ] exp/tri4a/decode_test/cer_14_0.5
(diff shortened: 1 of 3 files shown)


danpovey:
Thanks.  But would you mind testing it- seeing how it affects the results?

naxingyu:
Yes, I'm running the test. That's why I put WIP in the title. It's a
placeholder.

naxingyu:
@danpovey Here are the GMM results.

test CER% | non-cluster | cluster
---------|------------|--------
mono   | 33.82 | 36.41
tri1 (delta) | 19.39 | 18.76
tri2  (larger delta) | 19.23 | 18.64
tri3a (lda+mllt) | 17.14 | 17.04
tri4a (sat) | 13.64 | 13.82
tri5a  (larger sat) | 12.23 | 12.12

One thing worth noticing, after using clustered non-silence-phones, the `number of transition-ids` is increased from 6306 to 10160 and `number of transition-states` is increased from 3149 to 5076, while `number of pdfs` barely changes. This result in an larger graph in the tri5a experiment, 205M, compared with 175M in the non-clustered setting.

danpovey:
Thanks.   Let's wait to see the effect on the chain system.

naxingyu:
Comparing chain models trained using online pitch with right context 25.

test CER% | non-cluster | cluster
-- | -- | --
chain | 7.53 | 7.45
+online decoding | 7.55 | 7.43
+per-utt | 8.03 | 7.92

The size of decoding graph of the chain models barely change.

danpovey:
Oh, OK.  That's great.  Would you mind changing  the RESULTS file and the results in the chain script that you ran?  You can compare the two things in a comment there and explain what you were comparing.

naxingyu:
I'm thinking of running a non-online recipe as well. Because the recommended chain recipe of aishell is run_tdnn_1a(offline pitch), not run_tdnn_2a(online pitch). I have both offline and online results in the general RESULTS file.
Regarding the results in the chain script, do you mean comparison w/ or w/o clustered phones?

danpovey:
OK.
and yes.

danpovey:
Great, thanks.  I think those changes may mostly be random noise.    Merging, as it's the right way to set this up.

PR #2866, opened 2018-11-24
Remove redundant for loop in ProcessNonemitting while push state into queue_

mrlittlepig:
The for loop can be removed, new state can be push in queue_ when call FindOrAddToken func

### DIFF
--- a/src/decoder/lattice-faster-decoder.cc
+++ b/src/decoder/lattice-faster-decoder.cc
@@ -68,6 +68,7 @@ void LatticeFasterDecoderTpl<FST, Token>::InitDecoding() {
   Token *start_tok = new Token(0.0, 0.0, NULL, NULL, NULL);
   active_toks_[0].toks = start_tok;
   toks_.Insert(start_state, start_tok);
+  queue_.push_back(start_state);
   num_toks_++;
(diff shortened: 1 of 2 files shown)


danpovey:
I think this is probably not right; it looks to me like this would result in the wrong types of tokens (emitting vs. non-emitting) being processed in the wrong place.

mrlittlepig:
Look this state in the for loop, in which Elem is added at any FindOrAddState, and tokens has also separated after emitting processing, and all added token state is been used in non-emitting processing
```
for (const Elem *e = toks_.GetList(); e != NULL;  e = e->tail) {
    StateId state = e->key;
    if (fst_->NumInputEpsilons(state) != 0)
      queue_.push_back(state);
  }
```

mrlittlepig:
@danpovey emitting process current frame, non-emitting process next frame, so I think there is nothing wrong logically

danpovey:
I don't understand what you are saying.  I am closing this for now, but if you manage to persuade this guy
Zhehuai Chen <chenzhehuai@foxmail.com>
that your PR is correct, I am willing to reconsider.

mrlittlepig:
OK，Thank you

PR #2880, opened 2018-11-27
[WIP} windows cuda build failure

btiplitz:
When building with windows and cuda, the build fails with undefined types.  The errors are simulated from linux by removing the stdint.h include file which is not present in windows, thus the same error occurs.  I've updated the include to occur in both windows/non-windows

../cudamatrix/cu-matrixdim.h(35): error: identifier "uint32_t" is undefined

../cudamatrix/cu-kernels-ansi.h(765): error: identifier "uint16_t" is undefined

../cudamatrix/cu-kernels-ansi.h(769): error: identifier "uint8_t" is undefined

../cudamatrix/cu-kernels-ansi.h(778): error: identifier "uint8_t" is undefined

../cudamatrix/cu-kernels-ansi.h(784): error: identifier "uint16_t" is undefined

../cudamatrix/cu-kernels-ansi.h(790): error: identifier "uint8_t" is undefined

cu-kernels.cu(3630): error: identifier "uint8_t" is undefined

cu-kernels.cu(3634): error: identifier "uint16_t" is undefined

cu-kernels.cu(5414): error: "uint16_t" is not a type name

cu-kernels.cu(5434): error: "uint8_t" is not a type name

cu-kernels.cu(5445): error: "uint8_t" is not a type name

cu-kernels.cu(5455): error: "uint16_t" is not a type name


danpovey:
Thanks!
since the c++11 standard now seems to include stdint.h (also available as
`<cstdint>`), and therefore it should be present on windows, I think that
whole ifdef can be removed.

btiplitz:
are you thinking remove the 1st part and just use the 2nd part (and update the include>.  My windows build system is automatic, so testing updates are very painful but I could test out the change.  I'll know in the morning

danpovey:
Yes, remove the MS-specific stuff.

btiplitz:
The code change did compile on windows overnight

PR #2884, opened 2018-11-28
Adding sampling to nbest

vimalmanohar:
@danpovey 
I wrote the code, but I'm not really sure how to test it. It seems to work without errors.


danpovey:
Thanks, Mahsa will test it.

PR #2887, opened 2018-11-28
example self-contained example for KWS for minilibrispeech

jtrmal:
Compared to the old KWS example on WSJ, this uses a public corpus, does not depend on any external tools (F4DE), uses much simpler file formats (no rttms, no xmls) and actually does scoring (again usign in-built kaldi's tools).


ckn123:
Dr dan,My kaldi ASR model is kaldi/egs/thchs30.
I created exf.xml according to the command given by your repro, and I created kwlist.xml manually. However, for rttm I don't know how to create it. Can you guide me? Thank you!

Issue #2852, opened 2018-11-19
TImer.cc change breaks windows build


btiplitz:
In timer.cc:81, the destructor for Profiler fails as this does not exist on windows

danpovey:
@huangruizhe can you please make a PR to fix this?  Ask someone for help if you don't know how.

huangruizhe:
ok

btiplitz:
@huangruizhe any progress.  I could make a PR myself it that helps.  I see 2 other problems on the windows side.  1 is uint8, etc is undefined.  We seem to have a fix for that I blew away when merging.  I also have another error I'm investigating (all for windows only)

huangruizhe:
@btiplitz Thanks for the reminder and patience! I am new to this and I am working on it, and hopefully will deliver it today.

PR #2888, opened 2018-11-29
fixed: In timer.cc:81, the destructor for Profiler fails as this does…

huangruizhe:
… not exist on windows
#2852

### DIFF
--- a/src/base/timer.h
+++ b/src/base/timer.h
@@ -53,7 +53,7 @@ class Timer {
  private:
   LARGE_INTEGER time_start_;
 };
-}
+
 
 #else
 #include <sys/time.h>
@@ -88,6 +88,8 @@ class Timer {
   struct timezone time_zone_;
 };
 
+#endif
+
 class Profiler {
  public:
   // Caution: the 'const char' should always be a string constant; for speed,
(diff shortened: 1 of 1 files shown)


src/base/timer.h
```diff
@@ -88,6 +88,8 @@ class Timer {
   struct timezone time_zone_;
 };
 
+#endif
+
 class Profiler {
  public:
```
danpovey: These changes seem unrelated to the original issue about class Profiler. I think the fix is to simply add `{}` after
`~Profiler()`

huangruizhe: Oh, it seems in the original file, the class Profiler will only be enabled under the Unix settings -- that is, when `if defined(_MSC_VER) || defined(MINGW)` (line 26) is false (starting from line 58).
Can we just add add `{}` after `~Profiler()`?

huangruizhe: If just add `{}` after `~Profiler()`, there will be no class definition for class Profiler in windows settings

danpovey: Oh sorry, I see it now, your fix might be right.
@btiplitz  can you please test?

btiplitz: I’ll try it in the morning

PR #2889, opened 2018-11-29
Fix max duration check in sad_to_segments.py

dogancan:


danpovey:
Thanks!  @vimalmanohar  please check.

PR #2891, opened 2018-11-29
Fix speech duration calculation in sad_to_segments.py

dogancan:


dogancan:
@vimalmanohar Could you check if this change looks ok?

vimalmanohar:
It looks fine to merge. Thanks for the PR.

PR #2897, opened 2018-12-03
WIP Fix bugs in Dihard 2018

HuangZiliAndy:
I ran the experiment again and found a serious bug in the script. And I also found some small things related to python. (for example python3 map returns a generator instead of a list) Thanks!


HuangZiliAndy:
@david-ryan-snyder Do you have time to take a look? Thanks!

HuangZiliAndy:
Hi David!
    I met some problems when I was using python 3. It seems that python3 map returns a generator instead of a list. I have tested current code using both python2.7 and python3 and there is no problem. The operation is quite safe.

Best,
Zili

david-ryan-snyder:
@HuangZiliAndy, OK, thanks.

@danpovey, I think this is OK to merge.

PR #2901, opened 2018-12-05
Fix harcoded job numbers in AMI egs 

akashmjn:
1. Fixes a hardcoded --nj flag in run.sh 
2. Minor changes to all the run_tdnn* scripts that call train.py using a hardcoded number of max GPU jobs that can cause job failures. Added another nj_gpu flag to handle this 

Happens to include an (optional) .gitignore addition to ignore downloaded wav directories as well.


danpovey:
Thanks for the PR, but I don't want to merge this because it might give people the impression that they could expect to change the number of jobs and for the results to be the same.  That isn't the case.  A better way to do it is to change the GPUs to exclusive mode (`nvidia-smi -c 3`) and instead of --use-gpu=yes  to train.py, use --use-gpu=wait.  That way it will use the number of jobs the script requested, but not run out of memory.

akashmjn:
Cool, thanks for the clarification - makes sense. 
Just want to add that after trying out what was suggested in the logs (setting GPUs to exclusive mode + the use-gpu=wait option), this was finally what had gotten things to work in my case. 
In exclusive mode, after some iterations, the GPUs tended to get locked and processes were waiting indefinitely for them to free up. 
Unsure if that's particular to my setup (Ubuntu 14, Cuda 8, 4xQuadroM6000 GPUs) or a general issue.

danpovey:
It might be a hardware issue.  Sometimes it depends on the power draw.
But you can test by running
nvidia-smi
what process is using the GPU.

PR #2910, opened 2018-12-12
[not-for-merge] Implementation of variable sized word embeddings for the RNNLM

RuABraun:
This implements the core idea of [this paper](https://openreview.net/forum?id=ByxZX20qFQ). With it I can have a vocab size of 140 000 while the embeddings only take up 65MB (in RAM as well as on disk, unlike when using character level features). For comparison 140000 x 512 x 4 / 1e6 = 286MB. I get good WER results with this (haven't had time to do a proper comparison across different model and embedding sizes, but it's better than both word/character level embeddings (size 512) at a vocab size of 60 000).

I'm opening this so that those interested can try it out, and so others have an example of how to make changes to the RNNLM training code. I do not have time for cleaning this up or making it nicer (refactoring), so unless there are bugs I do not intend to work on this PR, hope that is alright.

The model has three inputs and outputs (`input`, `inputmed` and `inputsmall`, same style for the output). The variables for the different embeddings have to be set in `prepare_rnnlm_dir.sh`, `rnnlm-train-adapt` and `rnnlm-compute-prob-adapt`. 

The lines changed count is misleading, rather than introducing a lot of if else branches I just copied a lot of classes and modified them, the amount of actually new code added is not that high.

### DIFF
--- a/scripts/rnnlm/prepare_rnnlm_dir.sh
+++ b/scripts/rnnlm/prepare_rnnlm_dir.sh
@@ -16,6 +16,10 @@ unigram_factor=100.0     # Option used when pruning the LM used for sampling.
                          # You can increase this, e.g. to 200 or 400, if the LM used
                          # for sampling is too big, causing rnnlm-get-egs to
                          # take up too many CPUs worth of compute.
+adaptive=false
+largeembsize=10000
+medembsize=50000
+smallembsize=80000
 
 . utils/parse_options.sh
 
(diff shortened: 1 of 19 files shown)


danpovey:
That's interesting.  @hainan-xv perhaps you could have a look and let me know what you think?

For your information, my current ambition/plan is to move towards a more word-piece-based decoding method, i.e. to work with graphs where the word-ids correspond to fragments of words (in graphemic systems), or fragments of pronunciations of words (in phonetic systems, e.g. in English where graphemic systems won't work as well for typical amounts of training data).  Of course there are some non-trivial issues to solve here, such as, in languages like English, how you would train such models when most of your LM training data is text and not transcribed speech... so this might not happen on the timescale of a few months, and it might not happen at all.

hainan-xv:
This seems very interesting. I will take a look.

Nimitz14:
Happy to see some interest! Realized there isn't an example model config, this is the one I used (for my use case I have to keep the model small):

```
input dim=$embedding_dim name=input
input dim=128 name=inputmed
input dim=64 name=inputsmall

affine-layer name=embresizesmall input=inputsmall dim=$embedding_dim
affine-layer name=embresizemed input=inputmed dim=$embedding_dim
affine-layer name=embresizelarge input=input dim=$embedding_dim

relu-renorm-layer name=tdnn1 input=Sum(embresizelarge, embresizemed, embresizesmall) dim=$embedding_dim
fast-lstmp-layer name=lstm1 cell-dim=$embedding_dim recurrent-projection-dim=$lstm_rpd non-recurrent-projection-dim=$lstm_nrpd
relu-renorm-layer name=tdnn2 dim=$embedding_dim
output-layer name=output include-log-softmax=false dim=$embedding_dim
output-layer name=outputmed include-log-softmax=false input=tdnn2 dim=128
output-layer name=outputsmall include-log-softmax=false input=tdnn2 dim=64
```

PR #2925, opened 2018-12-20
Python 2/3 compatibility

desh2608:
Original PR: https://github.com/kaldi-asr/kaldi/pull/2909

Changes:
- removed dependence on "builtin" library
- made changes according to suggestions by @vimalmanohar in original PR


danpovey:
@vimalmanohar sorry, this may be a bit tedious but would you mind having a look through this?  You can focus on the shared scripts and just spot check the example scripts if you want to save time.

vimalmanohar:
I'm looking at it.

egs/wsj/s5/steps/libs/nnet3/train/common.py
```diff
@@ -564,7 +565,7 @@ def get_model_combine_iters(num_iters, num_epochs,
         in the final model-averaging phase.  (note: it's a weighted average
         where the weights are worked out from a subset of training data.)"""
 
-    approx_iters_per_epoch_final = num_archives/num_jobs_final
+    approx_iters_per_epoch_final = float(num_archives) / num_jobs_final
```
mthrok: Isn't adding `float` here redundant after `from __future__ import division`?

desh2608: Thanks for pointing out. I will resolve this once @vimalmanohar suggests which one to retain.

vimalmanohar:
Adding float is better for readability. import division can remain as it
causes no harm, and is applicable for future code we might write in
python3.

danpovey:
@vimalmanohar let me know when this is ready to merge.

vimalmanohar:
I went through some parts of it and will go through the rest. But there are some non-trivial changes needed when it comes to dealing with unicode since python 2 and python 3 do it differently. Even the futurize library would not support dealing with these. I think python 3 requires that text read are in some encoding and by default this is ASCII. But depending on the language, some times the text file can have non-ascii. And this is a problem for python 3 I think.

desh2608:
@vimalmanohar Thanks for your comments. I'll make these changes and also extrapolate the same to all other files.

egs/babel/s5b/local/lonestar.py
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env python
+from __future__ import print_function
+from future import standard_library
```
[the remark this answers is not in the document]
desh2608: Some external libraries (like pylauncher in this case) were refactored or extended from Python 2 to 3, but were not backported. This allows the Python3 APIs of these libraries to be compatible with Python2. Since the original code was in Python2 anyway, we are anyway just using the functions which were available then. So this does not affect anything. I'll remove these imports.

egs/babel/s5d/local/lexicon/make_unicode_lexicon.py
```diff
@@ -629,9 +630,9 @@ def write_lexicon(baseforms, encoded_transcription, outfile, sil_lex=None,
         for idx, w in enumerate(baseforms):
             # This is really just for BABEL in case <hes> is written as a word
             if(w[0].lower() == "<hes>"):
-                f.write("%s\t<hes>\n" % (unicode(w[0])))
```
desh2608: Marked for unicode.

egs/babel/s5d/local/lexicon/make_word_list.py
```diff
@@ -85,7 +85,7 @@ def main():
     # Print the word list
     with codecs.open(args.word_list, "w", encoding="utf-8") as f:
         for word, count in words:
-            f.write("%d %s\n" % (count, unicode(word)))
```
desh2608: Marked for unicode.

egs/cifar/v1/image/get_allowed_lengths.py
```diff
@@ -133,7 +134,7 @@ def main():
                 "Coverage rate: {}%".format(start_dur, end_dur,
                                       100.0 - args.coverage_factor * 2))
     logger.info("There will be {} unique allowed lengths "
-                "for the images.".format(int(math.log(end_dur / start_dur) /
+                "for the images.".format(int((math.log(float(end_dur)/start_dur))/
```
desh2608: This is done because `start_dur` and `end_dur` are ints.

egs/hub4_english/s5/local/data_prep/process_1995_bn_annotation.py
```diff
@@ -230,7 +230,7 @@ def run(args):
                         start_time = story_end_time
                     segments = process_story_content(
                         args, reco_id,
-                        ' '.join([unicode(x) for x in s.children]),
```
desh2608: Marked for unicode.

egs/hub4_english/s5/local/data_prep/process_1996_csr_hub4_lm_filelist.py
```diff
@@ -83,7 +83,7 @@ def process_file_lines(lines, out_file_handle):
                 for x in para.contents:
                     try:
                         if x.name is None:
-                            normalized_text = normalize_text(unicode(x))
```
desh2608: Marked for unicode.

egs/hub4_english/s5/local/data_prep/process_na_news_text.py
```diff
@@ -85,7 +85,7 @@ def process_file_lines(lines, out_file_handle):
                 continue
             for para in art.find_all('p'):
                 assert para.name == 'p'
-                text = ' '.join([unicode(x).strip() for x in para.contents])
```
desh2608: Marked for unicode.

egs/vystadial_cz/online_demo/pykaldi-online-latgen-recogniser.py
```diff
@@ -31,14 +33,14 @@ def write_decoded(f, wav_name, word_ids, wst):
     if wst is not None:
         decoded = [wst[w] for w in best_path]
     else:
-        decoded = [unicode(w) for w in best_path]
```
desh2608: Marked for unicode.

egs/wsj/s5/utils/reverse_arpa.py
```diff
@@ -179,10 +179,10 @@
         elif n == 2:
           revprob = revprob + offset # add <s> weight to bigrams starting with <s>
       if (prob[1] != inf): # only backoff weights from not newly created ngrams
-        print revprob,rev_ngram.encode("utf-8"),back
+        print(revprob,rev_ngram.encode("utf-8"),back)
```
desh2608: Marked for unicode.

misc/maintenance/cpplint.py
```diff
@@ -2073,7 +2074,7 @@ def GetLineWidth(line):
     The width of the line in column positions, accounting for Unicode
     combining characters and wide characters.
   """
-  if isinstance(line, unicode):
```
desh2608: Marked for unicode.

danpovey:
A bunch of these have
`from __future__ import print_function`
but are invoked as `#!/usr/bin/env python3`.
Would you mind reverting changes on files that are invoked as python3?  (Obviously check that the resulting changes are harmless).

desh2608:
> A bunch of these have
> `from __future__ import print_function`
> but are invoked as `#!/usr/bin/env python3`.
> Would you mind reverting changes on files that are invoked as python3? (Obviously check that the resulting changes are harmless).

Done @danpovey

vimalmanohar:
Maybe we want to make these scripts usable in both python 2 and 3? Or if
it's ok to make it just usable in python 3, some things might be simpler.
Like I think the unicode handling are very different in both.

PR #2931, opened 2018-12-21
Parallel lat

PCerles:


danpovey:
I'm afraid your changes break the build of other Kaldi programs.  
You probably won't want to do the lattice-determinize-phone-pruned in this case, you can just do regular determinization, like lattice-determinize does.
To do this right you might need to copy and modify certain classes.
I'm not sure that, if this were fixed, I would merge this, though, as I don't think there are use-cases right now.

PR #2932, opened 2018-12-21
[src] Fix incorrect invocation of mutex in nnet-batch-compute code

danpovey:


danpovey:
@luitjens this bug may have affected you guys.

PR #2938, opened 2018-12-25
Online2 NNet3 TCP server program

danijel3:
Several people asked for this and I feel like it would be a nice addition to the project.

The protocol is much simpler than the audio-server program I did a while ago - audio in -> text out.

The way it's made now is nice for a live demo (I added some commands to the doxygen docs), but may still lack some features for real-life use.

The main issue I have is that the new decoder is slightly different than before. The old decoder had a way to check which part of the output is final and which is "partial". This time, I can only check the current best path every N seconds (eg once per second of input audio). I use endpointing to determine when to finalize decoding.

Now, what would be really nice it to have online speech detection and speaker diarization included with this, but I know it's probably not happening too soon. What can be done (and I may do it myself if I find time) is a multithreaded version of the program with shared acoustic model and FST. Also, I bet it could be possible to combine the grammar version with the server to allow runtime vocabulary modification.

I also have a web interface that works with this server, but I'm not sure if it would fit the main Kaldi repo, so I'll probably make a separate repo for that (if anyone wants it).

I'm open to comments and suggestions.


danpovey:
@hhadian since you have done things of this nature, perhaps you could try to find time to review it?  Not just from code details, but whether the whole approach is a direction we want to go in.  No hurry.

src/online2bin/online2-net-nnet3-latgen-faster.cc
```diff
@@ -0,0 +1,398 @@
...
+    delete[] samp_buf_;
+    samp_buf_ = new int16[len];
+  }
+
+  size_t to_read = len;
+  has_read_ = 0;
+  while (to_read > 0) {
+    ssize_t ret = read(client_desc_, static_cast<void *>(samp_buf_ + has_read_), to_read * sizeof(int16));
+    if (ret <= 0)
+      break;
```
hhadian: This loop won't end unless an error occurs or the client shuts down its end of connection (TCP half-shutdown).

src/online2bin/online2-net-nnet3-latgen-faster.cc
```diff
@@ -0,0 +1,398 @@
...
+
+        SingleUtteranceNnet3Decoder decoder(decoder_opts, trans_model,
+                                            decodable_info,
+                                            *decode_fst, &feature_pipeline);
+
+        std::vector<std::pair<int32, BaseFloat> > delta_weights;
+
+        while (true) {
+
+          eos = !server.ReadChunk(chunk_len);
```
hhadian: Are you sure the eos part works with `nc`? AFAIK `nc` won't send a TCP half-shutdown, so you'll never get eos unless you stop `nc` in which case the client won't receive the final result (i.e. L248) because it's disconnected.

src/online2bin/online2-net-nnet3-latgen-faster.cc
```diff
@@ -0,0 +1,398 @@
...
+
+std::string LatticeToString(const Lattice &lat, fst::SymbolTable *word_syms) {
+  LatticeWeight weight;
+  std::vector<int32> alignment;
+  std::vector<int32> words;
+  GetLinearSymbolSequence(lat, &alignment, &words, &weight);
+
+  std::string msg = "";
+  for (size_t i = 0; i < words.size(); i++) {
+    msg += word_syms->Find(words[i]) + " ";
```
hhadian: Probably better to handle the case where `words[i]` is not found (e.g. show some error and return something meaningful to the client).

src/online2bin/online2-net-nnet3-latgen-faster.cc
```diff
@@ -0,0 +1,398 @@
...
+
+            server.disconnect();
+            break;
+
+          }
+
+          if (decoder.EndpointDetected(endpoint_opts)) {
+            server.Write("\n");
+            feature_pipeline.GetAdaptationState(&adaptation_state);
+            break;
```
hhadian: Are you sure no part of the output is missed? When breaking here, it seems to me that there might be new results that have not been sent to the client yet (e.g. if `check_update` is large).

hhadian: ...and BTW, the decoder is going to be destroyed here without `FinalizeDecoding` having been called on it.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Well, the samples I added in the docs worked fine. Regardless if the connection was interrupted abruptly (ctrl-c) or simply reached the end of file, the client (nc) disconnected cleanly and waited to accept the next connection. Did you have any problem running the samples?

Upon further inspection, the ctrl-c case does timeout for a few seconds, but it does go on waiting for the next client after that.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Ok, will do.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: I'll have to think about this a bit more. I'll get back to you on that ASAP.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Ok, I added a select w/timeout for reading. I'm pretty sure it's not really necessary for writing - it's unlikely we will overflow the output buffer, but I can do that as well if you think it's necessary. I tested the program and it's stable as usual.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Fixed this.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Ok, I made sure to finalize decoding after every speech segment and output the real best path from the final lattice.

I do realize a small semantic problem here. I used the name from the other binary (online2-wav-nnet3-latgen-faster), but I'm not really interested in generating a lattice here (and sending it over TCP). This program is intended more to preview the decoder output in real-time. Do you think I should rename it to something like "online2-net-nnet3-decode-faster"?

hhadian: That makes sense. Actually, it might be better to include TCP in the name-- something like `online2-nnet3-tcp-decoder` or `-server`.

src/online2bin/online2-net-nnet3-latgen-faster.cc
hhadian: I don't think that's necessary.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Ok, I renamed it to "online2-tcp-nnet3-decode-faster" so it syncs up with other names better.

src/online2bin/online2-net-nnet3-latgen-faster.cc
danijel3: Still trying to figure this out. Select doesn't seem to detect this CTRL-c scenario. Changing socket read timeout or keep alive timeout doesn't help either. My Linux simply goes through select, blocks in read for ~5 seconds and exists. Not sure what would unblock it sooner.

You think this should be really implemented in multiple threads? Isn't there a better option?

danijel3: Ok, I fixed it!

I looked at the sourcecode of netcat and seems poll is better for this kind of scenario than select. I went with poll for now and it works a treat.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,451 @@
...
+            decoder.FinalizeDecoding();
+
+            CompactLattice lat;
+            decoder.GetLattice(true, &lat);
+            std::string msg = LatticeToString(lat, word_syms);
+
+            server.Write(msg + "\n");
+            if (adapt_speaker)
+              feature_pipeline.GetAdaptationState(&adaptation_state);
+            break;
```
dogancan: There is another small issue here regarding the audio samples that were accepted by the feature pipeline but not yet processed into feature frames due to feature pipeline producing feature vectors only after the needed future audio context is available. As it is, the last few samples of audio already accepted by the feature pipeline will effectively be dropped. These dropped samples also have an effect on the first few feature frames extracted for the next segment. So the feature vectors will be extracted as if there was a small gap between the two segments determined by endpointing. The correct thing to do here is to figure out what audio samples are yet to be processed by the feature pipeline and accept those samples after reconstructing the feature pipeline before any new audio samples are accepted. You can take a look at what we have done in [this pykaldi example](https://github.com/pykaldi/pykaldi/blob/master/examples/asr/nnet3-online-recognizer.py#L108-L124) for the exact sequence of things to be done. Then again, this is a rather small issue that probably won't have an appreciable effect on decoding results.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,451 @@
...
+
+                server.Write(msg + '\r');
+              }
+
+              next_check += check_update;
+            }
+
+          } else {
+            feature_pipeline.InputFinished();
+
```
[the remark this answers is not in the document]
danijel3: Good point. Usually wouldn't notice that since most tests use files with a portion of silence in the end, but there may be examples that would fail.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Ok, I think I got it. Since we recreate the feature pipeline object for each new segment (which is needed for updating the adaptation state) we loose the last samples from the previous chunk that were still possibly speech.

I don't get one thing though in your code:

` remainder = data[offset:i + chunk_size]`

The offset is calculated purely using the values of the feature pipeline. So it's the offset from the beginning of the last segment, but here you are using it as if it is the offset from the beginning of the file. Shouldn't it be:

` remainder = data[i+offset:i + chunk_size]`

dogancan:
Yes it seems our code had a bug there!

dogancan:
I mean what you are saying in words is correct, what the code change does is not.

dogancan:
I updated the pykaldi example. The offset is again from the beginning of the file but now it is an accumulator. Let me know if it does not make sense to you.

nshmyrev:
@dogancan this is a great catch, I didn't notice that before. Actually I think it is also a design problem, in your pykaldi code it is simple to retrieve the remaining samples but in many cases processing and finalization belongs to different methods or executed in different places of the code. In that case access to audio buffer is already lost. It would be nice to have something like feature_pipeline.GetRemainingSamples then.

dogancan:
Unfortunately, feature pipeline needs to be reinitialized whenever decoder is reinitialized since the decoder starts counting the frames from 0 after the `init_decoding` call.

danpovey:
It might be necessary to store parts of the waveform outside the
feature-pipeline object.  Or if you can think of a way to extend it to
support getting the remainder, I could have a look.
I don't think it should be important in most applications to ensure that
the frame boundaries are exactly right between decoder invocations.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: I fixed this in the last commit. It's a bit different in my code, since I don't remember the complete audio, but just the last chunk, but the principle is the same.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -425,6 +449,26 @@ Vector<BaseFloat> TcpServer::getChunk() {
...
 
+void TcpServer::SaveRemainder(int remainder) {
+  rem_.Resize(static_cast<MatrixIndexT>(remainder));
+
+  int32 offset = has_read_ - remainder;
+  for (size_t i = offset; i < has_read_; i++)
+    rem_(i - offset) = static_cast<BaseFloat>(samp_buf_[i]);
+}
+
+bool TcpServer::HasReaminder() {
```
dogancan: typo: HasReaminder -> HasRemainder

There are a few more occurrences of this in this file.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -425,6 +449,26 @@ Vector<BaseFloat> TcpServer::getChunk() {
   return buf;
 }
 
+void TcpServer::SaveRemainder(int remainder) {
```
dogancan: probably better to name the argument remainder_size.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Refactored.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Sure

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -286,6 +307,9 @@ int main(int argc, char *argv[]) {
 
             decoder.FinalizeDecoding();
 
+            server.SaveRemainder(
+                samp_pipeline - feature_pipeline.NumFramesReady() * feature_pipeline.FrameShiftInSeconds() * samp_freq);
+
```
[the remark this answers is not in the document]
danijel3: No problem. It shouldn't hurt anything.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -307,8 +307,9 @@ int main(int argc, char *argv[]) {
 
             decoder.FinalizeDecoding();
 
-            server.SaveRemainder(
-                samp_pipeline - feature_pipeline.NumFramesReady() * feature_pipeline.FrameShiftInSeconds() * samp_freq);
+            int32 processed = feature_pipeline.NumFramesReady() * feature_pipeline.FrameShiftInSeconds() * samp_freq;
+            if (processed < samp_pipeline)
```
dogancan: We also need to make sure `samp_pipeline - processed <= chunk_len`. That's why I suggested clipping. Otherwise the `offset` variable in the `SaveRemainder` method might become negative and `samp_buf_[i]` might be out of bounds.

dogancan:
@danpovey 
I have been thinking about the best way of handling the feature pipeline when an endpoint is determined. With the current online feature pipeline, we need to reset the pipeline whenever we call `decoder.InitDecoding()` since the decoder starts counting frames from 0 when this method is called. There are two problems associated with resetting the pipeline in the case of an endpoint:

1. The remainder needs to be manually carried to the new pipeline. As you noted, this is not important in most applications and I would be very surprised if it had any appreciable effect on decoding results. Looking back at the changes @danijel3 had to make to handle the remainder, it seems we are better off not carrying the remainder and live with this small defect.

2. Whenever we reset the feature pipeline, we lose the left feature context. Unlike the first issue, this has an effect on decoding results since acoustic models typically make use of left context.

I think we can solve both of these issues by introducing a feature offset member variable to the feature pipeline. This feature offset would affect the `IsLastFrame`, `NumFramesReady` and `GetFrame` methods of the pipeline. We would update this variable when we want to reset the decoder, i.e. call `decoder.InitDecoding()`, but not reset the feature pipeline, e.g. in the case of an endpoint. Then, we would reset the pipeline only when we reach the end of the feature pipeline. 

What do you think? Do you see any issues with this design?

dogancan:
What I am proposing to do touches only `OnlineFeaturePipeline` and `OnlineNnet2FeaturePipeline` classes. Effectively, we will have the option of setting a frame offset in these two classes and this offset will be taken into consideration by `IsLastFrame`, `NumFramesReady` and `GetFrame` methods. We won't need to touch other parts of the code base. In fact, that is why I think this is the best way to go. Do you think there is a need for this functionality unless we are dealing with an online feature pipeline? If I understand correctly, what you are suggesting is a much deeper change.

dogancan:
To make things more concrete, I am suggesting something like the following:

```C++
void OnlineNnet2FeaturePipeline::SetFrameOffSet(int32 frame_offset) {
  frame_offset_ = frame_offset;
}

bool OnlineNnet2FeaturePipeline::IsLastFrame(int32 frame) const {
  return final_feature_->IsLastFrame(frame + frame_offset_);
}

int32 OnlineNnet2FeaturePipeline::NumFramesReady() const {
  return final_feature_->NumFramesReady() - frame_offset_;
}

void OnlineNnet2FeaturePipeline::GetFrame(int32 frame,
                                          VectorBase<BaseFloat> *feat) {
  return final_feature_->GetFrame(frame + frame_offset_, feat);
}
```

danpovey:
Oh OK, that change does look fairly minimal, yes, you can make a PR for
that (or suggest to whoever is working on the online decoding stuff to do
it).  You might want to add a check that the frame offset is in the correct
range, and don't forget to initialize it in the constructor.  And make sure
to document SetFrameOffset().

dogancan:
@danijel3 See the change in #3008. I think with this change all you need to do when an endpoint is detected is to call `feature_pipeline.SetFrameOffset(decoder.NumFramesDecoded())` before calling `decoder.InitDecoding()`. So you don't need to reconstruct the feature pipeline and you don't need to get/set adaptation state.

dogancan:
@danijel3 scratch my earlier comment. We need to keep a frame offset value and add the number of decoded frames to that value every time we have an endpoint. Also I just noticed that after this change we need to adjust the delta weights used for silence weighting. I am in a hurry at the moment but I will leave a more detailed comment later.

danpovey:
Perhaps it would be more elegant to make UpdateFrameWeights() a function of
feature_pipeline directly, and have that be responsible for dealing with
the frame offset if necessary.

dogancan:
@danpovey 
It seems like we cannot decode multiple utterances separated with endpoints without redoing the left context nnet computation for each such utterance using `SingleUtteranceNnet3Decoder` class since we reconstruct the decoder object, hence the decodable object, every time we hit an endpoint. Is this accurate?

I understand that `SingleUtteranceNnet3Decoder` was not designed with the intention of decoding multiple utterances given a feature pipeline. So, I suggest we add an `InitDecoding` method to `SingleUtteranceNnet3Decoder`, which simply calls the `InitDecoding` method of the underlying decoder, to handle this use case and maybe also rename the class to `OnlineNnet3Decoder` or something similar. This way, we could avoid reconstructing the decoder whenever we hit and endpoint and simply call `InitDecoding` instead. What do you think?

danpovey:
The decoder object itself does not really have any significant resources in
it and was not designed to be reused (might cause bugs)... I think it might
make more sense to reallocate it, but keep the decodable object if
necessary, if one were to go down that path.

dogancan:
I agree. In fact, I just realized the change in #3008 is not enough for what I was trying to do. After `InitDecoding` is called, the decoder will once again get the likelihoods for the first frame from the decodable object. As it is, the decodable object does not know about the frame offset, only the feature pipeline knows about it, so the likelihoods will be computed as if there is no left context which defeats the purpose.

dogancan:
Hey @danijel3. Do you mind pulling in the changes from #3008 and updating your decoder to work with the new API? If you have already pulled that PR branch before the latest changes, you will have to get rid of your local copy before pulling since I force pushed the changes to the PR branch.

With the new API, handling endpoints should be significantly simpler. Major changes are as follows:

1. We do NOT reconstruct the decoder when we hit an endpoint. We simply update the `frame_offset` by adding `decoder.NumFramesDecoded()` to it and call `decoder.InitDecoding(frame_offset)`.

2. We do NOT reconstruct the feature pipeline when we hit an endpoint since we are using the same decoder.

3. We do NOT get/set the adaptation state when we hit an endpoint since we will continue using the same feature pipeline.

4. We reconstruct the silence weighting object when we hit an endpoint as before. When silence weighting is active, we call `feature_pipeline.UpdateFrameWeights(delta_weights, frame_offset * frame_subsampling factor)` to weight ivector frames.

Let me know if you have any difficulty incorporating these changes. Note that the new code changes are not tested so there might be bugs, especially in some boundary cases.

danijel3:
Okay, I'll give it a go and let you know.

danijel3:
Okay, now I figured out I should update the `frame_offset` after `FinalizeDecoding` instead (so generally only once per segment), but I keep hitting this assertion the third time `InitDecoding`:
```
ASSERTION_FAILED (online2-tcp-nnet3-decode-faster[5.5.209-4d77]:SetFrameOffset():decodable-online-looped.cc:113) : '0 <= frame_offset && frame_offset <= NumFramesReady()' 
```

dogancan:
Can you push the code you have so far?

Yes, you need to update frame offset only when you hit an endpoint. Can you print out the values of frame offset and the number of frames ready right before that failing assertion? I mean print them out every time you call initdecoding so we can see the progression.

danijel3:
Okay, I pushed the changes. It compiles fine but dies after a few endpoint matches.

I added a debug to src/nnet3/decodable-online-looped.cc:113 to see and I get this:
```
LOG (online2-tcp-nnet3-decode-faster[5.5.209~1-4d77]:SetFrameOffset():decodable-online-looped.cc:113) frame_offset: 0 NFR 0
LOG (online2-tcp-nnet3-decode-faster[5.5.209~1-4d77]:main():valdi-online/online2-tcp-nnet3-decode-faster.cc:318) EPD frame_offset+= 133
LOG (online2-tcp-nnet3-decode-faster[5.5.209~1-4d77]:SetFrameOffset():decodable-online-looped.cc:113) frame_offset: 133 NFR 133
LOG (online2-tcp-nnet3-decode-faster[5.5.209~1-4d77]:main():valdi-online/online2-tcp-nnet3-decode-faster.cc:318) EPD frame_offset+= 168
LOG (online2-tcp-nnet3-decode-faster[5.5.209~1-4d77]:SetFrameOffset():decodable-online-looped.cc:113) frame_offset: 301 NFR 168
ASSERTION_FAILED (online2-tcp-nnet3-decode-faster[5.5.209~1-4d77]:SetFrameOffset():decodable-online-looped.cc:114) : '0 <= frame_offset && frame_offset <= NumFramesReady()'
```

EPD is "endpoint detected" and NFR is "NumFramesReady"

danijel3:
Do you want me to prepare some data/models and scripts to simplify testing? I can upload them to GDrive or somewhere similar.

dogancan:
Yep, that was a bug in the frame offset checking. I just pushed a fix for it to the other PR branch.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,509 @@
...
+
+          }
+
+          if (decoder.EndpointDetected(endpoint_opts)) {
+
+            decoder.FinalizeDecoding();
+            KALDI_LOG << "EPD frame_offset+= " << decoder.NumFramesDecoded();
+            frame_offset += decoder.NumFramesDecoded();
+
+            int32 processed =
```
dogancan: Same as before. We can get rid of all remainder related code.

dogancan:
Regarding your question about getting/setting the adaptation state, I agree with what you are saying. My suggestion was based on the assumption that each separate stream was a single speaker since we don't have online diarization yet. My point was that we don't need to update adaptation state in that scenario.

danijel3:
Okay, I'll fix the code above as soon as I get the chance.

Wrt the adaptation issue, as I understand, currently we have the following options:

1. if we have a single speaker throughout the recording, we want the system to adapt continuously and therefore we don't need to do anything in the code, just leave it as it

2. if we want to assume that each segment is a new speaker, we need to **reset** the adaptation step after each endpoint, by (for example) setting it to a default adaptation state created at the start of the program 

3. in the future, when we get online diarization, we will keep a map of the style `[speaker] -> [adaptation state]` and use that to reset the adaptation to a proper state at the start of each new segment

Is that about right?

Also, all these changes - will they have to be incorporated into other online2bin programs?

dogancan:
I agree with your assessment of the adaptation state management.

> Also, all these changes - will they have to be incorporated into other online2bin programs?

Let's not worry about them in this PR. Those executables assume that input is already segmented into relatively short utterances. Splitting each input utterance into multiple output utterances might be undesirable in that scenario.

@danpovey might also want to comment on these.

danpovey:
Regarding resetting the adaptation state or not, there is a config variable
-- something like max-remembered-frames -- which you can set that controls
how much it remembers from before... likely an intermediate value will work
best if you are not sure if the speaker state is changing.

danijel3:
Ok, I updated everything, removed remainders, but still have a problem.

I am unable to use `SetAdaptationState` cause I hit an assertion at online-ivector-feature.cc:429:
```
 KALDI_ASSERT(num_frames_stats_ == 0 &&
               "SetAdaptationState called after frames were processed.");
```

I wanted to use SetAdaptationState to reset the adaptation after endpoint but without recreating the feature pipeline.

danpovey:
OK, it looks like you  can't do that.
It may be easier to just modify the ivector extractor code so that the
limit 'max-remembered-frames' is applied continuously rather than just at
utterance boundaries.
I think we should probably keep that separate from this PR though.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,462 @@
...
+              decoder.GetLattice(true, &lat);
+
+              std::string msg = LatticeToString(lat, word_syms);
+
+              server.Write(msg + "\n");
+            } else
+              server.Write("\n");
+
+            server.Disconnect();
+            break;
```
dogancan: Seems like this break statement is not taking us out of the second loop. I mean we want to go to line 206 here.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,462 @@
...
+          eos = !server.ReadChunk(chunk_len);
+
+          if (!eos) {
+            Vector<BaseFloat> wave_part = server.GetChunk();
+            feature_pipeline.AcceptWaveform(samp_freq, wave_part);
+            samp_pipeline += wave_part.Dim();
+            samp_count += chunk_len;
+
+            if (silence_weighting.Active() &&
+                feature_pipeline.IvectorFeature() != NULL) {
```
dogancan: We still need the first two lines that used be here. I mean the lines which compute delta weights.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: It's not a big deal, because, the next time this loop is entered the read will fail and the program does exit gracefully.

But I will fix it by setting `eos=true` after server disconnect, which will exit the outer loop as well.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Yea, my bad. I don't really use silence weighing in my model so I missed that in my tests :(

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,460 @@
...
+        decoder.InitDecoding(frame_offset);
+
+        OnlineSilenceWeighting silence_weighting(
+            trans_model,
+            feature_info.silence_weighting_config,
+            decodable_opts.frame_subsampling_factor);
+
+        std::vector<std::pair<int32, BaseFloat>> delta_weights;
+
+        int32 samp_pipeline = 0;//this is used to figure out the offset for the remainder
```
dogancan: We can remove `samp_pipeline`. It doesn't have a use any more.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,460 @@
...
+
+        while (true) {
+
+          eos = !server.ReadChunk(chunk_len);
+
+          if (!eos) {
+            Vector<BaseFloat> wave_part = server.GetChunk();
+            feature_pipeline.AcceptWaveform(samp_freq, wave_part);
+            samp_pipeline += wave_part.Dim();
```
dogancan: remove `sample_pipeline`.

dogancan:
I think it makes sense to have the option of carrying the speaker state between different feature streams. This would be useful when each stream is relatively short, e.g. an utterance, and the speaker changes between separate streams are infrequent.

It might also make sense to accept the speaker ID as part of the request but I am not sure whether we want to do that as part of this PR.

danijel3:
I finally managed to complete a proper demo of the online decoding for Kaldi. It took me this long cause I had to learn a lot about asyncio and Python, but I'm finally satisfied by it and would like to share it with you.

The project is currently located at https://github.com/danijel3/KaldiWebrtcServer as it's a bit unrelated to the main Kaldi project, but I don't mind it being incorporated with the original project or moved the the "kaldi-asr" account or whatever. If you manage to run it, you'll notice I used the original Kaldi branding and I intend to leave it like that.

What the demo does is it allows you to open a website on a modern browser, talk to the microphone and get text appearing - kinda like what you get with Google Web Speech API. Unlike Google, it is based on  WebRTC which is a protocol used for video-conferencing/chat and is supported in most modern browsers out-of-the-box.

To make it easy to run the demo, I have dockerized everything, so you don't need to install anything apart from Docker and the docker-compose program. After downloading the two files as mentioned in the repo above and running one command (`docker-compose up -d`) you can open the browser and try it out. 

The above demo uses Aspire models from the Kaldi website. It works pretty decently (even on a cheap laptop).

I'd appreciate if you let me know if you find this sort of thing useful (I spent a lot of time debugging the damn guy). 

And also, if it's something worth considering, it'd be neat to finally merge this PR so that all of this becomes part of the main fork.

danpovey:
Thanks  a lot for this!!
@hhadian and @dogancan, if and when you have time it would be great to get your comments.
Just from looking at the diff, it seems super self contained and doesn't pull in any weird dependencies, which I like.

danpovey:
Thanks.  Let me wait till Hossein at least glances over it too; it should be good to merge then.

dogancan:
I am doing another round of review. Accepted the previous one so @danijel3 will have an easier time telling the new ones.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+    po.Read(argc, argv);
+
+    if (po.NumArgs() != 4) {
+      po.PrintUsage();
+      return 1;
+    }
+
+    std::string nnet3_rxfilename = po.GetArg(1),
+        fst_rxfilename = po.GetArg(2),
+        word_syms_rxfilename = po.GetArg(3),
```
dogancan: word_syms cannot be read from an rxfilename, it need to be regular filename.

word_syms_rxfilename -> word_syms_filename

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+    // to modify the nnet to accept iVectors at intervals.
+    nnet3::DecodableNnetSimpleLoopedInfo decodable_info(decodable_opts,
+                                                        &am_nnet);
+
+    KALDI_LOG << "Loading FST...";
+
+    fst::Fst<fst::StdArc> *decode_fst = ReadFstKaldiGeneric(fst_rxfilename);
+
+    fst::SymbolTable *word_syms = NULL;
+    if (!word_syms_rxfilename.empty())
```
dogancan: word_syms_rxfilename -> word_syms_filename

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+      if (!(word_syms = fst::SymbolTable::ReadText(word_syms_rxfilename)))
```
dogancan: word_syms_rxfilename -> word_syms_filename

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+        KALDI_ERR << "Could not read symbol table from file "
+                  << word_syms_rxfilename;
```
dogancan: word_syms_rxfilename -> word_syms_filename

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+    KALDI_LOG << "Loaded evertyhing. Waiting for data...";
```
dogancan: evertyhing -> everything

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+    signal(SIGPIPE, SIG_IGN); //ignore SIGPIPE to avoid crashing when socket forecfully disconnected
```
dogancan: socket forecfully disconnected -> socket is forcefully disconnected

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+    server.Listen(port_num);
+
+    while (true) {
+
+      server.Accept();
+
+      int32 samp_count = 0;//this is used for output refresh rate
+      size_t chunk_len = static_cast<size_t>(chunk_length_secs * samp_freq);
+
+      int32 check_update = static_cast<int32>(samp_freq * output_freq);
```
dogancan: output_freq -> output_period

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+            decoder.FinalizeDecoding();
+            frame_offset += decoder.NumFramesDecoded();
+
+            if (decoder.NumFramesDecoded() > 0) {
+              CompactLattice lat;
+              decoder.GetLattice(true, &lat);
+
+              std::string msg = LatticeToString(lat, word_syms);
+
+              server.Write(msg + "\n");
```
dogancan: See comment above about eol argument.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+    while (true) {
+
+      server.Accept();
+
+      int32 samp_count = 0;//this is used for output refresh rate
+      size_t chunk_len = static_cast<size_t>(chunk_length_secs * samp_freq);
+
+      int32 check_update = static_cast<int32>(samp_freq * output_freq);
+      int32 next_check = check_update;
```
dogancan: Maybe it would be better to rename these variables:
check_update -> check_period
next_check -> check_count

src/online2bin/online2-tcp-nnet3-decode-faster.cc
dogancan: Did you have an opportunity to check if silence weighting works correctly?

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+                std::string msg = LatticeToString(lat, word_syms);
+
+                server.Write(msg + '\r');
+              }
+
+              next_check += check_update;
+            }
+
+          } else {
```
dogancan: I think it would be better/simpler to change this else clause to `if (eos)` and move it above since it ends with a break statement. So the new logic would be

```C++
while (true) {
  eos = !server.ReadChunk(chunk_len);

  if (eos) {
    // handle end of stream
    break;
  }

  Vector<BaseFloat> wave_part = server.GetChunk();
  // handle chunk

  if (decoder.EndpointDetected(endpoint_opts)) {
    // handle endpoint
  }
}
```

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+  ~TcpServer();
+
+  bool Listen(int32 port);  //start listening on a given port
+  int32 Accept();  //accept a client and return its descriptor
+
+  bool ReadChunk(size_t len); //get more data and return false if end-of-stream
+
+  Vector<BaseFloat> GetChunk(); //get the data read by above method
+
+  bool Write(const std::string &msg); //write to accepted client
```
dogancan: I think it would be better to accept an eol argument (which would be either \r or \n in current code) to append to the end of the message so we avoid unnecessary string concatenation.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+            if (decoder.NumFramesDecoded() > 0) {
+              CompactLattice lat;
+              decoder.GetLattice(true, &lat);
+
+              std::string msg = LatticeToString(lat, word_syms);
+
+              server.Write(msg + "\n");
+            } else
+              server.Write("\n");
```
dogancan: See comment above about eol argument.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+            decoder.AdvanceDecoding();
+
+            if (samp_count > next_check) {
+              if (decoder.NumFramesDecoded() > 0) {
+                Lattice lat;
+                decoder.GetBestPath(false, &lat);
+
+                std::string msg = LatticeToString(lat, word_syms);
+
+                server.Write(msg + '\r');
```
dogancan: See comment above about eol argument.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+          if (decoder.EndpointDetected(endpoint_opts)) {
+
+            decoder.FinalizeDecoding();
+            frame_offset += decoder.NumFramesDecoded();
+
+            CompactLattice lat;
+            decoder.GetLattice(true, &lat);
+            std::string msg = LatticeToString(lat, word_syms);
+
+            server.Write(msg + "\n");
```
dogancan: See comment above about eol argument.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+  buf.Resize(static_cast<MatrixIndexT>(has_read_));
+
+  for (int i = 0; i < has_read_; i++)
+    buf(i) = static_cast<BaseFloat>(samp_buf_[i]);
+
+  return buf;
+}
+
+bool TcpServer::Write(const std::string &msg) {
```
dogancan: See comment above about eol argument.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+
+      int32 check_update = static_cast<int32>(samp_freq * output_freq);
+      int32 next_check = check_update;
+
+      int32 frame_offset = 0;
+
+      bool eos = false;
+
+      OnlineIvectorExtractorAdaptationState adaptation_state(
+          feature_info.ivector_extractor_info);
```
dogancan: If we are not going to have the option of carrying adaptation state between different recordings, we can remove this line since as it is adaptation_state is not being used anywhere.

hhadian:
Will review it in a few days - - I am traveling right now.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,457 @@
...
+      nnet3::CollapseModel(nnet3::CollapseModelConfig(), &(am_nnet.GetNnet()));
+    }
+
+    // this object contains precomputed stuff that is used by all decodable
+    // objects.  It takes a pointer to am_nnet because if it has iVectors it has
+    // to modify the nnet to accept iVectors at intervals.
+    nnet3::DecodableNnetSimpleLoopedInfo decodable_info(decodable_opts,
+                                                        &am_nnet);
+
+    KALDI_LOG << "Loading FST...";
```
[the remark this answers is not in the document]
danijel3: Okay, I agree with the rest, but just like to point out that this step is in practice usually most time consuming step of initialization. I'll set it to KALDI_VLOG(1) for now.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: I'll remove that line as it is usually unnoticed anyway. The "Waiting for client..." line is most essential.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: I can't seem to find what you are referring to...

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Yea, I'll have to figure out how to test this. The models that I have don't rely on silence weighing.

I'll fix the code as suggested.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Ok, to make code simple, I'll add a WriteLn function with an optional eol argument and call Write from within that function.

danijel3:
@dogancan I made all the modifications mentioned above.

danijel3:
Just to expand on the netcat issue, mentioned a while back, somewhere above. It is true that netcat doesn't send shutdown automatically on EOF because certain servers (when they can interpret the contents of the data they receive) the server is the one that shuts down the connection. In our case, we accept raw streaming audio and can't determine if it's done or not. Only way to determine that would be to introduce some sort of "protocol", which is something I really wanted to avoid this time. This means, that the client is responsible for terminating the connection and not the server. Thankfully, netcat has the '-N' switch to do just that. Furthermore, the timeout option will remain available if someone really wants to use it, although personally, I'd avoid it if possible. It has to be set pretty high to avoid any false alarms and that somewhat defeats the purpose of having the timeout in the first place.

src/doc/online_decoding.dox
```diff
@@ -438,6 +438,70 @@ and downloadable models that can be used with online nnet3 decoding, please
 see http://kaldi-asr.org/models.html (the first model there, the ASPIRE model,
 includes instructions in a README file).
 
+\subsection online_decoding_nnet3_net TCP server for nnet3 online decoding
+
+The program to run the TCP sever is online2-net-nnet3-latgen-faster located in the
```
dogancan: Seems like we missed one :)

online2-net-nnet3-latgen-faster -> online2-tcp-nnet3-decode-faster

Also let's rename the subsection to online_decoding_nnet3_tcp.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,458 @@
...
+ private:
+  struct ::sockaddr_in h_addr_;
+  int32 server_desc_, client_desc_;
+  int16 *samp_buf_;
+  size_t buf_len_, has_read_;
+  pollfd client_set_[1];
+  int read_timeout_;
+};
+
+std::string LatticeToString(const Lattice &lat, const fst::SymbolTable *word_syms) {
```
dogancan: Let's make word_syms a constant reference instead of a constant pointer. Same for the other version of this function.

dogancan:
@danijel3 I think the PR looks pretty good. I added a couple of new comments but nothing major.

@danpovey What is your view on having a command line option to carry the adaptation_state between recordings?

src/online2bin/online2-tcp-nnet3-decode-faster.cc
danijel3: Ok, I did this, but then I noticed that "word_sysm" is created as a pointer (line 193) using the ReadText function. Shall we leave it like before or derefence it every time we call LatticeToString?

src/doc/online_decoding.dox
danijel3: fixed

src/online2bin/online2-tcp-nnet3-decode-faster.cc
dogancan: I think it is best to create word_syms as a pointer and dereference it when calling LatticeToString.

danpovey:
Thanks.  I hope to merge in a few days, once Hossein has a quick look and confirms that he doesn't still have any concerns.

danpovey:
sorry pressed wrong button-- reopened again.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,458 @@
...
+              decoder.GetLattice(true, &lat);
+
+              std::string msg = LatticeToString(lat, *word_syms);
+
+              server.WriteLn(msg);
+            } else
+              server.Write("\n");
+
+            server.Disconnect();
+            eos = true;
```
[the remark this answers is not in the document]
danijel3: Are you referring to line 252 above? The one sending a single newline character?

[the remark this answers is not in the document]
dogancan: I think he is referring to line 255.

danijel3:
> Some very minor comments and a question: I know this program is single-threaded but I was wondering what would happen when a second client tries to connect while a client is already connected? It seems to me that the 2nd client will be rejected, is that right?

Yes. The server is single threaded and will reject any clients that try to connect after the first one. It's by design as this program isn't meant to distribute the decoding resources. It's just a way to offload the ASR processing to a different process or machine. The TCP/IP is just a method for IPC here.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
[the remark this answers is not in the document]
danijel3: Oh yea, duh...

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -0,0 +1,458 @@
...
+            feature_info.silence_weighting_config,
+            decodable_opts.frame_subsampling_factor);
+
+        std::vector<std::pair<int32, BaseFloat>> delta_weights;
+
+        while (true) {
+
+          eos = !server.ReadChunk(chunk_len);
+
+          if (eos) {
```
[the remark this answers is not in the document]
danijel3: > @danijel3 I think the simplest solution for this is to change the return statement at line 415 to
> 
> ```
> return has_read_ > 0;
> ```
> What do you think?

Yea, I guess that's a better way of doing it.

hhadian:
This looks good to me.

danpovey:
Thanks a lot, guys!

PR #2942, opened 2018-12-26
Yomdle Korean OCR setup

aarora8:
It's using 30k utterances from yomdle English and Korean dataset. It is getting the line image from the MAR (minimum area rectangle)

- A Korean Hangul to Jamo decomposition: changing the lexicon to represent a character/sub-word as a sequence of decomposed character instead of a sequence of characters. It effectively reduces the number of symbols to recognize from 3000 to around 250.
- It is getting the line image from the MAR (minimum area rectangle).
- BPE based model with vocab size=700
- It is currently building the sub-word language model from corpus text shared during Scale workshop (5M lines). It is performing 6-gram pruned decoding and 6-gram unpruned rescoring.
- shared model with English and Korean line images.
- semi-supervised training (22k untranscribed line images)
- normalized scoring scripts
- decoding beam = 12


danpovey:
Thanks!
If you can find someone to check it and make sure it runs, that would be great.  See if anyone you are working with on this type of thing is willing to run it.
Dan

desh2608:
I have verified these results.

PR #2951, opened 2018-12-30
Spanish dnn rnn recipe

GoVivace:
Improved WER


egs/fisher_callhome_spanish/s5/local/chain/run_tdnn_1g.sh
```diff
@@ -0,0 +1,282 @@
+#!/bin/bash
+
+# 1g is like 1f but upgrading to a "resnet-style TDNN-F model", i.e.
+#   with bypass resnet connections, and re-tuned.
```
danpovey: Can you please include the WER results in this file, with comparison to the old one if possible, using the standard 'compare_wer.sh' type of scripts that we normally put in?

GoVivace: Will it be okay to skip a comparison. The problem is that The old recipe didn't quite work for me. So if we put a comparison in, that we don't know what we are comparing to. In the new recipe, there is a 1% accuracy difference between the latest run and the run that was done prior to check in. However, I can maybe report the current WER - that is replicated on JHU machines. Will that suffice?

danpovey:
OK, you can report the latest WER.

egs/fisher_callhome_spanish/s5/local/chain/run_tdnn_1g.sh
GoVivace: done

PR #2955, opened 2019-01-01
WIP: Using TDNN acoustic model and 4-gram big-dict baseline

desh2608:


egs/wsj/s5/local/rnnlm/tuning/run_lstm_tdnn_1c.sh
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash
+
+# Copyright 2012  Johns Hopkins University (author: Daniel Povey)  Tony Robinson
+#           2017  Hainan Xu
+#           2017  Ke Li
+
+# This is the same as 1b, with two exceptions: TDNN is used as the acoustic model,
+# and the baseline is big-dict 4-gram LM.
+
+# System                    lstm_tdnn_1b lstm_tdnn_1c
```
danpovey: Please clarify what command produced this output.

desh2608: I'm not sure if we can specify an affix for the RNNLM type used, so both results are obtained using `local/chain/compare_wer.sh exp/chain/tdnn1h_sp` but in different `chain` dirs (named manually).

danpovey: I don't understand what you mean.  I think you should talk to Hainan or one of the other more senior students and come up with a more specific proposal (or just solve the problem some other way).  In any case, any time WERs are displayed, the command that generated those WERs should be displayed, or if it was stitched together manually somehow, then explain how.

egs/wsj/s5/local/rnnlm/tuning/run_lstm_tdnn_1c.sh
```diff
@@ -0,0 +1,143 @@
...
+# Copyright 2012  Johns Hopkins University (author: Daniel Povey)  Tony Robinson
+#           2017  Hainan Xu
+#           2017  Ke Li
+
+# This is the same as 1b, with two exceptions: TDNN is used as the acoustic model,
+# and the baseline is big-dict 4-gram LM.
+
+# (Manually rename chain dirs for lstm_tdnn_1b and lstm_tdnn_1c as 
+# chain1b and chain1c)
+# local/chain/compare_wer.sh exp/chain1b/tdnn1g_sp exp/chain1c/tdnn1g_sp
```
danpovey: I think something is not right here.  You are just comparing two different runs of the tdnn1g model without RNNLM rescoring.  You should talk to a more senior student and have them explain to you how this works.

desh2608: Ok I'll talk to Ke about this.

PR #2960, opened 2019-01-03
Remove duplicate test mode option from class GeneralDropoutComponent

chihyang:
The duplicate member `GeneralDropoutComponent::test_mode_` shadows `RandomComponent::test_mode_` and makes `SetDropoutProportion` not work. This PR removes the former from `GeneralDropoutComponent`.

### DIFF
--- a/src/nnet3/nnet-general-component.h
+++ b/src/nnet3/nnet-general-component.h
@@ -910,8 +910,6 @@ class GeneralDropoutComponent: public RandomComponent {
 
   bool continuous_;
 
-  bool test_mode_;
-
   const GeneralDropoutComponent &operator
   = (const GeneralDropoutComponent &other); // Disallow.
 };


danpovey:
Thanks.  Please let me know when you have a run a recipe with this fix, to check it doesn't break somehow.

chihyang:
Hi Daniel,

We trained and tested the model with recipe `egs/wsj/s5/local/chain/egs/wsj/s5/local/chain`, here is the result with test mode off and on respectively:


- Test Mode Off

Sentence |    Word | Correct(%) | Substitute(%d) | Deletion(%) | Insertion(%) | Word Error(%) | Sentence Error(%) 
-------|--------|------|-----|-----|-----|-----|-------
19388 | 155237 | 96.3 | 3.3 | 0.3 | 0.2 | 3.9 |  18.2 

- Test Mode On

Sentence |    Word | Correct(%) | Substitute(%d) | Deletion(%) | Insertion(%) | Word Error(%) | Sentence Error(%) 
-------|--------|------|-----|-----|-----|-----|-------
19388 | 155237 | 96.3 | 3.4 | 0.3 | 0.2 | 3.9 |  18.2 

Seems this option is not a big problem even if it is not correctly set. I am a little curious about the dropout schedule. At the end of training, I found every dropout component shares a small dropout proportion (about 0.001) but not zero (the following is part from our model).

```
<ComponentName> tdnn1.dropout <GeneralDropoutComponent> <Dim> 1536 <BlockDim> 1536 <TimePeriod> 0 <DropoutProportion> 0.001423042 <Continuous> </GeneralDropoutComponent>
```

Shouldn't `DropoutProportion` be zero? Or this is by design (a number close to zero)?

danpovey:
OK, I'll merge.  The code that gets the dropout proportion uses floats, it may just be a roundoff effect.  Anyway it doesn't matter as that number is so tiny.

PR #2962, opened 2019-01-03
[build] Fix nvcc verbose flag in cuda builds

ryanleary:
Move verbose flag introduced in #2944 to nvcc commands not not gcc commands.


galv:
@danpovey This is ready.

PR #2964, opened 2019-01-03
madcat Arabic (optional extra corpus text for LM), IAM (making word-based and BPE based setup similar),  madcat Chinese (clean scripts, tuning)

aarora8:
Madcat Arabic:
- Adding an optional subset of Arabic Gigaword corpus text for 6-gram sub-word language modeling. It improved the WER from 8.05 to 5.67.

IAM (word-based setup):
- adding smaller HMM/GMM (4 states instead of 8) punctuation topology
- augmentation (shearing, contrast normalization)
- parameter tuning of chain scripts.

Madcat Chinese:
- parameter tuning of chain scripts
- cleaning scripts, adding tuning directory


aarora8:
@danpovey, I guess, it is ok for review/merge.

egs/iam/v1/local/chain/tuning/run_cnn_chainali_1b.sh
```diff
@@ -22,6 +22,7 @@ stage=0
 
 nj=30
 train_set=train
+decode_val=true
```
danpovey: is there a reason there are no results shown herE?

aarora8: I started the training scripts, I will surely update the result today.

PR #2972, opened 2019-01-06
[src] Compute a separate avg (start, end) interval for each sausage word

dogancan:
This PR implements the complete time information extension described in Appendix C.2 of ["Minimum Bayes Risk decoding and system combination based on a recursion for edit distance"](http://danielpovey.com/files/csl11_consensus.pdf), Haihua Xu, Daniel Povey, Lidia Mangu and Jie Zhu, Computer Speech and Language, 2011.

After this change, MBR decoding algorithm keeps a separate time interval for each word in a sausage bin and the words on best MBR hypothesis receive these per-word averages instead of per-bin averages.

### DIFF
--- a/src/lat/sausages.h
+++ b/src/lat/sausages.h
@@ -2,6 +2,7 @@
 
 // Copyright 2012  Johns Hopkins University (Author: Daniel Povey)
 //           2015  Guoguo Chen
+//           2019  Dogan Can
 
 // See ../../COPYING for clarification regarding multiple authors
 //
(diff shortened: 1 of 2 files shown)


src/lat/sausages.h
```diff
@@ -222,15 +232,20 @@ class MinimumBayesRisk {
   // paper.  We sort in reverse order on the second member (posterior), so more
   // likely word is first.
 
-  std::vector<std::pair<BaseFloat, BaseFloat> > times_;
+  std::vector<std::vector<std::pair<BaseFloat, BaseFloat> > > times_;
+  // The average start and end times for words in each confusion-network bin.
+  // This is like an average over arcs, of the tau_b and tau_e quantities in
```
danpovey: This comment looks like it was copied from below.  Is it correct?
The indexing is not clear to me (also the original code).

dogancan: Yes, the comment is correct. What used to be `times_` is now called `sausage_times_`.

```
times_          -> tau_b(q, w), tau_e(q, w)
sausage_times_  -> tau_b(q), tau_e(q)
```

danpovey:
So basically the only effect of this change is that the times are more
accurate?
Do you think I need to test this?

dogancan:
Yes, that is the intended effect. I tested it on some datasets I was working with and I saw better time boundaries overall. Especially in cases where audio is noisy or models are relatively weak. I started looking into this when I noticed time boundaries for some sausage bins did not contain the best words in those bins. I think this change should be a net positive.

danpovey:
merging.  the test failure is a travis-parallel-make issue, unrelated.

Issue #2225, opened 2018-02-16
Sed compatibility problems


danpovey:
I hope someone can find the time to help with this.
This is a sed compatibility issue that affects Macs and other BSD.  Commands like the following:
`sed -i'' 's/#1$//g' $lout/lexicon.txt`
wil fail on mac, due to the command being interpreted as an extension argument to the -i option.  (The '' is not seen by sed).  I have looked into this and there appears to be no way to get this to work in a compatible way without adding an extension.  I suggest to do this as, for instance:
```
sed -i.bak 's/#1$//g' $lout/lexicon.txt
rm $lout/lexicon.txt.bak 
```
... or you can remove the backup file later in the script, or perhaps better, explicitly reference the original location so it's not in-place at all.

There are quite a few `sed -ir` commands.   I don't know how those arose; 'r' is a strange backup extension and the letter 'r' is certainly not interpreted as an option (and anyway that option doesn't exist on mac).  I think those r's should be changed to .bak and the backup removed if posible.

I'd like to warn people that sed should be considered HIGHLY SUSPECT from a compatibility point of view, especially if you are using command-line switches.  perl is preferable if possible.

See below for affected locations, but please run this yourself with larger maxdepth in case there are more.
```
cd egs/wsj/s5
find ../.. -maxdepth 6 -name '*.sh' -exec grep 'sed -i' {} \; -print 2>/dev/null
     sed -ir "s!$p1!$p2!" $dir/segments
../../ami/s5/local/ami_ihm_scoring_data_prep.sh
     sed -ir "s:$p1:$p2:" $tmpdir/segments
../../ami/s5/local/ami_mdm_scoring_data_prep.sh
     sed -ir "s:$p1:$p2:" $tmpdir/segments
../../ami/s5/local/ami_sdm_scoring_data_prep.sh
     sed -ir "s!$p1!$p2!" $dir/segments
../../ami/s5b/local/ami_ihm_scoring_data_prep.sh
     sed -ir "s:$p1:$p2:" $tmpdir/segments
../../ami/s5b/local/ami_mdm_scoring_data_prep.sh
     sed -ir "s:$p1:$p2:" $tmpdir/segments
../../ami/s5b/local/ami_sdm_scoring_data_prep.sh
  sed -i "s/_non_normalized//g" $file
../../aspire/s5/local/multi_condition/prepare_impulses_noises.sh
sed -i'' 's/#1$//g' $lout/lexicon.txt
[...]

PR #2981, opened 2019-01-09
Sed compatibility 

desh2608:
To solve SED compatibility issue https://github.com/kaldi-asr/kaldi/issues/2225

Replaced all occurences of `sed -i` with `perl -i -pe`.
Tested on AMI recipe and seems to be working fine.


misc/maintenance/fix_cpplint_whitespace.sh
```diff
@@ -3,11 +3,11 @@
...
-  sed -i 's/} \/\//}  \/\//' $file # '} //' -> '}  //'
-  sed -i 's/for(/for (/' $file     # 'for(' -> 'for ('
-  sed -i 's/if(/if (/' $file       # 'if(' -> 'if ('
-  sed -i 's/\s\s*$//' $file        # 'remove white-space at the end of lines'
+  perl -i -pe 's/; \/\//;  \/\//' $file # '; //' -> ';  //'
+  perl -i -pe 's/{ \/\//{  \/\//' $file # '{ //' -> '{  //'
+  perl -i -pe 's/} \/\//}  \/\//' $file # '} //' -> '}  //'
+  perl -i -pe 's/for(/for (/' $file     # 'for(' -> 'for ('
+  perl -i -pe 's/if(/if (/' $file       # 'if(' -> 'if ('
+  perl -i -pe 's/\s\s*$//' $file        # 'remove white-space at the end of lines'
```
danpovey: I tested this and it seemed to  remove the final newline.

tools/extras/install_portaudio.sh
```diff
@@ -82,7 +82,7 @@ if [ -z "$MACOS" ]; then
 fi
 
 ./configure --prefix=`pwd`/install --with-pic
-sed -i.bk '40s:src/common/pa_ringbuffer.o::g; 40s:$: src/common/pa_ringbuffer.o:' Makefile
+perl -i.bk -pe '40s:src/common/pa_ringbuffer.o::g; 40s:$: src/common/pa_ringbuffer.o:' Makefile
```
danpovey: This didn't seem to work for me, perl said syntax error.
Sed and perl syntax can be quite different.

egs/yomdle_zh/v1/local/create_download.sh
```diff
@@ -43,4 +43,4 @@ local/create_line_image_from_page_image.py \
 
 echo "Downloading table for CangJie."
 wget -P $download_dir/ $cangjie_url || exit 1;
-sed -ie '1,8d' $download_dir/cj5-cc.txt
```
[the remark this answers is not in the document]
desh2608: Okay, I figured it out. This is supposed to delete lines 1 to 8 from the file.

tools/extras/install_portaudio.sh
desh2608: Yes, thanks for pointing this out. I am making this and other modifications now. Some of the other edits such as deleting matching patterns are easier using `grep -v` than with `perl`.

desh2608:
Made some further modifications:
* `grep -v` to delete strings matching a pattern
* Changed `perl` syntax for replacing strings or deleting specific line numbers
* `ex -s -c` (Vim executable command) to add a line at beginning of file

danpovey:
@desh2608  there is a conflict here, could you please resolve it?

Also it would be nice to have someone experienced to look over these changes and make sure they look safe.

desh2608:
Conflicts are resolved, but yes, perhaps it would be best if someone looks over the changes.

danpovey:
@kkm000 do you have time to check over this?

kkm000:
@danpovey Will do. So the goal is to get rid of sed entirely, because of differences in the wild, right?

kkm000:
@desh2608, give me about an hour or two to get to this. I spy a a couple problems in the change. One is grep cannot modify a file in-place, like `sed -i` or `perl -i`, but we can do everything in perl. Another is a subtle difference regarding the handling of `\n` at the end. In sed, the final `\n` is not whitespace but in perl it is, so deleting whitespace at EOL will glue lines together in perl but not in sed. I'm sure both your grep and ex uses could be replaced with perl. ex would be a whole new can of worms to open (it can be vi, vim, or, heaven save, the original ex, "a better ed", that both vi and vim emulate, with better or worse accuracy), let's better not use it! : ) 

perl has practically no quirks across systems, AFAIK, so it's a good idea to use it alone.

misc/maintenance/fix_cpplint_whitespace.sh
kkm000: @danpovey, as a project to use clang-tidy is under way, maybe we should not touch this file at all? It's a maintainer's file, so presumably sed is fine here for the time being, only a handful of people run it if ever.

danpovey:
That makes sense, you can leave that maintainer's file alone.

egs/ami/s5b/local/ami_sdm_scoring_data_prep.sh
```diff
@@ -122,7 +122,7 @@ if [ `cat $tmpdir/segments_to_fix | wc -l` -gt 0 ]; then
   while read line; do
      p1=`echo $line | awk -F'>' '{print $1}'`
      p2=`echo $line | awk -F'>' '{print $2}'`
-     sed -ir "s:$p1:$p2:" $tmpdir/segments
+     perl -i -pe "s:$p1:$p2:" $tmpdir/segments
```
kkm000: This and the 4 above are an exact match to existing behavior. I would only change the separator to `:` everywhere, the mismatch kinda sticks out (and the `!` is special, but not in non-interactive shell, but still I'd avoid it). Another thing I'd add is, to make sure only really the whole line is replaced, `"s:^$p1\$:$p2:"`. Note that the `$` EOL match assertion needs to be quoted `\n` because shell; try `echo "\$"`.

If it works, it works, you may stop here. But if you look at it, the whole script block is written kinda weirdly, tho. First, a file of complete replacements for the segment file is written to file `segment_replacements` (in another inline perl progam), in the form
```
old line 1>new line 1
old line 2>new line 2
```
etc. Then each line is read inside this loop, split into `p1`=_old line_ and `p2`=_new line_, and a separate pass over the segments file is done to replace them. It maybe ok if done once.

If you really want to improve it, you can change that above perl program to write another little perl program instead, so the `segment_replacements` would be output in the form
```
s/^old line 1$/new line 1/;
s/^old line 2$/new line 2/;
```
and so on. Mind the `;` at ends. Then this whole while loop can be replaced with
```
perl -i -pf segment_replacements $tmpdir/segments
```
which will do a single pass over the segments file. The `-f` is to read program from file `segment_replacements` instead of `-e` for an inline program.

This is all up to you; just how I would do it. Depends on your desire to spend little time hacking perl scripts : ). Srsly, I have an impression that @danpovey puts certain weight on the code's having a didactic value, so it may make sense in the end.

egs/babel/s5c/local/syllab/generate_syllable_lang.sh
```diff
@@ -118,8 +118,8 @@ ln -s lex.syllabs2phones.disambig.fst $out/L_disambig.fst
 echo "Validating the output lang dir"
 utils/validate_lang.pl $out || exit 1
 
-sed -i'' 's/#1$//g' $lout/lexicon.txt
-sed -i'' 's/#1$//g' $lout/lexiconp.txt
+perl -i'' -pe 's/#1$//g' $lout/lexicon.txt
+perl -i'' -pe 's/#1$//g' $lout/lexiconp.txt
```
kkm000: Two notes here:
1. Use just `-i`, not `-i''`.
2. The same program can be applied to multiple files with `-i`, so this can be done to avoid copy/paste
```
perl -i -pe 's/#1$//g' $lout/lexicon.txt $lout/lexiconp.txt
```

egs/bentham/v1/local/create_splits.sh
```diff
@@ -28,9 +28,9 @@ function split {
         echo $name $spkid >> $split_dir/utt2spk 
 	done < "$line_file"
     
-    sed -i '/^\s*$/d' $split_dir/images.scp
-    sed -i '/^\s*$/d' $split_dir/text
-    sed -i '/^\s*$/d' $split_dir/utt2spk
+    grep -v -e '^\s*$' $split_dir/images.scp
+    grep -v -e '^\s*$' $split_dir/text
+    grep -v -e '^\s*$' $split_dir/utt2spk
```
kkm000: The intent here is to delete whitespace-only lines, again, in a file in-place. The typical rerl trick for that is to output non-whitespace instead, like this:
```
perl -i -ne 'print if /\S/' $split_dir/images.scp $split_dir/text $split_dir/utt2spk
```
The `\S` stands for "any non-space character" (like any not matched by the lowercase `\s`). Thus, any line with a non-space is preserved.

egs/callhome_egyptian/s5/local/callhome_prepare_dict.sh
```diff
@@ -55,7 +55,7 @@ $dir/extra_questions.txt || exit 1;
 
 # Add prons for laughter, noise, oov
 for w in `grep -v sil $dir/silence_phones.txt`; do
-sed -i "/\[$w\]/d" $tmpdir/lexicon.3
+grep -v -e "\[$w\]" $tmpdir/lexicon.3
```
kkm000: Here, we are deleting lines with phones like `[noise]` from the lexicon. This can be written as
```
perl -i -ne "print unless /\[$w\]/" $tmpdir/lexicon.3
```

Here again, as many passes are done over the file as there are phones to delete. You may want to concoct something that results in e. g. 
```
perl -i -ne "print unless /\[(laughter|noice|clap)\]/"  $tmpdir/lexicon.3
```
seen by perl, to process the file in a single pass. But that's, again, up to you.

egs/gale_arabic/s5/local/gale_prep_dict.sh
```diff
@@ -25,8 +25,7 @@ echo SIL > $dir/optional_silence.txt
 cat $dir/lexicon.txt | cut -d ' ' -f2- | tr -s ' ' '\n' |\
 sort -u >  $dir/nonsilence_phones.txt || exit 1;
 
-
- sed -i '1i<UNK> SIL' $dir/lexicon.txt
+ ex -s -c '1i|<UNK> SIL' -c x $dir/lexicon.txt
```
kkm000: This you can also do with perl alone
```
perl -i -pe 'print "<UNK> SIL\n" if $.==1'  $dir/lexicon.txt
```
Since the printing of the `-p` happens *after* all other processing (see `perldoc perlrun` for the exact loop perl runs with the `-p`), the line is inserted *before* the first (`$. == 1`) line in the file.

egs/reverb/s5/local/download_se_eval_tool.sh
```diff
@@ -18,14 +18,14 @@ unzip REVERB_scores.zip -d local/REVERB_scores_source
...
-sed -i 's/SRMR_main/SRMR/g' prog/score_sim.m
-sed -i 's/+wb\ //g' prog/calcpesq.m
-sed -i 's/pesq_/_pesq_/g' prog/calcpesq.m
-sed -ie '30d;31d' prog/calcpesq.m
+perl -i -pe 's/wavread/audioread/g' SRMRToolbox/libs/preprocess.m
+perl -i -pe 's/SRMR_main/SRMR/g' prog/score_real.m
+perl -i -pe 's/SRMR_main/SRMR/g' prog/score_sim.m
+perl -i -pe 's/+wb\ //g' prog/calcpesq.m
+perl -i -pe 's/pesq_/_pesq_/g' prog/calcpesq.m
+perl -n -i -e 'print if $. != 30 && $. != 31' prog/calcpesq.m  # Delete lines 30 and 31 from file
```
kkm000: Depending on what exactly is deleted, you can prefer matching the lines being deleted by their content rather than line number, using `if /some regex/` instead of `if $. == 31`.

misc/maintenance/fix_cpplint_whitespace.sh
```diff
@@ -3,11 +3,11 @@
...
-  sed -i 's/} \/\//}  \/\//' $file # '} //' -> '}  //'
-  sed -i 's/for(/for (/' $file     # 'for(' -> 'for ('
-  sed -i 's/if(/if (/' $file       # 'if(' -> 'if ('
-  sed -i 's/\s\s*$//' $file        # 'remove white-space at the end of lines'
+  perl -i -pe 's/; \/\//;  \/\//' $file # '; //' -> ';  //'
+  perl -i -pe 's/{ \/\//{  \/\//' $file # '{ //' -> '{  //'
+  perl -i -pe 's/} \/\//}  \/\//' $file # '} //' -> '}  //'
+  perl -i -pe 's/for(/for (/' $file     # 'for(' -> 'for ('
+  perl -i -pe 's/if(/if (/' $file       # 'if(' -> 'if ('
+  perl -i -pe 's/[ \t]+$//' $file       # 'remove white-space at the end of lines'
```
kkm000: We are not changing this file, but generally, prefer single pass over the file. The substututions can be separated with the `;`, like
```
perl -e 's/foo/bar/g; s/baz/quux/; '
```
as many as you want to do over every line.

tools/extras/install_portaudio.sh
```diff
@@ -82,7 +82,7 @@ if [ -z "$MACOS" ]; then
 fi
 
 ./configure --prefix=`pwd`/install --with-pic
-sed -i.bk '40s:src/common/pa_ringbuffer.o::g; 40s:$: src/common/pa_ringbuffer.o:' Makefile
+perl -i.bk -pe 's:src/common/pa_ringbuffer.o:: if $.==40; s:$: src/common/pa_ringbuffer.o: if $.==40' Makefile
```
kkm000: What happens here is a filename removed from a specific Makefile line, which actually is
```
OTHER_OBJS = src/hostapi/oss/pa_unix_oss.o src/os/unix/pa_unix_hostapis.o src/os/unix/pa_unix_util.o src/common/pa_ringbuffer.o
```
To make the script more robust, let's better match it by content (I noted that above regarding an .m file):
```
perl -i -pe 's:src/common/pa_ringbuffer.o:: if /^OTHER_OBJS\s*=/' Makefile
```
The `/^OTHER_OBJS\s*=/` regex applies the substitution only to this assignment line. It is also unnecessary to handle the space specially, as make would not worry if you leave a whitespace at end of line.

Not our case, but if you wanted to remove multiple pattern, the alternative match operator `|` would help, like
```
perl -pe 's:src/common/pa_ringbuffer.o|src/common/something_else.o::g if /^OTHER_OBJS /' Makefile
```
This has a `g` at the end, as the goal is to apply it to all matches in the same string.

Again, this is not our case; just to note a technique.

egs/reverb/s5/local/download_se_eval_tool.sh
```diff
@@ -18,14 +18,14 @@ unzip REVERB_scores.zip -d local/REVERB_scores_source
...
-sed -i 's/wavread/audioread/g' SRMRToolbox/libs/preprocess.m
-sed -i 's/SRMR_main/SRMR/g' prog/score_real.m
-sed -i 's/SRMR_main/SRMR/g' prog/score_sim.m
-sed -i 's/+wb\ //g' prog/calcpesq.m
-sed -i 's/pesq_/_pesq_/g' prog/calcpesq.m
-sed -ie '30d;31d' prog/calcpesq.m
+perl -i -pe 's/wavread/audioread/g' SRMRToolbox/libs/preprocess.m
+perl -i -pe 's/SRMR_main/SRMR/g' prog/score_real.m
+perl -i -pe 's/SRMR_main/SRMR/g' prog/score_sim.m
+perl -i -pe 's/+wb\ //g' prog/calcpesq.m
```
kkm000: Ah, missed this one: the space need not be quoted with the `\`, it stands for itself. But the `+` likely needs to be quoted like `\+`, because it's a regex operator (one or more). Perl will be unhappy, even though it's the first character in the line, where it can't be an operator. sed is apparently ok with that?

egs/ami/s5b/local/ami_sdm_scoring_data_prep.sh
desh2608: Nice catch! I made the change for a single-pass replacement.

desh2608:
@kkm000 Thanks for your detailed comments! I made the changes that you suggested, including deleting by pattern rather than line numbers (except in 1 file where no such pattern existed). Also rewrote some code blocks to make changes in single pass through file rather than looping.

@danpovey I believe this can be merged now.

egs/ami/s5/local/ami_ihm_scoring_data_prep.sh
```diff
@@ -87,19 +87,12 @@ sort -k 2 $dir/utt2spk | utils/utt2spk_to_spk2utt.pl > $dir/spk2utt || exit 1;
...
            if ($pu eq $_[1] && $pt > $_[3]) {
-             print "$_[0] $_[2] $_[3] $_[4]>$_[0] $_[2] $pt $_[4]\n"
+             print "s/^$_[0] $_[2] $_[3] $_[4]\$/$_[0] $_[2] $pt $_[4]/;\n"
            }
-           $pu=$_[1]; $pt=$_[4]; 
+           $pu=$_[1]; $pt=$_[4];
          }' > $dir/segments_to_fix
-if [ `cat $dir/segments_to_fix | wc -l` -gt 0 ]; then
-  echo "$0. Applying following fixes to segments"
-  cat $dir/segments_to_fix
```
kkm000: Maybe it makes better sense to leave the diagnostic output? It's in a different format, but still as human-readable as before. The test can be simplified to `if [ -s $dir/segments_to_fix ]` (`-s` test for for non-empty file).

egs/reverb/s5/local/download_se_eval_tool.sh
```diff
@@ -18,14 +18,14 @@ unzip REVERB_scores.zip -d local/REVERB_scores_source
...
-sed -i 's/SRMR_main/SRMR/g' prog/score_sim.m
-sed -i 's/+wb\ //g' prog/calcpesq.m
-sed -i 's/pesq_/_pesq_/g' prog/calcpesq.m
-sed -ie '30d;31d' prog/calcpesq.m
+perl -i -pe 's/wavread/audioread/g' SRMRToolbox/libs/preprocess.m
+perl -i -pe 's/SRMR_main/SRMR/g' prog/score_real.m
+perl -i -pe 's/SRMR_main/SRMR/g' prog/score_sim.m
+perl -i -pe 's/\+wb //g' prog/calcpesq.m
+perl -i -pe 's/pesq_/_pesq_/g' prog/calcpesq.m
+perl -n -i -e 'print unless /*remove target file name/' prog/calcpesq.m
```
kkm000: When I try that line in bash, perl yells at me for not quoting the `*`, should be `\*` if it's part of the string.

desh2608: I guess I'll just remove the `*`. I thought the entire line needs to be matched so I was using the regex operator, but I checked now and I guess the `print unless /<string>/` allows partial line matching too.

desh2608:
@kkm000 Thanks for the comments. I have made these fixes.

kkm000:
LGTM, thanks!

Not everyone shares my fondness for bash, but I like it as a scripting programming language. It has it quirks, but what language doesn't, and it's a very powerful shell. As such, I appreciated the literally one-liner bashism to get the expression that matches all phones! : )

desh2608:
@danpovey Perhaps this can be merged now before conflicts arise.

danpovey:
Thanks, guys!   Will merge.

egs/ami/s5b/local/ami_sdm_scoring_data_prep.sh
desh2608: @kkm000 revisiting this old comment. The `perl` command was not making the required replacement, but when I changed `s/^old line 1$/new line 1/;` to `s/old line 1/new line 1/;` it worked successfully.

danpovey: @desh2608 I am having trouble understanding what you are saying here, perhaps be more specific.  The dollar sign is supposed to match end-of-line.

desh2608: I found that there was a space at the end. After adding that in the regex, it works fine.

PR #2987, opened 2019-01-12
WIP: [travis] Try splitting two tasks to remove duplicated tasks.

galv:
Not ready for merge. I don't know any way to run Travis other than a PR, so I'll be making commits to here to try to diagnose things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,9 @@ before_script:
   - ccache --zero-stats --max-size=3G
 
 env:
-  - CI_TARGETS="all ext"        # Job1: Build everything.
-  - CI_TARGETS="test"           # Job2: Test libraries. #### ext_test? adds 5min compile, runs 1 test.
+  - CI_TARGETS="all"        # Job1: Build everything.
+  - CI_TARGETS="ext"        # Job2: Build everything.
+  - CI_TARGETS="test"       # Job3: Test libraries. #### ext_test? adds 5min compile, runs 1 test.
 
 script:
   # See http://petereisentraut.blogspot.com/2011/05/ccache-and-clang.html and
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -40,16 +40,17 @@ before_script:
   - ccache --zero-stats --max-size=3G
 
 env:
-  - CI_TARGETS="all ext"        # Job1: Build everything.
-  - CI_TARGETS="test"           # Job2: Test libraries. #### ext_test? adds 5min compile, runs 1 test.
+  - CI_TARGETS="all"        # Job1: Build everything.
+  - CI_TARGETS="ext"        # Job2: Build everything.
```
danpovey: I'd be surprised if that were the issue, because surely 'make -j' would know that two identical tasks generated from different top-level targets were identical.  I thought it was smart.  But I guess we'll see.

galv: It makes sense.

```
all:
    $(MAKE) checkversion
    $(MAKE) kaldi.mk
    $(MAKE) mklibdir
    $(MAKE) subdirs
    $(MAKE) -C matrix test
    -echo Done
```

```
ext: ext_depend $(SUBDIRS) $(EXT_SUBDIRS)
    -echo Done
```

`ext` depends explicitly on SUBDIRS in its Makefile targets, while `all` launches a new make subprocess to `$(MAKE) subdirs` to fulfill that target.

The subprocess has no idea about the `ext` target being run by the parent process, so that's why we have cases where optimization.o (for example)is created twice.

I'm not sure why `all` is defined the way that it is, but I am fairly certain that this change will fix our Travis CI problems.

galv: It's because of PR #1409. We need to run the checkversion and kaldi.mk targets before everything else for correctness. The ordering wouldn't be captured if we just did:

```
all: checkversion kaldi.mk mklibdir subdirs
```

Because -j can begin running those targets in any order.

galv:
Yes, I agree with you on that. I can look at that later.

galv:
@danpovey This looks good to me now. If you look at the current travis log, optimization.o is now built only once.


A note about: `-fuse-ld=gold`: gold is a faster implementation of ld. It should speed up build times marginally.

PR #2993, opened 2019-01-16
Fix nan issue in ctm times

vimalmanohar:


danpovey:
@dogancan can you please see if you think this makes sense?

dogancan:
Yes, it makes sense. Thanks @vimalmanohar!

PR #2996, opened 2019-01-17
Fix 'sausage-time' issue which occurs with disabled MBR decoding.

vesis84:
- the issue was introduced in #2972 and was addressed in #2993

In case of '--decode-mbr=false', the reference word sequence 'R_' is 'frozen'.
And the times in ctm output were sometimes taken from another word in the same bin
(it was taken always from the 1st word in the same bin with best score gamma_).

Now, the times are taken from the 'times_[q][s]', where 's' 
is looked-up to be corresponding to the word in 'R_[q]'.

I am looking forward to see the more accurate time-boundaries thanks to #2972.
The difference is there, but I did not listen to the data yet ;)

### DIFF
--- a/src/lat/sausages.cc
+++ b/src/lat/sausages.cc
@@ -53,7 +53,17 @@ void MinimumBayesRisk::MbrDecode() {
       }
       // build the outputs (time, confidences),
       if (R_[q] != 0 || opts_.print_silence) {
-        one_best_times_.push_back(times_[q][0]);
+        // see which 'item' from the sausage-bin should we select,
+        // (not necessarily the 1st one when MBR decoding disabled)
+        int32 s = 0;
+        for (int32 j=0; j<gamma_[q].size(); j++) {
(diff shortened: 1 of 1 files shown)


danpovey:
Seems to make sense, but @dogancan please have a look also.

dogancan:
This makes sense. I think my mental model of the code when making these changes was that reading times from the sausage was done only if `decode_mbr` was true, but that is definitely not the case. Thanks for the fix @vesis84!

One small comment about the new code. Do we need that separate `s` variable? It seems like one best times can simply be extended inside that if clause. Is it possible for that if clause to never be true for a particular bin?

danpovey:
I'll merge this because I don't like having bugs in the checked-in code.

vesis84:
In theory yes, it could be written without the `s` variable.
But, the way I wrote it, it is easier to follow what is happenning,
although it is on more lines of code...

PR #2997, opened 2019-01-18
[src,scripts] Fix bug in grammar-fst graph building, w.r.t. disambig syms

danpovey:


danpovey:
Did not work; will try #3000 instead.

PR #2998, opened 2019-01-18
Spanish fisher data prep improvement

GoVivace:
Slight improvement to data preparation.


kkm000:
LGTM, thanks. Just before I merge: I understand the data are missing spaces after the `[noise]` token, this is what you are fixing, correct? Generally, stating what problem is solved in the PR description is very helpful. "Slight improvement" is much less so (we hardly open PRs to inflict slight breakage, so they all are improvements in the end :) )

PR #3001, opened 2019-01-19
Allow for a special case in the segments file

samsucik:
Previously, `validate_data_dir.sh` did not allow the `segments` file to contain lines with `end-time=-1`, for example:
`AR042_33-2 AR042_33 0.5 -1`
However, lines like this are valid; citing `extract-segments.cc`: 
> And \<end-time\> of -1 means the segment runs till the end of the WAV file.


danpovey:
Sorry, I can't merge this because those lines were rejected for a good reason.  There are a bunch of scripts that won't work correctly with those types of lines there.

samsucik:
Right, thanks for clarifying this, Dan. I will adapt the script locally for my recipe that works a lot with segments spanning entire utterances.

PR #3012, opened 2019-01-30
IAM, uw3: Adding small fix to also make unk decoding work for non-ascii characters

aarora8:
- fixing bug due to non-ascii characters present in non-English datasets


egs/iam/v1/local/unk_arc_post_to_transcription.py
```diff
@@ -46,11 +47,11 @@
 word_handle = open(args.words, 'r', encoding='latin-1')
 unk_handle = open(args.unk,'r', encoding='latin-1')
 if args.one_best_arc_post == '-':
-    arc_post_handle = sys.stdin
+    arc_post_handle = io.TextIOWrapper(sys.stdin.buffer, encoding='latin-1')
```
danpovey: I am thinking of converting these latin-1 things to utf-8, since it seems to be more painful than I had thought to emulate python2's encoding-agnostic behavior in python3, and I think these days it's safe to assume no-one is using any other encoding.  Can you see if this works with utf-8 as the encoding type?

aarora8: I tried using utf-8 encoding, somehow it is degrading the WER. I am trying to find the reason.

egs/rimes/v1/local/word/score.sh.unk
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
+# Copyright 2012-2014  Johns Hopkins University (Author: Daniel Povey, Yenda Trmal)
+# Apache 2.0
+
+# This script is like steps/scoring/score_kaldi_wer.sh except it transcribes the <unk>'s
+# using local/word/unk_arc_post_to_transcription.py and also it calls
```
danpovey: Please rename this to e.g. score_unk.sh.

egs/rimes/v1/local/word/train_lm.sh.word
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+# Copyright 2016  Vincent Nguyen
```
danpovey: same here.  E.g. train_word_lm.sh.

aarora8:
Sorry, this pull request also has few changes I made to get quick results for the word-based model for comparison with the BPE based system. It does not seem to be in the correct format (not similar to other setups, both BPE and word-based scripts in a single directory). The word-based system is around 13% relative worse as compared to BPE based system. I guess I should remove the changes other than the unk decoding fixes.

danpovey:
Thanks, is this good to merge?

aarora8:
Thanks, yeah, it is ok to merge.

PR #3018, opened 2019-02-01
Fix Visual Studio error due to alternate syntactic form of noreturn

daanzu:


danpovey:
@daanzu  what version of Visual Studio did you get the error on?  Looks like this change may break some versions of gcc.

daanzu:
@danpovey Visual Studio 2017 (15.9.4)

src/base/kaldi-error.cc
```diff
@@ -221,7 +221,7 @@ FatalMessageLogger::FatalMessageLogger(LogMessageEnvelope::Severity severity,
   }
 }
 
-[[ noreturn ]] FatalMessageLogger::~FatalMessageLogger() noexcept(false) {
+FatalMessageLogger::~FatalMessageLogger [[ noreturn ]] () noexcept(false) {
```
danpovey: @daanzu would you mind making a PR to make it more like the following, so we can support both compiler versions?  Obviously test that it works first.

```
#if defined(_MSC_VER)
FatalMessageLogger::~FatalMessageLogger [[ noreturn ]] () noexcept(false) 
#else
[[noreturn]] FatalMessageLogger::~FatalMessageLogger () noexcept(false) 
#endif
{
```

daanzu: 👍

PR #3024, opened 2019-02-06
[src] Add check for the [-30,30] range required by chain denominator

danpovey:


src/chain/chain-training.h
```diff
@@ -62,13 +70,19 @@ struct ChainTrainingOptions {
...
+                          leaky_hmm_coefficient(1.0e-05),
                           xent_regularize(0.0) { }
 
   void Register(OptionsItf *opts) {
     opts->Register("l2-regularize", &l2_regularize, "l2 regularization "
                    "constant for 'chain' training, applied to the output "
                    "of the neural net.");
+    opts->Register("out-of-range-regularize", &out_of_range_regularize,
+                   "Constant that controls how much we penalize the nnet output "
+                   "leaving the range [-30,30]... this is needed because we "
```
kkm000: nit: `"exceeding the range [-30,30]. This is needed because we "` sounds clearer to me, and the ellipsis just separates two complete sentences.

xiaohui-zhang:
do we still plan to merge this?

danpovey:
Thanks, I changed that string.  I hadn't decided to merge this as I had never seen conclusive evidence that it was helping, but it seems safer, on balance, to merge it than not.  I'll merge it now.

PR #3025, opened 2019-02-06
Fixes OnlineFeInput, which was broken by commit cc2469e8.

jdieguez:
We have reverted just the parts affecting OnlineFeInput.

Since ExtractWaveformRemainder was removed in commit [1180e46](https://github.com/kaldi-asr/kaldi/commit/1180e467c8ca273c7704199bd27cb734509e931e), we used that code directly in OnlineFeInput<E>::Compute().

We have also created a fake FrameExtractionOptions so that NumFrames() will still work as expected.

Note that we also support the snip_edges=false case (OnlineFeInput constructor now takes an additional optional parameter).

**NOTE**: This proposal does not need ExtractWaveformRemainder(). But it'd probably be cleaner to just revert  1180e46 and call ExtractWaveformRemainder() before Compute().

### DIFF
--- a/src/online/online-feat-input.h
+++ b/src/online/online-feat-input.h
@@ -31,6 +31,7 @@
 
 #include "online-audio-source.h"
 #include "feat/feature-functions.h"
+#include "feat/feature-window.h"
 
 namespace kaldi {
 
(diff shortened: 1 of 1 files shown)


jdieguez:
THis proposal does not need ExtractWaveformRemainder.

But I think it's probably cleaner to revert  1180e46 to recover ExtractWaveformRemainder(), and just call it before Compute().

danpovey:
OK.  This part of the code is not much used any more, so I think I am willing to take your work that this fixes the issue.  It's ready for merge, right?

jdieguez:
Thanks! Yes, it’s ready to merge.

PR #3031, opened 2019-02-11
update the docs:

vesis84:
- update the history (inputs from Honza Cernocky),
- fix URLs in nnet1 docs,
- add 'usage' to cuda-gpu-available.


src/doc/history.dox
```diff
@@ -64,11 +65,21 @@
  continued working to complete the toolkit and get a working set of training scripts.
  The code was released on May 14th, 2011.
 
+ An official announcement was done in an unofficial event at a meeting-room 
+ during ICASSP 2011 in Prague on Friday 27th May. Although the event was not 
+ largely advertized, the room was full and people were sitting even on the floor.
```
kkm000: s/advertized/advertised/, I believe. Quite fascinatingly, we did not corrupt this verb on this size of the pond. : )

src/nnetbin/cuda-gpu-available.cc
```diff
@@ -40,6 +40,17 @@ void TestGpuComputation() {
 #endif
 
 int main(int argc, char *argv[]) try {
+
+  /* only for Doxygen documentation, never shown in command line */
+  const char *usage =
+        "Test if there is a GPU available, and if the GPU setup is correct.\n"
+        "A GPU is acquired and a small computation is done\n."
```
kkm000: Extra `.` after `\n`.

src/doc/history.dox
danpovey: Guys, I would like to delete these two extra paragraphs.  I just think it's more detail than people would care to know.  I am OK with the other changes.

vesis84:
Hello, 
I updated the commit according to the comments (Dan, I shortened the paragraph to the bare minimum, saying it was presented in ICASSP 2011 plus the link to the recordings).
Thank you for the reviews!
K.

PR #3034, opened 2019-02-15
fix a typo for natural gradient

ahban:
this commit is simple. but it is essential for some people trying to understand the natural gradient method.

### DIFF
--- a/src/nnet3/natural-gradient-online.h
+++ b/src/nnet3/natural-gradient-online.h
@@ -375,7 +375,7 @@ namespace nnet3 {
    * Initialization *
 
    Now, a note on what we do on time t = 0, i.e. for the first minibatch.  We
-   initialize X_0 to the top R eigenvectors of 1/N X_0 X_0^T, where N is the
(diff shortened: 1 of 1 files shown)


src/nnet3/natural-gradient-online.h
```diff
@@ -375,7 +375,7 @@ namespace nnet3 {
    * Initialization *
 
    Now, a note on what we do on time t = 0, i.e. for the first minibatch.  We
-   initialize X_0 to the top R eigenvectors of 1/N X_0 X_0^T, where N is the
+   initialize R_0 to the top R eigenvectors of 1/N X_0 X_0^T, where N is the
```
danpovey: I think there might be another problem here.  I believe X is (N by D) where N is the minibatch size.  (Although elsewhere I seem to have used M for the minibatch size so that might be a better letter).  We want the eigenvectors of a D x D matrix, so it should be X_0^T X_0.    That appears in the lines immediately below, too.

danpovey:
closing this because I have some more extensive fixes I want to merge about those comments.

PR #3037, opened 2019-02-18
fix rand_r for Android

keli78:


danpovey:
Thanks, but this is about the re-entrant random number generators, which
avoids locks.  (rand_r).  It's about only calling rand_r on platforms where
it's supported.  I'm not sure that I want to swap rand() for std::rand(),
assuming they are different, as there isn't any problem that we've observed
with rand().

src/base/kaldi-math.cc
```diff
@@ -42,7 +42,7 @@ int32 RoundUpToNearestPowerOfTwo(int32 n) {
 static std::mutex _RandMutex;
 
 int Rand(struct RandomState* state) {
-#if defined(_MSC_VER) || defined(__CYGWIN__)
+#if !defined(_POSIX_THREAD_SAFE_FUNCTIONS)
```
danpovey: OK, I have been testing this using commands like
```
g++ -E <<EOF
#include <unistd.h>
_POSIX_THREAD_SAFE_FUNCTIONS
EOF
```
and it turns out that if you just include stdlib.h where rand_r is declared, the macro _POSIX_THREAD_SAFE_FUNCTIONS is not defined, you need to include unistd.h.  
 Now, we likely get this include indirectly from other files, but it would be good practice to
include it above, right after `#include <stdlib.h>`.

keli78: ok. you suggest to include this unix standard header as below?
```suggestion
#ifndef _MSC_VER
#include <stdlib.h>
#include <unistd.h>
#endif
```

danpovey: yes

keli78: just added this header.

PR #3038, opened 2019-02-18
Memory optimization for online feature extraction of long recordings

pzelasko:
It introduces a configurable maximum size of the vector which keeps the MFCC/PLP/filterbank feature vectors - the last N elements are retained, but are accessible with the original indexing. By default it doesn't remove anything i.e. the behaviour doesn't change.

### DIFF
--- a/src/feat/online-feature-test.cc
+++ b/src/feat/online-feature-test.cc
@@ -381,13 +381,15 @@ void TestRecyclingVector() {
   for (int i = 0; i != 100; ++i) {
     Vector <BaseFloat> data(1);
     data.Set(i);
-    full_vec.Store(new Vector<BaseFloat>(data));
-    shrinking_vec.Store(new Vector<BaseFloat>(data));
+    full_vec.PushBack(new Vector<BaseFloat>(data));
+    shrinking_vec.PushBack(new Vector<BaseFloat>(data));
   }
+  KALDI_ASSERT(full_vec.Size() == 100);
+  KALDI_ASSERT(shrinking_vec.Size() == 100);
 
   // full_vec should contain everything
   for (int i = 0; i != 100; ++i) {
-    Vector <BaseFloat> *data = full_vec.Retrieve(i);
+    Vector <BaseFloat> *data = full_vec.At(i);
     KALDI_ASSERT(data != nullptr);
     KALDI_ASSERT((*data)(0) == static_cast<BaseFloat>(i));
   }
@@ -396,7 +398,7 @@ void TestRecyclingVector() {
   int caught_exceptions = 0;
   for (int i = 0; i != 90; ++i) {
     try {
-      shrinking_vec.Retrieve(i);
+      shrinking_vec.At(i);
     } catch (const std::runtime_error &) {
       ++caught_exceptions;
     }
@@ -406,7 +408,7 @@ void TestRecyclingVector() {
 
   // shrinking_vec should contain the last 10 elements
   for (int i = 90; i != 100; ++i) {
-    Vector <BaseFloat> *data = shrinking_vec.Retrieve(i);
+    Vector <BaseFloat> *data = shrinking_vec.At(i);
     KALDI_ASSERT(data != nullptr);
     KALDI_ASSERT((*data)(0) == static_cast<BaseFloat>(i));
   }
--- a/src/feat/online-feature-test.cc
+++ b/src/feat/online-feature-test.cc
@@ -375,6 +375,43 @@ void TestOnlineAppendFeature() {
   }
 }
 
+void TestRecyclingVector() {
+  RecyclingVector full_vec;
+  RecyclingVector shrinking_vec(10);
+  for (int i = 0; i != 100; ++i) {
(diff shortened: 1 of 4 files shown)


danpovey:
@vimalmanohar do you have time to review this PR?

src/feat/online-feature.h
```diff
@@ -41,6 +41,32 @@ namespace kaldi {
...
+/// "remembered" so that regardless of the MAX_ITEMS setting, the user always
+/// provides the indices as if no deletion was being performed.
+/// This is useful when processing very long recordings which would otherwise
+/// cause the memory to eventually blow up when the features are not being removed.
+class RecyclingVector {
+public:
+  /// By default it does not remove any elements.
+  RecyclingVector(int items_to_hold = -1);
+
+  Vector<BaseFloat> *Retrieve(int index) const;
```
kkm000: Can this return a `const Vector<BaseFloat> &`? The object owns the vectors outright. This way they could not be `delete`d by mistake.

Also, I would maybe call the methods `At` and `PushBack`, so they hark back to `std::vector`.

src/feat/online-feature.h
```diff
@@ -41,6 +41,32 @@ namespace kaldi {
...
+
+  void Store(Vector<BaseFloat> *item);
```
kkm000: Please document it very clearly that the `item` becomes owned by this collection and must not be deleted.

src/feat/online-feature.cc
```diff
@@ -24,18 +24,55 @@
...
+              << "size = " << Size() << ")";
+  }
+  // 'at' does size checking.
+  return items_.at(index - first_available_index_);
+}
+
+void RecyclingVector::Store(Vector<BaseFloat> *item) {
+  // Internally, we keep at most 2x requested items_to_hold_ and clean up
+  // when this number is reached
+  const int size_before_erase = items_.size();
```
kkm000: I'm assuming the concern here is the efficiency, as to avoid droping the first element and moving the entire rig towards the head of the vector on every (overflowing) insertion. If so, consider `std::deque` as the internal container instead of the `std::vector`. Presumably, a library-implemented container maintains a good balance between overhead and efficiency of `at()` access. 2x the requested size may not always be a good heuristic.

src/feat/online-feature.cc
```diff
@@ -24,18 +24,55 @@
 
 namespace kaldi {
 
+RecyclingVector::RecyclingVector(int items_to_hold) : items_to_hold_(items_to_hold),
+                                                  first_available_index_(0) { }
```
kkm000: Maybe add a `KALDI_ASSERT(items_to_hold > 0);`. The behavior would be fascinating if this is zero. : )

src/feat/online-feature-test.cc
```diff
@@ -375,6 +375,43 @@ void TestOnlineAppendFeature() {
...
 
+void TestRecyclingVector() {
+  RecyclingVector full_vec;
+  RecyclingVector shrinking_vec(10);
+  for (int i = 0; i != 100; ++i) {
+    Vector <BaseFloat> data(1);
+    data.Set(i);
+    full_vec.Store(new Vector<BaseFloat>(data));
+    shrinking_vec.Store(new Vector<BaseFloat>(data));
+  }
```
kkm000: Please add KALDI_ASSERTs for `Size()` at this point. The size is computed non-trivially in the end.

src/feat/online-feature.h
pzelasko: +1 for `At` and `PushBack`. 

If we want to return `const &` here I'd go full way and store the `Vector<BaseFloat>`s by value - but last time I checked move operations were still not supported in Kaldi, and I didn't want to make this PR bigger than necessary. 

I don't want to go half-way because what prevents somebody from `PushBack`ing a null pointer, or deleting the object after he pushes it in here? Would gladly use `unique_ptr`, but didn't see it used in the codebase.

WDYT?

src/feat/online-feature.cc
pzelasko: +1 - eagle eyes ;)

src/feat/online-feature-test.cc
pzelasko: +1

src/feat/online-feature.cc
pzelasko: I've actually mapped 0 to -1, hopefully making it more user-friendly than program termination.

src/feat/online-feature.cc
pzelasko: I considered this before but wasn't sure. Since you're the second person to suggest it and it results in less code, I guess it's better though.

src/feat/online-feature.h
pzelasko: +1

danpovey:
Thanks!

src/feat/online-feature.h
kkm000: Yes, I agree, it should better be symmetric.

kkm000:
LGTM

PR #3058, opened 2019-02-27
[src,egs] Add support for two-pass agglomerative clustering.

dogancan:
@david-ryan-snyder @mmaciej2 

I was having trouble clustering some long recordings due to memory blowup so I added support for two pass clustering. Let me know what you guys think. 

Note that this PR also includes the change from #2971. I can remove that change from this PR if you guys don't like it.

### DIFF
--- a/egs/callhome_diarization/v1/diarization/cluster.sh
+++ b/egs/callhome_diarization/v1/diarization/cluster.sh
@@ -14,6 +14,8 @@ stage=0
 nj=10
 cleanup=true
 threshold=0.5
+max_spk_fraction=1.0
+first_pass_max_utterances=32767
 rttm_channel=0
 read_costs=false
 reco2num_spk=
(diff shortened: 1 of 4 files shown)


dogancan:
@danpovey @david-ryan-snyder @mmaciej2

Just pinging to see if there is any interest or comments.

danpovey:
Yeah, I was hoping  David or Matt would look at it, since I'm pretty busy.

dogancan:
Pinging again.

@danpovey Is there anyone else you can think of who can review this PR? I am trying to close open threads before the end of this month, since I don't know when I will have the opportunity again.

danpovey:
I'll make sure someone does it within the next few days.

david-ryan-snyder:
Sorry for the delays @dogancan.

I don't have any problem with this PR in principle. Could you confirm that this doesn't change the results for short recordings, and that it doesn't require updates to the way these scripts are called in existing recipes (e.g., callhome_diarization)?

dogancan:
This PR does not require updates to the way diarization scripts are called and does not change the results unless num-utterances > first-pass-max-utterances. 

By default first-pass-max-utterances = 32767, so with default settings you would need a ~10 hour recording for two pass clustering to kick in. Note that 32767 is not the recommended value, it is the maximum number of points single pass clustering implementation can handle correctly given enough memory (~100GB). The results of single and two pass clustering algorithms are different but they seem to be equally good for first-pass-max-utterances > ~1000. The time and space requirements of the two algorithms are dramatically different though.

david-ryan-snyder:
@danpovey, I think it's OK to merge this PR as is.

PR #3059, opened 2019-02-27
RNNLM scripts support for UTF-8 encoding

saikiranvalluri:


danpovey:
Is this PR in response to a particular problem that you noticed?
The way it is done right now may not be pretty, but it does actually
support utf-8 and other encodings, as well as other ASCII-compatible
non-utf-8 encodings.

I will still consider the PR though.

saikiranvalluri:
Respected sir,
I actually ran into errors related to : '*ASCII can't decode some char at
position ....*' while using the RNNLM scripts for the Spanish Gigaword text
corpus, which was extracted from Spanish broadcast news text. So, upon
experience, I solved this issue using the encodings changed to utf-8 in all
the related scripts.

Thanks,
Saikiran.

scripts/rnnlm/get_best_model.py
```diff
@@ -35,15 +36,15 @@
...
+best_iter=-1
+for i in range(num_iters):
     this_logfile = "{0}/log/compute_prob.{1}.log".format(args.rnnlm_dir, i)
     try:
-        f = open(this_logfile, 'r', encoding='latin-1')
+        f = open(this_logfile, 'r', encoding='utf-8')
     except:
         sys.exit(sys.argv[0] + ": could not open log-file {0}".format(this_logfile))
-    this_objf = -1000
+    this_objf=-1000
```
danpovey: I think you should revert this line change

scripts/rnnlm/get_best_model.py
```diff
@@ -52,10 +53,6 @@
             except Exception as e:
                 sys.exit(sys.argv[0] + ": line in file {0} could not be parsed: {1}, error is: {2}".format(
                     this_logfile, line, str(e)))
-    # verify this iteration still has model files present
```
danpovey: not sure why this was deleted?

scripts/rnnlm/lmrescore_pruned.sh
```diff
@@ -16,18 +16,16 @@ max_ngram_order=4 # Approximate the lattice-rescoring by limiting the max-ngram-
                   # the same ngram history and this prevents the lattice from 
                   # exploding exponentially. Details of the n-gram approximation
                   # method are described in section 2.3 of the paper
-                  # http://www.danielpovey.com/files/2018_icassp_lattice_pruning.pdf
-max_arcs=         # limit the max arcs in lattice while rescoring. E.g., 20000
+                  # http://www.danielpovey.com/files/2018_icassp_lattice_pruning.pdm
+max_arcs=499         # limit the max arcs in lattice while rescoring. E.g., 20000
```
danpovey: please revert these configuration changes!  Anyway you probably should have  just added the options when you invoked it.

scripts/rnnlm/lmrescore_pruned.sh
```diff
@@ -105,7 +97,6 @@ cp $indir/num_jobs $outdir
 
 $cmd JOB=1:$nj $outdir/log/rescorelm.JOB.log \
   lattice-lmrescore-kaldi-rnnlm-pruned --lm-scale=$weight $special_symbol_opts \
-    --lattice-compose-beam=$lattice_prune_beam \
```
danpovey: and revert this too, or explain why it's needed.    why did you remove the check above for word-list mismatch.  can you please go through this PR and make sure there are no more extraneous changes?

saikiranvalluri:
Sorry Dan,
my intention was to just replace 'latin-1' encodings with 'utf-8', I think
other changes crept in due to me taking the scripts from older version
kaldi. Please let me commit new version scripts into this PR.

Thanks,
Saikiran.

saikiranvalluri:
Hi Dan,
I just closed this PR for now, because I notice that even 'latin-1'
encoding scripts, from master branch, are working fine for Spanish text
based RNNLM training.
I will definitely let you know, if I face any problem with the
latest version Kaldi RNNLM scripts, in the future.
Sorry for consuming your time.

Thanks,
Saikiran.

saikiranvalluri:
But I see in website -
https://stackoverflow.com/questions/2708958/differences-between-utf8-and-latin1.
that utf8 is more worldly prepared than latin-1, when we see Japanese,
russian, ... languages.
So it might be safer with utf-8.

PR #3061, opened 2019-02-28
[WIP] Lattice-faster-decoder-combine

LvHang:
Combine ProcessEmitting() and ProcessNonemitting() into ProcessForFrame(). According to the modification, update GetRawLattice(). 
Put it here so that @chenzhehuai can review it easily.
Thanks.


src/bin/latgen-faster-mapped-combine.cc
```diff
@@ -0,0 +1,179 @@
+// bin/latgen-faster-mapped.cc
+
+// Copyright 2009-2012  Microsoft Corporation, Karel Vesely
+//                2013  Johns Hopkins University (author: Daniel Povey)
```
danpovey: This is OK for testing but eventually this should just be a change to lattice-faster-decoder.cc.  It's just a small optimization.

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,541 @@
...
+  // forward+backward cost is more than lat_beam away from the best path.  It's
+  // possible to prove that this is "correct" in the sense that we won't lose
+  // anything outside of lat_beam, regardless of what happens in the future.
+  // delta controls when it considers a cost to have changed enough to continue
+  // going backward and propagating the change.  larger delta -> will recurse
+  // less far.
+  void PruneActiveTokens(BaseFloat delta);
+
+  /// Processes nonemitting (epsilon) arcs and emitting arcs for one frame
+  /// together. Consider it as a combination of ProcessEmitting() and
```
danpovey: This comment is not good, because it refers to functions that do not exist any more.  Comments should explain the current design, not the difference from a previous design.

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,541 @@
...
+  /// plus one, where the frame-index is zero-based, as used in decodable object.
+  /// That is, the emitting probs of frame t are accounted for in tokens at
+  /// toks_[t+1].  The zeroth frame is for nonemitting transition at the start of
+  /// the graph.
+  StateIdToTokenMap cur_toks_;
+  StateIdToTokenMap next_toks_;
+
+  /// When we call GetRawLattice() in the middle of an utterance, we have to
+  /// process non-emitting arcs so that we need to recover it original status.
+  std::unordered_map<Token*, BaseFloat> recover_map_;  // Token pointer to tot_cost
```
danpovey: I don't understand your comments about this at all.
I also don't see where you clear this map.  That looks like a bug to me; remember, memory can be reused.

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,541 @@
...
+  bool recover_;
+  /// Indicate each frame is processed wholly or not. The size equals to
```
danpovey: what does this mean, "wholly"?

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,541 @@
...
+  /// That is, the emitting probs of frame t are accounted for in tokens at
+  /// toks_[t+1].  The zeroth frame is for nonemitting transition at the start of
+  /// the graph.
+  StateIdToTokenMap cur_toks_;
+  StateIdToTokenMap next_toks_;
+
+  /// When we call GetRawLattice() in the middle of an utterance, we have to
+  /// process non-emitting arcs so that we need to recover it original status.
+  std::unordered_map<Token*, BaseFloat> recover_map_;  // Token pointer to tot_cost
+  bool recover_;
```
danpovey: what is the recover_ variable for?

danpovey:
Zhehuai:
You are right that the cutoff may not be optimal, but this is an issue for
the emitting arcs as well, and I don't think it will affect nonemitting
arcs any worse than it already does for emitting ones.

I don't think the need to reprocess the last frame is a big problem, it's
only one frame.  The speed advantage of this method is that we don't need
to iterate over all the states twice, to process emitting and nonemitting
arcs separately.

However, I have a lot of concerns about (a) the design of this, with
recover_, and recover_map_, and (b) the quality of the comments.  It
doesn't look even close to something that I would merge, at this point, in
terms of clarity and documentation; in fact, it might be faster for me to
just take it, figure out what is going on, and rewrite it myself.  I am
hoping that Zhehuai can work with Hang on this.  I might in principle be OK
with the core design, but first I would need to be able to understand it,
and the comments do not help with that.  Maybe Zhehuai can understand it
and write the comments himself?

LvHang:
Dan:

In my design, the recover_ and recover_map_ will be used when we call GetRawLattice() in the middle of an utterance decoding. 
After processing the frame t, we can achive the complete token list of frame t and the emittion part of frame t+1. So we will call ProcessNonemitting() once in GetRawLattice() to generate the complete token list for frame t+1.  As GetRawLattice() will use "vector<TokenList> active_toks_" to build the raw lattice, so I update "active_toks_" directly in ProcessNonemitting().
But in PrcessForFrame(), the start point should be an emittion part token list. I have to get rid of the non-emittion tokens from "active_toks_" and recover their tot_costs(alpha) so that the behavior is consistent with normal case.

LvHang:
Dan, Zhehuai:

I update the functions and comments.

Now ProcessNonemitting() will take an argument "std::unordered_map<Token*, BaseFloat> recover_map". In normal case, it will be NULL. But when we call GetRawLattice() during decoding, we will pass a local variable "recover_map" of GetRawLattice() and then recover the "active_toks_" in the end of GetRawLattice(). So it will not introduce the linkages.

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1092 @@
...
+  // order to keep everything in a nice dynamic range. Reduce roundoff errors.
+  BaseFloat cost_offset = 0.0;
+
+  // First process the best token to get a hopefully
+  // reasonably tight bound on the next cutoff.  The only
+  // products of the next block are "next_cutoff" and "cost_offset".
+  // Notice: As the difference between the combine version and the traditional
+  // version, this "best_tok" is choosen from emittion tokens. Normally, the
+  // best token of one frame comes from an epsilon non-emittion. So the best
+  // token is a looser boundary. Use it to estimate a bound on the next cutoff.
```
chenzhehuai: So the best token for now is a looser boundary. We use it to estimate a bound on the next cutoff and we will update the cutoff once we have better tokens

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1092 @@
...
+template <typename FST, typename Token>
+void LatticeFasterDecoderCombineTpl<FST, Token>::ProcessForFrame(
+    DecodableInterface *decodable) {
+  KALDI_ASSERT(active_toks_.size() > 0);
+  int32 frame = active_toks_.size() - 1; // frame is the frame-index
+                                         // (zero-based) used to get likelihoods
+                                         // from the decodable object.
+  active_toks_.resize(active_toks_.size() + 1);
+
+  cur_toks_.clear();
```
chenzhehuai: it may be better to understand:

cur_toks_.swap(next_toks_);
next_toks_.clear();

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1092 @@
...
+        // no change flag is needed
+        Token *next_tok = FindOrAddToken(arc.nextstate, frame + 1, tot_cost,
+                                         tok, &next_toks_, NULL);
+        // Add ForwardLink from tok to next_tok. Put it on the head of tok->link
+        // list
+        tok->links = new ForwardLinkT(next_tok, arc.ilabel, arc.olabel,
+                                      graph_cost, ac_cost, tok->links);
+      }
+    }  // for all arcs
+    tok->in_current_queue = false;  // out of queue
```
chenzhehuai: let's do this in line 838 in case there is a state with self-loop

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1092 @@
...
+  }  // end of while loop
+  KALDI_VLOG(6) << "toks after: " << cur_toks_.size();
```
chenzhehuai: KALDI_VLOG(6) << "Number of tokens active on frame " << NumFramesDecoded()-1
                  << " is " << cur_toks_.size();

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1092 @@
...
+          *best_elem = iter->second;
+        }
+      }
+    }
+
+    BaseFloat beam_cutoff = best_weight + config_.beam,
+        min_active_cutoff = std::numeric_limits<BaseFloat>::infinity(),
+        max_active_cutoff = std::numeric_limits<BaseFloat>::infinity();
+
+    KALDI_VLOG(6) << "Number of tokens active on frame " << NumFramesDecoded()
```
chenzhehuai: KALDI_VLOG(6) << "Number of tokens emitting on frame " << NumFramesDecoded()-1

danpovey:
Regarding max-active: see how much difference it makes to the WER/RTF curve
if you just leave a single option max-active and redefine it as the total
number of active states.  If it's not a huge incompatibility I'd rather
keep it simple; there will be other slight differences in the pruning
behavior as well, that may somewhat cancel out this effect.

src/decoder/lattice-faster-decoder-combine.cc
LvHang: It's ok. But, actually, the ilabel of self-loop couldn't be epsilon, so I think it will not cause any problem. I have changed.

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,547 @@
...
+                   bool use_final_probs = true);
+
+  /// Outputs an FST corresponding to the raw, state-level
+  /// tracebacks.  Returns true if result is nonempty.
+  /// If "use_final_probs" is true AND we reached the final-state
+  /// of the graph then it will include those as final-probs, else
+  /// it will treat all final-probs as one.
+  /// The raw lattice will be topologically sorted.
+  /// The function can be called during decoding, it will take "next_toks_" map
+  /// and generate the complete token list for the last frame. Then recover it
```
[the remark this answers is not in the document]
LvHang: I guess you mean:
it will process emitting tokens from "next_toks_" map to generate both emitting and non-emitting tokens to get raw lattices.

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,547 @@
...
+  // possible to prove that this is "correct" in the sense that we won't lose
+  // anything outside of lat_beam, regardless of what happens in the future.
+  // delta controls when it considers a cost to have changed enough to continue
+  // going backward and propagating the change.  larger delta -> will recurse
+  // less far.
+  void PruneActiveTokens(BaseFloat delta);
+
+  /// Processes non-emitting (epsilon) arcs and emitting arcs for one frame
+  /// together. It takes the emittion tokens in "cur_toks_" from last frame.
+  /// Generates non-emitting tokens for current frame and emitting tokens for
```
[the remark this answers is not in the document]
LvHang: You're right. I noticed it before. The concept of "cur_toks_" in this version is different from the one in lattice-simple-decoder. This is a problem about viewpoint. In fact, I want to emphasize the start point and advance.
Anyway, I will change it. Thanks!

src/decoder/lattice-faster-decoder-combine.h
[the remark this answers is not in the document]
chenzhehuai: I think in this function, you process non-emit, after that, you get both emit (you have done before) and non-emit, right?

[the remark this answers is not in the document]
LvHang: No. The start point is a list of the emitting tokens. And then I generate non-emitting and emitting tokens.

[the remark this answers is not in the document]
LvHang: No. In GetRawLattice(), we will call ProcessNonemitting() once. The start point is a list of emitting tokens (i.e. "cur_toks_"), and generate non-emitting tokens for "cur_toks_".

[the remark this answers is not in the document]
chenzhehuai: yes, that's why I wanna make it more explicit and saying: 
it will process non-emitting arcs from "next_toks_" map to get both emission and non-emission tokens for getting raw lattices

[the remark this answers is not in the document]
LvHang: oh, you said "tokens" previously, so I missed it.

LvHang:
According to the suggestions of Dan and Zhehuai, update the comments and code.

danpovey:
thanks.  Zhehuai, please check carefully.
I also will wait to hear on the efficiency.
Ideally it would be nice to be able to separate the effects of stopping
using HashList, versus combining nonemitting/emitting propagation, but
perhaps that is too much to hope for.

LvHang:
I uploaded a simple testing script. It calls "nnet3-compute" to generate posterior, and then use "latgen-faster-mapped" or "latgen-faster-mapped-combine" to decode.
Anyway, I'm testing it.

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1093 @@
...
+  }
+
+  // Store the offset on the acoustic likelihoods that we're applying.
+  // Could just do cost_offsets_.push_back(cost_offset), but we
+  // do it this way as it's more robust to future code changes.
+  cost_offsets_.resize(frame + 1, 0.0);
+  cost_offsets_[frame] = cost_offset;
+
+  // Build a queue which contains the emittion tokens from previous frame.
+  std::vector<StateId> cur_queue;
```
danpovey: making cur_queue a class member would make this code faster, although it's a little ugly, because it would avoid reallocation.

src/decoder/lattice-faster-decoder-combine.cc
```diff
@@ -0,0 +1,1093 @@
...
+  for (IterType iter = prev_toks_.begin(); iter != prev_toks_.end(); iter++) {
+    cur_queue.push_back(iter->first);
+    iter->second->in_current_queue = true;
+  }
+
+  // Iterator the "cur_queue" to process non-emittion and emittion arcs in fst.
+  while (!cur_queue.empty()) {
+    StateId state = cur_queue.back();
```
danpovey: Processing the queue from the same end you push to is probably not optimal as it would cause you to follow chains of epsilons depth-first, which may require them to be searched again if the costs change later on.  If you use a std::dequeue and use both ends, it would likely be more optimal.

src/decoder/lattice-faster-decoder-combine.h
```diff
@@ -0,0 +1,547 @@
...
+  /// That is, the emitting probs of frame t are accounted for in tokens at
+  /// toks_[t+1].  The zeroth frame is for nonemitting transition at the start of
+  /// the graph.
+  StateIdToTokenMap cur_toks_;
+  StateIdToTokenMap next_toks_;
+
+  /// Gets the weight cutoff.
+  /// Notice: In traiditional version, the histogram prunning method is applied
+  /// on a complete token list on one frame. But, in this version, it is used
+  /// on a token list which only contains the emittion part. So the max_active
```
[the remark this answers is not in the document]
danpovey: Don't make this change just yet.  I don't want to cause confusion in the future by having two very similar ways to set the number of active states.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1153 @@
...
+  bucket_storage_begin_ = std::floor((best_cost_estimate - 15) * cost_scale);
+  first_occupied_bucket_index_ = bucket_storage_begin_ + bucket_size;
+}
+
+template<typename Token>
+void BucketQueue<Token>::Push(Token *tok) {
+  int32 bucket_index = std::floor(tok->tot_cost * cost_scale_);
+  int32 vec_index = bucket_index - bucket_storage_begin_;
+
+  if (vec_index < 0) {
```
danpovey: You can combine these two checks into a single instruction by making vec_index a size_t and making the comparison just:
```
  if (vec_index >= buckets.size()) {
 }
```
Obviously be careful then, because vec_index would be unsigned, you would have to cast it to a signed value to figure out how you needed to reallocate.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1153 @@
...
+    std::vector<std::vector<Token*> > tmp(buckets_);
+    buckets_.resize(tmp.size() + increase_size);
+    std::copy(tmp.begin(), tmp.end(), buckets_.begin() + increase_size);
+    // Update start point
+    bucket_storage_begin_ = bucket_index;
+    vec_index = 0;
+  } else if (vec_index > buckets_.size() - 1) {
+    KALDI_WARN << "Have to reallocate the BucketQueue. Maybe need to reserve"
+               << " more elements in constructor. Push back.";
+    buckets_.resize(vec_index + 1);
```
danpovey: resizing one by one is very wasteful; resize to vec_index + 10

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1153 @@
...
+  int32 bucket_index = std::floor(tok->tot_cost * cost_scale_);
+  int32 vec_index = bucket_index - bucket_storage_begin_;
+
+  if (vec_index < 0) {
+    KALDI_WARN << "Have to reallocate the BucketQueue. Maybe need to reserve"
+               << " more elements in constructor. Push front.";
+    int32 increase_size = - vec_index;
+    std::vector<std::vector<Token*> > tmp(buckets_);
+    buckets_.resize(tmp.size() + increase_size);
+    std::copy(tmp.begin(), tmp.end(), buckets_.begin() + increase_size);
```
danpovey: You need a margin here, there is no point increasing one by one, have `increase_size = - vec_index + 10`.
What you should be doing here is
```
    std::vector<std::vector<Token*> > tmp(buckets_.size() + increase_size);
    for (int32 i ... );
           tmp[i].swap(buckets_[i + offset]);
   // for some offset you would figure out.
```

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1153 @@
...
+  while(vec_index < buckets_.size()) {
+    // Remove the best token
+    best_tok = buckets_[vec_index].back();
+    buckets_[vec_index].pop_back();
+
+    if (buckets_[vec_index].empty()) {  // This bucket is empty. Update
+                                        // first_occupied_bucket_index_
+      int32 next_vec_index = vec_index + 1;
+      for(; next_vec_index < buckets_.size(); next_vec_index++) {
+        if(!buckets_[next_vec_index].empty()) break;
```
danpovey: space after for(, if( and while(

danpovey: @hang, please address this comment about whitespace

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1159 @@
...
+      }
+      first_occupied_bucket_index_ = bucket_storage_begin_ + next_vec_index;
+      vec_index = next_vec_index;
+    }
+
+    if (best_tok->in_queue) {  // This is a effective token
+      best_tok->in_queue = false;
```
danpovey: I would make 'best_tok' just a variable inside the loop called 'took', and at this point, instead of doing `break;`, just return.  And no 'else' statement.  Then below, instead of 'return best_tok;', just do 'return NULL; '

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1159 @@
...
+void BucketQueue<Token>::Push(Token *tok) {
+  int32 bucket_index = std::floor(tok->tot_cost * cost_scale_);
+  size_t vec_index = static_cast<size_t>(bucket_index - bucket_storage_begin_);
+
+  if (vec_index >= buckets_.size()) {
+    KALDI_WARN << "Have to reallocate the BucketQueue. Maybe need to reserve"
+               << " more elements in constructor.";
+    int32 offset = static_cast<int32>(vec_index);
+    // a margin here (e.g. 10);
+    int32 increase_size = offset >= 0 ? offset + 1 - buckets_.size() + 10 :
```
danpovey: the code around here is very ugly, see if you can make it clearer.  This point in the code is reached rarely, efficiency is not an issue.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1159 @@
...
+                                        - offset + 10;
+    buckets_.resize(buckets_.size() + increase_size);
+    
+    // Push front
+    if (offset < 0) {
+      std::vector<std::vector<Token*> > tmp(buckets_);
```
danpovey: This code here is not right at all.  Perhaps you misunderstood what I told you.  Please work with @chenzhehuai  on this.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1159 @@
...
+      }
+      // Update start point
+      bucket_storage_begin_ = bucket_index - 10;
+      vec_index = 10;
+    }
+  }
+
+  tok->in_queue = true;
+  buckets_[vec_index].push_back(tok);
+  if (vec_index < (first_occupied_bucket_index_ - bucket_storage_begin_))
```
danpovey: (oops, I thought this was on my own branch, did not realize it was on Kaldi-asr or I would have done it via review).  It looks like the expression `first_occupied_bucket_index_ - bucket_storage_begin_` is evaluated a lot. That probably means that it would be more efficient to have a variable first_occupied_vec_index_ instead.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1133 @@
...
+      int32 next_vec_index = vec_index + 1;
+      for (; next_vec_index < buckets_.size(); next_vec_index++) {
+        if (!buckets_[next_vec_index].empty()) break;
+      }
+      vec_index = next_vec_index;
+    }
+
+    if (tok->in_queue) {  // This is a effective token
+      tok->in_queue = false;
+      first_occupied_vec_index_ = vec_index;
```
danpovey: This statement doesn't have to be here as it doesn't change unless you go inside the `if (buckets_[vec_index].empty()) {` block.  Put it inside that block, after ` vec_index = next_vec_index;`.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1133 @@
...
+        }
+
+        // no change flag is needed
+        Token *next_tok = FindOrAddToken(arc.nextstate, frame + 1, tot_cost,
+                                         tok, &cur_toks_, NULL);
+        // Add ForwardLink from tok to next_tok. Put it on the head of tok->link
+        // list
+        tok->links = new ForwardLinkT(next_tok, arc.ilabel, arc.olabel,
+                                      graph_cost, ac_cost, tok->links);
+        if (best_token_in_next_frame_ == NULL ||
```
danpovey: This code can be optimized by combining with the code around line 873.   Move the
```if (tot_cost + adaptive_beam < next_cutoff) {
          next_cutoff = tot_cost + adaptive_beam;  // a tighter boundary for emitting
        }
```
to here,  and inside the if-statement, do:
```
    best_token_in_next_frame_ = next_tok
```
Then you don't need this extra if-statement.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1133 @@
...
+template<typename Token>
+void BucketQueue<Token>::Clear() {
+  for (size_t i = 0; i < buckets_.size(); i++) {
+    buckets_[i].clear();
+  }
+  first_occupied_vec_index_ = buckets_.size();
+}
+
+template<typename Token>
+void BucketQueue<Token>::SetBegin(BaseFloat best_cost_estimate) {
```
danpovey: Actually we don't need this SetBegin() thing or the best_cost_estimate.  Just initialize bucket_storage_begin_ to 0 and get rid of best_cost_estimate (treat as zero).  You can rely on the code in Push() to move bucket_storage_begin_ to be as negative as needed.   The costs are already normalized due to cost_offset_ so they will stay in the same range for the entire file.  Because you don't need this, you also don't need best_token_on_next_frame.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1133 @@
...
+  prev_toks_.swap(cur_toks_);
+  cur_toks_.clear();
+  if (prev_toks_.empty()) {
+    if (!warned_) {
+      KALDI_WARN << "Error, no surviving tokens on frame " << frame;
+      warned_ = true;
+    }
+  }
+
+  KALDI_ASSERT(best_token_in_next_frame_);
```
danpovey: You don't need best_token_in_next_frame_; get rid of all instances of it.  Instead you can work out the best cost on the current frame from the value `next_cutoff` had when you were doing ProcessForFrame() last time.
You can set up the code so that each time it processes a frame, it computes the cost_offset for the *next* frame, e.g., later on in this function (before updating adapative_beam), do
  ```
  BaseFloat best_cost_on_next_frame = next_cutoff - adaptive_beam;
  cost_offsets_.resize(frame + 2);
  cost_offsets_[frame + 1] = cost_offset;
```
Obviously you will have to initialize cost_offsets_ to size 1 with first element zero, in the function where you initialize everything.

File: src/decoder/lattice-faster-decoder-combine-bucketqueue.cc (hunk 26 above)
Comment (LvHang) [reply src/decoder/lattice-faster-decoder-combine-bucketqueue.cc]: Dan, I set bucket_storage_begin_ to (-15 * scale_cost_) as constructor in Clear and get rid of SetBegin. I think it's better to set it to 0, since the negative cost will be encountered. If we just set it to 0, the "push front re-allocate" operation will occur frequently.  It will affact the speed I think.

Comment (danpovey) [reply src/decoder/lattice-faster-decoder-combine-bucketqueue.cc]: You mean better than setting it to 0?  I guess my rationale was that any reallocation would only occur once or twice per utterance at most.   Leaving it at -15 is OK.

Comment (LvHang) [reply src/decoder/lattice-faster-decoder-combine-bucketqueue.cc]: Ok, I will set it to "-15 * cost_scale_".
Actually, SetBegin will be called after Clear to set bucket_storage_begin_ for each frame. If we set bucket_storage_begin_ to 0, it means we have to reallocate when we encounter any negative cost. I did a simple test, about half of the frames in each utterance will generate negative cost.

File: src/decoder/lattice-faster-decoder-combine-bucketqueue.cc [hunk 28]
```diff
@@ -0,0 +1,1123 @@
...
+      first_occupied_vec_index_ = vec_index;
+    }
+
+    if (tok->in_queue) {  // This is a effective token
+      tok->in_queue = false;
+      return tok;
+    }
+  }
+  return NULL;
+}
```
danpovey: I just realized something about Pop(), which is a way to simplify it and possibly make it slightly more efficient.
the idea is to change the semantics of first_occupied_vec_index_, and rename it.  (obviously, document it).  E.g.
```
  /// first_nonempty_bucket_index_ is an integer in the range [0, buckets_.size() - 1]
  /// which is not larger than the index of the first nonempty element of buckets_.
  int32 first_nonempty_bucket_index_;
```
For clarity, I would like to get rid of the distinction between vec_index and bucket_index, which
makes things unclear.  I'd like to have a formula where:
```
  size_t bucket_index = std::floor(tok->tot_cost * cost_scale_) + bucket_offset_;
```
so you'd rename bucket_storage_begin_ to bucket_offset_ and negate it.


Then Pop() would be simplified to:
 ``` 
 Token *Pop() {
   if (!buckets_[first_nonempty_bucket_index_].empty()) {
       Token *ans = buckets_[first_nonempty_bucket_index_].back();
        buckets_[first_nonempty_bucket_index_].pop_back();
        return ans;
    }
    // and the rest of the code goes here.
 }
```

File: src/decoder/lattice-faster-decoder-combine-bucketqueue.cc [hunk 29]
```diff
@@ -0,0 +1,1123 @@
...
+
+template<typename Token>
+void BucketQueue<Token>::Clear() {
+  for (size_t i = 0; i < buckets_.size(); i++) {
```
danpovey: You can start this loop from first_occupied_vec_index_.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1123 @@
...
+    buckets_[i].clear();
+  }
+  first_occupied_vec_index_ = buckets_.size();
```
danpovey: obviously if you make the changes to first_occupied_vec_index_ I mentioned above, this line would change, as it has to be < buckets_.size().

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1123 @@
...
+  prev_toks_.reserve(1000);
+  cur_toks_.reserve(1000);
+}
+
+
+template <typename FST, typename Token>
+LatticeFasterDecoderCombineTpl<FST, Token>::~LatticeFasterDecoderCombineTpl() {
+  ClearActiveTokens();
+  if (delete_fst_) delete fst_;
+  //prev_toks_.clear();
```
danpovey: remove these comments.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1123 @@
...
+  BaseFloat next_cutoff = std::numeric_limits<BaseFloat>::infinity();
+  // "cost_offset" contains the acoustic log-likelihoods on current frame in 
+  // order to keep everything in a nice dynamic range. Reduce roundoff errors.
+  BaseFloat cost_offset = cost_offsets_[frame];
+
+  // Iterator the "cur_queue_" to process non-emittion and emittion arcs in fst.
+  Token *tok = NULL;
+  int32 num_toks_processed = 0;
+  int32 max_active = config_.max_active;
+  for (; num_toks_processed < max_active && (tok = cur_queue_.Pop()) != NULL;
```
danpovey: I told you to put this && statement the other way round for a reason:
```
(tok = cur_queue_.Pop()) != NULL && num_toks_processed < max_active
```
It's faster that way.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1123 @@
...
+       num_toks_processed++) {
+    BaseFloat cur_cost = tok->tot_cost;
+    StateId state = tok->state_id;
+    if (cur_cost > cur_cutoff &&
+        num_toks_processed < config_.min_active) { // Don't bother processing
```
danpovey: This `<` RE min_active should be a `>`.  Think about it.  This will affect the speed as well as correctness.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
danpovey: sorry, cancel this comment; both conditions will normally be true so the order doesn't matter.

src/decoder/lattice-faster-decoder-combine-bucketqueue.h
```diff
@@ -0,0 +1,642 @@
...
+template<typename Token>
+class BucketQueue {
+ public:
+  /** Constructor. 'cost_scale' is a scale that we multiply the token costs by
+   *  before intergerizing; a larger value means more buckets.
+   *  'best_cost_estimate' is an estimate of the best (lowest) cost  that
+   *  we are likely to encounter (e.g. the best cost that we have seen so far).
+   *  It is used to initialize 'bucket_storage_begin_'.
+   */
+  BucketQueue(BaseFloat best_cost_estimate, BaseFloat cost_scale = 1.0);
```
LvHang: Dan, how are you thinking about remove "best_cost_estimate" from the constructor of BucketQueue?

danpovey:
Just treat it as zero, since we do in fact normalize costs to be not far
from zero on each frame.  It's not necessary; let the Push() code handle it.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
LvHang: Sorry, dan, I didn't get the difference between "first_nonempty_bucket_index_" and "first_occupied_vec_index_"

Dan, with the new design, I think the Pop() function still looks like current code. 
We cannot return the token directly as you wrote
```
if (!buckets_[first_nonempty_bucket_index_].empty()) {
  Token *ans = buckets_[first_nonempty_bucket_index_].back();
  buckets_[first_nonempty_bucket_index_].pop_back();
  return ans;
}
```
Because we cannot guarantee the 'ans' is an effective token.  We have to peek the next token.
```
while (first_nonempty_bucket_index_ < buckets_.size()) {
  Token *tok = buckets_[first_nonempty_bucket_index_].back();
  if (tok->in_queue) break;
  buckets_[first_nonempty_bucket_index_].pop_back();  //remove the invalid token
  if (buckets_[first_nonempty_bucket_index_].empty()) {
    first_nonempty_bucket_index_++;
    for (; first_nonempty_bucket_index_ < buckets_.size(); first_nonempty_bucket_index_++) {
      if(!bucket_[first_nonempty_bucket_index_].empty()) break;
    }
  }
}
```
At the same time, we still need "first_nonempty_bucket_index == buckets_.size()" to determinize the BucketQueue is empty at the beginning of the function.
```
if (first_nonempty_bucket_index == buckets_.size()) {  // no token left
  return NULL:
]
```

danpovey:
>
>
> In src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
> <https://github.com/kaldi-asr/kaldi/pull/3061#discussion_r269375112>:
>
> > +    if (buckets_[vec_index].empty()) { // This bucket is empty. Update vec_index
> +      int32 next_vec_index = vec_index + 1;
> +      for (; next_vec_index < buckets_.size(); next_vec_index++) {
> +        if (!buckets_[next_vec_index].empty()) break;
> +      }
> +      vec_index = next_vec_index;
> +      first_occupied_vec_index_ = vec_index;
> +    }
> +
> +    if (tok->in_queue) {  // This is a effective token
> +      tok->in_queue = false;
> +      return tok;
> +    }
> +  }
> +  return NULL;
> +}
>
> Sorry, dan, I didn't get the difference between
> "first_nonempty_bucket_index_" and "first_occupied_vec_index_"
>

I was referring the terminology in a comment in the code, which you got
from an email of mine.  They are the same, but my point is you don't have a
temporary in the code any more named 'bucket_index'.


> Dan, with the new design, I think the Pop() function still looks like
> current code.
> We cannot return the token directly as you wrote
>
> if (!buckets_[first_nonempty_bucket_index_].empty()) {
>   Token *ans = buckets_[first_nonempty_bucket_index_].back();
>   buckets_[first_nonempty_bucket_index_].pop_back();
>   return ans;
> }
>
> Because we cannot guarantee the 'ans' is an effective token.
>

OK, then check ans ans->in_queue == true before returning it.  But don't
start with a while
loop, because you are checking a condition that is not necessary to check
at that point
(because as I said, you would always ensure that
first_nonempty_bucket_index_ was
less than buckets_size(), precisely to avoid that check).


> while (first_nonempty_bucket_index_ < buckets_.size()) {
>   Token *tok = buckets_[first_nonempty_bucket_index_].back();
>   if (tok->in_queue) break;
>   buckets_[first_nonempty_bucket_index_].pop_back();  //remove the invalid token
>   if (buckets_[first_nonempty_bucket_index_].empty()) {
>     first_nonempty_bucket_index_++;
>     for (; first_nonempty_bucket_index_ < buckets_.size(); first_nonempty_bucket_index_++) {
>       if(!bucket_[first_nonempty_bucket_index_].empty()) break;
>     }
>   }
> }
>
> At the same time, we still need "first_nonempty_bucket_index ==
> buckets_.size()" to determinize the BucketQueue is empty at the beginning
> of the function.
>
> if (first_nonempty_bucket_index == buckets_.size()) {  // no token left
>   return NULL:
> ]
>
> Read more carefully what I said.  I am proposing to always ensure that
first_nonempty_bucket_index< buckets_.size()), and change what that
variable means.

LvHang:
Hi Dan,

I understand that you want to use the new "bucket_index" formula instead of the old "bucket_index" and "vec_index".

Let me explain what makes me confused. Firstly, we have the definition of "first_nonempty_bucket_index_".
/// first_nonempty_bucket_index_ is an integer in the range [0, buckets_.size() -1]
/// which is not larger than the index of the ifrst nonempty element of buckets_.
int32 first_nonempty_bucket_index_;

So the key problem is when and how to update "first_nonempty_bucket_index_".
Assume the Push() function adds the token into the BucketQueue one by one and sets "first_nonempty_bucket_index_" to the first non-empty bucket.
From your previous comments, I guess, after once successful Pop operation (ans->in_queue), you just leave the "first_nonempty_bucket_index_" there. And then, until the bucket is empty (buckets_[first_nonempty_bucket_index_].empty()), you will update it.   
 
(1) But what should we do if the Pop operation is not successed? We need a while loop to repeat the try until the BucketQueue is empty.
(2) How can we know the BucketQueue is empty? I think the most convenient way is "first_nonempty_bucket_index_ == buckets_.size()". But it is conflict with your definition. So I suggest use
the following definition.
```
/// first_nonempty_bucket_index_ is an integer whose valid range is [0, buckets_.size() -1]
/// which is not larger than the index of the ifrst nonempty element of buckets_.
/// If "first_nonempty_bucket_index_" corresponds to a value past the end of buckets_
/// (i.e. first_nonempty_bucket_index_ == buckets_.size(), we interpret it as "there are no 
/// buckets with entries."
int32 first_nonempty_bucket_index_;
```
And the code will be (please see the comments to know what I'm concerned.)
```
  93 template<typename Token>                                                                              
  94 Token* BucketQueue<Token>::Pop() {                                                                    
  95   while (true) {                                                                                      
  96     if (!bucket_[first_nonempty_bucket_index_].empty()) {                                             
  97       Token *ans = buckets_[first_nonempty_bucket_index_].back();                                     
  98       buckets_[first_nonempty_bucket_index_].pop_back();                                              
  99       if (ans->in_queue) {                                                                            
 100         ans->in_queue = false;                                                                        
 101         return ans;                                                                                   
 102       }                                                                                               
 103     }                                                                                                 
 104     // There are two cases can trigger the following code to update the                               
 105     // "first_nonempty_bucket_index_". One is last token is invalid (i.e.                             
 106     // ans->in_queue == false). The other is bucket is empty.                                         
 107     if (buckets_[first_nonempty_bucket_index_].empty()) {                                             
 108       first_nonempty_bucket_index_++;                                                                 
 109       for (; first_nonempty_bucket_index_ < buckets_.size();                                          
 110            first_nonempty_bucket_index_++) {                                                          
 111         if (!buckets_[first_nonempty_bucket_index_].empty()) break;                                   
 112       }                                                                                               
 113     }                                                                                                 
 114     // There are two cases will trigger the following code. One is                                    
 115     // first_nonempty_bucket_index_ == buckets_.size() [ But it is conflict with                      
 116     // the definition]. The other is we find an appropriate new                                       
 117     // "first_nonempty_bucket_index_".                                          
 118     if (first_nonempty_bucket_index_ == buckets_.size()) return NULL;                                 
 119   }                                                                                                   
 120 }                                                         
```

danpovey:
The whole point is to make the normal code path very fast.
Just set first_nonempty_bucket_index_ to buckets_.size() - 1 if it would
naturally be more than buckets_.size().   That way you never have to check
(in the normally taken code path) whether that index is valid or not.

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1085 @@
...
+      buckets_.resize(bucket_index + margin);
+      first_nonempty_bucket_ = &buckets_[first_nonempty_bucket_index_];
+    } else {  // less than 0
+      int32 increase_size = - static_cast<int32>(bucket_index) + margin;
+      buckets_.resize(buckets_.size() + increase_size);
+      // translation
+      for (size_t i = buckets_.size() - 1; i >= increase_size; i--) {
+        buckets_[i].swap(buckets_[i - increase_size]);
+      }
+      bucket_offset_ = bucket_offset_ + increase_size * cost_scale_;
```
danpovey: This line doesn't look right to me, there should be no `* cost_scale_`; bucket_offset_ is a number of buckets.
It might affect decoding results (although it would only impact frames on which the bucket list was resize).

src/decoder/lattice-faster-decoder-combine-bucketqueue.cc
```diff
@@ -0,0 +1,1085 @@
...
+  // do it this way as it's more robust to future code changes.
+  // Set the cost_offset_ for next frame, it equals "- best_cost_on_next_frame".
+  cost_offsets_.resize(frame + 2, 0.0);
+  cost_offsets_[frame + 1] = adaptive_beam - next_cutoff;
+
+  {  // This block updates adaptive_beam_
+    BaseFloat beam_used_this_frame = adaptive_beam;
+    Token *tok = cur_queue_.Pop();
+    if (tok != NULL) {
+      // The queue would only be nonempty if we hit the max-active constraint.
```
danpovey: actuall, change  this comment to 
```
// We hit the max-active contraint, meaning we effectively pruned to a beam
// tighter than `beam`.  Work out what this was, it will be used to update
// `adaptive_beam`.
```

PR #3065, opened 2019-03-02
Rnnlm scripts utf8 support

saikiranvalluri:
But I see in website - https://stackoverflow.com/questions/2708958/differences-between-utf8-and-latin1. that utf8 is more worldly prepared than latin-1, when we see Japanese, Russian, ... languages. So it might be safer with utf-8.


danpovey:
You'll see in the comments that latin-1 is simply a device to allow any ASCII-compatible encoding, which does include UTF-8.  But we have run into more 'gotchas' than I expected-- basically I am realizing that python3 is fundamentally hostile to dealing with string data in this way, so I am considering just accepting that utf-8 is the standard and doing things python3's way.

scripts/rnnlm/show_word_features.py
```diff
@@ -12,7 +12,7 @@
 # encoding as well as other encodings such as gbk, as long as the spaces are
```
danpovey: If you are going to switch to utf-8 you should remove these comments which explained why we were using latin-1.

scripts/rnnlm/get_word_features.py
```diff
@@ -41,7 +41,7 @@
 # return the vocab, which is a dict mapping the word to a integer id.
 def read_vocab(vocab_file):
     vocab = {}
-    with open(vocab_file, 'r', encoding="latin-1") as f:
+    with open(vocab_file, 'r', encoding="utf-8") as f:
         for line in f:
             fields = re.split(tab_or_space, line)
```
danpovey: and this tab_or_space stuff would disappear if using utf-8 encoding, it would just be line.split()

saikiranvalluri:
Respected sir,
I have integrated the above mentioned changes into the scripts and
committed into the PR.

Thanks,
Saikiran.

danpovey:
Thanks.  How have you tested this; have you made sure that it runs for you after the latest changes?  If not, please do, and LMK when done.

saikiranvalluri:
Sure sir,
I am training a Spanish RNNLM using these scripts and I will let you know
the results.

saikiranvalluri:
Yes, I verified that the scripts are working fine after the latest changes.

Thanks,
Saikiran.

saikiranvalluri:
Hi Dan, I verified that the scripts work fine in RNNLM training for Spanish ASR. Could you please consider merging this PR, if feasible?

saikiranvalluri:
Thank you sir,
please feel free to revert back to me, when you face any issues in the
scripts.

PR #3066, opened 2019-03-02
Add device options to enable tensor core math mode.

luitjens:
This commit allows us to enable the tensor cores on Volta/Turing based GPUs.

When you do this 32bit gemms are performed in 16 bit math using tensor cores.  This reduces precision but increases speed significantly.  For inference the loss in precision does not effect results.  By default this option is set to false making this an opt-in.

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -110,6 +110,13 @@ void CuDevice::Initialize() {
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
     CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
+    
+    if(device_options_.use_tensor_cores_) {
+      //Enable tensor cores in CUBLAS
+      //Note if the device does not support tensor cores this will fall back to normal math mode
+      CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_, CUBLAS_TENSOR_OP_MATH));
+    }
+
     // Initialize the cuSPARSE library
     CUSPARSE_SAFE_CALL(cusparseCreate(&cusparse_handle_));
     CUSPARSE_SAFE_CALL(cusparseSetStream(cusparse_handle_, cudaStreamPerThread));
@@ -525,6 +532,8 @@ CuDevice::~CuDevice() {
 // Each thread has its own copy of the CuDevice object.
 // Note: this was declared "static".
 thread_local CuDevice CuDevice::this_thread_device_;
+  
+CuDevice::CuDeviceOptions_t CuDevice::device_options_;
 
 // define and initialize the static members of the CuDevice object.
 int32 CuDevice::device_id_ = -1;
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -111,10 +111,11 @@ void CuDevice::Initialize() {
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
     CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
     
-    if(device_options_.use_tensor_cores_) {
-      //Enable tensor cores in CUBLAS
-      //Note if the device does not support tensor cores this will fall back to normal math mode
-      CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_, CUBLAS_TENSOR_OP_MATH));
+    if (device_options_.use_tensor_cores_) {
+      // Enable tensor cores in CUBLAS
+      // Note if the device does not support tensor cores this will fall back to normal math mode
+      CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_, 
+            CUBLAS_TENSOR_OP_MATH));
     }
 
     // Initialize the cuSPARSE library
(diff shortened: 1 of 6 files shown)


danpovey:
Hm.  Do you think this increases the chance of overflow?

luitjens:
What do you mean by overflow here?  floating point numbers cannot overflow.

danpovey:
I mean if they are large, perhaps due to an instability in training, like
10e+40 or so will overflow in regular floating point (becomes infinity), it
looks like 65504 is the largest representable value in fp16, and the
smallest is 5.96 × 10−8, which might in principle be a problem for
derivative computation during training.  This  is the kind of thing that
we'd have to at least test and see how it affects the training in various
architectures before we could merge it.

luitjens:
That is why this is an opt-in.  Training should not be using this feature.  It is meant for inference.  Note that if you accepted this patch as is it would not impact any existing code because that option is disabled by default.

danpovey:
Can you please explain in the comment how one would opt into this?  
Also, per google style guide, please add a space after the '//' in comments.

luitjens:
Added a comment on usage.  Here is an example:

CuDevice::RegisterDeviceOptions(&po);
CuDevice::Initialize();

luitjens:
In case it isn't clear after you register that in a binary you can enable tensor cores with this command line:

--cuda-use-tensor-cores=true

src/cudamatrix/cu-device.cc
```diff
@@ -110,6 +110,13 @@ void CuDevice::Initialize() {
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
     CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
+    
+    if(device_options_.use_tensor_cores_) {
```
danpovey: space after if and after //.

src/cudamatrix/cu-device.h
```diff
@@ -184,8 +184,28 @@ class CuDevice {
...
+    CuDevice::device_options_.Register(po);  
+  }
   ~CuDevice();
  private:
+
+  struct CuDeviceOptions_t {
+    bool use_tensor_cores_; //Enable tensor cores
+    CuDeviceOptions_t () : use_tensor_cores_(false) {};
+    void Register(OptionsItf *po) {
+      po->Register("cuda-use-tensor-cores",&use_tensor_cores_, "Enable FP16 tensor math. "
```
danpovey: space after comma.
Please add: "Only recommended for inference (decoding)."
I think it would be a good idea to have these options registered in programs that might be used for inference, e.g.
nnet3-compute, if you haven't already done that.

luitjens: fixed

src/cudamatrix/cu-device.cc
luitjens: fixed

src/cudamatrix/cu-device.cc
```diff
@@ -525,6 +532,8 @@ CuDevice::~CuDevice() {
 // Each thread has its own copy of the CuDevice object.
 // Note: this was declared "static".
 thread_local CuDevice CuDevice::this_thread_device_;
+  
+CuDevice::CuDeviceOptions_t CuDevice::device_options_;
```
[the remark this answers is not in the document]
luitjens: fixed

src/cudamatrix/cu-device.h
```diff
@@ -184,8 +184,28 @@ class CuDevice {
...
+  //  CuDevice::RegisterDeviceOptions(&po);
+  //  CuDevice::Initialize();
+  static void RegisterDeviceOptions(OptionsItf *po) {
+    CuDevice::device_options_.Register(po);  
+  }
   ~CuDevice();
  private:
+
+  struct CuDeviceOptions_t {
+    bool use_tensor_cores_; //Enable tensor cores
```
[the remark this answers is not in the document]
luitjens: fixed

luitjens:
I believe all issues are now fixed.

src/cudamatrix/cu-device.h
```diff
@@ -184,8 +184,30 @@ class CuDevice {
...
+  //  CuDevice::RegisterDeviceOptions(&po);
+  //  CuDevice::Initialize();
+  static void RegisterDeviceOptions(OptionsItf *po) {
+    CuDevice::device_options_.Register(po);  
+  }
   ~CuDevice();
  private:
+
+  struct CuDeviceOptions {
+    bool use_tensor_cores_; // Enable tensor cores
```
danpovey: should be use_tensor_cores, no trailing underscore (since it's a struct)... also two spaces before the //.

luitjens:
Going to scan binaries to see which i should add this to.  Do you have any you would like to see it added to?

src/nnet3bin/nnet3-compute-batch.cc
```diff
@@ -88,6 +88,7 @@ int main(int argc, char *argv[]) {
     }
 
 #if HAVE_CUDA==1
+    CuDevice::RegisterDeviceOptions(&po);
```
danpovey: This would not work because you are registering the args after calling po.Read(), it would
have to be called further above.
Note that the call CuDevice::RegisterDeviceOptions() does not exist if HAVE_CUDA is not 1.
I think the best solution would be to have the device options class declared outside of the ifdef
(e.g., towards the bottom of cu-device.h), outside of class CuDevice, so that we don't have
to worry about the option being undefined when run without CUDA being compiled in, which
would make sure the script wouldn't crash if that options was given.

luitjens: Why not just put another HAVE_CUDA block and move it before read?  This way it only gets registered if HAVE_CUDA=true?

luitjens: Please see the version I just committed and see if that works well?  If cuda is not used we don't want to still have that option in the command line right?

danpovey:
OK, I guess that works.

PR #3067, opened 2019-03-02
Add MergeTaskOutputGPU which keeps GPU data on the device.

luitjens:
Adding MergeTaskOutputGPU.

This keeps data on the device to avoid unnecessary data copies when subsequent decoding will be on the GPU.

### DIFF
--- a/src/nnet3/nnet-batch-compute.h
+++ b/src/nnet3/nnet-batch-compute.h
@@ -193,7 +193,7 @@ struct NnetBatchComputerOptions: public NnetSimpleComputationOptions {
 void MergeTaskOutput(
     const std::vector<NnetInferenceTask> &tasks,
     Matrix<BaseFloat> *output);
-void MergeTaskOutputGPU(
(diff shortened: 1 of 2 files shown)


src/nnet3/nnet-batch-compute.h
```diff
@@ -193,6 +193,9 @@ struct NnetBatchComputerOptions: public NnetSimpleComputationOptions {
 void MergeTaskOutput(
     const std::vector<NnetInferenceTask> &tasks,
     Matrix<BaseFloat> *output);
+void MergeTaskOutputGPU(
```
danpovey: Thanks.  You can just call it MergeTaskOutput() and let the function overloading take care of it, that's what I normally do for CPU/GPU things.  Incidentically, if it had the  suffix, it would be MergeTaskOutputGpu()... it's a Google style guide thing.

luitjens: Fixed

PR #3068, opened 2019-03-02
Remove stream0 copies. 

luitjens:
Remove stream0 copies.  
Semantics here are the same except these copies occur in the per-thread default stream which avoids unnecessary synchronization.

### DIFF
--- a/src/cudamatrix/cu-array-inl.h
+++ b/src/cudamatrix/cu-array-inl.h
@@ -212,7 +212,8 @@ void CuArrayBase<T>::SetZero() {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) {
     CuTimer tim;
-    CU_SAFE_CALL(cudaMemset(this->data_, 0, this->dim_ * sizeof(T)));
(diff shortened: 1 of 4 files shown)


src/cudamatrix/cu-array-inl.h
```diff
@@ -212,7 +212,8 @@ void CuArrayBase<T>::SetZero() {
 #if HAVE_CUDA == 1
   if (CuDevice::Instantiate().Enabled()) {
     CuTimer tim;
-    CU_SAFE_CALL(cudaMemset(this->data_, 0, this->dim_ * sizeof(T)));
+    CU_SAFE_CALL(cudaMemsetAsync(this->data_, 0, this->dim_ * sizeof(T),cudaStreamPerThread));
```
danpovey: space after comma, please (google style guide)-- and line break if going over 80 chars per line.

luitjens: should be fixed now

danpovey:
thanks, will merge, probably, when checks finish.

PR #3072, opened 2019-03-04
[src] Log nnet3 computation to VLOG, not std::cout

kkm000:
Use an auxiliary stream inserter class to print the nnet3 Computation instance to the logging stream.

Also remove the declaration of undefined 'operator<<()'.

### DIFF
--- a/src/nnet3/decodable-simple-looped.cc
+++ b/src/nnet3/decodable-simple-looped.cc
@@ -52,7 +52,6 @@ DecodableNnetSimpleLoopedInfo::DecodableNnetSimpleLoopedInfo(
   Init(opts, &(am_nnet->GetNnet()));
 }
 
-
 void DecodableNnetSimpleLoopedInfo::Init(
     const NnetSimpleLoopedComputationOptions &opts,
     Nnet *nnet) {
@@ -86,10 +85,8 @@ void DecodableNnetSimpleLoopedInfo::Init(
   CompileLooped(*nnet, opts.optimize_config, request1, request2, request3,
                 &computation);
   computation.ComputeCudaIndexes();
-  if (GetVerboseLevel() >= 3) {
-    KALDI_VLOG(3) << "Computation is:";
-    computation.Print(std::cerr, *nnet);
-  }
+  KALDI_VLOG(3) << "Computation is:\n"
+                << NnetComputationPrintInserter{computation, *nnet};
 }
 
 
--- a/src/nnet3/nnet-computation.h
+++ b/src/nnet3/nnet-computation.h
@@ -514,17 +514,22 @@ struct NnetComputation {
   NnetComputation(): need_model_derivative(false) { }
 };
 
-
-
-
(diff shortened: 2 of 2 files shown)


src/nnet3/decodable-simple-looped.cc
```diff
@@ -86,10 +85,8 @@ void DecodableNnetSimpleLoopedInfo::Init(
   CompileLooped(*nnet, opts.optimize_config, request1, request2, request3,
                 &computation);
   computation.ComputeCudaIndexes();
-  if (GetVerboseLevel() >= 3) {
-    KALDI_VLOG(3) << "Computation is:";
-    computation.Print(std::cerr, *nnet);
-  }
+  KALDI_VLOG(3) << "Computation is:\n"
+                << NnetComputationPrintInserter{computation, *nnet};
```
danpovey: I think the reason I did this to cerr directly rather than to a temporary string is that it is in danger of exhausting memory otherwise, it's very verbose.   I believe these macros use temporary strings and store the string somewhere... I don't think this PR is a good idea, for that reason.

kkm000: Just checked--about 45K characters, which is pretty much nothing. Actually, this is a cherry-picked changeset that has been in production here since January, 21 last year, if I'm to trust my Git. And we valgrind it thoroughly.

The string is deleted when the MessageLogger goes out of scope, as soon as this line ends executing; it's a constructed temporary without a name.

kkm000: You did actually declare the `operator<<` (without a definition), but printing requires both the computation and the nnet, so requires an inserter. It really looked like something left to implement later. So I just caught up where you left it : )

danpovey:
OK, but can  you do it manually using a temporary string, rather than using
that special class?
it will be easier to understand.
I don't remember why that thing was declared-- and thanks for finishing
it-- but I would rather just keep it simple.

kkm000:
I do not have access to the ostream inside the `<<` .. `<<` chain to pass it to the Print member. What I'd have to do would be
```c++
if (GetVerboseLevel() >= 3) {
  std::ostrstream temp;
  computation.Print(temp, nnet);
  KALDI_VLOG(3) << "The computation is\n" << temp.str();
}
```
which is plain ugly, hardly more readable, and allocates an extra copy of the whole formatted computation string. An inserter class, on the other hand, is pretty paradigmatic with streams, when formatting takes more than one thing to print, i. e. when an overloaded `operator<<` does not cut it. And naturally avoids the extra allocation.

kkm000:
Come think of it, the thing is rather minor, and I do not want to steal your time arguing about it. I just thought it would be helpful to other guys who use custom logging, and assumed it would be entirely non-contentious. Not worth arguing about. I can just take out the dead declaration if you really prefer.

danpovey:
it's OK, I'll merge it.

kkm000:
Thanks, Dan!

PR #3084, opened 2019-03-10
Python3 error: minor modifications to avoid error at Stage 20 of run.sh in Librispeech

nikhilm16:
Changes are made to "egs/wsj/s5/local/chain/tuning/run_tdnn_1g.sh" and " egs/wsj/s5/utils/data/perturb_data_dir_volume.sh" for running the scripts with Python3 as default python.


egs/wsj/s5/utils/data/perturb_data_dir_volume.sh
```diff
@@ -43,6 +43,26 @@ fi
 # foo-1 cat foo.wav | sox --vol 1.6 -t wav - -t wav - |    # volume perturbation done
 # bar-1 sox --vol 1.2 bar.wav -t wav - |                   # volume perturbation done
 # foo-2 wav-reverberate --additive-signals="sox --vol=0.1 noise1.wav -t wav -|" foo.wav |   # volume perturbation not done
+# volume_perturb_done=`head -n100 $data/wav.scp | python -c "
```
danpovey: thanks.  You can remove this commented block.

nikhilm16: I have removed the commented block.

egs/wsj/s5/utils/data/perturb_data_dir_volume.sh
```diff
@@ -52,18 +52,19 @@ for line in sys.stdin.readlines():
   parts = line.strip().split()
   if line.strip()[-1] == '|':
     if re.search('sox --vol', ' '.join(parts[-11:])):
-      print 'true'
+      print("True")
```
danpovey: please don't change the capitalization of these things.  The standard things being referred to here are /usr/bin/false and /usr/bin/true, the upper-cased versions don't work on Debian linux at least.

nikhilm16: for me the lower-cased true/false was not working. Only, True/False was working. So I write True/False to volume_perturb_done. 
  File "<string>", line 10, in <module>
NameError: name 'true' is not defined

danpovey: That's because you also changed the single quotes to double quotes.  It's part of a double-quoted string, so double-quotes disappear.

nikhilm16: Use of single quotes helped. I have modified the PR

danpovey:
thanks!

PR #3085, opened 2019-03-10
remove (dash) from default  uterance suffix from augment_data_dir.py

rezame:
remove − (dash) from default utterance suffix. in this code if set --utt-suffix " " add a dash - to end of utt.
(must change SRE16 run.sh based of this.)


danpovey:
@phanisankar-nidadavolu and @david-ryan-snyder  please check this.

danpovey:
@phanisankar-nidadavolu can you comment on this?

david-ryan-snyder:
I think it's pretty standard to use "-" as the delimiter in Kaldi. And changing this would require updating 3 speaker recognition recipes.

I don't think this change is needed in the main branch.

danpovey:
OK, closing.  Thanks for the PR though.

PR #3089, opened 2019-03-11
[src] Make sure expf() speed probe times sensibly

kkm000:
I was getting random slow expf() detection even in different runs of configure without any changes in command line or switches. In addition to the CPU governor failing to bump up the clock quickly enough, the probe as currently written times the loops in microsecond or sub-microsecond range on a fast CPU.

Here I make sure that (1) the CPU clock is hot and (2) the test runs for 20 to 60 ms, to make the timing at the very least non-random.

Personally, I doubt the test is valid and the problem exists at all. Any sensible compiler will emit a hardware exponentiation instruction anyway. `-O0` is not the way to test performance of anything. But at the very least, in this form we won't get *random* false positives : )

No positives detected with
```
for cc in g++-6 g++-7 g++-8 clang++-5.0 clang++-6.0 clang++-7; do
  ${cc} -std=c++11 -I.. -isystem /home/kkm/work/kaldi2/tools/openfst/include -O0 exp-test.cc -o exp-test &&
    sleep 1 &&
    echo "${cc}: $(taskset 1 ./exp-test)"
done
```
(or with `-march=native`).

### DIFF
--- a/src/probe/README.slow_expf
+++ b/src/probe/README.slow_expf
@@ -1,5 +1,6 @@
-On some machines, expf() turns out to be very slow: much slower than its double precision counterpart exp().
-Probably this is concerned with the version of glibc.
+On some machines, expf() turns out to be very slow: much slower than its double
+precision counterpart exp().  Probably this is concerned with the version of
+glibc.
 
 Here are a couple of examples:
 
@@ -21,5 +22,7 @@ configuration$ ./exp-test
 exp() time: 0.0028439
 expf() time: 0.00713329
 
-If slow behaviour is detected, then KALDI_NO_EXPF macro will be used, and the Exp() wrapper in base/kaldi-math.h will use exp() even for single precision floats.
-The behaviour of expf() is considered to be slow if it is slower than exp() by at least 10%.
+If slow behaviour is detected, then KALDI_NO_EXPF macro will be used, and the
+Exp() wrapper in base/kaldi-math.h will use exp() even for single precision
+floats.  The behaviour of expf() is considered to be slow if it is slower than
+exp() by at least 10%.
--- a/src/probe/exp-test.cc
+++ b/src/probe/exp-test.cc
@@ -17,35 +17,52 @@
 // See the Apache 2 License for the specific language governing permissions and
 // limitations under the License.
 
+// Read Makefile.slow_expf. This test must be compiled with -O0.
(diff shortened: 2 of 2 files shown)


kkm000:
I take back "the problem does not exist" and the "any sensible compiler" statements:  https://stackoverflow.com/questions/7420665/

Perhaps `-ffast-math -no-fassociative-math` is the way to go in general. glibc is apparently trying to be anally correct about IEEE arithmetic, at the expense of performance, to avoid even the LSB errors. I do not think we should care about the full 16 decimal places being correct.  But this needs some testing.

danpovey:
This expf thing was a real problem for a particular glibc version.
I think this

https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1663280

may reference that issue.

-ffast-math is actually a problem for kaldi code right now, it leads to
run-time errors in nnet3 compilation and in decoding.  The problem is that
certain parts of the code use, and test for, infinities, and -ffast-math
makes the assumption that infinities do not exist, and therefore optimizes
out all tests for them.  We would need a way to test for infinities that
was invulnerable to optimization.

When using icc, Nagendra (cc'd) did find a way to have fast math in general
but disable the no-infinities assumption; he found that the two following
options (for icc)

export CXXFLAGS="-fPIC -xHost -O3 -m64 -IPO -fp-model fast=2
-fno-finite-math-only"
export CXXFLAGS="-fPIC -xHost -O3 -m64 -IPO -fp-model fast=2"

led to the same decoding speed.  I don't recall whether the option
-fp-model fast=2 actually helped the speed in the first place.  I wouldn't
expect it to make a huge difference because our math-heavy stuff is done in
BLAS anyway.

There may be a similar option available in gcc.

kkm000:
@danpovey, yes, I think I already understand why glibc is doing its things the hard way. I spent a couple hours reading about IEEE math. It put me in the state of a nearly primordial awe...

Just to make it clear, this change is still very valid, and fixes a real problem. It does not change the logic of the original test, just makes sure we run enough iterations to trust the result. The worst part of the current probe is that the loop is too short, so the result is essentially random on a fast CPU.

danpovey:
OK, merging.  The tests are failing due to PortAudio issues.  You might
consider removing "ext" from travis (or finding out what the problem is).

Issue #3091, opened 2019-03-12
still got python3 compatibility issue in segmentation scripts


XIAOYixuan:
https://github.com/kaldi-asr/kaldi/blob/abd4869ca4f527b361e567c2eb29b06d0f3a6d9b/egs/wsj/s5/steps/segmentation/internal/merge_targets.py#L114
https://github.com/kaldi-asr/kaldi/blob/abd4869ca4f527b361e567c2eb29b06d0f3a6d9b/egs/wsj/s5/steps/segmentation/internal/merge_targets.py#L134

I tried this new script recently and my program crashed at line 134(as referenced above), the error log says `num_sources` is not an int. 
In python 3, though both `len(row)` and  `dim` are ints, and len(row) is divisible by dim, the type of the result is still "float", which caused that error.
![divi](https://user-images.githubusercontent.com/11681498/54169766-0aa40e00-44af-11e9-95a4-2c7e6f71b29a.png)

danpovey:
I think if you put `c = a // b` it should work.   Can you also change the shebang (first line) to python3, remove the `from __future__` statements (no longer needed), re-test, and if it works, make a PR?

XIAOYixuan:
sure, I'm re-testing now

kkm000:
Fixed in #3094.

PR #3094, opened 2019-03-12
Removed import __future__ and fixed a bug in merge_targets.py

XIAOYixuan:
https://github.com/kaldi-asr/kaldi/issues/3091  solved the problem mentioned in this issue, re-tested and everything works fine now


jtrmal:
The way you changed things means the script will not run under python2
You will need to change the shebang as well.
Y.

PR #3097, opened 2019-03-12
[doc] fix some typos in doc.

csukuangfj:


danpovey:
note: these changes only become live when I compile, which is not super often

PR #3114, opened 2019-03-15
[src] Adding GPU/CUDA lattice batched decoder + binary

hugovbraun:
Adding the GPU lattice, batched, decoder. We're currently at 3500 XRTF end-to-end on V100, while generating full lattices, using this decoder and a TDNN acoustic model. 
Using the same model, we get 381 XRTF on a dual socket Titanium Xeon, using all 96 hardware threads. 
We'll publish more performance results in the next few days, alongside with a docker container containing this branch ready to use, allowing everybody to easily run the binary and test performance.

This PR has two components:

- cudadecoder/ , which contains the decoder itself.
- cudadecoderbin/, which contains a binary using the decoder. It can be used directly or viewed as an example on how to use the (batched) cuda decoder.

We've been testing stability/correctness on a variety of models.

Other authors: @luitjens @ryanleary 
Many thanks to @danpovey @chenzhehuai @galv for their help during the pre-review process

### DIFF
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -12,8 +12,8 @@ endif
 
 TESTFILES =
 
-OBJFILES += batched-threaded-cuda-decoder.o decodable-cumatrix.o \
-            cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
+OBJFILES = batched-threaded-cuda-decoder.o decodable-cumatrix.o \
+           cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
 
 LDFLAGS += $(CUDA_LDFLAGS)
 LDLIBS += $(CUDA_LDLIBS)
@@ -21,9 +21,9 @@ LDLIBS += $(CUDA_LDLIBS)
 LIBNAME = kaldi-cudadecoder
 
 ADDLIBS = ../cudamatrix/kaldi-cudamatrix.a ../base/kaldi-base.a ../matrix/kaldi-matrix.a \
-	  ../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
-	  ../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
-	  ../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
+          ../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
+          ../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
+          ../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
 
 # Implicit rule for kernel compilation
 %.o : %.cu
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -3,10 +3,11 @@ all:
 EXTRA_CXXFLAGS = -Wno-sign-compare
 include ../kaldi.mk
 
+# Make sure we have CUDA_ARCH from kaldi.mk,
 ifeq ($(CUDA), true)
-	# Make sure we have CUDA_ARCH from kaldi.mk,
-	ifndef CUDA_ARCH
-	$(error CUDA_ARCH is undefined, run 'src/configure')
+  ifndef CUDA_ARCH
+    $(error CUDA_ARCH is undefined, run 'src/configure')
+  endif
 endif
 
 TESTFILES =
@@ -31,5 +32,3 @@ ADDLIBS = ../cudamatrix/kaldi-cudamatrix.a ../base/kaldi-base.a ../matrix/kaldi-
 	$(CUDATKDIR)/bin/nvcc -c $< -o $@ $(CUDA_INCLUDE) $(CUDA_FLAGS) $(CUDA_ARCH) -I../ -I$(OPENFSTINC)
 
 include ../makefiles/default_rules.mk
-
-endif
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -3,19 +3,17 @@ all:
 EXTRA_CXXFLAGS = -Wno-sign-compare
 include ../kaldi.mk
 
-# Make sure we have CUDA_ARCH from kaldi.mk,
 ifeq ($(CUDA), true)
-  ifndef CUDA_ARCH
-    $(error CUDA_ARCH is undefined, run 'src/configure')
-  endif
+
+# Make sure we have CUDA_ARCH from kaldi.mk,
+ifndef CUDA_ARCH
+  $(error CUDA_ARCH is undefined, run 'src/configure')
 endif
 
 TESTFILES =
 
-OBJFILES += batched-threaded-cuda-decoder.o decodable-cumatrix.o
-ifeq ($(CUDA), true)
-	OBJFILES += cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
-endif
+OBJFILES += batched-threaded-cuda-decoder.o decodable-cumatrix.o \
+            cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
 
 LDFLAGS += $(CUDA_LDFLAGS)
 LDLIBS += $(CUDA_LDLIBS)
@@ -23,12 +21,16 @@ LDLIBS += $(CUDA_LDLIBS)
 LIBNAME = kaldi-cudadecoder
 
 ADDLIBS = ../cudamatrix/kaldi-cudamatrix.a ../base/kaldi-base.a ../matrix/kaldi-matrix.a \
-../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
-../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
-../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
+	  ../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
+	  ../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
+	  ../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
 
-# Implicit rule for kernel compilation,
+# Implicit rule for kernel compilation
 %.o : %.cu
 	$(CUDATKDIR)/bin/nvcc -c $< -o $@ $(CUDA_INCLUDE) $(CUDA_FLAGS) $(CUDA_ARCH) -I../ -I$(OPENFSTINC)
 
 include ../makefiles/default_rules.mk
+
+else # ifeq ($(CUDA), true)
+depend:
+endif
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -1,0 +1,35 @@
+all:
+
+EXTRA_CXXFLAGS = -Wno-sign-compare
+include ../kaldi.mk
+
+ifeq ($(CUDA), true)
+	# Make sure we have CUDA_ARCH from kaldi.mk,
+	ifndef CUDA_ARCH
+	$(error CUDA_ARCH is undefined, run 'src/configure')
+endif
+
+TESTFILES =
+
+OBJFILES += batched-threaded-cuda-decoder.o decodable-cumatrix.o
+ifeq ($(CUDA), true)
+	OBJFILES += cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
+endif
+
+LDFLAGS += $(CUDA_LDFLAGS)
+LDLIBS += $(CUDA_LDLIBS)
+
+LIBNAME = kaldi-cudadecoder
+
+ADDLIBS = ../cudamatrix/kaldi-cudamatrix.a ../base/kaldi-base.a ../matrix/kaldi-matrix.a \
+../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
+../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
+../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
+
+# Implicit rule for kernel compilation,
+%.o : %.cu
+	$(CUDATKDIR)/bin/nvcc -c $< -o $@ $(CUDA_INCLUDE) $(CUDA_FLAGS) $(CUDA_ARCH) -I../ -I$(OPENFSTINC)
+
+include ../makefiles/default_rules.mk
+
+endif
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -12,7 +12,7 @@ endif
 
 TESTFILES =
 
-OBJFILES = batched-threaded-cuda-decoder.o decodable-cumatrix.o \
+OBJFILES = batched-threaded-nnet3-cuda-pipeline.o decodable-cumatrix.o \
            cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
 
 LDFLAGS += $(CUDA_LDFLAGS)
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -23,14 +23,11 @@ LIBNAME = kaldi-cudadecoder
 ADDLIBS = ../cudamatrix/kaldi-cudamatrix.a ../base/kaldi-base.a ../matrix/kaldi-matrix.a \
           ../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
           ../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
-          ../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
+          ../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a
 
 # Implicit rule for kernel compilation
 %.o : %.cu
 	$(CUDATKDIR)/bin/nvcc -c $< -o $@ $(CUDA_INCLUDE) $(CUDA_FLAGS) $(CUDA_ARCH) -I../ -I$(OPENFSTINC)
+endif
 
 include ../makefiles/default_rules.mk
-
-else # ifeq ($(CUDA), true)
-depend:
-endif
--- a/src/cudadecoder/README
+++ b/src/cudadecoder/README
@@ -1,0 +1,141 @@
+CUDADECODER USAGE AND TUNING GUIDE
+
+INTRODUCTION:
+
+The CudaDecoder was developed by NVIDIA with coordination from Johns Hopkins.
+This work was intended to demonstrate efficient GPU utilization across a range 
+of NVIDIA hardware from SM_35 and on.  The following guide describes how to 
+use and tune the decoder for your models.
+
+A single speech-to-text is not enough work to fully saturate any NVIDIA GPUs.
+To fully saturate GPUs we need to decode many audio files concurrently.  The
+solution provide does this through a combination of batching many audio files
+into a single speech pipeline, running multiple pipelines in parallel on the
+device, and using multiple CPU threads to perform feature extraction and 
+determinization.  Users of the decoder will need to have a high level 
+understanding of the underlying implementation to know how to tune the 
+decoder.  
+
+The interface to the decoder is defined in "batched-threaded-cuda-decoder.h".
+A binary example can be found in cudadecoderbin/batched-wav-nnet3-cuda.cc".
+Below is a simple usage example. 
+/*
+ *  BatchedThreadedCudaDecoderConfig batchedDecoderConfig;
+ *  batchedDecoderConfig.Register(&po);
+ *  po.Read(argc, argv);
+ *  ...
+ *  BatchedThreadedCudaDecoder CudaDecoder(batchedDecoderConfig);
+ *  CudaDecoder.Initialize(*decode_fst, am_nnet, trans_model);
+ *  ...
+ *
+ *  for (; !wav_reader.Done(); wav_reader.Next()) {
+ *    std::string key = wav_reader.Key();
+ *    CudaDecoder.OpenDecodeHandle(key, wave_reader.Value());
+ *    ...
+ *  }
+ *
+ *  while (!processed.empty()) {
+ *    CompactLattice clat;
+ *    CudaDecoder.GetLattice(key, &clat);
+ *    CudaDecoder.CloseDecodeHandle(key);
+ *    ...
+ *  }
+ *
+ *  CudaDecoder.Finalize();
+ */
+
+In the code above we first declare a BatchedThreadedCudaDecoderConfig
+and register its options.  This enables us to tune the configuration 
+options.   Next we declare the CudaDecoder with that configuration.
+Before we can use the CudaDecoder we need to initalize it with an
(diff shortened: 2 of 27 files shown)


src/cudadecoderbin/batched-wav-nnet3-cuda.cc
```diff
@@ -0,0 +1,301 @@
+// online2bin/online2-wav-nnet3-latgen-faster.cc
```
danpovey: fix this line please

luitjens:
Dan any idea why travis is failing with a link outside of the code this patchset touches?

/home/travis/build/kaldi-asr/kaldi/src/online/online-audio-source.cc:161: undefined reference to `PaUtil_GetRingBufferWriteAvailable'
/home/travis/build/kaldi-asr/kaldi/src/online/online-audio-source.cc:164: undefined reference to `PaUtil_WriteRingBuffer'
kaldi-online.a(online-audio-source.o): In function `OnlinePaSource':
/home/travis/build/kaldi-asr/kaldi/src/online/online-audio-source.cc:71: undefined reference to `PaUtil_InitializeRingBuffer'
kaldi-online.a(online-audio-source.o): In function `kaldi::OnlinePaSource::Read(kaldi::Vector<float>*)':
/home/travis/build/kaldi-asr/kaldi/src/online/online-audio-source.cc:122: undefined reference to `PaUtil_GetRingBufferReadAvailable'
/home/travis/build/kaldi-asr/kaldi/src/online/online-audio-source.cc:122: undefined reference to `PaUtil_GetRingBufferReadAvailable'
/home/travis/build/kaldi-asr/kaldi/src/online/online-audio-source.cc:138: undefined reference to `PaUtil_ReadRingBuffer'

danpovey:
Sorry, this is a pre-existing condition.
It has nothing to do with your changes.  The "make test" works, which means
it is basically working.

src/cudadecoder/batched-threaded-cuda-decoder.cc
```diff
@@ -0,0 +1,637 @@
...
+#include "base/kaldi-utils.h"
+#include <nvToolsExt.h>
+
+namespace kaldi {
+namespace CudaDecode {
+
+void BatchedThreadedCudaDecoder::Initialize(
+    const fst::Fst<fst::StdArc> &decode_fst, const nnet3::AmNnetSimple &am_nnet,
+    const TransitionModel &trans_model) {
+  KALDI_LOG << "BatchedThreadedCudaDecoder Initialize with "
```
btiplitz: you should also print out the num_worker_threads to be consistent

hugovbraun: thanks! also added the batch size

Plusmonkey:
Interesting, when compiling this branch, cuda-decoder.o happens error, /tools/openfst/include/fst/push.h:132:126: error: ‘g’ was not declared in this scope
       total_weight = GallicWeight(
                                                                                                                              ^
/tools/openfst/include/fst/push.h:132:193: error: template argument 2 is invalid
       total_weight = GallicWeight(
                                                                                                                                                                                                 ^
# --error 0x1 --
make: *** [cuda-decoder.o] Error 1

btiplitz:
@Plusmonkey I've had that exact error before.  I believe the cause was something in my tool chain was out of date.

danpovey:
We should figure this out and take care of it when configuring.  At least
to detect it.  Brett, can you tell him what diagnostics might be relevant?

btiplitz:
@danpovey I tracked down when I got this exact issue.  I was compiling with Ubuntu 14. The default compiler there is 4.8.4.  It seems there was a bug in 4.8.4 and I needed to have 4.8.5 to get past this.  Having 4.9.4 around, I upgraded to that version.  I looked in the changelog and I am unsure how I determined the issue was fixed in 4.8.5.  

@Plusmonkey can you tell me which version of gcc you are running.

btiplitz:
last comment was for @hugovbraun

btiplitz:
@hugovbraunOn line34 of cudadecoder/Makefile, please add (the non-cuda build fails without this) - sorry the browser at work won't allow in-line comments.
else
depend:

src/cudadecoderbin/Makefile
```diff
@@ -0,0 +1,27 @@
+all:
```
chenzhehuai: add this piece of code in tools/config/common_path.sh
@@ -22,4 +22,5 @@ ${KALDI_ROOT}/src/rnnlmbin:\
 ${KALDI_ROOT}/src/sgmm2bin:\
 ${KALDI_ROOT}/src/sgmmbin:\
 ${KALDI_ROOT}/src/tfrnnlmbin:\
+${KALDI_ROOT}/src/cudadecoderbin:\
 $PATH

btiplitz:
@danpovey windows build is almost done. 1 error at a time

hugovbraun:
Thanks @btiplitz and @chenzhehuai I'll add those shortly. Just found a bug that appeared after the commits squash, fixing that and then adding the new rules

hugovbraun:
The bug was with the aspire model and was due to the rebasing on top of last week's master. The good news is that rebasing on today's master solved the issue. 
@btiplitz when opening the file to apply your changes I noticed that something was wrong with the nested if statements (in that version: https://github.com/kaldi-asr/kaldi/blob/080c805665c54329f716a8b7f08378c51ba57299/src/cudadecoder/Makefile ). I fixed it, but I'm not sure what the "else" in your change must refer to now, could you check the latest ?

btiplitz:
@hugovbraun the else only runs on non gpu builds and basically does nothing.

hugovbraun:
@btiplitz updated, the non-cuda builds now have an empty depend target

cloudhan:
@Plusmonkey are you using CUDA 8.0? 

I also failed building this branch with CUDA 8.0.61 and gcc 5.4.0. But it compiles and runs smoothly with CUDA 9.0.

@hugovbraun Does this decoder officially support CUDA8?

luitjens:
we have not officially tested on cuda 8.   we can probably work around any
cuda 8.0 compiler errors though.

Plusmonkey:
@btiplitz  Hi , I use gcc version 4.8.5 20150623 (Red Hat 4.8.5-36) (GCC)

cloudhan:
@luitjens @Plusmonkey 
I drilled down the `/tools/openfst/include/fst/push.h:132:126: error: ‘g’ was not declared in this scope` and it seems to be related with compiler

in `include/fst/push.h`
```c++
      total_weight = GallicWeight(
          ptype & kPushRemoveCommonAffix
              ? total_weight.Value1()
              : StringWeight<Label, GallicStringType(gtype)>::One(),
          ptype & kPushRemoveTotalWeight ? total_weight.Value2()
                                         : Weight::One());
```

in the generated `cuda-decoder-kernels.cu.cpp.ii ` at line 214785
```
total_weight = GallicWeight((ptype & kPushRemoveCommonAffix) ? (total_weight.Value1()) : StringWeight< typename Arc::Label, (g == 0) ? STRING_LEFT : ((g == 1) ? STRING_RIGHT :        STRING_RESTRICT)> ::One(), (ptype & kPushRemoveTotalWeight) ? (total_weight.Value2()) : Weight::One());
```
Apparently `GallicStringType` has been expended, which is defined as 
```c++
constexpr StringType GallicStringType(GallicType g) {
  return g == GALLIC_LEFT
             ? STRING_LEFT
             : (g == GALLIC_RIGHT ? STRING_RIGHT : STRING_RESTRICT);
}
```
and unfortunately the formal parameter `g` was not replace by the actually parameter `gtype`.

If I manually replace the the `g` with `gtest` at line 214785 of `cuda-decoder-kernels.cu.cpp.ii `, this file compiles smoothly.

Plusmonkey:
@luitjens @cloudhan  Yes, I alse use cuda 8.0

hugovbraun:
Thanks @cloudhan for looking into that. I'll remove the openfst dependency for the kernel files, it is not really needed anyway. It should fix this issue.

hugovbraun:
I added wrappers around the kernels and I did some refactoring so that only the kernels are compiled by nvcc, other files are compiled directly by (pure) g++. OpenFst is never compiled by nvcc, fixing the cuda8.0 compilation issue.

danpovey:
Nice-- easy compilation is important.
We were trying to compile the `marian` project on the CLSP grid and it was no end of trouble because they use things like std::tuple in code that is compiled by nvcc.

Plusmonkey:
@hugovbraun Hi , Can you give an example about how to use this cuda decoder?

src/cudadecoder/decodable-cumatrix.h
```diff
@@ -0,0 +1,66 @@
...
+  DecodableCuMatrixMapped(const TransitionModel &tm,
+                          const CuMatrixBase<BaseFloat> &likes,
+                          int32 frame_offset = 0);
+
+  virtual int32 NumFramesReady() const;
+
+  virtual bool IsLastFrame(int32 frame) const;
+
+  virtual BaseFloat LogLikelihood(int32 frame, int32 tid) {
+    KALDI_ASSERT(false);
```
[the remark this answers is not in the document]
jtrmal: IMO because gcc is able to figure out (from our annotations) that KALDI_ASSERT does not return. But I might be mistaken.

btiplitz:
@Plusmonkey I've coded using this and run it through a ton of testing.  What are you looking for?  Are you trying to use kaldi scripts or code?

src/cudadecoder/decodable-cumatrix.h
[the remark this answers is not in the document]
btiplitz: Since it is declared as non-void the standard says you need a return.  I don't think the language says that you don't need one if it never returns.  Like if you have an infinite loop, it's also required.  gcc/g++ is usually pretty good on this, but with centos, the gcc version is 10 years old or more.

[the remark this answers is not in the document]
hugovbraun: Yes gcc only triggers a warning when the return is missing, always found it weird. I’ll fix it, thanks for the info

Plusmonkey:
@btiplitz Yes, I am trying to use kaldi scripts and code.

src/cudadecoder/Makefile
```diff
@@ -0,0 +1,35 @@
...
+../lat/kaldi-lat.a ../util/kaldi-util.a ../matrix/kaldi-matrix.a ../gmm/kaldi-gmm.a \
+../fstext/kaldi-fstext.a ../hmm/kaldi-hmm.a ../gmm/kaldi-gmm.a ../transform/kaldi-transform.a \
+../tree/kaldi-tree.a ../online2/kaldi-online2.a ../nnet3/kaldi-nnet3.a  \
+
+# Implicit rule for kernel compilation,
+%.o : %.cu
+	$(CUDATKDIR)/bin/nvcc -c $< -o $@ $(CUDA_INCLUDE) $(CUDA_FLAGS) $(CUDA_ARCH) -I../ -I$(OPENFSTINC)
+
+include ../makefiles/default_rules.mk
+
```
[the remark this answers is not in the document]
hugovbraun: It looks like that comment concerns an older commit. Now depend: should come directly from the include ../makefiles/default_rules.mk, which is always included (cuda/non-cuda). Is that ok with windows ?

bamba518:
@ hugovbraun if I want to use this with other packages that create binaries
https://github.com/jimbozhang/kaldi-gop
What steps would need to be done

cloudhan:
If I want to keep the beam and lattice-beam unchanged, then what is the key factor that will affect the decoding speed?

luitjens:
If you want to use just the decoder then you should look how we used it
in src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.cc

luitjens:
@Cloud Han,  i'm not sure i understand the question.  A lot of thing impact
decoding speed including but not limited to:

1) GPU being used
2) Model being tested
3) Audio being decoded

cloudhan:
@luitjens To be clear, I mean with the same machine, GPU, model, audio input, what will affect this decoder's throughput. there seems to be some option will affect the scheduling, e.g.
```
--max-batch-size=200
--batch_drain_size=10
--cuda-control-threads=2
--cuda-worker-threads=16
--max-outstanding-queue-length=4000
```
from the documentation, I can see the larger `max-batch-size`, the faster, the more memory hungry. 
Any other option will also affect the speed, aka, throughput.

As I can see from `nvidia-smi`, the power consumption is only 145/250W for a 1080Ti, and the temperature is only 45 degree C, so I don't think it is busy enough...

src/cudadecoder/cuda-decoder-common.h
```diff
@@ -0,0 +1,468 @@
...
+// Because keeping a bit more than max_active_ is expected, we add the tolerance
+// so that we can avoid triggering ApplyMaxActiveAndReduceBeam for just a few
+// tokens above the limit
+// at the end of the frame
+#define KALDI_CUDA_DECODER_MAX_ACTIVE_TOLERANCE 0.2
+
+#define KALDI_CUDA_DECODER_DIV_ROUND_UP(a, b) ((a + b - 1) / b)
+
+#define KALDI_CUDA_DECODER_ASSERT(val, recoverable)                     \
+  {                                                                     \
```
galv: Use do { } while(0) for these macros. It's important.

I know that there are some macros in kaldi that don't do this. But still.

src/cudadecoder/cuda-decoder-common.h
```diff
@@ -0,0 +1,468 @@
...
+#define KALDI_CUDA_DECODER_ONE_THREAD_BLOCK 1
+
+namespace kaldi {
+namespace cuda_decoder {
+
+// Returning the number of CTAs to launch for (N,M) elements to compute
+// M is usually the batch size
+inline dim3 KaldiCudaDecoderNumBlocks(int N, int M) {
+  dim3 grid;
+  // TODO MAX_NUM_BLOCKS.
```
galv: Why would we need a max number of blocks? Not all blocks need to be resident on the GPU at once, anyway.

src/cudadecoder/cuda-decoder-common.h
```diff
@@ -0,0 +1,468 @@
...
+  return grid;
+}
+
+typedef int32 StateId;
+typedef float CostType;
+// IntegerCostType is the type used in the lookup table d_state_best_cost
+// and the d_cutoff
+// We use a 1:1 conversion between CostType <--> IntegerCostType
+// IntegerCostType is used because it triggers native atomic operations
+// (CostType does not)
```
galv: See `15-2 Comparing Floating-Point Numbers Using Integer Operations` in Hacker's Delight Second Edition to learn more about the 1:1 çonversion.

src/cudadecoder/cuda-decoder-common.h
```diff
@@ -0,0 +1,468 @@
...
+typedef int32 IntegerCostType;
+typedef int32 LaneId;
+typedef int32 ChannelId;
+
+// On the device we compute everything by batch
+// Data is stored as 2D matrices (BatchSize, 1D_Size)
+// For example, for the token queue, (BatchSize, max_tokens_per_frame_)
+// DeviceMatrix owns the data but is not used to access it.
+// DeviceMatrix is inherited in DeviceLaneMatrix and DeviceChannelMatrix
```
galv: `the parent of`

src/cudadecoder/README
```diff
@@ -0,0 +1,141 @@
...
+model, and data being decoded.
+
+The following will briefly describe each parameter:
+
+BatchedThreadedCudaDecoderOptions:
+  cuda-control-threads:  Number of CPU threads simultaniously submitting work
+    to the device.  For best performance this should be between 2-4.
+  cuda-worker-threads:  CPU threads for worker tasks like determinization and
+    feature extraction.  For best performance this should take up all spare
+    CPU threads available on the system.
```
galv: So it should be `MY_NUM_CORES - cuda-control-threads`, if you assume that your server is otherwise free, right?

src/cudadecoder/README
```diff
@@ -0,0 +1,141 @@
...
+    each frame.  This buffer size cannot be exceed or reallocated.  As this
+    buffer gets closer to being exhausted the beam is reduced possibly reducing
+    quality.  This should be tuned according to the model and data.  For
+    example, a highly accurate model could set this values smaller to enable
+    more concurrent decodes.
+
+  cuda-control-threads:  Each control thread is a concurrent pipeline.  Thus
+    the GPU memory scales linearly with this parameter.  This should always be
+    at least 2 but should probably not be higher than 4 as more concurrent
+    pipelines leads to more driver contention reducing performance.
```
galv: Is there any public documentation on which driver versions reduce the amount of kernel-call lock contention?

src/cudadecoder/cuda-decoder-common.h
```diff
@@ -0,0 +1,468 @@
...
+  float floatVal;
+  // Should be optimized away by compiler
+  memcpy(&floatVal, &intVal, sizeof(float));
+  return floatVal;
+}
+
+// Hashmap value. Used when computing the hashmap in PostProcessingMainQueue
+struct __align__(16) HashmapValueT {
+  // Map key : fst state
+  int32 key;
```
galv: StateId would be better. You're guaranteed that it's an int32 since you have the __align__ pragma.

src/cudadecoder/cuda-decoder-common.h
```diff
@@ -0,0 +1,468 @@
...
+  // Number of tokens associated to that state
+  int32 count;
+  // minimum cost for that state + argmin
+  int2 min_and_argmin_int_cost;
```
galv: This is the best part of the whole code base. I just love how you can just shove the argmin into the 32 bits below the 32-bit score, and atomicMin on 64-bit integers preserves the argmin.

src/cudadecoder/cuda-decoder-kernels-utils.h
```diff
@@ -0,0 +1,215 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef KALDI_CUDA_DECODER_CUDA_DECODER_KERNELS_UTILS_H_
+#define KALDI_CUDA_DECODER_CUDA_DECODER_KERNELS_UTILS_H_
+
+// NO_KEY == -1 is ok, because all keys will be >= 0 (FST states)
+#define KALDI_CUDA_DECODER_HASHMAP_NO_KEY -1
+#define KALDI_CUDA_DECODER_HASHMAP_NO_VAL                 \
+  {                                                       \
```
galv: `do { } while(0)`

src/cudadecoder/cuda-decoder-kernels-utils.h
```diff
@@ -0,0 +1,215 @@
...
+  int32 intVal = __float_as_int(floatVal);
+  return (intVal >= 0) ? intVal : intVal ^ 0x7FFFFFFF;
+}
+
+__device__ __forceinline__ float orderedIntToFloat(int32 intVal) {
+  return __int_as_float((intVal >= 0) ? intVal : intVal ^ 0x7FFFFFFF);
+}
+
+// binsearch_maxle (device)
+// With L=[all indexes low<=i<=high such as vec[i]<= val]
```
galv: What is L here?

You should document that this code assumes that `vec[low] <= val <= vec[high-1] `, since it seems to assume that anyway.

src/cudadecoder/cuda-decoder-kernels-utils.h
```diff
@@ -0,0 +1,215 @@
...
+__device__ __inline__ int2 atomicAddI2(int2 *ptr, int2 val) {
+  unsigned long long int *ptr64 =
+      reinterpret_cast<unsigned long long int *>(ptr);
+  UInt64UnionInt2 uval, uold;
+  uval.i2 = val;
+  uold.ull = atomicAdd(ptr64, uval.ull);
+  return uold.i2;
+}
+
+// We should switch to native atom64 on atomicMinI2 and atomicSubI2
```
galv: Yes, why isn't this using [atomicMin](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicmin) overloaded for `unsigned long long int`? It seems like it would be way faster that way. Do you really need to support compute capability 3.0?

galv:
@hugovbraun Have you uploaded your GTC 2019 slides anywhere?

hugovbraun:
@cloudhan with beam and lattice-beam fixed, the most important factor for speed is the batch size. For now we don't support batch size > 200, but it will be part of a patch in the next weeks. You can try increasing cuda-control-threads for now. 
Printing a warning for cuda 8.0 is probably is good idea. 

@galv Thanks a lot for the review. Kernel call lock contention has been improved continuously since cuda 9. This is still work in progress. 
However the recommended path going forward is clearly to use batches as big as possible. For a good model/well aligned eval data (leading to the creation a few thousands tokens/frame), going from batch size 1 to batch size 200 is almost free for the decoder (constant time). For more challenging models the remaining CPU tasks on the main cuda control thread take more time (time linear w/ batch size), but I've got a patch in progress to remove those. Those tasks are the host2host memory copies, and GetRawLattice currently called on the main thread.

src/cudadecoder/cuda-decoder-kernels-utils.h
hugovbraun: Yes this is in the list. For 3.0 I guess it is still supported in the Kaldi makefiles (we can keep the duplicate and add a #if __CUDA_ARCH__)

src/cudadecoder/README
hugovbraun: Yes. We could load std::thread::hardware_concurrency() by default if you think it's a good idea

src/cudadecoder/cuda-decoder-common.h
hugovbraun: There's a few reasons:
- some kernels aggregate values locally in shared memory first (like the histogram kernel). If the same CTA is in charge of more data (iterating on thread_idx += gridDim.x*blockDim.x), using the shared memory for local aggregates, it will push only once the local aggregates to the global memory. Those global aggregates are atomics so it's a big deal.
- Some data is loaded only once in registers and reused when the CTA loops on thread_idx += gridDim.x*blockDim.x
- Lauching/terminating CTAs has a cost
- More specific to our case is: the amount of work we have to do for each frame depends on that frame (and the previous ones). Which means that each utterances in a batch will have a different number of tokens to process for instance. Because we launch a 2D grid, limiting the number of CTAs allow some load balancing to appear (if we always use the max to decide the x-dimension, all other utterances in the batch will launch CTAs at the end for nothing).

All of this is not important when we generate only a few thousands tokens / frame, but it really speeds up the kernels for challenging models (for this, we usually test on the aspire model evaluated on librispeech/test_other). It gives around 10x on histogram, 2x on expand, etc. 

And for older arch, max # of CTAs / dim[x|y] is ~65000. Even if this is very high for us it's still cleaner to put a clear upper limit

hugovbraun:
@galv GTC19 slides : [GTC19-Kaldi-Acceleration.pdf](https://github.com/kaldi-asr/kaldi/files/3073268/GTC19-Kaldi-Acceleration.pdf)
It does not contain anything too technical though. It was a high level presentation of what we've done and performance numbers in various configurations.

tanvirfuad:
I just wanted to know if there is any way/suggestion/advice to use this for the aspire model? 
By the way, thank you all of you for making this real

danpovey:
Thanks, guys!

lkf123010:
only RealTimeX 9.4x ，

batched-wav-nnet3-cuda --max-batch-size=200 --iterations=4 --cuda-control-threads=2 --cuda-worker-threads=50 --acoustic-scale=1.0 --beam=15.0 --max-active=10000 --lattice-beam=4.0 --ivector-silence-weighting.max-state-duration=40 --ivector-silence-weighting.silence-phones=1 --ivector-silence-weighting.silence-weight=0.00001 --config=/asr/kaldi/egs/aishell2/s5_nopitch/no_pitch_model/online_test/conf/online.conf --word-symbol-table=/asr/kaldi/egs/aishell2/s5_nopitch/no_pitch_model//words.txt /asr/kaldi/egs/aishell2/s5_nopitch/no_pitch_model//final.mdl /asr/kaldi/egs/aishell2/s5_nopitch/no_pitch_model//HCLG.fst 'scp:echo invite1_right2 invite1_right2.wav|' 'ark:|lattice-scale --acoustic-scale=10.0 ark:- ark:- | gzip -c >local/tmp/lat.invite1_right2.gz' 
LOG (batched-wav-nnet3-cuda[5.5.673-fa957]:SelectGpuId():cu-device.cc:223) CUDA setup operating under Compute Exclusive Mode.
LOG (batched-wav-nnet3-cuda[5.5.673-fa957]:FinalizeActiveGpu():cu-device.cc:308) The active GPU is [0]: Tesla M40 24GB	free:22689M, used:256M, total:22945M, free/total:0.988832 version 5.2

----


LOG (batched-wav-nnet3-cuda[5.5.673-fa957]:main():batched-wav-nnet3-cuda.cc:304) ~Group 3 completed Aggregate Total Time: 422.199 Audio: 3994.74 RealTimeX: 9.46175
LOG (batched-wav-nnet3-cuda[5.5.673-fa957]:main():batched-wav-nnet3-cuda.cc:315) Decoded 1 utterances, 0 with errors.
LOG (batched-wav-nnet3-cuda[5.5.673-fa957]:main():batched-wav-nnet3-cuda.cc:317) Overall likelihood per frame was 1.18383 per frame over 399468 frames.
LOG (batched-wav-nnet3-cuda[5.5.673-fa957]:main():batched-wav-nnet3-cuda.cc:320) Overall:  Aggregate Total Time: 422.199 Total Audio: 3994.74 RealTimeX: 9.46175

luitjens:
You only decoding a single utterance.  You need to be decoding thousands of
utterances so that they are done in parallel.

PR #3116, opened 2019-03-15
[scripts] Bug-fix for removing deleted words

psmit:
The type of --max-deleted-words-kept-when-merging in segment_ctm_edits.py
was a string, which prevented the mechanism from working altogether.

Question, why is this actually 1 by default; wouldn't it make more sense to remove deletes from the cleaned text?

### DIFF
--- a/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits.py
+++ b/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits.py
@@ -70,7 +70,7 @@ parser.add_argument("--min-split-point-duration", type=float, default=0.1,
                     help="""Minimum duration of silence or non-scored word
                     to be considered a viable split point when
                     truncating based on junk proportion.""")
-parser.add_argument("--max-deleted-words-kept-when-merging", type = str, default = 1,
+parser.add_argument("--max-deleted-words-kept-when-merging", type = int, default = 1,
                     help = "When merging segments that are found to be overlapping or "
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks for finding this.  

I think this is a tradeoff between trusting the reference (e.g. there may be short words really in there that weren't clearly pronounced) versus trusting the alignment process.

It is going to be an empirical question, whether it helps or not.  I would appreciate it if you could see, on some setup that you have, whether it seems to make any difference.

danpovey:
Can you be more clear on how the output differs, if at all, before vs. after this fix?

psmit:
Yes, I can. Before this fix, any 'del' words (present in ref not in hyp) will appear in the "cleaned" transcript. No matter how many of them there are after each other. I noticed, because I reduce the parameter to 0, and still there appeared deleted words in my cleaned transcripts.

The only place that `max_deleted_words_when_merging` is used is https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits.py#L458 : 

    if num_deleted_words > args.max_deleted_words_kept_when_merging:

and this always evaluates to `false` when the second argument is a string.

danpovey:
OK.  How long were these sequences of words?  E.g. what was the longest one
you noticed?

psmit:
On real data, the longest one I noticed was one, but some artificial tests showed that it would happen as well for longer sequences.

PR #3118, opened 2019-03-15
Fixing allow downsample online wav nnet3 decoding

bayethiernodiop:


src/feat/online-feature.cc
```diff
@@ -77,7 +77,7 @@ template<class C>
 void OnlineGenericBaseFeature<C>::AcceptWaveform(BaseFloat sampling_rate,
                                                  const VectorBase<BaseFloat> &waveform) {
   BaseFloat expected_sampling_rate = computer_.GetFrameOptions().samp_freq;
-  if (sampling_rate != expected_sampling_rate)
+  if (sampling_rate != expected_sampling_rate && !computer_.GetFrameOptions().allow_downsample)
```
danpovey: Actually it's more complicated than this because there is an allow_upsample flag too.

danpovey: you don't seem to be on the right path here, you are just ignoring the mismatch but doing nothing.  @jtrmal would you mind making a PR for this?

jtrmal:
I'll have a look.
y.

src/feat/online-feature.cc
bayethiernodiop: Ah ok sorry. anyway thx.

PR #3119, opened 2019-03-15
Made changes to the augmentation script to make it work for ASR and speaker ID

phanisankar-nidadavolu:


danpovey:
@david-ryan-snyder please LMK if this is good to merge

egs/swbd/s5c/local/make_musan.sh
```diff
@@ -0,0 +1,37 @@
...
+#   http://www.openslr.org/17/
+
+set -e
+in_dir=$1
+data_dir=$2
+use_vocals='Y'
+
+mkdir -p local/musan.tmp
+
+echo "Preparing ${data_dir}/musan..."
```
danpovey: I think I'd like this script to  exit with an error if it detects that what it is trying to create already exists.

egs/swbd/s5c/local/make_musan.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
+# Copyright 2015   David Snyder
+# Apache 2.0.
+#
+# This file is meant to be invoked by make_musan.sh.
```
danpovey: Since we are re-using this script I think it would be better to work on it a bit more carefully-- e.g. use argparse to parse the command line args (with something more standard than the 'Y' to use the vocals, i.e. make it a boolean flag)-- and have a proper usage message printed out if called wrongly.
Also, I think we can put this in a shared place, steps/augmentation/ or wherever, since other scripts will likely want to call it.  The same goes for the shell script-- let's put it somewhere shared, e.g. in the same place; have it check whether its input already exists; give it a usage message; and have it parse command line args properly (e.g. give it a --use-vocals option).  Also some kind of function-level documentation for this python script would be nice (e.g. a doc-string).

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
```diff
@@ -0,0 +1,242 @@
+#!/bin/bash
+
```
danpovey: put the normal apache header on this please, and your name.

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
```diff
@@ -0,0 +1,242 @@
...
+mkdir -p exp/nnet3
+
+if [ -e data/rt03 ]; then maybe_rt03=rt03; else maybe_rt03= ; fi
+
+if [ "$multi_style" == "true" ]; then
+  suffix=_ms
+fi
+
+if [ "$multi_style" == "true" ]; then
+  if [ $stage -le 1 ]; then
```
danpovey: I think you can break this up into more stages, in case it dies somewhere in the middle.  And please have more of the stages check whether their work has already been done and die with a suitable message if so.

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
```diff
@@ -0,0 +1,242 @@
...
+train_set=train_nodup
+
+. ./path.sh
+. ./utils/parse_options.sh
+
+mkdir -p exp/nnet3
+
+if [ -e data/rt03 ]; then maybe_rt03=rt03; else maybe_rt03= ; fi
+
+if [ "$multi_style" == "true" ]; then
```
danpovey: I think we should remove the multi-style option here, for purposes of experiments, since if people didn't want it, they just shouldn't call this.  I know this was useful for you... you can of course use it for your own experiments, but i would have though just settinng noise_list to "clean" would do what you want.
Also I don't like the name "noise_list" as reverb and clean are not noise.  Let's call it "augmentation_list" wherever it occurs.

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -49,6 +52,20 @@ def GetArgs():
...
+    elif args.utt_suffix is None and args.utt_prefix is not None:
+        args.utt_modifier_type = "prefix"
+        args.utt_modifier = args.utt_prefix
+    elif args.utt_suffix is not None and args.utt_prefix is None:
+        args.utt_modifier_type = "suffix"
+        args.utt_modifier = args.utt_suffix
+    else:
+        raise Exception("Trying to add both prefix and suffix. Choose either of them")
+
+    args.modify_spkr_id = (True if args.modify_spkr_id == "true" else False)
```
danpovey: An example of how to register boolean options:
```
    parser.add_argument("--chain.apply-deriv-weights", type=str,
                        dest='apply_deriv_weights', default=True,
                        action=common_lib.StrToBoolAction,
                        choices=["true", "false"],
                        help="")
```

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -33,9 +33,12 @@ def GetArgs():
                         help='Number of overlapping background noises that we iterate over. For example, if the input is "1:2:3" then the output wavs will have either 1, 2, or 3 randomly chosen background noises overlapping the entire recording')
     parser.add_argument('--fg-interval', type=int, dest = "fg_interval", default = 0,
                         help='Number of seconds between the end of one foreground noise and the beginning of the next.')
-    parser.add_argument('--utt-suffix', type=str, dest = "utt_suffix", default = "aug", help='Suffix added to utterance IDs.')
+    parser.add_argument('--utt-suffix', type=str, dest = "utt_suffix", default = None, help='Suffix added to utterance IDs.')
+    parser.add_argument('--utt-prefix', type=str, dest = "utt_prefix", default = None, help='Prefix added to utterance IDs.')
     parser.add_argument('--random-seed', type=int, dest = "random_seed", default = 123, help='Random seed.')
 
+    parser.add_argument('--modify-spkr-id', type=str, default = "false", choices=["true", "false"], help='Utt prefix or suffix would be added to the speaker id also (Used in ASR), in speaker id it is left unmodifed' )
```
danpovey: let's call this --modify-spk-id .. more consistent with spk2utt, etc.

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -176,21 +212,37 @@ def main():
...
+
+    CopyFileIfExists(input_dir + "/utt2lang", output_dir + "/utt2lang", args.utt_modifier_type, args.utt_modifier)
+    CopyFileIfExists(input_dir + "/utt2num_frames", output_dir + "/utt2num_frames", args.utt_modifier_type, args.utt_modifier)
+    CopyFileIfExists(input_dir + "/text", output_dir + "/text", args.utt_modifier_type, args.utt_modifier)
+    CopyFileIfExists(input_dir + "/segments", output_dir + "/segments", args.utt_modifier_type, args.utt_modifier, fields=[0, 1])
+    CopyFileIfExists(input_dir + "/vad.scp", output_dir + "/vad.scp", args.utt_modifier_type, args.utt_modifier)
+    CopyFileIfExists(input_dir + "/reco2file_and_channel", output_dir + "/reco2file_and_channel", args.utt_modifier_type, args.utt_modifier, fields=[0, 1])
+
+    if args.modify_spkr_id:
+        CopyFileIfExists(input_dir + "/spk2gender", output_dir + "/spk2gender", args.utt_modifier_type, args.utt_modifier)
```
danpovey: please try to mostly keep within 80 or 100 characters by breaking long lines.

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -117,16 +134,35 @@ def AugmentWav(utt, wav, dur, fg_snr_opts, bg_snr_opts, fg_noise_utts, \
             + start_times_str + " " + snrs_str + " - - |"
     return new_wav
 
-def CopyFileIfExists(utt_suffix, filename, input_dir, output_dir):
-    if os.path.isfile(input_dir + "/" + filename):
-        dict = ParseFileToDict(input_dir + "/" + filename,
+# This function generates a new id from the input id
```
danpovey: It's good that you are adding documentation, but let's do it with a doc-string, e.g.

```
def Foo(string):
     """ Foo is a function that does nothing.
               'string' is expected to be of type str, but is otherwise ignored.
          This function returns None.
     """
      return None
```

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
danpovey: Also can you please add an arg `use_ivectors=true` that wil give the option to skip all the i-vector-related parts of the script?  We may want to use this option later on when we switch to online CMN, but for now I guess we can leave that issue separate, I suppose there is no need to mix it up with this.

danpovey:
@david-ryan-snyder or @phanisankar-nidadavolu, perhaps you can comment?  And @phanisankar-nidadavolu, let me know if I should be merging this now.

phanisankar-nidadavolu:
Hello Dan,

I think it is a good observation. musan corpus is originally at 16KHz. When
we use this script  (musan.py) for augmenting 8k data (like swbd, sre16) we
downsample musan corpus to 8kHZ (like in the below line)

utt2wav_str = utt2wav_str + utt + " sox -t wav " + utt2wav[utt] + " -r 8k
-t wav - |\n"

if the corpus is at 16k (like SITW) this is not required. Ideal setup would
be to make sampling frequency optional and use same musan.py everywhere.
But since musan scripts are located in local directories, it is safe enough
to assume the sampling frequency is known in advance (you know whether to
resample it or not). Let me know what you think.

Regarding merging I will push my latest changes tonight and we can merge.


Phani

danpovey:
It sounds like a great idea.  Please, if you have time, either implement it
yourself or work with the person who raised the issue to do so if you feel
that is practical.

phanisankar-nidadavolu:
I will do it myself but where should I place it, steps/data or would it
still be in local directories?

Phani

david-ryan-snyder:
The make_musan.py script is just a data prep scrip in local. Since SWBD uses audio sampled at 8kHz we know what the sample rate should be for this script. I don't see why this needs to be an option.

danpovey:
Oh.  Since we will likely be relying on this in the data augmentation
scripts once we widely use them, I think it would make sense to put it
somewhere shared though, e.g. steps/augmentation/ ?

david-ryan-snyder:
If we consider MUSAN important enough to put into steps/ it makes sense to change the usage so that the sample rate is an option. If you go that route, you should be sure to change all of the recipes that use MUSAN (that will include all of the x-vector recipes) so that it uses the new script in steps/ and also be sure to remove the old script in local.

danpovey:
Agreed.

david-ryan-snyder:
Moving it to steps/augmentation or steps/data sounds like a good idea to me.

danpovey:
I prefer steps/augmentation/, as it's not a *generic* script for
manipulating data dirs.

danpovey:
Oh, I see, steps/data/ already has augmentation stuff.  OK, put it there.

phanisankar-nidadavolu:
I will do that.

Phani

david-ryan-snyder:
@phanisankar-nidadavolu, is this PR done and ready to be reviewed?

phanisankar-nidadavolu:
Yes it is. I ran a test run and the script has successfully reached TDNN
training stage.


Phani

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
```diff
@@ -0,0 +1,225 @@
...
+    data/${train_set} data/${train_set}_music
+
+  # Augment with musan_speech
+  steps/data/augment_data_dir.py --utt-prefix "babble" --modify-spk-id "true" \
+    --bg-snrs "20:17:15:13" --num-bg-noises "3:4:5:6:7" \
+    --bg-noise-dir "data/musan_speech" \
+    data/${train_set} data/${train_set}_babble
+
+  # Combine all the augmentation dirs
+  # This part can be simplified once we know what noise types we will add
```
david-ryan-snyder: Is this comment still relevant?

egs/wsj/s5/steps/data/make_musan.sh
```diff
@@ -0,0 +1,63 @@
...
+stage=0
+
+echo "$0 $@"  # Print the command line for logging
+
+if [ -f path.sh ]; then . ./path.sh; fi
+. parse_options.sh || exit 1;
+
+if [ $# -ne 2 ]; then
+    echo USAGE: $0 input_dir output_dir
+    echo input_dir is the path where the original musal corpus is located
```
david-ryan-snyder: "original musal" -> "MUSAN"

egs/wsj/s5/steps/data/make_musan.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# Copyright 2015   David Snyder
+#           2019   Phani Sankar Nidadavolu
+# Apache 2.0.
+#
+# This script creates the MUSAN data directory.
+# Consists of babble, music and noise files.
+# Used to create augmented data
+# The required dataset is freely available at http://www.openslr.org/17/
```
david-ryan-snyder: Could you add a comment that says something like:

`# The corpus can be cited as follows:
#
# @misc{musan2015,
#  author = {David Snyder and Guoguo Chen and Daniel Povey},
#  title = {{MUSAN}: {A} {M}usic, {S}peech, and {N}oise {C}orpus},
#  year = {2015},
#  eprint = {1510.08484},
#  note = {arXiv:1510.08484v1}
# }`

egs/swbd/s5c/local/chain/multi_style/run_tdnn_1a.sh
```diff
@@ -0,0 +1,256 @@
+#!/bin/bash
```
david-ryan-snyder: I think this script should go in into a new tuning directory, e.g., local/chain/multi_style/tuning/run_tdnn_1a.sh.

Then, in local/chain/multi_style, I suggest creating a symbolic link to run_tdnn_1a.sh called run_tdnn.sh. That would be analogous to what is done in other recipes, such as https://github.com/kaldi-asr/kaldi/tree/master/egs/swbd/s5c/local/chain .

Also, I suggest adding a more detailed comment at the top of this script that also includes some WER results. For example, see https://github.com/kaldi-asr/kaldi/blob/master/egs/swbd/s5c/local/chain/tuning/run_tdnn_lstm_1n.sh.

danpovey:
Thanks David for the detailed review!

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
phanisankar-nidadavolu: I think it is still relevant for now. We have not arrived at a general conclusion yet.

egs/swbd/s5c/local/chain/multi_style/run_tdnn_1a.sh
phanisankar-nidadavolu: Makes sense but instead of creating the sym link in local/chain/multi_style it is better to create in local/chain/run_tdnn_multistyle.sh, since all the other tend scripts are located there. What do you think?

danpovey:
Yes, maybe having the symlink point from local/chain/ makes sense.

egs/swbd/s5c/local/chain/multi_style/run_tdnn_1a.sh
david-ryan-snyder: That sounds better.

phanisankar-nidadavolu: I made the changes except for the inclusion of WERs. I will include them once the tdnn finishes training. In my old experiments I never decoded on train_dev.

david-ryan-snyder:
I have two more comments on the aesthetics of the PR while you're waiting for the WERs. 

I don't think we need the multi_style directory. This doesn't matter much for this particular PR, but as we start adding more scripts that use augmentation, deciding where these scripts will go will become more important. As I understand it, we plan on augmentation becoming part of the standard ASR recipe. I think it makes more sense to put theses scripts in the same place people are used to finding them, which is in chain/tuning. 
My suggestion is to move the new script to local/chain/tuning/run_tdnn_multistyle_1a.sh and make local/chain/run_tdnn_multistyle.sh be a symbolic link to that.

Also, in my opinion, it's better to refer to these recipes using the word "augmentation" rather than "multi style." In my opinion, using some form of the word "augment" makes it immediately clear what you're doing, and then the terminology would be consistent with the new directory steps/data/augmentation that you're adding. Also, "multi style" sounds old fashioned to me. What do you think?

david-ryan-snyder:
During the meeting today, @danpovey suggested that we refer to the recipes as "aug" instead of "multi style" or "augmentation." Seems nice and concise, and everyone show know what that means at first glance, I think.

vimalmanohar:
I think it is better to call it multi_condition or something to be consistent with existing recipes (in Aspire, AMI and Babel and probably many others) that already use that name for the same purpose. The scripts should go to local/multi_condition or local/chain/multi_condition/.

david-ryan-snyder:
@vimalmanohar, my objection to this is because augmentation is now going to become a standard part of all or most of our ASR recipes. I think it's going to be cumbersome to propagate this terminology into the names of scripts for future recipes. If we need a way to distinguish between the past recipes without augmentation, and new recipes that use augmentation, adding the suffix "aug" to the new scripts is a concise (and I think, more intuitive) way to indicate what the change is.

If we have to go with multi_*, I agree multi_condition is better than multi_style since it already exists.

danpovey:
I think aug / augmentation is a better name because eventually we will
likely include other augmentation methods in this setup.  and
multi_condition is a mouthful when used in directory names.

vimalmanohar:
I guess it makes sense then to use augmentation if all the new recipes have it. But perhaps it is better to have the new recipes in local/chain/augmentation/tuning and then create a symlink in local/chain or local/chain/augmentation.

danpovey:
I'd rather not add a new thing in the hierarchy if this is going to become
the default.  The idea is to have this be the normal recipe in
local/chain/run_tdnn.sh.  It might have to call a different
run_ivector_common.sh script though.

egs/wsj/s5/steps/data/augment_data_dir.py
```diff
@@ -117,23 +149,66 @@ def AugmentWav(utt, wav, dur, fg_snr_opts, bg_snr_opts, fg_noise_utts, \
...
+    if os.path.exists(output_dir):
+        data_lib.RunKaldiCommand("utils/fix_data_dir.sh {output_dir}".format(
+                                        output_dir = output_dir))
+        print("directory {} already exists, not creating it again".format(output_dir))
+        sys.exit()
+
+    fg_snrs = [int(i) for i in args.fg_snr_str.split(":")]
+    bg_snrs = [int(i) for i in args.bg_snr_str.split(":")]
     num_bg_noises = [int(i) for i in args.num_bg_noises.split(":")]
     reco2dur = ParseFileToDict(input_dir + "/reco2dur",
```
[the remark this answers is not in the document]
phanisankar-nidadavolu: Hi Vimal, the wrapper script prepare_multistyle_data.sh (now changed to run_ivector_aug.sh) first creates reco2dur file in the input directory and then we call this script to augment. I guess it is safe to assume that this file exists (at least for this recipe). Do you want me to modify the script to first check whether the file exists or not, and create the file if it does not exist?

egs/swbd/s5c/local/nnet3/prepare_multistyle_data.sh
```diff
@@ -0,0 +1,225 @@
+#!/bin/bash
+# Copyright 2019   Phani Sankar Nidadavolu
+# Apache 2.0.
+
+. ./cmd.sh
+
+set -e
+stage=0
```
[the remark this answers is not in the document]
phanisankar-nidadavolu: We have made use_ivectors an optional argument. I am not sure run_ivector_common.sh is a valid name in that case. We either have to remove use_ivectors flag and make training of ivectors a default option or name the script prepare_aug_data.sh. Let me know what you think.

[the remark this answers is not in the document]
danpovey: Hm, maybe prepare_aug_data.sh is OK then.  We may not end up using ivectors in all cases.  I don't have super strong opinoins about that.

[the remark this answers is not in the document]
danpovey: Maybe run_aug_common.sh?

egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -84,9 +84,6 @@ def GetArgs():
     return args
 
 def CheckArgs(args):
-    if not os.path.exists(args.output_dir):
```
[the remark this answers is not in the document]
phanisankar-nidadavolu: @vimalmanohar I don't see why. This part of the code is only creating output dir if it does not  exist. Anyways, I am modifying the code to fail if the output directory already exists.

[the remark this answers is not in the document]
danpovey: I don't think Vimal was asking for it to fail if the output dir exists (and I don't recommend it either).
I think he's pointing out that since, below, you do os.makedirs with no such guard, it will fail if it did not
exist.  It seems to have an `exist_ok` parameter which you can set to true.

phanisankar-nidadavolu:
Hmm. Got it.
Phani

phanisankar-nidadavolu:
run_aug_common.sh is good.

Phani

egs/wsj/s5/steps/data/reverberate_data_dir.py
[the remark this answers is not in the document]
vimalmanohar: The original version of the script will re-create the augmented data dir if the script is run again. But Phani modified it so it would do nothing if run again, which I think is unexpected behavior. It should either 1) re-create the augmented data dir, 2) fail with error saying it already exists.

egs/wsj/s5/steps/data/augment_data_dir.py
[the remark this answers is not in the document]
vimalmanohar: You can add it do this script. utils/data/get_reco2dur.sh already will check if file exists and create it only if not. So you can simply call that script.

phanisankar-nidadavolu:
Hello, the PR looks good with changes that you suggested. Main changes made: renaming to aug, moving around files, adding results and some other stuff.

PR #3125, opened 2019-03-17
Cleanup split_scp.pl

xsawyerx:
Hi,

I cleaned up this script because we were trying to use it at work and it was hard to follow.

[Each commit here is self-contained, so it is easier to review it per commit than the entire diff.]

While I added a lot of whitespaces, I did make several more significant changes:

* All variables are properly declared. This prevents typos, overwriting of data, etc. that is otherwise completely undetected. (And also cleaning up memory.) Then I was able to add `strict.pm`, which is a required pragma nowadays when developing in Perl.
* Fixed insecure filehandle usage: The code made it possible to overwrite files, exhaust resources, etc. I fixed that.
* Removing bareword filehandles: This is a general bad pattern of using globally-accessible non-cleanup'able filehandles. I changed all of those to proper variables, allowing them to be scoped, cleaned up properly, and more.
* Closed filehandles explicitly. Perl attempts to close filehandles implicitly but depending on Perl version, the errors are handled differently. The safest is to close it yourself.

### DIFF
--- a/egs/wsj/s5/utils/split_scp.pl
+++ b/egs/wsj/s5/utils/split_scp.pl
@@ -89,31 +89,40 @@ my ( %utt2spk, %spk_count, %spk_data, @scparray, @scpcount );
 if ( $utt2spk_file ) { # We have the --utt2spk option...
     open( my $utt_fh, "<", $utt2spk_file )
         || die "Failed to open utt2spk file $utt2spk_file";
+
     while (<$utt_fh>) {
         my @A = split;
         @A == 2 || die "Bad line $_ in utt2spk file $utt2spk_file";
         my ( $u, $s ) = @A;
         $utt2spk{$u} = $s;
     }
+
     close $utt_fh;
+
     open( my $in_fh, "<", $inscp ) || die "Opening input scp file $inscp";
+
     my @spkrs = ();
     while (<$in_fh>) {
         my @A = split;
         if ( @A == 0 ) { die "Empty or space-only line in scp file $inscp"; }
+
         my $u = $A[0];
         my $s = $utt2spk{$u};
+
         if ( !defined $s ) {
             die "No such utterance $u in utt2spk file $utt2spk_file";
         }
+
         if ( !defined $spk_count{$s} ) {
             push @spkrs, $s;
             $spk_count{$s} = 0;
             $spk_data{$s}  = []; # ref to new empty array.
         }
+
         $spk_count{$s}++;
         push @{ $spk_data{$s} }, $_;
     }
+
     close $in_fh;
 
     # Now split as equally as possible ..
@@ -126,9 +135,11 @@ if ( $utt2spk_file ) { # We have the --utt2spk option...
             "Refusing to split data because number of speakers $numspks is less "
             . "than the number of output .scp files $numscps";
     }
+
     for ( my $scpidx = 0; $scpidx < $numscps; $scpidx++ ) {
         $scparray[$scpidx] = []; # [] is array reference.
     }
+
     for ( my $spkidx = 0; $spkidx < $numspks; $spkidx++ ) {
         my $scpidx = int( ( $spkidx * $numscps ) / $numspks );
         my $spk = $spkrs[$spkidx];
@@ -152,11 +163,13 @@ if ( $utt2spk_file ) { # We have the --utt2spk option...
             # First try to reassign ending spk of this scp.
             if ( $scpidx < $numscps - 1 ) {
                 my $sz = @{ $scparray[$scpidx] };
+
                 if ( $sz > 0 ) {
                     my $spk   = $scparray[$scpidx]->[ $sz - 1 ];
                     my $count = $spk_count{$spk};
                     my $nutt1 = $scpcount[$scpidx];
                     my $nutt2 = $scpcount[ $scpidx + 1 ];
+
                     if (
                         abs( ( $nutt2 + $count ) - ( $nutt1 - $count ) )
                         < abs( $nutt2 - $nutt1 ) )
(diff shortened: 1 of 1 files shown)


danpovey:
that's good, but would you mind removing the whitespace just inside the ( and )?  I'd like it to look like the regular Kaldi coding style which is based on the Google style guide?  And I assume you tested this somehow?  LMK what kinds of things you did.

xsawyerx:
I don't mind changing the whitespaces between parenthesis. That's fine. (I can try to tweak perltidy to do this.)

As for testing: There don't seem to be any testing suite in place, so the code was currently tested by adding `strict.pm` and maintaining it, meaning I didn't introduce new variables that did not exist, or removed variables that were used.

Additionally, the changes I had made, other than the security fixes and the "strict-mode" are innocuous.

Today we are testing this with our pipeline and I will update this ticket once we do, to confirm it didn't break our usage.

xsawyerx:
After testing, we found a small thing to fix. Added a commit. I also added the whitespace changes you requested. (s/( /(/g and s/ )/)/g).

kkm000:
@danpovey, my pet peeve with this script has been the 0-based value of the -j option, so I end up with ugly stuff like  `"scp:utils/split_scp.pl -j $nj \$((JOB-1)) lats.scp |"`. It's too late to change the semantics of the existing switch, but maybe we can add say `-k`, just like `-j`, only 1-based? `-k $nj JOB` would have been much cleaner...

danpovey:
that makes sense, maybe we could call it -n.  need to document it very
clearly though.

jtrmal:
Wouldn't be easier just to add a switch '--one-based'?
Y.

jtrmal:
Oh, sorry, I realized you werent talking about making the numbers negative
but about a similar thing I was suggesting.
Y.

kkm000:
I third `--one-based`. I thought of `k` being one 1 more than `j` was easy to remember, but `--one-based` beats it with one hand! @xsawyerx, I would really appreciate if you could add this switch.

jtrmal:
I'm blushing :)

xsawyerx:
I don't mind implementing it. (Honestly, I would be happy to move it to using a proper arg parser, like Getopt::Long that comes in core Perl anyway.)

However, I do think that `--one-based` is a long flag that refers to implementation detail ("in your implementation, consider "-j" to start from 1") whereas `-n` is better at saying "This is the number of workers I would like for this thing."

I apologize for jumping in late. I just saw this now.

danpovey:
We agreed on --one-based already, so go ahead and implement that, it's not
important.

I suggest to just extend the existing code rather than use arg parsers, the
diff will be smaller and easier to check.

xsawyerx:
I disagree this is the right way to do it, but I'll accept the group decision. I'll submit a separate PR (I think that's cleaner) that will be based on this PR. (So this one should be merged if the other is to be merged cleanly.)

xsawyerx:
Submitted https://github.com/kaldi-asr/kaldi/pull/3162.

Meanwhile, I would like to create a branch with `Getopt::Long` to showcase how much better it would be to use a proper library to parse things.

I've tried different usages of the CLI to see if my code on that other PR worked, but found numerous inconsistencies.

```
# fails with "-j 2" at the beginning
$ perl split_scp.pl -j 2 in.scp out.scp out2.scp out3.scp
Argument "in.scp" isn't numeric in numeric lt (<) at split_scp.pl line 53.
Usage: split_scp.pl [--utt2spk=<utt2spk_file>] in.scp out1.scp out2.scp ...
 or: split_scp.pl -j num-jobs job-id [--utt2spk=<utt2spk_file>] in.scp [out.scp]
 ... where 0 <= job-id < num-jobs. at split_scp.pl line 72.

# works with "-j 2" at the end ("works" here means "ignored")
$ perl split_scp.pl in.scp out.scp out2.scp out3.scp -j 2
```

Additional weird stuff:
```
# mistake in input (not clear how it should be)
$ perl split_scp.pl -j 2 3 in.scp out.scp out2.scp out3.scp -j 2
Invalid num-jobs and job-id: 2 and 3 at split_scp.pl line 54.

# same
$ perl split_scp.pl -j 2 2 in.scp out.scp out2.scp out3.scp
Invalid num-jobs and job-id: 2 and 2 at split_scp.pl line 54.

# now it's a usage error because the parsing somehow failed
# (instead of the above examples in which we detect the user mistake)
$ perl split_scp.pl -j 2 1 in.scp out.scp out2.scp out3.scp
Usage: split_scp.pl [--utt2spk=<utt2spk_file>] in.scp out1.scp out2.scp ...
 or: split_scp.pl -j num-jobs job-id [--utt2spk=<utt2spk_file>] in.scp [out.scp]
 ... where 0 <= job-id < num-jobs. at split_scp.pl line 72.
```

danpovey:
OK, I'll clean up this PR.  @xsawyerx you have to bear in mind that these scripts are very old.  For more recent perl script we have used `getopt`, but most scripts we write from scratch these days are in Python.  We can't change the interface of these scripts because a vast number of scripts use them.  Our main concern here is to not break anything (and to not spend too much time verifying that your changes won't break anything, hence the desire to keep your changes minimal).

danpovey:
i mean close this PR.

xsawyerx:
I understand your need. I did not suggest any changes to the interface whatsoever. What I'm suggesting here is to edit the implementation of the parsing to make it work correctly.

The pastes I provide above prove that if you just move arguments around, we easily confuse the parsing code. This code, in a sense, is inconsistent and can provide a less-than-optimal experience for the user.

I have quite a bit of experience with Perl and I'm willing to whip your scripts into shape. None of this involves changing their interfaces. What it does include is fixing the bugs in them and making them do what we think they should, as well as making them testable and more readable, as well as maintainable.

I'm willing to provide the results in small steps and help review all of them, to ease their integration.

None of this is to take away from your value in this: It works and it should continue to work with the same interface. I agree with it fully.

PR #3127, opened 2019-03-17
Merged 'master' branch

desh2608:


danpovey:
@desh2608  could you please make this PR under another name with the same source branch?
To keep the history clean (and ensure we merge only with actual merge) I used `git push -f` to remove this merge from the history, but it has confused GitHub.

PR #3136, opened 2019-03-18
Spanish Gigaword text based POCOLM and RNNLM training recipe

saikiranvalluri:
We introduce the following features into the existing fisher_spanish recipe:
- Text processing (optional) scripts for Spanish Gigaword text corpus.
- Train a 3-gram POCOLM using the Fisher train and gigaword texts.
- POCOLM wordlist is derived using relative frequency of words in each corpus and metaparameter weights of each text corpus.
- OOVs of the POCOLM wordlist are added to ASR lexicon and RNNLM wordlist using seq2seq transformer based G2P model - https://github.com/cmusphinx/g2p-seq2seq
- RNNLM (optional) is trained, using the two text corpora as train sets and Fisher dev2 partition as dev set, for 5 epochs.
- During the test sets decode stage, after training the chain model, the extended ASR lexicon above is used to derive the graph. And RNNLM rescoring is also performed using the trained Gigaword RNNLM.

We achieved WER 20.84% WER on the Fisher Spanish test partition and 24.67% WER on Fisher dev partition, using the Gigaword text-based trained RNNLM rescoring, over the baseline 3-gram LM based decoded lattices.


danpovey:
s there a reason it doesn't make sense to just replace the current example with this?
I doubt too many people were using the old example.

Are you using a graphemic or phonemic lexicon?  A graphemic lexicon might be a reasonable choice in Spanish, for simplification.

saikiranvalluri:
> s there a reason it doesn't make sense to just replace the current example with this?
> I doubt too many people were using the old example.
> 
> Are you using a graphemic or phonemic lexicon? A graphemic lexicon might be a reasonable choice in Spanish, for simplification.

I am using the same Callhome Spanish rules based lexicon , which is same as the one used in the original recipe.

danpovey:
I think I would like you to extend the existing recipe rather than starting a new one.  Can you do that?

saikiranvalluri:
> I think I would like you to extend the existing recipe rather than starting a new one. Can you do that?

Sure sir.
Before that let me run the script e2e and make sure we get a better WER using the gigaword RNNLM,  than the original s5 recipe even with an extended vocabulary.

kkm000:
@saikiranvalluri, where we are on this?

PR #3141, opened 2019-03-19
[egs] Sample hold-out set fairly

kkm000:
The way the code in steps/nnet3/chain/get_egs.sh is currently written, the number of hold-out utterances in the validation set is unpredictable, and ends up much greater than requested.

In fact, it sampled away nearly 1/3 of my wonderfully 27-way perturbed toy training set... Caught the thief!

Also added `$0:` to all echo diagnostics, and squished two `wait`s that did not wait for anything.

### DIFF
--- a/egs/wsj/s5/steps/nnet3/chain/get_egs.sh
+++ b/egs/wsj/s5/steps/nnet3/chain/get_egs.sh
@@ -157,8 +157,8 @@ if [ -f $data/utt2uniq ]; then
        "includes all perturbed versions of the same source utterance."
   utils/utt2spk_to_spk2utt.pl $data/utt2uniq 2>/dev/null |
     awk -v max_utt=$num_utts_subset '{
-        for(n=1;n<=NF;n++) print $n;
-        printed += NF;
+        for (n=2;n<=NF;n++) print $n;
+        printed += NF-1;
         if (printed >= max_utt) nextfile; }' |
     sort > $dir/valid_uttlist
 else
--- a/egs/wsj/s5/steps/nnet3/chain/get_egs.sh
+++ b/egs/wsj/s5/steps/nnet3/chain/get_egs.sh
@@ -151,36 +151,40 @@ mkdir -p $dir/log $dir/info
 # Get list of validation utterances.
 frame_shift=$(utils/data/get_frame_shift.sh $data) || exit 1
 
-awk '{print $1}' $data/utt2spk | \
-  utils/shuffle_list.pl 2>/dev/null | head -$num_utts_subset > $dir/valid_uttlist
-
-len_uttlist=$(wc -l < $dir/valid_uttlist)
-if [ $len_uttlist -lt $num_utts_subset ]; then
-  echo "Number of utterances is very small. Please check your data." && exit 1;
+if [ -f $data/utt2uniq ]; then
+  # Must hold out all augmented versions of the same utterance.
+  echo "$0: File $data/utt2uniq exists, so ensuring the hold-out set" \
+       "includes all perturbed versions of the same source utterance."
+  utils/utt2spk_to_spk2utt.pl $data/utt2uniq 2>/dev/null |
+    awk -v max_utt=$num_utts_subset '{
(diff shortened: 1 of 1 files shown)


danpovey:
@kkm000 I am slightly worried about the variance of the diagnostic output being increased unduly, especially if the dataset is large (due to taking all copies of utterances).  The validation sets aren't that large; sometimes we get situations where the valid prob is better than the train prob.

kkm000:
@danpovey It's reverse, actually; I'm *fixing* the problem you mention, if I understand what you are saying. The current code selects N random (perturbed) utterances, and then closes this set over the utt2uniq equivalence. So if N is negligibly small compared to the total size of the set, and the set is K-way perturbed, this will multiply the size of the hold-out set by approximately K (assuming that randomly selected utterances belong to different uniques each, it's exactly K). In practice, this is less than K even for a sizable data set (the "birthday paradox"). So it's not right, we are asking so subset 300 by default, and ending up with somewhere between 300 and 900, and nobody knows how many.

In my case (I "perturbed" the data 9-way × 3-way speed-perturbed, K=27), the number selected by the current algorithm was over 6000. Not exactly the upper bound 8100, of course, but certainly too many, and this is a small, toy-size dataset to train in 30-40 minutes.

What I do here is arrange the uniq-2-utterance file, like

utt1 utt1-sp0.9 utt1-sp1.0 utt1-sp1.1
utt2 utt2-sp0.9 utt2-sp1.0 utt2-sp1.1

then add all utterance ids from each line (awk prints fields 2 to NF), and stop after printing at least the requested sample size. So I'm overshooting by O(1), no more than by K-1, and guarantee that the set is closed over unique equivalence (what you explained in kaldi-help). In my tests, I'm getting 324 utterance ids in the closed set for the K=27-way-perturbed set, and exactly 300 for 12-way set, so I've essentially tested that in both exactly-divisible and non-divisible edge cases.

kkm000:
The potentially breaking thing here is that the current egs rely on the abnormal behavior, in a sense. Only a few egs scripts override the `num_utts_subset` option to get_egs.sh, so in the standard speed-perturbed setup, they must be getting actually say ~800 hold-out utterances. Maybe it makes sense to bump the default from 300 to 800 with this change (and increase the number in the egs that do use the option)?

danpovey:
The existing script doesn't keep all the other copies of what you selected,
it just keeps the randomly chosen ones.  At least that was the intention.
So it's a "more random" held-out set than what your script does.   I have
to think about this.

kkm000:
No urgency at all, you seem immersed into the k-10 stuff. To make sure we're on the same page, I'll recapture here.

@danpovey [in kaldi-help](https://groups.google.com/d/msg/kaldi-help/i5aNz9VGYAM/BW9V3OnOAQAJ):
> [...] purpose [of utt2uniq] is to make sure that when you create a held-out set for diagnostics, you hold out all versions of those utterances; otherwise those diagnostics would not really affect how much overfitting was occurring. If you don't use utt2uniq, your validation diagnostics may seem better than they should really be, since the utts were seen in training (e.g. with different perturbation).

Both current and new algorithms do exactly this, only the new is also keeping the size of the held-out set  at or closely above the requested num_utts_subset, while the current selects the number of utts which is random and strictly in [num_utts_subset, (N in N-way perturbed)*num_utts_subset], likely closer to the upper bound. My thought is the "do what I say" behavior is better; the current behavior is likely [DWIM](http://catb.org/~esr/jargon/html/D/DWIM.html): the thing held out over 1/4 of my dataset; that was quite a surprise.

True, my setup is unusual, but people do all the weird stuff with their data. Data science should have been called data witchcraft; it's not science at all.

danpovey:
OK, I'll merge.

kkm000:
An what do you think about the default? It's now 300, while the actually selected number is like 800 for a 3-way perturbed set.

danpovey:
the default is OK.  If too large, it can make compute_prob_valid slow.

kkm000:
Sure, it's probably not a very important parameter to keep on tuning. Thanks!

PR #3145, opened 2019-03-19
cudamatrix: fixing a synchronization bug in 'normalize-per-row'

vesis84:
- we found it through 'NormalizeComponent' from nnet3 than misbehaved
  (unit tests were passing, because of using too small matrices for
  which cuda thread synchronization is not necessary),
- we created a new unit-test which detects the error we had,

- the bug manifested itself with this error message :

ASSERTION_FAILED
(nnet3-chain-train[5.5.230~1-bcfe3]:HouseBackward():qr.cc:124) Assertion
failed: (KALDI_ISFINITE(sigma) && "Tridiagonalizing matrix that is too
large or has NaNs.")

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -2552,25 +2552,29 @@ static void _normalize_per_row(Real *y, int y_stride, const Real *x,
   const int i = blockIdx.x;
   const int tid = threadIdx.x;
   const Real* x_row = x + i * x_d.stride;
+
   typedef cub::BlockReduce<Real, CU1DBLOCK> BlockReduceT;
   __shared__ typename BlockReduceT::TempStorage temp_storage;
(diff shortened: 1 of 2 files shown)


danpovey:
Thanks!  Do you think this is a new bug that was introduced when we switched to cub-based synchronization?

vesis84:
Yes, this was introduced with the 'cub'. 

It seems appropriate to use '__shared__' variable to communicate the partial results among threads within a block. And without the syncs there were some strange values in the output (0's, inf's, nan's). I could not find in the docs if 'cub' is doing the syncs by itself, or if a 'cub' call should be guarded by 2 syncs, but the 2nd option is safer...

danpovey:
Thanks.  I had a look at the other code using cub reduce, and I don't think you need the syncthreads before it.  As far as I can see, cub does any syncing it needs, and the syncing is only required to make sure the shared variable is broadcast everywhere (hence: after the call).

PR #3158, opened 2019-03-21
Adding constructors with arguments to decoder classes.

RuABraun:
This is a very small PR which adds constructors with arguments to the standard classes used for decoding.

I know the kaldi PR philosophy seems to be if it's not used in any recipe to not merge it, but this is something basically everybody who uses kaldi for in some sort of production setting (where your decode settings are fixed and you don't want to pass them from the command line all the time) has to add. I thought it might make sense to add this to the master branch...

If you want I can also add a new binary `online2-nnet3-chain-decode` with reasonable defaults with which one can go straight from a WAV to a ctm, while only inputting a model folder path. So the help string would roughly look like
```Does online decoding on wavfile.
<model-dir> requires final.mdl, HCLG.fst, words.txt, align_lexicon.int and mfcc.conf
Usage: nnet3-decode <model-dir> <wav-path> <out-ctm>
```

Just so the new constructors are used somewhere...

### DIFF
--- a/src/decoder/lattice-faster-decoder.h
+++ b/src/decoder/lattice-faster-decoder.h
@@ -62,6 +62,16 @@ struct LatticeFasterDecoderConfig {
                                 beam_delta(0.5),
                                 hash_ratio(2.0),
                                 prune_scale(0.1) { }
+
+  LatticeFasterDecoderConfig(BaseFloat beam, BaseFloat latbeam,
+      int32 max_active=std::numeric_limits<int32>::max(), int32 min_active=200,
+      int32 prune_interval=25, bool determinize_lattice=true, BaseFloat beam_delta=0.5,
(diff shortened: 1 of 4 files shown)


danpovey:
Sorry... thanks for the PR, but I don't want to merge it because it would make maintenance more complicated (i.e. we would need to make sure any new args were initialized in both constructors).
Actually the reason for these types of config classes is, following the Google style guide, to *avoid* having tons of functions with default arguments where one has to remember the position of everything.  For example, seeing `config.beam = 10.0;` is clear, while passing 10 as the 3rd arg of a constructor is less clear (even if more compact).  That's the rationale.

The example program you mention is an interesting idea, there are definitely people who would find this interesting (although I'm not sure that I want to nudge people in the direction of re-loading the model and graph each time they want to decode something).  If other people advocate for it, I'd consider it though.  Let's see.

RuABraun:
Okay I understand..

PR #3165, opened 2019-03-24
[src] Modified mel computations (cleaned up version)

danpovey:
to be tested.  Would be activated by setting
`modified=true`
in conf/mfcc_hires.conf

### DIFF
--- a/src/feat/mel-computations.cc
+++ b/src/feat/mel-computations.cc
@@ -34,78 +34,57 @@ MelBanks::MelBanks(const MelBanksOptions &opts,
                    const FrameExtractionOptions &frame_opts,
                    BaseFloat vtln_warp_factor):
     htk_mode_(opts.htk_mode) {
+  SetConfigs(opts, frame_opts, vtln_warp_factor);
+
   int32 num_bins = opts.num_bins;
   if (num_bins < 3) KALDI_ERR << "Must have at least 3 mel bins";
-  BaseFloat sample_freq = frame_opts.samp_freq;
-  int32 window_length_padded = frame_opts.PaddedWindowSize();
-  KALDI_ASSERT(window_length_padded % 2 == 0);
-  int32 num_fft_bins = window_length_padded / 2;
-  BaseFloat nyquist = 0.5 * sample_freq;
 
-  BaseFloat low_freq = opts.low_freq, high_freq;
-  if (opts.high_freq > 0.0)
-    high_freq = opts.high_freq;
-  else
-    high_freq = nyquist + opts.high_freq;
 
-  if (low_freq < 0.0 || low_freq >= nyquist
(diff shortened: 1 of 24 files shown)


RuABraun:
This is supposed to give a small but consistent performance improvement correct?

danpovey:
Yes.  I'd say around 0.2 to 0.3% absolute on average (at least when the WER
is in the 20 to 30% range).  It needs some more testing before I'm ready to
merge it.  Of course if you want to help, that's great.

danpovey:
And by testing, I am talking about the effect on WER, not just does it run.

RuABraun:
I could do a build on my own data (so you'd have to take my word for the results), if that works for you?

danpovey:
Yes, that works for me.

src/feat/mel-computations.cc
```diff
@@ -129,29 +108,93 @@ MelBanks::MelBanks(const MelBanksOptions &opts,
...
+ */
+void MelBanks::ComputeModifiedBins() {
+  int32 num_bins = center_freqs_.Dim();
+  for (int32 bin = 0; bin < num_bins; bin++) {
+    BaseFloat center_freq = center_freqs_(bin),
+        next_center = (bin == num_bins - 1 ?
+                       high_freq_ : center_freqs_(bin + 1));
+
+    BaseFloat d1 = (next_center - center_freq) * 1.1,
+              d2 = 60.0 + 50.0 * (center_freq / (center_freq + 700.0));
```
pegahgh: It should be breakpoint_ instead of 700.

danpovey: I was thinking of making this a separate issue from the breakpoint, so that when we tune it it doesn't have to have the same value. Can you compare 700 with 300 (our current breakpoitn) in your new experiments?  If 300 works better we can reconsider.

RuABraun:
Sorry about the delay on this:

Some info about the model: 100h with 13 layer tdnnf. No ivector. Decoding with a 1M ngram language model (no rescoring).

Because of how my data is acquired, I cannot always guarantee that there is no overlap in speakers between training/test, these are the test sets A, B, C. Some test sets I can guarantee it, those are X, Y, Z

|          	| X     	| Y     	| Z     	| A     	| B     	| C     	|
|----------	|-------	|-------	|-------	|-------	|-------	|-------	|
| baseline 	| 18.01 	| 17.57 	| 29.12 	| 19.57 	| 12.03 	| 23.82 	|
| modified 	| 18.70 	| 17.85 	| 29.56 	| 19.48 	| 12.02 	| 23.59 	|

As can see from the results it seems the modified version overfits more.

danpovey:
Great, thanks.  It turns out there may have been a problem with that particular code (i.e. there was a mismatch to what it should have been), although I don't know how important that actually was.
Hopefully at some point in the future I will have a fixed version.

PR #3170, opened 2019-03-26
Kaldi10: Add missing file (from PR #3169) + minor fixes 

hhadian:
Another note: In L206 of topology.cc, `Connect` is non-const so `Check()` is non-const. I can instead call `Connect` on a copy of each entry if you want (this is for checking that all the states in the  topology are on successful paths).


danpovey:
Doing Connect() on a copy is better.  Check() should rarely be called anyway.

danpovey:
... and you should be making sure that the num-states and num-arcs is identical after calling Connect().

hhadian:
Done.

PR #3175, opened 2019-03-27
Shuffle utt2uniq list before getting validation set

desh2608:
This is to fix an issue which was created in https://github.com/kaldi-asr/kaldi/pull/3141 where only the top utterances were selected while creating the validation set, if an utt2uniq file exists.


kkm000:
Thank you! I just noticed the flop and was going to send in a fix.

PR #3185, opened 2019-03-29
[WIP] added out-of-place functions

YiwenShaoStephen:


src/cudamatrix/cu-kernels.cu
```diff
@@ -2484,6 +2484,42 @@ static void _heaviside(Real*y, const Real*x, MatrixDim d, int src_stride) {
...
+}
+
+template<typename Real>
+__global__
+static void _pow(Real*y, const Real*x, Real power, MatrixDim d, int src_stride) {
+  int i = blockIdx.x * blockDim.x + threadIdx.x;
+  int j = blockIdx.y * blockDim.y + threadIdx.y;
+  int dst_index = i + j * d.stride, src_index = i + j * src_stride;
+  if (i < d.cols && j < d.rows) {
+    Real res;
```
danpovey: This function can be substantially simplified (I know this is the same as the previous one).
Just do:
```
y[dst_index] = pow(x[src_index], power);
```
Also please remove the existing _apply_pow kernel and the code that calls it (can just have ApplyPow() be inlined
to call Pow()).
You can do the same for _pow_abs vs. _apply_pow_abs.

src/cudamatrix/cu-matrix.cc
```diff
@@ -2485,6 +2485,49 @@ void CuMatrixBase<Real>::Heaviside(const CuMatrixBase<Real> &src) {
   }
 }
 
+template<typename Real>
```
danpovey: BTW, if you are brave, you could consider making some kind of multi-line macro defined
inside cu-matrix.cc that will do this, so you could just replace this implementation with
```
// define CuMatrixBase<Real>::Exp(const CuMatrixBase<Real> &src) 
KALDI_DEFINE_CUDA_UNARY_FUNCTION(Exp, cuda_exp);  
```

src/cudamatrix/cu-kernels.cu
YiwenShaoStephen: So when should we make conditions for power? Or is it not necessary any more?

danpovey:
There don't need to be any tests; this is for when someone knows the result
is valid.  They can call PowAbs() if they want to call abs() first.

src/cudamatrix/cu-kernels.cu
danpovey: not necessary any more.  It's up to the user to ensure it will give reasonable output.

danpovey:
@YiwenShaoStephen it looks like there is a conflict, could you please resolve?
@kangshiyin would you mind reviewing this?

YiwenShaoStephen:
@danpovey I've resolved the conflicts and make it pass all the tests.

src/matrix/kaldi-vector.h
```diff
@@ -119,6 +120,15 @@ class VectorBase {
   template<typename OtherReal>
   void CopyFromVec(const CuVectorBase<OtherReal> &v);
 
+  /// Applies floor to all elements. Returns number of elements
+  /// floored in floored_count if it is non-null.
```
danpovey: The only part of this PR that I am uncertain about is the floor/ceiling stuff.  I guess I'm concerned about the extra memory bandwidth from writing the same thing back to memory in case this and v point to the same data (v.Data() == this->data_).  Perhaps there could be an if/else inside the floor/ceiling functions to take care of that special case?  Also the documentation needs to be changed, it's not accurate now that the function changed.  Also (this was a pre-existing issue), the word "returns" is not appropriate there, it should be "outputs"

PR #3187, opened 2019-03-31
Add GitHub issue templates

kkm000:
Add two issue templates, with comments directing the user to the kaldi-help group if they have a Kaldi question.

I think it would be a better use for the tracker, and also that keeping Q/A in one easily searchable place would be be beneficial. I am seeing too many questions here, and always linger before closing them. Maybe this will keep us more manageable.

On the topic of manageability, I want to create project boards for codename Kaldi10, and for the current development and fixes. Is it ok?


danpovey:
OK.  The project board thing is a good idea too.

PR #3190, opened 2019-03-31
fix a bug in pruned lat-rescoring when input lattice has epsilons

hainan-xv:


src/lat/compose-lattice-pruned.cc
```diff
@@ -771,7 +771,15 @@ void PrunedCompactLatticeComposer::ProcessTransition(int32 src_composed_state,
   // Note: we expect that ilabel == olabel, since this is a CompactLattice, but this
   // may not be so if we extend this to work with Lattice.
   fst::StdArc lm_arc;
-  if (!det_fst_->GetArc(src_info->lm_state, olabel, &lm_arc)) {
+
+  // the input input lattice might nave epsilons
```
danpovey: typo here

src/lat/compose-lattice-pruned.cc
```diff
@@ -771,7 +771,15 @@ void PrunedCompactLatticeComposer::ProcessTransition(int32 src_composed_state,
...
+  if (olabel == 0) {
+    lm_arc.ilabel = 0;
+    lm_arc.olabel = 0;
+    lm_arc.nextstate = src_info->lm_state;
+    lm_arc.weight = fst::StdArc::Weight(0.0);
+  }
```
danpovey: should be `} else if ...`

PR #3195, opened 2019-04-01
minor bug (phones.txt instead of words.txt)

armusc:
as in the object
line 134 is modified, new phone disambiguation symbols in the grammar-fst lexicon update, are added  at the end of phones.txt not words.txt


kkm000:
LGTM. But I would not call it minor...

danpovey:
i modified the commit message in my squash merge

PR #3196, opened 2019-04-02
Move curand handle out of CuRand class and into CuDevice. 

luitjens:
This matches the behavior of other CUDA libraries like cublas, cufft, and cusparse.
This change is necessary to avoid redundant creation/destroying of the
random generator every-time someone creates a CuRand object.  In my measurements
I noticed that a single instantiation of CuRand<> could take up to 3 ms.

The CuRand object is now a thin wrapper around the generators and should
have very little overhead.

### DIFF
--- a/src/cudamatrix/cu-device.h
+++ b/src/cudamatrix/cu-device.h
@@ -26,6 +26,7 @@
 #if HAVE_CUDA == 1
 #include <cublas_v2.h>
 #include <cusparse.h>
+#include <curand.h>
 #include <map>
 #include <string>
 #include <iostream>
@@ -34,6 +35,7 @@
 #include "base/kaldi-common.h"
 #include "base/timer.h"
 #include "cudamatrix/cu-allocator.h"
+#include "cudamatrix/cu-common.h"
 
 namespace kaldi {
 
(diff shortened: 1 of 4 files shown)


danpovey:
Cool.  I assume you made sure the tests ran with GPUs?

luitjens:
I ran a single test on my system and it worked fine.  That isn't very
extensive.  Do you have internal tests that stress GPUs?

luitjens:
Pushed a new patch.  make test works in src with cuda enabled.

src/cudamatrix/cu-device.h
```diff
@@ -85,10 +85,12 @@ class CuDevice {
   inline curandGenerator_t GetCurandHandle() { return curand_handle_; }
 
   inline void SeedGpu() {
-		// To get same random sequence, call srand() before the method is invoked,
-		CURAND_SAFE_CALL(curandSetPseudoRandomGeneratorSeed(
-          curand_handle_, RandInt(128, RAND_MAX)));
-		CURAND_SAFE_CALL(curandSetGeneratorOffset(curand_handle_, 0));
+    if(CuDevice::Instantiate().Enabled()) {
```
danpovey: sorry very small comment, but space after 'if' please.

danpovey: actually its OK, I can change it from github directly.

danpovey:
will  merge once checks complete

luitjens:
thanks!

PR #3197, opened 2019-04-02
[WIP] cached_compiler I/O for nnet3-xvector-compute

stanleyguan:
This implements the idea brought up by @gorinars in [this thread](https://groups.google.com/forum/#!msg/kaldi-help/ROtSHHe3Z_I/smRJd7d9BgAJ). It enables more efficient x-vector extraction when the number of frames is known. This is most useful for situations where latency is important.

I don't expect this to be merged per discussion in the aforementioned thread. However, this may serve as an example of how it can be done, for those interested.


stanleyguan:
Is there a reason this was renamed with WIP? I intended the PR to be as is.

danpovey:
It looks harmless, so merging it.

PR #3199, opened 2019-04-03
Variational bayes based speaker diarization

saikiranvalluri:
After performing this recipe's VB resegmentation, over the Callhome Diarization recipe's baseline output rttm file (DER - 8.02%), we achieved DER = 6.21%.


egs/callhome_diarization/v1/local/dump_model.py
```diff
@@ -0,0 +1,127 @@
...
+    data_3dmatrix = []
+    data_matrix = []
+    data_array = []
+
+    for line in content:
+        line = line.strip('\n')
+        if line == "<SigmaInv> [":
+            break
+        if state == 0:
+            if line != "<M> 1024  [":
```
david-ryan-snyder: What is the 1024 referencing? Should this be configurable?

egs/callhome_diarization/v1/local/VB_resegmentation.py
```diff
@@ -0,0 +1,308 @@
...
+        duration_list = map(lambda x: '{0:.2f}'.format(x), duration_list)
+        print("PREDICTED {} SPEAKERS".format(len(duration_list)))
+        print("DISTRIBUTION {}".format(" ".join(duration_list)))
+        print("sp_out", sp_out)
+        print("L_out", L_out)
+
+        # Create the output rttm file and compute the DER after re-segmentation
+        create_rttm_output(utt, predicted_label, rttm_dir, args.channel)
+        if args.true_rttm_filename != "None":
+            cmd = "md-eval.pl -1 -c 0.25 -r {}/{}_true.rttm -s {}/{}_predict.rttm 2".format(temp_dir, utt, rttm_dir, utt)
```
david-ryan-snyder: In my opinion this script should produce an RTTM file. The evaluation of it should happen outside of this script, in the top-level run.sh.

egs/callhome_diarization/v1/local/VB_resegmentation.py
```diff
@@ -0,0 +1,308 @@
...
+    return 0
+
+def match_DER(string):
+    string_split = string.split('\n')
+    for line in string_split:
+        if "OVERALL SPEAKER DIARIZATION ERROR" in line:
+            return line
+    return 0
+
+def main():
```
david-ryan-snyder: This needs more documentation, along with a message describing it's usage, and some example arugments.

egs/callhome_diarization/v1/local/VB_resegmentation.py
```diff
@@ -0,0 +1,308 @@
+#!/usr/bin/env python
```
david-ryan-snyder: I think this script should probably go in diarization/ rather than local/

egs/callhome_diarization/v1/local/VB_resegmentation.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
```
david-ryan-snyder: This script should probably go in diarization/ rather than local/ because we'll want to be able to use it from other recipes as well.

egs/callhome_diarization/v1/local/dump_model.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
```
david-ryan-snyder: This probably also needs to go into diarization/

egs/callhome_diarization/v1/local/kaldi_io.py
```diff
@@ -0,0 +1,627 @@
+#!/usr/bin/env python
```
david-ryan-snyder: This is probably not the right place for this file. @HuangZiliAndy, do you know where this should go?

egs/callhome_diarization/v1/local/make_callhome.sh
```diff
@@ -70,4 +70,10 @@ utils/filter_scp.pl $data_dir/callhome1/wav.scp $data_dir/callhome/reco2num_spk
 utils/filter_scp.pl $data_dir/callhome2/wav.scp $data_dir/callhome/reco2num_spk \
   > $data_dir/callhome2/reco2num_spk
 
+# Here we tweak somethings to make VB resegmentation work for original data/callhome
+rm $$data_dir/callhome/segments
```
david-ryan-snyder: Extra $ sign

egs/callhome_diarization/v1/local/train_ivector_extractor_diag.sh
```diff
@@ -0,0 +1,166 @@
+#!/bin/bash
```
david-ryan-snyder: Why do we need this script? There should be a comment at the top describing that this is a copy of some original script, with various modifications for whatever reason. Also, this should probably go in diarization/ rather than local/ . @HuangZiliAndy can you also look at this please.

egs/callhome_diarization/v2/run.sh
```diff
@@ -53,7 +70,7 @@ if [ $stage -le 1 ]; then
   # callhome1 and callhome2.  Each partition is treated like a held-out
   # dataset, and used to estimate various quantities needed to perform
   # diarization on the other part (and vice versa).
-  for name in train callhome1 callhome2; do
+  for name in train callhome callhome1 callhome2; do
```
david-ryan-snyder: Is the dataset "callhome" just the combination of "callhome1" and "callhome2?" If so, it's wasteful to create features for the entire dataset, in addition to the two partitions. It would be better to create features for the two partitions and then use combine_data_dir.sh to combine them.

egs/callhome_diarization/v2/run.sh
```diff
@@ -92,7 +109,29 @@ if [ $stage -le 1 ]; then
...
-    data/sre_cmn data/sre_cmn_segmented
+				 data/sre_cmn data/sre_cmn_segmented
+
+  if [ $VB_resegmentation ]; then
+      # Prepare telephone and microphone speech from Mixer6.
+      local/make_mx6.sh $data_root/LDC2013S03 data/
+
+      # Prepare SRE10 test and enroll. Includes microphone interview speech.
+      # NOTE: This corpus is now available through the LDC as LDC2017S06.
+      local/make_sre10.pl $sre_root/SRE2010/eval/ data/
```
david-ryan-snyder: I don't understand why the VB resegmentation requires a separate set of datasets than the rest of the system. Why can't we train it on the same datasets used to train the x-vector DNN?

Unless there's a good reason, we should just use the initial datasets.

egs/callhome_diarization/v2/run.sh
```diff
@@ -356,3 +395,82 @@ if [ $stage -le 11 ]; then
...
+			       --true_rttm_filename "None" --initialize 1 \
+			       data/callhome $init_rttm_file $output_dir $output_dir/model/diag_ubm.pkl $output_dir/model/ie.pkl || exit 1;
+
+    # Compute the DER after VB resegmentation
+    cat $output_dir/rttm/* > $output_dir/predict.rttm
+    md-eval.pl -1 -c 0.25 -r $label_rttm_file -s $output_dir/predict.rttm 2> $output_dir/log/DER.log \
+	       > $output_dir/results/DER.txt
+    der=$(grep -oP 'DIARIZATION\ ERROR\ =\ \K[0-9]+([.][0-9]+)?' \
+	       $output_dir/results/DER.txt)
+    # After VB resegmentation, DER: 6.15%
```
david-ryan-snyder: It looks like you use the initial RTTM that uses the oracle number of speakers during clustering. We need clearly state that in the comments, and we also need to report results using the tuned stopping threshold.

src/ivectorbin/ivector-extractor-copy.cc
```diff
@@ -0,0 +1,64 @@
+// ivectorbin/ivector-extractor-init.cc
```
david-ryan-snyder: Out of date comment

src/ivectorbin/ivector-extractor-copy.cc
```diff
@@ -0,0 +1,64 @@
...
+
+// Copyright 2013  Daniel Povey
```
david-ryan-snyder: Update with your own name, organization, etc

src/ivectorbin/ivector-extractor-copy.cc
```diff
@@ -0,0 +1,64 @@
...
+#include "ivector/ivector-extractor.h"
+
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    using kaldi::int32;
+
+    const char *usage =
+        "Copy an ivector-extractor\n"
```
david-ryan-snyder: Maybe you can have a longer comment describing how this might be used. E.g., to copy the extractor to a text file.

egs/callhome_diarization/v1/local/kaldi_io.py
danpovey: What is this needed for?  I think we should either put it somewhere shared, or not use it at all.

egs/callhome_diarization/v2/run.sh
saikiranvalluri: > Added a few comments. Will add more later. For now, the main things are:
> 
> * General purpose scripts should go in diarization/, not local/. We might use these scripts for other recipes, e.g., DIHARD.
      ok I will move VB related scripts to a new folder - diarization.
> * I don't understand why the VB needs to be trained on different datasets than the ones we used to train the x-vector system. Seems to be complicating things too much.
      Yes, we used mx6 and sre10 data also in training ivector-extractor. So, shall we add these two sources in training x-vector system also?
> * Currently you're only reporting results using the oracle number of speakers. We definitely need to see results using a tuned stopping threshold, which is more realistic.
       Yes, with tuned stopping threshold, I get 8.21% DER before VB and 6.32% DER after VB.

egs/callhome_diarization/v1/local/dump_model.py
saikiranvalluri: I am not sure about this.

david-ryan-snyder: @saikiranvalluri, where'd this code come from originally?

saikiranvalluri: Nagendra sir provided me the scripts dump_model.py and kaldi_io.py

egs/callhome_diarization/v2/run.sh
saikiranvalluri: We used these recent versions of data - sre10 and mx6 in order to obtain the claimed results.

egs/callhome_diarization/v1/diarization/VB_resegmentation.sh
```diff
@@ -0,0 +1,103 @@
...
+ie_model=$5
+
+mkdir -p $output_dir/rttm
+
+sdata=$data_dir/split$nj;
+utils/split_data.sh $data_dir $nj || exit 1;
+
+if [ $stage -le 0 ]; then
+    $cmd JOB=1:$nj $output_dir/log/VB_resegmentation.JOB.log \
+      diarization/VB_resegmentation.py --true-rttm-filename $true_rttm_filename --max-speakers $max_speakers \
```
jprobichaud: `VB_resegmentation.py` doesn't support the `--true-rttm-filename` argument, are you committing the right versions?

david-ryan-snyder:
@HuangZiliAndy, @danpovey, I think this PR can be closed. If I am not mistaken, I believe @HuangZiliAndy merged his own PR that covers this functionality a few months ago.

PR #3200, opened 2019-04-03
Fix compilation of posterior.cc

galv:


galv:
@hhadian @danpovey FYI

src/hmm/transitions.h
```diff
@@ -170,6 +170,8 @@ class Transitions {
 
   inline int32 TransitionIdToPdfFast(int32 trans_id) const;
 
+  int32 TransitionIdToPhone(int32 trans_id) const;
```
hhadian: I don't think this function is needed because `TransitionIdInfo` is a struct and we can always use `InfoForTransitionId` to get any info re a transition.

galv: Right. I did not realize that struct was declared in the public portion of the Transitions class. I will make the change.

galv: Okay, it's done. Thank you for telling me that.

PR #3202, opened 2019-04-03
added extra disambiguous symbols into disambig.txt

jty016:
Or it may report error in validation phase


danpovey:
This is related to #3000 which was a partial fix for the same issue.

PR #3204, opened 2019-04-03
Tensorflow version update

langep:
Installing tensorflow for the tfrnnlm setup caused some issues for some users [https://groups.google.com/forum/#!searchin/kaldi-help/tfrnnlm%7Csort:date/kaldi-help/Zjjsq0dEiWo/oDNMDkeCAQAJ](https://groups.google.com/forum/#!searchin/kaldi-help/tfrnnlm%7Csort:date/kaldi-help/Zjjsq0dEiWo/oDNMDkeCAQAJ).

I verified that using a newer bazel version resolves the issue. Theoretically, the issue causing the problem described in the post linked above could be solved by using bazel 0.10 however tensorflow before version 1.7 has errors in the version checking code and does not recognize 0.10 as a newer bazel version than 0.5.4. 

Tensorflow should be backwards compatible according to their versioning system, so I decided to update the scripts to a newer version using:

tensorflow v1.12
bazel 0.15.0

I also corrected the `-rpath` option in the tfrnnlmbin Makefile to not rely on GNU specific syntax. However, there is still an issue compiling this on MacOS because of `-lrt` but this might be better suited for another PR.

### DIFF
--- a/src/tfrnnlmbin/Makefile
+++ b/src/tfrnnlmbin/Makefile
@@ -14,7 +14,8 @@ TENSORFLOW = $(shell pwd)/../../tools/tensorflow
 
 all:
 
-EXTRA_CXXFLAGS = -Wno-sign-compare -I$(TENSORFLOW)/bazel-tensorflow/external/protobuf/src \
+EXTRA_CXXFLAGS = -Wno-sign-compare \
+                 -I$(TENSORFLOW)/bazel-tensorflow/external/protobuf_archive/src
                  -I$(TENSORFLOW)/bazel-genfiles -I$(TENSORFLOW) \
                  -I$(TENSORFLOW)/tensorflow/contrib/makefile/downloads/eigen \
                  -I$(TENSORFLOW)/tensorflow/contrib/makefile/downloads/nsync/public \
@@ -31,7 +32,7 @@ TESTFILES =
 ADDLIBS = ../lat/kaldi-lat.a ../lm/kaldi-lm.a ../fstext/kaldi-fstext.a \
           ../hmm/kaldi-hmm.a ../tree/kaldi-tree.a ../util/kaldi-util.a \
           ../matrix/kaldi-matrix.a ../base/kaldi-base.a \
-          ../tfrnnlm/kaldi-tensorflow-rnnlm.a 
+          ../tfrnnlm/kaldi-tensorflow-rnnlm.a
 
 LDLIBS +=  -lz -ldl -fPIC -lrt
 LDLIBS += -L$(TENSORFLOW)/bazel-bin/tensorflow -ltensorflow_cc -ltensorflow_framework
--- a/src/tfrnnlmbin/Makefile
+++ b/src/tfrnnlmbin/Makefile
@@ -37,6 +37,6 @@ ADDLIBS = ../lat/kaldi-lat.a ../lm/kaldi-lm.a ../fstext/kaldi-fstext.a \
 LDLIBS +=  -lz -ldl -fPIC -lrt
 LDLIBS += -L$(TENSORFLOW)/bazel-bin/tensorflow -ltensorflow_cc -ltensorflow_framework
 
-LDFLAGS += -Wl,-rpath,$(shell pwd)/../../tools/tensorflow/bazel-bin/tensorflow/
+LDFLAGS += -Wl,-rpath,$(TENSORFLOW)/bazel-bin/tensorflow/
 
 include ../makefiles/default_rules.mk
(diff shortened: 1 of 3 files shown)


danpovey:
@hainan-xv can you please check this?

tools/extras/install_tensorflow_cc.sh
```diff
@@ -25,17 +25,17 @@ else
 fi
 
 
-[ ! -f bazel.zip ] && wget https://github.com/bazelbuild/bazel/releases/download/0.5.4/bazel-0.5.4-dist.zip -O bazel.zip
+[ ! -f bazel.zip ] && wget https://github.com/bazelbuild/bazel/releases/download/0.15.0/bazel-0.15.0-dist.zip -O bazel.zip
```
kkm000: I have rather a general question about this file. bazel is widely available on many distros package feeds. My (Ubuntu 18.04) offers me v0.24.1, this installs v0.15.1. And it depends on jdk, whoch I do not have (it's also packaged, but I would need it only to build bazel to build tensorflow, if I am not misunderstanding? Maybe we should check if (a) bazel is installed on the system and, if not, (b) mention that it might be available to install? Certainly in the section that complain about a missing jdk. Here we kind of unconditionally possibly too much unnecessary stuff that is likely available otherwise. I may well be mistaken, just from a superficial look at it.

langep: I initially tried to use the newest bazel version but tensorflow complained that the version was too new. It would be possible to check if the correct bazel version is on the path but I chose to follow the tensorflow documentation regarding tested installs and make minimal changes to the script.

src/tfrnnlmbin/Makefile
```diff
@@ -35,6 +36,6 @@ ADDLIBS = ../lat/kaldi-lat.a ../lm/kaldi-lm.a ../fstext/kaldi-fstext.a \
 LDLIBS +=  -lz -ldl -fPIC -lrt
 LDLIBS += -L$(TENSORFLOW)/bazel-bin/tensorflow -ltensorflow_cc -ltensorflow_framework
 
-LDFLAGS += -Wl,-rpath=$(shell pwd)/../../tools/tensorflow/bazel-bin/tensorflow/
+LDFLAGS += -Wl,-rpath,$(shell pwd)/../../tools/tensorflow/bazel-bin/tensorflow/
```
kkm000: I think we should also use `$(TENSORFLOW)` here too instead of `$(shell pwd)/../../tools` part. The variable is defined at start of this file to exactly that value.

tools/extras/install_tensorflow_cc.sh
```diff
@@ -25,17 +25,17 @@ else
...
 unzip ../bazel.zip
 ./compile.sh
 cd ../
 
 # now bazel is built
-git clone https://github.com/tensorflow/tensorflow
+[ ! -d tensorflow ] && git clone https://github.com/tensorflow/tensorflow
 cd tensorflow
-git checkout r1.4
+git checkout r1.12
```
kkm000: Since the clone is conditional now on the absence of the directory, please add `git fetch --tags` before checking out, or the tag may just not be there in the old repo clone.

tools/extras/install_tensorflow_cc.sh
kkm000: Oh, I see. Complained that bazel was too new, wow! Thanks for the clarification.

langep: If I remember correctly it requested to downgrade to at most 0.21 but I can double check again.

kkm000: Up to you entirely, thanks for this contribution in any case! Personally I usually attempt to make the setup scripts as painless to the user as reasonable, but then I won't object to your definition of "reasonable" in this sense :)

Oh, and I just checked what versions are available to me, and I think I got actually seriously confused. First of all, I have the Google's deb feed for bazel added, probably at some time in the past, and forgot about it. So it is not at the Ubuntu's feed. And when I checked their Package.gz manifests, by downloading the file directly, it seems that the only version available at the feed for 18.04 is Bazel 0.24.1. Looks like they either unpublish the old ones, or did not possibly publish anything older than that for 18.04 ever. So the attempt to match the installed version of Bazel looks now more like a shot in the dark; the user (well, at the least, this particular user) has not even an option to downgrade his Bazel to the required version.

So I suggest just leaving this thing as is, do not spend much time on that please. I'm sorry about this confusion! I actually just meant to ask a question; I did not realize that tf was so picky about the version of blaze it uses.

kkm000:
LGTM. Thank you!

tools/extras/install_tensorflow_cc.sh
langep: I initially worked on this because some of my colleagues wanted to try out the lattice rescoring with a custom tensorflow language model but we ended up using the recipe provided in the Kaldi recipes. Maybe it is best to recommend that approach to most of the users anyway.

PR #3209, opened 2019-04-05
Minimize shared mem usage

kangshiyin:
Minimize shared mem usage and remove unnecessary __sync that may hurt the performance.
All GPU tests have passed.


danpovey:
Thanks.  I assume you made sure the tests still pass?

kangshiyin:
Yes,` make -j test` is done with no errors.

PR #3212, opened 2019-04-07
wav-to-duration works when header is corrupted

YoavRamon:
When the header is corrupted just read the entire file and get the duration from the amount of data.

### DIFF
--- a/src/featbin/wav-to-duration.cc
+++ b/src/featbin/wav-to-duration.cc
@@ -76,7 +76,7 @@ int main(int argc, char *argv[]) {
       for (; !wav_reader.Done(); wav_reader.Next()) {
         std::string key = wav_reader.Key();
         const WaveInfo &wave_info = wav_reader.Value();
-        BaseFloat duration = 0.0
+        BaseFloat duration = 0.0;
         if (wave_info.IsStreamed()) {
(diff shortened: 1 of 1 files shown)


kkm000:
Thanks for thinking about improving Kaldi, but I do not think that can be a general enough solution to be included into Kaldi. I see that you've closed the PR, but just in case, so you don't lose any time on this. I've seen many WAV files that have additional data chunks, either before or more commonly after the main `data/wav ` chunk stream. If you *know for sure* that your data does not have an extra block at the end, then it would be safe to skip the known fixed header altogether, using, for example, as simple a fix as `tail --bytes=+K`, where *K* is the number of bytes to skip, and pipe into `sox` as "raw", headerless data (specify bit rate, word width, and other stuff sox needs to know). But as a general feature this would not be good, as users applying it as fix to their data (and it's always easy to slap in a command-line option without understanding the full implications of its use) may easily end up with corrupted features that would be tremendously hard to diagnose later. Also (as is the case with `sox`), you'd need to be able then to specify the bit rate, word length, sample encoding (signed/unsigned/float), byte order, number of channels--and this is only for PCM data! This is why there is `sox`, and let's better just use it. But still, thank you for thinking how you could help us!

PR #3213, opened 2019-04-07
[build] check for i686 as a valid prefix for Android triplets

### LINKED ISSUE #3208: Can't build kaldi for android x86
Android NDK toolchains use `i686` as prefix for 32-bit x86 compilers/linkers.
image

But kaldi only allows x86 to be specified as a target.
This check is located here: https://github.com/kaldi-asr/kaldi/blob/ddeac9837a8f1e390a75948d0f34931a8f145c77/src/configure#L994

Can we relax it a bit to include i686 as a target?

Dr-Desty-Nova:
Add check for i686 as a valid prefix for Android triplets. Fixes #3208
    
Kaldi configure script for 32-bit x86 expects x86 cross-compile
toolchains to start with x86.
    
Android compilers, however,  have triplets like i686-linux-android-gcc
and i686-linux-android-ar, using i686 as a prefix.
    
Fixes #3208

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -991,7 +991,8 @@ if is_set $HOST; then
   IFS='-' read -ra PARTS <<< "$HOST"
   # The first field in the PARTS list is the target architecture.
   TARGET_ARCH="$PARTS"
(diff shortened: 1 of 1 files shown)


kkm000:
LGTM, thanks!

PR #3219, opened 2019-04-10
Move some protected methods to private

yzmyyff:


src/decoder/lattice-faster-decoder.h
```diff
@@ -360,6 +357,48 @@ class LatticeFasterDecoderTpl {
...
+  /// if this is set, then the output of ComputeFinalCosts() is in the next
+  /// three variables.  The reason we need to do this is that after
+  /// FinalizeDecoding() calls PruneTokensForFrame() for the final frame, some
+  /// of the tokens on the last frame are freed, so we free the list from toks_
+  /// to avoid having dangling pointers hanging around.
+  bool decoding_finalized_;
+
+  /// For the meaning of the next 3 variables, see the comment for
+  /// decoding_finalized_ above., and ComputeFinalCosts().
+  unordered_map<Token*, BaseFloat> final_costs_;
```
kkm000: Should be `std::unordered_map`. I'm wondering how did the `std` namespace leak into global...

src/decoder/lattice-faster-decoder.h
```diff
@@ -368,6 +407,9 @@ class LatticeFasterDecoderTpl {
   //    Elem *tail;
   //  };
 
+  // Deletes the elements of the singly linked list tok->links.
+  inline static void DeleteForwardLinks(Token *tok);
```
kkm000: I do not really understand the use of `inline` on a forward declaration of a class member function. `inline` allows the function to have multiple identical definitions, but it's not defined inline anywhere. It seems redundant, unless I am really missing something?

src/decoder/lattice-faster-decoder.h
```diff
@@ -520,6 +498,31 @@ class LatticeFasterDecoderTpl {
...
+
+  std::vector<StateId> queue_;  // temp variable used in ProcessNonemitting,
+  std::vector<BaseFloat> tmp_array_;  // used in GetCutoff.
+
+  // fst_ is a pointer to the FST we are decoding from.
+  const FST *fst_;
+
+  // delete_fst_ is true if the pointer fst_ needs to be deleted when this
+  // object is destroyed.
+  bool delete_fst_;
```
kkm000: Since the value of this flag is defined in constructor, I'd rather declare it const for clarity. Please see if `config_` is also a candidate for being const.

src/decoder/lattice-faster-decoder.h
danpovey: RE std::unordered_map, that's a special case; it has to do with compiler compatibility.  There is an ifdef somewhere that has a using statement.  But I think that is no longer needed now that we require c++11, so std::unordered_map is better now yes.

kkm000: Ah yes, it was in tr1/ or like. Time for spring cleaning!

yzmyyff: I think it can be changed in other PR.

src/decoder/lattice-faster-decoder.h
yzmyyff: It's not a new method whose definition in its cpp file. Or you mean that we should remove ```inline``` in the declaration.

src/decoder/lattice-faster-decoder.h
yzmyyff: ```delete_fst_``` can be const.  There is a method ```SetOptions``` which change config_.

danpovey:
sorry, too many lines changed here..

PR #3229, opened 2019-04-14
[build] Remove references to deprecated MKL libs in gst_plugin

### LINKED ISSUE #2202: Purpose of mkl_p4n?
It seems that Intel deprecated this a while ago (MKL 11?), and so if you try to compile the gst-plugin (which the linker needed mkl_p4n for), you'll get an error saying that it can't find it. I removed the -lmkl_p4n flag from gst-plugin's Makefile and it compiled fine, and testing with alumae's kaldi-gstreamer-server seems to work without any issues either.
So before I submit a PR, I just wanted to know the purpose of this library?

kkm000:
These libraries are long gone, see
https://software.intel.com/en-us/articles/intel-math-kernel-library-intel-mkl-112-deprecations

Fixes #2202

### DIFF
--- a/src/gst-plugin/Makefile
+++ b/src/gst-plugin/Makefile
@@ -34,11 +34,6 @@ ifneq ($(wildcard ../../tools/portaudio/install/include/pa_linux_alsa.h),)
     EXTRA_LDLIBS += -lasound
 endif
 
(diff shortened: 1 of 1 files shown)


kkm000:
This complies with MKL, I'll just merge.

PR #3236, opened 2019-04-16
Revert "Make no warp sync assumption for CC7.x"

danpovey:
Reverts kaldi-asr/kaldi#3211

### DIFF
--- a/src/cudamatrix/cu-kernels.cu
+++ b/src/cudamatrix/cu-kernels.cu
@@ -949,18 +949,14 @@ static void _copy_cols_from_vec(Real* m_out, MatrixDim d, const Real* v_in) {
 // _trace_mat_mat reduce the partial sum to
 // value[blockIdx.y * gridDim.x + blockIdx.x]
 // It use shared mem to transpose matrix B to ensure coalesced memory access
-// 2D block (8x32) is used
 template<int TileDim, typename Real>
 __global__
 static void _trace_mat_mat(const Real* A, const Real* B, MatrixDim dA,
                            int B_stride, Real* value) {
   // Reuse shared mem and make indexing easier. "+1" to avoid bank conflict
-  const int kWarpSize = 32;
-  typedef cub::BlockReduce<Real, kWarpSize, cub::BLOCK_REDUCE_WARP_REDUCTIONS,
-      CU1DBLOCK / kWarpSize> BlockReduceT;
   __shared__ union {
     Real trans[TileDim][TileDim + 1];
-    typename BlockReduceT::TempStorage sum;
+    Real sum[CU1DBLOCK];
   } smem;
 
   // linear thread id;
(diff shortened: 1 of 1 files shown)


danpovey:
@luitjens  found a problem with this, I have to revert it.  @luitjens if you could post what you found it would be great.

luitjens:
I don't know the exact problem yet but when I run make test in cudamatrix with these changes i get failures that I don't get without the changes.  

Currently i'm seeing failures in the code below with #if 1  and not #if 0.

__global__
static void _add_diag_mat_mat_MNT(const Real alpha, const Real* M,
                                  const MatrixDim dim_M, const Real* N,
                                  const int stride_N, const Real beta,
                                  Real* v) {
#if 1
  typedef cub::BlockReduce<Real, CU1DBLOCK> BlockReduceT;
  __shared__ typename BlockReduceT::TempStorage smem;
  const int tid = threadIdx.x;
  const int i = blockIdx.x;
  const int m_start = i * dim_M.stride;
  const int n_start = i * stride_N;

  // Loop along the matrix row. Reduce to CU1DBLOCK elements per row.
  Real tsum = Real(0);
  for (int j = tid; j < dim_M.cols; j += CU1DBLOCK) {
    tsum += M[m_start + j] * N[n_start + j];
  }

  // Block reduce
  tsum = BlockReduceT(smem).Sum(tsum);

  // output 1 sum per thread block
  if (tid == 0) {
    v[i] = alpha * tsum + beta * v[i];
  }
#else
  __shared__ Real ssum[CU1DBLOCK];
  const int tid = threadIdx.x;
  const int i = blockIdx.x;
  const int m_start = i * dim_M.stride;
  const int n_start = i * stride_N;

  // Loop along the matrix row. Reduce to CU1DBLOCK elements per row.
  Real tsum = Real(0);
  for (int j = tid; j < dim_M.cols; j += CU1DBLOCK) {
    tsum += M[m_start + j] * N[n_start + j];
  }
  ssum[tid] = tsum;
  __syncthreads();

  // Tree reduce to 2x warpSize elements.
# pragma unroll
  for (int shift = CU1DBLOCK / 2; shift > warpSize; shift >>= 1) {
    if (tid < shift)
      ssum[tid] += ssum[tid + shift];
    __syncthreads();
  }

  // Warp reduce to 1 element. Threads implicitly synchronized within a warp.
  if (tid < warpSize) {
#   pragma unroll
    for (int shift = warpSize; shift > 0; shift >>= 1) {
      ssum[tid] += ssum[tid + shift];
    }
  }

  // output 1 sum per thread block
  if (tid == 0) {
    v[i] = alpha * ssum[0] + beta * v[i];
  }
#endif

I don't see anything obvious yet but i'll keep looking and report back when I know more.

luitjens:
So it seems like the failure is non-deterministic and not necessarily related to the code above.

danpovey:
So perhaps I should revert the revert?
Or do you want to wait till you figure out what the problem was?  I guess
it's no hurry to reapply Shiyin's patch, as
anyway things are not working right on the architecture where he was trying
to fix the problem.

luitjens:
I think we should remove this code from TOT and wait for a fully vetted
patch before reapplying.  The CUB reduce code all works fine.  I've test
them individually using make test and it passes.  However, this assumes
make test covers these functions.

I then went and disabled all warpsyncs and the code passes.  I then went
and selectively enabled some warp syncs.  Here is at least one area that
fails:

template<EnumTransformReduce TransReduceType, typename Real>
__global__
static void _group_transform_reduce(
    Real *y, const Real *x, const MatrixDim d, const int src_stride,
    const int group_size, const TransReduceOp<TransReduceType, Real> op) {

  __shared__ Real sreduction[CU1DBLOCK];
  const int i = blockIdx.x;
  const int x_start = i * src_stride;
  const int y_start = i * d.stride;
  const int threads_per_group = blockDim.x;

  // Reduce n groups per thread block
  const int n = blockDim.y;
  const int len = group_size * n;
  // linear thread id
  const int tid = threadIdx.y * threads_per_group + threadIdx.x;
  int j = threadIdx.y * group_size + threadIdx.x; // col-id of *x
  int group_id = threadIdx.y;                     // col-id of *y
  int group_end = x_start + (group_id + 1) * group_size;

  while (group_id < d.cols) {
    // reduce to threads_per_group elements per group
    int x_idx = x_start + j;
    Real treduction = op.Transform(x[x_idx]);
    x_idx += threads_per_group;
    while (x_idx < group_end) {
      treduction = op.Reduce(treduction, op.Transform(x[x_idx]));
      x_idx += threads_per_group;
    }
    sreduction[tid] = treduction;
    if (threads_per_group > warpSize) {
      __syncthreads();
    }

    //PROBLEM HERE:  when threads_per_group <= 32 this is warp synchronous
and writes to sreduction above may not be visible at reads below.

    // tree-reduce to 2x warpSize elements per group
#   pragma unroll
    for (int shift = threads_per_group / 2; shift > warpSize; shift >>= 1) {
      if (threadIdx.x < shift) {
        sreduction[tid] = op.Reduce(sreduction[tid], sreduction[tid +
shift]);
      }
      __syncthreads();
    }

    // Warp-reduce to 1 element per group.
    const int warp_reduce_size =
        threads_per_group / 2 < warpSize ? threads_per_group / 2 : warpSize;
    if (threadIdx.x < warp_reduce_size) {
#     pragma unroll
      for (int shift = warp_reduce_size; shift > 0; shift >>= 1) {
        Real buf = op.Reduce(sreduction[tid], sreduction[tid + shift]);
        __syncwarp();  //PROBLEM HERE assumes 0xFFFFFFFF
        sreduction[tid] = buf;
        __syncwarp();  //PROBLEM HERE assumes 0xFFFFFFFF
      }
    }

There are two issues in the code above.
  1)  When blockDim.x<=32 we don't call syncwarp before reading from shared
memory.
  2)  When doing the warp reduction we don't pass in the mask which is
wrong when blockDim.x<32.

 Everywhere in this patchset that __syncwarp() is called we are just using
the default 0xFFFFFFFF which is only safe if we can guarantee
blockDim.x>=32.
If it is not then we need to rework those kernels.

Overall i think we should move this code to CUB or cooperative groups and
get rid of the __syncwarp() stuff as it is very brittle.

See here for more details on the issues with warpsync:
https://devblogs.nvidia.com/using-cuda-warp-level-primitives/

If we don't want to move directly to CUB for these reductions we should at
minimum use cooperative groups which gives us the portable abstractions we
need to safely express these algorithms.  In particular we could use the
tiled partition:  https://devblogs.nvidia.com/cooperative-groups/

Note the reason i'm seeing these failures and you are not are likely that
this is a Volta+ issue and internally your hardware is not Volta+.  Do you
have any Turing boards you could test on?

danpovey:
We are getting some RTX boards but they are not installed yes, so no.
Sorry guys, I don't have the bandwidth to fully follow this discussion, I
hope others will sort it out.

PR #3237, opened 2019-04-17
Rm chain recipe (local/chain/run_tdnn_5o.sh)

indra622:
In RM recipe, 'local/chain/run_tdnn_5n.sh' script is deprecated because 'steps/nnet3/tdnn/make_configs.py' is no longer used. 

I modified stage 7 in 'local/chain/run_tdnn_5n.sh' in RM recipe to wsj's 'local/chain/tuning/run_tdnn_1g.sh' 's stage 15 (neural net configs)

As a result, it shows absolute 1.52% WER better than RESULTS file in RM recipe.


indra622:
P.S.
'local/chain/run_tdnn_5n.sh' is no more working properly because 'steps/nnet3/tdnn/make_configs.py' is deprecated.
KakaoTalk_20190416_003844649

danpovey:
Thanks!  Before I merge, would you mind also trying the setup from mini_librispeech?  That dataset is more similar in size to RM.

indra622:
@danpovey 

I was trying the same topology to mini_librispeech, it shows abolute 8.60% and 12.00% WER.

image

danpovey:
That is about 0.1% absolute better than the current recipe (see
run_tdnn.sh).
Probably not worth changing it over due to the extra size.
Please try it anyway for the RM setup, before I merge.
And we don't need the extra commits in the mini_librispeech setup (i.e.
keep the diff clean).

indra622:
Thanks to your comment. My apology to my mistake about mini_librispeech recipe.

I was trying to mini_librispeech's config, Its WER is 1.59%. (My proposed model is 1.52% WER)

Can you merge in specific commit id such 2c99e16 commit? If you can't, I'll open another pull request for merging. 

If you want, I request mini librispeech's config code, as another pull request.

danpovey:
Oh.  I don't think I can pick out commits without messing around with the
command line, please figure it out; you can check out just that commit and
then push to the branch of your PR with "force push", e.g. git push origin
+HEAD:your_branch_name (as it won't be a fast forward commit).

indra622:
Thank you, I found it.

egs/rm/s5/local/chain/run_tdnn_5o.sh
```diff
@@ -0,0 +1,179 @@
+#!/bin/bash
+
+# this script is a modified version of run_tdnn_5n.sh. It uses
+# a new configs convention for chain model after kaldi 5.2.
```
danpovey: can you please mention that the results are in the RESULTS file?  People normally look for them here.
Or duplicate them here.

danpovey: .. and also, please move all these tuning scripts to a subdirectory tuning/, and make a link run_tdnn.sh -> tuning/run_tdnn_5o.sh.

indra622: Thank you, I moved scripts and added mention in RESULTS file

danpovey: It doesn't look like you removed the scripts from their original location.  Also make sure the soft link run_tdnn.sh is there; it seems to be missing (you'll have to "git add" it).

indra622: I'm sorry, I missed. Thank you.

PR #3240, opened 2019-04-17
Update hkust_data_prep.sh

zh794390558:
add `set -e` for  `python -c "import mmseg" 2>/dev/null || \
  (echo "mmseg is not found. Checkout tools/extra/install_mmseg.sh" && exit 1;)` not exit as expected.

### DIFF
--- a/egs/hkust/s5/local/hkust_data_prep.sh
+++ b/egs/hkust/s5/local/hkust_data_prep.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -e
  
 . ./path.sh || exit 1;
 
--- a/egs/hkust/s5/local/hkust_data_prep.sh
+++ b/egs/hkust/s5/local/hkust_data_prep.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 . ./path.sh || exit 1;
 
--- a/egs/hkust/s5/local/hkust_data_prep.sh
+++ b/egs/hkust/s5/local/hkust_data_prep.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
-
+ 
 . ./path.sh || exit 1;
 
 if [ $# != 2 ]; then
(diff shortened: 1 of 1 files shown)


kkm000:
Thanks. `set -e` in itself is not wrong.

As for the code,
https://github.com/kaldi-asr/kaldi/blob/952bfd84d4325a50f7c64e16df66519c16581ece/egs/hkust/s5/local/hkust_data_prep.sh#L68-L72

the `echo` and `exit` are run in a subshell. The `exit` exits the subshell with code 1 and returns back to the program, which continues to run happily This is the actual bug.. What was intended here was running the echo and exit in a *group command* instead of a subshell:
```sh
python -c "import mmseg" 2>/dev/null || {
    echo "mmseg is not found. Checkout tools/extra/install_mmseg.sh"; exit 1; }
```
If you are up to helping with a better fix, please fix this bug. Also, move the whole check to an earlier point, so the script fails quickly if the module not found. I'd put as early as before the first `mkdir` at the line 18. And even better, I'd reword the error message for clarity, "Python module mmseg is not found. To install it, run tools/extra/install_mmseg.sh".

zh794390558:
thanks for you recommendation。

zh794390558:
please check.

egs/hkust/s5/local/hkust_data_prep.sh
```diff
@@ -14,6 +14,11 @@ hkust_text_dir=$2
 train_dir=data/local/train
 dev_dir=data/local/dev
 
+# transcripts normalization and segmentation
+# needs external tools
+python -c "import mmseg" 2>/dev/null || {
```
danpovey: I think invoking this as python2.7 would be better, since mmseg doesn't work on python3, IIRC.  Obviously we should make sure that the script that actually uses mmseg later on also invokes it that way.

zh794390558: Done.

PR #3241, opened 2019-04-17
add python symlink to .gitignore in cases python2 doesn't exist

mcalahan:
Compiled on MacOs Mojave with brew installed dependencies.  
python2 binary/link doesn't exist so the extra/check_dependencies.sh creates a symlink for python2 in tools/python/.

This location is not ignored in .gitignore so it causes git to assume directory is dirty.


kkm000:
LGTM, good catch, thanks!

PR #3243, opened 2019-04-17
[scripts] Enhance argument checks in nnet3/align_lats.sh

### LINKED ISSUE #3242: Alignments from lattices: why lattice-best-path, not lattice-oracle?
The alignment scripts (steps/align_fmllr_lats.sh, steps/nnet3/align_lats.sh) are using ` lattice-best-path` to make alignments from the newly produced lattices. I remember even adjusting the `--acoustic-scale` to match the best test decode, to get a bit cleaner alignments (they were not amazingly important, but since I had the number, why not). It kinda just struck me just now, should they rather use `lattice-oracle`, so that the ground truth transcriptions are used?

I did not try yet, mostly just asking, maybe there is a reason this is not the right idea that I am not aware of.
https://github.com/kaldi-asr/kaldi/blob/f8021d7650bec72c5b3fa8f5d0155bd83f79f04b/egs/wsj/s5/steps/align_fmllr_lats.sh#L158-L163

(and, come think of it, the core strep of these scripts, `*-latgen-faster`, is not using the ground truth either).

----

Bottom line: add a check for acoustic_weight==1.0 for a chain system in steps/nnet3/align_lats.sh.

kkm000:
When a chain model is detected (frame sample factor > 1), verify
that --acoustic_scale is 1.0, and options in --scale_opts include
--transition-scale=1.0 and --self-loop-scale=1.0. Print an
error and exit with status 1 when any of these checks do not pass.

Fixes #3242.

### DIFF
--- a/egs/wsj/s5/steps/nnet3/align_lats.sh
+++ b/egs/wsj/s5/steps/nnet3/align_lats.sh
@@ -92,12 +92,16 @@ if [ -f $srcdir/frame_subsampling_factor ]; then
   frame_subsampling_factor=$(cat $srcdir/frame_subsampling_factor)
   frame_subsampling_opt="--frame-subsampling-factor=$frame_subsampling_factor"
   cp $srcdir/frame_subsampling_factor $dir
-  if [ "$frame_subsampling_factor" -gt 1 ] && \
-     [ "$scale_opts" == "--transition-scale=1.0 --self-loop-scale=0.1" ]; then
-    echo "$0: frame-subsampling-factor is not 1 (so likely a chain system),"
-    echo "...  but the scale opts are the defaults.  You probably want"
-    echo "--scale-opts '--transition-scale=1.0 --self-loop-scale=1.0'"
-    sleep 1
+  if [[ $frame_subsampling_factor -gt 1 ]]; then
+    # Assume a chain system, check agrument sanity.
+    if [[ ! ($scale_opts == *--self-loop-scale=1.0* &&
+             $scale_opts == *--transition-scale=1.0* &&
+             $acoustic_scale = '1.0') ]]; then
+      echo "$0: ERROR: frame-subsampling-factor is not 1, assuming a chain system."
+      echo "... You should pass the following options to this script:"
+      echo "  --scale-opts '--transition-scale=1.0 --self-loop-scale=1.0'" \
+           "--acoustic_scale 1.0"
+    fi
   fi
 fi
 


kkm000:
E. g.
```
steps/nnet3/align_lats.sh: *** WARNING! ***
... frame-subsampling-factor is not 1, likely a chain system. If so,
... the scale_opts are incorrect.  You probably want:
  --scale-opts '--transition-scale=1.0 --self-loop-scale=1.0'
... the acoustic_scale is incorrect.  You probably want:
  --acoustic_scale 1.0
```

egs/wsj/s5/steps/nnet3/align_lats.sh
```diff
@@ -92,12 +92,27 @@ if [ -f $srcdir/frame_subsampling_factor ]; then
...
   cp $srcdir/frame_subsampling_factor $dir
-  if [ "$frame_subsampling_factor" -gt 1 ] && \
-     [ "$scale_opts" == "--transition-scale=1.0 --self-loop-scale=0.1" ]; then
-    echo "$0: frame-subsampling-factor is not 1 (so likely a chain system),"
-    echo "...  but the scale opts are the defaults.  You probably want"
-    echo "--scale-opts '--transition-scale=1.0 --self-loop-scale=1.0'"
-    sleep 1
+  if [[ $frame_subsampling_factor -gt 1 ]]; then
+    # Likely a chain system, check agrument sanity.
+    epilogue() { :; }
```
danpovey: Sorry this is just a little bit too much for me; I'd prefer the code be simple and self-explanatory.
Just printing the error message and exiting with status 1 is fine; people can always set slightly different values if they really want to use those (weird) values.

kkm000: This is what it did before w.r.t. the scale_opts, printed a warning and paused, did not exit. I just added the second check. Am I missing the idea?

danpovey: You can simplify it and have it just exit.  All this epilogue() / prologue() stuff is too complex-- just have it print the message and exit.

kkm000: Or do you mean the code style? Local helper functions?

kkm000: Exiting makes sense only we sure that this is in fact a chain model--are we?

I can redo the same logic without functions, if you think they are unreadable. To me, they are just shell variables, only functions, but I'm not going to insist. :)

Maybe it would be better then make a separate nnet3/chain/aligh_lats.sh, with the correct defaults? Or nor even defaults, just hardcoded required values. What do you think?

danpovey:
This isn't worth spending time arguing about.. please just do it the way I
ask.

kkm000:
I'm all for it, I am not arguing! I'm trying to understand what you ask. Do all the same checks, print the same message, do not use functions, and exit instead of the 'sleep 1' -- correct?

danpovey:
Yes.

kkm000:
Error message:
```
steps/nnet3/align_lats.sh: ERROR: frame-subsampling-factor is not 1, assuming a chain system.
... You must pass the following options to this script:
  --scale-opts '--transition-scale=1.0 --self-loop-scale=1.0' --acoustic_scale 1.0
```

egs/wsj/s5/steps/nnet3/align_lats.sh
```diff
@@ -92,12 +92,16 @@ if [ -f $srcdir/frame_subsampling_factor ]; then
...
-    echo "...  but the scale opts are the defaults.  You probably want"
-    echo "--scale-opts '--transition-scale=1.0 --self-loop-scale=1.0'"
-    sleep 1
+  if [[ $frame_subsampling_factor -gt 1 ]]; then
+    # Assume a chain system, check agrument sanity.
+    if [[ ! ($scale_opts == *--self-loop-scale=1.0* &&
+             $scale_opts == *--transition-scale=1.0* &&
+             $acoustic_scale = '1.0') ]]; then
+      echo "$0: ERROR: frame-subsampling-factor is not 1, assuming a chain system."
+      echo "... You must pass the following options to this script:"
```
danpovey: cool, except let's do s/must/should probably/.

kkm000: Done

kkm000: The English article has always been a mystery to me, but I thought I grokked the modalities at last... What a setback! :)))

danpovey:
Thanks! Merging.

kkm000:
Thanks again for pinpointing the error Dan! I was puzzled by it for quite a while!

Issue #3238, opened 2019-04-17
mmseg not support for python3 under hkust


zh794390558:

danpovey:
@naxingyu do you have any suggestion for what to do?

naxingyu:
I'll check.

naxingyu:
@zh794390558 for quick fix, you could try replace the module with this: https://github.com/bxshi/py3mmseg
The readme says it only works for python3.* and it only tested in python 3.3, and it hasn't been updated since 4 years ago. I'm really sorry that I don't have time to test this. If it works, it would be great if you could create a PR. Thanks.

danpovey:
Obviously you would have to invoke python as `python3` if you were relying
on it being python 3.

naxingyu:
Yes, the repo is just for test. If it works, I'll change it to accept both
py2&3.

naxingyu:
When I have time, I'll fork the original mmseg module and make it work for
both py2&3.

danpovey:
Actually I think Kaldi currently requires both python3 and python2.7, so a
quicker fix might simply be to invoke python as `python2.7`.  However in
the longer term we do want to only depend on python3; python2.7 is required
only for sequitur g2p right now, IIRC.

naxingyu:
You are right. I'll look into it.

kkm000:
FWIW, the configure still sets python 2.7 as the default.
https://github.com/kaldi-asr/kaldi/blob/299b111d05e9f628767773ca924c7a9c48f29603/tools/extras/check_dependencies.sh#L117-L128

kkm000:
On the same point, when I hacked the train.py scripts, I cried for the python3 `f'...'` interpolated strings. The code using `.format()` is supremely unreadable there.

danpovey:
Most of our recent scripts that use python explicitly invoke python3.  The
train.py scripts should be among these.
Making `python` the default is mostly to increase the changes of misc
things in egs directories working.

naxingyu:
There are cases where people installed kaldi before installing python3, and
check_dependency is never touched since it's not required for compile src.

kkm000:
@danpovey 
> our recent scripts that use python explicitly invoke python3. The train.py scripts should be among these.

Trains scripts use py2 (via its being the implicit default).
[...]

PR #3244, opened 2019-04-18
invoke python2.7 for mmseg

naxingyu:
For fixing #3238

### DIFF
--- a/egs/hkust/s5/local/hkust_data_prep.sh
+++ b/egs/hkust/s5/local/hkust_data_prep.sh
@@ -75,7 +75,7 @@ cat $train_dir/transcripts.txt |\
   sed -e 's/<noise>\(.\+\)<\/noise>/\1/g' |\
   sed -e 's/((\([^)]\{0,\}\)))/\1/g' |\
   local/hkust_normalize.pl |\
-  python local/hkust_segment.py |\
+  local/hkust_segment.py |\
   awk '{if (NF > 1) print $0;}' > $train_dir/text || exit 1;
 
 cat $dev_dir/transcripts.txt |\
@@ -84,7 +84,7 @@ cat $dev_dir/transcripts.txt |\
   sed -e 's/<noise>\(.\+\)<\/noise>/\1/g' |\
   sed -e 's/((\([^)]\{0,\}\)))/\1/g' |\
   local/hkust_normalize.pl |\
-  python local/hkust_segment.py |\
+  local/hkust_segment.py |\
   awk '{if (NF > 1) print $0;}' > $dev_dir/text || exit 1;
 
 # some data is corrupted. Delete them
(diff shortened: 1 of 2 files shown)


naxingyu:
I just found out that gale_mandarin recipe, which is way outdated (not even
nnet2), is also using mmseg... Does it deserve a fix?

kkm000:
My take is that keeping dead recipes in egs/ is not the best approach overall. If all it takes is the same 3-line fix, I would.

But that's a good general question. Since the egs are often part of published/reproducible results, they better be kept, but they become a maintenance burden. But Git remembers everything; anyone can get an eg and the codebase as they were when a recipe was published. It makes sense to me to keep only the README and RESULTS files for outdated egs, and add a note to the README how to get the old eg code by its commit (simple instructions to the user how to do it, like `git checkout deadc0de42`). `git blame RESULTS` is the best way to see the history of reference results at a glance; this file receives no other edits in practice; mentioning the command could be a good idea too, We probably should get to it when Dan has a breather.

danpovey:
I would prefer to keep the outdated egs somewhat alive, and in the repo in
the normal way, although perhaps we could add warnings.
The issue is that someone has done the work to set up the data preparation
scripts for those datasets,
and there is value in that even if the recipes are not up to date; and
people find those recipes and eventually update them if they are
interested.  Putting up more barriers to entry could impede that process.
For example, someone just submitted an update for the RM recipe which had
been outdated for ages.  If we had made that more inaccesible that might
not have happened.

kkm000:
@danpovey, this makes sense. Even outdated egs can serve as examples. This requires marking them as no longer maintained, and that the users have an easy way--they do, I should say rather say know the easy way to pull the historical working slice. And I also see how people submit PRs, on top of random outdated branches. Both these issues converge on a more efficient use of Git by the users. I understand why do you rebase/collapse when merging: this is mainly because how the patches are sent in, sometimes on top of random branches neglected for years. Merging there as normal topo merge commits would turn the log graph into a horryfying bundle of capellini.

So I have been pondering writing a guide on efficient Git use for Kaldi users and occasional contributors. We touch on that in the docs, but just too briefly. I'm going to start when I have less rush at work. Are you ok with putting it into the main repo Wiki section on GitHub? It can be collaboratively edited there, too. We could convert them into a docx by the end, or leave there--that we could decide later. Is this ok by you?

danpovey:
A guide for git usage is a good idea, I think.
I think it will be difficult to mark recipes as maintained or
not-maintained, because in reality it's a continuous
spectrum.  Telling people in a prominent place which recipes they should
look at first
(e.g. mini_librispeech, swbd/s5c) might be a good idea though.

kkm000:
Yup. I noticed most first-timers tend to start with the largest dataset they can get hold of (my first was tedlium). I like learning stuff the hard way, but I know for others it may be discouraging.

@naxingyu, by the way, FYI:
> For fixing #3238

The list of keywords GH understands: https://help.github.com/en/articles/closing-issues-using-keywords. "Fix", "fixes" or "fixed" would do it, but "fixing" they do not match.

naxingyu:
@kkm000 oh, didn't know that, thanks!

PR #3247, opened 2019-04-18
nnet3,xconfig_to_configs.py:

vesis84:
- in 'check_model_contexts()' when parsing "left-context,right-context",
  don't apply 'nnet3-copy --edits=' on 'init.raw' !!!
   - the 'init.raw' is used as feature transform prior to LDA estimation,
     and here we don't need to rename the output layers.
   - renaming the output layers is relevant to multilingual networks,
     where we can have output layers 'output-0 output-1 output-2',
     layer renaming 'output-0 -> output' is used to get the contexts.
     But we cannot use the same rename on 'init.raw', which does not
     have a component with name 'output-0'.

- We also avoid calling pipe with 'nnet3-info {} | head -n 4'
   - if nnet3-info fails, the pipeline still returns zero status
     and no Python stack-trace was displayed.


vesis84:
I tested this by running both a multi-lingual setup and a simple wsj setup... It works well in both cases...
K.

kkm000:
LGTM

PR #3248, opened 2019-04-18
wsj: exposing i-vector #jobs config to the master script,

vesis84:
- there is the 3layer granularity: threads in a binary, mid-level processes
  (summing stats and saving them), top-level processes (-nj).
- in our cluster it is difficult to get a JOB with too many threads
  (the jobs get stuck in the queue).
- if the settings are exposed into the top-level script, it will be
  easier to change to setup to make it work well by modifying the top
  level script only...

### DIFF
--- a/egs/wsj/s5/local/chain/tuning/run_tdnn_1g.sh
+++ b/egs/wsj/s5/local/chain/tuning/run_tdnn_1g.sh
@@ -53,7 +53,7 @@ xent_regularize=0.1
 dropout_schedule='0,0@0.20,0.5@0.50,0'
 
 # training chunk-options
-chunk_width=140,100,60
+chunk_width=140,100,160
 # we don't need extra left/right context for TDNN systems.
 chunk_left_context=0
 chunk_right_context=0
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/local/chain/tuning/run_tdnn_1g.sh
```diff
@@ -46,7 +53,7 @@ xent_regularize=0.1
 dropout_schedule='0,0@0.20,0.5@0.50,0'
 
 # training chunk-options
-chunk_width=140,100,160
+chunk_width=140,100,60
```
danpovey: can you not change the chunk_width?  This could affect the results.

vesis84: I was written in documentation, that the chunk-width should be a list of integers in descending order (try length 140, 100, ...), therefore the 160 did not make sense to me, which is why I changed it. Or am I wrong? What was the intention of using the values '140,100,160' ?

PR #3249, opened 2019-04-18
[make_voxceleb1.pl] Fix train/test data split for the latest version of the voxceleb1 dataset 

sunshines14:
I found that previous scripts do not work with the latest version of the voxceleb1 dataset.
Therefore, I fixed the script for the latest version as follows: 
1. 'Url' update to download the latest version (from official site)
2. Delete the script about 'vox1_meta.csv' no longer needed
3. Fix the code for the latest version's output (variable name equals make_voxceleb2.pl)

It is confirmed that successful output can be obtained from the script. 
ex) voxceleb1_test -- spk2utt, trials, utt2spk, wav.scp

### DIFF
--- a/egs/voxceleb/v1/local/make_voxceleb1.pl
+++ b/egs/voxceleb/v1/local/make_voxceleb1.pl
@@ -28,28 +28,45 @@ my @spkr_dirs = grep {-d "$data_base/voxceleb1_wav/$_" && ! /^\.{1,2}$/} readdir
 closedir $dh;
 
 if (! -e "$data_base/voxceleb1_test.txt") {
-  system("wget -O $data_base/voxceleb1_test.txt http://www.robots.ox.ac.uk/~vgg/data/voxceleb/meta/veri_test.txt");
+  system("wget -O $data_base/voxceleb1_test.txt http://www.openslr.org/resources/49/voxceleb1_test.txt");
+}
+
+if (! -e "$data_base/vox1_meta.csv") {
+  system("wget -O $data_base/vox1_meta.csv http://www.openslr.org/resources/49/vox1_meta.csv");
 }
 
 open(TRIAL_IN, "<", "$data_base/voxceleb1_test.txt") or die "Could not open the verification trials file $data_base/voxceleb1_test.txt";
+open(META_IN, "<", "$data_base/vox1_meta.csv") or die "Could not open the meta data file $data_base/vox1_meta.csv";
 open(SPKR_TEST, ">", "$out_test_dir/utt2spk") or die "Could not open the output file $out_test_dir/utt2spk";
 open(WAV_TEST, ">", "$out_test_dir/wav.scp") or die "Could not open the output file $out_test_dir/wav.scp";
 open(SPKR_TRAIN, ">", "$out_train_dir/utt2spk") or die "Could not open the output file $out_train_dir/utt2spk";
 open(WAV_TRAIN, ">", "$out_train_dir/wav.scp") or die "Could not open the output file $out_train_dir/wav.scp";
 open(TRIAL_OUT, ">", "$out_test_dir/trials") or die "Could not open the output file $out_test_dir/trials";
 
+my %id2spkr = ();
+while (<META_IN>) {
+  chomp;
+  my ($vox_id, $spkr_id, $gender, $nation, $set) = split;
+  $id2spkr{$vox_id} = $spkr_id;
+}
+
 my $test_spkrs = ();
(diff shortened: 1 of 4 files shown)


sunshines14:
+ if previous script is still needed, I can use this modified script to another version (ex. v2) rather than replacing the existing script.

danpovey:
Thanks a lot!! @david-ryan-snyder  can you please check this?

david-ryan-snyder:
Thanks @sunshines14.

Could you do what you suggested and copy this to a new script called make_voxceleb1_v2.pl?

Then, in the run.sh script, make the new script the default one. Comment out the the old version of the script, and above it write a short comment describing the situation. E.g., mention that if you downloaded the dataset soon after it was released, you will want to use the make_voxceleb1.pl script instead.

egs/voxceleb/v1/local/make_voxceleb1.pl
```diff
@@ -28,45 +28,28 @@
 closedir $dh;
 
 if (! -e "$data_base/voxceleb1_test.txt") {
-  system("wget -O $data_base/voxceleb1_test.txt http://www.openslr.org/resources/49/voxceleb1_test.txt");
-}
-
-if (! -e "$data_base/vox1_meta.csv") {
-  system("wget -O $data_base/vox1_meta.csv http://www.openslr.org/resources/49/vox1_meta.csv");
+  system("wget -O $data_base/voxceleb1_test.txt http://www.robots.ox.ac.uk/~vgg/data/voxceleb/meta/veri_test.txt");
```
david-ryan-snyder: Can you point the new URL to http://www.openslr.org/resources/49/voxceleb1_test_v2.txt instead? We prefer to host these things ourselves where possible.

sunshines14:
Thanks @david-ryan-snyder.

I fixed some code as you commented as follow:
1. I have made a new script called 'make_voxceleb1_v2.pl'.
2. In the run.sh script, the v2 script used as the default one.
3. In addition, I have made the code simpler, including adding <dataset> in 'make_voxceleb1_v2.pl'.

It was reconfirmed that successful output can be obtained from all of fixed scripts.
Thanks.

egs/voxceleb/v1/local/make_voxceleb1_v2.pl
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/perl
+#
+# Copyright 2018  Ewald Enzinger
+#           2018  David Snyder
```
david-ryan-snyder: You can credit yourself here, if you want.

egs/voxceleb/v1/run.sh
```diff
@@ -24,11 +24,14 @@ stage=0
 if [ $stage -le 0 ]; then
   local/make_voxceleb2.pl $voxceleb2_root dev data/voxceleb2_train
   local/make_voxceleb2.pl $voxceleb2_root test data/voxceleb2_test
-  # This script reates data/voxceleb1_test and data/voxceleb1_train.
+  # This script reates data/voxceleb1_test and data/voxceleb1_train for latest version of VoxCeleb1.
```
david-ryan-snyder: Can you fix the typo, reates -> creates while you're at it?

david-ryan-snyder:
Thanks @sunshines14! I just suggested you credit your work in the v2 perl script and fix a preexisting typo. Then we can merge it.

sunshines14:
I did it all. 
Thanks @david-ryan-snyder.

egs/voxceleb/v1/local/make_voxceleb1_v2.pl
```diff
@@ -0,0 +1,119 @@
...
+#           2019  Soonshin Seo
+#
+# Usage: make_voxceleb1.pl /export/voxceleb1 data/
```
david-ryan-snyder: Sorry to nitpick, could you fix this comment, the usage is for the old script, not the new one.
Also, while you're here, could you write a comment that briefly summarizes why we need this script. Just say what you said before in the run.sh. Something like "The VoxCeleb 1 corpus underwent several updates that changed the directory and speaker ID format. The script make_voxceleb1.pl works for the oldest version of the corpus. This script should be used if you've downloaded the corpus recently."

sunshines14: Fine. I agree with your opinion. I think it is an important point. 
Thanks.

david-ryan-snyder:
Thanks @sunshines14, looks good to me. @danpovey, I think it's fine to merge this.

PR #3252, opened 2019-04-20
[RM chain script] refactoring RM chain script

indra622:
Following  #3237 , I proposes this pull request for refactoring. 

Mainly changed;
1) In chain process, 'local/chain/run_tdnn.sh' script in RM egs is no longer generating High-resolution features because It does not use it.  (It has lower performance than no use version.)
2) Now, It uses model topology from mini_librispeech's. It shows a higher performance than wsj's under the appropriate number of epochs. (Minimum WER is 1.37%)
3) The run.sh introduces the recent chain scripts. (pointed by symbolic link.)

It is confirmed that the successful outputs can be obtained from run.sh script.

### DIFF
--- a/egs/rm/s5/RESULTS
+++ b/egs/rm/s5/RESULTS
@@ -235,7 +235,7 @@ for x in exp/nnet2_online_wsj/nnet_ms_a_smbr_0.00005/1/decode_*; do grep WER $x/
 %WER 2.71 [ 340 / 12533, 58 ins, 59 del, 223 sub ] exp/chain/tdnn_5n/decode/wer_4_0.0
 # Its topology of chain model is from mini_librispeech's.
 # It uses a new configs convention for chain model after kaldi 5.2.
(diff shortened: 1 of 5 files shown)


danpovey:
I notice that you also reduced the size of the model.  What was the WER with the model-size reduced but hires MFCCs?  I may prefer to keep hires MFCCs just for consistency with other setups.

indra622:
@danpovey 

I totally agree your mention for consistency. When I ran hires MFCCs but 13 bins, Its minimum record is 1.32% WER. (40 bin was shown about 1.8% WER in previous scripts. Additionally, previous rm chain scripts such as run_tdnn_5n.sh used hires MFCC just for ivector, but did not use for train/test.) It is a better performance than I proposed even the lower number of epoch. Thanks to your advise.

I made mfcc_hires.conf file, changed scripts to use hires MFCCs and update RESULTS.

danpovey:
Thanks!  Will merge.

Issue #3228, opened 2019-04-14
Why our approach to compiling and linking is the way it is (BLAS, mostly)?


kkm000:
Here's the rub. I has been working on adding MKL as the default math library for Kaldi, and am still catching the fallout (this is a good thing, I think I can clean up some old stuff which may no longer be required).

One thing that I cannot grok is why the configuration is not using more or less system defaults. If my system supplies a development package, then I am supposed to be able to just link to the library, it should be "just there".

I got a few docker images, installed, after some fighting with signing keys, the packaged openblas, and then the test like this "just compiles", and it is supposed to! The only variable is the location of the `cblas.h` file:

```
echo $'#include <openblas/cblas.h>\nint main(){return 0;}' | g++ -xc - -o /dev/null -l:libopenblas.so
```
(or `-l:libopenblas:a` if I want it static, or `-lopenblas` if I do not care).

But when I am looking at how configure (a) requires specifying `--openblas-root` and (b) attempts to find the exact library file, then bakes the .so path into the executables, I do not really understand what was the reason to go such a great length instead of just letting the tooling figure out where its libraries and includes are, at the very least, by default.

So my questions, basically,
1. Is the reason that this complexity was created in the first place still exists today?
2. If yes, what it is?
3. So, if the answer is no, can I just trust the compiler with figuring out everything, and Bob's your uncle?

Or, in other words, naively, if I was writing this script today, I'd basically ran the above compile line three times:
1. Using the location of a compiled OpenBlas in tools/OpenBLAS/install, as we building it as part of tools (by adding -L and -I paths). If successful, bake in the `-rpath` if using .so.
2. If not, same test without -L and -I, maybe with the variation between `#include <openblas/cblas.h>` and `#include <cblas.h>`. If compiles, we are ok without `-rpath`.

But since the whole caboodle exists, I think I am still missing something. What are the edge cases?

danpovey:
I don't have a lot experience of writing these types of installation
scripts, I am fine to refactor it.  That was just what seemed the obvious
way to me.  Also, until the 'configure' finishes, as it's currently
written, you wouldn't be able to 'make', so wouldn't be able to easily
test.   But if there is a standard way to do these things that's easier, go
ahead.

vesis84:
[...]

PR #3261, opened 2019-04-24
[build] Build and configure OpenBLAS; default to it on non-x64 machine

kkm000:
* Default to OpenBLAS on non-x64 platforms.
* Download and build specific release of OpenBLAS from GitHub.
* Configure looks in the OpenBLAS build location only; the user
  can still specify `--openblas-root=`.
* Refactor switch parsing, chiefly to distinguish whether the
  `--openblas-root=` switch was not given, or was but pointed
  to a non-existing directory.
* Trim some dead ATLAS-related code left after commit 3e77220b3.
* Name all ATLAS configuration functions `*_atlas_*`.
* Move common Linux CUDA and Speex configuration out of math
  library configuration functions.

Fix: #3222
Mention: #3228

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -1,46 +1,41 @@
 #!/bin/bash
 
-# This configure script is hand-generated, not auto-generated.
-# It creates the file kaldi.mk, which is %included by the Makefiles
-# in the subdirectories.
+# This configure script is hand-generated, not auto-generated.  It creates the
+# file kaldi.mk, which is %included by the Makefiles in the subdirectories.
 # The file kaldi.mk is editable by hand -- for example, you may want to
-# remove the options -g -O0 -DKALDI_PARANOID, or edit the
-# DOUBLE_PRECISION variable (to be 1 not 0).
-
+# uncomment the options -O0 -DKALDI_PARANOID, or edit the DOUBLE_PRECISION
+# variable (to be 1 not 0).
 
 #  Example command lines:
-# ./configure --shared  ## shared libraries.
 # ./configure
-# ./configure --mkl-root=/opt/intel/mkl
-# ./configure --mkl-root=/opt/intel/mkl --threaded-math=yes
-# ./configure --mkl-root=/opt/intel/mkl --threaded-math=yes --mkl-threading=tbb
-#        # This is for MKL 11.3, which does not seem  to provide Intel OMP libs
-# ./configure --openblas-root=../tools/OpenBLAS/install
-#        # Before doing this, cd to ../tools and type "make openblas".
+# ./configure --shared                # Build shared Kaldi libraries.
+# ./configure --mathlib=OPENBLAS      # Build and use OpenBLAS.
+#        # Before doing this, cd to ../tools and type "make -j openblas".
+# ./configure --openblas-root=/usr    # Use system OpenBLAS.
 #        # Note: this is not working correctly on all platforms, do "make test"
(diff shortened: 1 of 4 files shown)


kkm000:
Tested on linux with both

```
$ ./configure --mathlib=OPENBLAS && make -C matrix clean && make -C matrix all test
$ ./configure --mathlib=OPENBLAS --static-math && make -C matrix clean && make -C matrix all test
```

### Messages

Some messages are new/updated.

#### Failures

```
$ ./configure --mathlib=OPENBLAS
Configuring KALDI to use OPENBLAS.
Checking compiler g++ ...
Checking OpenFst library in /home/kkm/work/kaldi2/tools/openfst-1.6.9 ...
Checking cub library in /home/kkm/work/kaldi2/tools/cub-1.8.0 ...
./configure: FATAL: OpenBLAS not found in '../tools/OpenBLAS/install'.
** This is the only place we look for it. The best option is to build OpenBLAS
** tuned for your system and CPU. To do that, run the following commands:
**
**   cd ../tools; extras/install_openblas.sh
**
** Another option is to specify the location of existing OpenBLAS directory
** with the switch '--openblas-root='. However, even if a package is provided
** for your system, the packaged version is almost always significantly slower
** and often older than the above commands can fetch and build.
**
** You can also use other matrix algebra libraries. For information, see:
**   http://kaldi-asr.org/doc/matrixwrap.html
```

```
$ ./configure --openblas-root=
./configure: FATAL: '--openblas-root=': switch requires a value. See './configure --help'.
```

```
$ ./configure --openblas-root=/foo
./configure: FATAL: '--openblas-root=/foo': switch must specify an existing directory. See './configure --help'.
```

#### Success

```
$ ./configure --mathlib=OPENBLAS
Configuring KALDI to use OPENBLAS.
Checking compiler g++ ...
Checking OpenFst library in /home/kkm/work/kaldi2/tools/openfst-1.6.9 ...
Checking cub library in /home/kkm/work/kaldi2/tools/cub-1.8.0 ...
Your math library seems to be OpenBLAS from /home/kkm/work/kaldi2/tools/OpenBLAS/install.  Configuring appropriately.
Configuring dynamically loaded OpenBlas since --static-math=no (the default)
Successfully configured for Linux with OpenBLAS from /home/kkm/work/kaldi2/tools/OpenBLAS/install
WARNING: CUDA will not be used! If you have already installed cuda drivers
         and CUDA toolkit, try using the --cudatk-dir= option. A GPU and CUDA
         are required to run neural net experiments in a realistic time.
INFO: Configuring Kaldi not to link with Speex. Don't worry, it's only needed if
      you intend to use 'compress-uncompress-speex', which is very unlikely.
Kaldi has been successfully configured. To compile:

  make -j clean depend; make -j <NCPU>

where <NCPU> is the number of parallel builds you can afford to do. If unsure,
use the smaller of the number of CPUs or the amount of RAM in GB divided by 2,
to stay within safe limits. 'make -j' without the numeric value  may not limit
the number of parallel jobs at all, and overwhelm even a powerful workstation,
since Kaldi build is highly parallelized.
```

danpovey:
Great.  While we're on it, would you mind updating the "example command lines" at the top of the `configure` file with whatever command lines you think people are most likely to want to use?

kkm000:
Done.

danpovey:
Thanks!  Merging.

PR #3268, opened 2019-04-25
Fixes to cuda unit tests.

luitjens:


src/cudamatrix/cu-matrix-test.cc
```diff
@@ -2924,6 +2941,11 @@ static void UnitTestCuMatrixEqualElementMask() {
 }
 
 template<typename Real> void CudaMatrixUnitTest() {
+#if HAVE_CUDA == 1
+  // Instantiate CuDevice here again in case we are running multi-threaded
+  // This is necessary to set streams for cuda libraries.
+  CuDevice::Instantiate();
```
danpovey: This concerns me a bit; it's not how the documentation says you are supposed to call CuDevice()::Instantiate().
It's supposed to be called whenever we call CUDA stuff automatically, as in:
```
#if HAVE_CUDA == 1
...
...
```
in places like cu-matrix.cc.  Can you remind me what the problem was that this fixed?

luitjens: Every thread must call CuDevice::Instantiate() prior to launching any work.  Since this is a thread entry point I added it there.  This follows the pattern we are using in our threaded pipeline.   Failure to have each thread call this means that their CuDevice is not instantiated.  So imagine a case where Thread0 allocates cudaMemory then Thread1 launches a cublas kernel on that memory.  In this case thread1 would launch work with an invalid cublas handle because it was never initialized.    Perhaps this would be done automatically behind the scenes but it isn't clear to me that is the case.   If you think it is then we should be able to safely delete those lines.  I added those at the beginning of my debugging of these tests and did not verify if they fixed any real issues.  I just figured it was better to be 100% certain that the device was initialized per thread.

danpovey:
All things that call CUDA code call that automatically-- or should call it,
at any rate.
I'd rather not have that in the test, as in theory it should be redundant.
Otherwise I would have to document its necessity and change other code
maybe.

luitjens:
> All things that call CUDA code call that automatically-- or should call it, at any rate. I'd rather not have that in the test, as in theory it should be redundant. Otherwise I would have to document its necessity and change other code maybe.

I have removed that code.

danpovey:
Thanks!  Merging.

PR #3274, opened 2019-04-26
fix deviation in UnitTestCuMatrixSetRandn()

DongjiGao:
fixed the allowed_deviation in UnitTestCuMatrixSetRandn() by calculating the exact deviation of x^n


danpovey:
Thanks, will merge when tests complete.

PR #3278, opened 2019-04-28
[Kaldi10] Update codes related to pdf-class to be 1-based.

hhadian:


danpovey:
Thanks a lot!  You might want to fix the documentation if you see something easy.  Merging.

hhadian:
Will do from now on.

PR #3281, opened 2019-04-30
[src] Log per-utt metrics in compute-wer --verbose=2

kkm000:
At verbose level 2, log per-utterance length, distance,
insertions, deletions and substitutions in a format that
is easy to parse out from the logs.

### DIFF
--- a/src/bin/compute-wer.cc
+++ b/src/bin/compute-wer.cc
@@ -20,10 +20,10 @@
 // limitations under the License.
 
 #include "base/kaldi-common.h"
-#include "util/common-utils.h"
-#include "util/parse-options.h"
 #include "tree/context-dep.h"
+#include "util/common-utils.h"
 #include "util/edit-distance.h"
+#include "util/parse-options.h"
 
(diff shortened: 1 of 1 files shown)


kkm000:
Just to explain the intent, this helps me find the outliers that do much better or much worse in a decoding set w.r.t. reference counts, to gain some hints what kind of acoustics the model likes or dislikes the most, thus what kind of data to throw into the training pool.

danpovey:
@jtrmal can you please comment on this?  Not sure if there is already another mechanism for this.  Not that that would preclude merging it necessarily.

jtrmal:
There is definitely a way how to do it without it -- it's done in the
native scoring scripts (kaldi-score-wer.sh),  (I think the script is called
per_utt_details.pl)
y.

kkm000:
Oh, did not know that, thanks much Yenda! Dan, I'll drop this PR myself is Yenda's suggestion cuts it for me, or comment if it does not for any reason.

danpovey:
Thanks!  Closing for now.

PR #3285, opened 2019-05-02
Cleanup filehandle usage in split_scp.pl

xsawyerx:
Hi,

These two commits clean up three issues with the use of file-handles:

* Remove the (majorly) insecure use of two-arg `open()`
* Use proper variables for filehandles (instead of what ends up being global variables that overwrite each other)
* Close filehandles that were forgotten empty

This does *not* change any of the interface or any of the user experience.

### DIFF
--- a/egs/wsj/s5/utils/split_scp.pl
+++ b/egs/wsj/s5/utils/split_scp.pl
@@ -90,16 +90,17 @@ if ($num_jobs == 0) { # without -j option
 }
 
 if ($utt2spk_file ne "") {  # We have the --utt2spk option...
-    open(U, '<', $utt2spk_file) || die "Failed to open utt2spk file $utt2spk_file";
-    while(<U>) {
+    open($u_fh, '<', $utt2spk_file) || die "Failed to open utt2spk file $utt2spk_file";
+    while(<$u_fh>) {
         @A = split;
         @A == 2 || die "Bad line $_ in utt2spk file $utt2spk_file";
         ($u,$s) = @A;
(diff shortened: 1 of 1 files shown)


danpovey:
OK, I'll merge, but in general I'm not super enthusiastic about these kinds of changes because there is  always a small chance of error.

xsawyerx:
I understand the concern, but I would like to share my perspective:

* We assume old code dies out, but it rarely actually does, especially for a large project. Kaldi has a *lot* of Perl code. It won't go away with the wave of a hand. If it sticks and it's used, it should be accounted for and taken care of.
* We also assume old code that works continues to do so forever. (Why fix if it ain't broken?) But that's not true either. Insecure code will eventually be picked up by someone who will see it and figure out how to abuse it. Customers will then suffer a new problem that could have been fixed, if one were to review and fix even old code.
* We also assume that there's no point in improving non-security issues, but I think that's also not true. Code can be made more readable, faster, more flexible, have more features, and be more readable and maintained by others.

Taking this script for instance: I only started playing with it because a colleague at work was trying to use Kaldi for a promising project, but couldn't understand what is happening in this script. It's a shame to prevent fixes and improvements to this code, because it *does* affect people, even if we can't see it.

I'm offering to help improve the state of these files for free. All I ask is that you dedicate some time to review them. I'm willing to push as small commits as possible (leaving any whitespace changes to the very end) and I'm willing to write tests for it to satisfy all requirements. Then, if you see value in new features in it, I would be happy to implement them as well.

Thanks for reading.

PR #3287, opened 2019-05-02
[scripts] Use correct compile-time regex syntax in split_scp.pl

xsawyerx:
This is a single commit with two character change. A simple cleanup.

The commit message:

Regular expressions should be written with // to allow the language to do compile-time checks for the regexp pattern.

This is from `perldoc perlop`:

    If the right argument is an expression rather than a search pattern,
    substitution, or transliteration, it is interpreted as a search pattern at
    run time. Note that this means that its contents will be interpolated
    twice, so

        '\\' =~ q'\\';

    is not ok, as the regex engine will end up trying to compile the pattern
    "\", which it will consider a syntax error.


danpovey:
OK, merging, but next time please try to batch things up if they are not to fix a bug or something.

PR #3289, opened 2019-05-02
fix ReadLexiconp index

rezame:


egs/wsj/s5/steps/dict/internal/prune_pron_candidates.py
```diff
@@ -98,8 +98,8 @@ def ReadLexiconp(lexiconp_handle):
         if len(splits) < 3:
             raise Exception('Invalid format of line ' + line
                                 + ' in lexicon file.')
-        word = splits[1]
-        prob = float(splits[0])
+        word = splits[0]
```
danpovey: Oh, I see, that does look plausible.   @xiaohui-zhang  can you comment?

danpovey:
from @xiaohui-zhang it looks like I shouldn't merge this.

PR #3296, opened 2019-05-06
[egs]try to fix an error about qsub in thchs30

zhang-ray:


danpovey:
sorry, we tend to leave those defaults as queue.pl, that file is supposed to be changed by users.
If you have run.pl where you have a queue, it fails in a less obvious way.

PR #3305, opened 2019-05-09
(WIP) Callhome VB Wrapper

HuangZiliAndy:
First version of VB wrapper. Improve x-vector performance from DER 8.02% to 6.48%. The result is a little bit worse than my former one(6.15%). The reason is that I added Mixer6 data set in my former experiment. (We can discuss whether to include that)


HuangZiliAndy:
Also, I am not very sure about the proper way to deal with python environment in Kaldi. Seems the BUT's code requires python 2 and install some packages.

danpovey:
It's OK to say python, python2 or python3 (if just `python` then must be
compatible with both versions).  Regarding extra package dependencies, you
will find scripts that tell the user to `pip install` things.  Maybe "git
grep" for that.

egs/callhome_diarization/v2/run.sh
```diff
@@ -356,3 +358,50 @@ if [ $stage -le 11 ]; then
...
+  diarization/convert_VB_model.sh --cmd "$train_cmd" exp/diag_ubm_$num_components/final.dubm \
+    exp/extractor_diag_c${num_components}_i${ivector_dim}/final.ie exp/VB 
+fi
+
+if [ $stage -le 13 ]; then
+  output_rttm_dir=exp/VB/rttm
+  mkdir -p $output_rttm_dir || exit 1;
+  cat $nnet_dir/xvectors_callhome1/plda_scores/rttm \
+    $nnet_dir/xvectors_callhome2/plda_scores/rttm > $output_rttm_dir/x_vector_rttm
+  init_rttm_file=$output_rttm_dir/x_vector_rttm
```
david-ryan-snyder: Is the initial RTTM file here the one obtained using the oracle number of speakers or the tuned stopping threshold? If it isn't, it should be the latter, I think.

egs/callhome_diarization/v1/diarization/kaldi_io.py
```diff
@@ -0,0 +1,627 @@
+#!/usr/bin/env python
```
david-ryan-snyder: I don't think diarization/ directory is the best place for this. It probably belongs somewhere else. Possibly the code should be downloaded from wherever its hosted, rather than added to our repository. @danpovey would you have a suggestion? I believe this is a general set of tools for handling Kaldi files in python. In this PR, it's used by BUT's VB resegmentation code.

egs/callhome_diarization/v1/diarization/convert_VB_model.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
```
david-ryan-snyder: In my opinion, it would be a better design to move the functionality of this script into VB_resegmentation.py and remove both this script and the bash script convert_VB_model.sh. That way, VB_resegmentation.py would take Kaldi models directly as input, and the conversion into numpy objects would happen internally after loading the objects.

In general, I think we should prefer that Kaldi scripts have interfaces that use Kaldi objects. The idea that you need to first convert a Kaldi object into a non-Kaldi object and dump it to disk before loading it into another Kaldi script doesn't feel right to me. Of course if there's a reason we really have to do it this way, it can be kept as is, but it doesn't seem necessary to me.

egs/callhome_diarization/v1/diarization/VB_resegmentation.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python
```
david-ryan-snyder: Does this script work with python3? I believe that we've decided to accept the dependency on python3. So, if possible, I think we should replace python with python3 at the top of the script.

egs/callhome_diarization/v2/run.sh
```diff
@@ -356,3 +358,50 @@ if [ $stage -le 11 ]; then
...
+  # Train the i-vector extractor. The UBM is assumed to be diagonal.
+  diarization/train_ivector_extractor_diag.sh \
+    --cmd "$train_cmd --mem 35G" \
+    --ivector-dim $ivector_dim --num-iters 5 --apply-cmn false \
+    --num-threads 1 --num-processes 1 --nj 40 \
+    exp/diag_ubm_$num_components/final.dubm data/train \
+    exp/extractor_diag_c${num_components}_i${ivector_dim}
+
+  # Converts diagonal UBM and ivector extractor to numpy array format
+  diarization/convert_VB_model.sh --cmd "$train_cmd" exp/diag_ubm_$num_components/final.dubm \
```
david-ryan-snyder: As I mentioned earlier, I think it's better if the conversion happens inside of VB_resegmentation.py. That way, the script VB_resegmentation.sh can continue using Kaldi objects in its interface.

egs/callhome_diarization/v2/run.sh
```diff
@@ -356,3 +358,50 @@ if [ $stage -le 11 ]; then
...
+  cat $nnet_dir/xvectors_callhome1/plda_scores/rttm \
+    $nnet_dir/xvectors_callhome2/plda_scores/rttm > $output_rttm_dir/x_vector_rttm
+  init_rttm_file=$output_rttm_dir/x_vector_rttm
+
+  # VB resegmentation. In this script, I use the x-vector result to 
+  # initialize the VB system. You can also use i-vector result or random 
+  # initize the VB system.
+  diarization/VB_resegmentation.sh --nj 20 --cmd "$train_cmd --mem 10G" \
+    --initialize 1 \
+    data/callhome $init_rttm_file exp/VB exp/VB/diag_ubm.pkl exp/VB/ie.pkl || exit 1;
```
david-ryan-snyder: As I mentioned earlier, I think these pkl should be regular Kaldi objections, and the conversions should happen internally.

david-ryan-snyder:
I think @danpovey answered my question already. I think we should probably have a script in extras that pip installs the kaldi_io.py script.

egs/callhome_diarization/v1/diarization/kaldi_io.py
david-ryan-snyder: @HuangZiliAndy you should probably make a script in tools/extras, e.g., tools/extras/install_kaldi_io_py.sh that installs this. Probably you can use install_tensorflow_py.sh as an example.

egs/callhome_diarization/v2/run.sh
```diff
@@ -53,7 +55,7 @@ if [ $stage -le 1 ]; then
   # callhome1 and callhome2.  Each partition is treated like a held-out
   # dataset, and used to estimate various quantities needed to perform
   # diarization on the other part (and vice versa).
-  for name in train callhome1 callhome2; do
+  for name in train callhome1 callhome2 callhome; do
```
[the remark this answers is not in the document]
HuangZiliAndy: The thing is that callhome1 and callhome2 are features for short segments. For the callhome dataset we want the features for the entire dataset.

egs/callhome_diarization/v2/run.sh
HuangZiliAndy: Yes, it is the later

egs/callhome_diarization/v2/run.sh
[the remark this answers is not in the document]
david-ryan-snyder: This is fine then. If there isn't already, it's a good idea to have a comment somewhere (e.g., when you're preparing the datasets) describe the difference between callhome{1,2} and callhome.

egs/callhome_diarization/v2/run.sh
david-ryan-snyder: OK thanks.

egs/callhome_diarization/v1/diarization/convert_VB_model.py
HuangZiliAndy: Yes, I feel the same way. It looks ugly for me too. Do you know any convenient way to load Kaldi objects(diagonal UBM and i-vector extractor in my case) directly as python dictionary? Perhaps that is much elegant. So I don't need to convert it to a text file first, which is extremely ugly. I guess Dan @danpovey might have good ideas about that?

danpovey: I suspect the amount of work required to change this would be disproportionate to the benefit.
This VB resegmentaion may  not even be something we rely on in future.  So I'd say, just let sleeping dogs lie for now.

HuangZiliAndy: Okay, thanks Dan.

HuangZiliAndy:
Hi, David. I think I have corrected things you mentioned above. Mainly (1) python issues. (it supports python3 now with some small modifications) (2) delete the step to convert the model into pickle file. Something I am not very sure. (1)the original result I got is a little bit different from yours. (I was using your model on web but the result I got is different which is quite weird since there is not something random) (2)Do you think it is necessary to add Mixer6 in the training set of VB to get better performance? (6.15% v.s. 6.48%) @david-ryan-snyder 

Thanks!
Zili

ZimingY:
> Hi, David. I think I have corrected things you mentioned above. Mainly (1) python issues. (it supports python3 now with some small modifications) (2) delete the step to convert the model into pickle file. Something I am not very sure. (1)the original result I got is a little bit different from yours. (I was using your model on web but the result I got is different which is quite weird since there is not something random) (2)Do you think it is necessary to add Mixer6 in the training set of VB to get better performance? (6.15% v.s. 6.48%) @david-ryan-snyder
> 
> Thanks!
> Zili

Thanks for the work! The VB branch is really helpful. I am wondering when the VB resegmentation will be released, and if the i-vector model that is used to define VB will be released together.

Thanks!

david-ryan-snyder:
@HuangZiliAndy, I wouldn't worry about the small difference in performance. It's probably due to minor choices in the datasets, but I certainly wouldn't worry about adding in a dependency on Mixer 6 over it.

@danpovey, I think this PR is good to go. Can you merge it when you have a chance?

tools/extras/install_kaldi_io.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Install kaldi_io. Please refer to https://github.com/vesis84/kaldi-io-for-python 
+# for details.
+
+python3 -m pip install --user kaldi_io
```
danpovey: Can somehow explain to me how this invocation makes kaldi_io available?  Does this mean that Karel has somehow registered that module with some official repo?

david-ryan-snyder: I don't know how it works, but it seems like Karel must've done something like that. If you copy and paste python3 -m pip install --user kaldi_io to the terminal it finds the package and starts installing it.

PR #3308, opened 2019-05-09
Fix utils/split_scp.pl breakage

### LINKED ISSUE #3306: utils/split_scp.pl broken
After b17fc8447, `utils/split_scp.pl -j 16 3 src-lats.scp` outputs the chunk 4 of 16 to a regular file named `-` instead of STDOUT.

The reason is the "old" 2-argument form `open FH, ">-"` used to open a handle to STDOUT, while the "modern Perl" `open $fh, '>', '-'` creates a real file named `-` instead.

kkm000:
Close #3306

### DIFF
--- a/egs/wsj/s5/utils/split_scp.pl
+++ b/egs/wsj/s5/utils/split_scp.pl
@@ -205,7 +205,7 @@ if ($utt2spk_file ne "") {  # We have the --utt2spk option...
             }
             $count == $scpcount[$scpidx] || die "Count mismatch [code error]";
         }
-        close(F);
+        close($f_fh);
     }
 } else {
    # This block is the "normal" case where there is no --utt2spk
--- a/egs/wsj/s5/utils/split_scp.pl
+++ b/egs/wsj/s5/utils/split_scp.pl
@@ -65,7 +65,8 @@ for ($x = 1; $x <= 3 && @ARGV > 0; $x++) {
     }
 }
 
-if ($num_jobs <= 0 || $job_id - $one_based < 0 || $job_id - $one_based >= $num_jobs) {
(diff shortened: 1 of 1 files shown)


kkm000:
Hold on, did not cleanly merge

kkm000:
This is ready to go now. I also cleaned up error messages, and fixed the -j M N argument check with and without one-based.

danpovey:
Thanks!  Note: this is a fixed version of #3285 which was reverted in #3307.

kkm000:
Yes, I noticed, I clashed with the revert :)

danpovey:
You sure, right?

kkm000:
I just used this in a practical script, and update my scripts to --one-based between two runs, both went out ok; I have thorough error checks in my scripts to make sure data counts all match after processing, and they do. I do not have a practical utt2spk-based case to test at the moment, though (although the next step, merging alignments and lattices, that we once talked on the list, likely will use it).

kkm000:
> You sure, right?

So pretty much yes, and I am likely to use it more a few times today. I'll be merging a complex dataset, with lattices coming from separate alignments, so if anything comes up, I'll notice.

danpovey:
Remind me to-morrow, I'll merge then.

kkm000:
@danpovey:
> Remind me to-morrow, I'll merge then.

Yup. You did the right thing. :) There is more weirdness, fixing now. I'll push when all modes work, and then let you know.

kkm000:
@danpovey, this flies! (and with the updated combine_ali_dirs, which comes tomorrow).

danpovey:
OK.  @kkm000, FYI I usually prefer to use filter_scp.pl using the appropriate utt2spk file, instead of split_scp.pl.   It probably doesn't make much difference though.

kkm000:
@dpovey, so can we merge it?

PR #3315, opened 2019-05-11
[scripts] Rig combine_ali_dirs.sh to combine alignment lattices

kkm000:
Relevant discussion:
https://groups.google.com/forum/#!topic/kaldi-help/2uxfByEAmfw

### DIFF
--- a/egs/wsj/s5/steps/combine_ali_dirs.sh
+++ b/egs/wsj/s5/steps/combine_ali_dirs.sh
@@ -1,105 +1,209 @@
 #!/bin/bash
 # Copyright 2016  Xiaohui Zhang  Apache 2.0.
+# Copyright 2019  SmartAction (kkm)
 
-# This srcipt operates on alignment directories, such as exp/tri4a_ali
-# the output is a new ali dir which has alignments from all the input ali dirs
+# This script combines alignment directories, such as exp/tri4a_ali, and
+# validates matching of the utterances and alignments after combining.
 
 # Begin configuration section.
 cmd=run.pl
-extra_files=
-num_jobs=4
+nj=4
+combine_lat=true
+combine_ali=true
+tolerance=10
 # End configuration section.
-echo "$0 $@"  # Print the command line for logging
+echo "$0 $@"  # Print the command line for logging.
 
-if [ -f path.sh ]; then . ./path.sh; fi
-. parse_options.sh || exit 1;
-
-if [[ $# -lt 3 ]]; then
-  echo "Usage: $0 [options] <data> <dest-ali-dir> <src-ali-dir1> <src-ali-dir2> ..."
-  echo "e.g.: $0 --num-jobs 32 data/train exp/tri3_ali_combined exp/tri3_ali_1 exp_tri3_ali_2"
-  echo "Options:"
-  echo " --extra-files <file1 file2...>   # specify addtional files in 'src-ali-dir1' to copy"
-  echo " --num-jobs <nj>                  # number of jobs used to split the data directory."
-  echo " Note, files that don't appear in the first source dir will not be added even if they appear in later ones."
-  echo " Other than alignments, only files from the first src ali dir are copied."
-  exit 1;
-fi
-
-data=$1;
-shift;
-dest=$1;
-shift;
-first_src=$1;
-
-mkdir -p $dest;
-rm $dest/{ali.*.gz,num_jobs} 2>/dev/null
-
-cp $first_src/phones.txt $dest 2>/dev/null
+[[ -f path.sh ]] && . ./path.sh
+. parse_options.sh || exit 1
 
 export LC_ALL=C
 
-for dir in $*; do
-  if [ ! -f $dir/ali.1.gz ]; then
-    echo "$0: check if alignments (ali.*.gz) are present in $dir."
-    exit 1;
+if [[ $# -lt 3 ]]; then
+  cat >&2 <<EOF
+Usage: $0 [options] <data> <dest-dir> <src-dir1> <src-dir2> ...
+ e.g.: $0 --nj 32 data/train exp/tri3_ali_combined exp/tri3_ali_1 exp_tri3_ali_2
+Options:
+ --nj <nj>              # number of jobs to split combined archives [4]
+ --combine_ali false    # merge ali.*.gz if present [true]
+ --combine_lat false    # merge lat.*.gz if present [true]
+ --tolerance <int,%>    # maximum percentage of missing alignments or lattices
+                        # w.r.t. total utterances in <data> before error is
+                        # reported [10]
+
+The script checks that certain important files are present and compatible in all
+source directories (phones.txt, tree); other are copied from the first source
+(cmvn_opts, final.mdl) without much checking.
+
+Both --combine_ali and --combine_lat are true by default, but the script
+proceeds with a warning if directories do not contain either alignments or
+alignment lattices. Check for files ali.1.gz and/or lat.1.gz in the <dest-dir>
+after the script completes if additional programmatic check is required.
+EOF
+  exit 1;
+fi
+
+if [[ ! $combine_lat && ! $combine_ali ]]; then
+  echo "$0: at least one of --combine_lat and --combine_ali must be true"
+  exit 1
+fi
+
+data=$1
+dest=$2
+shift 2
+first_src=$1
+
+do_ali=$combine_ali
+do_lat=$combine_lat
+
+# Check if alignments and/or lattices are present. Since we combine both,
+# whichever present, issue a warning only. Also verify that the target is
+# different from any source; we cannot combine in-place, and a lot of damage
+# could result.
+for src in $@; do
+  if [[ "$(cd 2>/dev/null -P -- "$src" && pwd)" = \
+        "$(cd 2>/dev/null -P -- "$dest" && pwd)" ]]; then
+    echo "$0: error: Source $src is same as target $dest."
+    exit 1
+  fi
+  if $do_ali && [[ ! -f $src/ali.1.gz ]]; then
+    echo "$0: warning: Alignments (ali.*.gz) are not present in $src, not" \
+         "combining. Consider '--combine_ali false' to suppress this warning."
+    do_ali=false
+  fi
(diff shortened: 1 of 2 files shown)


danpovey:
Cool.  How much testing have you done?

kkm000:
I merged ali and lats from two and three directories, and then extracted scps from 11-way split .gz arks and verified (by eyeballing) that the split was on a speaker boundary, and that the total numbers matched. I did not run failure cases though (with missing alignments). Can do that.

egs/wsj/s5/steps/combine_ali_dirs.sh
```diff
@@ -1,105 +1,209 @@
...
+  # Assign all source gzipped archive names to an exported variable, one each
+  # per source directory, so that we can copy archives in a job per source.
+  src_id=0
+  for src in $@; do
+    let src_id++
+    nj_src=$(cat $src/num_jobs) || exit 1
+    # Create and export variable src_arcs_${src_id} for the job runner.
+    # Each numbered variable will contain the list of archives, e. g.:
+    # src_arcs_1="exp/tri3_ali/ali.1.gz exp/tri3_ali/ali.1.gz ..."
+    # ('printf' repeats its format as long as there are more arguments).
```
danpovey: I don't see the -v option for printf on my mac.

kkm000: Should be bash 3.2? I try to verify all stuff on docker with it, checks out here. Could you please try if `printf -v foo 'bar'` works?
```
$ docker run --rm -it bash:3.2
bash-3.2# printf -v foo 'bar'
bash-3.2# echo $foo
bar
```

kkm000:
Tested with some alignment missing too (the original half-sets are both 16-job):
```
$ steps/combine_ali_dirs.sh --nj 11 data/train/vcogs/train_650 exp/vcogs-sr1/chain/ali_train_650 exp/vcogs-sr0.70/chain/ali_train_650 exp/vcogs-sr1.00/chain/ali_train_650
steps/combine_ali_dirs.sh --nj 11 data/train/vcogs/train_650 exp/vcogs-sr1/chain/ali_train_650 exp/vcogs-sr0.70/chain/ali_train_650 exp/vcogs-sr1.00/chain/ali_train_650
steps/combine_ali_dirs.sh: note: Temporary directory exp/vcogs-sr1/chain/ali_train_650/temp.epruSO will not be deleted in case of script failure, so you could examine it for troubleshooting.
steps/combine_ali_dirs.sh: Gathering alignments from each source directory.
steps/combine_ali_dirs.sh: Splitting combined alignments into 11 archives on speaker boundary.
steps/combine_ali_dirs.sh: Combined 342468 alignments for 342468 utterances.
steps/combine_ali_dirs.sh: There were 0 alignments not matching any utterance.
steps/combine_ali_dirs.sh: There were 0 utterances (0%) without alignments.
steps/combine_ali_dirs.sh: Gathering lattices from each source directory.
steps/combine_ali_dirs.sh: Splitting combined lattices into 11 archives on speaker boundary.
steps/combine_ali_dirs.sh: Combined 342468 lattices for 342468 utterances.
steps/combine_ali_dirs.sh: There were 0 lattices not matching any utterance.
steps/combine_ali_dirs.sh: There were 0 utterances (0%) without lattices.
steps/combine_ali_dirs.sh: Stored combined alignments and lattices in exp/vcogs-sr1/chain/ali_train_650
$ echo 15 >exp/vcogs-sr1.00/chain/ali_train_650/num_jobs
$ steps/combine_ali_dirs.sh --nj 11 --combine-lat false data/train/vcogs/train_650 exp/vcogs-sr1/chain/ali_train_650 exp/vcogs-sr0.70/chain/ali_train_650 exp/vcogs-sr1.00/chain/ali_train_650
steps/combine_ali_dirs.sh --nj 11 --combine-lat false data/train/vcogs/train_650 exp/vcogs-sr1/chain/ali_train_650 exp/vcogs-sr0.70/chain/ali_train_650 exp/vcogs-sr1.00/chain/ali_train_650
steps/combine_ali_dirs.sh: note: Temporary directory exp/vcogs-sr1/chain/ali_train_650/temp.2yVs5W will not be deleted in case of script failure, so you could examine it for troubleshooting.
steps/combine_ali_dirs.sh: Gathering alignments from each source directory.
steps/combine_ali_dirs.sh: Splitting combined alignments into 11 archives on speaker boundary.
steps/combine_ali_dirs.sh: Combined 342468 alignments for 331766 utterances.
steps/combine_ali_dirs.sh: There were 0 alignments not matching any utterance.
steps/combine_ali_dirs.sh: There were 10702 utterances (3%) without alignments.
steps/combine_ali_dirs.sh: Stored combined alignments in exp/vcogs-sr1/chain/ali_train_650
```

danpovey:
Would you mind combining the diagnostics so they take up fewer lines?
My bash version:

bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.

printf -v foo 'bar' prints nothing for me on mac

Dan

egs/wsj/s5/steps/combine_ali_dirs.sh
```diff
@@ -1,105 +1,209 @@
...
+do_combine() {
+  local ark=$1 entities=$2 copy_program=$3
+  shift 3
+
+  echo "$0: Gathering $entities from each source directory."
+  # Assign all source gzipped archive names to an exported variable, one each
+  # per source directory, so that we can copy archives in a job per source.
+  src_id=0
+  for src in $@; do
+    let src_id++
```
danpovey: can you just do `src_id=$[src_id+1]`, which  everyone knows?  Sometimes you are a little *too* knowledgeable.

kkm000:
> Would you mind combining the diagnostics so they take up fewer lines?

You mean the last 3 lines? Sure.

> printf -v foo 'bar' prints nothing for me on mac

That's a good sign :) And 'echo $foo' prints 'bar' after it, just to be completely sure nothing weird is going on?

> GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)

My docker image for bash 3.2 has the same version. Weird, but its help does mention -v

```
bash-3.2# help printf
printf: printf [-v var] format [arguments]
    printf formats and prints ARGUMENTS under control of the FORMAT. FORMAT
    is a character string which contains three types of objects: plain
    characters, which are simply copied to standard output, character escape
    sequences which are converted and copied to the standard output, and
    format specifications, each of which causes printing of the next successive
    argument.  In addition to the standard printf(1) formats, %b means to
    expand backslash escape sequences in the corresponding argument, and %q
    means to quote the argument in a way that can be reused as shell input.
    If the -v option is supplied, the output is placed into the value of the
    shell variable VAR rather than being sent to the standard output.
bash-3.2# bash --version
GNU bash, version 3.2.57(1)-release (x86_64-pc-linux-musl)
Copyright (C) 2007 Free Software Foundation, Inc.
```

egs/wsj/s5/steps/combine_ali_dirs.sh
kkm000: Is `src_id=$((src_id+1))` ok? The `$[...]` syntax for arithmetic is deprecated as of bash 4.

danpovey:
Yes, (()) is OK.
RE the -v thing, yes it does print bar.  But remember we don't have an
explicit version dependency with bash (AFAIK), so we shouldn't use
version-dependent features, within reason of course.

kkm000:
> we don't have an explicit version dependency with bash

I do not think bash earlier than 3.2 exists in the wild any more, and 3.2 is only on the Mac. And configure checks for it.

```
[[ $BASH_VERSION < '3.2' ]] && {
  echo >&2 "bash version ${BASH_VERSION} is too old, cannot continue." \
           "You won't be able to run Kaldi recipes with it anyway." \
           "Please upgrade. bash version 3.2 or higher is required."
  exit 1;
}
```

danpovey:
But please work with this for a while, and remind me in a week or so, I don't think it's super urgent.  I'd rather wait and see if any problem comes up than merge right away.

kkm000:
This script is actually used only once in the whole egs/ dir:
```
$ find egs/ -type f | xargs grep 'combine_ali_dirs.sh'
egs/aspire/s5/local/multi_condition/run_nnet2_common.sh:  steps/combine_ali_dirs.sh data/train_rvb exp/tri5a_rvb_ali $ali_dirs || exit 1;
```
Not very low-level.

kkm000:
Arithmetics in `$(( ))`, diagnostic made more compact:
```
$ steps/combine_ali_dirs.sh --nj 16 data/train/vcogs8/train_650 exp/vcogs-temp/chain/ali_train_650 exp/vcogs-sr0.80/chain/ali_train_650 exp/vcogs-sr1.00/chain/ali_train_650
steps/combine_ali_dirs.sh --nj 16 data/train/vcogs8/train_650 exp/vcogs-temp/chain/ali_train_650 exp/vcogs-sr0.80/chain/ali_train_650 exp/vcogs-sr1.00/chain/ali_train_650
steps/combine_ali_dirs.sh: note: Temporary directory exp/vcogs-temp/chain/ali_train_650/temp.AFs6Gb will not be deleted in case of script failure, so you could examine it for troubleshooting.
steps/combine_ali_dirs.sh: Gathering alignments from each source directory.
steps/combine_ali_dirs.sh: Splitting combined alignments into 16 archives on speaker boundary.
steps/combine_ali_dirs.sh: Combined 342468 alignments for 342468 utterances. There were 0 utterances (0%) without alignments, and 0 alignments not matching any utterance.
steps/combine_ali_dirs.sh: Gathering lattices from each source directory.
steps/combine_ali_dirs.sh: Splitting combined lattices into 16 archives on speaker boundary.
steps/combine_ali_dirs.sh: Combined 342468 lattices for 342468 utterances. There were 0 utterances (0%) without lattices, and 0 lattices not matching any utterance.
steps/combine_ali_dirs.sh: Stored combined alignments and lattices in exp/vcogs-temp/chain/ali_train_650
```

@danpovey, I'll be using this script a few times during the week. We can let it stew for a while to make sure, if you prefer. Only all its uses are going to be boringly similar here, not the best for testing.

danpovey:
OK thanks, will merge now.

Issue #3303, opened 2019-05-08
Generate utt2num_frames with features


kkm000:
From kaldi-help, ["Purpose and use of utt2dur"](https://groups.google.com/forum/#!topic/kaldi-help/vyEL7qtXmhk):

@kkm000:
> I always carry [the `utt2dur`] file, compute it carefully to the fifth decimal or have a script compute it, but I just realized I do not understand how it is used and why is it needed.

@danpovey:
> Hm. It's needed by certain scripts.  Sometimes it's used as a relatively fast way to generate `utt2num_frames`, which is used by the nnet training scripts for purposes like deciding how many egs files to create.  However it's recommended to generate `utt2num_frames` directly when you dump the features, e.g. MFCC or PLP, as that code already knows and doesn't have to do big disk I/O to find out the info.
> We should probably change the default in the `make_*_feats.sh` scripts to generate those files by default, if it's not done already.  I would merge the PR.
> There may possibly be other things that utt2dur is used for, e.g. maybe in diarization.

@kkm000:
>D-oh. Many recipes generate it, sometimes at a huge expense of running the full pipeline (e. g. LibriSpeech which uses flac apparently pumps all 1k hours twice; make_utt2dur.sh only supports wav and sphere directly).
> I am preparing a dataset right at the moment; I'll run through with utt2num_frames and without utt2dur and fix what breaks, if anything.

kkm000:
@danpovey, quick questions if/when you have time.

1. `make_utt2dur.sh` converts `utt2num_frames` only as the last resort, i. e. only in case `wav.scp` is not present; otherwise it may run the whole pipeline. Is it sensible to prefer creating `utt2dur` from  `utt2num_frames`, if it is present? The only downside I see is the lower precision, down to frame shift, i. e. 10ms typically. Should be acceptable, I think?

2. Related to feature extraction in general. When I create data directories, I always place a copy the feature config file that was used to extract features along feats.scp, under the name matching feature type (e. g. `mfcc.conf`, even if conf/mfcc_hires.conf was used). Helps me a ton to avoid a mess-up of variously extracted features. Do you think we should just always to that? 

3. `utils/data/get_frame_shift.sh` first blindly creates `utt2dur` (expensive!), then checks if the file `frame_shift` exists. I'll rearrange the order, indeed, but a better solution seems to always create the file `frame_shift` along with features?

PR #3316, opened 2019-05-11
[src,scripts]Store frame_shift,utt2{dur,num_frames}, .conf with features

kkm000:
~~WIP because I'll change other `make_<feat>` steps in the same manner after this is reviewed.~~ @danpovey, please let me know what do you think about this approach. Together with #3313 this ensures that `get_utt2dur.sh` does not touch `wav.scp` if features have been already extracted, and completes in a second if invoked. It is in fact not called by any script from `train.py`; `get_frame_shift.sh` and `get_utt2num_frames.sh` are, and they already have data available, and do nothing lengthy. I just ran the full training of a chain model, and it did it as expected with this change.

Storing `mfcc.conf` in the feature directory is tangential to this change, but I found it very helpful when working with a collection of differently extracted feature directories.

**ADD:** Actually, I think I'll change the perl re to `/^--frame-shift=(\d+)/` . It's easier to read, and, since compute-mfcc-feats had completed successfully by that point, it's safe to assume the switch was not malformed, i. e. the value was a string of digits.

 --8<-- commit description follows --8<--

[src,scripts]Store frame_shift,utt2{dur,num_frames}, .conf with features

Generate utt2dur and utt2num_frames during feature extraction,
and store frame period in frame_shift file in feature directory.

Copy relevant .conf files used in feature extraction into
the conf/ subdirectory with features.

Add missing validations and options in some extraction scripts.

Close: #3303.

### DIFF
--- a/egs/wsj/s5/steps/make_mfcc.sh
+++ b/egs/wsj/s5/steps/make_mfcc.sh
@@ -10,23 +10,28 @@ nj=4
 cmd=run.pl
 mfcc_config=conf/mfcc.conf
 compress=true
-write_utt2num_frames=false  # if true writes utt2num_frames
+write_utt2num_frames=true  # If true writes utt2num_frames.
+write_utt2dur=true
 # End configuration section.
 
-echo "$0 $@"  # Print the command line for logging
+echo "$0 $@"  # Print the command line for logging.
 
 if [ -f path.sh ]; then . ./path.sh; fi
 . parse_options.sh || exit 1;
 
 if [ $# -lt 1 ] || [ $# -gt 3 ]; then
-   echo "Usage: $0 [options] <data-dir> [<log-dir> [<mfcc-dir>] ]";
-   echo "e.g.: $0 data/train exp/make_mfcc/train mfcc"
-   echo "Note: <log-dir> defaults to <data-dir>/log, and <mfccdir> defaults to <data-dir>/data"
-   echo "Options: "
-   echo "  --mfcc-config <config-file>                      # config passed to compute-mfcc-feats "
-   echo "  --nj <nj>                                        # number of parallel jobs"
-   echo "  --cmd (utils/run.pl|utils/queue.pl <queue opts>) # how to run jobs."
-   echo "  --write-utt2num-frames <true|false>     # If true, write utt2num_frames file."
+  cat >&2 <<EOF
+Usage: $0 [options] <data-dir> [<log-dir> [<mfcc-dir>] ]
+ e.g.: $0 data/train
+Note: <log-dir> defaults to <data-dir>/log, and
+      <mfcc-dir> defaults to <data-dir>/data.
+Options:
+  --mfcc-config <config-file>          # config passed to compute-mfcc-feats.
+  --nj <nj>                            # number of parallel jobs.
+  --cmd <run.pl|queue.pl <queue opts>> # how to run jobs.
+  --write-utt2num-frames <true|false>  # If true, write utt2num_frames file.
+  --write-utt2dur <true|false>         # If true, write utt2dur file.
+EOF
    exit 1;
 fi
 
@@ -63,10 +68,11 @@ required="$scp $mfcc_config"
 
 for f in $required; do
   if [ ! -f $f ]; then
-    echo "make_mfcc.sh: no such file $f"
+    echo "$0: no such file $f"
     exit 1;
   fi
 done
+
 utils/validate_data_dir.sh --no-text --no-feats $data || exit 1;
 
 if [ -f $data/spk2warp ]; then
@@ -92,11 +98,16 @@ else
   write_num_frames_opt=
 fi
 
+if $write_utt2dur; then
+  write_utt2dur_opt="--write-utt2dur=ark,t:$logdir/utt2dur.JOB"
+else
+  write_utt2dur_opt=
+fi
 
 if [ -f $data/segments ]; then
   echo "$0 [info]: segments file exists: using that."
 
-  split_segments=""
+  split_segments=
   for n in $(seq $nj); do
     split_segments="$split_segments $logdir/segments.$n"
   done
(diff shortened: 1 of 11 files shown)


egs/wsj/s5/steps/make_mfcc.sh
```diff
@@ -151,6 +151,12 @@ if $write_utt2num_frames; then
   rm $logdir/utt2num_frames.*
 fi
 
+# Store frame_shift and mfcc_config along with feats.scp.
+frame_shift=$(perl -ne 'if (/^--frame-shift=(.*?)([\s#]|$)/) {
+                          printf "%.3f", 0.001 * $1; exit; }' $mfcc_config)
+echo ${frame_shift:-'0.01'} > $data/frame_shift
+cp $mfcc_config $data/mfcc.conf
```
danpovey: If we are to store the config I'd prefer to put it in $data/conf/{mfcc,plp,mfcc_pitch}.conf, or whatever.  We should keep the scripts in sync (although we rarely use theothers).

kkm000: > If we are to store the config I'd prefer to put it in $data/conf/{mfcc,plp,mfcc_pitch}.conf, or whatever.

Will do.

> We should keep the scripts in sync (although we rarely use the others).

Sure, I wanted to run the approach by you first, this is why this change is only to one of them. So, are all 3 changes ok by you: 1) create utt2num_frames by default (this is what you suggested, so I guess it is); 2) pull frame_shift from config, or store the default; and 3) store the configs with the feature dataset (but in a conf/ subdirectory)?

danpovey:
Yes, all that is OK.  Put frame_shift in conf/ too.  Probably converting to
seconds would be  best.  You could make the script that figures out the
frame shift from the config a separate python script if you want.  BTW this
will allow us to simplify the script that figures out the frame shift.  (Or
at least give it a much easier and faster way to get it before trying the
other options).

kkm000:
> Put frame_shift in conf/ too

There are scripts already that put it/look for it in the $data. I can try to find them all, but that's already dangerous.

> Probably converting to seconds would be best.

Yes, it's already in seconds.

> BTW this will allow us to simplify the script that figures out the frame shift. (Or at least give it a much easier and faster way to get it before trying the other options).

Yes, I rearranged the checks in #3313. That script was doing it in a weird order (first called get_utt2dur always, then failed if feats.scp did not exist, and last checked if frame_shift is already there :) ).

danpovey:
From grepping around I couldn't find any such script.  Would likely be a
local script, since if mfcc.conf were officially supported, copy_data.sh
would support it.

kkm000:
We are talking about frame_shift, not mfcc.conf?

danpovey:
Oh you're right, don't move frame_shift if it already exists.
The reason for the conf/ thing is it will simplify finding whether such a
thing exists, and copying it, since the name is not known in advance.

kkm000:
```
$ grep -nH '/frame_shift' {utils,steps}/**/*.{sh,py,pl}
utils/data/get_frame_shift.sh:30:if [[ -s $dir/frame_shift ]]; then
utils/data/get_frame_shift.sh:31:  cat $dir/frame_shift
utils/data/get_frame_shift.sh:68:echo $frame_shift > $dir/frame_shift
utils/data/get_utt2dur.sh:44:elif [[ -s $data/frame_shift && -f $data/utt2num_frames ]]; then
utils/data/get_utt2dur.sh:46:  frame_shift=$(cat $data/frame_shift) || exit 1
utils/data/get_utt2dur.sh:110:  if [[ -s $data/frame_shift ]]; then
utils/data/get_utt2dur.sh:111:    frame_shift=$(cat $data/frame_shift) || exit 1
utils/data/get_utt2dur.sh:112:    echo "$0: overriding frame_shift=$frame_shift from file $data/frame_shift"
utils/data/subsegment_data_dir.sh:144:  if [ ! -s $srcdir/frame_shift ]; then
utils/data/subsegment_data_dir.sh:147:    frame_shift=$(cat $srcdir/frame_shift)
steps/conf/get_ctm_conf.sh:53:if [ -f $dir/../frame_shift ]; then
steps/conf/get_ctm_conf.sh:54:  frame_shift_opt="--frame-shift=$(cat $dir/../frame_shift)"
steps/conf/get_ctm_conf.sh:55:  echo "$0: $dir/../frame_shift exists, using $frame_shift_opt"
steps/get_ctm_conf.sh:53:if [ -f $dir/../frame_shift ]; then
steps/get_ctm_conf.sh:54:  frame_shift_opt="--frame-shift=$(cat $dir/../frame_shift)"
steps/get_ctm_conf.sh:55:  echo "$0: $dir/../frame_shift exists, using $frame_shift_opt"
steps/make_mfcc.sh:157:echo ${frame_shift:-0.01} > $data/frame_shift
steps/nnet3/decode_score_fusion.sh:225:if [ -f $srcdir/frame_shift ]; then
steps/nnet3/decode_score_fusion.sh:226:  cp $srcdir/frame_shift $dir/../
steps/nnet3/decode_score_fusion.sh:227:  echo "$0: $srcdir/frame_shift exists, copy $srcdir/frame_shift to $dir/../"
```

kkm000:
> The reason for the conf/ thing is it will simplify finding whether such a thing exists, and copying it, since the name is not known in advance.

Ah, right, makes sense, did not think about that! There are scripts like subset_data_dir, combine_data, etc.

kkm000:
@dpovey, could you please comment on the changes to make_mfcc.sh and compute-mfcc-feat.cc in the changeset 571b886 on this branch? This is where I added writing utt2dur by the binary itself, and using it in the script. I think that both write_utt2dur and write_utt2num_frames should default to true, is this ok by you?

kkm000:
@dpovey, done. I actually RAN THEM ALL, sorry for the shouty case, with and without segments. A bug fix for extract-segments comes in a separate PR.

danpovey:
Thanks a lot!!  I'll merge this.

PR #3319, opened 2019-05-13
Updated make_musan.* so that the desired sample rate can be specified as an input argument

osadj:


danpovey:
Perhaps @david-ryan-snyder  or @phanisankar-nidadavolu would have a comment.  I believe we now have a shared version of this script that supports the sampling rate option, but perhaps it is not merged yet?  If a PR is ready to merge I hope people will remind me.

phanisankar-nidadavolu:
@osadj In the latest commit make_musan.sh and make_musan.py are moved to steps/data and can accept sampling rate as optional argument. Please check those scripts.

osadj:
Thanks. The latest commit was pushed 2 hrs ago. I had sent the PR in the morning. Also, please update 'sre16/v2' egs to reflect this change.

osadj:
@phanisankar-nidadavolu I think it is much cleaner if you use 16000 as the default sampling rate (as opposed to None), in make_musan.py

danpovey:
Yeah, seems a valid point.

phanisankar-nidadavolu:
@osadj Thank you for the suggestion. I will make the change.

danpovey:
I am guessing we can close this as it's now reflected in what Phani committed.  I can re-open this if you say I'm wrong.

osadj:
Yes. Thanks.

PR #3321, opened 2019-05-14
[src] Fix sample rounding errors in extract-segments

kkm000:
With a segments file constructed from exact wave file durations
some segments came out one sample short. The reason is the
multiplication of the float sample frequency and double audio
time point is inexact. For example, float 8000.0 multiplied
by double 2.03 yields 16239.99999999999, one LSB short of the
correct sample number 16240.

Also changed all endpoint calculations so that they performed
in seconds, not sample numbers, as this does not require a
conversion in nearly every comparison, and report positions in
diagnostic messages also in seconds, not sample numbers.

### DIFF
--- a/src/featbin/extract-segments.cc
+++ b/src/featbin/extract-segments.cc
@@ -70,18 +70,18 @@ int main(int argc, char *argv[]) {
 
     RandomAccessTableReader<WaveHolder> reader(wav_rspecifier);
     TableWriter<WaveHolder> writer(wav_wspecifier);
-    Input ki(segments_rxfilename);  // no binary argment: never binary.
+    Input ki(segments_rxfilename);  // No binary argment: never binary.
 
     int32 num_lines = 0, num_success = 0;
 
     std::string line;
-    /* read each line from segments file */
+    // Read each line from the segments file.
     while (std::getline(ki.Stream(), line)) {
       num_lines++;
       std::vector<std::string> split_line;
-      // Split the line by space or tab and check the number of fields in each
-      // line. There must be 4 fields--segment name , reacording wav file name,
-      // start time, end time; 5th field (channel info) is optional.
+      // Split the line into whitespace-separated fields and verify their
+      // number. There must be 4 or 5 fields: segment name, reacording ID, start
+      // time, end time, and the optional channel number.
       SplitStringToVector(line, " \t\r", true, &split_line);
       if (split_line.size() != 4 && split_line.size() != 5) {
         KALDI_WARN << "Invalid line in segments file: " << line;
@@ -92,8 +92,8 @@ int main(int argc, char *argv[]) {
           start_str = split_line[2],
           end_str = split_line[3];
 
-      // Convert the start time and endtime to real from string. Segment is
-      // ignored if start or end time cannot be converted to real.
+      // Parse the start and end times as float values. Segment is ignored if
+      // any of end times is malformed.
       double start, end;
       if (!ConvertStringToReal(start_str, &start)) {
         KALDI_WARN << "Invalid line in segments file [bad start]: " << line;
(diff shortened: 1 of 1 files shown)


kkm000:
```c
#include <stdio.h>
void main(void) {
  double f1 = 2.03;
  float f2 = 8000.0f;
  float f = f1 * f2;
  double d = f1 * f2;
  int n = f1 * f2;
  printf ("f:%.5f d:%.12lf n:%d\n", f, d, n);
}
```
outputs
```
f:16240.00000 d:16239.999999999998 n:16239
```

danpovey:
Hm OK, thanks.  Looks good to me, so merging.

kkm000:
@dpovey, actually, I was getting utt2dur mismatch with the segments file generated from the very same utt2dur file obtained without segments when tested the new scripts, i. e. segments in the format `utt-id utt-id 0 <file-length>`, transformed with the simple `print $1, $1, 0, $2` with awk from utt2dur, and `soxi` was giving me both the correct sample counts and the length in seconds. The only thing in between was extract-segments, so I figured it was at fault, but it was quite a surprising find! Even more interesting, 2.03 * 8000.0 is exact when both are floats, but in double precision it's one LSB too small. And 8000.0 in IEEE arithmetic is exact. I was a bit scared at first that it would have been our string-to-double conversion that was giving fits, but fortunately the repro C code above does not use strtod and friends, just compile-time constant, so the problem proved to be inherent to IEEE representation. Same with all compilers I compiled it with. Something to keep in mind; the `+ 0.5` thing when rounding to an int is always important.

I tested it on a larger dataset, and got no discrepancy any more. The original error rate was about 1 in 120 utterances.

danpovey:
Thanks for the testing!

alumae:
This doesn't always work properly. For example:
file length: 3180.67
segment start: 3174.98
segment end: 3180.67
sampling frequency: 16000
length of file in samples: 50890710

The code derives the following start_samp and end_samp: 50799680 50890711. The latter is greater than the file length.

alumae:
Continuing my last comment: this causes error:
    
    ASSERTION_FAILED (extract-segments[5.5.341~1-a32be]:SubMatrix():kaldi-matrix.cc:1647) Assertion failed: (static_cast<UnsignedMatrixIndexT>(ro) < static_cast<UnsignedMatrixIndexT>(M.num_rows_) && static_cast<UnsignedMatrixIndexT>(co) < static_cast<UnsignedMatrixIndexT>(M.num_cols_) && static_cast<UnsignedMatrixIndexT>(r) <= static_cast<UnsignedMatrixIndexT>(M.num_rows_ - ro) && static_cast<UnsignedMatrixIndexT>(c) <= static_cast<UnsignedMatrixIndexT>(M.num_cols_ - co))

danpovey:
Oops.  Can you please be more specific about the scenario here?  E.g. by
what path is it obtaining, say, utt2dur or utt2num_frames or segments
information, and how is that causing a crash (e.g. what script?)

danpovey:
Oh sorry @alumae , I see it's a code issue not a script issue.

src/featbin/extract-segments.cc
```diff
@@ -129,74 +129,70 @@ int main(int argc, char *argv[]) {
...
       }
-      /*
-       * This function  return a portion of a wav data from the orignial wav data matrix
-       */
-      SubMatrix<BaseFloat> segment_matrix(wave_data, channel, 1, start_samp, end_samp-start_samp);
+
+      // Convert endpoints of the segment to sample numbers. Note that the
+      // conversion requires a proper rounding.
+      int32 start_samp = static_cast<int32>(start * samp_freq + 0.5f),
+          end_samp = static_cast<int32>(end * samp_freq + 0.5f);
```
danpovey: I think just adding here:
```
    if (end_samp > num_samp) 
       end_samp = num_samp;
```
should fix the issue.  @alumae do you have time?  Or if anyone else has time to make a PR.

alumae:
Yes, I did the same fix myself. Will do a PR.

kkm000:
@alumae, thank you! How did not I catch this? I ran it by a 650 hour set with not a single discrepancy!

kkm000:
Holy Molly! The file duration as a double is 3180.669375000000, but rounded to BaseFloat=float value is 3180.669(4). Enough to introduce another LSB error. IEEE arithmetic loves me!

PR #3322, opened 2019-05-14
Initial version of Docker images for (CPU and GPU versions)

mdoulaty:
(I'll updated the README file to point to the Kaldi repo on DockerHub before merging)

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -6,6 +6,7 @@ docker run -it mdoulaty/kaldi:latest bash
 ``` 
 
 Sample usage of the GPU based images:
+
 Note: use [nvidia-docker](https://github.com/NVIDIA/nvidia-docker) to run the GPU images.
 
 ```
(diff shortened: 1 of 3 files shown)


galv:
@danpovey Yep, this looks fine. Again, though, I'd like to see how this can integrate with a continuous integration system. That way, the Dockerfiles don't become abandoned.

mdoulaty:
re the CI integration, I'll try DockerHub first (for CPU only images)
for the GPU-based images I think we need a VM with GPU to build the image there. Probably have to write some scripts that create a VM somewhere, build, tag and push the image and then tear down the env. Then all we need to do on the CI side is to call this script whenever something's merged
(at icassp this week, probably will setup dockerhub next week)

danpovey:
OK, since @galv says this is fine I will merge.  Since I don't fully understand all the issues around this, I hope @mdoulaty  will continue to be available to help with it.

mdoulaty:
yes, I'll be available to continue working on this
will work with @galv for the automation of builds

PR #3323, opened 2019-05-15
[scripts] handle frame_shift and utt2num_frames in utils/

kkm000:
subset_data_dir.sh has been refactored thoroughly so that its
logic can be followed easier. It has been well tested in all
modes, and dogfooded.

All changes here are necessary to subset, combine and verify
utt2num_frames, and copy frame_shift to new directories where
necessary.

### DIFF
--- a/egs/wsj/s5/utils/copy_data_dir.sh
+++ b/egs/wsj/s5/utils/copy_data_dir.sh
@@ -103,6 +103,9 @@ fi
 if [ -f $srcdir/utt2dur ]; then
   utils/apply_map.pl -f 1 $destdir/utt_map <$srcdir/utt2dur >$destdir/utt2dur
 fi
+if [ -f $srcdir/utt2num_frames ]; then
+  utils/apply_map.pl -f 1 $destdir/utt_map <$srcdir/utt2num_frames >$destdir/utt2num_frames
+fi
 if [ -f $srcdir/reco2dur ]; then
   if [ -f $srcdir/segments ]; then
     cp $srcdir/reco2dur $destdir/reco2dur
@@ -116,7 +119,7 @@ fi
 if [ -f $srcdir/cmvn.scp ]; then
   utils/apply_map.pl -f 1 $destdir/spk_map <$srcdir/cmvn.scp >$destdir/cmvn.scp
 fi
-for f in stm glm ctm; do
+for f in frame_shift stm glm ctm; do
   if [ -f $srcdir/$f ]; then
     cp $srcdir/$f $destdir
   fi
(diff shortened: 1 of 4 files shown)


kkm000:
I think that I found all cases where the files frame_shift and utt2num_frames should be filtered and transferred to the destination. The check in combine_data_dirs.sh is a little bit overzealous: I refuse to combine dirs with differing frame_shift where feats.scp are also present. But at that point. I do not know if feats.scp is present in all directories, and, if not, then frame_shit is pretty much irrelevant. However, I do not think this will be a real trouble, as the file normally is a result of feature extraction, and it's probably uncommon to combine directories where not all have feats.scp *and* also have a different frame_shift. So I went for a clarity of the script vs ultimate correctness, which would make the logic harder to read.

egs/wsj/s5/utils/copy_data_dir.sh
```diff
@@ -116,7 +119,7 @@ fi
 if [ -f $srcdir/cmvn.scp ]; then
   utils/apply_map.pl -f 1 $destdir/spk_map <$srcdir/cmvn.scp >$destdir/cmvn.scp
 fi
-for f in stm glm ctm; do
+for f in frame_shift stm glm ctm; do
   if [ -f $srcdir/$f ]; then
     cp $srcdir/$f $destdir
```
danpovey: how about copying the conf/ dir?  Could perhaps just copy one of them?

danpovey:
Thanks!  I guess my main question is, how much testing have you done?

egs/wsj/s5/utils/copy_data_dir.sh
danpovey: actually the conf dir thing is not important.  But how many different options have you tested with?

kkm000: It cannot be sensibly merged, the features can be potentially extracted differently.

> But how many different options have you tested with?

Short answer is quite exhaustively. Depends on the level of changes. For this script, basically tested the rejection if frame_shift is different, and that the file is actually copied when present somewhere (and d-oh, that initially failed, because I misspelled it). For the subset_data_dir, which I've heavily rearranged/rewritten, I had to test it exhaustively with all options (and actually used it already, but that does not cover much) -- this is a very commonly used one. Even with fake ctm files, which it also subsets, and with comments in them, because it keeps the comments; I also ran validate_data_dir on every subset it created. For validate_data_dir, I added an extra line by hand to utt2num_frames, then moved it out of order, then mangled an ints in the second columns in the third run. For simple cases where I simply had to added utt2num_frames besides utt2dir into the long list of files to be conditionally copied, basically did not do at all (do not remember if there are any changes this simple, tho).

I'm always very careful with stuff that does not have automated testing, really. Remember how I discovered the one-sample discrepancy in utt2dur extraction with and without segments the other day: I actually diffed the extracted files in the extraction scripts. All of them, and each without the segments and with a trivial segments file. Also, I've rewritten these extractions scripts such that when I did e.g. `diff make_mfcc.sh make_fbank.sh`, the only differences shown were related to the actual extractor changes. (This is where I resisted the urge to change $mfcc_dir in one and $fbank_dir in the other to $feats_dir. But maybe I should have). Then `diff make_mfcc.sh make_mfcc_pitch.sh`, and from there on `diff make_mfcc_pitch.sh make_everyother_pitch.sh`. Found some minor discrepancies this way. So... let's say I test our scrips like I would test something that has no automated test.

If only we had tests for these! If you look at the front page of Kaldi on GitHub, where it says what is the main language for the project (near the long colored horizontal bar), you'll notice that GitHub thinks Kaldi is written in bash. I was thinking about writing at least some tests, for most commonly used scripts at the very least, but this is an immense task. Just try to think of exhaustive test cases for validate_data_dir...

I quickly looked into bash testing frameworks, they'd help, but the majority of a test it still creating test data and validating it, and this is all manual. If we only head two of me...

danpovey:
OK, thanks for all your work.  I'll merge.

PR #3324, opened 2019-05-15
fixed an error while creating musan directory

phanisankar-nidadavolu:


kkm000:
Thanks, a good catcch!

kkm000:
Looks like this has already been fixed a few minutes ago, but thanks anyway! : )

PR #3325, opened 2019-05-15
A new recipe on aidatatang_200zh corpus.

wly112:
We have exposed a speech corpus named "aidatatang_200zh" on openslr recently, and this recipe is conducted to demonstrate that this corpus is a reasonable data resource for Chinese Mandarin speech recognition research. We hope the corpus could help much more researchers in Chinese speech recognition.


danpovey:
You seem to have duplicated the steps/ dir instead of making a soft link.  Perhaps that's why you closed the PR...

wly112:
Yes, I have fixed this problem by replacing the directory with corresponding soft link. And @DatatangAI launched another new PR number #3326 on this issue.

PR #3331, opened 2019-05-17
[src] Fix rare case when segment end rounding overshoots file end in …

alumae:
…extract-segments


danpovey:
Fixes issue in #3321

PR #3334, opened 2019-05-20
change default of --modify-spk-id from True to False. 

phanisankar-nidadavolu:
If this change is not make spk-id scripts should be updated by adding  --modify-spk-id 'false' to augmentation args.


danpovey:
Back-compatibility fix for #3119

PR #3337, opened 2019-05-22
resolve  segmentation code  issues in python3

jtrmal:
I'm trying to segment half an hour recordings from TV and it's giving me some problems in case of long silences (or music) where no word is recognized.
I'm hoping @vimalmanohar  and perhaps @danpovey  will be able to look at the fixes if they can make sense. I'm particularly unsure about the changes in the clustering algorithm. The error messages weren't really useful in figuring out what's wrong.

### DIFF
--- a/egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
+++ b/egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
@@ -136,7 +136,7 @@ def wer(ctm_edit_lines):
         return float('inf')
     if num_words == 0 and num_incorrect_words == 0:
         return 0
-    return (float(num_incorrect_words) / num_words, -num_words)
+    return float(num_incorrect_words) / num_words
 
 
 def choose_best_ctm_lines(first_lines, second_lines,
@@ -144,9 +144,9 @@ def choose_best_ctm_lines(first_lines, second_lines,
     """Returns ctm lines that have lower WER. If the WER is the lines with
     the higher number of words is returned.
     """
-    i, best_lines = min((0, first_lines), (1, second_lines),
+    i, best_lines = min((0, first_lines),
+                        (1, second_lines),
                         key=lambda x: wer(x[1]))
-
     return i
 
 
@@ -308,7 +308,7 @@ def run(args):
         try:
             if len(ctm_edits_for_reco) == 0:
                 logger.warn('CTMs for recording %s is empty.',
-                             reco)
+                            reco)
                 continue   # Go to the next recording
 
             # Process CTMs in the recordings
--- a/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py
+++ b/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py
@@ -1331,8 +1331,7 @@ class SegmentsMerger(object):
                     if reject:
                         rejected_clusters.add(tuple(new_cluster))
                         continue
-
-                    heapq.heappush(heap, (-scoring_function(merged_segment),
+                    heapq.heappush(heap, ((-scoring_function(merged_segment), i),
                                           (merged_segment, i, new_cluster)))
 
                 candidate_index = -1
@@ -1527,7 +1526,7 @@ def get_segments_for_utterance(split_lines_of_utt, args, utterance_stats):
         _global_logger.debug("stage 0: segment %d = %s", i, x)
 
     if args.verbose > 4:
-        print ("Stage 0 [segment cores]:", file=sys.stderr)
+        print("Stage 0 [segment cores]:", file=sys.stderr)
         segments_copy = [x.copy() for x in segments]
         print_debug_info_for_utterance(sys.stderr,
                                        copy.deepcopy(split_lines_of_utt),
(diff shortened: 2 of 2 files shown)


danpovey:
Thanks... can you please mention what the original problem was?

jtrmal:
I will send it tomorrow, but it was very cryptic and not really related to
the issue at hand - something about operation > undefined for given type.

danpovey:
Sounds like it could be a python3 issue.

jtrmal:
Hm, that's a good point,let me check this tomorrow first.

egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
```diff
@@ -144,10 +144,14 @@ def choose_best_ctm_lines(first_lines, second_lines,
...
-    i, best_lines = min((0, first_lines), (1, second_lines),
-                        key=lambda x: wer(x[1]))
-
-    return i
+    if not first_lines:
+        return 1
+    elif not second_lines:
+        return 0
+    else:
+        i, best_lines = min((0, first_lines), (1, second_lines),
```
vimalmanohar: I think if line 139 is modified as:
`return float(num_incorrect_words) / num_words`, this issue would not happen. Maybe you could test that instead of this change here.

egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py
```diff
@@ -1914,21 +1914,26 @@ def process_data(args, oov_symbol, utterance_stats, word_stats):
...
+                        args.text_out, args.segments_out, cur_utterance,
+                        segments_for_utterance, oov_symbol=oov_symbol,
                         frame_length=args.frame_length)
+                    if args.ctm_edits_out is not None:
+                        print_debug_info_for_utterance(
+                            args.ctm_edits_out, split_lines_of_cur_utterance,
+                            segments_for_utterance, deleted_segments_for_utterance,
+                            frame_length=args.frame_length)
+                except Exception:
+                    _global_logger.error(
```
vimalmanohar: I am not sure if we expect it to fail but still continue. If it was only because of the previous issue, which will be fixed it is better to not make this change.

egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
jtrmal: Yes, this has helped. It's definitely cleaner solution.

egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py
jtrmal: This still fails:
```
 4 2019-05-23 04:01:47,426 [steps/cleanup/internal/segment_ctm_edits_mild.py:1397 - merge_clusters - ERROR ] Failed merging clusters [[-1], [0, 1], [2], [3], [4], [5], [6], [7], [8], [9], [10], [11], [12], [    13]]
  5 2019-05-23 04:01:47,426 [steps/cleanup/internal/segment_ctm_edits_mild.py:1947 - process_data - ERROR ] Error with utterance 190404_162901-00142500-00145500
  6 2019-05-23 04:01:47,426 [steps/cleanup/internal/segment_ctm_edits_mild.py:2060 - main - ERROR ] Failed segmenting CTM edits
  7 Traceback (most recent call last):
  8   File "steps/cleanup/internal/segment_ctm_edits_mild.py", line 2075, in <module>
  9     main()
 10   File "steps/cleanup/internal/segment_ctm_edits_mild.py", line 2041, in main
 11     word_stats=word_stats)
 12   File "steps/cleanup/internal/segment_ctm_edits_mild.py", line 1920, in process_data
 13     utterance_stats=utterance_stats)
 14   File "steps/cleanup/internal/segment_ctm_edits_mild.py", line 1551, in get_segments_for_utterance
 15     segments = merge_segments(segments, args)
 16   File "steps/cleanup/internal/segment_ctm_edits_mild.py", line 1428, in merge_segments
 17     args.max_intersegment_incorrect_words_length))
 18   File "steps/cleanup/internal/segment_ctm_edits_mild.py", line 1336, in merge_clusters
 19     (merged_segment, i, new_cluster)))
 20 TypeError: '<' not supported between instances of 'Segment' and 'Segment'
 21 # Accounting: time=1 threads=1
 22 # Finished at Thu May 23 04:01:47 EDT 2019 with status 1
```

jtrmal: (it does work in python2)

jtrmal: My reasoning about ignoring the problem was that I'm willing to receive one utterance/segment less results, but I want to get at least something (other utterances for which this doesn't fail), especially because I don't see a way for a user how to fix these issues.

vimalmanohar: It seems python3 is strict about comparisons. The issue is sometimes two scores are equal and then the next element in the tuple is used for comparison. 
Line 1336-1337 can be modified instead to pick the first one arbitrarily:
```
                    heapq.heappush(heap, ((-scoring_function(merged_segment), i)
                                          (merged_segment, i, new_cluster)))
```

I believe most normal corner cases are addressed and it shouldn't fail unless there is some bug. So it is better to fail so we find out there is a bug.

jtrmal: Yes, that did the job.

jtrmal:
I'll prepare update the PR during weekend.

jtrmal:
this should address #3355 as well.
Included suggestions from @vimalmanohar

PR #3340, opened 2019-05-23
add script to compute dev PPL on kaldi-rnnlm

hainan-xv:


danpovey:
Thanks.  @saikiranvalluri  can you please check over it and test?

PR #3341, opened 2019-05-24
fix:add a warning about doing cmvn for data.

bringtree:


egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -358,6 +358,8 @@ def train(args, run_opts):
                                right_context_final >= 0 else -1)
 
     default_egs_dir = '{0}/egs'.format(args.dir)
+    if (args.egs_dir is not None) and (args.cmvn_opts != "--norm-means=false --norm-vars=false"):
+        logger.warning("kaldi doesn't call get_egs.sh to cmvn the egs.")
```
danpovey: Can the warning say:
``` Warning: the --cmvn-opts option has no effect because we are not dumping egs```
You could put similar warnings for other options that only affect the egs dumping.  We recently had someone confused about the --frames-per-iter option.

Comment (bringtree) [reply egs/wsj/s5/steps/nnet3/chain/train.py]: I have updated the pull request, may you take a look when you are available, thank you!

File: egs/wsj/s5/steps/nnet3/chain/train.py [hunk 2]
```diff
@@ -358,6 +358,9 @@ def train(args, run_opts):
                                right_context_final >= 0 else -1)
 
     default_egs_dir = '{0}/egs'.format(args.dir)
+    if args.egs_dir is not None:
+        logger.warning("the --feat.cmvn-opts option has no effect because we are not dumping egs")
```
danpovey: Can you change this to just warn if those respective options are actually set?  Would need two more if statements inside this block.

bringtree: I have updated the pull request, may you take a look when you are available, thank you!

PR #3348, opened 2019-05-28
Add CUDA accelerated MFCC computation.

luitjens:
Creates a new directory 'cudafeat' for placing cuda feature extraction
components as they are developed.  Added a directory 'cudafeatbin' for
placing binaries that are cuda accelerated that mirror binaries
elsewhere.

This commit implements:
  feature-window-cuda.h/cu which implements a feature window on the device
    by copying it from a host feature window.
  feature-mfcc-cuda.h/cu which implements the cuda mfcc feature
    extractor.
  compute-mfcc-feats-cuda.cc which mirriors compute-mfcc-feats.cc

  There were also minor changes to other files.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -4,16 +4,16 @@
 
 SHELL := /bin/bash
 
-SUBDIRS = base matrix util feat tree gmm transform \
+SUBDIRS = base matrix util feat cudafeat tree gmm transform \
           fstext hmm lm decoder lat kws cudamatrix nnet \
-          bin fstbin gmmbin fgmmbin featbin \
+          bin fstbin gmmbin fgmmbin featbin cudafeatbin \
           nnetbin latbin sgmm2 sgmm2bin nnet2 nnet3 rnnlm chain nnet3bin nnet2bin kwsbin \
           ivector ivectorbin online2 online2bin lmbin chainbin rnnlmbin \
 	  cudadecoder cudadecoderbin
 
-MEMTESTDIRS = base matrix util feat tree gmm transform \
+MEMTESTDIRS = base matrix util feat cudafeat tree gmm transform \
           fstext hmm lm decoder lat nnet kws chain \
-          bin fstbin gmmbin fgmmbin featbin \
+          bin fstbin gmmbin fgmmbin featbin cudafeatbin \
           nnetbin latbin sgmm2 nnet2 nnet3 rnnlm nnet2bin nnet3bin sgmm2bin kwsbin \
           ivector ivectorbin online2 online2bin lmbin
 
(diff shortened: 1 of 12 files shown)


danpovey:
Did you check that the compilation/installation process works when CUDA is not used?
Can you let us know how much effect this has on the decoding speed?
And have you compared the features with the CPU-generated ones somehow (e.g. `compare-feats`?)  It should be relatively easy to make a test program which would compute them both ways and assert using ApproxEqual() that they are similar.
BTW I just merged a small bugfix for the MFCC computation, you may want to check if it affects your work.

luitjens:
Did you check that the compilation/installation process works when CUDA is
not used?
    No I don't have a non-cuda build.  I was hoping travis would catch
these sorts of errors and I would fix that way.

Can you let us know how much effect this has on the decoding speed?
    I don't have the numbers off hand and my tree is dirty with other
changes right now.  But from memory libri speech on the CPU would take a
minute or so.  On the GPU it takes a couple seconds.   Please note that
this patch set is part of a larger set of changes which will eventually put
the entire online feature extraction pipeline on the device.  We get the
best perf savings when do do those at the same time to avoid unnecessary
traffic across PCI-E.

And have you compared the features with the CPU-generated ones somehow
(e.g. compare-feats?) It should be relatively easy to make a test program
which would compute them both ways and assert using ApproxEqual() that they
are similar.
    I currently do that manually:
       ./compute-mfcc-feats
--config=/workspace/models/LibriSpeech/conf/mfcc.conf
"scp:/workspace/datasets/LibriSpeech/test_clean/wav_conv.scp"
ark,scp:mfcc.ark,mfcc.scp
       ./compute-mfcc-feats-cuda
--config=/workspace/models/LibriSpeech/conf/mfcc.conf
"scp:/workspace/datasets/LibriSpeech/test_clean/wav_conv.scp"
ark,scp:mfcc-cuda.ark,mfcc-cuda.scp
      ./compare-feats ark:mfcc.ark ark:mfcc-cuda.ark
   This says they match.  On top of that I have integrated this into some
other internal work and validated the results manually with printf and
diffs.  The results appear to be correct within machine precision for the
libri test case.

BTW I just merged a small bugfix for the MFCC computation, you may want to
check if it affects your work.
  I see this:
https://github.com/kaldi-asr/kaldi/commit/8c6cd31db8b8b4c4f9fb6b1d19b8c9e61e34dc58
  This should be a trivial fix for us.  I'll get that in.

One final note:  I pushed this a bit earlier than I normally would have
because I was getting requests to get it out sooner for testing.
 shaheenkdr will hopefully be able to report back any issues.

luitjens:
Looking at my patch set I already have the epsilon fix in my code.  I
probably found this when debugging or just got lucky.

luitjens:
went back and tested timing on Libri speech using the linux time command.

compute-mfcc-feats-cuda:  5.7 seconds (probably 5 of that is initializing CUDA).
compute-mfcc-feats: 45.7 seconds.

So approximate speedup is 65x 1 GPU vs 1 CPU thread.

francisr:
What happens if the wav rspecifier is a command that takes a non negligible amount of CPU time (like adding noise)?

luitjens:
the cuda implementation handles dithering already.

luitjens:
and to answer the larger question.  if something is time consuming on the
CPU we can consider putting it in the device.  I would need to understand
the use case for this and compare it against other priorities.

francisr:
I meant adding reverberation and point source noises. And also conversion from whatever format the file is into wav.  
If this is not properly streamed the GPU will just spend most of its time waiting for the data to come in.

luitjens:
so be it.  this is a part of a larger set of changes and it provides value
to many use cases.  if you have a use case that is not covered then you
should consider adding those components in CUDA.    right now those
features you are concerned about are not a target for my effort.
eventually they will be but for now I have other priorities.

in the coming weeks I'll upstream more changes including cmvn, splice,
ivector, and possibly fbank.  as we get more of these features on the
device we can more easily amortize the cost of preprocessing.

luitjens:
It is very isolated and I plan to keep it that way along with all of the other feature extraction code i'm working on.

PR #3354, opened 2019-05-31
[WIP] A* combine decoder

LvHang:
This PR is used to track the A* decoder and ask Dan in a more convenient way.

Firstly, this decoder is based on combine decoder with bucket queue. We use backward_cost(beta) to replace extra_cost. If set the beta[tok] for all the currently active tokens to -alpha[tok], the alpha-beta pruning is equivalent to extra_cost pruning. (I wrote an example in the google doc, Dan, you can check it. In fact, we use cost_offset to make the tot_cost close to 0. When we want to prune forwardlinks and tokens, we have to compare it with the best one's alpha+beta + config_.lattice_beam. From this direction, extra_cost is a little bit better.)
The final frame is a special case. I set beta[tok] = -alpha[tok] + (alpha[tok] + final_cost[tok] - best_cost_with_final). I'm trying to evident whether it is right.

The outline of the code is as follows. 
For each frame, we call ProcessForFrame (explore stage) to generate a semi-complete token list. 
After that, we do backfill periodically (say, backfill_interval = 5). In backfill stage, we will compute the beta for expanded token from frame T to frame T-14 (say, beta_interval = 15) and the un-expanded token will inherit the beta from the corresponding token.  The token whose alpha+beta is fall below beam will be pruned. [Note: I only get rid of a token and its whole forwardlinks. However, I will not prune any forwardlink of an existing token. Because I think we will not re-visit the HCLG, we should keep more opportunities for shadowed tokens]. For frames t-14 throught t-10, we expand all shadowed tokens. For frames t-9 throught t-1, only the best-in-class tokens will be expanded. (This is controlled by expand_best_interval, say, 10).
After the backfill stage, the frames, from t-10 to 0, are completly processed. I call t-10 as the complete_frame_. To save memory, for every prune_interval (say 20) frames, we will PruneForwardLinks and PruneTokens. This process can go back to 0 or T-25. I will do some tests.
At last, in FinalizeDecoding, we will expand the epsilon arcs for the final frame, compute betas for the last few frames, expand the shadowed tokens for the last few frames and prune with final cost.

Do you think the basic procedure is matched with your envisage? I'm trying to do some experiments on minilibrispeech. (Of course, I will refine the code. It's ugly right now.)

### DIFF
--- a/egs/mini_librispeech/s5/run.sh
+++ b/egs/mini_librispeech/s5/run.sh
@@ -199,6 +199,17 @@ if [ $stage -le 9 ]; then
   local/chain/run_tdnn.sh --stage 0
 fi
 
+if [ $stage -le 10 ]; then
+  for test in dev_clean_2; do
+    steps/decode_biglm_fast.sh --stage 1 --nj 38 --cmd "$decode_cmd" \
+    --online-ivector-dir exp/nnet3/ivectors_${test}_hires \
+    --beam 15.0 --lattice-beam 8.0
+    exp/chain/tree_sp/graph_tgsmall \
+    data/lang_test_tgsmall/G.fst data/lang_test_tglarge/G.fst \
+    data/${test}_hires \
+    exp/chain/tdnn1h_sp/decode_biglm2_${test}_diff_15_8_15_cutoff
+  done
+fi
 # local/grammar/simple_demo.sh
 
 # Don't finish until all background decoding jobs are finished.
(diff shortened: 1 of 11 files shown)


danpovey:
On the last frame, instead of `beta[tok] = -alpha[tok] + (alpha[tok] + final_cost[tok] - best_cost_with_final)`, I would just do `beta[tok] = -alpha[tok]`.  Note: this means that the best total-cost (beta + alpha) is zero, so the pruning with lattice_beam is trivial.

RE `I only get rid of a token and its whole forwardlinks. However, I will not prune any forwardlink of an existing token.`: be careful not to have forward links with dangling pointers to the Token on the following frame.

Instead of using the term "shadowed", let's just use the term "un-expanded".  This is easier to understand in terms of A*.

kkm000:
@LvHang, @danpovey, the description is very good for _what_, but entirely lacking the _why_. What is the general idea? Is this still alive?

danpovey:
I think we can close this, it was a decoder idea that we didn't get very clear gains from.

PR #3355, opened 2019-05-31
modify_wer_return

byuns9334:
original wer function returns a tuple, which seems to cause error

### DIFF
--- a/egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
+++ b/egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
@@ -136,8 +136,8 @@ def wer(ctm_edit_lines):
         return float('inf')
     if num_words == 0 and num_incorrect_words == 0:
         return 0
-    return (float(num_incorrect_words) / num_words, -num_words)
-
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/steps/cleanup/internal/resolve_ctm_edits_overlaps.py
```diff
@@ -136,8 +136,8 @@ def wer(ctm_edit_lines):
         return float('inf')
     if num_words == 0 and num_incorrect_words == 0:
         return 0
-    return (float(num_incorrect_words) / num_words, -num_words)
-
+    #return (float(num_incorrect_words) / num_words, -num_words)
+    return (float(num_incorrect_words) / num_words)
```
danpovey: Perhaps @vimalmanohar  could comment.

vimalmanohar: I think @jtrmal is also working on fixing the exact same issue.

danpovey: hm, so the proper fix isn't as simple as this?

danpovey:
Looks like this is resolved now, in #3337.  But thanks a lot.

PR #3373, opened 2019-06-05
Enhancements to the cudamatrix/cudavector classes.

luitjens:
Added CuSolver to the device class.  This is only supported with
    Cuda 9.1 or newer.  Calling CuSolver code without Cuda 9.1 or newer
    will result in a runtime error.

    This change required some changes to the build system which requires
    versioning the configure script. This forces everyone to reconfigure.
    Failure to reconfigure would result in linking and build errors on
    some systems.

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -110,6 +110,12 @@ void CuDevice::Initialize() {
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
     CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
+
+    #if CUDA_VERSION >= 9100
+    CUSOLVER_SAFE_CALL(cusolverDnCreate(&cusolverdn_handle_));
+    CUSOLVER_SAFE_CALL(cusolverDnSetStream(cusolverdn_handle_, 
+            cudaStreamPerThread));
+    #endif
     
     #if CUDA_VERSION >= 9000 
     if (device_options_.use_tensor_cores) {
@@ -130,6 +136,7 @@ void CuDevice::Initialize() {
     // To get same random sequence, call srand() before the constructor is invoked,
     CURAND_SAFE_CALL(curandSetGeneratorOrdering(
           curand_handle_, CURAND_ORDERING_PSEUDO_DEFAULT));
+    CURAND_SAFE_CALL(curandSetStream(curand_handle_, cudaStreamPerThread));
     SeedGpu();
   }
 }
@@ -263,6 +270,23 @@ void CuDevice::FinalizeActiveGpu() {
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
     CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
+    
+    #if CUDA_VERSION >= 9100 
+    CUSOLVER_SAFE_CALL(cusolverDnCreate(&cusolverdn_handle_));
+    CUSOLVER_SAFE_CALL(cusolverDnSetStream(cusolverdn_handle_,
+            cudaStreamPerThread));
+    #endif
+
+    #if CUDA_VERSION >= 9000 
+    if (device_options_.use_tensor_cores) {
+      // Enable tensor cores in CUBLAS
+      // Note if the device does not support tensor cores this will fall back to normal math mode
+      CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_, 
+            CUBLAS_TENSOR_OP_MATH));
+    }
+    #endif
+
+    
     // Initialize the cuSPARSE library
     CUSPARSE_SAFE_CALL(cusparseCreate(&cusparse_handle_));
     CUSPARSE_SAFE_CALL(cusparseSetStream(cusparse_handle_, cudaStreamPerThread));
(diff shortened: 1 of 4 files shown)


src/cudamatrix/cu-kernels.cu
```diff
@@ -3666,11 +3666,11 @@ void _cuda_mat_copy_range_clamped(
       // compute offset row
       int32_t r_in = r + row_start;
       // clamp if necessary
-      if (r_in < clamp_low) r_in = clamp_low;
-      if (r_in > clamp_high) r_in = clamp_high;
+      if (r_in < clamp_low) r_in=clamp_low;
```
danpovey: changes around these lines seem to only make it non-google-style-guide-compliant.

danpovey:
I am not sure about the softmax scale thing.  It seems to me that it makes the interface messy; and when would it be used?

luitjens:
Thanks to default parameters the interface from the matrix class is unchanged unless someone wants to use it.  It is useful in the ivector code that i'm working on and possibly elsewhere.  The alternative is I implement a second softmax function for that component only.  That however, adds more code maintenance.

src/cudamatrix/cu-kernels.cu
luitjens: sorry looks like a case of a merge error on my side. i'll fix and push.

luitjens:
To help justify this i'll provide more context.  In the OnlineIVector process we compute the posteriors, apply a soft max, and then and then scale it by posterior_scale.  

The scaling operation is very small (1 load, 1 store, 1 op per element).  Since we have 2 memory transactions and only 1 op it is highly memory bandwidth bound.  Fusing this operation into another kernel makes a lot of sense as it eliminates an extra kernel call which can easily add 10-15us latency.  We essentially get that operation for free.  

While at first it may seem odd to have a scaling operation in ApplySoftMax type operations I contend this isn't that odd of a change.  For example, blas operations tend to have a multiplier (i.e. AddMatMat has both alpha and beta).  Adding the scaling parameter increases the utility of the operation without sacrificing performance.  The increase in complexity is actually extremely small and from an API standpoint it is invisible to a user that doesn't want to use it thanks to C++ default parameters.

For code context i've included the ComputePosteriors routine from my ivector code which will be sent out sometime in the next week or so.

void IvectorExtractorFastCuda::ComputePosteriors(
    const CuMatrixBase<float> &feats, CuMatrix<float> *posteriors) {

  int num_frames = feats.NumRows();

  posteriors->Resize(num_frames, num_gauss_, kUndefined);

  posteriors->CopyRowsFromVec(ubm_gconsts_);

  CuMatrix<float> feats_sq(feats.NumRows(), feats.NumCols(), kUndefined);

  // using our own kernel here to avoid an extra memcpy.
  // ApplyPow unfortunately only works in place.
  compute_square(feats.NumRows(), feats.NumCols(), feats.Data(), feats.Stride(),
      feats_sq.Data(), feats_sq.Stride());

  posteriors->AddMatMat(1.0, feats, kNoTrans, ubm_means_inv_vars_, kTrans, 1.0);
  posteriors->AddMatMat(-0.5, feats_sq, kNoTrans, ubm_inv_vars_, kTrans, 1.0);

  // apply scaling factor
  posteriors->ApplySoftMaxPerRow(*posteriors, info_.posterior_scale);

  if (info_.max_count > 0) {
    // when max count > 0 we need to know the total posterior sum to adjust
    // the prior offset.  So calculate that here.
    get_matrix_sum_double_buffer(b_, posteriors->NumRows(), posteriors->NumCols(),
                   posteriors->Data(), posteriors->Stride(), tot_post_.Data());
  }
}

If there is still concern about this then I can implement our own ApplySoftMaxPerRow kernel with scaling but that would lead to code duplication increasing the overall code maintenance.

danpovey:
OK, I suppose I'm willing to accept it.  Not super happy about it though.
Make sure it's clearly documented (e.g. that it scales the output *after*
the softmax).

luitjens:
Thanks, i pushed some more documentation.  I phrased it as controlling the scaling range. 

For example:  default softmax rescales to the range of [0, 1] such that the results sum to 1.  This parameter allows a user to change the range to [0, scale] and sum to scale.  This at least matches my understanding of what softmax and a scale would do.  However, i'm not an expert and you should probably validate my comments are correct.

danpovey:
@luitjens, about the i-vector estimation code: the scale we use on the posteriors (normally 0.1) only affects the interaction with the prior.  For purposes of the extracted i-vectors, the effect would be the same as scaling up the prior term.  In ivector-extractor.cc, where it does:
` quadratic_term_.AddToDiag(1.0);`
you could do
` quadratic_term_.AddToDiag(1.0 / conf.acoustic_weight);`
and it should produce the same i-vectors; the only other thing that would change is the dynamic range of the objective function, which would increase by a factor 1.0 / conf.acoustic_weight; this would affect only diagnostics and, if relevant, stopping criteria, but it should be easy to adjust those.  Can you see if it's possible to do it that way?  The thing with the softmax just seems a bit kludgey to me.

luitjens:
looking closely at where the posteriors get used it appears that I can pull this operation into blas operations by playing with alpha beta parameters;

 gamma->AddRowSumMat(info_.posterior_scale, posteriors, 0.0f);
 X->AddMatMat(info_.posterior_scale, posteriors, kTrans, feats, kNoTrans, 0.0f);

Would this be more acceptable?

luitjens:
I have updated the PR to now only include the Cu-Solver changes.  Softmax changes have been removed.

src/cudamatrix/cu-device.cc
```diff
@@ -110,6 +110,12 @@ void CuDevice::Initialize() {
     // Initialize CUBLAS.
     CUBLAS_SAFE_CALL(cublasCreate(&cublas_handle_));
     CUBLAS_SAFE_CALL(cublasSetStream(cublas_handle_, cudaStreamPerThread));
+
+    #if CUDA_VERSION >= 9100
```
danpovey: Let's have the `#if` left-justified, I think it's more standard.  Same with other preprocessor directives you come across.

luitjens: pushed

PR #3389, opened 2019-06-12
Added binary compute-online-feats.

luitjens:
This binary computes features and ivectors using the online2 recipe.

### DIFF
--- a/src/online2bin/compute-online-feats.cc
+++ b/src/online2bin/compute-online-feats.cc
@@ -1,0 +1,119 @@
+// online2/compute-online-feats.cc
+//
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+// Justin Luitjens
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "base/kaldi-common.h"
+#include "feat/wave-reader.h"
+#include "online2/online-nnet2-feature-pipeline.h"
+#include "util/common-utils.h"
+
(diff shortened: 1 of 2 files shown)


luitjens:
Dan any idea why this check isn't running? It looks like it is just stuck.

danpovey:
Likely just temporary outage on the test platform.

src/online2bin/compute-online-feats.cc
```diff
@@ -0,0 +1,119 @@
+// online2/compute-online-feats.cc
+//
```
danpovey: Is this binary substantially different from  online2-wav-dump-features?

luitjens:
Looks like the main differences are

it outputs the ivector features and does not handle speaker2utt.   Mostly
this binary is used to validate the cuda implementation of the feature
pipeline.  It is nice to have a binary we can compare against.  If you
would like we could move it under cudafeatbin though that may not make
sense since it does not use CUDA.

danpovey:
I think I would prefer if you just modify the existing program.  Will avoid
bulking things up, the compiled
set of programs is already too large.   How about modifying the existing
program so that if you give it a 4th argument it will have a usage like
yours?  (Would still take a spk2utt-rspecifier that
would, in your normal usage, be a one to one map).

Suggested modified usage message:
```
        "Usage (1):\n"
        " online2-wav-dump-features [options] <spk2utt-rspecifier>
<wav-rspecifier> <feature-wspecifier>\n"
        "   The spk2utt-rspecifier can just be <utterance-id>
<utterance-id> if\n"
        "    you want to generate features utterance by utterance.\n"
        "Usage (2):\n"
        " online2-wav-dump-features [options] <spk2utt-rspecifier>
<wav-rspecifier> <feature-wspecifier> <ivector-wspecifier>\n"
         "   In this usage pattern, the features and ivectors will be
dumped separately (the ivectors\n"
         "   will be dumped as vectors after being taken from the last
frame of the last utterance\n"
         "   of each speaker\n"
        "Usage (3):\n"
        "     online2-wav-dump-features [options]
--print-ivector-dim=true\n"
        "See steps/online/nnet2/{dump_nnet_activations,get_egs.sh} for
examples.\n";
```

If you have testing scripts that you use, it might be nice to have them
checked in somewhere.

luitjens:
ok,  i think we can close this and i'll try to get back and add the features to the existing binary.  This is low priority since it doesn't impact functionality at all.

danpovey:
ok.

PR #3391, opened 2019-06-12
CUDA decoder performance patch

hugovbraun:


hugovbraun:
Performance patch for the GPU decoder. Includes performance optimizations, some changes to reduce memory usage, plus some new pipeline API functions (creating group of tasks, useful if running the pipeline as a service). It also computes some parameters (such as aux-q-capacity or main-q-capacity) automatically using other parameters (max-active), in order to simplify things for the user.

We now have end-to-end performance at 4,800 XRTF for the model librispeech evaluated on librispeech/clean, and 2,000 XRTF for the model aspire evaluated on librispeech/other (what we use as "worst case").

We've also measured the **decoder-only** performance in the range 16,500 XRTF-24,900XRTF on V100, 8,100 XRTF- 11,600XRTF for T4 (low=librispeech/test_other, high=librispeech/test_clean). It means that the end-to-end speed will be close to the one of the acoustic model used (in our pipelines, we use TDNNs).

Most of the changes are internal. The pipeline API changes are:

1. Possibility to add a callback when a creating a new task. This callback will be called once the final lattice is ready. This can be used by the user to perform tasks on the final lattice (such as rescoring, best path search, I/O, etc.). This callback will be called on the threadpool, so it must be threadsafe (cf example in cudadecoderbin/).

2. Tasks group. Useful if the pipeline is used as a service. The user can specific a group when creating a new task. Then it is possible to sync (wait for completion) on a specific group, or just any group that is completed. The API sync functions are:

- Blocking:
Waiting for a specific group to complete:
    WaitForGroup(group)
Waiting for any group to complete :
    std::string group_done = WaitForAnyGroup()
- Non-blocking:
Check if a group is done. If not, returns immediately (and returns false)
    bool IsGroupCompleted(group)
Same thing, but without specifying a group (any group completed):
    std::string group_done;
    bool IsAnyGroupCompleted(&group_done);

danpovey:
Great.  Is this pretty stable / tested, then?

hugovbraun:
It is stable on our most common test cases, but it hasn't been validated on the larger test suite yet

ramakrishnapallala:
@hugovbraun I am trying to run Aspire model with V100 with Librispeech dataset. I am getting WER around 75%. Can you tell me what are settiings of parameters (default_parameters.inc) like beam, lattic_beam, max_active etc..

I am using the model from: http://dl.kaldi-asr.org/models/0001_aspire_chain_model.tar.gz
I am running run_benchmark.sh script from "docker pull nvcr.io/nvidia/kaldi:19.06-py3"

danpovey:
It may be largely due to the language model mismatch.  You can't fix that
without
getting a more matched language model and vocabulary and rebuilding the
graph.
But I suspect you'll need more Kaldi experience before you can do that.

ramakrishnapallala:
@danpovey, Can you give some pointers on how can I tune the settings to get better performance? does it require re-training?

danpovey:
This isn't the place for this type of question.. in any case, in your case I think you need to do some background reading.. e.g. HTK book for speech recognition, "a bit of progress in language modeling" for language modeling.

ramakrishnapallala:
OK Thank you. If I any questions, can I raise GH issue or any other forum is available?

danpovey:
see here http://kaldi-asr.org/forums.html RE kaldi-help.
but you need to do some reading first before asking anything there.

ramakrishnapallala:
> see here http://kaldi-asr.org/forums.html RE kaldi-help.
> but you need to do some reading first before asking anything there.

Ok thank you.

ramakrishnapallala:
@hugovbraun @danpovey,  I would like to get one final clarity, The Aspire model which @hugovbraun  tested is the one mentioned in the following link or different one?

http://dl.kaldi-asr.org/models/0001_aspire_chain_model.tar.gz

hugovbraun:
It's this one : 
https://github.com/kaldi-asr/kaldi/tree/master/egs/aspire
We're using the pretrained model there:
https://kaldi-asr.org/models/m1 

Please note that if you use "run_benchmark.sh" from the NVIDIA Container it will use librispeech as a test dataset. So if you do so using the ASpIRE model, it will lead to the model mismatch @danpovey was talking about.

ramakrishnapallala:
Thank you @hugovbraun  for your reply.

In your commit description, you mentioned about testing the ASpIRE model with LibriSpeech test-other dataset. what was the WER you observed in your benchmark test runs? I am seeing WER of 75%

PR #3399, opened 2019-06-18
online2-tcp, produce time,

vesis84:
- modify the online2 tcp server to produce time of the segments
  (new option --produce-time),
- time displayed as (seconds): "{  0.00 -->  3.40 } <recognized text>"

We also introduce debug dumps of features and raw audio (verbose>=100).

In chain trainer we'd like to ask SGE both for GPU and RAM resources
(binary typically crashes when not enough RAW available to map GPU memory with RAM).

### DIFF
--- a/src/online2bin/online2-tcp-nnet3-decode-faster.cc
+++ b/src/online2bin/online2-tcp-nnet3-decode-faster.cc
@@ -85,6 +85,20 @@ std::string LatticeToString(const Lattice &lat, const fst::SymbolTable &word_sym
   return msg.str();
 }
 
+std::string GetTimeString(int32 t_beg, int32 t_end, BaseFloat time_unit) {
+  char buffer[100];
+  double t_beg2 = t_beg * time_unit;
+  double t_end2 = t_end * time_unit;
+  snprintf(buffer, 100, "%.2f %.2f", t_beg2, t_end2);
+  return std::string(buffer);
+}
+
+int32 GetLatticeTimeSpan(const Lattice& lat) {
+  std::vector<int32> times;
+  LatticeStateTimes(lat, &times);
+  return times.back();
+}
+
 std::string LatticeToString(const CompactLattice &clat, const fst::SymbolTable &word_syms) {
   if (clat.NumStates() == 0) {
     KALDI_WARN << "Empty lattice.";
@@ -132,6 +146,7 @@ int main(int argc, char *argv[]) {
     BaseFloat samp_freq = 16000.0;
     int port_num = 5050;
     int read_timeout = 3;
+    bool produce_time = false;
 
     po.Register("samp-freq", &samp_freq,
                 "Sampling frequency of the input signal (coded as 16-bit slinear).");
@@ -145,6 +160,8 @@ int main(int argc, char *argv[]) {
                 "Number of seconds of timout for TCP audio data to appear on the stream. Use -1 for blocking.");
     po.Register("port-num", &port_num,
                 "Port number the server will listen on.");
+    po.Register("produce-time", &produce_time,
+                "Prepend begin/end times between endpoints (e.g. '5.46 6.81 <text_output>', in seconds)");
 
     feature_opts.Register(&po);
     decodable_opts.Register(&po);
@@ -163,6 +180,9 @@ int main(int argc, char *argv[]) {
         word_syms_filename = po.GetArg(3);
 
     OnlineNnet2FeaturePipelineInfo feature_info(feature_opts);
+
+    BaseFloat frame_shift = feature_info.FrameShiftInSeconds();
+    int32 frame_subsampling = decodable_opts.frame_subsampling_factor;
 
     KALDI_VLOG(1) << "Loading AM...";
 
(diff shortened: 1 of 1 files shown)


src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -248,7 +284,13 @@ int main(int argc, char *argv[]) {
 
           Vector<BaseFloat> wave_part = server.GetChunk();
           feature_pipeline.AcceptWaveform(samp_freq, wave_part);
+
+          // dump 1st chunk of 'raw' audio (debug),
```
danpovey: I feel like a lot of these vlog statements are debug things that probably shouldn't remain.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -145,6 +168,8 @@ int main(int argc, char *argv[]) {
                 "Number of seconds of timout for TCP audio data to appear on the stream. Use -1 for blocking.");
     po.Register("port-num", &port_num,
                 "Port number the server will listen on.");
+    po.Register("produce-time", &produce_time,
+                "Send 'sentence' begin/end times based on end-points");
```
danpovey: This might be a good place to document what exactly is printed?

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -85,6 +86,27 @@ std::string LatticeToString(const Lattice &lat, const fst::SymbolTable &word_sym
   return msg.str();
 }
 
+std::string GetTimeString(int32 t_beg, int32 t_end, int32 subsamp) {
+  char buffer[100];
+  double t_beg2 = t_beg * subsamp / 100.;
```
danpovey: I don't think we should just assume that the frame shift is 0.01.  There must be a way to infer the actual frame shift from the feature_opts.  I'm not sure why it has to be %5.2f instead of %.2f?  Please document this with an example of the kind of string it returns.  And I'm not 100% convinced that the {, --> and } is helpful; I feel like just the plain numbers might be easier for client code to parse.  Also it's not clear to me that the end-time of the last word, as opposed to the length of time that was decoded, is the most useful thing.  I am starting to wonder whether it might make more sense just to let people make similar changes depending on their use-case.

egs/wsj/s5/steps/nnet3/chain/train.py
```diff
@@ -239,9 +239,9 @@ def process_args(args):
                    If you have GPUs and have nvcc installed, go to src/ and do
                    ./configure; make""")
 
-        run_opts.train_queue_opt = "--gpu 1"
+        run_opts.train_queue_opt = "--gpu 1 --mem 10G"
```
danpovey: I think these queue option changes may be pretty specific to the way your grid infrastructure is set up, e.g. the virtual memory settings or whatever -> not inclined to merge this.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
vesis84: Okay, but how can I get the 'length of time that was decoded'? Is there an easy way to measure the 'time-lenght' of the Lattice?

vesis84: Got it, getting alignment and measuring its length...

danpovey:
LatticeStateTimes() returns the length in frames

vesis84:
All the remarks are resolved. Calling for 2nd round of code review...

danpovey:
Looks reasonable to me.  @hhadian let me know if you think this is a reasonable/useful feature.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -265,7 +294,17 @@ int main(int argc, char *argv[]) {
             if (decoder.NumFramesDecoded() > 0) {
               Lattice lat;
               decoder.GetBestPath(false, &lat);
+              TopSort(&lat); // for LatticeStateTimes(),
               std::string msg = LatticeToString(lat, *word_syms);
+
+              // get time-span after previous endpoint,
+              if (produce_time) {
+                int32 t_beg = frame_offset;
+                int32 t_end = frame_offset + GetLatticeTimeSpan(lat);
```
hhadian: I'm not sure if `TopSort` and `LatticeStateTimes` are really needed here. Can't we compute `t_end` using `decoder.NumFramesDecoded()` and/or `samp_count`/`check_count`?

src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -145,6 +160,8 @@ int main(int argc, char *argv[]) {
                 "Number of seconds of timout for TCP audio data to appear on the stream. Use -1 for blocking.");
     po.Register("port-num", &port_num,
                 "Port number the server will listen on.");
+    po.Register("produce-time", &produce_time,
```
hhadian: Just a suggestion: I think "--add-timestamps" or "--timestamps" is a better choice. "produce-time" seems a bit confusing to me.

src/online2bin/online2-tcp-nnet3-decode-faster.cc
vesis84: yes, these are good points, the PR is changed now accordingly...

vesis84:
incorporated remarks from @hhadian , ready for next step...

PR #3402, opened 2019-06-20
Update Insert function of hashlist and decoders

LvHang:
The Insert function of hashlist has been changed from "inline void Insert(I key, T val)" to "inline Elem *Insert(I key, T val)". 
Differing from the original Insert function in which the user is in charge of asserting the element is not already present, the new Insert function checks, for each new element, whehter the key is already present in the hashtable, and if so, the element is not inserted, returning a pointer to this existing element. (i.e. The new Insert function has the ability of Find and Insert.)

As the change of Insert, the corresponding decoders are modified.
1. Change "vector<StateId> queue_" to "vector<const Elem* > queue_"
2. FindOrAddToken returns "Elem*" rather than "Token*", and the implementation changes a little bit.
3. Find is replaced by Insert in different decoders

I tested the change with steps/nnet3/decode.sh and steps/decode_biglm.sh on minilibrispeech. Comparing with previous codes, consistent results can be obtained.
Perhaps, I should do some more tests.

### DIFF
--- a/src/decoder/biglm-faster-decoder.h
+++ b/src/decoder/biglm-faster-decoder.h
@@ -397,18 +397,14 @@ class BiglmFasterDecoder {
             if (new_weight < next_weight_cutoff) {  // not pruned..
               PairId next_pair = ConstructPair(arc.nextstate, next_lm_state);
               Token *new_tok = new Token(arc, ac_weight, tok);
-              Elem *e_found = toks_.Find(next_pair);
+              Elem *e_found = toks_.Insert(next_pair, new_tok);
               if (new_weight + adaptive_beam < next_weight_cutoff)
                 next_weight_cutoff = new_weight + adaptive_beam;
-              if (e_found == NULL) {
-                toks_.Insert(next_pair, new_tok);
+              if ( *(e_found->val) < *new_tok ) {
+                Token::TokenDelete(e_found->val);
+                e_found->val = new_tok;
               } else {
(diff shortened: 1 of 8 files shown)


danpovey:
Please report some results of speed tests.  It would probably make sense to do this at a couple different optimization levels, e.g. 0 and 3.

LvHang:
The following form shows the results of speed tests.
I used "a08" on our grid. I think it is standalone when I used it. The tests are conducted on mini-librispeech test set which has 1089 utterances.
I use nnet3-latgen-faster-parallel (used by steps/nnet3/decoder.sh) which calls LatticeFasterDecoder (see column 2) to do test with JOB=1, so I think the result is exact.
I wrote a binary "latgen-biglm-faster-mapped" which calls LatticeFasterBiglmDecoder (see column 3) to do test with JOB=38 , because it is slow. And then sum over it. I think the result is ok, but maybe has a little bit deviation.

1089 utts | LatticeFasterDecoder | LatticeFasterBiglmDecoder
-- | -- | --
base_o0 | 872.817(0.476897) | 28183.957
new_o0 | 863.985(0.472071) | 27964.83
base_o1(default) | 335.094(0.183091) | 8958.29
new_o1 | 330.817(0.180754) | 8944.62
base_o3 | 326.563(0.17843) | 9015.15
new_o3 | 319.861(0.174768) | 8874.64

From the results, we can see the new version is a little bit faster than previous'.

PR #3417, opened 2019-06-26
fix configure bug about cusolver

LvHang:


src/configure
```diff
@@ -526,7 +526,7 @@ WARNING: CUDA will not be used!
     fi
     
     #add cusolver flags for newer toolkits
-    if [[ $CUSOLVER -eq true ]]; then
+    if [ $CUSOLVER -eq true ]; then
```
danpovey: I don't think this is right either, I tried it out.  The issue is -eq is for integers.  Should be ==.

luitjens: A quick search online shows these would all be acceptable:

if [ "$bool" = true ]; then
if [ "$bool" = "true" ]; then

if [[ "$bool" = true ]]; then
if [[ "$bool" = "true" ]]; then
if [[ "$bool" == true ]]; then
if [[ "$bool" == "true" ]]; then

if test "$bool" = true; then
if test "$bool" = "true"; then

I propose we go with:

if [ "$CUSOLVER" = true ]; then

Dan can you just edit the PR in the web browser and accept?  I'm not sure if I'd have the ability to push into this existing PR and I don't know what time zone LVHang is on.  Seems like modifying it in the web browser to look like the line above would be the quickest solution.

LvHang:
I checked the document you're right.
(Actually, I tried it after I modified yesterday.  Maybe it's just a coincidence.)
Anyway, updated it.

PR #3419, opened 2019-06-26
moreNj_2_splitScp

wanglong001:
using utils/split_scp.pl,  --NJ must be little than  num of speakers. it's not good for me to train. 

`
./utils/split_scp.2.pl: Refusing to split data because number of speakers 2 is less than the number of output .scp files 4
`

and i want to fix it. 
1. Previously in speaker units, now in different classes of utt units.
2. Different classes are more evenly distributed in each split


danpovey:
This would break certain things, e.g. in SAT training, FMLLR.  You can split with --per-utt if this becomes a constraint, but it may require script changes.

PR #3421, opened 2019-06-26
fix cuda_cmd in FarsDat

rezame:
it used in this line
https://github.com/kaldi-asr/kaldi/blob/524db1986e8e4f6b8a5e493e4c426d0a43b25bd4/egs/farsdat/s5/local/nnet/run_dnn.sh#L56


danpovey:
better if you change it in local/nnet/run_dnn.sh to say $cmd --gpu 1

PR #3435, opened 2019-06-29
[build] Fix the cross-compiling issue for Android under MacOS

rayworks:
The changes on the makefile fix an issue related to cross-compiling for Android on host MacOS when building dynamic libraries.

`clang80++: error: linker command failed with exit code 1 (use -v to see invocation)`
`make[1]: *** [libkaldi-matrix.dylib] Error 1`
`make: *** [matrix] Error 2`

See also the original reported issue in [here](https://groups.google.com/forum/#!topic/kaldi-help/5CRcoiOhFZo).

### DIFF
--- a/src/makefiles/default_rules.mk
+++ b/src/makefiles/default_rules.mk
@@ -3,11 +3,19 @@ SHELL := /bin/bash
 
 ifeq ($(KALDI_FLAVOR), dynamic)
   ifeq ($(shell uname), Darwin)
-    ifdef LIBNAME
-      LIBFILE = lib$(LIBNAME).dylib
+    ifdef ANDROIDINC # cross-compiling enabled on host MacOS
+      ifdef LIBNAME
+        LIBFILE = lib$(LIBNAME).so
+      endif
+      LDFLAGS += -Wl,-rpath -Wl,$(KALDILIBDIR)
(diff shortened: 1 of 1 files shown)


src/makefiles/default_rules.mk
```diff
@@ -3,11 +3,19 @@ SHELL := /bin/bash
...
 ifeq ($(KALDI_FLAVOR), dynamic)
   ifeq ($(shell uname), Darwin)
-    ifdef LIBNAME
-      LIBFILE = lib$(LIBNAME).dylib
+    ifdef ANDROIDINC # cross-compiling enabled on host MacOS
+      ifdef LIBNAME
+        LIBFILE = lib$(LIBNAME).so
+      endif
+      LDFLAGS += -Wl,-rpath -Wl,$(KALDILIBDIR)
+      EXTRA_LDLIBS += $(foreach dep,$(ADDLIBS), $(dir $(dep))$(notdir $(basename $(dep))).a)
```
danpovey: I'm surprised this is .a and not .so

rayworks: It looks like all the "kaldi-*.a" libraries will be generated along the way by current default make rules.  Since all the sub folders under `src` has already specified the .a files as its `ADDLIBS`, I simply take advantage of this to make the minimum changes.

rayworks: I've verified the changes on MacOS 10.14.5.

PR #3437, opened 2019-06-29
fix alignment bug which is caused by hashlist insert

LvHang:
Sorry for the inconvience


src/decoder/biglm-faster-decoder.h
```diff
@@ -400,11 +400,13 @@ class BiglmFasterDecoder {
...
-              if ( *(e_found->val) < *new_tok ) {
-                Token::TokenDelete(e_found->val);
-                e_found->val = new_tok;
-              } else {
-                Token::TokenDelete(new_tok);
+              if (e_found->val != new_tok) {
+                if (*(e_found->val) < *new_tok) {
+                  Token::TokenDelete(e_found->val);
+                  e_found->val = new_tok;
+                } else {
```
danpovey: OK, but I reverted your PR so you will have to mess with some git stuff to resolve this conflict.

LvHang: I see.

PR #3439, opened 2019-07-01
WIP: Add OnlineCMVN to Online NNET2 pipeline.  This is used in some models…

luitjens:
… (e.g.

CVTE).  This is optional and off by default.  It applies CMVN before
applying pitch.  This code is essentially copied out of
"online2/online-feature-pipeline.cc/h".

Dan i'm pushing this now as I'd like to discuss if you would be open to
this.  If this is accepted we will also emulate these changes in the
cuda pipeline.

Patch set provided by Levi Barnes.

### DIFF
--- a/src/online2/online-nnet2-feature-pipeline.cc
+++ b/src/online2/online-nnet2-feature-pipeline.cc
@@ -64,6 +64,17 @@ OnlineNnet2FeaturePipelineInfo::OnlineNnet2FeaturePipelineInfo(
       KALDI_WARN << "--online-pitch-config option has no effect "
                  << "since you did not supply --add-pitch option.";
   }  // else use the defaults.
+  if (config.cmvn_config != "") {
+    use_cmvn = true;
+    std::cout << "Reading cmvn config opts from " << config.cmvn_config << std::endl;
+    ReadConfigFromFile(config.cmvn_config, &cmvn_opts);
+    global_cmvn_stats_rxfilename = config.global_cmvn_stats_rxfilename;
+    if (global_cmvn_stats_rxfilename == "")
+      KALDI_ERR << "--global-cmvn-stats option is required with cmvn_conf.";
+  } else { // else user the defaults
+    use_cmvn = false;
+  }
+  
 
   if (config.ivector_extraction_config != "") {
     use_ivectors = true;
@@ -79,6 +90,27 @@ OnlineNnet2FeaturePipelineInfo::OnlineNnet2FeaturePipelineInfo(
 OnlineNnet2FeaturePipeline::OnlineNnet2FeaturePipeline(
     const OnlineNnet2FeaturePipelineInfo &info):
     info_(info) {
+  if (info.global_cmvn_stats_rxfilename != "")
+    ReadKaldiObject(info.global_cmvn_stats_rxfilename, &global_cmvn_stats_);
+  Init();
+}
+
+void OnlineNnet2FeaturePipeline::SetCmvnState(const OnlineCmvnState &cmvn_state) {
+  cmvn_feature_->SetState(cmvn_state);
+}
+
+void OnlineNnet2FeaturePipeline::GetCmvnState(OnlineCmvnState *cmvn_state) {
+  int32 frame = cmvn_feature_->NumFramesReady() - 1;
+  // the following call will crash if no frames are ready.
+  cmvn_feature_->GetState(frame, cmvn_state);
+}
+
+
+// Init() is to be called from the constructor; it assumes the pointer
+// members are all uninitialized but config_ and lda_mat_ are
+// initialized.
+void OnlineNnet2FeaturePipeline::Init() 
+{
   if (info_.feature_type == "mfcc") {
     base_feature_ = new OnlineMfcc(info_.mfcc_opts);
   } else if (info_.feature_type == "plp") {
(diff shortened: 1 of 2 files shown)


danpovey:
I would be open to it.  However I'd be happier if it came with scripting support, since I'm not aware that the current config-preparation scripts support this.  I don't know if there is any chance of getting someone to help with that?  Also we should do plenty of tests to make sure nothing is broken.

luitjens:
I'm not sure what you mean by scripting support here nor the config-prep
scripts you would need.  Can you elaborate?

danpovey:
I mean that normally when you do online decoding you would invoke
steps/online/nnet3/prepare_online_decoding.sh to create the config files.
Also, normally if you want to do online CMVN during test time you would
want to train in a matched way, else there might be a mismatch.  I don't
believe the training scripts currently have an option to do that.

luitjens:
Ok,  i'll have Levi look into adding script support to: /egs/wsj/s5/steps/online/nnet3/prepare_online_decoding.sh

The idea is this should be an opt-in.  So existing trained models should not have to worry about this.  People training models can look and see if there is value and opt-in if there is.  Clearly at least one model developer has found utility in this as CVTE does this.

LeviBarnes:
Does egs/wsj/s5/steps/online/prepare_online_decoding.sh not create cmvn.conf already? Did you mean to add that to another prepare_online_decoding.sh?

danpovey:
That script is for GMMs.
steps/online/nnet3/prepare_online_decoding.sh is the relevant script for
nnet3.
We will already be using similar configs for the ivector setup, but
currently the CMVN'ed features are not used for the features fed into the
nnet.
And I don't know how CVTE is training that model.
It may be OK to add the code support first and look at the rest later
though.

Dan

LeviBarnes:
Thanks for the clarification. Shall I add an issue to the repo so we don't lose track of this? How do you like to track open issues for Kaldi?

danpovey:
Sure.  Just open an issue via github.

vesis84:
Hi, I am interested in this too...
For me, the question is, how should the scripting support look like.

The simplest would be to create 'data' directory that will have the online-cmvn applied on it.
Then train and decode with 'cmvn' disabled, and mark somehow that online-cmvn was used. 
Then this mark would be used when exporting the model into the online2 infrastructure.

Dan, would this be acceptable?
(i really don't want to touch the 'train.py', which is the reason why I would do it this way)

K.

vesis84:
Another questions is if the online_cmvn should be applied on top of:
a) base_features
b) base+pitch features
and if the 'online-cmvn' should be used on input-features of i-vector extractor.

The simplest would be to apply it on top of 'base+pitch features', 
and leave the i-vectors as they are... (online cmn for computing UBM posteriors, and no cmvn in i-vector extration).

Would you agree with what i just wrote down?

danpovey:
> Another questions is if the online_cmvn should be applied on top of:
> a) base_features
> b) base+pitch features
> and if the 'online-cmvn' should be used on input-features of i-vector
> extractor.
>
> The simplest would be to apply it on top of 'base+pitch features',
> and leave the i-vectors as they are... (online cmn for computing UBM
> posteriors, and no cmvn in i-vector extration).
>

Hm, I think I would prefer to give the ivector extractor the same feats
that the recognition is seeing, beause otherwise it will
be learning offsets that are irrelevant for the actual decoding process.  I
know this is a pain, but I think it's a better approach.
This would be a boolean option to the ivector-related scripts (see below).

I would prefer to do it in train.py, at the feature extraction level-- it
will save a lot of disk space and I/O.
The i-vector extractor itself already does online CMN.  (Currently it uses
600 frames, but these are frames in the past.  )
So I wouldn't want to give already-CMN'ed features to the ivector extractor.

Also the pitch features are already mean normalized, but I think there is
no harm in doing it again, especially since we'll
be using a longer window (also, the POV feature is not normalized).

It would make sense, I think, to always use the same CMN for all parts, and
to hardcode in the ivector extractor scripts that if
CMN is used for the features, it's the same CMN options as used for the UBM
posteriors.  So it's effectively a boolean
option, which can be recorded in the directory and then transferred into
the config file used for online ivector extraction.

And we could, as a matter of practice, use the same config file for
normalizing the mfcc and pitch features; but in the online
decoding code this would be passed in separately (as in justin's PR)
because we can't rely on there being ivector features.






Dan

vesis84:
there is one complication, that pitch features are not forwarded to the i-vector extracor. we need to have global cmvn-statistics. can we rely on fact that i-vector extractor is getting the first N features of those that are going to NN input? (then we can use one global stats file and look at whola matrix or its window)

danpovey:
I think it's better to have two separate CMVN objects, one for the ivector
extractor and one for the regular features.  In practice the configs would
be the same, but code-wise, sharing them would be complicated, and I'd
rather not do that.
For the ivector extractor, there will be a boolean option to use the
CMVN'ed features when estimating the ivectors.

Dan

LeviBarnes:
Can I try to summarize? It sounds like Dan prefers to leave the current PR alone and add an additional flag to CMV normalize during ivector extraction. True? Will that confuse people when they ask for CMVN + ivector and don't get any CMVN?

I'm going to look at changes to prepare_online_decoding to build an appropriate cmvn.conf file. Then we'll combine those with this PR. Yes?

Dan, when you say "lots of tests" which are you thinking of specifically?

src/online2/online-nnet2-feature-pipeline.cc
```diff
@@ -64,6 +64,17 @@ OnlineNnet2FeaturePipelineInfo::OnlineNnet2FeaturePipelineInfo(
       KALDI_WARN << "--online-pitch-config option has no effect "
                  << "since you did not supply --add-pitch option.";
   }  // else use the defaults.
+  if (config.cmvn_config != "") {
```
danpovey: please clarify here and where you declare these objects, via comments, that this is the CMVN used for the features, not those used inside the ivector extractor (they may be different because we don't include pitch features when estimating the i-vectors).

danpovey:
Yes, I basically mean leaving this PR alone.  
For testing, you should at least run an online decoding setup without CMVN.  Preferably one with pitch too.  To verify that it doesn't break the current scenarios.

vesis84:
Hi, I start looking into this again. To summarize, the steps are following:
1) add support of online-cmvn to train and decode scripts for "i-vector + nnet3" systems,
2) update the 'steps/online/nnet3/prepare_online_decoding.sh',
3) update the C++ code of online2 decoding (this PR),

So, first we need to solve '1)', this will be done by another PR. We can keep this PR open, and mark is as a WIP = work in progress...

LeviBarnes:
Can we move this discussion to https://github.com/kaldi-asr/kaldi/pull/3462?

danpovey:
Closing as I think #3462 is the right one.

PR #3449, opened 2019-07-07
[src,scripts,egs]  Add code and example for SpecAugment in nnet3

danpovey:


egs/mini_librispeech/s5/local/chain/tuning/run_cnn_tdnn_1b.sh
```diff
@@ -0,0 +1,309 @@
...
+  linear_opts="l2-regularize=0.03 orthonormal-constraint=-1.0"
+  prefinal_opts="l2-regularize=0.03"
+  output_opts="l2-regularize=0.015"
+
+  mkdir -p $dir/configs
+  cat <<EOF > $dir/configs/network.xconfig
+  input dim=100 name=ivector
+  input dim=40 name=input
+
+  # this takes the MFCCs and generates filterbank coefficients.  The MFCCs
```
bringtree: MFCC?

egs/mini_librispeech/s5/local/chain/tuning/run_cnn_tdnn_1b.sh
```diff
@@ -0,0 +1,309 @@
...
+
+# Not testing the 'looped' decoding separately, because for
+# TDNN systems it would give exactly the same results as the
+# normal decoding.
+
+if $test_online_decoding && [ $stage -le 17 ]; then
+  # note: if the features change (e.g. you add pitch features), you will have to
+  # change the options of the following command line.
+  steps/online/nnet3/prepare_online_decoding.sh \
+    --mfcc-config conf/mfcc_hires.conf \
```
bringtree: MFCC?

egs/mini_librispeech/s5/local/chain/tuning/run_cnn_tdnn_1b.sh
danpovey: We dump as MFCC on disk because they are more compressible, we compress each coefficient to 1 byte, which is harder with filterbanks as they are highly correlated.  This component converts back to filterbanks.

PR #3454, opened 2019-07-09
fix various typos and errors in the comment.

### LINKED ISSUE #3459: Error comment in LogSub
See the code
https://github.com/kaldi-asr/kaldi/blob/b5385b46c135f8d1f2bf4f9924e287e6fd91cd23/src/base/kaldi-math.h#L226-L227

It actually returns `log(exp(x) - exp(y))`

csukuangfj:
Fix #3459

### DIFF
--- a/src/fstext/kaldi-fst-io.h
+++ b/src/fstext/kaldi-fst-io.h
@@ -87,7 +87,7 @@ fst::VectorFst<fst::StdArc> *ReadAndPrepareLmFst(std::string rxfilename);
 
 // This is a Holder class with T = VectorFst<Arc>, that meets the requirements
 // of a Holder class as described in ../util/kaldi-holder.h. This enables us to
-// read/write collections of FSTs indexed by strings, using the Table comcpet (
+// read/write collections of FSTs indexed by strings, using the Table concept (
 // see ../util/kaldi-table.h).
 // Originally it was only templated on T = VectorFst<StdArc>, but as the keyword
 // spotting stuff introduced more types of FSTs, we made it also templated on
--- a/src/lm/arpa-lm-compiler.h
+++ b/src/lm/arpa-lm-compiler.h
@@ -31,7 +31,7 @@ class ArpaLmCompilerImplInterface;
 
 class ArpaLmCompiler : public ArpaFileParser {
  public:
-  ArpaLmCompiler(ArpaParseOptions options, int sub_eps,
+  ArpaLmCompiler(const ArpaParseOptions& options, int sub_eps,
                  fst::SymbolTable* symbols)
       : ArpaFileParser(options, symbols),
         sub_eps_(sub_eps), impl_(NULL) {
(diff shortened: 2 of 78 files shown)


csukuangfj:
@danpovey 

I've collected lots of changes. Please have a review.

danpovey:
@desh2608 can you please check through these and comment if you see anything that doesn't look right?  If you say it's OK to merge, I'll merge.

src/fstext/kaldi-fst-io.h
```diff
@@ -87,7 +87,7 @@ fst::VectorFst<fst::StdArc> *ReadAndPrepareLmFst(std::string rxfilename);
 
 // This is a Holder class with T = VectorFst<Arc>, that meets the requirements
 // of a Holder class as described in ../util/kaldi-holder.h. This enables us to
-// read/write collections of FSTs indexed by strings, using the Table comcpet (
+// read/write collections of FSTs indexed by strings, using the Table concpet (
```
desh2608: Is this a typo?

src/lm/mikolov-rnnlm-lib.cc
```diff
@@ -759,7 +759,7 @@ void CRnnLM::restoreNet() {   // will read whole network structure
     }
   }
   if (filetype == TEXT) {
-    goToDelimiter(':', fi);    // direct conenctions
+    goToDelimiter(':', fi);    // direct connenctions
```
desh2608: typo

src/lm/arpa-lm-compiler.h
```diff
@@ -31,7 +31,7 @@ class ArpaLmCompilerImplInterface;
 
 class ArpaLmCompiler : public ArpaFileParser {
  public:
-  ArpaLmCompiler(ArpaParseOptions options, int sub_eps,
+  ArpaLmCompiler(const ArpaParseOptions& options, int sub_eps,
```
desh2608: I'm not sure about the intended effect of changes here (and similar ones elsewhere) for making the variables `const` and pointers. I hope you have got them approved by @danpovey.

src/fstext/kaldi-fst-io.h
csukuangfj: @desh2608 
yes, you're correct.

src/lm/mikolov-rnnlm-lib.cc
csukuangfj: fixed.

src/lm/arpa-lm-compiler.h
csukuangfj: & ---> to avoid unnecessary copy
const ---> we do not need to change it inside the constructor.

desh2608: Ah, I see, a pass by reference for the arguments which will not be changes inside the function.

desh2608:
@danpovey I think this can be merged now.

PR #3458, opened 2019-07-10
Fix feat-to-dim for wxfilename.

### LINKED ISSUE #3457: inconsistent behavior of feat-to-dim for wspecifier and wxfilename
See the code
https://github.com/kaldi-asr/kaldi/blob/b5385b46c135f8d1f2bf4f9924e287e6fd91cd23/src/featbin/feat-to-dim.cc#L51-L61

It prints the dimension of every utterance when it is a wspecifier.
However, it only prints the dim of the first utterance for a wxfilename.

csukuangfj:
Fix #3457.


danpovey:
sorry-- it was supposed to work like that.

PR #3472, opened 2019-07-17
Small fixes

aka-rider:
[src] Fix buffer overflow access (ASAN)
[src] fixed warning: moving a temporary object prevents copy elision
[scripts] tools/extras/check_dependencies.sh look for alternative MKL installation at ${MKL_ROOT}
[src] Fixed compilation error when DEBUG is defined

### DIFF
--- a/src/latbin/lattice-oracle.cc
+++ b/src/latbin/lattice-oracle.cc
@@ -148,7 +148,7 @@ void CountErrors(const fst::StdVectorFst &fst,
 
 bool CheckFst(const fst::StdVectorFst &fst, string name, string key) {
 #ifdef DEBUG
-  StateId numstates = fst.NumStates();
+  fst::StdArc::StateId numstates = fst.NumStates();
   std::cerr << " " << name << " has " << numstates << " states" << std::endl;
   std::stringstream ss;
   ss << name << key << ".fst";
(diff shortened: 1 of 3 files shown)


src/util/text-utils.cc
```diff
@@ -571,8 +571,8 @@ void ReadConfigLines(std::istream &is,
     size_t end = line.find_first_of('#');
     if (start == std::string::npos || start == end) continue;
     end = line.find_last_not_of(" \t", end - 1);
-    KALDI_ASSERT(end >= start);
-    lines->push_back(line.substr(start, end - start + 1));
+    if (end > start)
```
danpovey: I don't understand why the case where end == start would be a problem.  There would be a one-character string, no?  (Note, here, "end" is really last, not one past the last).

aka-rider: The main reason I made it because address sanitizer was complaining.

aka-rider: Ah, I see. Now it skips one-character values. I will revert this change, and will look at it later.

danpovey: I think this was a false alarm, because it can't see the full logic of the routine.  depends on behavior of those string functions

aka-rider: I removed this commit from the pull request

PR #3485, opened 2019-07-22
add --phone-symbol-table to prepare_lang_subword.sh

DongjiGao:
add --phone-symbol-table to prepare_lang_subword.sh and a small fix of prepare_dict_subword.sh

### DIFF
--- a/egs/wsj/s5/utils/subword/prepare_lang_subword.sh
+++ b/egs/wsj/s5/utils/subword/prepare_lang_subword.sh
@@ -59,7 +59,7 @@ position_dependent_phones=true
 share_silence_phones=false  # if true, then share pdfs of different silence
                             # phones together.
 sil_prob=0.5
-phone_symbol_table=             # if set, use a specified phone.txt file
+phone_symbol_table=             # if set, use a specified phones.txt file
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/utils/subword/prepare_lang_subword.sh
```diff
@@ -258,8 +276,15 @@ echo $ndisambig > $tmpdir/lex_ndisambig
 ( for n in `seq 0 $ndisambig`; do echo '#'$n; done ) >$dir/phones/disambig.txt
 
 # Create phone symbol table.
-echo "<eps>" | cat - $dir/phones/{silence,nonsilence,disambig}.txt | \
-  awk '{n=NR-1; print $1, n;}' > $dir/phones.txt
+if [[ ! -z $phone_symbol_table ]]; then
```
danpovey: You need to either make this single brackets or quote $phone_symbol_table.  This will crash if it's empty.

DongjiGao: I did some tests it seems it does not crash with empty variable...

danpovey: oh yes I see, I was testing the wrong way.  OK, will merge.

PR #3490, opened 2019-07-25
Replace LDA layer with delta and delta-delta features

desh2608:
Added scripts for WSJ and SWBD setups.


egs/wsj/s5/local/chain/tuning/run_tdnn_1h.sh
```diff
@@ -0,0 +1,362 @@
...
+  cat <<EOF > $dir/configs/network.xconfig
+  input dim=100 name=ivector
+  input dim=40 name=input
+
+  idct-layer name=idct input=input dim=40 cepstral-lifter=22 affine-transform-file=$dir/configs/idct.mat
+
+  no-op-component name=delta_1 input=Scale(-1.0, Offset(idct, -1))
+  no-op-component name=delta_2 input=Scale(1.0, Offset(idct, 1))
+  no-op-component name=delta_delta_1 input=Scale(-2.0, Offset(idct, 0))
+  no-op-component name=delta_delta_2 input=Scale(1.0, Offset(idct, -2))
```
danpovey: Hm, I think before I merge this I'd like to have some work done on the scripts, to make this much more compact.
Firstly, please don't use any more no-op-components than are absolutely necessary, they lead to redundant matrix copies.  I suspect no more than 2 of those, at most, are needed.  In fact, the following would compile with the least hassle:
 `dim-range-node name=idct_copy1 input=idct offset=0 dim=40`
 `dim-range-node name=idct_copy2 input=idct offset=0 dim=40`
(the syntax may be off slightly):
and then whenever you get an error due to something appearing in the sum with 2 different scales, just use a different copy.  And I'd like this to all be put in an xconfig layer so that we don't need to bulk up the config files.
If you're not sure how, talk to one of the others.  You could copy-and-modify something simple like the `no-op-component` in trivial_layers.py; you can leave it in trivial layers.py; you'll need to add a line to parser.py.

desh2608: Okay, I'll make those changes.

desh2608: Should I just fix the scale as 1.0 or have it as a parameter for the layer (along with a parameter to choose if we want to use batchnorm)? That would be more general, but I suppose in that case I would need to have several NoOpComponents.

RuABraun:
Why are these being added when the performance is worse?

danpovey:
The performance is about the same.  In the WSJ recipe the tdnn1h2_sp is a contrastive experiment, not the one being merged (and Desh, you should probably remove that when you check in).
Desh is refactoring it a bit though, to look nicer in the config file.

The basic reason is that this  'lda layer' thingy adds complexity to the script, as there has to be a special stats accumulation and trainng stage.  Since replacing with the more standard delta/delta-delta recipe seems to give the same performance, I think it's more desirable.  It would also make replication of the recipe within other frameworks easier.

The LDA thing does have the nice property that it just works, you don't have to worry about scaling or hand-tuning anything.  (In these recipes we have to be a bit careful about the scaling of the i-vectors, which, for instance changes depending whether the MFCCs are batch-normalized because batch normalization reduces the MFCCs' scale).  But IMO that's outweighed by the greater ease of implementation of the delta thing.

desh2608:
@danpovey I have made the changes as you suggested.

egs/wsj/s5/local/chain/tuning/run_tdnn_1h.sh
ishine: In training, delta feature is chunk-based, while in testing, it is utt-based.

danpovey: @ishine: that's not really true as any required left/right context is stored with the chunks.

PR #3497, opened 2019-07-29
adding tdnn-F recipe for AMI dataset

aarora8:
To do:
- run same setup with 40 dimension mfcc instead of 80 dimension mfcc and with 13 epochs instead of 9 epochs.


curly1:
So how does the mfcc_hires.conf look like for this recipe?

PR #3499, opened 2019-07-31
Add config file support for relative paths

btiplitz:
I know I've seen other request and we also like to use config files with relative path, so I've added some support for relative paths of the decoder where nested config files are often used

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.cc
@@ -28,7 +28,7 @@ namespace cuda_decoder {
 
 void BatchedThreadedNnet3CudaPipeline::Initialize(
     const fst::Fst<fst::StdArc> &decode_fst, const nnet3::AmNnetSimple &am_nnet,
-    const TransitionModel &trans_model) {
+    const TransitionModel &trans_model, const std::string &path_hint) {
   KALDI_LOG << "BatchedThreadedNnet3CudaPipeline Initialize with "
             << config_.num_control_threads << " control threads, "
             << config_.num_worker_threads << " worker threads"
@@ -38,7 +38,7 @@ void BatchedThreadedNnet3CudaPipeline::Initialize(
   trans_model_ = &trans_model;
   cuda_fst_.Initialize(decode_fst, trans_model_);
 
-  feature_info_ = new OnlineNnet2FeaturePipelineInfo(config_.feature_opts);
+  feature_info_ = new OnlineNnet2FeaturePipelineInfo(config_.feature_opts, path_hint);
   feature_info_->ivector_extractor_info.use_most_recent_ivector = true;
   feature_info_->ivector_extractor_info.greedy_ivector_extractor = true;
 
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.h
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.h
@@ -139,7 +139,8 @@ public:
  // allocates reusable objects that are common across all decodings
  void Initialize(const fst::Fst<fst::StdArc> &decode_fst,
                  const nnet3::AmNnetSimple &nnet,
-                 const TransitionModel &trans_model);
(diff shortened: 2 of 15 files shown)


danpovey:
I'm not so hot on this; it seems to me it would be easier to just make a soft link from whatever the directory the program is running in to the config directory and access it via that link?  Pathnames in there can anyway be relative or absolute but they will be interpreted relative to where the program itself is running.  It seems to me that this just introduces confusion.  Anyway it would need much better documentation.

btiplitz:
In many environments, software and model packs are distributed in from different and multiple sources.   Further every system admin has a unique layout on their systems complicating everything.  

But I have not actually put enough in to make this work on any kaldi binary or I don't think I have since I allow for a default value.  Where would you look for this to be documented ?

danpovey:
Might it be possible to have the binary itself change directory to the root
of the model directlry using chdir?
Assuming it wasn't run from there.  I mean, as a modification to whatever
binary you distribute.

btiplitz:
It does not.  The directory is something that can change from underneath the object as it is a program global.   Perhaps reworking the ReadKaldiObject such that is was aware of the ParseOptions object.  Then the path hint could just be stored there without any change to the remainder of the code would be possible. Seems like a Kaldi 10 thing ?  It could even be useful if the ParseOptions allowed for reading from an iostream vs a filename.  

I have a very minor workaround that I am trying on our end that only effect a single file but I though that option does not look elegant, it works and I only modify a single file.  In place of the filename, I pass the config file as a string (which goes through just fine pretending it is a filename)

danpovey:
I think I found a nice way to solve it.

Class ParseOptions has a function ReadConfigFile().
I recommend to add to class ParseOptions a string member:
  std::string cur_config_file_;   /// The rxfilename we are currently
reading from, if we are inside a call to ReadConfigFile(); else empty.

ReadConfigFile() would set that variable at the start, and set it to the
empty string before it exits.

Then we'd change ParseOptions::SetOption, so that when processing string
options, it would do:

   if (!cur_config_file_.empty())
      value = ReplaceDirVariable(value);

and we'd have a private member function as follows:
  /// Replaces any instance of the string "$cur_dir" with the directory
  /// of the filename in cur_config_file_.  Used when parsing config
  /// files that have the filenames of other config files in them.
  std::string ReplaceDirVariable(const std::string &value);

which you could figure out how to implement; be careful with anything
OS-specific, it should work on Windows.

This doesn't have to wait for kaldi10, as it won't break anything.

Dan

btiplitz:
Right now when reading a config file, there is no way to know that the option is a path or even yet another filename.  The option above depends on knowing which option is a directory.  I believe a new convention would need to be established.  I went through another path today and there is a 3 layer nesting where a config file points to a config file pointing to a data file.

In my work around (which I believe is cleaner than the chdir() call, I have to specify which options are paths.  When a config file option is specified, it seems to alway end in -config for the option name, but when the file read points to a data file, there is no standard thus when reading the option I am uncertain.  There is a slash in the option value.  If the file had something like CDIR or some reserved string, then one would know.  

I'm not seeing an easy way out of this.  So I'm withdrawing the PR

danpovey:
I don't see a problem with my proposal.  Yes, it would replace the string $cur_dir with the name of the directory enclosing the config file that's being read regardless of the meaning of that option string, but I don't see how this could ever cause a problem.  We don't put arbitrary text, like language modeling text, in option strings, so there's no reason that particular string would ever end up there unless it was intended.

btiplitz:
Ok, if you are saying literally to replace cur_dir, then yes I think that could work.

danpovey:
yes, wiht a dollar, like `$cur_dir`

btiplitz:
I'm liking my private change more and more.  Replacing the filename with a config file value and I also allow for a 2nd level of nesting, thus allowing a 2 level config file to be set in memory and not from disk (or merge 2 of them).  I'll still try the curdir to see how that does but that change would require all the config files to change so I can't use it until everyone upgrades.

danpovey:
If you are distributing models somehow you can always manually edit the
config files.
If you are distributing scripts there is no problem as you can just have
them use a soft link.

PR #3501, opened 2019-08-01
Fixed some bugs in mgb_data_prep.sh of mgb2_arabic

DongjiGao:
@mahsa7823 may need to update it to run the recipe.


mahsa7823:
> @mahsa7823 may need to update it to run the recipe.

Thanks. I'll check it out.

PR #3505, opened 2019-08-05
backport of chime5 script from s5b

jtrmal:
I have updated the s5 recipe as the s5b has an awful lot of dependencies, installs whole python conda environment plus uses offline nara_wpe who needs tens of gigabytes of ram to complete. The s5 is much more practical and runnable easily -- only because it uses obsolete configuration, there were some convergence issues. Using the proper way how to specify TDNN-F system has fixed that. As an added benefit, the absolute WER has improved by almost 10 % abs -- so it seems the nara_wpe is not that crucial -- seems to add maybe two or three percent abs?

### DIFF
--- a/egs/chime5/s5/local/chain/run_tdnn.sh
+++ b/egs/chime5/s5/local/chain/run_tdnn.sh
@@ -1,1 +1,1 @@
-tuning/run_tdnn_1a.sh
+tuning/run_tdnn_1b.sh
--- a/egs/chime5/s5/local/chain/tuning/run_tdnn_1b.sh
+++ b/egs/chime5/s5/local/chain/tuning/run_tdnn_1b.sh
@@ -1,0 +1,266 @@
+#!/bin/bash
+
+# Set -e here so that we catch if any executable fails immediately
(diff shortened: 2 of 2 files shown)


danpovey:
OK.  You don't want to make it 1b and move the soft link, like the normal pattern?  I'm not sure if you have results for the old setup?  (They weren't at the top of the script like normal)

jtrmal:
I can figure it out -- I don't have the file but I can pull it from the
history.
I felt like you'd prefer this way, as the previous recipe was not
functioning well (if at all) -- but I don't have anything against going
back to 1a and adding 1b
y.

jtrmal:
ok, done + rebased/squashed

danpovey:
OK thanks.  Will merge.

jtrmal:
eh, seems like I accidentally deleted the backported version... will look if I can pull it out from somwhere :/
Please revert, if possible.

danpovey:
let me just wait till you fix it... I dont want  too many commits to clutter the history, and this shouldn't affect anyone.

PR #3510, opened 2019-08-06
[egs] Fixed issue that leads to empty segment file

Shujian2015:
Fixed issue that leads to empty segment file on SSD disk (more detail: https://groups.google.com/forum/#!topic/kaldi-help/Ij3lQLCinN8)


danpovey:
thanks!

PR #3512, opened 2019-08-08
fixed a bug in egs/voxceleb/v1/local/make_voxceleb1_v2.pl when preparin…

jkchen79:
Fixed a bug in egs/voxceleb/v1/local/make_voxceleb1_v2.pl when preparing the file data/voxceleb1_test/trials. The wav filename extension should be removed. Otherwise, the script run.sh would fail at stage 6.

### DIFF
--- a/egs/voxceleb/v1/local/make_voxceleb1_v2.pl
+++ b/egs/voxceleb/v1/local/make_voxceleb1_v2.pl
@@ -7,7 +7,7 @@
 # Usage: make_voxceleb1_v2.pl /export/voxceleb1 dev data/dev
 #
 # The VoxCeleb1 corpus underwent several updates that changed the directory and speaker ID format.
-# The script 'make_voxceleb1.pl' works for the oldest version of the corpus. 
+# The script 'make_voxceleb1.pl' works for the oldest version of the corpus.
 # This script should be used if you've downloaded the corpus recently.
 
 if (@ARGV != 3) {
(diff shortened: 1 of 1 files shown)


egs/voxceleb/v1/local/make_voxceleb1_v2.pl
```diff
@@ -40,11 +40,11 @@
     my @rec_dirs = grep {-d "$data_base/$dataset/wav/$spkr_id/$_" && ! /^\.{1,2}$/} readdir($dh);
     closedir $dh;
     foreach (@rec_dirs) {
-	  my $rec_id = $_;
-	  opendir my $dh, "$data_base/$dataset/wav/$spkr_id/$rec_id/" or die "Cannot open directory: $!";
-	  my @files = map{s/\.[^.]+$//;$_}grep {/\.wav$/} readdir($dh);
-	  closedir $dh;
-  	  foreach (@files) {
+      my $rec_id = $_;
```
danpovey: this diff is hard to see because of all the whitespace changes.   Any chance of fixing that?

danpovey: actually its OK, I'll merge as-is.

jkchen79: I believe that only using whitespace for indentation is a better code style. I am sorry for that I didn't explain it earlier for those whitespace changes. Anyway, you finally got it.

PR #3516, opened 2019-08-12
cosmetic changes in extend_vocab_demo.sh

QIU-Shuo:


danpovey:
normally this change would be on the small side to give its own commit, but I'll merge.

PR #3523, opened 2019-08-17
[src] Hopefully make it possible to use empty FST in grammar-fst

danpovey:


jty016:
Hmm ... still I got the same message. Looking into more

danpovey:
Oh, wait.  I made a mistake RE how I asked you to prepare the empty FST.
Just do `fstcompile` from an empty file.

I was checking for emptiness with NumStates() == 0; the example I showed
you isn't empty in that sense.

jty016:
changed empty FST generation script like you said and it is working perfectly now :)
`echo "" | fstcompile > empty.fst`

jty016:
Just friendly reminder. I tested my pre-designed grammar FST by assigning empty FSTs on the all entries. With the full graph the CER was 8.5%  and with empty FSTs I got 12.8%.

PR #3524, opened 2019-08-17
[egs] Change cmd var to the value from cmd.sh

Shujian2015:


danpovey:
this is not how it's done, you have to pass it in via the --cmd option

egs/wsj/s5/steps/segmentation/detect_speech_activity.sh
```diff
@@ -17,7 +17,7 @@ if [ -f ./path.sh ]; then . ./path.sh; fi
 
 affix=  # Affix for the segmentation
 nj=32
-cmd=queue.pl
+cmd=$decode_cmd
```
danpovey: .. but the default should be run.pl, you could change that.
And make sure that wherever it is invoked in example scripts, the --cmd option is used.

PR #3526, opened 2019-08-20
[src] Add option to convert pdf-posteriors to phones

danpovey:


whaozl:
@melspectrum007 kaldi provides a very convenient tool actually.

how to obtain "pdf-to-Monophone transformation matrix" (transform.mat)?
you can use get_phone_post.sh to produce transform.mat and  from all training align data sets. and only run once.

and then, use nnet3-chain-compute-post or  nnet3-compute to calculate the phone probability on every frame.


note: the chain frame-shift is 0.03 or frame-subsampling-factor=3. This requires special attention.

PR #3528, opened 2019-08-20
Fix GetDeltaWeights for long-running online decoding

nshmyrev:
Currently bad things happen for long-running online decoding like in tcp decoder. The current code requests delta weights with feature_pipeline.NumFramesReady() that is the whole input pipeline size, meaning hundred thousands of delta weights are returned repeatedly for a long file of 1 hour length on every new utterance. Those delta weights are accumulated in priority queue again and again growing memory to several gigabytes per stream while it could be just a hundred of megabytes. The weight values are not properly estimated too as I believe since the offsets are mixed.

This patch tries to fix the situation. The function GetDeltaWeights uses decoder frame offsets which are reset for every new utterance. Pipeline frame offsets are calculated only when deltaweights are returned.

### DIFF
--- a/src/online2/online-ivector-feature.h
+++ b/src/online2/online-ivector-feature.h
@@ -482,7 +482,9 @@ class OnlineSilenceWeighting {
   //
   // The first_decoder_frame is the offset from the start of the stream in
   // pipeline frames when decoder was restarted last time. We do not change
-  // weight for frames earlier than first_decoder_frame.
+  // weight for the frames earlier than first_decoder_frame. Set it to 0 in
+  // case of compilation error to reproduce the previous behavior or for a
+  // single utterance decoding.
   //
   // How many frames of weights it outputs depends on how much "num_frames_ready"
   // increased since last time we called this function, and whether the decoder
--- a/src/online2/online-ivector-feature.h
+++ b/src/online2/online-ivector-feature.h
@@ -473,20 +473,29 @@ class OnlineSilenceWeighting {
   void ComputeCurrentTraceback(const LatticeFasterOnlineDecoderTpl<FST> &decoder);
 
   // Calling this function gets the changes in weight that require us to modify
-  // the stats... the output format is (frame-index, delta-weight).  The
-  // num_frames_ready argument is the number of frames available at the input
-  // (or equivalently, output) of the online iVector extractor class, which may
-  // be more than the currently available decoder traceback.  How many frames
-  // of weights it outputs depends on how much "num_frames_ready" increased
-  // since last time we called this function, and whether the decoder traceback
-  // changed.  Negative delta_weights might occur if frames previously
+  // the stats... the output format is (frame-index, delta-weight).
+  //
+  // The num_frames_ready argument is the number of frames available at
+  // the input (or equivalently, output) of the online iVector feature in the
+  // feature pipeline from the stream start. It may be more than the currently
+  // available decoder traceback.
+  //
+  // The first_decoder_frame is the offset from the start of the stream in
+  // pipeline frames when decoder was restarted last time. We do not change
+  // weight for frames earlier than first_decoder_frame.
+  //
+  // How many frames of weights it outputs depends on how much "num_frames_ready"
+  // increased since last time we called this function, and whether the decoder
+  // traceback changed.  Negative delta_weights might occur if frames previously
   // classified as non-silence become classified as silence if the decoder's
   // traceback changes.  You must call this function with "num_frames_ready"
   // arguments that only increase, not decrease, with time.  You would provide
   // this output to class OnlineIvectorFeature by calling its function
   // UpdateFrameWeights with the output.
+  //
+  // Returned frame-index is in pipeline frames from the pipeline start.
   void GetDeltaWeights(
-      int32 num_frames_ready_in,
+      int32 num_frames_ready, int32 first_decoder_frame,
       std::vector<std::pair<int32, BaseFloat> > *delta_weights);
 
  private:
--- a/src/online2bin/online2-tcp-nnet3-decode-faster.cc
+++ b/src/online2bin/online2-tcp-nnet3-decode-faster.cc
@@ -283,9 +283,9 @@ int main(int argc, char *argv[]) {
               feature_pipeline.IvectorFeature() != NULL) {
             silence_weighting.ComputeCurrentTraceback(decoder.Decoder());
             silence_weighting.GetDeltaWeights(feature_pipeline.NumFramesReady(),
+                                              frame_offset * decodable_opts.frame_subsampling_factor,
                                               &delta_weights);
-            feature_pipeline.UpdateFrameWeights(delta_weights,
-                                                frame_offset * decodable_opts.frame_subsampling_factor);
+            feature_pipeline.UpdateFrameWeights(delta_weights);
           }
 
           decoder.AdvanceDecoding();
--- a/src/online2/online-ivector-feature.cc
+++ b/src/online2/online-ivector-feature.cc
@@ -574,22 +574,22 @@ int32 OnlineSilenceWeighting::GetBeginFrame() {
 }
 
 void OnlineSilenceWeighting::GetDeltaWeights(
-    int32 num_frames_ready_in,
+    int32 num_frames_ready, int32 first_decoder_frame,
     std::vector<std::pair<int32, BaseFloat> > *delta_weights) {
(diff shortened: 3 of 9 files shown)


src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -282,10 +282,9 @@ int main(int argc, char *argv[]) {
           if (silence_weighting.Active() &&
               feature_pipeline.IvectorFeature() != NULL) {
             silence_weighting.ComputeCurrentTraceback(decoder.Decoder());
-            silence_weighting.GetDeltaWeights(feature_pipeline.NumFramesReady(),
+            silence_weighting.GetDeltaWeights(feature_pipeline.NumFramesReady(), frame_offset * decodable_opts.frame_subsampling_factor,
```
danpovey: Would be nice to put this new arg on its own line.. way over 80.

src/online2/online-ivector-feature.h
```diff
@@ -486,7 +486,7 @@ class OnlineSilenceWeighting {
   // this output to class OnlineIvectorFeature by calling its function
   // UpdateFrameWeights with the output.
   void GetDeltaWeights(
-      int32 num_frames_ready_in,
+      int32 num_frames_ready_in, int32 frame_offset_in,
```
danpovey: Need to have some good documentation on what this new parameter does.  
Will the t values in the delta-weights be relative to this frame offset, or not?

danpovey:
So how safe do you think this is to merge?

nshmyrev:
> So how safe do you think this is to merge?

Thanks, I think it is ok. Memory-wise it is much better. Accuracy is about the same.

Here is decoding of the stream before the patch:

```
    GB
5.273^                                                                       :
     |                                                                      @#
     |                                                               @@@    @#
     |                                                               @@     @#
     |                                                               @@     @#
     |                                                               @@     @#
     |                                                               @@ ::::@#
     |                                                     @@ @@@@   @@ : ::@#
     |                                                  @@@@  @ @@   @@ : ::@#
     |                                                  @@ @  @ @@   @@ : ::@#
     |                                              @@@@@@ @ :@ @@:::@@ : ::@#
     |                                              @ @ @@ @ :@ @@: :@@ : ::@#
     |                                           :::@ @ @@ @ :@ @@: :@@ : ::@#
     |                           ::::::::::::::::: :@ @ @@ @ :@ @@: :@@ : ::@#
     | :::::::::@::::::::::::::::: : :: :: ::: ::: :@ @ @@ @ :@ @@: :@@ : ::@#
     | ::: :: : @::  ::: : ::: ::: : :: :: ::: ::: :@ @ @@ @ :@ @@: :@@ : ::@#
     | ::: :: : @::  ::: : ::: ::: : :: :: ::: ::: :@ @ @@ @ :@ @@: :@@ : ::@#
     | ::: :: : @::  ::: : ::: ::: : :: :: ::: ::: :@ @ @@ @ :@ @@: :@@ : ::@#
     | ::: :: : @::  ::: : ::: ::: : :: :: ::: ::: :@ @ @@ @ :@ @@: :@@ : ::@#
     | ::: :: : @::  ::: : ::: ::: : :: :: ::: ::: :@ @ @@ @ :@ @@: :@@ : ::@#
   0 +----------------------------------------------------------------------->Ti
     0                                                                   4.786

```

This is decoding after the patch:

```
    GB
2.704^#                                                                       
     |#                                                                       
     |#                                                                       
     |#                                                              @@@@@@@@@
     |#                                                @@@@@@@@@@@@@@@@ @ @ @@
     |#                                 @@@@@@@@@@@@@@@@@ @ @@ @ @ @ @@ @ @ @@
     |#                    @@@@@@@@@@@@@@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#     @@::::@:::@@@@@@ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#:::::@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
     |#: : :@ ::: @ : @@@@ @ @ @@ @ @@ @@ @@@@ @@@@@@@ @@ @ @@ @ @ @ @@ @ @ @@
   0 +----------------------------------------------------------------------->Ti
     0                                                                   22.10

```

danpovey:
OK.  Should the accuracy/WER be identical?

nshmyrev:
It differs 20 +- 0.1% due to randomness probably on long file dataset, even for the same code with or without the change.

src/online2/online-ivector-feature.h
```diff
@@ -473,20 +473,29 @@ class OnlineSilenceWeighting {
...
+  // How many frames of weights it outputs depends on how much "num_frames_ready"
+  // increased since last time we called this function, and whether the decoder
+  // traceback changed.  Negative delta_weights might occur if frames previously
   // classified as non-silence become classified as silence if the decoder's
   // traceback changes.  You must call this function with "num_frames_ready"
   // arguments that only increase, not decrease, with time.  You would provide
   // this output to class OnlineIvectorFeature by calling its function
   // UpdateFrameWeights with the output.
+  //
+  // Returned frame-index is in pipeline frames from the pipeline start.
```
danpovey: Thanks.. can you please make it clear that someone should set first_decoder_frame to zero if they get a compilation error here?  Want to make it easier for people to recover from breaking their code.

nshmyrev: Maybe move it to the end and make it 0 by default? Then there will be no compilation issue. I was thinking about that but just not sure about the order of arguments.

danpovey:
That's against the style guide and I'd rather not make an exception.

danpovey:
Thanks!

JDongian:
@nshmyrev How did you make those amazing memory graphs? Curious if it's an open source tool.

nshmyrev:
@JDongian http://valgrind.org/docs/manual/ms-manual.html

JDongian:
Thank you!

rezame:
Hi
@danpovey @nshmyrev 
After this PR TCP port gets too many memory. I send a wave about 1.5 minutes length to TCP, It gets whole 8 G of my memory and crash!
My HCLG graph has about 300 MB size and my model has 30 MB size.
best regards

nshmyrev:
@rezame sorry, for me it works fine. Maybe you can provide a bit more information on how to reproduce your issue:

- Are you sure you recompiled everything after update?
- Are you using chain model with standard frame subsampling? What are your parameters in online.conf, etc?
- Does the crash happens with this particular file or with any file you are trying?

rezame:
@nshmyrev 
excuse me, many sorry about that, It's Ok for me.

PR #3529, opened 2019-08-21
Adding cmd.sh to script file (solving unbound variable $decode_cmd)

teinhonglo:
learn_lexicon_{greedy,bayesian}.sh


danpovey:
Thanks for the PR, but this is not how it's supposed to work.. the idea is you pass in that option as
 --cmd "$cmd" from the top-level run.sh.

PR #3540, opened 2019-08-26
[src,doc] Fix bug in new option of post-to-phone-post; skeleton of faq page

danpovey:


danpovey:
Thanks for noticing.. fixing it via PR.

PR #3557, opened 2019-08-31
[scripts, egs] new get_egs.sh did NOT work for aishell egs

han-xie:
I find that some recent updates of get_egs.sh did NOT work for aishell egs and reverted get_egs.sh back to commit 558f01023489751c71ee13b944f30fb0d30cfa4b.

After the reversion, it works fine for aishell egs.
--stage=7 option is also added for chain step in run.sh.


danpovey:
Thanks for figuring it out, but this is a bug that has already been fixed!  It's an awk compatibility issue (gawk versus regular awk).  That's why there is a conflict in this PR.
But thanks!

PR #3558, opened 2019-08-31
removed old Docker files

mdoulaty:
see docker in the root folder for the latest files


mdoulaty:
@danpovey this is not urgent, but when ever you've got sometime, please check this PR

PR #3561, opened 2019-09-03
fixed nested variable

KimJeongSun:
when I using queue.pl (using run.pl, there is no error), that part has "bad substitution" error.
I checked that problem solved after change the nested variable export method.
I would appreciate it if you review and apply.
Thanks.


kkm000:
@KimJeongSun, thank for the fix! It kinda fell off my radar.

PR #3565, opened 2019-09-03
[src] Temporarily turn off some model-collapsing code while investiga…

danpovey:
…ting a bug, possibly related to #3531 but likely only revealed by that.


danpovey:
This is a rather mysterious bug, I'm afraid.. if anyone saw inexplicably terrible nnet3 decodes since #3431 was merged, this could be related; but it's odd because #3431 was intended to fix precisely this problem in another setup.

Issue #3425, opened 2019-06-27
batched-wav-nnet3-cuda core dump when set --gpu-feature-extract=true


drawfish:
Below is the exec info:
````
batched-wav-nnet3-cuda --acoustic-scale=1.0 --add-pitch=true --beam=15.0 --feature-type=mfcc --frame-subsampling-factor=3 --frames-per-
chunk=150 --gpu-feature-extract=true --lattice-beam=8.0 --max-active=7000 --mfcc-config=config/mfcc.conf --online-pitch-config=config/pitch.conf --word-symbol-table=exp/nnet3_tdnn_online_game_transfer_final/gr
cc/wav.scp ark:/dev/null-kndiscount-5.0e-10/words.txt exp/nnet3_tdnn_online_game_transfer_final/tdnn-lstm/final.mdl exp/nnet3_tdnn_online_game_transfer_final/graph/HCLG.fst scp:data/dahua/mfc
batched-wav-nnet3-cuda --acoustic-scale=1.0 --add-pitch=true --beam=15.0 --feature-type=mfcc --frame-subsampling-factor=3 --frames-per-chunk=150 --gpu-feature-extract=true --lattice-beam=8.0 --max-active=7000
--mfcc-config=config/mfcc.conf --online-pitch-config=config/pitch.conf '--word-symbol-table=exp/nnet3_tdnn_online_game_transfer_final/graph/words.txt' exp/nnet3_tdnn_onl
ine_game_transfer_final/tdnn-lstm/final.mdl 'exp/nnet3_tdnn_online_game_transfer_final/graph/HCLG.fst' scp:data/dahua/mfcc/wav.scp ark:/dev/null
LOG (batched-wav-nnet3-cuda[5.5.402~1-53346]:SelectGpuId():cu-device.cc:223) CUDA setup operating under Compute Exclusive Mode.
LOG (batched-wav-nnet3-cuda[5.5.402~1-53346]:FinalizeActiveGpu():cu-device.cc:308) The active GPU is [6]: GeForce GTX 1080 Ti   free:10945M, used:227M, total:11172M, free/total:0.979682 version 6.1
LOG (batched-wav-nnet3-cuda[5.5.402~1-53346]:RemoveOrphanNodes():nnet-nnet.cc:948) Removed 5 orphan nodes.
LOG (batched-wav-nnet3-cuda[5.5.402~1-53346]:RemoveOrphanComponents():nnet-nnet.cc:847) Removing 11 orphan components.
[...]

PR #3568, opened 2019-09-04
[src] CUDA Online/Offline pipelines + light batched nnet3 driver

hugovbraun:
This is still WIP. Requires some cleaning, integrating the online mfcc into a separate PR (cf below), and some other things.

Implementing a low-latency high-throughput pipeline designed for online. It uses the GPU decoder, the GPU mfcc/ivector, and a new lean nnet3 driver (including nnet3 context switching on device).

- Online/Offline pipelines

The online pipeline can be seen as taking a batch as input, and then processing a very regular algorithm of calling feature extraction, nnet3, decoder, and postprocessing on that same batch, in a synchronous fashion (i.e. all of those steps will run when DecodeBatch is called. Nothing is sent to some async pipelines along the way). What happens when you run DecodeBatch is very regular, and because of that it is able to guarantee some latency constraints (because the way the code will be executed is very predicable). It also focus on being lean, avoiding reallocations or recomputations (such as recompiling nnet3).

The online pipeline takes care of computing [MFCC, iVectors], nnet3, decoder, postprocessing. It can either uses as input chunks of raw audio (and then compute mfcc->nnet3->decoder->postprocessing), or it can be called directly with mfcc features/ivectors (and then compute nnet3->decoder->postprocessing). The second possibility is used by the offline wrapper when use_online_ivectors=false.

The old offline pipeline is replaced by a new offline pipeline which is mostly a wrapper around the online pipeline. What it does is having an offline-friendly API (accepting full utterances as input instead of chunks), and has the possibility to pre-compute ivectors on the full utterance first (use_online_ivectors = false). It then calls the online pipeline internally to compute most of the work.  

The easiest way to test the online pipeline end-to-end is to call it through the offline wrapper for now, with use_online_ivectors = true. Please note that ivectors will be ignored for now in this full end-to-end online (i.e. when use_online_ivectors=true). That's because the GPU ivectors are not yet ready for online. However the pipeline code is ready. The offline pipeline with use_online_ivectors=false should be fully functional and returns the same WER than before.

- Light nnet3 driver designed for GPU and online

It includes a new light nnet3 driver designed for the GPU. The key idea is that it's usually better to waste some flops to compute things such as partial chunks or partial batches. For example for the last chunk (nframes=17) of an utterance, that chunk can be smaller than max_chunk_size (50 frames per default). It that case compiling a new nnet3 computation for that exact chunk size is slower than just running it for a chunk size of 50 and ignoring the invalid output. 

Same idea for batch_size: The nnet3 computation will always run a fixed minibatch size. It is defined as minibatch_size = std::min(max_batch_size, MAX_MINIBATCH_SIZE). MAX_MINIBATCH_SIZE is defined to be large enough to hide the kernel launch latency and increase the arithmetic intensity of the GEMMs, but not larger, so that partial batches will not be slowed down too much (i.e. avoiding to run a minibatch of size 512 where only 72 utterances are valid). MAX_MINIBATCH_SIZE is currently 128. We'll then run nnet3 multiple time on the same batch if necessary. If batch_size=512, we'll run nnet3 (with minibatch_size=128) four times.

The context-switch (to restore the nnet left and right context, and ivector) is done on device. Everything that needs context-switch is using the concept of channels, to be consistent with the GPU decoder.

Those "lean" approaches gave us better performance, and a drop in memory usage (total GPU memory usage from 15GB to 4GB for librispeech and batch size 500). It also removes the need for "high level" multithreading (i.e. cuda-control-threads).

- Parameters simplification

Dropping some parameters because the new code design doesn't require them (--cuda-control-threads, the drain size parameter). In theory the configuration should be greatly simplified (only --max-batch-size needs to be set, others are optional).

- Adding batching and online to GPU mfcc

The code in cudafeat/ is modifying the mfcc GPU code. MFCC features can now be batched and processed online (restoring a few hundreds frames of past audio for each new chunk). That code was implemented by @mcdavid109 (thanks!). We'll create a separate PR for this, it requires some cleaning, and a large part of the code is redundant with existing mfcc files.
GPU batched online ivectors and cmvn are WIP.

- Indicative measurements

When used with use_online_ivectors=false, that code reach 4,940 XRTF on librispeech/test_clean, with a latency around 6x realtime for max_batch_size=512 (latency would be lower with smaller max_batch_size).
One use case where that GPU pipeline can be used in a situation where only latency matters (and not throughput) is for instance on the jetson nano, where some initial runs were measured at 5-10x realtime latency for a single channel (max_batch_size=1) on librispeech/clean. Those measurements are indicative only - more reliable measurements will be done in the future.


pingpiang2019:
In cudadecoder/batched-static-nnet3.cc, we notice there is no right context padding at utt end. There used to be right context number of frames padded at utt end. Is this intended? We have some words dropped at the utt end.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
```diff
@@ -0,0 +1,310 @@
...
+
+  // Number of frames already computed in channel (before
+  // curr_batch_)
+  std::vector<int32> channel_frame_offset_;
+
+  // Parameters extracted from the models
+  int input_frames_per_chunk_;
+  int output_frames_per_chunk_;
+  BaseFloat seconds_per_chunk_;
+  BaseFloat samples_per_chunk_;
```
qijiaxing: Shall we use `int` for variable `samples_per_chunk`?

hugovbraun:
@pingpiang2019 The offline wrapper will take care of flushing the right context at the end. If you use it directly in online mode, then for now the best way is to send an extra chunk with silence in it - to flush the right context. It will be fixed at some point.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
hugovbraun: I'll change that. Thanks

ppamorim:
@hugovbraun Hi, any update on this. Thank you very much anyway!

hugovbraun:
@ppamorim yes, I've resumed work on this. Currently in the process in getting it ready for merge and more thorough testing. 
@pingpiang2019 FYI the right context is now flushed automatically internally (as long as you set the end_of_sequence boolean).Got some more testing to do and then I'll push the commit

pingpiang2019:
Does this mean we can't opt out of gpu_feature_extract anymore? Does the GPU feature extract support pitch? At some point, it doesn't, according to:
batched-wav-nnet3-cuda core dump when set --gpu-feature-extract=true #3425 
It seems our model with pitch crashes with similar backtrace and we can't fallback to cpu.

superliuwen:
Because there is no pitch cuda code, in current code we can set --gpu-feature-extract=false to use cpu to compute feature. But in your new code, this setting is removed. How can we still use cpu to compute features to support pitch? Thanks a lot!

hugovbraun:
Ok, that's an issue. Is pitch the only thing missing? We may need to add (back) the option for cpu fe for the offline pipeline.

superliuwen:
> Ok, that's an issue. Is pitch the only thing missing? We may need to add (back) the option for cpu fe for the offline pipeline.

@hugovbraun Thanks a lot for quick reply. So far we only find pitch missing, also feature extractions include plp, fbank are also not inlcude in GPU version, so CPU extraction is important in currect stage. Another choice is to use CPU to extract pitch and still using GPU to extract MFCC. And I read the code, it seems that offline pipeline and online pipline is the same in your new design, it means if the CPU fe is added in offline pipleline, the online can be supported, right? Thanks!

hugovbraun:
Ok, thanks for the info.
We can add CPU FE back into the offline pipeline to allow use of special features. The offline and online pipelines are the same for everything except for feature extraction (because features such as ivectors can take advantage of being able to see the full utterance, and not process things in an online way). 
For online support, we'll need to study things a bit more before making a decision.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -0,0 +1,375 @@
...
+void BatchedThreadedNnet3CudaOnlinePipeline::ReadParametersFromModel() {
+  // TODO remove that
+  feature_info_.reset(new OnlineNnet2FeaturePipelineInfo(config_.feature_opts));
+  feature_info_->ivector_extractor_info.use_most_recent_ivector = true;
+  feature_info_->ivector_extractor_info.greedy_ivector_extractor = true;
+
+  OnlineNnet2FeaturePipeline feature(*feature_info_);
+  // TODO clean following lines
+  input_dim_ = feature.InputFeature()->Dim();
+  ivector_dim_ = feature.IvectorFeature()->Dim();
```
twisteroidambassador: Does the current code require a MFCC+ivector model? I tried running `batched-wav-nnet3-cuda` with a model on hand that uses fbank without ivectors, and it SegFaults at this line.

al-zatv: my fbank-based model works fine (using batched-wav-nnet3-cuda executable).

al-zatv:
This implementation is little bit slower on my test model, than offline version from main kaldi repository.
Model is fbank40x based (without i-vectors), 9 tdnnf layers, HCLG 1.5 Gb. GPU is 1080Ti.  Tested via batched-wav-nnet3-cuda executable.

offline version (currently in main kaldi repo) : RTx=640
this implementation: RTx=565

Also, it is less than 4,940 XRTF that you mention earlier. May be it is caused by model (it is different from librispeech one; neural net is smaller but graph is bigger), or may be 1080Ti is worse than your GPU.

al-zatv:
in src/cudadecoderbin/batched-wav-nnet3-cuda.cc:229
```
        if (iterations > 0) {
          // make key unique for each iteration
          key = std::to_string(iter) + "-" + key;
        }
```
probably, should be `if (iterations > 1)`

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -0,0 +1,375 @@
...
+
+  // delete data used for decoding that corr_id
+  for (int32 i = 0; i < list_channels_last_chunk_.size(); ++i) {
+    uint64_t ichannel = list_channels_last_chunk_[i];
+    CorrelationID corr_id = list_corr_id_last_chunk_[i];
+    int32 ndeleted = corr_id2channel_.erase(corr_id);
+    KALDI_ASSERT(ndeleted == 1);
+    KALDI_ASSERT(thread_pool_->tryPush(
+        {&BatchedThreadedNnet3CudaOnlinePipeline::FinalizeDecodingWrapper, this,
+         ichannel, corr_id}));
```
twisteroidambassador: When recognizing a large batch of short utterances, with a low number of `--cuda-worker-threads` (8), after a while the above assert fails. Apparently this is because the task queues filled up.

It works if the assert is changed to basically `while (!tryPush()) {usleep(1000)};`, and this should exert backpressure to slow down the decode loop properly. Does this look reasonable?

hugovbraun: I'll update this. Thanks.

twisteroidambassador:
`batched-wav-nnet3-cuda` from this PR would crash sporadically. I'm using the same data, model and command line each time, some times it runs happily to completion, and sometimes it terminates before recognizing the first utterance:

	LOG (batched-wav-nnet3-cuda[5.5.605-568892b]:SelectGpuId():cu-device.cc:223) CUDA setup operating under Compute Exclusive Mode.
	LOG (batched-wav-nnet3-cuda[5.5.605-568892b]:FinalizeActiveGpu():cu-device.cc:308) The active GPU is [0]: Tesla V100-PCIE-16GB	free:15507M, used:653M, total:16160M, free/total:0.959574 version 7.0
	LOG (batched-wav-nnet3-cuda[5.5.605-568892b]:RemoveOrphanNodes():nnet-nnet.cc:948) Removed 1 orphan nodes.
	LOG (batched-wav-nnet3-cuda[5.5.605-568892b]:RemoveOrphanComponents():nnet-nnet.cc:847) Removing 2 orphan components.
	LOG (batched-wav-nnet3-cuda[5.5.605-568892b]:Collapse():nnet-utils.cc:1472) Added 1 components, removed 2
	LOG (batched-wav-nnet3-cuda[5.5.605-568892b]:CheckAndFixConfigs():nnet3/nnet-am-decodable-simple.h:123) Increasing --frames-per-chunk from 50 to 51 to make it a multiple of --frame-subsampling-factor=3
	ERROR (batched-wav-nnet3-cuda[5.5.605-568892b]:AddMatMat():cu-matrix.cc:1317) cublasStatus_t 14 : "CUBLAS_STATUS_INTERNAL_ERROR" returned from 'cublas_gemm(GetCublasHandle(), (transB==kTrans? CUBLAS_OP_T:CUBLAS_OP_N), (transA==kTrans? CUBLAS_OP_T:CUBLAS_OP_N), m, n, k, alpha, B.data_, B.Stride(), A.data_, A.Stride(), beta, data_, Stride())'

	[ Stack-Trace: ]
	/home/kaldi-user/kaldi/src/lib/libkaldi-base.so(kaldi::MessageLogger::LogMessage() const+0x8b7) [0x2b918602ecfd]
	./batched-wav-nnet3-cuda(kaldi::MessageLogger::LogAndThrow::operator=(kaldi::MessageLogger const&)+0x11) [0x42719f]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudamatrix.so(kaldi::CuMatrixBase<float>::AddMatMat(float, kaldi::CuMatrixBase<float> const&, kaldi::MatrixTransposeType, kaldi::CuMatrixBase<float> const&, kaldi::MatrixTransposeType, float)+0x30d) [0x2b918307460d]
	/home/kaldi-user/kaldi/src/lib/libkaldi-nnet3.so(kaldi::nnet3::TdnnComponent::Propagate(kaldi::nnet3::ComponentPrecomputedIndexes const*, kaldi::CuMatrixBase<float> const&, kaldi::CuMatrixBase<float>*) const+0x200) [0x2b91824aaa4e]
	/home/kaldi-user/kaldi/src/lib/libkaldi-nnet3.so(kaldi::nnet3::NnetComputer::ExecuteCommand()+0x1bc) [0x2b91823c77fe]
	/home/kaldi-user/kaldi/src/lib/libkaldi-nnet3.so(kaldi::nnet3::NnetComputer::Run()+0x158) [0x2b91823c8c44]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudadecoder.so(kaldi::cuda_decoder::BatchedStaticNnet3::RunNnet3(kaldi::CuMatrix<float>*, int)+0x4f7) [0x2b91813db1c7]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudadecoder.so(kaldi::cuda_decoder::BatchedStaticNnet3::RunBatch(std::vector<int, std::allocator<int> > const&, std::vector<float*, std::allocator<float*> > const&, int, std::vector<float*, std::allocator<float*> > const&, std::vector<int, std::allocator<int> > const&, std::vector<bool, std::allocator<bool> > const&, kaldi::CuMatrix<float>*, std::vector<std::vector<std::pair<int, float*>, std::allocator<std::pair<int, float*> > >, std::allocator<std::vector<std::pair<int, float*>, std::allocator<std::pair<int, float*> > > > >*)+0x17e) [0x2b91813dc91a]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudadecoder.so(kaldi::cuda_decoder::BatchedThreadedNnet3CudaOnlinePipeline::RunNnet3(std::vector<int, std::allocator<int> > const&, std::vector<float*, std::allocator<float*> > const&, int, std::vector<int, std::allocator<int> > const&, std::vector<bool, std::allocator<bool> > const&, std::vector<float*, std::allocator<float*> > const&)+0x36) [0x2b91813c88b6]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudadecoder.so(kaldi::cuda_decoder::BatchedThreadedNnet3CudaOnlinePipeline::DecodeBatch(std::vector<unsigned long, std::allocator<unsigned long> > const&, std::vector<float*, std::allocator<float*> > const&, int, std::vector<int, std::allocator<int> > const&, std::vector<float*, std::allocator<float*> > const&, std::vector<bool, std::allocator<bool> > const&, std::vector<int, std::allocator<int> >*)+0xaa) [0x2b91813c9eea]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudadecoder.so(kaldi::cuda_decoder::BatchedThreadedNnet3CudaOnlinePipeline::DecodeBatch(std::vector<unsigned long, std::allocator<unsigned long> > const&, std::vector<kaldi::SubVector<float>, std::allocator<kaldi::SubVector<float> > > const&, std::vector<bool, std::allocator<bool> > const&)+0x686) [0x2b91813ca5dc]
	/home/kaldi-user/kaldi/src/lib/libkaldi-cudadecoder.so(kaldi::cuda_decoder::BatchedThreadedNnet3CudaPipeline::ComputeTasks()+0xba) [0x2b91813d8a22]
	./batched-wav-nnet3-cuda(std::thread::_Impl<std::_Bind_simple<std::_Mem_fn<void (kaldi::cuda_decoder::BatchedThreadedNnet3CudaPipeline::*)()> (kaldi::cuda_decoder::BatchedThreadedNnet3CudaPipeline*)> >::_M_run()+0x29) [0x426653]
	/lib64/libstdc++.so.6(+0xb5070) [0x2b91b2685070]
	/lib64/libpthread.so.0(+0x7e65) [0x2b918d00ce65]
	/lib64/libc.so.6(clone+0x6d) [0x2b91b2beb88d]

Any hints on how to pinpoint the problem?

hugovbraun:
@twisteroidambassador what kind of features are you using? Any chance you are using a model without ivectors?

hugovbraun:
@twisteroidambassador Just fixed a bug. However the case "model with mfcc but without ivectors" is still untested so there may be others.

twisteroidambassador:
@hugovbraun I was in fact using a f-bank model without i-vectors, so I had to transplant the original spectral feature code supporting fbank into the PR, and add a bunch of `if (has_ivector_)` checks.

I see that master now has new gpu online feature extraction natively supporting fbank. I'll try the latest version of this PR, see whether any more i-vector checks are necessary, and report back.

twisteroidambassador:
Attached patch that allows all three executables under `cudadecoderbin` to run with a fbank model without ivectors. It adds a bunch of ivector checks, and also fixes `model_frequency` in the online pipeline.

[kaldi-gpu-online-ivector.diff.txt](https://github.com/kaldi-asr/kaldi/files/4218530/kaldi-gpu-online-ivector.diff.txt)

However, the recognition result of `batched-wav-nnet3-cuda-online` are wrong, while the old executables produce the correct result.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
hugovbraun: Yes, it's not tested yet without ivectors

hugovbraun:
@twisteroidambassador by old executables do you mean cudadecoderbin/batched-wav-nnet3-cuda and cudadecoderbin/batched-wav-nnet3-cuda2, or just the version without the "2"?

twisteroidambassador:
@hugovbraun both `batched-wav-nnet3-cuda` and `batched-wav-nnet3-cuda2` produce correct results, while `batched-wav-nnet3-cuda-online` does not.

hugovbraun:
@twisteroidambassador Ok. batched-wav-nnet3-cuda2 calls the online pipeline behind the scene, to the exception of the feature extraction if --use-online-features is set to false (default). Could you try running batched-wav-nnet3-cuda2 with --use-online-features=true to try to isolate the bug?

hugovbraun:
@twisteroidambassador I can repro the problem locally with a fbank model with --use-online-features=true. Looks like we have a bug with our online fbank code. We'll take a look

twisteroidambassador:
@hugovbraun Yes, I can confirm that adding `--use-online-features=true` to the invocation of `batched-wav-nnet3-cuda2` produces incorrect recognition results, consistent with the results from `batched-wav-nnet3-cuda-online`.

hugovbraun:
@twisteroidambassador can you retry?

twisteroidambassador:
@hugovbraun Just tried the latest version, same behavior as before. `batched-wav-nnet3-cuda2` with `--use-online-features=true` and `batched-wav-nnet3-cuda-online` still produces incorrect recognition result.

hugovbraun:
@twisteroidambassador is the binary crashing? or is it always running successfully and returning incorrect results? Anything special about the model? We are currently testing with cvte (https://kaldi-asr.org/models/m2), do you know if this is comparable to what you have ?
Thanks for the feedback!

twisteroidambassador:
@hugovbraun The binary is not crashing, the runs look successful, it's just the results are wrong. Looking at the generated lattices, the wrong ones are much shorter.

I think the model is just a fbank-based chain model with a frame skip of 3, nothing special. I'll take a closer look later.

twisteroidambassador:
I downloaded the pre-trained cvte model, and as far as I can see the biggest difference between that and our own model is that cvte is trained on 16k audio, while ours is trained on 8k. I can run the cvte model by swapping out the model files, and using 16k audio.

And with the cvte model, the three executables produce four different results (`batched-wav-nnet3-cuda2` with both online and offline features). None of them are close to the correct result since the audio is of rather poor quality, and upsampled from 8k, so this is probably a poor test.

hugovbraun:
Thank you for your comments. I guess your model is not available publicly? 
Also @danpovey on this page https://kaldi-asr.org/models/m2 the cvte model shows a WER of 8.25%, is this indeed WER or CER? Just checking

hugovbraun:
@twisteroidambassador could you run the feature extraction binaries on your data and check that offline/online return the same values? 

```
src/cudafeatbin/compute-online-feats-batched-cuda --batch-size=80 --num-channels=80 --config=model/conf/online.conf  scp:path/to/wav.scp ark,scp:ivector-batched.ark,ivector-batched.scp ark,scp:feats-batched.ark,feats-batched.scp 
src/cudafeatbin/compute-online-feats-cuda --config=model/conf/online.conf  scp:path/to/wav.scp ark,scp:ivector.ark,ivector.scp ark,scp:feats.ark,feats.scp 
src/featbin/compare-feats scp:feats.scp scp:feats-batched.scp
```
 Thanks!

hugovbraun:
@danpovey we still have an accuracy issue with the model @twisteroidambassador is using, but could we start the review process?

src/cudadecoder/Makefile
```diff
@@ -13,8 +13,12 @@ endif
...
 TESTFILES =
 
-OBJFILES = batched-threaded-nnet3-cuda-pipeline.o decodable-cumatrix.o \
-           cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o
+OBJFILES = cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o \
+	   batched-threaded-nnet3-cuda-online-pipeline.o \
+	   batched-threaded-nnet3-cuda-pipeline.o \
+	   batched-threaded-nnet3-cuda-pipeline2.o \
+	   batched-static-nnet3.o batched-static-nnet3-kernels.o \
+	   deprecated/decodable-cumatrix.o
```
danpovey: Hm, I'm concerned that by messing with the standard directory structure, we may make maintaining the CMake and Windows builds harder.

src/cudadecoder/batched-static-nnet3-kernels.cu
```diff
@@ -0,0 +1,167 @@
...
+			int n_frames_in_context =
+				batch_assign.n_frames_already_in_context;
+			int n_frames_to_set = n_frames_in_context + params.total_nnet_right_context;  
+//			printf("will set %i frames, %i in context \n", n_frames_to_set, n_frames_in_context);
+
+			for (int iframe = blockIdx.y; iframe < n_frames_to_set;
+					iframe += gridDim.y) {
+				for (int idim = threadIdx.x; idim < params.input_dim; idim += blockDim.x) {
+					if(iframe < n_frames_in_context) {
+float f=						d_batch_slot_with_context[iframe * params.d_batch_with_context_frame_stride+ idim] = d_channel_context[iframe*params.d_all_context_frames_frame_stride + idim];
```
danpovey: very strange whitespace-- perhaps your tab settings were wrong?  Please make sure to convert all tabs to spaces, as google style guide mandates.  And left-justified `float f=` is wrong, plus shoul dbe space after =.

LeviBarnes:
@twisteroidambassador, does your model use CMVN?

twisteroidambassador:
@hugovbraun I tried to run `compute-online-feats-batched-cuda` like this:

    ./compute-online-feats-batched-cuda --batch-size=100 --num-channels=100 --feature-type=fbank --fbank-config=fbank.conf scp:...

and it aborts with this message:

    ASSERTION_FAILED (compute-online-feats-batched-cuda[5.5.657-30c41]:main():compute-online-feats-batched-cuda.cc:280) Assertion failed: (wave_data.SampFreq() == sample_freq)

Adding some log output before this line reveals that `sample_freq == 16000` when our audio data is 8k. It looks like the feature pipeline has the sampling rate wrong.

Our `fbank.conf` is:

	--window-type=hamming
	--sample-frequency=8000 
	--low-freq=64
	--high-freq=3800
	--dither=1
	--num-mel-bins=80

twisteroidambassador:
@LeviBarnes No, our model does not use CMVN.

twisteroidambassador:
Found the problem. Here: https://github.com/kaldi-asr/kaldi/blob/a2573871ba185b8fd83ec5e66270a9a2301e4300/src/cudafeat/online-batched-feature-pipeline-cuda.cc#L39-L41

The pipeline always grabs `frame_opts_` from `info_.mfcc_opts` even when using fbank. Adding some `if`s here fixed the wrong recognition results as far as I can see.

(I tried to come up with a clever way of only using one set of if statements here, since just a few lines below is another set of ifs. It would have worked in Python; too bad C++ is strongly typed.)

hugovbraun:
Thanks a lot @twisteroidambassador for finding the problem! We can take care of fixing this in cudafeat, except if you are interested in upstreaming it yourself

src/cudadecoder/Makefile
hugovbraun: Ok, I'll switch it back to the cudadecoder/ directory with some comments. We'll remove those files in a few months

src/cudadecoder/batched-static-nnet3-kernels.cu
hugovbraun: My guess is my new clang-format setup doesn't work properly with .cu files. I'll fix those

LeviBarnes:
Good catch, @twisteroidambassador. When you make that fix, can you then compare features from the batched and non-batched cuda pipelines with your model?

hugovbraun:
We should open a separate github issue now that @twisteroidambassador has localized the problem in a code already merged in master

src/cudadecoder/batched-static-nnet3-kernels.cu
hugovbraun: It was using llvm style for cuda files. Back to google style now.

twisteroidambassador:
[kaldi-cudafeat-samprate.diff.txt](https://github.com/kaldi-asr/kaldi/files/4254727/kaldi-cudafeat-samprate.diff.txt)

Here's a patch that fixes cudafeat. Strange that when doing an incremental `make` the computed features are still wrong, but a complete `make clean` and rebuild fixed it properly.

@LeviBarnes I compared features from `featbin/compute-fbank-feats`, `cudafeatbin/compute-online-feats-cuda` and `cudafeatbin/compute-online-feats-batched-cuda`, and all their outputs are similar.

LeviBarnes:
I submitted PR https://github.com/kaldi-asr/kaldi/pull/3958 which is similar to yours, @twisteroidambassador. Thanks for the fix!

twisteroidambassador:
I was just thinking, what will happen if a correlation ID is repeated in a batch? i.e. if a batch is [(corrID 1, chunk 0-50), (corrID 1, chunk 50-100), (corrID 1, chunk 100-150)], does it still work correctly, error out, or do something unexpected?

twisteroidambassador:
Also, is there an expected accuracy drop with the new online pipeline? I compared the recognition results between `cudadecoderbin/batched-wav-nnet3-cuda-online` and `cudafeatbin/compute-online-feats-batched-cuda` -> `nnet3bin/nnet3-latgen-faster-batch`, and there are some substantial differences. Presumably the only difference between these two pipelines is whether the decoder is run on GPU or CPU.

hugovbraun:
@twisteroidambassador You need to use at most one chunk per corr_id per batch. Looks like I forgot to put a comment about this. We'll add an assert at some point.
Regarding WER, we usually see a difference in the order of magnitude of 0.0x of a percentile. For instance, For the librispeech model we use, we see WER around 5.54% instead of 5.50%. That's something we plan to look at at some point. If the difference you're seeing is larger, you can try increasing --max-active on the GPU version.

twisteroidambassador:
@hugovbraun I have been using the exact same config options for GPU and CPU decoders, including setting a low-ish `--max-active`. Does the GPU decoder need a higher `--max-active` value to achieve the same accuracy as the CPU decoder?

danpovey:
I think some documentation on how to tune the GPU decoder would be good, if
it doesn't already exist.

al-zatv:
@hugovbraun I have a question about receiving partial results. My use case is pretty usual: to listen to audio channels, to send partial results back often, and to show final results after it. Like multi-channel online2-tcp-nnet3-decode-faster. 
But I don't see interface to take partial results from BatchedThreadedNnet3CudaOnlinePipeline. I see CudaDecoder::GetBestPath(), so I think my task is solvable. I wrote BatchedThreadedNnet3CudaOnlinePipeline::GetBestPath as follows, but it is slow and fails often. How to do it right?
`
void BatchedThreadedNnet3CudaOnlinePipeline::GetBestPath(
    const std::vector<CorrelationID> &corr_ids,
    std::vector<Lattice *> &fst_out_vec, bool use_final_probs) 
{
    std::vector<ChannelId> channels;
    channels.reserve(corr_ids.size());
  
    for (int i = 0; i < corr_ids.size(); ++i) {
        int corr_id = corr_ids[i];
        auto it = corr_id2channel_.find(corr_id);
        KALDI_ASSERT(it != corr_id2channel_.end());
        int ichannel = it->second;
        channels.push_back(ichannel);
    }
    
    cuda_decoder_->GetBestPath(channels,fst_out_vec,false);

}
`

twisteroidambassador:
With some more testing, I determined that with my current model and config (specifically beam, lattice-beam and max-active options), the limiting factor is beam. After loosening the constraints gradually until the one-best result no longer changes, for some utterances the result from GPU decoder is still not the same as that from the CPU decoder.

So, I guess my question is: Should I expect the GPU and CPU decoder to produce identical output for the same utterance, either when constrained to the same pruning options such as beam, or when unconstrained?

danpovey:
No they are not identical.  Some slightly different design decisions were
made in the GPU decoder.

twisteroidambassador:
Is there any documentation, discussion, mail archive, etc. where I can learn about the algorithmic and design differences between the GPU and CPU decoder? I'll also try reading the code, but a higher level description would be very welcome.

hugovbraun:
@twisteroidambassador @danpovey You are right, we need to write an up to date "How to use" guide for the decoder itself.
In short:
* Regarding parameters the main difference concerns --max-active. For the CPU decoder, that parameter will determine the max number of FST _states_ kept at each iteration. For the GPU decoder, that parameter will determine the max number of FST _arcs_ kept at each iteration. It is usually necessary to set the GPU parameter higher
* All other decoding parameters should behave the same way. @twisteroidambassador Your comment about beam is interesting. Increasing --beam (vs cpu) improved WER but not --max-active?
* For (very) challenging cases (bad audio) or an union ["challenge case" and "high beam"] you may need to rise the queue capacities with --main-q-capacity and --aux-q-capacity (--help will help you with what to set, try 2*default). Otherwise the mechanism responsible for overflow prevention will trigger and reduce the beam. It is usually not necessary to manually change the capacity of those queues.

The final results are not exactly the same as the CPU ones (due to the reasons listed above), but are described as "virtually identical" by our partners. 

@al-zatv implementing an efficient way to get back partial results after each output frames is next on our list (with endpointing, which is a similar change). As of today there's no real way to get back partial results (you could rely on the "normal" GetRawLattice, but it will be slow)

twisteroidambassador:
@hugovbraun I reran the tests to be sure. Yes, starting from my "stock" settings (`--max-active=7000 --beam=6.0 --lattice-beam=8.0`), raising --max-active to 14000 does not change the recognition results (cpu vs cpu, gpu vs gpu), the one-best results are identical within random variations. (By random variation, I mean when running batched-wav-nnet3-cuda-online repeatedly with same everything, the one-best results are not completely identical, out of the ~400 utterances usually 2~3 are different.) Raising --beam to 9.0 does result in significantly different one-best results. So I believe the --beam setting is the limiting factor in my stock settings.

Given that, it's strange that at stock settings, one-best cpu vs gpu is quite different, at ~50 utterances. And looking at the generated lattice files, when --determinize-lattice=false the gpu ark is at ~5MB while the cpu ark is at ~50MB, and with --determinize-lattice=true the gpu ark is ~700KB while the cpu ark is ~1.4MB. (These are binary ark files.)

As for --main-q-capacity and --aux-q-capacity, I have not seen any related warning messages printed to the console, so I don't think those were hit.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
```diff
@@ -0,0 +1,366 @@
...
+  }
+};
+
+class BatchedThreadedNnet3CudaOnlinePipeline {
+ public:
+  using CorrelationID = uint64_t;
+  BatchedThreadedNnet3CudaOnlinePipeline(
+      const BatchedThreadedNnet3CudaOnlinePipelineConfig &config,
+      const fst::Fst<fst::StdArc> &decode_fst,
+      const nnet3::AmNnetSimple &am_nnet, const TransitionModel &trans_model)
```
twisteroidambassador: From what I can see, both am_nnet and trans_model must remain valid for the entire lifetime of the pipeline, correct?

hugovbraun: Yes

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
```diff
@@ -0,0 +1,366 @@
...
+      : config_(config),
+        max_batch_size_(config.max_batch_size),
+        trans_model_(&trans_model),
+        am_nnet_(&am_nnet),
+        word_syms_(NULL) {
+    config_.compute_opts.CheckAndFixConfigs(am_nnet_->GetNnet().Modulus());
```
twisteroidambassador, line 135: Is this missing a call to `config_.CheckAndFixConfigs()`? Without it, the default value of --cuda-worker-threads (-1) will lead to an error.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -0,0 +1,478 @@
...
+      }
+      {
+        std::lock_guard<std::mutex> lk(stdout_m_);
+        KALDI_LOG << "OUTPUT: " << oss.str();
+      }
+    }
+
+    // if ptr set and if callback func callable
+    if (callback && *callback) {
+      (*callback)(dlat);
```
twisteroidambassador, line 463: Would it be better if the callback is always called, even when the lattice is empty? Not knowing whether the callback will fire depending on the contents of the utterance makes programming against it difficult.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
hugovbraun: Thanks for catching this! I'll update it

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
hugovbraun: I see. You're right. I'll take those lines out of the if

twisteroidambassador:
How does the pipeline handle acoustic / language model scaling? When the `--acoustic-scale` argument is given on the command line, are the scores scaled accordingly during nnet3 computations and decoding, and if so, is the lattice scaled back before output?

twisteroidambassador:
And a related question. Since `BatchedThreadedNnet3CudaOnlinePipelineConfig` uses `NnetSimpleComputationOptions`, it inherits its description on the `acoustic-scale` option, which says "caution: is a no-op if set in the program nnet3-compute". Does this caution also apply to the other nnet3-* programs?

(I was looking through nnet3-latgen-faster-batch.cc for scaling operations, to see whether that explains why the results are different from the gpu pipeline. I found several places where it seems to scale inversely w.r.t. acoustic_scale before output, but couldn't find where it scales proportionally to acoustic_scale. It was rather confusing.)

hugovbraun:
@twisteroidambassador the acoustic scale is applied there: https://github.com/hugovbraun/kaldi/blob/0b865f664e4e2e71bae0ecb5c94436047937ea09/src/cudadecoder/batched-static-nnet3.cc#L288

danpovey:
Did not realize this was ready.  Merging.

twisteroidambassador:
Just found out that, when calling DecodeBatch(wave_samples), if one of the SubVectors in wave_samples has length 0, the next call to RunNnet3 will freeze.
It can be worked around easily by passing a SubVector of length 1 instead, but that feels rather awkward.
Since the only way to end an utterance is to pass in an additional chunk, maybe it should accept 0-length chunks? Or maybe there can be another method to end an utterance without requiring a chunk?

hugovbraun:
I think you're right, sending a chunk of length 0 should be valid and allow you to end an utterance. We'll make the change. Thanks for reporting that.

For future issues, it may be better to create a new github issue to simplify tracking.

kli017:
hello If is it possible to modifiy the decode to a sever which could support multiply request from different client?

Issue #3543, opened 2019-08-28
Enable tighter control of downloaded dependencies


iskunk:
Hello,

I am using Kaldi in a large organization, with fairly tight rules concerning how third-party software is used, built, etc. The approach that the Kaldi build system uses to handle dependencies like sctk, OpenFST et al. is convenient to help the user get up and running quickly, but in a regulated environment, it's a bit too fast and loose for comfort.

I would like to propose a couple points of new functionality, across the various scripts under `tools/`, to the end of maintaining the convenient behavior currently implemented, while allowing it to be easily switched into a stricter mode more suitable for such an environment:

1. Check whether an environment variable `WGET` is set; if so, then use it, otherwise default its value to `wget`. Not only does this allow using a wget executable not in one's PATH, this allows one to set a value of `false`, effectively nulling the ability of the build system to download anything.

2. Check an environment variable `DOWNLOAD_DIR` (first-draft name); if set, then copy the required source tarball from there instead of downloading same. (This way, the tarballs can be downloaded and vetted in advance of the Kaldi build.)

I have an implementation of this in the current Git tree. [Here is a preliminary patch](https://github.com/kaldi-asr/kaldi/files/3548616/kaldi-dep-dl-revamp.patch.txt), for initial review. (I can prepare a PR if desired, but would need to know how it should be structured, etc.)

There are a few related changes included in the above:
* Various URLs have been updated to the `https://` form where supported;
* Dropped the `--no-check-certificate` option wherever it was not needed (this is an option that you definitely don't want to use unless absolutely necessary);
* Added variables containing package versions, and then replaced various copy-pasted version strings with these, to simplify package version bumps.

danpovey:
That all looks fine to me.  You can go ahead and make a PR.

iskunk:
I'm happy that it's to your liking :-)

I will work on getting a PR together, and hope to have that submitted within a few days.

danpovey:
Resolved in #3573

PR #3573, opened 2019-09-05
[scripts] Enable tighter control of downloaded dependencies (#3543)

iskunk:
Pull request submitted as discussed in #3543.

Summary of changes:
* Defined environment variables `GIT` and `WGET` which can be set to appropriate invocations of `git(1)` and `wget(1)` respectively that will be used by the various download scripts. Alternately, they can be set to a value like `false` to disable use of the respective command. If the variables are undefined, the scripts set the expected default value of `git` and `wget`.
* Defined environment variable `DOWNLOAD_DIR` which can be set to a directory containing the various tarballs and package files downloaded by the Kaldi scripts. If the variable is set, then the scripts will copy the necessary files from this directory instead of downloading them.
* Upgraded `http://` URLs to `https://` where supported. In a few cases, minor tweaks have been made to reflect current upstream usage (e.g. `readthedocs.org` -> `readthedocs.io`).
* Removed `--no-check-certificate` option from `wget` invocations, to avoid defeating HTTPS security. (I did not find any cases where this option was needed in order to allow a successful download.)
* Moved package version strings into variables to simplify changing their value.
* Fixed minor script irregularities as encountered.

I have tested most of these changes (a handful cannot be run in my testing environment). The general intent is that, if the aforementioned environment variables remain unset, then the scripts will continue to work as they have previously.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -5,11 +5,16 @@ CC ?= gcc        # used for sph2pipe
 # CXX = clang++  # Uncomment these lines...
 # CC = clang     # ...to build with Clang.
 
+WGET ?= wget
+
 # Note: OpenFst requires a relatively recent C++ compiler with C++11 support,
 # e.g. g++ >= 4.7, Apple clang >= 5.0 or LLVM clang >= 3.3.
 OPENFST_VERSION ?= 1.6.7
 CUB_VERSION ?= 1.8.0
 OPENBLAS_VERSION ?= 0.3.5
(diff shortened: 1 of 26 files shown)


danpovey:
I think I'll merge this and if any errors crop up we'll fix them as they come.

iskunk:
Thanks @danpovey. Feel free to pull me in on any issues that arise as a result of these changes.

PR #3579, opened 2019-09-09
WIP: Chime5/6 with Paderborn guided source separation

vimalmanohar:
This is in preparation for a new Chime recipe.

@sw005320


jtrmal:
@sw005320 @vimalmanohar  it would be a shame to let this die off

danpovey:
We can consider just merging it with any problems it has, assuming it
doesn't break anything else.
Let's get a comment from Shinji about this, at least.

jtrmal:
I don't object, but it should be at least runnable, I think.
y.

danpovey:
We could just rely on users to fix the problems, maybe?  Anyway it's the
kind of task that will only attract experts.

aarora8:
Hi,
Our recent chime6 setup in the kaldi master is the same as this one. This
setup was also for chime6 baseline
Thanks,
Ashish

jtrmal:
ashish, thanks -- are you saying this can be closed as the interesting bits
have been ingested to other recipe?
y.

aarora8:
Yeah, it can be closed as the interesting bits have been ingested to chime6
recipe.

jtrmal:
thanks!
y.

PR #3580, opened 2019-09-10
CMake Build System

cloudhan:
This suppress PR #3100 

Summary:
This is still in early stage. I now managed to build on Ubuntu 16.04 with CMake 3.15. 
You should be able to build with
```bash
cd tools
make install -j32
mkdir -p ../build && cd ../build
cmake .. && cmake --build . -- -j32
```

### STAGE 1:
- [x] Almost everything build.
- [x] Install targets
- [x] Special handling test executables
- [x] Export targets
- [x] ~~kaldi10 branch~~ DO NOT have time for it now.
- [x] ~~fix gmmbin target requirement cleverer.~~ NOT worth it now.
- [x] version.h

### STAGE 2:
- [x] ~~Remove the generator.~~ Generate on the fly and keep script for now.
- [x] ~~Move specific target requirement into sub-directories' `CMakeLists.txt`~~ **WONTFIX** due to generator script is still in using.
- [x] Handling 3rdparty libraries
    - [x] In repo build OpenFST libraries on Windows
    - [x] Require user to build OpenFST manually on other platform
    - [x] use in repo source for cub


cloudhan:
@galv Have some suggestion. Current the config.in is empty

<details><summary>folder structure</summary>

```
├── include
│   └── kaldi
│       ├── base
│       │   ├── io-funcs.h
│       │   ├── io-funcs-inl.h
│       │   ├── kaldi-common.h
│       │   ├── kaldi-error.h
│       │   ├── kaldi-math.h
│       │   ├── kaldi-types.h
│       │   ├── kaldi-utils.h
│       │   └── timer.h
│       ├── chain
│       │   ├── chain-datastruct.h
│       │   ├── chain-den-graph.h
│       │   ├── chain-denominator.h
│       │   ├── chain-generic-numerator.h
│       │   ├── chain-kernels-ansi.h
│       │   ├── chain-numerator.h
│       │   ├── chain-supervision.h
│       │   ├── chain-training.h
│       │   └── language-model.h
│       ├── cudadecoder
│       │   ├── batched-threaded-nnet3-cuda-pipeline.h
│       │   ├── cuda-decodable-itf.h
│       │   ├── cuda-decoder-common.h
│       │   ├── cuda-decoder.h
│       │   ├── cuda-decoder-kernels.h
│       │   ├── cuda-decoder-kernels-utils.h
│       │   ├── cuda-fst.h
│       │   ├── decodable-cumatrix.h
│       │   └── thread-pool.h
│       ├── cudafeat
│       │   ├── feature-online-cmvn-cuda.h
│       │   ├── feature-spectral-cuda.h
│       │   ├── feature-window-cuda.h
│       │   ├── online-cuda-feature-pipeline.h
│       │   ├── online-ivector-feature-cuda.h
│       │   └── online-ivector-feature-cuda-kernels.h
│       ├── cudamatrix
│       │   ├── cu-allocator.h
│       │   ├── cu-array.h
│       │   ├── cu-array-inl.h
│       │   ├── cublas-wrappers.h
│       │   ├── cu-block-matrix.h
│       │   ├── cu-common.h
│       │   ├── cu-compressed-matrix.h
│       │   ├── cu-device.h
│       │   ├── cu-kernels-ansi.h
│       │   ├── cu-kernels.h
│       │   ├── cu-math.h
│       │   ├── cu-matrixdim.h
│       │   ├── cu-matrix.h
│       │   ├── cu-matrix-inl.h
│       │   ├── cu-matrix-lib.h
│       │   ├── cu-packed-matrix.h
│       │   ├── cu-rand.h
│       │   ├── cu-sparse-matrix.h
│       │   ├── cu-sp-matrix.h
│       │   ├── cu-tp-matrix.h
│       │   ├── cu-value.h
│       │   └── cu-vector.h
│       ├── decoder
│       │   ├── biglm-faster-decoder.h
│       │   ├── decodable-mapped.h
│       │   ├── decodable-matrix.h
│       │   ├── decodable-sum.h
│       │   ├── decoder-wrappers.h
│       │   ├── faster-decoder.h
│       │   ├── grammar-fst.h
│       │   ├── lattice-biglm-faster-decoder.h
│       │   ├── lattice-faster-decoder.h
│       │   ├── lattice-faster-online-decoder.h
│       │   ├── lattice-simple-decoder.h
│       │   ├── simple-decoder.h
│       │   └── training-graph-compiler.h
│       ├── feat
│       │   ├── feature-common.h
│       │   ├── feature-common-inl.h
│       │   ├── feature-fbank.h
│       │   ├── feature-functions.h
│       │   ├── feature-mfcc.h
│       │   ├── feature-plp.h
│       │   ├── feature-spectrogram.h
│       │   ├── feature-window.h
│       │   ├── mel-computations.h
│       │   ├── online-feature.h
│       │   ├── pitch-functions.h
│       │   ├── resample.h
│       │   ├── signal.h
│       │   └── wave-reader.h
│       ├── fstext
│       │   ├── context-fst.h
│       │   ├── deterministic-fst.h
│       │   ├── deterministic-fst-inl.h
│       │   ├── determinize-lattice.h
│       │   ├── determinize-lattice-inl.h
│       │   ├── determinize-star.h
│       │   ├── determinize-star-inl.h
│       │   ├── epsilon-property.h
│       │   ├── epsilon-property-inl.h
│       │   ├── factor.h
│       │   ├── factor-inl.h
│       │   ├── fstext-lib.h
│       │   ├── fstext-utils.h
│       │   ├── fstext-utils-inl.h
│       │   ├── fst-test-utils.h
│       │   ├── grammar-context-fst.h
│       │   ├── kaldi-fst-io.h
│       │   ├── kaldi-fst-io-inl.h
│       │   ├── lattice-utils.h
│       │   ├── lattice-utils-inl.h
│       │   ├── lattice-weight.h
│       │   ├── pre-determinize.h
│       │   ├── pre-determinize-inl.h
│       │   ├── prune-special.h
│       │   ├── prune-special-inl.h
│       │   ├── push-special.h
│       │   ├── rand-fst.h
│       │   ├── remove-eps-local.h
│       │   ├── remove-eps-local-inl.h
│       │   ├── table-matcher.h
│       │   └── trivial-factor-weight.h
│       ├── gmm
│       │   ├── am-diag-gmm.h
│       │   ├── decodable-am-diag-gmm.h
│       │   ├── diag-gmm.h
│       │   ├── diag-gmm-inl.h
│       │   ├── diag-gmm-normal.h
│       │   ├── ebw-diag-gmm.h
│       │   ├── full-gmm.h
│       │   ├── full-gmm-inl.h
│       │   ├── full-gmm-normal.h
│       │   ├── indirect-diff-diag-gmm.h
│       │   ├── mle-am-diag-gmm.h
│       │   ├── mle-diag-gmm.h
│       │   ├── mle-full-gmm.h
│       │   ├── model-common.h
│       │   └── model-test-common.h
│       ├── hmm
│       │   ├── hmm-test-utils.h
│       │   ├── hmm-topology.h
│       │   ├── hmm-utils.h
│       │   ├── posterior.h
│       │   ├── transition-model.h
│       │   └── tree-accu.h
│       ├── ivector
│       │   ├── agglomerative-clustering.h
│       │   ├── ivector-extractor.h
│       │   ├── logistic-regression.h
│       │   ├── plda.h
│       │   └── voice-activity-detection.h
│       ├── kws
│       │   ├── kaldi-kws.h
│       │   ├── kws-functions.h
│       │   └── kws-scoring.h
│       ├── lat
│       │   ├── arctic-weight.h
│       │   ├── compose-lattice-pruned.h
│       │   ├── confidence.h
│       │   ├── determinize-lattice-pruned.h
│       │   ├── kaldi-lattice.h
│       │   ├── lattice-functions.h
│       │   ├── minimize-lattice.h
│       │   ├── phone-align-lattice.h
│       │   ├── push-lattice.h
│       │   ├── sausages.h
│       │   ├── word-align-lattice.h
│       │   └── word-align-lattice-lexicon.h
│       ├── lm
│       │   ├── arpa-file-parser.h
│       │   ├── arpa-lm-compiler.h
│       │   ├── const-arpa-lm.h
│       │   ├── kaldi-rnnlm.h
│       │   └── mikolov-rnnlm-lib.h
│       ├── matrix
│       │   ├── cblas-wrappers.h
│       │   ├── compressed-matrix.h
│       │   ├── jama-eig.h
│       │   ├── jama-svd.h
│       │   ├── kaldi-blas.h
│       │   ├── kaldi-matrix.h
│       │   ├── kaldi-matrix-inl.h
│       │   ├── kaldi-vector.h
│       │   ├── kaldi-vector-inl.h
│       │   ├── matrix-common.h
│       │   ├── matrix-functions.h
│       │   ├── matrix-functions-inl.h
│       │   ├── matrix-lib.h
│       │   ├── optimization.h
│       │   ├── packed-matrix.h
│       │   ├── sparse-matrix.h
│       │   ├── sp-matrix.h
│       │   ├── sp-matrix-inl.h
│       │   ├── srfft.h
│       │   └── tp-matrix.h
│       ├── nnet
│       │   ├── nnet-activation.h
│       │   ├── nnet-affine-transform.h
│       │   ├── nnet-average-pooling-component.h
│       │   ├── nnet-blstm-projected.h
│       │   ├── nnet-component.h
│       │   ├── nnet-convolutional-component.h
│       │   ├── nnet-frame-pooling-component.h
│       │   ├── nnet-kl-hmm.h
│       │   ├── nnet-linear-transform.h
│       │   ├── nnet-loss.h
│       │   ├── nnet-lstm-projected.h
│       │   ├── nnet-matrix-buffer.h
│       │   ├── nnet-max-pooling-component.h
│       │   ├── nnet-multibasis-component.h
│       │   ├── nnet-nnet.h
│       │   ├── nnet-parallel-component.h
│       │   ├── nnet-parametric-relu.h
│       │   ├── nnet-pdf-prior.h
│       │   ├── nnet-randomizer.h
│       │   ├── nnet-rbm.h
│       │   ├── nnet-recurrent.h
│       │   ├── nnet-sentence-averaging-component.h
│       │   ├── nnet-trnopts.h
│       │   ├── nnet-utils.h
│       │   └── nnet-various.h
│       ├── nnet2
│       │   ├── am-nnet.h
│       │   ├── combine-nnet-a.h
│       │   ├── combine-nnet-fast.h
│       │   ├── combine-nnet.h
│       │   ├── decodable-am-nnet.h
│       │   ├── get-feature-transform.h
│       │   ├── mixup-nnet.h
│       │   ├── nnet-component.h
│       │   ├── nnet-compute-discriminative.h
│       │   ├── nnet-compute-discriminative-parallel.h
│       │   ├── nnet-compute.h
│       │   ├── nnet-compute-online.h
│       │   ├── nnet-example-functions.h
│       │   ├── nnet-example.h
│       │   ├── nnet-fix.h
│       │   ├── nnet-functions.h
│       │   ├── nnet-limit-rank.h
│       │   ├── nnet-nnet.h
│       │   ├── nnet-precondition.h
│       │   ├── nnet-precondition-online.h
│       │   ├── nnet-stats.h
│       │   ├── nnet-update.h
│       │   ├── nnet-update-parallel.h
│       │   ├── online-nnet2-decodable.h
│       │   ├── rescale-nnet.h
│       │   ├── shrink-nnet.h
│       │   ├── train-nnet-ensemble.h
│       │   ├── train-nnet.h
│       │   └── widen-nnet.h
│       ├── nnet3
│       │   ├── am-nnet-simple.h
│       │   ├── attention.h
│       │   ├── convolution.h
│       │   ├── decodable-online-looped.h
│       │   ├── decodable-simple-looped.h
│       │   ├── discriminative-supervision.h
│       │   ├── discriminative-training.h
│       │   ├── natural-gradient-online.h
│       │   ├── nnet-am-decodable-simple.h
│       │   ├── nnet-analyze.h
│       │   ├── nnet-attention-component.h
│       │   ├── nnet-batch-compute.h
│       │   ├── nnet-chain-diagnostics.h
│       │   ├── nnet-chain-example.h
│       │   ├── nnet-chain-training.h
│       │   ├── nnet-combined-component.h
│       │   ├── nnet-common.h
│       │   ├── nnet-compile.h
│       │   ├── nnet-compile-looped.h
│       │   ├── nnet-compile-utils.h
│       │   ├── nnet-component-itf.h
│       │   ├── nnet-computation-graph.h
│       │   ├── nnet-computation.h
│       │   ├── nnet-compute.h
│       │   ├── nnet-convolutional-component.h
│       │   ├── nnet-descriptor.h
│       │   ├── nnet-diagnostics.h
│       │   ├── nnet-discriminative-diagnostics.h
│       │   ├── nnet-discriminative-example.h
│       │   ├── nnet-discriminative-training.h
│       │   ├── nnet-example.h
│       │   ├── nnet-example-utils.h
│       │   ├── nnet-general-component.h
│       │   ├── nnet-graph.h
│       │   ├── nnet-nnet.h
│       │   ├── nnet-normalize-component.h
│       │   ├── nnet-optimize.h
│       │   ├── nnet-optimize-utils.h
│       │   ├── nnet-parse.h
│       │   ├── nnet-simple-component.h
│       │   ├── nnet-test-utils.h
│       │   ├── nnet-training.h
│       │   └── nnet-utils.h
│       ├── online
│       │   ├── online-audio-source.h
│       │   ├── onlinebin-util.h
│       │   ├── online-decodable.h
│       │   ├── online-faster-decoder.h
│       │   ├── online-feat-input.h
│       │   └── online-tcp-source.h
│       ├── online2
│       │   ├── onlinebin-util.h
│       │   ├── online-endpoint.h
│       │   ├── online-feature-pipeline.h
│       │   ├── online-gmm-decodable.h
│       │   ├── online-gmm-decoding.h
│       │   ├── online-ivector-feature.h
│       │   ├── online-nnet2-decoding.h
│       │   ├── online-nnet2-decoding-threaded.h
│       │   ├── online-nnet2-feature-pipeline.h
│       │   ├── online-nnet3-decoding.h
│       │   ├── online-speex-wrapper.h
│       │   └── online-timing.h
│       ├── rnnlm
│       │   ├── rnnlm-compute-state.h
│       │   ├── rnnlm-core-compute.h
│       │   ├── rnnlm-core-training.h
│       │   ├── rnnlm-embedding-training.h
│       │   ├── rnnlm-example.h
│       │   ├── rnnlm-example-utils.h
│       │   ├── rnnlm-lattice-rescoring.h
│       │   ├── rnnlm-test-utils.h
│       │   ├── rnnlm-training.h
│       │   ├── rnnlm-utils.h
│       │   ├── sampler.h
│       │   ├── sampling-lm-estimate.h
│       │   └── sampling-lm.h
│       ├── sgmm2
│       │   ├── am-sgmm2.h
│       │   ├── am-sgmm2-project.h
│       │   ├── decodable-am-sgmm2.h
│       │   ├── estimate-am-sgmm2-ebw.h
│       │   ├── estimate-am-sgmm2.h
│       │   └── fmllr-sgmm2.h
│       ├── transform
│       │   ├── basis-fmllr-diag-gmm.h
│       │   ├── cmvn.h
│       │   ├── compressed-transform-stats.h
│       │   ├── decodable-am-diag-gmm-regtree.h
│       │   ├── fmllr-diag-gmm.h
│       │   ├── fmllr-raw.h
│       │   ├── fmpe.h
│       │   ├── lda-estimate.h
│       │   ├── lvtln.h
│       │   ├── mllt.h
│       │   ├── regression-tree.h
│       │   ├── regtree-fmllr-diag-gmm.h
│       │   ├── regtree-mllr-diag-gmm.h
│       │   └── transform-common.h
│       ├── tree
│       │   ├── build-tree.h
│       │   ├── build-tree-questions.h
│       │   ├── build-tree-utils.h
│       │   ├── clusterable-classes.h
│       │   ├── cluster-utils.h
│       │   ├── context-dep.h
│       │   ├── event-map.h
│       │   └── tree-renderer.h
│       └── util
│           ├── basic-filebuf.h
│           ├── common-utils.h
│           ├── const-integer-set.h
│           ├── const-integer-set-inl.h
│           ├── edit-distance.h
│           ├── edit-distance-inl.h
│           ├── hash-list.h
│           ├── hash-list-inl.h
│           ├── kaldi-cygwin-io-inl.h
│           ├── kaldi-holder.h
│           ├── kaldi-holder-inl.h
│           ├── kaldi-io.h
│           ├── kaldi-io-inl.h
│           ├── kaldi-pipebuf.h
│           ├── kaldi-semaphore.h
│           ├── kaldi-table.h
│           ├── kaldi-table-inl.h
│           ├── kaldi-thread.h
│           ├── parse-options.h
│           ├── simple-io-funcs.h
│           ├── simple-options.h
│           ├── stl-utils.h
│           ├── table-types.h
│           └── text-utils.h
└── lib
    ├── cmake
    │   └── kaldi
    │       ├── kaldi-config.cmake
    │       └── kaldi-config-version.cmake
    ├── libkaldi-base.a
    ├── libkaldi-chain.a
    ├── libkaldi-cudadecoder.a
    ├── libkaldi-cudafeat.a
    ├── libkaldi-cudamatrix.a
    ├── libkaldi-decoder.a
    ├── libkaldi-feat.a
    ├── libkaldi-fstext.a
    ├── libkaldi-gmm.a
    ├── libkaldi-hmm.a
    ├── libkaldi-ivector.a
    ├── libkaldi-kws.a
    ├── libkaldi-lat.a
    ├── libkaldi-lm.a
    ├── libkaldi-matrix.a
    ├── libkaldi-nnet2.a
    ├── libkaldi-nnet3.a
    ├── libkaldi-nnet.a
    ├── libkaldi-online2.a
    ├── libkaldi-online.a
    ├── libkaldi-rnnlm.a
    ├── libkaldi-sgmm2.a
    ├── libkaldi-transform.a
    ├── libkaldi-tree.a
    └── libkaldi-util.a

```
</details>

cloudhan:
@danpovey will the above folder structure, particularly `bin`, break recipe scripts nastily?

danpovey:
I sent this by email several days ago but just got a bounce:

For the time being, if the CMakeLists.txt files in the individual directories are automatically generated I would prefer not to have them checked in.  (It might cause difficulties keeping them in sync.)   

Also please put clear instructions in the INSTALL file about the two possible methods for installation (make-based vs. cmake-based).

The script that generates the CMakeLists files should perhaps be in a subdirectory somewhere, so as not to clutter the main directory?   E.g. cmake/ ?

I wanted some clarification on why, exactly the bin/ directories must be merged when we use CMake?

My concern is that if we were to merge this as-is, after that we'd have to modify two build systems each time we added or removed a file.  Realistically we probably couldn't ditch the `make`-based system right away, even assuming we agreed that was the best thing to do long-term.

cloudhan:
> For the time being, if the CMakeLists.txt files in the individual directories are automatically generated I would prefer not to have them checked in. (It might cause difficulties keeping them in sync.)

My plan is finally drop the generator. It simply serve me to quick make general changes to all subdirectories. IMO, keeping a generator that generate meta build system scripts which itself generate another bunch build scripts is clearly, absolutely wrong thing. We also loss the fine-grained control of them.

> I wanted some clarification on why, exactly the bin/ directories must be merged when we use CMake?

Just a matter of taste. At build time, they are not merged anyway. they lay in `<build-dir>/src/<subdir-name>/<target-name>`, if you don't install them, the directory structure is identical to the make build system except for a separate build directory. If that matters for installing, I will make them all install analogous to the `include` dir

> My concern is that if we were to merge this as-is, after that we'd have to modify two build systems each time we added or removed a file. Realistically we probably couldn't ditch the `make`-based system right away, even assuming we agreed that was the best thing to do long-term.

My suggest is that do it completely, and remove the old ones. Maybe not in the master branch, but on the kaldi10 branch. Anyway this PR is still incomplete.

jtrmal:
I disagree with the first one. That would mean maintaining two separate
buildsystems -- at least for the time being I'd suggest having only the
generator (which will be easier to maintain). If the time comes and we will
decide to switch to Cmake, we can always generate the cmakes one last time.
y.

danpovey:
I tried this on a Mac, but it couldn't seem to find the BLAS libraries (should be available via the accelerate framework).
I managed to get past that problem but in a hacky way as I am not sure how it's supposed to be done.
I changed the MATHLIB variable setting to
`set(MATHLIB "Accelerate" CACHE STRING "OpenBLAS|MKL|Accelerate")`
and later did:
```
elseif(MATHLIB STREQUAL "Accelerate")
    message("Using accelerate library")
    find_package(BLAS REQUIRED)
    find_package(LAPACK REQUIRED)
    add_definitions(-DHAVE_CLAPACK=1)
    link_libraries(${BLAS_LIBRARIES} ${LAPACK_LIBRARIES})
endif()
```
but couldn't figure out how to get that option to be chosen if it was not made the default.

I also found that it did not. work quite right if CUDA was not being used:
`ld: library not found for -lkaldi-cudamatrix`

cloudhan:
`find_package(BLAS)` and `find_package(LAPACK)` all use a variable `BLA_VENDOR` to determine the blas/lapack vendor to use. From https://github.com/Kitware/CMake/blob/master/Modules/FindBLAS.cmake or
https://github.com/Kitware/CMake/blob/master/Modules/FindLAPACK.cmake you can see the vendor name is `APPLE`. If it is not define, it will try to find one available listed in FindBLAS/FindLAPACK cmake script, which should be your case.

cloudhan:
To set the option from commandline, simple invoke cmake with `-DNAME=VALUE`
e.g.
```
cmake -DMATHLIB=MKL ..
```
will config with MKL enabled

danpovey:
Thanks!   Please make sure it's up to date with master, e.g. merge or rebase.

cloudhan:
I rebased onto the current master.  Switched to generating cmakelists on the fly. Still exists many rough edges. And would you mind pointing me out why kaldi-cudamatrix failed, is it because cuda is not supported on Mac OS and should be disabled generally?

danpovey:
It looked like it never attempted to build it.  I think it's because
cudamatrix/ is built conditionally in the top-level CMakeLists.txt; it
should be built whether or not CUDA is enabled.  (But not the other 2
CUDA-related dirs.)

cloudhan:
Is it a gcc specific problem or only with some version?
```
In file included from ../matrix/cblas-wrappers.h:27:
../matrix/kaldi-matrix.h:577:21: error: declaration of 'Real' shadows template parameter
  template<typename Real>
                    ^
../matrix/kaldi-matrix.h:48:19: note: template parameter is declared here
template<typename Real>
                  ^
../matrix/kaldi-matrix.h:578:22: error: function template partial specialization is not allowed
  friend Real kaldi::TraceMatMat<Real>(const MatrixBase<Real> &A,
                     ^          ~~~~~~
2 errors generated.
```
Anyway, I will walkaround it for now.

cloudhan:
Not exactly. The previous patch to `kaldi-matrix.h` was to walkaround a MSVC+NVCC coupled problem.

When
```
  // template<typename Real>   // <---- comment out this line
  friend Real kaldi::TraceMatMat<Real>(const MatrixBase<Real> &A,
      const MatrixBase<Real> &B, MatrixTransposeType trans);  // tr (A B)
```

during cuda library compilation on Windows
```
feature-spectral-cuda.cu
D:/kaldi/src/cudafeat/..\matrix/kaldi-matrix.h(579): error C2955: 'kaldi::MatrixBase': use of class template requires template argument list
D:/kaldi/src/cudafeat/..\matrix/kaldi-matrix.h(49): note: see declaration of 'kaldi::MatrixBase'
D:/kaldi/src/cudafeat/..\matrix/kaldi-matrix.h(815): note: see reference to class template instantiation 'kaldi::MatrixBase<Real>' being compiled
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\chrono(670): note: see reference to class template instantiation 'std::chrono::duration<double,std::ratio<1,1>>' being compiled
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\chrono(177): note: see reference to class template instantiation 'std::chrono::duration<__int64,std::ratio<1,1000000000>>' being compiled
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\chrono(648): note: see reference to class template instantiation 'std::chrono::time_point<std::chrono::steady_clock,std::chrono::steady_clock::duration>' being compiled
D:/kaldi/src/cudafeat/..\matrix/kaldi-matrix.h(579): error C2063: 'kaldi::TraceMatMat': not a function
D:/kaldi/src/cudafeat/..\matrix/kaldi-matrix-inl.h(34): note: see reference to class template instantiation 'kaldi::MatrixBase<Real>' being compiled
        with
        [
            Real=float
        ]
CMake Error at cuda_compile_1_generated_feature-spectral-cuda.cu.obj.cmake:279 (message):
  Error generating file
  D:/kaldi/build/src/cudafeat/CMakeFiles/cuda_compile_1.dir//./cuda_compile_1_generated_feature-spectral-cuda.cu.obj
```

MSVC happily accept either form. Shit happens again 🤦‍

cloudhan:
### To correctly handle targets export

- [ ] In all subdirectories, export to a single target `kaldi-exports`.
- [ ] Install `kaldi-exports` in top level cmake
- [ ] export fst and stop use nvtoolext as a target (this create an export dependency)
- [ ] in config, handle external dependencies for installed targets

cloudhan:
@kkm000 I might need to patch your winport openfst for a totally painless kaldi cmake build.
The current problem is, I added openfst as a subdirectory with an EXCLUDE_FROM_ALL flag. This makes the subdirectory targets as needed. It cause install failure of the fst target, which is requested by @galv and me. I might create a new fork and patch it ahead.

cloudhan:
It seems again the openfst cmake is generated. Since only lib is needed, I might better add fst lib as a target from out of source.

cloudhan:
@danpovey Building and installing should be pretty painless now, in terms of model user. For ASR model building, external tools from OpenFST is not included. It shouldn't be handled in cmake anyway

danpovey:
Thanks!
Does this interfere at all with the current build process- i.e. could they coexist for the time being, if I merged this?
At some point I think I saw this overwriting src/Makefile, which is non-ideal but not deadly.

cloudhan:
Not at all. It is fully independent. There are some minor code change only to walkaround platform specific compiler error. Since they pass the CI build, it won't be a problem. All build related script resides in the `cmake` directory with an additional `CMakeLists.txt` file in the top-level. CMakeLists.txt in subdirectories are generated on the fly and is properly add to `.gitignore`. As long as future development do not break current file naming convention, directory structure in a nasty way, they should coexist well.

cloudhan:
BTW, there is no overwriting `src/Makefile` at all. I just read the Makefile and try to parse the library dependencies for `target_link_libraries`, since there is no way to automatically figure it out.

INSTALL
```diff
@@ -4,6 +4,9 @@ This is the official Kaldi INSTALL. Look also at INSTALL.md for the git mirror i
 (1)
 go to tools/  and follow INSTALL instructions there.
 
-(2) 
+(2)
 go to src/ and follow INSTALL instructions there.
 
+Alternatively, a new cmake based build system is supported with faster build speed.
+Note, it may not be well tested and some features are missing currently.
```
danpovey: Can you be clearer about what the options are?  E.g. is the cmake thing a replacement for (1) and (2) or just (2)?

You could structure it as:
 
Option A:  "make"-based build (this is well tested on multiple platforms)
...

Option B: "cmake"-based build.  This has been added more recently and
   may not be as well tested.
...

cloudhan: @danpovey Any suggestion? There a "Resolve conversation". And I am wondering, for github best practice, should you or me click it?

cloudhan:
let me link some issues: #3704 #3086

danpovey:
@jtrmal  any chance you could look  at this and merge if you are happy? I am super busy.

jtrmal:
ok, but I will have a look in the next few days -- @cloudhan, please ping
me if I won't be active for a long time

INSTALL
kkm000: @cloudhan, I'd go with the original commenter marking a conversation resolved by default, unless it's a truly trivial case (e. g., if you implemented their suggestion in code as-is). My thinking is, if I have some reservations and start a sub-discussion, then it's upon me to drop these reservation, given the arguments (or come to other solution, but clearly consensual). Again, I'm pretty fine with trivial matters (e.g, I raised doubts but obviously missed that a variable was initialized, and someone pointed at this fact). But for less than trivial question, the discussion stewardship, I think, should be on the topic starter.

cmake/INSTALL.md
```diff
@@ -0,0 +1,34 @@
...
+## Build with Ninja Generator
+``` bash
+mkdir -p build && cd build
+cmake -GNinja -DCMAKE_INSTALL_PREFIX=../dist ..
+cmake --build . --target install
+```
+
+After built, you can find all installed files in <your_repo_root>/dist
+
+# For Advance User
```
danpovey: advanced users. Please clarify how you would supply those options.

danpovey: And I hope the current scripts still work?  E.g. there are no path changes required?

INSTALL
cloudhan: @kkm000 got it, that make sense. Thanks for commenting.

cmake/INSTALL.md
cloudhan: Unfortunately, since we are doing an out of source build, it will not work out of box. But due to all executables are installed to a unified bin directory, it will be pretty easy to fix for the users. Pls see the updated file.

.gitignore
```diff
@@ -149,3 +149,6 @@ GSYMS
 /tools/cub-1.8.0/
 /tools/cub
 /tools/python/
+
+/src/**/CMakeLists.txt
```
galv: I'm afraid filtering this out is not a good idea. It will confused people.

.gitignore
```diff
@@ -149,3 +149,6 @@ GSYMS
...
+/build
```
galv: I suggest `/build*` Part of the value of cmake is the ability to have multiple coexisting builds.

pytorch acts similarly: https://github.com/pytorch/pytorch/blob/455b5c1a7d1503c03853e37b5a2caf60948359eb/.gitignore#L186

cmake/Utils.cmake
```diff
@@ -0,0 +1,40 @@
+macro(normalize_path in_path)
```
galv: It's hard to understand why both macros are needed here.

galv: You should use `include_guard()`. Note that it's cmake 3.10, though.

src/base/kaldi-error.cc
```diff
@@ -33,7 +33,11 @@
 
 #include "base/kaldi-common.h"
 #include "base/kaldi-error.h"
+
+// KALDI_GIT_HEAD is useless currently in full repo
```
galv: I'm a little skeptical of getting rid of KALDI_GIT_HEAD altogether. It shouldn't be hard to retain it. I feel like some downstream user of kaldi like @kkm000 (or whoever added this) might require it for some reason.

src/cudamatrix/cu-array.h
```diff
@@ -189,8 +189,8 @@ class CuSubArray: public CuArrayBase<T> {
   CuSubArray(const T* data, MatrixIndexT length) {
     // Yes, we're evading C's restrictions on const here, and yes, it can be used
     // to do wrong stuff; unfortunately the workaround would be very difficult.
-    CuArrayBase<T>::data_ = const_cast<T*>(data);
-    CuArrayBase<T>::dim_ = length;
+    this->data_ = const_cast<T*>(data);
```
galv: I don't think you should be making these kinds of changes in your PR. I understand that both of these constructs are the same, but most people won't, and will be confused by this kind of change in a seemingly irrelevant PR.

src/matrix/kaldi-matrix.h
```diff
@@ -574,6 +574,9 @@ class MatrixBase {
   void SymPosSemiDefEig(VectorBase<Real> *s, MatrixBase<Real> *P,
                         Real check_thresh = 0.001);
 
+#if defined(__NVCC__) && defined(_MSC_VER)
```
galv: Can you explain the purpose of this? I do not understand why it is here at all.

CMakeLists.txt
```diff
@@ -0,0 +1,195 @@
...
+
+    find_package(NvToolExt REQUIRED)
+    include_directories(${NvToolExt_INCLUDE_DIR})
+    link_libraries(${NvToolExt_LIBRARIES})
+
+    find_package(CUB REQUIRED)
+    include_directories(${CUB_INCLUDE_DIR})
+endif()
+
+add_definitions(-DKALDI_NO_PORTAUDIO=1)
```
galv: You should attach this to the specific target that uses it (the executable created by online-net-client.cc) with `https://cmake.org/cmake/help/v3.0/command/target_compile_definitions.html`. Otherwise, this definition is inherited by literally everything. I'm not sure whether or not an installed kaldi library would attach it to all clients depending on the kaldi library. The point is that it's confusing, and modern cmake's target-based commands leave no room for confusion.

CMakeLists.txt
```diff
@@ -0,0 +1,195 @@
...
+    set(KALDI_CUDA_NVCC_FLAGS "--default-stream=per-thread;-std=c++${CMAKE_CXX_STANDARD}")
+    if(MSVC)
+        list(APPEND KALDI_CUDA_NVCC_FLAGS "-Xcompiler /permissive-,/FS,/wd4819,/EHsc,/bigobj")
+        list(APPEND KALDI_CUDA_NVCC_FLAGS "-Xcompiler /wd4018,/wd4244,/wd4267,/wd4291,/wd4305")
+        if(BUILD_SHARED_LIBS)
+            list(APPEND CUDA_NVCC_FLAGS_RELEASE -Xcompiler /MD)
+            list(APPEND CUDA_NVCC_FLAGS_DEBUG -Xcompiler /MDd)
+        endif()
+    else()
+    #     list(APPEND KALDI_CUDA_NVCC_FLAGS "-Xcompiler -std=c++${CMAKE_CXX_STANDARD}")
```
galv: I'm surprised that this isn't necessary.

CMakeLists.txt
```diff
@@ -0,0 +1,195 @@
...
+        set(ENV{CUDAHOSTCXX} ${CMAKE_CXX_COMPILER})
+    endif()
+    if(NOT DEFINED CUDA_HOST_COMPILER)
+        set(CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
+    endif()
+endif()
+
+find_package(CUDA)
+if(CUDA_FOUND)
+    set(CUB_ROOT_DIR "tools/cub")
```
galv: `"${CMAKE_PROJECT_DIR}/tools/cub"`

cmake/third_party/openfst.cmake
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 2.8.2)
+project(openfst-download NONE)
+
+include(ExternalProject)
+ExternalProject_Add(openfst
+    GIT_REPOSITORY https://github.com/kkm000/openfst
+    GIT_TAG winport
```
galv: You would need to use the full hash actually, not the git tag, so that cmake doesn't have to ping the server on every rebuild to make sure that the remote's definition of the hash hasn't changed. (Meanwhile, hashes are immutable.)

cmake/FindICU.cmake
```diff
@@ -0,0 +1,428 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
galv: What uses this?

cmake/FindLAPACK.cmake
```diff
@@ -0,0 +1,430 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
galv: Where did you copy this from? Why not use the one distributed with cmake? https://github.com/Kitware/CMake/blob/master/Modules/FindLAPACK.cmake

INSTALL
```diff
@@ -1,9 +1,16 @@
 This is the official Kaldi INSTALL. Look also at INSTALL.md for the git mirror installation.
-[for native Windows install, see windows/INSTALL]
+[Option 1 in the following dose not apply to native Windows install, see windows/INSTALL or following Option 2]
```
galv: `dose` -> `does`

cmake/FindLAPACK.cmake
cloudhan: They are copied from latest cmake repo with minor modification. And it is needed for, iirc, OpenBLAS.

cmake/FindICU.cmake
cloudhan: It is copied from latest cmake repo with minor modification for this build system to work with older version cmake, say, CMake 3.5 on Ubuntu 16.04. ICU is needed of a third party library for i18n.

cmake/third_party/openfst.cmake
cloudhan: I am still wondering why there is no mirror in kalid-asr project namespace.

CMakeLists.txt
cloudhan: It is desired at first, but not achievable at the moment.
See
https://github.com/kaldi-asr/kaldi/pull/3580#issuecomment-531456387 and https://github.com/kaldi-asr/kaldi/pull/3580#issuecomment-531459028 .

In summary, the `CMakeLists.txt` generator is kept at the moment. The `CMakeLists.txt`s in `src/**` are all generated on the fly.

cloudhan: It is desired at first, but not achievable at the moment.
See
https://github.com/kaldi-asr/kaldi/pull/3580#issuecomment-531456387 and https://github.com/kaldi-asr/kaldi/pull/3580#issuecomment-531459028 .

In summary, the `CMakeLists.txt` generator script is kept at the moment. The `CMakeLists.txt`s in `src/**` are all generated on the fly.

.gitignore
cloudhan: Apparently, you are confused. `/src/**/CMakeLists.txt` are all generated on the fly.

src/cudamatrix/cu-array.h
cloudhan: This is needed for successfully compilation CUDA related code on windows. They are absolutely relevant with this PR. There are some weird issue with two-phase name lookup in Windows version of nvcc.

src/matrix/kaldi-matrix.h
cloudhan: This is needed for successfully compilation CUDA related code on windows. They are absolutely relevant with this PR. There are some weird issue with template friend function in a class template in Windows version of nvcc. This is simple an ugly walkaround.

CMakeLists.txt
cloudhan: yeah

cmake/Utils.cmake
cloudhan: see latest file

cloudhan: see comment in latest file.

cmake/third_party/openfst.cmake
cloudhan: Updated to use a tag.

src/cudamatrix/cu-array.h
galv: Oh, wow.  Silly Windows!

cloudhan: To be clear, it is NVIDIA’s problem. Actually, I find msvc is much more reliable nowadays.

cmake/third_party/openfst.cmake
danpovey: RE the mirror: @kkm000  is a Kaldi contributor so we don't have to worry about him changing his OpenFST repo in a way that breaks stuff.  And Kaldi install is already dependent on github being up, so we don't need to worry about that being down (plus it's super reliable).

danpovey:
@jtrmal  or @kkm000 or @galv can you please see where we are with this?  I'd like to merge unless we can point to something that needs to be changed.

cloudhan:
@danpovey I forgot to update openfst to refer a hash. I might use a non-patched version for windows. Will try it soon.

jtrmal:
LGTM
I'd prefer to get the source code changes in first, as a separate commit and the cmake infrastructure after that, though.

danpovey:
I agree with @jtrmal that it should be split up into 2 PRs.  But other than that I think we're good to go!

cloudhan:
OK, so I rebased onto a recent master (I should not have push this rebasing result, though). Then cherry-picked all Windows NVCC fix/walkaround commits in #3741. And then rebase again without those commits.

danpovey:
Fantastic!  Merging.. it's all additions, so certainly won't break anything.  Thanks a lot for this-- it was a lot of work!

PR #3596, opened 2019-09-17
Update apply-cmvn-online.cc

ferb2015:
There may be only one "num_done++" here


src/online2bin/apply-cmvn-online.cc
```diff
@@ -128,7 +128,7 @@ int main(int argc, char *argv[]) {
         tot_t += feats.NumRows();
         feature_writer.Write(utt, normalized_feats);
         
-        num_done++;
+        //num_done++;
```
danpovey: please just remove the line rather than commenting it.

danpovey:
Hm, looks like this has been fixed since then.

PR #3602, opened 2019-09-19
rm transfer learning script bug

Szu-JuiChen:
- As discussed in https://groups.google.com/d/msg/kaldi-help/IwOApGnmi3Q/WvQNFYgZAQAJ, fixed parameter for final model combination error


danpovey:
Oh... this isn't really the right place to do the fix because your problem was with e2e training, which has a greater variety of minibatch sizes; and those scripts you pointed me to are not specific to e2e.
I don't think this is a bug per se; it's just that you didn't set minibatch sizes suitably for these e2e setups.   Eventually we might have a separate option for the minibatch size for the final combination.  Anyway, closing this...

Szu-JuiChen:
Actually I have the same problem in this discussion: https://groups.google.com/forum/#!searchin/kaldi-help/transfer$20learning%7Csort:date/kaldi-help/Ac87ueFcbzk/oy72OWPvAwAJ
So I encounter this minibatch problem while using egs/rm/s5/local/chain/tuning/run_tdnn_wsj_rm_1c.sh and the modification did fix the problem...

PR #3611, opened 2019-09-23
Fixed online2-tcp-nnet3-decode-faster.cc poll_ret checks

babakrkb:
Negative return value for `poll_ret` was not checked and `Socket error` is raised every time we pass an audio file to the TcpServer.

### DIFF
--- a/src/online2bin/online2-tcp-nnet3-decode-faster.cc
+++ b/src/online2bin/online2-tcp-nnet3-decode-faster.cc
@@ -431,7 +431,7 @@ bool TcpServer::ReadChunk(size_t len) {
       KALDI_WARN << "Socket timeout! Disconnecting...";
       break;
     }
-    if (client_set_[0].revents != POLLIN) {
+    if (poll_ret < 0) {
       KALDI_WARN << "Socket error! Disconnecting...";
       break;
     }


src/online2bin/online2-tcp-nnet3-decode-faster.cc
```diff
@@ -431,7 +431,7 @@ bool TcpServer::ReadChunk(size_t len) {
       KALDI_WARN << "Socket timeout! Disconnecting...";
       break;
     }
-    if (client_set_[0].revents != POLLIN) {
```
danpovey: @hhadian what do you think about this?

hhadian: It seems good to me. At least AFAIK `revents` is a bit mask and should not be compared like this (e.g. other bits in addition to POLLIN might be set). Maybe it's better to keep this if-block too but like this: `if (!(client_set_[0].revents & POLLIN))`.
Babak, can you please see if this will work too in your tests?
Anyway more checks might be needed, I don't remember all details right now (see http://man7.org/linux/man-pages/man2/poll.2.html)

babakrkb: @hhadian Yes it works too.

PR #3498, opened 2019-07-31
WIP: adding possibility to have 'online-cmn' on input of 'nnet3' models

vesis84:
Links: This code is related to issue #3461 (specification of code changes), and PR #3439 that will add the support of online-cmvn frontend in the online2 C++ codebase.

Characteristics of this PR:
- online-cmvn can be used with or without ivector features,
- it is demonstrated for the 'rm' databse,
- the online-cmn is supported for both 'chain' or 'cross-entropy' nnet3 models,
- the WER's for chain RM models were identical, see below.

- without cmn,
%WER 1.60 [ 201 / 12533, 17 ins, 34 del, 150 sub ] exp/chain/tdnn_5o/decode/wer_3_0.0
%WER 8.11 [ 1016 / 12533, 76 ins, 140 del, 800 sub ] exp/chain/tdnn_5o/decode_ug/wer_7_0.0
- with online-cmn,
%WER 1.57 [ 197 / 12533, 21 ins, 29 del, 147 sub ] exp/chain/tdnn_5o_online-cmn/decode/wer_2_0.5
%WER 8.11 [ 1017 / 12533, 59 ins, 160 del, 798 sub ] exp/chain/tdnn_5o_online-cmn/decode_ug/wer_8_0.0

Dan, any comments, suggestions? 
What should be the next 'egs' I should demonstrate this on? 'wsj' or something else?
Are you happy with the way how the 'online-cmvn' support is added? 

Not expecting this PR to merge yet, first we should add a demo for another recipe 
and if someone could help with testing, that would be fantastic...

Thanks,
K.

PR #3615, opened 2019-09-25
ivector, make the online-cmvn carry stats over utterances from same speaker.

vesis84:
This resolves issue identified after accepting PR #3498 (slight performance degradation)


danpovey:
Just a note: see this comment
https://github.com/kaldi-asr/kaldi/pull/3498#issuecomment-534492227
for how this affects results.  This only affects results when online-cmvn is used.  (It improves them slightly, it seems)

PR #3616, opened 2019-09-25
Lookahead graph decoding

nshmyrev:
This is not yet fully tested, but decoding works fine, almost same accurate and slightly lower speed with -O3 (much slower with -O1). OpenGrm ngram fsts still to be integrated.


danpovey:
Cool!  This was a feature that was missing for a long time.

src/nnet3bin/nnet3-latgen-faster-lookahead.cc
```diff
@@ -0,0 +1,243 @@
...
+#include "fstext/fstext-lib.h"
+#include "decoder/decoder-wrappers.h"
+#include "nnet3/nnet-am-decodable-simple.h"
+#include "nnet3/nnet-utils.h"
+#include "base/timer.h"
+
+namespace fst {
+
+template<class Arc>
+ComposeFst<Arc> TableComposeFst(
```
danpovey: Probably shouldn't have Table in the name, it referred to a specific composition matcher I was using.
Maybe CreateComposedFst()?

src/nnet3bin/nnet3-latgen-faster-lookahead.cc
```diff
@@ -0,0 +1,243 @@
...
+  try {
+    using namespace kaldi;
+    using namespace kaldi::nnet3;
+    typedef kaldi::int32 int32;
+    using fst::SymbolTable;
+    using fst::Fst;
+    using fst::StdArc;
+
+    const char *usage =
+        "Generate lattices using nnet3 neural net model.\n"
```
danpovey: make doc more specific to this!

RuABraun:
How much is slightly slower? 10% or 1%?

nshmyrev:
> How much is slightly slower? 10% or 1%?

More like 10%. Still building librispeech to test properly

egs/wsj/s5/steps/nnet3/decode_lookahead.sh
```diff
@@ -0,0 +1,147 @@
...
+# End configuration section.
+
+echo "$0 $@"  # Print the command line for logging
+
+[ -f ./path.sh ] && . ./path.sh; # source the path.
+. utils/parse_options.sh || exit 1;
+
+if [ $# -ne 3 ]; then
+  echo "Usage: $0 [options] <graph-dir> <data-dir> <decode-dir>"
+  echo "e.g.:   steps/nnet3/decode.sh --nj 8 \\"
```
sih4sing5hong5: decoding_lookahead.sh?

egs/wsj/s5/utils/mkgraph_lookahead.sh
```diff
@@ -0,0 +1,174 @@
...
+  [ "$1" == "--mono" -o "$1" == "--left-biphone" -o "$1" == "--quinphone" ] && shift && \
+    echo "WARNING: the --mono, --left-biphone and --quinphone options are now deprecated and ignored."
+  [ "$1" == "--remove-oov" ] && remove_oov=true && shift;
+  [ "$1" == "--transition-scale" ] && tscale=$2 && shift 2;
+  [ "$1" == "--self-loop-scale" ] && loopscale=$2 && shift 2;
+done
+
+if [ $# != 3 ]; then
+   echo "Usage: utils/mkgraph.sh [options] <lang-dir> <model-dir> <graphdir>"
+   echo "e.g.: utils/mkgraph.sh data/lang_testexp/tri1 exp/tri1/lgraph"
```
sih4sing5hong5: mkgraph_lookahead.sh?

egs/wsj/s5/steps/nnet3/decode_lookahead.sh
al-zatv: $0 !

egs/wsj/s5/utils/mkgraph_lookahead_vocab.py
```diff
@@ -0,0 +1,12 @@
...
+import sys
+
+wmap = {}
+for line in open(sys.argv[1]):
+    items = line.split()
+    wmap[items[0]] = items[1]
+
+for line in open(sys.argv[2]):
+    items = line.split()
+    print items[0], wmap.get(items[1], items[1])
```
galv: I really suggest making this a a python3-compatible print.

src/fstext/fstext-utils.h
```diff
@@ -392,7 +392,23 @@ void RhoCompose(const Fst<Arc> &fst1,
...
+ComposeFst<Arc> LookaheadComposeFst(const Fst<Arc> &ifst1,
+                                    const Fst<Arc> &ifst2) {
+  typedef LookAheadMatcher< StdFst > M;
+  typedef AltSequenceComposeFilter<M> SF;
+  typedef LookAheadComposeFilter<SF, M>  LF;
+  typedef PushWeightsComposeFilter<LF, M> WF;
+  typedef PushLabelsComposeFilter<WF, M> ComposeFilter;
+
+  fst::CacheOptions cache_opts(true, 1 << 26LL);
+  ComposeFstOptions<StdArc, M, ComposeFilter> opts(cache_opts);
```
galv: Hi @nshmyrev this is very interesting. I read a paper called [A Generalized Composition Algorithm for Weighted Finite-State Transducers](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/35539.pdf) about a month ago. It mentions that its dynamic runtime composition is provided in the OpenFST library. Does `ComposeFst` implement its functionality?

In the introduction, you can read about 3 problems the paper discusses with "normal" composition of HCL and G on the fly in section 1.

nshmyrev: Yes, this is the implementation of the described paper from openfst integrated with kaldi decoding.

jtrmal:
yes, I agree, that reminds me -- nowadays we want all new python scripts to
be in py3
y.

nshmyrev:
The current problem to solve is that recipes like http://bozskyfilip.blogspot.com/2017/11/how-make-hcl-and-g-graphs-and-on-fly.html suggest to remove disambig system from HCL, which is a bad idea since it breaks determinization during decoding during online composition.

Looks like I have to write another filter to remove disambig symbols during composition instead.

src/fstext/fstext-utils-inl.h
```diff
@@ -114,6 +114,52 @@ void GetInputSymbols(const Fst<Arc> &fst,
...
+       }
+private:
+  kaldi::ConstIntegerSet<I> to_remove_set_;
+};
+
+// Lookahead composition is used for optimized online
+// composition of FSTs during decoding. See
+// nnet3/nnet3-latgen-faster-lookahead.cc
+template<class Arc, class I>
+ArcMapFst<Arc, Arc, RemoveSomeInputSymbolsMapper<Arc, I> > LookaheadComposeFst(const Fst<Arc> &ifst1,
```
nshmyrev: Not sure how to simplify it, need help from C++ expert on how to specialize a template.

nshmyrev:
More or less complete now, please review.

src/fstext/fstext-utils-inl.h
galv: Why do you want to specialize this template function?

nshmyrev: Too long probably, could be something like 

```
template<class Arc, class I>
ComposedNodisambigFst<Arc, I> LookaheadComposeFst(const Fst<Arc> &ifst1
```

galv:
```
template<class Arc, class I>
using ComposedNodisambigFst = ArcMapFst<Arc, Arc, RemoveSomeInputSymbolsMapper<Arc, I> > 
```

nshmyrev: Very good, thank you!

src/fstext/fstext-utils-inl.h
```diff
@@ -114,6 +114,55 @@ void GetInputSymbols(const Fst<Arc> &fst,
...
+using LookaheadFst = ArcMapFst<Arc, Arc, RemoveSomeInputSymbolsMapper<Arc, I> >;
+
+// Lookahead composition is used for optimized online
+// composition of FSTs during decoding. See
+// nnet3/nnet3-latgen-faster-lookahead.cc
+template<class Arc, class I>
+LookaheadFst<Arc, I> LookaheadComposeFst(const Fst<Arc> &ifst1,
+                                         const Fst<Arc> &ifst2,
+                                         const std::vector<I> &to_remove) {
+  typedef LookAheadMatcher< StdFst > M;
```
galv: `< StdFst >` -> `<StdFst>`

src/fstext/fstext-utils-inl.h
```diff
@@ -114,6 +114,55 @@ void GetInputSymbols(const Fst<Arc> &fst,
...
+  typedef AltSequenceComposeFilter<M> SF;
+  typedef LookAheadComposeFilter<SF, M>  LF;
+  typedef PushWeightsComposeFilter<LF, M> WF;
+  typedef PushLabelsComposeFilter<WF, M> ComposeFilter;
+
+  fst::CacheOptions cache_opts(true, 1 << 26LL);
+  ComposeFstOptions<StdArc, M, ComposeFilter> opts(cache_opts);
+
+  RemoveSomeInputSymbolsMapper<Arc, I> mapper(to_remove);
```
galv: I'm not sure why this mapper is necessary here... You're essentially constructing HCLG here. The input symbols are going to be HMM transition ids, right? We only need to use disambiguation symbols in the case of the lexicons' input symbols (phonemes), right? Can you explain the purpose of this? It seems like you may accidentally remove valid transition ID's accidentally as it is now.

src/fstext/fstext-utils-inl.h
galv: How did you know that these were the right filters to use? Did you copy-paste from some sample inside OpenFST? I just want to double-check. I've read most of the paper, but I am not familiar with this part of the OpenFST codebase.

src/fstext/fstext-utils-inl.h
nshmyrev: Without disambig symbols HCL is not deterministic anymore, so it is problematic to compose it with G, online or offline. Just like in static recipe, we remove disambig symbols only on the last stage after composition.

src/fstext/fstext-utils-inl.h
nshmyrev: That all from the sources. I actually just changed it to more high-level openfst API, same effect as before.

nshmyrev:
The current problem I have is that without fstpushspecial in HCL construction the online decoder does not detect silence, another state has higher score for silence frames. With fstpushspecial it works fine, but less accurate overall (about 14->16.5% WER). With original fstpush instead of fstpushspecial it is accurate and detect silence properly but for some lexicons it doesn't return, it fails to push and hangs. I need to figure out why push makes silence detector work, probably composefst doesn't push properly.

danpovey:
In theory it should be totally invariant to pushing as long as the beams
are large enough

nshmyrev:
I figured out the problems were due to log semiring in determinization of non-stochastic graphs, now it is fast and accurate. I updated the results too.

More or less ready for review in my opinion, please take a look.

tools/extras/install_opengrm.sh
```diff
@@ -0,0 +1,55 @@
...
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+# WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+# MERCHANTABLITY OR NON-INFRINGEMENT.
+# See the Apache 2 License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script attempts to install Speex, which is needed by someone who wants
```
danpovey: outdated comment

.gitignore
```diff
@@ -151,6 +151,8 @@ GSYMS
 /tools/cub-1.8.0/
 /tools/cub
 /tools/python/
+tools/ngram-1.3.7.tar.gz
```
danpovey: would prefer /tools instead of tools,  for consistency with the other patterns

egs/wsj/s5/steps/nnet3/decode_lookahead.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/bash
+
+# Copyright 2012-2015  Johns Hopkins University (Author: Daniel Povey).
+# Apache 2.0.
+
+# This script does decoding with a neural-net with lookahead composition of HCL and G graphs.
```
danpovey: please fix author in copyright info.
If possible an example script for a smaller setup might be easier to demo to people (eg mini_librispeech) but not necessary.

egs/wsj/s5/utils/mkgraph_lookahead.sh
```diff
@@ -0,0 +1,208 @@
...
+        fstarcsort --sort_type=ilabel | \
+        fstconvert --fst_type=const > ${dir}/Gr.fst.$$
+    fi
+    mv $dir/Gr.fst.$$ $dir/Gr.fst
+    cp $lang/words.txt $dir/ || exit 1;
+  fi
+else
+  if [[ ! -s $dir/Gr.fst || $dir/Gr.fst -ot $arpa ]]; then
+    # Opengrm builds acceptors, so we need to reorder words in symboltable
+    utils/mkgraph_lookahead_vocab.py ${dir}/relabel ${lang}/words.txt > ${dir}/words.txt
```
danpovey: Please use utils/apply_map.pl instead here.

egs/wsj/s5/utils/mkgraph_lookahead.sh
```diff
@@ -0,0 +1,208 @@
...
+# place of -o
+if [[ $# != 3 && $# != 4 ]]; then
+   echo "Usage: $0 [options] <lang-dir> <model-dir> [<arpa_file>] <graphdir>"
+   echo "e.g.: $0 data/lang data/local/lm.gz exp/tri1 db/trigram.lm.gz exp/tri1/lgraph"
+   echo " Options:"
+   echo " --remove-oov       #  If true, any paths containing the OOV symbol (obtained from oov.int"
+   echo "                    #  in the lang directory) are removed from the G.fst during compilation."
+   echo " --transition-scale #  Scaling factor on transition probabilities."
+   echo " --self-loop-scale  #  Please see: http://kaldi-asr.org/doc/hmm.html#hmm_scale."
+   echo " --compose-graph    #  Compile composed graph for testing with stadnard decoders (default: false)"
```
danpovey: standard

src/fstext/kaldi-fst-io.cc
```diff
@@ -42,6 +42,12 @@ VectorFst<StdArc> *ReadFstKaldi(std::string rxfilename) {
   return fst;
 }
 
+// Register const fst to load it automatically. Other types like
+// olabel_lookahead or ngram or compact_fst should be registered
+// through OpenFst registration API.
+static fst::FstRegisterer<VectorFst<StdArc>> VectorFst_StdArc_registerer;
```
danpovey: Can you explain what this is for and how it worked previously?

nshmyrev: FST has a registration mechanism (http://www.openfst.org/doxygen/fst/html/register_8h.html) for loading the FST depending on the header type. If type is registered, you can simply load FST - vector, const, lookahead, ngram as

`Fst<StdArc> *fst = Fst<StdArc>::Read(ki.Stream(), ropts);
`
And when you link to extra so files, types are updated automatically.

Before that it was explicit

  ```
FstReadOptions ropts("<unspecified>", &hdr);
  Fst<StdArc> *fst = NULL;
  if (hdr.FstType() == "const") {
    fst = ConstFst<StdArc>::Read(ki.Stream(), ropts);
  } else if (hdr.FstType() == "vector") {
    fst = VectorFst<StdArc>::Read(ki.Stream(), ropts);
  }
```

and we will have to add more cases like:

  else if (hdr.FstType() == "ngram") {
    fst = NgramFst<StdArc>::Read(ki.Stream(), ropts);
  }

etc.

File: egs/wsj/s5/steps/nnet3/decode_lookahead.sh (hunk 12 above)
Comment (nshmyrev) [reply egs/wsj/s5/steps/nnet3/decode_lookahead.sh]: Sure, I will add mini version too.

File: egs/wsj/s5/utils/mkgraph_lookahead.sh (hunk 14 above)
Comment (nshmyrev) [reply egs/wsj/s5/utils/mkgraph_lookahead.sh]: Fixed

File: tools/extras/install_opengrm.sh (hunk 10 above)
Comment (nshmyrev) [reply tools/extras/install_opengrm.sh]: Fixed

File: .gitignore (hunk 11 above)
Comment (nshmyrev) [reply .gitignore]: Fixed

File: egs/librispeech/s5/local/lookahead/run_lookahead.sh [hunk 16]
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
+. path.sh
```
danpovey: Please use ./path.sh in case . is not on PATH

danpovey:
We have to make sure that if the user does `git pull` and then `make`, either it builds or it gives an informative error.. think I mentioned this before but not sure where.  Please confirm that that is the case.  It may be necessary to increment CONFIGURE_VERSION.  Also you might want to have a look into the CMake-based build and see if that needs any changes.. dont know if you are familiar with CMake?  (I am not).

nshmyrev:
> We have to make sure that if the user does git pull and then make, either it builds or it gives an informative error.. think I mentioned this before but not sure where. Please confirm that that is the case. It may be necessary to increment CONFIGURE_VERSION. Also you might want to have a look into the CMake-based build and see if that needs any changes.. dont know if you are familiar with CMake? (I am not).

Sure, I'll check both default build and cmake

egs/wsj/s5/utils/mkgraph_lookahead.sh
nshmyrev: Done

egs/librispeech/s5/local/lookahead/run_lookahead.sh
nshmyrev: Done

egs/wsj/s5/steps/nnet3/decode_lookahead.sh
nshmyrev: Done, mini_librispeech also added

nshmyrev:
> We have to make sure that if the user does git pull and then make, either it builds or it gives an informative error.. think I mentioned this before but not sure where. Please confirm that that is the case. It may be necessary to increment CONFIGURE_VERSION. Also you might want to have a look into the CMake-based build and see if that needs any changes.. dont know if you are familiar with CMake? (I am not).

I looked on this again. The thing is that openfst has all required code in headers already, so it doesn't really matter how do you compile it, it works fine. The added .so plugins are loaded dynamically and required only if you actually want to use FSTs of lookahead type.

So the check is only needed if you actually want to run lookahead demo script, so I added them to the local/lookahead/run_lookahead.sh demo scripts.

danpovey:
Thanks!  So if you think it's ready to merge now, I will trust that it's all OK and will merge.

src/fstext/fstext-utils-inl.h
Malkovsky: How does it affect online composition in terms of search accuracy/speed and fst cache size?

nshmyrev: We specifically disable cache for disambig removal, so no additional memory is used. As per speed drop, it is maybe 1% or less by measurement, certainly not critical. It is a simple filter anyway.

Overall, this whole thing is not very optimal as I feel it by many different reasons, thought it is a subject of much longer post I'll write soon.

Malkovsky: Thanks for the response, will look forward for the post!

nshmyrev:
> Thanks! So if you think it's ready to merge now, I will trust that it's all OK and will merge.

Should be good I think, thank you. I'll be ready to resolve issues if any.

yondu22:
Is there a demo on how to use the lookahead model, for example passing a list of words/phrases

sih4sing5hong5:
> Is there a demo on how to use the lookahead model, for example passing a list of words/phrases

There is some script like `egs/librispeech/s5/local/lookahead/run_lookahead.sh` in this PR.

nshmyrev:
In the vosk-api code you can check how recognizer is constructed from a list of words:

https://github.com/alphacep/vosk-api/blob/01eeec498da8332a8d5f9e9bfa7cb7ba30cdbfcb/src/kaldi_recognizer.cc#L49

https://github.com/alphacep/vosk-api/blob/master/python/example/test_local_words.py

YuriiMytiai:
@nshmyrev Hi!
There is small mismatch in https://github.com/kaldi-asr/kaldi/pull/3616/files#diff-e0ca53f8e34f1163b10eb26ed14ab144R55: `num-threads` parameter exists in the **help message** but script doesn't accept it. You probably need to remove this message from the **help**

nshmyrev:
@YuriiMytiai thank you for a great catch, I fixed this in my branch, will push on the next update!

src/fstext/fstext-utils-inl.h
francisr: > Overall, this whole thing is not very optimal as I feel it by many different reasons, thought it is a subject of much longer post I'll write soon.

I'm interested in a write up on this, are you still planning to do it?

nshmyrev: Thanks guys, I finally wrote it here https://alphacephei.com/nsh/2020/03/27/lookahead.html

wwbnjsace:
i have two question ,the first :
when i use "vosk-api-master/training/local/chain/run_tdnn.sh" train a model ,the model size is about 20M but the release model size is 70M ([vosk-model-en-us-0.22-lgraph](https://alphacephei.com/vosk/models/vosk-model-en-us-0.22-lgraph.zip) ,can u tell me why?

the second : when i use my 20M model ,i decode a file in final.mdl and HCLG.fst ,the output is normal ; but when i utils/mkgraph_lookahead.sh generate the HCLr.fst ,i decode a file by final and HCLr.st in a dynamicly way by language model,the output is none , can you tell me why?

src/fstext/kaldi-fst-io.cc
LRY1994: how to register  olabel_lookahead type

Fst::Read: Unknown FST type olabel_lookahead (arc type = standard)

PR #3627, opened 2019-09-30
[tools] Corrected the path of env.sh in the message

ajd12342:
The message "***() Please source the tools/extras/env.sh in your path.sh to enable it" was corrected to  "***() Please source the tools/env.sh in your path.sh to enable it" because the script is executed when pwd=tools/ and the env.sh is created in the tools/ directory.


developerkl:
()

ajd12342:
When would this be accepted?

PR #3629, opened 2019-10-01
5.4

devYRPauli:
Kindly review them.


danpovey:
Not sure what this is.  The old versions are just archives of what Kaldi was then.. it doesn't make sense to merge a newer version into an older.

PR #3630, opened 2019-10-01
[WIP] Add GPU CUDA for parallel decoding

raditya1710:


danpovey:
Can you please say something about what this is?
There are some aspects of this that definitely would cause a problem to be checked in, for example:
  - Large data files included in the PR.  This would bulk up the repo and is a no-go.
  - I can see merge conflicts
  - This is likely not as good as the GPU decoder stuff that NVidia has been doing (because they are NVidia).. but we can discuss this later.

Issue #3339, opened 2019-05-22
incorrect data directory when using utils/prepare_extended_lang.sh with the parameter  --word-list


jtrmal:
Hi, the way utils/prepare_extended_lang.sh behaves makes (in some cases) the lang dir incorrect if the parameter --word-list parameter is supplied.
The problem is that the prepare_lang is called first, which means that oov.int and the *.int and *.csl files in phone subdir are generated using the original word-map and only after that the map is regenerated.

jtrmal:
probably @xiaohui-zhang could check if I'm right.
IMO, there is no easy fix, because it goes down to L.fst -- I think the course of fix would be to add a parameter to prepare_lang.sh

jtrmal:
@DongjiGao following your latest PR #3485, I wonder if you'd be able to look at this?
Thanks :)

DongjiGao:
OK, I will take a look.

Dongji

On Jul 22, 2019, at 4:37 PM, jtrmal <notifications@github.com<mailto:notifications@github.com>> wrote:


@DongjiGao<https://github.com/DongjiGao> following your latest PR #3485<https://github.com/kaldi-asr/kaldi/pull/3485>, I wonder if you'd be able to look at this?
Thanks :)

xiaohui-zhang:
Thanks @jtrmal for raising this problem. Actually, I created that script before @danpovey made this much better one (interface-wise): https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/utils/lang/extend_lang.sh  The only functionality my version has, while Dan's newer version doesn't have is that, the newer version doesn't support lexiconp_silprobs.txt. I would suggest @DongjiGao try to port this functionality into Dan's version, and delete my version (IIRC it's not being used by any official script anyway).

danpovey:
@DongjiGao is there any chance you could address this issue?

DongjiGao:
I can do that. I am preparing a presentation for the reading group on Monday. I will fix it after the reading group.

jtrmal:
I assume this is resolved.

PR #3632, opened 2019-10-02
fix the issue #3339 about extend_lang.sh by adding support of lexiconp_silprob

DongjiGao:
delete original prepare_extended_lang.sh


egs/wsj/s5/utils/lang/extend_lang.sh
```diff
@@ -79,11 +83,26 @@ tmpdir=$dir/temp
...
-# TODO: more checking.
-if [ $(basename $lexicon) != lexiconp.txt ]; then
-  echo "$0: currently this script only supports the lexiconp.txt format; your lexicon"
-  echo " ... has to have that filename."
+silprob=false
+
+if [ $(basename $lexicon) == "lexiconp_silprob.txt" ]; then
+  silprob=true
+  if [ -z $silprob_file ] ; then
+    echo "silprob_file not provided, check $srcdir"
```
danpovey: you should say "checking $srcdir" , not "check $srcdir".

danpovey:
I think you can just add a warning at the top of prepare_extended_lang.sh that it's deprecated rather than delete it.. it might be used from other scripts.

PR #3636, opened 2019-10-03
Change the GPU policy to wait instead of yes

flashhack:
This is needed to avoid failures when using more number jobs than number of GPUs. Previously when --use-gpu=yes, when all GPUs are occupied, the job will error out in 20s. Changing to the policy to wait will have correct queuing behavior. @xiaohui-zhang


danpovey:
I have some reservations about this (would prefer to set it from outside) but it seems harmless so I'll merge it.

PR #3637, opened 2019-10-04
added call to AdvanceDecoding() inside Decode()

Bar-BY:


src/decoder/lattice-faster-decoder.cc
```diff
@@ -78,7 +78,7 @@ void LatticeFasterDecoderTpl<FST, Token>::InitDecoding() {
 template <typename FST, typename Token>
 bool LatticeFasterDecoderTpl<FST, Token>::Decode(DecodableInterface *decodable) {
   InitDecoding();
-
+  AdvanceDecoding(decodable);
```
danpovey: Please remove the while loop below this; it's now redundant because it's done in AdvanceDecoding().
And see if there are any other decoders templated in the same way that have the same issue (grep for Decode).

danpovey:
Thanks!  Can you do some basic testing?  E.g. run those decoders and make sure they don't crash?  They will have quite similar interfaces.. the simpler ones may take fewer args and fewer options though.

PR #3638, opened 2019-10-04
Added tri3b and chain training for Aurora4

desh2608:
The existing Aurora4 recipe only had training upto tri2b. Added tri3b (LDA+MLLT+SAT) and chain training based on latest WSJ recipe. Also added decoding results on test_0166 set in addition to test_eval92 set.


danpovey:
Thanks!  Merging.

PR #3639, opened 2019-10-04
[src] Fix bug in resampling checks for online features

entn-at:
The `--allow-upsample` and `--allow-downsample` options were not handled correctly in the code that handles resampling in online feature computation.

Compare with the check for offline feature computation:
https://github.com/kaldi-asr/kaldi/blob/743eb23d897e8b1e78b306e920b906ddfe792c8f/src/feat/feature-common-inl.h#L40

### DIFF
--- a/src/feat/online-feature.cc
+++ b/src/feat/online-feature.cc
@@ -88,10 +88,10 @@ void OnlineGenericBaseFeature<C>::MaybeCreateResampler(
   if (resampler_ != nullptr) {
     KALDI_ASSERT(resampler_->GetInputSamplingRate() == sampling_rate);
     KALDI_ASSERT(resampler_->GetOutputSamplingRate() == expected_sampling_rate);
-  } else if (((sampling_rate > expected_sampling_rate) &&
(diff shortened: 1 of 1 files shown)


nshmyrev:
@entn-at Can we please revisit this, here:

```
             ((sampling_rate > expected_sampling_rate) &&
              !computer_.GetFrameOptions().allow_upsample)) {	              computer_.GetFrameOptions().allow_upsample)) {
```
we convert from sampling_rate to expected_sampling_rate means if sampling_rate is bigger than expected_sampling_rate we actually downsample, not upsample.

it should be 

```
             ((sampling_rate > expected_sampling_rate) &&
              !computer_.GetFrameOptions().allow_upsample)) {	              computer_.GetFrameOptions().allow_downsample)) {
```

danpovey:
You're right, it seems to be backwards.  If @entn-at doesn't submit a PR, I hope you remind me about this, Nickolay!

entn-at:
@nshmyrev you are right, my apologies, the conditions in my previous fix were backwards. I opened a PR #3792.

PR #3641, opened 2019-10-08
Bump OpenBLAS version to 0.3.7 and enable locking

sayint-ai:
Bumps OpenBLAS version to 0.3.7 for improved performance with Zen architecture. Also sets USE_LOCKING which fixes issue when calling a single-threaded OpenBLAS from several threads in parallel


sayint-ai:
Looks like there is already a pull request pending due to branch name being common. Closing this request and recreating a new one with a different branch name

PR #3538, opened 2019-08-26
Use OpenBLAS 0.3.7 that allows USE_LOCKING flag for better multithrea…

ahmetaa:
…ding support.

PR #3642, opened 2019-10-08
Bump OpenBLAS version to 0.3.7 and enable locking

sayint-ai:
Bumps OpenBLAS version to 0.3.7 for improved performance with Zen architecture. Also sets USE_LOCKING which fixes issue when calling a single-threaded OpenBLAS from several threads in parallel


sayint-ai:
This is a duplicate of https://github.com/kaldi-asr/kaldi/pull/3538. However since that request is pending over the last few days due to issues with conflicts in branch names, I'm creating a new pull request

PR #3643, opened 2019-10-08
[src] Add batched xvector computation

danpovey:


danpovey:
TODO: test this and write the script.  I have not run this code.

danpovey:
Merging this, but we still need to try this in the x-vector extraction scripts, e.g. see whether we can replace the old version with this one.

danpovey:
That is expected. The question is whether there is a consistent or substantial change in performance.

PES2g:
Sorry to delete the last message, as i think it's better to compare final performance. In my initial experiment for speaker diarization, there's little performance regression, DER is 26.40 for nnet3-xvector-compute. 27.44 is for nnet3-xvector-compute-batched. lower is better.

danpovey:
What is the duration of these utterances?
You might want to try a significantly longer chunk-length and see if that affects things.
Right now it's 1.5 seconds I think.

PES2g:
In my experiments, i follow the same dihard 2018 diarization egs in kaldi. During training plda, i use 3 seconds as chunk size, and during test i use 1.5 seconds as chunk size.

danpovey:
As far as the test data is concerned, the fact that your chunk size during
test (1.5 seconds) is exactly the same as the default chunk size of the
batched computation (1.5 seconds) should be a good thing, and should mean
that the two ways of computing it would give the same result, at least for
the test data.  At least they should be very close.  Can you somehow figure
out how close they are, on average?

For the PLDA-computation data, they won't be exactly the same.  In fact,
because the batched-computation will overlap the chunks by the total
context of the nnet (to ensure that each hidden-frame appears in the
average at least once), the effective chunk size is less than 1.5 seconds
and it might be more optimal to choose, say, 1.7 seconds or 3 seconds as
the chunk size.

It would be good if you could separate out the PLDA-estimation and
test-time effects through another experiment.

PES2g:
I run another experiment. I train the plda model using xvector extracted by nnet3-xvector-compute. And test data is using nnet3-xvector-compute-batched. The result is 26.65(nnet3-xvector-compute-batched) vs 26.40(nnet3-xvector-compute). The difference is small. From the code, there are two difference. 1. different threshold for padding, in nnet3-xvector-compute, there is another variable min-chunk-size, which is threshold for padding. But in nnet3-xvector-compute-batched, only chunk-size is threshold for padding. 2. different padding model, in nnet3-xvector-compute, padding on both sides, but in nnet3-xvector-compute-batched, padding on the right side.

danpovey:
OK.  That difference may just be noise.  That could point to the larger
difference you saw previously being due to the effect on the PLDA
estimation.  Perhaps you could try using a 3-second chunk size for
nnet3-xvector-compute-batched when training the PLDA.

PR #3654, opened 2019-10-14
Fixed inaccurate log in online decoding

Shujian2015:


src/online2/online-timing.cc
```diff
@@ -52,7 +52,7 @@ void OnlineTimingStats::Print(bool online){
     // real-time factor.
     double real_time_factor = (total_time_taken_ - total_time_waited_) /
         total_audio_;
-    KALDI_LOG << "Timing stats: real-time factor for offline decoding was "
+    KALDI_LOG << "Timing stats: real-time factor for online decoding was "
```
danpovey: I'd rather not change it.. that is actually telling you the real-time factor you'd get if it wasn't an online system, i.e. if you didn't have to wait for the input.

Issue #3656, opened 2019-10-15
[scripts] The timecode string of vad result is not consistent


sih4sing5hong5:
I ran scripts like `callhome_diarization/v2/diarization/vad_to_segments.sh`, got `segment`
```
tong0000000-ku0000000-0000-0229 tong0000000 0 2.3
tong0000000-ku0000000-0261-0561 tong0000000 2.61 5.61
```
The timecode in filename is `0229`, but for end time is `2.3`

The bug maybe in https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/utils/segmentation.pl#L356-L366.

jtrmal:
I assume this is resolved.

PR #3657, opened 2019-10-15
[scripts] bug-fix for segmentation timecode

sih4sing5hong5:
#3656 

I'm not familar with perl.
Is there any better solution?

### DIFF
--- a/egs/wsj/s5/utils/segmentation.pl
+++ b/egs/wsj/s5/utils/segmentation.pl
@@ -353,7 +353,7 @@ sub print_segments {
     $i *= 10;
     $num_digits++;
   }
-  $format_str = "%0${num_digits}d"; # e.g. "%05d"
+  $format_str = "%0${num_digits}.0f"; # e.g. "%05.0f" round the same as seconds
 
   foreach $s (@segments) {
     my ($start,$end) = split(":", $s);


egs/wsj/s5/utils/segmentation.pl
```diff
@@ -353,7 +353,7 @@ sub print_segments {
     $i *= 10;
     $num_digits++;
   }
-  $format_str = "%0${num_digits}d"; # e.g. "%05d"
+  $format_str = "%0${num_digits}.0f"; # e.g. "%05.0f" round the same as seconds
```
jtrmal: I don't understand the change.... Wouldn't it generate the same output as the previous one?
If you need correct rounding, you could also do 
$start_seconds * 100 + 0.5
or
int($start_seconds * 100 + 0.5)

kkm000: @jtrmal, Looks like perl does the rounding. So there will be rounding instead of truncating. But I do not know what exact consequences this change would cause.
```
$ perl -e 'printf "%03.0f %03.0f\n", 1.4999, 1.5'
001 002
$ perl -e 'printf "%03d %03d\n", 1.4999, 1.5'
001 001
```
Also, you've closed #3656 as resolved. So should this change go in or not?

> jtrmal commented on 20 Nov 2019
> I assume this is resolved.

jtrmal: good question... I'd assume, since there wasn't any feedback, that the change is not necessary

kkm000: Thanks, I'll close it then.

PR #3666, opened 2019-10-19
Write all output to a single lattice writer instead of one per

luitjens:
iteration.

We are seeing that small corpus are dominated by writer Open/Close.

A better solution is to modify the key with the iteration number and
then handle the modified keys in scoring.  Note that if you have just a
single iteration the key is not modified and thus behavior is as
expected.

### DIFF
--- a/src/cudadecoderbin/batched-wav-nnet3-cuda.cc
+++ b/src/cudadecoderbin/batched-wav-nnet3-cuda.cc
@@ -90,13 +90,13 @@ void FinishOneDecode(const std::string &utt, const std::string &key,
                      int64 *num_frames, double *tot_like,
                      CompactLatticeWriter *clat_writer,
                      std::mutex *clat_writer_mutex, std::mutex *stdout_mutex,
-                     CompactLattice &clat) {
+                     const bool write_lattice, CompactLattice &clat) {
   nvtxRangePushA("FinishOneDecode");
   GetDiagnosticsAndPrintOutput(utt, word_syms, clat, stdout_mutex, num_frames,
                                tot_like);
-  {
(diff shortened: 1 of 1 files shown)


src/cudadecoderbin/batched-wav-nnet3-cuda.cc
```diff
@@ -90,13 +90,14 @@ void FinishOneDecode(const std::string &utt, const std::string &key,
                      int64 *num_frames, double *tot_like,
                      CompactLatticeWriter *clat_writer,
                      std::mutex *clat_writer_mutex, std::mutex *stdout_mutex,
-                     CompactLattice &clat) {
+                     const bool write_lattice, CompactLattice &clat) {
   nvtxRangePushA("FinishOneDecode");
   GetDiagnosticsAndPrintOutput(utt, word_syms, clat, stdout_mutex, num_frames,
                                tot_like);
+  if(write_lattice)
   {
```
danpovey: can you put the { on the same line as the if?  And space after 'if'.
Closing the clat_writer at the end is not needed, the destructor will do it.

src/cudadecoderbin/batched-wav-nnet3-cuda.cc
```diff
@@ -232,8 +227,10 @@ int main(int argc, char *argv[]) {
         std::string utt = wav_reader.Key();
         std::string key = utt;
 
-        // make key unique for each iteration
-        key = key + "-" + std::to_string(iter);
+        if(iterations>0) {
```
danpovey: space after if and around >

luitjens:
just did these, rebased, and pushed.

PR #3668, opened 2019-10-20
[src,build] Add minimal example to demonstrate pybind11 binding

danpovey:


pzelasko:
Perhaps you'll find scikit-build helpful - it's a utility for building different kinds of native extensions for Python with support for CMake (when/if you decide to migrate to CMake) https://github.com/scikit-build/scikit-build

danpovey:
Thanks.
I know that pybind11 building is very easy with CMake, but I wanted to
fully understand the actual process before handing it off to an opaque
build system.  We're working in parallel on the CMake build.
Actually it turns out that the build process for pybind11 is really quite
simple, even with conventional Makefiles.

src/pybind/README.md
```diff
@@ -4,7 +4,11 @@ you can cd to here and do:
 # make
 # python3
 >>> import kaldi_pybind as k
->>> str(k.FloatVector(10))
-' [ 0 0 0 0 0 0 0 0 0 0 ]\n'
->>>
+>>> import numpy as np
+>>> a = k.FloatVector(10)
+>>> b = np.array(a, copy_data = False)
```
naxingyu: `copy_data` is an invalid argument.

naxingyu: should be `copy`

kkm000:
@danpovey, is pybind11 still a thing?

revdeluxe:
> Not sure. With the experience gained from k2, I think we can do a better job about wrapping kaldi to Python (e.g, integration with PyTorch).

Rustlang has cargo and it's quick too. I ran into some runtime issues with Cmake last night. (June 7,2022)
I think wrapping it using rust's cargo would be faster in runtime and buildtime.(But i admit it's a learning curve. but at least errors are handled in-code.)

A suggestion nothing more nothing less;

csukuangfj:
> > Not sure. With the experience gained from k2, I think we can do a better job about wrapping kaldi to Python (e.g, integration with PyTorch).
> 
> Rustlang has cargo and it's quick too. I ran into some runtime issues with Cmake last night. (June 7,2022) I think wrapping it using rust's cargo would be faster in runtime and buildtime.(But i admit it's a learning curve. but at least errors are handled in-code.)
> 
> A suggestion nothing more nothing less;

Could you have a look at 
- http://github.com/k2-fsa/k2
- https://github.com/k2-fsa/sherpa

We are putting more effort into the above two projects.

revdeluxe:
> > > Not sure. With the experience gained from k2, I think we can do a better job about wrapping kaldi to Python (e.g, integration with PyTorch).
> > 
> > 
> > Rustlang has cargo and it's quick too. I ran into some runtime issues with Cmake last night. (June 7,2022) I think wrapping it using rust's cargo would be faster in runtime and buildtime.(But i admit it's a learning curve. but at least errors are handled in-code.)
> > A suggestion nothing more nothing less;
> 
> Could you have a look at
> 
> * [http://github.com/k2-fsa/k2](https://github.com/k2-fsa/k2)
> * https://github.com/k2-fsa/sherpa
> 
> We are putting more effort into the above two projects.

Ill see what i can share...

PR #3672, opened 2019-10-21
use data type size_t to avoid integer overflow

jiedanzhu2307:
use data type size_t to avoid integer overflow

tested on local machine, everything is good

### DIFF
--- a/src/lat/word-align-lattice-lexicon.cc
+++ b/src/lat/word-align-lattice-lexicon.cc
@@ -116,7 +116,7 @@ class LatticeLexiconWordAligner {
     size_t Hash() const {
       VectorHasher<int32> vh;
       const int32 p1 = 11117, p2 = 90647, p3 = 3967, p4 = 3557; // primes.
-      int32 ans = 0;
+      size_t ans = 0;
       for (int32 i = 0; i < static_cast<int32>(transition_ids_.size()); i++) {
         ans *= p1;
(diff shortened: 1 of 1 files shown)


jtrmal:
I'm confused about this -- does it really matter that it overflows? It's supposed to be hash afterall...

jiedanzhu2307:
@jtrmal,  if we use gcc -O0 or -O1 to debug word alignment,  it would crash due to integer overflow.

danpovey:
Integer overflow is normal and shouldn't cause a crash.  Were you using some unusual compiler options or debugger?

danpovey:
... but we can still merge this, a higher-order hash is no problem and the return value is already size_t.

jtrmal:
yes, I agree -- perhaps the original implementation returning limited
"dynamic range" is not completely right, I'm not sure.
@jiedanzhu2307 what version of gcc? Are you setting -fwrapv or -ftrapv or
the negated versions of these options?
y.

jiedanzhu2307:
@jtrmal @danpovey  the compiler is gcc 5.5 and there are some compiler options configured for the debugger with different optimization levels (sorry I am not sure at the debugger internal config)

danpovey:
OK, it looks like the issue is that signed integer-arithmetic overflow technically leads to undefined behavior in the C++ standard.

PR #3681, opened 2019-10-24
Avoid create token with infinity AM cost.

sipida:
As stated in the comments. The new code avoid the situation token is created with infinity cost in the forward link structure.


danpovey:
... actually, can you please see if there are similar lines in other decoders that could be changed the same way?

kkm000:
Fixed in #3773. Thanks!

PR #3682, opened 2019-10-25
Use buffer protocol for matrix

DongjiGao:


danpovey:
this should go to the pybind branch in my repo actually.. not here!

PR #3684, opened 2019-10-25
Kids asr

Fei00Wu:
ASR recipe for CMU_Kids and CSLU_Kids.


danpovey:
just some stuff I noticed from a glance:
don't have caps, e.g. in tdnnF.
Make sure all the example scripts you check in, e.g. in local/chain/tuning/, have actually been run and have relevant results from YOUR database present in a comment at the top.  For a new recipe you would normally check in just a simgle example script, probably numbered 1a.

danpovey:
I'm assuming we merged this via a different PR.

PR #3685, opened 2019-10-26
fix the folder name in run_tdnn_discriminative.sh

jiangj-apptek:
In egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh, the mfcc_hires folder is not referred anywhere else.

### DIFF
--- a/egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh
+++ b/egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh
@@ -95,8 +95,8 @@ if [ $frame_subsampling_factor -ne 1 ]; then
 
     data_dirs=
     for x in `seq -$[frame_subsampling_factor/2] $[frame_subsampling_factor/2]`; do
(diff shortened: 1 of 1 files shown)


egs/librispeech/s5/local/chain/run_tdnn_discriminative.sh
```diff
@@ -96,7 +96,7 @@ if [ $frame_subsampling_factor -ne 1 ]; then
     data_dirs=
     for x in `seq -$[frame_subsampling_factor/2] $[frame_subsampling_factor/2]`; do
       steps/shift_feats.sh --cmd "$train_cmd --max-jobs-run 40" --nj 350 \
```
danpovey: This is not really the fix I wanted, sorry.. I think you should remove that arg (mfcc_hires) and change steps/shift_feats.sh (which is deprecated) to utils/data/shift_feats.sh.

jiangj-apptek: Ahha, got it.

PR #3686, opened 2019-10-26
Remove pitch from nnet3 recipe

naxingyu:
This is related to a discussion in the forums about learning pitch from raw feature. This is also necessary fix for this recipe because in the current version, convolution components does not treat the pitch features properly. Also slight improvements can be seen from the result diff.


danpovey:
Is there any chance you could make sure the results are shown at the top of the chain chaining script in a comment?  E.g. add a modified version of egs/mini_librispeech/s5/local/chain/compare_wer.sh and use that to display it?

naxingyu:
Done. In the previous setup, pitch is not corretly used, as I used idct to deal with the whole feature set, it was a mistake. But anyways, removing pitch should do no harm.

naxingyu:
@danpovey I was trying to change "WER" to "CER" in the chain recipe header, do I proceed?

danpovey:
sure.

naxingyu:
It is committed  #3687

PR #3688, opened 2019-10-28
Reset on control thread instead of worker threads.

luitjens:
Resetting on worker thread would mean the thread deallocating the memory
would be different than the thread that allocated/used the memory.

It is unclear if this is a supported use case.  In the worst case this
could lead to correctness issues.  In the best case it would lead to
excessive synchronization.  Either way we should have the thread that
allocated the memory also deallocate the memory.


luitjens:
please hold on merging while i test more.  Not sure why this would break but i need to test more.

jtrmal:
Just closing and opening to force build on Travis, trying to debug something.

luitjens:
please close this.  Going to resubmit with a more complete patch that fixes an existing corruption issue.

Issue #3690, opened 2019-10-28
Need configure option to control run-time assertions


iskunk:
As I understand it, the Kaldi source has assertions that can be enabled/disabled at compile time in one of three forms:

* Default: Some assertions are enabled

* `-DKALDI_PARANOID`: Enables additional, potentially CPU-intensive assertion checks (the [documentation](https://kaldi-asr.org/doc/error.html) currently states that this is enabled by default, but that does not appear to be the case)

* `-DNDEBUG`: Disables all assertions

This modality can have a significant impact on the operation of the resulting compiled libraries, and so an option for it ought to be present in the configure script, to facilitate its discovery and customization.

(Currently, `-DKALDI_PARANOID` is added in commented-out form to `kaldi.mk` without any description, and nothing is said about `NDEBUG`.)

danpovey:
You're probably right.  You could perhaps add configure options?  If you have time?
@cloudhan it's something that could be considered for the CMake build as well.
@cloudhan, It just occurred to me that we need very clear and complete instructions for how to configure the CMake build, possibly in cmake/README.md.    The problem with CMake (versus our configure script) is that it's enormously harder to discover what the options are.

iskunk:
I could submit a patch. Do you have any preferences for what such an option should be named?

(Side note: CMake is actually not bad about option discovery, if you make use of its GUI or ncurses frontends. Either will present a list of options, as well as a checkbox to hide/show an "advanced" subset. As far as I'm aware, it only lacks a command-line equivalent to `configure --help`.)

danpovey:
Perhaps --debug-level N with N = 0, 1 or 2 and 1 being the default?

iskunk:
I was, in fact, thinking along similar lines :-)  Let me get back to you on this...

cloudhan:
For cmake, with `-DCMAKE_BUILD_TYPE=Release`, `NDEBUG` is automatically defined. Opt level is default to O3. There are other build types, e.g.,  `RelWithDebInfo` will build with debug symbol, `NDEBUG` is defined, Opt level is usually O2. These options is platform dependent, and I think they should go into the toolchain configuration as in [vpckg](https://github.com/microsoft/vcpkg/tree/master/scripts/toolchains).

Adding these definitions should be easy. All defs go into the top-level `CMakeLists.txt` will do the thing.

[...]

PR #3700, opened 2019-11-05
[src] Add --debug-level=N configure option to control assertions (#3690)

iskunk:
Proposed implementation of `--debug-level=N` option for Kaldi `src/configure` script, as discussed in #3690. Feedback welcome.

### DIFF
--- a/src/makefiles/linux_atlas_ppc64le.mk
+++ b/src/makefiles/linux_atlas_ppc64le.mk
@@ -1,5 +1,8 @@
 # ATLAS specific Linux ppc64le configuration
 
+ifndef DEBUG_LEVEL
+$(error DEBUG_LEVEL not defined.)
+endif
 ifndef DOUBLE_PRECISION
 $(error DOUBLE_PRECISION not defined.)
 endif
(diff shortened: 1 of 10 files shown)


danpovey:
Looks great!
Let's wait for the travis build to finish, or at least give it a chance, as it's there to detect build issues.

src/makefiles/linux_atlas_ppc64le.mk
```diff
@@ -1,5 +1,8 @@
 # ATLAS specific Linux ppc64le configuration
 
+ifndef DEBUG_LEVEL
+$(error DEBUG_LEVEL not defined.)
+endif
```
kkm000: Not a show-stopper, but I am wondering why do we carry this (and many similar checks) over so many files. Yes, kaldi.mk better be sane, but on the other hand, all these copy-paste sanity checks are so numerous that they nearly defeat their own purpose.

iskunk: I agree. I just followed existing conventions for the feature work, but yes, this is a clear case where duplicate code ought to be factored out.

danpovey:
@kkm000 says it's OK, so merging...

PR #3705, opened 2019-11-07
[egs] Change make_rttm.py to read/write files with UTF-8 encoding

entn-at:
The benefits offered by `common_lib.smart_open` would get lost with this change; however, they aren't used in any of the currently checked-in recipes (the only call location is in `cluster.sh` in the same directory, which uses regular file paths).


danpovey:
Looks fine to me- merging.

PR #3653, opened 2019-10-14
Splitting data evenly when utt2dur exists

RuABraun:
This change means that when using `split_data.sh`, if an utt2dur file exists the data will be distributed evenly (while keeping speakers in the same split). 

Opening the PR following discussion here: https://groups.google.com/forum/#!topic/kaldi-help/sbgQcmyIOwk

Note the split will not be even if one is just splitting by utterance, but I could add that. Thought I should first open the PR and see whether the general approach is acceptable.

Sometimes a warning will appear: `filter_scps.pl: warning: some input lines were output to multiple files [OK if splitting per utt]`, this is because when working with long audio files if the speakers go in different splits the audio files will go in both (and were therefore "output to multiple files"). Maybe that warning should be changed.

PR #3710, opened 2019-11-09
Split_scp.pl will try and redistribute speakers when there is an imbalance in the preliminary split

RuABraun:
This should fix the issues #3653 had.

However, this makes things a good deal more complicated, and while it worked for me on a dataset I manufactured to have a major imbalance in speaker durations, I'm nervous there may be something I missed, would be nice if some people could try it out.  
And it can be a lot slower.

Another option would be to just detect when there is an imbalance and crash with a message telling people to use `utils/data/modify_speaker_duration.sh`. Not ideal either of course I know but just throwing it out there... I think people with datasets that will cause issues will tend to be more experienced users and would appreciate the tip (it will probably speed their builds up).

### DIFF
--- a/egs/wsj/s5/utils/split_data.sh
+++ b/egs/wsj/s5/utils/split_data.sh
@@ -16,13 +16,18 @@
 # limitations under the License.
 
 split_per_spk=true
+allow_uneven_split=false
 if [ "$1" == "--per-utt" ]; then
   split_per_spk=false
   shift
 fi
+if [ "$1" == "--allow-uneven-split" ]; then
+    allow_uneven_split=true
+    shift
+fi
 
 if [ $# != 2 ]; then
-  echo "Usage: $0 [--per-utt] <data-dir> <num-to-split>"
+  echo "Usage: $0 [--per-utt] [--allow-uneven-split] <data-dir> <num-to-split>"
   echo "E.g.: $0 data/train 50"
   echo "It creates its output in e.g. data/train/split50/{1,2,3,...50}, or if the "
   echo "--per-utt option was given, in e.g. data/train/split50utt/{1,2,3,...50}."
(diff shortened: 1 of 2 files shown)


danpovey:
Looks a bit complicated to me; I'd be more comfortable detecting where,
say, smallest-split-duration is less than largest-split-duration/2 AND
largest_split_duration > 1 hour, and dying with that informative error
message.  People should be doing that anyway.

However, I think you should also accept the change that I proposed.

The issue with the current code is that it systematically rounds up.  I.e.
each split is slightly larger than the desired length and inevitably there
will be no data left for the last few splits if the speaker lengths are
quite long and there are many splits.

danpovey:
... although perhaps an option to split_data.sh like --allow-uneven-splits
that could override that crash would be useful.
There may be situations, with test data, where people won't want to mess
with the speaker identities.

jtrmal:
I wonder if it would be useful to revert the change and add a new option
(--balance-duration) to split-data-dir.sh to avoid regressions?
y.

danpovey:
I see what you mean, but people would never give that option.  And if we
could balance things, it would be a significant benefit.

jtrmal:
That I do not argue about -- I completely agree.
y.

RuABraun:
Okay, I'll make a new commit which will die with a useful message when the condition you mentioned fails.

> The issue with the current code is that it systematically rounds up.  I.e.
each split is slightly larger than the desired length and inevitably there
will be no data left for the last few splits if the speaker lengths are
quite long and there are many splits.

Fair point, will add as well.

danpovey:
Thanks!  You've tested it, I assume?

RuABraun:
Not enough, I'll post a comment when I feel confident it's all good.

RuABraun:
Good to go now.

PR #3711, opened 2019-11-10
Change the Python diagnostic scripts to accept non-ASCII UTF-8 phone set

vdp:
Hi,
not sure if I'm missing something obvious, but apparently these scripts won't work if the phone names are outside of the ASCII range. It's a bit surprising that no one else is having this problem, but it's likely that:
* people (perhaps prudently) avoid non-ASCII symbols in their phone sets
* these scripts are only for diagnostic purposes, so not critical for a recipe's execution

Summary of the changes:
* sys.stdout is replaced, so that it outputs UTF-8 symbols w/o complaining and w/o escaping them even when LC_ALL is set to 'C', which is the usual environment for Kaldi's recipes
* phones.txt is read as UTF-8 encoded file
* most of the "strings" are replaced w/ u"strings" for compatibility with Python2 where Unicode is not the default

The background of this is that I'm fiddling w/ a toy Bulgarian AM(weekend project) and I'm using cyrillic letters as phone names, b/c more they feel more intuitive to me as a native speaker.

### DIFF
--- a/egs/wsj/s5/steps/diagnostic/analyze_lattice_depth_stats.py
+++ b/egs/wsj/s5/steps/diagnostic/analyze_lattice_depth_stats.py
@@ -9,6 +9,15 @@ from __future__ import division
 import argparse
 import sys, os
 from collections import defaultdict
+from io import open
+import codecs
+
+# reference: http://www.macfreek.nl/memory/Encoding_of_Python_stdout
+if sys.version_info.major == 2:
+    sys.stdout = codecs.getwriter('utf-8')(sys.stdout, 'strict')
+else:
+    assert sys.version_info.major == 3
+    sys.stdout = codecs.getwriter('utf-8')(sys.stdout.buffer, 'strict')
 
 
 parser = argparse.ArgumentParser(description="This script reads stats created in analyze_lats.sh "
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/steps/diagnostic/analyze_lattice_depth_stats.py
```diff
@@ -29,13 +38,13 @@
 # set up phone_int2text to map from phone to printed form.
 phone_int2text = {}
 try:
-    f = open(args.lang + "/phones.txt", "r");
+    f = open(args.lang + "/phones.txt", "r", encoding='utf-8')
```
jtrmal: this wouldn't work under python2, I believe (it does not have the encoding= option

jtrmal:
IMO, if you are touching the files (and willing to spend the time on it), you could do it plain "python3" script (perhaps changing the hashbang would be sufficient).  And you could simplify the getwriter logic.
Other than that, LGTM

egs/wsj/s5/steps/diagnostic/analyze_lattice_depth_stats.py
vdp: Hi Yenda, thanks for having a look!
You are right that the built-in open() doesn't have an "encoding" option. That's what the "from io import open" line is- i.e. we are re-binding the 'open' variable to point to a function that has this option and works for both Python2 and Python3. I've tested the scripts w/ Python versions 2.7 and 3.6.

vdp:
Hmm, I am not sure I understand the meaning of your "plain python3" suggestion... I thought the scripts are supposed to be backward compatible for the time being(it would certainly make sense to drop the Python2 baggage for the next-gen Kaldi Dan is working on).
Was it decided to not support Python2 even for the current version?

jtrmal:
I didn't notice that line, sorry!

danpovey:
Kaldi is already dependent on python3 so in general python3-only is what I would choose these days.  But since Vassil has already tested it for both it's no problem.  I'll merge.  Thanks!

PR #3717, opened 2019-11-14
[scripts] Removing changes to split_scp.pl from #3653, #3710

danpovey:


danpovey:
Please see this kaldi-help thread for details

https://groups.google.com/d/msgid/kaldi-help/407fd82b-95c1-4652-ba4a-9dc064b12737%40googlegroups.com?utm_medium=email&utm_source=footer

Below is a potentential patch that might fix it (assuming one were to revert this PR first).
BUT I need to understand why these things failed first before I would be comfortable to stick with these changes.
```
b19:utils: git diff -b HEAD split_scp.pl
diff --git a/egs/wsj/s5/utils/split_scp.pl b/egs/wsj/s5/utils/split_scp.pl
index a25dbf55a..ceab97e97 100755
--- a/egs/wsj/s5/utils/split_scp.pl
+++ b/egs/wsj/s5/utils/split_scp.pl
@@ -184,10 +184,13 @@ if ($utt2spk_file ne "" && $utt2dur_file ne "" ) {  # --utt2spk and --utt2dur
             }
         }
     }
+    if ($scpidx < $numscps) {
+      $scp2dur[$scpidx] = $dursum_current;
+    }

     $smallest_dur = $splitdur;
     $largest_dur = $splitdur;
-    for($scpidx = 0; $scpidx < $numscps; $scpidx++) {
+    for ($scpidx = 0; $scpidx < $numscps; $scpidx++) {
         $scpdur = $scp2dur[$scpidx];
         if ($scpdur > $largest_dur) {
             $largest_dur = $scpdur;
@@ -200,9 +203,9 @@ if ($utt2spk_file ne "" && $utt2dur_file ne "" ) {  # --utt2spk and --utt2dur
     if ($allow_uneven_split != 1) {
         if (($smallest_dur < $largest_dur / 2 && $largest_dur > 3600) ||
             $smallest_dur == 0.0) {
-            die "Trying to split data while taking duration into account leads to a " .
+            die "$0: Trying to split data while taking duration into account leads to a " .
                 "severe imbalance in splits. This happens when there is a lot more data " .
-                "for some speakers than for others.\n" .
+                "for some speakers than for others (smallest,largest) dur are $smallest_dur,$largest_dur.\n" .
                 "You should use utils/data/modify_speaker_duration.sh to fix that.\n"
         }
     }
```

RuABraun:
Sorry about the bother...

danpovey:
No problem... I should have checked more carefully.

danpovey:
If you want you can make a new PR with my patch applied.  I think I understand now what happened, and I think with that patch it shouldn't happen again.

danpovey:
@desh2608 can you please make a PR to put back these changes?  I think I understand why it failed now.
You'd do
git revert 7e7760175f5
Then apply my patch above by:
cd to top-level of kaldi
patch -p1 
(paste my patch above and press ctrl-D.  You may have to change the 1 to 0,,2, etc., depending on your dir and whether I made a mistake).

There is another issue though.
Some shell scripts do not properly check the return status of split_data.sh.  You can find them with:
find steps -exec grep split_data.sh {} \; -print
If the script does not check it and does not have
set -e
somewhere in it, then please add
`|| exit 1`
after the invocations.
See if there are any Python scripts with the same issue, as well.

desh2608:
@danpovey Okay I'm on it.

PR #3725, opened 2019-11-20
Applied patch for split_scp.pl

desh2608:
* Reverted commit https://github.com/kaldi-asr/kaldi/commit/7e7760175f5bb85391fe6967908a7b100ab296f5
* Applied patch from https://github.com/kaldi-asr/kaldi/pull/3717#issuecomment-553682775 
* Added `|| exit 1;` to uses of `split_data.sh` within `steps/`


danpovey:
Thanks . Because last time was painful, could you please make sure the mini_librispeech recipe runs from start to finish?  Don't want to break anything.

desh2608:
@danpovey I tried running mini-librispeech with the new changes, but it throws the error at [this stage](https://github.com/kaldi-asr/kaldi/blob/eb28a6ae35ccb332df1215d0bfc353cec03df76d/egs/wsj/s5/steps/align_fmllr_lats.sh#L57): 
```
steps/align_fmllr_lats.sh --nj 75 --cmd queue.pl --mem 2G data/train_clean_5_sp data/lang exp/tri3b exp/chain_online_cmn/tri3b_train_clean_5_sp_lats
utils/split_scp.pl: Trying to split data while taking duration into account leads to a severe imbalance in splits. This happens when there is a lot more data for some speakers than for others (smallest,largest) dur are 0,2038.449806.
You should use utils/data/modify_speaker_duration.sh to fix that.
```

I tried adding `--allow-uneven-split` to the `split_data.sh` call inside `align_fmllr_lats.sh` but then it creates empty splits. Also, should the `utils/data/modify_speaker_duration.sh` in that error message be `utils/data/modify_speaker_info.sh`

RuABraun:
Trying it out it looks like there are a significant number of speakers with a longer duration than the target split duration, which makes splitting evenly difficult obviously, and in this case leads to the last split having no data. 

The primary goal of `--allow-uneven-split` was to avoid having to change speaker IDs while allowing for some splits to be a lot smaller than others. It can still lead to a crash when a split ends up empty. To reduce the likelihood of a failure (guaranteeing no failure is a bit tricky I think) occurring one could try moving a speaker from a split with several to an empty split as I did in #3653 , although one would want to make sure the speaker you move has a non-trivial amount of data (otherwise you end up with the crash that guy had that started this whole fiasco).

> Also, should the utils/data/modify_speaker_duration.sh in that error message be utils/data/modify_speaker_info.sh

Yeah it should be, my mistake.

danpovey:
I think we should modify it so that when assigning speakers to splits, it
keeps track of how many speakers remain; and it will disallow adding a
speaker to the current split if the number of remaining speakers is less
than the number of remaining splits.

RuABraun:
That seems like a nice solution. I didn't get the error anymore after changing things:
```
diff --git a/egs/wsj/s5/utils/split_scp.pl b/egs/wsj/s5/utils/split_scp.pl
index 69f158d..96db9eb 100755
--- a/egs/wsj/s5/utils/split_scp.pl
+++ b/egs/wsj/s5/utils/split_scp.pl
@@ -172,10 +172,13 @@ if ($utt2spk_file ne "" && $utt2dur_file ne "" ) {  # --utt2spk and --utt2dur
     for my $spk (sort (keys %spk2utt)) {
         $scpcount[$scpidx] += $spk_count{$spk};
         push @{$scparray[$scpidx]}, $spk;
+        $numspks--;
         $dur = $spk2dur{$spk};
         $dursum += $dur;
         $dursum_current += $dur;
-        if ($dursum >= $splitdur * ($scpidx + 1) && $dursum_current > 10.0) {
+
+        $num_split_left = $numscps - $scpidx - 1;
+        if (($dursum >= $splitdur * ($scpidx + 1) && $dursum_current > 10.0) || $numspks == $num_split_left) {
             $scp2dur[$scpidx] = $dursum_current;
             $scpidx += 1;
             $dursum_current = 0.0;
```

desh2608:
I made the changes (patch from @RuABraun) and confirmed that it runs without error on mini-librispeech.

egs/wsj/s5/steps/align_fmllr_lats.sh
```diff
@@ -54,7 +54,7 @@ sdata=$data/split$nj
 
 mkdir -p $dir/log
 echo $nj > $dir/num_jobs
-[[ -d $sdata && $data/feats.scp -ot $sdata ]] || split_data.sh $data $nj || exit 1;
+[[ -d $sdata && $data/feats.scp -ot $sdata ]] || split_data.sh --allow-uneven-split $data $nj || exit 1;
```
danpovey: After thinking about this more, I think it will cause too many problems to existing recipes if we make it an error to have uneven splits.  Best to have it just be a warning.  (Should go to stderr, obviously: print STDERR "string...";)

desh2608: Sorry, I didn't quite get what you are asking. Do you mean instead of the `die` [here](https://github.com/kaldi-asr/kaldi/blob/6d06c1bedcc90802a221bbae507799391beccf2e/egs/wsj/s5/utils/split_scp.pl#L209), there should be a Warning printed to STDERR instead?

danpovey:
Yes.

danpovey:
And of course remove the option --allow-uneven-split

desh2608:
I have made the changes.

egs/wsj/s5/utils/split_data.sh
```diff
@@ -16,20 +16,26 @@
...
   shift
 fi
+if [ "$1" == "--allow-uneven-split" ]; then
+    allow_uneven_split=true
+    shift
+fi
 
 if [ $# != 2 ]; then
-  echo "Usage: $0 [--per-utt] <data-dir> <num-to-split>"
+  echo "Usage: $0 [--per-utt] [--allow-uneven-split] <data-dir> <num-to-split>"
```
danpovey: The --allow-uneven-split is not necessary any more.

desh2608: Ok I'll remove this option from everywhere.

danpovey:
Thanks a lot.  Please test mini_librispeech one more time.  I want to be extra careful as this could break a lot of things.

desh2608:
I ran mini-librispeech again. It runs fine without any errors (I did get an error about creating a directory in b14 node, but that might be because b14 and several others are not accessible. Ran alright when I removed b14 from the list).

danpovey:
Ha-- it's Hopkins' spineless administration to blame.  They are so scared of the student radicals that they are making it impossible for the sysadmins to get into Garland out of ours.

PR #3739, opened 2019-11-26
Cuallocator backwards logic

btiplitz:
This is a partial fix to the memory allocator.  The memory allocator has an increasing list of free blocks and uses this when determining the next largest block, but it uses the 1st one and not the last.  This is NOT a complete fix in that the largest_free_block_ does not consider regions.  When the 2nd region is allocated, the logic appears to fail as the next largest block is grabbed from the subregion being replaced and thus I believe the largest_free_block_ code is not safe across a 2nd malloc from the gpu


danpovey:
You are right, this was wrong.  Can you please point to the line of code that you still feel was wrong though?

danpovey:
And a reminder to everyone who has PRs up: please remind me if you feel I need to do something and it's been a couple of days.  I am not always as proactive as I'd like to be.

PR #3745, opened 2019-11-30
fix slurm.pl error

naxingyu:
The `don't` in the comment cause this error every time slurm.pl is called.
```
sh: 13: Syntax error: Unterminated quoted string
```


kkm000:
Interesting, never hit that one. But I've been bitten before by the fact that Perl's backticks use hard-coded 'sh', not respecting user's $SHELL, with not even an option to change it.

FWIW, I rewrote slurm.pl nearly from scratch for my cloud setup, that I am still working on releasing, time allows. @danpovey, what do you think--is it safer to check it in under a different name, like slurm2.pl? The original script seems to have been cloned with minor changes from queue.pl, which, in turn, has a lot of quirks that someone (you, likely?) added fighting unrefined manners of a particular NFS setup on a particular cluster. I had zero issues with NFSv4 after having them removed.

danpovey:
I think I would be OK to just replace slurm.pl and if anyone has problems
we can tell them to look in the git history and try the older version.

PR #3746, opened 2019-12-01
Revert "Applied patch for split_scp.pl"

danpovey:
Reverts kaldi-asr/kaldi#3725


danpovey:
Sorry guys, my fault, I don't think we can merge this PR after all.  See discussion here:
@desh

https://groups.google.com/d/msgid/kaldi-help/765e5097-be46-4597-bc5a-76e72c0fab1a%40googlegroups.com.

It's an issue of switching over and breaking existing setups.  We should remember to make this change whenever we make another breaking change (e.g. kaldi10)l

Issue #3734, opened 2019-11-23
Move CheckAndFixConfigs() outside of NnetBatchComputer


hugovbraun:
We currently have a function [CheckAndFixConfigs()](http://kaldi-asr.org/doc/classkaldi_1_1nnet3_1_1NnetBatchComputer.html#aeb4fbab0a662101fff04c90e166ae8e0) in the NnetBatchComputer that makes sure that frames_per_chunk is a multiple of frame_subsampling_factor. However that functionality is useful outside of just the NnetBatchComputer. Would it be acceptable to move it to the [NnetSimpleComputationOptions](http://kaldi-asr.org/doc/structkaldi_1_1nnet3_1_1NnetSimpleComputationOptions.html) ? I can take care of it, just need the green light.

danpovey:
Sure you can move it to NnetSimpleComputationOptions().

danpovey:
I mean NnetSimpleComputationOptions

hugovbraun:
Ok, will do

PR #3749, opened 2019-12-02
Move CheckAndFix to config struct

hugovbraun:
Related to #3734


hugovbraun:
First commit is just clang-format formatting to C++ coding style. The changes are in the second commit

danpovey:
Hm, I'm OK with the main change but I don't like the clang formatting thing.  It already had a fairly consistent coding style (but different from what clang is doing), and if we were to consistently apply the clang thing everywhere it would pollute the git history and make it hard to track down meaningful changes.

hugovbraun:
Ok, I'll revert this on this PR and the other one

hugovbraun:
Revert done

hugovbraun:
Testing ok

PR #3758, opened 2019-12-05
[egs] Add recipes for CN-Celeb

csltstu:
This branch adds an example scripts for CN-Celeb speaker verification (egs/cnceleb). The CN-Celeb corpus is required, and can be downloaded from Openslr http://www.openslr.org/82/.

This example scripts contains two subdirectories. The subdirectories v1 and so on are different speaker recognition recipes. The example in v1 demonstrates a standard approach using a full-covariance GMM-UBM, iVectors, and a PLDA backend. The example v2 demonstrates DNN speaker embeddings with a PLDA backend.


danpovey:
@david-ryan-snyder @leibny who can review this?

egs/cnceleb/v1/local/choose_utts_to_combine.py
```diff
@@ -0,0 +1,387 @@
+#!/usr/bin/env python
+
+# Copyright 2016  Vijayaditya Peddinti
+#           2016  Johns Hopkins University (author: Daniel Povey)
+#           2019  Tsinghua University (author: Jiawen Kang)
+#
+# Copied from utils/data/internal/choose_utts_to_combine.py
```
david-ryan-snyder: What, if any, is the difference between this file and the one in utils/data/internal/choose_utts_to_combine.py?

If the differences are small, then I think it would be better to add some options to the original script to support whatever functionality is lacking, rather than copying the whole file and modifying it.

egs/cnceleb/v1/local/combine_short_segments.sh
```diff
@@ -0,0 +1,179 @@
...
+
+# This script copies and modifies a data directory while combining
+# segments whose duration is lower than a specified minimum segment
+# length.
+#
+# Note: this does not work for the wav.scp, since there is no natural way to
+# concatenate segments; you have to operate on directories that already have
+# features extracted.
+
+# Copied from utils/data/combine_short_segments.sh
```
david-ryan-snyder: As with the previous script, would you mind explaining what the difference is between this script and the one in utils? 

In general, it's best to try and make things work with existing scripts in the utils/ or steps/ directory, rather than making a new one in local/ if it does almost the same thing. If the utils/ version doesn't quite give you what you want, you might find that you can provide some option to the script that supports the required functionality.

egs/cnceleb/v1/local/make_cnceleb.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# Copyrigh       2017  Ignacio Viñals
```
david-ryan-snyder: Copyrigh -> Copyright

egs/cnceleb/v1/run.sh
```diff
@@ -0,0 +1,132 @@
...
+# See ../README.txt for more info on data required. The recipe uses
+# CN-Celeb/dev for training the UBM, T matrix and PLDA, and CN-Celeb/eval
+# for evaluation. The results are reported in terms of EER and minDCF,
+# and are inline in the comments below.
+
+. ./cmd.sh
+. ./path.sh
+set -e
+mfccdir=`pwd`/_mfcc
+vaddir=`pwd`/_vad
```
david-ryan-snyder: I don't think I've seen these directories preceded by an underscore before... It seems unusual in Kaldi recipes. I would consider changing that to be consistent with the other recipes.

egs/cnceleb/v2/run.sh
```diff
@@ -0,0 +1,151 @@
...
+    --num-utts=ark:$nnet_dir/xvectors_eval_enroll/num_utts.ark \
+    "ivector-copy-plda --smoothing=0.0 $nnet_dir/xvectors_train_comb/plda - |" \
+    "ark:ivector-mean ark:data/eval_enroll/spk2utt scp:$nnet_dir/xvectors_eval_enroll/xvector.scp ark:- | ivector-subtract-global-mean $nnet_dir/xvectors_train_comb/mean.vec ark:- ark:- | transform-vec $nnet_dir/xvectors_train_comb/transform.mat ark:- ark:- | ivector-normalize-length ark:- ark:- |" \
+    "ark:ivector-subtract-global-mean $nnet_dir/xvectors_train_comb/mean.vec scp:$nnet_dir/xvectors_eval_test/xvector.scp ark:- | transform-vec $nnet_dir/xvectors_train_comb/transform.mat ark:- ark:- | ivector-normalize-length ark:- ark:- |" \
+    "cat '$eval_trials_core' | cut -d\  --fields=1,2 |" $nnet_dir/scores/cnceleb_eval_scores || exit 1;
+
+  # CN-Celeb Eval Core:
+  # EER: 14.70%
+  # minDCF(p-target=0.01): 0.6814
+  # minDCF(p-target=0.001): 0.7979
```
david-ryan-snyder: It's quite unexpected that a well designed x-vector system would produce worse results than a traditional i-vector system. 

I suggest changing this PR so that it only includes the v1 recipe for now. Postpone including the v2 recipe until you know what's wrong with the x-vector system or until you have a better understanding of how to train these kinds of systems. I haven't had time to look at everything that could be improved in the recipe, but I notice that the PR doesn't even include data augmentation in the x-vector training recipe, which we know is an essential step for achieving good performance.

david-ryan-snyder:
Also tagging Ewald (@entn-at) incase he has any additional comments.

egs/cnceleb/v2/run.sh
david-ryan-snyder: In addition to augmentation, you might wish to include some other free, wideband resource, like VoxCeleb in the DNN training data. My guess is that this will substantially improve the x-vector system performance, even though it introduces a language mismatch. You'll still want to use the in-domain data for training the PLDA model, though.

Alternatively, if you feel the recipe should consist of nothing but the CN-Celeb dev and eval datasets, then I would again consider removing the v2 recipe, but retain the v1 recipe. If you do choose to use only this data, then it could be the case that the v2 recipe is simply not appropriate for it.

egs/cnceleb/v1/local/make_cnceleb.sh
csltstu: Thanks a lot !

egs/cnceleb/v1/run.sh
csltstu: Thanks a lot. I have changed that to be consistent with previous recipes.

egs/cnceleb/v2/run.sh
csltstu: Agree! I have removed the v2 recipe.

Actually, we have tried our best to improve the performance on the x-vector system if only choosing the CN-Celeb dev for x-vector training. Data augmentation similar as egs/VoxCeleb does not work in CN-Celeb training in our experiment. While in-domain PLDA training with the x-vector trained with VoxCeleb is effective
Related experimental results can be seen [https://arxiv.org/abs/1911.01799](url).

We hope the recipe only consist of the CN-Celeb dev and eval datasets, and accept your suggestion to remove the v2 recipe.

egs/cnceleb/v1/local/combine_short_segments.sh
csltstu: Thanks a lot. 
The only difference between this script and the one in utils/ is that we add the '--no-text' options in 'utils/validate_data_dir.sh' because here is no text file. 
It is OK to add an option in 'utils/validate_data_dir.sh' or in 'utils/data/combine_short_segments.sh' to ignore checking text file ? Do you think which one is better ?

david-ryan-snyder: I think it would be best to add a --no-text option to utils/data/combine_short_segments.sh. Have it default to false, that way the original functionality is maintained. If it is true, then you can add the --no-text option to the call to utils/validate_data_dir.sh.

egs/cnceleb/v1/local/choose_utts_to_combine.py
csltstu: Thanks a lot.
The only difference between this script and the one in utils/ is that we changed the return value of the function GetUtteranceGroups(). 
For the one in utils/, it will combine utterances across different speakers to satisfy the
duration constraint. This will involve some errors in spk2utt. So here we just combine utterances from the same speaker to ensure the correctness of the combined spk2utt.

egs/cnceleb/v1/local/combine_short_segments.sh
danpovey: I just had a look at it.  David's suggestion was reasonable, but what I'd like to be done is to add the --no-text option to there combine_short_segments.sh calls validate_data_dir.sh (it will still check the text file if it's there), and when doing:
`utils/apply_map.pl -f 2- $srcdir/text <$dir/utt2utts > $dir/text`
just check that the text file is present in $srcdir before doing that.

egs/cnceleb/v1/local/choose_utts_to_combine.py
david-ryan-snyder: I think you'll have to add some option that controls this. I'm not sure what the best name for this would be. Maybe call it --merge-within-speakers-only, and have it default to false? I can't think of a more succinct option name at the moment... as long as the "help" text describes the purpose of the option it should be OK. 

Then you will probably want pass an option to GetUtteranceGroups that toggles between the two ways of merging. If this becomes complicated, you might decide that it's cleaner to make another function, maybe call it GetUtteranceGroupsWithinSpeaker. Then call whichever function is appropriate based on the value of merge_within_speakers_only.

egs/cnceleb/v2/run.sh
david-ryan-snyder: Thanks! You could always decide to add an x-vector recipe later... and then I'd suggest doing whatever produces the best performance, which as you said, appears to be training the DNN on VoxCeleb (or perhaps VoxCeleb + CN-Celeb?) and the backend on CN-Celeb.

egs/cnceleb/v1/local/combine_short_segments.sh
csltstu: Thanks a lot. I have updated it.

egs/cnceleb/v2/run.sh
csltstu: Thanks! I will try it later to make the best performance on x-vector.

egs/cnceleb/v1/local/choose_utts_to_combine.py
csltstu: Thanks a lot. 
I have modified the scripts 'choose_utts_to_combine.py' and 'combine_short_segments.sh' in utils/, and remove them in local/. 

Besides, I found there is an error in our previous experiment due to the 'choose_utts_to_combine.py', and I will rebuild the i-vector and x-vector egs and update the experimental results. 
I will feed it back in tomorrow.

csltstu:
Dear Dan and David, I have completed all these issues, and the current recipes in this PR are ready to review.

danpovey:
Thanks, I prefer that David check this but if he doesn't respond by Friday, ping me.

david-ryan-snyder:
Thanks @csltstu, I will get to this soon (probably on Tuesday).

egs/cnceleb/v1/local/nnet3/xvector/prepare_feats_for_egs.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
csltstu: Thanks a lot. I have removed it from the PR.

egs/cnceleb/v1/local/nnet3/xvector/run_xvector.sh
```diff
@@ -0,0 +1 @@
+tuning/run_xvector_1a.sh
```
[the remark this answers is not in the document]
csltstu: Thanks a lot. I have removed it from the PR.

egs/cnceleb/v1/local/nnet3/xvector/tuning/run_xvector_1a.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
csltstu: Thanks a lot. I have removed it from the PR.

egs/wsj/s5/utils/data/combine_short_segments.sh
```diff
@@ -16,6 +16,7 @@
 
 # begin configuration section
 cleanup=true
+speaker_only=false
```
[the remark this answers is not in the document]
csltstu: Thanks a lot. I have improved the comment from the PR.

egs/wsj/s5/utils/data/internal/choose_utts_to_combine.py
```diff
@@ -216,12 +222,14 @@ def SelfTest():
 # This function figures out the grouping of utterances.
 # The input is:
 # 'min_duration' which is the minimum utterance length in seconds.
+# 'merge_within_speakers_only' which is the option to control if 
+# utterances were only combined from the same speaker.
```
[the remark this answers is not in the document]
csltstu: Thanks a lot. I have improved the comment from the PR.

egs/wsj/s5/utils/data/internal/choose_utts_to_combine.py
```diff
@@ -37,6 +37,12 @@
 
 parser.add_argument("--min-duration", type = float, default = 1.55,
                     help="Minimum utterance duration")
+parser.add_argument("--merge-within-speakers-only", type = str, default = 'false',
+                    choices = ['true', 'false'],
+                    help="If true, utterances were only combined from the same speaker."
+                    "It may be useful for the speaker recognition task."
+                    "If false, utterances were preferentially combined from the same speaker,"
+                    "and then combined across different speakers.")
```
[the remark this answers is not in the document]
csltstu: Thanks a lot. I have modified it from the PR.

david-ryan-snyder:
Thanks!

@danpovey this seems OK to merge in my opinion. You might want to first eyeball the changes to combine_short_segments.sh (https://github.com/kaldi-asr/kaldi/pull/3758/files#diff-4a1a2ae29145e1f58d85c67af9bfb05b ) and choose_utts_to_combine.py (https://github.com/kaldi-asr/kaldi/pull/3758/files#diff-7ab3032a776d9e4edf8cda30ee0a9625) before merging, since these changes potentially touch other recipes.

csltstu:
Thanks a lot !

PR #3760, opened 2019-12-05
Adding --file-skip option.  This is useful for debugging.

luitjens:
For example when you want to run a single file in a dataset you can
combine --file-skip=X --file-limit=Y to select the exact file for
processing without having to modify the underlying dataset.

Also reformatted some areas to match style guidelines

### DIFF
--- a/src/cudadecoderbin/batched-wav-nnet3-cuda.cc
+++ b/src/cudadecoderbin/batched-wav-nnet3-cuda.cc
@@ -122,6 +122,7 @@ int main(int argc, char *argv[]) {
 
     std::string word_syms_rxfilename;
 
+    int file_skip = 0;  // For debugging.  Let's us skip to a specific file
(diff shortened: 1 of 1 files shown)


danpovey:
I don't think I want to merge this as you could easily do it by modifying the scp files.  For instance, instead of `--file-skip=10` you could replace `scp:wav.scp` with `'scp:tail -n +10 wav.scp|'`

luitjens:
What if the input data is in a .ark and not a .scp?

danpovey:
It's unusual to have .wav files in an archive; do you ?
Normally they would come from an scp file in at least part of the pipeline.
Also i don't really want to merge things that are primarily for your debugging.

luitjens:
ok,  i'll revert the changes on our side and see if I can find a way to do it in the job scripts in the future.

PR #3761, opened 2019-12-05
Add nnet2 Chunking on GPU

btiplitz:
The nnet2 code for chunking needs to support the operation on a GPU


btiplitz:
@danpovey any thoughts on this change

src/nnet2bin/nnet-am-compute.cc
```diff
@@ -115,15 +115,15 @@ int main(int argc, char *argv[]) {
...
+      if (chunk_size > 0 && chunk_size < feats.NumRows()) {      
+        NnetComputationChunked(nnet, cu_feats, chunk_size, &cu_output);
       } else {
-        CuMatrix<BaseFloat> cu_feats(feats);
-        CuMatrix<BaseFloat> cu_output(output);
         NnetComputation(nnet, cu_feats, pad_input, &cu_output);
-        output.CopyFromMat(cu_output);
       }
-
+      output.CopyFromMat(cu_output);
```
danpovey: This is all fine, but please make this `output.Swap(&cu_output);`.  This will avoid actually copying the data if we are using CPU.

btiplitz: output is a Matrix where cu_output is a CuMatrix and might be on the gpu.  I would think using a swap would not work if the memory is on the gpu ?

danpovey:
No, Swap() does the right thing.

btiplitz:
@danpovey I update the code to swap

PR #3762, opened 2019-12-08
Fix minor bug in steps/data/reverberate_data_dir.py

HuangZiliAndy:
Hi Dan, I notice that there is some problems with this script. The original version cannot produce the exact result. You can reproduce the bug with /export/c10/hzili1/tools/kaldi_11_21/egs/sre16/v2/debug_reverb.sh. Basically the problem is that you cannot assume that after converting set to list the order is always the same.

Thanks,
Zili

### DIFF
--- a/egs/wsj/s5/steps/data/reverberate_data_dir.py
+++ b/egs/wsj/s5/steps/data/reverberate_data_dir.py
@@ -136,7 +136,10 @@ def pick_item_with_probability(x):
         collection (list or dictionary) where the values contain a field called probability
     """
     if isinstance(x, dict):
-        plist = list(set(x.values()))
+        keylist = list(x.keys())
+        keylist.sort()
+        random.shuffle(keylist)
+        plist = [x[k] for k in keylist]
     else:
         plist = x
     total_p = sum(item.probability for item in plist)


egs/wsj/s5/steps/data/reverberate_data_dir.py
```diff
@@ -136,7 +136,10 @@ def pick_item_with_probability(x):
         collection (list or dictionary) where the values contain a field called probability
     """
     if isinstance(x, dict):
-        plist = list(set(x.values()))
+        keylist = list(x.keys())
+        keylist.sort()
+        random.shuffle(keylist)
```
danpovey: Hm.  I don't understand why the order has to be deterministic, but doesn't random.shuffle give you different results each time?  I don't understand how this fixes the issue if nondeterministic order was the problem.

HuangZiliAndy: Since random seed is set in this script, so you don't need to worry about this.

HuangZiliAndy: I think for better reproducibility, the order has to be deterministic. Or you will mix an audio with a different noise file every time you run the experiment.

PR #3763, opened 2019-12-08
[egs] some fixes in Chime6 baseline system

aarora8:
- Fixing scoring (fixing number of deletions, outputting WER similar to compute-wer routine)
- adding a routine to select the best eval WER based on best lmwt and wip of development set.
- making diarization scoring similar to DIHARD II challenge
- removing test sets from ivectors script so that decoding and training can be independent
@sw005320


aarora8:
@danpovey, If you can get a chance, can you please review it. Thanks.

sw005320:
LGTM. @danpovey, it would be great if you review it soon for the merge so that we officially announce this recipe for the CHiME-6 challenge.

PR #3766, opened 2019-12-10
Added multithreaded version of chain-generic-numerator to remove CPU bottleneck

akshaysubr:
Multi-threading using C++11 threads to optimize the numerator graph's forward-backward algorithm to remove serial CPU bottleneck while training e2e chain models.

Added a command line option `--multithreaded-numerator` to toggle between single and multi-threaded execution. Tested and compared new implementation to the old implementation to ensure correctness. Overall speedup of 5-10x for just the forward-backward algorithm and 1.3-1.5x for a full training loop.

This PR also includes adding the ability to turn on mixed precision compute using tensor cores and added some NVTX markers in the code that are activated by a compile time flag `-DUSE_NVTX`.


src/chain/chain-generic-numerator.cc
```diff
@@ -221,35 +229,70 @@ bool GenericNumeratorComputation::ForwardBackward(
...
+  std::vector<std::thread> workers(nthreads);
+  for (int thread = 0; thread < nthreads; ++thread)
+    // Launch all threads
+    workers[thread] = std::thread(thread_lambda, thread, num_sequences, num_sequences_per_thread);
+  for (int thread = 0; thread < nthreads; ++thread) {
+    // Join threads back in
+    workers[thread].join();
+    // Reduce thread values to a single value
+    partial_loglike += partial_loglike_mt[thread];
+    ok = ok && ok_mt[thread];
```
danpovey: Does this actually create threads here, or uses some kind of thread pool?  Seems like creating threads might be kind of heavyweight.

src/chain/chain-training.cc
```diff
@@ -125,6 +126,7 @@ void ComputeChainObjfAndDerivE2e(const ChainTrainingOptions &opts,
   }
 
   if (xent_output_deriv != NULL) {
+    NVTX_RANGE("Resize");
```
danpovey: ?

src/chainbin/nnet3-chain-train.cc
```diff
@@ -53,6 +52,9 @@ int main(int argc, char *argv[]) {
                 "yes|no|optional|wait, only has effect if compiled with CUDA");
 
     opts.Register(&po);
+#if HAVE_CUDA==1
+    CuDevice::RegisterDeviceOptions(&po);
```
danpovey: If your changes require configuration to be effective, I think I would rather you set the defaults to be fast.

src/chain/chain-generic-numerator.cc
akshaysubr: Yes, threads are created within `ForwardBackward(...)` and the number of threads created is set to be the available hardware concurrency and the total number of sequences are split across these threads. There is an overhead to creating threads this way, but since the `GenericNumeratorComputation` object is created once per iteration and `ForwardBackward` is called only once during its lifetime, this design seemed reasonable.

src/chainbin/nnet3-chain-train.cc
akshaysubr: Okay, the multithreaded flag is set to `False` currently, will change that to `True` but do you also want the tensor core flag default to be `True` since this has an impact on numerical roundoff?

src/chain/chain-training.cc
akshaysubr: Remnant of debugging. Removed this.

danpovey:
Hm, it might need some debugging, if we set the tensor-core thing to true by default. 
I'm thinking the training might be affected or might be more likely to fail various 
assertions.
Tricky decision.
Does it cause a crash on older hardware if you set that flag to true?
@luitjens or @hugovbraun do you have any opinions on this?

hugovbraun:
It won't crash on older hardware (the tensor core flag will be ignored by cublas). However since it can possibly damage accuracy (the math is done in half precision), I think an opt-in would be better.

luitjens:
I agree with Hugo we should opt into tensor cores

src/chain/chain-generic-numerator.cc
jtrmal: I'm a bit worried about setting these things automatically -- hardware concurency is one thing and user limits placed on resources (e.g. in SGE/Slurm) are other things. This reminds me the OMP mentality that causes issues in many similar environments.
So I think it would be a nice thing to have but I don't think it should be default.

src/chainbin/nnet3-chain-train.cc
jtrmal: see my comment above

src/chain/chain-generic-numerator.cc
akshaysubr: So, make this also opt-in? Or set a better default value?

jtrmal:
Yes, that would be my preference (make it user configurable number of
threads).
Y

akshaysubr:
Changed the command line argument from a multithreading flag to a thread count. Setting thread count to 0 uses full hardware concurrency and the default is to use 1 thread.

danpovey:
I'm sorry, I seem to have neglected this PR.  (Reminder: please ping me if I drop the ball).
I am thinking that you could make the thread-count default to the minimum of 4, and the full hardware concurrency?  If that's possible?  The reason I think it's OK to default to >1 is that the numerator computation doesn't dominate anyway.

akshaysubr:
@danpovey I think that is reasonable. I've made the change.

src/chain/chain-denominator.cc
```diff
@@ -231,6 +234,7 @@ void DenominatorComputation::Beta(int32 t) {
 }
 
 BaseFloat DenominatorComputation::Forward() {
+  NVTX_RANGE(__func__);
```
danpovey: is this debug code?

akshaysubr: No, this is instrumentation for the profiler. The `NVTX_RANGE` macro is set when compiled with `-DUSE_NVTX` otherwise it's empty.

danpovey:
Great, thanks!  Merging.

PR #3768, opened 2019-12-11
shrink channel vectors instead of vector holding those vectors

luitjens:


luitjens:
Dan any chance we can get this merged?

PR #3769, opened 2019-12-11
Fix bug in NumPendingTasks() 

luitjens:
NumPendingTasks didn't consider padding giving incorrect numbers in some corner cases

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.h
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.h
@@ -198,8 +198,8 @@ public:
  // Check if any group is available. If one is available, set its name in *group
  bool IsAnyGroupCompleted(std::string *group);
  inline int NumPendingTasks() {
-   return (tasks_back_ - tasks_front_ + config_.max_pending_tasks + 1) %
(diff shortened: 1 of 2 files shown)


src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.h
```diff
@@ -198,8 +198,8 @@ class BatchedThreadedNnet3CudaPipeline {
  // Check if any group is available. If one is available, set its name in *group
  bool IsAnyGroupCompleted(std::string *group);
  inline int NumPendingTasks() {
-   return (tasks_back_ - tasks_front_ + config_.max_pending_tasks + 1) %
-          (config_.max_pending_tasks + 1);
+   return (tasks_back_ - tasks_front_ + config_.max_pending_tasks + 10) %
```
danpovey: Wherever a constant like this appears in the code I like to have an explanation of where it came from.  E.g. does this need to be in sync with some other number elsewhere in the code?

luitjens: removed the magic constant and put it into the config.  I didn't add a command line option as I don't think it is something users should be changing but changing via an app might be useful for testing.

PR #3771, opened 2019-12-16
Upgrade TensorFlow RNN to 2.0

scottcjt:
* Some adjustments to CMakeLists

Some adjustments are made to make it easier to compile CPU-only build on MacOS, including disabling CUDA and using Apple BLAS by default. Also integrates *CTest* to auto collect test binaries - `make test` or `ctest` to run them.

* Upgrade TensorFlow to v2.0

Upgrade TensorFlow RNNLM (training & rescore) code to use v2.0. I tried to use as many v2-flavor code to better integrate with other (possible) tf-v2 code in the future.

Some detailed use of this model in C++ is not quite clear to me, but I don't have wsj dataset so I cannot test it end-to-end.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,8 +6,16 @@ include(GNUInstallDirs)
 include(Utils)
 include(third_party/get_third_party)
 
+# Should update cmake to a more recent version which supports FindPython3.
+find_package(PythonInterp)
+if(NOT PYTHON_EXECUTABLE OR PYTHON_VERSION_MAJOR LESS 3)
+    message(WARNING "Needs python3 to auto-generate most CMake files, but not found. "
+                    "Will try `python3` directly...")
+    set(PYTHON_EXECUTABLE "python3")
+endif()
+
 message(STATUS "Running gen_cmake_skeleton.py")
-execute_process(COMMAND python
+execute_process(COMMAND ${PYTHON_EXECUTABLE}
     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/gen_cmake_skeleton.py"
     "${CMAKE_CURRENT_SOURCE_DIR}/src"
     "--quiet"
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -6,8 +6,16 @@ include(GNUInstallDirs)
 include(Utils)
 include(third_party/get_third_party)
 
+# Should update cmake to a more recent version which supports FindPython3.
```
danpovey: The CMake build system is optional.  Should way say that it's required when using TensorFlow?

scottcjt: No, TensorFlow only officially supports Bazel. We are free to use whatever we want.

So far I choose CMake because it is easier on my macbook - where many build tools or libraries on Linux are absent here.

danpovey: I just meant when using TensorFlow with Kaldi.  It's OK, not everyone will use this, so being limited to the CMake build setup is OK.

scottcjt:
@danpovey Do you think there is some other stuff to be done? I considered the pure RNN script is totally covered by the LSTM one, so I will leave it there.

danpovey:
Thanks a lot!  Since you think it's ready to merge  I'll merge, since I'm sure it was an improvement on the previous status.  If any issues arise we'll address them then.  Thanks!!

behnamasefi:
Hi All,
 
I see that we install `tensorflow_gpu-1.2.0` in `install_tensorflow_py.sh` script. My understanding from this PR is that the current code is compatible with TF2. Would anyone know if we install the latest `tensorflow_gpu-2.1.0` we can still use the current scripts and codes to train a RNNLM and rescore lattices? 

Thanks,

danpovey:
It looks to me like currently the TensorFlow part will only work if you use
the CMake build.

behnamasefi:
Thanks a lot @danpovey . But I guess it still uses `tensorflow_gpu-1.2.0` and just wondering if someone has tried the latest Tensorflow version with the current code. @scottcjt might knows more?
cheers,

danpovey:
No it doesn't, it uses a more recent version, 2.0 I think.  Check in cmake/.

scottcjt:
They are tested against tensorflow-2.0 (use CMake instead). 2.1.0 should be
fine.

Note that if you are training language model, there is no need to build
tensorflow.
For the Python training code, just use pip install to try out.


behnamasefi <notifications@github.com> 於 2020年2月18日 週二 下午5:09寫道：

PR #3780, opened 2019-12-20
fix tools/Makefile about wget

songmeixu:
If first url fail, next url will output with wrong filename (`openfst-1.6.7.tar.gz.1` e.g.):

image

Modification: keep the normal name by `-O`

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -151,7 +151,7 @@ cub:
 	if [ -d "$(DOWNLOAD_DIR)" ]; then \
 		cp -p "$(DOWNLOAD_DIR)/cub-$(CUB_VERSION).zip" .; \
 	else \
-		$(WGET) -T 10 -t 3 -c https://github.com/NVlabs/cub/archive/$(CUB_VERSION).zip; \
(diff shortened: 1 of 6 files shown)


tools/Makefile
```diff
@@ -98,7 +98,7 @@ openfst-$(OPENFST_VERSION).tar.gz:
 		cp -p "$(DOWNLOAD_DIR)/openfst-$(OPENFST_VERSION).tar.gz" .; \
 	else \
 		$(WGET) -T 10 -t 1 http://www.openfst.org/twiki/pub/FST/FstDownload/openfst-$(OPENFST_VERSION).tar.gz || \
-		$(WGET) -T 10 -t 3 https://www.openslr.org/resources/2/openfst-$(OPENFST_VERSION).tar.gz; \
+		$(WGET) -T 10 -t 3 -O openfst-$(OPENFST_VERSION).tar.gz https://www.openslr.org/resources/2/openfst-$(OPENFST_VERSION).tar.gz; \
```
danpovey: Hm. I think an easier and better fix would be to add the --continue flag.  You can add it to all of those wget commands.

songmeixu: It's a better idea, and i change those case math regex `\|\|\s*\\*\s*(wget|\$\(WGET\))`. I think it's all of that.

danpovey:
OK I'll merge, if there are more fixes I'll merge them later.

PR #3782, opened 2019-12-21
bugs of using u0* type dir

hangtingchen:
For beamformit, the u0* type input dir can not be parsed `utils/make_absolute.sh`, just use the input audio dir is ok

### DIFF
--- a/egs/chime6/s5_track1/local/prepare_data.sh
+++ b/egs/chime6/s5_track1/local/prepare_data.sh
@@ -22,7 +22,11 @@ fi
 
 set -e -o pipefail
 
-adir=$(utils/make_absolute.sh $1)
+if [ $mictype != "ref" ];then
+  adir=$(utils/make_absolute.sh $1)
+else
+  adir=$1
+fi
 jdir=$2
 dir=$3
 


danpovey:
Can you explain the issue in more detail?  Could be an issue that should be fixed in make_absolute.sh.

hangtingchen:
Hi,
For example,
```shell
cd egs/wsj/s5
for i in `seq 1 3`;do 
  mkdir -p debug_make_absolute/u0$i
done
utils/make_absolute.sh debug_make_absolute/u0*
```
It returns like this `***/egs/wsj/s5/debug_make_absolute/u01`. The expected result is `***/egs/wsj/s5/debug_make_absolute/u0*`

danpovey:
OK, it's behaving as expected because bash does the expansion of that expression with star, so make_absolute.sh never even sees it.  Also make_absolute.sh operates on directories that exist, not meta-expressions.
I'll have to wait for someone in the chime team to comment before merging.

danpovey:
@sw005320 can you please comment or ask someone to comment?

sw005320:
@aarora8, could you comment on it for us?

aarora8:
Thanks, yeah with the above modification the local/decode.sh script is now working well with both beamformit and gss enhancement.

kkm000:
@aarora8, so should this be merged?

aarora8:
Sorry, I just noticed. Yeah it can be merged.

PR #3788, opened 2019-12-24
Add support to zero-copy Kaldi CuVector/CuMatrix to PyTorch's GPU tensor

csukuangfj:
Reopend from https://github.com/danpovey/kaldi/pull/93

### DIFF
--- a/src/pybind/cudamatrix/cu_vector_pybind.cc
+++ b/src/pybind/cudamatrix/cu_vector_pybind.cc
@@ -36,11 +36,7 @@ void pybind_cu_vector(py::module& m) {
         .def("Set", &PyClass::Set, py::arg("value"))
         .def("Add", &PyClass::Add, py::arg("value"))
         .def("Scale", &PyClass::Scale, py::arg("value"))
-        .def("__getitem__", [](const PyClass& v, int i) { return v(i); })
-        .def("to_dlpack", [](PyClass* v) {
-          // we use the name `to_dlpack` because PyTorch uses the same name
-          return CuVectorToDLPack(v);
-        });
+        .def("__getitem__", [](const PyClass& v, int i) { return v(i); });
   }
   {
     using PyClass = CuVector<float>;
@@ -48,7 +44,11 @@ void pybind_cu_vector(py::module& m) {
         .def(py::init<>())
         .def(py::init<MatrixIndexT, MatrixResizeType>(), py::arg("dim"),
              py::arg("MatrixResizeType") = kSetZero)
-        .def(py::init<const VectorBase<float>&>(), py::arg("v"));
+        .def(py::init<const VectorBase<float>&>(), py::arg("v"))
+        .def("to_dlpack", [](PyClass* v) {
+          // we use the name `to_dlpack` because PyTorch uses the same name
+          return CuVectorToDLPack(v);
+        });
     // TODO(fangjun): wrap other methods when needed
   }
   {
--- a/src/pybind/cudamatrix/cu_vector_pybind.cc
+++ b/src/pybind/cudamatrix/cu_vector_pybind.cc
@@ -37,50 +37,52 @@ void pybind_cu_vector(py::module& m) {
         .def("Add", &PyClass::Add, py::arg("value"))
         .def("Scale", &PyClass::Scale, py::arg("value"))
         .def("__getitem__", [](const PyClass& v, int i) { return v(i); })
+        .def("to_dlpack", [](PyClass* v) {
 #if HAVE_CUDA == 1
-        .def("to_dlpack",
-             [](PyClass* v) {
-               // we use the name `to_dlpack` because PyTorch uses the same name
+          // we use the name `to_dlpack` because PyTorch uses the same name
 
-               // the created `managed_tensor` will be freed in
-               // `DLManagedTensorDeleter`, so no memory leak here.
-               auto* managed_tensor = new DLManagedTensor();
-               managed_tensor->manager_ctx = nullptr;
+          // the created `managed_tensor` will be freed in
+          // `DLManagedTensorDeleter`, which does not free `data`,
+          // so no memory leak here
+          auto* managed_tensor = new DLManagedTensor();
+          managed_tensor->manager_ctx = nullptr;
 
-               // setup the deleter to free allocated memory.
-               // refer to
-               // https://github.com/pytorch/pytorch/blob/master/torch/csrc/Module.cpp#L361
-               // for how and when the deleter is invoked.
-               managed_tensor->deleter = &DLManagedTensorDeleter;
+          // setup the deleter to free allocated memory.
+          // refer to
+          // https://github.com/pytorch/pytorch/blob/master/torch/csrc/Module.cpp#L361
+          // for how and when the deleter is invoked.
+          managed_tensor->deleter = &DLManagedTensorDeleter;
 
-               auto* tensor = &managed_tensor->dl_tensor;
-               tensor->data = v->Data();
-               tensor->ctx.device_type = kDLGPU;
-               tensor->ctx.device_id = CuDevice::GetCurrentDeviceId();
+          auto* tensor = &managed_tensor->dl_tensor;
+          tensor->data = v->Data();
+          tensor->ctx.device_type = kDLGPU;
+          tensor->ctx.device_id = CuDevice::GetCurrentDeviceId();
 
-               tensor->ndim = 1;
+          tensor->ndim = 1;
 
-               tensor->dtype.code = kDLFloat;
-               tensor->dtype.bits = 32;  // single precision float
-               tensor->dtype.lanes = 1;
+          tensor->dtype.code = kDLFloat;
+          tensor->dtype.bits = 32;  // single precision float
+          tensor->dtype.lanes = 1;
 
-               // `shape` and `strides` are freed in `DLManagedTensorDeleter`,
-               // so no memory leak here.
-               tensor->shape = new int64_t[1];
-               tensor->shape[0] = v->Dim();
+          // `shape` and `strides` are freed in `DLManagedTensorDeleter`,
+          // so no memory leak here.
+          tensor->shape = new int64_t[1];
+          tensor->shape[0] = v->Dim();
 
-               tensor->strides = new int64_t[1];
-               tensor->strides[0] = 1;
-               tensor->byte_offset = 0;
+          tensor->strides = new int64_t[1];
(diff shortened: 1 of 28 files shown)


csukuangfj:
I agree.

Fixed.

danpovey:
Thanks.  How have you tested it?  Have you tested that it deletes the object appropriately?
(And that it doesn't delete it if a reference to the matrix exists elsewhere)?

csukuangfj:
They are two kinds of deletion here.

One is for the DLManagedTensor. It is wrapped in a PyCapsule and its lifetime depends on
the PyCapsule. 

If the PyCapsule has not been consumed when it goes out of scope, then I
tested that the deleter for `DLManagedTensor` is called.

If the PyCapsule is consumed by PyTorch by 
```python
tensor = from_dlpack(v.to_dlpack())
del tensor
```
`del tensor` will invoke `DLManagedTensor`'s deleter. This is also been tested.

Note that the deleter of `DLManagedTensor` has nothing to do with `Vector`/`Matrix`;
it frees only its `strides`,`shapes` and itself.

The other one is for `DLPackVector`,  since I use `new` to create it and return it
to Python with `py::return_value_reference::take_ownership`.

I have tested that

```
tensor = torch.arange(2).float()
v = kaldi.DLPackFloatSubVector.from_dlpack(to_dlpack(tensor))

del v
```

`del v` will invoke the destructor of `DLPackFloatSubVector`. This has been tested.
If I do not use `del v`, when `v` goes out of scope, its destructor is also invoked. This has
also been tested.

src/pybind/cudamatrix/cu_vector_pybind.cc
```diff
@@ -34,11 +35,30 @@ void pybind_cu_vector(py::module& m) {
...
+  {
+    using PyClass = CuVector<float>;
+    py::class_<PyClass, CuVectorBase<float>>(m, "FloatCuVector")
+        .def(py::init<>())
+        .def(py::init<MatrixIndexT, MatrixResizeType>(), py::arg("dim"),
+             py::arg("MatrixResizeType") = kSetZero)
+        .def(py::init<const VectorBase<float>&>(), py::arg("v"))
+        .def("to_dlpack", [](PyClass* v) {
+          // we use the name `to_dlpack` because PyTorch uses the same name
+          return CuVectorToDLPack(v);
```
danpovey: What I was thinking here was that we'd have this DLManagedTensor do an incref on `v` when it is created, and a decref on `v` when it is deleted.  Of course, I was thinking we'd do this for just Vector, and not VectorBase.  The idea is that the Vector itself is the owner of the memory it owns, and deleting the Vector itself is the only valid way of freeing that memory.   The thing I am now sure, about, though, is how we get access to the Python object that we are part of here.  I don't know the invocation.

danpovey: .. of course the other option would be to have the creation of the DLManagedTensor be destructive of the original Vector or Matrix, and to have the DLManagedTensor itself own the data pointer.  But I think it's preferable to have it non-destructive, if we can find a way to incref the python object.  I imagine that these pybind-wrapped classes must somehow 'live inside' a Python object.

csukuangfj: That is easy.

The returned object is a `py::capsule` (https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L1145)

which is a subclass of `py::handle` (https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L174)

which has `inc_ref` method (https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L190)

The question is why we need to to this in Python ?

In C++, `SubVector` borrows a pointer from an outside object, it is the callers
respponsibility to keep the outside object alive while the `SubVector` is still in use.

In Python, the caller has to ensure the `Matrix` is alive if he/she calls 
`dlpack = matrix.to_dlpack()`.

csukuangfj: Here is the PyTorch's implementation for `to_dlpack`

https://github.com/pytorch/pytorch/blob/1dfb8752a62de49b1d270aa0c84d05aee07e51db/torch/csrc/Module.cpp#L370
```cpp
PyObject *THPModule_toDLPack(PyObject *_unused, PyObject *data)
{
  HANDLE_TH_ERRORS
  THPUtils_assert(THPVariable_Check(data), "data must be a Tensor");
  DLManagedTensor* dlMTensor = at::toDLPack(THPVariable_Unpack(data));
  return PyCapsule_New(dlMTensor, "dltensor", DLPack_Capsule_Destructor);
  END_HANDLE_TH_ERRORS
}
```

https://github.com/pytorch/pytorch/blob/1dfb8752a62de49b1d270aa0c84d05aee07e51db/aten/src/ATen/DLConvertor.cpp#L162
```cpp
// This function returns a shared_ptr to memory managed DLpack tensor
// constructed out of ATen tensor
DLManagedTensor* toDLPack(const Tensor& src) {
  ATenDLMTensor* atDLMTensor(new ATenDLMTensor);
  atDLMTensor->handle = src;
  atDLMTensor->tensor.manager_ctx = atDLMTensor;
  atDLMTensor->tensor.deleter = &deleter;
  atDLMTensor->tensor.dl_tensor.data = src.data_ptr();
  int64_t device_id = 0;
  if (src.is_cuda()) {
    device_id = src.get_device();
  }
  atDLMTensor->tensor.dl_tensor.ctx = getDLContext(src, device_id);
  atDLMTensor->tensor.dl_tensor.ndim = src.dim();
  atDLMTensor->tensor.dl_tensor.dtype = getDLDataType(src);
  atDLMTensor->tensor.dl_tensor.shape =
      const_cast<int64_t*>(src.sizes().data());
  atDLMTensor->tensor.dl_tensor.strides =
      const_cast<int64_t*>(src.strides().data());
  atDLMTensor->tensor.dl_tensor.byte_offset = 0;
  return &(atDLMTensor->tensor);
}
```

https://github.com/pytorch/pytorch/blob/1dfb8752a62de49b1d270aa0c84d05aee07e51db/aten/src/ATen/DLConvertor.cpp#L151
```cpp
struct ATenDLMTensor {
  Tensor handle;
  DLManagedTensor tensor;
};
```

It does not increase the reference count of the underlying tensor.
The caller has to ensure it is alive when the PyCapsule is in use.

danpovey:
Yeah that's true about in C++, but I felt it would be more Pythonic to have
the memory management happen automatically, because Python programmers are
not accustomed to worry about memory.
What I was thinking was that we can automatically keep the Matrix alive by
having the DLManagedTensor object do an incref on the Python object holding
that Matrix when it is created, and decref when it is destroyed.
From here:
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html
it says you can do:
```
MyClass *cls = ..;
py::object obj = py::cast(cls)
```
In fact, from
https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv46object
it looks like simply  having the DLManagedTensor have a py::object as a
member may do the incref and decref automatically.

src/pybind/cudamatrix/cu_vector_pybind.cc
danpovey: I think the thing that's different here is that class Tensor internally has some kind of shared_ptr.  (I've looked into PyTorch code and I think this is true, but it may be renamed to some kind of PyTorch-specific shared_ptr).
So the Tensor itself doesn't have to be kept alive.  Kaldi's memory management is a bit older and simpler, the Matrix itself has to be kept alive if you want the memory to not be deleted, so the only way to ensure this is to incref that Matrix.

csukuangfj:
I agree with you.

From Bjarne Stroustrup
> C makes it easy to shoot yourself in the foot; C++ makes it harder, but when you do it blows your whole leg off.

Increase the reference count for the Python users 
can make it harder for them to make a mistake.

The question is how do we let the DLManagedTensor know where its data come from.
We have to keep a pointer to the source object.

I am afraid we cannot modify the structure definition of `DLManagedTensor`, neither
add nor remove a field.

csukuangfj:
I think I find a way to solve the probleme from `CuPy`.

danpovey:
We could have the manager_ctx void* pointer just contain the PyObject*
pointer to that PyObject that owned the Matrix, and the deleter do a decref
manually.

csukuangfj:
Here is the `CuPy`'s implementation for `to_dlpack`

https://github.com/cupy/cupy/blob/master/cupy/core/dlpack.pyx#L81
```python
cpdef object toDlpack(ndarray array) except +:
    cdef DLManagedTensor* dlm_tensor = \
        <DLManagedTensor*>stdlib.malloc(sizeof(DLManagedTensor))
...

at line https://github.com/cupy/cupy/blob/master/cupy/core/dlpack.pyx#L117
    dlm_tensor.manager_ctx = <void *>array
    cpython.Py_INCREF(array)
    dlm_tensor.deleter = deleter
```

And inside the `deleter` https://github.com/cupy/cupy/blob/master/cupy/core/dlpack.pyx#L70
```python
cdef void deleter(DLManagedTensor* tensor) with gil:
    if tensor.manager_ctx is NULL:
        return
    stdlib.free(tensor.dl_tensor.shape)
    cpython.Py_DECREF(<ndarray>tensor.manager_ctx)
    stdlib.free(tensor)
    tensor.manager_ctx = NULL
```
We can save the object pointer to  the field`manager_ctx`

csukuangfj:
`CuPy` did as you suggested.

csukuangfj:
I will fix it tomorrow.

danpovey:
Thanks.  Do you know which C++ version introduced the `<some_type_you_are_casting_to>` version of casting?  I haven't seen that syntax before.  I *assume* it's a cast.

csukuangfj:
what is `<some_type_you_are_casting_to>`?? Is it one of the following:

- static_cast
- dynamic_cast
- const_cast
- reinterpret_cast,

I am aware of the above 4 casts in C++.

danpovey:
oh sorry, I thought I was looking at C++ code, that was pyx code.  Thought it was some new syntax; it changes every five minutes.

csukuangfj:
I agree.
 
C++ is moving very fast since C++11.

C++20 is a brand new C++ to old C++98/03 programmers.

csukuangfj:
Now if `Vector` or `Matrix` invokes `to_dlpack`, its reference count
is increased by one and you can safely run `del vector` or `del matrix`
without worrying about segmentation fault.

danpovey:
Fantastic, thanks!! Merging.

PR #3789, opened 2019-12-25
Pybind chain training opts

### LINKED ISSUE #3784: wrap ChainTraining options
to kaldi10

naxingyu:
Not sure if this is what you asked for @danpovey 
This closes #3784

### DIFF
--- a/src/pybind/chain/chain_training_pybind.cc
+++ b/src/pybind/chain/chain_training_pybind.cc
@@ -1,0 +1,51 @@
+// pybind/chain/chain_training_pybind.cc
+
+// Copyright 2019   Microsoft Corporation (author: Xingyu Na)
+
+// See ../../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "chain/chain_pybind.h"
+
+#include "chain/chain-training.h"
+
+using namespace kaldi::chain;
+
+void pybind_chain_training(py::module& m) {
+  py::class_<ChainTrainingOptions>(m, "ChainTrainingOptions")
+    .def(py::init<>())
+    .def_readwrite("l2_regularize", &ChainTrainingOptions::l2_regularize,
+                   "l2 regularization constant on the 'chain' output; the actual term added to"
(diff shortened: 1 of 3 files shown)


src/pybind/chain/chain_training_pybind.cc
```diff
@@ -0,0 +1,51 @@
...
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "chain/chain_pybind.h"
```
danpovey: Guys, is there a good reason why we are using underscores rather than dashes in these filenames?  Doesn't seem very consistent.  The dashes were for google style guide reasons I think.

danpovey:
Looks good, thanks!

src/pybind/chain/chain_training_pybind.cc
csukuangfj: I guess Dongji used this style first, for example, `kaldi_pybind.cc` instead of `kaldi-pybind.cc`.

I followed `kaldi_pybind.cc`.

danpovey:
OK.  There may well be a good reason for it, e.g. to do with CMake or with
Python module names.   I'll wait will we hear from others.

csukuangfj:
So should we follow `kaldi_pybind.cc` or change it and other files to use `kaldi-pybind.cc` style ?

danpovey:
Maybe someone could experiment with changing them to "-" and see if the
build and test procedure still works.  There may be some reason why it
wouldn't.

src/pybind/chain/chain_training_pybind.cc
naxingyu: Yep, I followed `kaldi_pybind.cc` too. I think we can switch at certain point later.

csukuangfj:
To the best of my knowledge, there won't be any problems if you change `kaldi_pybind.cc`
to `kaldi-pybind.cc`.

naxingyu:
Maybe it was for distinguishing pybind subfolders from others?

csukuangfj:
every file inside `pybind` is ended with `_pybind.h` or `_pybind.cc`,
so I think this will be no confusing if you switch from `_` to `-`

naxingyu:
I just tried changing `kaldi_pybind.cc` to `kaldi-pybind.h` while the name of the module is still `kaldi_pybind`. Nothing breaks.

naxingyu:
Only the file names should be changed, right? The module and function names seems compliant. @danpovey

csukuangfj:
and the first line in every file should match its filename.
and the Makefile should be changed accordingly.

naxingyu:
Of source. I mean everything related to the source file names, not the
functions, members, module, etc.

csukuangfj:
Yes, I think so.

danpovey:
Guys, I haven't *completely* decided on this.  If someone comes up with a
good argument why it should stay the same, I'm OK with that.  Just wanted
to see if it would break stuff. It's odd to have different styles in the
code.
CMake might have opinions about this stuff too (?)

PR #3793, opened 2019-12-25
Add LF-MMI training pipeline in PyTorch with kaldi pybind

csukuangfj:
`run.sh` should be working now. 

It gives you a reasonable WER/CER if you execute `run.sh` from stage==0 to the end.
The following is the result from a sample run
```
steps/scoring/score_kaldi_cer.sh --stage 2 --cmd run.pl data/fbank_pitch/dev exp/chain/graph exp/chain/decode_res/dev
steps/scoring/score_kaldi_cer.sh: scoring with word insertion penalty=0.0,0.5,1.0
+ echo './local/score.sh: Done'
./local/score.sh: Done
==> exp/chain/decode_res/test/scoring_kaldi/best_cer <==
%WER 10.16 [ 10640 / 104765, 660 ins, 700 del, 9280 sub ] exp/chain/decode_res/test/cer_10_1.0

==> exp/chain/decode_res/test/scoring_kaldi/best_wer <==
%WER 19.13 [ 12324 / 64428, 1169 ins, 1896 del, 9259 sub ] exp/chain/decode_res/test/wer_12_0.5
==> exp/chain/decode_res/dev/scoring_kaldi/best_cer <==
%WER 7.93 [ 16289 / 205341, 692 ins, 1007 del, 14590 sub ] exp/chain/decode_res/dev/cer_10_1.0

==> exp/chain/decode_res/dev/scoring_kaldi/best_wer <==
%WER 16.29 [ 20806 / 127698, 2125 ins, 2876 del, 15805 sub ] exp/chain/decode_res/dev/wer_11_0.0
```

Only the neural network part is based on PyTorch, everything else is in Kaldi.
It can be used as a starting point and you can try different fancy networks in PyTorch
without the need to change Kaldi.

Furthermore, kaldi pybind has no C++ source-level dependencies on PyTorch and you can
choose any version of PyTorch. We have tested the pipeline with
```
>>> import torch
>>> torch.__version__
'1.3.0.dev20191006'
```


naxingyu:
aishell2 is not free for anyone. It is better to create an example from free, and smaller data.

csukuangfj:
@naxingyu 
Is aishell free ?

naxingyu:
Yes. Or starting from the toy data mini_librispeech might be more
convenient for debugging.

csukuangfj:
I have more experience with `aishell` and I have already had a baseline with it using Kaldi LF-MMI.
Once the training with `aishell` is OK, I will switch to `mini_librispeech`.

I will switch to `aishell` now.

egs/aishell/s10/chain/model.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Create a network like the above one
+class ChainModel(nn.Module):
+
+    def __init__(feat_dim,
+                 output_dim,
+                 lda_mat_filename,
+                 hidden_dim=625,
+                 kernel_size_list=[1, 3, 3, 3, 3, 3],
+                 dilation_list=[1, 1, 1, 3, 3, 3],
```
danpovey: You shouldn't be using the dilation here, there should be a subsamplng via a factor of 3 in the middle of the network; can be done using the 'stride' of conv1d.  That is what Kaldi does too, but it's implicit via dependency tracking.

egs/aishell/s10/chain/model.py
```diff
@@ -0,0 +1,148 @@
...
+        self.batch_norms = nn.ModuleList([
+            nn.BatchNorm1d(num_features=hidden_dim) for i in range(num_layers)
+        ])
+
+        stride = frame_subsampling_factor
+
+        self.prefinal_chain_tdnn = nn.Conv1d(in_channels=hidden_dim,
+                                             out_channels=hidden_dim,
+                                             kernel_size=1,
+                                             stride=stride)
```
csukuangfj: @danpovey 
the subsampling is here.

This kind of layer
```
relu-batchnorm-layer name=tdnn4 input=Append(-3,0,3) dim=625
```
needs a dilation of `3`.

egs/aishell/s10/chain/model.py
csukuangfj: See the comment below.

We have to use dilation here.

danpovey:
No, you don't need the dilation.  The kernels are not really dilated, they
are normal 3x1 kernels, it's due to differences in how the frameworks work.

csukuangfj:
for kernel_size == 3, dilation == 1, at time `t`, it takes inputs [t - 1, t, t + 1]

for kernel_szie == 3, dilation == 3, at time `t`, it takes inputs [t - 3, t, t + 3]

I do not understand why we do NOT need dilation here.

csukuangfj:
```
  relu-batchnorm-layer name=tdnn1 dim=625
  relu-batchnorm-layer name=tdnn2 input=Append(-1,0,1) dim=625
  relu-batchnorm-layer name=tdnn4 input=Append(-3,0,3) dim=625
```
Do the above three layers have the same dilation and kernel size ?

csukuangfj:
The number of input frames is `model_left_context + model_right_context + seq_len`

and the number of expected output frames is `50`, for my specific model.
My approach, i.e., with dilation, gives me exactly 50 frames in the output.

If I change dilation to 1, there will be problems for `nnet_output` in the chain loss function.

danpovey:
I'm 100% sure.
There must be something wrong you are doing there, e.g. not taking into
account the frame subsampling factor.
It's a dependency graph on time indexes where it works out what needs to be
computed.

csukuangfj:
First version of LF-MMI in PyTorch with kaldi pybind.

It seems working. The following is the log output from this statement
https://github.com/mobvoi/kaldi/blob/fangjun-pybind11/egs/aishell/s10/chain/train.py#L73

```python
        if batch_idx % 10 == 0:
            print('Process {}/{}({:.6f}%) global average objf: {:.6f} over {} \
                    frames, current batch average objf: {:.6f} over {} frames, epoch {}'
                  .format(
                      batch_idx, len(dataloader),
                      float(batch_idx) / len(dataloader),
                      total_objf / total_weight, total_frames,
                      float(objf_l2_term_weight[0].item()) /
                      objf_l2_term_weight[2].item(), num_frames, current_epoch))
```

```
LOG ([5.5.15~3-94484]:SelectGpuDevice():cu-device.cc:172) Select gpu id 6 out of 7 (starting from 0)
WARNING ([5.5.15~3-94484]:SelectGpuDevice():cu-device.cc:187) Not in compute-exclusive mode.  Suggestion: use 'nvidia-smi -c 3' to set compute exclusi
ve mode
LOG ([5.5.15~3-94484]:FinalizeActiveGpu():cu-device.cc:357) The active GPU is [6]: TITAN Xp     free:11555M, used:641M, total:12196M, free/total:0.947
442 version 6.1
epoch 0, learning rate 0.001
Process 0/3161(0.000000%) global average objf: -1.066138 over 6400.0  frames, current batch average objf: -1.066138 over 6400 frames, epoch 0
Process 10/3161(0.003164%) global average objf: -1.047951 over 63616.0  frames, current batch average objf: -1.020809 over 6400 frames, epoch 0
Process 20/3161(0.006327%) global average objf: -1.017514 over 119168.0 frames, current batch average objf: -0.912189 over 3840 frames, epoch 0
Process 30/3161(0.009491%) global average objf: -0.992124 over 176384.0  frames, current batch average objf: -0.939771 over 6400 frames, epoch 0
Process 40/3161(0.012654%) global average objf: -0.976169 over 231936.0  frames, current batch average objf: -0.903288 over 3840 frames, epoch 0
Process 50/3161(0.015818%) global average objf: -0.965672 over 290048.0  frames, current batch average objf: -0.927077 over 6400 frames, epoch 0
Process 60/3161(0.018981%) global average objf: -0.958138 over 347264.0  frames, current batch average objf: -0.929169 over 6400 frames, epoch 0
Process 70/3161(0.022145%) global average objf: -0.950811 over 402816.0 frames, current batch average objf: -0.955712 over 6400 frames, epoch 0
Process 80/3161(0.025308%) global average objf: -0.945554 over 458368.0  frames, current batch average objf: -0.916387 over 6400 frames, epoch 0
Process 90/3161(0.028472%) global average objf: -0.941090 over 513920.0  frames, current batch average objf: -0.944927 over 4736 frames, epoch 0
Process 100/3161(0.031636%) global average objf: -0.938691 over 571136.0  frames, current batch average objf: -0.852855 over 3840 frames, epoch 0
Process 110/3161(0.034799%) global average objf: -0.937187 over 629248.0  frames, current batch average objf: -0.909436 over 6400 frames, epoch 0
```

You can see that the objf is increasing!

TODOS:
- stride == 3 will be moved to the last layer with (-1, 0, 1)
- Support for save/load checkpoint will be added
- Support for inference only mode will be added
- Support for tensorboard to visualize training loss

egs/aishell/s10/chain/model.py
csukuangfj: now dilation is repaced with stride. thanks

danpovey:
Thanks!  I'm too busy to look at this in detail but I'll merge as-is so you guys can continue to make progress.

csukuangfj:
Now it is able to do training, inference, decoding, and scoring.

Currently, it assumes that all required files are pre-generated, i.e., HCLG.fst, words.txt, cegs.*.scp, etc.

I will add scripts to generate the required files as soon as possible.

Part of the training log is as follows (you can find it in the attached file)

```
2019-12-27 17:20:34,235 INFO [train.py:109] ./chain/train.py --device-id 7 --dir /data/fangjunkuang/open-source/dan/kaldi/egs/aishell/s10/exp --feat-dim 43 --hidden-dim 625 --checkpoint= --is-training 1 --kernel-size-list 1, 3, 3, 3, 3, 3 --lda-mat-filename /cache/fangjun/chain/model/lda.mat --log-level info --output-dim 4464 --stride-list 1, 1, 3, 1, 1, 1 --train.cegs-dir /cache/fangjun/chain/merged_egs --train.den-fst /cache/fangjun/chain/model/den.fst --train.egs-left-context 13 --train.egs-right-context 13 --train.l2-regularize 5e-4 --train.lr 1e-3 --train.num-epochs 6
2019-12-27 17:20:40,332 INFO [train.py:174] epoch 0, learning rate 0.001
2019-12-27 17:20:40,574 INFO [train.py:92] Process 0/3161(0.000000%) global average objf: -1.233406 over 6400.0 frames, current batch average objf: -1.233406 over 6400 frames, epoch 0
2019-12-27 17:20:57,111 INFO [train.py:92] Process 100/3161(3.163556%) global average objf: -0.546051 over 571136.0 frames, current batch average objf: -0.358299 over 3840 frames, epoch 0
2019-12-27 17:21:13,781 INFO [train.py:92] Process 200/3161(6.327112%) global average objf: -0.421692 over 1138432.0 frames, current batch average objf: -0.227188 over 6400 frames, epoch 0
2019-12-27 17:21:30,091 INFO [train.py:92] Process 300/3161(9.490668%) global average objf: -0.359247 over 1701504.0 frames, current batch average objf: -0.210956 over 4736 frames, epoch 0
2019-12-27 17:21:46,525 INFO [train.py:92] Process 400/3161(12.654223%) global average objf: -0.319904 over 2268800.0 frames, current batch average objf: -0.192609 over 6400 frames, epoch 0
2019-12-27 17:22:03,253 INFO [train.py:92] Process 500/3161(15.817779%) global average objf: -0.292344 over 2836096.0 frames, current batch average objf: -0.153050 over 6400 frames, epoch 0
2019-12-27 17:22:19,806 INFO [train.py:92] Process 600/3161(18.981335%) global average objf: -0.271463 over 3400832.0 frames, current batch average objf: -0.174807 over 6400 frames, epoch 0
2019-12-27 17:22:36,163 INFO [train.py:92] Process 700/3161(22.144891%) global average objf: -0.255235 over 3967360.0 frames, current batch average objf: -0.149902 over 6400 frames, epoch 0
```


[log-train-2019-12-27-17-20-34.txt](https://github.com/kaldi-asr/kaldi/files/4005666/log-train-2019-12-27-17-20-34.txt)

danpovey:
Amazing!  Well done!
I would appreciate if the others can help review this and help flag any
issues or design points that we might want to discuss.
I have an offsite meeting today and to-morrow and won't have much time to
look at it over the weekend.

egs/aishell/s10/run.sh
```diff
@@ -0,0 +1,320 @@
...
+  steps/train_lda_mllt.sh --cmd $train_cmd \
+   2500 20000 data/train data/lang exp/tri2_ali exp/tri3a || exit 1
+fi
+
+if [[ $stage -le 13 ]]; then
+  steps/align_fmllr.sh --cmd $train_cmd --nj $nj \
+    data/train data/lang exp/tri3a exp/tri3a_ali || exit 1
+fi
+
+if [[ $stage -le 14 ]]; then
```
naxingyu: Stages GT 14 can go to `local/run_chain.sh`.

egs/aishell/s10/run.sh
```diff
@@ -0,0 +1,320 @@
...
+kernel_size_list="1, 3, 3, 3, 3, 3" # comma separated list
+stride_list="1, 1, 3, 1, 1, 1" # comma separated list
+
+log_level=info # valid values: debug, info, warning
+
+if [[ $stage -le 24 ]]; then
+  echo "training..."
+
+  mkdir -p exp/chain/train/tensorboard
+  # train_checkpoint=./exp/exp/train/epoch-5.pt
```
naxingyu: `train_checkpoint` can be retrieved from the pt files.

egs/aishell/s10/run.sh
csukuangfj: Done.

egs/aishell/s10/run.sh
csukuangfj: Done.

naxingyu:
LGTM

danpovey:
Cool!  Since @naxingyu says its OK I will merge as I dont have time to look closely and dont want to delay you guys.  Thanks!!

egs/aishell/s10/chain/chain_loss.py
```diff
@@ -0,0 +1,95 @@
...
+            # i.e., stride == cols; otherwise, it will be re-allocated in Kaldi
+        else:
+            # (fangjun): we have to zero them out manually!
+            g_nnet_output_deriv_tensor.zero_()
+            g_xent_output_deriv_tensor.zero_()
+
+        # it contains [objf, l2_term, weight] and will be returned to the caller
+        objf_l2_term_weight_tensor = torch.zeros(3).float()
+
+        nnet_output = kaldi.CuSubMatrixFromDLPack(to_dlpack(nnet_output_tensor))
```
danpovey: I'm thinking it would be nice to have a shorter convenience mechanism for
`kaldi.CuSubMatrixFromDLPack(to_dlpack(nnet_output_tensor))`.
e.g.
`kaldi.PytorchToCuSubMatrix(x)`
?
what do you think?

csukuangfj: I agree.

For example, in some file `util.py`, we can write
```python
def from_tensor(tensor):
   m = kaldi.CuSubMatrixFromDLPack(to_dlpack(nnet_output_tensor))
   return m
```

Then we can use

```python
from util import from_tensor
nnet_output = from_tensor(nnet_output_tensor)
```

csukuangfj:
If we implement `kaldi.PytorchToCuSubMatrix(x)` in c++, then we would have C++ source-level
dependencies on PyTorch.

If we put it in `kaldi.py`, then everything `import kaldi` would also depend on PyTorch.

egs/aishell/s10/chain/egs_dataset.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # contains a list of a 3D array
+        # of shape [batch_size, seq_len, feat_dim]
+        feature_list = []
+
+        supervision_list = []
+        for b in batch:
+            key, rxfilename = b
+            key_list.append(key)
+            eg = read_nnet_chain_example(rxfilename)
```
danpovey: Hm... I'm a bit concerned about speed, here.   I think it would  be preferable to have an iterable-style dataset.  The issue is that it will repeatedly have to do fopen(), fseek(), fclose(), while if it were just reading the archive sequentially it would be just read() each time.

danpovey: Something else I should point out is that with the newer 'chain2'-style egs dir (to be merged), we rely more on scp files, which makes it easier to randomize things and combine previously dumped egs.  But it's still important to use a SequentialTableReader because we go to a lot of trouble to make sure that most of the reads are sequential (the scp entries mostly point to successive egs in an archive).

egs/aishell/s10/chain/inference.py
```diff
@@ -0,0 +1,85 @@
...
+            m = kaldi.FloatMatrix(m)
+            writer.Write(key, m)
+
+        if batch_idx % 10 == 0:
+            logging.info('Processed batch {}/{} ({:.6f}%)'.format(
+                batch_idx, len(dataloader),
+                float(batch_idx) / len(dataloader) * 100))
+    writer.Close()
+    logging.info('confidence is saved to {}'.format(
+        os.path.join(args.dir, 'confidence.scp')))
```
danpovey: These are not confidences, they are (unnormalized) logprobs so I think logprobs.scp would be a better name.

csukuangfj: I will change it to `logprobs.scp`.

But actually, there is no `log_softmax` applied to the output, should we drop `log` in `logprobs.scp` ?

```
output-layer name=output include-log-softmax=false dim=$num_targets max-change=1.5
```
You can see that `include-log-softmax=false` is false for `nnet_output`.

danpovey:
Hm, they are kind of unnormalized logprobs.  Actually they are
pseudo-log-likelihoods (they don't have the prior term).
Maybe nnet_output.scp would be more correct?

csukuangfj:
Fixed in https://github.com/kaldi-asr/kaldi/pull/3817

PR #3801, opened 2019-12-26
allowed_durations computation in standalone script

vimalmanohar:
Add a standalone script for computation of allowed_durations.txt and allowed_lengths.txt for e2e-LFMMI training. This is to use the functionality separate from utils/data/perturb_speed_to_allowed_lengths.py.

### DIFF
--- a/egs/wsj/s5/utils/data/get_allowed_durations.py
+++ b/egs/wsj/s5/utils/data/get_allowed_durations.py
@@ -1,0 +1,219 @@
+#!/usr/bin/env python3
+
+# Copyright     2017  Hossein Hadian
+#               2019  Facebook Inc. (Author: Vimal Manohar)
+# Apache 2.0
+
+
+""" This script generates a set of allowed lengths of utterances
+    spaced by a factor (like 10%). This is useful for generating
+    fixed-length chunks for chain training.
+"""
+
+import argparse
+import os
+import sys
+import copy
+import math
+import logging
+
+sys.path.insert(0, 'steps')
+import libs.common as common_lib
+
+logger = logging.getLogger('libs')
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter("%(asctime)s [%(pathname)s:%(lineno)s - "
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/data/get_allowed_durations.py
```diff
@@ -0,0 +1,194 @@
...
+                              "%(funcName)s - %(levelname)s ] %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+def get_args():
+    parser = argparse.ArgumentParser(description="""
+    This script creates a list of allowed durations of
+    utterances for flatstart LF-MMI training
+    corresponding to input data directory 'data_dir'
+    and writes it to allowed_durations.txt file in the
```
danpovey: This usage message does not very clearly explain what the script does.  It would be nice to know, at the very least, what the output looks like, e.g. what the format is..

vimalmanohar: I fixed the usage message. Ideally utils/data/perturb_speed_to_allowed_lengths.py. could be modified to use this. But I didn't want to affect other scripts without extensively testing it.

egs/wsj/s5/utils/data/get_allowed_durations.py
```diff
@@ -0,0 +1,219 @@
...
+    these lengths. Also writes an allowed_lengths.txt file,
+    which will be used by the e2e chain egs preparation script.
+
+    The perturbed data directory (created by a downstream script)
+    that only contains utterances of these allowed durations,
+    along with the corresponding allowed_lengths.txt are
+    consumed by the e2e chain egs preparation script.
+    See steps/nnet3/chain/e2e/get_egs_e2e.sh for how these are used.
+
+    Both the allowed_durations.txt and allowed_lengths.txt are formatted to
```
danpovey: mismatch "allowed_durations.txt" vs. "allowed_durs.txt".  Also the usage message for 'dir' only mentions 'allowed_lengths.txt'.  The usage message should be very clear that these are the two outputs of this script.
I.e. what this script does is to create these two files.

vimalmanohar: I updated the usage message.

PR #3803, opened 2019-12-30
Replace MatrixWriter with CompressedMatrixWriter.

csukuangfj:
Now it takes 4 times less space to save the confidence.ark


csukuangfj:
@danpovey @naxingyu 
please have a review.

danpovey:
Is this the default now, and is it configurable?  Before making it the default we should investigate whether there is any WER impact, because I'm not sure precisely how easily compressible nnet log-posteriors (assuming that's what we are talking about) would be.  Also I guess I assumed that we might mostly be dealing with pipes for the nnet-inference -> decoder stuff?  Although I can definitely see scenarios where writing the nnet output to disk might make sense.

csukuangfj:
It saves the output of the fully connected layer without applying log softmax.

I am not sure how far it affects the final WER/CER but it indeed reduces the file size about 4x
for the AIShell test and dev dataset.
- test: 20 G --> 5.1 G
- dev: 36 G --> 9.3 G

If the decoder is wrapped, we can output `lat.1.gz` directly to disk.

danpovey:
Mm, yeah... make sure it's just an option though, at this point.  (BTW,
readers can handle both compressed and uncompressed matrices so
compatibility is no issue).

csukuangfj:
I will change it to be configurable and let the user decide whether the output should
be **lossily** compressed or not.

danpovey:
Thanks.

egs/aishell/s10/chain/inference.py
```diff
@@ -47,7 +47,14 @@ def main():
     specifier = 'ark,scp:{filename}.ark,{filename}.scp'.format(
         filename=os.path.join(args.dir, 'confidence'))
 
-    writer = kaldi.CompressedMatrixWriter(specifier)
+    if args.save_as_compressed == 1:
```
danpovey: Hm, it doesn't seem super Pythonic to have 0/1 instead of a boolean.  I know reading boolean args in argparse is a pain, but there's a way to do it...

csukuangfj: in the function `_check_inference_args `
```python
args.save_as_compressed = (args.save_as_compressed == 1)
```
to change it from `int` to `bool`
would this be fine ?

danpovey:
Yes.

egs/aishell/s10/chain/options.py
```diff
@@ -25,6 +25,13 @@ def _set_inference_args(parser):
                         type=int,
                         default=-1)
 
+    parser.add_argument(
+        '--save-as-compressed',
```
danpovey: What I was talking about was mostly about changing ths command line arg to true/false and the type of the variablr to bool.

csukuangfj: no problem, I will change it.

csukuangfj: Done.

danpovey:
Thanks!  Merging.

PR #3806, opened 2019-12-31
sourcing LD_PRELOAD

naxingyu:


tools/extras/install_mkl.sh
```diff
@@ -254,6 +254,16 @@ ConfigLdSo() {
     ldconfig )
 }
 
+(
+  set +u
```
danpovey: please clarify in a comment what this does

tools/extras/install_mkl.sh
```diff
@@ -254,6 +254,16 @@ ConfigLdSo() {
...
+
+  echo "export LD_PRELOAD:=/opt/intel/mkl/lib/intel64/libmkl_def.so:\
```
danpovey: I'll merge this, but I'd like someone to remind me why we are using LD_PRELOAD and not LD_LIBRARY_PATH.

naxingyu: Not sure. LD_LIBRARY_PATH just doesn't work...

danpovey: so you did something like "export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/intel/mkl/lib/intel64/" ?

tools/extras/install_mkl.sh
naxingyu: done

tools/extras/install_mkl.sh
naxingyu: Yes, it doesn't work.

naxingyu: wait, let me try again

naxingyu: yep. double checked, doesn't work.

danpovey: OK.  MKL is weird.  Document the +u, why it's needed and what it does, and I'll merge.

danpovey: oh, you removed it..

PR #3807, opened 2020-01-01
add support to read/write  `*.npy` format.

### LINKED ISSUE #3626: I/O and NumPy
This is kind of a big issue.

I want to modify the I/O in kaldi10 to use the NumPy format wherever possible (at least for the binary version; for the text version it would be the way NumPy prints the arrays when you "print" them in Python).  The above is for the core data; but the way of combining into bigger files would still be done using the ark/scp mechanism.

The following
https://github.com/rogersce/cnpy
might be a good place to look for reference.

Bear in mind that the tensor directory does not currently compile.
For the time being the contents of the tensor directory would only be used for I/O.

csukuangfj:
Fix #3626

Now it is able to read/write `*.npy` files.

Note that it supports only little endian, single/double precision, 1-D/2-D numpy arrays with C order.


danpovey:
This is amazingly fast!  Sorry I have had other stuff to do today, I will
try to look at this in more detail tomorrow.
As regards numpy versions that you asked before, I think 3.0 would make
most sense.  1.0 is only for legacy which we don't care about,
and 3.0 only differs from 2.0 by supporting UTF-8 rather than just ASCII
strings, which seems unproblematic to support.

danpovey:
Well, sorry-- support 2.0 and 3.0.

csukuangfj:
No problems.

I will add support for 2.0 and 3.0 this weekend.

Should we support also big endian ?

src/matrix/test_data/generate_numpy_array_test_data.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def main():
+    data = np.arange(8).astype(np.float32)
+    np.save('float_vector.npy', data)
+    np.save('double_vector.npy', data.astype(np.float64))
+
+    data = data.reshape(2, 4).astype(np.float32)
+    np.save('float_matrix.npy', data)
+    np.save('double_matrix.npy', data.astype(np.float64))
```
naxingyu: The 'npy' files should be removed in the commit if they are generated during test.

src/matrix/numpy-array.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+template <typename Real>
+class NumpyArray {
+ public:
+  static constexpr const char* kMagicPrefix = "\x93NUMPY";
+  static constexpr int kMagicLen = 6;
+  // TODO(fangjun): We support only format 1.0 at present.
+  // Support for 2.0 and 3.0 can be added later when needed.
+  static constexpr int8_t kMajorVersion = 1;
+  static constexpr int8_t kMinorVersion = 0;
```
naxingyu: `constexpr` static members are supported since C++17. Is it really necessary?

src/matrix/numpy-array.cc
```diff
@@ -0,0 +1,270 @@
...
+  end_pos = header.find(",", start_pos);
+  KALDI_ASSERT(end_pos != header.npos);
+
+  auto fortran_order = header.substr(start_pos, end_pos - start_pos);
+
+  // WARNING(fangjun): we support only C order
+  KALDI_ASSERT(fortran_order == "False");
+
+  // for `shape`
+
```
naxingyu: extra new line

src/matrix/numpy-array.h
csukuangfj: Removed to an unnamed namespace.

src/matrix/numpy-array.cc
csukuangfj: Done.

src/matrix/test_data/generate_numpy_array_test_data.py
csukuangfj: My intention was that the test is run by `make test` and the user needn't generate
the test data in advance. In addition, it needs to install `numpy` to run this file
and not everyone using kaldi has `numpy` installed on his/her computer.

csukuangfj:
It turns out that numpy uses 1.0 by default for *.npy files.

I've added support for 2.0 but I do not know how to generate 2.0 format npy files so
this is not tested.

As far as I know, if you use only English characters, utf-8 has the same encoding with ASCII.
Do we really need to support utf-8 encoding in Kaldi, e.g., Chinese characters ?

csukuangfj:
By the way, we assume that it is little endian by default.

Do we need to support big endian?

danpovey:
Yeah, there should be big endian support (e.g. we definitely want to support ARM).

csukuangfj:
I will add big-endian support in the weekend.

danpovey:
Thanks!!  BTW I am having a closer look at your code now.  It is taking me
a while because sometimes I need to look things up in order to understand
everything.

csukuangfj:
this file
https://github.com/numpy/numpy/blob/master/numpy/lib/format.py
contains all the information you needed, especially the following lines https://github.com/numpy/numpy/blob/master/numpy/lib/format.py#L90-L180

danpovey:
Thanks. I saw that stuff before.  Actually I may have commented on the wrong PR here.  I was referring to your earlier code, about LF-MMI.

danpovey:
I had a look through this, I have no major problems with it (although I guess there are aspects that aren't finished).  This isn't as high priority to merge as the one with the updated results though, as it's not really needed yet.  I have to figure out how and when this stuff will be used.  Also, we should make sure that the *reading* code treats these NumPy arrays and regular matrix/vectors the same where possible, just like we can read a compressed matrix as if it is a regular matrix, or double as if it is float.

csukuangfj:
I only have time to do this pullrequest during the Chinese New Year. I think the only unfinished part
is to support big-endian.

danpovey:
Thanks!!!

csukuangfj:
Now it supports both big and little endian `*.npy` files

danpovey:
Fantastic!  Let me check the patch to-morrow...

danpovey:
Great!  
Before I merge this I have to figure out where this fits into the larger picture and also which branch to merge to.  One issue is, we should probably make it so that regular Kaldi matrices and vectors can read this format.  We might even make it so they write this format by default, at least when in binary mode.

src/matrix/test_data/generate_numpy_array_test_data.py
kkm000: I would rather check in some. Otherwise, the read test gets the dependency on the write test: you cannot run the read test in isolation, and also if the write test fails. Small binary files are totally ok.

kkm000: Better than a dependency of the test on numpy. We do not have such a dependency now, so pre-generated files are better by me.

kkm000:
@danpovey, maybe we should check this in. Adding the support for dual format is a separate work that can be done later, I think. This is getting outdated, and that's a lot of work, with tests and everything.

@csukuangfj, what about tools support? eg. copy-vectors or copy-matrix -- will these able to read numpy files transparently?

csukuangfj:
@kkm000 
What would the output of`copy-numpy-array`  look like? Should it accept the same options as `copy-matrix`?

danpovey:
It looks like it uses a separate reader/writer object now.
You're right, it would be more useful if it could somehow be activated from the command line.
But perhaps doing this just for copy-feats would be enough, i.e. give it an option to write and/or read in NumPy mode.  (?)

kkm000:
@csukuangfj 
> What would the output of `copy-numpy-array` look like? Should it accept the same options as copy-matrix?

Good question! We have a proliferation of binaries, so if we could roll that into `copy-feats`, as Dan suggested. What type of numpy arrays we support? Features are usually stored in tables, and the tables have no indication about their format. So I'm kinda torn. Either an extra binary, or an extra letter in the `ark:` specifier, akin to the current `,t` for text files.

Extending `copy-vector` and `copy-matrix` it would be more versatile, but leads to the same problem. Do we have a use case for tables of npy files? The tricky point is, we currently have 2 formats: text and binary, and 2 modes for storage: single file or table. For single file, binary/text output is controlled by the `--binary` switch for output (default text), while for a table, it's an added `,t` for text, default is binary. So, from the general tool support level, this needs some design, really. And I cannot put that anywhere close to the top of my queue now, unfortunately (and neither probably has Dan). Technically, this applies to features too, but we never practically store them as individual files.

I think we autodetect text/binary when reading either type, there is a magic header byte 0x07 preceding a binary file, if I'm not mistaken. Do the .npy files have a magic value at start of file? If they do, it should be a piece of cake to integrate as a full-fledged third I/O type in addition to the current text and binary files.

@danpovey, notwithstanding the above, is it ok to merge this part? Looks like a self-contained API, even though without a binary support.

danpovey:
I guess I'd be OK to merge in principle, but right now I don't see a clear
use case so if it were up to me I probably wouldn't merge.

src/matrix/test_data/generate_numpy_array_test_data.py
csukuangfj: Pre-generated files are checked in along with this pullrequest.

kkm000:
LGTM.

@danpovey, I'll merge. It's a high-quality C++, very readable and certainly a well-aligned implementation with a good unit test coverage. We started gung-ho about it, then priorities kinda changed so the requirement had become a meh; but our I/O is a low-maintenance code, and I think it's possible we may need this some day.

danpovey:
OK.

src/matrix/test_data/generate_numpy_array_test_data.py
kkm000: Thanks, it's better to decouple them this way. I discovered I had no numpy once I attempted to generate training progress plots of a model that was unstable, a year and a half of using that machine for training; usually I just gleaned all the progress info from logs (when everything converges a expected, it's easy to grep and see the pattern). So numpy by itself is a very non-requirement for Kaldi. :)

PR #3809, opened 2020-01-03
[src] fix beam and cutoff calculation bug when #active_tokens < min_a…

lizezheng:
There is a chance #active tokens is less than min active config, and this will lead to pruning beam and cutoff be inf, which is not disireable.
Fix this bug


danpovey:
Why is it not desirable?
I.e. what bad results does this cause?

lizezheng:
Hmm

lizezheng:
Hmm

lizezheng:
Hmm , should not call it a bug. When adaptive beam and  cur cutoff are both inf, then it seems that all the emitting and non-emitting arc will survive for next frame , and #active_tokens will increase asap to satisfy the min_active_token constraint, which seems to be better.

danpovey:
? That's exactly the desired behavior.  If min-active was hit, there must have been only a very small number of active tokens anyway.

PR #3814, opened 2020-01-06
Fix pybind11 test error w/o IntelMKL

naxingyu:
Resolve [this](https://github.com/kaldi-asr/kaldi/commit/277bbafab3304eb23a1950ce385579c33d1080dc#r36670826)
@csukuangfj @danpovey

### DIFF
--- a/src/pybind/Makefile
+++ b/src/pybind/Makefile
@@ -114,14 +114,18 @@ clean:
 	-rm -rf __pycache__
 	-rm -f $(CCFILES_OBJS)
 	-rm -f .depend.mk
+	-rm -f .mkl_env.mk
 
-test: all
+mkl:
+	-grep "LD_PRELOAD" ../../tools/env.sh > .mkl_env.mk
+
+test: mkl
 	python3 tests/test_dlpack_subvector.py
 	python3 tests/test_kaldi_pybind.py
 	make -C chain test
 	make -C cudamatrix test
 	make -C dlpack test
-	$(eval -include ../../tools/env.sh)
+	$(eval -include .mkl_env.mk)
 	make -C feat test
 	make -C fst test
 	make -C matrix test
--- a/src/pybind/Makefile
+++ b/src/pybind/Makefile
@@ -119,7 +119,7 @@ clean:
 mkl:
 	-grep "LD_PRELOAD" ../../tools/env.sh > .mkl_env.mk
 
-test: mkl
+test: all mkl
 	python3 tests/test_dlpack_subvector.py
 	python3 tests/test_kaldi_pybind.py
 	make -C chain test
--- a/src/pybind/Makefile
+++ b/src/pybind/Makefile
@@ -121,7 +121,7 @@ test: all
 	make -C chain test
 	make -C cudamatrix test
 	make -C dlpack test
-	$(eval include ../../tools/env.sh)
+	$(eval -include ../../tools/env.sh)
 	make -C feat test
 	make -C fst test
 	make -C matrix test
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,6 @@ GSYMS
 # They are added here to avoid accidently checkin.
 /src/**/CMakeLists.txt
 /build*
+
+# /src/pybind/
+/src/pybind/.mkl_env.mk


src/pybind/Makefile
```diff
@@ -121,7 +121,7 @@ test: all
 	make -C chain test
 	make -C cudamatrix test
 	make -C dlpack test
-	$(eval include ../../tools/env.sh)
+	$(eval -include ../../tools/env.sh)
```
csukuangfj: I am afraid this does not work.

The error message I encountered is

```bash
$ make test
python3 tests/test_dlpack_subvector.py
make: python3: Command not found
Makefile:119: recipe for target 'test' failed
make: *** [test] Error 127
```

The file `../../tools/env.sh` always exists, where you use MKL or not.

You should change `env.sh`

1. Use `:=` instead of `=`
2. Replace `$PATH` with `${PATH}`

naxingyu: `env.sh` does not always exist. And the error message you show seems irrelevent. I test this fix with or without MKL install, both passed.

naxingyu: See in [this line](https://github.com/kaldi-asr/kaldi/commit/277bbafab3304eb23a1950ce385579c33d1080dc#diff-8c42b80642e8e01f0c92a912760fd5beR259) that it is `:=` not `=`

csukuangfj: What I meant was `PATH`, NOT `LD_PRELOAD`.

Although you did not change `PATH`, but you included it to `Makefile`, `$PATH`
is not valid in `Makefile`, you have to use `${PATH}`.

naxingyu: Double confirmed, even if `env.sh` is not clean, i.e. you have previous installation which generated `env.sh` with `=`, it won't raise any error regardless of you compile against IntelMKL or not.

naxingyu: Could you paste your `env.sh`?

naxingyu: `$PATH` is not used in the IntelMKL export, so it does not make a difference whether it is valid in `Makefile` or not.

csukuangfj: I see.
My `env.sh` is different from yours!

I have installed kaldi lm and this line caused the error for me:

https://github.com/kaldi-asr/kaldi/blob/752df8c700abf7a5488c07f4025fe18c27e244d3/tools/extras/install_kaldi_lm.sh#L31

I guess you only have one line `env.sh` that is why `=` is OK. But I have two lines.

`:=` is not valid in bash; but if you have multiple lines in `env.sh`, `=` is a lazy assignment,
I guess it is an infinite recursive substitution for `PATH`.

csukuangfj:
`env.sh` is generated dynamically and may be different for different users.

I think it is not a good idea to include `env.sh` directly in Makefile since you will
never know what is inside it.

naxingyu:
I see. That's not an issue to include generated files. `kaldi.mk` is also generated. I can generate `env.mk` in current folder by copying the IntelMKL export line only. That should work.

csukuangfj:
I think it is better to embed `mkl` into the generated filename, e.g., `mkl_env.mk`.

naxingyu:
@csukuangfj Try the update.

danpovey:
Guys, I think it's best to not bother with these kinds of tricks, and
instead just change the configure script to store that dir in a variable.
And then you can modify files like pybind/chain/Makefile to make use of
that variable.
Going back and forth with these brittle scripts will just waste your time
in the long run.

naxingyu:
@danpovey yea, but this should be merge cause it will fix the error.

naxingyu:
once @csukuangfj verifies the fix.

csukuangfj:
I do not see any problems in your approach.

The advantage of putting it in `env.sh` instead of `kaldi.mk` is that it can be reused in `path.sh`.

But if you can put it in `env.sh`, you can also save the related information in `kald.mk`.
There will be no conflicts.

naxingyu:
put it in `env.sh` is for the convenience of developing pybind11 style recipes.

naxingyu:
If the MKL dependencies were used in compilation, it'll go to `kaldi.mk`. For now, only used to pass the test.

src/pybind/Makefile
```diff
@@ -114,14 +114,18 @@ clean:
 	-rm -rf __pycache__
 	-rm -f $(CCFILES_OBJS)
 	-rm -f .depend.mk
+	-rm -f .mkl_env.mk
 
-test: all
+mkl:
+	-grep "LD_PRELOAD" ../../tools/env.sh > .mkl_env.mk
+
+test: mkl
```
csukuangfj: You can write
```
test: all mkl
```
here and no need to delete `all`

naxingyu: resolved.

csukuangfj: I would prefer that you generate `mkl_env.mk` during compilation and use

```
ifdef MKLROOT
-include mkl_env.mk
endif
```

`MKLROOT` is defined in `kaldi.mk`
and remove the `mkl` rule in the Makefile.

danpovey:
@naxingyu the setup is broken for regular use in experiments now because `a:=b` is not a valid bash syntax.  I'm not going to accept your way of doing it anyway, so why waste time?

csukuangfj:
Your commit fixes my problem.

By the way, if you added `export LD_PRELOAD` to your `.bashrc` while installing MKL, then there is no need to add this to Makefile.

yes, `:=` is not valid in bash.

danpovey:
Adding things to .bashrc is an extremely bad pattern.  You can prevent people logging in, for instance.  All configuration values should be local.  Just do it how I said.  This is getting to be a waste of time.

src/pybind/Makefile
danpovey: There's no need for a separate .mk file for the mkl stuff; kaldi.mk already contains all sorts of variables and there's no need to create a new file every time we add a new variable.   The `configure` script can just add whatever variables are needed to kaldi.mk.

naxingyu:
I tried to fix this because it cause `make test` fail and I do not want to block anyone. And I don't think I am the only one that use IntelMKL. The issue you talk about can be solved by using `=` in `env.sh` and change it to `:=` in `.mkl_env.mk`. @danpovey do I make that change or leave it alone? If you don't like it, I can close this PR immediately.

danpovey:
I prefer you close this PR as it's wasting our time.  @csukuangfj can make a local fix so he can continue working, and you can work on a proper fix separately.

danpovey:
Sorry, @naxingyu ... I am a little sick which put me in a bad mood.

naxingyu:
@danpovey no problem. I see @csukuangfj has a more reasonable solution. #3817

PR #3815, opened 2020-01-06
Wrap kaldi_pybind to kaldi package.

csukuangfj:
Now everything lives in the `kaldi` package and the user
does not need to use `kaldi_pybind` any more.


csukuangfj:
@danpovey @naxingyu 
please have a review.

src/pybind/kaldi/__init__.py
```diff
@@ -0,0 +1,32 @@
+import os
+import sys
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), os.pardir))
+sys.path.insert(0, os.path.dirname(__file__))
+
+from kaldi_pybind import *
+
```
danpovey: I suppose there are going to be a lot of decisions about which things to put in the global `kaldi` namespace.  But it's OK.

csukuangfj: There are lots of classes in `kaldi_pybind` and it is a pain to list each one explicitly.

danpovey: Sure, that's OK; kaldi is just a thin wrapper for kaldi_pybind.  I was actually talking about the lines below, where you import things like Matrix.  But as I said, it's OK.

danpovey:
anyway this looks fine; merging.

PR #3817, opened 2020-01-06
Fix the MKL LD_PRELOAD issue for kaldi pybind test.

csukuangfj:
@danpovey @naxingyu 
please have a review.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -1244,6 +1244,15 @@ elif [ "`uname`" == "Linux" ]; then
       echo MKLLIB = $MKLLIBDIR >> kaldi.mk
     fi
     echo >> kaldi.mk
+
+    echo "MKL_LD_PRELOAD := $MKLROOT/lib/intel64/libmkl_def.so \\
+                  $MKLROOT/lib/intel64/libmkl_avx2.so \\
+                  $MKLROOT/lib/intel64/libmkl_core.so \\
+                  $MKLROOT/lib/intel64/libmkl_intel_lp64.so \\
+                  $MKLROOT/lib/intel64/libmkl_intel_thread.so" >> kaldi.mk
+
+    echo >> kaldi.mk
+
     check_exists makefiles/linux_x86_64_mkl.mk
     cat makefiles/linux_x86_64_mkl.mk >> kaldi.mk
     echo "MKLFLAGS = ${MKL_LINK_LINE} ${THREADING_LINE} $EXTRA_LIBS " >> kaldi.mk
--- a/src/pybind/Makefile
+++ b/src/pybind/Makefile
@@ -97,6 +97,10 @@ LDFLAGS += -Wl,-rpath=$(CURDIR)/../../tools/openfst/lib
 
 LIBFILE=$(LIBNAME)$(LIBFILE_EXTENSION)
 
+ifdef MKL_LD_PRELOAD
+LD_PRELOAD += $(MKL_LD_PRELOAD)
+export LD_PRELOAD
+endif
 
 .PHONY: all clean test
 
@@ -138,6 +142,6 @@ depend:
 	for f in $(CCFILES); do \
            $(CXX) -M -MT "$$(dirname $$f)/$$(basename -s .cc $$f).o" \
            $(CXXFLAGS) $$f >> .depend.mk; \
-        done
+      done
 
 -include .depend.mk
--- a/egs/aishell/s10/chain/inference.py
+++ b/egs/aishell/s10/chain/inference.py
@@ -43,7 +43,7 @@ def main():
     model.eval()
 
     specifier = 'ark,scp:{filename}.ark,{filename}.scp'.format(
-        filename=os.path.join(args.dir, 'confidence'))
+        filename=os.path.join(args.dir, 'nnet_output'))
 
     if args.save_as_compressed:
         Writer = kaldi.CompressedMatrixWriter
@@ -83,8 +83,8 @@ def main():
                 float(batch_idx) / len(dataloader) * 100))
 
     writer.Close()
-    logging.info('confidence is saved to {}'.format(
-        os.path.join(args.dir, 'confidence.scp')))
+    logging.info('pseudo-log-likelihood is saved to {}'.format(
+        os.path.join(args.dir, 'nnet_output.scp')))
 
 
 if __name__ == '__main__':
--- a/egs/aishell/s10/local/decode.sh
+++ b/egs/aishell/s10/local/decode.sh
@@ -20,13 +20,13 @@ post_decode_acwt=10  # can be used in 'chain' systems to scale acoustics by 10
 . parse_options.sh || exit 1
 
 if [ $# != 4 ]; then
-    echo "Usage: $0 [options] <graph-dir> <trans_model> <confidence_scp> <decode-dir>"
+    echo "Usage: $0 [options] <graph-dir> <trans_model> <nnet_output_scp> <decode-dir>"
     exit 1
 fi
 
 graphdir=$1
 trans_model=$2
-confidence_scp=$3
+nnet_output_scp=$3
 dir=$4
 
 if [[ ! -d $graphdir ]]; then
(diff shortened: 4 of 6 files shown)


src/configure
```diff
@@ -1244,6 +1244,15 @@ elif [ "`uname`" == "Linux" ]; then
       echo MKLLIB = $MKLLIBDIR >> kaldi.mk
     fi
     echo >> kaldi.mk
+
+    echo "MKL_LD_PRELOAD := $MKLROOT/lib/intel64/libmkl_def.so" >> kaldi.mk
```
naxingyu: $MKLROOT -> $(MKLROOT)

src/pybind/Makefile
```diff
@@ -97,6 +97,12 @@ LDFLAGS += -Wl,-rpath=$(CURDIR)/../../tools/openfst/lib
 
 LIBFILE=$(LIBNAME)$(LIBFILE_EXTENSION)
 
+ifdef MKL_LD_PRELAOD
+EMPTY :=
+SPACE := $(EMPTY) $(EMPTY)
```
naxingyu: Does `SPACE := " "` do the same?

csukuangfj: No.

You can use `$(info $(some_variable))` to view the value of `some_variable` in `Makefile`.

Neither can you write
```
SPACE := $(EMPTY)
```

**nor**
```
SPACE := $(EMPTY) $(EMPTY) $(EMPTY)
```

It has to be exactly two `$(EMPTY)`. I do NOT know why but it works!

naxingyu: Hmm... I tried `SPACE := " "` and tests passed..

csukuangfj: Did you set the LD_PRELOAD environment variable?

naxingyu: No

naxingyu: `echo $LD_PRELOAD` is empty

csukuangfj: I just tried `SPACE := " "` and the tests also passed.

But if you print the value of `LD_PRELOAD`, it is not colon `:` separated. I do
not know why spaces sperated values also work for `LD_PRLOAD`.

What I mentioned above is that you can only get colon separated `LD_PRELOAD`
using the aforementioned approached.

csukuangfj: If you use `SPACE := " "`, then please put a statement right after `LD_PRELOAD +=`
```
$(error $(LD_PRELOAD))
```
you will see that it prints a list of space separated paths.

src/configure
naxingyu: oh, so you are echoing the whole path, ok

src/pybind/Makefile
naxingyu: OK. np

src/configure
csukuangfj: $MKLROOT, $(MKLROOT) and ${MKLROOT} are both fine in **bash**.

src/pybind/Makefile
csukuangfj: mark as resolved.

src/configure
csukuangfj: mark as resolved.

src/configure
```diff
@@ -1244,6 +1244,15 @@ elif [ "`uname`" == "Linux" ]; then
...
+    echo "MKL_LD_PRELOAD += $MKLROOT/lib/intel64/libmkl_avx2.so" >> kaldi.mk
```
danpovey: Hm, it seems to me it  would be more straightforward to do
"echo MKL_LD_PRELOAD := $MKLROOT/lib/intel64/libmkl_def.so:\
$MKLROOT/lib/intel64/libmkl_avx2.so:\
.... "  >> kaldi.mk
then no need for substitution magic.

csukuangfj: I also tried the suggested approach. The problem is that the resulting `configure`
file cannot be properly indented, e.g.,

```
"echo MKL_LD_PRELOAD := $MKLROOT/lib/intel64/libmkl_def.so:\
<no space here>$MKLROOT/lib/intel64/libmkl_avx2.so:\
<no space here>.... "  >> kaldi.mk
```

csukuangfj: In addition, if indentation is not the problem, the resulting `kaldi.mk` would have a very long line for `MKL_LD_PRELOAD` which harms readability.

csukuangfj: @danpovey @naxingyu 
It happens that `LD_PRELOAD` accepts both colon `:` and space-separated paths.

Please refer to http://man7.org/linux/man-pages/man8/ld.so.8.html

>LD_PRELOAD
>              A list of additional, user-specified, ELF shared objects to be
>             loaded before all others.  This feature can be used to selec‐
>             tively override functions in other shared objects.
>
>             The items of the list can be separated by spaces ****or**** colons,
>             and there is no support for escaping either separator.

csukuangfj: Should it be marked as resolved?

src/configure
```diff
@@ -1244,6 +1244,15 @@ elif [ "`uname`" == "Linux" ]; then
...
+    echo "MKL_LD_PRELAOD += $MKLROOT/lib/intel64/libmkl_core.so" >> kaldi.mk
```
danpovey: Typo: PRELAOD.

I don't think you needed to worry about the indentation.  It's OK if not all lines are  properly indented.   But if spaces work fine, that's OK with me.

csukuangfj: thanks, I'll fix the typo. It seems that not all `*.so` are needed since I did NOT include all
of them because of the typo.

csukuangfj: fixed.

src/configure
```diff
@@ -1244,6 +1244,15 @@ elif [ "`uname`" == "Linux" ]; then
       echo MKLLIB = $MKLLIBDIR >> kaldi.mk
     fi
     echo >> kaldi.mk
+
+    echo "MKL_LD_PRELOAD := $MKLROOT/lib/intel64/libmkl_def.so \\
+                  $MKLROOT/lib/intel64/libmkl_avx2.so \\
+                  $MKLROOT/lib/intel64/libmkl_core.so \\
+                  $MKLROOT/lib/intel64/libmkl_intel_lp64.so \\
+                  $MKLROOT/lib/intel64/libmkl_intel_thread.so" >> kaldi.mk
```
kkm000, line 1252: Question: when we run Kaldi binaries, we use no threading (sequential), and this calls for multithreaded BLAS.  I missed most of the fun; are we going to always use threaded BLAS backend for Python surface API? Is this something hardcoded into/required by PyTorch?

danpovey: We don't ever want to use multi-threaded BLAS, actually; it just spins its wheels for no real speed gain.   Perhaps this should be changed, but I don't know the interaction with PyTorch.

kkm000:
I'll try to figure it out. I'd rather use libmkl_rt.so and dynamic loading.
This has been recommended by Intel over the LD_PRELOAD hack.

src/configure
kkm000: I found that PyTorch depends on numpy, and numpy has a weird mechanism for selecting a BLAS/LAPACKE libraries at runtime. The order can be specified at build time, and MKL goes the last by default. Maybe there is an environment variable to select a preferred one, I do not know yet. I just built my own numpy from source. But as far as MKL is concerned, the default is to build against libmkl_rt.so, the Intel's suggested way of calling MKL from Python. The default is Intel threading; for sequential, the env variable `MKL_THREADING_LAYER=SEQUENTIAL` should be set. Perhaps the best solution is to add a configuration option to Kaldi's configure for that. Maybe it should be hidden, IDK, or at the least well-documented. I'm kinda reluctant to use this mechanism in general, because it adds a dependency on the environment variable. AFAICR, It's also an option to set the threading via an API call; maybe adding that call to the common initialization code for all Kaldi binaries (ParseOption is one common point) will do the trick.

In general, it's a good idea to build Kaldi libraries this way; the only thing I want to not change is how binaries work, i.e. no requirement to set an environment variable.

What do you think?

danpovey: Thanks for the background!!  I don't think this is very critical; we will probably mostly be using this branch for testing and comparison purposes, and possibly as a base for future work, rather than aiming to merge it to master in the near future; we'll be focusing on the k2 direction.

PR #3818, opened 2020-01-06
code edit for resolving compatibility of python3

KimJeongSun:
that python subprocess open code need [universal_newlines=True] option at python3.

i checked it in my env and it's working well in python 2 and 3.


egs/wsj/s5/steps/segmentation/internal/merge_segment_targets_to_recording.py
```diff
@@ -178,7 +178,7 @@ def run(args):
                 cmd = ("copy-feats --binary=false {mat_fn} -"
                        "".format(mat_fn=targets[utt]))
                 p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
-                                     stderr=subprocess.PIPE)
+                                     stderr=subprocess.PIPE, universal_newlines=True)
```
danpovey: In the docs I see this:
`The universal_newlines argument is equivalent to text and is provided for backwards compatibility.`
https://docs.python.org/3/library/subprocess.html?highlight=subprocess

danpovey: I think this should be `text`, not `universal_newlines`.  And I don't  see how the changes to combine_ali_dirs.sh relate to this.

PR #3828, opened 2020-01-08
Support to construct CuSubMatrix/CuSubVector from DLPack without GPU support

csukuangfj:


src/pybind/dlpack/dlpack_pybind.cc
```diff
@@ -192,18 +168,15 @@ py::capsule MatrixToDLPack(py::object obj) {
...
 }
 
-py::capsule CuVectorToDLPack(py::object obj) {
-#if HAVE_CUDA == 1
-  auto* v = obj.cast<CuVector<float>*>();
-  auto* managed_tensor =
-      CreateDLManagedtensor(kDLGPU, CuDevice::GetCurrentDeviceId(), v->Data());
-
+template <typename V>
+py::capsule VectorToDLPackImpl(py::object obj, DLManagedTensor* managed_tensor,
```
danpovey: Is it necessary to have this order of arguments?  The google style guide requires output arguments to be last, so unless there is a good reason, it would be better to have `managed_tensor` be the last arg.  (And I'm not sure how `obj` fits into all of this).

csukuangfj: I agree.

I will change it now.

csukuangfj: Now it is changed to pass `py::object` by pointer and `DLManagedTensor*` is the last parameter.

danpovey:
Thanks!   Let me wait for one of the other guys to review before I merge.  Mainly as a mechanism to familiarize others with the code...

csukuangfj:
Before merging, could you please ask @jtrmal to configure the travis CI for the pybind branch and to test whether it works for this pullrequest.

PR #3832, opened 2020-01-09
WIP: Sync pybind11 with master

naxingyu:
sync for CI @csukuangfj @jtrmal


naxingyu:
CI is running, but pybind is not in Makefile yet.

jtrmal:
I will work on it in half an hour or so

csukuangfj:
CI will run pybind due to my previous commit.
https://github.com/kaldi-asr/kaldi/commit/653595deadfc7b7b9ff2d427e8e6a6d55136838b#diff-ae1b6a2f4e258618fdcae2f154c1cc31R81

csukuangfj:
Refer to https://travis-ci.com/kaldi-asr/kaldi/jobs/273849448#L1890

It is indeed installing pybind11.

danpovey:
An odd compilation error:
```
/usr/bin/ld: /usr/lib/llvm-3.8/bin/../lib/LLVMgold.so: error loading plugin: /usr/lib/llvm-3.8/bin/../lib/LLVMgold.so: cannot open shared object file: No such file or directory
```

csukuangfj:
I guess this is related to `-flto` flag with clang using the gold linker.
Switch to gcc may solve this problem

danpovey:
We should get it working with Clang too.  Maybe someone would try to work
out the proper options?
I understand the gold linker does LTO itself, so perhaps the -flto option
should not be included when using clang.
It could be done at the configure-script level maybe, by defining some
variable for compilation options to pybind11.
We should also at some point work on the CMake build for pybind11.

naxingyu:
Saw this solution, not sure if it works
```
LDFLAGS += -fuse-ld=gold
```

danpovey:
You could test it by manually editing the Makefile, and if it works, do it
via a variable in the configure script.

naxingyu:
Something like this
```
ifeq ($(CXX),clang)
  LDFLAGS += -fuse-ld=gold
fi
```

naxingyu:
But `-flto` is not used gloabally, and it might slow down compile time if configured for all subdirs. I'm thinking of adding it in `pybind/Makefile`

danpovey:
The problem with those kinds of Makefile conditionals is that they don't
really handle the general case when there might be cross-compilation or a
specific version of clang or specific pathname is specified.  I think it's
easier to do it in the configure script but if you can figure out a
reasonable (and general) way to test for the clang family from a Makefile,
that's OK too.

I was not talking about using -flto globally, I was talking about defining
a new variable with extra options for pybind compilation.

naxingyu:
oh, I see.

naxingyu:
I'll try and squash my commits once it is ready. @danpovey

naxingyu:
Turns out `-fuse-ld=gold` is already in the flags... Will look for other solutions.

csukuangfj:
will remove `-flto` from the following two lines work?
https://github.com/kaldi-asr/kaldi/blob/653595deadfc7b7b9ff2d427e8e6a6d55136838b/src/pybind/Makefile#L29-L30

naxingyu:
It might. But isn't it necessary for pybind?

csukuangfj:
No, `lto (link time optimization)` is just a sort of optimization and can be turned off.

naxingyu:
I know optimizations can be turned off, but I mean this comment
```
# pybind11 is heavily templated and generates code that is bloated before optimization.
# -flto is link time optimization which apparently is important.
```

csukuangfj:
From the Pybind11 doc (https://github.com/pybind/pybind11/blob/master/docs/basics.rst):
```
$ c++ -O3 -Wall -shared -std=c++11 -fPIC `python3 -m pybind11 --includes` example.cpp -o example`python3-config --extension-suffix`
```
`-flto` is not necessary to compile pybind11. 

Remove it may increase the size of the `.so` file.

naxingyu:
ok. What I don't get is that whoever wrote this comment, why add it in the first place?

naxingyu:
@danpovey @jtrmal  now travis pass. Please see if this solution is acceptable. I'll clean commit history later.

naxingyu:
@danpovey I've squashed my changes.

danpovey:
Thanks.. BTW for syncing master we shouldn't squash others' changes, we want to keep the history.  But its OK this time.

naxingyu:
@danpovey I only select and squash my own commits. You can see the commit history from [here](https://github.com/kaldi-asr/kaldi/pull/3832/commits)

danpovey:
Oops, I did squash and merge instead of merge; will revert.

PR #3834, opened 2020-01-10
Show how to decode in python with kaldi pybind and wrap warp-ctc

csukuangfj:
Now you can decode in python with kaldi pybind.

We have ported [`src/bin/latgen-faster-mapped.cc`][1] to kaldi pybind. You can find
an example showing how to decode here: 
https://github.com/mobvoi/kaldi/blob/fangjun-int-vector-holder/src/pybind/tests/test_latgen_faster_mapped.py

The output of the above demo script is as follows:

```sh
kaldi/src/pybind/tests$ ./test_latgen_faster_mapped.py
a.out --acoustic-scale=1.0 --allow-partial=true --beam=12.0 --lattice-beam=4.0 --max-active=7000 --max-mem=50000000 --min-active=200
lattice-scale --acoustic-scale=10 ark:- ark:-
BAC009S0764W0121 甚至 出现 交易 几乎 停滞 的 情况
LOG (a.out[5.5.733~4-4c87]:DecodeUtteranceLatticeFaster():decoder-wrappers.cc:375) Log-like per frame for utterance BAC009S0764W0121 is 1.23619 over 140 frames.
BAC009S0764W0122 一二 线 城市 虽然 也 处于 调整 中
LOG (a.out[5.5.733~4-4c87]:DecodeUtteranceLatticeFaster():decoder-wrappers.cc:375) Log-like per frame for utterance BAC009S0764W0122 is 1.29542 over 137 frames.
LOG (lattice-scale[5.5.733~4-4c87]:main():lattice-scale.cc:107) Done 2 lattices.
.
----------------------------------------------------------------------
Ran 1 test in 0.561s
OK
```

The decoded results are in Chinese, since we have used AIShell as the example.
You can replace relevant files, e.g., HCLG.fst, words.txt, trans_model, etc, in
the demo script with your own. 

-----

We removed the constructor and `Open` for `kaldi::Input` since both of them require a `bool*` argument, but Pybind11 does not support it.

Refer to
https://pybind11.readthedocs.io/en/stable/advanced/functions.html#allow-prohibiting-none-arguments
>Even when .none(true) is specified for an argument, None will be converted to a nullptr only for custom and opaque types. Pointers to built-in types (double *, int *, …) and STL types (std::vector<T> *, …; if pybind11/stl.h is included) are copied when converted to C++ (see Overview) and will not allow None as argument. To pass optional argument of these copied types consider using std::optional<T>

and 
- https://github.com/pybind/pybind11/pull/1760
- https://github.com/pybind/pybind11/issues/1865

In our previous setting, `bool*` was always `non NULL` in C++ even if we passed a `None` from Python.

In addition, we have added `read_vec_int`, `read_vec_flt` and `read_mat`, taking the same function name from
-  https://github.com/vesis84/kaldi-io-for-python/blob/master/kaldi_io/kaldi_io.py#L153
- https://github.com/vesis84/kaldi-io-for-python/blob/master/kaldi_io/kaldi_io.py#L217
- https://github.com/vesis84/kaldi-io-for-python/blob/master/kaldi_io/kaldi_io.py#L376

@danpovey @naxingyu @jtrmal @songmeixu @qindazhu Please have a review

[1]: https://github.com/kaldi-asr/kaldi/blob/master/src/bin/latgen-faster-mapped.cc


src/pybind/util/kaldi_io_pybind.cc
```diff
@@ -30,29 +30,40 @@ void pybind_kaldi_io(py::module& m) {
...
-             py::arg("rxfilename"))
+        // the constructor and `Open` both require a `bool*` argument
+        // but pybind11 does not support passing a pointer to a primitive
+        // type, only pointer to customized type is allowed.
+        //
+        // For more information, please refer to
+        // https://github.com/pybind/pybind11/pull/1760/commits/1d8caa5fbd0903cece06ae646447fff9b4aa33c0
+        // https://github.com/pybind/pybind11/pull/1760
+        //
+        // Were it be `bool*`, would it always be non-NULL in C++!
```
danpovey: In comments like this, before you explain *why* this code does what it does, explain what it does.  (I.e. that it returns a pair of bools [opened, binary], where `opened` is True if the stream was successfully opened, and `binary` is True if the the stream was opened and in binary format).  Don't assume that the reader (a) is familiar with pybind11, which is extremely difficult, and (b) has time to dig through Kaldi code to see the type and meaning of Input::Open().

csukuangfj: @danpovey 
thank you for the comment.

I have changed the documentation.

Since I am lazy to create a local branch so I started to add pybind wrapper for
LatticeArc based on this pullrequest.

danpovey:
OK thank, no problem.  LMK when you think this is ready to merge.

csukuangfj:
@danpovey 

please have a review.

danpovey:
Thanks!! Will look at it within a couple of hours, I hope.

src/pybind/decoder/decodable_matrix_pybind.cc
```diff
@@ -0,0 +1,113 @@
...
+#include "decoder/decodable-matrix.h"
+
+using namespace kaldi;
+
+namespace {
+
+void pybind_decodable_matrix_scale_mapped(py::module& m) {
+  using PyClass = DecodableMatrixScaledMapped;
+  py::class_<PyClass, DecodableInterface>(m, "DecodableMatrixScaledMapped")
+      .def(py::init<const TransitionModel&, const Matrix<BaseFloat>&,
```
danpovey: This is an issue with the underlying code, not just with your wrapping.  Any time we have a const Matrix as an arg to some kind of function or constructor, it should really be MatrixBase for greater generality.  This seems to limit the re-use of PyTorch arrays, since presumably requiring a Matrix would make it necessary to copy the data (?) Please let me know if I am right about the need for copying the data.

csukuangfj: I agree with you. But some APIs in kaldi accepting `const Matrix<float>*` takes the owner
ship of the passed pointer, for instance, 
https://github.com/kaldi-asr/kaldi/blob/1358f383937f99ca47bbd6acd6f28831868a6fc4/src/decoder/decodable-matrix.h#L51
will delete the passed pointer in the destructor
https://github.com/kaldi-asr/kaldi/blob/1358f383937f99ca47bbd6acd6f28831868a6fc4/src/decoder/decodable-matrix.h#L75-L77

To avoid uncessary copy, can we add a new API taking `const MatrixBase<float>&`
as input ?

csukuangfj: Oh, I think `const Matrix<BaseFloat>*` and `const Matrix<BaseFloat>&` will not copy data since it is a pointer/reference.

csukuangfj: The downside is that it cannot be constructed from a numpy array.

danpovey:
Yes, that's the point. Making it MatrixBase should make SubMatrix work as
well.
You could make a PR to Kaldi master with the changes to the underlying
code, too, if you (or one of the others) has time.

src/pybind/decoder/decodable_matrix_pybind.cc
csukuangfj: Another point that I feel unnatural in kaldi is that there is no table writer for `MatrixBase`
and `VectorBase`. If we want to write a `MatrixBase` to file using a table writer, we have
to convert it to  `Matrix<float>`

csukuangfj:
I will do it in the coming week if no one opens a pullrequest for this.

danpovey:
Thanks.
It might be possible to just change the templates in table-types.h for
those writers (MatrixWriter, etc.) to use the base-classes.
It might work fine- someone should test it.  Haowen, could you please try
it out?

qindazhu:
dpovey, ok, I'll take it on

csukuangfj:
resolved conflicts. Now CI is running for this pullrequest.

csukuangfj:
@danpovey 
are there any comments for this pullrequest? I know it is a huge pullrequest and
needs time to review the code.

src/pybind/fst/fst_pybind.h
```diff
@@ -21,6 +21,176 @@
 
 #include "pybind/kaldi_pybind.h"
 
+#include "fst/fst.h"
+
+template <typename A>
+void pybind_state_iterator_base_impl(py::module& m,
```
danpovey: just curious where all this kind of thing is needed.  Doing things like state iteration in Python will be very slow.

src/pybind/hmm/hmm_topology_pybind.cc
```diff
@@ -0,0 +1,104 @@
...
+
+  using State = HmmTopology::HmmState;
+  py::class_<State>(
+      hmm, "HmmState",
+      "A structure defined inside HmmTopology to represent a HMM state.")
+      .def(py::init<>())
+      .def(py::init<int>(), py::arg("pdf_class"))
+      .def(py::init<int, int>(), py::arg("forward_pdf_class"),
+           py::arg("self_loop_pdf_class"))
+      .def_readwrite("forward_pdf_class", &State::forward_pdf_class)
```
danpovey: Is it really necessary to add bindings down to this detailed level?

csukuangfj: I intended to expose such APIs to Python so that what we can do in C++, we can also do it in Python. I haven't thought of a situation where such low-level bindings can be useful.

I can remove such bindings now and add them when they are really needed.

src/pybind/fst/fst_pybind.h
csukuangfj: I just wrapped it for completeness.

Do we need to iterate an FST in Python? If not, I can remove them.

al-zatv: imho, it can be useful sometimes in scripts.

src/pybind/hmm/hmm_topology_pybind.cc
danpovey: Its OK, was just checking, we can keep them.

src/pybind/ctc/ctc_pybind.cc
```diff
@@ -0,0 +1,215 @@
...
+                                 DLPackCuSubVector<float>* workspace,
+                                 ctcOptions options) {
+  options.stream = nullptr;  // use the default stream
+  ctcStatus_t status = compute_ctc_loss(
+      activations.Data(), gradients->Data(), flat_labels.Data(),
+      label_lengths.Data(), input_lengths.Data(), alphabet_size, minibatch,
+      costs->Data(), workspace->Data(), options);
+  return status;
+
+  // we use the default CUDA stream by setting it to nullptr, i.e., 0
```
danpovey: This comment is too far from the code it refers to.  Also, how would this interact with multi-GPU training?  Isn't that a common scenario?  And right now, how would our Kaldi LF-MMI code deal with that?

csukuangfj:
We just committed the pybind wrapper for warp-ctc from https://github.com/baidu-research/warp-ctc/

It has no dependencies on PyTorch.

Although PyTorch has its own [torch.nn.CTCLoss](https://pytorch.org/docs/stable/nn.html#torch.nn.CTCLoss), warp-ctc is still in use today.
For example, ESPnet still uses it (https://github.com/espnet/espnet/issues/700)

It is interesting that there is no public benchmark comparing the speed of warp-ctc
and PyTorch's CTC.

We will wrap warp-ctc to PyTorch and will try to be as similar with torch.nn.CTCLoss as possible
so that you can freely switch to any one you like.

We will contribute recipes for CTC later.

src/pybind/fst/fst_pybind.h
csukuangfj: So we can just keep it and let the user choose whether to use it or not?

src/pybind/ctc/ctc_pybind.cc
csukuangfj: This function is invoked inside
https://github.com/kaldi-asr/kaldi/pull/3834/commits/5ae7abc1907cc69014674c345cbb8d817bafe209#diff-a3c9284f584c43207ad2eb032edbfee6R88

so we can always use the stream corresponding to the current GPU device selected in PyTorch.

I think it can be scaled without any modifications to mutiple GPUs.

csukuangfj: I thought users will normally not want to see the implementation details so
I put the comment at the end.

Should I move it to the beginning ?

src/pybind/ctc/ctc.mk
```diff
@@ -0,0 +1,29 @@
+
+WGET ?= wget
+
+# This commit is the latest commit available as of 2019.01.15 .
+# Update it if needed.
+COMMIT_ID := bc29dcfff07ced1c7a19a4ecee48e5ad583cef8e
```
danpovey: About this CTC stuff, can we talk about what the options were here?  I see that Warp-CTC has PyTorch bindings already.  What are the advantages of loading it via the Kaldi module?

src/pybind/ctc/ctc_pybind.cc
```diff
@@ -0,0 +1,215 @@
...
+      .def_readwrite("num_threads", &ctcOptions::num_threads,
+                     "used when loc == CTC_GPU, which stream the kernels "
+                     "should be launched in")
+      .def_readwrite("blank_label", &ctcOptions::blank_label,
+                     "the label value/index that the CTC calculation should "
+                     "use as the blank label");
+
+  const char* compute_ctc_loss_help_str =
+      "Compute the connectionist temporal classification loss between a "
+      "sequence of probabilities and a ground truth labeling on **CPU**.  "
```
danpovey: This help string is being used for both CPU and GPU code but says it's for CPU.

src/pybind/ctc/ctc_pybind_test.py
```diff
@@ -0,0 +1,324 @@
...
+
+import kaldi
+from kaldi import ctc
+
+try:
+    import torch
+except ImportError:
+    print('This test needs PyTorch.')
+    print('Please install PyTorch first.')
+    print('PyTorch 1.3.0dev20191006 has been tested and is guaranteed to work.')
```
danpovey: change 'guaranteed' to 'known'

src/pybind/ctc/ctc.mk
csukuangfj: I can think of the following advantages:

(1) It does not depend on PyTorch and can be used for any version of PyTorch.

(2) If kaldi pybind switches to other frameworks, we can still use warp-ctc,
as long as the framework implements the DLPack interface.

(3) The users are not required to install warp-ctc PyTorch bindings by themselves,
which will eliminate lots of issues encountered during the installation (see https://github.com/SeanNaren/warp-ctc/issues).

It is straightforward to wrap warp-ctc and why not to wrap it in kaldi pybind?

danpovey: OK cool.

src/pybind/decoder/decoder_wrappers_pybind.cc
```diff
@@ -0,0 +1,72 @@
...
+#include "decoder/decoder-wrappers.h"
+
+using namespace kaldi;
+
+namespace {
+
+template <typename FST>
+void pybind_decode_utterance_lattice_faster_impl(
+    py::module& m, const std::string& func_name,
+    const std::string& func_help_doc = "") {
```
danpovey: Can you explain to me why this func_help_doc string is passed in and not just hardcoded?

csukuangfj: I find that the current approach is more flexible.

If it is hardcoded, then `fst::StdFst` and `fst::GrammarFst`
will share the same help documentation.

If later we want to add a new type fst, it will use the same
help message and we are not able to provide a customized help message.

src/pybind/ctc/ctc_pybind.cc
csukuangfj: fixed.

src/pybind/ctc/ctc_pybind_test.py
csukuangfj: fixed.

src/pybind/ctc/ctc_pybind.cc
csukuangfj: fixed. Moved to the beginning.

csukuangfj:
Now the CI passed!

danpovey:
Great! If you want to, @csukuangfj , you can merge it yourself; I don't want the branches to get too far out of sync.

csukuangfj:
thanks. i will merge now. should I use squash ?

danpovey:
yes, squash please.

PR #3843, opened 2020-01-13
[scripts] fix python comp. in librispeech textnorm

zngguvnf:
During text normalization for librispeech my calculation crashed
because `$PYTHON` in normalize_text.sh is calling the system
python and not the python from my virtual enviroment.

Is there any reason to use `$PYTHON` and not `python`?
When inside a virtual env `python` will point to this python (and pip packages).
When outside a virtual env `python` will point to the system python like now.


danpovey:
That script hasn't been touched in a long time.. I think your changes are fine, merging.

PR #3845, opened 2020-01-14
recipe for GALE Arabic speech translation project

DongjiGao:
This is for the ASR part of the speech translation project. It is similar to the current GALE Arabic recipe but with more data for acoustic and language model training.


danpovey:
Is any of the translation stuff somehow baked into this script, i.e. are there aspects that are specific translation?  If not I might want you to call it s5c or whatever the next number is, but you should clarify how it differs.  E.g. did we do something with subwords in s5b?  And how do the WERs differ?

DongjiGao:
> Is any of the translation stuff somehow baked into this script, i.e. are there aspects that are specific translation? If not I might want you to call it s5c or whatever the next number is, but you should clarify how it differs. E.g. did we do something with subwords in s5b? And how do the WERs differ?

There's no specific translation stuff. So I think I should call it s5d (s5b use graphemic lexicon and  s5c is the subword version of s5b).
s5c (subword) is slightly worse than s5b.  
s5d use more training data from GALE than s5b and the WER decrease from 16.47 to 14.89. 
I also run a subword version of s5d, it is also slightly worse than the word system after RNNLM-rescoring. So I did not check that in.

danpovey:
Looks OK to me.  Do you think it needs testing before merge?

DongjiGao:
> Looks OK to me. Do you think it needs testing before merge?

I have already done some tests. I think it's good to be merged.

danpovey:
Thanks a lot!!  Merging.

PR #3848, opened 2020-01-14
get gpu automatically rather than specify device_id

fanlu:
Hi, @csukuangfj , please have a review.
if you have implemented this func already, please let me know. I will discard this pr.
else then We can post  python training task in sge queue when we using this method.

### DIFF
--- a/src/pybind/cudamatrix/cu_device_pybind.cc
+++ b/src/pybind/cudamatrix/cu_device_pybind.cc
@@ -37,7 +37,7 @@ void pybind_cu_device(py::module& m) {
         py::arg("device_id"));
 
   m.def("SelectGpuId",
-        [](std::string use_gpu) {
+        [](const std::string& use_gpu) {
 #if HAVE_CUDA == 1
           CuDevice::Instantiate().SelectGpuId(use_gpu);
 #else
--- a/src/pybind/cudamatrix/cu_device_pybind.cc
+++ b/src/pybind/cudamatrix/cu_device_pybind.cc
@@ -36,6 +36,16 @@ void pybind_cu_device(py::module& m) {
         },
         py::arg("device_id"));
 
+  m.def("SelectGpuId",
+        [](std::string use_gpu) {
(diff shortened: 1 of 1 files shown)


src/pybind/cudamatrix/cu_device_pybind.cc
```diff
@@ -36,6 +36,16 @@ void pybind_cu_device(py::module& m) {
         },
         py::arg("device_id"));
 
+  m.def("SelectGpuId",
+        [](std::string use_gpu) {
```
csukuangfj: `const std::string& use_gpu`

The reason to use a specified GPU rather than an automatically selected
GPU is that we want to share data between PyTorch GPU tensor and CuSubMatrix/CuSubVector.

csukuangfj: Your commit may be useful in some cases. @danpovey

fanlu: Is there a plan to submit pytorch traning task distributed?

csukuangfj: We indeed have one internally, which can use either horovod or DDP (Distributed Data Parallel).
I do not know whether @danpovey is interested in distributed training with PyTorch.

fanlu: Cool!

danpovey:
@csukuangfj this looks to me like it wouldn't break anything.  How about you merge it if you have no objection?
RE distributed training: yes of course that's of interest.. we have to figure out what will be easy to support, given that people will be running on very different infrastructure.

csukuangfj:
yes, I agree.

csukuangfj:
I think I do not have permission to merge it.

csukuangfj:
For the distributed training, should we contribute for DDP training, which is supported by PyTorch (https://pytorch.org/tutorials/intermediate/ddp_tutorial.html).

csukuangfj:
merging

PR #3856, opened 2020-01-19
fix --cmd param errors of aishell scripts in Pytorch training

qindazhu:
some `$train_cmd` should be `"$train_cmd"` when they are used as the parameters


qindazhu:
BTW, there are some files related `warp-ctc`, should we add them to `.gitignore`? 

`      
      src/pybind/ctc/warp-ctc 
       src/pybind/ctc/warp-ctc-bc29dcfff07ced1c7a19a4ecee48e5ad583cef8e/
        src/pybind/ctc/warp-ctc.tar.gz
`

danpovey:
OK with me.

Issue #3870, opened 2020-01-22
installing kaldi facing issue.


NarasimmanSaravana1994:
While running make command .. I'm getting error ..

                                          ^
# --error 0x1 --
Makefile:32: recipe for target 'cu-kernels.o' failed
make[1]: *** [cu-kernels.o] Error 1
make[1]: Leaving directory '/home/administrator/kaldi/src/cudamatrix'
Makefile:136: recipe for target 'cudamatrix' failed
make: *** [cudamatrix] Error 2

My system using CUDA version 10.0 . Is there any issue in that or any other way ot install CUDA 10 supported installation step......

Thanks

danpovey:
PR #3871 seems to be responsible for this.  I reverted it.  Through my email I'm seeing other comments in this PR but they don't appear here... github email can sometimes be unreliable.

Issue #3860, opened 2020-01-20
Use const Fst using ReadFstKaldiGeneric in align-mapped


OhmVikrant:
Reading const Fst should be allowed for align-mapped.

OhmVikrant:
You can refer to [](https://github.com/kaldi-asr/kaldi/commit/33255ed224500f55c8387f1e4fa40e08b73ff48a) for the necessary changes.

PR #3878, opened 2020-01-25
[src] Allow usage of const Fst (#3871, #3870, #3860)

OhmVikrant:
VectorFst is not required. Simply const Fst can be used for ReadFstKaldi.
I tried to build it again and found no issues.


danpovey:
This is not compiling.  I think it's better if you create an issue stating what the problem was you were trying to fix, and maybe someone else can fix it (or help you fix it).

PR #3881, opened 2020-01-26
Fix shebangs

esdeboer:
Replace #!/bin/bash with #!/bin/usr/env bash in all .sh files for portability.

Contains a maintenance script which will go through all .sh files in the project and do the replacement.  The script runs fine on my linux and freeBSD (that's why it is a bit more complicated, sed -i doesn't work on BSD)


danpovey:
You have changed the mode on a lot of files, please revert those changes!
I don't see where your new script is, likely because the diff is too big to display, but maybe after you fix the mode it will be OK.  On bsd/mac you need to give a suffix to the -i command, e.g. sed -i.bak

esdeboer:
The script is in the second commit:
https://github.com/kaldi-asr/kaldi/pull/3881/commits/054d0c145c53f517520ec75fc392b01d63d300ba
I'll check the mode changes, I thought I fixed those. Do you want a commit which fixes, or shall I ammend the first commit and force push?

jtrmal:
N normally we squash anyway during merge, so I don't think it matters...
y.

esdeboer:
Fixed the script, the modes shouldn't have changed now.

PR #3884, opened 2020-01-27
[scripts] Make adjust_unk_graph.sh convert to const fst

nishchalb:
I think having it const is usually the desired case. An alternative could be adding an argument to the script for whether to do the conversion or not


jtrmal:
This makes sense to me,  but let's wait for Dan...  But I think your're right

PR #3885, opened 2020-01-27
Fixed scoring for SAD and diarization

desh2608:
Thanks to @boeddeker for identifying this error in the scoring scheme.


boeddeker:
@desh2608 Are the diarization results obtained with the uploaded models? I got slightly different numbers:
```
File               DER    JER
S02_U06.ENH      58.66  63.53
S09_U06.ENH      64.92  75.95
*** OVERALL ***  61.00  69.74
```

jtrmal:
@desh2608 you know about the conflict, right?

desh2608:
@jtrmal Yes, I think the difference may be because of evaluation with/without array synchronization (stage 0 of decode). I will confirm this and update this PR if necessary. In the meantime, I will add WIP.

egs/chime6/s5_track2/local/decode.sh
```diff
@@ -8,11 +8,11 @@
 # Apache 2.0
 
 # Begin configuration section.
-nj=8
+nj=10
 decode_nj=10
 stage=0
 sad_stage=0
-diarizer_stage=0
+diarizer_stage=1
```
danpovey: this is correct?

desh2608: Stage 0 basically keeps Array U06 and discards all others, so it's a matter of choice. But yeah, for the baseline, we should start at stage 0. I'll fix it with a few other things in my next commit. This PR is still WIP.

egs/chime6/s5_track2/local/diarize.sh
```diff
@@ -91,23 +93,35 @@ fi
...
-  cd ..
+  mkdir -p $out_dir/scoring
+  for session in ${session_ids}; do
+    ref_rttm_path=$out_dir/scoring/rttm_ref.$session
+    grep "$session" $ref_rttm > $ref_rttm_path
+    if ! [ -s $ref_rttm_path ]; then
+      rm $ref_rttm_path
+      continue
+    fi
+    for array in ${array_ids}; do
```
boeddeker: @desh2608 Why does `diarize.sh` distinguish between the arrays?
When I understand the CHiME-6 website correctly the users can use all arrays to obtain the `hyp_rttm`.
For simplicity the baseline only took `U06`. So this is a very simple array selection algorithm.

desh2608: Yes, if you are using selected arrays you can just update the `array_ids` accordingly. This is mostly just for scoring convenience than anything else, and also makes the scoring of SAD and diarization similar.

desh2608: @boeddeker @sw005320 Can you review these changes and see if they make sense to you?

desh2608:
@boeddeker @sw005320 Can you review these changes and see if they make sense to you?

egs/chime6/s5_track2/local/diarize.sh
boeddeker: I still don't get it, why distinguishing between `array_ids`.
In CHiME-5 and DIHARD it is important to distinguish between the arrays, because each array has start and end times for each utterance. So it was important to have the correct reference array.
But in CHiME-6 all arrays have the same start and end times.
So the reference `rttm` file is independent of the array ID.

The code a few lines lower calls `score.py` for each array.
In the baseline this is fine, because it takes a single array per session (`U06`).
When you have an array selection algorithm the reference array changes inside a session and `score.py` is called on a subset of a session.
I am not sure, but probably this will yield invalid scores.

Inside the `rttm` file I would expect that the `File ID` only contains the session name.
```
...
SPEAKER S01 1    1.92   10.86 <NA> <NA> P01 <NA>
...
```
When the `File ID` contains the array id (`S01_U06.ENH`) it should also contain the channel, when someone only works on single channel data.

desh2608: > I still don't get it, why distinguishing between `array_ids`.
> In CHiME-5 and DIHARD it is important to distinguish between the arrays, because each array has start and end times for each utterance. So it was important to have the correct reference array.
> But in CHiME-6 all arrays have the same start and end times.
> So the reference `rttm` file is independent of the array ID.

Yes, and the reference RTTM is indeed independent of the array ID. This loop over array ids is for scoring the output of all hypotheses for all arrays.
> 
> The code a few lines lower calls `score.py` for each array.
> In the baseline this is fine, because it takes a single array per session (`U06`).
> When you have an array selection algorithm the reference array changes inside a session and `score.py` is called on a subset of a session.
> I am not sure, but probably this will yield invalid scores.

If you have a fancy array selection mechanism, you can just name the session id with some suffix of your choice (say `_combined`, and then score it using `array_ids="_combined"`. I don't see why  this would yield invalid scores.
> 
> Inside the `rttm` file I would expect that the `File ID` only contains the session name.
> 
> ```
> ...
> SPEAKER S01 1    1.92   10.86 <NA> <NA> P01 <NA>
> ...
> ```
> 
> When the `File ID` contains the array id (`S01_U06.ENH`) it should also contain the channel, when someone only works on single channel data.

sw005320: I understood @boeddeker's point that it would be better to provide the array-id-independent reference RTTM for CHiME-6 but I think it is OK for me to use the current script. We may leave this script as it is but add the comment that the reference RTTM files are array-id-independent. Also, again it is true that we may make it channel-dependent for various possibilities, but as a **first** baseline method, we chose a simple pipeline (performing beamforming for each array and pick up one array manually), and within this method, the current solution is reasonable.

desh2608:
@danpovey I think this can be merged.

egs/chime6/s5_track2/local/diarize.sh
boeddeker: I thought this may be a good time to remove all these `grep`s and `sed`s and directly write the desired file to the hdd.

@sw005320 I am fine with the baseline. Selecting a reference array is a reasonable choice.
In DIHARD2 it was also fine to perform the SAD and diarization on each array.
But in CHiME-6 the goal is the WER, so we have to select one array for the baseline and cannot use all arrays at the same time.

> If you have a fancy array selection mechanism, you can just name the session id with some suffix of your choice (say _combined, and then score it using array_ids="_combined". I don't see why this would yield invalid scores.

I know that I can do this. But what is, when someone uses `_U0X` where X is the selected array?
Then your `rttm` file splits over all 6 arrays?

My point is, I do not see a motivation for the default value `array_ids="U01 U02 U03 U04 U05 U06"`.
The `out_dir` argument should already contain all information (e.g. reference array, dereverberation and beamforming).

When I am alone with my opinion, then keep it as it is.

desh2608: @boeddeker I know it's not a pretty solution, and I am open to discussing alternatives.

desh2608: @boeddeker I have changed the scoring in the diarization according to your suggestion. Now it takes just a single ref rttm and a single hyp rttm and evaluates them.

However, the scoring for SAD (stage 4 in local/decode.sh) still presents a problem. Do you think I should change that to be similar to the diarization scoring as well?

desh2608: I have now updated both scorings to be independent of arrays.

egs/chime6/s5_track2/local/decode.sh
```diff
@@ -8,13 +8,15 @@
 # Apache 2.0
 
 # Begin configuration section.
-nj=8
+nj=10
 decode_nj=10
```
boeddeker: Does `decode_nj=10` works for you? In CHiME-6 `S09_U05` (also in CHiME-5) and `S01_U03` (chime6-synchronisation has problems) are missing.

Later in this file the loop `for mictype in u01 u02 u03 u04 u06; do` ignores `U05` for DEV and EVAL. In DEV we can then use 10 jobs. Since in EVAL `S01_U03` is missing I got an error, because I was only allowed to use 9 jobs with `run.pl`.

egs/chime6/s5_track2/local/diarize.sh
boeddeker: Thanks. The new code looks resonable to me.

egs/chime6/s5_track2/local/decode.sh
desh2608: Good catch. I'll change that to be equal to the number of lines in utt2spk.

PR #3886, opened 2020-01-28
[scripts] add python script to process xml file 

DongjiGao:
Add python script to process xml file and modified related scripts to use it if xml binary is not installed. According to what @jtrmal mentioned.

### DIFF
--- a/egs/gale_arabic/s5d/local/check_tools.sh
+++ b/egs/gale_arabic/s5d/local/check_tools.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# check whether bs4 and lxml is installed
+if ! python -c "import bs4" 2>/dev/null; then
+  echo "$0: BeautifulSoup4 not installed, you can install it by 'pip install beautifulsoup4' if you prefer to use python to process xml file" 
+  exit 1;
+fi
+
+if ! python -c "import lxml" 2>/dev/null; then
+  echo "$0: lxml not installed, you can install it by 'pip install lxml' if you prefer to use python to process xml file"
+  exit 1;
+fi
+
+echo "both BeatufileSoup4 and lxml are installed in python"
+exit 0
--- a/egs/gale_arabic/s5d/local/check_tools.sh
+++ b/egs/gale_arabic/s5d/local/check_tools.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 # check whether bs4 and lxml is installed
-if ! python -c "import bs4" 2>/dev/null; then
+if ! python3 -c "import bs4" 2>/dev/null; then
   echo "$0: BeautifulSoup4 not installed, you can install it by 'pip install beautifulsoup4' if you prefer to use python to process xml file" 
   exit 1;
 fi
 
-if ! python -c "import lxml" 2>/dev/null; then
+if ! python3 -c "import lxml" 2>/dev/null; then
   echo "$0: lxml not installed, you can install it by 'pip install lxml' if you prefer to use python to process xml file"
   exit 1;
 fi
--- a/egs/gale_arabic/s5d/local/process_xml.py
+++ b/egs/gale_arabic/s5d/local/process_xml.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python
+
+import sys
+import argparse
+
+try:
+    from bs4 import BeautifulSoup
+except ImportError as error:
+    print(sys.argv[0] + ": ImportError,", error)
+    print(sys.argv[0] + ": BeautifulSoup4 is needed to process xml file. You can install it by 'pip install beautifulsoup4'.")
+    sys.exit(1)
+
+def get_args():
+    parser = argparse.ArgumentParser(description="""This script process xml file.""")
+    parser.add_argument("xml", type=str, help="""Input xml file""") 
+    parser.add_argument("output", type=str, help="""output text file""") 
+    args = parser.parse_args()
+    return args
+
+def process_xml(xml_handle, output_handle):
+    soup = BeautifulSoup(xml_handle, "xml")
+    for segment in soup.find_all("segment"):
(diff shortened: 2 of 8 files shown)


egs/gale_arabic/s5d/local/process_xml.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
```
jtrmal: please write 'python3' explicitly in both scripts (i.e. `#!/usr/bin/env python3`)

egs/gale_arabic/s5d/local/process_xml.py
```diff
@@ -0,0 +1,41 @@
...
+
+import sys
+import argparse
+
+try:
+    from bs4 import BeautifulSoup
+except ImportError as error:
+    print(sys.argv[0] + ": ImportError,", error)
+    print(sys.argv[0] + ": BeautifulSoup4 is needed to process xml file. You can install it by 'pip install beautifulsoup4'.")
```
jtrmal: This is nice, but It would be good to add a new script, local/check_tools.sh that would check for the packages also (and would be run in the beginning of the run.sh. For parsing xml, you will need lxml package also (which is the reason it would be better to add the script and check for it explicitly)

egs/mgb2_arabic/s5/local/mgb_data_prep.sh
```diff
@@ -39,11 +33,21 @@ head -500 $train_dir/wav_list > $train_dir/wav_list.short
 set -e -o pipefail
 
 xmldir=$db_dir/train/xml/bw
-cat $train_dir/wav_list | while read basename; do
+if [ -z $(which xml) ]; then
```
jtrmal: `which` should be avoided, see https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then for rationale and how to do it correctly

egs/mgb2_arabic/s5/local/mgb_data_prep.sh
```diff
@@ -39,11 +33,21 @@ head -500 $train_dir/wav_list > $train_dir/wav_list.short
...
+  done
+else
+  echo "xml found, using it"
+  cat $train_dir/wav_list | while read basename; do
     [ ! -e $xmldir/$basename.xml ] && echo "Missing $xmldir/$basename.xml" && exit 1
     xml sel -t -m '//segments[@annotation_id="transcript_align"]' -m "segment" -n -v  "concat(@who,' ',@starttime,' ',@endtime,' ',@WMER,' ')" -m "element" -v "concat(text(),' ')" $xmldir/$basename.xml | local/add_to_datadir.py $basename $train_dir $mer
     echo $basename $wavDir/$basename.wav >> $train_dir/wav.scp
-done 
+  done
+fi
```
jtrmal: Did you check both approaches produce the same outputs?

egs/mgb2_arabic/s5/local/process_xml.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
```
jtrmal: I didn't check this file but please include the changes from the previous py file -- I assume the files are either identical or very similar

jtrmal:
@danpovey To explain -- we had some feedback from the sponsor side about the xmlstar tool -- I checked and it quite a hassle to install it manually, so I asked Dongji to provide python script to do the same -- both packages (bs4 and lxml) are easily installable using pip

egs/gale_arabic/s5d/local/process_xml.py
jtrmal: Oh and please make sure it actually does run under py3 if youused py2.7

danpovey:
I think it's better to have it invoke python3 in the shebang.  Would
require less testing, and we already have a python3 dependency.

jtrmal:
yeah, I agree

egs/gale_arabic/s5d/local/process_xml.py
DongjiGao: OK, it runs under python3, I will write python3.

egs/gale_arabic/s5d/local/process_xml.py
DongjiGao: OK

egs/mgb2_arabic/s5/local/mgb_data_prep.sh
DongjiGao: Yes, but in different order

egs/mgb2_arabic/s5/local/mgb_data_prep.sh
DongjiGao: OK, thanks. Will fix that.

egs/mgb2_arabic/s5/local/mgb_data_prep.sh
DongjiGao: > Did you check both approaches produce the same outputs?

Yes, but in different order.

egs/gale_arabic/s5d/local/check_tools.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# check whether bs4 and lxml is installed
+if ! python -c "import bs4" 2>/dev/null; then
```
danpovey: if the scripts will run as python3, you should test this with python3.

egs/mgb2_arabic/s5/local/check_tools.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# check whether bs4 and lxml is installed
+if ! python -c "import bs4" 2>/dev/null; then
```
danpovey: .. same here, assuming these scripts are also invoked as python3.

danpovey:
OK, will merge

PR #3891, opened 2020-01-29
Upstream online batched pipeline

luitjens:
This completes the online pipeline.  Code has been tested on V100 heavily and lightly on other GPUs.  This completes the online pipeline.  Next phase will implement a binary which uses the online pipeline in our batched decoder.  That should be coming soon.


danpovey:
Great!  I dont see any obvious problems that would mess up the build, so I'll merge this.  Dan

Issue #3893, opened 2020-01-30
Receiving Compilation Error: compute-mfcc-online-batched-cuda.cc:18:31: fatal error: cuda_profiler_api.h: No such file or directory


codeforequity-at:
For the latest master branch, I am receiving a compilation error - i have no cuda installed, but it doesn't make a difference whether i am using the "--use-cuda=no" arg or not.

```
g++ -M -std=c++11 -I.. -isystem /opt/kaldi/tools/openfst-1.6.7/include -O1 -Wno-sign-compare -Wall -Wno-sign-compare -Wno-unused-local-typedefs -Wno-deprecated-declarations -Winit-self -DKALDI_DOUBLEPRECISION=0 -DHAVE_EXECINFO_H=1 -DHAVE_CXXABI_H -DHAVE_OPENBLAS -I/opt/kaldi/tools/OpenBLAS/install/include -msse -msse2 -pthread -g -fPIC apply-batched-cmvn-online-cuda.cc compute-mfcc-online-batched-cuda.cc compute-online-feats-cuda.cc compute-online-feats-batched-cuda.cc compute-fbank-feats-cuda.cc compute-mfcc-feats-cuda.cc apply-cmvn-online-cuda.cc >> .depend.mk
compute-mfcc-online-batched-cuda.cc:18:31: fatal error: cuda_profiler_api.h: No such file or directory
compilation terminated.
../makefiles/default_rules.mk:149: recipe for target 'depend' failed
make[1]: *** [depend] Error 1
make[1]: Leaving directory '/opt/kaldi/src/cudafeatbin'
Makefile:127: recipe for target 'cudafeatbin/depend' failed
make: *** [cudafeatbin/depend] Error 2
The command '/bin/bash -c cd /opt/kaldi/src && make depend && make' returned a non-zero code: 2
```

jtrmal:
Are you absolutely sure you have the latest version? An error like this was
fixed a week ago.
Y.

jtrmal:
I was able to reproduce, will fix
y.

codeforequity-at:
yes - i am building a docker image with a fresh clone

jtrmal <notifications@github.com> schrieb am Do., 30. Jän. 2020, 15:56:

> Are you absolutely sure you have the latest version? An error like this was
> fixed a week ago.
> Y.
>

jtrmal:
I created #3895 to resolve it -- will merge once I finish test compile

codeforequity-at:
[...]

Issue #3894, opened 2020-01-30
Build fails when making kaldi/src with configure --shared --use-cuda=no (docker and bare)


devspyrosv:
Hello!)
I am using commit #743eb23d8 [egs] Added tri3b and chain training for Aurora4 (#3862)

When I try to build the docker image using a modified version of the provided Dockerfile the build fails during the last 2 make commands. 

The same thing happens if I try to build it in a clean Ubuntu 16.04 machine.

In order to reach this point I had to make some changes to the Dockerfile. 
In essence:

* I have added gfortran (by the way there is something wrong with the commit that fixes gfortran) and gawk.

* I copy srilm.tgz /opt/kaldi/tools/ .

* I use the multistage build technique to save some time (https://docs.docker.com/develop/develop-images/multistage-build/).

So when this point is reached: 
```
 cd /opt/kaldi/src && \
    ./configure --shared --use-cuda=no && \
    make depend -j 2 && \
    make -j 2
```
it fails with: 

```
g++ -M -std=c++11 -I.. -isystem /opt/kaldi/tools/openfst-1.6.7/include -O1 -Wno-sign-compare -Wall -Wno-sign-compare -Wno-unused-local-typedefs -Wno-deprecated-declarations -Winit-self -DKALDI_DOUBLEPRECISION=0 -DHAVE_EXECINFO_H=1 -DHAVE_CXXABI_H -DHAVE_MKL -I/opt/intel/mkl/include -m64 -msse -msse2 -pthread -g -fPIC apply-batched-cmvn-online-cuda.cc compute-mfcc-online-batched-cuda.cc compute-online-feats-cuda.cc compute-online-feats-batched-cuda.cc compute-fbank-feats-cuda.cc compute-mfcc-feats-cuda.cc apply-cmvn-online-cuda.cc >> .depend.mk
compute-mfcc-online-batched-cuda.cc:18:31: fatal error: cuda_profiler_api.h: No such file or directory
 #include <cuda_profiler_api.h>
                               ^
compilation terminated.
../makefiles/default_rules.mk:149: recipe for target 'depend' failed
make[1]: *** [depend] Error 1
make[1]: Leaving directory '/opt/kaldi/src/cudafeatbin'
make: *** [cudafeatbin/depend] Error 2
make: *** Waiting for unfinished jobs....
Makefile:127: recipe for target 'cudafeatbin/depend' failed
make[1]: Leaving directory '/opt/kaldi/src/featbin'
[...]

PR #3895, opened 2020-01-30
[src]add compile guard to enable  compilation even without CUDA enabled

jtrmal:
still testing it out


jtrmal:
Merging -- resolves #3893 and #3894

PR #3896, opened 2020-01-30
removing note in ivector comments and explaining magic constant

luitjens:


luitjens:
Dan,  it would be good to have someone review the logic in this kernel for calculating the prior scale change.  This code differs from online CPU code a bit and I left myself a note to track down why.  I think i understand it and left a comment.

Primarily we always have old_num_frames = 0 because we don't carrying linear/quadratic from frame to frame like the CPU version does.  

Things worth validating in this:
1) Is it correct to set num_old_frames = 0
2) We don't compute tot_post because we don't filter any points.  Thus it is always equal to number frames.
3) We scale the number of frames by posterior_scale because tot_post would have been scaled to that.
4)  Inititalizing prior_scale_change to 1.0f makes sense.  I was a little concerned that it should be posterior_scale but experimentally i don't think this is the case.

Thanks,
Justin

danpovey:
It looks OK although since there are no old stats if I were doing it I would have simplified it a bit.  But it's OK.
However you do seem to assume max_count_ > 0.  If not, I think it would give wrong answers.  You might at least want to throw an informative error if max_count_ is <= 0.

luitjens:
I'm not sure where I assume max_count != 0.  I have a branch checking that so we don't divide by it.  This is essentially the same code that i found in ivector-extractor.cc.  What am I missing?

Thanks,

jtrmal:
closeable?

PR #3904, opened 2020-02-04
Update decodable-matrix.h

zh794390558:


danpovey:
thanks... it would have been better to combine both PRs into one, but I will merge.

PR #3906, opened 2020-02-04
wsj align_si.sh fail when analyze_alignments.sh failed

boeddeker:
Is there a guideline, when Kaldi shouldn't fail, when a subcommand failed?

I ran the CHiME6 eg and `analyze_alignments.sh` failed, but the eg continued.

I am not sure if the `run.sh` should fail when `steps/diagnostic/analyze_alignments.sh` failed.

### DIFF
--- a/egs/wsj/s5/steps/align_si.sh
+++ b/egs/wsj/s5/steps/align_si.sh
@@ -101,6 +101,6 @@ else
       "$feats" "ark,t:|gzip -c >$dir/ali.JOB.gz" || exit 1;
 fi
 
-steps/diagnostic/analyze_alignments.sh --cmd "$cmd" $lang $dir
+steps/diagnostic/analyze_alignments.sh --cmd "$cmd" $lang $dir || exit 1;
 
 echo "$0: done aligning data."


danpovey:
analyze_alignments.sh sometimes fails for harmless reasons, and is purely diagnostic, so I'd rather leave it as it is.  But thanks!

boeddeker:
Ah, ok.

I got the error
```
gzip: >: No such file or directory
gzip: exp/mono_ali/phone_stats.1.gz: No such file or directory
```
for each sub-job and thought this is an error.

danpovey:
That's definitely unexpected.  Check what /bin/sh points to.  It should be
bash, for Kaldi to work correctly.

boeddeker:
Thanks, for your feedback. Then I have to investigate what went wrong.
The file does not exist, so maybe the error occurred earlier.

My `/bin/sh` is bash:
```
kaldi/egs/chime6/s5_track1$ /bin/sh --help
GNU bash, Version 4.2.46(2)-release-(x86_64-redhat-linux-gnu)
```

boeddeker:
Just for documentation:
I caused the error, because I replaced `queue.pl` with a self-written script for openccs (Special HPC scheduler).
The self-written script parsed the input wrongly (This was the first test of the script).
So no one else will get the same error.

PR #3908, opened 2020-02-04
[src] make cudafeatbin optional

jtrmal:
I'm still working on it (will have to test with cuda), but any feedback welcomed from @danpovey  and @luitjens

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -70,7 +70,7 @@ Configuration options:
   --static              Build and link against static libraries [default=no]
   --shared              Build and link against shared libraries [default=no]
   --use-cuda            Build with CUDA [default=yes]
-  --without-cudadecoder Build without CUDA decoder [default=no]
+  --with-cudadecoder    Build with CUDA decoder [default=yes]
   --cudatk-dir=DIR      CUDA toolkit directory
   --cuda-arch=FLAGS     Override the default CUDA_ARCH flags. See:
          https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#nvcc-examples.
@@ -744,7 +744,7 @@ debug_level=1
 double_precision=false
 dynamic_kaldi=false
 use_cuda=true
-without_cudadecoder=0
+with_cudadecoder=1
 static_fst=false
 static_math=false
 threaded_atlas=false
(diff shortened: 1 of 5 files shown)


danpovey:
Great!
@jtrmal, it looks to me like cudadecoder/ and cudadecoderbin/ now depend on cudafeat, so I think it might be better to just make all the cuda decoding-related stuff optional.

src/cudadecoder/Makefile
```diff
@@ -4,6 +4,7 @@ EXTRA_CXXFLAGS = -Wno-sign-compare
 include ../kaldi.mk
 
 ifeq ($(CUDA), true)
+ifeq ($(WITH_CUDAFEAT), 1)
```
danpovey: It might be better to do `ifneq ($(WITH_CUDAFEAT), 0)` since that would behave like the current Makefile behaves for already-configured setups.

src/configure
```diff
@@ -818,6 +819,15 @@ do
   --use-cuda=no)
     use_cuda=false;
     shift ;;
+  --with-cudafeat)
```
danpovey: We should perhaps call this --with-cuda-decoding since that is the main reason people would want to use that part of the codebase.. might be less confusing.

jtrmal:
Yeah... I didn't understand the scope in the beginning. Will rename.
Y.

jtrmal:
updated... By default the decoder is built. I believe the configure version does not have to be upgraded as the changes should be compatible with the older configuration

src/configure
```diff
@@ -70,6 +70,7 @@ Configuration options:
   --static              Build and link against static libraries [default=no]
   --shared              Build and link against shared libraries [default=no]
   --use-cuda            Build with CUDA [default=yes]
+  --without-cudadecoder Build without CUDA decoder [default=no]
```
danpovey: sorry, I don't like double negatives, can we make this --with-cudadecoder=yes by default?

jtrmal: No problem, but with-feature-x and without-feature-x are completely common flags in configures...

jtrmal:
updated and tested

src/cudadecoder/Makefile
```diff
@@ -26,9 +24,19 @@ ADDLIBS = ../cudamatrix/kaldi-cudamatrix.a ../base/kaldi-base.a ../matrix/kaldi-
...
+
+
 # Implicit rule for kernel compilation
 %.o : %.cu
 	$(CUDATKDIR)/bin/nvcc -c $< -o $@ $(CUDA_INCLUDE) $(CUDA_FLAGS) $(CUDA_ARCH) -I../ -I$(OPENFSTINC)
+
+else
+all:
+		$(warning "Not building cudadecoder extension -- to build with it, configure with --with-cudadecoder[=true]")
+endif
```
danpovey: can you try to check that `depend` is not run in this case also?

jtrmal:
I'm not sure I understand. The rule should be executed but because the
variables are empty, there should be no dependencies generated.
But I will check.
Y.

danpovey:
still says WIP

jtrmal:
I checked the process for .depend.mk and the files will be always the same -- the dependencies are generated from a wildcard. Only the targets will not get run during make (they will not exist). 
Can that remain as it is?

danpovey:
sure!  you can merge yourself if you think it's ready.. maybe remove the WIP first though.

jtrmal:
Merging.

PR #3910, opened 2020-02-04
WIP [src] Ingnore most of the cuda dirs if cuda is  not configured

jtrmal:
This is a second version ignoring all the directories when the CUDA is not configured.
The patch is smaller, but  for the other approach, the similar patch could be devised, i.e. the size of the pach is not a subject of discussion -- I'm merely testing idfferent ways how to do this

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -4,16 +4,15 @@
 
 SHELL := /bin/bash
 
-SUBDIRS = base matrix util feat cudafeat tree gmm transform \
+SUBDIRS = base matrix util feat tree gmm transform \
           fstext hmm lm decoder lat kws cudamatrix nnet \
-          bin fstbin gmmbin fgmmbin featbin cudafeatbin \
+          bin fstbin gmmbin fgmmbin featbin \
           nnetbin latbin sgmm2 sgmm2bin nnet2 nnet3 rnnlm chain nnet3bin nnet2bin kwsbin \
-          ivector ivectorbin online2 online2bin lmbin chainbin rnnlmbin \
-          cudadecoder cudadecoderbin
+          ivector ivectorbin online2 online2bin lmbin chainbin rnnlmbin 
 
-MEMTESTDIRS = base matrix util feat cudafeat tree gmm transform \
+MEMTESTDIRS = base matrix util feat tree gmm transform \
           fstext hmm lm decoder lat nnet kws chain \
-          bin fstbin gmmbin fgmmbin featbin cudafeatbin \
+          bin fstbin gmmbin fgmmbin featbin \
           nnetbin latbin sgmm2 nnet2 nnet3 rnnlm nnet2bin nnet3bin sgmm2bin kwsbin \
           ivector ivectorbin online2 online2bin lmbin
 
@@ -28,6 +27,12 @@ EXT_SUBDIRS = online onlinebin  # python-kaldi-decoding
 EXT_SUBDIRS_LIB = $(filter-out %bin, $(EXT_SUBDIRS))
 
 include kaldi.mk
+
+ifeq ($(CUDA), 1)
+	SUBDIRS += cudafeatbin cudadecoderbin cudadecoder cudafeat
+	MEMTESTDIRS += cudafeatbin cudadecoderbin cudadecoder cudafeat
+else
+endif
 
 # Reset the default goal, so that the all target will become default
 .DEFAULT_GOAL :=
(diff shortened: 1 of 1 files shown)


jtrmal:
Again @danpovey  and @luitjens feel free to comment.
This one is not fully tested, it's more a sketch of the approach.

danpovey:
This would fix some of the issues but not all of them.. what I was thinking is making the GPU decoding stuff optional even if CUDA is installed, because lots of people will want to do normal neural net training, but there are extra compilation failures from the GPU decoding (which is used
by a smaller subset of users).

jtrmal:
Ok, that would make the previous pr preferable
Y.

luitjens:
I'd like to step back and make sure what is being proposed actually solves the various issues.  First let's discuss the types of failures and their causes.

1)  make depend failures:  caused by running make depends on cuda* directories when cuda is not enabled.  These directories have includes that don't exist.
     Fixes:  A) #if HAVE_CUDA all code B) don't run make depends on cuda* directories when cuda is not enabled.
      Out of these B seems the most robust and something we should do.

2)  avx512 failure:  Due to GCC 5.4 compatibility issue with ICC/MKL
     Fixes:  A)  disable cudafeats build unless requested.  B) don't allow GCC 5.4 with ICC/MKL in configure.
      A) doesn't actually fix anything.  It hides the bug for some users but a change in CudaMatrix may expose the same issue in the future.  Users who want cudafeat would still hit this same issue.  In addition, in the future NVIDIA may look into optimizing the training path which would potentially mean using cudafeat outside of inference.  I think A is a work around that doesn't fix anything.  B) is better.

3)  Using new CUDA features that are not compatible with older toolkits.   This is likely going to be an issue as we move forward and take advantage of new hardware and features of CUDA.  
     Fixes:   A) #if CUDA_VERSION newer features and throw asserts if users try to use newer code. B) require newer toolkits.   C) Disable certain cuda directories unless they have a newer toolkit version.
     A) is robust but new code will expose bugs that need to be fixed as we don't test every combination of CUDA toolkits.  When developing we try to restrict usage of newer features to only when necessary.  But we shouldn't constrain ourselves to this entirely.   
     B) forcing this is hard but at some point we do need a minumum.  We should evaluate what the minimum is every year and bump up as necessary.   I'd like to understand why certain users are still on old toolkits and see if it is necessary or just an oversight.
     C)  This is the most robust.  What if we add a minum cuda version to the make files in each directory.  cudafeat would be 9010,  cudafeatbin would be 1000.  When the CUDA version is too old the compiling and make depends in that directory is disabled and maybe a note is printed during configure or a README is put into the directory.  This is robust and removes #if branches from the code.

So in summary.  My proposed fixes are:

1B,  2B,  3C.

These don't all need to be addressed in the same patch set and if we agree on this path i'd suggest implementing these as separate patch sets.

danpovey:
Here's a different proposal: how about we always attempt to build all
directories, but we add logic to the Makefile so that if we are not in the
travis build the cudadecoder and cudafeat directories don't kill the entire
build but are built *if possible*, and a suitable message is printed at the
end?
That way things get tested, but failures that may not matter to the user
don't hold up other things.

luitjens:
wouldn't this mean that make would always return success in those
directories even if they have bad code?  I much prefer make to return an
error when an error exists and we handle those errors.

danpovey:
People will likely still report it on the list anyway.
I don't want people's use of Kaldi to be prevented by errors in these
features that are only used by power-users.

luitjens:
That is also what i'm trying to do.  I'd rather fix the underlying issues
than ignore them though.   In my opinion from personal experience with
other software ignoring building errors in the build process leads to many
headaches.  I've seen this done with bad build systems in the past and it
often leads to cases where things appear to build correctly but you get
obscure corruptions as your build is inconsistent.  I suspect if you do
this you will actually increase the email traffic from users with problems
but the problems will be much harder to figure out because you will get
messages like.  I run X but it crashes with some undefined behavior.

danpovey:
How about a compromise: we make it so the cudafeat stuff is built by
default but have an option that can turn that off,
so that people can still build if they have those gcc 5.4 issues, for
instance.

Dan

jtrmal:
that sounds ok to me, but it depends how Justin feels about it, I guess...
y.

luitjens:
That is fine.  This is better as it at least has people report when their
are build issues so that we can fix but gives them a quick work around.

jtrmal:
I'm closing this, as the changes have been done to the other part.

PR #3918, opened 2020-02-07
bug fix

hajime9652:
```
choose_utts_to_combine.py: combined 994636 utterances to 780401 utterances while respecting speaker boundaries, and then to 780398 utterances with merging across speaker boundaries.
  File "<string>", line 25
    print uniq, uniq2orig_uniq[uniq]
             ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(uniq, uniq2orig_uniq[uniq])?
```

### DIFF
--- a/egs/wsj/s5/utils/data/combine_short_segments.sh
+++ b/egs/wsj/s5/utils/data/combine_short_segments.sh
@@ -158,7 +158,7 @@ while changed:
                  changed = True
 
 for uniq in sorted(uniq2orig_uniq.keys()):
-    print uniq, uniq2orig_uniq[uniq]
+    print(uniq, uniq2orig_uniq[uniq])
 ' > $dir/uniq_to_orig_uniq
   rm $dir/uniq_sets
 


egs/wsj/s5/utils/data/combine_short_segments.sh
```diff
@@ -158,7 +158,7 @@ while changed:
                  changed = True
 
 for uniq in sorted(uniq2orig_uniq.keys()):
-    print uniq, uniq2orig_uniq[uniq]
+    print(uniq, uniq2orig_uniq[uniq])
```
danpovey: Can you please also change the python invocation to python3?  Otherwise this will break systems with python2 as the default.

hajime9652: I used python3 and got this error. Kaldi require python2 as the default?
Or please tell me the script instead for this.

danpovey: I was talking about line 137, change python to python3

hajime9652: I explicitly changed python to python3 and got the same error.
Should I use python2?

hajime9652: I ran this by python2.
In this case, it might be due to using conda python. I try to inspect this error.
If I clear this error,  I do another PR.
thx.

danpovey:
What I meant was that you should keep your other code change but change
python to python3.

hajime9652:
> What I meant was that you should keep your other code change but change python to python3.

I'm sorry it's my misunderstanding.

danpovey:
Please still make the pull request: it was a bug as we require Kaldi to
work whether python is python2 or python3.

hajime9652:
@danpovey 
I made new PR. please check #3926 
I could not reopen this PR, sorry for my confusing.

PR #3924, opened 2020-02-11
[src] add compile guards in case cuda is not configured

jtrmal:


jtrmal:
compile went through, merging...

PR #3926, opened 2020-02-12
Kaldi to work whether python is python2 or python3.

hajime9652:
#3918


danpovey:
This is still  not right.  I told you what was needed: your ORIGINAL PATCH plus change `python` at line 113 (IIRC) to `python3`.  I don't have time to explain why that is what's needed, but it is.

hajime9652:
Can you please make sure that everything is correct?

danpovey:
good now, thanks!

hajime9652:
Thank you for your support as ever.

PR #3927, opened 2020-02-12
fix incremental decoder bug

naxingyu:
@chenzhehuai 
not sure if this is what you meant?

### DIFF
--- a/src/decoder/lattice-incremental-online-decoder.cc
+++ b/src/decoder/lattice-incremental-online-decoder.cc
@@ -98,7 +98,7 @@ typename LatticeIncrementalOnlineDecoderTpl<FST>::BestPathIterator LatticeIncrem
     // it a fatal error for now.
     KALDI_WARN << "No final token found.";
   }
-  if (final_cost_out == NULL)
+  if (final_cost_out != NULL)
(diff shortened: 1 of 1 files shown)


naxingyu:
oh, I should @danpovey since you wrote this.

src/decoder/lattice-incremental-online-decoder.cc
```diff
@@ -98,7 +98,7 @@ typename LatticeIncrementalOnlineDecoderTpl<FST>::BestPathIterator LatticeIncrem
     // it a fatal error for now.
     KALDI_WARN << "No final token found.";
   }
-  if (final_cost_out == NULL)
+  if (final_cost_out != NULL)
     *final_cost_out = best_final_cost;
   return BestPathIterator(best_tok, this->NumFramesDecoded() - 1);
 }
```
danpovey: yes but please do the same for lattice-incremental-decoder.cc

naxingyu: I do not see this error in lattice-incremental-decoder.cc, where exactly?

danpovey: oh sorry, it's just here.

PR #3930, opened 2020-02-13
WIP: [build] Build binary directories via cmake too.

franck-exceenis:


danpovey:
Why add `add_library(kaldi-fst ALIAS kaldi-fstext)` (I don't see other
add_library invocations in
CMakeLists.txt).

franck-exceenis:
This is because `kaldi-fst` doesn't exist and `add_library(kaldi-fst ALIAS kaldi-fstext)` only create an alias for it. But actually I discovered that one executable actualy requires `kaldi-decoder` because it needs `fst::GrammarFst` class.
The other way to do the job is to rename de directory `fstbin` into `decoderbin`, but I don't know the impacts over the other way to build (with configure script). This way is the smallest change required to make it work.

franck-exceenis:
I looked at the python script (`cmake/gen_cmake_skeleton.py:252`) that generates the `CMakeLists.txt` in `fstbin` directory and it links every executable with `kaldi-fst` being `kaldi-` plus the directory name removing `bin`.
I successfully built the project on several architectures.
On raspberry, for example, you can try it using:
```
cmake -DCMAKE_BUILD_TYPE:STRING=Release -DBLA_PREFER_PKGCONFIG:BOOL=TRUE -DLAPACK_openblas_LIBRARY:PATH=/usr/lib/arm-linux-gnueabihf/liblapack.so -DKALDI_BUILD_TEST:BOOL=FALSE ..
make
```

franck-exceenis:
Maybe a cmake build should be added in the CI to check this. I don't know so much about Travis, but I can try to do it if you want

danpovey:
Sorry I think I'll need to hear from @cloudhan or @galv or someone else who understands the CMake build before merging this.  It doesn't look right to me, and I don't see why fstext has to be called fst.  What problem was this intending to fix?

danpovey:
.. and regarding adding a CMake build to Travis: I'd definitely consider merging it.  I think Travis is already running 2 things in parallel so maybe a 3rd would be allowed.  @jtrmal might know if our plan supports it.

jtrmal:
that change does not make sense to me neither -- what are you trying to fix?
I think a third build would be done after one of those previous ones finishes

franck-exceenis:
It just fixes that, on master, you can't build (via cmake) the executables in `fstbin` directory. Is there a way I don't see to do it? Please let me know.

jtrmal:
I'll look into it -- you well might be right, it's just the suggested
solution is very surprising
y.

cloudhan:
@franck-exceenis you can add additional dependencies to a specific target simply by adding it to the dictionary in https://github.com/kaldi-asr/kaldi/blob/793191be209357cd22dd20f4958a570098ac0cf8/cmake/gen_cmake_skeleton.py#L49-L51
Add decoder to the dict and use `add_library(kaldi-fst ALIAS kaldi-fstext)` should fix your problem.

danpovey:
I don't understand why the ALIAS thing is needed.  What is different about
the 'fstext' dir and why is it not just called 'kaldi-fstext' ?

cloudhan:
@danpovey OK, let me explain it for you.

In the CMakeLists.txt python generator `gen_cmake_skeleton.py`

1. For a library directory. e.g., `base`, `decoder`, will generate cmake library targets named `kaldi-base`, `kaldi-decoder` respectively. So in this case `fstext` will result a `kaldi-fstext`. 
2. For a executable in `*bin/` folder, e.g., `featbin/apply-cmvn.cc` the executable target name will be the filename `apply-cmvn` it will depend on a primary library `kaldi-feat`.

The reason for this is that we can simply parse the dependencies for a library by reading the Makefile. Then when adding executable targets, by linking with the primary library (this actually introduce a dependency tree by the cmake target system), we solve most dependency problem. But there are exceptions, there might be some additional dependencies, I solve them by adding to `get_exe_additional_depends`.

In this case, the folder `fstbin`, have many executable targets created, and they all link to the target `kaldi-fst` which dose not exist. To solve it, there are 3 viable methods I can come up with (maybe more):
1. rename `fstbin` to `fstextbin`
2. `add_library(kaldi-fst ALIAS kaldi-fstext)`
3. write some code in `gen_cmake_skeleton.py`
and the second one is just simple.

danpovey:
Hm.
I don't see it as a general rule that a XXXbin directory depends on an XXX
directory.  I think it would be easier to parse the Makefile in src/ to get
the dependencies.  Yenda wrote a script like that to generate the Windows
build files.

cloudhan:
I know that, but then we lose the fine grain control. 
say `XXXbin/foo.cc` depends on `kaldi-XXX` and XXXbin/bar.cc depends on `kaldi-XXX`, `kaldi-decoder`, `kaldi-feat`, `kaldi-gmm`

In the current setting, there will be no redundant dependency for `foo`. In your suggestion, both `foo` and `bar` will have to depends on `kaldi-decoder`, `kaldi-feat`, `kaldi-gmm`.

danpovey:
So how are you finding those fine-grained dependencies?  And why is that
method not already sufficient?  I.e. why do you need to add the rule that
XXXbin depends on XXX?

franck-exceenis:
@cloudhan the dictionary thing worked. I pushed it.

franck-exceenis:
Actually I found almost all the binary directories are not built via cmake.
So I rename the pull request to fix all of them.

cloudhan:
@danpovey  by adding them to the Python dict I mentioned in previous comments

CMakeLists.txt
```diff
@@ -213,6 +213,10 @@ if(CUDA_FOUND)
     add_subdirectory(src/cudadecoderbin)
 endif()
 
+add_library(kaldi-fst ALIAS kaldi-fstext)
```
danpovey: I don't like adding an alias.  It will be confusing IMO.  Why not explicitly add the dependency that fstbin depends on fstext?

franck-exceenis: I agree, I think the best solution is actually to not generate these `CMakeLists.txt` and add them for versionning so we can explicitly write the dependencies for each target. Is there a problem with this?

galv: I think that they wanted the Makefiles to be the source of truth, so that cmake and make don't become out-of-sync. Of course, that is a complicated endeavour.

cloudhan: Original design was to checkin all generated `CMakeLists.txt`s and remove the generator afterward, aka, make the cmake build system first class. But it will introduce maintenance burden. So for now, the generator is kept and `CMakeLists.txt`s are generated on the fly.

cloudhan:
@franck-exceenis I managed to add all bin directories to cmake. Could you try [cmake-add-bins](https://github.com/cloudhan/kaldi/tree/cmake-add-bins) branch out?

cloudhan:
I think that is specific to kaldi, I never see other project produce as many executables as kaldi does. Even large project like llvm only generate ~90 exe and kaldi have ~500 now. Maintaining such large amount of executable targets will certainly be a problem.

@franck-exceenis master branch should build all binary directories via cmake now.

danpovey:
I am assuming this is made unnecessary by #3935

franck-exceenis:
Ok, thank you

PR #3935, opened 2020-02-17
add missing bin directories to cmake build system

cloudhan:
Almost all bin directories are missing from cmake as mentioned in #3930 . But simple changes won't fix it. Update the generator to fix it.


danpovey:
Thanks!  Merging.

PR #3936, opened 2020-02-17
fixed a bug in TraceBackBestPath()

mahsa7823:
Resolve difference between GetBestPath() in online decoder and CompactLatticeShortestPath().

### DIFF
--- a/src/decoder/lattice-faster-online-decoder.cc
+++ b/src/decoder/lattice-faster-online-decoder.cc
@@ -135,24 +135,28 @@ typename LatticeFasterOnlineDecoderTpl<FST>::BestPathIterator LatticeFasterOnlin
   Token *tok = static_cast<Token*>(iter.tok);
   int32 cur_t = iter.frame, ret_t = cur_t;
   if (tok->backpointer != NULL) {
+    BaseFloat best_graph_cost = std::numeric_limits<BaseFloat>::infinity();
     ForwardLinkT *link;
     for (link = tok->backpointer->links;
          link != NULL; link = link->next) {
       if (link->next_tok == tok) { // this is the link to "tok"
-        oarc->ilabel = link->ilabel;
-        oarc->olabel = link->olabel;
-        BaseFloat graph_cost = link->graph_cost,
-            acoustic_cost = link->acoustic_cost;
-        if (link->ilabel != 0) {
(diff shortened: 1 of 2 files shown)


src/decoder/lattice-faster-online-decoder.cc
```diff
@@ -135,24 +135,28 @@ typename LatticeFasterOnlineDecoderTpl<FST>::BestPathIterator LatticeFasterOnlin
...
          link != NULL; link = link->next) {
       if (link->next_tok == tok) { // this is the link to "tok"
-        oarc->ilabel = link->ilabel;
-        oarc->olabel = link->olabel;
-        BaseFloat graph_cost = link->graph_cost,
-            acoustic_cost = link->acoustic_cost;
-        if (link->ilabel != 0) {
-          KALDI_ASSERT(static_cast<size_t>(cur_t) < this->cost_offsets_.size());
-          acoustic_cost -= this->cost_offsets_[cur_t];
-          ret_t--;
```
dophist, line 149: @mahsa7823 @danpovey 
might be a bug here: the for loop may hit "ret_t--" multiple times, but single call of traceback() shouldn't push back the timestamp more than once. (0 for non-emitting frame, 1 for emitting frame, in theory). The corrupted timestamp will trigger the above ASSERTION failure in some cases, which may further crash the decoder during runtime.

src/decoder/lattice-faster-online-decoder.cc
```diff
@@ -135,24 +135,28 @@ typename LatticeFasterOnlineDecoderTpl<FST>::BestPathIterator LatticeFasterOnlin
...
+          }
+          oarc->weight = LatticeWeight(graph_cost, acoustic_cost);
+          best_graph_cost = graph_cost;
         }
-        oarc->weight = LatticeWeight(graph_cost, acoustic_cost);
-        break;
       }
     }
-    if (link == NULL) { // Did not find correct link.
+    if (link == NULL &&
```
dophist, line 158: The post condition of the above "for loop termination" ensures (link == NULL), so here the first condition seems to be redundent.

src/decoder/lattice-faster-online-decoder.cc
danpovey: Oops.  Can you please make a fix?

dophist: No problem, I'll create a PR this week.

PR #3937, opened 2020-02-18
fixed a bug; print at stage14

SunSki:
I made it available in python3

### DIFF
--- a/egs/swbd/s5c/local/chain/multi_condition/run_tdnn_aug_1a.sh
+++ b/egs/swbd/s5c/local/chain/multi_condition/run_tdnn_aug_1a.sh
@@ -128,7 +128,7 @@ if [ $stage -le 14 ]; then
   echo "$0: creating neural net configs using the xconfig parser";
 
   num_targets=$(tree-info $treedir/tree |grep num-pdfs|awk '{print $2}')
-  learning_rate_factor=$(echo "print 0.5/$xent_regularize" | python)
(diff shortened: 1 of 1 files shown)


egs/swbd/s5c/local/chain/multi_condition/run_tdnn_aug_1a.sh
```diff
@@ -128,7 +128,7 @@ if [ $stage -le 14 ]; then
   echo "$0: creating neural net configs using the xconfig parser";
 
   num_targets=$(tree-info $treedir/tree |grep num-pdfs|awk '{print $2}')
-  learning_rate_factor=$(echo "print 0.5/$xent_regularize" | python)
+  learning_rate_factor=$(echo "print(0.5/$xent_regularize)" | python3)
```
danpovey: this works the same in python2 and python3, I don't see why the change would be needed.

SunSki: This error was happened in my case, so I changed like this
----------
  stage14
----------
run_tdnn_aug_ernie_a.sh: creating neural net configs using the xconfig parser
tree-info exp/chain/tri5_7d_tree_aug/tree
  File "<stdin>", line 1
    print 0.5/0.1
            ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(0.5/0.1)?

danpovey:
Oh the parens were missing.
Actually the change is OK, Kaldi already depends on python3.

PR #3945, opened 2020-02-19
Add a new binary for chain e2e

YiwenShaoStephen:
Add a new binary: chainbin/chain-make-num-fst-e2e
It is used to converts chain e2e numerator fst (containing transition-ids) to fst (containing pdf-ids,
zero-based, eps-removed, and normalized by the normalization fst).
This step is usually in prepare_egs stage while this binary would be necessary for other outside toolkits (e.g. pychain) that directly use fst instead of egs.

### DIFF
--- a/src/chainbin/chain-make-num-fst-e2e.cc
+++ b/src/chainbin/chain-make-num-fst-e2e.cc
@@ -1,0 +1,121 @@
+// chainbin/chain-make-num-fst-e2e.cc
+
+// Copyright 2019  Johns Hopkins University (author: Yiwen Shao)
+
+// See ../../COPYING for clarification regarding multiple authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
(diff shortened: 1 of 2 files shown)


src/chainbin/chain-make-num-fst-e2e.cc
```diff
@@ -0,0 +1,121 @@
+// chainbin/chain-make-num-fst-e2e.cc
+
+// Copyright 2019  Johns Hopkins University (author: Yiwen Shao)
```
danpovey: Just put Yiwen Shao as the author as you are not an employee (or if you have a part-time job with Hopkins, at least you are not doing this in that status)

src/chainbin/chain-make-num-fst-e2e.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "base/kaldi-common.h"
+#include "gmm/am-diag-gmm.h"
+#include "hmm/transition-model.h"
+#include "hmm/hmm-utils.h"
+#include "util/common-utils.h"
+#include "fst/fstlib.h"
+
+namespace kaldi {
+
+bool FstTransitionToPdf(const fst::StdVectorFst &fst_transition,
```
danpovey: I think it should be .. ToPdfPlusOne(), since these are pdf+1.  Also I think this conflicts with your comment or commit message in which you say they are zero-based, these seem to be one-based.

src/chainbin/chain-make-num-fst-e2e.cc
```diff
@@ -0,0 +1,121 @@
...
+
+}
+
+int main(int argc, char *argv[]) {
+  using namespace kaldi;
+  typedef kaldi::int32 int32;
+  try {
+    const char *usage =
+      "Converts chain e2e numerator fst (containing transition-ids) to fst (containing pdf-ids, \n"
+      "zero-based, eps-removed, and normalized by the normalization fst) \n"
```
danpovey: here is where it says zero-based.  Also this doesn't remove epsilons, it just checks that there are no epsilons.. might be good to clarify.   should say pdf-ids + 1.

YiwenShaoStephen:
Thanks for the comments! I've made modifications according to it.

PR #3946, opened 2020-02-20
Update am-diag-gmm.h

zh794390558:


danpovey:
I'll leave this open in case you find any more typos in the next couple of weeks; I prefer to batch up  such fixes rather than making lots of separate commits.

zh794390558:
@danpovey ok.

PR #3940, opened 2020-02-19
add PyTorch's DistributedDataParallel training.

csukuangfj:
support distributed training across multiple GPUs.

TODOs:

- there are lots of code duplicates

----

Part of the training log

```
2020-02-19 13:55:10,646 INFO [ddp_train.py:160] Device (1) processing 1100/4724(23.285351%) global average objf: -0.225449 over 6165760.0 frames, current batch average objf: -0.130735 over 6400 frames, epoch 0
2020-02-19 13:55:55,251 INFO [ddp_train.py:160] Device (0) processing 1200/4724(25.402202%) global average objf: -0.216779 over 6732672.0 frames, current batch average objf: -0.123979 over 3840 frames, epoch 0
2020-02-19 13:55:55,252 INFO [ddp_train.py:160] Device (1) processing 1200/4724(25.402202%) global average objf: -0.216412 over 6738176.0 frames, current batch average objf: -0.132368 over 4736 frames, epoch 0
```

The training seems working.

PR #3947, opened 2020-02-20
WIP: begin to add CTC training with kaldi pybind and PyTorch.

csukuangfj:
working in progress.


egs/aishell/s10b/local/convert_text_to_labels.py
```diff
@@ -0,0 +1,156 @@
...
+import os
+
+import kaldi
+
+
+def get_args():
+    parser = argparse.ArgumentParser(description='convert text to labels')
+
+    parser.add_argument('--lexicon-filename', dest='lexicon_filename', type=str)
+    parser.add_argument('--tokens-filename', dest='tokens_filename', type=str)
```
danpovey: Please use the standard OpenFST symbol-table format for these tokens.
I'm open to other opinions, but since we'll probably have these symbols present in FSTs I think symbol 0 should be reserved for <eps> and <blk> should be 1, and we can just apply an offset of 1 when interpreting the nnet outputs.

egs/aishell/s10b/local/convert_text_to_labels.py
```diff
@@ -0,0 +1,156 @@
...
+                # we choose only the first one and drop other alternatives
+                lexicon[word] = phones
+
+    return lexicon
+
+
+def read_tokens(filename):
+    '''
+    Returns:
+        a dict whose keys are phones and values are phone indices
```
danpovey: Should clarify the format of the file.

egs/aishell/s10b/local/token_to_fst.py
```diff
@@ -0,0 +1,61 @@
...
+
+import argparse
+import os
+
+
+def get_args():
+    parser = argparse.ArgumentParser(
+        description='convert tokens.txt to tokens.fst')
+
+    parser.add_argument('--tokens-txt-filename',
```
danpovey: Please clarify what the input is.  I.e. what the format of tokens.txt is.

egs/aishell/s10b/local/convert_text_to_labels.py
danpovey: ... if the format is already the symbol-table format, bear in mind that the order of lines is actually arbitrary;what matters is the integer there.

egs/aishell/s10b/local/convert_text_to_labels.py
```diff
@@ -0,0 +1,156 @@
...
+
+    parser.add_argument('--lexicon-filename', dest='lexicon_filename', type=str)
+    parser.add_argument('--tokens-filename', dest='tokens_filename', type=str)
+    parser.add_argument('--dir', help='input/output dir', type=str)
+
+    args = parser.parse_args()
+
+    assert os.path.isfile(args.lexicon_filename)
+    assert os.path.isfile(args.tokens_filename)
+    assert os.path.isfile(os.path.join(args.dir, 'text'))
```
danpovey: There needs to be much more clarity on what this program is supposed to do, either as an extended
command or in the usage message.  It needs to be clear what the file formats are, for instance.
An example might be the easiest way.

egs/aishell/s10b/local/convert_text_to_labels.py
csukuangfj: I reuse the notation from EESEN (https://github.com/srvk/eesen), which calls
`phones.txt` as `tokens.txt`.

`tokens.txt` is acutally a phone symbol table, with

```
<eps> 0
<blk> 1
other phones
```

The code here does not pose any constraint on the order of lines. What
matters here is only the integer of symbols. The first two integers `0` and `1`
are reserved. I think `0` is reserved for `<eps>`. Here I reserve `1` for
the blank symbol.

The script generating `tokens.txt` has considered the above constraint.

---

 Since there is a `T` in `TLG.fst`, I keep using `tokens.txt` here instead
of `phones.txt`.  I can switch to `phones.txt` if you think that is more natural
in kaldi.

egs/aishell/s10b/local/convert_text_to_labels.py
csukuangfj: thanks. I will add more documentation.

egs/aishell/s10b/local/convert_text_to_labels.py
csukuangfj: I will write example usage in the documentation to illustrate the purpose of the program.

csukuangfj:
I am intending to use Baidu's warp-ctc (https://github.com/baidu-research/warp-ctc)
or PyTorch's builtin CTCLoss (https://pytorch.org/docs/stable/nn.html#torch.nn.CTCLoss).

Both of them do not support words with multiple pronunciations. I currently use only
the pronunciation of a word when it first appears and ignore other alternative pronunciations.

Can we first implement a baseline that considers only one pronunciation? 
Since this approach is the easiest one and we can reuse existing APIs to compute ctc loss.

danpovey:
Sure, implementing a baseline is fine.

egs/aishell/s10b/local/token_to_fst.py
csukuangfj: Fixed in the newly added documentation.

egs/aishell/s10b/local/convert_text_to_labels.py
csukuangfj: Fixed.

csukuangfj:
The loss drops from 83 to 4.8 after 100 batches and stops decreasing. I am trying
to find out the reason.

danpovey:
For the CTC system you'll normally want to  end with an LSTM layer so it
can make the output spiky.

csukuangfj:
@danpovey thanks


My current network architecture is

```
layer0: input-batchnorm

layer1: lstm
layer2: projection + tanh

layer3: lstm
layer4: projection + tanh

layer5: lstm
layer6: projection + tanh

layer7: lstm
layer8: projection + tanh

layer9: prefinal-affine
layer10: log-softmax
```

Do you mean I should replace `layer9: prefinal-affine` with an LSTM layer ?

danpovey:
No it should be OK as long as there are LSTM layers in there..  You could
try removing the prefinal layer though.
You may have to play with the l2 and learning rates a bit.

csukuangfj:
@danpovey 
do we need to **normalize** the coefficients?

That is, to replace
- [-1, 0, 1]
- [1, 0, -2, 0, 1]

with

- [-0.5, 0, 0.5]
- [0.25, 0, -0.5, 0, 0.25]

danpovey:
Normalization is not necessary because this component would typically be
followed by a batchnorm component.
BTW, regarding the Conv1d-type layout, make sure it is documented.

csukuangfj:
Still working in progress.

Now if you use only 8 wave files for training and decoding, you can
get a CER as low as 0.04, which verifies that the pipleline is working.

I find that the LSTM model is difficult to converge. I'm going to replace it with TDNN-F.

csukuangfj:
Decode results for the current pullrequest are as follows

| | this pullrequest| haowen's PyTorch(#3925) | fanlu's PyTorch (#3940) with online cmvn | haowen's kaldi (#3925)|
|--|---|---|---|---|
|test cer| 12.91 | 7.86 | 7.31 | 7.08 |
|test wer|21.90 | 16.56 | 15.97 | 15.72|
|dev cer| 11.81 | 6.47 | 6.16 |5.99|
|dev wer| 20.46 | 14.45 | 14.01 |13.86|

The first column uses nearly the same TDNN-F model architecture as the remaining columns
except that it does not have `xent` regularizer. In addition, the first column uses CTC loss
instead of chain loss.

It takes about 59 minutes per epoch and the decode results are for the 12th epoch.
The ctc loss  value is `0.086`.

There is still a big gap in CER/WER of this pullrequet compared with the chain model.
I will add an LSTM layer before the output affine layer, add spectral augmentation,
and run the training again.

danpovey:
@csukuangfj  If you are implementing CTC-CRF and if I understand it correctly from this paper
http://oa.ee.tsinghua.edu.cn/~ouzhijian/pdf/ctc-crf.pdf
it is the same as LF-MMI except there is no context dependency and the self-loop (blank symbol) is shared between all phones, and there is no optional silence.  The current tree-building mechanism in Kaldi doesn't allow for one pdf to be shared and the other not; however, I do remember doing experiments with making the blank shared vs. not shared (I don't recall how) and not shared was a bit better.
It should be possible to use the same forward-backward code for both numerator and denominator, for CTC-CRF as LF-MMI.

csukuangfj:
@danpovey 

thanks. I am still in the learning mode and have read the implementation of Kaldi's
denominator computation. I find that the denominator part of CTC-CRF is adapted
from Kaldi's code. It takes more than 3 days for CTC-CRF for the AIShell dataset
to get a reasonable CER, which is too slow and unacceptable. I am trying to figure
out the reason and to reuse as much code from Kaldi as possible.

csukuangfj:
@danpovey 
I've read the denominator implementation of CTC CRF and find that it is
a re-implementation of the chain denominator computation with the following differences:

- It does not perform 100 iterations of the FST to get the initial probability for each state;
only the start states take the probability from the state weight; all other states have
initial probability 0;  The initial state probabilities are not normalized.

- it has no leaky hmm

- it performs the computation in the log space

Different from the chain training part, examples in the same batch of ctc training
do not have the same number of  frames, `DenominatorComputation` in `src/chain`
cannot be used directly for ctc training.

I would like to implement a class `CtcDenominatorCompuation` that can handle
examples with different sequence lengths in the same batch.

Would the above differences make a significant impact on the final training result?

danpovey:
@csukuangfj I don't think those things would make a significant difference.   Also I don't think that name is really appropriate.  Firstly it isn't CTC (the special characteristic of CTC is that it is normalized to 1 so needs no denominator computation).   Regarding the different sequence lengths... it would be interesting to allow the denominator computation to handle different sequence lengths (the numerator as well).  The difficulty is doing this efficiently.   GPU programming is only efficient if you can form suitable batches.   What I was thinking was, it might be better to still have batches with fixed-size elements, but instead focus on allowing the numerator computation to work with that.  If we had to break sentences into pieces, we could either handle it by constraining to an FST like Kaldi's current implementation of chain training, or use the regular CTC-type FST but allow all states to be initial and final, so the sequence can start and end in the middle.

csukuangfj:
I am trying to sort examples by their sequence lengths so that there is as less padding
as possible in the same batch.

-------

> What I was thinking was, it might be better to still have batches with fixed-size elements, but instead focus on allowing the numerator computation to work with that. 

I'm still learning the internals of chain training. We can come to this when the handling
of different sequence lengths is finished; at least, it can be used as a baseline.

PR #3932, opened 2020-02-16
Fixed Path to Sequitur Executable

Kadle11:
While working on VoxForge, I found that the path for sequitur executable is not ```$SEQUITUR/bin```, It is ```$SEQUITUR```.

PR #3948, opened 2020-02-20
[tools] use PYTHONUSERBASE in extras/install_sequitur.sh

jtrmal:
addresses #3932


danpovey:
@jtrmal please merge when done

PR #3950, opened 2020-02-22
fix bug in kws_search.sh for writing results 

thoshith-s:
results file are in gzip format. Unzip is required before doing operations on those files.


danpovey:
Merged but you should probably be using the more recent babel recipes (e.g. s5d).

jtrmal:
I agree -- please use s5d
y.

thoshith-s:
@jtrmal Thanks sure will do. But these bugs are common among all the babel recipes which I've mentioned in pr #3951.

jtrmal:
That might be true, even though I think some of them might be fixed. But it
would be better to fix the latest version of the scripts anyway.
y.

thoshith-s:
I've even updated latest version including s5d, which also includes the above bugs.

jtrmal:
Oh, great, I'll try to find time this week.
Y.

PR #3951, opened 2020-02-22
bug fixes for KWS setup in babel recipe, generation of confusion matr…

thoshith-s:
bug fixes for KWS setup in babel recipe 
1 . generation of confusion matrix (compute-wer compatibility issue) stats out is required.
2 . writing results local/kws_search.sh and utils/write_kwslist.pl

### DIFF
--- a/src/bin/compute-wer.cc
+++ b/src/bin/compute-wer.cc
@@ -25,6 +25,41 @@
 #include "tree/context-dep.h"
 #include "util/edit-distance.h"
 
+
+
+namespace kaldi {
+
+
+template<typename T>
+void PrintAlignmentStats(const std::vector<T> &ref,
+                         const std::vector<T> &hyp,
+                         T eps,
+                         std::ostream &os) {
+  // Make sure the eps symbol is not in the sentences we're aligning; this would
+  // not make sense.
+  KALDI_ASSERT(std::find(ref.begin(), ref.end(), eps) == ref.end());
+  KALDI_ASSERT(std::find(hyp.begin(), hyp.end(), eps) == hyp.end());
+
+  std::vector<std::pair<T, T> > aligned;
+  typedef typename std::vector<std::pair<T, T> >::const_iterator  aligned_iterator;
+
+  LevenshteinAlignment(ref, hyp, eps, &aligned);
+  for (aligned_iterator it = aligned.begin();
+       it != aligned.end(); ++it) {
+    KALDI_ASSERT(!(it->first == eps && it->second == eps));
+    if (it->first == eps) {
+      os << "insertion " << it->second << std::endl;
+    } else if (it->second == eps) {
+      os << "deletion " << it->first << std::endl;
+    } else if (it->first != it->second) {
+      os << "substitution " << it->first << ' ' << it->second << std::endl;
+    } else {
+      os << "correct " << it->first << std::endl;
+    }
+  }
+ }
+}
+
 int main(int argc, char *argv[]) {
   using namespace kaldi;
   typedef kaldi::int32 int32;
@@ -34,8 +69,9 @@ int main(int argc, char *argv[]) {
         "Compute WER by comparing different transcriptions\n"
         "Takes two transcription files, in integer or text format,\n"
         "and outputs overall WER statistics to standard output.\n"
+        "Optionally, the third argument can be used to obtain detailed statistics\n"
         "\n"
-        "Usage: compute-wer [options] <ref-rspecifier> <hyp-rspecifier>\n"
+        "Usage: compute-wer [options] <ref-rspecifier> <hyp-rspecifier> [<stats-out>]\n"
         "E.g.: compute-wer --text --mode=present ark:data/train/text ark:hyp_text\n"
         "See also: align-text,\n"
         "Example scoring script: egs/wsj/s5/steps/score_kaldi.sh\n";
@@ -54,13 +90,18 @@ int main(int argc, char *argv[]) {
 
     po.Read(argc, argv);
 
-    if (po.NumArgs() != 2) {
+    if (po.NumArgs() < 2 || po.NumArgs() > 3) {
       po.PrintUsage();
       exit(1);
     }
 
     std::string ref_rspecifier = po.GetArg(1);
     std::string hyp_rspecifier = po.GetArg(2);
+
+    Output stats_output;
+    bool detailed_stats = (po.NumArgs() == 3);
+    if (detailed_stats)
+      stats_output.Open(po.GetOptArg(3), false, false);  // non-binary output
 
     if (mode != "strict" && mode != "present" && mode != "all") {
       KALDI_ERR << "--mode option invalid: expected \"present\"|\"all\"|\"strict\", got "
@@ -95,6 +136,11 @@ int main(int argc, char *argv[]) {
       num_ins += ins;
       num_del += del;
       num_sub += sub;
+      
+      if (detailed_stats) {
+          const std::string eps = "";
+          PrintAlignmentStats(ref_sent, hyp_sent, eps, stats_output.Stream());
+      }
 
       num_sent++;
       sent_errs += (ref_sent != hyp_sent);
--- a/egs/babel/s5/local/kws_search.sh
+++ b/egs/babel/s5/local/kws_search.sh
@@ -147,7 +147,7 @@ if [ $stage -le 2 ]; then
     gunzip -c ${kwsoutdir}_LMWT/result.* \| \
       utils/write_kwslist.pl  --Ntrue-scale=$ntrue_scale --flen=0.01 --duration=$duration \
         --segments=$datadir/segments --normalize=true --duptime=$duptime --remove-dup=true\
-        --map-utter=$kwsdatadir/utter_map --digits=3 \
+        --map-utter=$kwsdatadir/utter_id --digits=3 \
         - ${kwsoutdir}_LMWT/kwslist.xml || exit 1
 fi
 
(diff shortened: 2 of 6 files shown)


danpovey:
Thanks... I think @jtrmal should check this.

thoshith-s:
Welcome. Below are the comments which are helpful in reviewing the above commit.
In babel recipe,
1.  local/generate_confusion_matrix.sh line number 67, compute-wer function is used which passes three arguments which leads to compatibility issue with current version of compute-wer

2. In local/kws_search.sh, results file are in gzip format. But the current version expects it to be in text format

3. In local/kws_search.sh, stage 2 and 3 map-utter is passed with utter_map file which is to be replaced with utter_id
     utter_map:
             <utterance-id> <utterance-id>
     utter-id:
             <utterance-id> <seq-id>  
             eg: 1-2000-890.wav 29

4. In utils/write_kwslist.pl, the utter mapping is read as <utterance-id> --> <seq-id> which should be 
<seq-id>  --> <utterance-id>
        `$utter_mapper{$col[0]} = $col[1];`
	... changed to
	`$utter_mapper{$col[1]} = $col[0];`

5. In utils/write_kwslist.pl, line 196-217,
`my @KWS;
		while (<$source>) {
		  chomp;
		  my @col = split(" ", $_);
		  @col == 5 || die "$0: Bad number of columns in raw results \"$_\"\n";
		  my $kwid = shift @col;
		  my $utter = $col[0];
		  my $start = sprintf("%.2f", $col[1]*$flen);
		  my $dur = sprintf("%.2f", $col[2]*$flen-$start);
		  my $score = exp(-$col[3]);

		  if ($segment) {
		    $start = sprintf("%.2f", $start+$tbeg{$utter});
		  }
		  if ($map_utter) {
		    my $utter_x = $utter_mapper{$utter};
		    die "Unmapped utterance $utter\n" unless $utter_x;
		    $utter = $utter_x;
		  }

		  push(@KWS, [$kwid, $utter, 1, $start, $dur, $score, ""]);
		}
`

changed to 
`my @KWS;
		while (<$source>) {
		  chomp;
		  my @col = split(" ", $_);
		  @col == 5 || die "$0: Bad number of columns in raw results \"$_\"\n";
		  my $kwid = shift @col;
		  my $utter = $col[0];
		  my $start = sprintf("%.2f", $col[1]*$flen);
		  my $dur = sprintf("%.2f", $col[2]*$flen-$start);
		  my $score = exp(-$col[3]);

		  if ($map_utter) {
		    my $utter_x = $utter_mapper{$utter};
		    die "Unmapped utterance $utter\n" unless $utter_x;
		    $utter = $utter_x;
		  }
		  
		  if ($segment) {
		    $start = sprintf("%.2f", $start+$tbeg{$utter});
		  }

		  push(@KWS, [$kwid, $utter, 1, $start, $dur, $score, ""]);
		}
`
Here, segment file is based on utterance-id not on the seq-id, when tried to get start time using seq-id, it will lead to error. 
Therefore first your seq-id is to be updated with utterance-id to continue processing.

kkm000:
@jtrmal, could you PTAL? Dan though you were the right person to review this.

jtrmal:
put on my todo

jtrmal:
yes, LGTM

src/bin/compute-wer.cc
```diff
@@ -25,6 +25,41 @@
 #include "tree/context-dep.h"
 #include "util/edit-distance.h"
 
+
+
+namespace kaldi {
+
+
+template<typename T>
+void PrintAlignmentStats(const std::vector<T> &ref,
```
jtrmal, line 34: I don't see this capability/function used in any external tool. Plus I'm sure we already do have this capability. So could you please remove this from this PR and eventually, if you think it's a "worthy" change, create a new PR, please?

kkm000: @jtrmal, could you please help us find the tool? Interesting, once I sent a PR to print similar data from the very same program at verbosity 2+, only in the raw form, and you then told me we already have that. I remember I could not find it.

jtrmal:
it's in steps/score_kaldi.sh

```
122     $cmd $dir/scoring_kaldi/log/stats1.log \
123       cat $dir/scoring_kaldi/penalty_$best_wip/$best_lmwt.txt \| \
124       align-text --special-symbol="'***'"
ark:$dir/scoring_kaldi/test_filt.txt ark:- ark,t:- \|  \
125       utils/scoring/wer_per_utt_details.pl --special-symbol "'***'" \|
tee $dir/scoring_kaldi/wer_details/per_utt \|\
126        utils/scoring/wer_per_spk_details.pl $data/utt2spk \>
$dir/scoring_kaldi/wer_details/per_spk || exit 1;
```

kkm000:
Looks like this is nearly there but abandoned. I'll take over, the changes seem trivial. @jtrmal, you ok with that?

jtrmal:
You know what?  Let me handle it. I was asked to revisit the recipes so I
will have to dig through my mess again anyway :)
y.

kkm000:
Bingo!

I'm trying to grok the difference between `en_US.UTF-8` and `C.UTF-8`. ATM the main dissimilarity is that I have the latter on my system but not the former, but I'm trying to bridge the gap...

jtrmal:
the differences, IIRC are in chr(127) and up... But I don't recall exactly.
Maybe currencies and stuff like that.

kkm000:
@jtrmal ah, I was thinking about a note on the different PR. This is bizarre: https://github.com/kaldi-asr/kaldi/pull/4130/files#diff-683416a3e05ed8af8425a06a31b0f2e4566d71dff67b3aae0f970427f834d58dR130

grep using the C locale finds non-printable non-spaces in Turkish text, but when using the EN locale, doesn't. UTF-8 operates on codepoints, and Unicode categories are assigned to them, AFAIK, in a locale-independent way w.r.t spacing and being printable. E.g., the codepoint U+001F (called `US`, "terminate and justify line", like stretch it from left to right margin) is a generic Control (`C`), and also more specific `Cc`, which is non-printable, but I'm not 100% sure that it's a non-space.  But this is all language-independent! `Ll` and `Lu` may be locale-dependent, but I'm unsure. Locale affects collation (C : codepoint -> integer for sorting) and case equivalence classes, but should not make codepoints invisible.

And no one provided a repro.

jtrmal:
yeah, no idea what that was fixing
y.

kkm000:
On my Debian 11 distro, there is no difference.

From /usr/share/i18n/locales/C:
```
LC_CTYPE
copy "i18n"
translit_start
include "translit_combining";""
translit_end
END LC_CTYPE
```
From /usr/share/i18n/locales/en_US:
```
LC_CTYPE
copy "en_GB"
END LC_CTYPE
```
From /usr/share/i18n/locales/en_GB:
```
LC_CTYPE
copy "i18n"
translit_start
include "translit_combining";""
translit_end
END LC_CTYPE
```
Both end up with `copy "i18n"` and `include "translit_combining"`, the same files. `LANG=C.UTF-8` is correct, its LC_TYPE is very generic. Their distro may be borked, but fixing that is beyond our control.

The `translit_combining` removes all diacritics. All like _really all_, including `U+00011A36: ZANABAZAR SQUARE SIGN CANDRABINDU WITH ORNAMENT` (remind of it to the ppl who think that Czech diacritics are weird). This should not affect Turkish printable/non-printable w.r.t. the LC_TYPE definition, whether the code is normalized or not.

FWIW, I see no problem with UTF-8 text in Turkish, unless I drop down to ASCII:
```
kkm@buba:~/.tmp$ LANG=en_US.UTF-8 grep '[^[:print:][:space:]]' futbol.txt
kkm@buba:~/.tmp$ LANG=C.UTF-8 grep '[^[:print:][:space:]]' futbol.txt
kkm@buba:~/.tmp$ LANG=C grep '[^[:print:][:space:]]' futbol.txt
Futbol, on birer oyuncudan olu▒▒an iki tak▒▒m aras▒▒nda, kendine ▒▒zg▒▒ k▒▒resel bir topla oynanan tak▒▒m
sporudur. 21. y▒▒zy▒▒l itibar▒▒yla 200'▒▒n ▒▒zerinde ▒▒lkede 250 milyonu a▒▒k▒▒n oyuncu taraf▒▒ndan
....
```

kkm000:
@jtrmal, I was thinking about #4193. Sorry, this ticket is unrelated.

jtrmal:
I figured :)

danpovey:
I think the issue is that not all systems have all locales available, and it probably defaults to C if you choose one that is not available.
See this PR:
https://github.com/kaldi-asr/kaldi/pull/4612/files

kkm000:
Every Linux system necessarily has at least one broken locale. I've seen LC_ALL=en_US.UTF-8 not working, with diagnostics like "cannot set locale LC_TYPE", while LANG=en_US.UTF-8 did work and set it just fine. I would rather use `LANG=` or `LC_CTYPE=` in this script than `LC_ALL`. The less you ask of a locale, the safer the code is. The complete list of locale facets include a lot of bizarre stuff, like postal address formatting and prefixes for "Mr.", "Ms." and unknown sex addressee. Later CLDR releases include locally adopted blood glucose measurement units.

`en_US.UTF-8` is ambiguous in many ways, as it can be asked nonsensical questions. Is GREEK QUESTION MARK a printable character? American English does not have it. Up to the locale's designer, I guess. In Debian and therefore Ubuntu, it is: en_US and C both extend the category definitions to the full Unicode codepoint set. I'd say that if a system does not have the "neutral" C.UTF-8 locale _extended with a complete Unicode character definition_ (which is not at all guaranteed), everything else is even a worse guesswork. 

Oh, and amazingly, ISO 8859-7 Latin+Greek 8-bit encoding does define the Greek question mark...

Also, my `locale -a` prints `C.UTF-8` and `en_US.utf8`. All case variants and both `utf8` and `utf-8` are accepted on input, but not canonicalized when printed. So the `locale -a | grep` is also error-prone in #4612.

jtrmal:
cresolved via #4633

PR #3952, opened 2020-02-22
[src] Trim spaces of src/bin/compute-gop.cc

jimbozhang:
Trim redundant spaces.


danpovey:
Thanks, but I don't want to waste a commit on this whitespace change.. prefer to fix these things when there are other changes to be made.

PR #3953, opened 2020-02-22
merging utterances based on the start time and end time to form a paragraph.

aarora8:
It addresses two issues raised in the CHiME group. 
- The first issue was a different result was obtained using the pre-trained models. Around 2% degradation in WER in the dev set and around 0.8% for the eval set, however, in my run the degradation was around 3% for dev set. It was possibly because the previous results were reported without array synchronization.
- The second issue was sorting based on the start time and end time was not performed while merging utterances from each speaker for each session to form a paragraph. 

In the S01.json file of the eval set, array U02  was the ref in the starting and then in the middle but in the code, all utterances of array U02 were merged together in the starting. It caused incorrect scoring for the eval set. With this fix the result changed from 85.42 to 78.08 for the eval set.

Results after running the setup with the pre-trained model
Dev:  %WER 84.33 [ 49653 / 58881, 1529 ins, 35813 del, 12311 sub ]
Eval: %WER 85.42 [ 47093 / 55132, 1583 ins, 32671 del, 12839 sub ]

Results after fixing scoring and running the setup with the pre-trained model
Dev:  %WER 84.33 [ 49653 / 58881, 1529 ins, 35813 del, 12311 sub ]
Eval: %WER 78.08 [ 43046 / 55132, 957 ins, 32045 del, 10044 sub ]
@sw005320

### DIFF
--- a/egs/chime6/s5_track2/RESULTS
+++ b/egs/chime6/s5_track2/RESULTS
@@ -14,5 +14,5 @@ Dev     57.15     83.96
 Eval    54.12     80.33
 
 # ASR nnet3 tdnn+chain
-Dev:  U06 %WER 81.18 [ 58881 / 47798, 1638 ins, 30528 del, 15632 sub ]
-Eval: U06 %WER 85.39 [ 55132 / 47076, 1107 ins, 27768 del, 18201 sub ]
+Dev:  %WER 84.33 [ 49653 / 58881, 1529 ins, 35813 del, 12311 sub ]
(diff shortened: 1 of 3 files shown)


danpovey:
OK, I assume you guys have some kind of internal process to give the OK for these things, please let me know when it's ready to merge.

sw005320:
> OK, I assume you guys have some kind of internal process to give the OK for these things, please let me know when it's ready to merge.

This is OK and ready to merge.
As a record, I also put a [thread](https://groups.google.com/d/msg/chime5/B0w2AN9cDfE/RQxHfhDOAAAJ) in the CHiME challenge google group discussion about it.

PR #3964, opened 2020-02-29
Revert "Better kernels for row sum and tr(AB^T)"

danpovey:
Reverts kaldi-asr/kaldi#3954


danpovey:
See  https://groups.google.com/d/msgid/kaldi-help/97e2c75a-5185-4ef0-a19e-3e763e6a0c93%40googlegroups.com

PR #3966, opened 2020-03-01
training with multiple machines

qindazhu:
- Remove `run_cleanup_segmentation.sh` as it does not make the result better.
- Remove duplicate code and support single-GPU/multiple-GPU/multiple-machine training in one script
- Add online-cmvn as @fanlu showed before

## Result
|          | tdnn_1c_rd_rmc_rng_without_cleanup | tdnn_1c_rd_rmc_rng |
| --------- |--------------------------| --------------------------------|
| dev_cer  |              5.92                               |5.99                      |
| dev_wer |                 13.71                            |13.86                     |
| test_cer  |              7.03                                   |7.08                        |
| test_wer |          15.35                                 |15.72                      |

|          | TDNN-F(Pytorch, Adam, fanlu's previous result )  | TDNN-F(Pytorch, Adam, this pullrequest with 4GPU) | 
| --------- |  ----------------------- | --------------------------| 
| dev_cer  |     6.16                 |              6.29                             |
| dev_wer |     14.01              |                  14.10                           |
| test_cer  |     7.31                 |          7.57                                 |
| test_wer |    15.97                 |          15.80                                 |

PR #3765, opened 2019-12-10
Newnnet3

mrsrikanth:
Hi @danpovey ,

Here is the PR for the chain2 recipes.

Srikanth

PR #3969, opened 2020-03-02
support ivector training in pytorch model

fanlu:
update latest result

||TDNN-F(Pytorch, Adam, fanlu's previous result ) | TDNN-F(Pytorch, Adam, haowen's previous result with 4GPU)| this pull request base on haowen's pr #3966 |
|-- | -- | -- | -- |
|dev_cer | 6.16 | 6.29|6.18|
|dev_wer | 14.01 | 14.10|13.96|
|test_cer | 7.31 | 7.57|7.59|
|test_wer | 15.97 | 15.80|15.86|


egs/aishell/s10/chain/feat_dataset.py
```diff
@@ -55,21 +56,36 @@ def _add_model_left_right_context(x, left_context, right_context):
...
 
-    def __init__(self, feats_scp):
+    def __init__(self, feats_scp, ivector_scp=None):
         assert os.path.isfile(feats_scp)
 
         self.feats_scp = feats_scp
 
-        # items is a list of [key, rxfilename]
-        items = list()
+        # items is a dict of {key: [key, rxfilename, ivec]}
```
csukuangfj: `[key, feat_rxfilename,e ivec_rxfilename]` is better.

egs/aishell/s10/chain/feat_dataset.py
```diff
@@ -55,21 +56,36 @@ def _add_model_left_right_context(x, left_context, right_context):
 
 class FeatDataset(Dataset):
 
-    def __init__(self, feats_scp):
+    def __init__(self, feats_scp, ivector_scp=None):
```
csukuangfj:
```python
if ivector_scp:
  assert os.path.isfile(ivector_scp)
```

egs/aishell/s10/chain/feat_dataset.py
```diff
@@ -81,6 +97,8 @@ def __getitem__(self, i):
 
     def __str__(self):
         s = 'feats scp: {}\n'.format(self.feats_scp)
+        if self.ivector_scp:
```
csukuangfj: `self.ivector_scp` should be initialized to `None` at the begining of the `__init__` function.

egs/aishell/s10/chain/egs_dataset.py
```diff
@@ -151,13 +151,17 @@ def __call__(self, batch):
                     self.egs_left_context + self.egs_right_context
 
             # TODO(fangjun): support ivector
```
csukuangfj: this TODO can be removed.

egs/aishell/s10/local/run_ivector_common.sh
```diff
@@ -0,0 +1,102 @@
...
+  # Also extract iVectors for the test data, but in this case we don't need the speed
+  # perturbation (sp) or small-segment concatenation (comb).
+  for data in dev test; do
+    steps/online/nnet2/extract_ivectors_online.sh --cmd "$train_cmd" --nj 10 \
+      data/${data}_hires exp/nnet3${nnet3_affix}/extractor \
+      exp/nnet3${nnet3_affix}/ivectors_${data}_hires
+  done
+fi
+
+exit 0;
```
csukuangfj: this file should end with a newline.

danpovey:
Thanks a lot for reviewing!

egs/aishell/s10/chain/feat_dataset.py
```diff
@@ -55,21 +56,36 @@ def _add_model_left_right_context(x, left_context, right_context):
...
+        items = dict()
 
         with open(feats_scp, 'r') as f:
             for line in f:
                 split = line.split()
                 assert len(split) == 2
-                items.append(split)
-
-        self.items = items
+                uttid, rxfilename =split
```
csukuangfj: there should be a space `=split` --> `= split`.

You can use some tools to format your files, e.g., `pip install yapf == 0.27.0`

If you are using vim, it can be configured to auto-format you files on saving.

egs/aishell/s10/chain/feat_dataset.py
```diff
@@ -120,10 +142,16 @@ def __call__(self, batch):
             feat = splice_feats(feat)
             feat_list.append(feat)
             # no need to sort the feat by length
+            ivector_list.append(ivector)
```
csukuangfj: `ivector` is undefined if `ivector_rxfilename` is `None`.

egs/aishell/s10/chain/feat_dataset.py
```diff
@@ -120,10 +142,16 @@ def __call__(self, batch):
             feat = splice_feats(feat)
             feat_list.append(feat)
             # no need to sort the feat by length
```
csukuangfj: this comment is for `pad_sequence`.

egs/aishell/s10/chain/feat_dataset.py
fanlu: this data_dict form is from #3947 , self.items will be ites.values(), Should I change the comment ?

egs/aishell/s10/chain/feat_dataset.py
fanlu: OK

egs/aishell/s10/chain/feat_dataset.py
fanlu: OK

egs/aishell/s10/chain/feat_dataset.py
csukuangfj: yes, please.
`rxfilename` was used here because there was only `feats.scp`.
Now ivector is added, so `rxfilename` is a confusing name here.

danpovey:
Guys, I just want to mention something...
I think it would be better if we shifted (not necessarily right now..) to,
instead of exposing the Kaldi egs as a Dataset,
exposing them as a DataLoader.  That way we could use the existing
command-line tools for things like shuffling and
time-shifting, and it will be much more efficient for I/O.
The idea is that the dataloader would, on every epoch, create a suitable
command line and read from it as a pipe.
If it was a distributed data-loader, probably the easiest way to do it
would be to make sure there is an appropriately
split scp file and give it the appropriate one.  We could the scripts here
https://github.com/kaldi-asr/kaldi/pull/3765
to generate the scp files.  I want to merge this soon; one option is to
merge into pybind11 first to test it.

fanlu:
@csukuangfj I have fixed the code with your suggestion. please have a look.
@csukuangfj @qindazhu  please test this pr at your convenience due to Dan's advice. Thanks.

qindazhu:
OK, I'll run it after it's merged.

qindazhu:
I'll take a look at that pr and start to do this.

danpovey:
OK, merging.

danpovey:
Great, thanks!   Firstly, just doing the merge and figuring out how to use
those newer scripts to prepare the egs would be a great start.

PR #3973, opened 2020-03-03
Better kernels for row sum and tr(AB^T): ver. 2

akshaysubr:
Updated PR based on https://github.com/kaldi-asr/kaldi/pull/3954 with the only addition being a fixed LSTM test in `cudamatrix/cu-math-test` that is not sensitive to round-off errors in `TraceMatMat`. The test fix is the 2nd option mentioned in https://github.com/kaldi-asr/kaldi/pull/3954#issuecomment-593668435.


danpovey:
Thanks!!  Merging.

PR #3976, opened 2020-03-04
support load kaldi model in python

fanlu:
@csukuangfj I have test all calculation between kaldi tdnn_1c's model and pytorch kernel(2,2) model, the previous issue is talked in this pr #3940  
After load kaldi model to pytorch, we can get similar result now.
|exp kaldi | test cer | test wer | dev cer | dev wer|
|---|---|---|---|---|
|tdnn_1c| 6.72 | 15.14 | 5.73 | 13.53|
|tdnn_1c load in pytorch BN eps:1e-3(default kaldi config)|6.76|15.20|5.75|13.54|
|tdnn_1c load in pytorch BN eps:1e-5(default pytorch config)|6.86|15.39|5.85|13.68|

First, I use [tool](https://github.com/XiaoMi/kaldi-onnx)  to do this load job, but it's too slow. 
So I want to implement a pybind function to load kaldi's model. And this is more intuitive. What do you think?
But I have problem with `GetComponent` now, sorry about that I am not good at pybind11 and C++.
If you have time, please help me or give me some hint to do this job. Thanks

***

Thanks for @csukuangfj 's help, it is worked.


danpovey:
What is the problem with GetComponent()
You might want to start with existing model-evaluation code, such as in nnet3-compute.

fanlu:
Hi, Dan, I have tested nnet3-compute.cc completely. 
The problem of `GetComponent ` is that this function will return a point to Component.
I don't know how to implement it in pybind. 
after add `.def("GetComponent", &PyClass::GetComponent, py::arg("c"))` in nnet_nnet_pybind.cc
I have got an error msg in make procedure. Maybe I must to read pybind document about this. 
```
template argument deduction/substitution failed:
nnet3/nnet_nnet_pybind.cc:47:63: note:   mismatched types 'pybind11::detail::initimpl::pickle_factory<Args ...>' and 'const char [13]'
      .def("GetComponent", &PyClass::GetComponent, py::arg("c"))
```

csukuangfj:
That is because `GetComponent` is an overloadded method.
You have to specify which one you want to wrap.

fanlu:
should I Wrap all below Component? Is there an easiest way?
```
// static
Component* Component::NewComponentOfType(const std::string &component_type) {
  Component *ans = NULL;
  if (component_type == "SigmoidComponent") {
    ans = new SigmoidComponent();
  } else if (component_type == "TanhComponent") {
    ans = new TanhComponent();
  } else if (component_type == "SoftmaxComponent") {
    ans = new SoftmaxComponent();
  } else if (component_type == "LogSoftmaxComponent") {
    ans = new LogSoftmaxComponent();
  } else if (component_type == "RectifiedLinearComponent") {
    ans = new RectifiedLinearComponent();
  } else if (component_type == "NormalizeComponent") {
    ans = new NormalizeComponent();
  } else if (component_type == "PnormComponent") {
    ans = new PnormComponent();
  } else if (component_type == "AffineComponent") {
    ans = new AffineComponent();
  } else if (component_type == "LinearComponent") {
    ans = new LinearComponent();
  } else if (component_type == "NaturalGradientAffineComponent") {
    ans = new NaturalGradientAffineComponent();
  } else if (component_type == "PerElementScaleComponent") {
    ans = new PerElementScaleComponent();
  } else if (component_type == "NaturalGradientPerElementScaleComponent") {
    ans = new NaturalGradientPerElementScaleComponent();
  } else if (component_type == "PerElementOffsetComponent") {
    ans = new PerElementOffsetComponent();
  } else if (component_type == "SumGroupComponent") {
    ans = new SumGroupComponent();
  } else if (component_type == "FixedAffineComponent") {
    ans = new FixedAffineComponent();
  } else if (component_type == "FixedScaleComponent") {
    ans = new FixedScaleComponent();
  } else if (component_type == "FixedBiasComponent") {
    ans = new FixedBiasComponent();
  } else if (component_type == "NoOpComponent") {
    ans = new NoOpComponent();
  } else if (component_type == "ClipGradientComponent") {
    ans = new ClipGradientComponent();
  } else if (component_type == "ElementwiseProductComponent") {
    ans = new ElementwiseProductComponent();
  } else if (component_type == "ConvolutionComponent") {
    ans = new ConvolutionComponent();
  } else if (component_type == "TdnnComponent") {
    ans = new TdnnComponent();
  } else if (component_type == "MaxpoolingComponent") {
    ans = new MaxpoolingComponent();
  } else if (component_type == "PermuteComponent") {
    ans = new PermuteComponent();
  } else if (component_type == "DistributeComponent") {
    ans = new DistributeComponent();
  } else if (component_type == "CompositeComponent") {
    ans = new CompositeComponent();
  } else if (component_type == "RepeatedAffineComponent") {
    ans = new RepeatedAffineComponent();
  } else if (component_type == "BlockAffineComponent") {
    ans = new BlockAffineComponent();
  } else if (component_type == "NaturalGradientRepeatedAffineComponent") {
    ans = new NaturalGradientRepeatedAffineComponent();
  } else if (component_type == "StatisticsExtractionComponent") {
    ans = new StatisticsExtractionComponent();
  } else if (component_type == "StatisticsPoolingComponent") {
    ans = new StatisticsPoolingComponent();
  } else if (component_type == "ConstantFunctionComponent") {
    ans = new ConstantFunctionComponent();
  } else if (component_type == "ConstantComponent") {
    ans = new ConstantComponent();
  } else if (component_type == "DropoutComponent") {
    ans = new DropoutComponent();
  } else if (component_type == "DropoutMaskComponent") {
    ans = new DropoutMaskComponent();
  } else if (component_type == "GeneralDropoutComponent") {
    ans = new GeneralDropoutComponent();
  } else if (component_type == "SpecAugmentTimeMaskComponent") {
    ans = new SpecAugmentTimeMaskComponent();
  } else if (component_type == "BackpropTruncationComponent") {
    ans = new BackpropTruncationComponent();
  } else if (component_type == "LstmNonlinearityComponent") {
    ans = new LstmNonlinearityComponent();
  } else if (component_type == "BatchNormComponent") {
    ans = new BatchNormComponent();
  } else if (component_type == "TimeHeightConvolutionComponent") {
    ans = new TimeHeightConvolutionComponent();
  } else if (component_type == "RestrictedAttentionComponent") {
    ans = new RestrictedAttentionComponent();
  } else if (component_type == "SumBlockComponent") {
    ans = new SumBlockComponent();
  } else if (component_type == "GruNonlinearityComponent") {
    ans = new GruNonlinearityComponent();
  } else if (component_type == "OutputGruNonlinearityComponent") {
    ans = new OutputGruNonlinearityComponent();
  } else if (component_type == "ScaleAndOffsetComponent") {
    ans = new ScaleAndOffsetComponent();
  }
  if (ans != NULL) {
    KALDI_ASSERT(component_type == ans->Type());
  }
  return ans;
}
```

csukuangfj:
Please refer to
https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/pybind/fstext/lattice_weight_pybind.cc#L59
for  how to wrap overloaded method.

csukuangfj:
It is needed if you want to build a nnet3 model for a config file.

Note that it is a static method.

csukuangfj:
Please refer to
https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/pybind/fstext/lattice_weight_pybind.cc#L80
for how to wrap static methods.

Note that in your case, the function  returns a pointer. You have to use the flag
`take_ownership`.

danpovey:
I would hope it would be possible to wrap only the base-class of Component
(and maybe one or two other classes in that hierarchy, but not all of them).

src/pybind/nnet3/nnet_nnet_pybind.cc
```diff
@@ -0,0 +1,49 @@
...
+          "Also see the function ReadEditConfig() in nnet-utils.h (it's made a"
+          "non-member because it doesn't need special access)."
+     )
+     .def(py::init<>())
+     .def("Read", &PyClass::Read, py::arg("is"), py::arg("binary"))
+     .def("GetComponentNames", &PyClass::GetComponentNames, 
+     "returns vector of component names (needed by some parsing code, for instance).",
+     py::return_value_policy::reference)
+     .def("GetComponentName", &PyClass::GetComponentName, py::arg("component_index"))
+     .def("Info", &PyClass::Info)
```
csukuangfj: please add the comment for `Info`. Later we will
generate documentation automatically from them.

src/pybind/nnet3/nnet_nnet_pybind.cc
```diff
@@ -0,0 +1,49 @@
+// pybind/nnet3/nnet_nnet_pybind.cc
+
+// Copyright 2019   JD AI, Beijing, China (author: Lu Fan)
```
csukuangfj: 2020

src/pybind/kaldi/io_util.py
```diff
@@ -82,3 +82,19 @@ def read_transition_model(rxfilename):
     ki.Close()
 
     return trans_model
+
+
+def read_nnet_model(rxfilename):
```
csukuangfj: `read_nnet3_model` is better since there exist also `nnet1` and `nnet2`.

csukuangfj:
```
.def_static("NewComponentOfType", &PyClass:: NewComponentOfType, py::return_value_policy::take_ownership) 
```
will do the job.

You do **NOT** need to wrap other components to wrap this function.

csukuangfj:
If you have time, please write a `nnet_nnet_pybind_test.py`
to test your code.

fanlu:
Should I wrap init a nnet3 model from a config file? I want a nnet3 model just readed from a binary model file.
And after follow your suggestion, I wraped `GetComponent` like
```
.def("GetComponent", (kaldi::nnet3::Component(*)(int32 c))(&PyClass::GetComponent), py::arg("c"), py::return_value_policy::take_ownership)
```
and this give me an error msg
```
nnet3/nnet_nnet_pybind.cc:48:87: error: address of overloaded function with no contextual type information
      .def("GetComponent", (kaldi::nnet3::Component(*)(int32 c))(&PyClass::GetComponent), py::arg("c"), py::return_value_policy::take_ownership)
```

csukuangfj:
```
.def("GetComponent", (kaldi::nnet3::Component(*)(int32 c))(&PyClass::GetComponent), py::arg("c"), py::return_value_policy::take_ownership)
```

what is the function type that you're trying to convert?

csukuangfj:
```
(ret_type (class_name::*)(arg_list) [optional_const_specifier])
```

The example I gave is for a global function. In your case, it is a class member function.

csukuangfj:
`&PyClass::GetComponent` returns a pointer.
`(type_converted_to) (&PyClass::GetComponent)`, you have to figure out the appropriate `type_converted_to`

fanlu:
Note that what kaldi-onnx dose. And I want to get a dict from kaldi's model file like the code below.
```
with open(filename, 'r') as f:
        p = Nnet3Parser(f)
        p.run()
        for component in p._components:
            kaldi_dict[component["name"]] = component
```
So I suspect that  I can use `GetComponent` to get Component and get children Component's params after range(`NumComponents`)

fanlu:
Can `type_converted_to` to be `Component`?
> `&PyClass::GetComponent` returns a pointer.
> `(type_converted_to) (&PyClass::GetComponent)`, you have to figure out the appropriate `type_converted_to`

csukuangfj:
As you can see from the following

https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/nnet3/nnet-nnet.h#L128-L133

`GetComponent` is an overloaded method. You have to specify explicitly which method
is going to be wrapped; otherwise, there is no way for pybind11 to infer the prototype of
`GetComponent`.

You can use either

```cpp
.def("GetComponent", (Component* (PyClass::*)(int32) )&PyClass::GetComponent, py::arg("c"),py::return_value_policy::reference)
```
or
```cpp
.def("GetComponent", (const Component* (PyClass::*)(int32) const )&PyClass::GetComponent, py::arg("c"), py::arg("c"),py::return_value_policy::reference)
```

Or you can use both.

Since the return type is `Component*`, you have to wrap it first.
https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/nnet3/nnet-component-itf.h#L114

fanlu:
I have got the Component Object. As Dan's advice, Is there an convenient way to cast Component to Child class If I do not wrap all Child Class?
image

```
FixedAffineComponent

  const CuMatrix<BaseFloat> &LinearParams() const { return linear_params_; }
  const CuVector<BaseFloat> &BiasParams() const { return bias_params_; }
```

csukuangfj:
@fanlu 
You can wrap any subclasses of Component that you are using.

francisr:
> First, I use [tool](https://github.com/XiaoMi/kaldi-onnx) to do this load job, but it's too slow.
What's slow about the tool? The conversion or the runtime?

fanlu:
Nnet3Parser(f).run()
> > First, I use [tool](https://github.com/XiaoMi/kaldi-onnx) to do this load job, but it's too slow.
> > What's slow about the tool? The conversion to ONNX or the runtime?

csukuangfj:
Is the conversion process slow?
Or after the conversion, is the runtime slow?

csukuangfj:
You can split `Nnet3Parser(f).run()` into

```
start = start_time
model = Nnet3Parser(f)
stop = end_time
--> the time for the conversion process

start =  start_time
model.run()
stop = stop_time
---> time for the runtime.
```

fanlu:
the time of conversion and runtime are 0.0004012584686279297s and 97.50831198692322s, respectively

src/pybind/nnet3/nnet_nnet_pybind.cc
```diff
@@ -19,31 +19,90 @@
...
 #include "nnet3/nnet-component-itf.h"
+#include "nnet3/nnet-convolutional-component.h"
+#include "nnet3/nnet-nnet.h"
+#include "nnet3/nnet-normalize-component.h"
+#include "nnet3/nnet-simple-component.h"
 
 using namespace kaldi;
 using namespace kaldi::nnet3;
 
+template <typename... Args>
```
csukuangfj: You do not need to redefine it here.

It is already defined in
https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/pybind/kaldi_pybind.h#L30-L31

src/pybind/nnet3/nnet_nnet_pybind.cc
```diff
@@ -19,31 +19,90 @@
...
+      .def("Info", &Comp::Info,
+           "Returns some text-form information about this component, for "
+           "diagnostics. Starts with the type of the component.  E.g. "
+           "\"SigmoidComponent dim=900\", although most components will have "
+           "much more info.")
+      .def_static("NewComponentOfType", &Comp::NewComponentOfType,
+                  py::return_value_policy::take_ownership);
+
+  using BNC = kaldi::nnet3::BatchNormComponent;
+  py::class_<BNC>(m, "BatchNormComponent")
```
csukuangfj: BatchNormComponent is a subclass of Component.

Please refer to kaldi-matrix for how to specify a base class
in Pybind11.

src/pybind/nnet3/nnet_nnet_pybind.cc
```diff
@@ -19,31 +19,90 @@
...
-     "returns vector of component names (needed by some parsing code, for instance).",
-     py::return_value_policy::reference)
-     .def("GetComponentName", &PyClass::GetComponentName, py::arg("component_index"))
-     .def("Info", &PyClass::Info)
-     .def("NumComponents", &PyClass::NumComponents)
-     .def("NumNodes", &PyClass::NumNodes)
-     // .def("GetComponent", &PyClass::GetComponent, py::arg("c"))
-     ;
+  using Comp = kaldi::nnet3::Component;
+  py::class_<Comp>(m, "Component",
```
csukuangfj: `Component` is from the file `nnet-component-itf`. Please create a new file
to wrap it.

When a class in Kaldi is in `foo.h`,  please wrap it in `foo_pybind.h`
and `foo_pybind.cc`

csukuangfj:
Please add unittest files.

fanlu:
Sure
> Please add unittest files.

I have problem about how to extract a numpy from kaldi_pybind.FloatCuMatrix or kaldi_pybind.FloatCuVector?

csukuangfj:
@fanlu 
Please refer to

https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/pybind/dlpack/dlpack_pybind_test_gpu.py#L94-L120

Since numpy does not support DLPack.
You have to first convert a FloatCuVector to dlpack, then to a PyTorch tensor, 
and finally to a numpy array object by using `.numpy()` method of a Pytorch tensor.

fanlu:
I have some problem with `from_dlpack()` now.
image

I have changed to_dlpack() from `CuMatrix` to `CuMatrixBase` also, I am not sure it is a right way.
@csukuangfj Maybe you can give me more implement details.

src/nnet3/nnet-normalize-component.h
```diff
@@ -224,6 +224,10 @@ class BatchNormComponent: public Component {
   const CuVector<BaseFloat> &Offset() const { return offset_; }
   const CuVector<BaseFloat> &Scale() const { return scale_; }
 
+  virtual const CuVector<double> &StatsSum() const { return stats_sum_; }
```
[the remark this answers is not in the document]
csukuangfj: Why do you make it a `virtual` method?

src/pybind/nnet3/nnet_normalize_component_pybind.cc
```diff
@@ -0,0 +1,35 @@
...
+
+void pybind_nnet_normalize_component(py::module& m) {
+  using PyClass = kaldi::nnet3::BatchNormComponent;
+  py::class_<PyClass>(m, "BatchNormComponent")
+      .def("Type", &PyClass::Type)
+      .def("StatsSum", &PyClass::StatsSum)
+      .def("StatsSumsq", &PyClass::StatsSumsq)
+      .def("Count", &PyClass::Count)
+      .def("Offset", &PyClass::Offset)
+      .def("Scale", overload_cast_<>()(&PyClass::Scale, py::const_));
```
csukuangfj: `py::return_value_policy:reference` and probably it will solve your `cuda runtime error`.

src/pybind/nnet3/nnet_simple_component_pybind.cc
```diff
@@ -0,0 +1,46 @@
...
+
+#include "nnet3/nnet-simple-component.h"
+
+using namespace kaldi::nnet3;
+
+void pybind_nnet_simple_component(py::module& m) {
+  using FAC = FixedAffineComponent;
+  py::class_<FAC>(m, "FixedAffineComponent")
+      .def("Type", &FAC::Type)
+      .def("LinearParams", &FAC::LinearParams)
```
csukuangfj: return value policy :: reference.

Please check your code to add the above flag.

src/nnet3/nnet-normalize-component.h
```diff
@@ -224,6 +224,10 @@ class BatchNormComponent: public Component {
...
+  virtual const CuVector<double> &StatsSumsq() const { return stats_sumsq_; }
+  virtual const double &Count() const { return count_; }
```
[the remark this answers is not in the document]
csukuangfj: For primitive types, there is no need to return a `const &`.

src/nnet3/nnet-normalize-component.h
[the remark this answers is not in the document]
fanlu: there is no method of name `StatsSum ` in base calss, Should I use keyword override?

fanlu:
because in class `TdnnComponent` the return type of get linear_params_ method is `CuMatrixBase`
```
CuMatrixBase<BaseFloat> &LinearParams() { return linear_params_; }
```
should I implement a new method like
```
CuMatrix<BaseFloat> &LinearParams() { return linear_params_; }
```

> @fanlu
> 
> > I have changed to_dlpack() from CuMatrix to CuMatrixBase also, I am not sure it is a right way.
> 
> Please refer to Dan's comment ([#3788 (comment)](https://github.com/kaldi-asr/kaldi/pull/3788#commitcomment-36561683)):
> 
> > I think it might be worth having a to_dlpack specifically in class Matrix, rather than in MatrixBase. Class Matrix owns its memory, so if the deleter were to do a Py_DECREF on the Python object where that Matrix lives, the memory would be managed appropriately. I'm concerned that with the current approach, it's just a segmentation fault waiting to happen because it doesn't automatically ensure that the pointer is still valid/allocated.

csukuangfj:
@fanlu 
The underlying type is
https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/nnet3/nnet-convolutional-component.h#L603

Please show me the return type in **Python** using `print()` and `type()`.
I guess pybind11 is smart enough to infer the return type and will **NOT** give you an object of type `CuMatrixBase`.

fanlu:
```
print(component.LinearParams())
<kaldi_pybind.FloatCuMatrixBase object at 0x7f2890160ef0>

print(type(component.LinearParams()))
<class 'kaldi_pybind.FloatCuMatrixBase'>
```

src/nnet3/nnet-normalize-component.h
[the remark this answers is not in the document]
csukuangfj: No.

Then please remove `virtual`.

csukuangfj:
I think it is a mistake in the c++ code to use `CuMatrixBase`:
https://github.com/kaldi-asr/kaldi/blob/cc63cca97ecd4b54d1b3f1dd9994a3117c999404/src/nnet3/nnet-convolutional-component.h#L554-L558

@danpovey 
Can we replace it with `CuMatrix`?

fanlu:
the same style in class LinearComponent 
```
CuMatrixBase<BaseFloat> &Params() { return params_; }
const CuMatrixBase<BaseFloat> &Params() const { return params_; }
CuMatrix<BaseFloat> params_;
```

csukuangfj:
> print(component.LinearParams())
<kaldi_pybind.FloatCuMatrixBase object at 0x7f2890160ef0>

> print(type(component.LinearParams()))
<class 'kaldi_pybind.FloatCuMatrixBase'>

What is the type of `component` ??? @fanlu

fanlu:
```
print(component)
<kaldi_pybind.nnet3.TdnnComponent object at 0x7f1b17fff9f0>

print(type(component))
<class 'kaldi_pybind.nnet3.TdnnComponent'>
```

csukuangfj:
For the following program

```cpp
class Base {
 public:
  virtual std::string Hello() const = 0;

  virtual ~Base() = default;

 protected:
  Base() = default;
};

class Child : public Base {
 public:
  Child() {}
  std::string Hello() const override { return "hello"; }
};

class Test {
 public:
  Test() = default;
  const Base& GetChild() const { return c_; }

 private:
  Child c_;
};
```

The return type of `GetChild()` in Python  is `Child`.

----

Please check your code again.

It should return a type of `FloatCuMatrix`

src/nnet3/nnet-normalize-component.h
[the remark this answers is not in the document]
fanlu: @csukuangfj 
```
  CuVector<double> &StatsSum() { return stats_sum_; }
  CuVector<double> &StatsSumsq() { return stats_sumsq_; }
  double Count() { return count_; }
```
I have a problem here
```
TypeError: Unable to convert function return value to a Python type! The signature was
        (self: kaldi_pybind.nnet3.BatchNormComponent) -> kaldi::CuVector<double>
```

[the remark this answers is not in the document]
csukuangfj: Please paste the link to the code.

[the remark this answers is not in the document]
csukuangfj: Do not use double for any matrix types in Kaldi for pybind.

src/pybind/nnet3/nnet_nnet_pybind_test.py
```diff
@@ -3,38 +3,155 @@
 # Copyright 2020 JD AI, Beijing, China (author: Lu Fan)
 # Apache 2.0
 
+import kaldi
```
csukuangfj: please rearrange the import order:

```python
import <standard modules>

** blank line **

import <third party modules>

** blank line **

import <your own modules>
```

src/pybind/nnet3/nnet_nnet_pybind_test.py
```diff
@@ -3,38 +3,155 @@
...
+import unittest
+from torch.utils.dlpack import to_dlpack
+from torch.utils.dlpack import from_dlpack
 import os
 import sys
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), os.pardir))
 
-import unittest
+try:
+    import torch
```
csukuangfj: please place `from torch ....` after this line.

src/pybind/nnet3/nnet_nnet_pybind_test.py
```diff
@@ -3,38 +3,155 @@
...
+            print('Kaldi is not compiled with CUDA! Skip it')
+            return
+
+        device_id = 0
+
+        # Kaldi and PyTorch will use the same GPU
+        kaldi.SelectGpuDevice(device_id=device_id)
+        kaldi.CuDeviceAllowMultithreading()
+
+        final_mdl = 'src/pybind/nnet3/final.mdl'
```
csukuangfj: please change it to `final.mdl`

This test is going to be run inside `pybind/nnet3` directory.

Append a line to `pybind/nnet3/Makefile`.

src/pybind/nnet3/nnet_nnet_pybind_test.py
```diff
@@ -3,38 +3,155 @@
...
-                # stats_sum = from_dlpack(component.StatsSum().to_dlpack())
-                # stats_sumsq = from_dlpack(component.StatsSumsq().to_dlpack())
-                # print(stats_sum.shape)
-                pass
-            elif "LinearComponent" == comp_type:
-                linear_params = from_dlpack(component.LinearParams().to_dlpack())
-                print(linear_params.shape)
+                self.assertEqual(linear_params.shape, (3448, 4))
+                self.assertEqual(bias_params.shape, (3448,))
+
```
csukuangfj: there should be an `else` statement at the end.

You can place an `assert` statement or raise an exception there.

src/nnet3/nnet-normalize-component.h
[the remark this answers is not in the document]
fanlu: but stats_sum_'s and stats_sumsq_ 's type are CuVector<double>
and offset_ scale_ also give me this bug
```
print(component.Offset().to_dlpack())
<capsule object "dltensor" at 0x7f28404c1780>

print(from_dlpack(component.Offset().to_dlpack()))
RuntimeError: CUDA error: invalid argument
```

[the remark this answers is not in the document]
csukuangfj: You do not need `stats_sum_` and `stats_sumsq_` in inference mode.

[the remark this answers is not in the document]
csukuangfj: Have you invoked `SetTestMode(true)` before 
`print(from_dlpack(component.Offset().to_dlpack()))` ?

Please re-check your code.

src/pybind/nnet3/nnet_normalize_component_pybind.cc
```diff
@@ -0,0 +1,36 @@
...
+
+void pybind_nnet_normalize_component(py::module& m) {
+  using PyClass = kaldi::nnet3::BatchNormComponent;
+  py::class_<PyClass, Component>(m, "BatchNormComponent")
+      .def("StatsSum", &PyClass::StatsSum, py::return_value_policy::reference)
+      .def("StatsSumsq", &PyClass::StatsSumsq, py::return_value_policy::reference)
+      .def("Count", &PyClass::Count)
+      .def("Eps", &PyClass::Eps)
+      .def("Offset", &PyClass::Offset, py::return_value_policy::reference)
+      .def("Scale", overload_cast_<>()(&PyClass::Scale, py::const_),
```
csukuangfj: Apparently, `SetTestMode` is missing.
`Offset` and `Scale` is empty by default and you cannot
assign an empty DLPack to `from_dlpack`.

src/nnet3/nnet-normalize-component.h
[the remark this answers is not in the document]
fanlu: the StatsMean and StatsVar in kaldi's model are equivalent to running_mean, running_var in pytorch, respectively.
but in kaldi's Read and Write function, there are some other operator, like 
```
Read
stats_sumsq_.AddVecVec(1.0, stats_sum_, stats_sum_, 1.0);
stats_sum_.Scale(count_);
stats_sumsq_.Scale(count_);
```
```
Write
CuVector<BaseFloat> mean(stats_sum_), var(stats_sumsq_);
  if (count_ != 0) {
    mean.Scale(1.0 / count_);
    var.Scale(1.0 / count_);
    var.AddVecVec(-1.0, mean, mean, 1.0);
  }
```
these operator are not necessary for pytorch's model, just use mean and var rather than offset and scale.

src/nnet3/nnet-normalize-component.h
```diff
@@ -224,8 +224,8 @@ class BatchNormComponent: public Component {
   const CuVector<BaseFloat> &Offset() const { return offset_; }
   const CuVector<BaseFloat> &Scale() const { return scale_; }
 
-  CuVector<double> &StatsSum() { return stats_sum_; }
-  CuVector<double> &StatsSumsq() { return stats_sumsq_; }
+  CuVector<BaseFloat> Mean();
```
csukuangfj: CuVector<BaseFloat> Mean()  *******const********;

csukuangfj: Batchnorm is acutally a affine transform in the inference mode.

Why don't you use `offset` and `scale`?

src/nnet3/nnet-normalize-component.h
```diff
@@ -224,8 +224,8 @@ class BatchNormComponent: public Component {
...
+  CuVector<BaseFloat> Var();
   double Count() { return count_; }
```
csukuangfj:
```cpp
double Count() const
BaseFloat Eps() const
```

src/nnet3/nnet-normalize-component.h
fanlu: In pytorch's forward calculation, Only we need is put the mean and var to a dict, and use `model.load_state_dict(dst_state_dict)` to load this two variable to pytorch's model, then pytorch can do offset and scale by itself.

csukuangfj:
+1

PR #3979, opened 2020-03-05
Create origin.yml

sabazmn:


jtrmal:
What is it for? Thanks.
Y.

kkm000:
This is just a copy of the Actions example. We are not likely to switch to it, because building with CUDA and MKL on 2vCPU allocated for open-source projects would be too slow. I'm thinking about using a cloud platform to speed up testing and builds, but certainly copying and unrelated example is not the way to start the whole process. Sorry, but it probably does not make sense yet.

PR #3983, opened 2020-03-06
sync master to pybind11

qindazhu:


danpovey:
This has no changes but I'm merging to avoid future merge problems.

PR #3984, opened 2020-03-06
MKL on Archlinux

o-alexandre-felipe:
Changed tools/install_mkl.sh and src/configure in order to be able to compile using MKL in systems using pacman from archlinux.


danpovey:
Thanks!  merging.

PR #3985, opened 2020-03-07
fix errors/types in doc of chain denominator.

csukuangfj:


src/chain/chain-den-graph.cc
```diff
@@ -120,7 +120,6 @@ void DenominatorGraph::SetInitialProbs(const fst::StdVectorFst &fst) {
       avg_prob(num_states);
   cur_prob(fst.Start()) = 1.0;
   for (int32 iter = 0; iter < num_iters; iter++) {
-    avg_prob.AddVec(1.0 / num_iters, cur_prob);
```
danpovey: Sorry, I can't accept this change, because then the initial state of the FST might get zero weight.
The first iter is more critical than the last iter.  The fixes to comments are OK though.

csukuangfj: OK, I'll revert it.

Is the result of the last iteration ignored intentionally?

csukuangfj: Can I copy it to line 122, i.e., outside the for loop?
Then we can use the result from the last iteration.

danpovey: No, don't change it please.  The extra time taken is tiny, and the current code is simple.

csukuangfj: Reverted.

PR #3989, opened 2020-03-09
Allow install_mkl.sh to take 'arch' as a type of package manager

BurnhamG:
This follow-up to #3984 allows the use of 'arch' in the -p flag to the install_mkl.sh script, and also checks if /etc/os-release contains 'arch.'


kkm000:
LGTM, thanks!

Issue #3992, opened 2020-03-11
wav-reverberate WaveData  output issue


CAIQT:
In [wav-reverberate.cc](https://github.com/kaldi-asr/kaldi/blob/344cdcb060b7d563d86ba22ffa0a5b8a80659482/src/featbin/wav-reverberate.cc#L373), when preparing Output ko for out_wave,
`Output ko(output_wave_file, false);`
may be changed into:
`Output ko(output_wave_file, true, false);`
to write without header.

Without this,  suppose we read a wav file without doing any changes and save to another position, then the contents of the saved file and the original file will not be the same because of the extra header.

danpovey:
I think you might be right... can you try changing it and verify that it doesn't break a recipe that uses it?  I am looking at the code and I am having a hard time seeing how it worked in the first place.

danpovey:
... and also could you make a PR?  Let me know if you won't be able to do this and I can ask someone else.

PR #3993, opened 2020-03-11
Fix wav-reverberate.cc WaveData output. Issue #3992

CAIQT:
Fix issue #3992


danpovey:
Is this tested at all?  you compiled it?  Ran it?

CAIQT:
Hi dan, I have compiled and run.  No unittest required. 
I use the following command line:
`wav-reverberated 1.wav 2.wav`
and then check whether the two wav files' md5sums are the same.

PR #3999, opened 2020-03-17
expose egs as Dataloader

qindazhu:
Expose egs as a Dataloader in PyTorch, training time now decreased from 150mins to 90mins for 6 epochs with 4 workers. 

### **RESULT**
 ||TDNN-F(Pytorch, Adam, delta dropout without ivector ) from @fanlu | TDNN-F(Pytorch, Adam, delta dropout without ivector ) this PR 2nd run | TDNN-F(Pytorch, Adam, delta dropout without ivector ) this PR 1st run | this Pr with commit 0d8aada to make dropout go to zero at the end |
|--|--|--|--|--|
|dev_cer|6.10|6.13|6.18|6.12
|dev_wer|13.86|13.89|13.96|13.92
|test_cer|7.14|7.19|7.20|7.26
|test_wer|15.49|15.54|15.66|15.63
|training_time|151mins|88mins|84mins|

WER/CER increase may come from:
-  **Shuffle**, we do not shuffle egs-minibatch during each epoch.
-  **Dropout**, we use `pseudo_epoch` (one scp file is one pseudo-epoch) to compute `data_fraction` in dropout, that is relatively too much coarse-grained than using `batch_idx`

Note that I have tried  `copy|shuffle|merge` in dataloader (see code below"), but seems that it will take as much time as (or even a little more time than) the original approach (egs as a Dataset), I may do further experiment to look into this:
```
 scp_rspecifier = scp_file_to_process
 egs_rspecifier = 'ark,bg:nnet-chain-copy-egs --frame-shift .. scp:scp_rspecifier ark:- | \
                        nnet3-chain-shuffle-egs --buffer-size .. --srand .. ark:- ark:- | \
                        nnet3-chain-merge-egs --minibatch-size .. ark:- ark:- |'
 with SequentialNnetChainExampleReader(egs_rspecifier) as example_reader:
       for key, eg in example_reader:
             batch = self.collate_fn(eg)
             yield pseudo_epoch, batch
```

### **TODO**
- [ ] Split egs to more scp files (currently 56 files) to see whether it will get fine-grained in `data_fraction` of dropout or not.
- [ ] Do further experiment and trace for approach `copy|shuffle|merge` in dataloader to confirm the bottleneck of this approach.
- [ ] Profile first epoch of training to see why it takes so much time, as we can see now that first epoch training would take most part of time of the whole training time, no matter what approach (egs as dataset or dataloader) we use.

### DIFF
--- a/egs/aishell/s10/chain/egs_dataloader.py
+++ b/egs/aishell/s10/chain/egs_dataloader.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python3
+
+# Copyright 2020 Xiaomi Corporation, Beijing, China (author: Haowen Qiu)
+# Apache 2.0
+
+from multiprocessing import Process
+import datetime
+import glob
+import os
+
+import numpy as np
+import torch
+import torch.distributed as dist
+
+from torch.utils.data import DataLoader
+from torch.utils.data import Dataset
+
+from kaldi import SequentialNnetChainExampleReader
+import kaldi
+import kaldi_pybind.nnet3 as nnet3
+
+from common import splice_feats
+
+def get_egs_dataloader(egs_dir_or_scp,
+                       egs_left_context,
+                       egs_right_context,
+                       frame_subsampling_factor=3,
+                       world_size=None,
+                       local_rank=None):
+    '''
+    world_size and local_rank is for DistributedDataParallel training.
+    '''
+    dataset = NnetChainExampleScpDataset(egs_dir_or_scp)
+
+    collate_fn = NnetChainExampleCollateFunc(
+        egs_left_context=egs_left_context,
+        egs_right_context=egs_right_context,
+        frame_subsampling_factor=frame_subsampling_factor)
+
+    if local_rank is not None:
+        sampler = torch.utils.data.distributed.DistributedSampler(
+            dataset, num_replicas=world_size, rank=local_rank, shuffle=False)
+    else:
+        sampler = torch.utils.data.SequentialSampler(dataset)
+    
+    dataloader = NnetChainExampleDataLoader(dataset,
+                            sampler=sampler,
+                            collate_fn=collate_fn)
+    return dataloader
+
+
+class NnetChainExampleScpDataset(Dataset):
+
+    def __init__(self, egs_dir_or_scp):
+        '''
+        If egs_dir_or_scp is a directory, we assume that there exist many cegs.*.scp files
+        inside it.
+        '''
+        if os.path.isdir(egs_dir_or_scp):
+            self.scps = glob.glob('{}/cegs.*.scp'.format(egs_dir_or_scp))
+        else:
+            self.scps = [egs_dir_or_scp]
+
+        assert len(self.scps) > 0
+
+    def __len__(self):
+        return len(self.scps)
+
+    def __getitem__(self, i):
+        return self.scps[i]
+
+    def __str__(self):
+        s = 'num egs scp files: {}\n'.format(len(self.scps))
+        return s
+
+
+class NnetChainExampleDataLoader(object):
+    '''
+    Nnet chain example data loader, provides an iterable over the given scp files.
+
+    Arguments:
+        dataset (Dataset): scp files from which to load the egs.
+        sampler (Sampler): defines the strategy to draw samples
+            from the dataset.
+        collate_fn (callable): creates a batch from mergerd eg.
+
+    '''
+
+    def __init__(self, dataset, sampler, collate_fn):
+
+        self.dataset = dataset
+        self.sampler = sampler
+        self.collate_fn = collate_fn
+
+    def __len__(self):
+        return len(self.sampler)
+
+    def __iter__(self):
+        # iterates over one scp file in a `pseudo_epoch`
+        for pseudo_epoch, sample_idx in enumerate(self.sampler):
+            # one sample is one scp file
+            egs_rspecifier = 'scp:' + self.dataset[sample_idx]
+            with SequentialNnetChainExampleReader(egs_rspecifier) as example_reader:
+                for key, eg in example_reader:
+                    batch = self.collate_fn(eg)
+                    yield pseudo_epoch, batch
+
+
+class NnetChainExampleCollateFunc:
+
+    def __init__(self, egs_left_context, egs_right_context,
+                 frame_subsampling_factor=3):
+
+        '''
+        egs_left_context is from egs/info/left_context
+        egs_right_context is from egs/info/right_context
+        '''
+        assert egs_left_context >= 0
+        assert egs_left_context >= 0
+
+        # currently support either no subsampling or
+        # subsampling factor to be 3
+        assert frame_subsampling_factor in [1, 3]
+
+        self.egs_left_context = egs_left_context
+        self.egs_right_context = egs_right_context
+        self.frame_subsampling_factor = frame_subsampling_factor
+        
+    def __call__(self, eg):
+        '''
+        eg is a batch as it has been merged
+        '''
+        assert eg.inputs[0].name == 'input'
+        assert len(eg.outputs) == 1
+        assert eg.outputs[0].name == 'output'
+
+        # the length of below two return values will be 1,
+        # as we have merged egs into one batch
+
+        # TODO(haowen): remove the list wrapping
+        feature_list = []
+        supervision_list = []
+
+        supervision = eg.outputs[0].supervision
+        supervision_list.append(supervision)
+
+        batch_size = supervision.num_sequences
+        frames_per_sequence = (supervision.frames_per_sequence *
+                               self.frame_subsampling_factor) + \
+            self.egs_left_context + self.egs_right_context
+
+
+        _feats = kaldi.FloatMatrix()
+        eg.inputs[0].features.GetMatrix(_feats)
+        feats = _feats.numpy()
+
+        if len(eg.inputs) > 1:
+            _ivectors = kaldi.FloatMatrix()
+            eg.inputs[1].features.GetMatrix(_ivectors)
+            ivectors = _ivectors.numpy()
+
+        assert feats.shape[0] == batch_size * frames_per_sequence
+
+        feat_list = []
+        for i in range(batch_size):
+            start_index = i * frames_per_sequence
+            if self.frame_subsampling_factor == 3:
+                shift = np.random.choice([-1, 0, 1], 1)[0]
+                start_index += shift
+
+            end_index = start_index + frames_per_sequence
+            start_index += 2  # remove the leftmost frame added for frame shift
+            end_index -= 2  # remove the rightmost frame added for frame shift
+            feat = feats[start_index:end_index:, :]
+            if len(eg.inputs) > 1:
+                repeat_ivector = torch.from_numpy(
+                    ivectors[i]).repeat(feat.shape[0], 1)
+                feat = torch.cat(
+                    (torch.from_numpy(feat), repeat_ivector), dim=1).numpy()
+            feat_list.append(feat)
+
+        batched_feat = np.stack(feat_list, axis=0)
+        assert batched_feat.shape[0] == batch_size
+
+        # -4 = -2 -2
+        # the first -2 is from extra left/right context
+        # the second -2 is from lda feats splicing
+        assert batched_feat.shape[1] == frames_per_sequence - 4
+        if len(eg.inputs) > 1:
+            assert batched_feat.shape[2] == feats.shape[-1] + ivectors.shape[-1]
+        else:
+            assert batched_feat.shape[2] == feats.shape[-1]
+
+        torch_feat = torch.from_numpy(batched_feat).float()
+        feature_list.append(torch_feat)
+
+        return feature_list, supervision_list
+
+
+def _test_nnet_chain_example_dataloader():
+    scp_dir = 'exp/chain_pybind/tdnn_sp/egs_chain2_for_training'
+    _test_dataloader_iter(scp_dir)
+
+def _test_dataloader_iter(scp_dir_or_file):
+    egs_left_context = 29
+    egs_right_context = 29
+    frame_subsampling_factor = 3
+
+    dataloader = get_egs_dataloader(
+        scp_dir_or_file,
+        egs_left_context,
+        egs_right_context,
+        frame_subsampling_factor)
+    
+    for i in range(2):
+        batch_idx = 0
+        for pseudo_epoch, batch in dataloader:
+            print('{}: epoch {}, pseudo_epoch {}, batch_idx {}'.format(
+                datetime.datetime.now(), i, pseudo_epoch, batch_idx))
+            batch_idx = batch_idx + 1
+            feature_list, supervision_list = batch
+            assert feature_list[0].shape == (128, 204, 120) \
+                or feature_list[0].shape == (128, 144, 120) \
+                or feature_list[0].shape == (128, 165, 120)
+            assert supervision_list[0].weight == 1
+            assert supervision_list[0].num_sequences == 128  # minibach size is 128
+
+
+if __name__ == '__main__':
+    _test_nnet_chain_example_dataloader()
--- a/egs/aishell/s10/chain/egs_dataloader.py
+++ b/egs/aishell/s10/chain/egs_dataloader.py
@@ -12,7 +12,6 @@ import numpy as np
 import torch
 import torch.distributed as dist
 
-from torch.utils.data import DataLoader
 from torch.utils.data import Dataset
 
 from kaldi import SequentialNnetChainExampleReader
@@ -182,9 +181,6 @@ class NnetChainExampleCollateFunc:
         batched_feat = np.stack(feat_list, axis=0)
         assert batched_feat.shape[0] == batch_size
 
-        # -4 = -2 -2
-        # the first -2 is from extra left/right context
-        # the second -2 is from lda feats splicing
         assert batched_feat.shape[1] == frames_per_sequence - 4
         if len(eg.inputs) > 1:
             assert batched_feat.shape[2] == feats.shape[-1] + ivectors.shape[-1]
--- a/egs/aishell/s10/chain/egs_dataloader.py
+++ b/egs/aishell/s10/chain/egs_dataloader.py
@@ -38,9 +38,10 @@ def get_egs_dataloader(egs_dir_or_scp,
 
     if local_rank is not None:
         sampler = torch.utils.data.distributed.DistributedSampler(
-            dataset, num_replicas=world_size, rank=local_rank, shuffle=False)
+            dataset, num_replicas=world_size, rank=local_rank, shuffle=True)
     else:
-        sampler = torch.utils.data.SequentialSampler(dataset)
+        #sampler = torch.utils.data.SequentialSampler(dataset)
+        sampler = torch.utils.data.RandomSampler(dataset)
     
     dataloader = NnetChainExampleDataLoader(dataset,
                             sampler=sampler,
@@ -133,15 +134,8 @@ class NnetChainExampleCollateFunc:
         assert len(eg.outputs) == 1
         assert eg.outputs[0].name == 'output'
 
-        # the length of below two return values will be 1,
-        # as we have merged egs into one batch
-
-        # TODO(haowen): remove the list wrapping
-        feature_list = []
-        supervision_list = []
 
         supervision = eg.outputs[0].supervision
-        supervision_list.append(supervision)
 
         batch_size = supervision.num_sequences
         frames_per_sequence = (supervision.frames_per_sequence *
@@ -188,13 +182,12 @@ class NnetChainExampleCollateFunc:
             assert batched_feat.shape[2] == feats.shape[-1]
 
         torch_feat = torch.from_numpy(batched_feat).float()
-        feature_list.append(torch_feat)
 
-        return feature_list, supervision_list
+        return torch_feat, supervision
 
 
 def _test_nnet_chain_example_dataloader():
-    scp_dir = 'exp/chain_pybind/tdnn_sp/egs_chain2_for_training'
+    scp_dir = 'exp/chain_pybind/tdnn_sp/egs_chain2'
     _test_dataloader_iter(scp_dir)
 
 def _test_dataloader_iter(scp_dir_or_file):
@@ -214,12 +207,12 @@ def _test_dataloader_iter(scp_dir_or_file):
             print('{}: epoch {}, pseudo_epoch {}, batch_idx {}'.format(
                 datetime.datetime.now(), i, pseudo_epoch, batch_idx))
             batch_idx = batch_idx + 1
-            feature_list, supervision_list = batch
-            assert feature_list[0].shape == (128, 204, 120) \
-                or feature_list[0].shape == (128, 144, 120) \
-                or feature_list[0].shape == (128, 165, 120)
-            assert supervision_list[0].weight == 1
-            assert supervision_list[0].num_sequences == 128  # minibach size is 128
+            feature, supervision = batch
+            assert feature.shape == (128, 204, 120) \
+                or feature.shape == (128, 144, 120) \
+                or feature.shape == (128, 165, 120)
+            assert supervision.weight == 1
+            assert supervision.num_sequences == 128  # minibach size is 128
 
 
 if __name__ == '__main__':
--- a/egs/aishell/s10/chain/model.py
+++ b/egs/aishell/s10/chain/model.py
@@ -70,7 +70,11 @@ output-layer name=output-xent dim=3456 learning-rate-factor=5.0 l2-regularize=0.
 
 
 def constrain_orthonormal_hook(model, unused_x):
-    if model.training == False:
+    if not model.training:
+        return
+    
+    model.ortho_constrain_count = (model.ortho_constrain_count + 1) % 2
+    if model.ortho_constrain_count != 0:
         return
 
     with torch.no_grad():
@@ -100,6 +104,8 @@ class ChainModel(nn.Module):
 
         assert len(kernel_size_list) == len(subsampling_factor_list)
         num_layers = len(kernel_size_list)
+        
+        self.ortho_constrain_count = 0
 
         input_dim = feat_dim * 3 + ivector_dim
         
--- a/egs/aishell/s10/chain/train.py
+++ b/egs/aishell/s10/chain/train.py
@@ -30,7 +30,7 @@ from common import save_checkpoint
 from common import save_training_info
 from common import setup_logger
 from device_utils import allocate_gpu_devices
-from egs_dataset import get_egs_dataloader
+from egs_dataloader import get_egs_dataloader
 from libs.nnet3.train.dropout_schedule import _get_dropout_proportions
 from model import get_chain_model
 from options import get_args
@@ -40,9 +40,9 @@ def get_objf(batch, model, device, criterion, opts, den_graph, training, optimiz
     total_weight = 0.
     total_frames = 0.  # for display only
     
-    key_list, feature_list, supervision_list = batch
-    assert len(key_list) == len(feature_list) == len(supervision_list)
-    batch_size = len(key_list)
+    feature_list, supervision_list = batch
+    assert len(feature_list) == len(supervision_list)
+    batch_size = len(feature_list)
     for n in range(batch_size):
         feats = feature_list[n]
         assert feats.ndim == 3
@@ -98,7 +98,7 @@ def get_validation_objf(dataloader, model, device, criterion, opts, den_graph):
  
     model.eval()
 
-    for batch_idx, batch in enumerate(dataloader):
+    for batch_idx, (pseudo_epoch, batch) in enumerate(dataloader):
         objf, weight, frames = get_objf(
             batch, model, device, criterion, opts, den_graph, False) 
         total_objf += objf
@@ -115,8 +115,13 @@ def train_one_epoch(dataloader, valid_dataloader, model, device, optimizer, crit
     total_frames = 0.  # for display only
 
     model.train()
-    for batch_idx, batch in enumerate(dataloader):
-        data_fraction = (batch_idx + current_epoch *
+    # iterates over one training scp file in one `pseudo_epoch`, 
+    # so one `pseudo_epoch` may contain many `batch`.
+    for batch_idx, (pseudo_epoch, batch) in enumerate(dataloader):
+        # `len(dataloader)` returns the number of `pseudo_epoch`
+        # in the current worker, that is the number of scp files
+        # we will process in this worker.
+        data_fraction = (pseudo_epoch + current_epoch *
                          len(dataloader)) / (len(dataloader) * num_epochs)
         _, dropout = _get_dropout_proportions(
             dropout_schedule, data_fraction)[0]
@@ -129,11 +134,12 @@ def train_one_epoch(dataloader, valid_dataloader, model, device, optimizer, crit
 
         if batch_idx % 100 == 0:
             logging.info(
-                'Device ({}) processing {}/{}({:.6f}%) global average objf: {:.6f} over {} '
+                'Device ({}) processing batch {}, current pseudo-epoch is {}/{}({:.6f}%), '
+                'global average objf: {:.6f} over {} '
                 'frames, current batch average objf: {:.6f} over {} frames, epoch {}'
                 .format(
-                    device.index, batch_idx, len(dataloader),
-                    float(batch_idx) / len(dataloader) * 100,
+                    device.index, batch_idx, pseudo_epoch, len(dataloader),
+                    float(pseudo_epoch) / len(dataloader) * 100,
                     total_objf / total_weight, total_frames,
                     curr_batch_objf / curr_batch_weight, 
                     curr_batch_frames, current_epoch))
@@ -154,21 +160,21 @@ def train_one_epoch(dataloader, valid_dataloader, model, device, optimizer, crit
             if tf_writer:
                 tf_writer.add_scalar('train/global_valid_average_objf',
                                  total_valid_objf / total_valid_weight,
-                                 batch_idx + current_epoch * len(dataloader))
+                                 pseudo_epoch + current_epoch * len(dataloader))
         # rank == None means we are not using ddp
-        if (rank == None or rank == 0) and batch_idx % 100 == 0 and tf_writer != None:
+        if (rank is None or rank == 0) and batch_idx % 100 == 0 and tf_writer is not None:
             tf_writer.add_scalar('train/global_average_objf',
                                  total_objf / total_weight,
-                                 batch_idx + current_epoch * len(dataloader))
+                                 pseudo_epoch + current_epoch * len(dataloader))
             tf_writer.add_scalar(
                 'train/current_batch_average_objf',
                 curr_batch_objf / curr_batch_weight,
-                batch_idx + current_epoch * len(dataloader))
+                pseudo_epoch + current_epoch * len(dataloader))
             
             tf_writer.add_scalar(
                 'train/current_dropout',
                 dropout,
-                batch_idx + current_epoch * len(dataloader))
+                pseudo_epoch + current_epoch * len(dataloader))
 
             state_dict = model.state_dict()
             for key, value in state_dict.items():
@@ -183,8 +189,7 @@ def train_one_epoch(dataloader, valid_dataloader, model, device, optimizer, crit
 
                 tf_writer.add_scalar(
                     'train/parameters/{}'.format(key), frobenius_norm,
-                    batch_idx + current_epoch * len(dataloader))
-
+                    pseudo_epoch + current_epoch * len(dataloader))
     return total_objf / total_weight
 
 def main():
@@ -198,23 +203,23 @@ def main():
             process_job(learning_rate, local_rank=local_rank)
         else:
             proc = []
-            if args.device_ids != None:
+            if args.device_ids is not None:
                 assert len(args.device_ids) >= args.world_size
             for i in range(args.world_size):
-                device_id = None if args.device_ids == None else args.device_ids[i]
+                device_id = None if args.device_ids is None else args.device_ids[i]
                 p = Process(target=process_job, args=(learning_rate, device_id, i))
                 proc.append(p)
                 p.start()
             for p in proc:
                 p.join()
     else:
-        device_id = None if args.device_ids == None else args.device_ids[0]
+        device_id = None if args.device_ids is None else args.device_ids[0]
         process_job(args.learning_rate, device_id)
 
 
 def process_job(learning_rate, device_id=None, local_rank=None):
     args = get_args()
-    if local_rank != None:    
+    if local_rank is not None:    
         setup_logger('{}/logs/log-train-rank-{}'.format(args.dir, local_rank),
                  args.log_level)
     else:
@@ -222,11 +227,11 @@ def process_job(learning_rate, device_id=None, local_rank=None):
 
     logging.info(' '.join(sys.argv))
 
-    if torch.cuda.is_available() == False:
+    if not torch.cuda.is_available():
         logging.error('No GPU detected!')
         sys.exit(-1)
 
-    if device_id == None:
+    if device_id is None:
         devices = allocate_gpu_devices(1)
         if len(devices) < 1:
             logging.error('Allocate GPU failed!')
@@ -334,10 +339,7 @@ def process_job(learning_rate, device_id=None, local_rank=None):
 
             if tf_writer:
                 tf_writer.add_scalar('learning_rate', curr_learning_rate, epoch)
-            
-            if dataloader.sampler and isinstance(dataloader.sampler, DistributedSampler):
-                dataloader.sampler.set_epoch(epoch)
-
+           
             objf = train_one_epoch(dataloader=dataloader,
                                    valid_dataloader=valid_dataloader,
                                    model=model,
@@ -383,7 +385,6 @@ def process_job(learning_rate, device_id=None, local_rank=None):
                             learning_rate=curr_learning_rate,
                             objf=objf,
                             local_rank=local_rank)
-
             epoch_info_filename = os.path.join(args.dir,
                                                'epoch-{}-info'.format(epoch))
             save_training_info(filename=epoch_info_filename,
@@ -394,7 +395,6 @@ def process_job(learning_rate, device_id=None, local_rank=None):
                                best_objf=best_objf,
                                best_epoch=best_epoch,
                                local_rank=local_rank)
-
     except KeyboardInterrupt:
         # save the model when ctrl-c is pressed
         model_path = os.path.join(args.dir,
--- a/egs/aishell/s10/chain/train.py
+++ b/egs/aishell/s10/chain/train.py
@@ -36,57 +36,48 @@ from model import get_chain_model
 from options import get_args
 
 def get_objf(batch, model, device, criterion, opts, den_graph, training, optimizer=None, dropout=0.):
-    total_objf = 0.
-    total_weight = 0.
-    total_frames = 0.  # for display only
-    
-    feature_list, supervision_list = batch
-    assert len(feature_list) == len(supervision_list)
-    batch_size = len(feature_list)
-    for n in range(batch_size):
-        feats = feature_list[n]
-        assert feats.ndim == 3
+    feature, supervision = batch
+    assert feature.ndim == 3
 
-        # at this point, feats is [N, T, C]
-        feats = feats.to(device)
-        if training:
-            nnet_output, xent_output = model(feats, dropout=dropout)
-        else:
-            with torch.no_grad():
-                nnet_output, xent_output = model(feats)
+    # at this point, feature is [N, T, C]
+    feature = feature.to(device)
+    if training:
+        nnet_output, xent_output = model(feature, dropout=dropout)
+    else:
+        with torch.no_grad():
+            nnet_output, xent_output = model(feature)
 
-        # at this point, nnet_output is: [N, T, C]
-        # refer to kaldi/src/chain/chain-training.h
-        # the output should be organized as
-        # [all sequences for frame 0]
-        # [all sequences for frame 1]
-        # [etc.]
-        nnet_output = nnet_output.permute(1, 0, 2)
-        # at this point, nnet_output is: [T, N, C]
-        nnet_output = nnet_output.contiguous().view(-1,
-	       					nnet_output.shape[-1])
+    # at this point, nnet_output is: [N, T, C]
+    # refer to kaldi/src/chain/chain-training.h
+    # the output should be organized as
+    # [all sequences for frame 0]
+    # [all sequences for frame 1]
+    # [etc.]
+    nnet_output = nnet_output.permute(1, 0, 2)
+    # at this point, nnet_output is: [T, N, C]
+    nnet_output = nnet_output.contiguous().view(-1,
+                                            nnet_output.shape[-1])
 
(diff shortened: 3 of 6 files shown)


danpovey:
Great!!
RE dropout, I don't think it matters much how fine-grained it is, but make sure it goes to exactly zero at the end.
I'm not sure whether the slight difference in results is statistical noise or not.  Maybe more runs would clarify??  It's not necessarily true that shuffling order helps performance; I've seen the claim that deterministic order works better, IIRC.

qindazhu:
> Great!!
> RE dropout, I don't think it matters much how fine-grained it is, but make sure it goes to exactly zero at the end.
> I'm not sure whether the slight difference in results is statistical noise or not. Maybe more runs would clarify?? It's not necessarily true that shuffling order helps performance; I've seen the claim that deterministic order works better, IIRC.

OK, Let me run the training more times and check the result, thanks! I'll check the dropout as well.

qindazhu:
@danpovey run again makes the result a little better, updated the table above.

danpovey:
OK.  I'm still not 100% convinced that the results difference is just noise, but I'll merge anyway as the speed improvement is considerable, we can figure it out later.

egs/aishell/s10/chain/egs_dataloader.py
```diff
@@ -0,0 +1,230 @@
...
+                feat = torch.cat(
+                    (torch.from_numpy(feat), repeat_ivector), dim=1).numpy()
+            feat_list.append(feat)
+
+        batched_feat = np.stack(feat_list, axis=0)
+        assert batched_feat.shape[0] == batch_size
+
+        # -4 = -2 -2
+        # the first -2 is from extra left/right context
+        # the second -2 is from lda feats splicing
```
csukuangfj: this comment is not correct any more.

Initially, line 172 is `start_index += 1`, now it is `+= 2`.

We can simply remove this comment.

egs/aishell/s10/chain/egs_dataloader.py
```diff
@@ -0,0 +1,230 @@
...
+from multiprocessing import Process
+import datetime
+import glob
+import os
+
+import numpy as np
+import torch
+import torch.distributed as dist
+
+from torch.utils.data import DataLoader
```
csukuangfj: DataLoader is imported but not used.
This line can be deleted.

csukuangfj:
@qindazhu 
Do you know why the training time is dropped from 150 minutes to 90 minutes
by replacing the built-in `DataLoader` with the customized `DataLoader`?

qindazhu:
> @qindazhu
> Do you know why the training time is dropped from 150 minutes to 90 minutes
> by replacing the built-in `DataLoader` with the customized `DataLoader`?

In the previous approach(Dataset with built-in Dataloader), we read minibacth egs one by one with opening the corresponding ark file again and again.
In the current approach, we read minibatch egs with a `NnetChainExampleReader` which means we open one ark file only one time in one epoch, so it is expected to get faster in training.

## 
However, according to my testing, both of the two approaches suffer from a same issue as I mentioned above: the first epoch of training takes much more time than the following epochs (no stable, but easy to repro. I'm doing some testings to find the reason, may related to cache or something). I profile it and find that io speech in first epoch is really too slow, that is, reading egs (see below lines) is very slow in the first epoch:
https://github.com/kaldi-asr/kaldi/blob/e267a4c156db6860912586cffa6738714e977da5/egs/aishell/s10/chain/egs_dataloader.py#L102-L103
https://github.com/kaldi-asr/kaldi/blob/e267a4c156db6860912586cffa6738714e977da5/src/pybind/kaldi/table.py#L153-L154

As I have no exclusive machines at my hand, @falu @csukuangfj，could you run the two approaches on your machines and report the time stats？You can add below code to get function call stats of every epoch:
```
from pyinstrument import Profiler 
...
profiler = Profiler()
profiler.start()

objf = train_one_epoch(dataloader=dataloader,...

profiler.stop()
logging.info('device {}, epoch {}, profile {}'.format(device_id, epoch, profiler.output_text(unicode=True, color=True)))
```
you also need to install `pyinstrument` (better than `cProfiler`) with `pip`. Here is the output on machines for epoch 0 and epoch 1 (this pull request running, ddp with 4 GPUs on one machine).

epoch0


epoch1

Note that you could use `htop` or some other tools to watch the `DISK_READ_SPEECH`.

csukuangfj:
@qindazhu 
thanks.

The current customized dataloader does not support multithreads. The previous one
supports `num_wokers`, which I guess is implemented via multiple processes.
 
Can we implement a multithread dataloader in C++ and wrap it to Python using pybind11?


I will run benchmarks following the code you posted.

danpovey:
The issue about the 1st iteration being slow will be to do with how
Xiaomi's cluster works.  It does local caching of files; file access can be
very slow the first time.
I don't think multi-threaded dataloaders should be necessary, because it
should be fast enough with just one thread.  But we'd have to test that.
If it were multi-threaded, it would probably make sense to have different
threads consume different scp files.

qindazhu:
@csukuangfj I agree with @danpovey 's point, as accoding to my experiment, except epoch 0, epoch 1-5 could be done in 50 minutes on Xiaomi's cluster, that means if we can make epoch 0 faster,  the speed of training is fast enough. As the profiler shows in above image of epoch 2, I think multiple num_workers would not make much improvment in speed.

BTW, if it is necessary to support multi-threaded dataloader, we could do all of things in Python, there's no need to wrap C++, as python just calls into C++ `ExampleReader` which is the main IO bottleneck. We can do just same thing as the built-in `Dataloader` does.

fanlu:
On our cluster, the training time taken by epoch 0 and epoch 1 are similar. Maybe I have no caching strategy on disk and the speed is very slow. Let my try to run benchmark code to figure it out.
image
image

qindazhu:
Thanks fanlu. Can you kindly report the training time of the origianl approach (built-in Dataloader) as well for comparion?

fanlu:
These are training time of epoch 0 and epoch 1 with built-in dataloader
image
image

qindazhu:
OK, so if there is no cache, the speed improvement is about 30mins for 6 epoch? Many thanks, @fanlu

fanlu:
Maybe the differences is not very large between this two dataloader. It's must be the clip_grad_norm_ takes a lot of time in training.

qindazhu:
Yes, maybe you could attach the log generated by `pyinstrument` to see the time stats of each function call after your benchmark running done. Thanks!

fanlu:
these are `pyinstrument` log of epoch 0 and epoch 1 on one machine
image

image

danpovey:
Couple observations:
  'forward' in chain_loss.py is taking way too much time.  Maybe that code
needs to be re-optimized for this architecture?  Or maybe it's being called
wrong, like in a loop or something?
 constrain_orthonormal is taking too much time.  You can run it every two
to four iterations to save time, it doesn't need to run every time.

danpovey:
Also it looks to me like whenever it calls C++ code it gets counted as
[self] which is not very helpful.  We should figure out a way to get that
info.

qindazhu:
Yes, the first and second `self` take so much time. Anyway, let me try to get an exclusive machine and run&profile it further. BTW @fanlu, IO speech on your cluster is so fast, reading 110G data in 86 seconds, more than 1G/s!!! Adoring!!

fanlu:
Maybe @csukuangfj can get `constrain_orthonormal` per 4 iterations when using `register_forward_pre_hook` back. 
BTW, I am very curious about the disk speed also. We have a team to maintain the shared file system

csukuangfj:
@qindazhu 

You can change
https://github.com/kaldi-asr/kaldi/blob/3f64b592c3b0c1c5dfd2069f7876ac54185a0433/egs/aishell/s10/chain/model.py#L72-L80

First, add a member variable in 

https://github.com/kaldi-asr/kaldi/blob/3f64b592c3b0c1c5dfd2069f7876ac54185a0433/egs/aishell/s10/chain/model.py#L96

```python
self.count = 0
```

Then, inside https://github.com/kaldi-asr/kaldi/blob/3f64b592c3b0c1c5dfd2069f7876ac54185a0433/egs/aishell/s10/chain/model.py#L72-L74

```python
model.count = (model.count + 1) % 4

if model.count != 0:
   return
```

Note that we cannot follow Kaldi's c++ implementation to use a random number here
because of distributed training.  We have to ensure that the model parameters
on every GPU before forwarding are the same.

qindazhu:
ok, I'll try tomorrow.

csukuangfj:
@qindazhu 
I've run the training but get a different result for CER. The code is nearly the same with you.
I only specified which 4 GPUs are used.

Screen Shot 2020-03-20 at 07 10 22

 ||TDNN-F(Pytorch, Adam, delta dropout without ivector ) from @fanlu | TDNN-F(Pytorch, Adam, delta dropout without ivector ) this PR 2nd run | TDNN-F(Pytorch, Adam, delta dropout without ivector ) this PR 1st run | this Pr with commit 0d8aada to make dropout go to zero at the end | this PR run by fangjun|
|--|--|--|--|--|--|
|dev_cer|6.10|6.13|6.18|6.12|7.02|
|dev_wer|13.86|13.89|13.96|13.92|14.17|
|test_cer|7.14|7.19|7.20|7.26|8.04|
|test_wer|15.49|15.54|15.66|15.63|15.73|
|training_time|151mins|88mins|84mins||71mins|

------

The training time is read from tensorboard. The relative time of the last step of `train/current_batch_average_objf` is 1 hour 11 minutes.
Screen Shot 2020-03-20 at 07 16 43

---------

The training time for epoch 0 and epoch 1 does not differ too much. Both of them are about 13 minutes. See the screenshots below.

Screen Shot 2020-03-20 at 07 19 23

Screen Shot 2020-03-20 at 07 19 32

------

I will run the benchmark later.

qindazhu:
@csukuangfj can you check your `best-epoch-info` file? I guess it did not save the checkpoint of `epoch 5`, and did you run steps `20`  that does `align_eg_numbers`?

https://github.com/kaldi-asr/kaldi/blob/e267a4c156db6860912586cffa6738714e977da5/egs/aishell/s10/local/run_chain.sh#L228-L232

csukuangfj:
@qindazhu 

I set `stage==0` in both `run.sh` and `local/run_chain.sh`, so it should run through all
stages.

`cat exp/chain_pybind/tdnn_sp/train/best-epoch-info`

```bash
model_path: exp/chain_pybind/tdnn_sp/train/best_model.pt
epoch: 5
learning rate: 4.0960000000000014e-05
objf: -0.04955414119370598
best objf: -0.04955414119370598
best epoch: 5
```

qindazhu:
OK. It is strange that you get nearly the same `objf` with me but so bad dev/test result
```
model_path: exp/chain_pybind/tdnn_sp/train/best_model.pt
epoch: 5
learning rate: 4.0960000000000014e-05
objf: -0.05007000607657938
best objf: -0.05007000607657938
best epoch: 5

```

And it seems that CER is terriblely bad, is there any script change of your local `compute_cer`? I suspect that you would get bad CER as well with the original built-in Dataloader approach.
@fanlu could you show your running result here for comparion?

csukuangfj:
I checked out your branch and made the following changes. I do NOT think such changes
will have any impact on the final CER.

```
diff --git a/egs/aishell/s10/cmd.sh b/egs/aishell/s10/cmd.sh
index cf68c5e..49b529c 100644
--- a/egs/aishell/s10/cmd.sh
+++ b/egs/aishell/s10/cmd.sh
@@ -10,9 +10,10 @@
 # conf/queue.conf in http://kaldi-asr.org/doc/queue.html for more information,
 # or search for the string 'default_config' in utils/queue.pl or utils/slurm.pl.

-export train_cmd="queue.pl -q all.q --mem 4G"
-export decode_cmd="queue.pl -q all.q --mem 4G"
-export mkgraph_cmd="queue.pl -q all.q --mem 8G"
-export cuda_train_cmd="queue.pl -q g.q --mem 4G"
-export cuda_inference_cmd="queue.pl -q g.q --mem 4G"
+
+export train_cmd="run.pl"
+export decode_cmd="run.pl"
+export mkgraph_cmd="run.pl"
+export cuda_train_cmd="run.pl"
+export cuda_inference_cmd="run.pl"

diff --git a/egs/aishell/s10/local/run_chain.sh b/egs/aishell/s10/local/run_chain.sh
index 56a848c..1659832 100755
--- a/egs/aishell/s10/local/run_chain.sh
+++ b/egs/aishell/s10/local/run_chain.sh
@@ -5,14 +5,14 @@

 set -e

-stage=22
+stage=0

 nj=10

 train_set=train
 gmm=tri3
 nnet3_affix=_pybind
-tree_affix=
+tree_affix=
 tdnn_affix=
 train_affix=
 online_cmvn=true
@@ -262,7 +262,7 @@ if [[ $stage -le 21 ]]; then

   num_epochs=$num_epochs
   lr=1e-3
-
+
   # use_ddp = false & world_size = 1: training model with one GPU
   # use_ddp = true & use_multiple_machine = false: training model with multiple GPUs on a single machine
   # use_ddp = true & use_multiple_machine = true:  training model with GPU on multiple machines
@@ -271,14 +271,14 @@ if [[ $stage -le 21 ]]; then
   world_size=$num_workers
   use_multiple_machine=false
   # you can assign GPUs with --device-ids "$device_ids"
-  # device_ids="4, 5, 6, 7"
+  device_ids="0, 1, 2, 3"
   if $use_multiple_machine ; then
     # suppose you are using Sun GridEngine
     cuda_train_cmd="$cuda_train_cmd --gpu 1 JOB=1:$world_size $dir/$train_dir/logs/job.JOB.log"
   else
     cuda_train_cmd="$cuda_train_cmd --gpu $world_size $dir/$train_dir/logs/train.log"
   fi
-
+
   $cuda_train_cmd python3 ./chain/train.py \
         --bottleneck-dim $bottleneck_dim \
         --checkpoint=${train_checkpoint:-} \
diff --git a/egs/aishell/s10/run.sh b/egs/aishell/s10/run.sh
index fb6d6e1..3f1260f 100755
--- a/egs/aishell/s10/run.sh
+++ b/egs/aishell/s10/run.sh
@@ -21,9 +21,10 @@
 . ./path.sh

 data=/home/asr-corpus/public/aishell
+data=/home/fangjunkuang/data/aishell
 data_url=www.openslr.org/resources/33

-nj=30
+nj=10

 stage=13
 . utils/parse_options.sh || exit 1;
@@ -100,5 +101,5 @@ if [[ $stage -le 12 ]]; then
 fi

 if [[ $stage -le 13 ]]; then
-  local/run_chain.sh --nj $nj
+  local/run_chain.sh --nj $nj
 fi
```

fanlu:
this is my result with that the exp start at stage 13.
image

qindazhu:
Thanks @fanlu.

@csukuangfj Yes, I think those changes will not affect the results. Can you run your original approach to see the result? Starting from `stage=21` (training stage) should be fine. Just checkout your original `chain/train.py` file from branch `pybind11`

```
git checkout your-local-pybind11-branch -- chain/train.py
```

csukuangfj:
@qindazhu 
I am running the benchmark right now. Once it is finished, I will revert your changes and run it again.

qindazhu:
OK. I have gotten similar benchmark result with fanlu's on a local machine (cannot get 1G/s read speed on our cluster...). It seems now the customized `Dataloader` will not lead to better speed performance if IO is not the bottleneck. I'm now running the  `quarter_version_constrain_orthonormal` and will paste the result after it's done.

fanlu:
@qindazhu @danpovey In my exps, the shuffling order strategy helps performance with aishell dataset. Maybe because this dataset is very small. 
We can shuffle the scps in random order also if we can only read a scp sequentially.
and `torch.nn.utils.clip_grad_norm_` can also helps the performance. just a little bit, less than 0.1, but it cost too much time in calculation.

|exp new loader|test cer | test wer | dev cer | dev wer|glb|validation|output-affine|
|---|---|---|---|---|---|---|---|
|no ivector new loader clip_norm_10 shuffle|7.15|15.48|6.02|13.76| -0.047656|-0.045833|40.32|

image

fanlu:
hi @qindazhu, if customized `Dataloader` do not lead to better speed performance, should we use another version that you have posted? then we can reduce the disk space usage
```
 scp_rspecifier = scp_file_to_process
 egs_rspecifier = 'ark,bg:nnet-chain-copy-egs --frame-shift .. scp:scp_rspecifier ark:- | \
                        nnet3-chain-shuffle-egs --buffer-size .. --srand .. ark:- ark:- | \
                        nnet3-chain-merge-egs --minibatch-size .. ark:- ark:- |'
 with SequentialNnetChainExampleReader(egs_rspecifier) as example_reader:
       for key, eg in example_reader:
             batch = self.collate_fn(eg)
             yield pseudo_epoch, batch
```
> OK. I have gotten similar benchmark result with fanlu's on a local machine (cannot get 1G/s read speed on our cluster...). It seems now the customized `Dataloader` will not lead to better speed performance if IO is not the bottleneck. I'm now running the `quarter_version_constrain_orthonormal` and will paste the result after it's done.

qindazhu:
Ok, I run it with shuffle before but with performance gain less than 0.05 so I remove it, so maybe that's becasue of noise. 

Yes, we will definitely remove `clip_grad_norm_` as @aadps is working on `max-change`

qindazhu:
@fanlu, yes, we can try this approach `copy|shuffle|merge`. However, there is a trap now of this approach, as Pytorch DDP assumes each worker should load same `num-minibatches` in one epoch, so we need to get the number of egs of each worker before Pytorch script starting and make sure each worker holds the same num-egs, otherwise it will hang after `epoch 5` (actually it hangs at the end of each epoch, but `enumerate(dataloder)` of the next epoch will reset it..).

But anyway, for now, at least we can test the speed instead of the performace caused by this issue (will not save checkpoint of `epoch 5`)

qindazhu:
I mean, if that approach can get at least same speed with current approach as it merges egs again and again. I have not tested the speed on a local machine, but it is expected to be a little slower. Of course, the good is disk usage saving.

fanlu:
Yse, I see, Maybe `drop_last=True` like `pytorch` is needed in egs generating procedure. Is there a way to use raw_egs?

qindazhu:
actually no as far as I know, but we can write this as it's not so complex, I mean after we test this approach and find the speed and performace is as good as the original approach.

csukuangfj:
Screen Shot 2020-03-20 at 12 23 00 PM
Screen Shot 2020-03-20 at 12 23 23 PM

> Note that you could use htop or some other tools to watch the DISK_READ_SPEECH.

what is `DISK_READ_SPEECH` ?

I am unable to get a properly formatted output of `pyinstrument.` Only epoch 1 is available
and is shown in the above screenshots.

The above screenshots are from `cat train.log`. There is an error for 

> logging.info('device {}, epoch {}, profile {}'.format(device_id, epoch, profiler.output_text(unicode=True, color=True)))

Screen Shot 2020-03-20 at 12 25 47 PM

csukuangfj:
By the way, what is `[self]` in the profiling output ?

qindazhu:
sorry, that's a typo, `DISK_READ_SPEED`

you can just do `export LANG=en_US.UTF8`  and run it. I guess Unicode is not supported on your system by default.

Regarding 'self', I'm now trying to get the detail info about it, I'll sync with all of you then.

qindazhu:
||TDNN-F(Pytorch, Adam, delta dropout without ivector ) , built-in dataloader | TDNN-F(Pytorch, Adam, delta dropout without ivector )  built-in dataloader,  quarter_version_constrain_orthonormal | 
|--|--|--|
|dev_cer|6.09|6.61
|dev_wer|13.89|14.56
|test_cer|7.14|7.83
|test_wer|15.52|16.34
|training_time|56 mins | 52mins

Note that all runnings are on a local machine, seems it's not worth decreasing the frequency of `constrain_orthonormal`. 

Let me focus on the `[self]` issue.  Guys, you can just do your work, do not waste so much time on it now. I'll update detail later. Thanks!

danpovey:
RE LANG=en_US.UTF8 :  it's important for all Kaldi scripts to not rely on
externally set locales: that is, if a script needs a certain locale it
should set it itself.  (But that may not be the issue here).

Regarding random shuffling: one option is to create a fairly large number
of .scp files and, as you mentioned, randomize the order of scp files.
Another option would be to give randomize_egs.sh a --num-repeats option,
and instead of doing
utils/shuffle_list.pl  $processed_egs_dir/train.scp  ...
it would do:
for n in $(seq $num_repeats); do cat $processed_egs_dir/train.scp; done |
utils/shuffle_list.pl   ...

HOWEVER: I just realized that steps/chain2/randomize_egs.sh does not do
what I thought it does.
It was supposed to ensure locality on disk by shuffling blocks of
consecutive minibatches, but as you can
see from the command above, it just randomizes them globally.  So in fact
my rationale for the change we
made (that Haowen implemented) is incorrect, with that script as it is.  It
wouldn't be that hard to fix that
though.

qindazhu:
Thanks Dan. Regarding `locales`, the issue here is from a profile tool `pyinstrument` which will not included in our script finally, so it's just OK.

RE `randomize_egs`, let me do line-by-line profiling first as it seems the bottleneck now is `get_objf`, and then fix the issue of current `randomize_egs.sh`

qindazhu:
Here is line-by-line time stats of `[self]` in `get_objf` , you can see that it takes much time to do `line 87: objf_l2_term_weight.detach().cpu() `
test

<pre>
Timer unit: 1 s

Total time: 409.24 s
./chain/train.py
Function: get_objf at line 41

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    41                                           def get_objf(batch, model, device, criterion, opts, den_graph, training, optimizer=None, dropout=0.):
    42      2422          0.0      0.0      0.0      total_objf = 0.
    43      2422          0.0      0.0      0.0      total_weight = 0.
    44      2422          0.0      0.0      0.0      total_frames = 0.  # for display only
    45
    46      2422          0.0      0.0      0.0      key_list, feature_list, supervision_list = batch
    47      2422          0.0      0.0      0.0      assert len(key_list) == len(feature_list) == len(supervision_list)
    48      2422          0.0      0.0      0.0      batch_size = len(key_list)
    49      4844          0.0      0.0      0.0      for n in range(batch_size):
    50      2422          0.0      0.0      0.0          feats = feature_list[n]
    51      2422          0.0      0.0      0.0          assert feats.ndim == 3
    52
    53                                                   # at this point, feats is [N, T, C]
    <b>54      2422          4.6      0.0      1.1          feats = feats.to(device)</b>
    55      2422          0.0      0.0      0.0          if training:
    56      2422         79.9      0.0     19.5              nnet_output, xent_output = model(feats, dropout=dropout)
    57                                                   else:
    58                                                       with torch.no_grad():
    59                                                           nnet_output, xent_output = model(feats)
    60
    61                                                   # at this point, nnet_output is: [N, T, C]
    62                                                   # refer to kaldi/src/chain/chain-training.h
    63                                                   # the output should be organized as
    64                                                   # [all sequences for frame 0]
    65                                                   # [all sequences for frame 1]
    66                                                   # [etc.]
    67      2422          0.0      0.0      0.0          nnet_output = nnet_output.permute(1, 0, 2)
    68                                                   # at this point, nnet_output is: [T, N, C]
    69      2422          0.1      0.0      0.0          nnet_output = nnet_output.contiguous().view(-1,
    70      2422          0.1      0.0      0.0                                                 nnet_output.shape[-1])
    71
    72                                                   # at this point, xent_output is: [N, T, C]
    73      2422          0.0      0.0      0.0          xent_output = xent_output.permute(1, 0, 2)
    74                                                   # at this point, xent_output is: [T, N, C]
    75      2422          0.1      0.0      0.0          xent_output = xent_output.contiguous().view(-1,
    76      2422          0.0      0.0      0.0                                                 xent_output.shape[-1])
    77      2422          0.0      0.0      0.0          objf_l2_term_weight = criterion(opts, den_graph,
    78      2422          0.0      0.0      0.0                                     supervision_list[n], nnet_output,
    79      2422         92.7      0.0     22.7                                     xent_output)
    80      2422          0.0      0.0      0.0          objf = objf_l2_term_weight[0]
    81      2422          0.0      0.0      0.0          if training:
    82      2422          1.8      0.0      0.4              optimizer.zero_grad()
    83      2422         40.0      0.0      9.8              objf.backward()
    84      2422          6.3      0.0      1.5              clip_grad_value_(model.parameters(), 5.0)
    85      2422         27.5      0.0      6.7              optimizer.step()
    86
    <b>87      2422        156.0      0.1     38.1          objf_l2_term_weight = objf_l2_term_weight.detach().cpu() </b>
    88
    89      2422          0.1      0.0      0.0          total_objf += objf_l2_term_weight[0].item()
    90      2422          0.0      0.0      0.0          total_weight += objf_l2_term_weight[2].item()
    91      2422          0.0      0.0      0.0          num_frames = nnet_output.shape[0]
    92      2422          0.0      0.0      0.0          total_frames += num_frames
    93
    94      2422          0.0      0.0      0.0      return total_objf, total_weight, total_frames
</pre>

A strange thing is that if I commment `objf_l2_term_weight.detach().cpu()` related code, then `feats = feats.to(device)` will take most of time...I run this for several times but get the same behavior

<pre>
Timer unit: 1 s

Total time: 324.423 s
./chain/train.py
Function: get_objf at line 41

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    41                                           def get_objf(batch, model, device, criterion, opts, den_graph, training, optimizer=None, dropout=0.):
    42      2422          0.0      0.0      0.0      total_objf = 0.
    43      2422          0.0      0.0      0.0      total_weight = 0.
    44      2422          0.0      0.0      0.0      total_frames = 0.  # for display only
    45
    46      2422          0.0      0.0      0.0      key_list, feature_list, supervision_list = batch
    47      2422          0.0      0.0      0.0      assert len(key_list) == len(feature_list) == len(supervision_list)
    48      2422          0.0      0.0      0.0      batch_size = len(key_list)
    49      4844          0.0      0.0      0.0      for n in range(batch_size):
    50      2422          0.0      0.0      0.0          feats = feature_list[n]
    51      2422          0.0      0.0      0.0          assert feats.ndim == 3
    52
    53                                                   # at this point, feats is [N, T, C]
   <b> 54      2422         74.2      0.0     22.9          feats = feats.to(device)</b>
    55      2422          0.0      0.0      0.0          if training:
    56      2422         81.4      0.0     25.1              nnet_output, xent_output = model(feats, dropout=dropout)
    57                                                   else:
    58                                                       with torch.no_grad():
    59                                                           nnet_output, xent_output = model(feats)
    60
    61                                                   # at this point, nnet_output is: [N, T, C]
    62                                                   # refer to kaldi/src/chain/chain-training.h
    63                                                   # the output should be organized as
    64                                                   # [all sequences for frame 0]
    65                                                   # [all sequences for frame 1]
    66                                                   # [etc.]
    67      2422          0.0      0.0      0.0          nnet_output = nnet_output.permute(1, 0, 2)
    68                                                   # at this point, nnet_output is: [T, N, C]
    69      2422          0.1      0.0      0.0          nnet_output = nnet_output.contiguous().view(-1,
    70      2422          0.1      0.0      0.0                                                 nnet_output.shape[-1])
    71
    72                                                   # at this point, xent_output is: [N, T, C]
    73      2422          0.0      0.0      0.0          xent_output = xent_output.permute(1, 0, 2)
    74                                                   # at this point, xent_output is: [T, N, C]
    75      2422          0.1      0.0      0.0          xent_output = xent_output.contiguous().view(-1,
    76      2422          0.0      0.0      0.0                                                 xent_output.shape[-1])
    77      2422          0.0      0.0      0.0          objf_l2_term_weight = criterion(opts, den_graph,
    78      2422          0.0      0.0      0.0                                     supervision_list[n], nnet_output,
    79      2422         92.2      0.0     28.4                                     xent_output)
    80      2422          0.0      0.0      0.0          objf = objf_l2_term_weight[0]
    81      2422          0.0      0.0      0.0          if training:
    82      2422          1.7      0.0      0.5              optimizer.zero_grad()
    83      2422         40.5      0.0     12.5              objf.backward()
    84      2422          6.7      0.0      2.1              clip_grad_value_(model.parameters(), 5.0)
    85      2422         27.3      0.0      8.4              optimizer.step()
    86
   <b> 87                                                   #objf_l2_term_weight = objf_l2_term_weight.detach().cpu()
    88
    89      2422          0.0      0.0      0.0          total_objf += 1 #objf_l2_term_weight[0].item()
    90      2422          0.0      0.0      0.0          total_weight += 1 #objf_l2_term_weight[2].item()</b>
    91      2422          0.0      0.0      0.0          num_frames = nnet_output.shape[0]
    92      2422          0.0      0.0      0.0          total_frames += num_frames
    93
    94      2422          0.0      0.0      0.0      return total_objf, total_weight, total_frames
</pre>

**NOTE**
- Validation running is disabled in all testings.
- The first picture is from `pyinstrument` and the first table is from `line_profiler`, I get them in two seperate runnings as it seems they can not co-exist in one running, thus you may notice there are small time difference for function calls between them.

danpovey:
It could be about asynchronous launching of kernels, and the sync to CPU
causes it to wait for the entire queue.
I think pytorch has some kind of synchronous mode that you can use to get
more informative timing stats.

qindazhu:
OK, I will.

##
BTW, here is the stats of `forward` in `chain_loss.py`, the `[self]` inside it is just doing `chain.ComputeChainObjfAndDeriv`

<pre>
Total time: 91.9981 s
/home/dev-qiuhaowen/kaldi/egs/aishell/s10/chain/chain_loss.py
Function: forward at line 19

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    19                                               @staticmethod
    20                                               def forward(ctx, opts, den_graph, supervision, nnet_output_tensor,
    21                                                           xent_out_unused_tensor):
    22                                                   '''
    23                                                   opts:          Struct containing options
    24
    25                                                   den_graph:     The denominator graph, derived from denominator fst.
    26
    27                                                   supervision:   The supervision object, containing the supervision
    28                                                                  paths and constraints on the alignment as an FST.
    29
    30                                                   nnet_output_tensor:   The output of the neural net; dimension must
    31                                                                         equal ((supervision.num_sequences * supervision.frames_per_sequence)
    32                                                                         by den_graph.NumPdfs()).  The rows are ordered
    33                                                                         as: all sequences for frame 0; all sequences for
    34                                                                         frame 1; etc.
    35
    36                                                   xent_out_unused_tensor: It MUST have the same structure as `nnet_output_tensor`;
    37                                                                           it is passed but NOT used directly in this
    38                                                                           function; it is still needed because we need
    39                                                                           to do back propagation for it.
    40                                                   '''
    41                                                   global g_nnet_output_deriv_tensor
    42                                                   global g_xent_output_deriv_tensor
    43
    44      2422          0.0      0.0      0.0          if g_nnet_output_deriv_tensor is None or g_nnet_output_deriv_tensor.shape != nnet_output_tensor.shape:
    45                                                       # preallocate GPU memory in PyTorch and it will be shared with Kaldi
    46      1253          0.1      0.0      0.1              g_nnet_output_deriv_tensor = torch.zeros_like(nnet_output_tensor)
    47      1253          0.0      0.0      0.0              g_xent_output_deriv_tensor = torch.zeros_like(
    48      1253          0.0      0.0      0.0                  nnet_output_tensor).contiguous()
    49                                                       # (fangjun): Note that g_xent_output_deriv needs to be contiguous,
    50                                                       # i.e., stride == cols; otherwise, it will be re-allocated in Kaldi
    51                                                   else:
    52                                                       # (fangjun): we have to zero them out manually!
    53      1169          0.0      0.0      0.0              g_nnet_output_deriv_tensor.zero_()
    54      1169          0.0      0.0      0.0              g_xent_output_deriv_tensor.zero_()
    55
    56                                                   # it contains [objf, l2_term, weight] and will be returned to the caller
    57      2422          0.1      0.0      0.1          objf_l2_term_weight_tensor = torch.zeros(3).float()
    58
    59      2422          0.0      0.0      0.0          nnet_output = kaldi.PytorchToCuSubMatrix(to_dlpack(nnet_output_tensor))
    60
    61      2422          0.0      0.0      0.0          nnet_output_deriv = kaldi.PytorchToCuSubMatrix(
    62      2422          0.0      0.0      0.0              to_dlpack(g_nnet_output_deriv_tensor))
    63
    64      2422          0.0      0.0      0.0          xent_output_deriv = kaldi.PytorchToCuSubMatrix(
    65      2422          0.0      0.0      0.0              to_dlpack(g_xent_output_deriv_tensor))
    66
    67      2422          0.0      0.0      0.0          objf_l2_term_weight = kaldi.PytorchToSubVector(
    68      2422          0.0      0.0      0.0              to_dlpack(objf_l2_term_weight_tensor))
    69
    <b>70      2422          0.0      0.0      0.0          chain.ComputeChainObjfAndDeriv(opts=opts,
    71      2422          0.0      0.0      0.0                                         den_graph=den_graph,
    72      2422          0.0      0.0      0.0                                         supervision=supervision,
    73      2422          0.0      0.0      0.0                                         nnet_output=nnet_output,
    74      2422          0.0      0.0      0.0                                         objf_l2_term_weight=objf_l2_term_weight,
    75      2422          0.0      0.0      0.0                                         nnet_output_deriv=nnet_output_deriv,
    76      2422         91.0      0.0     98.9                                         xent_output_deriv=xent_output_deriv) </b>
    77
    78      2422          0.0      0.0      0.0          ctx.save_for_backward(g_nnet_output_deriv_tensor,
    79      2422          0.0      0.0      0.0                                g_xent_output_deriv_tensor)
    80
    81      2422          0.0      0.0      0.0          objf_l2_term_weight_tensor = objf_l2_term_weight_tensor.to(
    82      2422          0.7      0.0      0.7              nnet_output_tensor.device)
    83      2422          0.0      0.0      0.0          return objf_l2_term_weight_tensor

</pre>

danpovey:
Ah OK, thanks.

qindazhu:
After added `torch.cuda.synchronize()`, we can get more informative timing stats, you can see that it takes most of time to do `forward` and `backward`

<pre>
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    41                                           def get_objf(batch, model, device, criterion, opts, den_graph, training, optimizer=None, dropout=0.):
    ...
    55      2422          0.0      0.0      0.0          if training:
   <b> 56      2422          0.1      0.0      0.0              torch.cuda.synchronize()
    57      2422         80.3      0.0     18.3              nnet_output, xent_output = model(feats, dropout=dropout)
    58      2422         44.0      0.0     10.1              torch.cuda.synchronize()</b>

    79      2422          0.0      0.0      0.0          objf_l2_term_weight = criterion(opts, den_graph,
    80      2422          0.0      0.0      0.0                                     supervision_list[n], nnet_output,
    81      2422         47.3      0.0     10.8                                     xent_output)
    82      2422          0.0      0.0      0.0          objf = objf_l2_term_weight[0]
    83      2422          0.0      0.0      0.0          if training:
    84      2422          1.7      0.0      0.4              optimizer.zero_grad()
    <b>85      2422          0.1      0.0      0.0              torch.cuda.synchronize()
    86      2422         40.2      0.0      9.2              objf.backward()
    87      2422        185.5      0.1     42.4              torch.cuda.synchronize()</b>
    88      2422          6.3      0.0      1.4              clip_grad_value_(model.parameters(), 5.0)
    89      2422          0.1      0.0      0.0              torch.cuda.synchronize()
    90      2422         26.9      0.0      6.2              optimizer.step()
    91
    92      2422          0.1      0.0      0.0          objf_l2_term_weight = objf_l2_term_weight.detach().cpu()
    93
    94      2422          0.1      0.0      0.0          torch.cuda.synchronize()

   100      2422          0.0      0.0      0.0      return total_objf, total_weight, total_frames
</pre>

## Forward
Detailed stats in `forward` are captured as below, **It seems we can do few things to accelerate `forward`.**
<pre>
<b>File: ./chain/model.py </b>
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   155                                               def forward(self, x, dropout=0.):

   180
   <b>182     31486          0.1      0.0      0.1          for i in range(len(self.tdnnfs)):
   183     29064         27.4      0.0     33.6              x = self.tdnnfs[i](x, dropout=dropout)
   184
   185                                                   # at this point, x is [N, C, T]
   186
   187      2422         38.0      0.0     46.7          torch.cuda.synchronize()</b>

   189

   216      2422          0.0      0.0      0.0          return nnet_output, xent_output
</pre>
<pre>
<b> File: ./chain/tdnnf_layer.py: FactorizedTDNN->forward</b>
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   234                                               def forward(self, x, dropout=0.):
   235                                                   # input x is of shape: [batch_size, feat_dim, seq_len] = [N, C, T]
   236     29064          0.1      0.0      0.1          assert x.ndim == 3
   237
   238                                                   # save it for skip connection
   239     29064          0.0      0.0      0.0          input_x = x
   240
   241     29064          4.1      0.0      4.7          torch.cuda.synchronize()
   242     29064          6.3      0.0      7.3          x = self.linear(x)
   243                                                   # at this point, x is [N, C, T]
   244
   245     29064         21.7      0.0     25.0          torch.cuda.synchronize()
   246     29064          5.3      0.0      6.2          x = self.affine(x)
   247                                                   # at this point, x is [N, C, T]
   248
   249     29064          9.6      0.0     11.1          torch.cuda.synchronize()
   250     29064          1.3      0.0      1.5          x = F.relu(x)
   251
   252     29064          3.3      0.0      3.8          torch.cuda.synchronize()
   253                                                   # at this point, x is [N, C, T]
   254
   255     29064          6.1      0.0      7.1          x = self.batchnorm(x)
   256
   257     29064          9.0      0.0     10.3          torch.cuda.synchronize()
   258                                                   # at this point, x is [N, C, T]
   259
   260     29064          4.7      0.0      5.4          x = self.dropout(x, alpha=dropout)
   261
   262     29064          4.3      0.0      5.0          torch.cuda.synchronize()
   263     29064          0.3      0.0      0.3          if self.linear.kernel_size > 1:
   264     26642          2.8      0.0      3.2              x = self.bypass_scale * input_x[:, :, self.s:-self.s:self.s] + x
   265                                                   else:
   266      2422          0.3      0.0      0.3              x = self.bypass_scale * input_x[:, :, ::self.s] + x
   267     29064          7.5      0.0      8.7          torch.cuda.synchronize()
   268     29064          0.0      0.0      0.0          return x
</pre>
<pre>
<b>File: ./chain/tdnnf_layer.py: called by self.linear in forward of tdnnf</b>

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    98                                               def constrain_orthonormal(self):

   116
   117     36330          1.1      0.0      2.5          torch.cuda.synchronize()
   118     36330         31.6      0.0     72.7          w = _constrain_orthonormal_internal(w)
   119
   120     36330          1.5      0.0      3.4          torch.cuda.synchronize()
</pre>


## Backward
With `torch.autograd.profiler.profile`, we can inspect the cost on CPU and GPU of `backward`. Stats for one minibatch are showd as below table, you can see that **`cudnn_convolution_backward` takes much time.**
<pre>
Name                                 Self CPU total %  Self CPU total   CPU total %      CPU total        CPU time avg     CUDA total %     CUDA total       CUDA time avg    Number of Calls
-----------------------------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------
<b>CudnnConvolutionBackward             2.18%            465.799us        32.23%           6.881ms          245.746us        31.33%           51.178ms         1.828ms          28
cudnn_convolution_backward           30.05%           6.415ms          30.05%           6.415ms          229.111us        31.24%           51.039ms         1.823ms          28</b>
SliceBackward                        8.43%            1.800ms          19.74%           4.214ms          117.067us        5.66%            9.255ms          257.081us        36
NativeBatchNormBackward              2.39%            509.803us        6.40%            1.366ms          80.358us         3.98%            6.497ms          382.193us        17
native_batch_norm_backward           4.01%            856.286us        4.01%            856.286us        50.370us         3.93%            6.417ms          377.494us        17
MmBackward                           0.76%            162.236us        2.82%            601.736us        150.434us        2.99%            4.893ms          1.223ms          4
mm                                   1.33%            283.364us        1.33%            283.364us        35.421us         2.94%            4.805ms          600.576us        8
MulBackward0                         1.79%            382.652us        5.44%            1.160ms          46.420us         2.72%            4.444ms          177.767us        25
mul                                  3.64%            777.844us        3.64%            777.844us        31.114us         2.65%            4.323ms          172.933us        25
zeros                                2.78%            593.889us        9.42%            2.012ms          54.373us         1.92%            3.139ms          84.825us         37
ReluBackward0                        2.40%            511.411us        5.19%            1.108ms          73.875us         1.81%            2.950ms          196.676us        15
threshold_backward                   2.79%            596.721us        2.79%            596.721us        39.781us         1.76%            2.876ms          191.760us        15
zero_                                4.85%            1.036ms          4.85%            1.036ms          28.009us         1.68%            2.742ms          74.115us         37
add                                  2.14%            456.957us        2.14%            456.957us        35.151us         1.60%            2.615ms          201.176us        13
torch::autograd::AccumulateGrad      2.94%            627.360us        9.55%            2.040ms          41.632us         0.46%            745.442us        15.213us         49
LogSoftmaxBackward                   0.07%            15.789us         0.39%            82.241us         82.241us         0.37%            605.184us        605.184us        1
_log_softmax_backward_data           0.31%            66.452us         0.31%            66.452us         66.452us         0.37%            601.088us        601.088us        1
KaldiChainObjfFunctionBackward       0.31%            65.607us         0.65%            137.886us        137.886us        0.33%            536.576us        536.576us        1
mul_                                 0.34%            72.279us         0.34%            72.279us         36.139us         0.30%            491.520us        245.760us        2
add_                                 6.62%            1.413ms          6.62%            1.413ms          28.829us         0.30%            489.463us        9.989us          49
UnsqueezeBackward0                   2.71%            578.189us        6.42%            1.371ms          24.924us         0.25%            406.505us        7.391us          55
sum                                  0.87%            185.343us        0.87%            185.343us        46.336us         0.22%            356.351us        89.088us         4
reshape                              0.61%            129.685us        1.45%            308.553us        30.855us         0.19%            314.367us        31.437us         10
UnsafeViewBackward                   0.19%            40.705us         0.88%            187.212us        46.803us         0.18%            290.816us        72.704us         4
clone                                0.16%            33.325us         0.16%            33.325us         33.325us         0.15%            238.592us        238.592us        1
SqueezeBackward1                     1.36%            291.147us        3.01%            642.265us        22.938us         0.13%            209.939us        7.498us          28
squeeze                              3.71%            792.636us        3.71%            792.636us        14.412us         0.09%            143.361us        2.607us          55
SelectBackward                       0.26%            56.032us         0.66%            141.866us        141.866us        0.09%            140.288us        140.288us        1
empty                                1.79%            381.562us        1.79%            381.562us        10.312us         0.06%            102.403us        2.768us          37
slice                                2.22%            473.224us        2.22%            473.224us        13.145us         0.05%            86.035us         2.390us          36
ViewBackward                         0.31%            66.095us         1.07%            228.141us        38.023us         0.05%            74.752us         12.459us         6
unsqueeze                            1.64%            351.118us        1.64%            351.118us        12.540us         0.04%            68.607us         2.450us          28
PermuteBackward                      0.39%            83.092us         0.91%            194.985us        24.373us         0.04%            58.369us         7.296us          8
torch::autograd::CopyBackwards       0.33%            69.496us         0.51%            108.531us        18.089us         0.03%            46.080us         7.680us          6
unsigned short                       0.98%            209.220us        0.98%            209.220us        13.076us         0.02%            39.938us         2.496us          16
AddBackward0                         0.32%            69.244us         0.32%            69.244us         4.328us          0.02%            35.852us         2.241us          16
TBackward                            0.18%            38.669us         0.43%            91.753us         22.938us         0.02%            30.719us         7.680us          4
view                                 0.95%            201.977us        0.95%            201.977us        15.537us         0.02%            29.698us         2.284us          13
permute                              0.52%            111.893us        0.52%            111.893us        13.987us         0.01%            19.457us         2.432us          8
to                                   0.18%            39.035us         0.18%            39.035us         6.506us          0.01%            15.360us         2.560us          6
select                               0.07%            14.784us         0.07%            14.784us         14.784us         0.01%            14.336us         14.336us         1
torch::autograd::GraphRoot           0.04%            7.866us          0.04%            7.866us          7.866us          0.00%            5.120us          5.120us          1
_unsafe_view                         0.08%            16.493us         0.08%            16.493us         16.493us         0.00%            3.072us          3.072us          1
-----------------------------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------  ---------------
Self CPU time total: 21.350ms
CUDA time total: 163.374ms
</pre>

danpovey:
Mm.  The only easy opportunity for speedup that I see is to call the orthonormal-constraint code only every few minibatches.

qindazhu:
> TDNN-F(Pytorch, Adam, delta dropout without ivector ) , built-in dataloader	TDNN-F(Pytorch, Adam, delta dropout without ivector ) built-in dataloader, quarter_version_constrain_orthonormal
> dev_cer	6.09	6.61
> dev_wer	13.89	14.56
> test_cer	7.14	7.83
> test_wer	15.52	16.34
> training_time	56 mins	52mins
> Note that all runnings are on a local machine, seems it's not worth decreasing the frequency of `constrain_orthonormal`.
> 
> Let me focus on the `[self]` issue. Guys, you can just do your work, do not waste so much time on it now. I'll update detail later. Thanks!

However, according to my previous experiment of `quarter_version_constrain_orthonormal`, seems it's not worth decreasing the frequency of `orthonormal-constraint`.

csukuangfj:
@qindazhu 
How about `**half**_version_constrain_orthonormal`?

danpovey:
I really doubt it would make such a difference to run it every 4th iteration.  I suspect a bug.

qindazhu:
I run those three version (`normal, half, quarter`) with same randomized scp files, here are the results

 ||TDNN-F(Pytorch, Adam, delta dropout without ivector ) scp_dataloader, normal_version_constrain_orthonormal |  half_version | quarter_version|
|--|--|--|--|
|dev_cer|6.22|6.18|6.25|
|dev_wer|14.02|13.95|14.10|
|test_cer|7.24|7.21|7.41|
|test_wer|15.67|15.61|15.85|
|training_time|55 mins | 53mins|51mins|

Seems it does exist some data bugs in my last running.

The better performace of `half_version` may come from nosie.

However, the performace loss of `quarter_version` is still notable. Considering that the speech performance gain is not so much, I think we should keep the default version, or with a samll minibatch?

qindazhu:
It seems that more scp files or shuffle would not make a big difference according to my experiments. All below results are with 4 workers and customized scp dataloader.

|| 14*4 = 56 scp files, no shuffle | 14*4, shuffle | 28*4, no shuffle | 28*4, shuffle | 41*4, no shuffle | 41*4, shuffle |
|--|--|--|--|--|--|--|
|dev_cer| 6.19| 6.10|6.08|6.15|6.11|6.12|
|dev_wer|14.01|13.90|13.91|13.95|13.84|13.92|
|test_cer|7.25|7.21|7.22|7.27|7.25|7.26|
|test_wer|15.65|15.54|15.68|15.61|15.58|15.64|

danpovey:
Cool, thanks!  Yes, it seems doing the orthonormal constraint more rarely doesn't make much speed difference, but let's keep doing it only every other frame, for speed (a lot of small differences will add up).
And it looks like there is a gain from shuffling the scp files, which is nice!  
LMK when you want me to merge this PR.

qindazhu:
Added a new commit for doing orthonormal constraint every other minibatch, @danpovey it's ready to merge now.
> Cool, thanks! Yes, it seems doing the orthonormal constraint more rarely doesn't make much speed difference, but let's keep doing it only every other frame, for speed (a lot of small differences will add up).
> And it looks like there is a gain from shuffling the scp files, which is nice!
> LMK when you want me to merge this PR.

danpovey:
Thanks!  Merging.

PR #4007, opened 2020-03-22
chime6: train_stats_sad_1a.sh add missing quotes

boeddeker:
I got the error `train_raw_rnn.py: error: unrecognized arguments: --norm-vars=false`.

I think the problem is the whitespace in line 65 (code: `cmvn_opts="--norm-means=false --norm-vars=false"`) and the missing quotes in line 101.

@aarora8 Could you check that this change is correct?
I am wondering, because when my bash understanding is correct, nobody should be able to execute the baseline.


aarora8:
Thanks yeah, it is correct, @desh2608 recently also found this bug.

desh2608:
Well, actually, @pzelasko found it!

desh2608:
@danpovey I think this can be merged.

PR #4008, opened 2020-03-23
fixing typos in bn recipe

keighrim:


egs/bn_music_speech/v1/local/make_bn.py
```diff
@@ -33,13 +33,13 @@
   for line in music_fi:
     left, right = line.rstrip().split(" ")
     segments = "{0}{1}-music-{2} {1} {3} {4}\n".format(segments, utt, count, left, right)
-    utt2spk = "{0}{1}-music-{2} {1}-music-{2}".format(utt2spk, utt,count)
+    utt2spk = "{0}{1}-music-{2} {1}-music-{2}\n".format(utt2spk, utt,count)
     count += 1
```
kkm000, line 37: Thanks, the `-wav` and the second `music` are good catches!

What worries me about this file is the way it organizes writes. Are the files large? 50-100 lines kinda ok as is, if 1000 or more, Python will generate O(N²) total length of every string and then have to garbage-collect them (you get `line1\n`, then `line1\nline2\n` and drop `line1\n` on the floor, then `line1\nline2\nline3\n` and drop `line1\nline2\n`... last lines become huge). This may become surprisingly slow. In this script it would be much more reasonable to write the files line-by-line.

There are a couple nice tricks that you could use to make the script more succinct. Scoping `with` can open all files at once using comma:
```py
with open(os.path.join(out_dir, "utt2spk"), 'w') as utt2spk_fi,
     open(os.path.join(out_dir, "segments"), 'w') as segments_fi:
```
and is more succinct that open-write-close. Also, you can get the line count and the line in a single `for`, if you like that more:
```py
  for count, line in enumerate(music_fi, 1):
```
`1` here is the starting value for `count` (defaulit is 0). `enumerate()` prepends the counter to each successive read, returning an immutable tuple. It's a built-in function that requires no `import`.

keighrim: Made another commit with those improvements you brought up.

egs/bn_music_speech/v1/local/make_bn.py
```diff
@@ -14,35 +14,26 @@
...
-    count += 1
-utt2spk_fi = open(os.path.join(out_dir, "utt2spk"), 'w')
-utt2spk_fi.write(utt2spk)
-segments_fi = open(os.path.join(out_dir, "segments"), 'w')
-segments_fi.write(segments)
+    for utt in utts:
+      music_filename = utt + "_music.key.refined"
+      speech_filename = utt + "_speech.key.refined"
+      music_fi = open(os.path.join(out_dir, music_filename), 'r').readlines()
+      speech_fi = open(os.path.join(out_dir, speech_filename), 'r').readlines()
```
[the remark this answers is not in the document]
keighrim: I greped all open statements in `*.py` and applied the same *fix* to all of them.

kkm000:
Thanks!

Issue #4010, opened 2020-03-23
chime6: train_diarizer.sh: missing local/make_voxceleb1_v2.pl


boeddeker:
In the file `kaldi/egs/chime6/s5_track2/local/train_diarizer.sh` the comments suggest to use  `local/make_voxceleb1_v2.pl` when voxceleb was downloaded recently.

https://github.com/kaldi-asr/kaldi/blob/b1d809386c17a9630d786b8fb0f67ed3bc173811/egs/chime6/s5_track2/local/train_diarizer.sh#L44-L45

The problem is, that this file does not exist in `chime6`.

I found this file 3 times in this repr. One symlink and two files.
```
.../kaldi $ find -iname "make_voxceleb1_v2.pl"
./egs/dihard_2018/v2/local/make_voxceleb1_v2.pl
./egs/dihard_2018/v1/local/make_voxceleb1_v2.pl
./egs/voxceleb/v1/local/make_voxceleb1_v2.pl
.../kaldi $ ll ./egs/dihard_2018/v2/local/make_voxceleb1_v2.pl
lrwxrwxrwx 1 cbj cbj 35 Jan 28 14:42 ./egs/dihard_2018/v2/local/make_voxceleb1_v2.pl -> ../../v1/local/make_voxceleb1_v2.pl*
.../kaldi $ ll ./egs/dihard_2018/v1/local/make_voxceleb1_v2.pl
-rwxrwxr-x 1 cbj cbj 4,5K Jan 28 14:42 ./egs/dihard_2018/v1/local/make_voxceleb1_v2.pl*
.../kaldi $ ll ./egs/voxceleb/v1/local/make_voxceleb1_v2.pl
-rwxrwxr-x 1 cbj cbj 4,6K Jan 28 14:42 ./egs/voxceleb/v1/local/make_voxceleb1_v2.pl*
```
After comparing the files, I think the files in `dihard_2018` have a bug, because `voxceleb` has one bugfix commit that was not applied to `dihard_2018`.

Also the other `make_voxceleb*` files are copies. Wouldn't it be better to use symlinks?

I am not sure, who is responsible for this, maybe @aarora8, @desh2608 or @sw005320.

desh2608:
Using `egs/voxceleb/v1/local/make_voxceleb1_v2.pl` should work. As for whether symlinks should be used for this instead of copying over the file to local, @david-ryan-snyder is better suited to make that decision.

desh2608:
[...]

PR #4012, opened 2020-03-23
Fixed voxceleb data prep scripts in other egs

desh2608:
Copied updated scripts from `voxceleb/v1/local` to `dihard_2018/v1/local` and `chime6/s5_track2/local/` (raised in https://github.com/kaldi-asr/kaldi/issues/4010)


johnjosephmorgan:
Could you add the line that sources utils/parse_options.sh so that stages work in the voxceleb recipes?

johnjosephmorgan:
I have another request for the voxceleb prep scripts.
The make_voxceleb1_v2.pl downloads the voxceleb1_test_v2.txt file.
Could you change the directory where it is downloaded?
Could you download it to $out_dir instead of $data_base?
The $data_base directory contains the voxceleb corpus.
We download this corpus and store it on a read only file system.

PR #4014, opened 2020-03-24
Correct typo in endpoint comment.

apengwin:
Fix a typo in comment in `online-endpoint.h`. It says there are four endpoint rules, but there actually are five.


danpovey:
Nobody expects the Spanish Inquisition!
Amongst the endpointing rules are...

PR #4018, opened 2020-03-27
Update wer_output_filter

wxt1997:
You should make this change when you use python3.


egs/thchs30/s5/local/wer_output_filter
```diff
@@ -12,7 +12,7 @@ for l in sys.stdin:
     ll=l.split()
     lk=ll[0]
     for v in ll[1:]:
-        v = v.decode('utf-8')
+        v = v.encode('utf-8').decode('utf-8')
```
danpovey: thanks.. can you please change the sheband /usr/bin/env python to say /usr/bin/env python3?  Then we dont have to worry about python2 compatibility.

PR #4020, opened 2020-03-30
fix false removal of 1st field of clean text

naxingyu:
thanks to Xuegong Bian


naxingyu:
@danpovey plz merge this, thx.

PR #4028, opened 2020-04-04
Update Makefile

jinsongzhao:


songmeixu:
The original makefile is alright, isn't it?

PR #4047, opened 2020-04-20
Fix return condition and add resize code in online-audio-source.cc

KimJeongSun:
i used this source for online-audio-client. but when i process last frame of audio. 
that frame is skipped because of this condition. 
i think it's because `pos_ == src_.Dim()` at last frame.
so i changed condition of return. and add resize code for last frame.
it works well in my environment. 
I would appreciate it if you review.

### DIFF
--- a/src/online/online-audio-source.cc
+++ b/src/online/online-audio-source.cc
@@ -177,12 +177,14 @@ bool OnlineVectorSource::Read(Vector<BaseFloat> *data) {
     if (data->Dim() == subsrc.Dim()) {
       data->CopyFromVec(subsrc);
     } else {
+      data->Resize(n_elem);
       for (int32 i = 0; i < subsrc.Dim(); ++i)
         (*data)(i) = subsrc(i);
     }
     pos_ += n_elem;
+    return true;
   }
-  return (pos_ < src_.Dim());
+  return false;
 }
 
 }  // namespace kaldi
--- a/src/online/online-tcp-source.cc
+++ b/src/online/online-tcp-source.cc
@@ -140,6 +140,7 @@ bool OnlineTcpVectorSource::Read(Vector<BaseFloat> *data) {
   int32 n_read = b_read / 2;
 
   short* s_frame = (short*) frame;
+  data->Resize(n_read);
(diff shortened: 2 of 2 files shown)


src/online/online-audio-source.cc
```diff
@@ -177,12 +177,14 @@ bool OnlineVectorSource::Read(Vector<BaseFloat> *data) {
     if (data->Dim() == subsrc.Dim()) {
       data->CopyFromVec(subsrc);
     } else {
+      data->Resize(n_elem);
```
danpovey, line 180: @nshmyrev  @hhadian  can you guys please check this?  I'm suspicious because it looks to me like whoever wrote this code intended that `data` should not be resized.  Incidentally, in this kind of situation you should pass in a pointer to  SubVector, not Vector.  (However, I do think this code has another issue, which is that it fails to zero the unset part of *data.)

nshmyrev: Hi

I have checked the issue. The only place where it is used is [OnlineFeatureInput](https://github.com/kaldi-asr/kaldi/blob/5968b4cc03f9deccfd566962d3bba96bad8ce522/src/online/online-feat-input.h#L325) and data->Resize is not harmful there. Moreover, some implementations like GstOnlineSource [already resize](https://github.com/kaldi-asr/kaldi/blob/5968b4cc03f9deccfd566962d3bba96bad8ce522/src/gst-plugin/gst-audio-source.cc#L110). TcpInputSource [don't resize](https://github.com/kaldi-asr/kaldi/blob/5968b4cc03f9deccfd566962d3bba96bad8ce522/src/online/online-tcp-source.cc#L143).

Logically I'd better resize than add zero.

So I would:

1. Keep this change as is
2. Add a note on the method documentation comment that the data can be resized.
3. Fix TcpInputSource too.

danpovey:
Thanks.  Jeong-Sun, can you please fix TcpInputSource too?

KimJeongSun:
I added same code in `online-tcp-source.cc`.
Please check when you have time. 
Thank you for review.

Issue #4049, opened 2020-04-20
About a small issue in multi_cn - $cd


underdogliu:
Dear all, 

I am trying with `multi_cn` egg and found that `local/aishell_download_and_untar.sh` did not work on my side. After some tweaking I found that after downloading when untar-ing waves, seems like after $cd (line 74 and 91), the script did not return to previous directory, which shall be the case according to what I observed. By adding `cd -` after corresponding process (for example, add it after line 86 and 95 like below) it worked. However, I am not sure if it is actually a bug or it just happened to me. If so, similar issues needs to be fixed for all `*download_and_untar.sh` scripts. @naxingyu 

Regards

```
......
cd $data

if ! tar -xvzf $part.tgz; then
  echo "$0: error un-tarring archive $data/$part.tgz"
  exit 1;
fi
cd -

touch $data/$part/.complete

if [ $part == "data_aishell" ]; then
  cd $data/$part/wav
  for wav in ./*.tar.gz; do
    echo "Extracting wav from $wav"
    tar -zxf $wav && rm $wav
  done
  cd -
fi
.......
```

naxingyu:
The script should work presuming that the $data path is absolute, as it is called in run.sh.

underdogliu:
ok then I got no problem. Thanks for specifying. But still maybe it will be safe to call `cd -` in case anybody would like to modify script outside or happen to specify the source audio directory to be a relative one? Or we can add some comments at `run.sh`. I know using full path is sort of common practice but sometimes there are users who does not have access to like `/mnt/` and they have to download data to their own home space on server.

naxingyu:
Thanks for the solution. I think add comments is enough, since the scripts
in local are bound to respective example, not generally utilized. Would you
mind do this?

alex-ht:
`pushd` and `popd` are good stuff

underdogliu:
@naxingyu OK I think I will do it and make PR.

@alex-ht Just out of curiosity, is there any difference between `pushd/popd` to `cd/cd -`? Because I used to see `pushd`, but for most of the cases it was used in docker execution scripts ad function blocks in bash.

alex-ht:
every script should avoid changing working dir, then it can be easily integrated with others. If necessary, using `pushd/popd` structure instead of `cd` can keep traceability. that's why this bug happened. just my opinion

underdogliu:
[...]

PR #4056, opened 2020-04-28
Modify `multi_cn` downloading scripts

underdogliu:
From issue #4049  

This is a modification based on experiment from other projects where I have to use those open-sourced data. I found under my settings, `cd $dir` would not work well if not adding `cd -` after each of it. I chatted with @naxingyu on this and made this PR. @alex-ht proposed to use `pushd/popd` ops instead but I personally regard such pair as ones more widely used under docker and bash scripts with more functions involved.


underdogliu:
sorry this PR was a mistake by wrongly copying previous work from my side which was not optimal as discussed. Please ignore this PR.

PR #4073, opened 2020-05-17
Revert "[scripts] Fix issue with invoking sed, for mac"

danpovey:
Reverts kaldi-asr/kaldi#4072


danpovey:
I had to revert this as I remembered/realized too late that on Linux you cannot pass command line options with /usr/bin/env foo -opt.  (You can on mac).

Given that on linux sed is only at /bin/sed and on mac it's only in /usr/bin/sed, and we seem to need the -f option, I'm afraid that I see no way to make these scripts compatible, unless they are changed to bash script that invoke inline sed.  

@jtrmal do you have a chance to look at it?

jtrmal:
I'll try today
y.

PR #4074, opened 2020-05-17
Fix stats dump when some utts are missing in bestpath due to errors

nshmyrev:
This is not optimal solution yet, so not really recommended for a merge.

The problem is that when some data is broken, lattice-best-path might return error on some lats and skip them in the output. Then paste which merges align output with depth stats gets unsyncronized and the whole thing fails. 

This patch fixes the problem but it creates the single per-frame alignment dump file. Maybe there should be adidtional filter script instead. I'd write it in Python, not sure about perl.

### DIFF
--- a/egs/wsj/s5/steps/diagnostic/analyze_lats.sh
+++ b/egs/wsj/s5/steps/diagnostic/analyze_lats.sh
@@ -62,24 +62,12 @@ $cmd $dir/log/analyze_alignments.log \
 grep WARNING $dir/log/analyze_alignments.log
 echo "$0: see stats in $dir/log/analyze_alignments.log"
 
-
-# note: below, some things that would be interpreted by the shell have to be
-# escaped since it needs to be passed to $cmd.
-# the 'paste' command will paste together the phone-indexes and the depths
-# so that one line will be like utt-id1 phone1 phone2 phone3 .. utt-id1 depth1 depth2 depth3 ...
-# the following command computes counts of pairs (phone, lattice-depth) and outputs lines
-# containing 3 integers representing:
-#   phone lattice_depth, count[phone,lattice_depth]
-$cmd JOB=1:$num_jobs $dir/log/lattice_best_path.JOB.log \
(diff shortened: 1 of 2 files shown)


danpovey:
If it's not super slow it may be OK.  Make sure that file gets cleaned up though.

egs/wsj/s5/steps/diagnostic/analyze_lats.sh
```diff
@@ -62,32 +62,20 @@ $cmd $dir/log/analyze_alignments.log \
...
-# containing 3 integers representing:
-#   phone lattice_depth, count[phone,lattice_depth]
-$cmd JOB=1:$num_jobs $dir/log/lattice_best_path.JOB.log \
-  ali-to-phones --per-frame=true "$model" "ark:gunzip -c $dir/ali_tmp.JOB.gz|" ark,t:- \| \
-  paste /dev/stdin '<(' gunzip -c $dir/depth_tmp.JOB.gz  ')'  \| \
-  perl -ane '$half=@F/2;for($i=1;$i<$half;$i++){$j=$i+$half;$count{$F[$i]." ".$F[$j]}++;}
-  END{for $k (sort keys %count){print "$k $count{$k}\n"}}' \| \
-  gzip -c '>' $dir/depth_stats_tmp.JOB.gz
+$cmd $dir/log/dump_ali_frame.log \
+  ali-to-phones --per-frame=true "$model" "ark:gunzip -c $dir/ali_tmp.*.gz|" "ark,t:|gzip -c >$dir/ali_frame_tmp.gz"
```
kkm000, line 66: Just wondering about the speed of this. Since ali_tmp.*.gz is already sharded, this and the next commands can be ran as parallel jobs (or possibly even a single parallel job per shard). Would it provide a reasonable speed improvement?

danpovey:
.. I know this merge was questionable, but I figure if there are any problems we can fix them afterward.  Right now the script si not so robust.

PR #4077, opened 2020-05-19
Add optional aishell2 for multi_cn

### LINKED ISSUE #4049: About a small issue in multi_cn - $cd

naxingyu:
And some bug fixes that does not change result. Besides, this PR also closes #4049

### DIFF
--- a/egs/multi_cn/s5/README.md
+++ b/egs/multi_cn/s5/README.md
@@ -5,6 +5,7 @@ This is a Chinese speech recognition recipe that trains on all Chinese corpora o
 * Primewords (99 hours)
 * ST-CMDS (110 hours)
 * THCHS-30 (26 hours)
+* optional AISHELL2 (~1000 hours) if available
 
 This recipe was developed by Xingyu Na (Microsoft Corporation) and Hui Bu (AISHELL Foundation).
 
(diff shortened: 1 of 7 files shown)


naxingyu:
@danpovey ready to merge

PR #4096, opened 2020-06-06
Fix build with latest version of openfst

sandsmark:
What it says on the tin


sandsmark:
OK, openfst is apparently not backwards compatible (as in they deprecate things at the same time they introduce them).

Should I leave this open for when you update the bundled openfst?

kkm000:
Yes, please. I'm looking into making Kaldi compatible with either, your changeset should be helpful!

kkm000:
Xref #4131

sandsmark:
if they hadn't changed `FstPrinter` it would be fairly trivial to add a couple of ifdefs to support both, but because of that change it would end up a complete mess...

kdorichev:
Is it still a requirement to use **OpenFST** **1.6.7** and the latest **1.8.0** is not yet supported?

danpovey:
I think the only way we can make it compatible with the new version of Openfst without forcing an instant upgrade is to add, e.g. in fstext/fstext-utils.h, some kind of wrapper for Openfst's Print that uses an #ifdef of some kind to test the Openfst version.

kkm000:
Superseded by #4927

PR #4097, opened 2020-06-07
[doc] Fix typos in command line help

rouseabout:


jtrmal:
thanks! these are great things to have but no-one wants to do it, so I really appreciate your effort.

PR #4098, opened 2020-06-07
fix ami_train_lms.sh

qiujiali:
- Add the `order` option for all `ngram` commands, so that the script works for higher-order (>3) n-gram LMs
- Fix the hard-coded lambda value for LM interpolation


kkm000:
LGTM, thanks!

PR #4107, opened 2020-06-16
[src] minor change of the next weight cutoff

keli78:


danpovey:
Thanks!
The biglm decoder was always slow and I wonder to what extent this was the reason.
Can you try to run steps/decode_biglm.sh somewhere and see what the speed difference is?
Will merge.

keli78:
Sure. Will do.

PR #4108, opened 2020-06-16
Create LICENSE

joellehanna97:


danpovey:
We can't really do this as the contributors would all need to agree, and this is impractical.  It already has  LICENSE.txt, with Apache 2.0.

PR #4115, opened 2020-06-18
[src] fix timestamp bug in commit fd88b1f in TraceBackBestPath()

dophist:


danpovey:
Thanks!   You think it's OK to merge?

dophist:
Dan, I'm testing it on several large test sets, it may take some time. I'll notice you when this PR is ready to merge.

danpovey:
Thanks!

dophist:
Update:

test set: total 15k utterances from 5 test sets, covering various domains and acoustic conditions

1. timestamp assertion crash in TraceBackBestPath():
  I did capture utterances that can reproducibly crash the decoder runtime due to timestamp assertion failure, although it's rare. And after this PR, no more crashes.

2. code correctness:
  during trace back, say Token B backpoints to Token A, and A may have multiple forward links to Token B. To retrieve the correct forward link, link's graph cost is not enough, I think we should pick the best link with lowest (acoustic + graph) cost.  Code changes are already included in this PR.

3. consistency of lattice-based & traceback-based best path.
  I can still observe 4(out of 15k) utterances, that their traceback best path differs from lattice best path, yielding slightly different recognition text.
By checking the best path cost of these 4 utterances:

    | | traceback best path cost | lattice best path cost|
    |-|-|-|
    | utt1 |  -266.950012 | -266.949829 |
    | utt2 | -160.107300 | -160.107193 |
    | utt3 | -382.689575 |-382.689362 |
    | utt4 | -136.480682 |-136.480576 |

    Considering the precision of BaseFloat type, and the proportion of the in consistency (4/~15k), I think it's more of a problem how lattice and traceback algorithms pick one-best from "approximately equal paths". Not a bug.

dophist:
@danpovey Dan
testing results are reported above, I think this PR is ready to merge.

PR #4116, opened 2020-06-19
Bug fixed: affine transform case in BatchedIvectorExtractorCuda::LDATransform

jty016:
The LDA matrix can legitimately have a extra column, for the bias. So the last column should be optionally handled like the way implemented in cudafeat/online-ivector-feature-cuda.cc


jty016:
@danpovey I made a PR according to this [thread](https://github.com/kaldi-asr/kaldi/commit/5a706bc5a8cfb5a106532bfdcb7e2c5ab83ff75f)

PR #4117, opened 2020-06-19
modify makefile, and solve libtool issue with openfst for gcc >= 5

songmeixu:


src/configure
```diff
@@ -194,7 +194,7 @@ function check_compiler {
              You need g++ >= 4.7, Apple clang >= 5.0 or LLVM clang >= 3.3."
   else
     COMPILER_VER_INFO=$($COMPILER --version 2>/dev/null)
-    if [[ $COMPILER_VER_INFO == *"g++"* ]]; then
+    if [[ $COMPILER_VER_INFO == *[gc]"++"* ]]; then
```
songmeixu, line 197: the name of cpp compiler like that from conda-build: "x86\_64-conda\_cos6-linux-gnu-c++", ends with "c++"

src/cudadecoderbin/Makefile
```diff
@@ -12,6 +12,7 @@ endif
 
 LDFLAGS += $(CUDA_LDFLAGS)
 LDLIBS += $(CUDA_LDLIBS)
+EXTRA_LDLIBS += -lrt
```
songmeixu, line 15: for a build error like: "batched-wav-nnet3-cuda-online.o: undefined reference to symbol 'clock\_gettime@@GLIBC\_2.2.5'"

src/makefiles/cuda_64bit.mk
```diff
@@ -13,5 +13,10 @@ CUDA_FLAGS = --machine 64 -DHAVE_CUDA \
              -std=c++11 -DCUDA_API_PER_THREAD_DEFAULT_STREAM  -lineinfo \
              --verbose -Xcompiler "$(CXXFLAGS)"
 
+ifeq ($(shell test -e $(CUDATKDIR)/lib64/libcudart_static.a && echo -n yes),yes)
```
songmeixu, line 16: As the conda env install libs, bins, headers of all packages into its root path, the dirname for libs is "lib".

I think we need conda env to prepare a proper toolchain (CXX, MKL, CUDA, etc.) As conda can provide a standalone virtual env like a container of docker.

src/makefiles/cuda_64bit.mk
```diff
@@ -13,5 +13,10 @@ CUDA_FLAGS = --machine 64 -DHAVE_CUDA \
...
 CUDA_LDFLAGS += -L$(CUDATKDIR)/lib64 -Wl,-rpath,$(CUDATKDIR)/lib64
+else
+CUDA_LDFLAGS += -L$(CUDATKDIR)/lib -Wl,-rpath,$(CUDATKDIR)/lib
+endif
+
```
songmeixu, line 21: As the conda env install libs, bins, headers of all packages into its root path, the dirname for libs is "lib".

I think we need conda env to prepare a proper toolchain (CXX, MKL, CUDA, etc.) As conda can provide a standalone virtual env like a container of docker.

src/makefiles/default_rules.mk
```diff
@@ -82,7 +82,7 @@ endif
 	$(MAKE) -C ${@D} ${@F}
 
 clean:
-	-rm -f *.o *.a *.so $(TESTFILES) $(BINFILES) $(TESTOUTPUTS) tmp* *.tmp *.testlog
+	-rm -f *.o *.a *.so *.dylib $(OBJFILES) $(TESTFILES) $(BINFILES) $(TESTOUTPUTS) tmp* *.tmp *.testlog
```
songmeixu, line 85: add more "should-be-there" stuff to delete

tools/Makefile
```diff
@@ -86,6 +86,12 @@ else
 endif
 
 openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION) | check_required_programs
+	# modify src/extensions/far/Makefile.am to make it work with gcc >= 5, while keep its timestamp to avoid autoreconf
+	$(eval make_file=openfst-$(OPENFST_VERSION)/src/extensions/far/Makefile.in)
+	$(eval STAMP="$(shell date -r $(make_file))")
+	sed -i -r 's~am__DEPENDENCIES_1 =$$~& libfstfar.la~' $(make_file)
+	sed -i -E 's~(@LDADD = libfstfarscript.la) \.\.~\1 libfstfar.la ..~' $(make_file)
+	touch -d $(STAMP) $(make_file)
```
songmeixu, line 94: Modify the makefile of one openfst extension: far, to make it works with gcc >= 5.

This issue is related to the (libtool Inter-library dependencies)\[https://www.gnu.org/software/libtool/manual/libtool.html#Inter_002dlibrary-dependencies\] problem. For shared libraries that depend on another shared lib, this problem would appear.

src/makefiles/default_rules.mk
```diff
@@ -50,7 +50,7 @@ $(LIBFILE): $(LIBNAME).a
 	ln -sf $(shell pwd)/$@ $(KALDILIBDIR)/$@
   else ifeq ($(shell uname), Linux)
         # Building shared library from static (static was compiled with -fPIC)
-	$(CXX) -shared -o $@ -Wl,--no-undefined -Wl,--as-needed  -Wl,-soname=$@,--whole-archive $(LIBNAME).a -Wl,--no-whole-archive $(LDFLAGS) $(LDLIBS)
+	$(CXX) -shared -o $@ -Wl,--as-needed  -Wl,-soname=$@,--whole-archive $(LIBNAME).a -Wl,--no-whole-archive $(LDFLAGS) $(LDLIBS)
```
songmeixu, line 53: "--no-undefined: It reports the unresolved symbols that are not resolved immediately, at linking stage."

---

is not suggested to commonly use as here. We can control all needed libs are listed in link cmd. However, we cannot require third-party packages. As the openfst depends on automake, and libtool has an issue with a shared lib that links to another shared lib. So any target link with this openfst shared library would have build error like: "fst.h:746: undefined reference to `fst::FstHeader::Write"

songmeixu: The no-undefined flag also is not help for optimized the lib sizes, that is the intension when it was added in one pull.

tools/extras/check_dependencies.sh
```diff
@@ -136,7 +138,7 @@ case $(uname -m) in
     # installed in an alternative location (this is unlikely).
     MKL_ROOT="${MKL_ROOT:-/opt/intel/mkl}"
     if [ ! -f "${MKL_ROOT}/include/mkl.h" ] &&
-         ! echo '#include <mkl.h>' | $CXX -I /opt/intel/mkl/include -E - >&/dev/null; then
+         ! echo '#include <mkl.h>' | $CXX -I /opt/intel/mkl/include -E - &>/dev/null; then
```
songmeixu, line 141: here are two formats for redirecting standard output and standard error:

          &>word

   and

          >&word

   Of the two forms, the first is preferred.  This is semantically equivalent to

          >word 2>&1

songmeixu: to make shellcheck happy

tools/Makefile
rolzy: Hello!

I came across a bug at this code block. This section does not work if the locale of the machine is set to another language. The date of the file at line 91 will be outputted in another language, which will not work in the successive touch command. 

```
touch -d "2018Õ╣┤  3µ£ê  3µùÑ Õ£ƒµø£µùÑ 00:08:40 UTC" openfst-1.6.7/src/extensions/far/Makefile.in
touch: invalid date format ÔÇÿ2018Õ╣┤  3µ£ê  3µùÑ Õ£ƒµø£µùÑ 00:08:40 UTCÔÇÖ
Makefile:89: recipe for target 'openfst-1.6.7/Makefile' failed
```

I suggest to set the locale to English at line 91 so the date is outputted correctly:
`$(eval STAMP="$(LANG=en_us_88591; shell date -r $(make_file))")`

songmeixu: Yeah, it's a bug. Thank you. Also `LANG=zh_CN.utf8 date` doesn't work for `3.18.6-2.el7.centos.x86_64` in my test. Anyway, it's OK to change to `.. LANG=C date ..`. The change has been done 812596dfe365658d4ba852cd999a4fb032aea3f7.

PR #4122, opened 2020-06-22
rm extra ')' in egs/iam/v2/local/gen_topo.py

songmeixu:


egs/iam/v2/local/gen_topo.py
```diff
@@ -68,23 +90,27 @@
...
 print("</ForPhones>")
-if(args.num_sil_states > 1):
+if args.num_sil_states > 1:
     transp = 1.0 / (args.num_sil_states - 1)
-    
+
     state_str = "<State> 0 <PdfClass> 0 "
     for x in range(0, (args.num_sil_states - 1)):
-        state_str = "{}<Transition> {} {} ".format(state_str, x, transp))
+        state_str = "{}<Transition> {} {} ".format(state_str, x, transp)
```
songmeixu, line 98: Here is the extra `)`. Other changes just follow the [black](https://github.com/psf/black) format style.

egs/iam/v2/local/gen_topo.py
```diff
@@ -68,23 +90,27 @@
...
     state_str = state_str + "</State>"
     print(state_str)
 
     for x in range(1, (args.num_sil_states - 1)):
-        state_str = "<State> {0} <PdfClass {0} ".format(x))
+        state_str = "<State> {0} <PdfClass {0} ".format(x)
```
songmeixu, line 103: Here is one extra `)`. Other changes just follow the [black](https://github.com/psf/black) format style.

egs/iam/v2/local/gen_topo.py
```diff
@@ -68,23 +90,27 @@
...
         for y in range(1, args.num_sil_states):
-            state_str = "{}<Transition> {} {} ".format(state_str, y, transp))
+            state_str = "{}<Transition> {} {} ".format(state_str, y, transp)
```
songmeixu, line 105: Here is one extra `)`. Other changes just follow the [black](https://github.com/psf/black) format style.

kkm000:
@songmeixu, thanks for the fix! Could you please split off the part that fixes syntax errors as a separate PR? I am not aware that we agreed to use a particular Python coding style--I do not think we ever had this discussion. If it comes to this, the one adopted by Google (with 2 space indent vs 4) is a much better approach, as it aligns with our C++ and Bash indents, which makes reading of the overall codebase easier. Speaking of this coding style, to me personally splitting the `f(x)` into 4 lines makes the code less readable. Both out C++ and Bash code is pretty densely written, so a page in the text editor contains a lot of it.

Also, the comments in GitHub are for code reviews. Again, a golden rule is if a change cannot be read by the reviewer without additional commentary, it does not pass the readability test. All the comments here are essentially lost when the code is committed. If something needs explaining, the explanation should be in code comment, not put into comment intended for revewer's comments. Another downside of (ab)using the review comments is that GitHub reports the number of comments if any exists, so when I have time to review code, I would gloss over the list of PRs and think "oh, there is already a discussion here, I'll better pick one not yet reviewed."

The purpose of the overall change should be put into the overall PR description. Again, it's a good self-check: if you're finding that you are putting too much loosely related stuff into the PR description, or it is not clear what code changes multiple items are referring to, it's a sure sign the change should be split into multiple PRs.

In this case, a change that fixes syntax error in a separate PR would ask for no more than "Fix syntax errors in xyz.py", and self-evident from the change, easy to review and non-contentious, "thanks" and merged.

A PR reformatting the code would be "Reformat code according to the XYZ standard"--I'm not sure if it would be agreed upon and accepted, but at the least it's purpose is clear from a single sentence. I believe changes like the latter better be discussed ahead of proposing a PR,

songmeixu:
@kkm000 Very sorry for committing more than the title shows in this PR. I was not going to commit the *.py style changes. I found a bug while formatting them, thus I PR it. Days later, I pushed the format changes to my branch, and I just forgot it's in PR. I certainly wouldn't commit that many changes without community discussions. Sorry about this again, I would keep caution next time. And I am only able to make a new PR #4136 only with fixing the bug. 
 
---

About the py style, you are right and remind me how to suggest it, thank you very much!!

kkm000:
Nothing to be sorry about, you're good! We developed some productivity practices over years, but they're not formalized or documented, so I'm just explaining the rationale behind them.

PR #4127, opened 2020-06-23
upgrade trrnnlm installer

kunazure:
upgrade trrnnlm installer to tensorflow 2.0.0


danpovey:
Thanks!  Looks harmless so I will merge.  This part doesn't get used very often.

PR #4129, opened 2020-06-23
[egs] Add --am-dir option to make_weighted_den_fst.sh

rslish:
Using the --am-dir option allows you to build the denominator using a tree and transition model other than the ones used to create the alignments.


kkm000:
LGTM, thanks!

PR #4130, opened 2020-06-24
Update validate_data_dir.sh

alex-ht:
- line 127: skip text validation when `no_text` is true.
- line 130: `grep '[^[:print:][:space:]]'` says Chinese characters are unprintable.

### DIFF
--- a/egs/wsj/s5/utils/validate_data_dir.sh
+++ b/egs/wsj/s5/utils/validate_data_dir.sh
@@ -124,10 +124,10 @@ if [ ! -f $data/text ] && ! $no_text; then
 fi
 
 num_utts=`cat $tmpdir/utts | wc -l`
-if [ -f $data/text ]; then
+if ! $no_text; then
   if ! $non_print; then
     # compatible with bin/align-text
-    n_non_print=$(grep -c '[^[:print:][:space:]]' $data/text) && \
+    n_non_print=$(LC_ALL="C.UTF-8" grep -c '[^[:print:][:space:]]' $data/text) && \
     echo "$0: text contains $n_non_print lines with non-printable characters" &&\
     exit 1;
   fi


kkm000:
Thanks @alex-ht! I just nearly removed the whole check. This way it's better!

egs/wsj/s5/utils/validate_data_dir.sh
```diff
@@ -124,10 +124,10 @@ if [ ! -f $data/text ] && ! $no_text; then
 fi
 
 num_utts=`cat $tmpdir/utts | wc -l`
-if [ -f $data/text ]; then
+if ! $no_text; then
   if ! $non_print; then
     # compatible with bin/align-text
-    n_non_print=$(grep -c '[^[:print:][:space:]]' $data/text) && \
+    n_non_print=$(LC_ALL="C.UTF-8" grep -c '[^[:print:][:space:]]' $data/text) && \
```
xiaohui-zhang, line 130: this isn't working for turkish. We should use en_us.UTF-8 here. Also I suggest we make non_print as true here, so that ppl only turn this on when necessary. Otherwise failing the program even if there's one offending line is too strict.

xiaohui-zhang: cc @vimalmanohar

danpovey: This has already been fixed, see #4130

xiaohui-zhang: hmm no.. Here we are using C.UTF-8, while it's still failing for turkish characters.. en_us.UTF-8 works

jtrmal:
#4193 should address/fix the stuff, it would be great if you could
test/review
y.

egs/wsj/s5/utils/validate_data_dir.sh
alex-ht: Can you provide a test case?

kkm000: The correct setting here is `LANG=C.UTF-8`. grep does not care what is the first day of the week in the locale (if the C locale has them at all). either `LANG=` or `LC_CTYPE=` is enough.

Issue #4124, opened 2020-06-22
linux extension to sed in tools Makefile


johnjosephmorgan:
I'm trying to run make in the tools directory on MacOSX. The run is failing because sed is using a linux extension, namely the -i option.

kkm000:
Huh, I thought we got rid of them all. Can you pinpoint where, what is being built? tools/Makefile downloads a few programs, so it may not be as easy.

kkm000:
Introduced in 57bc88acf #4117

johnjosephmorgan:
tools/Makefile lines 92 and 93

johnjosephmorgan:
misc/fix_cpplint.sh and misc/fix_bash.sh

kkm000:
Sure, but the latter two are maintainer scripts, and not important. The Makefile one is a recent change, and the one that needs to be fixed. Gimme a few hours, I'll get to it.

johnjosephmorgan:
The -d option is also a linux extension that is not in bsd.

songmeixu:
Hi, the patch was introduced by me. And I have made a PR in #4135. Instead of reverting it, it will be much appreciated if someone could help review it. 🤝

songmeixu:
Sorry, @kkm000. I just read the #4133. Now, let me explain why I committed this patch there.

PR #4133, opened 2020-06-25
[build] Undo OpenFST Makefile patch hack from #4177

kkm000:
Please treat the fix as provisional at this time.

I could not reproduce the problem with the configuration as close as was reported in https://groups.google.com/d/msg/kaldi-help/5smFF2RW9do/P3Ml5iITAgAJ, with the exception of the compiler packaging version: I used g++ 7.3.1 (Red Hat 7.3.1-6) while the problem was reported against the package version 7.3.1-2.

I also could not find similar reports anywhere else. Patching the automake-generated Makefile is a cure worse than the disease, and I certainly disfavor this solution. It will hit us hard when we upgrade OpenFST. And, since the variable `OPENFST_VERSION` can be set in make's command line, this solution is certainly asking for a trouble.

@songmeixu, were you able to actually reproduce the problem? I spent a couple of hours trying everything I could to recreate the exact same build environment, but to no avail. If you have an exact repro, please share, and maybe I'll figure out a prettier fix. If not, let's just roll back this change.

Close: #4124

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -86,12 +86,6 @@ else
 endif
 
 openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION) | check_required_programs
-	# modify src/extensions/far/Makefile.am to make it work with gcc >= 5, while keep its timestamp to avoid autoreconf
-	$(eval make_file=openfst-$(OPENFST_VERSION)/src/extensions/far/Makefile.in)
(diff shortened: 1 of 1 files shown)


kkm000:
@songmeixu, just reminding, you were going to point to a repro which this change was fixing. I know you're busy, but configure is broken on the Macs at the moment.

PR #4134, opened 2020-06-25
[build] Fix RedHat package names

kkm000:
python2.7 package is named python27, and gfortran gcc-gfortran.


danpovey:
Great.. you can check these kinds of thing in yourself if you don't think
it's likely to be dangerous.

kkm000:
@danpovey, sure, but I always want a second pair of eyes on my changes. Just an "LGTM" is enough.

I found these mismatches while bootstrapping Fedora 27 in Docker -- docker base images come pretty minimal. Of course I ran them again in a fresh container.

PR #4135, opened 2020-06-25
change to portable usage of `sed, touch` for macos/BSD and GNU.

songmeixu:


kkm000:
Looks our letters crossed in the mail :)

Could you please read my question in #4133? I understand the solution, but I am not certain we have a problem that it solves.

PR #4146, opened 2020-06-30
[src] CudaDecoder endpointing

hugovbraun:
Built on top of #4101. Probably better to look at the diff once #4101 has been merged.

Implements endpointing directly into the cuda decoder. 

Uses the rules as defined in online2/online-endpoint.h. From a user point of view, setting the high level parameters on endpointing and passing a vector in the DecodeBatch should be enough:

```
  void DecodeBatch(const std::vector<CorrelationID> &corr_ids,
                   const std::vector<SubVector<BaseFloat>> &wave_samples,
                   const std::vector<bool> &is_first_chunk,
                   const std::vector<bool> &is_last_chunk,
                   std::vector<std::string *> *partial_hypotheses = NULL,
                   std::vector<bool> *end_points = NULL)
```
Parameters (from online2/online-endpoint.h):

```
--endpoint.rule1.max-relative-cost : This endpointing rule requires relative-cost of final-states to be <= this value (describes how good the probability of final-states is). (float, default = inf)
--endpoint.rule1.min-trailing-silence : This endpointing rule requires duration of trailing silence(in seconds) to be >= this value. (float, default = 5)
--endpoint.rule1.min-utterance-length : This endpointing rule requires utterance-length (in seconds) to be >= this value. (float, default = 0)
--endpoint.rule1.must-contain-nonsilence : If true, for this endpointing rule to apply there mustbe nonsilence in the best-path traceback. (bool, default = false)
```

Easiest way to test is to pass `--print-endpoints=true` to the binary `src/cudadecoderbin/batched-wav-nnet3-cuda-online`

Internally, it provides all the necessary metrics (relative cost, number of silence phones on current best path, total length). Those rules can be modified through the command line parameters.

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
@@ -85,6 +85,11 @@ void BatchedThreadedNnet3CudaOnlinePipeline::AllocateAndInitializeData(
     cuda_decoder_->SetThreadPoolAndStartCPUWorkers(
         thread_pool_.get(), config_.num_decoder_copy_threads);
   }
+
+  cuda_decoder_->SetOutputFrameShiftInSeconds(
+      feature_info_->FrameShiftInSeconds() *
+      config_.compute_opts.frame_subsampling_factor);
+
   n_samples_valid_.resize(max_batch_size_);
   n_input_frames_valid_.resize(max_batch_size_);
   n_lattice_callbacks_not_done_.store(0);
@@ -218,7 +223,8 @@ void BatchedThreadedNnet3CudaOnlinePipeline::DecodeBatch(
     const std::vector<SubVector<BaseFloat>> &wave_samples,
     const std::vector<bool> &is_first_chunk,
     const std::vector<bool> &is_last_chunk,
-    std::vector<const std::string *> *partial_hypotheses) {
+    std::vector<const std::string *> *partial_hypotheses,
+    std::vector<bool> *end_points) {
   nvtxRangePushA("DecodeBatch");
   KALDI_ASSERT(corr_ids.size() > 0);
   KALDI_ASSERT(corr_ids.size() == wave_samples.size());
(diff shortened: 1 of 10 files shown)


kkm000:
Thanks! So, you'd say that both #4101 and this one are ready to merge?

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -250,6 +256,8 @@ void BatchedThreadedNnet3CudaOnlinePipeline::DecodeBatch(
         "You need to set --word-symbol-table to use partial hypotheses");
     cuda_decoder_->AllowPartialHypotheses();
   }
+  if (end_points) cuda_decoder_->AllowEndPointing();
```
kkm000, line 259: Naming of functions varies between `Endpoint` and `EndPoint`. We generally use "endpoint" as a single word, so the former would be better.
```suggestion
  if (end_points) cuda_decoder_->AllowEndpointing();
```

src/cudadecoder/cuda-decoder.h
```diff
@@ -236,14 +241,30 @@ class CudaDecoder {
...
 
-  void AllowPartialHypotheses() { generate_partial_hypotheses_ = true; }
+  void AllowPartialHypotheses() {
+    partial_traceback_ = generate_partial_hypotheses_ = true;
+  }
+
+  void AllowEndPointing() {
+    KALDI_ASSERT(
+        "You must call SetOutputFrameShiftInSeconds() to use endpointing" &&
+        frame_shift_seconds_ != FLT_MAX);
```
kkm000, line 251:
```suggestion
    if (frame_shift_seconds_ == FLT_MAX) {
      KALDI_ERR << "You must call SetOutputFrameShiftInSeconds() "
                << "to use endpointing";
    }
```

src/cudadecoder/cuda-decoder.h
```diff
@@ -463,10 +486,16 @@ class CudaDecoder {
   CudaDecoderConfig config_;
   fst::SymbolTable *word_syms_;       // for partial hypotheses
   bool generate_partial_hypotheses_;  // set by AllowPartialHypotheses
+  bool endpointing_;
+  bool partial_traceback_;
+  BaseFloat frame_shift_seconds_;
+
+  std::set<int> silence_phones_;
```
kkm000, line 493:
```suggestion
  std::set<int32> silence_phones_;
```

hugovbraun:
@kkm000 The known issue was fixed and we don't have any known bugs now

kkm000:
Thanks much! Merging.

auzxb:
A bug happened in this commit

`cuda-decoder.h:851:3: error: ‘atomic_int32_t’ in namespace ‘std’ does not name a type
   std::atomic_int32_t n_partial_traceback_threads_todo_;`

kkm000:
@auzxb, this is fixed in #4173 today. Please open a new issue when reporting a big, we won't notice otherwise.

PR #4161, opened 2020-07-10
fix typo in mkgraph.sh

glynpu:


egs/wsj/s5/utils/mkgraph.sh
```diff
@@ -54,7 +54,7 @@ mkdir -p $dir
 # (note: the [[ ]] brackets make the || type operators work (inside [ ], we
 # would have to use -o instead),  -f means file exists, and -ot means older than).
 
-required="$lang/L.fst $lang/G.fst $lang/phones.txt $lang/words.txt $lang/phones/silence.csl $lang/phones/disambig.int $model $tree"
+required="$lang/L_disambig.fst $lang/G.fst $lang/phones.txt $lang/words.txt $lang/phones/silence.csl $lang/phones/disambig.int $model $tree"
```
danpovey, line 57: Dunno why you changed L.fst to L_disambig.fst.  I'm concerned this may break somethign that I can't anticipate right now.

glynpu: Because L.fst is not used in this script.
L_disambig.fst  is composed with G.fst to generate LG.fst. 
So I think it's more appropriate to check the existance of L_disambing.fst.

PR #4164, opened 2020-07-13
TS-VAD based diarization in CHiME-6

desh2608:
The implementation is from @ipmedenn and it has also been tested by @aarora8 and me.


src/featbin/paste-vectors.cc
```diff
@@ -0,0 +1,138 @@
...
+
+}
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    using namespace std;
+
+    const char *usage =
+        "Paste vector files \n"
```
danpovey, line 60: This is not specific: maybe 'Concatenate vectors' might be clearer?

kkm000: @danpovey, we have `concat-*` vs. `paste-*` semantics, and this one looks like it's `paste-*`-type? I.e., for
vec1
```
utt-1 [ 1 2 ]
utt-2 [ 3 4 5 ]
```
and vec2
```
utt-1 [ 11 12 ]
utt-2 [ 13 14 15 ]
```
paste (vec1 vec2) is
```
utt-1 [ 1 2 11 12 ]
utt-2 [ 3 4 5 13 14 15 ]
```
Did I grok what the tool was doing?

danpovey: From a  UNIX-command-line-manipulating text point of view you're right, but (a) these aren't text files necessarily, (b) someone may not make that mental connection.  These are vectors, "concatenate" is something that makes sense from a vector standpoint.

desh2608:
@kkm000 thanks for the detailed comments! @ipmedenn could you make these changes when you have some time?

ipmedenn:
> @kkm000 thanks for the detailed comments! @ipmedenn could you make these changes when you have some time?

@desh2608 yes I will. Should I make it in kaldi_chime6 repo or in https://github.com/desh2608/kaldi/tree/ts_vad_final?

src/featbin/multiply-vectors.cc
```diff
@@ -0,0 +1,173 @@
...
+        != kNoRspecifier) {
+      // We're operating on tables, e.g. archives.
+
+      // Last argument is output
+      string wspecifier = po.GetArg(po.NumArgs());
+      BaseFloatVectorWriter vector_writer(wspecifier);
+
+      // First input is sequential
+      string rspecifier1 = po.GetArg(1);
+      SequentialBaseFloatVectorReader input1(rspecifier1);
```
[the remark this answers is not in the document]
ipmedenn: @kkm000 These names have been copypasted from paste-feats.cc, so is it necessary to rename?

src/featbin/paste-vectors.cc
ipmedenn: @danpovey should I rename paste-vectors to concat-vectors, or you meant only usage string?

desh2608:
> > @kkm000 thanks for the detailed comments! @ipmedenn could you make these changes when you have some time?
> 
> @desh2608 yes I will. Should I make it in kaldi_chime6 repo or in https://github.com/desh2608/kaldi/tree/ts_vad_final?

@ipmedenn You can push it to this branch (ts_vad_final) so that this PR will get updated automatically.

src/featbin/multiply-vectors.cc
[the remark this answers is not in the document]
kkm000: Yes, please.

src/featbin/paste-vectors.cc
kkm000: @danpovey, I'm kinda torn on this, the naming would be inconsistent either. We already have binaries with a similar semantics called `paste-*`, and dissimilar called `concat-*`. Maybe `combine-vectors` would work?

Yes, of course I understand these are not usually text files.

danpovey:
I think paste- is fine for the name, for consistency; just change the error
message to clarify what it actually does.

ipmedenn:
> 
> 
> > > @kkm000 thanks for the detailed comments! @ipmedenn could you make these changes when you have some time?
> > 
> > 
> > @desh2608 yes I will. Should I make it in kaldi_chime6 repo or in https://github.com/desh2608/kaldi/tree/ts_vad_final?
> 
> @ipmedenn You can push it to this branch (ts_vad_final) so that this PR will get updated automatically.

@desh2608 it seems that I have no permission to commit in your repo

desh2608:
I believe all the comments have now been addressed by @ipmedenn.

ipmedenn:
@kkm000 check the updated versions, please.

kkm000:
@ipmedenn, Thanks, I'll do my best to review today.

src/featbin/paste-vectors.cc
```diff
@@ -0,0 +1,129 @@
...
+
+}  // namespace kaldi
+
+int main(int argc, char *argv[]) {
+  try {
+    using namespace kaldi;
+    using namespace std;
+
+    const char *usage =
+        "Concatenate vector files \n"
```
kkm000, line 54: Let's put it this way (feel free to break into lines 60-100 characters on output each paragraph, so they are nice and readable):

>Append to vectors the first archive the vectors under the same key from the second and following archives, and output the result.
>
>NOTE: Keys absent from the first archive will not be present in output, even if they occur in other input archives."

The space before '\n' is not needed.

src/featbin/paste-vectors.cc
```diff
@@ -0,0 +1,129 @@
...
+    po.Read(argc, argv);
+
+    if (po.NumArgs() < 3) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    if (ClassifyRspecifier(po.GetArg(1), NULL, NULL)
+        != kNoRspecifier) {
+      // We're operating on tables, e.g. archives.
```
kkm000, line 69: This program can only work on archives. This `if` closes, and the program exists silently if this condition is false.

Reverse the condition and print an error, e.g. 
```suggestion
    if (ClassifyRspecifier(po.GetArg(1), NULL, NULL) != kNoRspecifier) {
      KALDI_ERR << "This program can operate only on tables, not on rxfiles";
    }
```
and then unindent the rest. `KALDI_ERR` will exits the program.

ipmedenn:
@kkm000 Thank you! I've made these changes. @desh2608 could you merge my PR to your repo?

kkm000:
LGTM, thanks guys, merging this!

PR #4170, opened 2020-07-15
Set cmd to run.pl instead of queue.pl.

### LINKED ISSUE #4150: set default cmd to run.pl instead of queue.pl
The cmd variable in the script steps/segmentation/detect_speech_activity.sh is set to queue.pl. Could we set the default to run.pl instead?

johnjosephmorgan:
Sets cmd to default to run.pl instead of queue.pl.

Closes #4150


kkm000:
Thanks!

PR #4171, opened 2020-07-15
Removed python command to enable python3 from shebang.

johnjosephmorgan:
Removed python command to enable python3 from shebang.


kkm000:
LGTM, thanks.

PR #4172, opened 2020-07-16
add use-gpu option

viannecc:
Support "use-gpu" option when computes rnnlm perplexity.

### DIFF
--- a/scripts/rnnlm/compute_perplexity.sh
+++ b/scripts/rnnlm/compute_perplexity.sh
@@ -8,9 +8,9 @@ use_gpu=no
 . utils/parse_options.sh
 
 if [ $# != 2 ]; then
-  echo "Usage: $0 <rnn-dir> <input-text>"
+  echo "Usage: $0 [options] <rnn-dir> <input-text>"
   echo "Options: "
-  echo "  --use_gpu (yes|no|optional|wait)  # whether to use gpu."
+  echo "  --use_gpu (yes|no|optional|wait)  # whether to use gpu [no]."
   exit 1
 fi
 
@@ -51,4 +51,3 @@ ppl=$(rnnlm-sentence-probs ${opt} \
        awk '{for(i=2;i<=NF;i++) a+=$i; b+=NF-1}END{print exp(-a / b)}')
 
 echo "$0: perplexity is $ppl"
-
--- a/scripts/rnnlm/compute_perplexity.sh
+++ b/scripts/rnnlm/compute_perplexity.sh
@@ -3,10 +3,14 @@
 # This script computes perplexity of text on the specified RNNLM model. 
 
 [ -f ./path.sh ] && . ./path.sh
+
+use_gpu=no
 . utils/parse_options.sh
 
 if [ $# != 2 ]; then
   echo "Usage: $0 <rnn-dir> <input-text>"
+  echo "Options: "
+  echo "  --use_gpu (yes|no|optional|wait)  # whether to use gpu."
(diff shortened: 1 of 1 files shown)


scripts/rnnlm/compute_perplexity.sh
```diff
@@ -39,7 +43,9 @@ done
 
 special_symbol_opts=$(cat $dir/special_symbol_opts.txt)
 
-ppl=$(rnnlm-sentence-probs --normalize-probs=true \
+opt="--normalize-probs=true --use-gpu=${use_gpu}"
+
+ppl=$(rnnlm-sentence-probs ${opt} \
        $special_symbol_opts $dir/final.raw "$word_embedding" \
        <(cat $text_in | sym2int.pl $dir/config/words.txt | awk '{print "utt_id ", $0}') | \
        awk '{for(i=2;i<=NF;i++) a+=$i; b+=NF-1}END{print exp(-a / b)}')
```
kkm000, line 51: @viannecc, thanks, being able to pass the gpu option to the binary certainly won't hurt!

While reviewing the changes, I've spotted something that looked weird to me, a couple lines below. First `awk` adds a new field $1 to every line which is trivially 'utt_id', and the next `awk` in the pipeline sums up and counts fields _except_ the first, just added. I'd change this as below, but I want a second pair of eyes on the change. Can you think of anything wrong with it? I do not really know what this script really computes.
```suggestion
       <(cat $text_in | sym2int.pl $dir/config/words.txt |
         awk '{for(i=1;i<=NF;i++) a+=$i; b+=NF} END{print exp(-a / b)}' )
```
The `|` implies line continuation, `\` is not required, but that's purely cosmetic. It was the use of `awk` that baffled me.

scripts/rnnlm/compute_perplexity.sh
```diff
@@ -3,10 +3,14 @@
...
 
 [ -f ./path.sh ] && . ./path.sh
+
+use_gpu=no
 . utils/parse_options.sh
 
 if [ $# != 2 ]; then
   echo "Usage: $0 <rnn-dir> <input-text>"
+  echo "Options: "
+  echo "  --use_gpu (yes|no|optional|wait)  # whether to use gpu."
```
kkm000, line 13:
```suggestion
  echo "Usage: $0 [options] <rnn-dir> <input-text>"
  echo "Options: "
  echo "  --use_gpu (yes|no|optional|wait)  # whether to use gpu [no]."
```

scripts/rnnlm/compute_perplexity.sh
viannecc: Thanks for your fast review.

viannecc: Hi. Thanks for your fast review！
I just checked the source code of rnnlm-sentence-probs.cc, which demands each line of the text to be evaluated should be started with a key, here is 'utt_id'. If without the key, the results would be wrong.  So, 'utt_id' is needed and 'i' should begin at 2. 
The 'NF' item denotes log prob of '</s>' symbol. So ’NF-1‘ is right I think.
'|' is pipeline symbol, it shows the results of 'rnnlm-sentence-probs' would be inputs of the last awk command. 
'\' implies line continuation.

kkm000: Thanks for looking! I am not sure what is really going on, but as long as it works, let's not fix it! :)

kkm000:
LGTM, thanks!

PR #4183, opened 2020-07-19
[src] Correctrly treat subsampling factor != 1 in nnet3-get-egs

### LINKED ISSUE #4167: ./src/nnet3bin/nnet3-get-egs.cc with frame_subsampling_factor != 1
[From kaldi-help](https://groups.google.com/d/msgid/kaldi-help/1a2ec683-154a-4c82-95c6-39511c66fa8co%40googlegroups.com?utm_medium=email&utm_source=footer)

I am reading the code of nnet3-get-egs and I don't understand the following code.
```c++
    // Note: chunk.first_frame and chunk.num_frames will both be
    // multiples of frame_subsampling_factor.
    int32 start_frame_subsampled = chunk.first_frame / frame_subsampling_factor,
        num_frames_subsampled = chunk.num_frames / frame_subsampling_factor;

    Posterior labels(num_frames_subsampled);

    // TODO: it may be that using these weights is not actually helpful (with
    // chain training, it was not), and that setting them all to 1 is better.
    // We could add a boolean option to this program to control that; but I
    // don't want to add such an option if experiments show that it is not
    // helpful.
    for (int32 i = 0; i < num_frames_subsampled; i++) {
      int32 t = i + start_frame_subsampled;
      if (t < pdf_post.size())
        labels[i] = pdf_post[t];  
      for (std::vector<std::pair<int32, BaseFloat> >::iterator
               iter = labels[i].begin(); iter != labels[i].end(); ++iter)
        iter->second *= chunk.output_weights[i];
    }
    eg.io.push_back(NnetIo("output", num_pdfs, 0, labels, frame_subsampling_factor));
```

if chunk.first_frame = 0, chunk.num_frames = 15,  frame_subsampling_factor = 3, then start_frame_subsampled  = 0, num_frames_subsampled  = 5, and t will be 0, 1, 2, 3, 4, but the index will be 0 3 6 9 12. I don't know why the t not be  0 3 6 9 12. 

**Dan's comment**:

Instead of
```c++
       int32 t = i + start_frame_subsampled;
```
it should probably be
```c++
       int32 t = (i + start_frame_subsampled) * frame_subsampling_factor;
```
and instead of 
```c++
  iter->second *= chunk.output_weights[i];
```
should probably be
```c++
  iter->second *= chunk.output_weights[i * frame_subsampling_factor];
```

That code was, I think, never tested with frame_subsampling_factor != 1.

**kkm to himself**: check other get-egs binaries for the same issue.

kkm000:
Fixes #4167

----

Dan, I do not like how easy and error-prone it currently is to confuse subsampled and continuous indices. I'm thinking of a helper struct constructed like `Resample resample(factor,offset)` with methods `int32 resample.Down(int32)` and `int32 resample.Up(int32)`. I looked at the code if it were possible to have different, non-decayable separate types for continuous and downsampled indices, so that they cannot be assigned one to another, but it's probably not possible, as at some points we just simply do not know. But this helper I think I'll implement.

I read other binaries' code to check if they have the same problem, and it's not  very readable, and hard to trace. At the very least, an explicit object doing the trivial arithmetic should make the code a bit easier to grok.

### DIFF
--- a/src/nnet3bin/nnet3-get-egs.cc
+++ b/src/nnet3bin/nnet3-get-egs.cc
@@ -105,21 +105,21 @@ static bool ProcessFile(const GeneralMatrix &feats,
 
     Posterior labels(num_frames_subsampled);
 
-    // TODO: it may be that using these weights is not actually helpful (with
-    // chain training, it was not), and that setting them all to 1 is better.
-    // We could add a boolean option to this program to control that; but I
-    // don't want to add such an option if experiments show that it is not
-    // helpful.
-    for (int32 i = 0; i < num_frames_subsampled; i++) {
-      int32 t = i + start_frame_subsampled;
+    // TODO(danpovey): it may be that using these weights is not actually
+    // helpful (with chain training, it was not), and that setting them all to 1
+    // is better.  We could add a boolean option to this program to control
+    // that; but I don't want to add such an option if experiments show that it
+    // is not helpful.
+    for (int32 i = 0; i < num_frames_subsampled; ++i) {
+      int32 t = (i + start_frame_subsampled) * frame_subsampling_factor;
       if (t < pdf_post.size())
         labels[i] = pdf_post[t];
-      for (std::vector<std::pair<int32, BaseFloat> >::iterator
-               iter = labels[i].begin(); iter != labels[i].end(); ++iter)
-        iter->second *= chunk.output_weights[i];
+      for (auto& label_post : labels[i])
+        label_post.second *= chunk.output_weights[i * frame_subsampling_factor];
     }
 
-    eg.io.push_back(NnetIo("output", num_pdfs, 0, labels, frame_subsampling_factor));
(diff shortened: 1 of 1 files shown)


kkm000:
Thanks for checking! Was just going to e-mail you, but you noticed the PR.

Regarding the encapsulation of subsampling--the code should be more readable, the readability is one of our strengths. I'm mot very insistent on making these changes, but reading the code is not easy. I've gone through other binaries in nnet3bin yesterday looking for the same mistake, and, frankly, I cannot tell for sure it's not present in other files. Some spots I just could not decipher. That's not a good sign. I'm trying to think how to better explicitly mark which indices are downsampled at the least for human consumption. If I could also recruit a compiler's help by making them different, not implicitly convertible types, that would be ideal, but I already see where it's not going to work.

Breaking stuff--agree, any change to convoluted code has such risks, but balanced by decreasing chances of breaking stuff in the future.

I'll think more of this, but not rushing. Too much new stuff to do, too little time to catch up.

PR #4185, opened 2020-07-20
[src] Remove configure support for threaded math

### LINKED ISSUE #3192: Remove all threaded BLAS options from configure

kkm000:
Closes #3192

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -83,15 +83,11 @@ Configuration options:
   --mathlib=LIB         Math library [default=MKL|OPENBLAS, based on platform]
                         Supported libraries: ATLAS, MKL, CLAPACK, OPENBLAS.
   --static-math         Build with static math libraries [default=no]
-  --threaded-math       Build with multi-threaded math libraries [default=no]
-  --threaded-atlas      Build with multi-threaded ATLAS libraries [default=no]
   --atlas-root=DIR      ATLAS root directory [default=../tools/ATLAS/]
   --openblas-root=DIR   OpenBLAS root directory
   --clapack-root=DIR    CLAPACK root directory
   --mkl-root=DIR        MKL root directory
   --mkl-libdir=DIR      MKL library directory
-  --mkl-threading=LIB   MKL threading layer [default=sequential]
-                        Supported layers: sequential, iomp, tbb, gomp.
   --omp-libdir=DIR      OpenMP directory
   --speex-root=DIR      SPEEX root directory
   --speex-libdir=DIR    SPEEX library directory
@@ -169,12 +165,12 @@ ucase () { awk '{print toupper($0)}' <<<"$1" ; }
 
 function failure {
   echo "***configure failed: $* ***" >&2
-  if [ -f kaldi.mk ]; then rm kaldi.mk; fi
+  rm -f kaldi.mk
   exit 1;
 }
 
 function check_exists {
-  if [[ ! -f $1 ]]; then failure "$1 not found."; fi
+  [[ -f $1 ]] || failure "$1 not found."
 }
 
 function check_library {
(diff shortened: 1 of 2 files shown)


kkm000:
@jtrmal, if you have a few minutes, could you PTAL? I checked MKL, but I do not have Atlas handy, although changes there are minimal.

jtrmal:
LGTM, please merge

kkm000:
@jtrmal Thanks!

psmit:
Unfortunately, this PR broke static MKL compilation. Originally link lines for static MKL contained `-l/opt/intel/mkl/lib/intel64/libmkl_core.a` and this has been changed to `-lmkl`. For static compilation, the search path is not set, so this fails.

Possible fixes are:
1. Also set the search path (-L) for static compilation (like for non-static https://github.com/kaldi-asr/kaldi/blob/master/src/configure#L283). Not sure if that would solve the problem.
2. Restore the old link line by changing https://github.com/kaldi-asr/kaldi/blob/master/src/configure#L295 from:
```
linkline+=" -l$file"
```

to

```
if ! $static; then
 linkline+=" -l$file"
else
  linkline+="-l$libfile"
fi
```

I'm not sure what is the best option, but I'm happy to provide a PR if needed.

danpovey:
Would be great if you could provide a PR, Peter!  Sorry dont have an
opinion about which is best, you can decide.

kkm000:
@psmit, I've seen your PR, thanks. I am wondering what benefits you are getting form linking to MKL statically, and what are the binary sizes? If you have an approximate idea by how much this increases the binary sizes? They are about 20GB on Linux x64 with gcc and dynamic MKL.

I think—better said, thought—that we should to switch to their "single runtime library" linking, or otherwise there are conflicts when calling Kaldi libraries from Python. But that is not available in static. Understanding your use case would be very helpful for us.

daanzu:
@kkm000 I haven't tested this PR or @psmit's yet, but I statically link MKL to simplify distribution. My project only needs to distribute (from Kaldi) one shared library (which is used by Python via cffi) and a few CLI programs, and doing so is far simpler when linking MKL & OpenFST statically. I would hate to lose that ability.

psmit:
Thanks for merging and the extensive comment. 

Regarding MKL, I'm pretty sure all the architectures are in the binary, but that is more a guess than that I checked (or read that manual).

Many of the techniques mentioned, like multi-stage docker files, are indeed what we are using. It took us a lot of time to get our image sizes down, and for us, this was back then with the static MKL.

In the near future, I'm looking to bench things more on our side, if this happens I'll share the results and observations also here.

PR #4191, opened 2020-07-24
Fix potential use-after-free in HasBannedPrefixPlusDigits

orivej:
Clang 10 reports that the copy of the string returned by `siter.Symbol()` will be freed at the end of the line, so its c_str pointer will become invalid:
```
src/fstext/pre-determinize-inl.h:239:23: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
    const char *sym = siter.Symbol().c_str();
                      ^~~~~~~~~~~~~~
```

### DIFF
--- a/src/fstext/pre-determinize-inl.h
+++ b/src/fstext/pre-determinize-inl.h
@@ -236,15 +236,15 @@ inline bool HasBannedPrefixPlusDigits(SymbolTable *symTable, std::string prefix,
   const char *prefix_ptr = prefix.c_str();
   size_t prefix_len = strlen(prefix_ptr);  // allowed to be zero but not encouraged.
   for (SymbolTableIterator siter(*symTable); !siter.Done(); siter.Next()) {
-    const char *sym = siter.Symbol().c_str();
-    if (!strncmp(prefix_ptr, sym, prefix_len)) {  // has prefix.
+    const string &sym = siter.Symbol();
+    if (!strncmp(prefix_ptr, sym.c_str(), prefix_len)) {  // has prefix.
       if (isdigit(sym[prefix_len])) {  // we don't allow prefix followed by a digit, as a symbol.
         // Has at least one digit.
         size_t pos;
         for (pos = prefix_len;sym[pos] != '\0'; pos++)
           if (!isdigit(sym[pos])) break;
         if (sym[pos] == '\0') {  // All remaining characters were digits.
-          if (bad_sym != NULL) *bad_sym = (std::string) sym;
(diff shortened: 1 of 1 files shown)


src/fstext/pre-determinize-inl.h
```diff
@@ -236,15 +236,15 @@ inline bool HasBannedPrefixPlusDigits(SymbolTable *symTable, std::string prefix,
...
   size_t prefix_len = strlen(prefix_ptr);  // allowed to be zero but not encouraged.
   for (SymbolTableIterator siter(*symTable); !siter.Done(); siter.Next()) {
-    const char *sym = siter.Symbol().c_str();
-    if (!strncmp(prefix_ptr, sym, prefix_len)) {  // has prefix.
+    const string &sym = siter.Symbol();
+    if (!strncmp(prefix_ptr, sym.c_str(), prefix_len)) {  // has prefix.
       if (isdigit(sym[prefix_len])) {  // we don't allow prefix followed by a digit, as a symbol.
         // Has at least one digit.
         size_t pos;
         for (pos = prefix_len;sym[pos] != '\0'; pos++)
```
kkm000, line 244: Please also change this to 
```suggestion
        for (pos = prefix_len; pos < sym.length(); ++pos)
```
because accessing the string beyond the last character using the `operator[]()` is technically an UB. `length()` is guaranteed O(1) by the C++11 and beyond (and the compiler knows that the string is const in any case).

src/fstext/pre-determinize-inl.h
```diff
@@ -236,15 +236,15 @@ inline bool HasBannedPrefixPlusDigits(SymbolTable *symTable, std::string prefix,
...
           if (!isdigit(sym[pos])) break;
         if (sym[pos] == '\0') {  // All remaining characters were digits.
```
kkm000, line 246: Same here:
```suggestion
        if (pos == sym.length()) {  // All remaining characters were digits.
```

src/fstext/pre-determinize-inl.h
orivej: `siter.Symbol()` returns a copy of the string, and this temporary string may be freed at the end of the expression.

C++11 specifies that `str[str.length()]` returns '\0': see [operator\[\]](https://en.cppreference.com/w/cpp/string/basic_string/operator_at). Should I still change this as you suggest?

kkm000: Thanks, I did not know that C++11 defined this! No, no changes required then, LGTM.

Issue #4157, opened 2020-07-06
validate_data_dir.sh causes some errors for non-ascii characters


sw005320:
I have encountered the following issues when we use `validate_data_dir.sh`:
```
utils/validate_data_dir.sh: text contains 13489 lines with non-printable character
```
It happens almost all non-English corpora.
It seems that this error depends on the Linux (probably locale?) environment, and I had a problem in one environment but it did not happen in the other environment, but multiple people reported the same issue.

This check seems to be introduced recently from the following changes and I did not encounter the issue before.
https://github.com/kaldi-asr/kaldi/commit/e6729879ff1a4d677fd9efe3777308c8faca8130#diff-312b1ebd7cf78ac1a6fd63ebbe575800
https://github.com/kaldi-asr/kaldi/commit/31c2baeb53737f11f9a06a7ca9cc27d9b0e6438a#diff-312b1ebd7cf78ac1a6fd63ebbe575800 

Since `utils/validate_data_dir.sh` is used for various places (e.g., it is called from `copy_data_dir.sh`), it is difficult to solve this by only adding `--non-print`

aarora8:
I can try to work on it and try to fix it. I can try to recreate it on COE (where it happened) and from @danpovey @jtrmal  @m-wiesner suggestion can try to fix it.

johnjosephmorgan:
While you're at it, could you look at #4144. I think it might be related.

o-alexandre-felipe:
This issue is actually related to a check I introduced.
The C++ makes some assumptions about tokens https://github.com/kaldi-asr/kaldi/blob/00625e85130ace8baa24ee1c59423268916919fa/src/util/text-utils.cc#L105 this check was supposed to ensure that the assumed condition holds for the text data. If a better support for UTF-8 is required I would suggest to change the tokenization as well.

danpovey:
I think the issue is not that fundamental, but because the locale C.UTF-8
is not always defined.
In any case I think the easiest fix would be to simply remove the check.
We already check for that kind of compatibility issue in utils/
validate_text.pl, e.g. see validate_utf8_whitespaces.
(Checks that there are no whitespaces that would not be treated as
whitespaces by the C++ code).

Actually,  @o-alexandre-felipe, IIRC you never told us precisely what issue
you were trying to solve.
I should have asked though.

danpovey:
I think the only things that IsToken() is rejecting are non-printable ASCII
characters, and 255, which is already checked.
[...]

PR #4193, opened 2020-07-27
WIP -- untested, should address #4157

jtrmal:
@kkm000 @danpovey 
guys how about this... The discussion under the original issue went a bit far, so I just want to check, if this would be minimal change fulfilling your expectations

### DIFF
--- a/egs/wsj/s5/utils/validate_data_dir.sh
+++ b/egs/wsj/s5/utils/validate_data_dir.sh
@@ -6,17 +6,15 @@ no_feats=false
 no_wav=false
 no_text=false
 no_spk_sort=false
-non_print=false
 
 
 function show_help
 {
-      echo "Usage: $0 [--no-feats] [--no-text] [--non-print] [--no-wav] [--no-spk-sort] <data-dir>"
+      echo "Usage: $0 [--no-feats] [--no-text] [--no-wav] [--no-spk-sort] <data-dir>"
       echo "The --no-xxx options mean that the script does not require "
       echo "xxx.scp to be present, but it will check it if it is present."
       echo "--no-spk-sort means that the script does not require the utt2spk to be "
       echo "sorted by the speaker-id in addition to being sorted by utterance-id."
-      echo "--non-print ignore the presence of non-printable characters."
       echo "By default, utt2spk is expected to be sorted by both, which can be "
       echo "achieved by making the speaker-id prefixes of the utterance-ids"
       echo "e.g.: $0 data/train"
@@ -29,9 +27,6 @@ while [ $# -ne 0 ] ; do
       ;;
     "--no-text")
       no_text=true;
-      ;;
-    "--non-print")
-      non_print=true;
       ;;
     "--no-wav")
       no_wav=true;
--- a/egs/wsj/s5/utils/validate_data_dir.sh
+++ b/egs/wsj/s5/utils/validate_data_dir.sh
@@ -125,12 +125,6 @@ fi
 
 num_utts=`cat $tmpdir/utts | wc -l`
 if ! $no_text; then
-  if ! $non_print; then
-    # compatible with bin/align-text
-    n_non_print=$(LC_ALL="C.UTF-8" grep -c '[^[:print:][:space:]]' $data/text) && \
-    echo "$0: text contains $n_non_print lines with non-printable characters" &&\
-    exit 1;
-  fi
   utils/validate_text.pl $data/text || exit 1;
   check_sorted_and_uniq $data/text
   text_len=`cat $data/text | wc -l`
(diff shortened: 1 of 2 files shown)


egs/wsj/s5/utils/validate_data_dir.sh
```diff
@@ -125,12 +125,6 @@ fi
 
 num_utts=`cat $tmpdir/utts | wc -l`
 if ! $no_text; then
```
danpovey, line 127: can remove the --no-print option, which was introduced with the PR that added this.

egs/wsj/s5/utils/validate_text.pl
```diff
@@ -103,16 +93,17 @@ sub check_allowed_whitespace {
...
   if ($is_utf) {
     my $has_invalid_whitespaces = validate_utf8_whitespaces(\@lines);
     if ($has_invalid_whitespaces) {
       print STDERR "$0: ERROR: text file '$filename' contains disallowed UTF-8 whitespace character(s)\n";
       return 0;
     }
+    return 1;
   }
-  return 1;
+  return 0;
```
danpovey, line 106: i think this change may not be right.. it does `check_allowed_whitespace(..) or exit 1`.

egs/wsj/s5/utils/validate_text.pl
```diff
@@ -83,12 +73,12 @@ sub validate_utf8_whitespaces {
...
-      print STDERR "$0: The line for utterance $utt_id contains CR (0x0D) character\n";
+      print STDERR "$0: The line number $i (key $utt_id) contains CR (0x0D) character (we do not support Windows/DOS-style end-of-line characters)\n";
       return 1;
     }
-    $current_line =~ s/[\x{0009}\x{000a}\x{0020}]/./g;
+    $current_line =~ s/[\x{0009}\x{000a}\x{0020}\x{007f}\x{00ff}]/./g;
     if ($current_line =~/\s/) {
-      print STDERR "$0: The line for utterance $utt_id contains disallowed Unicode whitespaces\n";
+      print STDERR "$0: The line number $i (key $utt_id) contains disallowed Unicode whitespaces\n";
       return 1;
```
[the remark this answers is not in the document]
jtrmal: good point, let me find out what re classes are available.
y.

egs/wsj/s5/utils/validate_text.pl
jtrmal: IMO the change is right -- if the text is not UTF-or-ASCII, we have to return false, so that ' or exit 1' gets executed

egs/wsj/s5/utils/validate_data_dir.sh
jtrmal: removed

egs/wsj/s5/utils/validate_text.pl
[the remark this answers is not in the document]
jtrmal: resolved

danpovey:
thanks.. any chance you could point at least one person who asked about
this, to your PR to have them help test?

jtrmal:
@sw005320 @MartinKocour @o-alexandre-felipe 
guys, could you please test it out?

egs/wsj/s5/utils/validate_text.pl
```diff
@@ -53,18 +48,13 @@ sub get_utf8_or_bytestream {
       $is_utf_compatible = $is_utf_compatible && defined($decoded_text);
       push @unicode_lines, $decoded_text;
     } else {
-      #print STDERR "WARNING: the line $raw_text cannot be interpreted as UTF-8: $decoded_text\n";
-      ;
+      print STDERR "ERROR: the line number $lineno (containing $raw_text) cannot be interpreted as a valid UTF-8 or (7-bit) ASCII\n";
+      return (0, $unicode_lines)
```
sw005320, line 52:
```suggestion
      return (0, @unicode_lines)
```

jtrmal:
@MartinKocour @o-alexandre-felipe it would be good, if you verify for your use-case, too (as you were providing fixes to the original script)

danpovey:
I think that would better belong in fix_data_dir.sh, because those
properties might be relied on by code that would otherwise crash.

kkm000:
shut up stale bot.

kkm000:
@jtrmal friendly ping. Is this still relevant?

I slapped a "stale-exclude" label on it, but I can turn this bi-monthly annoyer back if you want to postpone it.

danpovey:
I'd be OK to merge if Yenda thinks so, in that I don't see obvious  problems; however I haven't looked super closely and it looks like the original issue was a can of worms.  But I'm all for getting rid of the use of grep with its locale dependency.

kkm000:
I have another gripe with grep: its exit code is non-zero not only in case
of error, but also when nothing was found, and the $? is the ugliest
variable name there is in Bash. I find myself using Perl in place of grep,
Awk and sed more and more often.

jtrmal:
I propose to close it without merging then
y.

danpovey:
@jtrmal but this PR was *removing* the use of grep, not adding it.

danpovey:
@jtrmal how about you remove "draft", I merge, and we address any problems later?

jtrmal:
there is a conflict with the current version  and it's been laying around for half a year... I doubt it's needed -- the people having the issues didn't test it and no new issues were reported... the validation is such a finely tuned mess that it's hard to get things fixed without breaking something else...

Issue #4178, opened 2020-07-17
CUDA online pipeline segfaults


twisteroidambassador:
I wrote a online recognition server using BatchedThreadedNnet3CudaOnlinePipeline. When I throw many concurrent requests at it, the program sometimes exit with a segmentation fault. I managed to grab a backtrace using `catchsegv`, and used `addr2line` to identify the lines of code. (Some line numbers seems off by one.)

Backtrace:
$HOME/kaldi/src/lib/libkaldi-cudadecoder.so(_ZN5kaldi12cuda_decoder11CudaDecoder23AddFinalTokensToLatticeEiPSt6vectorISt4pairIiNS0_9InfoTokenEESaIS5_EEPSt13unordered_mapIiNS1_15RawLatticeStateESt4hashIiESt8equal_toIiESaIS3_IKiSA_EEEPN3fst9VectorFstINSK_6ArcTplINSK_16LatticeWeightTplIfEEEENSK_11VectorStateISP_SaISP_EEEEE+0x15d)[0x4a209] [kaldi/src/cudadecoder/cuda-decoder.cc:1211]
https://github.com/kaldi-asr/kaldi/blob/2c7e78f0757120a965e671f2843c7c07ac7141d4/src/cudadecoder/cuda-decoder.cc#L1210-L1211
$HOME/kaldi/src/lib/libkaldi-cudadecoder.so(_ZN5kaldi12cuda_decoder11CudaDecoder36ConcurrentGetRawLatticeSingleChannelEiPN3fst9VectorFstINS2_6ArcTplINS2_16LatticeWeightTplIfEEEENS2_11VectorStateIS7_SaIS7_EEEEE+0x2b9)[0x4a82f] [kaldi/src/cudadecoder/cuda-decoder.cc:1553]
https://github.com/kaldi-asr/kaldi/blob/2c7e78f0757120a965e671f2843c7c07ac7141d4/src/cudadecoder/cuda-decoder.cc#L1544-L1553
$HOME/kaldi/src/lib/libkaldi-cudadecoder.so(_ZN5kaldi12cuda_decoder38BatchedThreadedNnet3CudaOnlinePipeline16FinalizeDecodingEim+0x38)[0x5d30c] [kaldi/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc:417]
[...]

PR #4196, opened 2020-07-29
Return empty lattice for zero frames decoded

hugovbraun:
Bug #4178. Currently if we try to call GetRawLattice on a channel with 0 frames decoded it will crash. It should now return an empty lattice instead.
Note: The bug still exists for GetBestPath. I'll fix it as soon as we confirm the fix for GetRawLattice.

### DIFF
--- a/src/cudadecoder/cuda-decoder.cc
+++ b/src/cudadecoder/cuda-decoder.cc
@@ -1007,6 +1007,8 @@ void CudaDecoder::GetBestCost(const std::vector<ChannelId> &channels,
   // Waiting for the copy
   cudaStreamSynchronize(compute_st_);
   for (int32 ilane = 0; ilane < nlanes_used_; ++ilane) {
+    int ichannel = channels[ilane];
+    KALDI_ASSERT(NumFramesDecoded(ichannel) > 0);
     int2 minarg = h_lanes_counters_.lane(ilane)->min_int_cost_and_arg;
     // Min cost in that channel last token queue
     CostType min_cost = orderedIntToFloatHost(minarg.x);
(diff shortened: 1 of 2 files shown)


src/cudadecoder/cuda-decoder.cc
```diff
@@ -1585,7 +1594,11 @@ void CudaDecoder::ConcurrentGetRawLatticeSingleChannel(const ChannelId ichannel,
       "You need to call PrepareForGetRawLattice before "
       "ConcurrentGetRawLatticeSingleChannel" &&
       best_cost_idx >= 0);
+  fst_out->DeleteStates();  // reset out lattice
   const int32 nframes = NumFramesDecoded(ichannel);
+  // If no frames were decoded, leaving the lattice empty and return
+  if (nframes == 0) return;
+
```
twisteroidambassador, line 1601: Moving the assert below the `nframes == 0` check allow it to run properly.

hugovbraun: Great, thanks

hugovbraun:
Thanks @twisteroidambassador for checking. I'll clean up the patch now (and fix GetBestPath)

hugovbraun:
Cleaned up the patch. Did not fix GetBestPath as we're going to remove that version anyway (replaced by partial hypotheses).

@twisteroidambassador any chance you could check that everything is working with the latest commit? Thanks!

twisteroidambassador:
Looks like it's working correctly. Thanks!

danpovey:
still marked draft.

hugovbraun:
Removed the draft. Thanks!

twisteroidambassador:
Should this PR be merged? Or is the issue already taken care of in another pull request?

hugovbraun:
It's ready for merge on our end.

PR #4198, opened 2020-07-30
[fix:cudadecoder]application hangs if numFrames == 0

spike-liu:
cuda_pipeline->WaitForGroup() will never return if numFrames == 0, so maintain all_group_tasks_not_done variable and notify_all if nothing is left in the group.


hugovbraun:
Makes sense. Thanks for your help!

spike-liu:
You are welcome, @hugovbraun.

kkm000:
@hugovbraun, my understanding is that this code is being/has been deprecated in favor of batched-...-pipeline2.cc? Is this moot now? Is it supported on Jetson only, or not at all?

kkm000:
We've deprecated batched-threaded-nnet3-cuda-pipeline.cc, but have not removed it yet. So I'm merging this fix. The CI failure is unrelated.

PR #4210, opened 2020-08-06
[src] Dynamic batcher for cuda online pipeline

hugovbraun:
Still a draft but getting close.

Adding a dynamic batcher for the cuda online pipeline. The dynamic batcher takes care of automatically creating the batches for the pipeline. It allows the user to just call dynamic_batcher.Push(new chunk) with whatever chunk is ready, from whichever audio channel. The batcher will take care of the rest.

```dynamic_batcher.Push(CorrelationID corr_id, bool is_first_chunk, bool is_last_chunk, SubVector<Basefloat> wave_part);```

With corr_id a unique uint64_t identifier for each channel. Results are returned using callbacks:

```
cuda_pipeline.SetBestPathCallback()
cuda_pipeline.SetLatticeCallback()
```

And that's all. The rest is handled by the batcher and the pipeline.

The best path callback also returns partial hypotheses and endpointing. Both callbacks are optional (but you should of course define at least one).
If the lattice callback is not set, the lattice will not be generated, reducing the pressure on the CPU. Please note that recent GPUs come with a fairly large amount of memory (~40GB), which might allow some users to skip the lattice rescoring and just use larger FSTs instead.

The online example binary has been rewritten to use the new batcher and simplify understanding of how it works. I'll add comments on methodology in the file. 
For now, the important thing to know is that we've chosen the most comprehensive definition of latency, which is "time between speaker is done talking and the callback with the result is running". 
If we start streaming a speaker at t1, and that speaker corresponds to an audio file of duration d, then we consider the speaker as done talking at t2=t1+d. We then get the final result back at t3. Latency is t3-t2. It includes all possible reason for delay (delay due to dynamic batching, processing of all chunks, context flush, final result generation, etc.). Except network, which is outside the scope of this work for now.

Here are some latency numbers in single GPU measured with the eg model "librispeech" on librispeech/test_clean:

```
=== 3500 parallel users ===
V100
Latencies (s):	Avg		90%		95%		99%
 		0.117		0.161		0.172		0.206
Ampere GPU
Latencies (s):	Avg		90%		95%		99%
 		0.069		0.093		0.101		0.118

=== 7000 parallel users ===
Ampere GPU
Latencies (s):	Avg		90%		95%		99%
		0.225		0.312		0.355		0.437
```

### DIFF
--- a/src/cudadecoder/Makefile
+++ b/src/cudadecoder/Makefile
@@ -18,7 +18,7 @@ OBJFILES = cuda-decoder.o cuda-decoder-kernels.o cuda-fst.o \
 	   batched-threaded-nnet3-cuda-pipeline.o \
 	   batched-threaded-nnet3-cuda-pipeline2.o \
 	   batched-static-nnet3.o batched-static-nnet3-kernels.o \
-	   decodable-cumatrix.o
+	   cuda-online-pipeline-dynamic-batcher.o decodable-cumatrix.o
 
 LIBNAME = kaldi-cudadecoder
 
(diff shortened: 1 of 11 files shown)


danpovey:
Cool!
I hope it won't break existing APIs?

hugovbraun:
It's a layer on top of the existing work. It doesn't change the existing APIs for the pipeline (it uses them).

danpovey:
OK, cool.  Let us know when ready.

hugovbraun:
@danpovey this is ready

danpovey:
Thanks!  Would you mind removing the "draft" label?

hugovbraun:
@danpovey done

PR #4225, opened 2020-08-19
[src] fix passing std::ostream by ref

andreytkachenko:
openfst expecting that by ref:
```c++
// /include/fst/script/print-impl.h
template <class Arc>
class FstPrinter {
 public:
  using StateId = typename Arc::StateId;
  using Label = typename Arc::Label;
  using Weight = typename Arc::Weight;
  ...

  // Prints FST to an output stream.
  void Print(std::ostream &ostrm, const std::string &dest) {
      ...
  }
```


andreytkachenko:
I have tested with OpenFst 1.7.9, but current supported version (from travis build) is 1.6.7. Closing

Issue #4227, opened 2020-08-21
align_ctm_ref.py correct match at (1,1) not in the result


wangsiw1:
Hello all. I am new to kaldi and recently working on a project utilizing part of kaldi functions. One is segment long utterances, and when running align_ctm_ref.py, it seems it will not append correct match at position (1,1) to the result due to this part of code:

https://github.com/kaldi-asr/kaldi/blob/a1502739d5c8f7d272d364f31807958b771f141e/egs/wsj/s5/steps/cleanup/internal/align_ctm_ref.py#L287-L291

Use --debug-only option and edit first letter of both REF and HYP to make them same. Output:
```
# Verbose=2
[random@host ~]$ ./align_ctm_ref_new.py --hyp-format=CTM --hyp=tmp.txt --ref=tmp.txt --output=tmp.txt --verbose=2 --debug-only=true --align-full-hyp
2020-08-21 07:38:32,270 [./align_ctm_ref_new.py:484 - test_alignment - INFO ] REF: AGCACACA
2020-08-21 07:38:32,270 [./align_ctm_ref_new.py:485 - test_alignment - INFO ] HYP: AGCCAT
Alignment G C - C A T
Exiting since --debug-only was true

# Verbose=3
[random@host ~]$ ./align_ctm_ref_new.py --hyp-format=CTM --hyp=tmp.txt --ref=tmp.txt --output=tmp.txt --verbose=3 --debug-only=true --align-full-hyp
...
2020-08-21 07:16:03,629 [./align_ctm_ref_new.py:272 - smith_waterman_alignment - DEBUG ] Alignment score: 8 for (7, 6)
0 -1 -2 -3 -4 -5 -6
0 2 1 0 -1 -2 -3
0 1 4 3 2 1 0
0 0 3 6 5 4 3
0 2 2 5 5 7 6
0 1 1 4 7 6 6
0 2 1 3 6 9 8
0 1 1 3 5 8 8
0 2 1 2 4 7 7
2020-08-21 07:16:03,629 [./align_ctm_ref_new.py:328 - smith_waterman_alignment - DEBUG ] Aligned output:
2020-08-21 07:16:03,629 [./align_ctm_ref_new.py:330 - smith_waterman_alignment - DEBUG ] (2,2)  -  (3,3)  -  (4,3)  -  (5,4)  -  (6,5)  -  (7,6)
[...]

PR #4229, opened 2020-08-21
Fix the problem that Smith-Waterman alignment might miss (1, 1) in align_ctm_ref.py (#4227)

DongjiGao:
Fix the problem that the Smith-Waterman alignment might miss (1, 1) during backtracing if they match

### DIFF
--- a/egs/wsj/s5/steps/cleanup/internal/align_ctm_ref.py
+++ b/egs/wsj/s5/steps/cleanup/internal/align_ctm_ref.py
@@ -498,7 +498,7 @@ def ctm_line_to_string(ctm_line):
 
 
 def test_alignment(align_full_hyp):
-    hyp = "CGCCAT"
+    hyp = "GCCAT"
     ref = "AGCACACA"
 
     verbose = 3
--- a/egs/wsj/s5/steps/cleanup/internal/align_ctm_ref.py
+++ b/egs/wsj/s5/steps/cleanup/internal/align_ctm_ref.py
@@ -288,9 +288,8 @@ def smith_waterman_alignment(ref, hyp, similarity_score_function,
                     or (prev_ref_index, prev_hyp_index) == (0, 0)):
                 score = H[ref_index][hyp_index]
                 if score != 0:
-                    ref_word = ref[ref_index-1]
-                    hyp_word = hyp[hyp_index-1]
-                    assert ref_word != eps_symbol and hyp_word != eps_symbol
+                    ref_word = ref[ref_index-1] if ref_index > 0 else eps_symbol
+                    hyp_word = hyp[hyp_index-1] if hyp_index > 0 else eps_symbol
                     output.append((ref_word, hyp_word, prev_ref_index,
                         prev_hyp_index, ref_index, hyp_index))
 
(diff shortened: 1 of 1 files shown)


wangsiw1:
@DongjiGao everything works fine except line 290, where it is possible to have negative score in the algorithm, and in my test case it raised exception at assertion on line 293 for an alignment with a word and a eps symbol. No problem other than that.
https://github.com/kaldi-asr/kaldi/blob/55e98edf82368177cfc5daeb596e891845a6225c/egs/wsj/s5/steps/cleanup/internal/align_ctm_ref.py#L290-L293

```
# Output log
2020-08-24 03:24:48,586 [./align_ctm_ref_new.py:329 - smith_waterman_alignment - ERROR ] Unexpected entry (0,0) -> (0,1), word_1, <eps>
2020-08-24 03:24:48,586 [./align_ctm_ref_new.py:599 - run - ERROR ] Alignment failed for recording 022-00018122-00000000-00000225 with ref = word_1 word_2 word_3 word_4 and hyp = <eps> word_1 word_2 word_3 word_4 <eps>
2020-08-24 03:24:48,586 [./align_ctm_ref_new.py:615 - main - ERROR ] Failed to align ref and hypotheses; got exception 
  File "./align_ctm_ref_new.py", line 291, in smith_waterman_alignment
    assert ref_word != eps_symbol and hyp_word != eps_symbol
AssertionError

During handling of the above exception, another exception occurred:

  File "./align_ctm_ref_new.py", line 612, in main
    run(args)
  File "./align_ctm_ref_new.py", line 579, in run
    align_full_hyp=args.align_full_hyp)
  File "./align_ctm_ref_new.py", line 330, in smith_waterman_alignment
    raise RuntimeError("Unexpected result: Bug in code!!")
RuntimeError: Unexpected result: Bug in code!!
```

DongjiGao:
> @DongjiGao everything works fine except line 290, where it is possible to have negative score in the algorithm
Thanks. I will fix that.
> and in my test case it raised exception at assertion on line 293 for an alignment with a word and a eps symbol. 
The eps_symbol only indicates there's empty in the alignment. It's a bit weird to have that symbol in your text. I think it might cause confusion. @danpovey what do you think?

wangsiw1:
Those symbols are from decoding result(hyp) rather than our text input(ref). There is some silence in our audio input and I guess those eps_symbols were used to indicate the silent parts?

DongjiGao:
> Those symbols are from decoding result(hyp) rather than our text input(ref). There is some silence in our audio input and those eps_symbols were used to indicate the silent parts, which should be fine I guess? Please correct me if I'm wrong.

I am talking about the eps_symobl ("-" by default) in align_ctm_ref.py. It is different from "<eps>" (which is silence) in your result. You might have set --eps-symbol="<eps>"?

danpovey:
Dongji, are you sure that eps aren't expected in the hyp?  I thought we
might map silence to that?  But not sure about this.

wangsiw1:
We are using steps/cleanup/segment_long_utterances_nnet3.sh in our pipeline where --eps-symbol is set to "<eps>" by default.

DongjiGao:
> We are using steps/cleanup/segment_long_utterances_nnet3.sh in our pipeline where --eps-symbol is set to "" by default.

Oh, I see. I was wrong. I will fix that. Thanks!

wangsiw1:
No problem, thanks for help!

DongjiGao:
> Dongji, are you sure that eps aren't expected in the hyp? I thought we might map silence to that? But not sure about this.

You are right. The eps is allowed in hyp.

Issue #4228, opened 2020-08-21
Docker images no longer built


kir4h:
While checking https://hub.docker.com/r/kaldiasr/kaldi/tags, I noticed Docker images haven't been updated in the latest 3 months.

I was trying to understand what is broken and ended up in https://github.com/kaldi-asr/kaldi/issues/3284 where the Docker discussion started and there is a link to https://github.com/mdoulaty/kaldi-image-builder that I guess has been used for pushing those images.

From that conversation, [it seems](https://github.com/kaldi-asr/kaldi/issues/3284#issuecomment-499549272) @mdoulaty is running this on his account, but I'm not sure if this is automated or just a manual trigger from his side. So I guess maybe @mdoulaty can shed some light on what is the current status (as maybe the reason is that there was an automatic procedure that is now stopped, or there was a manual procedure that is no longer done to trigger the builds)

Thank you!

kir4h:
> @kir4h, @mdoulaty, since you guys automate the process, could you please continue work on reducing the image size? Here's an easy 1G catch for ya).
> 
> ```
> $ du -hc $(find /opt/intel -name '*.a') | tail -1
> 936M    total
> $ du -hs ~/work/kaldi/.git
> 139M    /home/kkm/work/kaldi/.git
> ```
> 
> If you want to go a bit bolder and shave another 350MB, you can remove MKL kernels for pre-AVX CPUs (Pentium 4 and early Core/Core 2) CPUs, and Knight Landing MIC accelerators (hardly anyone has them, they were sparklingly unsuccessful, and very likely won't virtualize in Docker anyway). It's also safe to remove threading layer libraries and the iLP64 interface, since we link against the sequential layer only (libmkl_sequential.so, does not have `_thread` in file name) and LP64:
> 
> ```
> $ du -hc $(find /opt/intel -type f -regex '.*\(_mc.?\|_mic\|_thread\|_ilp64\)\.so') | tail -1
> 352M    total
> ```
> 
> This and the *.a removes more than half of MKL disk space use.

These are straightforward, if ok I will create the PR straight away

> Also, it often happens that latest commits break stuff. I'd release a week-old code, because it would be more stable. You may need some Git-fu for this; to get the SHA of the latest 1-week-old commit, use `git rev-list`, e. g.
> 
> ```
> $ git rev-list --first-parent -n1 --until=1week master
> 8eb569178d6926b979726df6b1a8b99c0784a52d
[...]

PR #4231, opened 2020-08-24
Further reductions on Docker image size

kir4h:
Signed-off-by: Andres Alvarez <kir4h.gh@gmail.com>

Following the suggestions by @kkm000  in https://github.com/kaldi-asr/kaldi/issues/4228#issuecomment-678452211, some extra deletions for further reduction on the image size. 

We are going down from `5.26GB` to `3.88GB` on the regular one and from `8.21GB` to `6.83GB` on the gpu version

```
kir4h/kaldi                                             gpuv2                           8be3bd9043ff        4 minutes ago       6.83GB
kir4h/kaldi                                             v2                              9ae588678bac        46 minutes ago      3.88GB
kir4h/kaldi                                             gpu                             d2fe407ffe3b        4 days ago          8.21GB
kir4h/kaldi                                             latest                          141039043d6f        4 days ago          5.26GB
kaldiasr/kaldi                                          gpu-latest                      0b92c53f2ef1        3 months ago        14.9GB
kaldiasr/kaldi                                          latest                          b44678467882        3 months ago        11.6GB
```

### DIFF
--- a/docker/debian9.8-cpu/Dockerfile
+++ b/docker/debian9.8-cpu/Dockerfile
@@ -35,6 +35,9 @@ RUN git clone --depth 1 https://github.com/kaldi-asr/kaldi.git /opt/kaldi && \
     ./configure --shared && \
     make depend -j $(nproc) && \
     make -j $(nproc) && \
-    find /opt/kaldi  -type f \( -name "*.o" -o -name "*.la" -o -name "*.a" \) -exec rm {} \;
+    find /opt/kaldi -type f \( -name "*.o" -o -name "*.la" -o -name "*.a" \) -exec rm {} \; && \
+    find /opt/intel -type f -name "*.a" -exec rm {} \; && \
+    find /opt/intel -type f -regex '.*\(_mc.?\|_mic\|_thread\|_ilp64\)\.so' -exec rm {} \; && \
+    rm -rf /opt/kaldi/.git
 WORKDIR /opt/kaldi/
 
--- a/docker/ubuntu16.04-gpu/Dockerfile
+++ b/docker/ubuntu16.04-gpu/Dockerfile
@@ -35,7 +35,10 @@ RUN git clone --depth 1 https://github.com/kaldi-asr/kaldi.git /opt/kaldi && \
     ./configure --shared --use-cuda && \
     make depend -j $(nproc) && \
     make -j $(nproc) && \
-    find /opt/kaldi  -type f \( -name "*.o" -o -name "*.la" -o -name "*.a" \) -exec rm {} \;
+    find /opt/kaldi  -type f \( -name "*.o" -o -name "*.la" -o -name "*.a" \) -exec rm {} \; && \
+    find /opt/intel -type f -name "*.a" -exec rm {} \; && \
+    find /opt/intel -type f -regex '.*\(_mc.?\|_mic\|_thread\|_ilp64\)\.so' -exec rm {} \; && \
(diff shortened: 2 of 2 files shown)


daanzu:
Regarding not including support for pre-AVX CPUs (as discussed in the above referenced comment), let me throw in one vote for instead maintaining support for pre-AVX CPUs, unless it saves a lot of space. (An old server in a lab with a bunch of pre-AVX cores and RAM is great for running some big non-critical Kaldi jobs.) I figure the Docker images are meant to be easy to pull and run on whatever hardware you have, and anyone very particular about the size of the image would presumably already want to modify it to only include the parts of Kaldi that they are actually using.

kir4h:
> Regarding not including support for pre-AVX CPUs (as discussed in the above referenced comment), let me throw in one vote for instead maintaining support for pre-AVX CPUs, unless it saves a lot of space. 

It´s around 357MB according to the comment. Lots for a regular Docker image, but not so much for a 4GB image

> I figure the Docker images are meant to be easy to pull and run on whatever hardware you have, and anyone very particular about the size of the image would presumably already want to modify it to only include the parts of Kaldi that they are actually using.

I'm pretty new to Kaldi so I'm not yet even aware of the different parts. I guess if we manage to improve the way we are building Kaldi Docker image (and we can build different images we can then combine into different final images) we could even have several Docker images for different scenarios (not currently doable because building the image is extremely heavy)

@daanzu @kkm000 should I leave the pre-AVX part then for the time being?

jtrmal:
IMO  AVX is a minimum reasonable HW capability
y.

daanzu:
> It´s around 357MB according to the comment. Lots for a regular Docker image, but not so much for a 4GB image

I think that 350MB figure also includes the savings from removing the MIC/threading/iLP64 stuff as well (which all seem eminently worthless to keep), in addition to just the pre-AVX support. Assuming the `_mc.?` files are the pre-AVX, that is only ~130MB, but I don't know how `_mc`/`_mc2`/`_mc3` differ (maybe some level of SSE? the user manual isn't clear).

```
45M     /opt/intel/compilers_and_libraries_2020.0.166/linux/mkl/lib/intel64_lin/libmkl_mc.so
11M     /opt/intel/compilers_and_libraries_2020.0.166/linux/mkl/lib/intel64_lin/libmkl_vml_mc2.so
11M     /opt/intel/compilers_and_libraries_2020.0.166/linux/mkl/lib/intel64_lin/libmkl_vml_mc3.so
47M     /opt/intel/compilers_and_libraries_2020.0.166/linux/mkl/lib/intel64_lin/libmkl_mc3.so
11M     /opt/intel/compilers_and_libraries_2020.0.166/linux/mkl/lib/intel64_lin/libmkl_vml_mc.so
```

kkm000:
@kir4h, @daanzu, yes, different sets of SSE instructions, and I estimated 350M all together with that stuff. And 130MB is still130 MB more than nothing. Given that releases are frequent, that may be a factor. As for an "old lab full of Pentium 4s", everyone can build their own image, or, even better, their own Kaldi, as we always did before Docker was a fad. I would not inconvenience the absolute majority of users with an extra size. I agree with @jtrmal, a non-AVX CPU is hard to come by these days.

danpovey:
@kkm please take care of merging this when appropriate...

kkm000:
@dpovey, sure, I'm not missing it—just usually do not do code reviews right after I wake up and just before crashing :)

kkm000:
Merging, thanks.

PR #4233, opened 2020-08-25
Fix data location changing of Gale Arabic data on CLSP grid and some small problem

DongjiGao:


egs/gale_arabic/s5d/run.sh
```diff
@@ -68,16 +73,16 @@ if [ $stage -le 0 ]; then
...
-
-  utils/prepare_lang.sh data/local/dict "<UNK>" data/local/lang data/lang
-
+#  local/prepare_data.sh $options
+#
+#  echo "$0: Preparing lexicon and LM..." 
+#  local/prepare_dict.sh
+#
+#  utils/prepare_lang.sh data/local/dict "<UNK>" data/local/lang data/lang
+#
```
danpovey, line 82: please uncomment this; would be better to add "stage" option.

danpovey:
Thanks!  Re the other PR, with the eps thing, please let me know when you think it's tested.

PR #4239, opened 2020-08-27
[src] CUDA 11 configure

hugovbraun:
Adding support for cuda 11 in configure script


lv-gh:
Is MIN_UNSUPPORTED_GCC_VER="9.0" requirement for CUDA 11 intentional? AFAICT CUDA 11.0 officially supports gcc 9.0:
https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-compiler-new-features

PR #4242, opened 2020-08-27
[src] Using C++14 for CUDA

hugovbraun:
Using C++14 for cuda instead of C++11. Required by CUB. The CXXFLAGS are still using c++11.

### DIFF
--- a/src/makefiles/cuda_64bit.mk
+++ b/src/makefiles/cuda_64bit.mk
@@ -10,8 +10,8 @@ CXXFLAGS += -DHAVE_CUDA -I$(CUDATKDIR)/include -fPIC -pthread -isystem $(OPENFST
 CUDA_INCLUDE= -I$(CUDATKDIR)/include -I$(CUBROOT)
 CUDA_FLAGS = --machine 64 -DHAVE_CUDA \
              -ccbin $(CXX) -DKALDI_DOUBLEPRECISION=$(DOUBLE_PRECISION) \
-             -std=c++11 -DCUDA_API_PER_THREAD_DEFAULT_STREAM  -lineinfo \
-             --verbose -Xcompiler "$(CXXFLAGS)"
+             -std=c++14 -DCUDA_API_PER_THREAD_DEFAULT_STREAM  -lineinfo \
+             --verbose -Wno-deprecated-gpu-targets
 
 ifeq ($(shell test -e $(CUDATKDIR)/lib64/libcudart_static.a && echo -n yes),yes)
 CUDA_LDFLAGS += -L$(CUDATKDIR)/lib64 -Wl,-rpath,$(CUDATKDIR)/lib64
--- a/src/configure
+++ b/src/configure
@@ -339,16 +339,20 @@ Either your CUDA is too new or too old."
       GCC_VER=$($COMPILER -dumpversion)
       GCC_VER_NUM=$(echo $GCC_VER | sed 's/\./ /g' | xargs printf "%d%02d%02d")
       case $CUDA_VERSION in
-        7_*)
-          MIN_UNSUPPORTED_GCC_VER="5.0"
-          MIN_UNSUPPORTED_GCC_VER_NUM=50000;
-          CUSOLVER=false
-        ;;
-        8_*)
-          MIN_UNSUPPORTED_GCC_VER="6.0"
-          MIN_UNSUPPORTED_GCC_VER_NUM=60000;
-          CUSOLVER=false
-        ;;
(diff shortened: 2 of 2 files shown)


danpovey:
Mm, @kkm000 and @hugovbraun, what is the chance this will cause compilatin problems, do you think?

kkm000:
No, not at all. I think we're long overdue. All major compilers support C++14 (and C++17, too). We may need to bump the minimum compiler versions in configure. I can take care of this.

src/makefiles/cuda_64bit.mk
```diff
@@ -10,8 +10,8 @@ CXXFLAGS += -DHAVE_CUDA -I$(CUDATKDIR)/include -fPIC -pthread -isystem $(OPENFST
 CUDA_INCLUDE= -I$(CUDATKDIR)/include -I$(CUBROOT)
 CUDA_FLAGS = --machine 64 -DHAVE_CUDA \
              -ccbin $(CXX) -DKALDI_DOUBLEPRECISION=$(DOUBLE_PRECISION) \
-             -std=c++11 -DCUDA_API_PER_THREAD_DEFAULT_STREAM  -lineinfo \
-             --verbose -Xcompiler "$(CXXFLAGS)"
+             -std=c++14 -DCUDA_API_PER_THREAD_DEFAULT_STREAM  -lineinfo \
+             --verbose -Wno-deprecated-gpu-targets
```
kkm000, line 14: Just to make sure, are you removing `-Xcompiler "$(CXXFLAGS)"` intentionally here?

hugovbraun:
Regarding cuda, c++14 was added with cuda 9.0, c++17 with cuda 11.0. We shouldn't have any compilation issues on the cuda side, as we are always using the same compiler (nvcc). 
It compiles without issues with cuda-9.0 and cuda-11.0. That's with the cuda flags with c++14, and cxxflags with c++11 (this PR).

src/makefiles/cuda_64bit.mk
hugovbraun: Yes, they are no longer needed - I suspect we added the -Xcompiler back when we had more CPU code in .cu files. We've since moved the pure CPU code to .cc files to speed up compilation. 
It also prevents us from the config "cuda in c++14, rest of the code in c++11", as the c++11 flag in CXXFLAGS overrides the c++14 flag

kkm000:
@hugovbraun, so this change is good to go even without switching the rest to C++11? I do not see why it should be, just confirming.

src/makefiles/cuda_64bit.mk
kkm000: Yeah, just saw the very first comment. Actually, good riddance to this flag—I remember tweaking compilation flags and splitting what goes to nvcc and what should not. Thank you for this!

hugovbraun:
However we are still supporting cuda 7 and cuda 8 in the configure script. I guess we should either remove them, or add a fallback to c++11 for those versions. However it means we'll have to keep CUB behind. What do you think @kkm000 @danpovey ?

@kkm000 yes, we can keep the rest with c++11 for now. CUB is made of .cuh files so switching only the cuda flags is enough.

kkm000:
@danpovey: This just suddenly come to my mind. Our production code has been compiled with C++14 for years, as the MS compiler dropped C++11 in v13 or v14, but years ago. This does not include CUDA, but includes a lot of nnet3 training stuff, as I could not cleanly separate training and inference code in nnet3, and left it to the linker to deal with it.

kkm000:
@hugovbraun, what about other platforms out there there besides x64? I remember aarch64 could also be compiled with CUDA. Should be no problem there? I have no idea even where to get one—is this the new Amazon CPU, Graviton or something?

@danpovey, I'm pro dropping CUDA 8-. I think CUDA 9 came out in 2012 or 2013. Probably no support above K40/80 in CUDA 8.

jtrmal:
LGTM

danpovey:
I guess removing CUDA 8 support is OK with me.  But I prefer if you remove
CUDA 8 support by commenting out parts of the configure script, and adding
a comment there that links to this PR and mentions it was about the c++14
flag (assuming that is true and there weren't other issues).
That will make it possible for someone who really wants to, to enable CUDA
8 support.

hugovbraun:
@kkm000 jetson devices will compile cuda code on arm but also use nvcc to do so, so we're good on that front.

Commented the CUDA 7 and CUDA 8 lines and added some details on how to revert those changes.

kkm000:
LGTM!

PR #4244, opened 2020-08-27
Revert "[src] adding missing cusparse errors (#4241)"

danpovey:
This reverts commit 18f0a52722a8f4ccfaf446f30d54e74d8e7bde22.


danpovey:
Reverted because it was causing build errors for older CUDA toolkit versions.

PR #4248, opened 2020-08-28
[build] aarch64 cuda11

hugovbraun:
Complement to #4239, which was missing the aarch64 part.


kkm000:
@hugovbraun, there is no capabilitiy 7.5 or 8.0 for aarch64, is this correct? I just really have no idea—does it depend on the platform?

PR #4250, opened 2020-08-31
[build] Fix CUDA includes

hugovbraun:
Fix for #4242. Probably had a stale build during testing. Sorry about that. Adding some missing includes for cuda build.

### DIFF
--- a/src/makefiles/cuda_64bit.mk
+++ b/src/makefiles/cuda_64bit.mk
@@ -7,7 +7,7 @@ endif
 
 CXXFLAGS += -DHAVE_CUDA -I$(CUDATKDIR)/include -fPIC -pthread -isystem $(OPENFSTINC)
 
-CUDA_INCLUDE= -I$(CUDATKDIR)/include -I$(CUBROOT)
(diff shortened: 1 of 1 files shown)


hugovbraun:
@kkm000 if you're around

kkm000:
@danpovey, @hugovbraun, sorry, how did I manage to miss that one!

danpovey:
Its OK, but we have another problem I think, that -fPIC is not included in the CUDA_FLAGS as it is in CXXFLAGS.
Of course ideally we'd only include that option when building dynamic libraries, not sure why it's there regardless.

hugovbraun:
Thanks Dan, I'll take a look.

danpovey:
Its OK we fixed this.
It is odd, though, that we have the -fPIC flag (for both CPU and CUDA flag
sets) regardless of whether we're doing the dynamic build, in cuda_64bit.mk.
In some other .mk templates we have this:
src/makefiles/linux_atlas_arm.mk-ifeq ($(KALDI_FLAVOR), dynamic)
src/makefiles/linux_atlas_arm.mk:CXXFLAGS += -fPIC
src/makefiles/linux_atlas_arm.mk-endif

PR #4264, opened 2020-09-11
global_cmvn_stats in 'OnlineNnet2FeaturePipelineInfo'

vesis84:
- changing so that the global_cmvn_stats are kaldi::Matrix and not a std::string with a  filename
- this allows to assign the matrix directly in C++ without loading it from a file in OS file system

- this change is handy for kaldi integrators, when all the model/config files are loaded from a package,
- but backward compatibility might be an issue for those who worked directly with 'OnlineNnet2FeaturePipelineInfo'...

### DIFF
--- a/src/cudafeat/online-batched-feature-pipeline-cuda.cc
+++ b/src/cudafeat/online-batched-feature-pipeline-cuda.cc
@@ -65,12 +65,10 @@ OnlineBatchedFeaturePipelineCuda::OnlineBatchedFeaturePipelineCuda(
   }
 
   if (info_.use_cmvn) {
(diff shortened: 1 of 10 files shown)


danpovey:
Looks OK to me; @nshmyrev will this break anything that you know of?

danpovey:
Karel, it looks OK to me but I want to wait a couple days before merging to
see if e.g. Nickolay or Kirill have comments.

vesis84:
okay np. let's see what they will respond.

nshmyrev:
No problem for me!

danpovey (commented):
LGTM!

PR #4273, opened 2020-09-21
Clapack support

mhu-coder:
Hi,
While porting Kaldi to Web Assembly, we, at Inria, faced the issue of interfacing Kaldi with CLAPACK. To successfully compile Kaldi with CLAPACK, CBLAS and (reference) BLAS, we ended up making some minor modifications the source code of the matrix library.

In this PR, we are submitting the changes we made to some of the build scripts in Kaldi so that it can be built against our version of CLAPACK, which is available at https://gitlab.inria.fr/kaldi.web/clapack-wasm.

What do you think?

PS: We also reorganized and bound the online decoder in `online2bin/online2-tcp-nnet3-decode-faster.cc` with javascript. The end result is running at https://kaldi-web.loria.fr, which runs the online decoder in a web browser. Would you like a separate PR to add these changes?

### DIFF
--- a/src/makefiles/linux_clapack.mk
+++ b/src/makefiles/linux_clapack.mk
@@ -13,6 +13,10 @@ ifndef OPENFSTLIBS
 $(error OPENFSTLIBS not defined.)
 endif
 
+CLAPACKLIBS = $(CLAPACKROOT)/CLAPACK-3.2.1/lapack.a $(CLAPACKROOT)/CLAPACK-3.2.1/libcblaswr.a \
+	      $(CLAPACKROOT)/CBLAS/lib/cblas.a \
+	      $(CLAPACKROOT)/f2c_BLAS-3.8.0/blas.a $(CLAPACKROOT)/libf2c/libf2c.a
+
 CXXFLAGS = -std=c++11 -I.. -isystem $(OPENFSTINC) -O1 $(EXTRA_CXXFLAGS) \
            -Wall -Wno-sign-compare -Wno-unused-local-typedefs \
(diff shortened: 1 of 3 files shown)


danpovey:
Great!  Will merge.  Is there any way you can show us what you've done for the  online decoder?  I'd rather not commit before seeing the scope.

src/makefiles/linux_clapack.mk
```diff
@@ -13,6 +13,10 @@ ifndef OPENFSTLIBS
 $(error OPENFSTLIBS not defined.)
 endif
 
+CLAPACKLIBS = $(CLAPACKROOT)/CLAPACK-3.2.1/lapack.a $(CLAPACKROOT)/CLAPACK-3.2.1/libcblaswr.a \
+	      $(CLAPACKROOT)/CBLAS/lib/cblas.a \
+	      $(CLAPACKROOT)/f2c_BLAS-3.8.0/blas.a $(CLAPACKROOT)/libf2c/libf2c.a
+
```
petronny, line 19: Please remove these hard coded libs.

These will break builds on many linux systems.

If you need to modify the ld flags.

Please use
```
LDFLAGS='/path/to/your/static/libs.a' ./configure
```
or
```
LDFLAGS='-lcblas -llapack' ./configure
```
when compiling the packages.

danpovey: Thanks!
We very rarely compile with CLAPACK.  We'd need to test any changes to the Makefiles.  If you have time to do this and make a PR then great, but this isn't a high priority right now.

PR #4276, opened 2020-09-22
Online2 tcp web assembly

mhu-coder:
Hi,
To cross-compile the online decoder in `online2-tcp-nnet3-decode-faster` to Web Assembly (WASM), I extracted the online decoder from the main function and put it in a class.

Then I added a binding to Javascript, encapsulated in a `#ifndef __EMSCRIPTEN__`preprocessor directive. The `__EMSCRIPTEN__` flag is set automatically by the WASM compiler I am using (emscripten). Using the resulting binaries with javascript require additional steps that are really javascript though...

I also made some cosmetic changes to comply with Google cpplint's rules although I am not too sure about how to solve the line length issues at lines 5, 613 and 686.

What do you think?


danpovey:
Cool!
If we were to merge this we'd probably want the program named something different, to avoid
possibly breaking  the current program.  Also maybe leave it not compiled by default, to head off any
potential compilation failures due to missing dependencies.
But let me wait and see what comments others may have.

sikoried:
Just some nit picking: There's a whole lot of editor-induced code changes (mostly line wraps or rearranged comments), which should be reverted (so that the PR is limited to the actual code changes/additions).

mhu-coder:
Hi, thanks for the feedback!
I reverted all the changes made on `online2-tcp-nnet3-decode-faster` and put the modifications in a new file, named `online2-tcp-nnet3-decode-faster-reorganized`, which is not listed in the Makefile.

@sikoried : I also undid the formatting changes but I have 2 complaints from CodeFactor which is asking lines 344 and 426, which are blank, to be removed. However, these blank lines are in `online2-tcp-nnet3-decode-faster`. What would be the best solution?

sikoried:
I guess deleting  blank lines is fine since they don't potentially affect semantics (whereas other edits "look" like changes).

jtrmal:
I wonder if you have a comparison how fast/slow the WASM is compared to the native code? Any ballpark number is fine -- I'm just curious.

mhu-coder:
> I wonder if you have a comparison how fast/slow the WASM is compared to the native code? Any ballpark number is fine -- I'm just curious.

Below, a table of the processing time in WASM and native (both linked against CLAPACK, compiled without the `-msse` nor the `-msse2` flag).

|Input audio duration (s)| WASM | Native |
|:--------------------:|:----------:|:---------:|
| 8 | 1.5 | 1.0 |
| 312 | 55 | 40 |

These numbers are not very scientifically computed though. I've just loaded a model and timed the processing of an audio file on my laptop.

jtrmal:
Yeah I understand, but it's pretty impressive IMO.
Y.

jtrmal:
would it be possible to name the file _emscripten instead of _reorganized?

jtrmal:
also I wonder if there should be a specific target or "if" condition in the make files (or both)

mhu-coder:
> would it be possible to name the file _emscripten instead of _reorganized?

Sure.

> also I wonder if there should be a specific target or "if" condition in the make files (or both)

I can export `TARGET_ARCH` from `src/configure` to `src/kaldi.mk` and add an "if" statement to `src/online2bin/Makefile` to have `online2-tcp-nnet3-decode-faster-emscripten.cc` compiled only when `TARGET_ARCH` is set to `WASM`. Would that work?

jtrmal:
yeah, that seems reasonable choice to me
y.

mhu-coder:
Hi, I made the modifications but had to export `TARGET_ARCH` from `src/configure` as `ARCH` as the variable `TARGET_ARCH` is already used in `LINK.o` in `src/makefiles/default_rules.mk` although I am not sure where it is exported from.

I also removed the `pthread` flag when compiling to WASM and using CLAPACK as pthreads are not widely supported by web browsers yet (see [emscripten doc](https://emscripten.org/docs/porting/pthreads.html) and also because the build to WASM currently fails when the flag is set.

With these changes, the build of Kaldi to WASM is successful on my machine. However, there are additional steps to build to WASM from scratch. More specifically:
* openfst needs do be compiled to WASM. This step requires one `make` command and to disable the availability check of zlib in `tools/extras/check_dependencies.sh` :confounded: 
* CLAPACK compiled to WASM. I am not sure if ATLAS or OpenBLAS would compile as easily/without changes.

Do you think the current support to build to WASM is sufficient or would you also include the build of openfst and clapack to WASM as well?

jtrmal:
I'll have to think about it...  No idea ATM
@danpovey? @kkm?

danpovey:
Kirill is better at these compilation issues than me, hopefully he can help
with that.

jtrmal:
I'd propose this -- extend the kaldi root README.md with description how
the other stuff should be compiled.
Please do not forget about some example :)
y.

mhu-coder:
Thanks for the suggestion! As the description is rather lengthy, I added a link to another document at the end of the README.md. Please feel free to comment and provide feedback on this other document :).

jtrmal:
thanks -- I was hoping someone else will chime in on this.
LGTM

danpovey:
Let's keep this open.   @kkm000 if you can comment at some point it would be great.   IDK whether merging will cause build problems?

danpovey:
Merging this as it looks like it won't break anything.

kkm000:
Sorry I'm late to the party, but LGTM.

check_dependencies.sh wants way too much stuff. Why does everyone need to install svn? Does it still even exist? :) If only a few recipes use it, should not be a common dependency. 

I'd rather demote the failure of check_depndencies from error to warning.

Issue #4220, opened 2020-08-13
VoxCeleb1 dataset folder structure not compatible with SITW recipe


mmmmayi:
Hello Daniel,

I tried to run [SITW/v2](https://github.com/kaldi-asr/kaldi/tree/master/egs/sitw/v2) recipt with the latest VoxCeleb dataset. However, there is a bug error reported when I run run.sh:
`Cannot open directory: No such file or directory at local/make_voxceleb1.pl line 56.
` 
I think it happened because of line 40 in run.sh, and seemsly a path of 'VoxCeleb1/voxceleb1_wav' should exists, but I only have  'VoxCeleb1/dev' and 'VoxCeleb1/test'. I think this because of updating of VoxCeleb. Could you please help me to update the script? Thank you

danpovey:
I dunno- I know chnages in voxceleb data have been an issue but dont know
if your data is the old or new one, or if this recipe is supposed to be up
to date.

mmmmayi:
I used the latest version. It works well with make_voxceleb1_v2.pl in voxceleb/v2

danpovey:
It would be great if you could figure out how to resolve the issue using
that other data-prep script and make a PR so it works for the current
voxceleb but can be made to work for the older release via a commented-out
command in the run.sh.

mmmmayi:
I wrote a make_voxceleb1_v2.pl for sitw according to egs/voxceleb:
```
if (@ARGV != 2) {
  print STDERR "Usage: $0 <path-to-voxceleb1> <path-to-data-dir>\n";
  print STDERR "e.g. $0 /export/voxceleb1 data/\n";
  exit(1);
}

($data_base, $out_dir) = @ARGV;
my $out_dir = "$out_dir/voxceleb1";

if (system("mkdir -p $out_dir") != 0) {
  die "Error making directory $out_dir";
}

# This file provides the list of speakers that overlap between SITW and VoxCeleb1.
if (! -e "$out_dir/voxceleb1_sitw_overlap.txt") {
  system("wget -O $out_dir/voxceleb1_sitw_overlap.txt http://www.openslr.org/resources/49/voxceleb1_sitw_overlap.txt");
}

if (! -e "$data_base/vox1_meta.csv") {
  system("wget -O $data_base/vox1_meta.csv http://www.openslr.org/resources/49/vox1_meta.csv");
}

# sitw_overlap contains the list of speakers that also exist in our evaluation set, SITW.
[...]

PR #4287, opened 2020-10-07
edit shell for sitw database

mmmmayi:
As mentioned [here](https://github.com/kaldi-asr/kaldi/issues/4220), I wrote a make_voxceleb1_v2.pl for egs/sitw/v2 according to egs/voxceleb


danpovey:
Thanks!  Looks OK so I'll merge and if and when someone finds a problem with it, we'll address it then.

PR #4290, opened 2020-10-09
Updates for smooth chain2

nshmyrev:
Small changes to make chain2 in mini_librispeech smooth:

1. Always recreate a tree, no need to exit
2. Leave xconfig stage stanalone
3. Do not rely on init/final.mdl which is created after get_raw_egs on stage 22, use trans_mdl directly instead. Fixes situation when trans_mdl is already updated but final.mdl not yet updated on script rerun
4. Quote opts so they are not treated as test options
5. Assume we can have just 1 job
6. Do not run diagnostics on background, creates issues with gpu memory sharing on single machine.

### DIFF
--- a/egs/wsj/s5/steps/nnet3/chain2/compute_preconditioning_matrix.sh
+++ b/egs/wsj/s5/steps/nnet3/chain2/compute_preconditioning_matrix.sh
@@ -76,7 +76,7 @@ if [ $stage -le 2 ]; then
         $ldafolder/lda.mat $ldafolder/lda_stats || exit 1
 
     rm $ldafolder/lda_stats
-    ln -rfs $ldafolder/lda.mat $ldafolder/configs/lda.mat
+    ln -sf $ldafolder/lda.mat $ldafolder/configs/lda.mat
 fi
 
 echo "$0: Finished computing LDA transform"
--- a/egs/wsj/s5/steps/nnet3/chain2/compute_preconditioning_matrix.sh
+++ b/egs/wsj/s5/steps/nnet3/chain2/compute_preconditioning_matrix.sh
@@ -76,7 +76,7 @@ if [ $stage -le 2 ]; then
         $ldafolder/lda.mat $ldafolder/lda_stats || exit 1
 
     rm $ldafolder/lda_stats
-    ln -rs $ldafolder/lda.mat $ldafolder/configs/lda.mat
+    ln -rfs $ldafolder/lda.mat $ldafolder/configs/lda.mat
 fi
 
 echo "$0: Finished computing LDA transform"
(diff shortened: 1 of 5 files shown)


danpovey:
Thanks a lot!!  Hoping for more contributions like this from you in future!

egs/wsj/s5/steps/nnet3/chain2/get_raw_egs.sh
```diff
@@ -130,8 +130,8 @@ latdir=$3
 dir=$4
 
 tree=$chaindir/${lang}.tree
-trans_mdl=$chaindir/init/${lang}.mdl  # contains the transition model and a nnet, but
-                                   # we won't be making use of the nnet part.
+trans_mdl=$chaindir/init/${lang}_trans.mdl  # contains the transition model and a nnet, but
```
danpovey, line 133: is the comment still accurate?

egs/wsj/s5/steps/nnet3/chain2/train.sh
```diff
@@ -190,7 +190,7 @@ while [ $x -lt $num_iters ]; do
             --print-interval=10  \
            "nnet3-copy --learning-rate=$lrate $dir/${x}.raw - |" $den_fst_dir \
            "ark:nnet3-chain-copy-egs $egs_opts scp:$egs_dir/${name}_subset.scp ark:- | nnet3-chain-merge-egs $multilingual_eg_opts --minibatch-size=1:64 ark:- ark:-|" \
-           $dir/${next_x}_${name}.mdl || touch $dir/.error_diagnostic &
+           $dir/${next_x}_${name}.mdl || touch $dir/.error_diagnostic
```
danpovey, line 193: Not 100% ideal from speed perspective - perhaps make it an option?  [could do it as '&' and then just wait if the option is a certain way..] Open to making it like this by default.  Dont know how fast it is...

egs/wsj/s5/steps/nnet3/chain2/get_raw_egs.sh
nshmyrev: Ok, I removed this since it doesn't have any sense now

egs/wsj/s5/steps/nnet3/chain2/train.sh
nshmyrev: Yes, now should be better

nshmyrev:
> Hoping for more contributions like this from you in future!

Thank you, I hope that too as well!

egs/wsj/s5/steps/nnet3/chain2/compute_preconditioning_matrix.sh
```diff
@@ -76,7 +76,7 @@ if [ $stage -le 2 ]; then
         $ldafolder/lda.mat $ldafolder/lda_stats || exit 1
 
     rm $ldafolder/lda_stats
-    ln -rs $ldafolder/lda.mat $ldafolder/configs/lda.mat
+    ln -rfs $ldafolder/lda.mat $ldafolder/configs/lda.mat
```
danpovey, line 79: can we remove the pre-existing option 'r'?   It isn't present on mac.

nshmyrev: Done

PR #4296, opened 2020-10-16
enable linking from both libs, the heuristic malfunctioned sometimes

jtrmal:
I don't see any reason why adding both library directories should be problematic... and fixes mine and Kirill's problem with CI when the static libs are removed


kkm000:
The exact problem was that not everyone had all CUDA libraries installed. So the test for either .a or .so alone was not enough, and, come think of it, not even required; ld will just ignore missing libs.

Validated with CUDA in lib64.

PR #4324, opened 2020-11-08
fix scripts/rnnlm/prepare_rnnlm_dir.sh

LvHang:
rnnlm/get_unigram_counts.sh is absent.
Actually, we use rnnlm/ensure_counts_present.sh to generate the '.counts' files, so replace corresponding commands.


kkm000:
LGTM

PR #4334, opened 2020-11-13
[scripts] Fix to support <end-time> of segments be -1 in utils/validate_data_dir.sh

Jarvan-Wang:
extract-segments support <end-time> be -1, which means the segment runs till the end of the WAV file
but utils/validate_data_dir.sh does not support that.

### DIFF
--- a/egs/wsj/s5/utils/validate_data_dir.sh
+++ b/egs/wsj/s5/utils/validate_data_dir.sh
@@ -179,7 +179,7 @@ if [ -f $data/wav.scp ]; then
     check_sorted_and_uniq $data/segments
     # We have a segments file -> interpret wav file as "recording-ids" not utterance-ids.
     ! cat $data/segments | \
-      awk '{if (NF != 4 || $4 <= $3) { print "Bad line in segments file", $0; exit(1); }}' && \
+      awk '{if (NF != 4 || ( $4 > 0 && $4 <= $3 ) ) { print "Bad line in segments file", $0; exit(1); }}' && \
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/utils/validate_data_dir.sh
```diff
@@ -179,7 +179,7 @@ if [ -f $data/wav.scp ]; then
     check_sorted_and_uniq $data/segments
     # We have a segments file -> interpret wav file as "recording-ids" not utterance-ids.
     ! cat $data/segments | \
-      awk '{if (NF != 4 || $4 <= $3) { print "Bad line in segments file", $0; exit(1); }}' && \
+      awk '{if (NF != 4 || ( $4 > 0 && $4 <= $3 ) ) { print "Bad line in segments file", $0; exit(1); }}' && \
```
kkm000, line 182: Just a suggestion: awkly use of awk gets rid of the if and a couple of pairs of braces.
```suggestion
      awk 'NF != 4 || ($4 > 0 && $4 <= $3) {
             print "Bad line", NR, "in segments file:", $0; exit(1); }' && \
```

kkm000 (commented):
LGTM, suggestion optional.

jtrmal:
LGTM

danpovey:
Actually there's a reason we don't support that.  There are several scripts that it causes problems for.
If you want to add segments to data that didn't previously have it you should call utils/data/get_segments_for_data.sh.
That should automatically be called if you use combine_data.sh on data some of which has segments and some of which doesn't.

PR #4335, opened 2020-11-14
Fix the backoff logic bug in GetProbForWord

LvHang:
Fix the bug in GetProbForWord.
When we construct the 'backoff' history states, we erase the 'head' word from the history word sequence gradually. So the function argument--"states" is arranged from the high order to low order. So we should iterate it from the end.
I use a 4-gram case as the example, as the 3-gram (normal case) only push one element into the 'states' argument so it will not trigger any problem. 
Say we are dealing with ‘(w1, w2, w3, w4)’，the history of it is '(w1, w2, w3)'. There are two backoff history states will be pushed into the 'states' variable. states[0] = (w2, w3), state[1] = (w3).
In original code, we will compute "ans = P(w4) * bow(w2, w3) + P(w4|w2, w3) * bow(w3) + P(w4|w3)".
I think the correct one should be "ans = P(w4) * bow (w3) + P(w4|w3) * bow(w2, w3) + P(w4|w2, w3)" as the new code.

### DIFF
--- a/src/rnnlm/sampling-lm-estimate.cc
+++ b/src/rnnlm/sampling-lm-estimate.cc
@@ -539,7 +539,7 @@ BaseFloat SamplingLmEstimator::GetProbForWord(
   // compute the probability from lowest to highest order.
   KALDI_ASSERT(word > 0 && word < static_cast<int32>(unigram_probs_.size()));
   BaseFloat ans = unigram_probs_[word];
-  for (size_t i = 0; i < states.size(); i++) {
+  for (size_t i = states.size() - 1; i >= 0; i--) {
     const HistoryState *state = states[i];
     ans *= state->backoff_count / state->total_count;
     Count c;
--- a/src/rnnlm/sampling-lm-estimate.cc
+++ b/src/rnnlm/sampling-lm-estimate.cc
@@ -539,7 +539,7 @@ BaseFloat SamplingLmEstimator::GetProbForWord(
   // compute the probability from lowest to highest order.
   KALDI_ASSERT(word > 0 && word < static_cast<int32>(unigram_probs_.size()));
   BaseFloat ans = unigram_probs_[word];
-  for (int32 i = states.size() - 1; i >= 0; i--) {
+  for (int32 i = states.size() - 1; i >= 0; --i) {
     const HistoryState *state = states[i];
     ans *= state->backoff_count / state->total_count;
     Count c;
--- a/src/rnnlm/sampling-lm-estimate.cc
+++ b/src/rnnlm/sampling-lm-estimate.cc
@@ -539,7 +539,7 @@ BaseFloat SamplingLmEstimator::GetProbForWord(
   // compute the probability from lowest to highest order.
   KALDI_ASSERT(word > 0 && word < static_cast<int32>(unigram_probs_.size()));
   BaseFloat ans = unigram_probs_[word];
-  for (size_t i = states.size() - 1; i >= 0; i--) {
+  for (int32 i = states.size() - 1; i >= 0; i--) {
     const HistoryState *state = states[i];
     ans *= state->backoff_count / state->total_count;
     Count c;


danpovey:
Can you see how it affects convergence of the RNNLM training?

LvHang:
Sure, I will use a 4-gram model on librispeech to compare them.

src/rnnlm/sampling-lm-estimate.cc
```diff
@@ -539,7 +539,7 @@ BaseFloat SamplingLmEstimator::GetProbForWord(
   // compute the probability from lowest to highest order.
   KALDI_ASSERT(word > 0 && word < static_cast<int32>(unigram_probs_.size()));
   BaseFloat ans = unigram_probs_[word];
-  for (size_t i = 0; i < states.size(); i++) {
+  for (int32 i = states.size() - 1; i >= 0; i--) {
```
kkm000, line 542: nit: While you're at it, our coding style requires that loops only use pre-{increment, decrement} form,  {`++i`, `--i`} resp. BTW, this is another idiomatic C[++] reverse array iteration, up to you:
```suggestion
  for (int32 i = states.size(); --i >= 0;) {
```

kkm000 (changes requested):

danpovey:
Don't think I like `for (int32 i = states.size(); --i >= 0;) `... kind of
makes a mockery of the `for` syntax, IMO. Pre-decrement is OK with me...

LvHang:
I tried to compare the convergence situation between the original code and the new code, where I generate a 4-gram sampling LM. In general, I think there is no difference. 
image
I attached the "rnnlm_objective vs iterations" graph. The "rnnlm_lstm_1b" corresponds to the original code and "rnnlm_lstm_1c" corresponds the new one. The lines are alomost coincident.

And I use the two models to rescore a chain model. The 'wer's are twisted as follows.
%WER 7.79 [ 4077 / 52343, 553 ins, 352 del, 3172 sub ] exp/chain_cleaned/tdnn_1d_sp/decode_test_other_tglarge_rnnlm_1b_nbest_rescore/wer_14_0.0
%WER 7.56 [ 3957 / 52343, 455 ins, 406 del, 3096 sub ] exp/chain_cleaned/tdnn_1d_sp/decode_test_other_tglarge_rnnlm_1b_rescore/wer_13_0.5
%WER 7.82 [ 4094 / 52343, 556 ins, 337 del, 3201 sub ] exp/chain_cleaned/tdnn_1d_sp/decode_test_other_tglarge_rnnlm_1c_nbest_rescore/wer_13_0.0
%WER 7.52 [ 3936 / 52343, 428 ins, 447 del, 3061 sub ] exp/chain_cleaned/tdnn_1d_sp/decode_test_other_tglarge_rnnlm_1c_rescore/wer_14_0.5

There is an rag at the last few iterations on ‘lstm_nonlin’ figures that I cannot explain. It seems the original one is stable or this phenomenon is caused by occasional sampling cases.
image

LvHang:
In addition,I doubt the commands in "rnnlm/train_rnnlm.sh" may not very well.

```$cmd $queue_gpu_opt $queue_thread_opt $dir/log/train.$x.$n.log rnnlm-train.....```
In the default setup, it will become ```utils/queue.pl --gpu 1 --num-threads 6 $dir/log/train.$x.$n.log rnnlm-train | ```

Actually, The code want to generate the sampling egs in parallel with multi-threads and then pipe the data to 'rnnlm-train' program to train the model. But, in a sge grid, I think the above command tries to ask 6 gpus but only 1 is used as the "smp" environment. But I haven't come up with a good solution to deal with it.

danpovey:
That depends on the queue configuration.  If the gpu resource is per-job (something like qconf -c and change a certain value to JOB), then it will work correctly.
I'll merge the change.

LvHang:
I will try to find out how to make it working correctly. Currently, when my labmates want to submit a multi-gpus pytorch program, they use "--gpu 1 --smp N". I think they may can use a right way "--gpu N" after the change. I still need to learn more. Anyway, thanks a lot.

Issue #4311, opened 2020-10-28
Sporadic timeouts when downloading sctk tools


markus-franke:
When building Kaldi we sometimes get download timeouts for SCTK in the tools folder.

```
Unable to establish SSL connection.
make: *** [Makefile:124: sctk-2.4.10-20151007-1312Z.tar.bz2] Error 4
make: *** Waiting for unfinished jobs....
Unable to establish SSL connection.
```

The current timeout is set to 10 seconds (see snippet below).

**tools/Makefile:124**
```
sctk-$(SCTK_VERSION).tar.bz2:
	if [ -d "$(DOWNLOAD_DIR)" ]; then \
		cp -p "$(DOWNLOAD_DIR)/sctk-$(SCTK_VERSION).tar.bz2" .; \
	else \
		$(WGET) -T 10 https://www.openslr.org/resources/4/sctk-$(SCTK_VERSION).tar.bz2; \
	fi
```

Increasing the timeout to e.g. 20 seconds leads to a successful download and build afterwards.

Wouldn't it make sense to increase those timeouts in general to avoid sporadic timeouts for slow endpoints?

slckl:
I'm also seeing kaldi builds fail because of this, but I'm not sure if 20 seconds will cut it. 
It seems the openslr resources get hammered more at certain times of day, which makes Kaldi builds unpleasantly flaky, particularly because it doesn't fail the whole build as soon as this download fails, at least with -j4.

danpovey:
I am tuning the apache config on openslr to hopefully make it more stable.

kkm000:
They're officially on GitHub now, so we can pull the tarball from there.

kkm000:
I'm changing the makefile to pull SCTK from GitHub.

PR #4336, opened 2020-11-15
[tools] Prefer tarballs from GitHub for reliability

kkm000:
* Software distributed on GitHub is downloaded from GitHub:
  - cub (official): rework rules to avoid excessive re-downloads, and change tarball format to .tar.gz to match other software.
  - SCTK (official)
  - sph2pipe (mirror at burrmill, stable)

* Fix build of `sph2pipe`. Some tarballs of the very same version 2.5 did not accept the patch from `extras/` cleanly, and `make -j` _N_ for large _N_ stopped on a prompt from patch(1) that has scrolled far up, creating an impression that make had hung. The currently used mirror does not require a patch, and cleanly builds with its own Makefile. Rework build rules to use the Makefile, and remove the patch.

* Retire `openblas` rule: it no longer invokes `install_openblas.sh`, and directs the user to use the script instead. The rule was incorrect because it ignored `OPENBLAS_VERSION` manifestly set in the Makefile

* Remove a patch for SCTK: its only purpose had been to add missing #include's and declarations, which generated warnings in C code. Instead, suppress all compilation warnings during build.

Fixes: #4311
Fixes: #4323


danpovey:
OK with me!  You can merge if you think it's OK..

jtrmal:
LGTM

PR #4337, opened 2020-11-15
[tools] Sanitize extra scripts and gitignores

kkm000:
* Fork off `tools/.gitignore` from the root file.
* Make gitignores sensible by using wildcards. Listing every possible versioned package name had been unmaintainable.
* Make all extra scripts executable, and trim EOL/EOF whitespace.


danpovey:
Is the empty file install_nkf.sh new?

kkm000:
Should not be. Something is wrong. Let me figure out.

kkm000:
The GitHub's message "empty file" is nonsense, it's an empty diff, not empty file. Three script files did not have the executable bit set. Of these, tools/extras/install_pfile_utils.sh had other changes, and is shown normally on GitHub. The other two, install_{nkf,liblbfgs}.sh, where the only change is their mode bits, have the message "empty file". The diff of this file correctly shows only a mode change.

I'll double-check after merge that they won't become empty :)

```
~/work/kaldi[2011-sanitize-tools-extra]$ git show -- tools/extras/install_nkf.sh
commit e4c24a1cd857e9a29359c53b5c246ec8049ee2f4
Author: kkm <kkm@pobox.com>
Date:   Sat Nov 14 17:40:21 2020 -0800

    [tools] Sanitize extra scripts and gitignores

    * Fork off tools/.gitignore from the root file.
    * Make gitignores sensible by using wildcards. Listing every
      possible versioned package name had been unmaintainable.
    * Make all extra scripts executable, and trim EOL/EOF whitespace.

diff --git a/tools/extras/install_nkf.sh b/tools/extras/install_nkf.sh
old mode 100644
new mode 100755
```

danpovey:
Ah, OK.  LGTM, then.

kkm000:
Thanks, and both files are indeed non-empty after merge! :)

PR #4341, opened 2020-11-18
fix compilation failure on  std::atomic_int32_t with gcc/++<7.0

freewym:
The error msg is:

../cudadecoder/cuda-online-pipeline-dynamic-batcher.h:64:8: error: ‘atomic_int32_t’ in namespace ‘std’ does not name a type
   std::atomic_int32_t n_chunks_in_queue_;  // equals to chunks.size(), lock free

I see it was recently introduced in https://github.com/kaldi-asr/kaldi/commit/cc79575e2a7e028728a6f5b444ede9214e13881c.

It is suggested from the Internet to use std::atomic<std::uint32_t> instead. I tried and the compilation was passed.


danpovey:
FYI, @hugovbraun ... merging for now though.

hugovbraun:
Thanks for fixing this - according to https://en.cppreference.com/w/cpp/atomic/atomic those two are direct aliases, so the fix works with me.

PR #4342, opened 2020-11-18
chain2 small fix: no groups-per-minibatch option

LvHang:
In normal local/chain2/*.sh, the "groups-per-minibatch" option is passed to steps/chain2/train.sh. But the train.sh doesn't have corresponding option. I found it is same as "minibatch" which is used in command "nnet3-chain-merge-egs". So replace the "minibatch" with "groups_per_minibatch"

### DIFF
--- a/egs/babel/s5d/local/chain2/run_tdnn.sh
+++ b/egs/babel/s5d/local/chain2/run_tdnn.sh
@@ -297,7 +297,7 @@ if [ $stage -le 18 ]; then
     --initial-effective-lrate $initial_effective_lrate \
     --final-effective-lrate $final_effective_lrate \
     --max-param-change $max_param_change \
-    --groups-per-minibatch 128 \
+    --minibatch-size 128 \
     --l2-regularize 0.00005 \
     --num-jobs-initial $num_jobs_initial --num-jobs-final $num_jobs_final \
      $common_egs_dir $dir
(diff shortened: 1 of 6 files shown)


danpovey:
@mrsrikanth can you please review?

mrsrikanth:
@LvHang  Thanks for pointing this out. I think train.sh should have minibatch_size as it is. It should be local/chain2/*.sh that should always pass minibatch_size.  I see that 4 scripts are doing that should be changed:

```
wsj/s5/local/chain2/tuning/run_tdnn_1i.sh
swbd/s5c/local/chain2/tuning/run_tdnn_7k.sh
mini_librispeech/s5/local/chain2/tuning/run_tdnn_1a.sh
babel_multilang/s5/local/chain2/run_tdnn.sh
babel/s5d/local/chain2/run_tdnn.sh
```

Is that something you can modify instead?

LvHang:
@mrsrikanth , so I think you mean change the "groups-per-minibatch" in local/chain2/*.sh to "minibatch_size" rather than change "minibatch_size" in steps/nnet3/chain2/train.sh. I did the latter as we use 'num_chunk_per_minibatch' in traditional chain script, so I felt "groups-per-minibatch" is not a too bad name.
Anyway, I have changed the 5 scripts in corresponding local/chain2/ and recover the steps/nnet3/chain2/train.sh.

Issue #4262, opened 2020-09-08
configure / make fail to find MKL on Ubuntu 20.04


4gra:
Steps to reproduce:

```
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
[...]
$ sudo apt install intel-mkl-full gfortran libsox-dev [...etc]
$ git clone https://github.com/kaldi-asr/kaldi.git
$ ls -l /usr/lib/x86_64-linux-gnu/libmkl_*
$ ls -l /usr/include/mkl/
  (i.e. they exist)
$ cd tools
$ extras/check_dependencies.sh
./extras/check_dependencies.sh: Intel MKL is not installed. Run extras/install_mkl.sh to install it.
 [...]
```

I realised the approval of check_dependencies wasn't required to actually make it work, so I just went ahead and made the tools, which worked fine.  But then MKL hit a snag again on building kaldi proper:

```
$ cd ../src
$ ./configure --shared
    (doesn't work)
$ ./configure --shared --mkl-libdir=/usr/lib/x86_64-linux-gnu
   (also doesn't work)
$ CXXFLAGS="-I/usr/include/mkl/" ./configure --shared --mkl-libdir=/usr/lib/x86_64-linux-gnu
MKL configured libs: -L/usr/lib/x86_64-linux-gnu -Wl,-rpath=/usr/lib/x86_64-linux-gnu -lmkl_intel_lp64 -lmkl_core -lmkl_sequential -ldl -lpthread -lm
MKL include directory: -I/usr/lib/x86_64-linux-gnu/../../include
Using Intel MKL as the linear algebra library.
mkl-test.cc:21:10: fatal error: mkl.h: No such file or directory
   21 | #include <mkl.h>
      |          ^~~~~~~
compilation terminated.
***configure failed: Cannot validate the MKL switches ***
```

so, I patched `./configure`:

```diff
diff --git a/src/configure b/src/configure
index e0ac02115..da1253e26 100755
--- a/src/configure
+++ b/src/configure
@@ -266,6 +266,8 @@ configure_mkl_includes() {
   local mklroot=$1 mkllibdir=$2
   if [[ -d $mklroot/include ]]; then
     echo "-I$1/include"
+  elif [[ -d /usr/include/mkl ]]; then
+    echo "-I/usr/include/mkl"
   elif [[ -d $mkllibdir/../../include ]]; then
     echo "-I$mkllibdir/../../include"
[...]

PR #4350, opened 2020-11-23
[configure] Recognize MKL installed from Ubuntu package (#4262)

kkm000:
Ubuntu 20.04+ packages an unknown 20.x MKL version. The installation in /opt/intel is still preferred, but if none were found or supplied by the user, the package is probed as well, last in the order of probing.

All 8 combination of ({static,dynamic} Kaldi) × ({static,dynamic} MKL) × ({Ubuntu-packaged,normal /opt/intel} MKL) built successfully and passed the matrix/ lib tests.

Fixes: #4262

### DIFF
--- a/src/makefiles/linux_x86_64_mkl.mk
+++ b/src/makefiles/linux_x86_64_mkl.mk
@@ -1,13 +1,8 @@
 # MKL specific Linux configuration
 
-# We have tested Kaldi with MKL version 10.2 on Linux/GCC and Intel(R) 64
-# architecture (also referred to as x86_64) with LP64 interface layer.
-
 # The linking flags for MKL will be very different depending on the OS,
-# architecture, compiler, etc. used. The correct flags can be obtained from
+# architecture, compiler, etc. used. In case configure did not cut it, use
 # http://software.intel.com/en-us/articles/intel-mkl-link-line-advisor/
-# Use the options obtained from this website to manually configure for other
-# platforms using MKL.
 
 ifndef DEBUG_LEVEL
 $(error DEBUG_LEVEL not defined.)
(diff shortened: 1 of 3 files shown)


danpovey:
Cool!  Merge these kinds of things as you see fit...

tools/extras/check_dependencies.sh
```diff
@@ -136,18 +136,21 @@ fi
...
+    if ! [[ -f "${MKL_ROOT}/include/mkl.h" ]] &&
+       # Ubuntu 20+ has an MKL package
+       ! pkg-config mkl-dynamic-lp64-seq --exists &>/dev/null; then
+      echo "$0: Intel MKL does not seem to be installed."
+      if [[ $(uname) = Linux ]]; then
+        echo $' ... Run extras/install_mkl.sh to install it. Some distros' \
+             $'(e.g., Ubuntu 20.04) provide\n ... a version of MKL via'    \
+             $'the package manager, but verify that it is up-to-date.'
       else
-        echo "$0: Intel MKL is not installed. Download the installer package for your
```
jtrmal, line 149: I'd propose to leave the part ""$0: Intel MKL is not installed. " in (to make the script outputs consistent).

src/makefiles/linux_x86_64_mkl.mk
```diff
@@ -53,5 +42,8 @@ ifeq ($(findstring clang,$(COMPILER)),clang)
 CXXFLAGS += -Wno-mismatched-tags
 endif
 
-LDFLAGS = $(EXTRA_LDFLAGS) $(OPENFSTLDFLAGS) -rdynamic
-LDLIBS = $(EXTRA_LDLIBS) $(OPENFSTLIBS) $(MKLFLAGS) -lm -lpthread -ldl
+# As late as possible to allow the user to do what they want.
+CXXFLAGS += $(EXTRA_CXXFLAGS)
```
jtrmal, line 46: the thing with changing might have positive (or negative) impacts in the "active" parameters (IIRC gcc will override the values by the latest one if there are multiple definitions of the same parameters)

jtrmal:
except for those two details which I'm mentioning just to be  assessed, LGTM

tools/extras/check_dependencies.sh
kkm000: I just split the message in two echoes, "$0: Intel MKL is not installed." is printed on its own line a few lines above, out of pure aesthetics, :) The messages were too ragged when printed. I'm trying to wrap all errors no longer than 100, max. 120 characters. Ppl's terminal windows vary in width wildly.

Hint, hint, tangentially, that's what I do, because my terminal windows usually end up too distended:
```
export MANWIDTH=108  MANOPT=--no-hyphenation
```

src/makefiles/linux_x86_64_mkl.mk
kkm000: I know, but this is where we previously violated the (conventional?) semantics of `EXTRA_*FLAGS`  variables. They really should trump everything else. Due to their previous place, I doubt we'd break anyone: override flags had no effect, except to `-O1` which used to came before EXTRA_CXXFLAGS. No one could sensibly put any overrides into this variable in their build scripts anyway. For non-overrides, there is no change.

jtrmal:
yeah, that makes sense. I agree

kkm000:
Cool, I'll merge it then.

src/makefiles/linux_x86_64_mkl.mk
danpovey: FYI, nvcc will just crash if you give it different -O options like -O1 and later -O2.  (I'm not saying this PR would cause that problem, just FYI).

kkm000:
Ugh...

PR #4352, opened 2020-11-24
fixed relative path to openfst in install_opengram.sh

GeorgeTsaplin:
running `./install_opengram.sh` leads to error "configure: error: fst/fst.h header not found"


GeorgeTsaplin:
install_opengram.sh should be running from `tools/` folder instead of `tools/extras`

PR #4367, opened 2020-12-05
Opensat20 large vocabulary

huangruizhe:
In the current opensat20 recipe, we are using CMU dict, which had a vocabulary of 125k words. However, as we hoped to cover more words in KWS and reduce OOVs, we have been using a larger 241k vocabulary for the OpenSAT evaluation. The 241k vocabulary is created from Safe-T and OpenSubtitles corpus. 
The added scripts in this PR reproduce the 241k vocabulary generation process.
@aarora8 @jtrmal


danpovey:
Cool!  Is there anyone over there who can review this?
If not I'm inclined to just merge and fix any problems later on.

huangruizhe:
I think @jtrmal and @aarora8 will double check this PR. Besides, this should be merged after Ashish's opensat20 recipe is merged.

danpovey:
OK.

jtrmal:
I just skimmed through it -- I think it suffers from the same issues as the
previous Ashish's PR -- the AMI and other scripts should be either
symlinked or (if modified), you should be adding your name to authors and
perhaps mention the changes done. Can you please look into it?
y.

huangruizhe:
Thanks! Sure, will do.

PR #4371, opened 2020-12-08
[src] Segmentation in Cuda offline binary

hugovbraun:
Adding the option to do the segmentation directly into the binary. Some users reported that doing the segmentation as a preprocessing step was becoming a bottleneck. Turned on by setting --segmentation=true. Other options can be used to define the segments length and overlap between segments.

Results from different segments of the same utterance are combined in a callback. We're probably provide a default "combine operator" per default, at least for a text output. 

For the lattice output, any ideas on what we should do is welcome: either use some kind of "MergeLattice" op if that already exists, or leave it up to the user. Or probably write each segment lattice as an individual entry in the output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.5)
 project(kaldi)
 
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/src CACHE PATH "Install path prefix." FORCE)
+endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH}")
 include(GNUInstallDirs)
 include(Utils)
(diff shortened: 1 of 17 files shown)


danpovey:
Cool!  I think individual segments might be easiest (maybe encode times in key name?)
If you did want to append them, look for fst::Concat().

hugovbraun:
Added some things: 

1) Support for CTM output at the segment level and at the full utterance level. I wrote a simple "combine" operator that tries to take advantage of a small overlap between segments. If CTM output is used, the binary will use that combine operator and write the full utterance CTM text output to file.

2) Added a "Lattice Postprocessor" module. This is to improve ease of use (no need to write C++ code in callbacks for postprocessing tasks). We can add more tools in the future, for now it supports scale lattice/word insertion penalty/word align. This is a pure C++ file calling other Kaldi modules so it is simple to maintain/improve. 
It is configured through a config file passed by _--lattice-postprocessor-rxfilename_. 

For now the 'WordAlignLattice" step is mandatory, so the "--word-boundary-rxfilename=" option is mandatory in that file. @danpovey, could we skip that step or do we have to do it? After that step, we use a MBR.

As long as word-boundary-rxfilename is set, it is now possible to push long audio files to the binary, and get a single CTM output per utterance written to the output

3) For more advanced users, segments lattice output can be used. Lattices will be written under the key "_[utt_key]-[segment_offset_in_seconds]_". It is up to the user to avoid collisions with existing keys (by using utterance keys of fixed length, for instance).

Switch between "CTM" output and "lattice" output is done by detecting a wspecifier in the output_filename (lattice) or not (ctm).

There's a need for documentation + a good example. Would it make sense to have a script such as ./run_gpu.sh in egs/aspire/s5/ (or else) which deploys the trained model to the GPU?

95% of the cleanup is done, but more testing is necessary.

danpovey:
Yes I think a script would be great.
Aspire setup is OK.. something like mini_librispeech is more accessible,
but anything is OK.

hugovbraun:
@danpovey this is ready. Everything is disabled per default for backward compatibility. I'll add the deploy script with another PR

danpovey:
Thanks!  Can you remove the 'draft' tag?

hugovbraun:
done. thanks!

PR #4375, opened 2020-12-09
Fix a bug of getting result in CudaDecoder

heibaidaolx123:
The last arc in path is not included, leading to incomplete results. @danpovey


danpovey:
Thanks!  @hugovbraun perhaps you could review?

hugovbraun:
Looks good! Thanks @heibaidaolx123 
Btw, if you are using this for streaming, the streaming pipeline now provides an interface to get the best path text output (including for intermediate results). It might interest you.

heibaidaolx123:
Thank you for the nice work of CudaDecoder, @hugovbraun 
I'll check that out.
By the way, is there any plan to add looped AM computaion to CudaDecoder?

PR #4380, opened 2020-12-10
[src] missing init in OnlineIvectorExtractionInfo(), fixes #4379

### LINKED ISSUE #4379: OnlineIvectorExtractionInfo constructor misses initializing online_cmvn_iextractor
The constructor `OnlineIvectorExtractionInfo()` misses setting a member variable `this->online_cmvn_iextractor = false`.  This variable was introduced in ed1d139a8. 

I am aware that the comments say that this constructor should not be used, but we use it anyways—we load the content of all member variables ourselves from code, and not from files specified in a OnlineIvectorExtractionConfig object.  

The fix is trivial, I will submit a PR as soon as this bug report has received a number.

davidavdav:
It took me the better part of the week to find out why our models showed performance regression when our code was linked to more recent kaldi builds.  

I learned a lot about constructors, and finally linking against Kaldi on a mac to demonstrate both the bug and a fix.  See also my comments in #4379 why we implicitly use this constructor That Should Never Be Used.

### DIFF
--- a/src/online2/online-ivector-feature.cc
+++ b/src/online2/online-ivector-feature.cc
@@ -97,7 +97,7 @@ void OnlineIvectorExtractionInfo::Check() const {
 
 // The class constructed in this way should never be used.
 OnlineIvectorExtractionInfo::OnlineIvectorExtractionInfo():
-    ivector_period(0), num_gselect(0), min_post(0.0), posterior_scale(0.0),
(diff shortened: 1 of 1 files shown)


danpovey:
Thanks!!

danpovey:
BTW valgrind would have found the use of uninitialized memory, but IDK if something like that exists for mac.

PR #4382, opened 2020-12-10
Adding TF32 tensor core support

akshaysubr:
This PR adds [TF32](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/) support to `cudamatrix`. TF32 is a compute type with the same dynamic range as single precision and is much more suitable to training, especially in cases where half precision training results in NaN errors due to the lower dynamic range. TF32 support needs CUDA 11+ and Ampere+ GPU architecture, provides up to 8x higher matrix multiply throughput without any significant loss in accuracy. TF32 tensor cores are enabled using the flag `--cuda-use-tf32-compute`.

This PR also updates the use of some deprecated CUDA API for enabling tensor core support.

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -118,12 +118,32 @@ void CuDevice::Initialize() {
 #endif
     
 #if CUDA_VERSION >= 9000 
+#if CUDA_VERSION >= 11000
     if (device_options_.use_tensor_cores) {
       // Enable tensor cores in CUBLAS
       // Note if the device does not support tensor cores this will fall back to normal math mode
-      CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_, 
-            CUBLAS_TENSOR_OP_MATH));
+      // CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_, 
+      //       CUBLAS_TENSOR_OP_MATH));
+      cublas_compute_type_ = CUBLAS_COMPUTE_32F_FAST_16F; 
+      cublas_gemm_algo_ = CUBLAS_GEMM_DEFAULT_TENSOR_OP;
+    } else {
+      // Enable TF32 compute mode in CUBLAS by default for tensor core acceleration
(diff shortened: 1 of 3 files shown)


danpovey:
Cool!  I assume `make test` works with this change?
Is this mainly geared towards training?  I wonder whether we should make the flag default to true.

akshaysubr:
`make test` works with this change. And, yes, this is mainly geared towards training. With FP16 math, some more sensitive routines like QR and Householder reflections can have dynamic range issues unless a more sophisticated mixed precision approach is implemented. With TF32, all data is still in single precision, it is only the compute that is at the lower precision and without any impact to dynamic range. I'm seeing < 1% change in the loss value with and without TF32 for `egs/swbd/s5c`.

I agree that making the flag default to true would be ideal to get out of the box performance. I think only the Cholesky test in `cu-tp-matrix-test` fails with the flag on by default. Is there any particular reason for a custom Cholesky implementation instead of using a library function like [this](https://docs.nvidia.com/cuda/cusolver/index.html#cusolverDnXpotrf)?

danpovey:
Mm.  The cholesky stuff was probably implemented before CUDA lapack existed.  It's unlikely that we are using the tpmatrix version, since we don't really use packed matrices with CUDA.  Let me merge this as-is for now and we can try to get people to test with it set to true in the code by default.  There are many architectures and I'm concerned some of them may be more susceptible to roundoff than others.

danpovey:
https://groups.google.com/d/msgid/kaldi-help/6c6ea934-9a20-4d4a-8ed1-1e76341a8899n%40googlegroups.com.

danpovey:
Guys,
We had some compilation problems with earlier toolkit versions like 9.0 (cublasGemmBatchedEx seems to have toolkit version dependency, although this isn't mentioned in the documentation).  In addition we have some runtime problems with GPUs with compute capability less than 5.0, see this
 https://groups.google.com/d/msgid/kaldi-help/6c6ea934-9a20-4d4a-8ed1-1e76341a8899n%40googlegroups.com.
I think the best thing is that I revert this for now, and we re-merge once the problems are fixed.

PR #4388, opened 2020-12-14
Create average_rnnlm.sh

rickychanhoyin:
script to average rnnlm models at user define model iterations


scripts/rnnlm/average_rnnlm.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+#
+# Created by HY Chan (2020)
```
danpovey, line 3: Seems to me it would be more user-friendly to add a later stage to the RNNLM training script to call this automatically (possibly controlled by an option)?  This script isn't so easy to discover or, once discovered, to understand how to use.

PR #4396, opened 2020-12-18
Revert "[src] Fix up the legacy dlopen-based CUDA driver calls"

danpovey:
Reverts kaldi-asr/kaldi#4387


danpovey:
I reverted this because people were getting linking errors, apparently due to -lcuda not being included on the link line.
I don't fully understand how to fix this issue (e.g. because I can't see where libcuda.so is on the system I have a GPU on) so I'm reverting the PR for now.

achimnol:
Usually libcuda is either under /usr/local/cuda (may be symlinked to a specific cuda version directory) or mounted from the host when using nvidia-docker. I’ll review by myself again and revise the PR to include a proper link option, but if you are able to use nvcc there should be no problems.

PR #4401, opened 2020-12-23
Revert "Adding TF32 tensor core support"

danpovey:
Reverts kaldi-asr/kaldi#4382


danpovey:
See #4382 for discussion

PR #4409, opened 2021-01-07
RATS SAD Recipe

johnjosephmorgan:
This is a recipe that trains and evaluates a Speech Activity Detector on the rats_sad corpus.


danpovey:
Cool!
I hope some people involved in diarization can make comments?

egs/sad_rats/s5/Note.txt
```diff
@@ -0,0 +1,46 @@
...
+1. s5/steps
+This directory is symlinked to ../../wsj/s5/steps
+It contains standard Kaldi scripts that have been optimized over many years.
+2. s5/utils
+This directory is symlinked to:
+../../wsj/s5/utils
+It also contains scripts that have been optimized over many years.
+3. local
+This directory contains scripts that were taken from other Kaldi recipes and modified to work for the rats_sad corpus.
+- s5/local/rconvert_rttm_to_utt2spk_and_segments.py
```
desh2608, line 25: Typo *rconvert -> convert

egs/sad_rats/s5/Note.txt
```diff
@@ -0,0 +1,46 @@
...
+- s5/local/detect_speech_activity.sh
+this script performs SAD with the trained models.
+- s5/local/segmentation/tuning/tdnn_lstm_sad_1a.sh
+This file contains parameter settings and the command used to train a Long Short-Term Memory Neural Network model for SAD.
+the afix 1a denotes that this file represents an experiment that uses the parameter settings given in the file.
+This file is symlinked to:
+- s5/local/segmentation/run_lstm.sh
+
+According to the pattern set in other Kaldi recipes a new experiment with different parameter settings would be written in a file named something lie:
+s5/local/segmentation/tuning/tdnn_lstm_sad._1b.sh
```
desh2608, line 42: Typo: *sad._1b -> sad_1b

egs/sad_rats/s5/Note.txt
```diff
@@ -0,0 +1,46 @@
...
+and symlinked to
+s5/local/segmentation/run_lsmt.sh
```
desh2608, line 44: Typo: *lsmt -> lstm

egs/sad_rats/s5/README.txt
```diff
@@ -0,0 +1,9 @@
+This recipe builds a Speech Activity Detection (SAD)system.
```
desh2608, line 1: Could you combine the contents of this file and `Note.txt`?

egs/sad_rats/s5/RESULTS.txt
```diff
@@ -0,0 +1,121 @@
+command line (run on 2020 Nov 9 at 14:27:18) Version: 22  /home/john/kaldi_jjm/egs/sad_rats/s5/../../../tools/sctk/bin/md-eval.pl -1 -c 0.25 -r data/dev-1/rttm.annotation -s exp/sad_1a/dev-1/sad.rttm
```
desh2608, line 1: * Since the `sctk` path is already added in `path.sh`, perhaps you can replace the absolute path with just `md-eval.pl` so that it is easier for others to understand how to evaluate. 

* Since we only care about the missed speech and false alarm, you can just show those results by piping the output of md-eval to `awk '/(MISSED|FALARM) SPEECH/'`.

* Are you sure it is fair to evaluate with the `-1` and `-c 0.25` arguments. Recall that `-1` only evaluates single speaker regions (a SAD should detect speech even if more than 1 speakers are speaking), and 0.25 is a fairly permissive collar, especially considering that the annotations in RATS SAD corpus are done in 3 passes (including 2 manual passes). I leave this to your better judgement.

egs/sad_rats/s5/local/get_speech_targets.py
```diff
@@ -0,0 +1,153 @@
...
+                        format""")
+
+    args = parser.parse_args()
+
+    if args.frame_shift < 0.0001 or args.frame_shift > 1:
+        raise ValueError("--frame-shift should be in [0.0001, 1]; got {0}"
+                         "".format(args.frame_shift))
+    return args
+
+class Segment:
```
desh2608, line 63: I believe you borrowed this from the `steps/overlap/get_overlap_targets.py`. Since the time I pushed that code, I have realized that it may be more "pythonic" to use a namedtuple here instead of this class. Perhaps something like the following:

```
from collections import namedtuple

Segment = namedtuple('Segment', 'reco spk start dur end')
```

And later on, in the for loop of line 103, you can do the following:

```
start = float(3)
duration = float(4)
end = start + duration
segments.append(Segment(fields[1], fields[7], start, duration, end))
```

This is better because each segment here is an immutable record, and so it is well represented by a namedtuple (which is immutable). We can also do the same through classes by using a `dataclass` as follows:

```
from dataclasses import dataclass

@dataclass(frozen=True)
class Segment:
    recoid: str,
    spkid: str,
    start: float,
    duration: float,
    end: float
```

But I think a `namedtuple` is more convenient here.

egs/sad_rats/s5/local/get_speech_targets.py
```diff
@@ -0,0 +1,153 @@
...
+        self.start_time = start_time
+        if (dur is None):
+            self.end_time = end_time
+            self.dur = end_time - start_time
+        else:
+            self.dur = dur
+            self.end_time = start_time + dur
+        self.label = label
+
+def groupby(iterable, keyfunc):
```
desh2608, line 82: Again, this is something I have realized recently. Since we are only using this function at a single place, we don't really need it. Instead, you can remove this and use `itertools.groupby()` in line 108. Note that `itertools.groupby()` needs a sorted iterable (which is why this function was created in the first place, so instead of simply passing `segments`, you would need to do `sorted(segments, key=lambda x: x.recoid)`.

egs/sad_rats/s5/local/get_transform_probs_mat.py
```diff
@@ -0,0 +1,74 @@
...
+
+    parser.add_argument("--sil-in-speech-weight", type=float,
+                        default=0.0,
+                        help="The fraction of silence probability "
+                        "to add to speech")
+    parser.add_argument("--speech-in-sil-weight", type=float,
+                        default=0.0,
+                        help="The fraction of speech probability "
+                        "to add to silence")
+    parser.add_argument("--garbage-in-speech-weight", type=float,
```
desh2608, line 33: The `garbage` related options can be removed since they are unused.

egs/sad_rats/s5/local/make_mfcc.sh
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env bash
```
desh2608, line 1: Could you mention how this script is different from `steps/make_mfcc.sh` and why it is needed? You can add this description in the comments at the top.

egs/sad_rats/s5/local/prepare_data.py
```diff
@@ -0,0 +1,168 @@
...
+        for line in f.readlines():
+            fields = line.strip().split()
+            segments.append(Segment(fields))
+    return segments
+
+
+def find_audios(wav_path, file_list):
+    # Get all .flac file names from audio directory
+    command = 'find %s -name "*.flac"' % (wav_path)
+    wavs = subprocess.check_output(command, shell=True).decode('utf-8').splitlines()
```
desh2608, line 92: Using `subprocess()` with `shell=True` can be malicious. Can you check if using `glob()` can work here in a reasonable time? (If it's much slower, then it's okay to keep using subprocess)

egs/sad_rats/s5/local/rats_sad_tab_prep.sh
```diff
@@ -0,0 +1,28 @@
...
+echo "Output is written to data/local/annotations/{train,dev-1,dev-2}."
+  exit 1;
+fi
+
+dir=$1
+
+mkdir -p data/local/annotations
+
+for fld in train dev-1 dev-2; do
+  find $dir/$fld/sad -type f -name "*.tab" | xargs cat > \
```
desh2608, line 26: `xargs` behaves slightly differently on Linux and MacOS, when the piped input is empty, but in this case if you have verified that isn't the case, then it should be okay.

egs/sad_rats/s5/run.sh
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env bash
+# Copyright   2021   ARL (Author: John Morgan)
+# Apache 2.0.
+
+# This recipe builds a Speech Activity Detection system on the rats_sad corpus.
+# The LDC identifyer for the rats_sad corpus is LDC2015S02.
```
desh2608, line 6: *identifier

egs/sad_rats/s5/run.sh
```diff
@@ -0,0 +1,123 @@
...
+
+    utils/utt2spk_to_spk2utt.pl data/$fld/utt2spk > data/$fld/spk2utt
+    utils/fix_data_dir.sh data/$fld
+  done
+fi
+
+if [ $stage -le 3 ]; then
+  # Extract MFCCs for training data set
+  echo "$0: Extract features for train data directory."
+  local/make_mfcc.sh --nj $nj --cmd "$train_cmd"  --write-utt2num-frames true \
```
desh2608, line 54: If you are training the SAD on the `whole` data directory (for which you extract MFCCs in stage 5), do you still need to exract the MFCCs here?

egs/sad_rats/s5/run.sh
```diff
@@ -0,0 +1,123 @@
...
+      $dir/$fld/segments \
+    $dir/$fld/sad.rttm
+  done
+fi
+
+if [ $stage -le 10 ]; then
+  # Evaluate SAD output
+  for fld in $test_sets; do
+    echo "$0: evaluating $fld output."
+    md-eval.pl -1 -c 0.25 -r data/$fld/rttm.annotation \
```
desh2608, line 119: Here, you should either evaluate without the `-1 -c 0.25` or make a comment justifying why this is okay.

desh2608:
@johnjosephmorgan I have finished reviewing and have made some comments. Let me know when you are done addressing them.

egs/sad_rats/s5/local/get_speech_targets.py
johnjosephmorgan: I tried changing to namedtuple,but I ran into problems. I'm now reluctant to fix something that is not broken.

egs/sad_rats/s5/local/get_speech_targets.py
johnjosephmorgan: Done

egs/sad_rats/s5/local/get_speech_targets.py
desh2608: Sorry, I made a small error in the code above. It should be: 

```
segments.append(
  Segment(
    reco = fields[1],
    spk = fields[7],
    start = start, 
    dur = duration, 
    end = end
))
```

i.e., you have to specify the key-value pairs when creating the namedtuple object

egs/sad_rats/s5/local/rats_sad_tab_prep.sh
johnjosephmorgan: I did not have problems with xargs in this script on linux. I did not go end to end on MacOSX.

egs/sad_rats/s5/local/get_speech_targets.py
johnjosephmorgan: I made the change you suggested and it looks like the namedtuple data structure works now.

egs/sad_rats/s5/local/make_mfcc.sh
johnjosephmorgan: Since extracting segment-level features is not necessary, I removed this script.

egs/sad_rats/s5/run.sh
johnjosephmorgan: I evaluate without -1 and -c 0. The results are not as god, but I think they are still OK.

egs/sad_rats/s5/run.sh
johnjosephmorgan: The segment-level feature extraction step was unnecessary. I removed that stage.

johnjosephmorgan:
@desh2608 I addressed your comments. . I ran the recipe end to end and I get similar results.

danpovey:
Great!  Merging.  Thanks, jjm; and thanks Desh for reviewing!

PR #4411, opened 2021-01-07
Set 8 frames per chunk by default

nshmyrev:
By default we use 20 frames per chunk which results in tensor size 7 for chain models with subsampling 3. With 24 frames per chunk we get tensors of size 8, which ends in slightly faster matrix multiplication.

### DIFF
--- a/src/nnet3/decodable-simple-looped.h
+++ b/src/nnet3/decodable-simple-looped.h
@@ -54,7 +54,7 @@ struct NnetSimpleLoopedComputationOptions {
   NnetSimpleLoopedComputationOptions():
       extra_left_context_initial(0),
       frame_subsampling_factor(1),
-      frames_per_chunk(20),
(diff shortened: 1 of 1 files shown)


danpovey:
But surely this depends on the amount of context, etc.?
I'm concerned it may cause regressions of latency that may surprise people.

nshmyrev:
> But surely this depends on the amount of context, etc.?

On higher levels on DNN there is no context, just number of frames. The sizes for small TDNN models for sgemm are like 2192 (tied states) x 192 (Factor dimension) x 8 (frames, was 7 before). One can see it with computation.debug.

> I'm concerned it may cause regressions of latency that may surprise people.

Its just 0.04 second difference.

danpovey:
mm, OK, will merge.

nshmyrev:
Thank you!

PR #4415, opened 2021-01-11
Add a few fixes for undefined behaviour

gibiansky:
In building and running Kaldi, I found a few cases of undefined behaviour being triggered. Running with a UB sanitizer, these caused my runs to fail. Two of them have to do with signed plus unsigned integer arithmetic overflowing signed integers, and one of them is memset(NULL, 0). The last change here is an include change to match the style used everywhere else in Kaldi (this also caused a failure in my use of Kaldi).


danpovey:
thanks!

PR #4416, opened 2021-01-12
Adding TF32 tensor core support

akshaysubr:
This PR adds [TF32](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/) support to `cudamatrix`. TF32 is a compute type with the same dynamic range as single precision and is much more suitable to training, especially in cases where half precision training results in NaN errors due to the lower dynamic range. TF32 support needs CUDA 11+ and Ampere+ GPU architecture, provides up to 8x higher matrix multiply throughput without any significant loss in accuracy. TF32 tensor cores are enabled using the flag `--cuda-use-tf32-compute`.

This PR also updates the use of some deprecated CUDA API for enabling tensor core support.

Also fixed CUDA 9 build issues and should also work with older GPU architectures.

### DIFF
--- a/src/cudamatrix/cu-device.cc
+++ b/src/cudamatrix/cu-device.cc
@@ -134,8 +134,10 @@ void CuDevice::Initialize() {
         cublas_gemm_algo_ = CUBLAS_GEMM_DEFAULT_TENSOR_OP;
     }
 #else
-    CUBLAS_SAFE_CALL(cublasSetMathMode(cublas_handle_,
-            CUBLAS_TENSOR_OP_MATH));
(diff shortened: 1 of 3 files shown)


akshaysubr:
@danpovey This should fix the CUDA 9 and SM 3.x issues described in https://github.com/kaldi-asr/kaldi/pull/4382 and https://github.com/kaldi-asr/kaldi/issues/4399

danpovey:
Thanks a lot!!  I'll merge; please check your github mentions in case there are compilation or runtime issues that we need your help with!

PR #4422, opened 2021-01-14
Revision of #4387

achimnol:
This is a revision of #4387 reverted by #4396, with fixes to the Makefile fragments to correctly link the CUDA driver symbols.
Since the NVIDIA driver and CUDA toolkit may be installed separately (e.g., during `docker build`, the container images have the CUDA toolkit only but not the driver), NVIDIA ships a symbol-only stub libraries for the driver to allow compilation of CUDA programs that uses the driver API directly.
Those symbol-only stub libraries are located under `$(CUDATKDIR)/lib64/stubs`, and at runtime they are resolved to actual driver-provided files (e.g., in containers launched via `nvidia-docker` with GPU assignments).

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -39,7 +39,7 @@
 
 # This should be incremented after any significant change to the configure
 # script, i.e. any change affecting kaldi.mk or the build system as a whole.
-CONFIGURE_VERSION=11
+CONFIGURE_VERSION=12
 
 # We support bash version 3.2 (Macs still ship with this version as of 2019)
 # and above.
--- a/src/cudamatrix/cu-allocator.cc
+++ b/src/cudamatrix/cu-allocator.cc
@@ -237,7 +237,7 @@ void* CuMemoryAllocator::MallocFromSubregion(SubRegion *subregion,
   block->t = t_;
   allocated_block_map_[block->begin] = block;
   allocated_memory_ += (block->end - block->begin);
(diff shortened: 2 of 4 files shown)


danpovey:
I'd rather get comments from the NVidia guys like @hugovbraun before merging this, it makes me a little nervous.
One concern is that kaldi.mk is generated from that make file, so if people do `git pull` without reconfiguring, their build may fail.  That can be made more elegant by bumping CONFIGURE_VERSION in `configure` to 12.  But IDK whether this is a recommended way to do things and how system-independent it is.

hugovbraun:
It looks good to me. Considering that this driver API call is within `#if HAVE_CUDA` block, and that @achimnol uses the stubs, it looks clean. stubs library have been available for a while, so no problem there.  

I'm curious how you got it to work without linking cuda in #4387 ? That file is compiled with gcc, not nvcc.

achimnol:
> It looks good to me. Considering that this driver API call is within `#if HAVE_CUDA` block, and that @achimnol uses the stubs, it looks clean. stubs library have been available for a while, so no problem there.
> 
> I'm curious how you got it to work without linking cuda in #4387 ? That file is compiled with gcc, not nvcc.

We were doing custom build based on some specialized Docker setup and there had an external link configuration.
Thanks for confirmation!

PR #4428, opened 2021-01-20
modefy component copy constructor function

datemoon:
modefy DropoutMaskComponent,SpecAugmentTimeMaskComponent,GeneralDropoutComponent copy constructor function.


danpovey:
makes the test_mode_ get copied properly.

Issue #4430, opened 2021-01-22
Why is it by default to augment speaker labels in speed perturbation?


underdogliu:
Hi all! I am now running a recipe with data augmentation via speed perturbation. However even though I used to run it in previous experiments without any doubt, this time I found that it augments speaker labels as well as utterances. i.e. by default number of speakers became 3 times larger as well (by setting 0.9, 1.0, 1.1 as mapping parameter). I am wondering if this is necessary. If there is about on this feature, I can create a PR on disabling speaker augmentation by default. 

@danpovey @hhadian what do you think?

danpovey:
It doesn't make sense to share CMVN stats between different speeds of the
same data, or to treat them as the same speaker for speaker adaptation.

underdogliu:
@danpovey Thanks for the reply Dan! Well the thing is sometimes we use such scripts for something else than speaker adaptation in ASR. And for such "other-than-ASR" applications it may not be necessary to treat diff. speed-perturbed data belonging to pseudo-speakers.

Also I didn't find related "It doesn't make sense...." statements or practical supportive statements in the [original paper from Tom Ko.](https://www.danielpovey.com/files/2015_interspeech_augmentation.pdf) I know it would be too time-consuming to explain this trivial thing but any hint on justification would be helpful. Thanks!

danpovey:
mm, there is a spk–prefix option or similar that you could remove.
it is not a natural transformation that we encounter in the wild.

underdogliu:
yeah true. I am not making it totally different but offering an option actually. Will try some tests and make a PR later (it involves segments but I think it is irrelevant to this issue).

PR #4432, opened 2021-01-22
[WIP] Adding option on augmenting speaker labels for speed perturbation

underdogliu:
This is a PR related to discussion #4430 . I simply add an option in 3-way speed perturbation and single speed perturbation scripts, to decide whether we expand speaker labels or not. This is a rather trivial issue for ASR because we need it for speaker adaptation using I-vector, but for other applications (e.g. speaker verification) where users would like to use speed perturbation as part of data augmentation, this can be cumbersome.

As can be seen in the changes, ideally if we set options by default, nothing will change - augmentation on speaker labels will perform as usual. If we set `$include_spk_prefix` to `false`, we will not augment speakers while augmenting data with prefix `sp-`.

I have ran it on my own mock dataset with Kaldi file structure and it works. **I will test it on more large-scale dataset but just in case (although technically irrelevant), I cannot test it on egs whose Kaldi paths has `segments` due to not owning corresponding resources.**

@danpovey sorry for creating this PR if from our discussion in issue cited above you are actually feeling not good about this change. Again if everything works well, nothing will change for ASR applications. The option is for other stuff. But if it sounds not very persuasive, please make sharp comments.


egs/wsj/s5/utils/data/perturb_data_dir_speed_3way.sh
```diff
@@ -53,8 +59,8 @@ echo "... would be very slow, and you might need them."
 utils/data/get_utt2dur.sh ${srcdir}
 utils/data/get_reco2dur.sh ${srcdir}
 
-utils/data/perturb_data_dir_speed.sh 0.9 ${srcdir} ${destdir}_speed0.9 || exit 1
-utils/data/perturb_data_dir_speed.sh 1.1 ${srcdir} ${destdir}_speed1.1 || exit 1
+utils/data/perturb_data_dir_speed.sh --include-spk-prefix $include_spk_prefix 0.9 ${srcdir} ${destdir}_speed0.9 || exit 1
+utils/data/perturb_data_dir_speed.sh --include-spk-prefix $include_spk_prefix 1.1 ${srcdir} ${destdir}_speed1.1 || exit 1
```
danpovey, line 63: this may be problematic in terms of data sorting.  generally speaking it's safest to always ensure that speaker ids are prefixes of utterance-ids .  It would be better to make the utterance affix a postfix instead of a prefix, in the case where no speaker prefix is used.

underdogliu: @danpovey yes this morning I found such issue with one of the SRE datasets. Will handle it.

underdogliu:
@danpovey changes made and worked for SRE, switchboard and voxceleb. I also added a `utils/fix_data_dir.sh` at the beginning of 3-way perturbation script because sometimes if the files in the source directory are not in the original order, speed perturbation augmentation without speaker augmentation will fail. Hope that doesn't affect anything.

danpovey:
OK.   You've tested it with the default setting?  (Without your option?)

underdogliu:
@danpovey both with and without. But it may be good if somebody who has access to other data can perform tests on their kaldi directories for data as well, especially data with `segments` presented. Again this should be an orthogonal issue to such file but it does not harm to do some sanity checks. Anybody can do it maybe on AMI dataset? I don't have it in place.

danpovey:
I suggest we just merge as-is and fix any problems later.

underdogliu:
@danpovey ok we will see if there is any problem. If there is please ping me.

PR #4434, opened 2021-01-26
threading-fix

ngoel17:
Fix an issue that manifests with ivector-extractor crashing because the thread is not joinable when machines are under very heavy sustained loads


src/util/kaldi-thread.h
```diff
@@ -209,6 +209,9 @@ class TaskSequencer {
   void Wait() { // You call this at the end if it's more convenient
     // than waiting for the destructor.  It waits for all tasks to finish.
     if (thread_list_ != NULL) {
+      while (!thread_list_->thread.joinable()) {
+	sleep(1);
```
danpovey, line 213: can you please indent this 2 spaces?  also below.

ngoel17: done. Sorry about the delay.

PR #4440, opened 2021-01-27
Librispeech data preparation optionally with text

### LINKED ISSUE #4437: Do we want to put small prefix on Librispeech speakers? And making text creation as optional
Hi all! So I am using a mixture of multiple datasets but found a problem when sorting out speakers with LibriSpeech and one of the SRE datasets. I found the reason was LibriSpeech use pure integer IDs for speakers. While for SRE datasets which are not completely free it looks not making too much trouble, I am wondering if we can add a small prefix at beginning of each speaker and utterance, for example `lbi-$spk_id` for speakers, `lbi-$spk_id-$chapter_id` for utterances.

Also, I am wondering if we can make text creation in LibriSpeech as optional since for tasks other than ASR we don't usually need text.

If the answer to both (or one of them) are yes, I can make a PR. For text creation by default we will still create them. Any thoughts or sharp comments?

underdogliu:
This is related to issue #4437 where I proposed an optional on data preparation without text. Such change should not affect the whole librispeech pipeline since by default we still have text file for speech recognition tasks. I already tested it on the 3 partitions of librispeech and `utils/fix_data_dir.sh` works for all of them. More tests and further sharp comments or even strong objections are welcomed. 

I made such trivial change for the benefit of ASR usage on multiple datasets as well as other speech tasks, where users of this script may suffer from purely integer speaker ids.

Changes made:
- Add `lbi-` prefix to speaker, utterance and text identifiers
- Make text creation optional (generate text by default)

Close #4437


egs/librispeech/s5/local/data_prep.sh
```diff
@@ -2,11 +2,19 @@
...
 
+no_text=false
+
+. ./utils/parse_options.sh
+
 if [ "$#" -ne 2 ]; then
   echo "Usage: $0 <src-dir> <dst-dir>"
   echo "e.g.: $0 /export/a15/vpanayotov/data/LibriSpeech/dev-clean data/dev-clean"
+  echo "Options:"
+  echo "   --no-text <true|false>           # Decide to disregard text or not."
```
danpovey, line 16: Could you remove the negation, and call the option prepare_text, default true?

underdogliu: changes made and tested on the 3 partitions.

kkm000:
LGTM. Merging, thanks!

PR #4441, opened 2021-01-28
[src,script,egs] the gop_speechocean762 recipe

jimbozhang:
For now, there is no public dataset for the pronunciation scoring task yet. Researchers have to use their private datasets in their papers.

Thanks to Speechocean's help, a dataset for the pronunciation scoring task is ready and allowed to be released for free. The dataset names `speechocean762`, which consists of 5000 English sentences. All the speakers are non-native, and their mother tongue is Mandarin. Five human experts made the scores in phoneme-level, word-level, and sentence-level. To avoid subjective bias, each expert scores dependently under the same metric. It will be great if it can be uploaded to OpenSLR.

Correspondingly, I wrote this `gop_speechocean762` recipe to replace the old `gop` recipe. In this recipe, phoneme-level scoring is illustrated and tested. The result was placed in `egs/speechocean762/s5/RESULT`. It is not a perfect performance, but I think it is okay as the baseline for researchers in their papers.

For phoneme-level scoring, a unique phoneme sequence is required as the target pronunciation. Therefore, we have to use the linear LG (without multiple pronunciations) in the forced alignment. For this purpose, the  `compile-train-graphs-without-lexicon` was added. This tool is similar to `compile-train-graphs` but does not load lexicon, and it requires the phone-level transcript as input.

`TrainingGraphCompiler` was also reconstructed because duplicate codes were in the member functions `TrainingGraphCompiler::CompileGraph` and `TrainingGraphCompiler::CompileGraphs`. 

The dataset has not been uploaded to OpenSLR yet, but there is a tar file on Xiaomi's server, and users can use it directly if they have the accessing permission:

```
# egs/gop_speechocean762/s5/local/download_and_untar.sh

...
# If you have permission to access Xiaomi's server, you would not need to
# download it from OpenSLR
path_on_mi_server=/home/storage07/zhangjunbo/share/data/$corpus_name.tar.gz
if [ -f $path_on_mi_server ]; then
  cp $path_on_mi_server $data/$corpus_name.tar.gz
fi
...
```


qindazhu:
+2  for the public dataset, there was a very long time that we have no a public dataset for pronunciation scoring task, I believe it will accelerate the research and application of computer-assisted language learning (CALL)

egs/gop_speechocean762/s5/run.sh
```diff
@@ -0,0 +1,212 @@
...
+  # For example, [ 27 -5.382001 ] means the GOP of the pure-phone 27 (it corresponds to the
+  # phone "OW", according to "$dir/phones-pure.txt") is -5.382001, indicating the audio
+  # segment of this phone should be a mispronunciation.
+  #
+  # The gop-base features are in matrix format:
+  # 4446-2273-0031  [ -0.2462088 -10.20292 -11.35369 ...
+  #                   -8.584108 -7.629755 -13.04877 ...
+  #                   ...
+  #                   ... ]
+  # The row number is the phone number of the utterance. In this case, it is 17.
```
qindazhu, line 169: I would like to say, the number of rows is the number of phones..

egs/gop_speechocean762/s5/run.sh
```diff
@@ -0,0 +1,212 @@
...
+  for part in train test; do
+    steps/make_mfcc.sh --nj $nj --mfcc-config conf/mfcc_hires.conf \
+      --cmd "$cmd" data/$part || exit 1;
+    steps/compute_cmvn_stats.sh data/$part || exit 1;
+    utils/fix_data_dir.sh data/$part
+  done
+fi
+
+if [ $stage -le 4 ]; then
+  # Extract ivector
```
qindazhu, line 74: No hurry, but it would be better to remove ivector later as it may not help for Pronunciation scoring task

src/bin/compute-gop.cc
```diff
@@ -63,7 +63,17 @@
 
 namespace kaldi {
 
-/** FrameLevelLpp compute a log posterior for pure-phones by sum the posterior
+int32 PhoneNum(const std::vector<std::set<int32> > &pdf2phones) {
```
qindazhu, line 66: It's a small issue, just noted we can do this faster, as `std::set` is sorted, we can get phone_num just as below:

```
for(auto &pdf: pdf2phones){
if(!pdf.empty()) phone_num = std::max(phone_num, 1+ pdf.rbegin());
}
```

egs/gop_speechocean762/s5/RESULT
```diff
@@ -0,0 +1,12 @@
+MSE: 0.15
```
qindazhu, line 1: It  would be better to add documentation for this file, e.g. MSE(mean_sequared_error), and what `0/1/2` means, what `macro/weighted avg` etc.

src/bin/compute-gop.cc
jimbozhang: Thanks haowen, I didn't know `std::set` is sorted. @wangyongqing0731 has modified it as your suggestion, but I think it should be `1+ *pdf.rbegin()` rather than `1+ pdf.rbegin()`. :sweat_smile:

egs/gop_speechocean762/s5/run.sh
jimbozhang: Yes, I agree. The ivector extraction in this recipe is only for reusing the librispeech model.

egs/gop_speechocean762/s5/RESULT
jimbozhang: I just updated the documents of both this file and the `README` in the `speechocean762` tar.

jimbozhang:
Thanks Haowen's review. Before merging, I think we need to upload the dataset to OpenSLR. The dataset tar file is in Xiaomi's server: `/home/storage07/zhangjunbo/share/data/speechocean762.tar.gz`, with the size 520810923.

jtrmal:
@danpovey can you copy it to /mnt/resources1/upload? I'll handle the rest.
y.

danpovey:
OK.  Might be a while though.

jimbozhang:
Thanks @jtrmal , here's some information of this dataset:

**Identifier (maybe):** SLR100
**Name:** speechocean762
**Category:** Speech
**Summary:** Pronunciation scoring dataset, labeled independently by five human experts.
**License:** Apache License v.2.0
**About this resource:** Pronunciation scoring is a crucial technology in computer-assisted language learning (CALL) systems. The pronunciation quality scores might be given at phoneme-level, word-level, and sentence-level for a typical pronunciation scoring task. However, there is no public dataset for this task. Researchers use their private datasets in their papers. This corpus aims to provide a free public dataset for the pronunciation scoring task. This corpus consists of 5000 English sentences. All the speakers are non-native, and their mother tongue is Mandarin. Half of the speakers are Children, and the others are adults. The information of age and gender are provided. Five experts made the scores. To avoid subjective bias, each expert scores independently under the same metric.

jtrmal:
identifier will be 101, I'm already in process of downloading a very cool
corpus for SLR100

can you please create info.txt and about.html as described here:
http://openslr.org/contributions.html
feel free to mail them to me directly, do not worry about creating an
archive with them
y.

jimbozhang:
Sure. I'm doing it and will notify you when ready.

> identifier will be 101, I'm already in process of downloading a very cool corpus for SLR100 can you please create info.txt and about.html as described here: http://openslr.org/contributions.html feel free to mail them to me directly, do not worry about creating an archive with them y.
> […](#)

danpovey:
I uploaded the data.

jimbozhang:
Hi @jtrmal , I mailed you with the attachments `about.html` and `info.txt`. Please kindly check them~

jimbozhang:
I think this version is ready to merge. @danpovey

danpovey:
Great!

PR #4442, opened 2021-01-28
Remove kaldiio dependence

desh2608:
Earlier, VBx used the `kaldiio` package while spectral clustering used `kaldi_io`. This PR removes dependence on `kaldiio`, so that read/write is more consistent across different diarization schemes.


danpovey:
Cool!  Merging.

PR #4446, opened 2021-02-04
Recipe for NSC corpus (Singaporean English, ~1000h of recordings)

pzelasko:
I prepared this some time ago and then forgot about it. Maybe it's useful for somebody. It is not super-well tuned (mimics LibriSpeech recipe mostly). The numbers seem right though, @sw005320 got similar results with ESPnet. As a fun fact this is probably the first Kaldi recipe that uses Lhotse (in a very limited way) 🙃


danpovey:
Great!
Merging, if there are problems we'll address them later.

PR #4447, opened 2021-02-05
modify pipeline for efficiency

KimJeongSun:
In original code, whole lats are extracted by 1 process per $src. 
This way takes lot more time when handling some large database splited by $nj_src.
so it need to modify code.
I would appreciate if you review it.


danpovey:
Looks OK to me, I assume you tested ?

KimJeongSun:
yes i'm using this code.

PR #4448, opened 2021-02-06
skip_kenlm_module_in_cmake

dophist:
Fixing:
* CMake building failure due to automatic detection of KenLM as a building target(mentioned in https://github.com/kaldi-asr/kaldi/pull/4247)

Code Changes:
* cmake/gen_cmake_skeleton.py: excludes KenLM related source files from CMake building targets

Testing:
* MacOS: Now CMake builds smoothly.

@nshmyrev Nickolay, please help to review and see if this fixes the problem in your building environment. I believe a clean-removal of out-of-tree "build dir", and rebuild with CMake from the beginning should be fine now.

### DIFF
--- a/cmake/gen_cmake_skeleton.py
+++ b/cmake/gen_cmake_skeleton.py
@@ -26,7 +26,7 @@ def is_bin_dir(d):
     return d.endswith("bin")
 
 def get_files(d):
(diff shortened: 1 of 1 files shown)


dophist:
@danpovey I think we can merge this, I've tested it in MacOS & Ubuntu.

danpovey:
Thanks!!

Issue #4303, opened 2020-10-21
Fail to make in kaldi/src: CUSPARSE_SPMM_CSR_ALG2 has not been declared in this scope


yangwanzhao:
I fail to 'make' for kaldi/src. There is a bug in kaldi/src/cudamatrix/cublas-wrappers.h related to the last update (support for Cuda 11). I am using Cuda 10.2 and the error says CUSPARSE_SPMM_CSR_ALG2 has not been declared in this scope. 

Do I need to use Cuda 11 to build it?

dophist:
same issue here, probably due to code change in https://github.com/kaldi-asr/kaldi/commit/c0ac791cebeba3d222caf1322c139ce3db880fb6

danpovey:
Oops.  Change _CSR_ALG2  to _DEFAULT and it should compile. @qmmn can you please fix it with an ifdef?
I think
`#if CUDA_VERSION >= 11000)`
would mean  CUDA version >= 11.0.00 (assuming that's where that starts to be defined).

qmmn:
@danpovey 
Yessss, fixed and PRed.

yesmung:
Hi, 
Checking for the same error, I came across this page.
And he was able to confirm that the error was resolved with @qmmn commit as below.

https://github.com/kaldi-asr/kaldi/pull/4304/commits/24008354746a5b409f3c77e1787b15c57b49c2e6

Thanks!!!

jimbozhang:
> @danpovey
> Yessss, fixed and PRed.

Has the PR been merged? Seem this problem still exists in the latest master branch.

danpovey:
It looks like no-one actually did craete the PR.  You could perhaps create it based on that ocmmit?

kkm000:
I am closing this issue for now. The discussion in  #4461 seems to suggest that you have the issue resolved.  If you believe that your issue has not been addressed, please feel free to ping me, and I'll reopen it. @-mention me please for a faster response!

PR #4304, opened 2020-10-21
Fix cusparseSpMM backward compatibility

qmmn:
Fix a compilation error with Cuda version < 11.0 on cusparseSpMM's argument:
- Cuda version < 11.0: CUSPARSE_MM_ALG_DEFAULT
- Cuda version >= 11.0: CUSPARSE_SPMM_CSR_ALG2

PR #4461, opened 2021-02-20
[src] fixed an issue of building with cuda11

jimbozhang:
For the issue #4303 
Tested with both cuda-10.1 and cuda-11.0.

### DIFF
--- a/src/cudamatrix/cublas-wrappers.h
+++ b/src/cudamatrix/cublas-wrappers.h
@@ -278,25 +278,14 @@ inline cusparseStatus_t cusparse_csrmm2(cusparseHandle_t handle,
   if (status != CUSPARSE_STATUS_SUCCESS) return status;
 
   size_t buffer_size;
-#if CUDA_VERSION >= 11000
-  status = cusparseSpMM_bufferSize(handle, transA, transB, alpha, matA, matB,
-                                   beta, matC, valType, CUSPARSE_SPMM_CSR_ALG2,
-                                   &buffer_size);
-#else
   status = cusparseSpMM_bufferSize(handle, transA, transB, alpha, matA, matB,
                                    beta, matC, valType, CUSPARSE_MM_ALG_DEFAULT,
                                    &buffer_size);
-#endif
   if (status != CUSPARSE_STATUS_SUCCESS) return status;
 
   void *buffer = (buffer_size > 0) ? CuDevice::Instantiate().Malloc(buffer_size) : NULL;
-#if CUDA_VERSION >= 11000
-  status = cusparseSpMM(handle, transA, transB, alpha, matA, matB, beta, matC,
-                        valType, CUSPARSE_SPMM_CSR_ALG2, buffer);
-#else
   status = cusparseSpMM(handle, transA, transB, alpha, matA, matB, beta, matC,
                         valType, CUSPARSE_MM_ALG_DEFAULT, buffer);
-#endif
 
   if (status != CUSPARSE_STATUS_SUCCESS) return status;
   if(buffer)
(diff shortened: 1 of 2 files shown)


src/makefiles/cuda_64bit.mk
```diff
@@ -10,7 +10,7 @@ CXXFLAGS += -DHAVE_CUDA -I$(CUDATKDIR)/include -fPIC -pthread -isystem $(OPENFST
 CUDA_INCLUDE= -I$(CUDATKDIR)/include -I$(CUBROOT) -I.. -isystem $(OPENFSTINC)
 CUDA_FLAGS = --compiler-options -fPIC --machine 64 -DHAVE_CUDA \
              -ccbin $(CXX) -DKALDI_DOUBLEPRECISION=$(DOUBLE_PRECISION) \
-             -std=c++14 -DCUDA_API_PER_THREAD_DEFAULT_STREAM -lineinfo \
+             -std=c++11 -DCUDA_API_PER_THREAD_DEFAULT_STREAM -lineinfo \
```
danpovey, line 13: is this change supposed to be here?

src/cudamatrix/cublas-wrappers.h
```diff
@@ -278,25 +278,14 @@ inline cusparseStatus_t cusparse_csrmm2(cusparseHandle_t handle,
   if (status != CUSPARSE_STATUS_SUCCESS) return status;
 
   size_t buffer_size;
```
danpovey, line 280: This seeems to be just removing the original support for ALG2... and you said you were compiling with 10.2; in this case surely CUDA_VERSION should not be 11000?  Maybe you changed the installation in some default place but failed to reconfigure?

dophist:
@jimbozhang, for reference, it seems you are reverting the code change in https://github.com/kaldi-asr/kaldi/pull/4304 which fixed the issue in https://github.com/kaldi-asr/kaldi/issues/4303

src/makefiles/cuda_64bit.mk
jimbozhang: With the `devtoolset-7-gcc` of the CentOS Developer Toolset, when I uses `c++14`, the building fails:
https://paste.ubuntu.com/p/N2xWXZ2jtm/

I haven't found why, but using `c++11` could address this problem. :flushed:

danpovey:
Seems to be an issue with certain versions of CUDA 10.1 combined with certain versions of gcc.
I think easiest fix would be to upgrade CUDA to a more recent version of 10.1.
That change would break the Kaldi build in other respects, so we can't do that.
  
https://forums.developer.nvidia.com/t/cuda-10-1-nvidia-youre-now-fixing-gcc-bugs-that-gcc-doesnt-even-have/71063/10

Let's focus on the errors you got with CUDA 11.

jimbozhang:
I compiled it on CUDA 11.0, and I found there is no `CUSPARSE_SPMM_CSR_ALG2 ` in `cusparse.h`:

```bash
$ grep -n CUSPARSE_MM_ALG_DEFAULT /usr/local/cuda/include/cusparse.h
5846:    CUSPARSE_MM_ALG_DEFAULT = 0,

$ grep -n CUSPARSE_SPMM_CSR_ALG2 /usr/local/cuda/include/cusparse.h
(nothing)
```

And the version information of `cusparse.h` is:
```c++
//##############################################################################
//# CUSPARSE VERSION INFORMATION
//##############################################################################

#define CUSPARSE_VER_MAJOR 11
#define CUSPARSE_VER_MINOR 0
#define CUSPARSE_VER_PATCH 0
#define CUSPARSE_VER_BUILD 191
#define CUSPARSE_VERSION (CUSPARSE_VER_MAJOR * 1000 + \
                          CUSPARSE_VER_MINOR *  100 + \
                          CUSPARSE_VER_PATCH)

```

But the CUDA version seems right ...
```bash
$ grep CUDA_VERSION /usr/local/cuda/targets/x86_64-linux/include/cuda.h
#define CUDA_VERSION 11000
```

danpovey:
According to this
https://docs.nvidia.com/cuda/archive/11.0/cusparse/index.html
it should be supported.
You are showing 2 different directories there.  I think there must be some
kind of mixup.

jimbozhang:
Yes, I have noticed the document. Thanks dan, and I'll recheck my CUDA installation.

Issue #4462, opened 2021-02-24
VB resegmentation gives dimension mismatch error due to improper use of coo_matrix


mdrpanwar:
Hi all,
Thanks a lot for this wonderful toolkit. I want to bring to your notice an issue with the VB resegmentation code that I came across while working on speaker diarization. I used callhome dataset, extracted mfcc features using kaldi, and towards the end of my pipeline while performing resegmentation, I got the following error:

![Screenshot from 2021-02-12 01-52-20](https://user-images.githubusercontent.com/39766613/108947749-1df21900-7687-11eb-950b-445f7414eed2.png)

I was using linux and just to be sure, I tried to reproduce the same issue on Mac using [the official docker image](https://hub.docker.com/r/kaldiasr/kaldi). However, it executed perfectly on Mac without failing. Upon inspecting the files, I found that files `egs/callhome_diarization/v1/diarization/VB_resegmentation.py` and `egs/callhome_diarization/v1/diarization/VB_diarization.py` are different in the [git repository](https://github.com/kaldi-asr/kaldi) and the [docker image](https://hub.docker.com/r/kaldiasr/kaldi). The one in docker image is some previous version.
Even though a lot has changed and been renamed in these files, the significant change that led to this issue was in [lines 131-132 of VB_diarization.py](https://github.com/kaldi-asr/kaldi/blob/master/egs/callhome_diarization/v1/diarization/VB_diarization.py#L131-L132): 
`  F_s =coo_matrix((((X[zeta.row]-m[zeta.col])*zeta.data[:,np.newaxis]).flat,
                   (zeta.row.repeat(D), zeta.col.repeat(D)*D+np.tile(range(D), len(zeta.col)))))`

The `coo_matrix` must be sent a shape parameter as well (as in the docker image version of this file). If not, it infers the shape of output matrix from the index arrays passed, which will not always agree with the desired output shape and therefore result in a dimension mismatch during matrix multiplication. I changed the above to the following:
` F_s =coo_matrix((((X[zeta.row]-m[zeta.col])*zeta.data[:,np.newaxis]).flat,
                   (zeta.row.repeat(D), zeta.col.repeat(D)*D+np.tile(range(D), len(zeta.col)))), shape=(nframes, D*C))`

[...]

PR #4463, opened 2021-02-24
passing shape parameter to coo_matrix to have fixed dimensional output matrix

mdrpanwar:
The `coo_matrix` must have a shape parameter to govern the dimensions of the output matrix, or else this results in a dimension mismatch during matrix multiplication in [line 133 in VB_diarization.py](https://github.com/kaldi-asr/kaldi/blob/master/egs/callhome_diarization/v1/diarization/VB_diarization.py#L133).

For detailed stack trace of the error and related discussion see #4462.


danpovey:
Thanks!!

PR #4467, opened 2021-03-03
read_entire_file as parameter in get-utt2dur.sh

oplatek:
I described the problem here: https://github.com/espnet/espnet/issues/3038


kkm000:
LGTM, merging, thanks.

You normally shouldn't need to re-read the dataset to get the utt_2dur file. I left the pointers in a comment https://github.com/espnet/espnet/issues/3038#issuecomment-808667650

PR #4470, opened 2021-03-08
Chime6 track1

aarora8:


egs/chime6/s5b_track1/run.sh
```diff
@@ -0,0 +1,307 @@
+#!/usr/bin/env bash
+#
+# Based mostly on the TED-LIUM and Switchboard recipe
```
danpovey, line 3: I think we should have a comment here on in a README about what "track 1" is, people in the future or who don't know about the eval may find this confusing.

aarora8: Thank you, I added the details.

danpovey:
@sw005320 if you could check over this would be good (can fix any issues in another PR..)

PR #4483, opened 2021-03-23
Changes needed to get Kaldi compiled on Win and

jtrmal:
*usleep does not exist on windows and this_thread::sleep is platform-independent way how to do this.
*the new Intel OneAPI (rebranded MKL and Compiler) changed a bit the directory structure. Also should be backward compatible.
*Will be grateful if someone will test on linux...

Signed-off-by: Jan "Yenda" Trmal <yenda@kalditek.com>

### DIFF
--- a/cmake/FindBLAS.cmake
+++ b/cmake/FindBLAS.cmake
@@ -457,9 +457,15 @@ if(BLA_VENDOR MATCHES "Intel" OR BLA_VENDOR STREQUAL "All")
         endif()
       endif()
       set(BLAS_mkl_LIB_PATH_SUFFIXES
-          "compiler/lib" "compiler/lib/${BLAS_mkl_ARCH_NAME}_${BLAS_mkl_OS_NAME}"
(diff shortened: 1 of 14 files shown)


jtrmal:
closing

jtrmal:
@kkm000 you might want to consider merging this, if it's to your taste...
y.

kkm000:
Looks like I have a cleaner one already. I'm waiting on #4490 to go in first. Essentially, I got rid of the dependency on windows.h in the file where Sleep() is defined (forgot what it was). Also, I replaced these defines with more concisely named const doubles. 

To be honest,
```c++
std::this_thread::sleep_for(std::chrono::microseconds(KALDI_CUDA_DECODER_WAIT_FOR_CPU_FEATURES_THREADS_US));
```
is not to my taste. :) Also, the selection of the clock is left to an implementation here; an explicit cast into `chrono::high_resolution_clock` is more robust.

I'll go through it carefully to check if I've missed anything. 

There were a couple of keywords like `__inline__` and `__align__` that the compiler frowned upon in the pedantic mode.

PR #4069, opened 2020-05-14
[WIP] Add HIMIA (openslr-85) recipe speaker verification

underdogliu:
Hi all,

So here is a recipe of speaker verification experiments for HIMIA (openslr-85) data set. Here, different from original paper, we used x-vector recipe here. We enquired speech data from `multi_cn` that has speaker information (in fact all of them do have). AISHELL2 can be included as well but we made it optional since not everybody owns it.

Note: for scripts cited from `multi_cn`, because we do not have to use textual information here I made slight changes on them. Besides, I remember @naxingyu was planning to make some changes on them so I will keep those on track and changes can be made later (before or after merging).

Note2: The name of such egs is `hi_mia_sv` but feel free to suggest better names. I named it because there is already a similar PR from @nwpuaslp on KWS using same data.

README file is under active writing and hope to be finished soon. Meanwhile people can review script I have written or run experiments to see if there is any unexpected bug there. I will keep things tracked and fixed.

Thanks

PR #4484, opened 2021-03-24
HI-MIA update: complementing trial list format and text-independent results

underdogliu:
Dear all,

In case you are interested on doing work on HIMIA under kaldi, from some of my recent experiments I noticed that the trial list created is at speaker-level. I cannot recall if it has been updated sometime or not because it used to work as confirmed by @jimbozhang from PR #4069 but I do the update on TI scripts with slight fixture on README. 

Please do not merge yet. This PR is under testing phase. Thanks a lot.


underdogliu:
@danpovey @jimbozhang I think it is ready to merge. Please have a check and make sharp comments if you find anything does not look right. Thanks a lot!

PR #4490, opened 2021-04-05
[src] making the cuda dynamic batcher long stream friendly

hugovbraun:
The function Push(..) of the dynamic batcher now guarantees that a deep copy of the samples will be done. So that the caller doesn't have to deal with lifetime/deallocation of the samples. 
Taking the opportunity to also improve performance of the set "batcher + pipeline".

A "TODO' is intentionally left in the code. A line can be added to improve performance of the H2D copy, but I'll take care of it in a future commit.

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
@@ -22,7 +22,9 @@
 #define KALDI_CUDA_DECODER_WAIT_FOR_AVAILABLE_CHANNEL_US 1000
 
 #include "cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h"
+
 #include <nvToolsExt.h>
+
 #include "feat/feature-window.h"
 #include "lat/lattice-functions.h"
 #include "nnet3/nnet-utils.h"
@@ -57,6 +59,10 @@ void BatchedThreadedNnet3CudaOnlinePipeline::AllocateAndInitializeData(
     h_all_ivectors_.Resize(max_batch_size_, ivector_dim_, kSetZero,
                            kStrideEqualNumCols);
   }
+
+  // Set d_features_ptrs_, d_ivectors_ptrs_, features_frame_stride_
+  // to be used with DecodeBatch
+  SetFeaturesPtrs();
 
   d_all_log_posteriors_.Resize(max_batch_size_ * output_frames_per_chunk_,
                                trans_model_->NumPdfs(), kUndefined);
@@ -158,11 +164,8 @@ bool BatchedThreadedNnet3CudaOnlinePipeline::TryInitCorrID(
   return true;
 }  // namespace cuda_decoder
 
-void BatchedThreadedNnet3CudaOnlinePipeline::ComputeGPUFeatureExtraction(
-    const std::vector<int> &channels,
-    const std::vector<SubVector<BaseFloat>> &wave_samples,
-    const std::vector<bool> &is_first_chunk,
-    const std::vector<bool> &is_last_chunk) {
+void BatchedThreadedNnet3CudaOnlinePipeline::CompactWavesToMatrix(
+    const std::vector<SubVector<BaseFloat>> &wave_samples) {
   for (int i = 0; i < wave_samples.size(); ++i) {
     const SubVector<BaseFloat> &src = wave_samples[i];
     int size = src.Dim();
@@ -171,10 +174,18 @@ void BatchedThreadedNnet3CudaOnlinePipeline::ComputeGPUFeatureExtraction(
     BaseFloat *wave_dst = h_all_waveform_.RowData(i);
     std::memcpy(wave_dst, wave_src, size * sizeof(BaseFloat));
   }
+}
+
+void BatchedThreadedNnet3CudaOnlinePipeline::ComputeGPUFeatureExtraction(
+    const std::vector<int> &channels, const Matrix<BaseFloat> &h_all_waveform,
+    const std::vector<int> &n_samples_valid,
+    const std::vector<bool> &is_first_chunk,
+    const std::vector<bool> &is_last_chunk) {
   // CopyFromMat syncs, avoiding it
-  KALDI_ASSERT(d_all_waveform_.SizeInBytes() == h_all_waveform_.SizeInBytes());
-  cudaMemcpyAsync(d_all_waveform_.Data(), h_all_waveform_.Data(),
-                  h_all_waveform_.SizeInBytes(), cudaMemcpyHostToDevice,
+  KALDI_ASSERT(d_all_waveform_.SizeInBytes() == h_all_waveform.SizeInBytes());
+  // Note : we could have smaller copies using the actual channels.size()
+  cudaMemcpyAsync(d_all_waveform_.Data(), h_all_waveform.Data(),
+                  h_all_waveform.SizeInBytes(), cudaMemcpyHostToDevice,
                   cudaStreamPerThread);
 
   KALDI_ASSERT(channels.size() == is_last_chunk.size());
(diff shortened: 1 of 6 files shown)


src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -273,13 +300,26 @@ void BatchedThreadedNnet3CudaOnlinePipeline::DecodeBatch(
...
     channels = &channels_;
     ListIChannelsInBatch(corr_ids, channels);
   }
 
+  partial_hypotheses_ = in_partial_hypotheses;
+  end_points_ = in_end_points;
+  {
+    std::lock_guard<std::mutex> lk(map_callbacks_m_);
+    if (!best_path_callbacks_.empty()) {
+      // If best path callbacks are in use, always activate partial hyp and endp
```
kkm000, line 317: I do not understand why best path callback implies partial hypotheses and endpointing. In the binary sample both are gladly ignored in case the user did not ask for them. Are they inseparable?
https://github.com/kaldi-asr/kaldi/blob/38b80369ded5609b39a7dc8482f29c865e0cd3f7/src/cudadecoderbin/batched-wav-nnet3-cuda-online.cc#L162-L185

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
```diff
@@ -44,10 +51,76 @@ CudaOnlinePipelineDynamicBatcher::~CudaOnlinePipelineDynamicBatcher() {
...
+    const SubVector<BaseFloat> &wave_samples_subv) {
+  std::lock_guard<std::mutex> lk(next_batch_and_backlog_m_);
+
+  // Try to add the chunk is the batch right away
+  if (!TryAddChunkToNextBatchDeepCopy(corr_id, is_first_chunk, is_last_chunk,
+                                      wave_samples_subv)) {
+    // If we cannot add the chunk to the next batch,
+    // put it in the backlog
+    Vector<BaseFloat> wave_samples(wave_samples_subv);  // deep copy
+    backlog_.push_back(
```
kkm000, line 63: 1. Am I understanding that the backlog essentially holds all the data pushed so far and not fitting the batch, with an essentially unbounded growth? In my current design, I own the waveforms anyway.
2. The deep copying here seems to increase the time spent under the lock. Memory locality here is not great already, as the total amount of data pumped causes L2 misses, and L1 is not even in the game, so memory accesses are slow (this is unavoidable, I think). And this lock is a contention point for all streaming channels, and at the same time must be obtained by the BatcherThreadLoop every 100 us to avoid starving the device. This effect seems to be counteracted by the filling from the backlog in the same thread, without the need to obtain the log again. Does this provide an overall win for the contention on this single lock?

I'm wondering because my use is different from realtime streaming in the online example: I'm streaming much faster than realtime.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
```diff
@@ -44,10 +51,76 @@ CudaOnlinePipelineDynamicBatcher::~CudaOnlinePipelineDynamicBatcher() {
...
+        {corr_id, is_first_chunk, is_last_chunk, std::move(wave_samples)});
+  }
+  n_chunks_not_done_.fetch_add(1, std::memory_order_release);
```
kkm000, line 66: If my understanding is correct, `n_chunks_not_done_` is not _Synchronized-With_ other atomic operations anywhere. If that's the case, `std::memory_order_relaxed` is more appropriate on every access to it. No real difference on the x64 architecture, as it has quite a strict memory ordering, but ARM and even more so POWER9 reorder accesses wildly. I think we support ARM.

If you put it alone by itself into a 64-byte structure aligned at a 64-byte boundary, it won't cause any false sharing, and will flush only a single cacheline. Since it's one per decoder, not much memory waste, but the accesses will be faster.

kkm000:
If it's good to go by you, I'll commit?

I think I should be able to compare the perf of the existing/new code w.r.t locking. I can't promise 100%, as my code may change due to this change, but that's does not seem likely.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
hugovbraun: Both are used as arguments for the best path callbacks, and once we have one having the other is pretty cheap anyway

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
hugovbraun: Yes, that would behave poorly if each stream is pushing audio faster than realtime. The assumption here is that most chunks should be placed directly into the batch, skipping the backlog. But that doesn't work if we remove the "realtime" part. Do you own the waveforms because you own the entire audio file anyway ?

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
hugovbraun: I added std::memory_order_release to protect the backlog_ write from being reordered after. Is it not necessary?

hugovbraun:
It's good to go by me. Thanks for the review!

kkm000:
@hugovbraun, I'm going to send it my changes today. I wanted us to avoid touching the same files to decrease conflicts. This is to make the code compile on Windows, and improving readability (all time-valued constants in double seconds).

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
kkm000: The more I'm thinking about this, the more confused I feel... I was probably wrong and you're right, the backlog item counter should use the Acquire-Release semantics. In any case, better safe than sorry :)

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
kkm000: I don't think this should behave poorly because of the backlog. Both sleeping and acquiring a lock are expensive operations, and backfilling happens while the lock is held. I really don't know right now, I'm only starting on perfing the whole pipeline right now. 

I do not own the entire file, it's streamed, but in quite hefty buffers. Also, there are segment in and out points, so every stream is jerky and contains multiple utterances often separated by silence. It's complicated.

All in all, let me play with it first; I'll certainly try many scenarios, and will keep you in the loop. The new code is simpler than the previous version, much easier to reason about the parts which are prone to locking contention, backlog blow-up etc., so I'll give it a good dusting! :)

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
kkm000: Got it, thanks!

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
kkm000: I was certainly wrong. After reading the C++11 spec, it is indeed true that both the compiler and the HW are allowed to reorder across atomics with the relaxed memory semantics.

I attempted to see if the compile reordering happens, and did not find any on godbolt that would. https://godbolt.org/z/djcnx6zab -- uncommenting line 25 pessimizes both x64 and armv8 code from 2 wide to 4 narrow stores, 2 before the atomic increment and 2 after, as the code is written. Then I found this: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4455.html -- quite an interesting read encouraging the compiler writers to optimize atomics. 6 year past, it did not happen. But it's not only allowed, but kinda encouraged, too.

PR #4498, opened 2021-04-19
CHiME-6 track 2 JHU-CLSP submission

desh2608:
* Rename s5b_track2 to s5c_track2 (for TS-VAD)

New s5b_track2 recipe (based on JHU-CLSP submission to CHiME-6 track 2):
* Multi-array posterior fusion for VAD
* Overlap-aware VB resegmentation (using Pyannote based overlap detection)
* Multi-array GSS using diarization output
* RNNLM rescoring


danpovey:
Great!
Do you think this needs much checking?

desh2608:
@danpovey I ran it start to end so it should be good. Could you just look at the new src file `src/bin/matrix-max.cc`?

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+      const std::vector<std::string> &uttlist = spk2utt_reader.Value();
+      if (uttlist.empty()) {
+        KALDI_ERR << "Speaker with no utterances.";
+      }
+      Matrix<BaseFloat> spk_max;
+      int32 utt_count = 0;
+
+      // All matrices may be unequal size. First we drop last
+      // rows of matrices and make them the same size.
+      int32 min_length = std::numeric_limits<std::int32_t>::max();;
```
kkm000, line 73:
```suggestion
      int32 min_length = std::numeric_limits<int32>::max();
```
The template argument should be of the type which is being queried. C++ primitive types cannot be derived, only aliased, so the specialization for an actual type that `kaldi::int32` aliases will be used.
Also a nit: extra `;`.

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+        }
+      }
+      for (size_t i = 0; i < uttlist.size(); i++) {
+        std::string utt = uttlist[i];
+        if (!matrix_reader.HasKey(utt)) {
+          KALDI_WARN << "No matrix present in input for utterance " << utt;
+          num_utt_err++;
+        } else {
+          if (utt_count == 0)
+            spk_max.Resize(min_length, matrix_reader.Value(utt).NumCols());
```
kkm000, line 87: nit: multiline loop/conditional's body must be a block.
```suggestion
          if (utt_count == 0) {
            spk_max.Resize(min_length, matrix_reader.Value(utt).NumCols());
          }
```

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+        if (matrix_reader.HasKey(utt)) {
+          min_length = std::min(min_length, matrix_reader.Value(utt).NumRows());
+        }
+      }
+      for (size_t i = 0; i < uttlist.size(); i++) {
+        std::string utt = uttlist[i];
+        if (!matrix_reader.HasKey(utt)) {
+          KALDI_WARN << "No matrix present in input for utterance " << utt;
+          num_utt_err++;
+        } else {
```
kkm000, line 85: This `else`  block is long, and has nested `if`s. I'd rather use a `continue` above and save an indent. That would make the body easier to read.

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+          if (utt_count == 0)
+            spk_max.Resize(min_length, matrix_reader.Value(utt).NumCols());
+          if (matrix_reader.Value(utt).NumRows() > min_length) {
+            KALDI_WARN << "Cropping rows of matrix " << utt << " from "
+                        << matrix_reader.Value(utt).NumRows() << " to " 
+                        << min_length;
+
```
kkm000, line 92: Empty line.
```suggestion
```

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+        num_spk_done++;
+      }
+    }
+
+    KALDI_LOG << "Computed max for " << num_spk_done << " speakers ("
+              << num_spk_err << " with no utterances), consisting of "
+              << num_utt_done << " utterances (" << num_utt_err
+              << " absent from input).";
+
+    return (num_spk_done != 0 ? 0 : 1);
```
kkm000, line 116: The parens are superfluous.
```suggestion
    return num_spk_done != 0 ? 0 : 1;
```

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+          SubMatrix<BaseFloat> temp(matrix_reader.Value(utt), 0, min_length, 
+              0, matrix_reader.Value(utt).NumCols());
+          spk_max.Max(temp);
+          num_utt_done++;
+          utt_count++;
+        }
+      }
+      if (utt_count == 0) {
+        KALDI_WARN << "Not producing output for speaker " << spk
+                    << " since no utterances had matrices";
```
kkm000, line 103: nit: extra space
```suggestion
        KALDI_WARN << "Not producing output for speaker " << spk
                   << " since no utterances had matrices";
```

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
...
+        KALDI_ERR << "Speaker with no utterances.";
+      }
+      Matrix<BaseFloat> spk_max;
+      int32 utt_count = 0;
+
+      // All matrices may be unequal size. First we drop last
+      // rows of matrices and make them the same size.
+      int32 min_length = std::numeric_limits<std::int32_t>::max();;
+      for (size_t i = 0; i < uttlist.size(); i++) {
+        std::string utt = uttlist[i];
```
kkm000, line 75: nit: `++i` (and below).
Alternatively, since the loop variable `i` isn't used in either loop body, both can be elevenized as
```suggestion
      for (const std::string& utt : uttlist) {
```

src/bin/matrix-max.cc
```diff
@@ -0,0 +1,121 @@
+// featbin/matrix-mean.cc
+
+// Copyright 2013-2014  Daniel Povey
```
kkm000, line 3:
```suggestion
// Copyright 2021  Desh Raj
```
:)

kkm000 (changes requested):

desh2608:
@kkm000 thanks for your comments. I have made the changes you suggested.

kkm000 (approved):

kkm000:
Cool, LGTM! If you're comfortable with the scripting part, I'll merge after the CI run completes.

Issue #4297, opened 2020-10-17
Build problems with recent -std=c++14 option in CUDA


danpovey:
There is some discussion here
 https://groups.google.com/d/msgid/kaldi-help/e4a7cc57-5c07-4888-9322-359f13589cb9n%40googlegroups.com.
e.g. with gcc 4.8.5 this didn't work.
we may have to update the configure script.

kkm000:
What should we do: bump the minimum required compiler version unconditionally, or only if CUDA 11 is used? @danpovey, @jtrmal, what do you think?

GCC 4.8.5 was released June 23, 2015, 5.5 years ago.

kkm000:
X-Ref: #4245

jtrmal:
IMO would be fine to bump the version of the compiler unconditionally...
y.

kkm000:
Looks like 4.8.5 comes with Centos 7. I never used it. Is it possible (and easy enough) to get a newer toolchain on that OS?

jtrmal:
it might be, but people using centos7 will be probably very conservative,
so it might be significant obstacle
y.

kkm000:
People using Sun Solaris are even more conservative. We need to draw the line somewhere.

jtrmal:
I'm not arguing about that :)
y

jtrmal:
actually, I'm not arguing at all :p
y.

kkm000:
https://linuxize.com/post/how-to-install-gcc-compiler-on-centos-7/#installing-multiple-gcc-versions

kkm000:
@jtrmal, what is the lowest Apple clang we need—do we need to bump it, too? I'll check gcc and LLVM clang, but I have no Mac.

jtrmal:
this might help -- I'm not quite sure what would be the baseline:

https://trac.macports.org/wiki/XcodeVersionInfo

setting it to 10 or maybe 9 should be fine IMO
according to this
https://update.omnigroup.com/
most versions of macs are like three "releases" old at most
y.

kkm000:
[...]

Issue #4245, opened 2020-08-27
Update to C++14


kkm000:
Check compiler support for minimum versions in configure.

The rest is fully compatible; use `std::make_unique<T>`, omitted from C++11 by mistake, whenever feasible.

kkm000:
X-Ref: #4242 4242

PR #4503, opened 2021-04-21
[build] Bump minimum compiler versions and enable C++14

kkm000:
```
===================================
                Min. req. version:
Compiler       Current       New
===========    =======     ========
GCC             4.9.1   >    5.0
LLVM Clang      3.3     >    3.5
XCode Clang     5.0     >    6.0
===================================
```
At the same time, upgrade OpenFST to 1.7.2 (which I've been using for quite a while without a glitch).

/cc @hugovbraun

Close: #4297
Close: #4245

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -77,7 +77,7 @@ else ifeq ($(OS),Windows_NT)
   # This new OS path is confirmed working on Windows 10 / Cygwin64.
   openfst_add_CXXFLAGS = -g -O2 -Wa,-mbig-obj
 else
-  openfst_add_CXXFLAGS = -g -O3
+  openfst_add_CXXFLAGS = -g -O2
 endif
 
 openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION)
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -17,8 +17,7 @@ SCTK_GITHASH = 20159b5
 
 # Default features configured for OpenFST; can be overridden in the make command line.
 OPENFST_CONFIGURE ?= --enable-static --enable-shared --enable-far \
-                     --enable-ngram-fsts --enable-lookahead-fsts \
-                     --disable-dependency-tracking --with-pic
+                     --enable-ngram-fsts --enable-lookahead-fsts --with-pic
 
 OPENFST_VER_NUM := $(shell echo $(OPENFST_VERSION) | sed 's/\./ /g' | xargs printf "%d%02d%02d")
 ifeq ("$(shell expr $(OPENFST_VER_NUM) \< 10607)","1")
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -7,9 +7,7 @@ CC ?= gcc        # used for sph2pipe
 
 WGET ?= wget
 
-# Note: OpenFst requires a relatively recent C++ compiler with C++11 support,
-# e.g. g++ >= 4.7, Apple clang >= 5.0 or LLVM clang >= 3.3.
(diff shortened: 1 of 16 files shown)


tools/Makefile
```diff
@@ -19,12 +17,13 @@ SCTK_GITHASH = 20159b5
 
 # Default features configured for OpenFST; can be overridden in the make command line.
 OPENFST_CONFIGURE ?= --enable-static --enable-shared --enable-far \
-                     --enable-ngram-fsts --enable-lookahead-fsts --with-pic
+                     --enable-ngram-fsts --enable-lookahead-fsts \
+                     --disable-dependency-tracking --with-pic
```
galv, line 21: Is `--disable-dependency-tracking` really a good idea? You are basically assuming that users won't modify and develop openfst source code, which is something I've done in the past.

tools/Makefile
```diff
@@ -78,13 +77,13 @@ else ifeq ($(OS),Windows_NT)
...
 else
-  openfst_add_CXXFLAGS = -g -O2
+  openfst_add_CXXFLAGS = -g -O3
 endif
 
 openfst-$(OPENFST_VERSION)/Makefile: openfst-$(OPENFST_VERSION)
 	cd openfst-$(OPENFST_VERSION)/ && \
 	./configure --prefix=`pwd` $(OPENFST_CONFIGURE) CXX="$(CXX)" \
-		CXXFLAGS="$(CXXFLAGS) $(openfst_add_CXXFLAGS)" \
+		CXXFLAGS="$(openfst_add_CXXFLAGS) $(CXXFLAGS)" \
```
galv, line 86: Given that you've made this change to swap the order of `openfst_add_CXXFLAGS` and `CXXFLAGS`, can you change -O3 back to -O2 in `openfst_add_CXXFLAGS`? Right now, a semi-large change is adding different, unrelated changes. If you really want -O3 for your use case, you can do `CXXFLAGS="-O3" make`

galv (commented):
excellent forward progress

tools/Makefile
kkm000: When automatically compiling it as part of Kaldi, I think it's a good idea. If you really hack OpenFST in context of Kaldi to the point dependencies change, you'll probably rerun configure anyway many times.

tools/Makefile
kkm000: Good point! Done.

tools/Makefile
jtrmal: I think the --disable-dependency-tracking  flag should be removed. I don't want to optimize for the time of machine. I want to optimize for time of people who are trying to track down an issue.

jtrmal:
LGTM, with the exception of the tracking dependecies flag

tools/Makefile
kkm000: Done.

kkm000:
@jtrmal, ok, I probably underestimate the importance of dependencies. Anyway, the build with libtool is so slow that the speedup from skipping 'make depend' is negligible anyway. FWIW, later versions of OpenFST come with Bazel build files, but no CMake :( The library (w/o bins) builds is about 6-7 seconds with CMake/ninja if `-j` arg (and the # of hyperthreads) equals or exceeds the number of `.cc` files in `src/lib`. And, alas, there is no way to build it with `-std=c++14`. When we upgrade the whole rig to CMake, providing our own CMake build framework for OpenFST seems to be the best approach (given that you wrote it already for the Windows port, it's a zero effort task, with a sizable build time benefit).

I'll merge when Travis completes, thanks for looking!

PR #4506, opened 2021-04-24
[src] Compile cudadecoder binaries on Windows

kkm000:
* Make `BatchedThreadedNnet3CudaOnlinePipeline::CorrelationID`  public for binaries in src/cudadecoderbin to compile (fallout from #4490 -- our CI runs w/o CUDA).
* Replace Win32-specific `Sleep()` with portable C++ `std::chrono`  classes in src/base/kaldi-utils.*. Change `Sleep(float)` arg type to `double` because this is what these classes take and return when converting to/from time in seconds.
* Use C++ `double` constants for timers instead of sesquipedalian macros. Express time invariably in seconds.
* Move includes under `HAVE_CUDA` conditionals in files without code that would make no sense without CUDA. Check for `HAVE_CUDA` always after include guards for consistency.
* Issue an `#error` if attempting to compile CUDA-dependent code with `HAVE_CUDA` not set (not yet all files, only these I had to touch).
* Add, reword some and reformat commentary and help strings in src/cudadecoderbin/batched-wav-nnet3-cuda-online.cc

Accidental touch-paint fixes:
* Replace `usleep()` with `kaldi::Sleep()` in src/util/kaldi-table-test.cc.
* Fix escape non-sequence `\%` where `%` was intended in strings in src/nnet3/nnet-utils.cc.

@hugovbraun, PTAL.

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
@@ -15,13 +15,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if HAVE_CUDA == 1
-
-#define KALDI_CUDA_DECODER_WAIT_FOR_CALLBACKS_US 10000
-#define KALDI_CUDA_DECODER_WAIT_FOR_CPU_FEATURES_THREADS_US 1000
-#define KALDI_CUDA_DECODER_WAIT_FOR_AVAILABLE_CHANNEL_US 1000
+#if !HAVE_CUDA
+#error CUDA support must be configured to compile this library.
+#endif
 
 #include "cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h"
+
+#include <mutex>
+#include <numeric>
+#include <tuple>
 
 #include <nvToolsExt.h>
 
@@ -31,6 +33,11 @@
 
 namespace kaldi {
 namespace cuda_decoder {
+
+const double kSleepForCallBack = 10e-3;
+const double kSleepForCpuFeatures = 1e-3;
+const double kSleepForChannelAvailable = 1e-3;
+
 void BatchedThreadedNnet3CudaOnlinePipeline::Initialize(
     const fst::Fst<fst::StdArc> &decode_fst) {
   ReadParametersFromModel();
@@ -115,7 +122,7 @@ void BatchedThreadedNnet3CudaOnlinePipeline::SetBestPathCallback(
 }
 
 bool BatchedThreadedNnet3CudaOnlinePipeline::TryInitCorrID(
-    CorrelationID corr_id, int wait_for) {
+    CorrelationID corr_id, int32 wait_for_us) {
   bool inserted;
   decltype(corr_id2channel_.end()) it;
   std::tie(it, inserted) = corr_id2channel_.insert({corr_id, -1});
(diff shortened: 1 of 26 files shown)


src/cudadecoder/thread-pool-light.h
```diff
@@ -99,7 +100,7 @@ class ThreadPoolLightWorker {
         (curr_task_.func_ptr)(curr_task_.obj_ptr, curr_task_.arg1,
                               curr_task_.arg2);
       } else {
-        usleep(1000);  // TODO
+        Sleep(1e-3f);  // TODO
```
jtrmal, line 103: just imo, this does not really help readability
maybe either rename the function to something like "SleepSeconds" or add defines so that it says Sleep (1000 *kMiliseconds) or something else :)
y

jtrmal: Or both rename the function and add the defines/constants
y.

kkm000: I thought about that. There is a much better solution. C++11 (indirectly) support units of measure through user-defined literals. The indirect part is it's up to the implementor to make sure that, for example, mass added to time fails to compile (by treating different units as isomorphic but distinct fields, e.g. the `duration` type that does not add to either unitless double value or and instance of the `mass`, but allows to multiply them, so they commute categorically over `duration×mass`). A full support for arbitrary units¹ would probably consist of a lot of C++17 variadic template trickery, but I vaguely remember Boost had a stab at it.

Anyhow, C++14 stdlib++ includes [duration literals constructors](https://en.cppreference.com/w/cpp/chrono/operator%22%22ms) that support duration units; you could be able to write `Sleep(250ms)`, or pass `const auto kDelay = 250ms;` to it, etc.

When I have a breather, I am going to add an overload to `Sleep()` that would accept the `duration` type, and the rest is already provided by the choro library (essentially, the same implementation as I added minus the first line, although durations are be tricky when converted between different clocks—they are intrinsically defined by clocks). Maybe we're able to deprecate the non-dimensional version some day.

I do not think there is so much ambiguity, as the second is the SI unit, and, besides, most of other absolute durations in Kaldi are in seconds by default (such as frame offset etc.). Anyway, it's inconsistent, and migrating to dimensioned duration may make the code more robust and easier to use. But that's when I have a breather and take care of at least the worst of another 18 issues assigned to me :)

---

¹ My favorite unit of measure (volume, to be specific) that I ever came across is the _buttload_. The definition shifted over time, place, and the exact liquid volume of which was being measured, but an oft cited conversion equals 1 buttload of wine, and sometimes beer, to 4 barrels, so that's certainly quite a _buttload_ of booze, in the modern sense of the word.

hugovbraun:
Looks great, thanks for doing that!

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -31,6 +33,11 @@
 
 namespace kaldi {
 namespace cuda_decoder {
+
+const double kSleepForCallBack = 10e-3;
```
hugovbraun, line 37: Probably compiles to the same thing, but couldn't we use constexpr here in theory?

kkm000: Yes and yes. :) I haven't seen a single compiler treating const qualification and constexpr declaration of simple numeric types differently: before constexpr, const was a de-facto guarantee of the literal constant substitution, and undoing that, although legal per C++11, would affect a lot of existing code. But constexpr indeed provides the compile-time evaluation guarantee.

I ultimately want to use [the `chrono` literals](https://en.cppreference.com/w/cpp/chrono/operator%22%22us) for time constants anyway, like `Sleep(250ms)` or `constexpr auto kDelay = 250ms;`. They are too sweet and readable not to use them. :) Do you mind if we leave them as is in the meantime?

hugovbraun: I agree that literals would be perfect - but for now it's fine. I'll update some other parts of the code to make it more C++14-like too (like finally being able to use std::make_unique :) )

kkm000: Oh yeah, the delightsome long coveted `std::make_unique` may at last be used!

I once heard someone referring to C++14 as "C++11 service pack 1".

And since we're now allowed to use C++14, we can finally begin craving the C++17 features! :-D

PR #4520, opened 2021-05-08
[src] Assert matrix size before memcpy into it

kkm000:
* `KALDI_ASSERT` enough space in a matrix column before `memcpy()`ing into it.
* [chore] rearrange the include guard to make including the file invariant w.r.t. `HAVE_CUDA` defined, and rename the guard preprocessor variable to the coding conventions.

The pipeline object has an accessor that returns the correct number of samples per chunk, but I slapped together a quick POC code that had no reference to the pipeline at the point I called `Push()`, computed the wrong size, and the "quick" part turned into a day-long bug hunting. The OED has both _slow_ and _stupid_ as antonyms to _quick_ for a good reason...  :-0

/cc @hugovbraun

### DIFF
--- a/src/cudadecoder/cuda-online-pipeline-dynamic-batcher.h
+++ b/src/cudadecoder/cuda-online-pipeline-dynamic-batcher.h
@@ -15,16 +15,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


danpovey:
LGTM

Issue #4518, opened 2021-05-05
Debian10 docker image build fail


mzegla:
Hi,

I'm using debian10-cpu Dockerfile to build docker image with Kaldi - https://github.com/kaldi-asr/kaldi/blob/master/docker/debian10-cpu/Dockerfile

It worked fine before, but when I tried to build it today I got errors:

```
make -C cudadecoder/ depend
make[1]: Entering directory '/opt/kaldi/src/cudadecoder'
rm -f .depend.mk
g++ -M -std=c++14 -I.. -isystem /opt/kaldi/tools/openfst-1.7.2/include -O1 -Wall -Wno-sign-compare -Wno-unused-local-typedefs -Wno-deprecated-declarations -
Winit-self -DKALDI_DOUBLEPRECISION=0 -DHAVE_EXECINFO_H=1 -DHAVE_CXXABI_H -DHAVE_MKL -I/opt/intel/mkl/include -m64 -msse -msse2 -pthread -g -fPIC -Wno-sign-c
ompare batched-threaded-nnet3-cuda-pipeline.cc cuda-decoder.cc decodable-cumatrix.cc cuda-fst.cc batched-static-nnet3.cc batched-threaded-nnet3-cuda-online-
pipeline.cc cuda-online-pipeline-dynamic-batcher.cc lattice-postprocessor.cc batched-threaded-nnet3-cuda-pipeline2.cc >> .depend.mk
batched-threaded-nnet3-cuda-pipeline.cc:19:2: error: #error CUDA support must be configured to compile this library.
 #error CUDA support must be configured to compile this library.
  ^~~~~
batched-threaded-nnet3-cuda-pipeline.cc:26:10: fatal error: nvToolsExt.h: No such file or directory
 #include <nvToolsExt.h>
          ^~~~~~~~~~~~~~
compilation terminated.
cuda-decoder.cc:19:2: error: #error CUDA support must be configured to compile this library.
 #error CUDA support must be configured to compile this library.
  ^~~~~
cuda-decoder.cc:29:10: fatal error: cuda_runtime_api.h: No such file or directory
 #include <cuda_runtime_api.h>
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
cuda-fst.cc:19:2: error: #error CUDA support must be configured to compile this library.
 #error CUDA support must be configured to compile this library.
  ^~~~~
cuda-fst.cc:24:10: fatal error: cuda_runtime_api.h: No such file or directory
[...]

PR #4527, opened 2021-05-12
fix fail Kaldi install on CPU: CUDA error

BadriAhmed:
fix fail Kaldi install on CPU: CUDA support is required to compile this library on issue #4518


danpovey:
Thanks!

PR #4530, opened 2021-05-14
[cmake] Fix build when used via add_subdirectory

funous:
Change all uses of path variables to consistently use CMAKE_CURRENT_*
variables. Previously a combination of CMAKE_* and CMAKE_CURRENT_*
variables was used, which lead to error when Kaldi was added to another
cmake project using add_subdirectory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(kaldi)
 
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-  set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/src CACHE PATH "Install path prefix." FORCE)
+  set(CMAKE_INSTALL_PREFIX ${CMAKE_CURRENT_SOURCE_DIR}/src CACHE PATH "Install path prefix." FORCE)
 endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH}")
 include(GNUInstallDirs)
@@ -156,7 +156,7 @@ if(CUDA_FOUND)
     link_libraries(${NvToolExt_LIBRARIES})
 
     get_third_party(cub)
-    set(CUB_ROOT_DIR "${CMAKE_BINARY_DIR}/cub")
+    set(CUB_ROOT_DIR "${CMAKE_CURRENT_BINARY_DIR}/cub")
     find_package(CUB REQUIRED)
     include_directories(${CUB_INCLUDE_DIR})
 endif()
@@ -170,15 +170,15 @@ if(${KALDI_USE_PATCH_NUMBER})
 endif()
 
 # get_third_party(openfst)
-# set(OPENFST_ROOT_DIR ${CMAKE_BINARY_DIR}/openfst)
+# set(OPENFST_ROOT_DIR ${CMAKE_CURRENT_BINARY_DIR}/openfst)
 # include(third_party/openfst_lib_target)
 find_library(OpenFST_LIBRARY
-            NAMES fst 
-            PATHS ${CMAKE_SOURCE_DIR}/tools/openfst/lib 
+            NAMES fst
+            PATHS ${CMAKE_CURRENT_SOURCE_DIR}/tools/openfst/lib
             REQUIRED)
 find_path(OpenFST_INCLUDE_DIR
-            NAMES "fst/fst.h" 
-            PATHS "${CMAKE_SOURCE_DIR}/tools/openfst/include"
+            NAMES "fst/fst.h"
+            PATHS "${CMAKE_CURRENT_SOURCE_DIR}/tools/openfst/include"
             REQUIRED)
 get_filename_component(OpenFST_LIBRARY_DIR ${OpenFST_LIBRARY} DIRECTORY CACHE)
 mark_as_advanced(FORCE OpenFST_INCLUDE_DIR OpenFST_LIBRARY OpenFST_LIBRARY_DIR)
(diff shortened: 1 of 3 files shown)


kkm000:
Thank you, I think we're on the right track! In fact, Modern CMake pattern recommends detecting whether the project is compiled standalone or as part of the larger project. The differences may be huge: for example, by default it makes more sense to build binaries, but do not do that by default in Kaldi as a subproject (look at the `libfmt`, linked below, how its options are defaulted to `${FMT_MASTER_PROJECT}`. It would also be super-nice in the end to export [interface properties](https://cmake.org/cmake/help/v3.19/command/add_library.html#interface-libraries) with [library aliases](https://cmake.org/cmake/help/v3.19/command/add_library.html#alias-libraries) for the stuff built within Kaldi in this case.

[Here's an approach taken by a small `fmt` library.](https://github.com/fmtlib/fmt/blob/57280762b6e60ef4ddafe6c6807bbd77d4f4e997/CMakeLists.txt#L8-L16)

Now, I'm not asking for everything below from the start; just want to show a general direction. However much or little of this you want to do, it appreciated. Please at the very least introduce a boolean-ish variable we can start on using later, like `Kaldi_IS_SUBPROJECT` (or its negation, whichever fits the best). You nearly did it already. :) It will save you a lot of repeated string comparisons, and CMake is not a pretty language.

Feel free to bump CMake minimum to 3.19. Reliable CUDA support is provided only starting with this version. Installing any version in user's local directory is just unpacking a tarball (or running an .msi installer on Windows, which upgrades the CMake keeping all settings and takes barely 5 minutes to download and install; macOS .dmg is also released with every minor point), so I'm against this "if CMake version >= ..." gymnastics. We'll add a script to ensure the correct CMake, like we always do for all external components.

The rest is an optional, but really great stuff which would be important.

Note how the `fmt` library [carefully separates PRIVATE and INTERFACE properties](https://github.com/fmtlib/fmt/blob/57280762b6e60ef4ddafe6c6807bbd77d4f4e997/CMakeLists.txt#L237-L248). properties on dependencies. This is a very simple concept to grok: PRIVATE include and library dependencies do not propagate to targets using the library, while INTERFACE ones do. If I, as a user, add Kaldi as a subdirectory and simply write `add_executable(myprog main.cc kaldi::kaldi_libs)`:
 * I don't want to automatically acquire `-DHAVE_EXECINFO_H` and other internal compiler defines while compiling `myprog`, as they are used exclusively by non-inlined code inside Kaldi objects, and suppose I do not care.
 * but I do very likely want `-DHAVE_MKL` and `-DHAVE_CUDA` in my build, since they are needed both for Kaldi headers to properly compile in my project and also give me useful info, as I'll use these toolkit, and I'll use the fact that Kaldi has already located them for me. These are INTERFACE compiled defines.

PUBLIC is a simply a shortcut (think of `PUBLIC = PRIVATE | INTERFACE;`) to set _things_ to be both PRIVATE (needed to compile Kaldi libs) and INTERFACE (needed to compile other stuff including Kaldi via a subdirectory) the same time; one rarely needs explicit INTERFACEs. E. g., `HAVE_MKL` is the compiler macro definition needed by both Kaldi code itself _and_ by its users, so it's `PRIVATE | INTERFACE => PUBLIC`. The _things_ are compile defines and include directories, libraries and their locations—basically, anything you tell the targets to both successfully build and keep for their users as dependencies. For example, if Kaldi is compiled with MKL, the same simple one-liner `add_executable(myprog main.cc kaldi::kaldi_libs)` will set my correct compilation flags to locate MKL includes (probably as `-isystem <dir>`, library directories via `-L`, and will add `-l....` for each one I indirectly need for the whole executable to work; the same goes for OpenFST, and even CUDA. Behold the power of modern CMake: this is a _great_ way to make the life of the Kaldi-as-a-library consumer much easier! No more finding out all the variables and painfully stuffing them into my dependencies:  `${Kaldi_BASE_LIBRARY}`, `${Kaldi_BASE_LIBDIR}`, and so on—all 17 of them, a the the very least.

The ALIAS library is a conventional way, a good pattern to export a library or set of libraries from a subdirectory project: it's all the same as the one it aliases, _except_ its properties are sealed and read-only.

[Here's a more complex, but thoroughly documented example](http://tudat.tudelft.nl/developerGuide/applicationCMakeLists/applicationCMakeLists.html) from a significantly more complex project than Kaldi, the TU Delft astrodynamic toolbox. I dunno what it does, but it's indeed a huge project.

funous:
Thanks for the reply. In this PR I am not really aiming to fix all the issues with using Kaldi as a subproject, mostly just the bare minimum to make it usable in at least some cases. All the things you mention are definitely relevant, but I feel it asks for much bigger changes than the scope of my PR. It really just fixes one issue I am having right now, I'm afraid I won't have the time to improve this much more, apart from minor fixes. I think it makes sense to raise a general issue with using Kaldi as a cmake subproject, and this PR could be a step in that direction. The information in your comment would be a great start I think, as it does cover a lot of the work that'd be nice to do. I am happy for this PR to be postponed until the whole set of improvements is done, however I think it can be used independently as a small step towards that goal. What do you think about this approach?

With these changes (or similar changes on an older commit to be precise), I am able to use Kaldi via add_subdirectory. I still have to do a few tweaks in the parent project (I'm using MKL, so I have to set `MATHLIB` to MKL, export `MKLROOT` and use `include_directories(<path-to-mkl-headers>`). Then if I use `add_subdirectory` with `EXCLUDE_FROM_ALL`, I am able to use things like `target_link_libraries(my-target kaldi-nnet3)`. I do list all the dependencies of kaldi-nnet3 explicitly (other `kaldi-*` targets), I am not sure if it's actually needed on latest master of Kaldi. However in the end, I am able to compile `my-target` without building any Kaldi executables, just the necessary object files for the Kaldi libraries it's using. I could try to provide a minimal example of this to help anyone trying to set-up something similar. In my case I don't mind leaking things like `-DHAVE_EXECINFO_H` into the parent build, perhaps that makes my solution quite simplistic.

> Please at the very least introduce a boolean-ish variable we can start on using later, like Kaldi_IS_SUBPROJECT (or its negation, whichever fits the best). You nearly did it already. :) It will save you a lot of repeated string comparisons, and CMake is not a pretty language.

I can introduce it, but it'd have no use currently. With my changes, using Kaldi as the top-level project is unchanged, as in such case `CMAKE_CURRENT_SOURCE_DIR` has the same value as `CMAKE_SOURCE_DIR` (same goes for BINARY_DIR). Unless I'm missing something, my changes shouldn't change the current behaviour at all when Kaldi is not used as a subproject. But it should improve the behaviour when it is used as one. I am not introducing any conditionals, so I am not sure what the variable would be used for.

danpovey:
@kkm000 is there any way we can merge these and leave the other issues for a later PR? 
I just want to make sure we're making forward progress..

kkm000:
Sure, let's do that. Thanks @funous for the change, Dan is right, any progress is beneficial.

PR #4535, opened 2021-05-19
Sprakbanken swe fix

giampierosalvi:
The changes in this branch make the sprakbanken_swe recipe run. This is still not the version that was used to obtain the results reported in the RESULTS file. It obtains on average about 5 percent points lower accuracy that what reported. This may be due to a large number of missing words in the dictionary, mostly due to misspelling in the transcriptions of the data set. Also, the scripts seem to optimize insertion penalty and language weight to the same test data that is used for the final results, which is not correct.


danpovey:
Sorry that we didn't respond to this earlier.
This may not be perfect but I think it's OK to merge as-is rather than bothering you with changes.. I'll merge.

PR #4539, opened 2021-05-25
[src] thread-pool-light.h: Change shared_ptr to weak_ptr to break memory cycle

galv:
ThreadPoolLightWorker has a cycle of strong references to implement its "work stealing functionality".

This causes a memory leak.

Testing: Apply this change in thread-pool-light.h

```
virtual ~ThreadPoolLightWorker() { KALDI_LOG << "GALVEZ:~ThreadPoolLightWorker() for thread " << std::this_thread::get_id(); Stop(); }
```

With this PR, you will see this:

```
LOG (batched-wav-nnet3-cuda2[5.5.930~2-ca99]:~ThreadPoolLightWorker():cudadecoder/thread-pool-light.h:122) GALVEZ:~ThreadPoolLightWorker() for thread 140704129964672
LOG (batched-wav-nnet3-cuda2[5.5.930~2-ca99]:~ThreadPoolLightWorker():cudadecoder/thread-pool-light.h:122) GALVEZ:~ThreadPoolLightWorker() for thread 140704129964672
```

Without this PR, you won't see that information print out. So that
proves that the destructor of ThreadPoolLightWorker never executes
(because of the stale reference due to the cycle).

I messed with using -fsanitize=leak on GCC7.5 to try to prove the
existence of this memory leak (to prevent such errors in the future),
but it doesn't appear to be smart enough to catch the bug. Unfortunately, 
I do not know a good way to unit test these sorts of issues.

### DIFF
--- a/src/cudadecoder/thread-pool-light.h
+++ b/src/cudadecoder/thread-pool-light.h
@@ -1,7 +1,7 @@
 // cudadecoder/cuda-decoder.h
 //
 // Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
-// Hugo Braun, Justin Luitjens, Ryan Leary
+// Hugo Braun, Justin Luitjens, Ryan Leary, Daniel Galvez
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


galv:
@kkm000 here you go

FYI @hugovbraun

kkm000:
LGTM, thanks!

PR #4540, opened 2021-05-26
Consistent static/share external dependency

rickychanhoyin:
Consistent static/share external dependency


src/configure
```diff
@@ -1278,7 +1287,7 @@ or try another math library, e.g. --mathlib=OPENBLAS (Kaldi may be slower)."
     failure "Unsupported linear algebra library '$MATHLIB'"
   fi
   $use_cuda && configure_cuda
-  linux_configure_speex
+  linux_configure_speex $static_speex
```
kkm000, line 1290: I do not understand this change. Since you are also [removing quite ugly lines 410-413](https://github.com/kaldi-asr/kaldi/pull/4540/files#diff-f1a9b6510d3d745ebd9aae86235d5fc93750e53f24bbc5042e62ab15584c0f02L410-L413), the function `linux_configure_speex` no longer uses $1, or any other argument references. Probably not required?

kkm000 (changes requested):
Thanks, this is a useful option, and your handling of its value is in line with the rest of configure's library static/shared options!

src/configure
rickychanhoyin: lines 410-413 is not necessary. line1290 in the new file pass directly as argument

kkm000: What I mean is the function `linux_configure_speex` now does not take arguments. It simply uses the global value of `$static_speex`. You do not need to pass it to the function. The addition of an argument here is unnecessary. It is ignored anyway. Otherwise, the change is totally LGTM!

rickychanhoyin: yes, you are right. Can you help modify it such that this  thread can be closed ? Thanks

kkm000: Of course. Wondering why you couldn't. GitHub lets edit files right on the web.

kkm000 (approved):

kkm000:
LGTM. Goes in as soon as CI passes tests.

PR #4543, opened 2021-05-27
chain/align_lats.sh, add execute permission 'x'

vesis84:


vesis84:
did it have the premission missing on prupose (deprecated), or was it a bug ?

kkm000:
Of course, that was a bug. LGTM, thanks!

PR #4545, opened 2021-05-28
`rm -r $dest` fail when $dest not exists

kimdwkimdw:


egs/wsj/s5/utils/combine_data.sh
```diff
@@ -30,7 +30,7 @@ shift;
 
 first_src=$1;
 
-rm -r $dest 2>/dev/null
+rm -rf $dest 2>/dev/null
```
danpovey, line 33: Rather than adding `-f`, which can be a little dangerous, overriding user preferences, I prefer to add "|| true" to such statements.  
Note: this statement would only have caused a problem if you had done
`set -e`
for instance, in your .bash_profile.

kimdwkimdw: Thanks! I'll add only `|| true`

kimdwkimdw: I've checked with `set -e`, `set -u`, `set -o pipefail`

Issue #4542, opened 2021-05-27
Write sctd++ allocator for std containers in device memory


kkm000:
### I want to:
1. Unsee this:
```c++
  StateId* d_arc_nextstates_;
  d_arc_nextstates_ = static_cast<StateId *>(
      CuDevice::Instantiate().Malloc(arc_count_ * sizeof(*d_arc_nextstates_)));
```
2. Never spend another day hunting for a missing call to this:
```
  KALDI_ASSERT(d_arc_nextstates_ &&
               "Please call CudaFst::Initialize() before calling Finalize()");
  CuDevice::Instantiate().Free(d_arc_nextstates_);
```

### Instead, I want simply this:
```
std::vector<StateId, CuDevice::Allocator> d_arc_nextstates_;
```

----
@hugovbraun, @danpovey, are you with me? Do you see any downsides? The guarantee of memory contiguity was added to `std::vector` spec starting with C++17, but I know of no implementation that would chunk `vector`s internally for any C++ language version.

danpovey:
We have something- not often used- called Array, which does something a
little like that.
I'm pretty sure it's a bad idea to attempt to overload standard library
things for GPU, because they have so much going on, and the contents must
be assumed to live on CPU.

kkm000:
@danpovey, it has nothing to do with "overloading standard library." It's the interface designed _exactly_ for this use case. The template for vector has the second parameter which is defaulted to the default allocator. There are other allocators available, for example, a thread-aware allocator in TBB. Ours is a perfect use case for an allocator.

danpovey:
But std::vector has many functions that would break if you were to do
that.  Like back(), operator [], and so on.
The majority of them would crash.

kkm000:
You mean an attempt to dereference the pointers. Not `back()` per se, rather what you do with it. Sure. But that's the case with raw cudaMalloc'd pointers, too. I'm not saying that `push_back()` would work. Just using the std::vector as a container that guarantees  alignment, deallocation on destruction, and not mixing host/device allocs, [as we just did](https://github.com/kaldi-asr/kaldi/pull/4517). Not a fully-functional vector, of course. We have the CuVector for that. Too bad stdc++ does not define a more blobsy container--that would be ideal.

BTW, I think that craftily crafted definitions of allocator's `pointer` and `reference` types may render these methods unusable at compile time.

[...]

PR #4548, opened 2021-05-28
[src] Fix leaking CudaFst object's device memory

kkm000:
`BatchedThreadedNnet3CudaOnlinePipeline` never freed a `CudaFst` object which it initialized. Freeing it caused an error returned from `::cudaFree` (with our allocator disabled).

The root cause of this problem was copying of the `CudaFst` in `CudaDecoder`'s constructor. I rewrote it so that `CudaDecoder` only stores a reference to a `CudaFst` instance, and the object itself is uniquely owned by the pipeline (as `CudaDecoder` already is). 

This allowed folding of separate `CudaFst::Initialize` and `CudaFst::Finalize` methods into its respective constructor and destructor.

Then I added a proof-of-concept `unique_device_ptr`, a specialization of `std::unique_ptr` for device-allocated memory. The type is declared privately in, and only used by `CudaFst`. The `CudaFst::Finalize` had been removed entirely, and the memory is freed in the object's destructor, as members are destroyed. The specialized smart pointer is defined as 

```c++
  template<typename T>
  struct CuDeleter {
    constexpr CuDeleter() noexcept = default;
    void operator()(T* ptr) const { CuDevice::Instantiate().Free(ptr); }
  };

  /// A uniquely owned, movable pointer to device-allocated memory.
  template<typename T>
  using unique_device_ptr = std::unique_ptr<T, CuDeleter<T>>;
```

Technically this does not provide any safety against e.g. dereferencing the device pointer on host, only controls the allocation lifetime. But my two-day leak hunting safari suggests that this thing alone is quite useful.

I think we can also minimize ugly alloc size computation by adding members to CuDevice, Malloc-like, but templated on an element type and taking the number of elements as an argument, and returning `unique_device_ptr`. (A raw pointer is detachable with unique_device_ptr::release(), but that's explicit and more readable).

@danpovey, @jtrmal, please comment on this approach.

My hands are itchy for a single-member `device_memptr` structure, not mutually implicitly convertible with a regular pointer. How many different ::cudaXXX() functions which take pointers do we use? No more than a dozen, I think. We could wrap them so that they take raw pointers for host, `device_memptr`s for device memory. And if the thing also knew it was pinned and unregistered itself on destruction of the wrapper, I'd jump over the moon. In short, I want to make writing incorrect CUDA host-side code as hard as possible. :)

Other minor cleanup:
* Match the types in `CudaFst` to those in `monsterstruct DeviceParams` (like `int` vs `int32`), and use explicitly sized integer types generally.
* Use safe pointers for `KernelParams` and `DeviceParams` instead of new/delete.
* Move definitions of Arc, Weight, Label and StateId into `CudaFst` and out of the namespace scope, where they probably ended up by mistake.
* Doxygenate some comments, very non-systematically, rather only those I had to touch.

@hugovbraun, @galv, PTAL.

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
@@ -90,9 +90,8 @@ void BatchedThreadedNnet3CudaOnlinePipeline::AllocateAndInitializeData(
     feature_pipelines_.resize(config_.num_channels);
   }
 
-  // Decoder
-  cuda_fst_ = std::make_shared<CudaFst>();
-  cuda_fst_->Initialize(decode_fst, trans_model_);
+  // Decoder.
+  cuda_fst_ = std::make_unique<CudaFst>(decode_fst, trans_model_);
   cuda_decoder_.reset(new CudaDecoder(*cuda_fst_, config_.decoder_opts,
                                       max_batch_size_, config_.num_channels));
   if (config_.num_decoder_copy_threads > 0) {
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
@@ -413,7 +413,7 @@ class BatchedThreadedNnet3CudaOnlinePipeline {
   // HCLG graph : CudaFst object is a host object, but contains
   // data stored in
   // GPU memory
-  std::shared_ptr<CudaFst> cuda_fst_;
+  std::unique_ptr<CudaFst> cuda_fst_;
   std::unique_ptr<CudaDecoder> cuda_decoder_;
 
   std::unique_ptr<ThreadPoolLight> thread_pool_;
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-pipeline.cc
@@ -23,8 +23,13 @@
 #define SLEEP_BACKOFF_S ((double)SLEEP_BACKOFF_NS / 1e9)
 
 #include "cudadecoder/batched-threaded-nnet3-cuda-pipeline.h"
+
+#include <memory>
+
 #include <nvToolsExt.h>
+
 #include "base/kaldi-utils.h"
+#include "cudadecoder/cuda-fst.h"
 
 // This pipeline is deprecated and will be removed. Please switch to
 // batched-threaded-nnet3-cuda-pipeline2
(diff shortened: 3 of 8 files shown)


hugovbraun:
Looks way cleaner, thanks! The unique_device_ptr is a great idea. I also agree that using a templated version of malloc with a count is better (readibility, avoid cast/overflow issues, etc.)

For the cuda_fst, do you think we should use a ref in the cudadecoder or a shared_ptr ?

hugovbraun:
@kkm000 did you notice any other memory leak or that was it ?

kkm000:
@hugovbraun,

> did you notice any other memory leak or that was it ?

I was debugging a huge leaks in connection with pipeline creation and destruction, and cuda-memcheck found 6 large leaks per cycle; finding the leaking object was a no-brainer to spot. Speaking of only create-destroy cycle of the pipeline and the dynamic batcher, I'm sure these are all there were.

I'm now debugging a weird memory corruption problem, but I'm not sure if it's my code at fault or not. It's a very weird behavior: exactly the third constructed pipeline crashes. I added `assert`s at the point where ASAN found the out-of-bound `vector` access, and found an inexplicable weirdness.

Maybe you can recall something similar from debugging this code? `CudaDecoder::AddFinalTokensToLattice` is called for the first time after adding the initial state to the lattice FST. At this point inside it:

https://github.com/kaldi-asr/kaldi/blob/2d57eb92c6c7f2670ac09c2fa19e93b9ba004b1f/src/cudadecoder/cuda-decoder.cc#L1284-L1297

the `offset` pulled at line 1293 has an absolutely insane value, over 250000. I am still trying to track down how it might have gotten there.

In any case, I'll certainly cuda-memcheck everything for leaks when it works.

> For the cuda_fst, do you think we should use a ref in the cudadecoder or a shared_ptr?

Do you see a use case for a single `CudaFst` shared by an unknown number of multiple `CudaDecoder`s, which are hard to account for otherwise? In the current implementation, both objects are uniquely owned by the pipeline instance, so to me, there is nothing wrong with giving one a pointer or a reference to another. This pattern is just everywhere in Kaldi. 

In general, I always prefer unique ownership, like is our case. In my rules of thumb book, `shared_ptr` is a solution only if an object has multiple independent uncoordinated owners. In our case, there is a clear single owner, so I would not use a `shared_ptr`. Think of the recent @galv's fix to the thread-local pools holding shared pointers to each other :) Besides, `shared_ptr` is a more heavy-weight, heap-allocating object, and its rules w.r.t completeness of the argument type at the point of definition, and even the exact type whose destructor is called if its template parameter is a base not equal to the more derived type of the object pointer which it actually owns are different, compared to the `unique_ptr`. I'm using them where they belong, but trying to stay away from if I can.

kkm000:
@danpovey, what are your thoughts?  Additional templated `Malloc()` methods to allocate typed array-sized device memory chunk and return it wrapped into the `unique_device_ptr`? Looks seriously cleaner to me cleaner than what I rather hastily proposed #4542, and w/o the drawbacks you highlighted there (which I agree with, having thought about them for a while—too much semantic abuse). Like

```c++
class CudaFst {
  unique_device_ptr<int32> d_some_array_;

  CudaFst(size_t n_elements) {
    d_some_array_ = CuDevice::Instance()::Malloc<int32>(n_elements);
  }
};
```
Plus wrappers like this for other cudaAlloc variants used around the CuAllocator.

PR #4544, opened 2021-05-27
Fix make depend without cuda

nshmyrev:
Currently `make depend` without cuda is broken. Originally broken in 5de039aeaacb90241d0362f113eeb88c5c46943d 

```
shmyrev@kappa ~/t/kaldi/src/cudadecoderbin $ make depend
rm -f .depend.mk
g++ -M -std=c++14 -I.. -isystem /home/shmyrev/t/kaldi/tools/openfst-1.7.2/include -O1  -Wall -Wno-sign-compare -Wno-unused-local-typedefs -Wno-deprecated-declarations -Winit-self -DKALDI_DOUBLEPRECISION=0 -DHAVE_EXECINFO_H=1 -DHAVE_CXXABI_H -DHAVE_ATLAS -I/home/shmyrev/t/kaldi/tools/ATLAS_headers/include -msse -msse2 -pthread -g batched-wav-nnet3-cuda2.cc batched-wav-nnet3-cuda.cc batched-wav-nnet3-cuda-online.cc >> .depend.mk
batched-wav-nnet3-cuda-online.cc:23:2: error: #error CUDA support must be configured to compile this binary.
   23 | #error CUDA support must be configured to compile this binary.
      |  ^~~~~
batched-wav-nnet3-cuda-online.cc:31:10: fatal error: cuda.h: No such file or directory
   31 | #include <cuda.h>
      |          ^~~~~~~~
compilation terminated.
../makefiles/default_rules.mk:149: recipe for target 'depend' failed
make: *** [depend] Error 1
```

PR #4551, opened 2021-05-29
[make] Fix 'make depend' without CUDA

kkm000:
1. Order `SUBDIRS` alphabetically.
2. Obtain `MEMTESTDIRS` by set subtraction: there are only 3 members of `SUBDIRS` not in it.
3. Add a phony 'libs' target to build libs only. Just in case.
4. Abbreviate a common dependency 'BMU = case matrix util', and order all directory dependencies of each target alphabetically (near the end of `src/Makefile`; e.g., `decoder: $(BMU) fstext gmm hmm lat transform tree`). There are still too many extra declared (they are transitive, after all), but freeze me in hell if I want to untangle it. I'll better save time to transition the build to CMake.
5. Use `true` and `false` as values for the variable `with_cudadecoder` in `configure`, as all other variables do, not 0 and 1. It used to be unlike all others. Change Makefiles in related subdirs to use the new value.
6. Bump `CONFIGURE_VERSION` to 14 because stale `kaldi.mk` will have 0 or 1.
7. Never set `WITH_CUDADECODER = true` in `kaldi.mk` if `CUDA` is `false`. The previous default disrespected the detection of CUDA, and this was the root cause of issue #4544.
8. Make the `depend` target phony.
9. Demote `sgmm2` and `sgmm2bin` to EXTRA_SUBDIRS. They aren't used in egs.

Testing, ad-hock script, all pass: 

```bash
#!/bin/bash

# Place this into the kaldi root, lest 'git clean' munches it, the test_all.log and yourself!

set -u

rm -f test_all.log
exec &> >(tee -a test_all.log)

separator() {
  for i in {1..3}; do
    echo "################################################################################"
  done
}

cd src
pwd

export CXX=clang++-11 EXTRA_CXXFLAGS='-fcolor-diagnostics -Wall -Wpedantic' LDFLAGS='-fuse-ld=lld-11'

# No CUDA.
time (set -x
      git clean . -fx &&
        ./configure --use-cuda=no --shared &&
        make -j20 depend &&
        make -j12
      printf '\n***Exit code: %d\n' $?)
find -name .depend.mk | sort
separator

# CUDA, no cudadecoder.
time (set -x
      git clean . -fx &&
        ./configure --use-cuda=yes --cuda-arch=-arch=sm_75 --with-cudadecoder=no --shared &&
        make -j20 depend &&
        make -j12
      printf '\n***Exit code: %d\n' $?)
find -name .depend.mk | sort
separator

# CUDA, with cudadecoder.
time (set -x
      git clean . -fx &&
        ./configure --use-cuda=yes --cuda-arch=-arch=sm_75 --with-cudadecoder=yes --shared &&
        make -j20 depend &&
        make -j12
      printf '\n***Exit code: %d\n' $?)
find -name .depend.mk | sort
```

CC: @nshmyrev 
X-Ref: supersedes #4544. 
Close: #4544

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -4,30 +4,40 @@
 
 SHELL := /bin/bash
 
-SUBDIRS = base matrix util feat cudafeat tree gmm transform \
-          fstext hmm lm decoder lat kws cudamatrix nnet \
-          bin fstbin gmmbin fgmmbin featbin cudafeatbin \
-          nnetbin latbin sgmm2 sgmm2bin nnet2 nnet3 rnnlm chain nnet3bin nnet2bin kwsbin \
-          ivector ivectorbin online2 online2bin lmbin chainbin rnnlmbin \
-          cudadecoder cudadecoderbin
+# Please keep sorted alphabetically, and start each with a letter which
+# is different from the first one in the last word one the row above it.
+SUBDIRS := base bin chain chainbin cudamatrix decoder		\
+           feat featbin fgmmbin fstbin fstext			\
+           gmm gmmbin hmm                               	\
+           ivector ivectorbin kws kwsbin	        	\
+           lat latbin lm lmbin matrix				\
+           nnet nnetbin nnet2 nnet2bin nnet3 nnet3bin		\
+           online2 online2bin rnnlm rnnlmbin            	\
+           transform tree util
 
-MEMTESTDIRS = base matrix util feat cudafeat tree gmm transform \
-          fstext hmm lm decoder lat nnet kws chain \
-          bin fstbin gmmbin fgmmbin featbin cudafeatbin \
-          nnetbin latbin sgmm2 nnet2 nnet3 rnnlm nnet2bin nnet3bin sgmm2bin kwsbin \
-          ivector ivectorbin online2 online2bin lmbin
+# Will also build these if configured with --with-cudadecoder
+# (default 'true' if CUDA is used, else 'false')
+CUDADECODER = cudafeat cudafeatbin cudadecoder cudadecoderbin
 
+MEMTESTDIRS = $(filter-out chainbin cudamatrix rnnlmbin $(SUBDIRS))
 CUDAMEMTESTDIR = cudamatrix
 
-SUBDIRS_LIB = $(filter-out %bin, $(SUBDIRS))
-
-KALDI_SONAME ?= libkaldi.so
-
 # Optional subdirectories
-EXT_SUBDIRS = online onlinebin  # python-kaldi-decoding
+EXT_SUBDIRS := online onlinebin sgmm2 sgmm2bin  # python-kaldi-decoding
 EXT_SUBDIRS_LIB = $(filter-out %bin, $(EXT_SUBDIRS))
 
 include kaldi.mk
+
+ifeq ($(CUDA), true)
+ifeq ($(WITH_CUDADECODER), true)
+SUBDIRS += $(CUDADECODER)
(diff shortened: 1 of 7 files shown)


src/Makefile
```diff
@@ -130,52 +140,60 @@ depend: $(addsuffix /depend, $(SUBDIRS))
...
-# python-kaldi-decoding: base matrix util feat tree gmm transform sgmm2 fstext hmm decoder lat online
-cudafeat: base matrix util gmm transform tree feat cudamatrix online2
-cudafeatbin: base matrix util gmm transform tree feat cudamatrix cudafeat online2
-online: decoder gmm transform feat matrix util base lat hmm tree
-online2: decoder gmm transform feat matrix util base lat hmm tree ivector cudamatrix nnet2 nnet3 chain
-kws: base util hmm tree matrix lat
-cudadecoder:  cudamatrix cudafeat online2 nnet3 ivector feat fstext lat chain transform
-cudadecoderbin: cudadecoder cudafeat cudamatrix online2 nnet3 ivector feat fstext lat chain transform
+util:   base matrix
+feat:      $(BMU) gmm transform tree
```
danpovey, line 172: This change might affect misc/maintenance/find_missing_dependencies.sh and reorder_addlibs.sh, not many people know about those scripts so it's not a big deal but just FYI.

src/Makefile
```diff
@@ -130,52 +140,60 @@ depend: $(addsuffix /depend, $(SUBDIRS))
 ext_depend: check_portaudio
 	-for x in $(EXT_SUBDIRS); do $(MAKE) -C $$x depend; done
 
+# 'make libs' to skip binaries, and build only libraries.
+.PHONY: libs $(SUBDIRS_LIBS)
```
danpovey, line 144: SUBDIRS_LIBS only appears these two times, elsewhere it's SUBDIRS_LIB; you didn't mean SUBDIRS_LIB?  same with SUBDIRS_BINS?

kkm000: Done. Thanks fro spotting.

src/Makefile
kkm000: Ugh. As soon as I get a breather, I'll rewrite the build to use CMake. What's interesting, CMake itself warns you if you give it too many depedendencies that left unused. It also has a control to define how many times give a linker static libraries, so reordering is also not an issue (-llib1 -llib2 -llib1 -llib2 etc).

Fixing things like this one is a real PITA, and the more complex it gets, the more maintenance pain it is. The current effort started with a rewrite of everything top-down. I'd rather start bottom-up: Kaldi main codebase first, and then all the bells and whistles like automatically downloading dependencies etc.

CMake 2 used to be ugly. CMake post 3.14 ("dubbed "the modern CMake") is very different. And I compile CUDA code with CMake 3.19 natively (still need to call a package to find CUDA libraries, but that's true for C++ too), but NVCC is invoked with all defined flags, uses the defined C++ compiler etc.) And ninja is a much better build system for CMake: instead of a ton of unreadable Makefiles, it generates one huge unreadable file for ninja :) Surprisingly, ninja works even on Windows, too.

And CMake can be installed anywhere by untarring a tar, so we can just unpack any version we rely on in tools/

I tried many build systems, and the main take-away is that build is hard. Not that I like CMake, it has its own twists too, but currently it's probably the least painful and ugly build system.

kkm000:
@danpovey, @jtrmal, PTAL.

danpovey:
Good with me too-- thanks!

Issue #4529, opened 2021-05-13
[CMake] Can't link Openfst library


vzxxbacq:
I'm developing a gst plugin with kaldi cmake, but I couldn't link openfst when I build project. So I remove CACHE option in get_filename_component in CMakeLists.txt line 185 as follow:
```
get_filename_component(OpenFST_LIBRARY_DIR ${OpenFST_LIBRARY} DIRECTORY CACHE)
message("OpenFST_LIBRARY: ${OpenFST_LIBRARY}")
message("OpenFST_LIBRARY_DIR: ${OpenFST_LIBRARY_DIR}  (with CACHE option)")
get_filename_component(OpenFST_LIBRARY_DIR ${OpenFST_LIBRARY} DIRECTORY)
message("OpenFST_LIBRARY_DIR: ${OpenFST_LIBRARY_DIR}  (without CACHE option)")
```
And the output is:
```
OpenFST_LIBRARY: /home/laptop/code/kaldi/tools/openfst/lib/libfst.so
OpenFST_LIBRARY_DIR:   (with CACHE option)
OpenFST_LIBRARY_DIR: /home/laptop/code/kaldi/tools/openfst/lib (without CACHE option)
```
I'm not sure this is a bug of kaldi, so I put the linking part in the CMakeLists.txt of custom gst plugin
```
set(GST_KALDI_DEPS kaldi-online2 kaldi-lat kaldi-decoder kaldi-feat kaldi-transform
        kaldi-gmm kaldi-hmm kaldi-tree kaldi-matrix  kaldi-util kaldi-base kaldi-lm
        kaldi-nnet2 kaldi-nnet3 kaldi-cudamatrix kaldi-ivector kaldi-fstext kaldi-chain)

target_link_libraries(gst-plugin-test
        ${GST_KALDI_DEPS}
        ${GST_LIBRARIES})
```

kkm000:
The CMake support is very experimental, and has been written in an old, 2.x style. I'm afraid we cannot support it in general, and more so under Linux. It had been written long ago as a proof-of-concept. And we've all but dropped support for the gst plugin.

Under Linux, I suggest going normal `configure` and `make` route. You need to enable extra components in `configure`, IIRC, to include the login into compilation. It won't probably compile cleanly; if you come up with sensible, quality patches, please send a PR.

[...]

PR #4558, opened 2021-06-08
Using FetchContent_Declare to automatically setup OpenFST at configure time.

vzxxbacq:
fix https://github.com/kaldi-asr/kaldi/issues/4529

### DIFF
--- a/cmake/third_party/openfst.cmake
+++ b/cmake/third_party/openfst.cmake
@@ -1,14 +1,27 @@
-cmake_minimum_required(VERSION 2.8.2)
-project(openfst-download NONE)
+cmake_minimum_required(VERSION 3.14)
+include(FetchContent)
 
-include(ExternalProject)
-ExternalProject_Add(openfst
-    GIT_REPOSITORY https://github.com/kkm000/openfst
-    GIT_TAG 0bca6e76d24647427356dc242b0adbf3b5f1a8d9 # tag win/1.7.2.1
-    SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/openfst"
-    BINARY_DIR ""
-    CONFIGURE_COMMAND ""
-    BUILD_COMMAND ""
-    INSTALL_COMMAND ""
-    TEST_COMMAND ""
+FetchContent_Declare(
+        openfst
+        # GIT_REPOSITORY  https://github.com/kkm000/openfst
+        # GIT_TAG         0bca6e76d24647427356dc242b0adbf3b5f1a8d9 # tag win/1.7.2.1
+        GIT_REPOSITORY  https://github.com/vzxxbacq/openfst
+        GIT_TAG 1b5a9c4bbeb4bd09861f3d9c5cf4cc39e2f8b39c  # temp address, fixed cmake path error
 )
+
+FetchContent_GetProperties(openfst)
+if(NOT openfst_POPULATED)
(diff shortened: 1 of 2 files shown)


kkm000:
Looks like a definite improvement, thanks! Does it build binaries as well? And did you test it on Linux, too?

Some recipes require FAR extensions of OpenFST (the configure switches in our tools/Makefile aren't exactly correct, I suspect). But this is not a show-stopper, we can take care of that next.

cmake/third_party/openfst.cmake
```diff
@@ -1,14 +1,25 @@
...
-    BUILD_COMMAND ""
-    INSTALL_COMMAND ""
-    TEST_COMMAND ""
+cmake_minimum_required(VERSION 3.14)
+include(FetchContent)
+
+FetchContent_Declare(
+        openfst
+        GIT_REPOSITORY  https://github.com/kkm000/openfst
+        GIT_TAG         338225416178ac36b8002d70387f5556e44c8d05 # tag win/1.7.2.1
```
danpovey, line 7: tag win/1.7.2.1... does this mean it's specifically for Windows?

vzxxbacq: the comment , I test on ubuntu and WSL(arch)

> Looks like a definite improvement, thanks! Does it build binaries as well? And did you test it on Linux, too?
> 
> Some recipes require FAR extensions of OpenFST (the configure switches in our tools/Makefile aren't exactly correct, I suspect). But this is not a show-stopper, we can take care of that next.

I have built the project on ubuntu and WSL (arch) using ninja build. The configure switches depend on the CMakeLists.txt in the OpenFST project.

vzxxbacq:
@kkm000  
I have built the project on ubuntu and WSL (arch) using ninja build. The configure switches depend on the CMakeLists.txt in the OpenFST project.

cmake/third_party/openfst.cmake
vzxxbacq: The tag is the newest commit of the winport branch not specifically for Windows, I forgot to remove the old comment, sorry.

danpovey:
I'm OK to merge it as is if it would work.

kkm000:
> The configure switches depend on the CMakeLists.txt in the OpenFST project.

Yup, and that is kind of the gray areas of CMake. There is no generally accepted best practices for controlling the subproject's publicly exposed controls. The "strong-arm" approach is to set all (or, at the least, all required) exposed variables before including the subdirectory. This may be confusing, because project's cached options that you see in `ccmake` or the GUI are not those which are applied during build, and changes that the user applies to the cache have no effect. Another is to test if the variables are cached, and seed them only if they aren't. Sometimes, a combination of the two is necessary. 

But I digress here. I just want to be sure we build what is necessary for Kaldi _by default_. In other words, if the user just runs `cmake -B build_dir && cmake --build build_dir`. For OpenFST, this means FAR extensions and all corresponding binaries. We just need to make sure these are build in thise very default case. Let's commit this change so it does not blow up, and then look into this requirement and make sure it's satisfied.

For Kaldi proper, only the base libfst is required as a dependency; base FST binaries are obviously needed for all egs, and some call for more than the basic set. I think that the only extension required is FAR (binaries, not libraries: if Kaldi is built without binaries—there's a CMake option for that, and I think it's a good idea to have this option—then no extensions are needed).

PR #4560, opened 2021-06-12
[src] Remove unused class members

kkm000:
I'm trying to take care of compiler warnings about unused class data members. In a few cases, it's obvious. In most cases, however, the variables should probably have been used. This is WIP; I did not completely remove everything. In some cases, I commented out the variables and their initialization; in some, I did but left in constructor parameters that used to initialize them. It should be one way or another—either completely remove them, including extra constructor arguments, or figure out why are they unused, restore them and fix the bigger problem. `embedding_config_` in rnnlm-training.* is especially suspicious.

@danpovey, @hugovbraun, need your help with that. PTAL at the parts of code you know, and let me know which way should I proceed with them. Please mark only those than need to be changed; let's by default assume that the changes which are not commented on are good to go.

### DIFF
--- a/src/online2/online-nnet3-wake-word-faster-decoder.h
+++ b/src/online2/online-nnet3-wake-word-faster-decoder.h
@@ -30,12 +30,9 @@
 
 namespace kaldi {
 
-// Extends the definition of FasterDecoder's options to include additional
-// parameters.
+/// Extends the FasterDecoder's options with additional parameters
+/// (currently none).
 struct OnlineWakeWordFasterDecoderOpts : public FasterDecoderOptions {
-
-  OnlineWakeWordFasterDecoderOpts() {}
-
   void Register(OptionsItf *opts, bool full) {
     FasterDecoderOptions::Register(opts, full);
   }
@@ -52,9 +49,8 @@ struct OnlineWakeWordFasterDecoderOpts : public FasterDecoderOptions {
 class OnlineWakeWordFasterDecoder : public FasterDecoder {
  public:
   OnlineWakeWordFasterDecoder(const fst::Fst<fst::StdArc> &fst,
-                              const OnlineWakeWordFasterDecoderOpts &opts,
-                              const TransitionModel &trans_model)
-      : FasterDecoder(fst, opts), opts_(opts), trans_model_(trans_model) {}
+                              const OnlineWakeWordFasterDecoderOpts &opts)
+      : FasterDecoder(fst, opts), opts_(opts) {}
 
   // Makes a linear graph, by tracing back from the last "immortal" token
   // to the previous one
@@ -69,7 +65,6 @@ class OnlineWakeWordFasterDecoder : public FasterDecoder {
   // and then (possibly multiple times) AdvanceDecoding().
   void InitDecoding();
 
-  
  private:
   // Returns a linear fst by tracing back the last N frames, beginning
   // from the best current token
@@ -84,7 +79,6 @@ class OnlineWakeWordFasterDecoder : public FasterDecoder {
   void UpdateImmortalToken();
 
   const OnlineWakeWordFasterDecoderOpts opts_;
-  const TransitionModel &trans_model_; // needed for trans-id -> phone conversion
   Token *immortal_tok_;      // "immortal" token means it's an ancestor of ...
   Token *prev_immortal_tok_; // ... all currently active tokens
   KALDI_DISALLOW_COPY_AND_ASSIGN(OnlineWakeWordFasterDecoder);
--- a/src/rnnlm/rnnlm-training.cc
+++ b/src/rnnlm/rnnlm-training.cc
@@ -34,7 +34,6 @@ RnnlmTrainer::RnnlmTrainer(bool train_embedding,
                            nnet3::Nnet *rnnlm):
     train_embedding_(train_embedding),
     core_config_(core_config),
-    embedding_config_(embedding_config),
     objective_config_(objective_config),
     rnnlm_(rnnlm),
     core_trainer_(NULL),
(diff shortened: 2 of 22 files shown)


src/rnnlm/rnnlm-training.h
```diff
@@ -69,7 +69,7 @@ class RnnlmTrainer {
   */
   RnnlmTrainer(bool train_embedding,
                const RnnlmCoreTrainerOptions &core_config,
-               const RnnlmEmbeddingTrainerOptions &embedding_config,
+               const RnnlmEmbeddingTrainerOptions &/* unused embedding_config*/,
```
danpovey, line 72: It seems like it is used, but not via the class member, instead it is given to the trainer like this:
```
    embedding_trainer_ = new RnnlmEmbeddingTrainer(embedding_config,
                                                   embedding_mat_);
```
(which keeps it as a const reference).
I think we can remove the class member.

kkm000: Thanks! Could you please look at another one: https://github.com/kaldi-asr/kaldi/pull/4560/files#diff-b3543f4e3f4dc99d000d8a05e473e533426da611c5eb58e660d6f85fe5347817 , the `TransitionModel OnlineWakeWordFasterDecoder::trans_model_` member in online2/online-nnet3-wake-word-faster-decoder.h? I never looked at the wake word stuff, and am not even sure what the term means exactly. If it what I think it is, an okaygoogle bracketed by silence, then it should be a more or less distant kin to decoder's endpointing, then the transition model would be likely needed.

The rest of suspiciously unused stuff is in @hugovbraun's _UnübertroffenezauberlicheCUDAkunst_ realm.

src/rnnlm/rnnlm-training.cc
```diff
@@ -34,7 +34,7 @@ RnnlmTrainer::RnnlmTrainer(bool train_embedding,
                            nnet3::Nnet *rnnlm):
     train_embedding_(train_embedding),
     core_config_(core_config),
-    embedding_config_(embedding_config),
+    /* TODO(danpovey): unused: embedding_config_(embedding_config),*/
```
danpovey, line 37: you can jsut delete this class member.

src/online2/online-nnet3-wake-word-faster-decoder.h
```diff
@@ -53,8 +53,8 @@ class OnlineWakeWordFasterDecoder : public FasterDecoder {
  public:
   OnlineWakeWordFasterDecoder(const fst::Fst<fst::StdArc> &fst,
                               const OnlineWakeWordFasterDecoderOpts &opts,
-                              const TransitionModel &trans_model)
-      : FasterDecoder(fst, opts), opts_(opts), trans_model_(trans_model) {}
```
danpovey, line 57: I guess the trans_model arg can be deleted, although you might have to change the calling code as well.  if there are no compilation errors, OK to delete.

kkm000: Yes, it goes without saying that calling code also changes.

I'm also wondering if it's worth it to declare this constructor `[[deprecated("Use the 42-argument constructor")]]` for some value of 42 instead of just removing it? I'm cool breaking internal interfaces, but do not like breaking the public API. `OnlineWakeWordFasterDecoder` is certainly useful at inference time, hence in production.

A last one I want to run by you. I noticed in the commit that I merged today: a public constructor
```c++
NnetChainModel2::NnetChainModel2(const NnetChainTraining2Options& /* unused */,
                                 Nnet *nnet, const std::string &den_fst_dir)
    : nnet(nnet), den_fst_dir_(den_fst_dir) {}
```
which became unused when I removed unused member `const NnetChainTraining2Options &opts_` (but the same `NnetChainTraining2Options` to `NnetChainTrainer2` struct is used, and it looks like it's where it should have been from the start. But again, don't want to break a public API. It's unlikely the _trainer_ will be used with anything but own Kaldi binary, but the extra argument is on the `NnetChainModel2`. 

Am I reading the code right, Is this is a training-time only wrapper around `Nnet*`? Then the extra parameter would be ok to remove w/o going through deprecation. Or is it a useful inference-time public class?

danpovey: OnlineWakeWordFasterDecoder  is probably not used enough yet, to worry about breaking APIs.
Sorry I don't recall very much about the NnetChainModel2 code, you'll have to figure it out yourself.  It's likely not very much used right now.

kkm000: Thanks! I won't worry about the OWWFD then, and will figure out the other one, no biggie.

kkm000:
@hugovbraun, I and @danpovey figured out everything outside of CUDA decoder and friends. PTAL at the changes when you have time. Are these unused members simply leftovers of something that did not materialize, or their non-use is a genuine problem that must be fixed? I run analysis like this on the codebase once in a while, but not often and manually, and want to make sure you're at the very least aware of them. Nothing urgent or blocking here, this is just a bit of cleanup.

danpovey:
I think these changes are OK.

hugovbraun:
LGTM, those are old data members from previous iterations on the code. They were left behind. Thanks!

kkm000:
This is ready, @danpovey, PTAL. I reformatted files related to chain2 that I had to touch, so they at least do not stick beyond 80 coulmns. Looks like they had been quite hastily committed, a lot of construction dust still there. I 

also fixed warnings in cudadecoder/thread-pool-light.h, one about the order of initialization not matching the order of members, and another about the pessimizing std::move (which I have a vague recollection of having added... :)

And yes, `NnetChainModel2` is only a training-time wrapper around `Nnet`, so no breakage of public API here.

danpovey (commented):
LGTM!

PR #4562, opened 2021-06-12
[src] Don't leak new'd NnetChainModel2::LanguageInfo

kkm000:
* Mainly, taking heed from static analysis about unpaired delete.
* Remove an unused data member.
* Move the file path combining utility functions into the .cc file, as they are static and do not semantically belong in the class.
* Adjust formatting to the coding style (not exhaustively).

### DIFF
--- a/src/nnet3/nnet-chain-training2.cc
+++ b/src/nnet3/nnet-chain-training2.cc
@@ -20,22 +20,21 @@
 // limitations under the License.
 
 #include "nnet3/nnet-chain-training2.h"
+
 #include "nnet3/nnet-utils.h"
 
 namespace kaldi {
 namespace nnet3 {
 
 NnetChainTrainer2::NnetChainTrainer2(const NnetChainTraining2Options &opts,
-                                   const NnetChainModel2 &model, 
-                                   Nnet *nnet):
-    opts_(opts),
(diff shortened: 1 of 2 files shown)


src/nnet3/nnet-chain-training2.cc
```diff
@@ -284,103 +290,84 @@ void NnetChainTrainer2::ProcessOutputs(bool is_backstitch_step2,
...
-void NnetChainModel2::GetPathname(const std::string &dir,
-                                   const std::string &name,
-                                   int32 job_id,
-                                   const std::string &suffix,
-                                   std::string *pathname) {
-  std::ostringstream str;
-  str << dir << '/' << name << '.' << job_id << '.' << suffix;
-  *pathname = str.str();
+// TODO(kkm): Currently unused. Delete or use if intended.
+#if 0
```
danpovey, line 342: delete/cleanup?
Otherwise looks OK.

kkm000: It's unused and is fairly trivial 2-line long helper, so I better kill it. I'm no fan of dead code, and many times so when it's trivial, 2-line long, and a helper whose help is unasked for. The rest I'm merging then.

PR #4566, opened 2021-06-16
[src] Use static_assert for KALDI_COMPILE_TIME_ASSERT

kkm000:
Remove `template<bool B> class KaldiCompileTimeAssert` because `static_assert` is now a language declaration. Before C++17 a second argument for the message is required (lame, lame!), which is, lacking anything better, is obtained by stringizing the function-like macro argument. For C++17, the second argument is optional, and the semantics is identical.


jtrmal:
LGTM

PR #4567, opened 2021-06-16
[src] Initialize CuDevice member; cleanup cu-common.{h,cc}

kkm000:
* Initialize `CuDevice::curand_handle_` to `NULL` in constructor
* Remove stray include guards from cu-common.cc
* Rearrange `#include`s per coding guidelines
* Exclude more `#include`s if !HAVE_CUDA in cu-common.{cc,h}


jtrmal:
LGTM

kkm000:
Looks like I excluded a bit too many includes... Will fix now.

PR #4568, opened 2021-06-16
[src] Assert empty workqueue on ThreadPoolLight destruction

kkm000:
Other minor tweaks:
 * Convert runtime asserts to static asserts in template.
 * Remove ThreadPoolLight::nworkers_ as it's equal to the
   .size() of the worker vector, and was used only once.
 * Replace pointer indirection with std::move and make
   ThreadPoolLightWorker::thread_ a direct class member.
 * Update to coding guidelines and IWYU.

### DIFF
--- a/src/cudadecoder/thread-pool-light.h
+++ b/src/cudadecoder/thread-pool-light.h
@@ -19,15 +19,15 @@
 #define KALDI_CUDADECODER_THREAD_POOL_LIGHT_H_
 
 #include <atomic>
+#include <memory>
 #include <thread>
 #include <vector>
-
-#include "util/stl-utils.h"
 
 namespace kaldi {
 namespace cuda_decoder {
 
-const double kSleepForWorkerAvailable = 1e-3;
+constexpr double kSleepForWorkAvailable = 1e-3;
+constexpr double kSleepForWorkerAvailable = 1e-3;
 
 struct ThreadPoolLightTask {
   void (*func_ptr)(void *, uint64_t, void *);
@@ -39,20 +39,17 @@ struct ThreadPoolLightTask {
 template <int QUEUE_SIZE>
 // Single producer, multiple consumer
 class ThreadPoolLightSPMCQueue {
-  static const unsigned int QUEUE_MASK = QUEUE_SIZE - 1;
+  static constexpr unsigned int QUEUE_MASK = QUEUE_SIZE - 1;
   std::vector<ThreadPoolLightTask> tasks_;
   std::atomic<int> back_;
   std::atomic<int> front_;
-  int inc(int curr) { return ((curr + 1) & QUEUE_MASK); }
+  static int inc(int curr) { return ((curr + 1) & QUEUE_MASK); }
 
  public:
-  ThreadPoolLightSPMCQueue() {
-    KALDI_ASSERT(QUEUE_SIZE > 1);
-    bool is_power_of_2 = ((QUEUE_SIZE & (QUEUE_SIZE - 1)) == 0);
-    KALDI_ASSERT(is_power_of_2);  // validity of QUEUE_MASK
-    tasks_.resize(QUEUE_SIZE);
-    front_.store(0);
-    back_.store(0);
+  ThreadPoolLightSPMCQueue() : tasks_(QUEUE_SIZE), front_(0), back_(0) {
+    KALDI_COMPILE_TIME_ASSERT(QUEUE_SIZE > 1);
+    constexpr bool is_power_of_2 = ((QUEUE_SIZE & (QUEUE_SIZE - 1)) == 0);
+    KALDI_COMPILE_TIME_ASSERT(is_power_of_2);  // validity of QUEUE_MASK
   }
 
   bool TryPush(const ThreadPoolLightTask &task) {
@@ -70,23 +67,25 @@ class ThreadPoolLightSPMCQueue {
   bool TryPop(ThreadPoolLightTask *front_task) {
     while (true) {
       int front = front_.load(std::memory_order_relaxed);
-      if (front == back_.load(std::memory_order_acquire))
+      if (front == back_.load(std::memory_order_acquire)) {
         return false;  // queue is empty
+      }
       *front_task = tasks_[front];
       if (front_.compare_exchange_weak(front, inc(front),
-                                       std::memory_order_release))
+                                       std::memory_order_release)) {
         return true;
+      }
     }
   }
 };
 
-class ThreadPoolLightWorker {
+class ThreadPoolLightWorker final {
   // Multi consumer queue, because worker can steal work
   ThreadPoolLightSPMCQueue<512> queue_;
   // If this thread has no more work to do, it will try to steal work from
   // other
-  std::unique_ptr<std::thread> thread_;
-  bool run_thread_;
+  std::thread thread_;
+  volatile bool run_thread_;
   ThreadPoolLightTask curr_task_;
   std::weak_ptr<ThreadPoolLightWorker> other_;
 
@@ -94,9 +93,9 @@ class ThreadPoolLightWorker {
     while (run_thread_) {
       bool got_task = queue_.TryPop(&curr_task_);
       if (!got_task) {
-          if (auto other_sp = other_.lock()) {
-              got_task = other_sp->TrySteal(&curr_task_);
-          }
+        if (auto other_sp = other_.lock()) {
+          got_task = other_sp->TrySteal(&curr_task_);
+        }
       }
       if (got_task) {
         // Not calling func_ptr as a member function,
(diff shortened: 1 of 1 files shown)


src/cudadecoder/thread-pool-light.h
```diff
@@ -70,33 +67,35 @@ class ThreadPoolLightSPMCQueue {
...
-class ThreadPoolLightWorker {
+class ThreadPoolLightWorker final {
   // Multi consumer queue, because worker can steal work
   ThreadPoolLightSPMCQueue<512> queue_;
   // If this thread has no more work to do, it will try to steal work from
   // other
-  std::unique_ptr<std::thread> thread_;
-  bool run_thread_;
+  std::thread thread_;
+  volatile bool run_thread_;
```
jtrmal, line 88: isn't volatile deprecated? Not sure what the intention of this is

jtrmal (commented):
LGTM (with perhaps the volatile thing questionmark)

src/cudadecoder/thread-pool-light.h
kkm000: Not as of C++20. I've seen a check for a non-volatile boolean flag reordered before the loop. Like this
```c++
bool run;
```
```c++
while(run) {
  body();
}
```
turned by the compiler at -O3 (a pretty recent Clang, IIRC) into this
```c++
if (run) {
  while (true) {
    body();
  }
}
```
A pretty sensible optimization, if you think about it. Depends on what the compiler knows how the variable changes; most compilers are conservative, and if the variable _could_ be changed outside of the compilation unit, they assume it is volatile. The problem is, this cannot be reliably inferred.

kkm000: Not our case (the member is private, but is set to `false` in the same compilation unit), but who knows, this way its more future-proof. `volatile` prevents exactly this case of optimization, so it's sure thing it won't hurt.

jtrmal:
ok

src/cudadecoder/thread-pool-light.h
galv: I 100% agree that we need volatile, or at least std::atomic here.

Unfortunately, volatile is in fact deprecated as of C++20. I would need to read the relevant part of C++ Concurrency in Action 2nd Edition before I know what the appropriate thing to substitute is, though (pretty sure we just need std::atomic<bool> and to call load() on it, but I'm not 100% certain. In particular, it depends upon whether or not we are using the correct atomic memory model semantics.)

galv:
Not sure about this one. I don't meant to doubt you, it's just that I'm not quite sure that current code does in fact guarantee that the queues of work to do are empty by the end of destruction without staring more at it.

src/cudadecoder/thread-pool-light.h
kkm000: Ok, @galv, @jtrmal, I think it deserves a deeper analysis. 

### Part I: The Good and The Ugly

C++20 "deprecates" `volatile` where behavior is really hard or impossible to define, but it does not affect the simple flag pattern, as used here, in any way. The WG21 discussion paper you are looking for is P1152. AFAICR, the R0 of the paper proposed a more widespread deprecation, but only a few cases made it into the language in the end.

In other words, we want get rid of the ugly of it, and allow clearly defined and readable uses.
 
Let's start with a dense quote from pretty authoritative [cppreference.com]( https://en.cppreference.com/w/cpp/language/cv), and for each use I can write a simple example, to show that the deprecated cases are clear as mud. Text is quote, code blocks are mine.

Some uses of `volatile` are deprecated:
 *  lvalue of volatile type as operand of built-in increment/decrement operators;
    ```c++
    volatile int i = 0;
    ++i;
    // How many times memory is accessed? On some hardware, it's
    int tmp = i; i = tmp + 1;
    // This may not end with i incremented by 1.
    // People 
    ```
 *  lvalue of volatile type as left operand of built-in direct/compound assigment operators, unless the direct assigment expression appears in an unevaluated context or is a discarded-value expression;
    ```c++
    int k, m = 0;
    volatile int i = 0;
    k = i = m;  // lvalue of i is not discarded in the (i = m); k is undefined. This is UB.
    // People expect k==m, but that's not the case!
    i = m;  // lvalue evaluated in a discarded-value context, OK. It;s a single store op.
    // This is also our case of 'run = false;': a simple store.
    ```
 * volatile object type as function parameter type or return type
    ```c++
    // What is the difference between these functions?
    void a (std::atomic_short s);
    void a (volatile std::atomic_short s);
    // The answer to the exercise is left to the reader, because I forgot it.
    ``` 
 * volatile qualifier in structured binding declaration. 

The last item does not affect us (the previous also does not; I put it as an illustration how mind-boggling the syntactically and, we can only hope, semantically legal use cases may become).

The basic idea is to restrict the uses of `volatile` that (a) confusing to people and/or (b) confusing to the compiler, w.r.t hardware-dependent behavior (is increment a single instruction, or multiple instruction where you can be interrupted, while the variable changes, and then store the corrupt value?)

For static or member variables, the rule is very simple:
---

#### `volatile` is simply a directive to the compiler to always perform a load before use.

---

This has not changed, not in C++20, and still not considered for a change in C++23.

### Part II. The Bad and The Whatever

`volatile` is, and has only been a type qualifier for the compiler. It prohibits the optimization I shown before by following the above rule (you must load the `run` variable before every test, on each loop iteration. But, as [cppreference.com]( https://en.cppreference.com/w/cpp/language/cv) says,

> `volatile` objects [are] suitable for communication with a signal handler, but not with another thread of execution, see `std::memory_order`

And this is what everyone on the x86 architecture think it does: communicates the stop event to the thread. This is a source of misunderstanding of the feature, leading to coding errors. The features existence makes it more dangerous and more often misused than used where it belongs (or, at the least, OK). This is, in turn, the reason why the C++ communitu wants it out. But what it's useful for?

> Every [...] read or write operation [...] made through a glvalue expression of volatile-qualified type is treated as a **visible side-effect for the purposes of optimization** [the _as-if_ rule, bold mine -kkm] (that is, within a single thread of execution, **volatile accesses cannot be optimized out or reordered with another visible side effect that is sequenced-before or sequenced-after the volatile access** [bold mine -kkm]. This makes volatile objects suitable for communication with a signal handler [...]

Also, volatile does not prevent processor reordering, since it does not imply any type of barrier. The documentation is clear on it as well. As long as it's satisfies the _As-If_, i.e. does not have side effect in the thread that would be out-of-order, as specified above, it's free to be reordered.

But what is it that makes the signal handler different? Nothing! The (non-) problem is, you cannot build a sensible communication between two threads using volatile guarantees only. All you can do is raise a flag once. Yes, you can also color it etc (pass the whole `volatile` data structure). But you do not know whether the thread noticed your flag or not. There is not even a temporal coherence here, so the question loses its semantics and becomes nonsensical. But otherwise, this is the same thing. You raise a flag, and receive response by _other, real synchronization means_. An attempt to wave flags at each other, however, will not work. 

Amazingly, given it's semantics, volatile is a legit way to pass _non-synchronized_ shutdown notification to the thread. This is exactly same as the signal flag scenario, and about the only sane use of `volatile` in multithreaded programming: flip the flag, let the thread eventually load it, and exit. The _other, real synchronization means_ in this case is the call to `thread::join()`. No one knows whether the thread had seen the flag, including the thread itself [[This surprising statement is stretching the semantics of "knows": the thread exits in the end, so we infer it "knew" that the the flag was set. It's the "when" of "knows" is entirely undefined, so there exists no temporal state in which the thread may know that it knows that the flag is set]]. As soon as the caller `thread::join()` returns, it learns that the thread has seen the flag (and, in theory, the thread that just exited learns of this event too, but this has zero practical implications). This is a normal Synchronizes-With relation, except it non-temporal.

But who cares whether the `thread::join()` _Synchronizes-With_ the even of the setting a Boolean flag temporally or non-temporally? We just want to stop the thundering thread. Either way, it works. So, whatever. `volatile` works here.

kkm000:
@galv, When you are faced with a bug like the three-is-a-charm one, you want asserts() everywhere.

I haven't seen this assert() ever triggering. Maybe my other changes guarantee it. But that was my goal: make shutdown deterministic.

src/cudadecoder/thread-pool-light.h
kkm000: @galv, @jtrmal, @danpovey, @hugovbraun, guys, holy Guacamole, I laughed to tears, you'll enjoy that!!! https://godbolt.org/z/E5E3ecPWf

Try all 4 combinations of {`volatile` at line 8 commented out / not commented out} × { Clang 11 / Clang 12 }. This is x64, no exotics.

Look how differently Clang11 and Clang12 optimize this program. With `volatile`, it works as expected and prints some large number. Without `volatile`, Clang11 optimizes `stop` check out of the loop to do it once, and the program runs forever ("Killed - processing time exceeded" in 1-2s). Clang12 optimizes it even further: it removes the sleep, initializes `stop` to false, guts out the `threadfunc` to a single `ret` (it's not gonna run anyway, as `stop` is false right from the start), and the program prints 0 because there is not a single instruction that would increment the `counter`.

Without `volatile`, we're getting the good old UB. `volatile` prohibits such extreme reordering. **It's absolutely important that in all places where we flagging the thread to stop that the flag variable is marked `volatile`.**

I tried Clang 13 (branch) and -std=c++2b. `volatile` is not going anywhere; as expected, it yells at `++count` if you make it volatile, as this use of `volatile` is no longer allowed. Same as c++20, no change.

Also note that non-volatile counter is loaded once, incremented in a register and stored when the loop exits. When its volatile, the `add <something> 1` instruction is used instead, to increment a memory address. So it works both ways, for load as well as for store. (you need the non-UB program, i.e. with `volatile bool stop` for that, and c++14 or 17, because 20 and above does not like the `++` with a volatile variable, and since it's another UB, the compiled code does not have to be sensical).

Issue #4556, opened 2021-06-04
CUDA decoder H2H threads behavior intended?


kkm000:
@hugovbraun, the question is most likely for you. I am not sure I correctly understand the behavior of H2H threads (which are separate from the post-processing thread pool), and is it intended. Some of this applies only if  `CudaDecoder::n_threads_used_ > 1`, but even with just one thread (my case, currently), there are possible issues.

All the `n_threads_used_` H2H threads are sitting there waiting for work

https://github.com/kaldi-asr/kaldi/blob/ea2b433dfdd7eab4c7a665ef46f7e87a2e4a782e/src/cudadecoder/cuda-decoder.cc#L1831-L1836

signaled by the `n_h2h_main_task_todo_cv_` CV

https://github.com/kaldi-asr/kaldi/blob/ea2b433dfdd7eab4c7a665ef46f7e87a2e4a782e/src/cudadecoder/cuda-decoder.cc#L1978-L1987

and to initiate this work, `CudaDecoder::LaunchH2HCopies()` increments `n_h2h_task_not_done_` by  `n_threads_used_`, sets  `n_h2h_main_task_todo_` to equal exactly `n_threads_used_`, and signals _all_ H2H threads to start their thing:

https://github.com/kaldi-asr/kaldi/blob/ea2b433dfdd7eab4c7a665ef46f7e87a2e4a782e/src/cudadecoder/cuda-decoder.cc#L1814-L1825

The H2H thread(s) decrease `n_h2h_main_task_todo_` by the same `n_threads_used_` immediately and proceeds. `n_h2h_task_not_done_` is decremented near the end of `ComputeH2HCopies()` under a mutex, to complete the CV semaphore pattern.

[...]

PR #4569, opened 2021-06-16
[src] Deterministic shutdown and error checks to 11 in CudaDecoder

kkm000:
* Wait for D2H then H2H transfers to dry out before stopping the pump.
* Assert no H2H tasks after it stops.
* Assert that an offset into vector is within bounds (in #4556 scenario, it's not).
* Add `CU_SAFE_CALL` to every CUDA call.
* Use compile time asserts on constants.
* The rest is reformatting comments, IWYU and coding conventions.

### DIFF
--- a/src/cudadecoder/cuda-decoder.cc
+++ b/src/cudadecoder/cuda-decoder.cc
@@ -22,14 +22,27 @@
 #include "cudadecoder/cuda-decoder.h"
 
 #include <algorithm>
-#include <map>
-#include <sstream>
+#include <atomic>
+#include <functional>
+#include <iterator>
+#include <list>
+#include <memory>
+#include <mutex>
+#include <stack>
+#include <string>
+#include <thread>
 #include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
 
 #include <cuda_runtime_api.h>
 #include <nvToolsExt.h>
 
+#include "base/kaldi-utils.h"
 #include "cudadecoder/cuda-decoder-kernels.h"
+#include "cudamatrix/cu-common.h"
 #include "online2/online-endpoint.h"
 #include "util/text-utils.h"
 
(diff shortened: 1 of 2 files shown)


kkm000:
@galv or @hugovbraun, maybe you want to have a look.

hugovbraun:
LGTM. Did the WaitForH2H() .. end up helping? I agree it's necessary to call them in the destructor, however in practice they should be called by the last lattice generation

kkm000:
@hugovbraun, thanks for looking! None of my latest commits ended up helping. I just untangled the heap of modifications related to things that could potentially help. Most of them simply eliminate races during shutdown (potential races: as you note, some of them are unlikely). Still, I needed to be sure that everything is squeaky clean on this front. I decided to send these changes, because an extra correctness didn't kill the cat. But none of these have any effect on #4556, not a tiny bit. I hit a brickwall.

For another test, I commented out freeing cuBLAS and friends handles in CuDevice destructor, that also did not help. I suspected stuff may be done out of order there, as the instances are thread_local and destroyed very late, and there are a lot of threads involved at the moment of overlap: my own, those of the decoder, gRPC etc. But nope, that wasn't it.

I hit a brickwall with this one. We are not doing the the code that is going to production, and it's probably not even a big overall throughput gain, but it still leaves that lingering feeling that _something is probably not right_. It would be super to get to the bottom of it, eventually. @galv suggested a few checks on that ticket that I could try to see if anything changes.

Regarding your faster memcpy question, if you want to share what is the actual problem that you're solving, that would make more sense. I skimmed the Intel Optimization Manual, and there are a few techniques, from alternating processing and loading data so that the prefetcher would figure out which memory is going to be accessed next and what is its pitch pattern, and load-ahead it in LLC/L2, to instructions performing non-temporal streaming from RAM into the CPU buffer (8 cachelines long, IIRC), bypassing all caches (a streamed application, like a video encoder, benefits from it, as it churns streamed data which it uses only once, so saving the cache for computation local data is a huge win). One can even bypass the prefetcher and do explicit PREFETCHxxx instructions at the right time, but that pushed us down nearly to the assembly level. It all depends on the data use pattern, the CPU generation etc—same things as with the GPU, different CPU microarchitectures require specific tuning, and even the same cores come with x2-x3 difference in L3 cache size (Xeon vs i7/i9).

If you feel more comfortable chatting about that by e-mail, mine is always in the profile, kkm@pobox.com. Please feel free to e-mail any time. I like tinkering with this stuff and getting the most oomph out of the largest CPUs, so that sounds like a great problem to solve!

PR #4570, opened 2021-06-16
[src] Shutdown CUDA online pipeline deterministically

kkm000:
Also reformat string literals that ended up split into too many
pieces over too many lines, and add commentary on the importance
of data member ordering within the class.

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
@@ -68,7 +68,7 @@ struct BatchedThreadedNnet3CudaOnlinePipelineConfig {
     po->Register("max-batch-size", &max_batch_size,
                  "The maximum execution batch size."
                  " Larger = better throughput, but slower latency.");
-    po->Register("num-channqels", &num_channels,
+    po->Register("num-channels", &num_channels,
                  "The number of parallel audio channels. This is the maximum"
                  " number of parallel audio channels supported by the pipeline."
                  " This should be larger than max_batch_size.");
@@ -415,11 +415,11 @@ class BatchedThreadedNnet3CudaOnlinePipeline {
   // Ordering of the cuda_fst_ w.r.t. thread_pool_ and the decoder is important:
   // order of destruction is bottom-up, opposite to the order of construction.
   // We want the FST object, which is entirely passive and only frees device
-  // FST representation when destroyed, to survove both the thread pool and the
+  // FST representation when destroyed, to survive both the thread pool and the
   // decoder, which both may perform pending work during destruction. Since no
   // new work may be fed into this object while it is being destroyed, the
   // relative order of the latter two is unimportant, but just in case, FST must
-  // stay around until the otther two are positively quiescent.
+  // stay around until the other two are positively quiescent.
 
   // HCLG graph. CudaFst is a host object, but owns pointers to the data stored
   // in GPU memory.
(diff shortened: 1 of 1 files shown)


kkm000:
@galv or @hugovbraun, maybe you want to eyeball this one, too.

jtrmal:
LGTM

jtrmal:
Well, actually, I find the PR name confusing, esp. cs the changes are in the object construction :)

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
```diff
@@ -66,24 +66,22 @@ struct BatchedThreadedNnet3CudaOnlinePipelineConfig {
...
-                 "The maximum execution batch size. "
-                 "Larger = Better throughput slower latency.");
-    po->Register("num-channels", &num_channels,
-                 "The number of parallel audio channels. This is the maximum "
-                 "number of parallel audio channels supported by the pipeline"
-                 ". This should be larger "
-                 "than max_batch_size.");
+                 "The maximum execution batch size."
+                 " Larger = better throughput, but slower latency.");
+    po->Register("num-channqels", &num_channels,
```
danpovey, line 71: typo

kkm000:
@jtrmal, it's very straightforward: constructors of members are called in the order they appear in the class, top to bottom, and the destructors are in the reverse order. So I just moved the CudaFst variable, a "passive" object that just frees device memory when destructed, above "active" objects which (potentially) wait for threads to complete in their destructors and (may) need the long-living device representation of the CudaFst during destruction. I don't think the relative order of the other two matters; at the least, they did not affect the enigmatic three-is-a-charm issue #4556. I just think the more deterministic device-host operations are, the better. So this does not really fix anything, just cuts off some branches when reasoning about it (the other PR does the same in a different area).

I'm just dumping out my changes to keep local modifications to the minimum, whether they're helpful or not. The more deterministic the shutdown of the pipeline + the batcher is, the better. This scenario is not covered in the corresponding binary.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
```diff
@@ -415,22 +412,36 @@ class BatchedThreadedNnet3CudaOnlinePipeline {
   // Only used if feature extraction is run on the CPU
   std::vector<std::unique_ptr<OnlineNnet2FeaturePipeline>> feature_pipelines_;
 
-  // HCLG graph : CudaFst object is a host object, but contains
-  // data stored in
-  // GPU memory
+  // Ordering of the cuda_fst_ w.r.t. thread_pool_ and the decoder is important:
+  // order of destruction is bottom-up, opposite to the order of construction.
+  // We want the FST object, which is entirely passive and only frees device
+  // FST representation when destroyed, to survove both the thread pool and the
```
danpovey, line 418: survive

danpovey (commented):
noticed couple typos.   one is channqels

kkm000:
_(Grumbles.)_ That's a VSCode spellchecker for ya. The only one working I could only find. I suspect it used either heuristics or, I am starting to suspect and fear, an ML model. Misses _survove_ and thinks _chanquels_, _recostruct_ and _otther_ are words (I'm a grumpy otter, and I'm unhappy with that!), but yells on some common medium-frequency words. VSCode gives plugins access to clang-assisted pre-parsed syntax (while Emacs essentially uses a beefed-up regex engine), so it can certainly tell a comment or a string from an identifier.

Emacs + flyspell-prog-mode (with ispell backend) is da best! I do normally spell-check code with it after VSCode. It does not think Doxygen comments are comments and must be spell-checked, tho...

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
kkm000: Done.

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.h
kkm000: Done + 2 more.

galv (approved):
I remember looking into the what-order-are-member-destructors-run rules as well, back in #4517 (since the relevant cuda error appeared in a destructor, but I wasn't sure which one).

This looks fine to me. Please merge. It is probably better to merge this as is. If I find anything unusual in my internal work, I will speak up.

kkm000:
Thanks @galv! Please also have a look at #4569, it's even simpler to review, very regular changes. Mostly CU_SAFE macros everywhere, 4 asserts in a row at the point where the memory is corrupted in #4556, and deterministic shutdown of device streams and H2H pumps. It fixes nothing, only removes a possible synchronization issues from the scope of the issue.

PR #4571, opened 2021-06-18
On the fly contextual adaptation,

vesis84:
adding code for doing 'lattice boosting' and 'HCLG boosting' to do on-the fly contextual adaptation by WFST composition

- this PR contains the code to do the composition, each utterance can be adapted with different FST
- this PR does not contain the code to prepare the WFST 'adaptation graphs',
- for HCLG boosting the graph should be simple (single state graph), for more complicated boosting graphs the HCLG o B composition is too slow...

### DIFF
--- a/egs/wsj/s5/steps/nnet3/decode_compose.sh
+++ b/egs/wsj/s5/steps/nnet3/decode_compose.sh
@@ -22,8 +22,8 @@ min_active=200
 ivector_scale=1.0
 lattice_beam=8.0 # Beam we use in lattice generation.
 iter=final
-num_threads=1 # if >1, will use gmm-latgen-faster-parallel
-use_gpu=false # If true, will use a GPU, with nnet3-latgen-faster-batch.
+#num_threads=1 # if >1, will use gmm-latgen-faster-parallel
+#use_gpu=false # If true, will use a GPU, with nnet3-latgen-faster-batch.
               # In that case it is recommended to set num-threads to a large
               # number, e.g. 20 if you have that many free CPU slots on a GPU
               # node, and to use a small number of jobs.
--- a/egs/wsj/s5/steps/nnet3/decode_compose.sh
+++ b/egs/wsj/s5/steps/nnet3/decode_compose.sh
@@ -48,7 +48,7 @@ echo "$0 $@"  # Print the command line for logging
 if [ $# -ne 3 ]; then
   echo "Usage: $0 [options] <graph-dir> <data-dir> <decode-dir>"
   echo "e.g.:   steps/nnet3/decode.sh --nj 8 \\"
-  echo "--online-ivector-dir exp/nnet2_online/ivectors_test_eval92 \\"
+  echo "      --online-ivector-dir exp/nnet2_online/ivectors_test_eval92 \\"
   echo "    exp/tri4b/graph_bg data/test_eval92_hires $dir/decode_bg_eval92"
   echo "main options (for others, see top of script file)"
   echo "  --config <config-file>                   # config containing options"
@@ -67,7 +67,7 @@ fi
 graphdir=$1
 data=$2
 dir=$3
-srcdir=`dirname $dir`; # Assume model directory one level up from decoding directory.
+srcdir=$(dirname $dir)  # Assume model directory one level up from decoding directory.
 model=$srcdir/$iter.mdl
 
 [ -z "$boosting_graphs" ] && echo "Error, \$boosting_graphs have to be set !" && exit 1
--- a/egs/wsj/s5/steps/nnet3/decode_compose.sh
+++ b/egs/wsj/s5/steps/nnet3/decode_compose.sh
@@ -1,0 +1,181 @@
+#!/usr/bin/env bash
+
+# Copyright 2021       Brno University of Technology (Author: Karel Vesely).
+# Copyright 2012-2015  Johns Hopkins University (Author: Daniel Povey).
+# Apache 2.0.
+
+# This script does decoding with a neural-net.
+# It calls 'nnet3-latgen-faster-compose', which does on-the-fly boosting
+# of HCLG graph by composing it with per-utterance boosting graphs (pre-existing).
+
+# Begin configuration section.
+stage=1
+nj=4 # number of decoding jobs.
+acwt=0.1  # Just a default value, used for adaptation and beam-pruning..
+post_decode_acwt=1.0  # can be used in 'chain' systems to scale acoustics by 10 so the
+                      # regular scoring script works.
+cmd=run.pl
+beam=15.0
+frames_per_chunk=50
+max_active=7000
+min_active=200
+ivector_scale=1.0
+lattice_beam=8.0 # Beam we use in lattice generation.
+iter=final
+num_threads=1 # if >1, will use gmm-latgen-faster-parallel
+use_gpu=false # If true, will use a GPU, with nnet3-latgen-faster-batch.
+              # In that case it is recommended to set num-threads to a large
+              # number, e.g. 20 if you have that many free CPU slots on a GPU
+              # node, and to use a small number of jobs.
+scoring_opts=
+skip_diagnostics=false
+skip_scoring=false
+extra_left_context=0
+extra_right_context=0
+extra_left_context_initial=-1
+extra_right_context_final=-1
+online_ivector_dir=
+minimize=false
+
+boosting_graphs=
+# End configuration section.
+
+echo "$0 $@"  # Print the command line for logging
+
+[ -f ./path.sh ] && . ./path.sh; # source the path.
+. utils/parse_options.sh || exit 1;
+
+if [ $# -ne 3 ]; then
+  echo "Usage: $0 [options] <graph-dir> <data-dir> <decode-dir>"
+  echo "e.g.:   steps/nnet3/decode.sh --nj 8 \\"
+  echo "--online-ivector-dir exp/nnet2_online/ivectors_test_eval92 \\"
+  echo "    exp/tri4b/graph_bg data/test_eval92_hires $dir/decode_bg_eval92"
+  echo "main options (for others, see top of script file)"
+  echo "  --config <config-file>                   # config containing options"
+  echo "  --nj <nj>                                # number of parallel jobs"
+  echo "  --cmd <cmd>                              # Command to run in parallel with"
+  echo "  --beam <beam>                            # Decoding beam; default 15.0"
+  echo "  --iter <iter>                            # Iteration of model to decode; default is final."
+  echo "  --scoring-opts <string>                  # options to local/score.sh"
+  echo "  --num-threads <n>                        # number of threads to use, default 1."
+  echo "  --use-gpu <true|false>                   # default: false.  If true, we recommend"
+  echo "                                           # to use large --num-threads as the graph"
+  echo "                                           # search becomes the limiting factor."
+  exit 1;
+fi
+
+graphdir=$1
+data=$2
+dir=$3
+srcdir=`dirname $dir`; # Assume model directory one level up from decoding directory.
+model=$srcdir/$iter.mdl
+
+[ -z "$boosting_graphs" ] && echo "Error, \$boosting_graphs have to be set !" && exit 1
+
+extra_files=
+if [ ! -z "$online_ivector_dir" ]; then
+  steps/nnet2/check_ivectors_compatible.sh $srcdir $online_ivector_dir || exit 1
+  extra_files="$online_ivector_dir/ivector_online.scp $online_ivector_dir/ivector_period"
+fi
+
+utils/lang/check_phones_compatible.sh {$srcdir,$graphdir}/phones.txt || exit 1
+
+for f in $graphdir/HCLG.fst $data/feats.scp $model $extra_files; do
+  [ ! -f $f ] && echo "$0: no such file $f" && exit 1;
+done
+
+sdata=$data/split$nj;
+if [ -f $srcdir/cmvn_opts ]; then
+    cmvn_opts=`cat $srcdir/cmvn_opts`
+else
+    cmvn_opts="--norm-means=false --norm-vars=false"
+fi
+
+#thread_string=
+#if $use_gpu; then
+#  if [ $num_threads -eq 1 ]; then
+#    echo "$0: **Warning: we recommend to use --num-threads > 1 for GPU-based decoding."
+#  fi
+#  thread_string="-batch --num-threads=$num_threads"
+#  queue_opt="--num-threads $num_threads --gpu 1"
+#elif [ $num_threads -gt 1 ]; then
+#  thread_string="-parallel --num-threads=$num_threads"
+#  queue_opt="--num-threads $num_threads"
+#fi
+queue_opt="--num-threads 1" # 1 thread, we do on-the-fly boosting, the binary has no multi-threading...
+
+mkdir -p $dir/log
+[[ -d $sdata && $data/feats.scp -ot $sdata ]] || split_data.sh $data $nj || exit 1;
+echo $nj > $dir/num_jobs
+
+## Set up features.
+if [ -f $srcdir/online_cmvn ]; then online_cmvn=true
+else online_cmvn=false; fi
+
+if ! $online_cmvn; then
+  echo "$0: feature type is raw"
+  feats="ark,s,cs:apply-cmvn $cmvn_opts --utt2spk=ark:$sdata/JOB/utt2spk scp:$sdata/JOB/cmvn.scp scp:$sdata/JOB/feats.scp ark:- |"
+else
+  echo "$0: feature type is raw (apply-cmvn-online)"
+  feats="ark,s,cs:apply-cmvn-online $cmvn_opts --spk2utt=ark:$sdata/JOB/spk2utt $srcdir/global_cmvn.stats scp:$sdata/JOB/feats.scp ark:- |"
+fi
+
+if [ ! -z "$online_ivector_dir" ]; then
+  ivector_period=$(cat $online_ivector_dir/ivector_period) || exit 1;
+  ivector_opts="--online-ivectors=scp:$online_ivector_dir/ivector_online.scp --online-ivector-period=$ivector_period"
+fi
+
+if [ "$post_decode_acwt" == 1.0 ]; then
+  lat_wspecifier="ark:|gzip -c >$dir/lat.JOB.gz"
+else
+  lat_wspecifier="ark:|lattice-scale --acoustic-scale=$post_decode_acwt ark:- ark:- | gzip -c >$dir/lat.JOB.gz"
+fi
+
+frame_subsampling_opt=
+if [ -f $srcdir/frame_subsampling_factor ]; then
+  # e.g. for 'chain' systems
+  frame_subsampling_opt="--frame-subsampling-factor=$(cat $srcdir/frame_subsampling_factor)"
+elif [ -f $srcdir/init/info.txt ]; then
+    frame_subsampling_factor=$(awk '/^frame_subsampling_factor/ {print $2}' <$srcdir/init/info.txt)
+    if [ ! -z $frame_subsampling_factor ]; then
+        frame_subsampling_opt="--frame-subsampling-factor=$frame_subsampling_factor"
+    fi
+fi
+
+if [ $stage -le 1 ]; then
+  $cmd $queue_opt JOB=1:$nj $dir/log/decode.JOB.log \
+    nnet3-latgen-faster-compose $ivector_opts $frame_subsampling_opt \
+     --frames-per-chunk=$frames_per_chunk \
+     --extra-left-context=$extra_left_context \
+     --extra-right-context=$extra_right_context \
+     --extra-left-context-initial=$extra_left_context_initial \
+     --extra-right-context-final=$extra_right_context_final \
+     --minimize=$minimize --max-active=$max_active --min-active=$min_active --beam=$beam \
+     --lattice-beam=$lattice_beam --acoustic-scale=$acwt --allow-partial=true \
+     --word-symbol-table=$graphdir/words.txt "$model" \
+     $graphdir/HCLG.fst "$boosting_graphs" "$feats" "$lat_wspecifier" || exit 1;
+fi
+
+
+if [ $stage -le 2 ]; then
+  if ! $skip_diagnostics ; then
+    [ ! -z $iter ] && iter_opt="--iter $iter"
+    steps/diagnostic/analyze_lats.sh --cmd "$cmd" $iter_opt $graphdir $dir
+  fi
+fi
+
+
+# The output of this script is the files "lat.*.gz"-- we'll rescore this at
+# different acoustic scales to get the final output.
+if [ $stage -le 3 ]; then
+  if ! $skip_scoring ; then
+    [ ! -x local/score.sh ] && \
+      echo "Not scoring because local/score.sh does not exist or not executable." && exit 1;
+    echo "score best paths"
+    [ "$iter" != "final" ] && iter_opt="--iter $iter"
+    local/score.sh $scoring_opts --cmd "$cmd" $data $graphdir $dir
+    echo "score confidence and timing with sclite"
+  fi
+fi
+echo "Decoding done."
+exit 0;
--- a/src/latbin/lattice-compose-fsts.cc
+++ b/src/latbin/lattice-compose-fsts.cc
@@ -88,8 +88,10 @@ int main(int argc, char *argv[]) {
         fst::ILabelCompare<StdArc> ilabel_comp;
         ArcSort(fst2, ilabel_comp);
       }
+      /* // THIS MAKES ALL STATES FINAL STATES! WHY?
       if (phi_label > 0)
         PropagateFinal(phi_label, fst2);
+      */
 
       fst::CacheOptions cache_opts(true, num_states_cache);
       fst::MapFstOptions mapfst_opts(cache_opts);
@@ -144,8 +146,10 @@ int main(int argc, char *argv[]) {
           fst::ILabelCompare<StdArc> ilabel_comp;
           fst::ArcSort(&fst2, ilabel_comp);
         }
+        /* // THIS MAKES ALL STATES FINAL STATES! WHY?
         if (phi_label > 0)
           PropagateFinal(phi_label, &fst2);
+        */
 
         // mapped_fst2 is fst2 interpreted using the LatticeWeight semiring,
         // with all the cost on the first member of the pair (since we're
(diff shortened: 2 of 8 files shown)


src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+          KALDI_WARN << "No boosting fst for utterance " << utt;
+          num_fail++;
+          continue;
+        } else {
+          boosting_fst = boosting_fst_reader.Value(utt); // copy,
+        }
+
+        timer_compose.Reset();
+
+        // RmEpsilon saved 30% of composition runtime...
```
kkm000, line 216: nit: duplicate periods at end of sentence.
```suggestion
        // RmEpsilon saved 30% of composition runtime.
```

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+        // - Note: we are loading 2-state graphs with eps back-link to the initial state.
+        if (boosting_fst.Properties(fst::kIEpsilons, true) != 0) {
+          fst::RmEpsilon(&boosting_fst);
+        }
+
+        // make sure boosting graph is sorted on ilabel,
```
kkm000, line 222: nit: a comma in place of the period (a few in the file).
```suggestion
        // Make sure boosting graph is sorted on ilabel.
```

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+        if (DecodeUtteranceLatticeFaster(
+                decoder, nnet_decodable, trans_model, word_syms.get(), utt,
+                decodable_opts.acoustic_scale, determinize, allow_partial,
+                &alignment_writer, &words_writer, &compact_lattice_writer,
+                &lattice_writer,
+                &like)) {
+          tot_like += like;
+          frame_count += nnet_decodable.NumFramesReady();
+          num_success++;
+        } else num_fail++;
```
kkm000, line 270: Blockify `else`
```suggestion
        } else {
          ++num_fail;
        }
```

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+    KALDI_LOG << "Composition time "<< elapsed_compose
+              << "s (" << (elapsed_compose * 100.0 / elapsed) << "%)";
+    KALDI_LOG << "Done " << num_success << " utterances, failed for "
+              << num_fail;
+    KALDI_LOG << "Overall log-likelihood per frame is "
+              << (tot_like / frame_count) << " over "
+              << frame_count << " frames.";
+
+    if (num_success != 0) return 0;
+    else return 1;
```
kkm000, line 290: Either
```suggestion
    if (num_success != 0) return 0;
    return 1;
```
or (reads even better to me, but both are ok by coding style)
```suggestion
    return num_success != 0 ? 0 : 1;
```

egs/wsj/s5/steps/nnet3/decode_compose.sh
```diff
@@ -0,0 +1,181 @@
...
+
+mkdir -p $dir/log
+[[ -d $sdata && $data/feats.scp -ot $sdata ]] || split_data.sh $data $nj || exit 1;
+echo $nj > $dir/num_jobs
+
+## Set up features.
+if [ -f $srcdir/online_cmvn ]; then online_cmvn=true
+else online_cmvn=false; fi
+
+if ! $online_cmvn; then
```
kkm000, line 115: `online_cmvn` is used only once.
```suggestion
## Set up features.
if [[ -f $srcdir/online_cmvn ]]; then
```

kkm000 (changes requested):

kkm000:
This is a super interesting observation, thanks @vesis84!
```
        //       OpenFst docs say that more specific iterators
        //       are faster than generic iterators. And in HCLG
        //       is usually loaded for decoding as ConstFst.
        //
        //       auto decode_fst_ = ConstFst<StdArc>(decode_fst);
        //
        //       In this way, I tried to cast VectorFst to ConstFst,
        //       but this made the decoding 20% slower.
```
Very related to #4564. While we're trying to squeeze 5% out of it...

LvHang:
Hm....This is interesting. 

IIRC, the "ConstFst" is actually faster to read ,write and access than "VectorFast" as it has less fragmentation, so the "ReadFstKaldiGeneric" related functions are added and cast the "VectorFst" HCLG to "ConstFst" one when the graph is built.
I checked my old emails and found the following paragraph which is forwarded by Dan, someone said 
"We find that const FSTs load from storage significantly faster than vector FSTs. They also take up less space in memory. Const FSTs also speed up decoding by a measurable amount, as they seem to traverse more quickly. Curiously, we see that const FSTs take up approximately 10% more space on disk when compared to vector FSTs."

For the specific iterators are faster than generic iterators, I remembered that's why the codes as follows are added
```
if (fst_->Type() == "const") {
  LatticeFasterDecoderTpl<fst::ConstFst<fst::StdArc>, Token> *this_cast =
    reinterpret_cast<LatticeFasterDecoderTpl<fst::ConstFst<fst::StdArc>, Token>* >(this);
  this_cast->AdvanceDecoding(decodable, max_num_frames);
....
```

kkm000:
@LvHang, AFAICR, `ConstFst` is not transformed in any way when loaded; it can be directly mapped with mmap(2) into the address space. This means it has offsets there the `VectorFst` has pointers. Offset arithmetics may impose certain overhead, although the 20% difference is hard to explain by it. Possibly, a loss of cache locality is at play here. A good profiler (Intel VTune is freely available now), when best of all running on real hardware (even if a VM has a vPMU, I've personally seen a lot of glitches with these; without PMU, this statistics is unavailable) will collect reliable cache miss statistics.

Also, it's important to do the comparison at the optimization turned to 11: A recent compiler, like GCC 10.2 or Clang 11 or 12, `-O3` is mandatory (especially with GCC, as -O2 does not imply loop unrolling, unlike clang), and all architecture advantages taken (e.g., if you run it on Skylake-X Xeon with AVX512, specify `-march=skylake-avx512`; the default `-msse2` does not cut it at all).

src/latbin/lattice-compose-fsts.cc
```diff
@@ -0,0 +1,194 @@
...
+      VectorFst<StdArc>* fst2 = fst::ReadFstKaldi(fst_rxfilename);
+      // mapped_fst2 is fst2 interpreted using the LatticeWeight semiring,
+      // with all the cost on the first member of the pair (since we're
+      // assuming it's a graph weight).
+      if (fst2->Properties(fst::kILabelSorted, true) == 0) {
+        // Make sure fst2 is sorted on ilabel.
+        fst::ILabelCompare<StdArc> ilabel_comp;
+        ArcSort(fst2, ilabel_comp);
+      }
+      /* // THIS MAKES ALL STATES FINAL STATES! WHY?
```
galv, line 91: Fishy line

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+          num_fail++;
+          continue;
+        } else {
+          boosting_fst = boosting_fst_reader.Value(utt); // copy,
+        }
+
+        timer_compose.Reset();
+
+        // RmEpsilon saved 30% of composition runtime...
+        // - Note: we are loading 2-state graphs with eps back-link to the initial state.
```
galv, line 217: I thought that these are one state graphs, with a bunch of self loops on that single state.

galv: This is another reason why I would want to see a recipe to create an appropriating boost fst.

galv:
> this PR does not contain the code to prepare the WFST 'adaptation graphs',

Is it a good idea to merge this PR as is without an example WFST that this is known to work with? Think about how there was an example recipe with the lookahead decoding recipe PR.

@LvHang @kkm000 I suspect something much, much simpler is going on. FST Composition creates a VectorFst as output, and that VectorFst is used only once (to decode a single utterance). Why convert to ConstFst if you're only going to run the Fst just once? Keep in mind that decoding with FSTs typically reads <10% of the FST states, so the overhead of conversion (with touches 100% of states and 100% of arcs) is not worth it here.

src/latbin/lattice-compose-fsts.cc
kkm000: Does the ρ-matcher treats ε specially? I vaguely remember it doesn't. If this is so, and it falls under the "everything else" category, I can guess why it might happen.

kkm000:
@galv, you're almost certainly right.

@vesis84, only the profiler knows the ultimate truth. When I profile a piece of sluggish code for the first time, I much more often get surprises 🎉🎉 than not. Like, InfoZip's unzip spends 90% of decompression time in libz's crc32() function. Who's expect that!

egs/wsj/s5/steps/nnet3/decode_compose.sh
```diff
@@ -0,0 +1,181 @@
...
+  exit 1;
+fi
+
+graphdir=$1
+data=$2
+dir=$3
+srcdir=`dirname $dir`; # Assume model directory one level up from decoding directory.
+model=$srcdir/$iter.mdl
+
+[ -z "$boosting_graphs" ] && echo "Error, \$boosting_graphs have to be set !" && exit 1
```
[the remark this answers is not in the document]
vesis84: yes, but i prefer to be more "explicit" here...

egs/wsj/s5/steps/nnet3/decode_compose.sh
```diff
@@ -0,0 +1,181 @@
...
+
+extra_files=
+if [ ! -z "$online_ivector_dir" ]; then
+  steps/nnet2/check_ivectors_compatible.sh $srcdir $online_ivector_dir || exit 1
```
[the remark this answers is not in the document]
vesis84: we were using everywhere else 'exit 1' since the beggining of kaldi,
this is essentially a copy of `steps/nnet3/decode.sh` and most of the code comes from there, 
including this part...
and yes, i know what 'set -eu' does...

egs/wsj/s5/steps/nnet3/decode_compose.sh
vesis84: i prefer the script not to deviate too much from `steps/nnet3/decode.sh`, people might be willing to diff against it...

egs/wsj/s5/steps/nnet3/decode_compose.sh
```diff
@@ -0,0 +1,181 @@
...
+  # e.g. for 'chain' systems
+  frame_subsampling_opt="--frame-subsampling-factor=$(cat $srcdir/frame_subsampling_factor)"
+elif [ -f $srcdir/init/info.txt ]; then
+    frame_subsampling_factor=$(awk '/^frame_subsampling_factor/ {print $2}' <$srcdir/init/info.txt)
+    if [ ! -z $frame_subsampling_factor ]; then
+        frame_subsampling_opt="--frame-subsampling-factor=$frame_subsampling_factor"
+    fi
+fi
+
+if [ $stage -le 1 ]; then
```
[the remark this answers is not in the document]
vesis84: this suggestion would break the 'kaldi style', everywhere else in kaldi scripts it is with square brackets...

src/latbin/lattice-compose-fsts.cc
```diff
@@ -0,0 +1,194 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+
+#include "base/kaldi-common.h"
+#include "util/common-utils.h"
+#include "fstext/fstext-lib.h"
+#include "lat/kaldi-lattice.h"
+#include "lat/lattice-functions.h"
```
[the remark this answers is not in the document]
vesis84: isn't it better to sort them logically ?
`base/...` `util/...` those are just some simple tools
`fstext/...` is for reading the boosting graphs
`lat/...` is for manipulating the latticest...

the order corresponds to the logic of the "flow" of the tool...

src/latbin/lattice-compose-fsts.cc
```diff
@@ -0,0 +1,194 @@
...
+        std::string key = lattice_reader1.Key();
+        KALDI_VLOG(1) << "Processing lattice for key " << key;
+        Lattice lat1 = lattice_reader1.Value();
+        ArcSort(&lat1, fst::OLabelCompare<LatticeArc>());
+        Lattice composed_lat;
+        if (phi_label > 0) PhiCompose(lat1, mapped_fst2, phi_label, &composed_lat);
+        else Compose(lat1, mapped_fst2, &composed_lat);
+        if (composed_lat.Start() == fst::kNoStateId) {
+          KALDI_WARN << "Empty lattice for utterance " << key << " (incompatible LM?)";
+          n_fail++;
```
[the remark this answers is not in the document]
vesis84: n_fail++ is okay, only for iterators it is good to use ++ as prefix...

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+      SequentialBaseFloatMatrixReader feature_reader(feature_rspecifier);
+
+      RandomAccessTableReader<fst::VectorFstHolder> boosting_fst_reader(boosting_fst_rspecifier);
+
+      // 'hclg_fst' is a single FST.
+      VectorFst<StdArc> hclg_fst;
+      {
+        auto hclg_fst_tmp = std::unique_ptr<Fst<StdArc>>(fst::ReadFstKaldiGeneric(hclg_fst_rxfilename));
+        hclg_fst = VectorFst<StdArc>(*hclg_fst_tmp); // Fst -> VectorFst, as it has to be MutableFst...
+        // 'hclg_fst_tmp' is deleted by 'going out of scope' ...
```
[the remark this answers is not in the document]
vesis84: i like more the old-school C++ constructor way, it is easier to understand...

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+        const Matrix<BaseFloat> *online_ivectors = NULL;
+        const Vector<BaseFloat> *ivector = NULL;
+        if (!ivector_rspecifier.empty()) {
+          if (!ivector_reader.HasKey(utt)) {
+            KALDI_WARN << "No iVector available for utterance " << utt;
+            num_fail++;
+            continue;
+          } else {
+            ivector = &ivector_reader.Value(utt);
+          }
```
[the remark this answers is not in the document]
vesis84: yes, it is 1 line longer, but also it is more clear... (and it was already like this in `src/nnet3bin/nnet3-latgen-faster.cc`)

src/nnet3bin/nnet3-latgen-faster-compose.cc
vesis84: well, it could be 1 state, and i made it as 2 state with <eps> link to state 1, so that the figures look better.
semantically, it is identical...

egs/wsj/s5/steps/nnet3/decode_compose.sh
```diff
@@ -0,0 +1,181 @@
...
+  fi
+fi
+
+
+# The output of this script is the files "lat.*.gz"-- we'll rescore this at
+# different acoustic scales to get the final output.
+if [ $stage -le 3 ]; then
+  if ! $skip_scoring ; then
+    [ ! -x local/score.sh ] && \
+      echo "Not scoring because local/score.sh does not exist or not executable." && exit 1;
```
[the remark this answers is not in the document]
vesis84: okay, `\` and `;` is not required, but it is also not illegal.

it was like that already in `steps/nnet3/decode.sh`, 
probably to make the code as much clear and explicit as possible...

src/latbin/lattice-compose-fsts.cc
vesis84: @galv What is wrong with that ? Do you react on the present `ArcSort` or the disabled `PropagateFinal` ?

vesis84:
Hi, thank you for the feedback. I adopted some of the suggestions.

Some other suggestion are pointing into the 'old' code that was taken, copied and modified.
I don't think we should deviate from the style of that original code too much, 
it would be good if it stays possible to diff against the original code.

I'll process your inputs and re-open later.
Thanks
Karel

src/latbin/lattice-compose-fsts.cc
danpovey: He may have been referring to the comment 
// THIS MAKES ALL STATES FINAL STATES! WHY?
... 
It is only relevant if the --phi-label option was specified (0 is not allowed for phi label),
 in which case it will use backoff semantics.  
I think it makes sense to keep that commented-out code since the phi stuff is supported.
It is intended that all states in G.fst should be final, because EOS is allowed from all states.

danpovey: .. PropagateFinal() fixes the semantics of EOS, because it is done as a final-prob, not an arc, so the phi-matcher does not naturally handle it.   (In k2 this kind of problem could not happen because final-probs are done as a special transition to the superfinal-state, so we wouldn't need special-purpose code).

galv: Hi @vesis84 the main concern is that you are trying to commit code with a commented out block that you apparently don't understand, from a software engineering point of view.

Also, I thought it was odd that you say that PropagateFinal makes all states final, but fst2 is expected to be a single-state FST based on what you wrote. In this case, I would expect you to set the only state in the FST to be final anyway. Unfortunately, I don't know how phi-symbol matching works, though, so I can't really say more about what your specific problem may be.

vesis84: In FST composition with ρ-matcher the ε symbols the output of the composition should be the same regardless if ε is removed or not. On the other hand, for me the runtime of composition was faster, when i removed ε from the boosting graph B in the HCLG o B composition.

src/latbin/lattice-compose-fsts.cc
```diff
@@ -0,0 +1,200 @@
+// latbin/lattice-compose-fsts.cc
+
+// Copyright 2020  Brno University of Technology; Microsoft Corporation
+
```
danpovey, line 4: Was this file actually intended to be added?  It seess to be the same as lattice-compose, with some relatively small changes, and essentially the same interface; I don't understand why it is a new program instead of a modification to the old one.

src/latbin/lattice-compose-fsts.cc
```diff
@@ -0,0 +1,200 @@
...
+    using fst::VectorFst;
+    using fst::StdArc;
+
+    const char *usage =
+        "Composes lattices (in transducer form, as type Lattice) with word-network FSTs.\n"
+        "Either with a single FST from rxfilename or with per-utterance FSTs from rspecifier.\n"
+        "The FST weights are interpreted as \"graph weights\" when converted into the Lattice format.\n"
+        "\n"
+        "Usage: lattice-compose-fsts [options] lattice-rspecifier1 "
+        "(fst-rspecifier2|fst-rxfilename2) lattice-wspecifier\n"
```
danpovey, line 42: I am thinking this can be done with a string arg called e.g. "--compose-with-fst", defaulting to "auto" which is the old behavior, meaning: rspecifier=lats, rxfilename=FST; and true/True or false/False is FST or lattice respectively.

src/latbin/lattice-compose-fsts.cc
vesis84: yes, they are similar, the original `lattice-compose` does compose 'N lattices' with 'N lattices', or 'N lattices' with '1 FST'...

yes, it could be told externally that there are rxpecifier has 'N lattices' by default or 'N FSTs' with some boolean option like '--read-fsts-from-rxspec', but that would be lead to a more complicated interface, which goes against the principle of 'copy and modify' to keep things clear that we were using in the past...

and yes i'll do the change, if it is needed

src/latbin/lattice-compose-fsts.cc
vesis84: I'd like to cosult about this `PropagateFinal(phi_label, fst2)`. 
Is it needed there ?

I did not want its behavior that all the states in the FST graph were becoming final states 
(2nd arg of FST composition) and the contextual adaptation was not working well due to that...

It was maybe helpful when composing lattices with lattices,
to make them better 'match' and avoid empty outputs?

Could that be also disabled by an CLI option (--propagate-final=false) ?

danpovey: I'd rather modify the existing program if it can be done while clearly not affecting existing usages, just to avoid bloat.
The final propagation is needed to have expected behavior when the FST represents a backoff n-gram language model, because if you don't have that, the proper semantics of backing off to lower order states doesn't happen when it comes to final-probs.   I'm not sure what your topology is.. perhaps you can describe it?

kkm000:
@vesis84 
> sometimes the code seems to be 'over-explicit', and this is for a good reason of simple readability. It might annoy some people, but also help other people.

Readability = good. I have not seen anyone annoyed because he had too little trouble reading code :) I've been annoyed by unreadable code myself. Unless you are referring to the comment about a stack object destroyed at end of scope, which is really excessive, I'd say. One needs to know basic C++ to read code.

> I don't think we should deviate from the style of that original code too much

I think that we should do [what we say we do](http://kaldi-asr.org/doc/style.html) and follow the coding style, even though we have old code which doesn't, and fix the old code when we touch it. Or stop saying that we have the standard. Being of two minds in this respect is the worst.

One's logical order of includes is another's illogical order. Alphabetical order is unambiguous. Imagine I edit something unrelated in this file and also casually rearrange includes because I think my order is more logical. You then rearrange them again. Last thing I want to see is the battle of commits. Unless you have any idea how to unambiguously standardize the "logical order," it's better to stick tight to written guidelines.

One day I'll unleash clang-format on the codebase, and it will ultimately squish all style arguments. Which is a good thing, as I want to spend the first day of the rest of my life kaldying, not arguing about our coding style. :)

egs/wsj/s5/steps/nnet3/decode_compose.sh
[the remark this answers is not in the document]
kkm000: Agree, these are harmless. And we don't have standardized Bash style, although Github thinks Kaldi is written in Bash. :)

src/latbin/lattice-compose-fsts.cc
```diff
@@ -0,0 +1,194 @@
...
+                "Number of states we cache when mapping LM FST to lattice type. "
+                "More -> more memory but faster.");
+    po.Read(argc, argv);
+
+    if (po.NumArgs() != 3) {
+      po.PrintUsage();
+      exit(1);
+    }
+
+    KALDI_ASSERT(phi_label > 0 || phi_label == fst::kNoLabel); // e.g. 0 not allowed.
```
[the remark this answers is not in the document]
kkm000: This is a misunderstanding, I'm not proposing to ignore it. `KALDI_ERR` also kills the program. You print a message which makes sense to the user: "--phi-label=0 (epsilon) is invalid". ASSERT is to assert invariants that result not from user actions, but from programming errors. And it prints what looks gibberish to the user. What does a message "phi_label > 0 || phi_label == fst::kNoLabel" mean to someone who runs the program? It's a message to programmer. Also, when people report assertion failures, we handle them differently, we know it's _not_ a user's error. There is a very clear semantic distinction.

src/latbin/lattice-compose-fsts.cc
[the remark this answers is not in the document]
kkm000: And even better to always use the prefix form in the discarded value context. I agree with the standard's justification that this reduces your own mental load: you don't have to remember whether or not it's an iterator.

src/nnet3bin/nnet3-latgen-faster-compose.cc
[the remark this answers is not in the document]
kkm000: Yeah, I'm not a big fan of the braces too, but this is the only case where they potentially prevent the [most vexing parse](http://www.fluentcpp.com/2018/01/30/most-vexing-parse/). I'm even less a fan of incomprehensible compiler diagnostics... :)

src/nnet3bin/nnet3-latgen-faster-compose.cc
```diff
@@ -0,0 +1,295 @@
...
+
+    double tot_like = 0.0;
+    kaldi::int64 frame_count = 0;
+    int num_success = 0, num_fail = 0;
+    // this compiler object allows caching of computations across
+    // different utterances.
+    CachingOptimizingCompiler compiler(am_nnet.GetNnet(),
+                                       decodable_opts.optimize_config);
+
+    KALDI_ASSERT(ClassifyRspecifier(hclg_fst_rxfilename, NULL, NULL) == kNoRspecifier);
```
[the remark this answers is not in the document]
kkm000: Oh. This is a bug. Don't put any code inside the ASSERT. It compiles to nothing if NDEBUG is defined.

https://github.com/kaldi-asr/kaldi/blob/9d235864c3105c3b72feb9f19a219dbae08b3a41/src/base/kaldi-error.h#L184-L194

danpovey:
that ClassifierRspecifier() call has no side effects, at least if last 2
args are null, so that should be OK.

vesis84:
Okay, i give up for the moment. And Dan's comment about extending `lattice-compose` is right.
I'll work on it later in a separate branch. And working contextual adaptation on lattice-level will be a good start.

One more thing, when i was going through Google papers, i noticed they are using sigma-matcher for contextual adaptation.
The boosting graph contains just the text snippets to be boosted and a sigma arc to match everything else.

I am not sure yet, how difficult would it be to add support of sigma-composition to kaldi for composing 
a lattice with FST having a sigma-arc. But the 'wirings' might be analogical to the phi-composition that
already is there... So, maybe one day... Let's see...

daanzu:
@vesis84 I am quite interested in this feature, and appreciate you making the PR. Regardless of whether it gets merged right now, I would be curious to see an full example including the "adaptation graph" FST. Even seeing a pre-prepared FST would be enlightening, if you don't want to release the code to generate them.

danpovey:
I'm going to see if I can find someone to address the issues in this PR.  I think it is an important topic that we should pursue.

jtrmal:
I'll try to respond this week if @danpovey or @kkm won't have more specific
ideas (sooner :)
y.

kkm000:
I think that boosting individual words would be a great thing already! We
used Kaldi decoder to automate customer service for multiple clients,
including clothing manufacturers and road assistance. The road assistance
application sometimes spit out stuff like "I need to change attire." A
simple lexicon boosting would have resolved that.

More significant HCLG changes on the fly would be likely expensive.

jtrmal:
yeah, I agree

jtrmal:
@vesis84  do you have the bandwidth to work on this? it would be a welcome feature

PR #4596, opened 2021-07-22
[cudadecoder] cuda decoder crash when `--cuda-worker-threads` option is not set

quadr:
In constructor of `BatchedThreadedNnet3CudaOnlinePipeline`, `BatchedThreadedNnet3CudaOnlinePipelineConfig` is check and fix config values. However `ThreadPoolLight` use `config` that is original config instead of `config_` is fixed.


jtrmal:
LGTM

kkm000:
LGTM. Excellent catch, thank you!

kkm000:
@hugovbraun, @galv, just FYI, so that you track and copy. The fix it totally solid.

PR #4598, opened 2021-07-23
Remove break statement to prevent leaving loop early before finishing processing rest of segments

wangsiw1:
At this stage of segmentation, it may happen that after previous segment was split, it triggers break statement and rest of segments in the list are discarded.

Consider there are 2 segments in the list 'segments', first one has duration of 17, and second one has 6. The input parameter "hard_max_segment_length" is 15, so first segment needs to be split again. After splitting the first one, the tool reaches StopIteration exception, and then just break out of for loop and left 6-second segment not append to 'new_segments'.

Not sure why was that break statement left there. Please correct me if there are other conditions that need to be considered.

### DIFF
--- a/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py
+++ b/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py
@@ -782,7 +782,6 @@ class Segment(object):
                     self.split_lines_of_utt, cur_start_index,
                     segment.end_index)
                 new_segments.append(new_segment)
-                break
         segments = new_segments
         return segments
 


kkm000:
Yeah, something is seriously borked in this implementation. The function description says
https://github.com/kaldi-asr/kaldi/blob/42e6d7a1543238cf63d357a5a75b2b687a2554c4/egs/wsj/s5/steps/cleanup/internal/segment_ctm_edits_mild.py#L657-L659

What actually happens it adds some part(?) of the segment that it failed to split, and breaks out of the loop. It does not seem to attempt to do what it promises. What does it _actually_ do in this case: does it add the whole 17-second-long segment back to the return list, or does it truncate it? If it doesn't truncate and discard any part of the long segment, it's probably ok. If it truncates it, then it certainly wants more work. Your fix is indeed correct, as discarding the rest of segments is a disaster. I'm just wondering what actually happens in this case _with_ your fix added.

wangsiw1:
Before fix the 17-second segment was split into two parts and not truncated. After removing the break, the other 6 second segment also appeared in the output, but that is the only case I tested

kkm000:
Ah, ok, cool!. LGTM, checking in, thanks for the fix!

PR #4610, opened 2021-08-17
#4574 Create Minimal Dependency Lattice Tools

galv:
Various lattice tools depend upon the (rather large) TransitionModel
interface to discover extra information, like phone boundaries, and so
on. Fortunately, none depends upon anything HMM-specific other than
LatticeForwardBackwardMmi and CompactLatticeToWordProns. Therefore, I
created an interface called TransitionInformation to capture the extra
information required by the functions in lat/. I prototyped a
CtcTransitionInformation subclass that allows us to use these
functions on the output of CTC acoustic model decoding, but I don't
provide that in this PR because it sits in an external repo.

The functions that require a larger interface than
TransitionInformation (i.e.,LatticeForwardBackwardMmi and CompactLatticeToWordProns) can be excluded by defining the macro "KALDI_MINIMAL_LATTICE_FUNCTIONS". This is primarily intended
to be used by people integrating kaldi into their larger C++
codebases with their own build systems.

In bazel, someone could do this:
```
cc_library(kaldi,
...
defines=[""],
...)
```
In cmake, someone could do this:
```
target_compile_definitions(kaldi PUBLIC
KALDI_MINIMAL_LATTICE_FUNCTIONS)
```

### DIFF
--- a/src/hmm/transition-model.h
+++ b/src/hmm/transition-model.h
@@ -157,11 +157,11 @@ class TransitionModel: public TransitionInformation {
   int32 SelfLoopOf(int32 trans_state) const;  // returns the self-loop transition-id, or zero if
   // this state doesn't have a self-loop.
 
-  inline int32 TransitionIdToPdf(int32 trans_id) const;
+  inline int32 TransitionIdToPdf(int32 trans_id) const final override;
   // TransitionIdToPdfFast is as TransitionIdToPdf but skips an assertion
   // (unless we're in paranoid mode).
   inline int32 TransitionIdToPdfFast(int32 trans_id) const;
-  const std::vector<int32>& TransitionIdToPdfArray() const override;
+  const std::vector<int32>& TransitionIdToPdfArray() const final override;
 
   int32 TransitionIdToPhone(int32 trans_id) const;
   int32 TransitionIdToPdfClass(int32 trans_id) const;
@@ -172,9 +172,6 @@ class TransitionModel: public TransitionInformation {
   bool IsFinal(int32 trans_id) const;  // returns true if this trans_id goes to the final state
   // (which is bound to be nonemitting).
   bool IsSelfLoop(int32 trans_id) const;  // return true if this trans_id corresponds to a self-loop.
-
-  /// Returns the total number of transition-ids (note, these are one-based).
-  inline int32 NumTransitionIds() const { return id2state_.size()-1; }
 
   /// Returns the number of transition-indices for a particular transition-state.
   /// Note: "Indices" is the plural of "index".   Index is not the same as "id",
--- a/src/hmm/transition-model.h
+++ b/src/hmm/transition-model.h
@@ -157,11 +157,11 @@ class TransitionModel: public TransitionInformation {
   int32 SelfLoopOf(int32 trans_state) const;  // returns the self-loop transition-id, or zero if
   // this state doesn't have a self-loop.
 
-  inline int32 TransitionIdToPdf(int32 trans_id) const final override;
+  inline int32 TransitionIdToPdf(int32 trans_id) const final;
   // TransitionIdToPdfFast is as TransitionIdToPdf but skips an assertion
   // (unless we're in paranoid mode).
   inline int32 TransitionIdToPdfFast(int32 trans_id) const;
-  const std::vector<int32>& TransitionIdToPdfArray() const final override;
+  const std::vector<int32>& TransitionIdToPdfArray() const final;
 
   int32 TransitionIdToPhone(int32 trans_id) const;
   int32 TransitionIdToPdfClass(int32 trans_id) const;
--- a/src/hmm/transition-model.h
+++ b/src/hmm/transition-model.h
@@ -161,6 +161,7 @@ class TransitionModel: public TransitionInformation {
   // TransitionIdToPdfFast is as TransitionIdToPdf but skips an assertion
   // (unless we're in paranoid mode).
   inline int32 TransitionIdToPdfFast(int32 trans_id) const;
+  const std::vector<int32>& TransitionIdToPdfArray() const override;
 
   int32 TransitionIdToPhone(int32 trans_id) const;
   int32 TransitionIdToPdfClass(int32 trans_id) const;
--- a/src/hmm/transition-model.h
+++ b/src/hmm/transition-model.h
@@ -158,7 +158,7 @@ class TransitionModel: public TransitionInformation {
   // this state doesn't have a self-loop.
 
   bool TransitionIdsEquivalent(int32_t trans_id1, int32_t trans_id2) const final;
-  bool TransitionIdIsStartOfToken(int32_t trans_id) const final;
+  bool TransitionIdIsStartOfPhone(int32_t trans_id) const final;
 
   // TransitionIdToPdfFast is as TransitionIdToPdfArray()[trans_id] but skips an assertion
   // (unless we're in paranoid mode).
--- a/src/hmm/transition-model.h
+++ b/src/hmm/transition-model.h
@@ -27,6 +27,7 @@
 #include "hmm/hmm-topology.h"
 #include "itf/options-itf.h"
 #include "itf/context-dep-itf.h"
+#include "itf/transition-information.h"
 #include "matrix/kaldi-vector.h"
 
 namespace kaldi {
@@ -120,7 +121,7 @@ struct MapTransitionUpdateConfig {
   }
 };
 
-class TransitionModel {
+class TransitionModel: public TransitionInformation {
 
  public:
   /// Initialize the object [e.g. at the start of training].
--- a/src/hmm/transition-model.h
+++ b/src/hmm/transition-model.h
@@ -157,8 +157,10 @@ class TransitionModel: public TransitionInformation {
   int32 SelfLoopOf(int32 trans_state) const;  // returns the self-loop transition-id, or zero if
   // this state doesn't have a self-loop.
 
-  inline int32 TransitionIdToPdf(int32 trans_id) const final;
-  // TransitionIdToPdfFast is as TransitionIdToPdf but skips an assertion
+  bool TransitionIdsEquivalent(int32_t trans_id1, int32_t trans_id2) const final;
+  bool TransitionIdIsStartOfToken(int32_t trans_id) const final;
+
+  // TransitionIdToPdfFast is as TransitionIdToPdfArray()[trans_id] but skips an assertion
   // (unless we're in paranoid mode).
   inline int32 TransitionIdToPdfFast(int32 trans_id) const;
   const std::vector<int32>& TransitionIdToPdfArray() const final;
@@ -322,13 +324,6 @@ class TransitionModel: public TransitionInformation {
 
   KALDI_DISALLOW_COPY_AND_ASSIGN(TransitionModel);
 };
-
-inline int32 TransitionModel::TransitionIdToPdf(int32 trans_id) const {
-  KALDI_ASSERT(
-      static_cast<size_t>(trans_id) < id2pdf_id_.size() &&
-      "Likely graph/model mismatch (graph built from wrong model?)");
-  return id2pdf_id_[trans_id];
-}
 
 inline int32 TransitionModel::TransitionIdToPdfFast(int32 trans_id) const {
   // Note: it's a little dangerous to assert this only in paranoid mode.
--- a/src/itf/transition-information.h
+++ b/src/itf/transition-information.h
@@ -46,7 +46,11 @@ class TransitionInformation {
   virtual bool IsFinal(int32_t trans_id) const = 0;
   virtual bool IsSelfLoop(int32_t trans_id) const = 0;
   virtual int32_t TransitionIdToPdf(int32_t trans_id) const = 0;
+  // Ideally, this would return a std::span, but it doesn't because
+  // kaldi doesn't support C++20. Sorry.
+  virtual const std::vector<int32_t>& TransitionIdToPdfArray() const = 0;
   virtual int32_t NumTransitionIds() const = 0;
+  virtual int32 NumPdfs() const = 0;
   virtual int32_t TransitionIdToHmmState(int32_t trans_id) const = 0;
 };
 
--- a/src/itf/transition-information.h
+++ b/src/itf/transition-information.h
@@ -21,6 +21,7 @@
 #define KALDI_ITF_TRANSITION_INFORMATION_H_
 
 #include <stdint.h>
+#include <vector>
 
 namespace kaldi {
 
@@ -99,7 +100,7 @@ class TransitionInformation {
    * TransitionIdToPdf(). Another way to look at this is as the number
    * of outputs over which your acoustic model does a softmax.
    */
-  virtual int32 NumPdfs() const = 0;
+  virtual int32_t NumPdfs() const = 0;
 };
 
 }  // namespace kaldi
--- a/src/itf/transition-information.h
+++ b/src/itf/transition-information.h
@@ -55,14 +55,15 @@ class TransitionInformation {
    * Returns true if this trans_id corresponds to the start of a
    * phone.
    */
-  virtual bool TransitionIdIsStartOfToken(int32_t trans_id) const = 0;
+  virtual bool TransitionIdIsStartOfPhone(int32_t trans_id) const = 0;
   /**
-   * Phone is understood to correspond not only to linguistic phones,
-   * but also graphemes, word pieces, byte pieces, etc. A better word
-   * would be Token. We use the word Phone for backwards compatibility
-   * with TransitionModel, which was designed only with phones in
-   * mind. Since TransitionInformation was added to subsume
-   * TransitionModel, we did not want to change the call site of every
+   * Phone is a historical term, and should be understood in a wider
+   * sense that also includes graphemes, word pieces, etc.: any
+   * minimal entity in your problem domain which is represented by a
+   * sequence of transitions with a PDF assigned to each of them by
+   * the model. In this sense, Token is a better word. Since
+   * TransitionInformation was added to subsume TransitionModel, we
+   * did not want to change the call site of every
    * TransitionModel::TransitionIdToPhone to
    * TransitionInformation::TransitionIdToToken.
    */
--- a/src/itf/transition-information.h
+++ b/src/itf/transition-information.h
@@ -38,27 +38,69 @@ namespace kaldi {
 class TransitionInformation {
  public:
   virtual ~TransitionInformation() {};
-  virtual int32_t TransitionIdToTransitionState(int32_t trans_id) const = 0;
   /**
-   * Phone should really be Token here, where Token could be a word
-   * piece, phone, or character. However, because the original
-   * class, TransitionModel, used Phone, we retain it for the sake
-   * of not changing all the callers of this method.
+   * Returns true if trans_id1 and trans_id2 can correspond to the
+   * same phone when trans_id1 immediately precedes trans_id2 (i.e.,
+   * trans_id1 occurss at timestep t, and trans_id2 ocurs at timestep
+   * 2) (possibly with epsilons between trans_id1 and trans_id2) OR
+   * trans_id1 ocurs before trans_id2, with some number of
+   * trans_id_{k} values, all of which fulfill
+   * TransitionIdsEquivalent(trans_id1, trans_id_{k})
+   *
+   * If trans_id1 == trans_id2, it must be the case that
+   * TransitionIdsEquivalent(trans_id1, trans_id2) == true
+   */
+  virtual bool TransitionIdsEquivalent(int32_t trans_id1, int32_t trans_id2) const = 0;
+  /**
+   * Returns true if this trans_id corresponds to the start of a
+   * phone.
+   */
+  virtual bool TransitionIdIsStartOfToken(int32_t trans_id) const = 0;
+  /**
+   * Phone is understood to correspond not only to linguistic phones,
+   * but also graphemes, word pieces, byte pieces, etc. A better word
+   * would be Token. We use the word Phone for backwards compatibility
+   * with TransitionModel, which was designed only with phones in
+   * mind. Since TransitionInformation was added to subsume
+   * TransitionModel, we did not want to change the call site of every
+   * TransitionModel::TransitionIdToPhone to
+   * TransitionInformation::TransitionIdToToken.
    */
   virtual int32_t TransitionIdToPhone(int32_t trans_id) const = 0;
+  /**
+   * Returns true if the destination of any edge with this trans_id
+   * as its ilabel is a final state (or if a final state is
+   * epsilon-reachable from its destination state).
+   */
   virtual bool IsFinal(int32_t trans_id) const = 0;
+  /**
+   * Returns true if *all* of the FST edge labeled by this trans_id
+   * have the same start and end states.
+   */
   virtual bool IsSelfLoop(int32_t trans_id) const = 0;
-  virtual int32_t TransitionIdToPdf(int32_t trans_id) const = 0;
-  // Ideally, this would return a std::span, but it doesn't because
-  // kaldi doesn't support C++20. Sorry.
+  int32_t TransitionIdToPdf(int32_t trans_id) const {
+    return TransitionIdToPdfArray()[trans_id];
+  }
+  /**
+   * Returns the contiguous array that backs calls to
+   * TransitionIdToPdf().
+   *
+   * Ideally, this would return a std::span, but it doesn't because
+   * kaldi doesn't support C++20 at the time this interface was
+   * written.
+   */
   virtual const std::vector<int32_t>& TransitionIdToPdfArray() const = 0;
   int32_t NumTransitionIds() const {
       return TransitionIdToPdfArray().size() - 1;
   }
+  /**
+   * Return the number of distinct outputs from
+   * TransitionIdToPdf(). Another way to look at this is as the number
+   * of outputs over which your acoustic model does a softmax.
+   */
   virtual int32 NumPdfs() const = 0;
-  virtual int32_t TransitionIdToHmmState(int32_t trans_id) const = 0;
 };
 
-} // namespace kaldi
+}  // namespace kaldi
 
 #endif // KALDI_HMM_TRANSITION_MODEL_H_
--- a/src/itf/transition-information.h
+++ b/src/itf/transition-information.h
@@ -103,4 +103,4 @@ class TransitionInformation {
 
 }  // namespace kaldi
 
-#endif // KALDI_HMM_TRANSITION_MODEL_H_
+#endif // KALDI_TRANSITION_INFORMATION_H_
--- a/src/itf/transition-information.h
+++ b/src/itf/transition-information.h
@@ -31,17 +31,20 @@ namespace kaldi {
  * without a dependency on the hmm/ and tree/ directories. For
  * example, you can consider creating a subclass that implements these
  * via extracting information from Eesen's CTC T.fst object.
+ *
+ * TransitionId values must be contiguous, and starting from 1, rather
+ * than 0, since 0 corresponds to epsilon in OpenFST.
  */
 class TransitionInformation {
  public:
   virtual ~TransitionInformation() {};
   virtual int32_t TransitionIdToTransitionState(int32_t trans_id) const = 0;
-    /**
-     * Phone should really be Token here, where Token could be a word
-     * piece, phone, or character. However, because the original
-     * class, TransitionModel, used Phone, we retain it for the sake
-     * of not changing all the callers of this method.
-     */
+  /**
+   * Phone should really be Token here, where Token could be a word
+   * piece, phone, or character. However, because the original
+   * class, TransitionModel, used Phone, we retain it for the sake
+   * of not changing all the callers of this method.
+   */
   virtual int32_t TransitionIdToPhone(int32_t trans_id) const = 0;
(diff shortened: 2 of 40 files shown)


danpovey:
Great-- thanks!
I am hoping someone like @kkm000 will have time to review this?

kkm000:
@danpovey Sure! 

@galv, Even before I had a look: the AM file format is a TransitionModel followed by one or another type of acoustic model (GMM, NNET...). If we do not change that as well, what do we win in the end? On the surface, you need the class defined even to read the AM at inference time. I probably missed part of the discussion.

kkm000:
@galv, are you on Kaldi Slack? I need to get up to date with this development. I feel that I miss some important piece here. We can do finger or voice, what's more convenient to you.

galv:
> @galv, Even before I had a look: the AM file format is a TransitionModel followed by one or another type of acoustic model (GMM, NNET...). If we do not change that as well, what do we win in the end? On the surface, you need the class defined even to read the AM at inference time. I probably missed part of the discussion.

@kkm000 I'm not sure how aware you are of this, but there are several users of kaldi that use kaldi as a library, rather than as the full-fledged toolkit. There are two challenges that these users have: (1) Code size and (2) Inability to use several kaldi tools with non-HMM-based models.

On point (2), the current kaldi decoders and lattice functions depend upon the TransitionModel interface, which basically assumes you have an HMM model defining your "transition topology". There's a somewhat hacky way I've discussed with a colleague to make a "CTC topology" TransitionModel definition, but it's sketchy. Another point is that CTC isn't the only model type out there. Ideally, we could postprocess lattices created by arbitrary topologies in K2, for example. With this change, we can simply create a suitable base class of TransitionInformation that takes in a K2 topology. Basically, I have quite a few users that want to use kaldi's lattice tools and decoders, without having to buy into its acoustic modeling toolkit as well.

On point (1), the lat/ directory's dependency on TransitionModel creates a dependency on hmm/ (which defines TransitionModel) and tree/ (which TransitionModel depends upon). We don't really need these for decoding with non-HMM models. We break the dependency for external builders of kaldi to avoid having to pull in as much files.

In fact, here are all of the kaldi sources we need in cmake with this change for a sample project that it is using the kaldi WFST cuda decoder. It's pretty small (in my opinion).

```
set(cuda_decoder_sources
  third_party/kaldi/src/base/io-funcs.cc   
  third_party/kaldi/src/base/kaldi-error.cc
  third_party/kaldi/src/base/kaldi-math.cc
  third_party/kaldi/src/base/kaldi-utils.cc
  third_party/kaldi/src/base/timer.cc
  third_party/kaldi/src/fstext/context-fst.cc
  third_party/kaldi/src/fstext/grammar-context-fst.cc
  third_party/kaldi/src/fstext/kaldi-fst-io.cc
  third_party/kaldi/src/fstext/push-special.cc
  third_party/kaldi/src/lat/compose-lattice-pruned.cc
  third_party/kaldi/src/lat/confidence.cc
  third_party/kaldi/src/lat/determinize-lattice-pruned.cc
  third_party/kaldi/src/lat/kaldi-lattice.cc
  third_party/kaldi/src/lat/lattice-functions.cc
  # third_party/kaldi/src/lat/lattice-functions-transition-model.cc                                                                                                                                       
  third_party/kaldi/src/lat/minimize-lattice.cc   
  third_party/kaldi/src/lat/phone-align-lattice.cc
  third_party/kaldi/src/lat/push-lattice.cc
  third_party/kaldi/src/lat/sausages.cc
  third_party/kaldi/src/lat/word-align-lattice.cc
  # third_party/kaldi/src/lat/word-align-lattice-lexicon.cc                                                                                                                                               
  third_party/kaldi/src/lm/arpa-file-parser.cc
  third_party/kaldi/src/lm/arpa-lm-compiler.cc
  third_party/kaldi/src/lm/const-arpa-lm.cc
  third_party/kaldi/src/lm/kaldi-rnnlm.cc
  third_party/kaldi/src/lm/kenlm.cc
  third_party/kaldi/src/lm/mikolov-rnnlm-lib.cc
  third_party/kaldi/src/matrix/compressed-matrix.cc
  third_party/kaldi/src/matrix/kaldi-matrix.cc
  third_party/kaldi/src/matrix/kaldi-vector.cc
  third_party/kaldi/src/matrix/matrix-functions.cc
  third_party/kaldi/src/matrix/numpy-array.cc
  third_party/kaldi/src/matrix/optimization.cc
  third_party/kaldi/src/matrix/packed-matrix.cc
  third_party/kaldi/src/matrix/qr.cc
  third_party/kaldi/src/matrix/sparse-matrix.cc
  third_party/kaldi/src/matrix/sp-matrix.cc
  third_party/kaldi/src/matrix/srfft.cc
  third_party/kaldi/src/matrix/tp-matrix.cc
  third_party/kaldi/src/util/kaldi-holder.cc
  third_party/kaldi/src/util/kaldi-io.cc
  third_party/kaldi/src/util/kaldi-semaphore.cc
  third_party/kaldi/src/util/kaldi-table.cc
  third_party/kaldi/src/util/kaldi-thread.cc
  third_party/kaldi/src/util/parse-options.cc
  third_party/kaldi/src/util/simple-io-funcs.cc
  third_party/kaldi/src/util/simple-options.cc
  third_party/kaldi/src/util/text-utils.cc
  third_party/kaldi/src/cudadecoder/cuda-decoder.cc
  third_party/kaldi/src/cudadecoder/cuda-decoder-kernels.cu
  third_party/kaldi/src/cudadecoder/online-endpoint.cc
  third_party/kaldi/src/cudadecoder/lattice-postprocessor.cc
  third_party/kaldi/src/cudadecoder/cuda-fst.cc
)
```

@kkm000 I am on kaldi2020.slack.com. I will ping you there.

A final note: I would like to make similar changes for decoder/ and cudadecoder/, but those are a bit more involved (rather than simple sed commands), because we can't just virtualize TransitionIdToPdfFast, which is intentionally an inline function.

kkm000:
Yeah, I'm not only "aware" of that, this is _exactly_ how I use Kaldi in production. I productized it 3 times already,  (counting sufficiently different projects, where we had to build from the ground up, not reuse previous code with tweaks), Linux and Windows, and the last one with cudadecoder. That's how you supposed to use it? Depending on what exactly _is_ in your prod. server? On of them compiled small LMs and SRGS on the fly, another did not. I'm kinda pessimistic that we would be able to chop the whole of Kaldi fine enough in a non-trivial way (one file per library) that would cover all the ways people use Kaldi in production.

I totally agree we need to untie the tighter spaghetti balls. You identified one, I have an issue with nnet3/, which is also guilty tightly coupling training-only code with inference-time code. I think this is a very good start. Maybe the next, lesser step, would be to go over all non-trivial files and figure out if they need to be split. In two _files_, not two _libraries_. RYO .a for .o is a no-brainer.

But I want to emphasize the importance not only of a good start, but that of a _good stop_, too. :) We need to address sharp pain points, but think well before proceeding with the rest of the codebase. There is no clear cut of what you need and what you do not need in prod. It depends on what you're building. For the ~~lasy~~ time-savvy, there is function-level linking, and LTO actually requires it.

In the end, if you go prod (w/o GPU), you do LTO, you do PGO, and all this stuff is sensible only in a fully-static compilation.

kkm000:
> I would like to make similar changes for decoder/ and cudadecoder/, but those are a bit more involved (rather than simple sed commands), because we can't just virtualize TransitionIdToPdfFast, which is intentionally an inline function.

I understand _what_. What I'm really fail to grok is _why_. By chopping source files you'll get more source files. Chalk that in red. I can't readily see the benefit that outweigh it enough to put it in the black.

galv:
> > I would like to make similar changes for decoder/ and cudadecoder/, but those are a bit more involved (rather than simple sed commands), because we can't just virtualize TransitionIdToPdfFast, which is intentionally an inline function.
> 
> I understand _what_. What I'm really fail to grok is _why_. By chopping source files you'll get more source files. Chalk that in red. I can't readily see the benefit that would outweigh it enough to put it in the black.

Hmm, just to confirm, the overarching concern isn't code size (just use -ffunction-sections and -fdata-sections). The overarching concern is being able to use the lattice tools and decoders that kaldi provides without having to depend upon the rest of kaldi. Speaking from experience, the kaldi cudadecoder has been a game-changer in other projects I have worked on. There is just no way to get high-throughput speech recognition on tens of thousands of hours of audio (https://openreview.net/forum?id=R8CwidgJ0yT) with modest compute requirements. We would like to reuse the cuda decoder for different languages; if we were to do that today without changing any C++ code, we would have to train a kaldi model from scratch in the language of interest, which includes all of the complicated parts about doing training effectively without NCCL or MPI, getting a pronunciation dictionary, etc. Because the cudadecoder depends upon TransitionModel for no reason other than convenience today, we cannot really do that, though.

galv:
> But I want to emphasize the importance not only of a good start, but that of a good stop, too. :) We need to address sharp pain points, but think well before proceeding with the rest of the codebase. There is no clear cut of what you need and what you do not need in prod. It depends on what you're building. For the lasy time-savvy, there is function-level linking, and LTO actually requires it.

I'm trying to be surgical, above all. Make the changes I need while not breaking any existing interface (explicit or implict) to the largest extent possible. I would like to follow this up with making cudadecoder/ and decoder/ directories depend upon TransitionInformation rather than TransitionModel so that they can be used by CTC (and more exotic) models. I don't really have any other changes in mind beyond that, however.

kkm000:
Sure, this is where I'm totally on the same foot with you! I just need to
really understand the problem. Can we chat tomorrow? What TZ are you in? Im
T, -0700. Have 0615-0745J open, will update you after if that does not work.

galv:
@kkm000 just to confirm, you are saying you are int he pacific time zone, and that you are free 6:15AM to 7:45 AM Pacific Time tomorrow, right?

kkm000:
Yep

kkm000:
> On point (2), the current kaldi decoders and lattice functions depend upon the TransitionModel interface, which basically assumes you have an HMM model defining your "transition topology". There's a somewhat hacky way I've discussed with a colleague to make a "CTC topology" TransitionModel definition, but it's sketchy.

Yeah, the "H" part of HCLG is trivial in CTC.

> With this change, we can simply create a suitable base class of TransitionInformation that takes in a K2 topology. Basically, I have quite a few users that want to use kaldi's lattice tools and decoders, without having to buy into its acoustic modeling toolkit as well.

Of the methods that you've collected under `TransitionInformation`, I think only `IsFinal()` calls methods on private `topo_`. If we add a bitset for it, we can chop it cleanly as a separate structure. All other private data are initialized from topology and CD.

kkm000:
@galv, I'll give the decoder/decodable code another go, just want to share couple ideas.

If you look at the interface like you see it for the first time, it should looks sort of weird: it uses concepts of "HMM state" and "Transition state" which do not really make sense without an HMM. If your idea is that someone would implement this interface for _their_ AM, neither concept may make sense in the user's case. We can get rid of these by exposing two different methods that make no use of these concepts.

First, both uses of  `TransitionIdToHmmState` in lat/, one in determinize-lattice-pruned.cc and another in lat/lattice-functions.cc, do essentially same thing:
```
      if ((x != 0) // has a transition-id on input..
          && (trans.TransitionIdToHmmState(x) == 0)
          && (!trans.IsSelfLoop(x))) {
        y = trans.TransitionIdToPhone(x);
      }
```
If you replace this with a method like `typename Lattice::Label TransitionIdToPhoneLabel(int32_t trans_id)` (the fansy typename is an `int` in OpenFST, without size) and return `fst::kNoLabel = -1` from it, you'll get rid of the HMM state concept, which does not really make sense in the scope of this interface.

Second, look how `TransitionIdToTransitionState` is used in word-align-lattice.cc, 3 times. It's always `TransitionIdToTransitionState(x) == TransitionIdToTransitionState(y)`. So this method could better have a more abstract semantics, reporting the equivalence of two transition IDs _in whatever sense the AM defines it_. For example, `bool AreTransitionIdsEquivalent(int32_t trans_id1, int32_t trans_id2)`.

And do not forget `const vector<int32_t>& TransitionIdToPdfArray() const = 0;` instead of `TransitionIdToPdf`. Please also document that transition IDs are indexed from 1, not 0.

I would not make `NumTransitionIds()` virtual. It's invariant return value is `this->TransitionIdToPdfArray().size() - 1`.  It has one and only one possible implementation.

The comment about the `Phone` being a misnomer is rather confusing to to someone reading the documentation and implementing the interface ("Phone should really be Token" makes no sense, because we don't have a thing called `Token` in the public interface; worse, there _are_ objects of class Token in the decoder code, so it's a double-whammy).  I would rephrase it like "Phone is understood in an extended sense and may not only be phone, but also... etc."

I do not see any problems in decoders using this interface. Looks like it separates nice and clean!

galv:
I removed conditionally including the .cc file here: 4887ae325a3cddbd72eb50aa069559cdd5007824

I also make the decoder/ directory support TransitionInformation in 97e19bd8318d8df3a3029f595eb568cd4b950f2c

Here is the relevant text of the second commit:
```
decodable-matrix.h's subclasses of DecodableInterface now support
TransitionInformation instead of TransitionModel. I could have added
support for TransitionInformation to every sublcass of
DecodableInterface (outside of the decoder/ directory), but all of
those DecodableInterface subclasses have a strong assumpetion on
kaldi (usually, a kaldi-trained acoustic model), which would imply
that you are using TransitionModel rather than TransitionInformation
anyway.

Since DecodableInterface is how decoders get information from
TransitionInformation anyway, we can now say that we can decode with
non-kaldi acoustic models' output with DecodableMatrixMapped.

I expanded the TransitionInformation class with two new interfaces:

virtual const std::vector<int32_t>& TransitionIdToPdfArray() const = 0;
virtual int32 NumPdfs() const = 0;

The first allows us to retain the fast lookups of
TransitionIdToPdfFast that TransitionModel had. Because
TransitionModel::TransitionIdToPdfFast is inline, straightforwardly
virtualizing TransitionInformation::TransitionIdToPdfFast would pay a
performance penalty in the tight loops of the decoder.

NumPdfs() is trivially computable if you have a map of transition ids
to pdfs, and some parts of decoder/ requested it, so I added it.
```

One thing I noticed is that DecodableInterface::LogLikelihood, which calls TransitionIdToPdfFast, is itself a virtual interface.

kkm000:
Thanks for noticing `NumPdfs()`, I forgot to mention it!

> DecodableInterface::LogLikelihood, which calls TransitionIdToPdfFast, is itself a virtual interface.

Aw shucks! :) Perhaps some templating is in order. Decoder perf has been a pain point for wide beams.

I'll be away most of the day today, will go over your changes tonight.

kkm000:
Wait a sec, is not LogLikelihood called once per frame and returns a vector of loglikes? I have to run now, sorry, can't check.

galv:
This are very good suggestions. Thank you @kkm000 . I'm looking into them!

galv:
> Wait a sec, is not LogLikelihood called once per frame and returns a vector of loglikes? I have to run now, sorry, can't check.

DecodableInterface::LogLikelihood() does not return a vector of loglikelihoods. CudaDecodableInterface does in fact do this, because virtual methods don't make sense in cuda: https://github.com/kaldi-asr/kaldi/blob/146bdff1c6e1d9adff3586136ee77d25cc6ccb24/src/cudadecoder/cuda-decodable-itf.h#L32

We cannot do that in general, because interfaces like DecodableMapped are not backed by a contiguous address space of log likelihoods.

src/itf/transition-information.h
```diff
@@ -0,0 +1,59 @@
...
+  virtual int32_t TransitionIdToPdf(int32_t trans_id) const = 0;
+  // Ideally, this would return a std::span, but it doesn't because
+  // kaldi doesn't support C++20. Sorry.
+  virtual const std::vector<int32_t>& TransitionIdToPdfArray() const = 0;
+  virtual int32_t NumTransitionIds() const = 0;
+  virtual int32 NumPdfs() const = 0;
+  virtual int32_t TransitionIdToHmmState(int32_t trans_id) const = 0;
+};
+
+}
```
kkm000, line 57:
```suggestion
}  // namespace kaldi
```

src/lat/lattice-functions-transition-model.h
```diff
@@ -0,0 +1,96 @@
...
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef KALDI_LAT_LATTICE_FUNCTIONS_TRANSITION_MODEL_H_
+#define KALDI_LAT_LATTICE_FUNCTIONS_TRANSITION_MODEL_H_
+
+#include <vector>
+#include <map>
```
kkm000, line 27: This should be alpha-sorted:
```suggestion
#include <map>
#include <vector>
```
The next block of includes, too.

src/lat/lattice-functions-transition-model.h
```diff
@@ -0,0 +1,96 @@
...
+bool CompactLatticeToWordProns(
+    const TransitionModel &tmodel,
+    const CompactLattice &clat,
+    std::vector<int32> *words,
+    std::vector<int32> *begin_times,
+    std::vector<int32> *lengths,
+    std::vector<std::vector<int32> > *prons,
+    std::vector<std::vector<int32> > *phone_lengths);
+
+} // end namespace kaldi
```
kkm000, line 94:
```suggestion
}  // namespace kaldi
```

src/lat/lattice-functions-transition-model.cc
```diff
@@ -0,0 +1,139 @@
...
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "lat/lattice-functions-transition-model.h"
+#include "lat/lattice-functions.h"
+#include "hmm/transition-model.h"
+#include "hmm/hmm-utils.h"
```
kkm000, line 28: nit: the main include goes first, separated by an empty line.
```suggestion
#include "lat/lattice-functions-transition-model.h"

#include "hmm/hmm-utils.h"
#include "hmm/transition-model.h"
#include "lat/lattice-functions.h"
```

src/lat/lattice-functions-transition-model.cc
```diff
@@ -0,0 +1,139 @@
...
+      prons->push_back(phones);
+      phone_lengths->push_back(plengths);
+
+      cur_time += length;
+      state = arc.nextstate;
+    }
+  }
+}
+
+} // end namespace kaldi
```
kkm000, line 139:
```suggestion
}  // namespace kaldi
```

src/lat/lattice-functions.h
```diff
@@ -28,14 +28,24 @@
...
 
 namespace kaldi {
 
+// Redundant with the typedef in hmm/posterior.h. Redeclaration of a
+// typedef is valid in C++, but not C. This is for the sake of users
+// who compile with the macro KALDI_MINIMAL_LATTICE_FUNCTIONS defined,
+// so that they compile the majority of the lattice functions (all but
+// those declared in lattice-functions-transition-model.h) without depending
+// upon the hmm/ or tree/ libraries.
+typedef std::vector<std::vector<std::pair<int32, BaseFloat> > > Posterior;
```
kkm000, line 47: If we're getting rid of `KALDI_MINIMAL_LATTICE_FUNCTIONS`. then probably this comment is redundant? Also, it's obviously C++, so there's hardly any point going on tangent about C typedef rules.

src/lat/lattice-functions.h
```diff
@@ -28,14 +28,24 @@
...
 #include "base/kaldi-common.h"
-#include "hmm/posterior.h"
 #include "fstext/fstext-lib.h"
-#include "hmm/transition-model.h"
 #include "lat/kaldi-lattice.h"
 #include "itf/decodable-itf.h"
+#include "itf/transition-information.h"
+#ifndef KALDI_MINIMAL_LATTICE_FUNCTIONS
+#include "lat/lattice-functions-transition-model.h"
+#endif // KALDI_MINIMAL_LATTICE_FUNCTIONS
```
kkm000, line 37: I thought we were getting rid of `KALDI_MINIMAL_LATTICE_FUNCTIONS`? There are two separate headers and two .cc files.
```suggestion
#include "lat/lattice-functions-transition-model.h"
```

Please also add
```c++
#include "lat/lattice-functions.h"
```
the first thing in the file while you're there--I think it's missing entirely

src/lat/word-align-lattice-lexicon.cc
```diff
@@ -571,7 +571,7 @@ void LatticeLexiconWordAligner::ProcessFinalForceOut() {
 }
 
 void LatticeLexiconWordAligner::ComputationState::Advance(
-    const CompactLatticeArc &arc, const TransitionModel &tmodel, LatticeWeight *weight) {
+    const CompactLatticeArc &arc, const TransitionInformation &tmodel, LatticeWeight *weight) {
```
kkm000, line 574: nit: over 80 columns.

src/lat/word-align-lattice-lexicon.cc
```diff
@@ -1038,8 +1038,18 @@ bool WordAlignLatticeLexicon(const CompactLattice &lat,
...
+        KALDI_WARN << "Lattice test via TestWordAlignedLattice does not support "
+                   << "non-TransitionModel TransitionInformation instances. No "
+                   << "testing will be done.";
+    }
+    // this function brings in Posterior and TransitionModel from
+    // "hmm/transition-model.h" and "hmm/hmm-utils.h". It would be
+    // nice to be able to fully remove those dependencies from this
+    // function, given that it is only for sanity checking.
+    else if (!TestWordAlignedLattice(lexicon_info, *transition_model, lat, *lat_out,
+                                     opts.allow_duplicate_paths)) {
```
kkm000, line 1052: This really comes out too ugly. Just remove the condition check and the call to `TestWordAlignedLattice` into latbin/lattice-align-words-lexicon.cc, the only place it's called from, and the function `TestWordAlignedLattice` itself into the TransitionModel-dependent lattice-functions .cc/.h file.

kkm000 (changes requested):

kkm000:
> `DecodableInterface::LogLikelihood()` does not return a vector of loglikelihoods. [...] We cannot do that in general, because interfaces like `DecodableMapped` are not backed by a contiguous address space of log likelihoods.

A very good catch, thanks. I'll think what we can do to avoid a virtual call there at least in some cases.

src/hmm/transition-model.h
```diff
@@ -160,6 +161,7 @@ class TransitionModel {
   // TransitionIdToPdfFast is as TransitionIdToPdf but skips an assertion
   // (unless we're in paranoid mode).
   inline int32 TransitionIdToPdfFast(int32 trans_id) const;
```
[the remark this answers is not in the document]
kkm000: > I could replace all these cases with creating a reference to the output of TransitionIdToPdfArray, and then indexing that array, but I'm not enthusiastic about it

Neither am I. I was only wondering if it died quietly,

>  Let me know if you have a very strong opinion.

I rarely have strong opinions about CLs from someone of your level of seniority in the project and in general. Consider all comments (except code style/typo fixes) as options for discussion. :)

kkm000:
> [...] things remaining:  
>  Fully document itf/transition-information.h

This would be nice, thanks. We can commit the PR first if you prefer, it's growing to the size that may cause conflicts, especially with the Hugo's large CL.

>  Consider removing TransitionIdToTransitionState and TransitionIdToHmmState, in favor of new interface functions. This will take some thought, so probably a few more days until this is done; I'm sorry. In particular, I have a subclass of TransitionInformation for CTC models which isn't visible here. I want to make sure that any interface changes work for that, as well as more exotic models, before committing.

I'm kinda repeating myself here, but when I looked at the interface for what it is (an interface of Kaldi's lattice decoding machinery to an _abstract_ AM), only the concept of transition ID made sense, and "HMM" certainly raises brows. However, we can theorize about abstractions and making the interface as clean as it gets, but when real life code hits the fan, sometimes _unexpected thing_ fly around. IMO, verifying the interface design concepts against an existing non-Kaldi AM is a great thing to do, and it's kinda lucky you have the private code which you can apply it to. So yeah, please, a fan test of the design is very much appreciated!

> Address https://github.com/kaldi-asr/kaldi/pull/4610/files/97e19bd8318d8df3a3029f595eb568cd4b950f2c#r694766325

This seems to lead nowhere, probably outdated by some commit. Do you remember what is it?

Also,

> 7526c9c Fix codefactor warnings.

the codefactor thing is both missing important stuff and chatters too much about perfectly valid things. So take it as mild suggestions at best. :)

src/itf/transition-information.h
```diff
@@ -31,29 +31,34 @@ namespace kaldi {
...
-  virtual int32_t NumTransitionIds() const = 0;
+  int32_t NumTransitionIds() const {
+      return TransitionIdToPdfArray().size() - 1;
+  }
   virtual int32 NumPdfs() const = 0;
   virtual int32_t TransitionIdToHmmState(int32_t trans_id) const = 0;
 };
 
-}
+} // namespace kaldi
```
kkm000, line 62:
```suggestion
}  // namespace kaldi
```

src/lat/lattice-functions-transition-model.cc
```diff
@@ -136,4 +136,4 @@ bool CompactLatticeToWordProns(
   }
 }
 
-} // end namespace kaldi
+} // namespace kaldi
```
kkm000, line 139:
```suggestion
}  // namespace kaldi
```

src/itf/transition-information.h
```diff
@@ -38,27 +38,69 @@ namespace kaldi {
...
+   * but also graphemes, word pieces, byte pieces, etc. A better word
+   * would be Token. We use the word Phone for backwards compatibility
+   * with TransitionModel, which was designed only with phones in
+   * mind. Since TransitionInformation was added to subsume
+   * TransitionModel, we did not want to change the call site of every
+   * TransitionModel::TransitionIdToPhone to
+   * TransitionInformation::TransitionIdToToken.
    */
   virtual int32_t TransitionIdToPhone(int32_t trans_id) const = 0;
+  /**
```
kkm000: I encourage you to leave the `Phone` alone. It does not matter _what_ you call it, the user adapts to terminology. It's not linguistically a phone anyway. Most if not all lexicons (TIMIT is likely the only exception) are phonemic (cat = /kat/, kitten = /ˈkitən/, butter = /ˈbətər/), not phonetic (cat = [kʰæʔ], [kʰæt̚], kitten = ['kʰıʔn̩], butter = ['bʌɾɚ], ['bʌɾɹ̩]). So it's properly a `Phoneme`, not `Phone`. I fully subscribe to the de Saussure's notion that « le lien unifiant le signifiant et le signifié est arbitraire [ in a more precise sense, _immotivé_, as he remarks further — kkm] ». In other words, I give no fuzz about it. 

But what happened now, `Phone` is still in `TransitionIdToPhone`, but the previous method calls the same thing `Token`. If you _strongly_ feel that `Phone` is _very_ misleading and must be renamed `Token`, it should be consistentl. This will have a fallout among the existing Kaldi user base, so please take that into a consideration. I'd leave it alone just for convenience of leaving it alone. :) The Doxy comment at L55 still calls it a phone...

I'd also rather omit the long excuse/explanation why "Phone". The documentation should be concise and precise. Something like
```
   * Phone is a historical term, and should be understood in a wider sense that
   * also includes graphemes, word pieces, etc.: any minimal entity in your
   * problem domain which is represented by a sequence of transitions with
   * a PDF assigned to each of them by the model.
```
"Historical" says it all without a lengthy explanation that reads like "we're sorry, we did not think about it." This is not very informative.

I can live with the precise linguistic term "word piece." :) No problem here.

I understand you don't want to change all `TransitionIdToPhone` to `TransitionIdToToken` right now because of the sheer scope of the change. My preference is leave the thing called `Phone`. If you strongly feel it has to be changed, let's commit this one and then change it. But as is, the presence of both `Token` and `Phone` in the same interface and with the same semantics is very confusing (especially given that TM is pretty loaded with intimately interconnected things: phone, state ID, transition ID, PDF ID), and I'm not really excited by adding a synonym to one of the terms in this set.

kkm000 (changes requested):
Thanks, that's cool! I feel we're quickly converging on something. I'm not going to mark syle nits anymore. One day I'll unleash a hungry clang-format on them all.

There's only one comment. We need to come to a decision on this Token vs Phone issue. Somehow the thing is now called both names, and that's not really good. So this is the last remaining issue. What's your take on it? To me, the exact name is unimportant, as long as it's unique.

src/itf/transition-information.h
galv: Phone it is. I value backwards compatibility too much to consider changing to Token.

Thanks for pointing out that `TransitionIdIsStartOfToken` should be consistent with TransitionIdToPhone.

galv:
This looks good to me at this point. I used it inside my internal build (and caught an issue: https://github.com/kaldi-asr/kaldi/pull/4610/commits/c5af1bac3419fbfe250712a917ff8f5729e525f0) and saw that the new lattice tools produce sensible output with ctc models. Let me know if there are remaining concerns.

src/itf/transition-information.h
kkm000: Cool! So I am merging this as soon as Travis passes? We build there with CUDA off, let's see if anything pops out.

kkm000 (approved):

PR #4614, opened 2021-08-20
fix_data_dir.sh in perturb_data_dir_speed.sh

lubacien:
Adding this removes an error that your data is not sorted, in my case.


danpovey:
OK, I can see how this might happen if the speaker ids were lower case.

PR #4618, opened 2021-08-30
use CompactLattice call GetLinearSymbolSequence use weight.String() out of order

datemoon:
modefy GetLinearSymbolSequence function, when you use CompactLattice,if you get weight and you use weight.String() get aligmnets will be reversed order ,change times parameter order and make sure weight.String() is same aligments.


galv:
This makes sense to me. I don't think we use any other fst weight types that are not commutative over multiplication, so it shouldn't break anything.

datemoon:
> I don't think we use any other fst weight types
yes, In kaldi coding , all code don't use CompactLattice weight.String() get aligment.When I call GetLinearSymbolSequence, I pass 
CompactLattice  as parameter, and I hope get aligments , but aligment is same word, so I have to use weight.String() get aligment , and at this moment I found weight.String() is out of oider. This problem don't cause kaldi problem, I came across it when I was redeveloping. So I hope fix tihs problem and so I don't have to modify th kaldi coda when I redeveloping.

PR #4621, opened 2021-09-03
[src] cublasGetStatusString overloaded

hugovbraun:
cublasGetStatusString will be declared in cublas headers in future cuda revisions. Renaming the Kaldi version to avoid collision (we need to keep it around for older cuda versions)


kkm000:
LGTM, thanks!

PR #4622, opened 2021-09-03
Make logits in cudadecoder const.

galv:
There is no reason why they couldn't be const. It was just an
oversight in the interface.


kkm000:
LGTM, thanks, merging!

PR #4623, opened 2021-09-03
Remove dependency of cuda decoder on cudamatrix and TransitionModel.

galv:
Instead, it depends on TransitionInformation.

We replace CuDevice::Insantiate calls withs raw cudaMalloc calls. This
shouldn't be a big deal, since global memory allocations occur only in
destructors. This could break users who have given >99% of the the
available memory to the global allocator, though.

With this, non-HMM based models can use the cuda decoder without
depending upon several kaldi dependencies.

Also remove cudadecoder's dependency on online endpointing. This is
simple enough dependency that I just copy and paste it into cudadecder/


danpovey:
Thanks for this and your other PRs!!
I'll wait for Kirill to check and merge though.

src/cudadecoder/cuda-fst.cc
```diff
@@ -106,27 +107,27 @@ void CudaFst::ComputeOffsets(const fst::StdFst &fst) {
...
 
 void CudaFst::AllocateData(const fst::StdFst &fst) {
-  d_e_offsets_.reset(static_cast<uint32*>(CuDevice::Instantiate().Malloc(
-      (num_states_ + 1) * sizeof(*d_e_offsets_))));
-  d_ne_offsets_.reset(static_cast<uint32*>(CuDevice::Instantiate().Malloc(
-      (num_states_ + 1) * sizeof(*d_ne_offsets_))));
-  d_final_.reset(static_cast<float*>(CuDevice::Instantiate().Malloc(
-      (num_states_) * sizeof(*d_final_))));
+  void *temp_pointer;
+  CU_SAFE_CALL(cudaMalloc(&temp_pointer, (num_states_ + 1) * sizeof(uint32)));
```
csukuangfj, line 111: Replacing `Malloc` with `cudaMalloc` will not use the cache provided by `CuMemoryAllocator`. Will it have any side effects with such a replacement?

galv: This allocation happens only in constructors and destructors, so there will be no performance problems. The only possible effect may be a failure to allocate, if the custom memory allocator owns too much memory. See "This could break users who have given >99% of the the
available memory to the global allocator, though." in my initial comment. You could solve this by decreasing this value: https://github.com/kaldi-asr/kaldi/blob/e6cc1eef1f6c55936d36e8febe9baff6c479f81b/src/cudamatrix/cu-allocator.h#L74

kkm000: I do not remember what happens when the initial default allocation of 50% runs out? Will the allocator grab all available memory, or some part of it?

Besides, what is wrong with the allocator? I certainly don't like what it does with per-thread instantiation (I had to hack it to explicitly uninitialize every thread to use cuda-memcheck; even when I disabled it, it held handles to cudaBlas and other stuff). Once in a while I'm thinking of reworking its per-thread TLS use and syntax (this `CuDevice::Instantiate().Malloc` everywhere is ugly), but never finding a good reason. Or excuse. 

Here it's really unimportant, CudaFst constructor takes ungodly CPU time (2-3 seconds in a well-optimized build). But the allocator is used in other places too, I think. Are you planning to replace it everywhere in cudadecoder/?

galv: > Besides, what is wrong with the allocator?

Basically, the allocator prevents us from using the low level parts of cudadecoder codebase as a library for external codebases. When I say low-level, I mean cuda-fst.cc, cuda-decoder.cc, cuda-decoder-kernels.cu, and 
lattice-postprocessor.cc. It prevents us from doing so because it (Or CuDevice, rather than CuAllocator) makes a hard assumption that there is only one GPU per process. This means that we could not run the cudadecoder in pytorch when it is using multiple GPUs, for example.

There is another, smaller problem caused by the allocator when using this as a standalone library: Memory fragmentation is more likely. The user could set "cuda-memory-proportion" too high, which makes that memory unavailable to the rest of your application. There isn't a straightforward, easy way to check for this sort of error, but it can cause a performance degredation. By using cudaMalloc(), we allocate only that memory which is necessary.

These are the only locations in the cuda-fst.cc, cuda-decoder.cc, cuda-decoder-kernels.cu, and 
lattice-postprocessor.cc that use the cuda allocator, so I am basically allowing this part of the codebase to become a standalone library. Note that we don't need to remove usage of the allocator from the nnet3-pipeline*.cc files. Those already have strong dependencies on other kaldi parts (like nnet3), so there is no real benefit from doing so.

In general, the slowdown caused by cudaMalloc (it is a synchronous call, meaning that the call won't return until all previous enqueue operations and it complete) is not a huge problem when it is called within the constructor of an object (as is the case for CudaFst), since this happens only once. A caching allocator is useful in cases like deep neural network training where you have to allocate different sized chunks in each layer and don't want to suffer the synchronous stop caused by cudaMalloc().

danpovey: I'm OK with the way this is done using cudaMalloc.

src/cudadecoder/online-endpoint.cc
```diff
@@ -0,0 +1,71 @@
+// online2/online-endpoint.cc
```
kkm000, line 1:
```suggestion
// cudadecoder/online-endpoint.cc
```

src/cudadecoder/online-endpoint.h
```diff
@@ -0,0 +1,173 @@
...
+#include <string>
+#include <vector>
+#include <deque>
+
+#include "base/kaldi-types.h"
+#include "itf/options-itf.h"
+#include "util/parse-options.h"
+
+namespace kaldi {
+/// @addtogroup  onlinedecoding OnlineDecoding
```
kkm000, line 34: Please put this into a different group. If there is none yet, `cudadecoder` will do.

src/cudadecoder/online-endpoint.h
```diff
@@ -0,0 +1,173 @@
...
+   [ Note: we're changing the decoders so that GetBestPath() is efficient
+   and does not require operations on the entire lattice. ]
+
+   For details of the default rules, see struct OnlineEndpointConfig.
+
+   It's up to the caller whether to use final-probs or not when generating the
+   best-path, i.e. whether to call decoder.GetBestPath(&lat, (true or false)),
+   but my recommendation is not to use them.  If you do use them, then depending
+   on the grammar, you may force the best-path to decode non-silence even though
+   that was not what it really preferred to decode.
```
kkm000, line 77: This needs an update.

src/cudadecoder/online-endpoint.cc
```diff
@@ -0,0 +1,71 @@
...
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cudadecoder/online-endpoint.h"
+
+namespace kaldi {
```
kkm000, line 22: I do not think we can get away with duplicate symbols for struct and function names. The simplest way is to put this simplified version into the same `kaldi::cuda_decoder` namespace as everything else here.

Can you think of a way to reuse code at least partially? E.g., split the original file into online-enpoint-lite and online-endpoint? At the very least, add a comment that this is a tear-off  copypasta of `online2/online-endpoint.h`. But of we can avoid cloning 500 lines, it would be much better.

src/cudadecoder/online-endpoint.h
```diff
@@ -0,0 +1,173 @@
...
+//
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY IMPLIED
+// WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+// See the Apache 2 License for the specific language governing permissions and
+// limitations under the License.
+
+
+#ifndef KALDI_ONLINE2_ONLINE_ENDPOINT_H_
```
kkm000, line 22:
```suggestion
#ifndef KALDI_CUDADECODER_ONLINE_ENDPOINT_H_
```

kkm000 (changes requested):

src/cudadecoder/cuda-fst.cc
kkm000: I see. The allocator is not used anywhere in cudadecoder/ except these places. Then it's certainly alright. And what about the cudafeat/ library?

galv: cudafeat/ library does use the caching allocator outside of constructors. See here: https://github.com/kaldi-asr/kaldi/blob/3b8a97d859464912ddc50877f7280048f654c275/src/cudafeat/online-ivector-feature-cuda.cc#L300

Not only that, there may be transitive dependencies on CuAllocator because of the usage of CuMatrix in that directory. So it would certainly be more complicated to "librarize" the cuda feature extractor software.

galv:
I redid this to remove cudadecoder/online-endpoint.cc. It seemed better to jut make online2/online-endpoint.cc work as is (well, almost as is; I had to remove the dependency on TransitionModel), which means no code duplication.

src/cudadecoder/cuda-fst.cc
kkm000: @galv,
> cudafeat/ library does use the caching allocator outside of constructors. [...] Not only that...

So what do you think would work the best for you? It's hard to build a decoder without this library: the only option is extracting features on the host. Is my understanding correct: the only _blocker-level_ problem is our lack of support for multiple devices in one process? It would be cool to retain its speed where allocations come and go, and make it working for you at the same time. I already said I'm looking for an excuse to rework it a bit... :) 

You mentioned fragmentation, and that's fair, but show me an efficient allocator not affected by it. I productized Kaldi a few times (with and without CUDA). On Windows, I got bitten by the new allocator in VS 2013, which minimized fragmentation at a huge perf hit (I switched to TBBmalloc), and the GNU libc one, which suffers from memory creep to OOM in a few hours, and large perf hit with tunables set to prevent it (I switched to Jemalloc). BTW, is cudaMalloc immune to fragmentation, i.e., fully virtualized in hardware like mmap(3), but with a smaller equivalent of the "page size," the minimum allocation granularity?

By the way, I'm wondering why doesn't CUDA offer a suballocator like ours out of the box. I read TensorFlow code, and they RYO a very similar one. I'll be surprised if Torch had not have one as well.

kkm000:
> I redid this to remove cudadecoder/online-endpoint.cc.

Super, thanks! I'll go over it and likely merge tomorrow.

kkm000 (approved):

kkm000:
@galv, Please think of all the complaints about our architecture of the CuAllocator that you can only think of. We won't as easy chop off cudafeat/ as we did cudadecoder/.

At the worst, you can keep the allocator interface (the .h file) but substitute your own .cc file for its implementation. But when it comes to the creepy `CuDevice::Instantiate().Malloc` this becomes kinda impossible. Even simply changing _that thing_ to `CuAllocator::Malloc()` would make more sense of it. The code is ancient, and has evolved with CUDA API kinda, ehm, organically. I find its interface very confusing, starting with the call to `Instantiate()` which in fact does not instantiate anything, except once. When I read code, I always stumble over and grumble at the name.

But I want to think about the best, not the worst changes that would make sense in the context of Kaldi, too.

One question to consider is whether you contemplate e.g. placing memory on an explicit device. Currently you're still always using a single device and a single context everywhere. When you mention multi-device support, I'm wondering how far you do want to go.

kkm000:
@hugovbraun, could you please share your thoughts, too? Basically, what do we want from CuAllocator in Kaldi. @galv freed cudadecoder code from it, but, IMO, cudafeat/ would be much tougher, and the suballocation may be useful. I'm thinking of making changes to CuDevice/CuAllocator more palatable. Basically, the comment https://github.com/kaldi-asr/kaldi/pull/4623#discussion_r705900609 and the one right above. Like, if you had a magic wand, how you'd transform these two components so that they are the most useful to production cudadecoder-based software?

PR #4626, opened 2021-09-10
Fix linking with tcmalloc.

### LINKED ISSUE #4625: build kaldi with tcmalloc ，cannot find -ltcmalloc_minimal
i use tools/install_tcmalloc.sh  to build  tcmalloc 
and ./congure --enable-tcmalloc  
but  when make，this is  build error：
/opt/rh/devtoolset-4/root/usr/libexec/gcc/x86_64-redhat-linux/5.3.1/ld: cannot find -ltcmalloc_minimal
why？

csukuangfj:
Fixes #4625

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -1307,7 +1307,7 @@ if $enable_tcmalloc ; then
     TCMALLOC_ROOT=$(rel2abs ../tools/gperftools)
     echo "Configuring tcmalloc since --enable-tcmalloc"
     if [ "$(uname)" == "Linux" ]; then
-      echo "LDLIBS += -Wl,-rpath=$TCMALLOC_ROOT/lib -ltcmalloc_minimal" >> kaldi.mk
+      echo "LDLIBS += -Wl,-rpath=$TCMALLOC_ROOT/lib -L$TCMALLOC_ROOT/lib -ltcmalloc_minimal" >> kaldi.mk
     else
       failure "Now, --enable-tcmalloc option is only supported for Linux"
     fi


src/configure
```diff
@@ -1307,7 +1307,7 @@ if $enable_tcmalloc ; then
     TCMALLOC_ROOT=$(rel2abs ../tools/gperftools)
     echo "Configuring tcmalloc since --enable-tcmalloc"
     if [ "$(uname)" == "Linux" ]; then
-      echo "LDLIBS += -Wl,-rpath=$TCMALLOC_ROOT/lib -ltcmalloc_minimal" >> kaldi.mk
+      echo "LDLIBS += -Wl,-rpath=$TCMALLOC_ROOT/lib -L$TCMALLOC_ROOT/lib -ltcmalloc_minimal" >> kaldi.mk
```
csukuangfj, line 1310: The CI should fail without `-L$TCMALLOC_ROOT/lib`.

kkm000: tcmalloc It's not enabled in CI. But your fix is just too correct not to merge it right away, thanks. :)

LvHang:
@csukuangfj , thank you Fangju, you're right. I miss the link-time error when I tested since the library has existed in my '/usr/local/lib' which is included in the '/etc/ld.so.conf' file.

Just leave some information here:
ld(1):
the ```-rpath```: (1) adds a directory to the runtime library search path. (2) is used when locating shared objects which are need by shared objects explicitly included in the line.
the ```-L```: adds a directory to the lists of paths that ```ld``` will search for archive libraries and ```ld``` control scripts.
gcc(1):
```-L```: For GNU linker, the directories searched include several standard system directories plus any that you specify with ```-L```.
```-Wl,option```: passes 'option' as an option to the linker. It seems that the ```-Wl,-L=dir``` can work as ```-Ldir``` also.

PR #4631, opened 2021-09-24
At the end of Finalize，When setting LatticeCallback, clean up the mapping of corr_id to channel in corr_id2 channel_

ryanwy:
if set LatticeCallback, at the end of Finalize, it is also necessary to clean up the corr_id to channel mapping in corr_id2 channel_. Otherwise, when TryInitCorrID is called again, no new channel will be allocated

### DIFF
--- a/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
+++ b/src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
@@ -479,9 +479,9 @@ void BatchedThreadedNnet3CudaOnlinePipeline::RunCallbacksAndFinalize(
         } else {
           // All done with this corr_ids. Cleaning up
           available_channels_.push_back(ichannel);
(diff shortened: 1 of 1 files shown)


kkm000:
Thanks Kitty, I think you caught a good edge case!

LGTM, but with a but. There is a pending large PR #4553 with a lot of changes touching the same area, from @hugovbraun, the original author and _de facto_ owner of this code. I'll leave it up to him to decide to merge your change as is, or incorporate into that quite a hefty PR to avoid merge conflicts. I just hold all changes to cudadecoder/ before that large PR is merged.

@hugovbraun, could you please review this CL?

/cc @galv

hugovbraun:
Great catch from both of you, it is indeed a bug fixed in #4553. We can merge this PR though, it's cleaner to have the fix in an atomic commit anyway.

kkm000:
@hugovbraun, thanks for taking a look. Yeah, #4553 continues to grow. :) I'd rather merge it sooner than later.

@ryanwy, thanks for the fix!

PR #4633, opened 2021-09-24
update and rebase #3951

jtrmal:
I just removed changes in compute-wer.cc that were including duplicate functionality


jtrmal:
checks ok, merging

PR #4635, opened 2021-09-28
[egs] If utt2num_frames already exists, don't run feat-to-len.

galv:
There are presumably places other than utils/subset_data_dir.sh that
can use this.

utt2num_frames can be produced by passing --write_utt2num_frames=true
to steps/make_mfcc.sh. This is much faster than running feat-to-len
after the fact.


kkm000:
LGTM, thanks, good catch! We once decided to always generate utt2dur and utt2num_frames, but there should have been a couple places I missed...

PR #4636, opened 2021-09-28
[egs] train_lda_mllt.sh: Fix out-of-memory errors when training on large datasets.

galv:
Previously, gmm-acc-mllt would load the entirety of the alignments
into memory. When training on very large datasets like The People's
Speech (>20,000 hours), this can result in each process (I was using 16) loading ~18 GiB into
memory. However, both the features and the alignments are in sorted
order (because they are derived from wav.scp and text, which start
sorted). Therefore, I add annotations to the alignments archive
specifier to ensure that values are discarded aftering being read ("o"
is unnecessary here for reducing memory usage but I add it anyway
because it's true).


kkm000:
LGTM

PR #4641, opened 2021-09-30
segments related bug fix

qmpzzpmq:
Please check it
I followed the description here:
>          "Extract segments from a large audio file in WAV format.\n"
>          "Usage:  extract-segments [options] <wav-rspecifier> <segments-file> <wav-wspecifier>\n"
>          "e.g. extract-segments scp:wav.scp segments ark:- | <some-other-program>\n"
>          " segments-file format: each line is either\n"
>          "<segment-id> <recording-id> <start-time> <end-time>\n"
>          "e.g. call-861225-A-0050-0065 call-861225-A 5.0 6.5\n"
>          "or (less frequently, and not supported in scripts):\n"
>          "<segment-id> <wav-file-name> <start-time> <end-time> <channel>\n"
>          "where <channel> will normally be 0 (left) or 1 (right)\n"
>          "e.g. call-861225-A-0050-0065 call-861225 5.0 6.5 1\n"
>          "And <end-time> of -1 means the segment runs till the end of the WAV file\n"
>          "See also: extract-feature-segments, wav-copy, wav-to-duration\n";

https://kaldi-asr.org/doc/extract-segments_8cc.html

### DIFF
--- a/egs/wsj/s5/utils/fix_data_dir.sh
+++ b/egs/wsj/s5/utils/fix_data_dir.sh
@@ -97,10 +97,9 @@ function filter_recordings {
     utils/filter_scp.pl $data/wav.scp $tmpdir/recordings > $tmpdir/recordings.tmp
     mv $tmpdir/recordings.tmp $tmpdir/recordings
 
-
-    cp $data/segments{,.tmp}; awk '{print $2, $1, $3, $4}' <$data/segments.tmp >$data/segments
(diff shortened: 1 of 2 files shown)


danpovey:
I think this is an attempt to have the "segments" file support a channel-id, like is supported in the binary extract-segments.
The concept is that a recording-id represents a particular channel, and for stereo data, you have to use sox commands in the wav file to extract the appropriate channel.
It would cause a bunch of other scripts to have problems if we were to support an extra field in the segments file.
Closing, sorry!

PR #4643, opened 2021-10-06
add 'src/cudafeatbin' to common_path.sh

qmpzzpmq:


galv:
@kkm000 We should approve this. I have a change just like it in my personal workspace right now.

kkm000:
LGTM, thanks @qmpzzpmq! (🤔 How can you only _pronounce_ that???)

@galv, thanks for looking. Please check if you have other changes that are useful to most users in your workplace! ;-)

PR #4645, opened 2021-10-09
[src] Optimize lattice-faster-decoder to use memory pools for tokens and links

danlark1:
Hi, this is a follow up optimization from EasyPerf contest https://easyperf.net/blog/2021/07/16/Performance-analysis-and-tuning-contest-5

I managed to optimize the performance by 13.5% in benchmark provided in
the contest.

We presented the (not so many) solutions in https://www.youtube.com/watch?v=jQG3QnhunZg.

I have couple of more things to optimize, for example, changing
nth_element to a little faster algorithm and prefetching opportunity in
ProcessEmitting. However, they totally do optimize at most 0.5% and
memory allocator is the most elegant, easy to understand approach to
save some cycles. But please do let me know if you want to see them.

In contest I used my implementation of memory allocator, however, I
found that openfst has its own which perfectly fits so that we can
use it in this decoder.

If you know other places where I should replace raw news and deletes
with the memory pool, let me also know.

### DIFF
--- a/src/decoder/lattice-faster-decoder.cc
+++ b/src/decoder/lattice-faster-decoder.cc
@@ -33,8 +33,8 @@ LatticeFasterDecoderTpl<FST, Token>::LatticeFasterDecoderTpl(
       delete_fst_(false),
       config_(config),
       num_toks_(0),
-      token_pool_(config.memory_pool_size),
-      forward_link_pool_(config.memory_pool_size) {
+      token_pool_(config.memory_pool_tokens_block_size),
+      forward_link_pool_(config.memory_pool_links_block_size) {
   config.Check();
   toks_.SetSize(1000);  // just so on the first frame we do something reasonable.
 }
@@ -46,8 +46,8 @@ LatticeFasterDecoderTpl<FST, Token>::LatticeFasterDecoderTpl(
       delete_fst_(true),
       config_(config),
       num_toks_(0),
-      token_pool_(config.memory_pool_size),
-      forward_link_pool_(config.memory_pool_size) {
+      token_pool_(config.memory_pool_tokens_block_size),
+      forward_link_pool_(config.memory_pool_links_block_size) {
   config.Check();
   toks_.SetSize(1000);  // just so on the first frame we do something reasonable.
 }
--- a/src/decoder/lattice-faster-decoder.h
+++ b/src/decoder/lattice-faster-decoder.h
@@ -50,7 +50,10 @@ struct LatticeFasterDecoderConfig {
   // tokens as we go.
   BaseFloat prune_scale;
 
-  int32 memory_pool_size;
+  // Number of elements in the block for Token and ForwardLink memory
+  // pool allocation.
+  int32 memory_pool_tokens_block_size;
+  int32 memory_pool_links_block_size;
 
   // Most of the options inside det_opts are not actually queried by the
   // LatticeFasterDecoder class itself, but by the code that calls it, for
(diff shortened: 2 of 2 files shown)


danlark1:
@jtrmal, FYI

jtrmal:
nice, thanks!
Looks great, IMO.
@kkm do you wanna review or merge right away?

danlark1:
Also note this implementation is not thread safe within one instance of the decoder

arlofaria:
Firstly, it might help to clarify that there seem to be at least 2 different versions of "TCMalloc":

1. https://github.com/gperftools/gperftools: this is what was recently added to Kaldi by @LvHang, although it seems like it might be a somewhat outdated project ([possibly described by this page last updated in 2007](https://gperftools.github.io/gperftools/tcmalloc.html)).
2. https://github.com/google/tcmalloc: this was more recently [released in 2020](https://abseil.io/blog/20200212-tcmalloc#read-more) and implements what's described in [the Google paper](https://research.google/pubs/pub50370/) referenced by @danlark1.

I've been building with the older TCMalloc and found that the memory pool introduced in this PR actually makes things a bit slower and uses slightly more memory.  I'll investigate further, but just checking: has anyone tested the interaction of using either of the TCMalloc libraries along with this memory pool?

However, I can also confirm that without (the older) TCMalloc, this memory pool indeed speeds up decoding! 👏 

Also, @danlark1 asked:
> If you know other places where I should replace raw news and deletes with the memory pool, let me also know.

Perhaps it could help in Kaldi's lattice determinization code?

Alchemist2077:
I have similer test result：
memory pool version is little faster then No menory pool version with glibc;
memory pool version is little slower then No menory pool version with tcmalloc(gperftools version);

kkm000:
@arlofaria, I've had success with Jemalloc on Linux and TBB Allocator on both Linux and Windows.

Jemalloc in the default configuration is not faster than glibc malloc, but it's consistent and avoids the server's "middle age crisis" that glibc runs into (a lot of nearly empty allocated page pools that are held by long-living allocation, the good old pool fragmentation problem but up a level). 

My timing in glibc benchmarks was inconsistent, too, and Jemalloc fixed that. I do not have graphs here, but e.g. 24 hour test with a ~5 min repeated exactly same load over and over, 700+ samples on 4 machines did not pass normal distribution statistics test even close, neither combined nor per machine, and when I graphed them all 4, it was all kid's scribbles with 4 colored pens, not Gaussians. 48 hours ran into OOM, exceeding memory than I set up on test machines (cloud), and memory consumption never stopped growing. Changing its tuning to prevent unbounded growth made it a quarter slower, but didn't make the perf predictable.

"Old" tcmalloc and tbbmalloc are both designed with heavy multithreading in mind. They are fast and predictable with out-of-the-box settings, with a comparable performance. But you pay a memory use penalty for that, they are not as compact. However,  I did not ever ran into the same unbounded growth problem with tbbmalloc, but have not a similar experimental measurement experience with tcmalloc. I never tried the "new" tcmalloc , because you need Bazel, Python and a few other major cities and large animals just to build it. Some people would [script this build](https://github.com/kkm000/tty_resize/blob/master/Makefile) with Bazel. Еv'ry man buckles his belt his ain gate.

To summarize: IMO, Jemalloc is the best for general use, and tcmalloc or tbbmalloc for the cases of the high rate of thead-affine allocations. On Windows, I'm using tbbmalloc by default, just because it's easy to set up: it comes with TBB, and Jemalloc requires compiling from source. All 3 allocators make the memory consumption counts shown by e.g. top or ps a  less useful figure, as they all delay releasing pages back to the OS.

arlofaria:
Thanks for sharing these tips!  Jemalloc does seem like a good recommendation, given the various tradeoffs and practical considerations (including ease of integration).  As a couple more observations: I've found that the old tcmalloc is particularly slow for systems with multiple physical CPU sockets; and I didn't go too far with tbbmalloc because it seems slightly complicated or not recommended for static linking.

PR #4648, opened 2021-10-22
CUDA feature generation: warn if the frequencies do not match

jtrmal:
@danpovey @kkm000 for your comments -- I want to try to add the on-the-fly resampling but that will impact the speed of the code (and what portion can be run on gpu). Any opinions if it's worth the effort or the error is enough?
Anyway, we will have to modify the other binaries in cudafeatbin/ as well, not just the mfcc

### DIFF
--- a/src/cudafeatbin/compute-fbank-feats-cuda.cc
+++ b/src/cudafeatbin/compute-fbank-feats-cuda.cc
@@ -66,7 +66,7 @@ int main(int argc, char *argv[]) {
       po.PrintUsage();
       exit(1);
     }
-    
+
     g_cuda_allocator.SetOptions(g_allocator_options);
     CuDevice::Instantiate().SelectGpuId("yes");
     CuDevice::Instantiate().AllowMultithreading();
@@ -76,7 +76,7 @@ int main(int argc, char *argv[]) {
 
     std::string output_wspecifier = po.GetArg(2);
 
-    // Fbank is implemented via the MFCC code path 
+    // Fbank is implemented via the MFCC code path
     CudaSpectralFeatures fbank(fbank_opts);
 
     SequentialTableReader<WaveHolder> reader(wav_rspecifier);
@@ -88,7 +88,7 @@ int main(int argc, char *argv[]) {
                    "needed if the vtln-map option is used.");
     RandomAccessBaseFloatReaderMapped vtln_map_reader(vtln_map_rspecifier,
                                                       utt2spk_rspecifier);
-    
+
     if (output_format == "kaldi") {
       if (!kaldi_writer.Open(output_wspecifier))
         KALDI_ERR << "Could not initialize output with wspecifier "
@@ -106,6 +106,11 @@ int main(int argc, char *argv[]) {
       num_utts++;
       std::string utt = reader.Key();
       const WaveData &wave_data = reader.Value();
+      if (wave_data.SampFreq() != fbank_opts.frame_opts.samp_freq) {
+        KALDI_ERR << "File: " << utt << " has an incompatible sampling "
+          << "frequency (config= " << fbank_opts.frame_opts.samp_freq
+          << " vs file=" << wave_data.SampFreq() << ".";
+      }
       if (wave_data.Duration() < min_duration) {
         KALDI_WARN << "File: " << utt << " is too short ("
                    << wave_data.Duration() << " sec): producing no output.";
--- a/src/cudafeatbin/compute-fbank-feats-cuda.cc
+++ b/src/cudafeatbin/compute-fbank-feats-cuda.cc
@@ -22,6 +22,8 @@
 #include "feat/wave-reader.h"
 #include "cudamatrix/cu-matrix.h"
 #include "cudamatrix/cu-vector.h"
+
+
 int main(int argc, char *argv[]) {
   try {
     using namespace kaldi;
@@ -76,7 +78,7 @@ int main(int argc, char *argv[]) {
 
     std::string output_wspecifier = po.GetArg(2);
 
-    // Fbank is implemented via the MFCC code path
+    // Fbank is implemented via the MFCC code path.
     CudaSpectralFeatures fbank(fbank_opts);
 
     SequentialTableReader<WaveHolder> reader(wav_rspecifier);
@@ -107,8 +109,8 @@ int main(int argc, char *argv[]) {
       std::string utt = reader.Key();
       const WaveData &wave_data = reader.Value();
       if (wave_data.SampFreq() != fbank_opts.frame_opts.samp_freq) {
-        KALDI_ERR << "File: " << utt << " has an incompatible sampling "
-          << "frequency (config= " << fbank_opts.frame_opts.samp_freq
+        KALDI_ERR << "File: " << utt << " has an mismatched sampling "
+          << "rate (config= " << fbank_opts.frame_opts.samp_freq
           << " vs file=" << wave_data.SampFreq() << ".";
       }
       if (wave_data.Duration() < min_duration) {
--- a/src/cudafeatbin/compute-fbank-online-batched-cuda.cc
+++ b/src/cudafeatbin/compute-fbank-online-batched-cuda.cc
@@ -34,7 +34,7 @@ using namespace kaldi;
 
 // This class stores data for input and output for this binary.
 // We will read/write slices of this input/output in an online
-// fasion.
+// fashion.
 struct UtteranceDataHandle {
   std::string utt;
   WaveData wave_data_in;
@@ -146,7 +146,7 @@ int main(int argc, char *argv[]) {
     int32_t ldf = max_chunk_frames;
 
     CudaOnlineBatchedSpectralFeatures fbank(feature_opts, max_chunk_frames,
-                                           num_channels, num_lanes);
+                                            num_channels, num_lanes);
 
     int32_t feat_dim = fbank.Dim();
 
@@ -187,8 +187,8 @@ int main(int argc, char *argv[]) {
       std::string utt = reader.Key();
       WaveData &wave_data = reader.Value();
       if (wave_data.SampFreq() != feature_opts.frame_opts.samp_freq) {
-        KALDI_ERR << "File: " << utt << " has an incompatible sampling "
-          << "frequency (config= " << feature_opts.frame_opts.samp_freq
+        KALDI_ERR << "File: " << utt << " has an mismatched sampling "
+          << "rate (config= " << feature_opts.frame_opts.samp_freq
           << " vs file=" << wave_data.SampFreq() << ".";
       }
 
@@ -266,7 +266,7 @@ int main(int argc, char *argv[]) {
 
       // process batch
       fbank.ComputeFeaturesBatched(d_lanes, lanes.size(), d_batch_wav_in,
-                                  sample_freq, vtln_warp, &d_batch_feats_out);
+                                   sample_freq, vtln_warp, &d_batch_feats_out);
 
       // copy feats to host
       cudaMemcpyAsync(h_batch_feats_out.Data(), d_batch_feats_out.Data(),
--- a/src/cudafeatbin/compute-fbank-online-batched-cuda.cc
+++ b/src/cudafeatbin/compute-fbank-online-batched-cuda.cc
@@ -146,7 +146,7 @@ int main(int argc, char *argv[]) {
     int32_t ldf = max_chunk_frames;
 
     CudaOnlineBatchedSpectralFeatures fbank(feature_opts, max_chunk_frames,
-                                            num_channels, num_lanes);
+                                           num_channels, num_lanes);
 
     int32_t feat_dim = fbank.Dim();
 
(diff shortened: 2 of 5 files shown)


kkm000:
I dunno if it's worth it, but it will certainly affect the CPU load much less than you seem to assume. The worst multirate conversion is 44.1 to 48 KHz (147/160 = (3×7^2)/(2^5×5)), and then you need to go a few times the ratio higher. I prefer IIR over FIR, as they are provably convergent, computationally efficient, even if you oversample to a more rational ratio, 'cus your multirate filter does not have to be as steep as it would with less oversampling. Note that @danpovey resampling code uses FIR, not requiring the approximation calisthenics but much more computationally intensive. And if you have multiple channels going at the same time, AVX2 and AVX512 do wonders (16 parallel channels per FMU). There's a reason Intel dropped sound processing primitives from the IPP: a good compiler does the same thing as good.

And yes, these algorithms are very yielding to the GPU processing (the amount of computation per sample is strictly fixed), but I'm not sure if it's helpful, unless you want to run a few hundreds of audio channels at once.

jtrmal:
I'm concerned mostly about managing people's expectations, about using
gpu/cpu I guess -- I think you are right about the speed/performance, since
we do not do any of those crazy polyphase resamplers.
y.

kkm000:
BTW, if you assume for the above ratio that 44100/48000≈\frac{3^2}/{2×5} (within 2%), you can do much better with SRC. The 2% difference will hardly affect the trained model in the end: the margin of various speakers natural F_1 pitch and syllable rate is way above this margin.

kkm000:
> we do not do any of those crazy polyphase resamplers

C'mon. It's the simplest thing to implement.

kkm000:
@Arlo, thanks, that's very helpful! "Sinc" likely implies a FIR? In principle, there is no limit to quality. I'm wondering were you up or downsampling? I'm asking b/c I once observed a weird effect: upsampling both training data and test set yielded a better WER. I used default sox FIR settings, and compared on telephony speech, as is vs 2x upsampled. Likely indicative of a suboptimal model (you can't extract information that's not in the signal, after all), but I could not find hyperparameters that would equalize out the effect.

That was back at nnet2 times, so I didn't play with different topologies, it wasn't that easy back then 

What was your experiment setup, if you don't mind my asking?

arlofaria:
Yes, I think "sinc" implies FIR in this case.

This was an nnet3 model trained from a fairly standard Switchboard+Fisher recipe (except no ivectors, and an LM trained on Web text) that should achieve about 5-10% WER with well-matched US English telephony test data.

But we tested on mismatched data: a small collection of diversely accented English from online videos of various quality.  These were MP4 files and the audio stream was AAC-compressed, I think at 96kbps CBR and a sample rate of maybe 44.1kHz.  The audio stream was converted to a WAV file using `ffmpeg`; it was further downsampled somehow to 16kHz, although I can't recall if that was done via `ffmpeg` or `sox`.  These were presented to the ASR system, which further downsampled with SRC to 8kHz to match the model sample rate.  So it's not a very clean experiment since the input was already decompressed and downsampled.  But what surprised me was that merely changing the SRC presets actually made a significant difference in accuracy:
```
                        WER
SRC_SINC_BEST_QUALITY:  24.7
SRC_SINC_FASTEST:       25.7
```

Hope that helps, and please do take these results with a grain of salt!  Ultimately our system design decision was to set the default at `SRC_SINC_FASTEST` anyway, but we allow an advanced option to change the preset.  For example:
```
# Download a 16kHz test file (recorded on Gen-1 AirPods on a MacBook, which is natively 16kHz).
wget mod9.io/hi.wav

# Send it to the ASR system, requesting a mismatched telephony model, but ensure best quality SRC.
(echo '{"asr-model":"en-US_phone","resample-mode":"best"}'; cat hi.wav) | nc mod9.io 9900
```
I can't seem to find a specific test file where the `resample-mode` makes a difference, but in theory it should be possible.

kkm000:
@arlofaria thanks, I understand now. The largest mismatch here, in my experience, is the types of voice compression (in the amplification dynamic sense, like an analog compressor of old times) used in telephony. It still does not explain a WER difference, but I would not hesitate to use the highest quality setting always. In the whole CPU consumption picture should have very little effect, especially when the multirate ratio is 1/2. 

I worked with mismatched telephony data it other way round: training a model for telephony from librispeech data. I "perturbed" the training data in multiple different ways, applying two different compression settings, used in VoLTE (essentially, RTP over data network, as LTE has not defined a voice channel on its own), plain old GSM compression, combination of these with pretty hot overload with hard limiter, and other DSP tricks (12-16 combinations in total) to get good results. These days, Fisher and Switchboard are dated, they are all landline recordings, which has different dynamics, but the calls we took were 95% cellular, so I tried to model modern telephony voice channel from minimally processed data; librispeech is good for that.

Another effect that makes sense to apply to training data is packet loss simulation, but I have no measurable results from this—a simple "honest" encode-drop-decode with real codecs changed the duration of audio, and then I never found time back then to ensure the pauses are correctly filled. These are also handled differently in the wideband WB-AMR decoder (VoLTE) than in GSM.

But, naturally, I paid zero attention to the SRC of librispeech 16 kHz data to 8kHz, which I tortured in next stages. It's, kinda, just SRC.

Many thanks for your recommendation of the library! I can play with it to understand the difference between the 3 quality settings. Many libraries for simple DSP algorithms carry over the concessions made at the time when a UP Pentium 4 machine was the mainstream. libsox indeed shows its age.

arlofaria:
> ... I would not hesitate to use the highest quality setting always. In the whole CPU consumption picture should have very little effect ...

Well, it's all relative :-)
For a single-threaded online ASR decoder, the 70xRT cost of best-quality SRC is indeed minimal.
But for batch-mode operation, we apply the SRC as a single-threaded pre-processing stage of the pipeline that prepares the audio stream for VAD, which will then distribute segments for further ASR decoding on multiple threads.  Unfortunately, the SRC becomes a bottleneck since the VAD operates at about 3,000 xRT.

arlofaria:
Oops, the speed measurements I quoted earlier appear to be incorrect!  I just re-measured more carefully now (not an ideal independent test of SRC, since it's still coupled to IO and ASR code) but I believe this is more accurate:
- `SRC_SINC_BEST_QUALITY` can be up to 115xRT (not 70xRT)
- `SRC_SINC_MEDIUM_QUALITY` can be up to 365xRT (not 150 xRT)
- `SRC_SINC_FASTEST` can be up to 790xRT (not 200xRT)
- `SRC_LINEAR` can be up to 3,750xRT (not 300xRT)

This seems more reasonable to me now, and hope it helps to clarify that SRC can be on the same order of speed as VAD.

jtrmal:
should be ready for review and merge

src/cudafeatbin/compute-fbank-feats-cuda.cc
```diff
@@ -106,6 +106,11 @@ int main(int argc, char *argv[]) {
       num_utts++;
       std::string utt = reader.Key();
       const WaveData &wave_data = reader.Value();
+      if (wave_data.SampFreq() != fbank_opts.frame_opts.samp_freq) {
+        KALDI_ERR << "File: " << utt << " has an incompatible sampling "
```
kkm000, line 110: To me, "incompatible" implies that there exists a sample rate which is "compatible" with that specified in the config file. While, strictly speaking it's true (there is exactly one), this usually suggests that there is more than than one. How about "incorrect", "unexpected", "bad" or  "mismatching"?

×4 in total.

src/cudafeatbin/compute-fbank-online-batched-cuda.cc
```diff
@@ -146,7 +146,7 @@ int main(int argc, char *argv[]) {
     int32_t ldf = max_chunk_frames;
 
     CudaOnlineBatchedSpectralFeatures fbank(feature_opts, max_chunk_frames,
-                                            num_channels, num_lanes);
+                                           num_channels, num_lanes);
```
kkm000, line 149: Oops. ×2 (the 2nd @ L256)
```suggestion
                                            num_channels, num_lanes);
```

src/cudafeatbin/compute-mfcc-feats-cuda.cc
```diff
@@ -22,6 +22,9 @@
 #include "feat/wave-reader.h"
 #include "cudamatrix/cu-matrix.h"
 #include "cudamatrix/cu-vector.h"
+
+
+
```
kkm000, line 27: nit: at least one too many. 1 or 2 are allowed.
```suggestion


```

src/cudafeatbin/compute-fbank-feats-cuda.cc
```diff
@@ -76,7 +76,7 @@ int main(int argc, char *argv[]) {
 
     std::string output_wspecifier = po.GetArg(2);
 
-    // Fbank is implemented via the MFCC code path 
+    // Fbank is implemented via the MFCC code path
```
kkm000, line 79: über-nit: period at end of sentence. [Yes, it's _that much_ strict](https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar)!

kkm000 (changes requested):
If you agree with any of the words choices I suggested, please use it and feel free to commit. The rest is only nitpicking on formatting.

kkm000:
@arlofaria 
>  VAD operates at about 3,000 xRT.

Holy Great Guacamole on a Pogo stick with a whistle!!! Is this... a GPU or a custom ASIC? :) Srsly, how do you do that, if not a trade secret, of course?

kkm000 (approved):

kkm000:
@arlofaria, if you interested: I gave the Rabbit code a quick look over. The high-quality setting uses about 350000 filter coefficients which I pulled into Mathematica and looked at its step and impulse responses. It gives a negligible pass-band ripple and the stop-band attenuation of -109dB (16-bit quantization noise is about ‒96db, so this is an overkill for ASR; it makes sense 24-bit integer or 32-bit float sample format only). The code is slow probably because it keeps the coefficients in one long array, taking every n-th during the processing passes, like `(for i=; ..; i+=n)`, which is a cache-buster. It could have been made many times faster if the coefficients were reshuffled into a matrix (after the frequency ratio is known, that's a one-time prep), and then the code rearranged a bit more to take advantage of the FMU (the unit  that implements AVX{,2,512}). Convolution is very efficient in the FMU if rewritten in a matrix op form, and compilers code patterns to emit the right code. Clang 13 emits even AVX512 instructions, and takes cache sizes into account if you give it a matching `-march=` switch.

So if it really bites you in terms of performance, you may patch its main loop. Even simply accessing elements in order, not in large `n` steps sometimes result in a 10 times boost in performance. This struck me as a nearly certain lossage of an efficient use of CPU hardware. (Clang 11 and above may also give you hits where to align data and rearrange loops so it can emit efficient AVX2 code, but this is more involved.) I'm talking about these loops:

https://github.com/libsndfile/libsamplerate/blob/4909e679b3e268a1403608cbd860f3895549b0d5/src/src_sinc.c#L393-L404

A careless use of an offset into the linear array of the coefficient and using the variable `increment` to jump over entries from the start to end essentially leaves you with the L2 cache being a deadweight. Rearranging the big array at runtime, after the increment is known, into two matrices (what they call the "left" and the "right" halves of the filter), such that the `increment_t increment, increment_t start_filter_index` in the function declaration become 1 and the the column number (assuming row-in-order layout), so that access is sequential, should not be hard after all the ratio computations are done above. The CPU predictor will quickly learn to refill L2 from L3 in advance, and Xeon's L3 is likely large enough to hold the whole coefficient array.

Don't trust me before you profile the code, but I do not see any other loops in audio path.

arlofaria:
@kkm000, wow that's a very impressive deep dive into the low-level code!  I'm not sure if that SRC project is very actively developed anymore; it seems to be a couple decades old and the original author might not be involved, though folks from libsndfile community are trying to maintain it.  Perhaps they might appreciate your expertise over there?

I was considering a more high-level approach to speed up SRC: https://github.com/libsndfile/libsamplerate/issues/6#issuecomment-957483544

> Is this ... a GPU or a custom ASIC? :) Srsly, how do you do that, if not a trade secret, of course?

Ha, nope, it's CPU and just an open-source **lib**rary **f**or **VAD** ;-)

kkm000:
@arlofaria, yeah, these things did not matter as much at the time when Pentium III was king. And integers were much faster than floats, indeed. Now I understand why they are so reserved about floating point. I thought all this convoluted arithmetics was done to guarantee every last bit of precision... But for the today's CPU, RAM is nearly a peripheral. An NVMe SSD drive is only 4-5 times slower than RAM!

> Perhaps they might appreciate your expertise over there?

Possibly. I wish there were at least two of me, tho... I have way too much on my plate already, unfortunately.

I'll have a look at that VAD library, thanks. Probably it's not a very sophisticated one? There are quite simple algorithms that take any burst of quasiperodic noise for voice. These are usually used in SIP phones, to save on bandwidth when there is positively no voice. When it comes to voice discrimination that does not trigger when you drop a pen on the table, that becomes more CPU-intensive. And when adapting on the fly to ignore that TV in background, 1000× RT is hardly reachable...

arlofaria:
No, that VAD is not too sophisticated but seems to be the industry-standard for WebRTC, I think.

For our use case it's actually better to tune it for high recall (i.e. mis-detecting noise as speech) rather than high precision (i.e. missing short-duration words or speech acts). We do not use it in a real-time scenario to activate low-latency online ASR decoding, for example as a voice assistant on an embedded device.  Instead, we use it as audio stream pre-processing to determine the non-speech regions at which to split the audio into batched segments for downstream ASR decoding distributed across concurrent processors on a relatively large server.

So we just need this VAD to be better than using a windowed energy threshold, and faster than the ASR speed x number of processors.  Since we can run ASR at upwards of 30xRT per CPU thread (with a very small AM, graph, and beams), and the largest servers can now have over 100 CPUs, that means VAD at 3,000xRT could become the bottleneck _in theory_, and maybe it should be optimized.  But _in practice_, the bottleneck will often be the latency and bandwidth of network-based storage, e.g. retrieving files from S3, especially if it's in a cold data tier and/or in a distant region ... and particularly if these are media files from which you need to decompress an audio stream while discarding the video stream.

PR #4649, opened 2021-10-23
[src] remove duplicate cleaning up operations of LatticeDeterminization

LvHang:
The "minimal_subset" of each "output_states_[i]" has been cleaned up in line 217-220. Remove the duplicate operations.

### DIFF
--- a/src/lat/determinize-lattice-pruned.cc
+++ b/src/lat/determinize-lattice-pruned.cc
@@ -223,10 +223,6 @@ template<class Weight, class IntType> class LatticeDeterminizerPruned {
          iter != initial_hash_.end(); ++iter)
       delete iter->first;
     { InitialSubsetHash tmp; tmp.swap(initial_hash_); }
-    for (size_t i = 0; i < output_states_.size(); i++) {
-      vector<Element> tmp;
-      tmp.swap(output_states_[i]->minimal_subset);
-    }
     { vector<char> tmp;  tmp.swap(isymbol_or_final_); }
     { // Free up the queue.  I'm not sure how to make sure all
       // the memory is really freed (no swap() function)... doesn't really


kkm000:
@LvHang, LGTM, thanks for spotting the repeated code fragment code fragment! I have a request for you. When I look at the code, it certainly shows its age. Apparently, most of it has been carried over from even pre-C++03 times. It deserves a good hot bath cleanup! If you have time to clean the whole file, that would be great. If not, that's fine too, I'll commit just your fix. We all have only so much time. Read notes below only if you are up to this janitorial job. 

---

When stdc++ was a toddler named STL, "copy and swap" and "swap and destruct" idioms were widely used. I buy it, because STL might have had neither a [`vector::clear() noexcept`](https://en.cppreference.com/w/cpp/container/vector/clear) method, nor a way to deallocate vector's memory.

```c++
{ vector<char> tmp; tmp.swap(isymbol_or_final_); }
```
(including that in the body of the loop the second copy of which you took out) are much more readable if rewritten as
```c++
isymbol_or_final_.clear();
```
But note that `clear()` does not shrink the vector's capacity. When the vectors are reused, it may be a good thing. Here, on the other hand, the intent is to deallocate the containers, for which `vector::shrink_to_fit()` (new in C++11) was added. You may opt for a small helper
```c++
namespace {

template <typename T>
void Deallocate(vector<T>* vec) {
  vec->clear();
  vec->shrink_to_fit();
}

}  //namespace
```
and use it like
```c++
    Deallocate(&isymbol_or_final_);
```
If you want to go fully standard-compliant to a iota, [`vector::shrink_to_fit()`](https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit) “...is a non-binding request to reduce `capacity()` to `size()`. It depends on the implementation whether the request is fulfilled,” although I've never seen an implementation that would not deallocate a zero-sized vector; the standard allows ignoring the call with minimal reduction in size in mind, e.g. from 1000 to 995 elements that may be rejected for efficiency. You can even keep the same idiom that is used in code currently, which guarantees deallocation, as long as it's swept under the carpet, into a separate helper (hidden from the public at large by an unnamed namespace, as above):

```c++
template <typename T>
void Deallocate(vector<T>* vec) {
  vector<int>().swap(*vec);
}
```
`tmp` does not have to be explicitly declared: an unnamed object will do. But still, it's much better readable, despite its being a one-liner, when hidden into its own function, which also scopes the unnamed object lifetime. I hope the modern compilers are smart enough to optimize out the actual construction of this object. That is what I'd probably go with.

As written, the code is noisy and hard to read.

There are also a lot of coding conventions that are violated. The really dangerous one is a single-statement for body on its own line, such as
https://github.com/kaldi-asr/kaldi/blob/6c85bcccbd83f68b2049cb8e0c5ac5086b5ec8bb/src/lat/determinize-lattice-pruned.cc#L202-L203

This one also uses post-increment `i++`, which is verboten (unless this is the intent); all increments in loops should be of the `++i` form. All taken together, this should be written as
```c++
 for (size_t i = 0; i < output_states_.size(); ++i) {
   delete output_states_[i];
 }
```

This
https://github.com/kaldi-asr/kaldi/blob/6c85bcccbd83f68b2049cb8e0c5ac5086b5ec8bb/src/lat/determinize-lattice-pruned.cc#L211-L214

should be just
```c++
      delete ifst_;
```
since deleting a null pointer is perfectly fine, and is a no-op.

Finally, the new `for` statement iterating construct designed to reduce monstrosities like
https://github.com/kaldi-asr/kaldi/blob/6c85bcccbd83f68b2049cb8e0c5ac5086b5ec8bb/src/lat/determinize-lattice-pruned.cc#L222-L224

to much less obtrusive
```c++
    for (auto& e : initial_hash_) delete e.first;
```
Please keep in mind the blocks: one-statement body is ok if on the same line, as above, but requires `{ ... }` when on its own line, even if it's still a single statement. That's one of the parts of coding style that I'm pushing the hardest for. The HeartBleed OpenSSL bug was caused by a misindentation of an accidentally duplicated `if()` single-statement body on a separate line, and the code passed all code reviews, including its scrutiny for security, _for 6 years_ with flying colors: it's really hard to spot!

jtrmal:
Only small comment -- Having function Deallocate taking a pointer to
something and then not "deleting" the pointer itself but something within
the pointed-to container seems like a bad naming to me :/
y.

kkm000:
@jtrmal, I thought about the name, and "deallocate" is a very standard stdc++ term. Every container has an allocator which allocates and deallocates internal memory for it. Do you have a better idea for the name?

Really, this is one of the places where I'd rather ignore the code style dictum and use a reference. Then the ambiguity disappears. Since it's a function that's local to the file, I think it's fine.

jtrmal:
DeallocateContent?
y.

kkm000:
DeallocateElements?
CleanAndShrink?

And how do you feel about conspiring against the coding style and use non-const reference here? It's clearly shows that the container itself is not deleted.

jtrmal:
Yeah, I like the first one best I guess.
Don't have strong/any opinion on the non-const ref :)
y.

kkm000:
To me, with a reference even unqualified `Deallocate` reads unambiguous:
```c++
// Code style exception: passing the argument by non-const reference improves
// readability compared to passing it by pointer.
template <typename T>
void Deallocate(vector<T>& vec) {
  vector<int>().swap(vec);
}
```
How's that by you? The signature clearly shows the vector is not `delete`d.

jtrmal:
yeah, I'm fine with that.
y.

LvHang:
Sorry for the late response. I will clean it up and test it this week. Bests.

danpovey:
sorry, merged this as it seemed a simple fix, and I think we have bigger problems than this in the lattice determinization, see
#4662

LvHang:
Hi all, sorry for the late reply. I had an accident so that I have suffered a minor fracture in my left hand.
Dan, I saw that "non-functional" determinization case. The process of "fstdeterminizestar" should be similar to latticedeterminzation. I will try to debug.

danpovey:
Thanks!

PR #4659, opened 2021-11-07
Allow override cxx/ar/ranlib when host is set

nshmyrev:
Currently when HOST environment is set, CXX, AR and RANLIB are hardcoded. This change allows to redefine them with environment variables which follows default automake practice.


jtrmal:
thanks, LGTM

PR #4653, opened 2021-10-27
CUDA 11.4+ in fact supports GCC 11

kkm000:
GCC 10 and 11 are supported starting from CUDA v11.4.

See https://docs.nvidia.com/cuda/archive/11.4.2/cuda-installation-guide-linux/index.html

PR #4666, opened 2021-11-23
remove sgmm2* from the bin target

jtrmal:
this will finally stop everything depending on portaudio. 
sgmm2 and online are in the ext depend and the way depends are compiled for ext depend will pull in portaudio issues into normal compilation


jtrmal:
@kkm000  at your leisure

kkm000:
I reworked this part in #4653, but it needs to be split into 2 PRs, and also breaks Travis, needs a script update. But we need a stopgap on this madness ASAP, thanks, I agree! :)

Issue #4661, opened 2021-11-10
Kaldi feedstock update on conda-forge


mmcauliffe:
Hello,

I've recently updated the [kaldi feedstock for conda-forge](https://github.com/conda-forge/kaldi-feedstock), and I was wondering if any Kaldi maintainers were interested in being added a maintainer there as well. 

To get it working with the conda-forge system, I made some [changes to the CMake build system](https://github.com/conda-forge/kaldi-feedstock/blob/master/recipe/cmake.patch).  To compile some notes on these:

* Kaldi on conda-forge doesn't link to `libfst.so`, but rather to `libfst.16.so` (to allow for installation beside the OpenFst conda-forge package which is 1.8.1)
* On Windows, OpenFst binaries are included as well as Kaldi binaries (OpenFst on Conda doesn't support Windows)
* Removed building of `online` and `online2` on Windows, as I was getting some compiler errors
* Added support for building `fgmmbin` on all platforms

I can make a PR for the ones that would be helpful generally (i.e. `fgmmbin` included in the CMake build).

jtrmal:
Yes, would be good if you create a PR... some of these things should be useful.
Thanks!

mmcauliffe:
@jtrmal PR created here: #4669

PR #4669, opened 2021-12-06
Update CMake build recipe allow dependencies from Conda

mmcauliffe:
Updated the CMake recipe to incorporate a cleaned up version of what I mentioned in #4661.  Changes only get triggered if `-DCONDA_ROOT` is specified, otherwise if should run the original code.

- Added support for building fgmm binaries, as those were previously excluded
- Added a check for building online* folders only on Mac and Linux
- Added components to installed files, to prevent any conflict with installing multiple versions of OpenFst
- I removed the FindBLAS and FindLAPACK scripts and bumped the minimum version to use the latest scripts distributed with CMake instead (I don't know if the ones in this repo had any patches applied)


CMakeLists.txt
```diff
@@ -228,6 +263,7 @@ endif()
 # add all native executables
 add_subdirectory(src/bin)
 add_subdirectory(src/gmmbin)
+add_subdirectory(src/fgmmbin)
```
danpovey, line 266: We normally exclude fgmmbin because it's basically never used.  In regular Makefile it's in an additional target "ext".
Might be a little wasteful to build it...

danpovey (commented):
Thanks!
We should probably have someone familiar with CMake comment on this.
One concern I have is that it may cause problems for people with older CMake versions
on their systems.  I wonder how feasible it would be to merge this later, e.g. after
a few months-- don't know if this would cause you inconvenience?

h-vetinari:
@mmcauliffe 

I just saw this PR, would be awesome indeed if this could be contained here directly (saving us from having to carry large patches in conda-forge). 🥳 

There's gonna be some more necessary adaptations regarding LAPACK(E) bindings. Could you rebase this PR (then I'd raise a PR against your PR)? Alternatively, I can rebase it and open a new PR (or push into this one, if you want to make me a collaborator on your fork).

h-vetinari:
@danpovey @mmcauliffe 
How would you feel about having a `libkaldi` target that only has the [libs](https://github.com/kaldi-asr/kaldi/blob/master/CMakeLists.txt#L190-L226) but not the [binaries](https://github.com/kaldi-asr/kaldi/blob/master/CMakeLists.txt#L228-L250)?

h-vetinari:
Thanks for the invitation to collaborate! Since your changes here differ from your patch in conda-forge, I kept the changes in this PR first and foremost. To verify the rebase - the last commit before rebasing will be 9b372f0b2948cc1edfdecd75602755a3e376e026, you can do:
```
git diff 54b4a13f34e6c169e2677443aacec737dec93ec0 9b372f0b2948cc1edfdecd75602755a3e376e026 -b
```
(i.e. ignoring whitespace changes).

I then added one more commit (320b4466705d88f4afb1735bb13a9f073eeec13c) on top for the lapacke-handling. Still might need some slight windows-adaptations, but I thought I'd provide this for discussion already.

jtrmal:
Thanks folks :)
Let me know if you will need any assistance. Or if you are waiting for
something from us (from kaldi devs)
y.

kkm000:
@mmcauliffe is the ball in our court now?

kkm000:
@danpovey:

> One concern I have is that it may cause problems for people with older CMake versions on their systems.

I think we should take an easy stance on CMake version. For both Windows and Linux there are drop-in tar.gz (maybe .zip for Windows) with identical file structure, bin/, share/man, share/cmake-XX.XX and doc/. For Mac, I do not understand the structure of the tar.gz file, it has additional directories with capitalized names, but probably it's designed to be as easy as possible. CMake looks for its runtime files relative to its own binary path when run. Also, there is an .msi package for Windows and a .dmg for Mac. Also, for Linux, the binaries are static and do not depend on anything in the system, even libc.

**I think that we should simply require a version not older than.** For the Kaldi "easy" install, the required version can be downloaded into tools/ just as we do to all other stuff. For other scenarios, the users can upgrade. I often unpack a new version into either `/usr/local` or `~/.local` over the previous one, then take out the old ./share/cmake-NN.NN versioned directory. On Windows, I use the MSI installer, but it would be an equally easy setup if I did not have permissions to run it (it's rare on Windows, I guess).

@danpovey, @jtrmal, do you see anything cheesy with this approach?

h-vetinari:
> @mmcauliffe is the ball in our court now?

I think this is good to go, yes. It contains @mmcauliffe fixes for windows, and everything conda-related is hidden away behind respective if-switches, so shouldn't concern anyone else. I'm still working on fixing some issues on the conda-forge side, but this patch would already be a huge upgrade, and I'll happily open another PR if some small incremental fixes on top still turn out to be necessary.

h-vetinari:
PS. the one part that clearly affects other is the CMake lower bound and removing the custom FindBLAS / FindLAPACK functions, but you were already addressing that above

csukuangfj:
> @danpovey, @jtrmal, do you see anything cheesy with this approach?

By the way, you can use either `pip install cmake` or `conda install cmake`, which also does not require root permission.

kkm000:
@csukuangfj, thanks. They both have 3.22.x, the latest (3.23 was released yesterday).

CMakeLists.txt
```diff
@@ -158,18 +188,26 @@ if(CUDA_FOUND)
...
     include_directories(${CUB_INCLUDE_DIR})
 endif()
 
 add_definitions(-DKALDI_NO_PORTAUDIO=1)
 
-include(VersionHelper)
-get_version() # this will set KALDI_VERSION and KALDI_PATCH_NUMBER
-if(${KALDI_USE_PATCH_NUMBER})
-    set(KALDI_VERSION "${KALDI_VERSION}.${KALDI_PATCH_NUMBER}")
+if(OVERRIDE_KALDI_VERSION)
```
kkm000, line 201: Consider using just `KALDI_VERSION`? Is the additional variable really needed?

CMakeLists.txt
```diff
@@ -158,18 +188,26 @@ if(CUDA_FOUND)
     include_directories(${NvToolExt_INCLUDE_DIR})
     link_libraries(${NvToolExt_LIBRARIES})
 
-    get_third_party(cub)
-    set(CUB_ROOT_DIR "${CMAKE_CURRENT_BINARY_DIR}/cub")
+    if (NOT CONDA_ROOT)
+        get_third_party(cub)
+        set(CUB_ROOT_DIR "${CMAKE_CURRENT_BINARY_DIR}/cub")
+    endif()
```
kkm000, line 194: Are you not relying on a pretty recent CUDA here? I honestly do not remember the version that came with CUB bundled, but it was likely 10 or even 11.

CMakeLists.txt
```diff
@@ -125,7 +150,12 @@ if(MSVC)
     endif()
 endif()
 
-find_package(CUDA)
+if (CONDA_ROOT)
+    find_package(CUDAToolkit)
```
kkm000, line 154: I think this should be unconditional for a recent CMake. 3.18, IIRC, is the lowest version that got this. 3.19 even supports CUDA as a project language. And then you do not need to `find_package(CUDA)`. As is, the `if` is kinda misleading: what is put inside it isn't because Conda, it's because CMake is recent.

cmake/gen_cmake_skeleton.py
```diff
@@ -276,13 +290,35 @@ def get_test_exe_name(filename):
...
+            EXPORT kaldi-targets
+            LIBRARY
+                DESTINATION ${{CMAKE_INSTALL_LIBDIR}}
+                COMPONENT kaldi
+            RUNTIME
+                DESTINATION ${{CMAKE_INSTALL_BINDIR}}
+                COMPONENT kaldi
+        )
+    endif()
+else() # Original functionality
```
kkm000, line 313: I don't think you need `if(MSVC)` if you also omit the `DESTINATION` clauses, which are set to their defaults here anyway (and in the `else()` clause below, too. Both .exe and .dll files are installed into the RUNTIME destination on Windows, i.e. shared library targets are RUNTIME, not LIBRARY. Feel free to add `COMPONENT kaldi` to the `install` command below. Then you won't need even the `if(CONDA_ROOT)`.

I am not sure why do you need to define everything as a component. I was thinking of _maybe_ using components eventually for `lib` and `extlib`. Is it required for the Conda build?

kkm000 (changes requested):
@h-vetinari, thanks for taking over! I added a few comments. Also, I do not understand what restrictions on the build process are imposed by Conda. Could you please list the bullet points that we need to maintain not to break it? 

My hands are itchy to ditch make and switch to CMake as the build system. It's a big rework, and I should be careful not to break you.

CMakeLists.txt
csukuangfj: > 3.19 even supports CUDA as a project language.

It's CMake 3.8, I think. We are already using CUDA as a project language in k2.
 See https://github.com/k2-fsa/k2/blob/master/CMakeLists.txt#L46

I agree we should throw away `find_package(CUDAToolkit)` as well as `find_package(CUDA)`.

cmake/third_party/openfst.cmake
```diff
@@ -14,12 +14,74 @@ if(NOT openfst_POPULATED)
...
-    install(TARGETS fst
-            EXPORT kaldi-targets
-            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-            RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+    if(CONDA_ROOT)
+
+        if(MSVC)
+            install(TARGETS fstarcsort fstclosure fstcompile fstcompose fstconcat fstconnect fstconvert fstdeterminize
```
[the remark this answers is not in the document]
mmcauliffe: So this was something I added due to the differences between OpenFst between Windows and other systems.  Conda has OpenFst-1.8.2 for Linux and Mac, but none for Windows (and Kaldi still depends on 1.7.X, right?).  The ideal would be to have Kaldi depend on the latest OpenFst and have OpenFst available on conda for all platforms.  I talked to Kyle Gorman back in December about porting the build code over to Cmake, but I haven't had the bandwidth to tackle that.  

Might be easier to simply use an updated version of your windows fork on conda's windows build.  From looking over the past discussions of upgrading to 1.8 last year, I got the impression that it would be a larger task.

CMakeLists.txt
mmcauliffe: Yeah, this conditional can probably get dropped, I was pretty conservative initially with preserving the cmake recipe as it was when conda wasn't specified.

cmake/gen_cmake_skeleton.py
mmcauliffe: The components go back to the other comment with the version of OpenFst on conda forge for Mac and Windows.  I used components to install just the kaldi binaries and libraries on those so to not conflict with OpenFst-1.8.2, but on Windows, OpenFst have the `kaldi` component so that they get installed alongside Kaldi.

h-vetinari:
> @kkm000: @h-vetinari, thanks for taking over! I added a few comments. Also, I do not understand what restrictions on the build process are imposed by Conda. Could you please list the bullet points that we need to maintain not to break it?
> 
> My hands are itchy to ditch make and switch to CMake as the build system. It's a big rework, and I should be careful not to break you.


Thanks a lot for taking a look! I'm happy that @mmcauliffe is already busy responding to some of your points, because I bascially only took his patch from conda-forge, resolved the conflicts upon rebasing, and then added only https://github.com/kaldi-asr/kaldi/pull/4669/commits/320b4466705d88f4afb1735bb13a9f073eeec13c.

You don't need to worry too much about not breaking conda-forge. We are able to carry patches, and more importantly, happy to push them back upstream (like in this PR). If any given release (or rather commit, based on how the patch-version here is generated) breaks us, we just don't release it, and can keep working together with you to fix it.

I would welcome a move to CMake, even if you need to break some eggs to make that omelette.

That said, what we rely on:
* ability to point to our own compilers and $PREFIX
* ability to inject our CMake args and pick up our CUDA builds
* ability to set various BLAS implementations

CMakeLists.txt
h-vetinari: This was added by @mmcauliffe, but I think the issue was (IIRC) that `KALDI_VERSION` doesn't get populated if there's no `git` installed.

cmake/third_party/openfst.cmake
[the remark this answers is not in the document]
h-vetinari: I started having a go at this in https://github.com/conda-forge/openfst-feedstock/pull/27

CMakeLists.txt
mmcauliffe: Yeah I think @kkm000 's point was that we could just set `KALDI_VERSION` via CMake `-DKALDI_VERSION` instead of introducing another variable, which was just an oversight on my part as I was learning how cmake works.  I'll fix that up.

CMakeLists.txt
mmcauliffe: Yeah I agree that'd be better.  I did try at one point to use CMake's cuda language support, but it was throwing some errors elsewhere, so I went with the more hacky approach just to get it working.  Would it be alright to switch over to using CUDA as a project language after this gets merged in, just in case those errors I hit haven't been fixed already by the work that @h-vetinari has done.

CMakeLists.txt
```diff
@@ -228,6 +268,7 @@ endif()
 # add all native executables
 add_subdirectory(src/bin)
 add_subdirectory(src/gmmbin)
+add_subdirectory(src/fgmmbin)
```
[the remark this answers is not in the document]
mmcauliffe: Ah ok, I was a little confused, as the fgmm bins are referenced in a number of places:

* Speaker id, language id, etc scripts for `sre08`, `lre07`, `callhome_diarization`, `gp`
* Some scripts in `egs/wsj/s5/steps`: [egs/wsj/steps/train_ubm.sh](https://github.com/kaldi-asr/kaldi/blob/cbed4ff688a172a7f765493d24771c1bd57dcd20/egs/wsj/s5/steps/train_ubm.sh#L117) and [egs/wsj/s5/steps/tandem/train_ubm.sh](https://github.com/kaldi-asr/kaldi/blob/cbed4ff688a172a7f765493d24771c1bd57dcd20/egs/wsj/s5/steps/tandem/train_ubm.sh)
* In [ivector-extract's usage example](https://github.com/kaldi-asr/kaldi/blob/cbed4ff688a172a7f765493d24771c1bd57dcd20/src/ivectorbin/ivector-extract.cc#L221)

So is the prototypical ivector/ubm training script the [egs/wsj/s5/steps/online/nnet2/train_ivector_extractor.sh](https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/online/nnet2/train_ivector_extractor.sh) and the others are/should be deprecated?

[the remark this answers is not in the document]
kkm000: @mmcauliffe D-oh! No, it's apparently current, my bad. 

@danpovey, I was under an impression the fgmmbin stuff was obsolete, but it's used when initializing the i-vector extractor with the D-UBM converted to full, unconditionally, e. g.

https://github.com/kaldi-asr/kaldi/blob/dd107fd594ac58af962031c1689abfdc10f84452/egs/wsj/s5/steps/online/nnet2/train_ivector_extractor.sh#L127-L136

called from

https://github.com/kaldi-asr/kaldi/blob/dd107fd594ac58af962031c1689abfdc10f84452/egs/mini_librispeech/s5/local/nnet3/run_ivector_common.sh#L103-L113

CMakeLists.txt
kkm000: @mmcauliffe, yes, I also do not think CUDA as a language should go into this changeset, there are too many little changes around enabling it, which won't affect the Conda-enabling things, so it's clearly separate. 

And, @csukuangfj, you cannot throw away `find_package(CUDAToolkit)` even if CUDA language is enabled. You still need the libraries. `find_package(CUDA)` is obsoleted by `find_package(CUDAToolkit)` though.

CMakeLists.txt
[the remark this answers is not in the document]
mmcauliffe: Oh, the `gmm-global-to-fgmm` is in `gmmbin`, not `fgmmbin`, so I don't think `fgmmbin` is necessary for anything in nnet2 or nnet3, `fgmmbin` calls are just in the `train_ubm.sh` scripts (within `wsj/steps`)

cmake/gen_cmake_skeleton.py
kkm000: I think our porting currently goes up to 1.7.2 or 1.7.3, we've never moved up to 1.8.2. And Kaldi does not compile with 1.8. The libfst.so by itself is likely to conflict, too. Is it better to tie Kaldi on Conda to the same version of FST that we're normally building against?

We can certainly define `COMPONENT kaldi` in https://github.com/kkm000/openfst, although I would rather prefer something more semantically sounding, for example, `kaldi-rt`, because the component's semantic is "runtime-only dependencies of Kaldi." By the way, I _think_ that besides the main library, the only extension that Kaldi really depends on is far binaries. Possibly the rho and phi matching binaries (I do not remember the component name, and I'm away from a real computer now, too) for the relatively new grammar support, but hardly anything beyond that. Let me double check and respond later. tools/Makefile is not an anchor of trust, it builds more than necessary.

Do you remember off the cuff if CMake can declare a component by purely combining other components?

What looks kinda weird is that the whole of Kaldi is also installed as a single component `kaldi`. Is it necessary?

mmcauliffe: Right, the linking component to libfst I changed here: https://github.com/mmcauliffe/kaldi/blob/master/cmake/third_party/openfst.cmake#L70, so that it links to the versioned `libfst.16.so` (if I'm remembering the version correctly).

Yeah, I think `kaldi-rt` makes sense for OpenFst.  I'm not 100% on how components interact together, I think you'd have to specify each one for installation if you specify any, but the main functionality that's necessary is being able to exclude `kaldi-rt` binaries for installation on mac and linux.

CMakeLists.txt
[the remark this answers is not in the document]
kkm000: @mmcauliffe Aah! This is it! Let me check where else fgmmbin could be used, if anywhere at all.

cmake/gen_cmake_skeleton.py
```diff
@@ -278,11 +292,10 @@ def get_test_exe_name(filename):
         ret.append("""
 install(TARGETS {tgt}
     EXPORT kaldi-targets
-    ARCHIVE DESTINATION ${{CMAKE_INSTALL_LIBDIR}}
-    LIBRARY DESTINATION ${{CMAKE_INSTALL_LIBDIR}}
-    RUNTIME DESTINATION ${{CMAKE_INSTALL_BINDIR}}
+    ARCHIVE DESTINATION ${{CMAKE_INSTALL_LIBDIR}} COMPONENT kaldi
+    LIBRARY DESTINATION ${{CMAKE_INSTALL_LIBDIR}} COMPONENT kaldi
+    RUNTIME DESTINATION ${{CMAKE_INSTALL_BINDIR}} COMPONENT kaldi
```
kkm000, line 297: Ditto w.r.t. `DESTINATION` and `COMPONENT`. This reduces to either
```
install(TARGETS {tgt} EXPORT kaldi-targets)
```
or
```
install(TARGETS {tgt} EXPORT kaldi-targets COMPONENT kaldi)
```
depending on the requrement of the Conda CMake bild to provide explicit component name(s).

cmake/third_party/openfst.cmake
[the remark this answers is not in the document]
kkm000: I will be working on a support for FST 1.8+ in Kaldi, but no ETA ATM, sorry.

cmake/third_party/openfst.cmake
```diff
@@ -14,12 +14,74 @@ if(NOT openfst_POPULATED)
...
+        else()
+
+            install(TARGETS fst
+                    LIBRARY
+                        DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                        COMPONENT kaldi
+                        NAMELINK_SKIP # Link to so with version to avoid conflicts with OpenFst 1.8.1 on conda
+                    )
+
+        endif()
```
kkm000, line 73: I'll investigate which of these binaries are in fact required.

BTW, you can build OpenFST binaries against a static OpenFST library. The libfst in fact has a tiny footprint, as the most of its magic is based on C++ templates and is thus header-only. I usually do exactly that. This will divorce the FST library used to build Kaldi from the one baked into the binaries. Building Kaldi fully static results in ~15GB of binaries, so probably not the best approach for packaging, but its required fst\* and far\* binaries subset isn't very large, and isn't reduced much by linking against a shared SO library, as it's really very small. You need to build the static library either using PIC, which is a much smaller performance hit on x64 than it used to be on x86, or use the large/huge memory model, so that all relative jump instructions are 64 bit in side. Doing neither may result in the linker complaining of being unable to fit the jmp offset into the space left in the binary by the compiler. Doing neither will likely result in linker computing a relocation overflowing the one generated by the compiler for the small memory model, of type `R_X86_64_32`, with an error message like "relocation truncated to fit: R_X86_64_32". The large memory model forces the compiler to generate relocations of type R_X86_64_64, not overflowable in a 64-bit address space, resulting in somewhat larger code; PIC will add an indirection for calls, but it is, again, handled much better by the 64-bit mode of CPU than it used to be in the 32-bit mode. Both approaches have a comparable speed, huge relocs beating PIC but only marginally, due to their negative impact on the L1C cache.

cmake/third_party/openfst.cmake
```diff
@@ -14,12 +14,74 @@ if(NOT openfst_POPULATED)
...
+    else() # Original functionality
+
+        install(DIRECTORY ${openfst_SOURCE_DIR}/src/include/ DESTINATION include/
+                FILES_MATCHING PATTERN "*.h")
+
+        install(TARGETS fst
+                EXPORT kaldi-targets
+                ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
kkm000, line 84: Ah, the dreaded `DESTINATION` again... :)

kkm000 (changes requested):
Thank you very much for your work guys. We'd certainly prefer integrating the support that you require from the build system. This has been the philosophy of Kaldi: we should support all build scenarios that expose Kaldi to a wider variety of people by making it as easy to start with as simple as possible. Installing Kaldi with a single pip command is certainly a boon.

Having dedicated reviewers from your side would be very helpful, so that we can tag you when there are big changes in our CMake scripts, so you can timely spot potential incompatibilities in advance. @mmcauliffe, @h-vetinari, do you want to volunteer for this role?

I'm marking this review as a request for changes, but not all of my comments are really change requests; some are just for a discussion.

Also, please open PRs against the default `winport` branch in https://github.com/kkm000/openfst/tree/winport for stuff that will reduce workarounds in the Kaldi repo.

h-vetinari:
> Thank you very much for your work guys. We'd certainly prefer integrating the support that you require from the build system. This has been the philosophy of Kaldi: we should support all build scenarios that expose Kaldi to a wider variety of people by making it as easy to start with as simple as possible. Installing Kaldi with a single pip command is certainly a boon.

That's a great philosophy to have! :)

> Having dedicated reviewers from your side would be very helpful, so that we can tag you when there are big changes in our CMake scripts, so you can timely spot potential incompatibilities in advance. @mmcauliffe, @h-vetinari, do you want to volunteer for this role?

I should be able to help test a given PR against the conda-forge infrastructure, and provide _some_ review through that.

h-vetinari:
PS. I had forgotten that I had had some more additions (https://github.com/kaldi-asr/kaldi/pull/4669/commits/3df87344561046547857fa51ad3a3636b1c54d0d) on the conda-forge side that I hadn't played back into this PR.

kkm000:
@h-vetinari 

> PS. I had forgotten that I had had some more additions (https://github.com/kaldi-asr/kaldi/commit/3df87344561046547857fa51ad3a3636b1c54d0d) on the conda-forge side that I hadn't played back into this PR.

Did you already transplant this change? I'm kinda losing track of updates to this PR.

h-vetinari:
> Did you already transplant this change? I'm kinda losing track of updates to this PR.

Yes, that's what the commit reference was for. 🙃 
Sorry if that was confusing.

kkm000:
@h-vetinari Ah, sure. I see it right above your previous comment. When _I_ click on it, it's not obvious where it is in relation to the PR branch. Normally GH shows a list of heads that the commit is reachable from (and this is likely how you see it), but I rather see a yellow box instead of that green one, with the text only, "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." (Funny, always say branch _in_ a repository, not like a branch _on_ a tree. I mean, on a real wooden tree, the only kind I'm aware of that grows with its roots down, crown up... Branches literally growing _on_ a repository are a bit too much for me. _Spring was in full swing, and branches were blooming with commits... (Shudders.)_  Maybe my imagination is just too vivid...)

So, what do you think we do next? Commit as is and figure out why `COMPONENT` later? The `COMPONENT` chief use is for selective build and/or packaging, e.g. one component for source, another for static libs and include files, a third one for runtime SO libs, etc. E.g, we can make binaries a component instead of toggling their build it with a CMake variable.

I think I'm going to seriously hack the CMake files along with those in our FST port repo, so it should not be hard to fix stuff later; CMake is kinda WIP anyway. But I still don't understand if the component somehow essential to your build process for the Py repo. Is it? Lumping everything into one makes little sense to me.

h-vetinari:
> When _I_ click on it, it's not obvious where it is in relation to the PR branch.

Sure, in the commit view you don't see that information (for branches in forks, as it the case for most PRs), but it's also there in the [commit](https://github.com/kaldi-asr/kaldi/pull/4669/commits) tab of the PR (I guess I could have used a better link to the same thing such as [this](https://github.com/kaldi-asr/kaldi/pull/4669/commits/3df87344561046547857fa51ad3a3636b1c54d0d) one)

> So, what do you think we do next? Commit as is and figure out why `COMPONENT` later?

Considering that CMake is WIP as you say, I think it would be beneficial to get this in and then build on top of it. I haven't understood the openfst integration yet, but in an ideal world, we could actually replace that with a separate build of the dynamic lib in conda-forge, see here. I was working on this in https://github.com/conda-forge/openfst-feedstock/pull/27 but ran into some errors - they look pretty trivial (wrong library names), but I haven't had time to investigate yet. If you have thoughts on that, your help would be highly appreciated.

kkm000:
@h-vetinari, we have a working CMake build on the winport branch. At least, it _should_ be working, and is not Windows-specific at all. The branch is somewhat behind current, though. Will fix. No ETA, but I'll bump it up.

And, finally, do you need the COMPONENT declared? Is it _really_ required? It will affect the CMake changes in the OpenFST port repo.

h-vetinari:
> And, finally, do you need the COMPONENT declared? Is it _really_ required? It will affect the CMake changes in the OpenFST port repo.

I was building predominantly on @mmcauliffe's patches. My impression was most of the changes were restricted to when building in conda, but where that's not the case, we should ask @mmcauliffe to comment.

mmcauliffe:
If we get rid of it, then when building for conda forge on linux and mac, we would have to make sure that the FST binaries aren't included in the distribution, otherwise they';ll conflict with the openfst-feedstock.

kkm000:
Got it. Now I grok the purpose of it, thanks!

kkm000 (approved):

PR #4672, opened 2021-12-17
添加posteriors matrix计算及decodable方法

ryanxiauic:


danpovey:
Not sure if you intended to make this PR?
If so it needs some cleanup.
What is the intended use case?

PR #4674, opened 2021-12-20
build 'portaudio' by default by 'tools/Makefile'

vesis84:
Hi Kyrill and Yenda,

- portaudio it is a dependency of 'src' build (not sure how much used though),

- till now 'src/Makefile' was touching back to 'tools' to install portaudio,
   - this was not ideal for the case of kaldi build without Internet connection as it breaks the build process...
   
(i usually build 'tools' on less powerful server that has Internet connection, 
and 'src' is built on manycore server that has no Internet connection)

This PR is fixing this issue.

Cheers,
K.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -26,14 +26,14 @@ ifeq ("$(shell expr $(OPENFST_VER_NUM) \< 10607)","1")
 endif
 
 .PHONY: all clean distclean
-all: check_required_programs cub openfst sctk_made sph2pipe
+all: check_required_programs cub openfst sctk_made sph2pipe portaudio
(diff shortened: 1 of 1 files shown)


danpovey:
Portaudio should only be needed if you build the target "ext" in src/Makefile.
This is not done by default.  Perhaps you need one of those "ext" tools for some reason and you added it to
your targets?

vesis84:
Aha, i see, the issue was just in my out-of-date version. I tried to compile current 'master', and it is okay.
(the dirs `online onlinebin sgmm2 sgmm2bin` are not compliled, and portaudio is dependency of `online`)
K.

PR #4675, opened 2021-12-21
[scripts] slurm.pl: Catch SIGTERM and SIGINT signals

zhuangweiji:


jtrmal:
that's great! I was missing this.

PR #4677, opened 2021-12-21
Add scripts support for channel in segments file

hhadian:
The channel column is supported by `extract-segments` binary but not in the scripts. This PR adds support in the scripts.


danpovey:
I don't think I want to accept this change because it is against the design of the data dir and would open up a can of worms..  The concept is that a "recording" represents a particular channel.  If recordings could represent multiple channels, lots of things would have to be changed.  So if your actual wav file or whatever has 2 channels, you are supposed to create two "recording-ids" that represent this; you can put sox commands in wav.scp that extract a particular channel from the wav file.

jtrmal:
I think this should be closed. Hossein/@hhadian thank for the effort anyway!

PR #4681, opened 2021-12-23
Expose API to wait for separate streams

nshmyrev:
If data from streams comes in irregular intervals it becomes important to understand how much data per-stream is processed. This API allows decoder to wait for particular stream results instead of waiting for the whole pack.

### DIFF
--- a/src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
+++ b/src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
@@ -64,6 +64,7 @@ void CudaOnlinePipelineDynamicBatcher::Push(
     backlog_.push_back(
         {corr_id, is_first_chunk, is_last_chunk, std::move(wave_samples)});
   }
+  n_chunks_per_corr_[corr_id] = n_chunks_per_corr_[corr_id] + 1;
   n_chunks_not_done_.fetch_add(1, std::memory_order_release);
 }
 
@@ -146,7 +147,9 @@ void CudaOnlinePipelineDynamicBatcher::BatcherThreadLoop() {
             curr_batch_->is_last_chunk);
         n_chunks_not_done_.fetch_sub(curr_batch_->Size(),
                                      std::memory_order_release);
-
+        for (auto const& corr_id: curr_batch_->corr_ids) {
+          n_chunks_per_corr_[corr_id]--;
+        }
         curr_batch_->Clear();
       }
 
@@ -165,6 +168,11 @@ void CudaOnlinePipelineDynamicBatcher::WaitForCompletion() {
   }
   // Waiting for pipeline to complete
   cuda_pipeline_.WaitForLatticeCallbacks();
+}
+
+int CudaOnlinePipelineDynamicBatcher::GetPendingChunks(CorrelationID corr_id) {
+  std::lock_guard<std::mutex> lk(next_batch_and_backlog_m_);
+  return n_chunks_per_corr_[corr_id];
 }
 
 }  // namespace cuda_decoder
--- a/src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
+++ b/src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
@@ -145,10 +145,19 @@ void CudaOnlinePipelineDynamicBatcher::BatcherThreadLoop() {
             curr_batch_->corr_ids, curr_batch_->h_all_waveform,
             curr_batch_->n_samples_valid, curr_batch_->is_first_chunk,
             curr_batch_->is_last_chunk);
-        n_chunks_not_done_.fetch_sub(curr_batch_->Size(),
-                                     std::memory_order_release);
-        for (auto const& corr_id: curr_batch_->corr_ids) {
-          n_chunks_per_corr_[corr_id]--;
+
+        {
+          // Update counts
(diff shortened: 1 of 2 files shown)


danpovey:
How well tested is this?  I'd be more comfortable merging after you've used it for a while, if the code is new.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.h
```diff
@@ -126,6 +127,7 @@ class CudaOnlinePipelineDynamicBatcher {
   std::vector<const std::string *> partial_hypotheses_;
   std::vector<bool> end_points_;
   std::atomic<std::uint32_t> n_chunks_not_done_;
+  std::map<CorrelationID, int> n_chunks_per_corr_;
```
csukuangfj, line 130: In C++ 11, `std::unordered_map` is preferred over `std::map` in most cases.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
```diff
@@ -167,5 +179,10 @@ void CudaOnlinePipelineDynamicBatcher::WaitForCompletion() {
   cuda_pipeline_.WaitForLatticeCallbacks();
 }
 
+int CudaOnlinePipelineDynamicBatcher::GetPendingChunks(CorrelationID corr_id) {
+  std::lock_guard<std::mutex> lk(next_batch_and_backlog_m_);
+  return n_chunks_per_corr_[corr_id];
```
csukuangfj, line 184: `n_chunks_per_corr_.at(corr_id);` is better  as it checks whether the given key exists or not.

This method is a getter I think, so you should check if the key exists or not.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
```diff
@@ -144,9 +145,20 @@ void CudaOnlinePipelineDynamicBatcher::BatcherThreadLoop() {
...
-                                     std::memory_order_release);
 
+        {
+          // Update counts
+          std::lock_guard<std::mutex> lk(next_batch_and_backlog_m_);
+          n_chunks_not_done_.fetch_sub(curr_batch_->Size(),
+                                       std::memory_order_release);
+          for (int i = 0; i < curr_batch_->corr_ids.size(); i++) {
+            CorrelationID corr_id = curr_batch_->corr_ids[i];
+            n_chunks_per_corr_[corr_id]--;
```
csukuangfj, line 156:
```suggestion
            --n_chunks_per_corr_[corr_id];
```

pre-decrement is preferred over post-decrement.
Also, for the counter `++i` in the above `for` loop.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
```diff
@@ -144,9 +145,20 @@ void CudaOnlinePipelineDynamicBatcher::BatcherThreadLoop() {
...
             curr_batch_->is_last_chunk);
-        n_chunks_not_done_.fetch_sub(curr_batch_->Size(),
-                                     std::memory_order_release);
 
+        {
+          // Update counts
+          std::lock_guard<std::mutex> lk(next_batch_and_backlog_m_);
+          n_chunks_not_done_.fetch_sub(curr_batch_->Size(),
+                                       std::memory_order_release);
+          for (int i = 0; i < curr_batch_->corr_ids.size(); i++) {
```
csukuangfj, line 154: You will receive compiler warnings complaining about comparing signed integer `i` with
unsigned type `corr_ids.size()`.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.h
nshmyrev: Fixed

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
nshmyrev: Done. Actually there is a catch here. If corr_id is finalized, we still need to return 0. I've added comment about it.

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
nshmyrev: Done

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
nshmyrev: Done

nshmyrev:
> How well tested is this? 

We run it for couple days, seems fine. We will test more.

@csukuangfj Thank you for review

src/cudadecoder/cuda-online-pipeline-dynamic-batcher.cc
```diff
@@ -167,5 +181,12 @@ void CudaOnlinePipelineDynamicBatcher::WaitForCompletion() {
   cuda_pipeline_.WaitForLatticeCallbacks();
 }
 
+int CudaOnlinePipelineDynamicBatcher::GetPendingChunks(CorrelationID corr_id) {
```
danpovey, line 184: IMO this should be named GetNumPendingChunks(), since it returns the number, not the actual chunks.

nshmyrev: Done

PR #4686, opened 2022-01-01
Add support for Continual Learning using chain/LF-MMI models

hhadian:
This adds support for Continual Learning (a.k.a domain expansion) to Kaldi's LF-MMI (i.e. chain) models. 
The changes to the core chain codes are minimal (and obviously back compatible).

Please see the paper for details: https://arxiv.org/abs/2110.07055
@danpovey


egs/cl_english/v1/run.sh
```diff
@@ -0,0 +1,88 @@
...
+  --exp exp_cvi_train_10hr --train-set cvi_train_10hr \
+  --src-mdl exp_fsh_train_10hr/chain/tdnn1a_noiv_small_sp/final.mdl
+
+# Do LWF continual learing
+local/chain/run_lwf_clean_1a.sh \
+  --src-tree-dir exp_fsh_train_10hr/chain/tree_sp_mono \
+  --dir exp_cvi_train_10hr/chain/tdnn_lwf1a_fsh2cvi_sp \
+  --exp exp_cvi_train_10hr --train-set cvi_train_10hr \
+  --src-mdl exp_fsh_train_10hr/chain/tdnn1a_noiv_small_sp/final.mdl \
+  --lwf-scale 0.8
```
danpovey, line 50: Do you really recommend LWF, from the paper I think you recommend LWF?
My feeling is it might be better to comment out the regular LWF stuff and only keep the recommended part of
the recipe as the main path.

danpovey (commented):
Some comments...

src/chainbin/nnet3-chain-add-post-to-egs.cc
```diff
@@ -0,0 +1,206 @@
...
+    using namespace kaldi::nnet3;
+    using namespace kaldi::chain;
+    typedef kaldi::int32 int32;
+    typedef kaldi::int64 int64;
+
+    const char *usage =
+        "This program reads the input nnet3 egs, computes nnet outputs for them and .\n"
+        "appends the outputs as NnetIos to the egs. For Continual Learning. Search for LWF.\n"
+        "\n"
+        "Usage:  nnet3-chain-copy-egs [options] <nnet-chain- model> <denominator-fst> <egs-rspecifier> <egs-wspecifier1> \n"
```
csukuangfj, line 90:
```suggestion
        "Usage:  nnet3-chain-add-post-to-egs [options] <nnet-chain-model> <denominator-fst> <egs-rspecifier> <egs-wspecifier1> \n"
```

egs/cl_english/v1/run.sh
hhadian: Thanks for the review!
I'll fix the logs.
In all the experiments that we tried, DenLWF worked better than regular LWF but I kept it for reference. I can make it default here.

hhadian:
All comments addressed. I kept one of the logs as it is a useful check to make sure the source model has softmax and it is within a WithProb(0.01) block so it won't be verbose.

hhadian:
I can try to prepare something when I have some free time. But why mini-librispeech? Just for easy reproduciblity?

jtrmal:
yeah
but anything goes, really... whatever will work for you
y.

PR #4689, opened 2022-01-04
Feature dim might differ from lda dim in CUDA ivector extractor

nshmyrev:
Currently code assumes both dimensions are the same, it is not the case for some models.


jtrmal:
lgtm. @danpovey?

PR #4692, opened 2022-01-25
Extending 'lattice-compose.cc' to compose with ark of fsts,

vesis84:
- This is a follow-up of: https://github.com/kaldi-asr/kaldi/pull/4571

- Refactoring 'lattice-compose.cc' to support composition with ark
  of fsts, so that it is done as Dan suggested before:

```
  I am thinking this can be done with a string arg called e.g.
  "--compose-with-fst", defaulting to "auto" which is the old behavior,
  meaning: rspecifier=lats, rxfilename=FST; and true/True or false/False
  is FST or lattice respectively.
```

- I added there possibility of rho-composition, which is useful for
  biasing lattices with word-sequences. Thanks to rho-composition,
  the biasing graph does not need to contain all words from lexicon.

- Would you be interested in an example how to use this?
  (i.e. create graphs from text file with python script
   using openfst as library, but that would need to change
   build of openfst to enable python extensions)

- Also which 'egs' recipe would be convenient to use it with?

### DIFF
--- a/src/latbin/lattice-compose.cc
+++ b/src/latbin/lattice-compose.cc
@@ -1,6 +1,7 @@
 // latbin/lattice-compose.cc
 
 // Copyright 2009-2011  Microsoft Corporation;  Saarland University
+//           2022  Brno University of Technology
 
 // See ../../COPYING for clarification regarding multiple authors
 //
@@ -16,7 +17,6 @@
 // MERCHANTABLITY OR NON-INFRINGEMENT.
 // See the Apache 2 License for the specific language governing permissions and
 // limitations under the License.
-
 
 #include "base/kaldi-common.h"
 #include "util/common-utils.h"
@@ -39,22 +39,34 @@ int main(int argc, char *argv[]) {
         "or lattices with FSTs (rspecifiers are assumed to be lattices, and\n"
         "rxfilenames are assumed to be FSTs, which have their weights interpreted\n"
         "as \"graph weights\" when converted into the Lattice format.\n"
+        "Or, rspecifier can be ark of biasing FSTs, see --compose-with-fst=true.\n"
         "\n"
         "Usage: lattice-compose [options] lattice-rspecifier1 "
         "(lattice-rspecifier2|fst-rxfilename2) lattice-wspecifier\n"
         " e.g.: lattice-compose ark:1.lats ark:2.lats ark:composed.lats\n"
-        " or: lattice-compose ark:1.lats G.fst ark:composed.lats\n";
+        " or: lattice-compose ark:1.lats G.fst ark:composed.lats\n"
+        " or: lattice-compose --compose-with-fst=true ark:1.lats\n"
+        "       ark:biasing.fsts ark:composed.lats\n";
 
     ParseOptions po(usage);
 
     bool write_compact = true;
     int32 num_states_cache = 50000;
     int32 phi_label = fst::kNoLabel; // == -1
+    int32 rho_label = fst::kNoLabel; // == -1
+    std::string compose_with_fst = "auto";
+
     po.Register("write-compact", &write_compact, "If true, write in normal (compact) form.");
     po.Register("phi-label", &phi_label, "If >0, the label on backoff arcs of the LM");
+    po.Register("rho-label", &rho_label,
+                "If >0, the label to forward fst1 paths not present biasing graph fst2. "
+                "(rho is input and output symbol on special arc in biasing graph)");
     po.Register("num-states-cache", &num_states_cache,
                 "Number of states we cache when mapping LM FST to lattice type. "
                 "More -> more memory but faster.");
+    po.Register("compose-with-fst", &compose_with_fst,
+                "(true|false|auto) For auto arg2 is: rspecifier=lats, rxfilename=fst "
+                "(old behavior), for true/false rspecifier is fst/lattice.");
     po.Read(argc, argv);
 
     if (po.NumArgs() != 3) {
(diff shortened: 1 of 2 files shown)


src/latbin/lattice-compose.cc
```diff
@@ -121,7 +164,23 @@ int main(int argc, char *argv[]) {
         }
       }
       delete fst2;
-    } else {
+    }
```
danpovey, line 167: Thanks!

Would you mind formatting the else in a more standard way that won't upset auto-formatters?
E.g.
```
} else {
  /* comment here */
}
```
.. I see why you did what you did, but let's be consistent with other code.
Are you confident that this won't break existing code?

src/latbin/lattice-compose.cc
```diff
@@ -39,22 +39,34 @@ int main(int argc, char *argv[]) {
         "or lattices with FSTs (rspecifiers are assumed to be lattices, and\n"
         "rxfilenames are assumed to be FSTs, which have their weights interpreted\n"
         "as \"graph weights\" when converted into the Lattice format.\n"
+        "Or, rspecifier can be ark of biasing FSTs, see --compose-with-fst=true.\n"
```
danpovey, line 42: This whole message is not very clear, especially with the modification.  Can write:

 Depending on the command-line arguments, either composes lattices with lattices,
 or lattices with a single FST or multiple FSTs (whose weights are interpreted as "graph weights").

Then change the lines below to:
```
Usage: lattice-compose [options] <lattice-rspecifier1> "
        "<lattice-rspecifier2|fst-rxfilename2|fst-rspecifier2> <lattice-wspecifier>\n"
 If the 2nd arg is an rspecifier, it is interpreted by default as a table of lattices,
 or as a table of FSTs if you specify --compose-with-fst=true.
```

src/latbin/lattice-compose.cc
```diff
@@ -63,14 +75,30 @@ int main(int argc, char *argv[]) {
...
     KALDI_ASSERT(phi_label > 0 || phi_label == fst::kNoLabel); // e.g. 0 not allowed.
+    KALDI_ASSERT(rho_label > 0 || rho_label == fst::kNoLabel); // e.g. 0 not allowed.
+    if (phi_label > 0 && rho_label > 0) {
+      KALDI_ERR << "You cannot set both 'phi_label' and 'rho_label' at the same time.";
+    }
+
+    { // convert 'compose_with_fst' to lowercase to support: true, True, TRUE
+      std::string tmp_lc(compose_with_fst);
+      std::transform(compose_with_fst.begin(), compose_with_fst.end(),
+                     tmp_lc.begin(), ::tolower);  // lc
```
csukuangfj, line 86:
```suggestion
                     compose_with_fst.begin(), std::tolower);  // lc
```
It can be an **in-place** operation.

src/latbin/lattice-compose.cc
```diff
@@ -121,7 +164,23 @@ int main(int argc, char *argv[]) {
...
+    else if (arg2_is_rxfilename && compose_with_fst == "false") {
+      // Would it make sense to do this? Not implementing...
+      KALDI_ERR << "Unimplemented...";
+    }
+
+    /**
+     * arg2 is rspecifier that contains a table of lattices
+     * - composing arg1 lattices with arg2 lattices
+     */
+    else if (not arg2_is_rxfilename &&
```
csukuangfj, line 181:
```suggestion
    else if (!arg2_is_rxfilename &&
```

csukuangfj (changes requested):

jtrmal:
was there a message why the 4-param std::transform cannot be used? I wonder
if there should be push_back_iterator(tmp_lc.begin())?
y.

vesis84:
it seems like the symbol `std::tolower` is overloaded, there is unary variant we want:
`int tolower( int ch )` in <cctype>
the other is two-arg:
`charT tolower( charT ch, const locale& loc )` from <locale>
so, the `std::transform` does not know how to resolve this...

vesis84:
the messages are:
```
/usr/local/include/c++/9.4.0/bits/stl_algo.h:4332:5: note:   template argument deduction/substitution failed:
lattice-compose.cc:84:71: note:   couldn’t deduce template parameter ‘_UnaryOperation’
   84 |       std::transform(str.begin(), str.end(), str.begin(), std::tolower);  // lc
```
and
```
/usr/local/include/c++/9.4.0/bits/stl_algo.h:4369:5: note:   template argument deduction/substitution failed:
lattice-compose.cc:84:71: note:   candidate expects 5 arguments, 4 provided
   84 |       std::transform(str.begin(), str.end(), str.begin(), std::tolower);  // lc
```

vesis84:
okay, i'll put there the lambda, i did not find any better solution...
(like wrapping `std::tolower` to resolve the overloading...)

vesis84:
interesting, this one works too:
`std::transform(str.begin(), str.end(), str.begin(), (int(*)(int))std::tolower);  // lc`

vesis84:
For the testing, i did test the variant: `ark:lattices - ark:fsts`, and it works well.

For the other combinations `ark:lattices - ark:lattices`, `ark:lattices - fst` and did not test it, 
but i did not change their code that was already there (just RhoCompose was added in the if-then-else).
And there are 3 composition types, Compose, PhiCompose and RhoCompose.

So, out of these 9 combinations, should i test everything?
Or, just some of them?

vesis84:
also, normally, in lattices there is normally no phi or rho symbol, but the PR code allows it...
(could there be possibly someone introducing phi/rho into a lattice ?)

vesis84:
I did some of the tests. I did the variants:
- lattice o fsts (rho composition, phi composition)
- lattice o 1fst (rho composition)
- lattices o lattices

In all these cases it produces lattice ark of about the same size as the input.
Plus for "lattice o fsts (rho composition)" it produced 1% WER improvement 
on my air-traffic communication task.

Otherwise, all the previous remarks to the code sholud be resolved now.
K.

src/latbin/lattice-compose.cc
```diff
@@ -34,27 +34,35 @@ int main(int argc, char *argv[]) {
...
     int32 num_states_cache = 50000;
     int32 phi_label = fst::kNoLabel; // == -1
+    int32 rho_label = fst::kNoLabel; // == -1
+    std::string compose_with_fst = "auto";
+
     po.Register("write-compact", &write_compact, "If true, write in normal (compact) form.");
     po.Register("phi-label", &phi_label, "If >0, the label on backoff arcs of the LM");
+    po.Register("rho-label", &rho_label,
+                "If >0, the label to forward fst1 paths not present in biasing graph fst2 "
+                "(rho is input and output symbol on special arc in biasing graph)");
```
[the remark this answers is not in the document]
vesis84: ok, done

danpovey:
OK, I'm going to merge this so that we have some forward movement.
It might be more ideal to have an example script, but we can work on that later.

PR #4694, opened 2022-02-03
Development container for VS code.

o-alexandre-felipe:
This will enable Windows users to run kaldi in a Linux environment by [developing inside a container](https://code.visualstudio.com/docs/remote/containers).

Basically you choose to open kaldi/docker/devcontainer in Visual studio code. It will build a linux image with the requirements for building kaldi.

It is also possible to specify other volumes in `docker-compose.yml` to share models, datasets, etc.


danpovey:
Thanks!
This doesn't look like it would break anything, so in the interest of moving forward I'll merge this as-is.

PR #4696, opened 2022-02-08
Fix SCTK compilation for Mac M1

pzelasko:
I doubt the `-march=native` flag in SCTK compilation has any observable effect for Kaldi users, but it breaks the compilation on M1 Macs where clang apparently doesn't have a registered tuning configuration for `native`.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -103,7 +103,7 @@ openfst_cleaned:
 
 #== SCTK =======================================================================
 
-SCTK_CXFLAGS = -w -march=native
+SCTK_CXFLAGS = -w
 SCTK_MKENV = CFLAGS="$(CFLAGS) $(SCTK_CXFLAGS)" \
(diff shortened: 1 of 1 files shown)


csukuangfj:
> I doubt the -march=native flag in SCTK compilation has any observable effect for Kaldi users, but it breaks the compilation on M1 Macs where clang apparently doesn't have a registered tuning configuration for native.

If in doubt, I would suggest removing that flag conditionally, i.e., remove it only when it is macOS.
```makefile
OS := $(shell uname -s)

SCTK_CXFLAGS = -w
ifneq ($(OS), Darwin)
SCTK_CXFLAGS += -march=native
endif
```

pzelasko:
Good idea, I added that.

pzelasko:
LGTM?

jtrmal:
🤷‍♂️  looks completely fine to me, merging...

PR #4698, opened 2022-02-12
renormalize line endings

h-vetinari:
In the context of iterating on the [patch](https://github.com/conda-forge/kaldi-feedstock/blob/master/recipe/cmake.patch) by @mmcauliffe in conda-forge, I noticed that several of the line endings in the repo are non-standard.

Git with autocrlf does a very good job of checking out files with the correct line endings (lf on unix, crlf on windows), but it seems in some places, windows line endings were forcefully checked in (or somehow remained from history).

The commit in this PR was generated by `git add --renormalize .` in order to renormalize the line endings-as-checked-in. No other changes.

You can verify this by using ⚙️ in the changes-tab to ignore whitespace changes.


danpovey:
All the CMake files had Windows line endings?   A shame!
It is a little sad that this change seems to lose the history of committers per line... this would
make things harder to interpret later on...

h-vetinari:
> It is a little sad that this change seems to lose the history of committers per line... this would make things harder to interpret later on...

I agree, but the github UI actually does a pretty good job these days to quickly skip through the blame.

grafik

And in this case, I'd say that it's a useful clean-up, and the deferring it will just accrue further effort (IMO).

danpovey:
OK, I suppose.  I'll merge.

h-vetinari:
Thanks!

PR #4708, opened 2022-03-09
[egs] Improve steps/align_si.sh commentary

HuangruiChu:
Since align_si.sh could be used to align monophone, it could deal with delta features.
Since align_si.sh could also be used to align triphone trained by "train_deltas.sh", it could deal with delta+ delta-delta features.

For LDA+MLLT features, I think "align_fmmlr.sh" could help deal with them.

### DIFF
--- a/egs/wsj/s5/steps/align_si.sh
+++ b/egs/wsj/s5/steps/align_si.sh
@@ -2,7 +2,8 @@
 # Copyright 2012  Johns Hopkins University (Author: Daniel Povey)
 # Apache 2.0
 
-# Computes training alignments using a model with delta or delta+delta-delta features.
+# Compute training alignments using a model with delta or
+# delta + delta-delta features.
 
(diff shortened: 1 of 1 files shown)


egs/wsj/s5/steps/align_si.sh
```diff
@@ -2,8 +2,7 @@
 # Copyright 2012  Johns Hopkins University (Author: Daniel Povey)
 # Apache 2.0
 
-# Computes training alignments using a model with delta or
-# LDA+MLLT features.
+# Computes training alignments using a model with delta or delta+delta-delta features.
```
kkm000, line 5:
```suggestion
# Compute training alignments using a model with delta or
# delta + delta-delta features.
```

kkm000 (approved):
I'll just commit this with the change. Hope you don't mind.

kkm000:
Thanks for spotting the incorrect script purpose statement!

PR #4710, opened 2022-03-09
Update with Kaldi master

abavisani:
PR to sync with kaldi master.


jtrmal:
This is probably some kind of mistake -- did you mean to open the PR to some other tree?

PR #4713, opened 2022-03-21
Fix vector casts by using explicit namespace

auroraanon38:
This code caused build-time errors due to lacking namespace imports or prefixes when I tried to build `kaldi`. This pull request fixes that.

The errors looked something like: `error: 'vector' does not name a type`

### DIFF
--- a/src/onlinebin/online-audio-server-decode-faster.cc
+++ b/src/onlinebin/online-audio-server-decode-faster.cc
@@ -308,9 +308,9 @@ int32 main(int argc, char *argv[]) {
         } else {
           std::vector<int32> word_ids;
           if (decoder.PartialTraceback(&out_fst)) {
-            GetLinearSymbolSequence(out_fst, static_cast<vector<int32> *>(0),
+            GetLinearSymbolSequence(out_fst, static_cast<std::vector<int32> *>(nullptr),
                                     &word_ids,
-                                    static_cast<LatticeArc::Weight*>(0));
+                                    static_cast<LatticeArc::Weight*>(nullptr));
             for (size_t i = 0; i < word_ids.size(); i++) {
               if (word_ids[i] != 0) {
                 WriteLine(client_socket,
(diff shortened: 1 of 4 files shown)


jtrmal:
LGTM

kkm000:
I'm wondering why not just `nullptr`? I thought it was supposed to implicitly convert to any pointer type.

https://en.cppreference.com/w/cpp/language/nullptr

> The keyword `nullptr` denotes the pointer literal. It is a [prvalue](https://en.cppreference.com/w/cpp/language/value_category) of type [std::nullptr_t](https://en.cppreference.com/w/cpp/types/nullptr_t). There exist [implicit conversions](https://en.cppreference.com/w/cpp/language/implicit_conversion) from `nullptr` to null pointer value of any pointer type and any pointer to member type.

jtrmal:
I think just because the person was just fixing the namespace lookup and
didn't want to bother about anything else, which I think is legit. Plus if
they wanted to fix all the issues, the patch would never got here... :)
y.

auroraanon38:
jtrmal ***@***.***> writes:

> I think just because the person was just fixing the namespace lookup and
> didn't want to bother about anything else, which I think is legit. Plus if
> they wanted to fix all the issues, the patch would never got here... :)
> y.
>

That's essentially it, yes.

I'm trying to get Kaldi back into a buildable state in Guix, so I just
went with the most obvious solution that didn't require diving into the
codebase.

kkm000:
This is how crap creeps in. I'm worrying nobody will get back to fix this. If you touch the codebase, it's an opportunity to do it right. All these casts convert 0 to a pointer type.

auroraanon38:
kkm000 ***@***.***> writes:

> @auroraanon38, I'm OK with an emergency fix, but maybe you could check if `nullptr`s work here? I'm 99.7% sure they would. :)

I'm not particularly used to C++, so I felt that minimizing the risk of
breaking things unexpectedly was desirable.

I'll give it a try.

auroraanon38:
kkm000 ***@***.***> writes:

> @auroraanon38, thank you, that will be really helpful!

I think what I just pushed should do what you meant. It also builds.

kkm000:
Thanks. Not exactly what I was thinking it could simplify to, but at the least build is fixed!

Issue #4712, opened 2022-03-20
OpenFst no longer produces a fst/types.h file since 1.8


auroraanon38:
According to the 1.8 news, [OpenFst no longer produces types.h shims](https://www.openfst.org/twiki/pub/FST/FstDownload/NEWS), which I have verified to be true.

This prevents building kaldi with a recent OpenFst version, namely 1.8.2, with errors like this:

    In file included from ../base/kaldi-error.h:34,
                     from ../base/kaldi-common.h:35,
                     from ../matrix/matrix-common.h:26,
                     from ../matrix/packed-matrix.h:25,
                     from ../matrix/tp-matrix.h:26,
                     from tp-matrix.cc:21:
    ../base/kaldi-types.h:44:10: fatal error: fst/types.h: No such file or directory
       44 | #include <fst/types.h>
          |          ^~~~~~~~~~~~~
    compilation terminated.

I have also verified that 1.7.9 does produce such a file.

auroraanon38:
I have begun some work on trying to bring the code to build with openfst 1.8, though some things need discussion with this project's main contributors & maintainers. Changes in function signatures and naming schemes happened.

https://github.com/auroraanon38/kaldi/tree/fix/update-kaldi-for-openfst-1.8

kkm000:
@auroraanon38, please be very careful with this. There was a long discussion and, IIRC, a decision to upgrade to an 1.7.x version breaking the public API in a minor update. Search issues/tickets. I can help you, but a few hours later.

What is the motivation for the upgrade? Will it give us a better functionality or a significant performance boost? I'm just asking, I do not know.

1.8 is quite a major change, and you mentioned that C++ is not your first language. FST is heavily templated API, and you _will_ need to hack the code that consumes it.

kkm000:
Here are must-read references related to the issue, possibly not in the best order. There are multiple issues that must be taken care of.

* https://github.com/kaldi-asr/kaldi/issues/4565#issuecomment-861250982 (and below)
* #4096
* https://github.com/kaldi-asr/kaldi/issues/4533#issuecomment-846226502

It's generally a good idea to search issues and PRs, open or closed, before committing to a chunk of work.

auroraanon38:
kkm000 ***@***.***> writes:

> @auroraanon38, please be very careful with this. There was a long discussion and, IIRC, a decision to upgrade to an 1.7.x version breaking the public API in a minor update. Search issues/tickets. I can help you, but a few hours later.
[...]

Issue #4393, opened 2020-12-16
Build is broken with openfst >=1.7.4


ivan3182:
in openfst >=1.7.4 pointer is replaced with reference in `fst::FstPrinter<Arc>::Print` argument

fst/script/print-impl.h:
```diff
1.7.3
-void Print(std::ostream *ostrm, const std::string &dest)
---
1.7.4
+void Print(std::ostream &ostrm, const std::string &dest)
```

**Build log:**
```
../util/kaldi-table-inl.h: In instantiation of ‘bool kaldi::SequentialTableReaderArchiveImpl<Holder>::Open(const string&) [with Holder = kaldi::BasicHolder<float>; std::string = std::__cxx11::basic_string<char>]’:
../util/kaldi-table-inl.h:490:16:   required from here
../util/kaldi-table-inl.h:501:20: warning: unused variable ‘rs’ [-Wunused-variable]
  501 |     RspecifierType rs = ClassifyRspecifier(rspecifier,
      |                    ^~
kaldi-lattice.cc: In function ‘bool kaldi::WriteCompactLattice(std::ostream&, bool, const CompactLattice&)’:
kaldi-lattice.cc:81:19: error: cannot convert ‘std::ostream*’ {aka ‘std::basic_ostream<char>*’} to ‘std::ostream&’ {aka ‘std::basic_ostream<char>&’}
   81 |     printer.Print(&os, "<unknown>");
      |                   ^~~
      |                   |
      |                   std::ostream* {aka std::basic_ostream<char>*}
In file included from ../fstext/fstext-utils.h:35,
                 from ../fstext/deterministic-fst-inl.h:25,
                 from ../fstext/deterministic-fst.h:333,
                 from ../fstext/context-fst.h:62,
                 from ../fstext/fstext-lib.h:23,
                 from ../lat/kaldi-lattice.h:24,
                 from kaldi-lattice.cc:22:
/usr/include/fst/script/print-impl.h:59:28: note:   initializing argument 1 of ‘void fst::FstPrinter<Arc>::Print(std::ostream&, const string&) [with Arc = fst::ArcTpl<fst::CompactLatticeWeightTpl<fst::LatticeWeightTpl<float>, int> >; std::ostream = std::basic_ostream<char>; std::string = std::__cxx11::basic_string<char>]’
   59 |   void Print(std::ostream &ostrm, const std::string &dest) {
      |              ~~~~~~~~~~~~~~^~~~~
[...]

Issue #4131, opened 2020-06-24
OpenFST FstPrinter.Print signature has changed between 1.7.2 and 1.7.7


kkm000:
From https://groups.google.com/forum/#!topic/kaldi-help/qRp-TquUPHA

```
/home/jupyter/rob/espnet/kaldi/tools/openfst-1.7.7/include/fst/script/print-impl.h:46:8: note: candidate: void fst::FstPrinter<Arc>::Print(std::ostream&, const string&) [with Arc = fst::ArcTpl<fst::CompactLatticeWeightTpl<fst::LatticeWeightTpl<float>, int> >; std::ostream = std::basic_ostream<char>; std::__cxx11::string = std::__cxx11::basic_string<char>]
   void Print(std::ostream &ostrm, const std::string &dest) {
        ^~~~~
/home/jupyter/rob/espnet/kaldi/tools/openfst-1.7.7/include/fst/script/print-impl.h:46:8: note:   no known conversion for argument 1 from ‘std::ostream* {aka std::basic_ostream<char>*}’ to ‘std::ostream& {aka std::basic_ostream<char>&}’
```

1.7.2 indeed had `void Print(std::ostream *ostrm, const string &dest)`:

https://github.com/kkm000/openfst/blob/0bca6e76d24647427356dc242b0adbf3b5f1a8d9/src/include/fst/script/print-impl.h#L46-L57

kkm000:
Blocked by kkm000/openfst#30

PR #4716, opened 2022-03-24
[WIP] Fix/update kaldi for openfst 1.8

auroraanon38:
WIP to bring Kaldi into a workable state with openfst 1.8 as per [this issue](https://github.com/kaldi-asr/kaldi/issues/4712)

Discussion and feedback required. As well as help with more complicated C++ modifications difficult with my current skills.

### DIFF
--- a/src/base/kaldi-types.h
+++ b/src/base/kaldi-types.h
@@ -39,24 +39,6 @@ typedef float   BaseFloat;
 // we find in the future lacks stdint.h
 #include <stdint.h>
 
-// for discussion on what to do if you need compile kaldi
-// without OpenFST, see the bottom of this this file
-#include <fst/types.h>
-
-namespace kaldi {
-  using ::int16;
-  using ::int32;
-  using ::int64;
-  using ::uint16;
-  using ::uint32;
-  using ::uint64;
-  typedef float   float32;
-  typedef double double64;
-}  // end namespace kaldi
-
-// In a theoretical case you decide compile Kaldi without the OpenFST
-// comment the previous namespace statement and uncomment the following
-/*
 namespace kaldi {
   typedef int8_t   int8;
   typedef int16_t  int16;
@@ -70,6 +52,5 @@ namespace kaldi {
   typedef float    float32;
   typedef double   double64;
 }  // end namespace kaldi
-*/
 
 #endif  // KALDI_BASE_KALDI_TYPES_H_
--- a/src/base/kaldi-types.h
+++ b/src/base/kaldi-types.h
@@ -39,18 +39,16 @@ typedef float   BaseFloat;
 // we find in the future lacks stdint.h
 #include <stdint.h>
 
-namespace kaldi {
-  typedef int8_t   int8;
-  typedef int16_t  int16;
-  typedef int32_t  int32;
-  typedef int64_t  int64;
+typedef int8_t   int8;
+typedef int16_t  int16;
+typedef int32_t  int32;
+typedef int64_t  int64;
 
-  typedef uint8_t  uint8;
-  typedef uint16_t uint16;
-  typedef uint32_t uint32;
-  typedef uint64_t uint64;
-  typedef float    float32;
-  typedef double   double64;
-}  // end namespace kaldi
+typedef uint8_t  uint8;
+typedef uint16_t uint16;
+typedef uint32_t uint32;
+typedef uint64_t uint64;
+typedef float    float32;
+typedef double   double64;
 
 #endif  // KALDI_BASE_KALDI_TYPES_H_
--- a/src/configure
+++ b/src/configure
@@ -284,7 +284,7 @@ Either your CUDA is too new or too old."
       GCC_VER=$($CXX -dumpversion)
       GCC_VER_NUM=$(echo $GCC_VER | sed 's/\./ /g' | xargs printf "%d%02d%02d")
       case $CUDA_VERSION in
-        # Disabling CUDA 7 and CUDA 8 because we now use C++14 to compile CUDA code
+        # Disabling CUDA 7 and CUDA 8 because we now use c++17 to compile CUDA code
         # It is still possible to use those cuda versions by switching back to C++11
         # in src/makefiles/cuda_64bit.mk and use CUB <= 1.8.0 (see tools/)
         # https://github.com/kaldi-asr/kaldi/pull/4242
--- a/src/fstext/fstext-utils-inl.h
+++ b/src/fstext/fstext-utils-inl.h
@@ -163,7 +163,7 @@ void RemoveSomeInputSymbols(const std::vector<I> &to_remove,
                             MutableFst<Arc> *fst) {
   KALDI_ASSERT_IS_INTEGER_TYPE(I);
   RemoveSomeInputSymbolsMapper<Arc, I> mapper(to_remove);
-  Map(fst, mapper);
+  ArcMap(fst, mapper);
 }
 
 template<class Arc, class I>
--- a/src/fstext/fstext-utils.h
+++ b/src/fstext/fstext-utils.h
@@ -113,7 +113,7 @@ void PushInLog(VectorFst<StdArc> *fst, uint32 ptype, float delta = kDelta) {
 template<class Arc>
 void MinimizeEncoded(VectorFst<Arc> *fst, float delta = kDelta) {
 
-  Map(fst, QuantizeMapper<Arc>(delta));
+  ArcMap(fst, QuantizeMapper<Arc>(delta));
   EncodeMapper<Arc> encoder(kEncodeLabels | kEncodeWeights, ENCODE);
   Encode(fst, &encoder);
   internal::AcceptorMinimize(fst);
--- a/src/fstext/kaldi-fst-io-inl.h
+++ b/src/fstext/kaldi-fst-io-inl.h
@@ -44,7 +44,7 @@ void WriteFstKaldi(std::ostream &os, bool binary,
     bool acceptor = false, write_one = false;
     FstPrinter<Arc> printer(t, t.InputSymbols(), t.OutputSymbols(),
                             NULL, acceptor, write_one, "\t");
-    printer.Print(&os, "<unknown>");
+    printer.Print(os, "<unknown>");
     if (os.fail())
       KALDI_ERR << "Stream failure detected writing FST to stream";
     // Write another newline as a terminating character.  The read routine will
--- a/src/fstext/kaldi-fst-io-inl.h
+++ b/src/fstext/kaldi-fst-io-inl.h
@@ -99,7 +99,7 @@ void ReadFstKaldi(std::istream &is, bool binary,
     fst->DeleteStates();
     string line;
     size_t nline = 0;
-    string separator = FLAGS_fst_field_separator + "\r\n";
+    string separator = FST_FLAGS_fst_field_separator + "\r\n";
     while (std::getline(is, line)) {
       nline++;
       vector<string> col;
--- a/src/fstext/lattice-utils-inl.h
+++ b/src/fstext/lattice-utils-inl.h
@@ -268,9 +268,9 @@ void ConvertFstToLattice(
     MutableFst<ArcTpl<LatticeWeightTpl<Real> > > *ofst) {
   int32 num_states_cache = 50000;
   fst::CacheOptions cache_opts(true, num_states_cache);
-  fst::MapFstOptions mapfst_opts(cache_opts);
+  fst::ArcMapFstOptions mapfst_opts(cache_opts);
   StdToLatticeMapper<Real> mapper;
-  MapFst<StdArc, ArcTpl<LatticeWeightTpl<Real> >,
+  ArcMapFst<StdArc, ArcTpl<LatticeWeightTpl<Real> >,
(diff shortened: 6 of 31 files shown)


auroraanon38:
Apparently the Travis template will also need updating, as it isn't aware of C++17

auroraanon38:
Currently encountering an issue with `operator+` for type conversion between basic string & basic string view.

```
./lat/arctic-weight.h:52:59: error: no match for ‘operator+’ (operand types are ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’} and ‘std::string_view’ {aka ‘std::basic_string_view<char>’})
   52 |     static const std::string type = std::string("arctic") +
      |                                          ~~~~~~~~~~~~~~~~~^
   53 |         FloatWeightTpl<T>::GetPrecisionString();
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~            
In file included from /gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/bits/stl_algobase.h:67,
                 from /gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/vector:60,
                 from ../lat/lattice-functions.h:27,
                 from kws-functions2.cc:21:
/gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/bits/stl_iterator.h:533:5: note: candidate: ‘template<class _Iterator> constexpr std::reverse_iterator<_Iterator> std::operator+(typename std::reverse_iterator<_Iterator>::difference_type, const std::reverse_iterator<_Iterator>&)’
  533 |     operator+(typename reverse_iterator<_Iterator>::difference_type __n,
      |     ^~~~~~~~
/gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/bits/stl_iterator.h:533:5: note:   template argument deduction/substitution failed:
In file included from ../kws/kaldi-kws.h:25,
                 from ../kws/kws-functions.h:27,
                 from kws-functions2.cc:22:
../lat/arctic-weight.h:52:59: note:   ‘std::string_view’ {aka ‘std::basic_string_view<char>’} is not derived from ‘const std::reverse_iterator<_Iterator>’
   52 |     static const std::string type = std::string("arctic") +
      |                                          ~~~~~~~~~~~~~~~~~^
   53 |         FloatWeightTpl<T>::GetPrecisionString();
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~            
In file included from /gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/bits/stl_algobase.h:67,
                 from /gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/vector:60,
                 from ../lat/lattice-functions.h:27,
                 from kws-functions2.cc:21:
/gnu/store/069aq2v993kpc41yabp5b6vm4wb9jkhg-gcc-10.3.0/include/c++/bits/stl_iterator.h:1578:5: note: candidate: ‘template<class _Iterator> constexpr std::move_iterator<_IteratorL> std::operator+(typename std::move_iterator<_IteratorL>::difference_type, const std::move_iterator<_IteratorL>&)’
 1578 |     operator+(typename move_iterator<_Iterator>::difference_type __n,
      |     ^~~~~~~~
```

Is it preferable for me to extend the operator to cover this specific case, or replace the calls with explicit conversions?

danpovey:
The tricky thing, and the reason we hvaen't done this earlier, is that we don't want to break existing builds with older OpenFST versions.  I assume OpenFST must export some kind of flag saying what the version is; we might be able to use ifdefs based on that....

danpovey:
.. and you can probably fix the operator+ problem by casting the second arg to std::string, i.e replace x.GetPrecisionString() with std::string(x.GetPrecisionString()).

danpovey:
Responding to: "Is it preferable for me to extend the operator to cover this specific case, or replace the calls with explicit conversions?", which is not appearing on the webpage for some reason..
Explicit conversions.  You shouldn't be extending the standard library for sure.

auroraanon38:
> The tricky thing, and the reason we hvaen't done this earlier, is that we don't want to break existing builds with older OpenFST versions. I assume OpenFST must export some kind of flag saying what the version is; we might be able to use ifdefs based on that....

Right, that's reasonable. I'll look to see if it has such a flag. The configure file has the version and might be putting it somewhere useful, but it might not be adding any such symbol inside the library itself.

> .. and you can probably fix the operator+ problem by casting the second arg to std::string, i.e replace x.GetPrecisionString() with std::string(x.GetPrecisionString()).

Noted, will do.

> [](https://github.com/kaldi-asr/kaldi/pull/4716#issuecomment-1077352876)Responding to: "Is it preferable for me to extend the operator to cover this specific case, or replace the calls with explicit conversions?", which is not appearing on the webpage for some reason..
> Explicit conversions. You shouldn't be extending the standard library for sure.

Yeah, I've gotten too used to the Lisp multimethod patterns.

> I have certain doubts about enforcing C++17

Is it possible to build C++ libraries with C++17 but use them in programs compiled with lower versions?

In any case you have a point that it could and *should* be turned into a conditional variable dependent on openfst use inside those Makefiles, rather than hardcoded as I did.

jtrmal:
Ad the C++17
Lots of the OpenFST stuff is templated and now written with the C++17
features, so I'm afraid you will need C++17 compliant compiler every time
you will need to compile something.
Unless there will be some compatibility layer or something....
y.

auroraanon38:
Noting that this PR also pertains to https://github.com/kaldi-asr/kaldi/issues/4131 and https://github.com/kaldi-asr/kaldi/issues/4393

auroraanon38:
> Ad the C++17 Lots of the OpenFST stuff is templated and now written with the C++17 features, so I'm afraid you will need C++17 compliant compiler every time you will need to compile something. Unless there will be some compatibility layer or something.... 
> y.

Alright.

danpovey:
I wonder whether, if it's not practical to make everything conditional, we could create a separate branch of Kaldi for use with the more recent OpenFST versions?

auroraanon38:
Daniel Povey ***@***.***> writes:

> I wonder whether, if it's not practical to make everything conditional, we could create a separate branch of Kaldi for use with the more recent OpenFST versions?

That would likely be the simplest and cleanest option.

kkm000:
I like many of these changes, not even related to OpenFST per se. We should certainly divorce Kaldi from OpenFST primitive types.

So far in this changeset, there is not so much stuff that would require complex shimming. 3 or 4 small functions will do. We support much larger divergence in CUDA. I do not think maintaining a forked branch is a viable option. Someone has to be responsible for replicating changes to it, and then our test pipelines are barely holding, and the amount of testing immediately doubles. We'll _require_ a lower version of OpenFST in the out-of-the-box, "standard" build, but _support_ 1.8+ and C++17 for custom builds. C++17, is fully supported.

@danpovey, @jtrmal, what do you think? There was another issue with changed command line arguments to one of the FST binaries, breaking scripts, somewhere between 1.7.2 and 1.7.7.

I do not really understand the benefits of using FST 1.8 as the default. Is it faster, or fixing any issues? If we're not using features added to higher versions, then the default for the Kaldi-as-toolkit can be 1.7.x, where x is the version right before the breaking command line change. Kaldi-as-library, which the users integrate, should not use command line tools, and is not affected.

jtrmal:
I'm ok with that

danpovey:
Sorry I missed this earlier,
@kkm000 I think your proposal is OK.  I suppose you are talking about making our codebase support both older and newer version of OpenFST.

jtrmal:
makes sense to me. Apologies if I sounded too negative. I dont even oppose
to switching to C++17, it just feels it would cut off quite a large portion
of the user base.
y..

kkm000:
Bad bot! Boo!

danpovey:
ping

georgthegreat:
@auroraanon38, is there some work still going on in this issue?
We have some patches that fix compatibility with newer versions of OpenFST.

I might upstream them though they will not bring perfect compatibility out of the box.

@danpovey, is there any kind of a todolist fixed somewhere which describes the whole migration process?

auroraanon38:
Yuriy Chernyshov ***@***.***> writes:

> @auroraanon38, is there some work still going on in this issue?
> We have some patches that fix compatibility with newer versions of OpenFST.
Nope, no work is currently ongoing, it is legitimately stale.

georgthegreat:
> Nope, no work is currently ongoing, it is legitimately stale.

Could you, please, elaborate a bit?
What is the rationale behind the stale?

Would you (or maybe someone else in the project) accept the PRs supporting newer versions of OpenFST (I assume this would mean dropping the support for any other version though).

jtrmal:
He, we would support but it would have to be done more carefully than just
dropping old code, I think
the reason is that the new OpenFST changes some interfaces and needs
C++17(?) standard
So resolving this is not trivial -- probably would need conditional
compilation and stuff
But if you just check how much code changes is in in C++ codes only, we/I
could then try to figure out the rest
y.

georgthegreat:
@jtrmal, I have just submitted PR #4829 with a chain of commits needed to support openfst version up to 1.8.2.

kkm000:
I apologize I fell out of the loop, but I want to remind that a command line of one tool has changed somewhere between minor v1.7 points, most probably in [1.7.2, 1.7.8] range. I forgot which binary was that.

I have not only nothing against bumping Kaldi to C++17, but all for it.  We haven't yet even embraced full C++11 support for system-independent language features, really: some stuff like aligned memory allocation and memory model in general, and PRNG have been standardized, so we can get rid of quite a few `#if WIN32` and such stuff. In fact, we wrote two very multithreaded and lock-contending Kaldi-based production servers; the first in C++14, xplat Linux/Windows; the second together with @jtrmal in C++17, without a _single_ system call in the main logic paths¹

I don't see much value in sticking with a 10-year-C++. Or even 5-year-C++, to that matter. In any case, Git tags and maintenance branches exist, and for a good measure.

¹ The three exceptions outside of the main course are (1) service startup/termination, of which there are [five standards](https://xkcd.com/927/): init, upstart, systemd and containers on Linux and Service Control Manager on Windows, (2) Windows-only built-in ultra-high throughput metric collector, with no Linux counterpart of similar performance, and (3) three different logging systems on Linux: syslog, systemd journal and in-container Docker logging, and yet the fourth on Windows.

danpovey:
OK, but is it possible to use C++17 with the older OpenFST version?
I think there is some desirability for keeping back-compatibility, and if it's not possible to support the newer OpenFST without a seamless upgrade process (e.g. enabled via #ifdefs on the OpenFST version number) then I'm not convinced the benefits outweigh the costs.

georgthegreat:
We used to compile OpenFST with c++17 (we currently compile it with c++20).

We are maintaining a huge monorepo with hundreds of opensource projects.
The only way to make opensource work at scale we were able to find so far is update everytihng to the most recent version and fix small incompatibilities by patches, then upstream patches.

So here I am, upstreaming our patches.

danpovey:
OK, and I appreciate the effort but... 
I think Kaldi is a bit of a special case.  At this point the project is mostly in "legacy maintenance" mode as the technology has
moved on a bit; and there are a lot of people who have incorporated Kaldi into their products and may want the latest fixes but not want a big hassle with updating OpenFST.

I'm wondering whether we could have a branch for this and do our best to keep it up to date with master?

rkjaran:
> OK, and I appreciate the effort but... I think Kaldi is a bit of a special case. At this point the project is mostly in "legacy maintenance" mode as the technology has moved on a bit; and there are a lot of people who have incorporated Kaldi into their products and may want the latest fixes but not want a big hassle with updating OpenFST.
> 
> I'm wondering whether we could have a branch for this and do our best to keep it up to date with master?

As someone who incorporates both Kaldi and OpenFST (>1.8) into products, the big hassle for us is having to deal with maintaining OpenFST related patches when pulling in new commits from Kaldi.

georgthegreat:
> and there are a lot of people who have incorporated Kaldi into their products and may want the latest fixes but not want a big hassle with updating OpenFST.

Sorry, but do you have any kind of statistics to proof this?
As far as I see from the conversation here, most people would like to get both kaldi AND the latest version of OpenFST
(I understand this might be just a survivorship bias as we are discussing OpenFST update).

> I'm wondering whether we could have a branch for this and do our best to keep it up to date with master?

Unfortunately, this will not work for us. As we want to get the latest fixes and kaldi does not publish any releases at all, we have to remain on the master branch, not some feature branch.

I think branch diversion will move us to a git-flow style of development (i. e. keep the latest release in master, while doing all the development in some branch). As to my taste, I find [trunk based development](https://trunkbaseddevelopment.com/) to be more clear and understandable way to develop.

auroraanon38:
Yuriy Chernyshov ***@***.***> writes:

> Could you, please, elaborate a bit?
Right, I meant that on my end there's no work ongoing.

Sorry for the ambiguity.

h-vetinari:
Not stale

h-vetinari:
Guess this has been superseded by #4829

kkm000:
Duplicate-of: #4829

@h-vetinari, thanks, it indeed is, marking it as such.

PR #4721, opened 2022-04-06
Remove cu-device include.

galv:
It was unused anyway.


galv:
Just a friendly ping on this. It is effectively a no-op change. It just allows those using kaldi as a library to avoid pulling in the cu-device dependency when using the cuda decoder.

jtrmal:
thanks, I will merge

PR #4730, opened 2022-04-17
Remove sgmm2 from gmm-est-fmllr-gpost dependencies

mmcauliffe:
Small fix from the CMake change, sgmm2 was still listed as a dependency of gmm-est-fmllr-gpost, despite not being used, so removing sgmm2 from being built caused linker errors.


kkm000:
LGTM, thanks!

PR #4731, opened 2022-04-19
Fix missing cblas and lapack external symbols for netlib

mmcauliffe:
One last small fix to allow compiling with Conda's BLAS infrastructure

### DIFF
--- a/cmake/third_party/openfst.cmake
+++ b/cmake/third_party/openfst.cmake
@@ -54,6 +54,7 @@ if(NOT openfst_POPULATED)
             )
 
             install(DIRECTORY ${openfst_SOURCE_DIR}/src/include/ DESTINATION include/
+            COMPONENT kaldi
                     FILES_MATCHING PATTERN "*.h")
 
             install(TARGETS fst
--- a/cmake/third_party/openfst.cmake
+++ b/cmake/third_party/openfst.cmake
@@ -54,7 +54,7 @@ if(NOT openfst_POPULATED)
             )
 
             install(DIRECTORY ${openfst_SOURCE_DIR}/src/include/ DESTINATION include/
-            COMPONENT kaldi
+                    COMPONENT kaldi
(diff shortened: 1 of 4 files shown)


mmcauliffe:
There were a couple more changes necessary to get Kaldi compiling correctly with Cuda on Windows (VS2019).  There is a code addition for src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc to add a function on windows to get it to compile, but everything else is in the CMake files.

cmake/third_party/openfst.cmake
```diff
@@ -54,6 +54,7 @@ if(NOT openfst_POPULATED)
             )
 
             install(DIRECTORY ${openfst_SOURCE_DIR}/src/include/ DESTINATION include/
+            COMPONENT kaldi
```
h-vetinari, line 57: The indent is off here...

src/cudadecoder/batched-threaded-nnet3-cuda-online-pipeline.cc
```diff
@@ -32,6 +32,23 @@
 #include "lat/lattice-functions.h"
 #include "nnet3/nnet-utils.h"
 
+// Necessary for supporting online2 libary on Windows
+#ifdef _WIN32
+  #include <windows.h>
+
+  void usleep(__int64 usec)
```
jtrmal, line 39: I would prefer this to be replaced by
```
std::this_thread::sleep_for(std::chrono::microseconds(usec));
```

jtrmal: that's standard C++(11, IIRC) and OS independent.

jtrmal:
@mmcauliffe would you mind making the change in the windows usleep implementation? Or do you have any other opinion?

danpovey:
ping

jtrmal:
replaced by 4777

PR #4733, opened 2022-04-25
FreeBSD build support

tpoindex:
Minimal patches to allow building with FreeBSD.    I've used these patches to build[ Vosk-API ](https://github.com/alphacep/vosk-api)for FreeBSD


danpovey:
Cool!  Will wait for one of the others to merge as  they look at the build system more carefully than me.

jtrmal:
would be great to test out the cmake also, @tpoindex, if you have bw

PR #4735, opened 2022-04-28
Update run_aug_common.sh

bofenghuang:
Hi, I think there is a missed argument `generate_alignments=true` for 3th stage

```
if [ $stage -le 3 ] && $generate_alignments; then
...
```

refer to https://github.com/kaldi-asr/kaldi/blob/master/egs/swbd/s5c/local/nnet3/run_ivector_common.sh#L7


jtrmal:
thanks! merging

PR #4736, opened 2022-04-29
Fix 'make valgrind'

hungry-foolish:
Fix the 'make valgrind' command so that it will work instead of giving something like:
Makefile:22: *** insufficient number of arguments (1) to function 'filter-out'.  Stop.


jtrmal:
This looks valid, let me try to close and reopen the PR just to restart travis

jtrmal:
@hungry-foolish  thanks!

PR #4737, opened 2022-05-03
minor fix in calculating per catetory stats

huangruizhe:
Hi Yenda, here is the fix. @jtrmal 

To explain more, without this fix, we could not distinguish `ref_time=0` or `ref_time=`. Thus the counting in the following `corr_sweep` and `fa_sweep` may be wrong.


jtrmal:
looks good, thanks!

PR #4738, opened 2022-05-06
[scripts] chain2 training: reset the error flag next time the script runs

jtrmal:
this is a minor change :)


jtrmal:
No objections from anyone, merging :)

PR #4739, opened 2022-05-08
AMI hotfix

athornton1618:
There's a couple issues with egs/ami (broken URLs, missing call to generate x-vector, etc). As part of a class at Columbia University, I worked out of that directory and cleaned up some of those dependencies. The corrections should become part of egs/ami for posterity.


danpovey (commented):
Thanks!
Is the dir egs/ami/s5c_apt2141 supposed to be added?  I would have thought you'd submit just your fixes?  There are things like MANIFEST.txt that don't look like they should be added, too..

athornton1618:
Oops! Yup, I meant to only push a subset of the changes here. (It was packaged in this state for the course I did this project for). Let's delete this PR and I'll make a new one later this week with just the relevant fixes.

PR #4744, opened 2022-05-17
[build] enable compilers gcc 10 and 11 when cuda>=11.4

jtrmal:
@kkm000 if you care...

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -43,27 +43,40 @@ KALDI_SONAME ?= libkaldi.so
 .DEFAULT_GOAL :=
 .PHONY: all mklibdir rmlibdir checkversion
 all: $(SUBDIRS) matrix/test
-	-echo Done
+	@echo Done
 
-mklibdir:
-	test -d $(KALDILIBDIR) || mkdir $(KALDILIBDIR)
+ifneq ($(KALDILIBDIR), )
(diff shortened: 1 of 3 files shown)


jtrmal:
I took your changes, fixed up two little detail and resolved conflicts. Wanna compile on linux also (probably during the evening) before I commit. FOr now, tested out on macos

jtrmal:
seems like now even Travis is happy. I'll merge

kkm000:
Sure, LGTM. I was looking at the travis script, which always installs portaudio even for the job that won't use it, but this part needs a bigger rework anyway.

kkm000:
If Travis is happy, and you tested on a Mac, I'm happy, too! Thanks, really! I'm sorry for dragging on with my Kaldi stuff.

I am fighting with my computer problems, on top of other things: three dead notebooks this year. Two of the late were brand new, one DOA, another in 6 weeks, and I just got a call from HP that they'll ship a new one not in 1 week, as said the standard "we received your untimely passed one" notification, and not in mid-July, as another SMS next day said, but in 3-4, but they're "working" on ASAP. My desktop is past due for an upgrade by 5 years or so, I guess... But if all else fails, I have a Raspberry Pi! :-)

jtrmal:
I thank _you_. I didn't do much. You did the work.
y.

PR #4761, opened 2022-06-22
Fix Windows build - Add cctype import to kaldi-utils.cc, remove mkl_intel_c.lib

Zohan:
Hello again!
I was able to get a successful build in VS2019, and online-gmm-decode-faster.exe running in Powershell successfully. Not too much to report, mkl_intel_c.lib is no longer a thing.
Let me know if I need to do an IFDEF and check for Win32 on the cctype import.
Thanks!

### DIFF
--- a/src/base/kaldi-utils.cc
+++ b/src/base/kaldi-utils.cc
@@ -22,7 +22,6 @@
 #include <chrono>
 #include <cstdio>
 #include <thread>
-#include <cctype>
 
 
 namespace kaldi {
--- a/src/base/kaldi-utils.cc
+++ b/src/base/kaldi-utils.cc
@@ -21,6 +21,7 @@
 #include <chrono>
 #include <cstdio>
 #include <thread>
+#include <cctype>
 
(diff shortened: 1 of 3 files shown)


masa-mitsu:
Hi @Zohan,
I have made the same modification as you to `kaldi-utils.cc` in #4757, so there are multiple `cctype` includes.
It would be better to remove one of them.

jtrmal:
Hi @Zohan, thanks for your contributuion, could you verify that your changes w.r.t. cctype.h are still needed even for the latest version of Kaldi? 
I would merge afterward.

Zohan:
> Hi @Zohan, thanks for your contributuion, could you verify that your changes w.r.t. cctype.h are still needed even for the latest version of Kaldi? I would merge afterward.

Have done so and confirmed there was a double import on <cctype>. I've removed my `kaldi-utils.cc` changes.

jtrmal:
Perfect, thank you very much!

PR #4763, opened 2022-07-14
chime6/s5c_track2: example code for pretrained model

boeddeker:
In chime6/s5c_track2/run.sh are comments, that the training can be skipped, with pretrained models.
To use the pretained models, there are code suggestions, but not all work.
They need human interaction.

This PR is a suggestion, to add commented code, that can be used inside the script,
so that the user doesn't have to kill the script, once it reaches the training code.

Feel free, to modify my suggestion, if you think it adds value to the run.sh script.
(Takes probably less of your time, when you directly change it)


jtrmal:
looks reasonable and it's comments that could help less experienced people, so I will merge

PR #4764, opened 2022-07-15
chime6: missing --cmd argument

boeddeker:
In `egs/chime6/s5c_track2/local/decode_ts-vad.sh` the script `local/segmentation/detect_speech_activity.sh` is called
without the --cmd arguments, so it uses by always the default queue.pl and not what the user specified.


jtrmal:
thanks, merging

PR #4766, opened 2022-07-19
egs/chime6/s5c_track2/local/ts-vad/diarize_TS-VAD_itX.sh: use python3

boeddeker:
I got the error
```
  File "local/ts-vad/convert_prob_to_rttm.py", line 32, in <module>
    import regex as re
ImportError: No module named regex
```
for the CHiME-6 recipe and I had issues to figure out, why I get this error, because the
first line of that script says it runs with python3
(I installed in every python3 env regex, but it didn't helped).

So I printed the python version, and it said, it's python2.
Then I tried to install regex in python2, but there I got errors like:

```
$ python2 -m pip install regex
...
regex_3/_regex.c:755:15: error: ‘Py_UCS1’ undeclared (first use in this function); did you mean ‘Py_UCS4’?
...
regex_3/_regex.c:26218:5: error: ‘PyModuleDef_HEAD_INIT’ undeclared here (not in a function)
...
```
I think it is easier to change to python3, especially since eol was 2020 for python2.

This PR changes all python calls in diarize_TS-VAD_itX.sh to use python3, as indicated in the first line of the script.

### DIFF
--- a/egs/chime6/s5c_track2/local/ts-vad/diarize_TS-VAD_it1.sh
+++ b/egs/chime6/s5c_track2/local/ts-vad/diarize_TS-VAD_it1.sh
@@ -140,7 +140,7 @@ scoring=$out/scoring
 hyp_rttm=$scoring/rttm
 if [ ! -f $scoring/.done ]; then
   if [ ! -f $hyp_rttm ]; then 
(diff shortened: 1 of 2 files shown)


desh2608:
I think it would work if you just run the script inside a conda python3 environment (`conda create -n my_env python=3.8`), since in that case the `python` command defaults to `python3`.

jtrmal:
I think mostly a history. Plus I think there is no PEP guaranteeing python2
will exist (I might be wrong, tho)
y.

jtrmal:
yeah, it will be great if you do that... python3 is in PEP now, I believe,
so should be just fine.
y.

PR #4774, opened 2022-08-12
remove obsoleted and C++17 incompatible unary_function<>, resolves #4732

### LINKED ISSUE #4732: Remove use of `std::unary_function`
This [facility](https://en.cppreference.com/w/cpp/utility/functional/unary_function) has been deprecated in C++11 and was removed in C++17. In particular, this means the code as-is currently doesn't compile with C++17, which can be problematic because that's often a necessary work-around for CUDA on windows, cf. e.g. https://github.com/microsoft/STL/issues/2020.

jtrmal:
@danpovey @h-vetinari  I took a stab at it -- seems like it's working for C++14, testing C++17 ATM

### DIFF
--- a/src/chainbin/nnet3-chain-shuffle-egs.cc
+++ b/src/chainbin/nnet3-chain-shuffle-egs.cc
@@ -22,6 +22,7 @@
 #include "util/common-utils.h"
 #include "hmm/transition-model.h"
 #include "nnet3/nnet-chain-example.h"
+#include <random>
 
 int main(int argc, char *argv[]) {
   try {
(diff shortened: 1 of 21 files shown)


jtrmal:
turns out `std::random_shuffle()` was also removed and needed to be replaced by `std:shuffle(begin, end, RNG)` which I did in the second commit. @danpovey  could you please check if this has any implication on the functionality? Things will be more random now... Tests ran fine both in C++14 and C++17

h-vetinari (approved):
Don't know the kaldi codebase well at all, but LGTM. Thanks!

jtrmal:
merging

PR #4779, opened 2022-08-17
Fix typo in gen_cmake_skeleton

mmcauliffe:
@jtrmal There was a typo in gen_cmake_skeletons.py, but with it fixed, all platforms are building: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=552548&view=results


jtrmal:
thank you and sorry for the error

PR #4780, opened 2022-08-17
Revert "remove obsoleted and C++17 incompatible unary_function<>, resolves #4732"

### LINKED ISSUE #4732: Remove use of `std::unary_function`

jtrmal:
Reverts kaldi-asr/kaldi#4774


jtrmal:
sorry, I messed up

PR #4781, opened 2022-08-18
automatically compile docker images and push to docker hub

jtrmal:
@mdoulaty, please have a look --  this has run for pushing into my personal space on docker hub.
Once you ok this, I will  play with it more to get it run monthly or something like that (and remove build on every PR and every merge, which sounds wasteful)

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -41,7 +41,7 @@ jobs:
     - name: Build and push
       run: |
           cd docker/ubuntu18.04-cuda10.0/
-          docker build --push --tag jtrmal/kaldi:gpu-latest --tag jtrmal/kaldi:gpu-ubuntu18.04-cuda10.0 .
+          docker build --push --tag kaldiasr/kaldi:gpu-latest --tag kaldiasr/kaldi:gpu-ubuntu18.04-cuda10.0 .
           
           
       
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -36,14 +36,15 @@ jobs:
           echo "Flags:     ${{ steps.buildx.outputs.flags }}"
           echo "Platforms: ${{ steps.buildx.outputs.platforms }}"
           
-#    - name: Build and push
-#      uses: docker/build-push-action@v3
-#      with:
-#        #context: "${{defaultContext}}:docker/ubuntu18.04-cuda10.0"
-#        #file: "{{defaultContext}}:docker/ubuntu18.04-cuda10.0/Dockerfile"
-#        push: true
-#        #tags: jtrmal/test:$(date +%s),jtrmal/gpu:latest
-          
+    - name: Build and push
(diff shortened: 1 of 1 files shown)


jtrmal:
there was an issue with secrets sharing within pull requeststs opened from a different fork so instead I just went ahead and made the changes

csukuangfj:
> there was an issue with secrets sharing within pull requeststs opened from a different fork so instead I just went ahead and made the changes

please have a look at https://github.com/k2-fsa/k2/blob/79d9c12b1571f02e6440b955eacf6bd15555cd3c/.github/workflows/nightly-cuda-ubuntu.yml#L21

```
    if: github.repository == 'k2-fsa/k2'
```

You can enable this GitHub action only if is invoked  from `kaldi-asr/kaldi`.

.github/workflows/docker-images.yml
```diff
@@ -0,0 +1,51 @@
...
+          echo "Name:      ${{ steps.buildx.outputs.name }}"
+          echo "Endpoint:  ${{ steps.buildx.outputs.endpoint }}"
+          echo "Status:    ${{ steps.buildx.outputs.status }}"
+          echo "Flags:     ${{ steps.buildx.outputs.flags }}"
+          echo "Platforms: ${{ steps.buildx.outputs.platforms }}"
+
+    - name: Build and push
+      run: |
+          cd docker/ubuntu18.04-cuda10.0/
+          docker build --push --tag kaldiasr/kaldi:gpu-latest --tag kaldiasr/kaldi:gpu-ubuntu18.04-cuda10.0 .
```
csukuangfj, line 51: Shall we put a date on the tag so that users can report bugs/issues on a specific tag?

jtrmal: sounds reasonable, I'll look into it
Do you suggest something like 
```
 --tag kaldiasr/kaldi:gpu-latest --tag kaldiasr/kaldi:gpu-ubuntu18.04-cuda10.0 --tag kaldiasr/kaldi:gpu-ubuntu18.04-cuda10.0-$(date %s)
 ```
?

csukuangfj: Yes, something like that. The main motivation is to let developers distinguish which version users are using when they are reporting issues.

It is not clear what the `latest` version refers to as users can install the `latest` version at different times.

jtrmal:
I'm gonna close this for a while, to stop generating faulty actions....
will have working/testing PR open in jtrmal/kaldi

.github/workflows/docker-images.yml
jtrmal: yeah, makes sense to me

mdoulaty: I think we should also keep a `latest` tag that always points to the latest image that's built

jtrmal:
I do that for the cpu build. I build gpu and CPU images in parallel jobs,
so they will be at docker hub within few minutes of each other
y.

PR #4782, opened 2022-08-18
[infra] docker images automatically using gh

jtrmal:
recreating the docker build PR


jtrmal:
@csukuangfj thanks for great example

jtrmal:
I\ll merge because the c++ runner is not affected by this and the docker build thing is not run (and been tested)

PR #4787, opened 2022-08-25
Move comment to create '$lang/tmp/LG.fst' (line 53) down

taku-takamatsu:
Move comment to create '$lang/tmp/LG.fst' (line 53) to where it's actually being created (line 96).

Ref: https://groups.google.com/g/kaldi-help/c/R051WkAuraY


kkm000:
Thanks!

PR #4789, opened 2022-08-31
Copy utt2lang when using utils/copy_data_dir.sh

jonghwanhyeon:
Currently `utils/copy_data_dir.sh` does not copy `utt2lang` to destination directory. However, `utils/subset_data_dir.sh` does as follows:
https://github.com/kaldi-asr/kaldi/blob/0fb502d10047ab702c664db4d7e175d282762539/egs/wsj/s5/utils/subset_data_dir.sh#L141-L142

Because I encountered this issue when making multi-language and multi-speaker TTS model, I wrote this PR.


kkm000:
LGTM. Nice catch, thank you!

PR #4790, opened 2022-09-07
Update run_blstm.sh

xu-gaopeng:
fix bug aspire run_blstm.sh

### DIFF
--- a/egs/aspire/s5/local/nnet3/run_blstm.sh
+++ b/egs/aspire/s5/local/nnet3/run_blstm.sh
@@ -89,9 +89,9 @@ if [ $stage -le 7 ]; then
   fast-lstmp-layer name=blstm2-backward input=Append(blstm1-forward, blstm1-backward) delay=2 $lstm_opts
 
   fast-lstmp-layer name=blstm3-forward input=Append(blstm2-forward, blstm2-backward) delay=-3 $lstm_opts
-  fast-lstmp-layer name=blstm3-backward input=Append(blstm2-forward, blstm2-backward delay=3 $lstm_opts
+  fast-lstmp-layer name=blstm3-backward input=Append(blstm2-forward, blstm2-backward) delay=3 $lstm_opts
 
-  output-layer name=output output-delay=$label_delay dim=$num_targets max-change=1.5
+  output-layer name=output input=Append(blstm3-forward, blstm3-backward) output-delay=$label_delay dim=$num_targets max-change=1.5
 EOF
   steps/nnet3/xconfig_to_configs.py --xconfig-file $dir/configs/network.xconfig --config-dir $dir/configs || exit 1
 fi


kkm000:
@xu-gaopeng Thanks for the catch, these are obviously typos. What's odd, the config seems not to have an `output-xent` layer. This is off. Have you ran the trainer, at least for a few iterations, just to test it?

egs/aspire/s5/local/nnet3/run_blstm.sh
```diff
@@ -89,9 +89,9 @@ if [ $stage -le 7 ]; then
   fast-lstmp-layer name=blstm2-backward input=Append(blstm1-forward, blstm1-backward) delay=2 $lstm_opts
 
   fast-lstmp-layer name=blstm3-forward input=Append(blstm2-forward, blstm2-backward) delay=-3 $lstm_opts
-  fast-lstmp-layer name=blstm3-backward input=Append(blstm2-forward, blstm2-backward delay=3 $lstm_opts
+  fast-lstmp-layer name=blstm3-backward input=Append(blstm2-forward, blstm2-backward) delay=3 $lstm_opts
 
-  output-layer name=output output-delay=$label_delay dim=$num_targets max-change=1.5
+  output-layer name=output input=Append(blstm3-forward, blstm3-backward) output-delay=$label_delay dim=$num_targets max-change=1.5
 EOF
```
kkm000, line 95: `output-layer name=output-xent ...` is missing. The model is indeed using the chain loss.

kkm000 (changes requested):

egs/aspire/s5/local/nnet3/run_blstm.sh
```diff
@@ -89,9 +89,9 @@ if [ $stage -le 7 ]; then
   fast-lstmp-layer name=blstm2-backward input=Append(blstm1-forward, blstm1-backward) delay=2 $lstm_opts
 
   fast-lstmp-layer name=blstm3-forward input=Append(blstm2-forward, blstm2-backward) delay=-3 $lstm_opts
-  fast-lstmp-layer name=blstm3-backward input=Append(blstm2-forward, blstm2-backward delay=3 $lstm_opts
+  fast-lstmp-layer name=blstm3-backward input=Append(blstm2-forward, blstm2-backward) delay=3 $lstm_opts
 
-  output-layer name=output output-delay=$label_delay dim=$num_targets max-change=1.5
+  output-layer name=output input=Append(blstm3-forward, blstm3-backward) output-delay=$label_delay dim=$num_targets max-change=1.5
```
kkm000, line 94: I'm wondering what does the label delay mean in a bidi network. It's likely intended to be 0 (and, generally, throughout the whole script). What sense does it make?

@jtrmal, do you have experience with the BLSTM training? Also, `include-log-softmax` is `true` by default. PTAL.

```suggestion
  output-layer name=output input=Append(blstm3-forward, blstm3-backward) include-log-softmax=false output-delay=0 dim=$num_targets max-change=1.5
```

kkm000 (dismissed):

egs/aspire/s5/local/nnet3/run_blstm.sh
jtrmal: I'm actually not sure. @danpovey  could you please help?

danpovey: 0 is the default so it's not really important here.

jtrmal:
`input=Append(blstm3-forward, blstm3-backward)` change makes sense, though,
or not?
y.

egs/aspire/s5/local/nnet3/run_blstm.sh
kkm000: @danpovey I'd rip it out. It's only confusing. But since the recipe predates CTC... Maybe whatever.

kkm000:
@jtrmal Yes, this one certainly does. The default is the previous line's output.

egs/aspire/s5/local/nnet3/run_blstm.sh
kkm000: @danpovey And `include-log-softmax=false` is needed here, right?

danpovey: we normally use include-log-softmax=false for the output layer for MPE, it saves a little time although it doesn't affect the result.

danpovey: .. also: output-delay should be 0 for blstm, yes.

jtrmal:
Thanks, merging. Sorry for the delay!
y.

PR #4809, opened 2022-12-06
Do not use ADL to invoke std::binary_search

georgthegreat:


jtrmal:
you would need to include <algorithm> explicitly, I think

georgthegreat:
@jtrmal, do you have any objections against merging this PR?

jtrmal:
No objection. Will try to get to it today.
Sorry for the delay.
Y.

PR #4820, opened 2023-02-03
egs/ami: Fix BUT path to wavs in AMI scripts, add beamformer config

vesis84:
- the audio data no longer exist in that path
- the beamformer config was missing in 'ami/s5b', it's taken from 'ami/s5'

(The path to audio files is specific to Brno University of Technology cluster, I kindly ask for a possibility to update the paths.)
(The beamformer config is not specific to Brno, the problem was there for everyone.)

### DIFF
--- a/egs/ami/s5/run_ihm.sh
+++ b/egs/ami/s5/run_ihm.sh
@@ -17,7 +17,7 @@ set -euxo pipefail
 # Path where AMI gets downloaded (or where locally available):
 AMI_DIR=$PWD/wav_db # Default,
 case $(hostname -d) in
-  fit.vutbr.cz) AMI_DIR=/mnt/matylda5/iveselyk/KALDI_AMI_WAV ;; # BUT,
+  fit.vutbr.cz) AMI_DIR=/mnt/matylda2/data/AMI_KALDI_DOWNLOAD ;; # BUT,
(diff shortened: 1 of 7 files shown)


jtrmal:
LGTM

kkm000 (approved):

kkm000:
Thanks guys!

kkm000:
While I have both of your guys attention: among other super-useful things, like _defenestratio_ as a way to round up a boring meeting that dragged on and on and on with no end in sight (I've attended too many of these, and at times entertained myself by imagining how I would end it the Prague way!), the notion of the set was also invented in Prague: https://mathshistory.st-andrews.ac.uk/Biographies/Bolzano/. I've never heard of that!

vesis84:
It is interesting idea to use _defenestratio_ in daily life. Actually in one workshop in Prague in 2014, we were jumping out of the window. That was in the ground floor, but we enjoyed it a lot... :-) 

In the history, there were also other "science personas" linked to Czechia, like astronomer Johannes Kepler https://en.wikipedia.org/wiki/Johannes_Kepler (Prague) and the father of genetics Johann Gregor Mendel https://en.wikipedia.org/wiki/Gregor_Mendel (Brno).

And from non-Czechia, I like the story of hiding the author's identity for the Student t-distribution created at the Guinness brewery in Dublin:
https://en.wikipedia.org/wiki/Student%27s_t-distribution#History_and_etymology

PR #4821, opened 2023-02-04
WSL PATH fixes

QThans:
for solving the "bad variable name" error which occurs in WSL.
The error is due to the WSL automatically appends the Windows path to the WSL/Linux path, and the Windows path includes space such as "Program file".
image


jtrmal:
Thanks for the PR but I'm not convinced this will resolve your issues -- most of the scripts were written with no regard for the possibility of spaces in the paths/filenames

QThans:
I tried. It's work normally now. This error is because there are some spaces in Windows path such as 'C:\Program Files (x86)', this PR just for solving this issue, not for other possibility of spaces.

PR #4823, opened 2023-02-12
Fix matrix data offset for large matrices

nshmyrev:
Funny enough it wasn't noticed before.

If we create a big matrix say 2 000 000 x 1280 (10Gb in size), RowData doesn't work properly and causes a segfault. It just overflows int32 while calculating offset.


nshmyrev:
Hold on merging this, cudamatrix also needs fix

jtrmal:
LGTM

jtrmal:
Merging, thanks!

nshmyrev:
Thank you!

PR #4828, opened 2023-02-23
More fixes of unwanted ADL usage of std algos

georgthegreat:
This continues the work started in #4809.


jtrmal:
I think this is fine. Thanks!

